PR #141, opened 2026-07-21
test(odbc-e2e): skip INFO-surfacing tests on the msodbcsql comparison leg

David-Engel:
## Description

`./run_e2e.sh --compare-with-msodbcsql` reruns the ODBC e2e suite against msodbcsql 18 and currently fails two INFO-token tests (added in #113) that assert **mssql-odbc-specific** INFO surfacing which msodbcsql does not replicate:

- `ExecDirectLiveTest.InfoMessagesSurfaceAsDiagnostics` — a no-result-set `PRINT` / low-severity `RAISERROR` batch returns `SQL_SUCCESS_WITH_INFO` on `SQLExecDirect`; msodbcsql defers these differently.
- `MoreResultsLiveTest.TrailingInfoBetweenResultSetsSurfacesWithHint` — an INFO message between two result sets surfaces on the `SQLMoreResults` advance as `SQL_SUCCESS_WITH_INFO`; msodbcsql surfaces between-result INFO differently (its leg hits SQLSTATE `24000` invalid cursor state).

Both are intended divergences, so they are gated with `SKIP_IF_COMPARING_MSODBCSQL()`.

To avoid conflicts, the `SKIP_IF_COMPARING_MSODBCSQL()` macro and its `run_e2e.sh` `ODBC_TEST_TARGET` wiring are taken **byte-for-byte from #119** (`dev/tkotian/SQLExecAndParams`), which introduces the same helper. Whichever PR merges first, the identical additions merge cleanly.

## Validation

Against a live SQL Server 2022 container:
- `./run_e2e.sh --compare-with-msodbcsql` → parity report shows both tests as `parity`, "Both runs passed".
- Direct binary run confirms each case reports `[ SKIPPED ]` on the msodbcsql leg (`ODBC_TEST_TARGET=msodbcsql`) and runs/passes on the mssql-odbc leg.

## Checklist

- [x] `run_e2e.sh` passes (mssql-odbc only) — 13/13
- [x] `run_e2e.sh --compare-with-msodbcsql` passes — both legs green
- [x] Shared helper matches #119 byte-for-byte

PR #145, opened 2026-07-22
Add ActiveDirectoryInteractive auth to mssql-odbc via mssql-auth (T3)

Vahid-b:
## Description

Implements **T3**: `Authentication=ActiveDirectoryInteractive` for mssql-odbc, matching the classic C++ `msodbcsql` driver.

> ### ⚠️ This PR was redesigned after review — please re-read
> The earlier revision hand-rolled an OAuth2 **authorization-code + PKCE** flow over a loopback listener and the system browser. That design has been **removed**. Interactive sign-in is now delegated to **`mssql-auth.dll` (OneAuth/MSQA)** — the same library msodbcsql loads.
>
> `auth/interactive.rs` is rewritten, `auth/msqa.rs` is new, and the loopback/PKCE/browser-launch code, its dependencies (`base64`, `getrandom`, `sha2`, `serde`, `serde_json`) and its tests are gone. Most of the earlier review discussion (PKCE, `state` validation, redirect classification, browser launching, zombie reaping) no longer applies to any code in this PR.

### Why the design changed

mssql-odbc presents **msodbcsql's own client id**. The two drivers therefore have to be indistinguishable to callers and to the service — a keyword that behaved differently between them would be a behavioural fork, not parity. Delegating to the same auth library yields identical prompts, identical caching semantics, and identical error text, and avoids this repo owning a hand-rolled OAuth implementation.

## Platform behaviour

| Platform | Behaviour |
|---|---|
| Windows | Interactive sign-in via `mssql-auth.dll` |
| Linux / macOS | Resolves to `ActiveDirectoryIntegrated` (msodbcsql's own fall-through) |

**msodbcsql's interactive support is Windows-only.** From the C++ source:

- `SNI_FedAuth.hpp:16` wraps the entire header in `#if !defined(MPLAT_BUILD) //No MSQA for non-windows clients`
- `sni/src/Makefile` omits `SNI_FedAuth` from `SRCLIST` — it is never compiled on Unix
- `SNI_FedAuth.cpp:249` loads the DLL via `LoadLibraryExA`; the repo contains no `dlopen` counterpart

On Unix the MSQA branch **and its `else` arm** both sit inside `#if !defined(XPLAT_ODBC_TODO)` (`Parse.cpp:3597-3660`), so the guard removes both and control falls through to `AzureADAuth` with `AKVCFG_AUTHMODE_INTEGRATED`. Nothing rejects the keyword earlier (`sqlcconn.cpp:4066-4078` accepts it on every platform).

An explicit *"interactive is not supported on this platform"* error was prototyped and **deliberately withdrawn**: because we present msodbcsql's client id, refusing a keyword msodbcsql accepts would be the one place the drivers visibly disagree. Improving on the fall-through is tracked separately rather than shipped as a silent divergence.

## How it works

`InteractiveTokenFactory` implements the existing mssql-tds `EntraIdTokenFactory` seam, so it plugs into the FedAuth handshake exactly like the T2 service-principal factory. `auth/msqa.rs` holds the `mssql-auth.dll` bindings and mirrors `SNI_FedAuth.cpp`:

- Authentication contexts are cached on **`username + "\0" + stsUrl`** (`SNI_FedAuth.cpp:127`), created with `ForcePrompt=TRUE`, and `ForcePrompt` is cleared after the first success — so a second connection reuses the context and does not re-prompt.
- `MSQAGetRequestStatus` decides whether interaction is needed; when it is, the window is hosted on a dedicated STA thread (`CoInitializeEx(COINIT_APARTMENTTHREADED)` + `MSQAUICreateHostWindow` + message pump), and the completion callback ends the pump via `PostThreadMessage(WM_QUIT)`.
- `MSQAReleaseAuthenticationContext` is **intentionally not called** for interactive (`SNI_FedAuth.cpp:777-782`) — the context *is* the token cache.
- OneAuth packs its `Status` into the high byte of the returned error. Only `NetworkTemporarilyUnavailable`, `ServerTemporarilyUnavailable` and `TransientError` are retryable, mirroring `IsTransientError` (`:301-306`). **`UserCanceled` is not transient**, so a cancelled sign-in fails once instead of reopening the window.

Parity parameters, all from `Parse.cpp:3601-3620`:

| Parameter | Value |
|---|---|
| Client id | `2c1229aa-16c5-4ff5-b46b-4f7fe2a2a9c8` (msodbcsql's ODBC client id) |
| Redirect URI | `https://sqlaad/` |
| WAM | off by default (`ADALuseWAM` defaults to 0, `sqlcconn.cpp:3417`) |
| Window title | `"Authenticate to database on %s"` (`IDS_AD_AUTH_DB`) |

## Login timeout

Also lands `SQL_ATTR_LOGIN_TIMEOUT` and a real `SQLGetConnectAttrW`. mssql-tds previously had a single `connect_timeout` bounding both the overall login deadline *and* each TCP connect, so allowing minutes for an MFA prompt also let an unreachable host hang that long. These are now split, matching msodbcsql:

- **`connect_timeout`** (default 15s) — per-TCP-connect cap; unreachable servers still fail fast.
- **`login_timeout: Option<u32>`** (new) — overall login deadline. `None` falls back to `connect_timeout`, so existing callers are unaffected; `Some(0)` waits indefinitely.

Interactive installs a 330s default **only when the app has not set its own** `SQL_ATTR_LOGIN_TIMEOUT`. If the deadline expires while a window is open, `CancelUiOnDrop` posts `WM_QUIT` so the window is torn down rather than orphaned.

## Tests

`cargo test -p mssql-odbc --lib` → **419 pass**. Unit tests cover context-cache keying, transient/permanent status classification, error-description reads, `configure_auth` dispatch on both platforms, and the login-timeout attribute (including `0` → infinite and preserving an app-set value).

Because Linux/macOS cannot be cross-compiled here (`openssl-sys`), the non-Windows path was validated by **inverting every `cfg(windows)`/`cfg_attr` gate** and building: clippy clean under `-D warnings`, 333 tests pass. That caught a real break — an unused-variable error that would only have appeared in Linux CI.

### Live verification

The sign-in window and token read cannot run in CI, so this was verified manually against a live Azure SQL server and Entra tenant:

| Check | Result |
|---|---|
| Interactive sign-in completes | `SQL_SUCCESS_WITH_INFO`; `SUSER_NAME()` returns the Entra account |
| Second connection does not re-prompt | **0.1s, no window** (vs 19.4s with the prompt) — context cache confirmed |
| Cancelling does not relaunch the prompt | exactly one window; `SQLSTATE 08001`, one failure |
| `SQL_ATTR_LOGIN_TIMEOUT` round-trips | set 300 → read back 300 |

## Known limitations

- **`mssql-auth.dll` is not redistributed by this driver.** Interactive works where msodbcsql 18 has already installed it into `System32`. Packaging is unresolved and needs a decision before GA.
- **A slow first sign-in can lose the connection.** TDS FedAuth requires the server's `FEDAUTHINFO` (carrying the STS URL) *before* a token can be acquired, so the connection idles for the whole sign-in; a multi-minute first-time MFA can exceed the Azure SQL gateway's idle timeout. The token is acquired successfully — the failure surfaces from the TDS login parser. msodbcsql prompts from the same point in the login exchange and shares this exposure, so it is treated as parity, not a defect here.
- **Token caching is per-process**, matching msodbcsql — nothing is persisted to disk. Cross-connection caching is tracked in [AB#46409](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46409).

## Related Issues

[AB#46067](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46067)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes — `cargo test -p mssql-odbc --lib` (419 pass) locally; `nextest`/`llvm-cov` unavailable on this machine, so full coverage runs in CI


mssql-odbc/src/auth/entra.rs
```diff
@@ -243,6 +244,20 @@ pub(crate) fn configure_auth(
...
+            let login_hint = (!resolved.user_name.is_empty()).then_some(resolved.user_name);
+            let factory = InteractiveTokenFactory::new(login_hint);
+            context.auth_method_map.insert(
+                TdsAuthenticationMethod::ActiveDirectoryInteractive,
+                Box::new(factory),
+            );
+            // The browser sign-in (with MFA) can take minutes, so disable the
+            // login-connect deadline; the flow keeps its own inner bound
+            // (REDIRECT_TIMEOUT) and stays cancellable. See AB#46067.
+            context.connect_timeout = 0;
```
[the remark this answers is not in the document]
Vahid-b: @-

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,655 @@
...
+    async fn create_token(
+        &self,
+        spn: String,
+        sts_url: String,
+        _auth_method: TdsAuthenticationMethod,
+    ) -> TdsResult<Vec<u8>> {
+        let login_hint = self.login_hint.clone();
+        let token = self
+            .token
+            .get_or_try_init(|| async move {
```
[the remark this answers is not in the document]
Vahid-b: @-

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,655 @@
...
+        login_hint,
+    })?;
+
+    info!("launching browser for interactive Entra sign-in");
+    debug!(%authorize_url, "interactive authorize URL");
+    open_browser(authorize_url.as_str()).map_err(|e| {
+        Error::ConnectionError(format!(
+            "failed to launch a browser for interactive sign-in: {e}. Open this URL manually to continue: {authorize_url}"
+        ))
+    })?;
```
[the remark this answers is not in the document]
Vahid-b: @-

mssql-odbc/src/auth/entra.rs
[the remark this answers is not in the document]
Vahid-b: @-

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,655 @@
...
+        .spawn()
+        .map(|_| ())
+}
+
+#[cfg(target_os = "macos")]
+fn open_browser(url: &str) -> std::io::Result<()> {
+    std::process::Command::new("open")
+        .arg(url)
+        .spawn()
+        .map(|_| ())
```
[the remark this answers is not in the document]
Vahid-b: @-

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,655 @@
...
+}
+
+#[cfg(all(unix, not(target_os = "macos")))]
+fn open_browser(url: &str) -> std::io::Result<()> {
+    std::process::Command::new("xdg-open")
+        .arg(url)
+        .spawn()
+        .map(|_| ())
```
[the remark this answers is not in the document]
Vahid-b: @-

Vahid-b:
Pushed `816589d0`: implemented the separate **`SQL_ATTR_LOGIN_TIMEOUT`** login timeout (full msodbcsql parity), replacing the earlier "enlarge `connect_timeout` to 330s" workaround.

**mssql-tds core (additive, back-compatible):**
- `ClientContext` gains `login_timeout: Option<u32>`. The outer login deadline now uses `login_timeout.unwrap_or(connect_timeout)`, so callers that only set `connect_timeout` are byte-for-byte unchanged. `connect_timeout` remains the per-TCP-connect cap.

**mssql-odbc:**
- `SQL_ATTR_LOGIN_TIMEOUT` is now honored (stored on the DBC, applied to `login_timeout` at connect; `0` → wait indefinitely) instead of being a silent no-op.
- Interactive installs the 330s login-timeout default **only when the app hasn't set `SQL_ATTR_LOGIN_TIMEOUT`**, and leaves `connect_timeout` at its default — so an unreachable server fails fast on the per-connect cap while the browser/MFA round-trip still fits.

**Net effect:** the black-hole-server concern from the earlier review is resolved properly rather than papered over.

Validation: `cargo bfmt` + workspace `cargo bclippy -D warnings` clean; `cargo test -p mssql-odbc --lib` → 346 pass; changed mssql-tds modules (`client_context`, `connection_provider`) pass. Full nextest/coverage in CI.

mssql-tds/src/connection_provider/tds_connection_provider.rs
```diff
@@ -269,8 +269,16 @@ impl TdsConnectionProvider {
...
+            // set, else falls back to `connect_timeout` so callers that only set
+            // the historical single knob keep the same behavior. `0` (either
+            // field) means "no deadline". `connect_timeout` still bounds each
+            // individual TCP-connect attempt, so an unreachable host fails fast
+            // even when the login deadline is large (e.g. interactive sign-in).
+            let login_timeout = context.login_timeout.unwrap_or(context.connect_timeout);
+            let deadline = match login_timeout {
+                1.. => Some(Instant::now() + Duration::from_secs(login_timeout.into())),
                 _ => None,
             };
```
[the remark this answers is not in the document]
Vahid-b: Fixed in 7b55b3c5. The login deadline is now computed at the top of `execute_action_chain`, before the shared-memory shortcut and SSRP/LocalDB resolution, so it bounds those phases too. The SM attempt uses the remaining login budget (`deadline - now`) instead of `connect_timeout`, so interactive sign-in against a local named instance is no longer cancelled after the default 15s. `connect_timeout` still bounds each individual TCP-connect attempt, and the change is back-compatible for callers that only set `connect_timeout` (`login_timeout` falls back to it).

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,657 @@
...
+        client_id,
+        scope,
+        redirect_uri: &redirect_uri,
+        state: &state,
+        challenge: &pkce.challenge,
+        login_hint,
+    })?;
+
+    info!("launching browser for interactive Entra sign-in");
+    debug!(%authorize_url, "interactive authorize URL");
```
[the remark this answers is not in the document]
Vahid-b: Fixed in 7b55b3c5. The debug log now emits only `scheme`/`host`/`path`; the query string (CSRF `state`, PKCE `challenge`, and `login_hint`) is no longer logged.

One note: the additional locations flagged (lines 312/420/431) don't actually log anything — they're the request read-timeout, the reqwest client builder, and a `response.status()` read respectively. Line 235 was the only statement logging the URL, so that's the only change needed.

mssql-odbc/src/api/set_connect_attr.rs
```diff
@@ -116,13 +117,20 @@ unsafe fn sql_set_connect_attr_w_impl(
...
             }
         }
+        SQL_ATTR_LOGIN_TIMEOUT => {
+            // Integer attribute: the SQLUINTEGER value is passed by value in the
+            // pointer slot (not a pointer to it). Store it so SQLDriverConnect
+            // can apply it to the TDS login deadline. `0` means "wait
+            // indefinitely" (mapped to no deadline at connect time).
+            let secs = value_ptr as usize as u32;
+            state.login_timeout = Some(secs);
+            debug!(secs, "SQLSetConnectAttrW: login timeout stored");
```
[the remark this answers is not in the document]
Vahid-b: Fixed in 7b55b3c5. Implemented `SQLGetConnectAttrW` (new `get_connect_attr.rs`) backed by `DbcState.login_timeout`: for `SQL_ATTR_LOGIN_TIMEOUT` it writes the stored value into the caller's `SQLUINTEGER` buffer, returning the driver default of 15 when unset and `0` when the app set "wait indefinitely". Added set/get round-trip, default-when-unset, and null-pointer rejection tests.

mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -0,0 +1,167 @@
...
+            string_length_ptr,
+        )
+    })
+}
+
+unsafe fn sql_get_connect_attr_w_impl(
+    connection_handle: SqlHandle,
+    attribute: SqlInteger,
+    value_ptr: SqlPointer,
+    _buffer_length: SqlInteger,
```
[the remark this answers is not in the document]
Vahid-b: Refactored in `2382c55c`. `get_connect_attr.rs` now follows the same panic-boundary → unsafe-shim → safe-core split as `get_data.rs`:

- `sql_get_connect_attr_w` — public `unsafe fn`, all-argument `debug!` + `ffi_entry!` panic boundary.
- `sql_get_connect_attr_w_impl` — `unsafe fn`, null-handle → `SQL_INVALID_HANDLE`, `handle_from_raw::<DbcHandle>`, `debug_assert_eq!` on the object type.
- `sql_get_connect_attr_w_safe(&DbcHandle, …)` — safe core: lock, `free_errors`, attribute match.

Good catch — the file had mirrored the older `set_connect_attr.rs`, which predates this convention.

mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -0,0 +1,167 @@
...
+                    0,
+                    "SQL_ATTR_LOGIN_TIMEOUT value pointer is null",
+                );
+                return SQL_ERROR;
+            }
+            // SQLUINTEGER attribute: write the current value into the caller's
+            // buffer. `Some(0)` reflects an app-set "wait indefinitely"; an
+            // unset attribute reports the driver default.
+            let secs = state.login_timeout.unwrap_or(DEFAULT_LOGIN_TIMEOUT_SECS);
+            unsafe { (value_ptr as *mut u32).write_unaligned(secs) };
```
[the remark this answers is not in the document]
Vahid-b: Done in `2382c55c`. The value write now goes through `write_if_some(value_ptr as *mut u32, secs)` (util.rs) instead of a raw `write_unaligned`. The explicit null-pointer check stays ahead of it (returns `SQL_ERROR` with a diagnostic), so `write_if_some` only ever runs on a validated non-null pointer.

mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -0,0 +1,167 @@
...
+    value_ptr: SqlPointer,
+    buffer_length: SqlInteger,
+    string_length_ptr: *mut SqlInteger,
+) -> SqlReturn {
+    debug!(
+        ?connection_handle,
+        attribute,
+        ?value_ptr,
+        "SQLGetConnectAttrW called",
+    );
```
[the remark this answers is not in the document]
Vahid-b: Fixed in `2382c55c`. The entry `debug!` now logs `buffer_length` and `string_length_ptr` alongside the handle, attribute, and value pointer, matching `get_data.rs`'s all-argument trace.

mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    match attribute {
+        SQL_ATTR_LOGIN_TIMEOUT => {
+            if value_ptr.is_null() {
+                error!("SQLGetConnectAttrW: SQL_ATTR_LOGIN_TIMEOUT value pointer is null");
+                post_sql_error(
+                    &mut state,
+                    SQLSTATE_HY009,
+                    0,
+                    "SQL_ATTR_LOGIN_TIMEOUT value pointer is null",
```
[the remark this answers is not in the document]
Vahid-b: Fixed in `2382c55c`. The null-`value_ptr` case now calls `post_diag(&mut state, ERR_INVALID_NULL_POINTER)` (sqlstate.rs) instead of an inline `post_sql_error(SQLSTATE_HY009, …)` literal, so the SQLSTATE and message come from the shared diagnostic constant.

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,666 @@
...
+        Error::ConnectionError(format!(
+            "failed to launch a browser for interactive sign-in: {e}"
+        ))
+    })?;
+
+    let code = tokio::time::timeout(REDIRECT_TIMEOUT, wait_for_redirect(&listener, &state))
+        .await
+        .map_err(|_| {
+            Error::ConnectionError("timed out waiting for interactive sign-in".into())
+        })??;
```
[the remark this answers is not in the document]
Vahid-b: Fixed in `2382c55c`. The browser-redirect wait is no longer a fixed 300s.

`InteractiveTokenFactory` now carries the effective `login_timeout` (set on the context by the interactive arm — the app's `SQL_ATTR_LOGIN_TIMEOUT`, or the 330s default when unset), and `redirect_wait_cap` derives the wait from it:

- `Some(n)` → `n` seconds (honors app values above or below the old fixed cap),
- `Some(0)` → no cap (honors an infinite login timeout),
- `None` → the `REDIRECT_TIMEOUT` fallback.

The provider already wraps `create_token` in `timeout(deadline)` where `deadline` is derived from the same `login_timeout`, so the inner (browser-wait) and outer (login-deadline) bounds now agree instead of the inner 300s silently winning. Added 3 unit tests for the cap derivation.

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,666 @@
...
+        .body(body)
+        .send()
+        .await
+        .map_err(|e| Error::ConnectionError(format!("token request failed: {e}")))?;
+
+    let status = response.status();
+    let bytes = response
+        .bytes()
+        .await
+        .map_err(|e| Error::ConnectionError(format!("failed to read token response: {e}")))?;
```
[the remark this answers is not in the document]
Vahid-b: Fixed in `2382c55c`. `read_body_capped` replaces `response.bytes()`: it rejects an advertised `Content-Length` over the cap, then streams the body via `chunk()` enforcing a 1 MiB `MAX_TOKEN_RESPONSE_BYTES` limit (real token responses are a few KB). Since the STS authority is taken from the server's FEDAUTHINFO, this closes the unbounded-read / memory-exhaustion exposure you flagged.

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,666 @@
...
+    reap(std::process::Command::new("xdg-open").arg(url).spawn()?);
+    Ok(())
+}
+
+/// `open`/`xdg-open` exit as soon as they hand the URL to the browser, but a
+/// dropped `Child` is never waited on and would linger as a zombie. Reap it on a
+/// detached thread so the sign-in flow is not blocked.
+#[cfg(unix)]
+fn reap(mut child: std::process::Child) {
+    std::thread::spawn(move || {
```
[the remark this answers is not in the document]
Vahid-b: Fixed in `2382c55c`. `reap` now uses `std::thread::Builder::new().name(…).spawn(…)` and handles the `Err`: on an OS thread-creation failure it logs and skips the wait (the child detaches — a rare zombie) instead of panicking. That keeps the shared library from unwinding across the FFI boundary, per the no-panic rule.

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,768 @@
...
+    let bytes = read_body_capped(response, MAX_TOKEN_RESPONSE_BYTES).await?;
+    let body = String::from_utf8_lossy(&bytes);
+
+    if !status.is_success() {
+        let detail = serde_json::from_str::<TokenErrorResponse>(body.as_ref())
+            .map(|e| format!("{}: {}", e.error, e.error_description))
+            .unwrap_or_else(|_| body.into_owned());
+        return Err(Error::ConnectionError(format!(
+            "interactive token exchange failed ({status}): {detail}"
+        )));
```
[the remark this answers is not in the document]
Vahid-b: Same root cause as the redirect-failure comment. A non-2xx from the `/token` endpoint returned `Error::ConnectionError` (transient), so the connect-retry loop relaunched the whole browser flow — even for permanent RFC 6749 §5.2 failures such as `invalid_grant`.

Fixed in 2a1415b3: the `!status.is_success()` path now returns `Error::Security(SecurityError::AuthenticationDenied(..))` (non-transient), so a token-exchange failure is surfaced to the caller instead of silently re-prompting. I classify the entire non-2xx case as non-transient deliberately: `create_token` is atomic (a fresh browser + consent on every call, no token cache), so a provider-level retry is never a valid recovery for a token-exchange failure — msodbcsql likewise surfaces the error rather than reopening the browser.

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,768 @@
...
+            RedirectOutcome::Failed(detail) => {
+                write_http_response(
+                    &mut stream,
+                    "400 Bad Request",
+                    "Sign-in failed. You can close this window and return to your application.",
+                )
+                .await;
+                return Err(Error::ConnectionError(format!(
+                    "interactive sign-in failed: {detail}"
+                )));
```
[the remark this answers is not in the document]
Vahid-b: Correct, and this was an active bug rather than theoretical. `RedirectOutcome::Failed` (user clicks Cancel / `access_denied`, `consent_required`, …) returned `Error::ConnectionError`, which `is_transient_connect_error` (mssql-tds/src/error/mod.rs:402-407) classifies as transient. The ODBC connect path uses `ClientContext`'s default `connect_retry_count = 1` (ODBC parses `ConnectRetryCount` but doesn't wire it into the context), so the provider loops `0..=1` = two attempts and re-runs `create_token` on the retry — the browser pops a second time before the connection finally fails.

Fixed in 2a1415b3: a `state`-matched authorization error now maps to `Error::Security(SecurityError::AuthenticationDenied(..))`, which is non-transient (it falls through the `_ => false` arm of `is_transient_connect_error`; pinned by the new `authentication_denied_is_not_transient` test). Genuinely transient transport faults around the sign-in stay `ConnectionError`. I reused the existing `SecurityError` variant instead of adding one, so there's no change to the public error surface.

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,768 @@
...
+}
+
+/// Reads an incoming request up to the end of the request line (first CRLF),
+/// bounded by [`READ_TIMEOUT`] and [`MAX_REQUEST_BYTES`], and returns its query
+/// string. Returns `None` on timeout, EOF, error, or a request with no query.
+async fn read_request_query(stream: &mut tokio::net::TcpStream) -> Option<String> {
+    let mut buf = Vec::new();
+    let mut chunk = [0u8; 1024];
+    loop {
+        let n = match tokio::time::timeout(READ_TIMEOUT, stream.read(&mut chunk)).await {
```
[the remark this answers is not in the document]
Vahid-b: Good catch. `READ_TIMEOUT` was applied *per* `read()` call, so a client that dribbles bytes (or just holds the socket open) reset the 5s timer on every chunk and could hold this **sequential** loopback handler for the entire login budget — and indefinitely once the login timeout is infinite (the `Some(0)` path added in the previous round). Because `wait_for_redirect` accepts one connection at a time, that starves the real browser callback.

Fixed in 2a1415b3: `read_request_query` now bounds the whole request-line read with a single `tokio::time::timeout_at(now + READ_TIMEOUT, …)` deadline instead of a per-read `timeout`, so total read time is capped regardless of how the bytes arrive. I split out `read_request_query_within(stream, budget)` and added two unit tests — a complete request line parses, and a one-byte-then-stall client is cut off well within the budget.

mssql-tds/src/connection_provider/tds_connection_provider.rs
```diff
@@ -143,10 +159,12 @@ impl TdsConnectionProvider {
...
             {
                 debug!("Trying Shared Memory before SSRP");
-                let timeout_duration = match context.connect_timeout {
-                    1.. => Some(Duration::from_secs(context.connect_timeout.into())),
-                    _ => None,
-                };
+                // Bound the shared-memory attempt (which wraps the full TDS/auth
+                // handshake) by the remaining login budget rather than
+                // `connect_timeout`, so interactive sign-in against a local named
+                // instance isn't cancelled after the default 15s.
```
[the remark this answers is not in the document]
Vahid-b: Good catch — this is a direct consequence of the round-3 change that made interactive-auth denials non-transient. The shared-memory shortcut runs the full TDS/auth handshake, so a cancelled interactive sign-in against a local named instance now surfaces as `AuthenticationDenied`, and the old `Err` arm fell through to SSRP/TCP for *every* error — relaunching the browser.

Fixed in 95bc1c90: the SM `Err` arm now only falls through when the failure is transient (`err.is_transient_connect_error()` — a transport-level problem such as SM not being available). A non-transient failure (the cancelled/denied login, or any permanent login error) returns immediately. That matches SNI semantics — a definitive login failure would recur on every transport — and stops the second browser.

mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -0,0 +1,175 @@
...
+            // buffer. `Some(0)` reflects an app-set "wait indefinitely"; an
+            // unset attribute reports the driver default.
+            let secs = state.login_timeout.unwrap_or(DEFAULT_LOGIN_TIMEOUT_SECS);
+            unsafe { write_if_some(value_ptr as *mut u32, secs) };
+            debug!(secs, "SQLGetConnectAttrW: login timeout returned");
+            SQL_SUCCESS
+        }
+        // Attributes not backed by stored state report success without writing,
+        // matching the historical stub behavior.
+        _ => SQL_SUCCESS,
```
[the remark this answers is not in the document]
Vahid-b: Agreed — returning `SQL_SUCCESS` without writing `value_ptr` leaves the caller's buffer undefined while claiming success, and it was inconsistent with the set-side, which already rejects unknown attributes with `HYC00`.

Fixed in 95bc1c90: the `_` arm now posts `HYC00` ("Connection attribute not supported") and returns `SQL_ERROR`, mirroring `sql_set_connect_attr_w`. Added an `unsupported_attribute_returns_error` test to cover it.

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,839 @@
...
+use mssql_tds::connection::client_context::{EntraIdTokenFactory, TdsAuthenticationMethod};
+use mssql_tds::core::TdsResult;
+use mssql_tds::error::Error;
+use mssql_tds::security::SecurityError;
+
+/// Well-known Microsoft public-client id used by msodbcsql / `SqlClient` for
+/// ActiveDirectory Interactive. It is registered with an `http://localhost`
+/// loopback redirect for native clients, so reusing it matches the classic
+/// driver and needs no per-app registration.
+const PUBLIC_CLIENT_ID: &str = "a94f9c62-97fe-4d19-b06d-472bed8d2bcb";
```
shiwanigupta0809: I didnt find this client id in msodbcsql repo. Is this a different client id that's intentionally used here? Just curious, as the comment says it's the client id used by msodbcsql.

Vahid-b: Switched to msodbcsql's actual ODBC client id, `2c1229aa-16c5-4ff5-b46b-4f7fe2a2a9c8` (`tds/Parse.cpp:3608` , `aadauth/AzureADAuth.cpp:839` ), so both drivers present one identity to Entra.



One gap remains, now documented on `PUBLIC_CLIENT_ID` : msodbcsql only signs in through the Windows WAM broker, so that app is registered for broker redirects only — `http://localhost` isn't registered yet. Until it is, the flow fails late, after the user has already completed MFA (`AADSTS50011`). The redirect timeout now points at the browser and names that error instead of expiring silently.



T3 is split into three work items so this PR isn't blocked on the registration turnaround:

| Work item | Scope |
| --- | --- |
| **AB#46067 — Part 1 (this PR)** | OAuth2 auth-code + PKCE loopback flow on all platforms; separate `SQL_ATTR_LOGIN_TIMEOUT` and new `SQLGetConnectAttr` for msodbcsql parity |
| **AB#46684 — Part 2** | Windows interactive via `mssql-auth` (OneAuth/WAM), matching msodbcsql. Its broker redirect is already registered, so it isn't blocked by Part 3. This PR's flow stays as the Windows fallback for contexts WAM can't serve, e.g. Session 0. |
| **AB#46683 — Part 3** | Add `http://localhost` to app `2c1229aa`. Unblocks Linux/macOS, where loopback is the only option — msodbcsql compiles its WAM path out off Windows, and this is what SqlClient uses there. |


On Windows this PR ships loopback as an interim, superseded by Part 2. A caller-supplied `client_id` override would sidestep the registration dependency, but msodbcsql has no such keyword and it would let a connection string redirect sign-in to an arbitrary app — out of scope in AB#46067.

I plan to merge Part 1 (current PR) now, documented as non-functional until AB#46683 lands, or hold it until the registration is in place.

Vahid-b:
Heads-up before you re-review: **this PR has been redesigned, so a fresh read is needed rather than a diff against your last review.**

The loopback + PKCE + system-browser flow is gone. Interactive sign-in now delegates to **`mssql-auth.dll` (OneAuth)** — the same library msodbcsql loads. `auth/interactive.rs` is rewritten, `auth/msqa.rs` is new, and `base64`/`getrandom`/`sha2`/`serde`/`serde_json` are dropped.

**Why:** we present msodbcsql's own client id, so the two drivers have to behave identically to callers and to the service. Hand-rolling our own OAuth flow behind that client id was the wrong call — same identity, different behaviour.

**What that means for your earlier comments:** the threads about PKCE, `state` validation, redirect classification, browser launching and child-process reaping no longer refer to any code in this PR. Nothing was dismissed — the code they applied to was deleted.

**Platform behaviour:** interactive is Windows-only, matching msodbcsql (`SNI_FedAuth` is excluded from the Unix `SRCLIST` and the DLL is loaded via `LoadLibraryExA` with no `dlopen` counterpart). On Linux/macOS the keyword falls through to `ActiveDirectoryIntegrated`, which is what msodbcsql itself does at `Parse.cpp:3597-3660`. I prototyped an explicit "not supported on this platform" error and withdrew it — refusing a keyword msodbcsql accepts would be the one place the drivers visibly disagree.

Also still in scope: `SQL_ATTR_LOGIN_TIMEOUT` plus a real `SQLGetConnectAttrW`.

Verified live against Azure SQL, not just unit tests: sign-in completes, a second connection reuses the cached context without prompting (0.1s vs 19.4s), and cancelling fails exactly once without reopening the window.

One open question worth your input: **`mssql-auth.dll` is not redistributed by this driver.** Interactive works where msodbcsql 18 has already put it in `System32`. Packaging needs a decision before GA — happy to take that to a separate thread.

Details and parity citations are in the updated PR description.

mssql-odbc/src/auth/msqa.rs
```diff
@@ -0,0 +1,676 @@
...
+    if created == MSQA_SUCCESS {
+        // The DSN test dialog can still hold the mouse; releasing it lets the
+        // user interact with the sign-in window (`SNI_FedAuth.cpp:470`).
+        let _ = unsafe { ReleaseCapture() };
+
+        let mut message = MSG::default();
+        while unsafe { GetMessageW(&mut message, None, 0, 0) }.as_bool() {
+            let _ = unsafe { TranslateMessage(&message) };
+            unsafe { DispatchMessageW(&message) };
+        }
```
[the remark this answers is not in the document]
Vahid-b: Good catch, fixed in 0f08e9c2.

Confirmed against the crate source: `GetMessageW` returns `BOOL` (`windows-0.58.0/src/Windows/Win32/UI/WindowsAndMessaging/mod.rs:1360`) and `BOOL::as_bool()` is `self.0 != 0` (`windows-0.58.0/src/extensions/Win32/Foundation/BOOL.rs:5`), so the `-1` error return was indeed being read as "message available".

The consequence was worse than a stray dispatch: `run_interactive_ui` blocks on `worker.join()`, so a persistent error would have spun the pump forever, hung the connection, and leaked the STA thread. The outer login deadline cancels the future but cannot kill an OS thread.

This is also a parity divergence - msodbcsql tests the value explicitly:

```cpp
// SNI_FedAuth.cpp:474
while (::GetMessage(&msg, NULL, 0, 0) > 0)
```

The three cases are now handled separately, with the failure path logging `GetLastError` before leaving the pump.

mssql-tds/src/connection/client_context.rs
```diff
@@ -225,7 +225,16 @@ pub struct ClientContext {
...
+    ///
+    /// Bounds each individual TCP-connect attempt (the network reach), not the
+    /// whole login. See [`Self::login_timeout`] for the overall login deadline.
     pub connect_timeout: u32,
+    /// Overall login deadline in seconds, covering the network connect, the TDS
+    /// handshake, and any auth token acquisition (e.g. the interactive browser
+    /// flow). `None` falls back to [`Self::connect_timeout`] (preserving the
+    /// historical single-knob behavior); `Some(0)` disables the deadline (wait
+    /// indefinitely). Maps to the ODBC `SQL_ATTR_LOGIN_TIMEOUT` attribute.
+    pub login_timeout: Option<u32>,
```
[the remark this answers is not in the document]
Vahid-b: The source-breaking mechanics are right, but the impact is not what the comment assumes, so I'm leaving `ClientContext` as-is.

`mssql-tds` is unpublished (`crates.io/api/v1/crates/mssql-tds` returns 404) and sits at `0.1.0`, so there are no downstream crates that could have a full struct literal to break. Every `ClientContext { .. }` literal in the repo lives inside `mssql-tds` itself; the sibling crates that consume it (`mssql-odbc`, `mssql-js`, `mssql-tds-cli`, `mssql-py-core`) build it through `Default` plus field assignment or `From`, never a literal. `cargo clippy --workspace --all-features --all-targets` is clean, which covers all of them.

So the "back-compatible" claim in the PR description means no in-tree caller needs edits, which holds. Pre-1.0 and unpublished, there is no external API contract to break yet.

Worth doing eventually, just not here: `#[non_exhaustive]` on `ClientContext` would make this class of change permanently additive. That is a deliberate API decision affecting every consumer and belongs in its own change rather than riding along with interactive auth.

mssql-odbc/src/auth/interactive.rs
Vahid-b: You were right to question it - the id you looked at was wrong, and it has since been corrected.

The PR has also been reworked since: interactive auth now goes through `mssql-auth.dll` (OneAuth/MSQA), the same library msodbcsql uses, rather than the browser/loopback flow that was here when you commented. The client id and redirect URI are now taken verbatim from msodbcsql:

```cpp
// Sql/Ntdbms/sqlncli/tds/Parse.cpp:3606-3608
L"https://sqlaad/",
*m_pLoginInfo->pCurrentConnectionId,
L"2c1229aa-16c5-4ff5-b46b-4f7fe2a2a9c8",   // ODBC client ID
```

Matching both matters beyond cosmetics: the client id is what a tenant's sign-in logs and conditional-access policies key on, so diverging would make this driver show up as a different application and potentially be blocked by policies written for the ODBC driver. There is now a unit test (`client_id_matches_msodbcsql`) pinning both values so they cannot drift.

Vahid-b:
## Review thread triage

I went back through all 23 open threads and validated each one against the code as it stands now rather than as it stood when the comment was written. Two pushes since: 1f0a7b69 and 0f08e9c2.

**One was a live bug and is fixed (0f08e9c2)** - `msqa.rs` message pump. `GetMessageW` returns `-1` on error, and `BOOL::as_bool()` is a plain `!= 0`, so the error return was read as "message available". Because `run_interactive_ui` blocks on the thread join, a persistent error would have hung the connection and leaked the STA thread. msodbcsql tests `> 0` (`SNI_FedAuth.cpp:474`); we now do too. Thanks Copilot - real find.

**One more parity gap I found while re-reading, fixed (1f0a7b69)** - `SQL_ATTR_LOGIN_TIMEOUT` was not clamped. msodbcsql caps at `MAX_QUERY_TIMEOUT` (`0xfffe`) and posts `01S02` "Login timeout changed" (`sqlcmisc.cpp:1735`). We accepted anything and narrowed with a raw `as u32`, so a request for 2^32 seconds wrapped to `0`, which the connect path reads as *wait indefinitely* - the opposite of the caller's intent. Clamping at pointer width before narrowing makes the wrap unreachable. Three tests added.

**14 were obsolete**, all describing the browser/loopback/PKCE flow that the MSQA pivot deleted. Verified by symbol, not assumption - `xdg-open`, `/usr/bin/open`, `TcpListener`, `reqwest`, `code_verifier`, `READ_TIMEOUT` and the authorization-URL logging are all gone from `mssql-odbc`. The `OnceCell` flagged in `interactive.rs` is gone as well; the one left in `entra.rs` caches the *credential*, not a token, so the staleness concern does not carry over.

**6 were fixed in earlier rounds and simply never marked resolved** - the `get_connect_attr.rs` cluster (layering now `ffi_entry!` -> unsafe shim -> safe core, `write_if_some`, full entry logging, canonical `ERR_INVALID_NULL_POINTER`, `HYC00` for unsupported attributes), the login-deadline placement, and the shared-memory `Err` arm now surfacing non-transient failures instead of falling through to SSRP and re-prompting.

**1 was valid but inert** - `ClientContext` field addition. Answered in-thread: the crate is unpublished and pre-1.0, and no sibling crate constructs it with a struct literal, so nothing can break. `#[non_exhaustive]` is worth doing, but as its own API change.

Those 22 are now resolved. I left @shiwanigupta0809's client-id question open since it is hers to close - short version: the id was wrong when you asked and is now taken verbatim from `Parse.cpp:3608`, with a test pinning it.

mssql-odbc/src/auth/msqa.rs
```diff
@@ -0,0 +1,693 @@
...
+        self.0.load(std::sync::atomic::Ordering::SeqCst)
+    }
+}
+
+/// Closes a sign-in window that is still open, used when the caller's login
+/// deadline expires.
+///
+/// The pump clears its id before it returns, so this cannot post to a thread id
+/// that has already been recycled by another thread.
+pub(super) fn cancel_ui(ui_thread_id: &UiThreadId) {
```
David-Engel: **Blocking:** `WM_QUIT` can be posted to a recycled thread id.

The doc comment above says "the pump clears its id before it returns, so this cannot post to a thread id that has already been recycled", but the load and the post are not synchronized:

```rust
let thread_id = ui_thread_id.get();   // reads X
// pump completes -> clear() -> thread exits -> OS recycles X
let _ = unsafe { PostThreadMessageW(thread_id, WM_QUIT, ...) };  // posts to a stranger
```

This is reachable whenever the login deadline expires at roughly the same moment the user finishes signing in. Inside a host application that owns its own message loops, posting `WM_QUIT` to an unrelated thread can look like the app quitting.

Making the id a `Mutex<u32>` closes the window, because the pump cannot clear (and therefore cannot exit) while a canceller holds the lock across its post:

```rust
pub(super) struct UiThreadId(Mutex<u32>);

pub(super) fn cancel_ui(ui: &UiThreadId) {
    let id = ui.0.lock().unwrap_or_else(|p| p.into_inner());
    if *id == 0 {
        return;
    }
    let _ = unsafe { PostThreadMessageW(*id, WM_QUIT, WPARAM(0), LPARAM(0)) };
}
```

Related: `ui_complete` carries the same hazard if OneAuth can ever fire the callback after a cancelled pump has exited but before `MSQADeleteRequest` runs. Worth confirming OneAuth guarantees the callback cannot outlive the request.

mssql-odbc/src/auth/entra.rs
```diff
@@ -243,6 +249,38 @@ pub(crate) fn configure_auth(
...
+        // msodbcsql has no interactive path off Windows: `SNI_FedAuth` is not
+        // compiled at all (its Makefile omits the translation unit) and the
+        // dispatch site is removed by `#if !defined(XPLAT_ODBC_TODO)`
+        // (`Parse.cpp:3597`). The request lands in the generic `AzureADAuth`
+        // block, whose `authMode` ternary (`:3657-3660`) has no Interactive arm
+        // and so resolves to `AKVCFG_AUTHMODE_INTEGRATED` — a Kerberos attempt
+        // against the STS. Report it as Integrated for the same reason; once
+        // that method is implemented this becomes msodbcsql's behaviour exactly.
+        #[cfg(not(windows))]
+        TdsAuthenticationMethod::ActiveDirectoryInteractive => {
```
David-Engel, line 305: **Blocking:** off Windows this reports a method the caller never asked for.

`configure_auth` returns `Err(ActiveDirectoryIntegrated)`, and `driver_connect.rs` formats that straight into the diagnostic: `"Authentication method ActiveDirectoryIntegrated is not yet supported"`. A Linux user who wrote `Authentication=ActiveDirectoryInteractive` now sees an error naming a keyword they never used, with nothing tying it back to what they asked for.

The parity argument justifies *resolving* to Integrated; it doesn't justify the message. Either carry the requested method through for the text, or post something like "ActiveDirectoryInteractive resolves to ActiveDirectoryIntegrated on this platform, which is not yet supported."

Everything else about this arm — including the reasoning in the comment — looks right.

mssql-odbc/src/auth/msqa.rs
```diff
@@ -0,0 +1,693 @@
...
+    let context = get_or_create_context(api, sts_url, login_hint, client_id, redirect_uri)?;
+    let resource_c = to_c_string(resource, "resource")?;
+
+    // A fresh correlation id per acquisition, so a failed sign-in can be
+    // located in the tenant's Entra sign-in logs.
+    let correlation_id = GUID::new().unwrap_or(GUID::zeroed());
+
+    // Only one sign-in window per context at a time: without this, two
+    // connections opening concurrently would each raise their own prompt.
+    let _serialized = context
```
David-Engel: **Suggestion:** this lock can pin blocking-pool threads for minutes.

`acquire_lock` is held for the entire interactive sign-in, up to 330s. A second connection's `spawn_blocking` task blocks here with no deadline of its own: its `CancelUiOnDrop` fires with `ui_thread_id == 0` (a no-op, since that connection never opened a window), the future is dropped, but the blocking task stays parked on the mutex until the first sign-in finishes. N concurrent connections park N blocking threads, starving other `spawn_blocking` users.

Bounding the wait — e.g. `try_lock` against the caller's remaining deadline — would let a cancelled connection release its thread.

mssql-tds/src/connection_provider/tds_connection_provider.rs
```diff
@@ -171,6 +189,19 @@ impl TdsConnectionProvider {
                         return Ok(client);
                     }
                     Err(err) => {
+                        // Only transient (transport-level) shared-memory failures
+                        // fall through to SSRP/TCP. A definitive failure — e.g. the
+                        // user cancelled interactive sign-in (`AuthenticationDenied`)
+                        // or the login was rejected — would recur on every transport,
+                        // and falling through would relaunch the interactive browser.
+                        // Surface it immediately.
+                        if !err.is_transient_connect_error() {
```
David-Engel, line 198: **Suggestion:** this is a broader behavior change than the PR title implies, and it's untested.

Returning `Err` on any non-transient shared-memory failure affects every Windows local-named-instance connection, not just interactive ones. `Error::Security(_)` and `Error::ProtocolError(_)` now abort instead of retrying over TCP, so (for example) a TLS negotiation failure specific to the SM transport no longer gets a TCP second chance.

I agree with the intent — falling through would relaunch the sign-in window — but the only new test is `authentication_denied_is_not_transient` in the error module, which doesn't exercise this branch. Worth a provider-level test covering both arms: transient falls through to SSRP, non-transient returns immediately.

mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -0,0 +1,207 @@
...
+        }
+        // Any other attribute is genuinely unsupported: surface HYC00 instead of
+        // claiming success while leaving the caller's buffer untouched. Mirrors
+        // the set-side (`sql_set_connect_attr_w`).
+        _ => {
+            error!(
+                attribute,
+                "SQLGetConnectAttrW: unsupported connection attribute"
+            );
+            post_sql_error(
```
David-Engel, line 139: **Suggestion:** the get side rejects attributes the set side accepts.

`sql_set_connect_attr_w` accepts `SQL_ATTR_ACCESS_MODE`, `SQL_ATTR_CONNECTION_TIMEOUT`, `SQL_ATTR_PACKET_SIZE`, and `SQL_ATTR_ANSI_APP` as no-ops, but they land in this arm and get `HYC00`. An application doing set-then-get sees a failure for a value the driver just claimed to accept.

Not a regression — the old implementation was a hard `SQL_ERROR` stub — but aligning the two sides (return the stored or default value for the same set the set side tolerates) would avoid a confusing asymmetry.

mssql-odbc/src/api/set_connect_attr.rs
```diff
@@ -116,11 +123,30 @@ unsafe fn sql_set_connect_attr_w_impl(
                 }
             }
         }
+        SQL_ATTR_LOGIN_TIMEOUT => {
```
David-Engel, line 126: **Suggestion:** missing the post-connect guard that `SQL_COPT_SS_ACCESS_TOKEN` has.

The access-token arm above correctly rejects a post-connect set with `HY011`. `SQL_ATTR_LOGIN_TIMEOUT` is equally connect-time only, but here it is silently accepted and stored where it will never be read for this connection. Same `state.connection_state != ConnectionState::Disconnected` check would cover it.

mssql-odbc/src/auth/msqa.rs
```diff
@@ -0,0 +1,693 @@
...
+/// process is always explicit; [`acquire_token`] clears it once a token has
+/// been obtained. Mirrors `MSQAAuthContextCache::getOrCreate`.
+fn get_or_create_context(
+    api: &'static MsqaApi,
+    sts_url: &str,
+    login_hint: &str,
+    client_id: &str,
+    redirect_uri: &str,
+) -> TdsResult<Arc<CachedContext>> {
+    let key = (login_hint.to_string(), sts_url.to_string());
```
David-Engel, line 309: **Suggestion:** worth documenting what an empty `UID` means here.

With no `UID` the key is `("", sts_url)`, and `ForcePrompt` is cleared after the first success, so a second connection in the same process silently authenticates as whoever signed in first — no prompt, and no way to switch accounts.

This matches msodbcsql's `username + "\0" + stsUrl` key, so I read it as intended parity rather than a defect. Because it's security-visible, it deserves an explicit line in the module docs or the PR's known-limitations list.

mssql-odbc/src/auth/msqa.rs
```diff
@@ -0,0 +1,693 @@
...
+static CONTEXT_CACHE: OnceLock<ContextCache> = OnceLock::new();
+
+/// Loads `mssql-auth.dll` and resolves the entry points, once per process.
+///
+/// The search is restricted to System32 (as msodbcsql does at
+/// `SNI_FedAuth.cpp:249`) so a DLL dropped next to the application cannot
+/// hijack authentication. The module is never freed: OneAuth spins up
+/// background state that must outlive individual connections.
+fn api() -> TdsResult<&'static MsqaApi> {
+    let loaded = MSQA_API.get_or_init(|| {
```
David-Engel: **Suggestion:** a failed load is cached for the life of the process.

`MSQA_API` is a `OnceLock<Result<..>>`, so once `mssql-auth.dll` is missing every later connection fails without retrying — including after the user installs msodbcsql 18 mid-session. Given that packaging is called out as unresolved, caching only the success and retrying the load on failure would be friendlier.

mssql-odbc/src/auth/msqa.rs
```diff
@@ -0,0 +1,693 @@
...
+/// `request` must be a live `HMSQAREQUEST` that no other thread is using for
+/// the duration of this call.
+unsafe fn pump_sign_in_window(
+    api: &'static MsqaApi,
+    request: HMsqaRequest,
+    title: &[u16],
+    ui_thread_id: &UiThreadId,
+) -> i32 {
+    // OneAuth's window is a COM single-threaded apartment object.
+    let com = unsafe { CoInitializeEx(None, COINIT_APARTMENTTHREADED) };
```
David-Engel, line 487: **Nit:** an HRESULT ends up in an MSQA status slot.

`return com.0` feeds an HRESULT into `describe_failure`, which then queries `MSQAGetErrorDescription` on a request OneAuth never failed, producing something like `"sign-in failed with status -2147417850"`. A distinct `Err` for the COM-init path would give a clearer message.

mssql-odbc/src/auth/msqa.rs
```diff
@@ -0,0 +1,693 @@
...
+/// Called by OneAuth with the `callback_data` supplied to
+/// `MSQAUICreateHostWindow`, which is always a thread id cast to a pointer.
+unsafe extern "system" fn ui_complete(_request: HMsqaRequest, data: *mut c_void) {
+    let thread_id = data as usize as u32;
+    let _ = unsafe { PostThreadMessageW(thread_id, WM_QUIT, WPARAM(0), LPARAM(0)) };
+}
+
+/// Reads the acquired token with the two-pass length-then-buffer protocol
+/// `MSQAGetAccessToken` uses.
+fn read_access_token(api: &'static MsqaApi, request: HMsqaRequest) -> TdsResult<String> {
```
David-Engel, line 588: **Nit:** the two-pass length convention deserves one line of comment.

The second call reuses the first call's `length` as the buffer size while allocating `length + 1`. If `MSQAGetAccessToken` ever reports the length *including* the terminator, `buffer.truncate(length)` would leave a trailing NUL inside the bearer token. The live verification suggests it excludes the terminator — stating which convention was confirmed would save the next reader from re-deriving it.

mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -0,0 +1,207 @@
...
+};
+use crate::api::util::write_if_some;
+use crate::error::{free_errors, post_sql_error};
+use crate::handles::{DbcHandle, HandleType, handle_from_raw};
+
+/// Login timeout reported when the application has not set
+/// `SQL_ATTR_LOGIN_TIMEOUT`. Mirrors `ClientContext::connect_timeout`'s default,
+/// which is what the connect path falls back to when no explicit login timeout
+/// is present.
+const DEFAULT_LOGIN_TIMEOUT_SECS: u32 = 15;
```
David-Engel: **Nit:** this duplicates `ClientContext::connect_timeout`'s default. The comment acknowledges the coupling; a shared constant would keep the two from drifting.

David-Engel (commented):
## Review summary

Adds `Authentication=ActiveDirectoryInteractive` to mssql-odbc by delegating to `mssql-auth.dll` (OneAuth/MSQA) on Windows, splits `connect_timeout` into a per-TCP-connect cap and a new overall `login_timeout`, and implements a real `SQLGetConnectAttrW` for `SQL_ATTR_LOGIN_TIMEOUT`.

The redesign away from a hand-rolled PKCE flow is the right call, and the FFI layer is careful: RAII request guard, dedicated STA thread, the `GetMessageW > 0` fix, embedded-NUL rejection, and traceable msodbcsql references throughout. `cargo test -p mssql-odbc --lib` passes locally on Linux (440 tests).

Findings are inline. Two I'd like resolved before merge:

1. **`WM_QUIT` can be posted to a recycled thread id** in `cancel_ui` — the load-then-post is unsynchronized, and in an ODBC driver hosted inside an arbitrary application this can tear down the host app's message loop.
2. **The off-Windows diagnostic names a method the caller never asked for** — a Linux user who writes `Authentication=ActiveDirectoryInteractive` gets "Authentication method ActiveDirectoryIntegrated is not yet supported", which is undiagnosable.

The rest are suggestions and nits.

### Tests

Coverage of the pure logic is good: clamping, `0` to infinite, app-set precedence, transient classification, cache-key round-trip, NUL rejection. The gap is that roughly 690 lines of `msqa.rs` — the pump, the cancel path, the token and error reads — are exercised only by the manual live run, and both blocking findings live in exactly that untested region. Given the 85% bar, consider extracting the cancel/thread-id lifecycle behind a small trait so the race is testable without the DLL.

Vahid-b:
Re-queued validation (build 164519). The previous run failed only on `mssql-py-core`'s `test_tracing_end_to_end`, a known intermittent flake: the test waits on `tracing_appender::non_blocking`'s background writer with a fixed `thread::sleep(100ms)`, so on a loaded agent the log file is only partially flushed when it is read. Same test failed the same way on PR #141 (build 162171) and PR #143 (build 163026), and the failing assertion moves between the first and second message across runs. Nothing in this PR touches `mssql-py-core`.

mssql-odbc/src/auth/msqa.rs
Vahid-b: Fixed in f69fc566.

You're right, and the doc comment was the tell — clearing before returning does nothing for a reader that already copied the id out. `UiThreadId` is now a `Mutex<u32>` and `post_quit` holds the guard across `PostThreadMessageW`, so the pump cannot reach `clear()` (and therefore cannot return) while a post is in flight.

On your OneAuth question: I could not find a guarantee either way, so I stopped depending on one. `ui_complete` previously posted to a bare thread id captured at window creation; it now goes through the same guarded path, so a callback arriving after a cancelled pump has exited sees id 0 and does nothing.

Worth recording that msodbcsql never had this exposure: `UIComplete` (`SNI_FedAuth.cpp:430`) is its only `PostThreadMessage` caller, and its pump only ever ends on that callback, so its target is always alive. The deadline cancellation is ours, so the race was ours too.

mssql-odbc/src/auth/entra.rs
Vahid-b: Fixed in f69fc566. Agreed — parity justifies the resolution, not the message.

`configure_auth` now returns `UnsupportedAuth { requested, resolved }` rather than a bare method, and `driver_connect` formats the two cases separately:

- unchanged: `Authentication method X is not yet supported`
- resolved: `Authentication method ActiveDirectoryInteractive resolves to ActiveDirectoryIntegrated on this platform, which is not yet supported`

The resolution itself is untouched, so the msodbcsql fall-through still holds. The off-Windows test now asserts the requested method survives into the error.

mssql-odbc/src/auth/msqa.rs
Vahid-b: Tracked as a follow-up rather than changed here — [AB#46980](https://dev.azure.com/SqlClientDrivers/mssql-rs/_workitems/edit/46980), under the auth story. Evidence for why it is lower risk than it first looks:

- msodbcsql serializes identically, with a plain blocking lock and no deadline (`SNI_FedAuth.cpp:638`, commented "we need to synchronize concurrent AcquireToken calls on the same context to only show one pop-up at a time").
- The long hold only happens on the **first** sign-in. `acquire_token` clears `ForcePrompt` after the first success, so later acquisitions on that context are answered from OneAuth's cache without a prompt and release the lock promptly.
- The hold is bounded rather than indefinite: if the first sign-in never completes, `CancelUiOnDrop` fires at the login deadline, the pump exits, and `acquire_token` returns and drops the lock.

So the exposure is N-1 blocking threads parked for the length of one human sign-in, and exhausting Tokio's 512-thread blocking pool needs >512 concurrent interactive connects in a single process — not a shape that occurs for a flow requiring a human at a desktop.

I also don't think `try_lock` against the deadline is the right fix: the waiter is parked for the same wall-clock time either way, it just fails at the end instead of proceeding. The fix that actually helps is making the wait cancellable — resolve the context in its own `spawn_blocking`, then take a `tokio::sync::Mutex` via `lock_owned().await` before the second one, so a caller that hits its deadline is dropped from the waiter queue and never occupies a blocking thread. That restructures `acquire_token` enough that I would rather not fold it into this PR.

mssql-tds/src/connection_provider/tds_connection_provider.rs
Vahid-b: Test added in f69fc566: `shared_memory_falls_through_only_on_transient_failures`, in the provider module, covering both arms.

You're right that the change is broader than the title suggests — it applies to every Windows local named-instance connect, and `Error::Security(_)` / `Error::ProtocolError(_)` now abort instead of getting a TCP second chance. The test pins exactly that split: `ConnectionError`, `TimeoutError` and `Io` fall through; `Security(AuthenticationDenied)` and `ProtocolError` do not.

Being straight about its limits: the branch needs a live local named instance to drive end to end, so the test asserts the policy the branch applies rather than exercising a real shared-memory transport. If you want the branch itself covered I would need to make `connect_with_transport_context` injectable, which I would rather do as its own change.

mssql-odbc/src/api/get_connect_attr.rs
Vahid-b: Fixed in f69fc566 — and it turned out to be a parity gap as well, not just an asymmetry. msodbcsql answers `SQL_ATTR_ACCESS_MODE` (`sqlcmisc.cpp:3038`), `SQL_ATTR_PACKET_SIZE` (`:3092`) and both timeouts (`:3389-3391`) on the get side.

Those three attributes are now stored on the DBC and reported back, so a set/get round-trip agrees. `SQL_ATTR_ANSI_APP` deliberately still returns `HYC00`: the Driver Manager sets it, ODBC defines no way to read it back, and msodbcsql has no get arm for it either.

Reading msodbcsql for this also surfaced a real bug on the set side — see my reply on your `SQL_ATTR_LOGIN_TIMEOUT` comment.

mssql-odbc/src/api/set_connect_attr.rs
Vahid-b: I dug into this one and I don't think the guard belongs on this attribute — msodbcsql deliberately does not have it.

In the same `switch`, the genuinely connect-time-only attributes carry an explicit connected check:

- `SQL_ATTR_PACKET_SIZE` (`sqlcmisc.cpp:1901-1906`): `if (lpdbc->hConn) { wError = IDS_S1_011; goto ErrorRet; }` — that is the HY011.
- `SQL_ATTR_ACCESS_MODE` (`:1696`): `if (lpdbc->hConn)`, with its own handling.

`SQL_ATTR_LOGIN_TIMEOUT` has neither (`:1733-1748`). It clamps, then calls `SetLoginTimeout(lpdbc, vParam)` unconditionally, which is a plain store (`sqlcprot.h:1568-1571`).

And the value is not dead. An ODBC connection handle is reusable, and our `SQLDisconnect` does not clear `login_timeout` — it drops the client and resets `connection_state` only (`disconnect.rs:77-80`) — so a login timeout set while connected applies to the next `SQLDriverConnect` on that handle. That is the normal pattern for a pooled DBC.

Added `login_timeout_after_connect_is_accepted` in f69fc566 to pin that behaviour, with the msodbcsql citation in the code.

That said, the comment found a real bug one arm over: we were silently accepting a post-connect `SQL_ATTR_PACKET_SIZE`, which msodbcsql rejects because packet size is fixed by the LOGIN7 handshake. That now returns HY011, with `packet_size_after_connect_is_rejected` covering it. Thanks — I would not have gone looking otherwise.

mssql-odbc/src/auth/msqa.rs
Vahid-b: Documented in f69fc566. The `msqa` module docs now carry a "Known limitation: account reuse when no UID is supplied" section spelling out that the key becomes `("", sts_url)`, that `ForcePrompt` is cleared after the first success, and that every later connection to the same authority silently reuses whoever signed in first with no way to switch accounts for the life of the process — with the guidance to supply `UID` to pin a connection to an account. `get_or_create_context` points at it.

Behaviour left alone, since it does match msodbcsql's `username + "\0" + stsUrl` key. Agreed it needed to be written down rather than inferred from the cache key.

mssql-odbc/src/auth/msqa.rs
Vahid-b: Fixed in f69fc566. `MSQA_API` is now `OnceLock<MsqaApi>` and only a successful load is cached. The load moved into `load_api()`, and a failure returns without ever initializing the cell, so a later connection re-attempts `LoadLibraryExA` — including after the user installs msodbcsql 18 mid-process.

If two threads race a first successful load, one table wins and the other is dropped; it is only fn pointers, and the extra `LoadLibraryExA` reference is deliberately never freed anyway.

mssql-odbc/src/auth/msqa.rs
Vahid-b: Fixed in f69fc566. `pump_sign_in_window` and `run_interactive_ui` now return `TdsResult<i32>`, and COM init failure returns `SecurityError::InternalError` carrying the HRESULT ("the interactive sign-in thread could not initialize COM (HRESULT 0x...)") instead of an `i32` that `describe_failure` would read as a OneAuth status. That also stops it querying `MSQAGetErrorDescription` about a request OneAuth never failed.

mssql-odbc/src/auth/msqa.rs
Vahid-b: Documented in f69fc566. The doc comment on `read_access_token` now states that the reported length is a character count that *excludes* the terminating NUL — which is why the second call allocates `length + 1` while still passing `length` as the buffer size, and why truncating to `length` yields the token alone — and spells out your failure mode: were it ever to include the terminator, the truncation would leave an embedded NUL inside the bearer token and the server would reject the login.

That matches what the live sign-in produced. I have not found it stated in `msqa_api.h`, so the comment records the assumption and its consequence rather than claiming a contract.

mssql-odbc/src/api/get_connect_attr.rs
Vahid-b: Fixed in f69fc566. There was no constant to share — `connect_timeout: 15` was a bare literal duplicated across both `ClientContext` constructors — so I added `DEFAULT_CONNECT_TIMEOUT_SECS` in `mssql-tds`, used it in both, and `DEFAULT_LOGIN_TIMEOUT_SECS` in the ODBC layer now derives from it. One definition, and the two can no longer drift apart.

mssql-odbc/src/auth/msqa.rs
Vahid-b: Reversing this one: fixed in 16b8d09e rather than deferred.

I sized it wrong the first time. I claimed it needed `acquire_token` restructured and called that too large for this PR — but on actually measuring it, it is two files and about 40 lines, and nothing else in the tree touches that lock. Everything I said about the exposure being bounded still holds; that argued the bug was low *severity*, not that it was expensive to fix, and I let the first conclusion carry the second.

What changed:

- `get_or_create_context` is now reachable through `resolve_context`, run in its own `spawn_blocking`, so the context is in hand before anything locks.
- `CachedContext::sign_in_lock` is an `Arc<tokio::sync::Mutex<()>>`, taken with `lock_owned().await` in `InteractiveTokenFactory::acquire`. A connection that hits its login deadline while another is signing in is now dropped from the waiter queue instead of parking a blocking-pool thread.
- `acquire_token` takes the resolved `&CachedContext` and no longer locks; the guard moves into the blocking closure.

One thing worth calling out, since it is not obvious from the diff: the guard is held *inside* the blocking task, not in the async caller. If the async side held it, a dropped future would release the lock the instant it unwound — while the cancelled sign-in was still tearing its window down — and the next waiter could open a second prompt on top of it. Moving it in keeps the original invariant (lock held for the true duration of the acquisition, cancellation included) while making only the *wait* cancellable.

Four tests, all timer-free and on a single-threaded runtime, which is deliberate: a lock that parks its waiter instead of yielding deadlocks that runtime outright, so a revert to a blocking mutex fails loudly rather than silently.

- `a_second_sign_in_waits_without_stalling_the_runtime`
- `an_abandoned_waiter_does_not_hold_up_the_next_sign_in` — the login-deadline case directly
- `a_panicking_acquisition_becomes_a_connection_error`
- `a_successful_acquisition_passes_its_value_through`

AB#46980 is closed as completed. Thanks for pressing on this.

David-Engel (commented):
## Re-review

All eleven findings from the previous pass are addressed, and several went further than what I asked for:

- The thread-id fix also re-routes `ui_complete` through the guarded path, so a late OneAuth callback can't post at a recycled thread either — not just the deadline canceller.
- The `HY011` question was answered better than I posed it: the guard belongs on `SQL_ATTR_PACKET_SIZE` (fixed by LOGIN7) and *not* on `SQL_ATTR_LOGIN_TIMEOUT`, with the msodbcsql citation showing it stores that one unconditionally on a reusable handle.
- The async `sign_in_lock` correctly keeps the guard inside the blocking task, so a dropped future can't release it while the window is still tearing down. The current-thread-runtime tests are a neat way to prove the waiter yields rather than parks.

`cargo test -p mssql-odbc --lib` passes (446) and `cargo clippy -p mssql-odbc -p mssql-tds --all-targets` is clean on Linux.

Two nits inline, neither merge-blocking.

### Where I'd still spend manual attention

Everything remaining is Windows-only FFI that CI cannot reach:

1. **`ui_complete`'s pointer lifetime.** The callback now dereferences `data` as `&UiThreadId`. The ordering in the code is right — `MSQADeleteRequest` runs inside `acquire_token` before the closure's `Arc` drops — but the safety argument rests on OneAuth never invoking the callback after the request is deleted, which is an assumption about a closed-source DLL. The fix traded a narrow stray-`WM_QUIT` window for a use-after-free invariant, so it's worth confirming the MSQA contract actually guarantees it.
2. **Live re-run of the deadline-expiry-with-window-open path,** since `post_quit` now holds a lock across `PostThreadMessageW` while the pump may be inside `DispatchMessageW`. `PostThreadMessageW` is documented as non-blocking so this should be fine, but it is exactly the path the change touches and no test can reach it.
3. **A DM smoke test for the new `SQL_ATTR_PACKET_SIZE` `HY011`,** which is new behaviour for every connection rather than just interactive ones.

mssql-odbc/src/auth/interactive.rs
```diff
@@ -0,0 +1,246 @@
...
+/// Closes an in-flight sign-in window if the acquisition future is dropped —
+/// which is what happens when the connection's login deadline expires.
+struct CancelUiOnDrop {
+    ui_thread_id: Arc<super::msqa::UiThreadId>,
+}
+
+impl CancelUiOnDrop {
+    /// Suppresses cancellation after the acquisition has already returned.
+    fn disarm(self) {
+        std::mem::forget(self);
```
David-Engel: **Nit:** `mem::forget(self)` leaks the `Arc<UiThreadId>` on every successful sign-in.

Forgetting the wrapper suppresses `Drop` for the `Arc` too, so the strong count is never decremented and the `UiThreadId` allocation is never freed. Small and bounded per connection, but unbounded across a long-lived process. Holding the `Arc` in an `Option` disarms without leaking:

```rust
struct CancelUiOnDrop {
    ui_thread_id: Option<Arc<super::msqa::UiThreadId>>,
}

impl CancelUiOnDrop {
    fn disarm(mut self) {
        self.ui_thread_id = None;
    }
}

impl Drop for CancelUiOnDrop {
    fn drop(&mut self) {
        if let Some(id) = &self.ui_thread_id {
            super::msqa::cancel_ui(id);
        }
    }
}
```

Worth flagging that this is coupled to the `ui_complete` pointer lifetime: the leak currently keeps the `UiThreadId` allocation alive forever, which masks the question of whether OneAuth can call back after `MSQADeleteRequest`. The ordering inside the blocking closure looks correct — `acquire_token` deletes the request before `ui_thread_id` drops — so removing the leak should be safe, but it's the reason to be sure of that contract rather than incidentally protected from it.

mssql-odbc/src/api/set_connect_attr.rs
```diff
@@ -116,14 +123,71 @@ unsafe fn sql_set_connect_attr_w_impl(
...
+            state.access_mode = value_ptr as usize as u32;
+            SQL_SUCCESS
+        }
+        SQL_ATTR_CONNECTION_TIMEOUT => {
+            // Shares msodbcsql's clamp with SQL_ATTR_LOGIN_TIMEOUT
+            // (`sqlcmisc.cpp:1733-1741`).
+            let requested = value_ptr as usize as u64;
+            state.connection_timeout = requested.min(MAX_LOGIN_TIMEOUT_SECS) as u32;
+            if requested > MAX_LOGIN_TIMEOUT_SECS {
+                post_diag(&mut state, WARN_LOGIN_TIMEOUT_CHANGED);
```
David-Engel: **Nit:** wrong noun for this attribute.

`WARN_LOGIN_TIMEOUT_CHANGED` renders as "Login timeout changed", but the application set `SQL_ATTR_CONNECTION_TIMEOUT` here. The SQLSTATE is right; only the text misnames what was clamped. Either a second `DiagMsg` for the connection timeout, or a generic "Option value changed" shared by both arms, would keep the diagnostic honest.

saurabh500:
Make sure to merge this branch with the latest origin/main to get the benefits of coverage reporting.

mssql-odbc/src/auth/interactive.rs
Vahid-b: Fixed in 24d2c13c. Straightforward miss on my part — `mem::forget(self)` suppresses `Drop` for the whole struct, field included, so the strong count never came back down. Took your `Option` shape.

On the coupling you flagged: I traced it, and the leak was never what kept the pointer valid. The `Arc` that guarantees it is the one moved into the blocking closure, not the canceller's clone:

- `RequestGuard` drops at the end of `acquire_token`, so `MSQADeleteRequest` runs *inside* the call.
- The closure's `Arc` drops only after `acquire_token` has returned.
- That ordering holds on every path, cancelled or not, because the closure owns the `Arc` for its whole body.

So the allocation outlives request deletion whether or not the canceller leaks, and removing the leak doesn't move that line. Worth being explicit that this is the invariant, though — I've said so in the commit message, since "the leak protects us" would have been the wrong reason to feel safe.

The `#[test]` I added asserts `Arc::strong_count == 1` after `disarm`, which fails against the old `mem::forget` version.

The remaining assumption — that OneAuth won't invoke the callback after `MSQADeleteRequest` — is unchanged and still an assumption about a closed-source DLL. It is the same one msodbcsql makes, with less protection: it passes a bare thread id (`SNI_FedAuth.cpp:430`), so a post-delete callback there would post at a possibly-recycled id, where ours hits a cleared `UiThreadId` and no-ops. Agreed it's worth confirming against the MSQA contract rather than reasoning from their code.

mssql-odbc/src/api/set_connect_attr.rs
Vahid-b: Fixed in 40c0f702 — with the caveat that this one is a deliberate divergence rather than a straight correction, so flagging it explicitly.

msodbcsql has the same defect. `SQL_ATTR_CONNECTION_TIMEOUT` and `SQL_ATTR_LOGIN_TIMEOUT` share a single arm, and the clamp posts `IDS_01_S02_05` for either (`sqlcmisc.cpp:1733-1741`):

```cpp
case SQL_ATTR_CONNECTION_TIMEOUT:
case SQL_ATTR_LOGIN_TIMEOUT:
    if (vParam > MAX_QUERY_TIMEOUT)
    {
        vParam = MAX_QUERY_TIMEOUT;
        PostSQLError(lpdbc, IDS_01_S02_05);
```

and `local.rc:45` defines it as:

```
IDS_01_S02_05   "Login timeout changed"
```

So our text was byte-identical to the driver we're replacing, wrong noun included. Given that, I checked before changing it rather than assuming — the call was to correct it, on the grounds that the SQLSTATE is what applications branch on and it is unchanged, while the text is human-facing only.

Added `WARN_CONNECTION_TIMEOUT_CHANGED` ("Connection timeout changed"), used it in that arm, and documented the divergence on the constant so nobody later "restores parity" without knowing it was deliberate. `connection_timeout_above_maximum_warns_about_the_connection_timeout` pins it, including a negative assertion that the login-timeout wording is not reused.

Vahid-b:
@saurabh500 Done — merged `origin/main` into the branch in ae97302c. Clean merge, no conflicts; the branch was two commits behind and has picked up #167, so the Windows ODBC C++ e2e coverage now flows into this PR's diff coverage.

Head is now 40c0f702.

David-Engel (approved):

PR #146, opened 2026-07-22
odbc: typed SQLGetData conversion core (P1) [AB#46578]

David-Engel:
## Summary

Implements **P1 — typed `SQLGetData`** for the mssql-odbc driver (User Story 46375, Task [AB#46578](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46578)), building the shared `ColumnValues → SQL_C_*` conversion core that `SQLBindCol` block fetch will reuse in P3.

> Merged with current `main` after #153 landed; the typed conversions are now layered on its cursor-based fetch.

## What's included

- **New `api/fetch_convert.rs`** — the value-level conversion core, decoupled from the statement handle/diagnostics: it writes the converted value and reports outcomes via `ConvError { OutOfRange (22003), Unsupported (HYC00) }`, leaving SQLSTATE posting to the caller.
- **Integer targets**: `tinyint`/`smallint`/`int`/`bigint`/`bit` → `SQL_C_TINYINT`..`SQL_C_SBIGINT`, `SQL_C_BIT` (signed + unsigned), range-checked (`22003` on overflow; bit accepts only 0/1).
- **Floating point**: numeric sources → `SQL_C_FLOAT` / `SQL_C_DOUBLE`.
- **GUID**: `uniqueidentifier` → `SQL_C_GUID` (`SQLGUID` via `Uuid::as_fields`).
- **Date/time**: `date`/`time`/`datetime`/`datetime2`/`datetimeoffset`/`smalldatetime` → `SQL_C_TYPE_DATE`/`TIME`/`TIMESTAMP`, `SQL_C_SS_TIME2`, `SQL_C_SS_TIMESTAMPOFFSET`. Uses a Hinnant civil-from-days helper (anchors verified against proleptic Gregorian ordinals) plus 1/300s-tick (`datetime`) and minute (`smalldatetime`) epoch handling.
- **Character targets** (`SQL_C_CHAR`/`SQL_C_WCHAR`): every scalar type via text — integers, floats, decimal/numeric, money/smallmoney (exact, no `f64` round-trip), xml, json, vector, and the date/time family as ISO-ish text.
- **Binary** (`SQL_C_BINARY`): binary/xml/json.
- **Truncation reporting**: `01004` when a value does not fit the caller's buffer, and `01S07` (fractional truncation) when a date/time target discards a component (`SQL_C_TYPE_DATE` dropping a non-zero time, `SQL_C_TYPE_TIME` dropping a fraction).
- **Character rendering pads to the column's declared scale** rather than trimming trailing zeros, so every row of a column renders the same width — matching msodbcsql, which applications rely on when sizing buffers from `SQLDescribeCol`.
- A `NULL` column reports `SQL_NULL_DATA` for any target (covers the `sql_variant` `SQL_C_BINARY` NULL probe).
- **Malformed payloads no longer panic.** `SqlXml::as_string`, `SqlJson::as_string` and the UTF-8 branch of `SqlString::to_utf8_string` unwrap (and `SqlXml` indexes `chunk[1]` of a `chunks(2)` iterator), so bad server bytes panicked while the statement mutex was held — poisoning it and bricking the handle. Character payloads are now decoded fallibly in the ODBC layer and invalid data reports SQLSTATE **22018** with `SQL_ERROR`, leaving the handle usable. The panicking helpers in `mssql-tds` are left for a separate fix since they're shared with other consumers.

## Scope / deferrals

- Full `sql_variant` underlying-type resolution depends on `SQLColAttributeW` (`SQL_CA_SS_VARIANT_TYPE`) and lands in **P2**. The NULL-detection half of the probe works here.
- `SQL_C_NUMERIC` output struct is not emitted; decimal/numeric is delivered as character text, which is what `mssql-python` requests.
- **Mandatory source-type conversions are tracked as P1a ([AB#47107](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47107)).** Appendix D requires conversions to all C types from every supported SQL type; `decimal`/`money`/character sources into numeric and date/time C targets are not implemented yet and return `HYC00`. The `ConvOk::Truncated` / `01S07` plumbing they need is already in place.
- E2e tests (`set_stmt_attr_test.cpp` + a live typed-`SQLGetData` test) are folded into this phase's follow-up per the P0 decision; they require a live SQL connection and land with the block-fetch path.
- **Fetch mechanics belong to #153 (now merged).** That PR replaced the fetch model: rows are no longer materialized, `SQLGetData` resumes a column cursor (`read_row_column` → `CursorColumn`) and streams PLP values (`read_active_plp_chunk`). This PR is layered on top of it: `write_captured_column` dispatches the typed C targets through the conversion core, replacing the `"Target type not yet implemented"` HYC00 guard, while the character targets keep #153's chunking and `partial_text_offset` resume. PLP columns are unaffected — they stream as bytes or text and never reach a typed target.

## Testing

491 unit tests pass (full mssql-odbc suite); `clippy --all-targets` and `fmt` clean. New coverage includes the conversion core (integers/floats/guid/date-time round-trips, range errors), and `SQLGetData` integration tests for binary full/chunked streaming, `SQL_NO_DATA` exhaustion, character chunking, timestamp struct, decimal/vector text, NULL handling, offset reset across a row change, and malformed xml/json/utf-8 payloads (including one asserting the statement still serves a subsequent successful `SQLGetData`).


mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,952 @@
...
+        ($ty:ty) => {{ <$ty>::try_from(v).map_err(|_| ConvError::OutOfRange)? }};
+    }
+
+    let ret = match target_type {
+        // SQL_C_TINYINT is signed per the ODBC C type mapping (same as
+        // SQL_C_STINYINT); SQL_C_UTINYINT is the unsigned form.
+        SQL_C_STINYINT | SQL_C_TINYINT => unsafe {
+            write_fixed(target_value_ptr, narrow!(i8), strlen_or_ind_ptr)
+        },
+        SQL_C_UTINYINT => unsafe { write_fixed(target_value_ptr, narrow!(u8), strlen_or_ind_ptr) },
```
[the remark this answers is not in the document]
David-Engel: Fixed in 23c6a2f3: SQL_C_TINYINT now maps to unsigned u8 (only SQL_C_STINYINT stays signed i8), so tinyint values 128-255 round-trip instead of failing with 22003. Added a >127 unsigned round-trip regression test plus a signed-target reject test.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,952 @@
...
+    strlen_or_ind_ptr: *mut SqlLen,
+) -> Result<SqlReturn, ConvError> {
+    let Some(v) = numeric_source_as_f64(value) else {
+        return Err(ConvError::Unsupported);
+    };
+    let ret = match target_type {
+        // SQL_C_FLOAT is 32-bit; the narrowing follows the ODBC contract (the
+        // driver may lose precision, mirroring the C cast an application would
+        // perform itself).
+        SQL_C_FLOAT => unsafe { write_fixed(target_value_ptr, v as f32, strlen_or_ind_ptr) },
```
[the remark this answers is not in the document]
David-Engel: Fixed in 23c6a2f3: convert_float_c now range-checks a finite f64 against the f32 range and returns ConvError::OutOfRange (22003) instead of writing infinity; an already-infinite source still passes through. Added an overflow regression test.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,952 @@
...
+//! and, later, `SQLBindCol` block fetch (P3). It is deliberately free of any
+//! statement-handle or diagnostic-list coupling: it writes the converted value
+//! into the caller's buffer and reports success/failure through [`ConvError`],
+//! leaving the SQLSTATE posting to the caller (whose diagnostic target differs
+//! between the two fetch paths).
+//!
+//! Scope so far: the fixed-width integer C targets from an integer source
+//! column (`tinyint` / `smallint` / `int` / `bigint` / `bit`). Unhandled
+//! source/target pairs return [`ConvError::Unsupported`] so callers can fall
+//! back to their existing paths (e.g. the character conversion in `get_data`).
```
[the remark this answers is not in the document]
David-Engel: Fixed in 23c6a2f3: the module doc now lists the floating-point, GUID, and date/time conversions in addition to integers.

mssql-odbc/src/api/get_data.rs
```diff
@@ -223,17 +433,47 @@ fn column_value_to_text(v: &ColumnValues) -> Option<String> {
...
         ColumnValues::Float(x) => Some(x.to_string()),
         ColumnValues::Bit(x) => Some(if *x { "1".into() } else { "0".into() }),
+        ColumnValues::Decimal(d) | ColumnValues::Numeric(d) => Some(d.to_string()),
+        ColumnValues::Money(m) => Some(money_scaled_to_string(
+            (i64::from(m.lsb_part) & 0xFFFF_FFFF) | (i64::from(m.msb_part) << 32),
+        )),
+        ColumnValues::SmallMoney(m) => Some(money_scaled_to_string(i64::from(m.int_val))),
         ColumnValues::String(s) => Some(s.to_utf8_string()),
+        ColumnValues::Xml(x) => Some(x.as_string()),
+        ColumnValues::Json(j) => Some(j.as_string()),
```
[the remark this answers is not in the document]
David-Engel: Good catch — confirmed all three panic: `SqlXml::as_string` indexes `chunk[1]` on a `chunks(2)` iterator (so an odd byte count panics) and unwraps `from_utf16`; `SqlJson::as_string` and the UTF-8 branch of `SqlString::to_utf8_string` unwrap `from_utf8`. Because these ran while the statement mutex was held, the panic was caught by the FFI panic boundary but left the mutex poisoned, so every subsequent call on that handle failed — the statement was effectively bricked.

Fixed in 4486d77f by decoding fallibly in the ODBC layer:

- `column_value_to_text` now returns `Result<String, TextError>`, distinguishing `Unsupported` (still HYC00) from `Malformed`.
- Malformed payloads post SQLSTATE **22018** (invalid character value for cast specification) and return `SQL_ERROR`, leaving the handle usable.
- `xml_to_text` requires a whole number of UTF-16 code units and uses `chunks_exact` + checked `from_utf16`; JSON and the `SqlString` UTF-8 branch use checked `from_utf8`. The UTF-16/LCID `SqlString` branches already decode through `encoding_rs`, which substitutes replacement characters rather than failing, so they are unchanged.

Added tests for malformed xml / json / utf-8 string, including one that asserts the statement still serves a subsequent successful `SQLGetData` — i.e. the mutex-poisoning regression is gone.

I deliberately left the panicking helpers in `mssql-tds` as-is to keep this PR scoped; they're shared with other consumers (mssql-py-core) and warrant their own fix.

mssql-odbc/src/api/get_data.rs
```diff
@@ -121,119 +146,379 @@ fn sql_get_data_safe(
...
 
-    if target_type != SQL_C_CHAR && target_type != SQL_C_WCHAR {
-        post_sql_error(
-            &mut stmt_state,
-            SQLSTATE_HYC00,
-            0,
-            "Target type not yet implemented",
+    // A repeated SQLGetData on a column whose value was already fully returned
+    // yields SQL_NO_DATA (the streaming cursor is exhausted).
+    if stmt_state.getdata_col == column_number && stmt_state.getdata_offset == usize::MAX {
```
shiwanigupta0809: This PR overlaps with https://github.com/microsoft/mssql-rs/pull/153.
#153 already handles the SQL_NO_DATA case and these other checks and uses ODBC style wire-stream state instead of an offset sentinel.

saurabh500:
Flagging a sequencing issue before this goes further, and proposing a way for it to make progress independently.

#153 (@shiwanigupta0809) reworks the same `SQLGetData` path from the other end: `SQLFetch` stops materializing rows, and columns are pulled on demand through a cursor, with PLP values streaming straight off the wire. That deletes `StmtState::current_row` — the field this PR's `set_current_row` exists to wrap — and rewrites the body of `sql_get_data_safe`. So the two PRs collide in `handles/stmt.rs`, `api/get_data.rs`, and the seven one-line `current_row = None` call sites.

The good news is the collision is confined to the plumbing. **Roughly half of this PR is untouched by #153.**

### Independent of #153 — can land now

- **`api/fetch_convert.rs`** (+1006). #153 never creates or touches this file. It takes `&ColumnValues` plus a target type and writes to a raw pointer; it has no `StmtState` coupling by design (as the module doc says). Under #153 it feeds off `last_captured: Option<(usize, ColumnValues)>` exactly as well as it does off `current_row` today. All the unit tests transfer verbatim.
- **`api/sqlstate.rs`** (+10) — purely additive, #153 doesn't touch the file.
- **`Cargo.toml`** uuid dev-dep.

### Independent of both PRs — arguably belongs upstream

The fallible character decode (`TextError`, `sql_string_to_text`, `xml_to_text`, the JSON branch) fixes three real panics: `SqlXml::as_string` indexes `chunk[1]` on a `chunks(2)` remainder and `.unwrap()`s `from_utf16`, and both `SqlString::to_utf8_string` (UTF-8 branch) and `SqlJson::as_string` `.unwrap()` a `from_utf8`. Any of those unwinding through the held `MutexGuard` poisons `stmt.inner`, after which every call on the handle returns `SQL_ERROR` with no diag record.

#153 inherits all three — its `column_value_to_text` still calls `s.to_utf8_string()`. And `mssql-js` / `mssql-py-core` call the same helpers. So this is worth doing once as `try_as_string`-style APIs in `mssql-tds` rather than as a local workaround in the ODBC layer, and neither PR should have to own it.

### Genuinely conflicting — but still valuable

The chunked-offset streaming (`getdata_col` / `getdata_offset`, `stream_text`, `stream_binary`, `finish_stream`) is sited on the materialized-row model that #153 removes. But it shouldn't be thrown away: #153 currently treats non-PLP `SQLGetData` as single-shot, so a truncated read marks the column consumed and the next call returns `SQL_NO_DATA` — probe-then-fetch and chunk-loop patterns lose data (raised on `get_data.rs:313` there). This PR already implements the resumable-offset fix, including the `SQL_C_WCHAR` code-unit-vs-byte distinction and `01004` on truncation, with tests. Re-siting it onto `last_captured` is a port, not a redesign.

### Suggested split

1. **This PR, reduced to `fetch_convert.rs` + `sqlstate.rs` + `Cargo.toml`.** Merge independently of #153. One wrinkle: with no caller, `-D warnings` will fail on `dead_code`, so it needs a minimal graft into today's `get_data.rs` routing typed targets only — no streaming, no `StmtState` changes. That graft is a small conflict for #153 to resolve instead of 874 lines.
2. **The fallible decode**, as a separate `mssql-tds` PR. Unblocks #153 and the other bindings too.
3. **The streaming offset**, folded into #153 or a follow-up on top of it, since it's the fix for the truncation issue there.

That way #153 lands the row-model change without racing this one, and @David-Engel isn't blocked behind it.

Separately — my question on `SQL_SS_TIMESTAMPOFFSET` (whether the struct fields should carry local time rather than the wire's UTC) lives in `fetch_convert.rs`, so it needs an answer regardless of which PR ends up carrying that file.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1006 @@
...
+fn hms_from_nanos(nanos: u64) -> (u16, u16, u16, u32) {
+    let secs = nanos / 1_000_000_000;
+    let fraction_ns = (nanos % 1_000_000_000) as u32;
+    (
+        (secs / 3600) as u16,
+        ((secs % 3600) / 60) as u16,
+        (secs % 60) as u16,
+        fraction_ns,
+    )
+}
```
saurabh500: 🔴 **Blocking — `time_nanoseconds` is not nanoseconds; every `time`/`datetime2`/`datetimeoffset` value is 100× too small**

`mssql-tds/src/datatypes/decoder.rs` documents its own contract:

> *"We need to convert to **100-nanosecond units** for consistency"*

…and multiplies each scale bucket up to 100 ns ticks (`0 => scaled_value * 10_000_000`, … `7 => scaled_value`). The field name is a misnomer — it holds **100 ns ticks**, not nanoseconds. Dividing by `1_000_000_000` here under-scales by exactly 100.

**Verified against `msodbcsql18` on SQL Server 2022** (same query, same buffer, `SQL_OV_ODBC3_80`):

| Query | Target | msodbcsql18 | this PR |
|---|---|---|---|
| `CAST('12:34:56.1234567' AS time(7))` | `SQL_C_SS_TIME2` | `12:34:56.123456700` | `00:07:32.961234567` |
| same | `SQL_C_CHAR` | `"12:34:56.1234567"` | `"00:07:32.9612345"` |
| `CAST('12:34:56' AS time(0))` | `SQL_C_SS_TIME2` | `12:34:56.000000000` | `00:07:32.960000000` |
| `CAST('2023-06-15 12:34:56.1234567' AS datetime2(7))` | `SQL_C_TYPE_TIMESTAMP` | `…12:34:56.123456700` | `…00:07:32.961234567` |

Check: `12:34:56.1234567` = 45 296.1234567 s. Observed `00:07:32.961234567` = 452.961… s = 45 296.1234567 / 100. ✔

**Fix**

```rust
/// (hour, minute, second, fraction_ns) from 100-nanosecond ticks since midnight.
fn hms_from_ticks_100ns(ticks: u64) -> (u16, u16, u16, u32) {
    let secs = ticks / 10_000_000;
    let fraction_ns = ((ticks % 10_000_000) * 100) as u32;
    (
        (secs / 3600) as u16,
        ((secs % 3600) / 60) as u16,
        (secs % 60) as u16,
        fraction_ns,
    )
}
```

Rename the function and both call sites (`:297`, `:306`, `:319`) so the unit is legible. Worth filing a follow-up in `mssql-tds` to rename `SqlTime::time_nanoseconds` → `time_ticks_100ns`; the name is what caused this, and `mssql-js`/`mssql-py-core` read the same field.

The `ColumnValues::DateTime` arm at `:333` is **unaffected** — it uses 1/300 s ticks in a separate code path, which is why `datetime` and `smalldatetime` matched msodbcsql exactly in the probe. Only the `hms_from_nanos` callers are wrong.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1006 @@
...
+            p.hour = h;
+            p.minute = mi;
+            p.second = s;
+            p.fraction_ns = f;
+            p.tz_hour = dto.offset / 60;
+            p.tz_minute = dto.offset % 60;
+            p.has_date = true;
+            p.has_time = true;
+            p.has_tz = true;
+        }
```
saurabh500, line 386: 🔴 **Blocking — `datetimeoffset` returns the UTC wall clock in fields that must hold local time**

`SqlDateTimeOffset.datetime2` holds **UTC**; `offset` (minutes) is what you *add* to reach the local wall clock that was originally written. This arm copies the UTC fields straight through and only populates `tz_hour`/`tz_minute`, so the struct claims "12:34 in +05:30" while actually carrying 07:04 — wrong wall clock *and* wrong instant, off by 2× the offset.

Three independent confirmations:

1. **msodbcsql18, measured.** `SELECT CAST('2023-01-01 12:34:56.1234567 +05:30' AS datetimeoffset(7))` → `SQL_C_SS_TIMESTAMPOFFSET` returns `2023-01-01 12:34:56.123456700 +05:30` — **local**. Negative offsets likewise: `'2023-06-15 12:00:00 -05:00'` → `12:00:00 -05:00`.
2. **This repo already encodes the semantics.** `mssql-js/__test__/datetime_conversion.spec.mjs` expects `'2023-01-01T12:34:56.1234567+05:30'` → `new Date('2023-01-01T07:04:56.123Z')`, i.e. local minus 5:30 = UTC.
3. **T-SQL docs:** `datetimeoffset` is *"processed, compared, sorted, and indexed in the server as in UTC."*

This also flows into `format_datetime_parts`, so `SQL_C_CHAR` is wrong by the same amount.

**Fix** — add the offset before splitting into fields, and carry the day rollover:

```rust
ColumnValues::DateTimeOffset(dto) => {
    let ticks = dto.datetime2.time.time_nanoseconds as i64
        + i64::from(dto.offset) * 60 * 10_000_000;
    const TICKS_PER_DAY: i64 = 864_000_000_000;
    let (day_adj, ticks) = (ticks.div_euclid(TICKS_PER_DAY), ticks.rem_euclid(TICKS_PER_DAY));
    let (y, m, day) = civil_from_days_since_0001(i64::from(dto.datetime2.days) + day_adj);
    let (h, mi, s, f) = hms_from_ticks_100ns(ticks as u64);
    // …
}
```

`div_euclid`/`rem_euclid` are what make negative offsets correct — plain `/` and `%` truncate toward zero and will produce a negative time-of-day for e.g. `00:30 -05:00`.

Day rollover must be covered both directions: `'2023-06-15 23:30:00 +05:30'` (msodbcsql returns `2023-06-15 23:30:00 +05:30`; this PR returns `00:10:48`) and an early-morning value with a negative offset. The `0001-01-01` / `9999-12-31` boundaries can overflow the civil date after adjustment — worth a guard.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1006 @@
...
+
+/// Why a value-level conversion could not be completed. The caller maps each
+/// variant to the appropriate SQLSTATE on its own diagnostic target.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum ConvError {
+    /// The value does not fit the requested C type (SQLSTATE `22003`).
+    OutOfRange,
+    /// This source/target pairing is not handled here (SQLSTATE `HYC00`), and
+    /// the caller should try another path or report "not implemented".
+    Unsupported,
```
saurabh500: 🟡 **`Unsupported` maps to the wrong SQLSTATE for ODBC SQL types**

The ODBC spec reserves `HYC00` ("Optional feature not implemented") for conversions from a **driver-specific** SQL type. When the source is a standard ODBC SQL type and the driver does not support the requested C target, the required state is **`07006` Restricted data type attribute violation**.

**msodbcsql18, measured:** `SELECT CAST(42 AS int)` → `SQL_C_TYPE_TIMESTAMP` returns `SQL_ERROR` / **`07006`**. This PR returns `HYC00`. Same for `varchar 'not-a-date'` → `SQL_C_TYPE_DATE`, where msodbcsql returns **`22018`**.

Applications branch on SQLSTATE — several drivers/ORMs treat `HYC00` as "driver too old, fall back" and `07006` as "this cast is illegal, surface to the user."

**Fix** — the enum currently conflates two meanings: "not handled *here*, try another path" (internal routing) and "not convertible at all" (terminal, user-visible). Split them:

```rust
pub(crate) enum ConvError {
    /// The value does not fit the requested C type (SQLSTATE `22003`).
    OutOfRange,
    /// This source/target pairing is not handled by this converter; the caller
    /// should try another path. Never surfaced to the application directly.
    NotHandledHere,
    /// The requested C type is not a legal target for this SQL type
    /// (SQLSTATE `07006`). Terminal.
    Restricted,
}
```

Then add `SQLSTATE_07006` + an `ERR_RESTRICTED_DATA_TYPE` `DiagMsg` in `sqlstate.rs`, alongside the `22003`/`22018` entries this PR already adds, and have `get_data.rs` only fall through to "not implemented" for `NotHandledHere`.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1006 @@
...
+fn integer_source_as_i128(value: &ColumnValues) -> Option<i128> {
+    match value {
+        ColumnValues::TinyInt(x) => Some(i128::from(*x)),
+        ColumnValues::SmallInt(x) => Some(i128::from(*x)),
+        ColumnValues::Int(x) => Some(i128::from(*x)),
+        ColumnValues::BigInt(x) => Some(i128::from(*x)),
+        ColumnValues::Bit(b) => Some(i128::from(*b)),
+        _ => None,
+    }
+}
```
saurabh500, line 80: 🟡 **Missing mandatory source types — `decimal`, `numeric`, `money`, `smallmoney`, `real`, `float`, and character sources all fail**

ODBC Appendix D: *"drivers are required to support conversions to **all** ODBC C data types from the ODBC SQL data types that they support."* This helper rejects every non-integer source, and `numeric_source_as_f64` (`:153`) rejects the exact decimal types. Neither accepts `ColumnValues::String`.

**msodbcsql18 vs. this PR, measured:**

| Query | Target | msodbcsql18 | this PR |
|---|---|---|---|
| `CAST(12345.6789 AS decimal(18,4))` | `SQL_C_DOUBLE` | `12345.6789` | `SQL_ERROR` `HYC00` |
| `CAST(1234.5678 AS money)` | `SQL_C_DOUBLE` | `1234.5678` | `SQL_ERROR` `HYC00` |
| `CAST(1234.99 AS float)` | `SQL_C_SLONG` | `1234` + `01S07` | `SQL_ERROR` `HYC00` |
| `CAST('123' AS varchar(10))` | `SQL_C_SLONG` | `123` | `SQL_ERROR` `HYC00` |
| `CAST(12345.6789 AS decimal(18,4))` | `SQL_C_SBIGINT` | `12345` + `01S07` | `SQL_ERROR` `HYC00` |

**Fix** — extend both widening helpers:

- `integer_source_as_i128`: accept `Decimal`/`Numeric`/`Money`/`SmallMoney` (truncate toward zero, and report `01S07` when a fractional part is dropped), `Real`/`Float` (truncate, `01S07`), and `String` (parse; `22018` on failure, `22003` on overflow).
- `numeric_source_as_f64`: accept `Decimal`/`Numeric`/`Money`/`SmallMoney` and `String`.

If these are deliberately sequenced to a later "P" milestone, that is a fine answer — please say so in the PR description and in `docs/typed-columnar-fetch-plan.md`, because right now they silently regress to `HYC00`, which is also the wrong state (see the `ConvError` comment above).

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1006 @@
...
+            write_fixed(
+                target_value_ptr,
+                SqlTimeStruct {
+                    hour: p.hour,
+                    minute: p.minute,
+                    second: p.second,
+                },
+                strlen_or_ind_ptr,
+            )
+        },
```
saurabh500: 🟡 **Truncating conversions must return `SQL_SUCCESS_WITH_INFO` + `01S07` (Fractional truncation)**

Discarding the date half (`SQL_C_TYPE_TIME`) or the time half (`SQL_C_TYPE_DATE`) is a truncation, and ODBC requires it be reported.

**msodbcsql18, measured** on `CAST('2023-06-15 12:34:56.1234567' AS datetime2(7))`:

| Target | msodbcsql18 | this PR |
|---|---|---|
| `SQL_C_TYPE_TIME` | `12:34:56`, **`SQL_SUCCESS_WITH_INFO` / `01S07`** | value only, **no diagnostic** |
| `SQL_C_TYPE_DATE` | `2023-06-15`, **`SQL_SUCCESS_WITH_INFO` / `01S07`** | value only, **no diagnostic** |

Note msodbcsql emits `01S07` for `SQL_C_TYPE_DATE` even though only *time* is dropped — it treats any lost temporal precision as fractional truncation. Numeric truncation behaves the same way (`float 1234.99` → `SQL_C_SLONG` gives `1234` + `01S07`), so this pairs with the missing-conversions comment above.

**Fix** — the current signature can't express this: `SqlReturn` alone carries no diagnostic, and `Result<(), ConvError>` has only two outcomes. Add a third:

```rust
pub(crate) enum ConvOk {
    Exact,
    Truncated,   // caller posts 01S07 and returns SQL_SUCCESS_WITH_INFO
}
// -> Result<ConvOk, ConvError>
```

`SQL_C_TYPE_DATE` should set `Truncated` whenever `has_time && (hour|minute|second|fraction_ns) != 0`; `SQL_C_TYPE_TIME` whenever `fraction_ns != 0`. Add `SQLSTATE_01S07` + `WARN_FRACTIONAL_TRUNCATION` to `sqlstate.rs`.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1006 @@
...
+        if p.fraction_ns != 0 {
+            // Render in 100 ns units (7 digits) and trim trailing zeros.
+            let hundred_ns = p.fraction_ns / 100;
+            let frac = format!("{hundred_ns:07}");
+            let frac = frac.trim_end_matches('0');
+            if !frac.is_empty() {
+                s.push('.');
+                s.push_str(frac);
+            }
+        }
```
saurabh500: 🟡 **Character rendering must pad to the column's declared scale, not trim trailing zeros**

Trimming produces a variable-length string. msodbcsql pads to the column scale and returns a fixed length for every row of that column — applications size buffers from `SQLDescribeCol` and compare fixed-width strings.

**msodbcsql18 vs. this PR, measured** (`SQL_C_CHAR`):

| Query | msodbcsql18 | this PR |
|---|---|---|
| `CAST('2023-06-15 12:00:00' AS datetime2(7))` | `"2023-06-15 12:00:00.0000000"` len **27** | `"2023-06-15 00:07:12"` len 19 |
| `CAST('2023-06-15 12:00:00.100' AS datetime2(3))` | `"2023-06-15 12:00:00.100"` len 23 | `"2023-06-15 00:07:12.001"` len 23 |
| `CAST('2023-06-15 12:00:00' AS datetime)` | `"2023-06-15 12:00:00.000"` len **23** | `"2023-06-15 12:00:00"` len 19 |
| `CAST('12:34:56.1234567' AS time(7))` | `"12:34:56.1234567"` len 16 | `"00:07:32.9612345"` len 16 |

(The wrong time values are the `hms_from_nanos` bug; the differing **lengths** are this issue. `smalldatetime` matched exactly — scale 0, nothing to pad.)

The ODBC core spec permits omitting a zero fraction, so this is a **parity** break rather than a spec break — but parity with msodbcsql is the bar this driver is held to.

**Fix** — `DateTimeParts` has no scale field, so this needs a small structural change. `SqlTime.scale` is already available on the source (`datetime` is always 3, `smalldatetime` 0, `date` has no time part):

```rust
pub(crate) struct DateTimeParts {
    // …
    pub scale: u8,
}
// in format_datetime_parts:
if p.scale > 0 {
    let hundred_ns = p.fraction_ns / 100;
    s.push_str(&format!(".{:07}", hundred_ns)[..=p.scale as usize]);
}
```

Populate `scale` in each `extract_datetime_parts` arm: `t.scale` / `dt.time.scale` / `dto.datetime2.time.scale`, `3` for `DateTime`, `0` for `SmallDateTime`.

Two things here are already **correct** per the msodbcsql docs and confirmed by the probe — please keep them: the space before the timezone offset, and the signed `+` for a zero offset.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1006 @@
...
+        .unwrap();
+        assert_eq!(
+            out,
+            SqlSsTime2Struct {
+                hour: 13,
+                minute: 45,
+                second: 30,
+                fraction: 123_456_700
+            }
+        );
```
saurabh500, line 961: 🟡 **These tests construct the input in the wrong unit, so they can't catch the 100× scaling bug**

`time_nanoseconds` holds **100 ns ticks** (see `decoder.rs`), but this builds the value as `secs * 1_000_000_000 + frac` — nanoseconds. The test then asserts against `hms_from_nanos`, which divides by `1e9`. Input and assertion are wrong in exactly compensating ways, so the test passes while the driver returns `00:07:32.961234567` for `12:34:56.1234567` against a real server.

`datetime2_to_timestamp_struct` at `:888` has the same shape.

**Fix** — build the input in ticks and keep the expected output as the real wall clock:

```rust
// 13:45:30.1234567 in 100 ns ticks since midnight
let ticks = ((13 * 3600 + 45 * 60 + 30) as u64) * 10_000_000 + 1_234_567;
// … expect SqlSsTime2Struct { hour: 13, minute: 45, second: 30, fraction: 123_456_700 }
```

Better still, add one case that decodes a real TDS `time(7)` payload through `mssql-tds`'s decoder rather than hand-constructing `SqlTime` — that is the only way this class of unit mismatch gets caught at the seam where it actually lives.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1006 @@
...
+                        },
+                    },
+                    offset: -330, // -05:30
+                }),
+                SQL_C_SS_TIMESTAMPOFFSET,
+                (&mut out as *mut SqlSsTimestampoffsetStruct).cast(),
+                &mut ind,
+            )
+        }
+        .unwrap();
```
saurabh500, line 1022: 🟡 **This test is blind to the `datetimeoffset` bug — it pins `time_nanoseconds: 0` and never asserts the time**

With a zero time-of-day and assertions limited to year/month/day/tz_hour/tz_minute, this passes identically whether the offset is applied or not. It is the reason the UTC-vs-local defect at `:317` reached a green build.

**Fix** — assert the full wall clock, with a nonzero time and both offset signs, using values measured from `msodbcsql18`:

```rust
// SELECT CAST('2023-01-01 12:34:56.1234567 +05:30' AS datetimeoffset(7))
// msodbcsql18 -> 2023-01-01 12:34:56.123456700 +05:30
// UTC as stored: 2023-01-01 07:04:56.1234567, offset = +330
assert_eq!((out.hour, out.minute, out.second), (12, 34, 56));
assert_eq!(out.fraction, 123_456_700);
assert_eq!((out.timezone_hour, out.timezone_minute), (5, 30));
```

Then add the two cases that only a correct implementation survives:

- **Positive rollover:** `'2023-06-15 23:30:00 +05:30'` — msodbcsql returns `2023-06-15 23:30:00 +05:30`; the date must not advance to the 16th. (This PR currently returns `00:10:48`.)
- **Negative rollover:** an early-morning local time with a negative offset, where UTC + offset goes below zero — this is what `rem_euclid` protects and plain `%` breaks.

saurabh500 (commented):
Inline comments below, all grounded in a side-by-side run against `msodbcsql18`.

**Method.** I built a raw-FFI ODBC probe (no Rust ODBC crate in the way), registered this PR's `msodbcsql18.dll` as a second driver, and ran 22 identical `SQLGetData` cases through both drivers against a local SQL Server 2022 — same SQL, same target C types, same buffers, `SQL_ATTR_ODBC_VERSION = SQL_OV_ODBC3_80`. Every finding below cites what msodbcsql actually returned rather than what the spec says it should.

**Headline:** the probe surfaced a blocker I would not have found by reading alone — `SqlTime::time_nanoseconds` holds **100-nanosecond ticks**, not nanoseconds, so `hms_from_nanos` under-scales every `time`, `datetime2`, and `datetimeoffset` value by exactly 100×. `12:34:56.1234567` comes back as `00:07:32.961234567`. The existing unit tests can't catch it because they construct the input in nanoseconds too — the bug and its tests are self-consistently wrong.

Summary of the 22 cases:

- 🔴 **2 blockers** — the 100× time scaling, and `datetimeoffset` returning UTC in the local-time fields (off by 2× the offset).
- 🟡 **4 parity gaps** — `HYC00` where msodbcsql returns `07006`/`22018`; missing mandatory conversions (`decimal`/`money`/`float`/character → `SQL_C_DOUBLE`/`SQL_C_SLONG`); no `01S07` on truncating conversions; character rendering trims trailing zeros where msodbcsql pads to the column scale.
- 🟡 **2 test gaps** that let the above through a green build.
- ✅ **Matching already:** `22003` range checking on both integer and float overflow, `datetime` and `smalldatetime` rendering, and the `SQLGUID` / `SQL_C_TINYINT` / timezone-format choices. The overall factoring of a shared SQL→C conversion core is the right shape — these are corrections within it, not a redesign.

Happy to share the probe harness if it is useful for building out the e2e parity suite.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Agreed — fixed in 5de38575. The enum genuinely conflated internal routing with a terminal, user-visible error, and I've split it as you suggested:

```rust
pub(crate) enum ConvError {
    OutOfRange,      // 22003
    NotHandledHere,  // internal routing; never surfaced directly
    Restricted,      // 07006, terminal
}
```

`Restricted` now covers the illegal pairings — a non-temporal column into a date/time C target (your `CAST(42 AS int)` → `SQL_C_TYPE_TIMESTAMP` case), a non-`uniqueidentifier` into `SQL_C_GUID`, and a value that lacks the component the target needs (e.g. a `time` column into `SQL_C_TYPE_DATE`). `get_data.rs` posts `ERR_RESTRICTED_DATA_TYPE` for those. `SQLSTATE_07006` and the `DiagMsg` already existed in `sqlstate.rs`, so no new constants were needed.

One deliberate distinction: `NotHandledHere` still surfaces as `HYC00`. That path is now reached only for conversions that are *required but not yet implemented* (e.g. `varchar '123'` → `SQL_C_SLONG`), where "optional feature not implemented" is the more honest answer than claiming the cast is illegal. Those pairings are tracked as **P1a / AB#47107** (see the reply on your Appendix D comment); once implemented, that path disappears rather than changing SQLSTATE.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Agreed — this is a real Appendix D gap, and your measurements match what the code does today.

Rather than expand this PR (which is already ~1.1k lines of conversion core and currently green), David and I have split it into its own task: **AB#47107 — "mssql-odbc | P1a: mandatory source-type conversions for SQLGetData (decimal/money/character sources)"**, in the current sprint. It captures exactly your table:

- `decimal`/`numeric` → numeric C targets (`numeric_source_as_f64` rejects the exact-decimal types today)
- `money`/`smallmoney` → numeric C targets
- character sources → numeric and date/time C targets (`'123'` → `SQL_C_SLONG`, `'2023-06-15'` → `SQL_C_TYPE_DATE`), with `22018` when the text isn't a valid literal
- `01S07` + `SQL_SUCCESS_WITH_INFO` for fractional truncation on lossy numeric conversions (needs a new `SQLSTATE_01S07`)

I've also documented it as a P1a section in `mssql-odbc/docs/typed-columnar-fetch-plan.md`, so the gap is explicit rather than implied. Until it lands these pairings return `HYC00` (not implemented) rather than silently doing the wrong thing.

Happy to fold it into this PR instead if you'd prefer it not ship incomplete — just say the word.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Confirmed and fixed in 5de38575.

I checked the semantics from two directions before changing anything: the decoder reads the `datetime2` component straight off the wire and attaches `offset` without adjusting, and `mssql-js/__test__/datetime_conversion.spec.mjs` expects `'2023-01-01T12:34:56.1234567+05:30'` → `new Date('2023-01-01T07:04:56.123Z')`. So the stored component is UTC and the local wall clock is UTC + offset — the struct was claiming a timezone while carrying UTC, exactly as you described.

The arm now adds the offset before splitting into fields, using `div_euclid`/`rem_euclid` for the day rollover so a negative offset borrows a day instead of producing a negative time-of-day. I also added the range guard you flagged: if the adjustment leaves `0001-01-01..9999-12-31` the conversion is rejected rather than producing a bogus civil date.

Tests cover both directions plus the boundary:
- negative offset borrowing a day (UTC `2000-02-29 00:00 -05:30` → local `2000-02-28 18:30`),
- positive offset rolling forward (UTC `2023-06-15 22:00 +05:30` → local `2023-06-16 03:30`),
- `0001-01-01 00:00` with a negative offset → rejected.

This flows through `format_datetime_parts`, so the `SQL_C_CHAR` rendering is corrected by the same change.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Confirmed and fixed in 5de38575 — thank you for the measured comparison, this was a real 100× error.

I verified the contract independently: `decoder.rs` scales every bucket to 100 ns ticks (`0 => scaled_value * 10_000_000` … `7 => scaled_value`), and its own test asserts `time_nanoseconds == 10_000_000` for **one second**. So the field is 100 ns ticks and `hms_from_nanos` under-scaled by exactly 100, as you said.

Replaced it with `hms_from_ticks_100ns` (`ticks / 10_000_000`, `fraction_ns = (ticks % 10_000_000) * 100`) and renamed all three call sites so the unit is legible. Tests updated to construct values in ticks.

You're also right that the `ColumnValues::DateTime` arm is unaffected — it uses the separate 1/300 s tick path, which is why `datetime`/`smalldatetime` matched in your probe. I left it alone.

Agreed on the follow-up to rename `SqlTime::time_nanoseconds` → `time_ticks_100ns` in `mssql-tds`; the name is exactly what caused this, and `mssql-js`/`mssql-py-core` read the same field. That's a cross-crate rename so I've kept it out of this PR.

mssql-odbc/src/api/get_data.rs
David-Engel: Good catch, and agreed — I've ceded the overlap to #153 in 5de38575.

Comparing the two: #153 owns the fetch *mechanics* (column-wise fetch, incremental PLP, `SQL_NO_DATA` semantics) with real ODBC wire-stream state, while this PR is the typed `ColumnValues → SQL_C_*` conversion matrix (`fetch_convert.rs`); #153's `get_data.rs` only references `SQL_C_CHAR`/`SQL_C_WCHAR`. Your wire-stream state is the better design here — my offset sentinel was explicitly the stopgap chosen when we deferred PLP, so it shouldn't land only to be ripped out.

Removed from this PR:
- `getdata_col` / `getdata_offset` on `StmtState` and the `SQL_NO_DATA` exhaustion check
- `stream_text` / `stream_binary` / `finish_stream` / `mark_column_done`
- the `set_current_row` refactor that existed only to reset that offset

Net effect: `stmt.rs`, `fetch.rs`, `close_cursor.rs`, `exec_direct.rs`, `execute.rs`, `more_results.rs`, `prepare.rs` and `get_type_info.rs` are now **byte-identical to main** in this branch, so the conflict surface with #153 is essentially just `get_data.rs`'s dispatch. This PR is down to 6 files.

`SQLGetData` here returns each value in one call and reports truncation with `01004`; chunked retrieval arrives with your PR. I've documented the boundary in `mssql-odbc/docs/typed-columnar-fetch-plan.md` so it's clear which PR owns what.

When #153 lands, the conversion core should slot in behind your chunk pipeline — happy to do that integration if it helps.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Correct, and this is the honest lesson from this review: the test pinned `time_nanoseconds: 0` and asserted only year/month/day plus the timezone fields, so it was structurally incapable of detecting whether the offset had been applied. It is exactly why the UTC-vs-local defect reached a green build.

Fixed in db7e1bcd. The suite now asserts the full wall clock, using your measured values:

- `datetimeoffset_matches_msodbcsql_wall_clock` — stored UTC `2023-01-01 07:04:56.1234567` with `offset = +330`, asserting `(12, 34, 56)`, `fraction == 123_456_700`, and `(+5, +30)`, matching `msodbcsql18`'s `2023-01-01 12:34:56.123456700 +05:30`.
- `datetimeoffset_positive_offset_rolls_day_forward` — UTC `2023-06-15 22:00 +05:30` → local `2023-06-16 03:30`.
- `datetimeoffset_negative_offset_rolls_day_backward` — UTC `2023-06-15 02:00 -05:00` → local `2023-06-14 21:00`, i.e. UTC + offset goes below zero. This is the case that `rem_euclid` protects and plain `%` breaks, as you noted.
- `datetimeoffset_out_of_range_after_offset_is_rejected` — `0001-01-01 00:00` with a negative offset, covering the boundary guard.

I confirmed these fail against the previous implementation rather than just passing against the new one.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: You're right, and this is the part I find most worth calling out: the tests were self-consistently wrong — input in nanoseconds, assertion against a converter that divided by 1e9 — so they passed while the driver was 100× off against a real server.

Fixed in 5de38575, using exactly the construction you suggested:

```rust
// 13:45:30.1234567 in 100 ns ticks since midnight
let ticks = ((13 * 3600 + 45 * 60 + 30) as u64) * 10_000_000 + 1_234_567;
// expects SqlSsTime2Struct { hour: 13, minute: 45, second: 30, fraction: 123_456_700 }
```

`datetime2_to_timestamp_struct` was rebuilt the same way, and the expected outputs are now real wall-clock values rather than derived from the same wrong assumption.

On your stronger suggestion — a case that decodes a real TDS `time(7)` payload through the `mssql-tds` decoder — I agree that's the only thing that catches this class of mismatch at the seam where it actually lives, since the bug was a unit contract between two crates. I haven't added it here because it belongs on the `mssql-tds` side of the boundary (and pairs naturally with the `time_nanoseconds` → `time_ticks_100ns` rename). Happy to add it as part of that follow-up, or in this PR if you'd rather it not wait.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Agreed — fixed in db7e1bcd, and thanks for calling out that this is a parity break rather than a spec break; that's the right bar.

Implemented as you outlined: `DateTimeParts` now carries `scale`, and `format_datetime_parts` pads to exactly that many digits instead of trimming:

```rust
if p.scale > 0 {
    let hundred_ns = p.fraction_ns / 100;
    let frac = format!("{hundred_ns:07}");
    s.push('.');
    s.push_str(&frac[..usize::from(p.scale).min(frac.len())]);
}
```

`scale` is populated per arm from `SqlTime::scale` (`t.scale` / `dt.time.scale` / `dto.datetime2.time.scale`), `3` for the legacy `datetime`, and `0` for `smalldatetime` — so a scale-0 column emits no fractional part at all.

Verified against your table: `datetime2(7)` `12:00:00` → `"2023-06-15 12:00:00.0000000"` (27), `datetime2(3)` → `"2023-06-15 12:00:00.100"` (23), `datetime` → `"2023-06-15 12:00:00.000"` (23), `smalldatetime` → `"2023-06-15 12:00:00"` (19). All four are now asserted in `character_rendering_pads_to_declared_scale`.

And noted — I've left the space before the timezone offset and the signed `+` for a zero offset exactly as they were.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Agreed — fixed in db7e1bcd. Silently dropping a component was wrong.

Took your suggested shape: converters now return `Result<ConvOk, ConvError>` with

```rust
pub(crate) enum ConvOk { Exact, Truncated }
```

`get_data.rs` posts `01S07` and returns `SQL_SUCCESS_WITH_INFO` on `Truncated`. Added `SQLSTATE_01S07` + `WARN_FRACTIONAL_TRUNCATION` to `sqlstate.rs`.

Applied to `SQL_C_TYPE_DATE` when a non-zero time is discarded, and to `SQL_C_TYPE_TIME` / `SQL_TIME_STRUCT` when a fraction is discarded (the struct has no fractional field). Tests cover both.

The numeric side (`float 1234.99` → `SQL_C_SLONG` → `1234` + `01S07`) needs the conversions themselves, which aren't implemented yet — that's tracked with the other Appendix D work in **AB#47107**. The `ConvOk`/`01S07` plumbing is now in place for it, so that task only has to add the conversions and return `Truncated`.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1288 @@
...
+pub(crate) unsafe fn convert_datetime_c(
+    value: &ColumnValues,
+    target_type: SqlSmallInt,
+    target_value_ptr: SqlPointer,
+    strlen_or_ind_ptr: *mut SqlLen,
+) -> Result<ConvOk, ConvError> {
+    // A date/time C target was requested for a non-temporal column: illegal.
+    let Some(p) = extract_datetime_parts(value) else {
+        return Err(ConvError::Restricted);
+    };
```
saurabh500: 🟡 **`Restricted` (07006) is now over-applied to character sources — msodbcsql converts these, so 07006 is a false claim**

The `ConvError` split landed correctly and `int` → `SQL_C_TYPE_TIMESTAMP` now returns `07006`, matching msodbcsql. But this `else` arm treats *every* source that `extract_datetime_parts` doesn't recognise as `Restricted`, and that bucket includes `ColumnValues::String`.

`07006` asserts the pairing is **illegal**. For character → date/time it isn't — it's legal and simply not implemented yet.

**Measured, same query both drivers:**

| Query | Target | msodbcsql18 | this PR |
|---|---|---|---|
| `CAST('2023-06-15' AS varchar(20))` | `SQL_C_TYPE_DATE` | `SQL_SUCCESS`, `2023-06-15` | `SQL_ERROR` `07006` |
| `CAST('not-a-date' AS varchar(20))` | `SQL_C_TYPE_DATE` | `SQL_ERROR` **`22018`** | `SQL_ERROR` `07006` |

The first row is the proof: msodbcsql *succeeds*, so the pairing is legal. msodbcsql only fails the second because the **text** is bad — `22018` is a data error, not a type error. Neither is `07006`.

This is a small regression against the previous commit, which returned `HYC00` here. `HYC00` was at least honest ("not implemented"); `07006` tells the application "never do this", which will make callers permanently disable a code path that P1a is about to make work. Your own P1a entry says character → date/time must eventually return `22018` on bad text, so 07006 is on the wrong axis entirely.

**Fix** — keep `Restricted` for sources that genuinely can never convert, and route character sources to `NotHandledHere` until P1a lands:

```rust
let Some(p) = extract_datetime_parts(value) else {
    // Character sources are a legal date/time conversion per Appendix D;
    // they are simply not implemented until P1a. Report "not implemented"
    // rather than claiming the pairing is restricted.
    return Err(match value {
        ColumnValues::String(_) => ConvError::NotHandledHere,
        _ => ConvError::Restricted,
    });
};
```

Worth applying the same reasoning in `convert_integer_c` / `convert_float_c`: `'123'` → `SQL_C_SLONG` currently reaches `NotHandledHere` → `HYC00`, which is already right — so this is only the datetime path that needs the carve-out.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -0,0 +1,1288 @@
...
+                    },
+                    strlen_or_ind_ptr,
+                )
+            };
+            // Dropping a non-zero time component is a truncation.
+            if p.has_time && (p.hour | p.minute | p.second) != 0 || p.fraction_ns != 0 {
+                ConvOk::Truncated
+            } else {
+                written
+            }
```
saurabh500, line 474: 🔵 **Nit — precedence makes this read as a bug even though it isn't**

`&&` binds tighter than `||`, so this parses as:

```rust
(p.has_time && (p.hour | p.minute | p.second) != 0) || (p.fraction_ns != 0)
```

The `fraction_ns` test escapes the `has_time` guard. It's harmless today because every `extract_datetime_parts` arm that sets `fraction_ns` also sets `has_time`, but that's an invariant held at a distance — a future arm setting a fraction without a time component would emit a spurious `01S07` on a date-only value.

**Fix** — make the grouping explicit:

```rust
```

saurabh500 (commented):
Re-reviewed at `db7e1bcd` and re-ran the same differential probe — 23 cases, identical SQL and buffers, this driver vs. `msodbcsql18` on SQL Server 2022.

**All six findings from my previous review are addressed, and both blockers are confirmed fixed against the real driver.**

| # | Finding | Status | Verified by |
|---|---|---|---|
| 🔴 | 100× tick scaling (`hms_from_nanos`) | **Fixed** | `time(7) '12:34:56.1234567'` → `SS_TIME2` now `12:34:56.123456700`, byte-identical to msodbcsql (was `00:07:32.961234567`) |
| 🔴 | `datetimeoffset` returned UTC in local fields | **Fixed** | `+05:30` → `12:34:56.123456700 +05:30`; `-05:00` and the `23:30 +05:30` rollover all match |
| 🟡 | `HYC00` where msodbcsql returns `07006` | **Fixed** | `int` → `SQL_C_TYPE_TIMESTAMP` now `07006` |
| 🟡 | Missing mandatory source conversions | **Deferred, tracked** | New P1a section + task 47107 in the plan doc |
| 🟡 | No `01S07` on truncation | **Fixed** | `datetime2` → `C_TYPE_TIME` and `→ C_TYPE_DATE` both now `SQL_SUCCESS_WITH_INFO` / `01S07` |
| 🟡 | Character output trimmed instead of padded | **Fixed** | `datetime2(7)` zero fraction → `"…12:00:00.0000000"` len **27**; `datetime` → len 23; `smalldatetime` unchanged at 19 |
| 🟡 | Tests built input in ns, masking the bug | **Fixed** | Rebuilt in ticks at `:921`/`:951` |
| 🟡 | `datetimeoffset` test pinned time to 0 | **Fixed** | `datetimeoffset_matches_msodbcsql_wall_clock` asserts the full wall clock; rollover covered both directions |

**Score: 20 of 23 cases now byte-identical to msodbcsql.** The 5 remaining `HYC00` cases are all the P1a deferral, which is the right call — the plan doc entry is specific about which pairings are missing and that `22018` is required on bad text, so the follow-up is well-specified rather than a vague TODO.

Two new comments below, neither blocking:

- 🟡 One genuine side effect of the `ConvError` split: the new `Restricted` arm catches character sources too, so `varchar` → `SQL_C_TYPE_DATE` now returns `07006`. msodbcsql *succeeds* on that pairing with valid text, so `07006` ("illegal cast") is a false claim where the old `HYC00` ("not implemented") was accurate. Small carve-out suggested inline.
- 🔵 A precedence grouping in the `01S07` condition that is correct today but fragile.

Worth calling out: splitting the streaming work out to #153 and re-scoping the plan doc around it is a much cleaner boundary than the original PR had, and the `div_euclid`/`rem_euclid` handling plus the `DATETIMEOFFSET_RANGE` guard went further than I suggested. The parity test comments citing the actual msodbcsql values are exactly what this driver needs more of.

Once the `07006`/character carve-out is in, this is good to merge from my side.

saurabh500 (approved):
I will leave it to author's discretion to fix these now, or later.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Agreed — the measured `SQL_SUCCESS` on `CAST('2023-06-15' AS varchar(20))` → `SQL_C_TYPE_DATE` is conclusive: the pairing is legal, so `07006` was a false claim and a regression against the previous commit. Fixed in ade49971 with exactly your carve-out:

```rust
let Some(p) = extract_datetime_parts(value) else {
    // Character sources are a legal date/time conversion per Appendix D and
    // land in P1a; report "not implemented" rather than claiming the
    // pairing is restricted. Everything else genuinely cannot convert.
    return Err(match value {
        ColumnValues::String(_) => ConvError::NotHandledHere,
        _ => ConvError::Restricted,
    });
};
```

So character → date/time is `HYC00` again, while `int` → `SQL_C_TYPE_TIMESTAMP` stays `07006`. Added a test asserting `HYC00` for `'2023-06-15'` → `SQL_C_TYPE_DATE` specifically, so this can't silently regress again.

And confirmed on your last point: `convert_integer_c` / `convert_float_c` already reach `NotHandledHere` for character sources, so the datetime path was the only one needing the carve-out.

Worth noting where this lands: P1a (AB#47107) implements character → date/time, so that branch replaces this carve-out with a real conversion returning `22018` on bad text — which is the axis you identified.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Fixed in ade49971 — parenthesised as suggested:

```rust
```

You're right that it was harmless today only because every arm that sets `fraction_ns` also sets `has_time` — an invariant held at a distance, and exactly the kind that a future arm quietly breaks. Better to make the grouping explicit than to rely on it.

David-Engel:
Validation caught one e2e failure, now fixed in d6a25572.

`GetDataLiveTest.UnsupportedColumnTypeHyc00PreservesValue` (added in #153) was anchored on `DATETIME` having no `column_value_to_text` arm. This PR adds that arm, so the first `SQLGetData` started succeeding — exactly the situation the test's own maintenance note anticipated:

> Once DATETIME becomes convertible the first SQLGetData will succeed and this test will pass for the wrong reason (no HYC00 reached). At that point re-point it at whatever column type is still unconvertible[.]

@shiwanigupta0809 — following that instruction, I re-pointed it at a non-PLP `VARBINARY(8)`. `ColumnValues::Bytes` has no `column_value_to_text` arm and none is planned in P1/P1a, so the HYC00-preserves-the-value contract is still exercised, on the same code path, with the assertions unchanged. The maintenance note now records the move and names the next re-point trigger (binary→hex). Shout if you'd rather it assert the target-type HYC00 path instead.

While here I added live coverage for two typed targets, since the unit tests build `SqlTime`/`SqlDateTime2` by hand and therefore cannot see wire-level unit errors:

- `IntColumnToSlongTarget` — `INT` → `SQL_C_SLONG`, value plus indicator.
- `Datetime2ToTimestampTargetKeepsFraction` — `DATETIME2(7)` → `SQL_C_TYPE_TIMESTAMP`, asserting `fraction == 123456700`. This is the only place the 100 ns tick vs. nanosecond mismatch found in review is observable end to end.

Both run unskipped against the msodbcsql comparison leg, so they double as parity checks.

saurabh500 (commented):
Re-review of the post-#153-merge changes (db7e1bcd..d6a25572). Verified end-to-end against msodbcsql with a differential probe on a fresh release build.

**In scope — clean.** Both findings from my previous review are fixed and confirmed on the wire:
- The `07006` character carve-out now routes `ColumnValues::String(_)` -> `NotHandledHere` (HYC00). The probe case "valid varchar -> SQL_C_TYPE_DATE" now returns HYC00, not the earlier false 07006.
- The `01S07` precedence grouping is fixed with explicit parens.

The #153 cursor integration in your `fetch.rs` / `stmt.rs` / `get_data.rs` changes reads correctly: the materialized `current_row` is gone, replaced by the lazy pull-cursor state (`row_positioned` / `last_captured` / `active_plp` / `partial_text_offset`); `begin_row` / `reset_row_stream` are well-named; the forward-only guard (07009 on backward access, SQL_NO_DATA on re-read of the just-consumed column) is coherent; and the fallible text decoders (`sql_string_to_text`, `xml_to_text`, json `from_utf8`) are wired through every caller in `column_value_to_text` — no panic path survives. I re-ran the 23-case parity probe: no regressions, byte-identical to the prior build, with the 5 remaining HYC00 cases being the tracked P1a numeric/character deferral.

**Out of scope — one pre-existing bug I want on your radar (not this PR).** While tracing the nvarchar(max) -> SQL_C_CHAR PLP transcode path that your typed work sits next to, I found a real buffer-overflow on surrogate-straddle: when a high surrogate is carried across a chunk boundary and pairs off ahead of 3-byte BMP text, `max_read` under-budgets and the transcoded UTF-8 overflows the caller buffer — `copy_with_nul` truncates mid-sequence and the dropped bytes are lost from the stream (silent corruption; a debug-build FFI panic via the `debug_assert!(!truncated)`).

`git blame` puts this on the #153 column-wise-fetch code now in `main` (commit `7e9dfadc`, Shiwani), so it is **not** yours to fix in this PR — flagging only because #146 builds directly on it and the existing `NvarcharMaxToCharChunkedAstralRoundTrip` test misses it (pure emoji expand to 2 bytes/unit and never exceed the budget). I reproduced it deterministically with `"A" + U+10437 + U+4F60` into a 7-byte buffer and filed it with the repro + fix as #211. No action needed here.

**Verdict: Approve with comments** — the in-scope typed-conversion work is correct and well-tested (the differential msodbcsql comparison leg with documented `SKIP_IF_COMPARING_MSODBCSQL` carve-outs is exactly the right way to pin parity). The one bug I found belongs to the #153 path and is tracked separately in #211.

saurabh500 (approved):

PR #147, opened 2026-07-23
Add mssql-python connection-string parity attributes to ODBC parser

Vahid-b:
## Description

Finishes connection-string compatibility parity with mssql-python (feature [AB#46372](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46372)) by extending the `SQLDriverConnect` parser to recognize and act on the full canonical ODBC Driver 18 keyword set that mssql-python normalizes to before handing the string to the driver.

Newly acted-on keywords, each wired into the mssql-tds `ClientContext` / `EncryptionOptions`:

| Keyword(s) | Landing spot | Value handling |
|---|---|---|
| `ServerSPN` | `server_spn` | verbatim |
| `ApplicationIntent` | `application_intent` | `ReadOnly` / `ReadWrite` (validated) |
| `MultiSubnetFailover` | `multi_subnet_failover` | `Yes` / `No` (validated) |
| `ConnectRetryCount` / `ConnectRetryInterval` | matching fields | integer |
| `KeepAlive` / `KeepAliveInterval` | `keep_alive_in_ms` / `keep_alive_interval_in_ms` | integer seconds → ms (×1000) |
| `IpAddressPreference` | `ipaddress_preference` | `IPv4First` / `IPv6First` / `UsePlatformDefault` (validated) |
| `PacketSize` | `packet_size` | integer bytes (→ u16, saturating) |
| `HostnameInCertificate` | `EncryptionOptions::host_name_in_cert` | verbatim |
| `ServerCertificate` | `EncryptionOptions::server_certificate` | verbatim path → `PathBuf` |

Also:
- Map the msodbcsql-native `Addr` / `Address` synonyms to `Server` (they were previously in the ignored set and silently dropped).
- Enumerated keys hard-reject invalid values (`E_FAIL` → `HY024`), matching how `Encrypt`/`Yes-No` already behave. Integer keys parse leniently — rejecting only non-numeric/negative input and leaving documented range clamping to mssql-tds so the parser can't diverge from the native driver.
- Collapse the `KEY_*` consts and the `classify_key` match into a single `MAPPED_KEYS` descriptor table (the code's own TODO sanctioned this once the mapped set grew).

Value validation ownership note: mssql-python value-validates nothing itself (it only *maps* `Authentication` for client-side token acquisition and passes everything else through), so mssql-odbc owns all connection-string value validation — which this PR implements.

## Related Issues

Closes [AB#46418](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46418)
Feature: [AB#46372](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46372)
Work item: https://dev.azure.com/SqlClientDrivers/mssql-rs/_workitems/edit/46418

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes — mssql-odbc suite is green (337/337 via nextest). The remaining workspace failures are mssql-tds live-DB integration tests that require a `.env`/SQL Server (untouched crate, environmental).
- [x] New/changed functionality has tests — 14 new parser unit tests + a gated end-to-end `NewConnectionAttributesParity` test
- [x] Public API changes are documented — `docs/connection_string_parser.md` updated (mapped-attributes table + value validation)

### DIFF
--- a/mssql-odbc/docs/connection_string_parser.md
+++ b/mssql-odbc/docs/connection_string_parser.md
@@ -58,7 +58,7 @@ For each key/value pair, repeated until end of input:
    - *Mapped* — a key we act on. First occurrence stores the value; duplicates are
      discarded (**first-wins**).
    - *Ignored* — recognized by msodbcsql but not acted on here (e.g. `Driver`,
-     `DSN`, `ApplicationIntent`, `OEMToANSI`). No warning.
+     `DSN`, `QuotedId`, `OEMToANSI`). No warning.
    - *Unknown* — not in the table. Raises an `01S00` warning, but **parsing
      continues** to the next pair; the value is parsed and discarded. Never fails
      and never halts the scan.
@@ -131,15 +131,61 @@ Only an invalid **value** on a recognized, validated key is a hard error: msodbc
 returns `E_FAIL` (→ `Err(InvalidAttrValue)`) and the connect fails. Unknown or
 invalid *keywords* never fail.
 
+## Mapped connection attributes
+
+`MAPPED_KEYS` is the single source of truth for the keywords the parser acts on.
+Each maps (case-insensitively) to a `ConnectionParams` field, which
+`driver_connect` then wires into the mssql-tds `ClientContext` / `EncryptionOptions`.
+Several spellings can share one slot, giving first-wins semantics across the whole
+synonym group (e.g. `Server` / `Addr` / `Address`).
+
+| Connection-string key(s) | mssql-tds landing spot | Value handling |
+|--------------------------|------------------------|----------------|
+| `Server` / `Addr` / `Address` | `create_client(server)` | verbatim (`server,port`) |
+| `Database` | `ClientContext::database` | verbatim |
+| `UID` | `ClientContext::user_name` | verbatim |
+| `PWD` | `ClientContext::password` | verbatim, redacted in logs |
+| `Authentication` | auth resolution | recognized-keyword set |
+| `Trusted_Connection` | integrated auth | `Yes` / `No` |
+| `Encrypt` | `EncryptionOptions::mode` | `Yes` / `Mandatory` / `No` / `Optional` / `Strict` |
+| `TrustServerCertificate` | `EncryptionOptions::trust_server_certificate` | `Yes` / `No` |
+| `HostnameInCertificate` | `EncryptionOptions::host_name_in_cert` | verbatim |
+| `ServerCertificate` | `EncryptionOptions::server_certificate` (path) | verbatim path |
+| `ServerSPN` | `ClientContext::server_spn` | verbatim |
+| `ApplicationIntent` | `ClientContext::application_intent` | `ReadOnly` / `ReadWrite` |
+| `MultiSubnetFailover` | `ClientContext::multi_subnet_failover` | `Yes` / `No` |
+| `IpAddressPreference` | `ClientContext::ipaddress_preference` | `IPv4First` / `IPv6First` / `UsePlatformDefault` |
+| `ConnectRetryCount` | `ClientContext::connect_retry_count` | integer |
+| `ConnectRetryInterval` | `ClientContext::connect_retry_interval` | integer |
+| `KeepAlive` | `ClientContext::keep_alive_in_ms` (×1000) | integer, seconds |
+| `KeepAliveInterval` | `ClientContext::keep_alive_interval_in_ms` (×1000) | integer, seconds |
+| `PacketSize` | `ClientContext::packet_size` (u16) | integer, bytes |
+
+This mirrors the canonical keyword set that mssql-python normalizes to before it
+hands the string to the driver, so a mssql-python application can target mssql-odbc
+without changing its connection strings.
+
 ## Value validation
 
 Whole-value, case-insensitive, exact match (not prefix, not `y`/`1`):
 
-- **Yes/No** keys (`TrustServerCertificate`, `Trusted_Connection`): `Yes` | `No`.
+- **Yes/No** keys (`TrustServerCertificate`, `Trusted_Connection`,
+  `MultiSubnetFailover`): `Yes` | `No`.
 - **`Encrypt`**: `Yes` | `Mandatory` | `No` | `Optional` | `Strict`.
+- **`ApplicationIntent`**: `ReadOnly` | `ReadWrite`.
+- **`IpAddressPreference`**: `IPv4First` | `IPv6First` | `UsePlatformDefault`.
+- **Integer** keys (`ConnectRetryCount`, `ConnectRetryInterval`, `KeepAlive`,
+  `KeepAliveInterval`, `PacketSize`): a non-negative integer. Documented ranges are
+  **not** enforced at parse time — out-of-range clamping is left to mssql-tds so the
+  parser cannot diverge from the native driver. A non-numeric or negative value is a
+  hard error (`E_FAIL`).
 - **`Authentication`**: delegated to `is_recognized_keyword`
   (`odbc_supported_auth_keywords.rs`) so the accept/reject set never drifts from
   mssql-tds. An empty value is a recognized reset.
+
+Pass-through string keys (`Server`/`Addr`/`Address`, `Database`, `UID`, `PWD`,
+`ServerSPN`, `HostnameInCertificate`, `ServerCertificate`) are stored verbatim with
+no value validation, exactly as msodbcsql does.
 
 ## Keys we recognize but do not act on
 
--- a/mssql-odbc/docs/connection_string_parser.md
+++ b/mssql-odbc/docs/connection_string_parser.md
@@ -155,11 +155,11 @@ synonym group (e.g. `Server` / `Addr` / `Address`).
 | `ApplicationIntent` | `ClientContext::application_intent` | `ReadOnly` / `ReadWrite` |
 | `MultiSubnetFailover` | `ClientContext::multi_subnet_failover` | `Yes` / `No` |
 | `IpAddressPreference` | `ClientContext::ipaddress_preference` | `IPv4First` / `IPv6First` / `UsePlatformDefault` |
-| `ConnectRetryCount` | `ClientContext::connect_retry_count` | integer |
-| `ConnectRetryInterval` | `ClientContext::connect_retry_interval` | integer |
-| `KeepAlive` | `ClientContext::keep_alive_in_ms` (×1000) | integer, seconds |
-| `KeepAliveInterval` | `ClientContext::keep_alive_interval_in_ms` (×1000) | integer, seconds |
-| `PacketSize` | `ClientContext::packet_size` (u16) | integer, bytes |
+| `ConnectRetryCount` | `ClientContext::connect_retry_count` | integer, clamped to 0–255 |
+| `ConnectRetryInterval` | `ClientContext::connect_retry_interval` | integer, clamped to 1–60 (seconds) |
+| `KeepAlive` | `ClientContext::keep_alive_in_ms` (×1000) | integer, seconds (saturating) |
+| `KeepAliveInterval` | `ClientContext::keep_alive_interval_in_ms` (×1000) | integer, seconds (saturating) |
+| `PacketSize` | `ClientContext::packet_size` (u16) | integer bytes, clamped to 512–32768 |
 
 This mirrors the canonical keyword set that mssql-python normalizes to before it
 hands the string to the driver, so a mssql-python application can target mssql-odbc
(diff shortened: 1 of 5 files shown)


mssql-odbc/docs/connection_string_parser.md
```diff
@@ -131,16 +131,62 @@ Only an invalid **value** on a recognized, validated key is a hard error: msodbc
...
-- **Yes/No** keys (`TrustServerCertificate`, `Trusted_Connection`): `Yes` | `No`.
+- **Yes/No** keys (`TrustServerCertificate`, `Trusted_Connection`,
+  `MultiSubnetFailover`): `Yes` | `No`.
 - **`Encrypt`**: `Yes` | `Mandatory` | `No` | `Optional` | `Strict`.
+- **`ApplicationIntent`**: `ReadOnly` | `ReadWrite`.
+- **`IpAddressPreference`**: `IPv4First` | `IPv6First` | `UsePlatformDefault`.
+- **Integer** keys (`ConnectRetryCount`, `ConnectRetryInterval`, `KeepAlive`,
+  `KeepAliveInterval`, `PacketSize`): a non-negative integer. Documented ranges are
+  **not** enforced at parse time — out-of-range clamping is left to mssql-tds so the
+  parser cannot diverge from the native driver. A non-numeric or negative value is a
```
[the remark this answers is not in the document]
Vahid-b: Updated. The design doc now describes the real behavior: the parser accepts any u32 and the driver clamps out-of-range values when mapping onto the ClientContext (PacketSize 512-32768, ConnectRetryCount 0-255, ConnectRetryInterval 1-60), rather than delegating range handling to mssql-tds.

mssql-odbc/src/api/driver_connect.rs
```diff
@@ -278,10 +280,53 @@ fn do_connect(
...
+            ApplicationIntent::ReadOnly
+        } else {
+            ApplicationIntent::ReadWrite
+        };
+    }
+    if let Some(multi_subnet_failover) = params.multi_subnet_failover {
+        context.multi_subnet_failover = multi_subnet_failover;
+    }
+    if let Some(count) = params.connect_retry_count {
+        context.connect_retry_count = count;
```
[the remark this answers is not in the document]
Vahid-b: Good catch. Fixed in 68f25f7e: the param->ClientContext mapping now lives in `apply_connection_params`, which clamps `ConnectRetryCount` to 0-255 (and `ConnectRetryInterval` to 1-60) at the boundary, so the downstream `connect_retry_count + 1` in mssql-tds can no longer overflow. Clamping rather than rejecting mirrors msodbcsql. Added a unit test that feeds `u32::MAX` and asserts the result stays addable.

mssql-odbc/src/connection/connection_string_parser.rs
```diff
@@ -546,6 +652,18 @@ fn is_yes(value: &str) -> bool {
     value.eq_ignore_ascii_case("yes")
 }
 
+/// Parse a non-negative integer attribute value. Rejects non-numeric or negative
+/// input with `InvalidAttrValue` (msodbcsql `E_FAIL`). Documented per-key ranges
+/// (e.g. PacketSize 512–32767) are intentionally not enforced here; range handling
```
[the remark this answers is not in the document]
Vahid-b: Fixed. The doc comment now reads 512-32768 (matching `DefaultClientContextValidator`, MAX = 32768) and no longer claims range handling is deferred to mssql-tds.

mssql-odbc/src/api/driver_connect.rs
```diff
@@ -278,10 +280,53 @@ fn do_connect(
             Some(e) if e.eq_ignore_ascii_case("strict") => EncryptionSetting::Strict,
             _ => EncryptionSetting::On, // ODBC default
         },
-        host_name_in_cert: None,
-        server_certificate: None,
+        host_name_in_cert: params.host_name_in_certificate.clone(),
+        server_certificate: params.server_certificate.as_deref().map(PathBuf::from),
```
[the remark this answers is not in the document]
Vahid-b: Done. Factored the mapping into `apply_connection_params(&mut ClientContext, &ConnectionParams)` and added direct unit tests asserting the HostnameInCertificate/ServerCertificate -> EncryptionOptions mappings, plus the enum and clamping paths, so they are no longer only reachable via a live-server e2e test.

mssql-odbc/src/connection/connection_string_parser.rs
```diff
@@ -234,6 +238,42 @@ impl ConnectionParams {
...
+            parts.push(format!("IpAddressPreference={ip_address_preference}"));
+        }
+        if let Some(packet_size) = self.packet_size {
+            parts.push(format!("PacketSize={packet_size}"));
+        }
+        if let Some(host_name_in_certificate) = &self.host_name_in_certificate {
+            parts.push(format!("HostnameInCertificate={host_name_in_certificate}"));
+        }
+        if let Some(server_certificate) = &self.server_certificate {
+            parts.push(format!("ServerCertificate={server_certificate}"));
```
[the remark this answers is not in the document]
Vahid-b: Addressed in 68f25f7e. This formatter only feeds the redacted diagnostic log line (never re-parsed), but the ambiguity was real. Added a `quote_odbc_value` helper that brace-quotes values containing `;{}=` and doubles any inner `}`, applied to the free-form fields including ServerCertificate, with tests.

mssql-odbc/src/connection/connection_string_parser.rs
```diff
@@ -234,11 +238,66 @@ impl ConnectionParams {
...
+            parts.push(format!("ApplicationIntent={application_intent}"));
+        }
+        if let Some(multi_subnet_failover) = self.multi_subnet_failover {
+            parts.push(format!(
+                "MultiSubnetFailover={}",
+                if multi_subnet_failover { "yes" } else { "no" }
+            ));
+        }
+        if let Some(connect_retry_count) = self.connect_retry_count {
+            parts.push(format!("ConnectRetryCount={connect_retry_count}"));
```
shiwanigupta0809, line 262: in msodbcsql ConnectRetryCount and ConnectRetryInterval are range-validated during parse  and rejected when out of range, but here we accept and clamp later. Is this clamping intentional here?
https://sqlclientdrivers.visualstudio.com/msodbcsql/_git/msodbcsql?path=/Sql/Ntdbms/sqlncli/odbc/sqlcconn.cpp&version=GBmaster&_a=contents&line=4136&lineStyle=plain&lineEnd=4136&lineStartColumn=9&lineEndColumn=36
https://sqlclientdrivers.visualstudio.com/msodbcsql/_git/msodbcsql?path=/Sql/Ntdbms/sqlncli/odbc/sqlcconn.cpp&version=GBmaster&_a=contents&line=4009&lineStyle=plain&lineEnd=4009&lineStartColumn=1&lineEndColumn=30

mssql-odbc/src/connection/connection_string_parser.rs
```diff
@@ -332,6 +423,40 @@ fn assign_value(
...
+            params.connect_retry_interval = Some(parse_uint(lower, value)?);
+        }
+        ConnAttrKey::KeepAlive => {
+            params.keep_alive = Some(parse_uint(lower, value)?);
+        }
+        ConnAttrKey::KeepAliveInterval => {
+            params.keep_alive_interval = Some(parse_uint(lower, value)?);
+        }
+        ConnAttrKey::IpAddressPreference => {
+            validate_attr(lower, value, IP_ADDRESS_PREFERENCE_VALUES)?;
```
shiwanigupta0809: msodbcsql falls back unknown values to IPv4First, while this parser currently rejects unknown values with HY024.  Is this strict rejection intentional? 
I didnt find [ODBC spec ](https://learn.microsoft.com/en-us/sql/connect/odbc/dsn-connection-string-attribute?view=sql-server-ver17) defined behavior for unknown IpAddressPreference values, so this is a product-parity decision

mssql-odbc/src/connection/connection_string_parser.rs
Vahid-b: You are right, thanks for the msodbcsql pointer. Changed in 61e66405: ConnectRetryCount (0-255) and ConnectRetryInterval (1-60) are now range-validated at parse time and rejected when out of range (matching sqlcconn.cpp) instead of clamping downstream. This also removes the `connect_retry_count + 1` overflow without needing a clamp. Added parser tests for the boundaries and the rejection.

mssql-odbc/src/connection/connection_string_parser.rs
Vahid-b: Good catch, aligned to msodbcsql in 61e66405. IpAddressPreference no longer rejects unknown values, they are accepted at parse and fall back to IPv4First when mapped onto the ClientContext (apply_connection_params). PacketSize still clamps, since that is the range mssql-tds accepts. Updated tests and docs to describe the fallback.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
```diff
@@ -457,3 +457,97 @@ TEST_F(DriverConnectLiveTest, ConnectionStringParserParityBehaviors) {
...
+    {
+        auto r = tryConnect(base + ";PacketSize=notanumber");
+        EXPECT_EQ(SQL_ERROR, r.rc);
+        EXPECT_TRUE(r.hasHY024);
+    }
+
+    // Out-of-domain IpAddressPreference -> HY024, connect fails.
+    {
+        auto r = tryConnect(base + ";IpAddressPreference=IPv7");
+        EXPECT_EQ(SQL_ERROR, r.rc);
```
shiwanigupta0809: This test also needs to be updated with the latest commit. PR pipeline is failing due to this.
Out-of-domain IpAddr is accepted and falls back to IPv4First

mssql-odbc/src/connection/connection_string_parser.rs
```diff
@@ -1092,6 +1269,201 @@ mod tests {
...
+            let s = format!("{key}=host1;UID=u;<PW>=p");
+            let (p, warn) = parse_connection_string(&cs(&s)).unwrap();
+            assert!(!warn, "{key} should map cleanly to Server");
+            assert_eq!(p.server, "host1", "key {key}");
+        }
+    }
+
+    #[test]
+    fn server_and_address_share_one_slot_first_wins() {
+        let (p, warn) = parse_connection_string("Server=first;Address=second;UID=u").unwrap();
```
shiwanigupta0809: parity nit: [msodbcsql](https://sqlclientdrivers.visualstudio.com/msodbcsql/_git/msodbcsql?path=/Sql/Ntdbms/sqlncli/odbc/sqlcconn.cpp&version=GBmaster&_a=contents&line=2965&lineStyle=plain&lineEnd=2970&lineStartColumn=1&lineEndColumn=66) prefers address over server. [spec ](https://learn.microsoft.com/en-us/sql/relational-databases/native-client/applications/using-connection-string-keywords-with-sql-server-native-client?view=sql-server-ver15&viewFallbackFrom=sql-server-ver17) mentions that too.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
Vahid-b: Fixed in 9b9c7abd. Updated this e2e case so IpAddressPreference=IPv7 now expects a successful connect with no 01S00, matching the parser change (unknown values are accepted and fall back to IPv4First). That was the cause of the failing Linux stage (and the coverage-report job, which was just timing out waiting for the artifact the failed stage never produced).

mssql-odbc/src/connection/connection_string_parser.rs
Vahid-b: Good call, implemented in 9b9c7abd. The server/addr/address group is now resolved so a non-empty Address/Addr takes precedence over Server regardless of position, and an empty Address falls back to Server (per the SNAC ODBC spec you linked). Added unit tests for both orderings, the empty-Address fallback, and first-wins within each spelling. Note this only affects callers passing both keys directly to the driver: mssql-python collapses the synonym group to a single canonical Server (first-wins) before it reaches ODBC, so the mssql-python path is unaffected either way.

Vahid-b: Correction on my earlier reply: I referenced the SQL Server Native Client (SNAC) public doc, but the authoritative parity source is msodbcsql. I've since verified the change directly against the msodbcsql source, and it matches:

- `sqlcconn.cpp` builds its login target (`ServerNameLoginInfo`) from `KEY_ADDR` when Address is present with a value, otherwise `KEY_SERVER` — i.e. non-empty Address/Addr wins over Server, empty/absent Address falls back to Server. That's exactly what the parser now does.
- Both spellings are recognized as `KEY_ADDR`: `L"ADDR"` and `INI_ADDRESS` (`#define INI_ADDRESS L"Address"`), matching our `addr`/`address` synonyms.

Updated the code comments and design doc to cite `sqlcconn.cpp` instead of the SNAC doc (04f5719e).

shiwanigupta0809 (approved):

mssql-odbc/src/api/driver_connect.rs
```diff
@@ -331,6 +335,66 @@ fn do_connect(
...
+    if let Some(pref) = &params.ip_address_preference {
+        context.ipaddress_preference = if pref.eq_ignore_ascii_case("ipv6first") {
+            IPAddressPreference::IPv6First
+        } else if pref.eq_ignore_ascii_case("useplatformdefault") {
+            IPAddressPreference::UsePlatformDefault
+        } else {
+            IPAddressPreference::IPv4First
+        };
+    }
+    if let Some(size) = params.packet_size {
```
saurabh500, line 392: Can you add a line to make sure that the library name in the Client Context is set to the ODBC driver for every connection? We want the ODBC driver name so that the connections are redirected appropriately in Private link scenarios ?

PR #83, opened 2026-06-29
Add Always Encrypted column-encryption support

David-Engel:
## Summary

Adds [Always Encrypted](https://learn.microsoft.com/sql/relational-databases/security/encryption/always-encrypted-database-engine) support to the `mssql-tds` driver: transparent encryption of query parameters and decryption of encrypted result-set columns, plus bulk copy into encrypted columns. The feature negotiates the Always Encrypted TDS feature extension at login, parses the column encryption key (CEK) table and per-column crypto metadata, and uses `sp_describe_parameter_encryption` to determine which parameters require encryption.

Always Encrypted is compiled unconditionally and stays inert unless the caller opts in at runtime, so existing builds and behavior are unchanged and the public API surface stays backward compatible.

## Runtime opt-in

Always Encrypted is additive and disabled by default at runtime. The connection only negotiates the feature when the caller sets the Column Encryption Setting (connection-level or per-command override); no encryption or decryption happens otherwise.

## Cryptography

The data path uses `AEAD_AES_256_CBC_HMAC_SHA256` for cell encryption and RSA-OAEP for CEK wrapping. Cryptography is provided by platform-native libraries — per the crypto compliance policy, which permits only OpenSSL, Apple native, and Windows native libraries — through a small `security::crypto` abstraction that selects a backend at compile time:

- **Windows** targets use CNG (`bcrypt.dll`) plus the CryptoAPI ASN.1 helpers (`crypt32.dll`) via the existing `windows-sys` bindings. This needs no `openssl-sys`, so default Windows (Schannel) builds require no system OpenSSL and no perl/NASM build tooling.
- **macOS** targets use Apple's Security.framework (`SecKey` for RSA and key generation) and CommonCrypto (AES-256-CBC, HMAC-SHA256, secure RNG), mirroring how the TLS layer already uses Secure Transport on Apple platforms. This needs no `openssl-sys` either, so default macOS builds require no system OpenSSL.
- **All other targets** (Linux, etc.) use OpenSSL, which is approved on every platform.

[...]

PR #148, opened 2026-07-23
Prepared-handle invalidation after transparent reconnect

Theekshna:
### Invalidate prepared-statement handles across a transparent reconnect

Prepared-statement handles are per-session: SQL Server restarts handle numbering on every new session, so after a transparent reconnect (Idle Connection Resiliency) any cached handle belongs to the dead session. Reusing it via `sp_execute` either fails with error 8179 ("Could not find prepared statement with handle") or, worse, silently aliases an unrelated handle. Before this PR, `SQLExecute` / `SQLFreeHandle` reused cached handles blindly across a reconnect.

This PR makes the server handle an implementation detail that never leaves `mssql-tds`. Callers hold an opaque, client-issued `StatementId`; the client owns the `StatementId -> handle` map and clears it on reconnect, so **"stale" simply becomes "absent from the map"** — there is no epoch to compare and no liveness decision for a caller to get wrong.

### Approach

- **Opaque identity.** `StatementId(u64)` is unique and never reused within the client that issues it (not yet globally unique across clients — tracked in [AB#47292](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47292)). `PreparedStatement { sql, id: Option<StatementId> }` carries no handle. The client owns `prepared_handles: HashMap<StatementId, i32>` and the parallel Always Encrypted describe map, both keyed the same way.
- **Recovery is owned by `execute_prepared`.** It reconnects once up front (mirrors msodbcsql `GetBatchCtxOrRecover`), charges the elapsed time against the command timeout, fails fast when recovery consumed the whole budget, then: id resolves to a live handle → reuse via `sp_execute`; absent → prepare via `sp_prepexec`. No RPC self-recovers after that decision, so a reconnect can never slip between "chose the handle" and "sent it" — the fix does not depend on server error 8179.
- **Capture has no caller step.** `sp_prepexec` arms `pending_capture`; `push_return_value` — the single funnel for every RETURNVALUE — writes `id -> handle` and pins the AE describe metadata when the token lands, on whichever drain path reads it.
- **Orphan release.** A superseded statement is passed as `&mut Option<StatementId>` and resolved to a handle *inside* `sp_prepexec`, piggybacked onto the by-ref `@handle` argument (matching msodbcsql `sqlccmd.cpp`). It is taken, and its map entries evicted, only at the serialization boundary; a pre-send failure leaves it releasable.
- **`unprepare` evicts before sending.** A failed `sp_unprepare` is ambiguous, and a map entry naming a plan the server already dropped would make the next execute fail with 8179 instead of re-preparing. Absent is the recoverable state.

### API changes (`mssql-tds`) — breaking, pre-release

- Public surface is `StatementId`, `PreparedStatement`, `execute_prepared()`, `unprepare()`.
- `PreparedHandle`, `session_handle`, `take_session_handle`, `capture_prepared_handle_into`, and `plan_prepared_execution` are **removed**. No compatibility shim: they exposed the raw handle and the caller-driven capture that made the aliasing and dropped-orphan bugs reachable. No release has shipped them (crate is `0.1.0`, `CHANGELOG` `[Unreleased]`).
- The four `sp_*` RPCs and `check_and_reconnect` are private; tests and the benchmark reach them through `#[doc(hidden)]` `*_for_test` wrappers under the existing `test-util` feature. `PreparedStatement` is not `Clone` (clones would share one id).

### ODBC layer (`mssql-odbc`)

- `StmtState` stores `PreparedPlan { stmt, marker_count }`; `pending_unprepare` is `Option<StatementId>`.
- `SQLExecute` delegates the whole reuse/re-prepare/recovery sequence to `execute_prepared()`. The three post-drain `capture_prepared_handle` call sites (`close_cursor`, `exec_common`, `more_results`) are **deleted** — capture is internal now.
- Parameter-marker rewriting (`?` → `@P1..@Pn`) happens once at `SQLPrepare`.

### Tests

- **`mssql-tds` unit:** `execute_prepared` branch matrix (first materialization / id-with-no-live-handle re-prepares with a *new* id / live-handle reuse), each asserting the RPC proc id **and** `@handle` bytes on the wire; `unprepare` release-vs-skip and both-map eviction on RPC failure; orphan retention across the send boundary; budget exhaustion; AE pinning; a regression test that a reused server handle id cannot collide across statements.
- **`mssql-odbc` unit:** orphan threading through staging and restoration on claim failure.
- **Live integration** (`test_rpc_results.rs`, 11 tests) covers `sp_prepare`/`sp_prepexec`/`sp_execute`/`sp_unprepare` incl. the `@handle` piggyback — verified against a local SQL Server.
- ⚠️ **Not covered by CI:** `connectivity.rs` is excluded by the `not (test(connectivity))` filter, and the ODBC e2e `StaleHandleAfterReconnectIsInvalidatedAndReprepared` still `GTEST_SKIP`s pending the transparent-reconnect (LOGIN7) fix in [AB#46631](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46631); enabling it is [AB#47099](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47099).

## Related Issues

Closes [AB#46527](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46527)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes — 1637 (`mssql-tds`) + 492 (`mssql-odbc`) lib tests
- [x] New/changed functionality has tests — *except the reconnect scenario itself; see the ⚠️ above*
- [x] Public API changes are documented — `CHANGELOG.md` Changed/Removed


mssql-tds/src/connection/tds_client.rs
```diff
@@ -4359,45 +4435,4 @@ mod tests {
 
         assert!(matches!(err, UsageError(message) if message.contains("@CE_POS_0")));
     }
```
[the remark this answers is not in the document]
Theekshna: Was removed accidentally during rebase. Restored.

mssql-odbc/src/api/execute.rs
```diff
@@ -66,54 +67,121 @@ fn sql_execute_safe(statement_handle: SqlHandle, stmt: &StmtHandle) -> SqlReturn
...
+    // exhausted budget must be turned into an immediate `HYT00` here (don't
+    // issue the RPC) rather than passed down. See `TdsClient::deduct_timeout`.
+    let timeout_sec: Option<u32> = None;
+
+    // Single recovery point (msodbcsql `GetBatchCtxOrRecover`): reconnect
+    // *before* the liveness decision below. No execute RPC recovers again, so a
+    // reconnect can't slip in and make the chosen handle stale mid-send.
+    let reconnect_elapsed = match dbc
+        .runtime
+        .block_on(client.check_and_reconnect(timeout_sec, None))
```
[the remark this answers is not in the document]
Theekshna: This has parity with msodbcsql driver. It ignores the info messages during reconnect too. Created a task to fix this in other places too - [Clear info messages from reconnect](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46613)

David-Engel (commented):
I think we should reconsider pushing prepared-handle epoch tracking and reprepare decisions into every `mssql-tds` consumer. A prepared handle is a session-scoped implementation detail, and callers should not need to coordinate `check_and_reconnect()`, `connection_recovery_count()`, timeout deduction, stale-handle cleanup, and re-preparation correctly.

Instead, could `mssql-tds` expose a logical `PreparedStatement` containing everything required to recreate its server handle?

```rust
pub struct PreparedStatement {
    sql: String,
    parameter_declarations: String,
    session_handle: Option<SessionHandle>,
}

struct SessionHandle {
    id: i32,
    recovery_epoch: u32,
}
```

The preferred public surface could be:

```rust
PreparedStatement::new(...)
TdsClient::prepare(...)
TdsClient::prepare_and_execute(...)
TdsClient::execute_prepared(...)
TdsClient::unprepare(...)
```

`execute_prepared` would own the complete recovery sequence:

1. Run ICR when necessary.
2. Deduct recovery time from the command timeout.
3. Compare the stored handle epoch with the current connection epoch.
4. If the handle is absent or stale, prepare it again using the retained SQL and parameter declarations.
5. Replace the internal handle and epoch.
6. Execute the valid handle.

This keeps the recovery and handle-liveness decision atomic within `mssql-tds`. It also avoids exposing bare server handle IDs that can silently alias an unrelated statement after reconnect.

For an explicit `sp_prepare` followed by `sp_execute` flow, consumer code would look like:

```rust
let mut statement = client
    .prepare(
        "SELECT @P1",
        "@P1 int",
        options,
    )
    .await?;

let mut results = client
    .execute_prepared(
        &mut statement,
        parameters![("@P1", 42)],
        options,
    )
    .await?;

while let Some(row) = results.next_row().await? {
    // Process row.
}

results.close().await?;
```

For a one-round-trip initial `sp_prepexec` followed by later `sp_execute` calls:

```rust
let mut statement = PreparedStatement::new(
    "SELECT @P1",
    "@P1 int",
);

let mut results = client
    .prepare_and_execute(
        &mut statement,
        parameters![("@P1", 42)],
        options,
    )
    .await?;

while let Some(row) = results.next_row().await? {
    // mssql-tds records the @handle RETURNVALUE in `statement`
    // when it reaches that token in the stream.
}

results.close().await?;

let mut results = client
    .execute_prepared(
        &mut statement,
        parameters![("@P1", 84)],
        options,
    )
    .await?;
```

The streaming result object can borrow both the client and statement while the response is active:

```rust
pub struct PreparedExecution<'a> {
    client: &'a mut TdsClient,
    statement: &'a mut PreparedStatement,
}
```

When it encounters the `sp_prepexec` handle token, it updates the statement's `SessionHandle`. `close()` must drain through that token when the consumer stops fetching early. If the stream fails before the handle arrives, the statement remains unmaterialized and the next execution prepares it again.

Cleanup should also operate on the logical object:

```rust
client.unprepare(statement, options).await?;
```

If its epoch is stale, `unprepare` only clears the local handle because the old session has already released the server resource.

I also suggest dropping the `execute_sp_prepare`, `execute_sp_prepexec`, `execute_sp_execute`, and `execute_sp_unprepare` names from the primary public API. They expose the current wire implementation and become inaccurate once an operation may transparently run a different procedure after ICR. The existing methods could become private `_raw` helpers.

This is a breaking API change, but this seems like the right time to make it while downstream usage is still limited. It establishes the correct ownership boundary now rather than requiring each consumer to implement the same subtle recovery protocol.

mssql-odbc/src/api/execute.rs
```diff
@@ -66,54 +67,121 @@ fn sql_execute_safe(statement_handle: SqlHandle, stmt: &StmtHandle) -> SqlReturn
...
+            error!(%e, "SQLExecute: reconnect failed");
+            return fail_with_tds(dbc, stmt, statement_handle, client, &e);
+        }
+    };
+    // Keep recovery + execution within one command-timeout budget.
+    let timeout_sec = TdsClient::deduct_timeout(timeout_sec, reconnect_elapsed);
+    let session_epoch = client.connection_recovery_count();
+
+    // msodbcsql `FIsReprepareRequired`: reuse vs. reprepare, judged by epoch.
+    match plan_execution(exec.handle, exec.drop_handle, session_epoch) {
```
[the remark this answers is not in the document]
Theekshna: Added a e2e test, but its skipped for now coz session recovery in mssql-tds is not working. Here's a work item to track it. - https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46631

Theekshna:
> I think we should reconsider pushing prepared-handle epoch tracking and reprepare decisions into every `mssql-tds` consumer. A prepared handle is a session-scoped implementation detail, and callers should not need to coordinate `check_and_reconnect()`, `connection_recovery_count()`, timeout deduction, stale-handle cleanup, and re-preparation correctly.
> 
> Instead, could `mssql-tds` expose a logical `PreparedStatement` containing everything required to recreate its server handle?
> 
> ```rust
> pub struct PreparedStatement {
>     sql: String,
>     parameter_declarations: String,
>     session_handle: Option<SessionHandle>,
> }
> 
> struct SessionHandle {
>     id: i32,
>     recovery_epoch: u32,
> }
> ```
> 
> The preferred public surface could be:
> 
> ```rust
> PreparedStatement::new(...)
> TdsClient::prepare(...)
> TdsClient::prepare_and_execute(...)
> TdsClient::execute_prepared(...)
> TdsClient::unprepare(...)
> ```
> 
> `execute_prepared` would own the complete recovery sequence:
> 
> 1. Run ICR when necessary.
> 2. Deduct recovery time from the command timeout.
> 3. Compare the stored handle epoch with the current connection epoch.
> 4. If the handle is absent or stale, prepare it again using the retained SQL and parameter declarations.
> 5. Replace the internal handle and epoch.
> 6. Execute the valid handle.
> 
> This keeps the recovery and handle-liveness decision atomic within `mssql-tds`. It also avoids exposing bare server handle IDs that can silently alias an unrelated statement after reconnect.
> 
> For an explicit `sp_prepare` followed by `sp_execute` flow, consumer code would look like:
> 
> ```rust
> let mut statement = client
>     .prepare(
>         "SELECT @P1",
>         "@P1 int",
>         options,
>     )
>     .await?;
> 
> let mut results = client
>     .execute_prepared(
>         &mut statement,
>         parameters![("@P1", 42)],
>         options,
>     )
>     .await?;
> 
> while let Some(row) = results.next_row().await? {
>     // Process row.
> }
> 
> results.close().await?;
> ```
> 
> For a one-round-trip initial `sp_prepexec` followed by later `sp_execute` calls:
> 
> ```rust
> let mut statement = PreparedStatement::new(
>     "SELECT @P1",
>     "@P1 int",
> );
> 
> let mut results = client
>     .prepare_and_execute(
>         &mut statement,
>         parameters![("@P1", 42)],
>         options,
>     )
>     .await?;
> 
> while let Some(row) = results.next_row().await? {
>     // mssql-tds records the @handle RETURNVALUE in `statement`
>     // when it reaches that token in the stream.
> }
> 
> results.close().await?;
> 
> let mut results = client
>     .execute_prepared(
>         &mut statement,
>         parameters![("@P1", 84)],
>         options,
>     )
>     .await?;
> ```
> 
> The streaming result object can borrow both the client and statement while the response is active:
> 
> ```rust
> pub struct PreparedExecution<'a> {
>     client: &'a mut TdsClient,
>     statement: &'a mut PreparedStatement,
> }
> ```
> 
> When it encounters the `sp_prepexec` handle token, it updates the statement's `SessionHandle`. `close()` must drain through that token when the consumer stops fetching early. If the stream fails before the handle arrives, the statement remains unmaterialized and the next execution prepares it again.
> 
> Cleanup should also operate on the logical object:
> 
> ```rust
> client.unprepare(statement, options).await?;
> ```
> 
> If its epoch is stale, `unprepare` only clears the local handle because the old session has already released the server resource.
> 
> I also suggest dropping the `execute_sp_prepare`, `execute_sp_prepexec`, `execute_sp_execute`, and `execute_sp_unprepare` names from the primary public API. They expose the current wire implementation and become inaccurate once an operation may transparently run a different procedure after ICR. The existing methods could become private `_raw` helpers.
> 
> This is a breaking API change, but this seems like the right time to make it while downstream usage is still limited. It establishes the correct ownership boundary now rather than requiring each consumer to implement the same subtle recovery protocol.

Makes sense. Thanks for the suggestion. I added the new `execute_prepared` function that deals with reconnect instead of offloading to user. I was earlier iffy about doing this wondering if other drivers will ever use it. But i think this is a common enough use case & a lot cleaner to do it this way.

David-Engel (dismissed):
Thanks for reworking this — moving the recovery protocol into `mssql-tds` is the right direction, and `PreparedStatement` / `PreparedHandle` is a much better boundary than making every caller track epochs. The epoch-planning logic itself is clean and well unit-tested.

I'm requesting changes: there are three correctness issues in the new managed path, plus a validation gap that means the scenario this PR targets isn't currently demonstrated by any enabled test.

Reviewed at `aad5f9047c8139a24b83930a10bf29a98a830476`.

---

### Blocking

**1. Managed prepared statements break on the second execution under Always Encrypted**

`execute_sp_prepexec_raw` encrypts parameters (`tds_client.rs:1675`) but never caches the resulting describe metadata under the handle the server returns. Only `execute_sp_prepare_raw` populates `prepared_param_encryption` (`tds_client.rs:1407`), and `execute_sp_execute_raw` requires that entry (`tds_client.rs:1809`).

Since `execute_prepared` materializes every statement via `sp_prepexec` and then reuses it via `sp_execute`, this fails deterministically whenever column encryption is active and the statement has parameters:

1. First `execute_prepared` → `sp_prepexec` succeeds, handle captured.
2. Second `execute_prepared` → `Reuse` → `sp_execute`.
3. Lookup misses → `ColumnEncryptionError` ("has no Always Encrypted parameter metadata... prepare it with execute_sp_prepare on this connection").

No reconnect is needed to hit this. The metadata gathered during `sp_prepexec` needs to be associated with the handle once the `@handle` RETURNVALUE is known, and this path needs an AE test that executes the same prepared statement twice.

**2. `session_epoch` can't distinguish handles from different connections**

`PreparedHandle` carries only a per-client recovery count (`tds_client.rs:3691`), and `plan_prepared_execution` compares only that count (`tds_client.rs:3779`). `PreparedStatement` is a public, `Clone` type with no tie to the client that materialized it.

A statement materialized on client A at epoch 0 and then passed to client B at epoch 0 is classified `Reuse`, so B issues `sp_execute` with a handle id that belongs to a different session — the exact aliasing hazard the PR sets out to eliminate, just reached by a different route. `unprepare()` has the same exposure and can release an unrelated plan.

The handle needs an identity that is unique per session (for example a monotonically-issued connection id combined with the recovery count), or the API needs to prevent a statement from being used with a client other than its owner.

**3. Re-preparation doesn't preserve how the statement was originally prepared**

`PreparedStatement` stores only SQL and an optional handle (`tds_client.rs:3717`), and `execute_prepared` routes every absent-or-stale statement through `sp_prepexec` (`tds_client.rs:1542-1547`).

`sp_prepare` compiles without execution values; `sp_prepexec` compiles with the supplied values. Because of parameter sniffing, they can produce different plans. A statement prepared via `sp_prepare` and later recovered via `sp_prepexec` can therefore come back with different performance characteristics after a reconnect — a silent behavior change rather than a visible failure, which makes it hard to diagnose in the field.

Worth recording the preparation strategy (and the parameter declarations needed to reproduce it) on the statement, then recovering with the matching sequence: `sp_prepare` + `sp_execute` for statements prepared that way, `sp_prepexec` for the combined form. Related: there is currently no managed entry point corresponding to `sp_prepare`, so that strategy isn't reachable through the new API at all.

---

### Should fix before merge

**4. A pending orphaned handle is lost when execution fails**

`stage_execution` takes `pending_unprepare` out of `StmtState` (`execute.rs:170`). It's restored when the connection claim fails (`execute.rs:80`), but the `execute_prepared` error path only restores `prepared_stmt` (`execute.rs:104-106`) before returning via `fail_with_tds` (`execute.rs:112`).

If the call fails before the piggybacked drop reaches the server (reconnect failure, encryption failure, serialization error), the driver forgets a still-live server-side plan and it stays allocated until disconnect. Restoring the orphan on that path — or having the API report whether the drop was consumed — would close this.

**5. The scenario this PR fixes has no enabled test**

The E2E test unconditionally skips in `SetUp` (`session_recovery_test.cpp:25`), and the live TDS test drives the raw APIs to assert that a stale handle *fails* with 8179 (`connectivity.rs:546-589`) rather than exercising `execute_prepared`.

The unit tests cover `plan_prepared_execution` and handle capture well, but nothing enabled demonstrates the full managed loop: reconnect → detect stale → re-prepare → capture new handle → reuse successfully. Given the description notes transparent reconnect isn't working yet in `mssql-tds`, it's worth being explicit about whether this lands ahead of that fix, and if so what evidence we have that the new path behaves as intended.

**6. The `_raw` methods are still public**

`execute_sp_prepare_raw` (1282), `execute_sp_unprepare_raw` (1441), `execute_sp_prepexec_raw` (1621), and `execute_sp_execute_raw` (1763) remain `pub`. Renaming signals intent, but consumers can still take the bare-handle path and reintroduce the class of bug this PR fixes.

Making them `pub(crate)` (or private) would make the boundary real. The blocker is the tests and benchmark that call them from outside the crate; protocol-level coverage could move into the crate where private access is available.

There's also an internal inconsistency: `execute_sp_prepare_raw` is documented as "never reconnects / caller-owned recovery" but still calls `check_and_reconnect` at `tds_client.rs:1300`.

---

### Smaller items

**7. Handle capture is still a two-step contract.** `execute_prepared` (`tds_client.rs:1525`) doesn't finish updating the statement — callers must drain, then remember `capture_prepared_handle_into` (`1558`). The ODBC layer already needs this at three separate places (`close_cursor.rs:216`, `exec_common.rs:303`, `more_results.rs:185`), and a drain path added later would silently drop handles. That's protocol ordering leaking through the managed API, and it's the footgun `parameters_plan.md` previously listed as something to remove.

A borrow-based execution guard won't work here, since ODBC returns to the application between the execute and the drain that carries the `@handle`. A shape that does: give `PreparedStatement` an `Arc<Mutex<Option<SessionHandle>>>` and have `execute_prepared` register a clone of it as the client's pending sink (replacing `expecting_prepare_handle`). `push_return_value` (`1873`) is already the single funnel for every RETURNVALUE, so it can write the handle — stamped with the session identity current at that moment — straight into the statement, with no caller step at all. Clear the sink in `begin_command`, on the reconnect reset (`341-343`), and in `close_query` (`3327`). That removes `capture_prepared_handle_into` and all three call sites, and it also means a failure after the server allocated a handle still records it instead of leaking the plan. Worth dropping the `Clone` derive (`3716`) at the same time, since a shared cell makes clones alias one server plan.

**8. A live orphan is silently discarded on the reuse path.** `plan_prepared_execution` ignores `orphaned` when the cached handle is live. The ODBC layer maintains an invariant that makes this unreachable, but for any other consumer the API accepts a still-live handle and drops it without releasing it. Either document the precondition on `execute_prepared` or release the orphan on that path.

---

### Description and docs

A few things that should be reconciled before merge:

- The description says the low-level calls "no longer self-reconnect"; `execute_sp_prepare_raw` still does (`tds_client.rs:1300`).
- The description doesn't mention that the `_raw` methods remain public, which is material to the layering story it presents.
- "New/changed functionality has tests" is checked, while the E2E test for the core scenario is skipped and the live test is CI-ignored. Worth calling that out explicitly instead.
- `mssql-odbc/docs/parameters_plan.md` still describes the ODBC layer performing recovery and calling `plan_execution`; that's the superseded design.
- The Related Issues section still contains the template placeholder `Fixes #<ISSUE_NUMBER>` and its commented instructions.
- Unconditional `sp_prepexec` recovery is presented as the intended design without noting the plan-shape implications from item 3.

---

### Verification

Built and ran against `aad5f904`:

- `cargo test -p mssql-odbc --lib` — 450 passed.
- `cargo test -p mssql-tds --lib` — 1595 passed, 4 failed. The failures are `certificate_validator` tests missing local fixture files, unrelated to this change.

The new unit tests are genuinely good — the wire-byte assertions in `unprepare_releases_live_handle` and the `plan_prepared_execution` matrix are exactly the right level of specificity. The gaps above are about the paths those tests don't reach.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1437,11 +1499,97 @@ impl TdsClient {
...
+    /// [`next_row`](ResultSet::next_row) / [`advance`](Self::advance).
+    pub async fn execute_prepared<'a>(
+        &mut self,
+        statement: &mut PreparedStatement,
+        named_params: Vec<RpcParameter>,
+        orphaned: Option<PreparedHandle>,
+        options: impl Into<ExecuteOptions<'a>>,
+    ) -> TdsResult<StatementResult> {
+        let mut opts = options.into();
+        let reconnect_elapsed = self.check_and_reconnect(opts.timeout, opts.cancel).await?;
```
saurabh500, line 1671: `check_and_reconnect` now runs *before* `begin_command()`, which inverts the invariant this file documents on `begin_command` (~:3263):

> Call this at the top of every public token-consuming command, before consuming the response (**and before `check_and_reconnect` where applicable**): a transparent reconnect repopulates login messages for the new session *after* this point, so those remain visible as part of the command that triggered the reconnect.

Every other path honors it — `execute` (:707), `execute_sp_executesql` (:1174), `execute_sp_prepare_raw` (:1299) and `begin_transaction` (:3415) all do `begin_command()` and then `check_and_reconnect`. Here the order is reversed: `execute_prepared` reconnects, the new session's login info messages land in `self.info_messages`, and then the raw call's `begin_command()` immediately clears them. So the one command that actually triggered the reconnect is the one that loses the reconnect's messages — the application never sees them through `info_messages()` / `SQLGetDiagRec`.

Suggested fix — move the command boundary into the managed method, ahead of recovery:

```rust
pub async fn execute_prepared<'a>(...) -> TdsResult<StatementResult> {
    let mut opts = options.into();

    // Command boundary must precede recovery: a reconnect repopulates
    // info_messages for the new session, and those belong to this command.
    self.begin_command();

    let reconnect_elapsed = self.check_and_reconnect(opts.timeout, opts.cancel).await?;
    // ...
}
```

That requires the inner `begin_command()` not to clear a second time. Two ways:

- **(a)** make it idempotent within a command — a `command_started` flag set by `begin_command` and cleared in `close_query` / on error;
- **(b)** split each raw body into the public `*_raw` (calls `begin_command`) and an inner `*_inner` that assumes the boundary is already open, with `execute_prepared` calling the inner form.

(b) is more code but keeps the invariant verifiable by reading one function instead of reasoning about call order across four.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1437,11 +1499,97 @@ impl TdsClient {
...
+        opts.timeout = Self::deduct_timeout(opts.timeout, reconnect_elapsed);
```
saurabh500, line 1672: `deduct_timeout` saturates to `Some(0)` meaning *budget exhausted*, but `timeout_to_duration` maps `Some(0)` to `None` meaning *no timeout* — so a reconnect that consumes the whole command budget produces an **unbounded** execution, the opposite of the intent. `opts.timeout` flows straight from here into the raw calls, which do exactly that conversion (`self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec)`).

The PR documents the hazard carefully in both helpers and notes it's latent because the ODBC layer passes `None`. I'd still fix it here rather than defer: `execute_prepared` is now the single consolidated producer of that `Some(0)`, and it's the function `SQL_ATTR_QUERY_TIMEOUT` will be wired into. Leaving a documented-but-live trap in the exact function a future PR will plumb user timeouts through is asking for it to be found in the field.

Suggested fix:

```rust
let reconnect_elapsed = self.check_and_reconnect(opts.timeout, opts.cancel).await?;
opts.timeout = Self::deduct_timeout(opts.timeout, reconnect_elapsed);
if opts.timeout == Some(0) {
    return Err(Error::Timeout(
        "command timeout exhausted by connection recovery".into(),
    ));
}
```

(Whatever the crate's canonical timeout error is — the point is failing fast instead of falling through into the `0 == infinite` convention.) Worth a unit test that a `Some(n)` budget fully consumed by recovery errors rather than executing.

Longer term, the underlying footgun is that two adjacent helpers assign opposite meanings to `Some(0)`. A small newtype (`enum Budget { None, Remaining(NonZeroU32), Exhausted }`) would make the mistake unrepresentable, but that's a follow-up, not this PR.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1437,11 +1499,97 @@ impl TdsClient {
...
+    /// Returns positioned on the first result; drain rows with
+    /// [`next_row`](ResultSet::next_row) / [`advance`](Self::advance).
+    pub async fn execute_prepared<'a>(
+        &mut self,
+        statement: &mut PreparedStatement,
+        named_params: Vec<RpcParameter>,
+        orphaned: Option<PreparedHandle>,
+        options: impl Into<ExecuteOptions<'a>>,
+    ) -> TdsResult<StatementResult> {
+        let mut opts = options.into();
```
saurabh500, line 1670: Minor, but the error the caller sees changes here. The raw methods check `has_open_batch()` and return `UsageError(ALREADY_EXECUTING_ERROR)` *before* touching the connection. `execute_prepared` reconnects first, so with a batch already open on a dead socket, `is_recovery_possible()` returns `false` and the caller now gets `ConnectionClosed` instead of the "already executing" usage error — a caller-programming-error reported as a connection failure.

Suggested fix — hoist the guard so the cheap, purely-local check happens before any I/O:

```rust
) -> TdsResult<StatementResult> {
    if self.execution_context.has_open_batch() {
        return Err(UsageError(ALREADY_EXECUTING_ERROR.to_string()));
    }

    let mut opts = options.into();
    // ...
```

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1456,8 +1604,21 @@ impl TdsClient {
...
+    ///
+    /// This method never reconnects. The expected flow is: the caller (1)
+    /// recovers the connection once with
+    /// [`check_and_reconnect`](Self::check_and_reconnect), (2) checks — via
+    /// [`connection_recovery_count`](Self::connection_recovery_count) — whether a
+    /// cached handle still belongs to the live session (passing it as
+    /// `drop_handle` only if so), then (3) calls this. Because the decision is
+    /// made just before the send and this method won't reconnect underneath it,
+    /// it can't prepare against a session the caller didn't account for.
+    /// `_raw`: a low-level wire call with caller-owned recovery — prefer the
```
saurabh500: The `_raw` suffix is being asked to carry a contract that a name can't hold, and it costs a breaking public rename to do it. Four concerns, then a concrete alternative.

**1. There are no production callers to warn.** Every call site outside this file is a test or a bench:

| Consumer | `_raw` calls |
|---|---|
| `mssql-tds/tests/*` | 28 |
| `mssql-tds-bench` | 5 |
| `mssql-odbc` / `mssql-js` / `mssql-py-core` | **0** |

So the rename churns the public API for an audience of one — this crate's own test suite — while leaving the actual footgun (`pub`) fully loaded. If nobody outside the crate should be calling these, the fix is visibility, not spelling.

**2. `raw` already means something else here.** Elsewhere in the crate — and in Rust generally (`from_raw`, `as_raw_fd`) — it means *unwrapped representation*:

- `cursor_types.rs:204` `pub fn from_raw(value: i32)`
- `sql_string.rs:124` `pub fn as_raw_wire_bytes(&self)`
- `bulk_load.rs:392` `pub async fn write_raw_bytes(..)`

Here it means *unmanaged lifecycle / caller-owned recovery*. Two incompatible senses of one word in one crate. Nothing about these calls is raw — they build typed RPCs, parse tokens, populate AE metadata, and return `StatementResult`.

**3. A suffix can't express a precondition.** What actually differs is a three-step protocol the caller must follow. The doc block right above needs ~10 lines of prose to say so, ending in "prefer the managed `execute_prepared`". If the name were doing the work, the paragraph wouldn't be needed. (Also `execute_` + `sp_prepare` + `_raw` is triply redundant.)

**4. It doesn't partition consistently.** `execute_sp_executesql` was not renamed despite being equally a direct wire call with caller-owned recovery. So "no suffix" doesn't mean "managed" — it means "four of the five got renamed," and a reader can't infer the rule.

---

**Suggested fix — use the `test-util` pattern already in this crate.** It exists precisely for "tests need it, shipped API shouldn't have it," and this PR already uses it correctly 2,000 lines down for `PreparedStatement::materialized_for_test` (:3751):

```rust
pub(crate) async fn execute_sp_prepexec(..) -> TdsResult<StatementResult> { .. }

#[cfg(any(test, feature = "test-util"))]
#[doc(hidden)]
pub async fn execute_sp_prepexec(..) -> TdsResult<StatementResult> { .. }
```

Benefits over the suffix: the constraint is enforced by the compiler rather than by convention; the public surface *shrinks* instead of growing; it disappears from rustdoc and from release builds (`resolver = "2"` keeps dev-dep features out of the normal graph); and the 33 test/bench call sites revert to their original names, so the breaking rename goes away entirely.

Precedent, for reference: `test-util` was added in bfe2e899 (#83) and extended across the crate boundary to `mssql-odbc`'s dev-dependencies in bb02ba65 (#143) for exactly this purpose. One caveat — `mssql-tds-bench` would need `features = ["test-util"]` on its `mssql-tds` dependency.

**If they must stay public**, I'd drop the suffix and put the caveat in the path so it can't be misread — `client.raw().sp_prepexec(..)` via a `RawRpc<'_>` wrapper. One place to document the contract, applies uniformly, and `execute_sp_executesql` can join without another rename.

**Minimum, if the suffix stays**: apply it to `execute_sp_executesql` too, and ship `#[deprecated(note = "renamed to ..._raw")]` shims for one release instead of a hard break.

---

**Related — `check_and_reconnect` probably shouldn't be `pub` either.** External callers: zero. Intra-crate: 10 in `cursor_ops.rs` (so `pub(crate)` covers them) plus one integration test (covered by the same `test-util` gate). It was made public so the ODBC layer could own recovery, but aad5f904 moved that logic into `execute_prepared` — the motivating caller no longer exists. `parameters_plan.md:64` still documents the superseded rationale ("`check_and_reconnect()` is now public so the caller can own the single recovery point") and should be updated either way.

mssql-odbc/src/handles/stmt.rs
```diff
@@ -51,24 +52,26 @@ pub(crate) struct StmtState {
...
-    pub(crate) prepared_sql: Option<String>,
+    /// The logical prepared statement stored by `SQLPrepare` (SQL text plus the
+    /// server handle once materialized). The server-side prepare is deferred to
+    /// `SQLExecute`. `Some` marks the statement as prepared; its
+    /// [`session_handle`](PreparedStatement::session_handle) is filled after the
+    /// first execute.
+    pub(crate) prepared_stmt: Option<PreparedStatement>,
+    /// Number of parameter markers in `prepared_stmt`, computed once at prepare
+    /// so `SQLExecute` builds the `@P1..@Pn` list without re-scanning the SQL.
+    pub(crate) prepared_marker_count: usize,
```
saurabh500: `prepared_marker_count` re-introduces the split state this PR just removed. The whole point of bundling SQL + handle into `PreparedStatement` is to have one owner; this adds a second field that must stay in lockstep with it, with nothing enforcing that.

Four sites now have to update both together — `prepare.rs:120-121`, `exec_direct.rs:115-116`, `get_type_info.rs:155-156`, and `StmtHandle::new`. Miss one and `SQLExecute` builds the wrong `@P1..@Pn` list at `execute.rs:161`: wrong parameter count, silently, no error. The PR's own test helper already tripped over this — `set_prepared` had to be updated to set both.

The marker count is a property of the prepared SQL, so it belongs with it. This is also the same gap behind @David-Engel's blocker #3: his proposed `PreparedStatement` carries `parameter_declarations` because the tds-side type is currently under-specified about what's needed to re-create the statement, so consumers bolt the missing pieces on alongside it.

Suggested fix, in `mssql-tds`:

```rust
pub struct PreparedStatement {
    sql: String,
    parameter_declarations: String, // or marker_count: usize
    session_handle: Option<PreparedHandle>,
}

impl PreparedStatement {
    pub fn new(sql: impl Into<String>, parameter_declarations: impl Into<String>) -> Self { /* ... */ }
    pub fn marker_count(&self) -> usize { /* ... */ }
}
```

`prepared_marker_count` then disappears from `StmtState` and the four sync points collapse to zero.

If you'd rather not widen the tds type in this PR, the local stopgap is a small struct so the two can only be set together:

```rust
pub(crate) struct PreparedPlan {
    stmt: PreparedStatement,
    marker_count: usize,
}

pub(crate) prepared: Option<PreparedPlan>,
```

mssql-odbc/src/handles/stmt.rs
```diff
@@ -112,18 +115,25 @@ impl StmtState {
...
+        };
+        let Some(handle) = prepared.take_session_handle() else {
+            // No materialized handle to release; the statement stays in place.
+            return;
+        };
+        debug_assert!(
+            self.pending_unprepare.is_none(),
+            "orphan_prepared_handle: a pending unprepare already exists"
+        );
+        self.pending_unprepare = Some(handle);
```
saurabh500: `debug_assert!` compiles out in release, and the next line unconditionally overwrites `self.pending_unprepare`. If the invariant is ever violated, the previously pending handle is dropped on the floor — a server-side plan that stays allocated until disconnect, with no error and no log entry to find it by.

This shape predates the PR, so I'm not treating it as introduced here. But it's worth fixing now for two reasons: the function was rewritten in this PR anyway, and the invariant it guards has just become load-bearing **across a crate boundary** — it's what makes @David-Engel's item #8 (live orphan silently discarded on the reuse path) unreachable from ODBC. A release-build violation used to leak one handle; now it can also hand `plan_prepared_execution` a state it doesn't handle.

```rust
if let Some(previous) = self.pending_unprepare.replace(handle) {
    debug_assert!(false, "orphan_prepared_handle: a pending unprepare already exists");
    error!(
        id = previous.id(),
        "orphan_prepared_handle: overwriting a pending unprepare — handle leaked until disconnect"
    );
}
```

Cheap, and it turns a silent leak into something greppable in a customer log.

mssql-odbc/src/handles/stmt.rs
```diff
@@ -51,24 +52,26 @@ pub(crate) struct StmtState {
...
-    /// execute, which flushes any pending drop first).
-    pub(crate) pending_unprepare: Option<i32>,
+    /// The live server handle of a prepared statement superseded by a
+    /// re-prepare / rebind / `SQLExecDirect`, awaiting release with
+    /// `sp_unprepare`. The drop is deferred to the next point that already
+    /// holds the TDS client (execute / exec-direct) or to statement free, so
+    /// bind/prepare stay I/O-free. Invariant: this is `None` whenever
+    /// `prepared_stmt` holds a live handle (a new handle can only be acquired by
+    /// an execute, which flushes any pending drop first).
+    pub(crate) pending_unprepare: Option<PreparedHandle>,
```
saurabh500: > Invariant: this is `None` whenever `prepared_stmt` holds a live handle

Worth cross-referencing, because this invariant is now load-bearing in a different crate. It's exactly the precondition `plan_prepared_execution` relies on to be correct — it's what makes the "live orphan silently discarded on the reuse path" case unreachable from ODBC, which is the reason @David-Engel's item #8 is a latent API hazard rather than a live bug today.

Right now it's stated only here. Suggest a pointer in both directions: note here that `TdsClient::execute_prepared` depends on it, and state it as an explicit precondition in the doc comment on `execute_prepared` (plus the `debug_assert!` suggested there), so a future change on either side can't quietly invalidate the other.

mssql-tds/src/connection/tds_client.rs
Theekshna: Fixed in `da8e70db`. I followed the existing `test-util` pattern:

- The four prepared RPC implementations are private and restored to their original unsuffixed internal names.
- Integration tests and `mssql-tds-bench` use `#[doc(hidden)]` `*_for_test` wrappers gated by `test-util`.
- The benchmark explicitly enables `test-util`.
- `check_and_reconnect` is private and has a test-only wrapper for the connectivity test.
- The superseded public-API rationale was removed from `parameters_plan.md`.

This removes the lifecycle-sensitive methods from production builds and rustdoc while retaining protocol-level test and benchmark coverage.

Theekshna:
> Thanks for reworking this — moving the recovery protocol into `mssql-tds` is the right direction, and `PreparedStatement` / `PreparedHandle` is a much better boundary than making every caller track epochs. The epoch-planning logic itself is clean and well unit-tested.
> 
> I'm requesting changes: there are three correctness issues in the new managed path, plus a validation gap that means the scenario this PR targets isn't currently demonstrated by any enabled test.
> 
> Reviewed at `aad5f9047c8139a24b83930a10bf29a98a830476`.
> 
> ### Blocking
> **1. Managed prepared statements break on the second execution under Always Encrypted**
> 
> `execute_sp_prepexec_raw` encrypts parameters (`tds_client.rs:1675`) but never caches the resulting describe metadata under the handle the server returns. Only `execute_sp_prepare_raw` populates `prepared_param_encryption` (`tds_client.rs:1407`), and `execute_sp_execute_raw` requires that entry (`tds_client.rs:1809`).
> 
> Since `execute_prepared` materializes every statement via `sp_prepexec` and then reuses it via `sp_execute`, this fails deterministically whenever column encryption is active and the statement has parameters:
> 
> 1. First `execute_prepared` → `sp_prepexec` succeeds, handle captured.
> 2. Second `execute_prepared` → `Reuse` → `sp_execute`.
> 3. Lookup misses → `ColumnEncryptionError` ("has no Always Encrypted parameter metadata... prepare it with execute_sp_prepare on this connection").
> 
> No reconnect is needed to hit this. The metadata gathered during `sp_prepexec` needs to be associated with the handle once the `@handle` RETURNVALUE is known, and this path needs an AE test that executes the same prepared statement twice.
> 
> **2. `session_epoch` can't distinguish handles from different connections**
> 
> `PreparedHandle` carries only a per-client recovery count (`tds_client.rs:3691`), and `plan_prepared_execution` compares only that count (`tds_client.rs:3779`). `PreparedStatement` is a public, `Clone` type with no tie to the client that materialized it.
> 
> A statement materialized on client A at epoch 0 and then passed to client B at epoch 0 is classified `Reuse`, so B issues `sp_execute` with a handle id that belongs to a different session — the exact aliasing hazard the PR sets out to eliminate, just reached by a different route. `unprepare()` has the same exposure and can release an unrelated plan.
> 
> The handle needs an identity that is unique per session (for example a monotonically-issued connection id combined with the recovery count), or the API needs to prevent a statement from being used with a client other than its owner.
> 
> **3. Re-preparation doesn't preserve how the statement was originally prepared**
> 
> `PreparedStatement` stores only SQL and an optional handle (`tds_client.rs:3717`), and `execute_prepared` routes every absent-or-stale statement through `sp_prepexec` (`tds_client.rs:1542-1547`).
> 
> `sp_prepare` compiles without execution values; `sp_prepexec` compiles with the supplied values. Because of parameter sniffing, they can produce different plans. A statement prepared via `sp_prepare` and later recovered via `sp_prepexec` can therefore come back with different performance characteristics after a reconnect — a silent behavior change rather than a visible failure, which makes it hard to diagnose in the field.
> 
> Worth recording the preparation strategy (and the parameter declarations needed to reproduce it) on the statement, then recovering with the matching sequence: `sp_prepare` + `sp_execute` for statements prepared that way, `sp_prepexec` for the combined form. Related: there is currently no managed entry point corresponding to `sp_prepare`, so that strategy isn't reachable through the new API at all.
> 
> ### Should fix before merge
> **4. A pending orphaned handle is lost when execution fails**
> 
> `stage_execution` takes `pending_unprepare` out of `StmtState` (`execute.rs:170`). It's restored when the connection claim fails (`execute.rs:80`), but the `execute_prepared` error path only restores `prepared_stmt` (`execute.rs:104-106`) before returning via `fail_with_tds` (`execute.rs:112`).
> 
> If the call fails before the piggybacked drop reaches the server (reconnect failure, encryption failure, serialization error), the driver forgets a still-live server-side plan and it stays allocated until disconnect. Restoring the orphan on that path — or having the API report whether the drop was consumed — would close this.
> 
> **5. The scenario this PR fixes has no enabled test**
> 
> The E2E test unconditionally skips in `SetUp` (`session_recovery_test.cpp:25`), and the live TDS test drives the raw APIs to assert that a stale handle _fails_ with 8179 (`connectivity.rs:546-589`) rather than exercising `execute_prepared`.
> 
> The unit tests cover `plan_prepared_execution` and handle capture well, but nothing enabled demonstrates the full managed loop: reconnect → detect stale → re-prepare → capture new handle → reuse successfully. Given the description notes transparent reconnect isn't working yet in `mssql-tds`, it's worth being explicit about whether this lands ahead of that fix, and if so what evidence we have that the new path behaves as intended.
> 
> **6. The `_raw` methods are still public**
> 
> `execute_sp_prepare_raw` (1282), `execute_sp_unprepare_raw` (1441), `execute_sp_prepexec_raw` (1621), and `execute_sp_execute_raw` (1763) remain `pub`. Renaming signals intent, but consumers can still take the bare-handle path and reintroduce the class of bug this PR fixes.
> 
> Making them `pub(crate)` (or private) would make the boundary real. The blocker is the tests and benchmark that call them from outside the crate; protocol-level coverage could move into the crate where private access is available.
> 
> There's also an internal inconsistency: `execute_sp_prepare_raw` is documented as "never reconnects / caller-owned recovery" but still calls `check_and_reconnect` at `tds_client.rs:1300`.
> 
> ### Smaller items
> **7. Handle capture is still a two-step contract.** `execute_prepared` (`tds_client.rs:1525`) doesn't finish updating the statement — callers must drain, then remember `capture_prepared_handle_into` (`1558`). The ODBC layer already needs this at three separate places (`close_cursor.rs:216`, `exec_common.rs:303`, `more_results.rs:185`), and a drain path added later would silently drop handles. That's protocol ordering leaking through the managed API, and it's the footgun `parameters_plan.md` previously listed as something to remove.
> 
> A borrow-based execution guard won't work here, since ODBC returns to the application between the execute and the drain that carries the `@handle`. A shape that does: give `PreparedStatement` an `Arc<Mutex<Option<SessionHandle>>>` and have `execute_prepared` register a clone of it as the client's pending sink (replacing `expecting_prepare_handle`). `push_return_value` (`1873`) is already the single funnel for every RETURNVALUE, so it can write the handle — stamped with the session identity current at that moment — straight into the statement, with no caller step at all. Clear the sink in `begin_command`, on the reconnect reset (`341-343`), and in `close_query` (`3327`). That removes `capture_prepared_handle_into` and all three call sites, and it also means a failure after the server allocated a handle still records it instead of leaking the plan. Worth dropping the `Clone` derive (`3716`) at the same time, since a shared cell makes clones alias one server plan.
> 
> **8. A live orphan is silently discarded on the reuse path.** `plan_prepared_execution` ignores `orphaned` when the cached handle is live. The ODBC layer maintains an invariant that makes this unreachable, but for any other consumer the API accepts a still-live handle and drops it without releasing it. Either document the precondition on `execute_prepared` or release the orphan on that path.
> 
> ### Description and docs
> A few things that should be reconciled before merge:
> 
> * The description says the low-level calls "no longer self-reconnect"; `execute_sp_prepare_raw` still does (`tds_client.rs:1300`).
> * The description doesn't mention that the `_raw` methods remain public, which is material to the layering story it presents.
> * "New/changed functionality has tests" is checked, while the E2E test for the core scenario is skipped and the live test is CI-ignored. Worth calling that out explicitly instead.
> * `mssql-odbc/docs/parameters_plan.md` still describes the ODBC layer performing recovery and calling `plan_execution`; that's the superseded design.
> * The Related Issues section still contains the template placeholder `Fixes #<ISSUE_NUMBER>` and its commented instructions.
> * Unconditional `sp_prepexec` recovery is presented as the intended design without noting the plan-shape implications from item 3.
> 
> ### Verification
> Built and ran against `aad5f904`:
> 
> * `cargo test -p mssql-odbc --lib` — 450 passed.
> * `cargo test -p mssql-tds --lib` — 1595 passed, 4 failed. The failures are `certificate_validator` tests missing local fixture files, unrelated to this change.
> 
> The new unit tests are genuinely good — the wire-byte assertions in `unprepare_releases_live_handle` and the `plan_prepared_execution` matrix are exactly the right level of specificity. The gaps above are about the paths those tests don't reach.



Thanks for the detailed review. Replying with the same numbering:

1. **Fixed.** `sp_prepexec` now retains its Always Encrypted describe metadata and pins it under the returned handle. Replacement, same-ID reuse, failure cleanup, and metadata pinning have focused tests.

2. **Tracked separately.** ODBC cannot move a statement between clients because an HSTMT belongs to one HDBC, but direct `mssql-tds` consumers can. Removing `Clone` would not prevent moving the statement. Robust client/session identity is tracked in [ADO 47098](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47098).

3. **No strategy switch exists in the current managed path.** Statements are initially materialized and recovered through `sp_prepexec`. A future managed `sp_prepare` path must retain its strategy and declarations so it can recover through `sp_prepare` + `sp_execute`.

4. **Fixed.** The orphan is now passed as `&mut Option<PreparedHandle>` and retained through reconnect, validation, parameter construction, and AE setup. It is consumed only at the serialization boundary. Later failures remain ambiguous because the server may have consumed it, matching msodbcsql. Both ownership outcomes have focused tests.

5. **Tracked, but not enabled yet.** The managed reconnect E2E remains skipped pending the underlying session-recovery fix in [ADO 46631](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46631). Enabling the complete reconnect → reprepare → capture → reuse test is tracked in [ADO 47099](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47099). This limitation is now documented explicitly.

6. **Fixed.** The four prepared RPC methods are now private implementation methods. Integration tests and the benchmark use `#[doc(hidden)]` `*_for_test` wrappers available only under the existing `test-util` feature. `check_and_reconnect` is private as well, with the same test-only wrapper pattern. These APIs no longer appear in production builds or rustdoc.

7. **Deferred.** I did not add an `Arc<Mutex<...>>` completion sink because it introduces clone aliasing, stale delivery, cancellation, and lock-order concerns. Explicit post-drain capture remains deterministic under ODBC's STMT/HDBC serialization. I audited close, advance, fetch, row-writer, and PLP paths so capture state is consistently completed or cleared.

8. **Fixed through an explicit contract.** `execute_prepared` now documents that an orphan may be supplied only when the statement has no live handle. ODBC maintains this invariant by storing a superseded handle only alongside an unmaterialized statement.

I also updated `parameters_plan.md` with the managed ownership model, AE lifecycle, orphan boundary, private raw API status, and tracked follow-ups.

mssql-tds/src/connection/tds_client.rs
Theekshna: Fixed in 6924c28a. `execute_prepared` (and `unprepare`) now call `begin_command()` before `check_and_reconnect`, and the inner RPCs take a `command_started: bool` — when `true` they skip both `begin_command` and the re-entrancy guard, so the reconnect's login/info messages land in this command's buffer and aren't cleared underneath it. I went with option (a) (the flag) over the `*_inner` split to keep churn down; the invariant is now "the managed entry point owns the single `begin_command`, and `command_started` just means don't reopen the boundary," and both managed entry points read identically (guard → `begin_command` → recover).

mssql-tds/src/connection/tds_client.rs
Theekshna: Fixed in 6924c28a — hoisted the `has_open_batch()` guard to the top of `execute_prepared`, before `check_and_reconnect`, so a re-entrant call on a dead socket now returns the `ALREADY_EXECUTING` `UsageError` before any I/O instead of `is_recovery_possible()` turning it into `ConnectionClosed`. Covered by `execute_prepared_open_batch_returns_usage_error_and_keeps_orphan` and `execute_prepared_open_batch_guard_is_epoch_agnostic`.

mssql-tds/src/connection/tds_client.rs
Theekshna: Fixed — `execute_prepared` now fails fast with a `TimeoutError` when the post-`deduct_timeout` budget is `Some(0)`, instead of falling through the `Some(0) == infinite` convention; I applied the same guard to `unprepare` (after the epoch skip, since a recovery that advanced the epoch means the handle is already gone and a timeout there would be misleading). Covered by `execute_prepared_errors_when_budget_exhausted_by_recovery` and `unprepare_errors_when_budget_exhausted_before_send` (the harness can't simulate a timed reconnect, so they drive the same `Some(0)` state via a zero budget). Filed the `Budget` newtype follow-up — which also extends the check to the other execute paths sharing this pattern — as [AB#47211](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47211).

mssql-odbc/src/handles/stmt.rs
Theekshna: Fixed with the local bundling struct — `StmtState` now holds a single `prepared: Option<PreparedPlan>` where `PreparedPlan { stmt, marker_count }` can only be constructed with both, so the four hand-synced sites collapse to zero and `SQLExecute` can't build the wrong `@P1..@Pn` count. I kept `marker_count` in the ODBC layer rather than widening the tds `PreparedStatement`, since it's a byproduct of the ODBC `?`→`@Pn` rewrite that `execute_prepared` never needs; David's `parameter_declarations` (blocker #3) is a genuine TDS concept and better handled as its own tds-side change than conflated with the marker count here.

mssql-odbc/src/handles/stmt.rs
Theekshna: Fixed — `orphan_prepared_handle` now uses `pending_unprepare.replace(handle)` and, on a release-build invariant violation, logs an `error!` with the leaked handle id (plus the `debug_assert!(false, …)`), so the silent drop becomes greppable.

mssql-odbc/src/handles/stmt.rs
Theekshna: Added the pointer in both directions: `StmtState::pending_unprepare`'s doc now notes that `mssql_tds::TdsClient::execute_prepared` relies on the invariant (it's what keeps a live orphan from being discarded on the `sp_execute` reuse path, via `plan_prepared_execution`), and `execute_prepared` gained a `# Precondition` section stating `orphaned = Some` only when the statement has no live handle, that `plan_prepared_execution` depends on it, and that the ODBC layer upholds it through `pending_unprepare` / `orphan_prepared_handle`. Used plain backticks rather than intra-doc links since `mssql-tds` can't link into `mssql-odbc` and `plan_prepared_execution` is private.

David-Engel (dismissed):
Re-reviewed at `ffe125be`. The fixes here are solid, and two are better than what I suggested: pinning the AE describe inside `push_return_value` (so it completes in the token funnel regardless of drain path), and evicting the dropped handle's entry *before* inserting the new one because the server can reuse the integer id — that's a real aliasing bug you anticipated and I didn't. `&mut Option<PreparedHandle>` consumed at the serialization boundary is also a cleaner answer to the orphan problem than either option I offered. Items 1, 4, and 6 are resolved, and I'm satisfied with the pushback on item 3: with no managed `sp_prepare` path, statements are materialized and recovered the same way, so there's no divergence today.

I do want to push back on the disposition of items 2, 7, and 8, because I think they're one problem rather than three, and the fix collapses all of them.

### The root cause

The handle's destination lives outside the client. `execute_prepared` borrows `&mut PreparedStatement` for the duration of the call, but the `@handle` RETURNVALUE arrives after the result set — often in a later FFI call entirely. So the client has the value and no way to reach the place it belongs, and everything downstream follows from that:

- **#7** exists because the caller has to close that gap manually (three call sites in ODBC today).
- **#2** exists because the handle therefore has to travel *with* the statement, carrying a session epoch that's only meaningful relative to one client.
- **#8** exists because `plan_prepared_execution` has to reason about two externally-supplied handles whose validity it can't verify.

Note that the AE fix in this PR is already internal capture, and it works for exactly one reason: `prepared_param_encryption` is a field on the client, so `push_return_value` can write to it with no borrow at all. The handle isn't special — its destination is just on the wrong side of the boundary.

### Suggested shape

Move the destination inside the client and give statements an opaque, client-issued identity:

```rust
/// Opaque, client-scoped identity for a materialized prepared statement.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub struct StatementId(u64);

pub struct PreparedStatement {
    sql: String,
    id: Option<StatementId>,
}

// TdsClient
next_statement_id: u64,
prepared_handles: HashMap<StatementId, i32>,
pending_capture: Option<StatementId>,
```

`execute_prepared` assigns an id if absent and records it in `pending_capture`; `push_return_value` writes `id -> handle` into the client's own map when the token lands — the same mechanism as the AE metadata, and it needs no `Arc`, no `Mutex`, and no cross-thread delivery, which I think addresses the concerns you raised against my earlier sink proposal.

What falls out:

- **`capture_prepared_handle_into` disappears**, along with all three ODBC call sites and the "call after drain" ordering rule.
- **The epoch disappears.** The map is cleared on reconnect exactly as `prepared_param_encryption` already is (`tds_client.rs:352`), so "stale" becomes "absent from the map." `plan_prepared_execution` becomes a lookup.
- **AB#47098 is closed structurally.** A statement from another client resolves to an id this client has never issued, so it's simply unmaterialized and gets re-prepared. No aliasing, no documentation to read.
- **#8 becomes unrepresentable.** With orphans identified the same way, the client resolves both sides itself instead of trusting a caller-supplied precondition.
- **`PreparedHandle` leaves the public API entirely**, which is what the `PreparedStatement` doc already claims ("callers never touch raw handle ids or epochs") but the current shape doesn't deliver — `id()` and `session_epoch()` are `pub`, ODBC stores `PreparedHandle` in `pending_unprepare`, and `exec_direct.rs`'s tests assert on both accessors.
- **The eviction-before-insert hazard stops existing.** That code is only necessary because the AE map is keyed by a server-reusable integer; keyed by `StatementId`, a fresh statement can never collide with a dropped one.

Costs, to be fair: one hash lookup per execute, which is noise next to a round trip; and a map entry lingering if a statement is dropped without unprepare, which is bounded and cleared on reconnect — no worse than the AE map's behavior today.

### On timing

I'd rather see this in this PR than tracked, for one reason: this is the public API shape, and it has exactly one consumer today. Every later adopter makes it more expensive, and the follow-up items currently filed against it (AB#47098, plus whatever #7 becomes) are all downstream of a shape we can still change cheaply. The blast radius is the same set of files this PR already rewrote.

### Still needed

The description no longer matches the code. It says the low-level calls are "renamed with a `_raw` suffix," documented as caller-owned recovery, and that `check_and_reconnect()` is now `pub` — `da8e70db` reversed all three. It also doesn't mention the AE metadata fix, orphan retention, or the timeout fast-fail, which are the most substantive changes since the last review.

### Verification at `ffe125be`

- `cargo test -p mssql-odbc --lib` — 450 passed.
- `cargo test -p mssql-tds --lib` — 1611 passed (up from 1595), same 4 pre-existing `certificate_validator` fixture failures, unrelated to this PR.
- The new coverage around orphan preservation, budget exhaustion, and AE pinning is good, and `abort_pending_prepare_capture` being wired into every failure path is real risk reduction for the current shape.

Theekshna:
> Re-reviewed at `ffe125be`. The fixes here are solid, and two are better than what I suggested: pinning the AE describe inside `push_return_value` (so it completes in the token funnel regardless of drain path), and evicting the dropped handle's entry _before_ inserting the new one because the server can reuse the integer id — that's a real aliasing bug you anticipated and I didn't. `&mut Option<PreparedHandle>` consumed at the serialization boundary is also a cleaner answer to the orphan problem than either option I offered. Items 1, 4, and 6 are resolved, and I'm satisfied with the pushback on item 3: with no managed `sp_prepare` path, statements are materialized and recovered the same way, so there's no divergence today.
> 
> I do want to push back on the disposition of items 2, 7, and 8, because I think they're one problem rather than three, and the fix collapses all of them.
> 
> ### The root cause
> The handle's destination lives outside the client. `execute_prepared` borrows `&mut PreparedStatement` for the duration of the call, but the `@handle` RETURNVALUE arrives after the result set — often in a later FFI call entirely. So the client has the value and no way to reach the place it belongs, and everything downstream follows from that:
> 
> * **#7** exists because the caller has to close that gap manually (three call sites in ODBC today).
> * **#2** exists because the handle therefore has to travel _with_ the statement, carrying a session epoch that's only meaningful relative to one client.
> * **#8** exists because `plan_prepared_execution` has to reason about two externally-supplied handles whose validity it can't verify.
> 
> Note that the AE fix in this PR is already internal capture, and it works for exactly one reason: `prepared_param_encryption` is a field on the client, so `push_return_value` can write to it with no borrow at all. The handle isn't special — its destination is just on the wrong side of the boundary.
> 
> ### Suggested shape
> Move the destination inside the client and give statements an opaque, client-issued identity:
> 
> ```rust
> /// Opaque, client-scoped identity for a materialized prepared statement.
> #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
> pub struct StatementId(u64);
> 
> pub struct PreparedStatement {
>     sql: String,
>     id: Option<StatementId>,
> }
> 
> // TdsClient
> next_statement_id: u64,
> prepared_handles: HashMap<StatementId, i32>,
> pending_capture: Option<StatementId>,
> ```
> 
> `execute_prepared` assigns an id if absent and records it in `pending_capture`; `push_return_value` writes `id -> handle` into the client's own map when the token lands — the same mechanism as the AE metadata, and it needs no `Arc`, no `Mutex`, and no cross-thread delivery, which I think addresses the concerns you raised against my earlier sink proposal.
> 
> What falls out:
> 
> * **`capture_prepared_handle_into` disappears**, along with all three ODBC call sites and the "call after drain" ordering rule.
> * **The epoch disappears.** The map is cleared on reconnect exactly as `prepared_param_encryption` already is (`tds_client.rs:352`), so "stale" becomes "absent from the map." `plan_prepared_execution` becomes a lookup.
> * **AB#47098 is closed structurally.** A statement from another client resolves to an id this client has never issued, so it's simply unmaterialized and gets re-prepared. No aliasing, no documentation to read.
> * **#8 becomes unrepresentable.** With orphans identified the same way, the client resolves both sides itself instead of trusting a caller-supplied precondition.
> * **`PreparedHandle` leaves the public API entirely**, which is what the `PreparedStatement` doc already claims ("callers never touch raw handle ids or epochs") but the current shape doesn't deliver — `id()` and `session_epoch()` are `pub`, ODBC stores `PreparedHandle` in `pending_unprepare`, and `exec_direct.rs`'s tests assert on both accessors.
> * **The eviction-before-insert hazard stops existing.** That code is only necessary because the AE map is keyed by a server-reusable integer; keyed by `StatementId`, a fresh statement can never collide with a dropped one.
> 
> Costs, to be fair: one hash lookup per execute, which is noise next to a round trip; and a map entry lingering if a statement is dropped without unprepare, which is bounded and cleared on reconnect — no worse than the AE map's behavior today.
> 
> ### On timing
> I'd rather see this in this PR than tracked, for one reason: this is the public API shape, and it has exactly one consumer today. Every later adopter makes it more expensive, and the follow-up items currently filed against it (AB#47098, plus whatever #7 becomes) are all downstream of a shape we can still change cheaply. The blast radius is the same set of files this PR already rewrote.
> 
> ### Still needed
> The description no longer matches the code. It says the low-level calls are "renamed with a `_raw` suffix," documented as caller-owned recovery, and that `check_and_reconnect()` is now `pub` — `da8e70db` reversed all three. It also doesn't mention the AE metadata fix, orphan retention, or the timeout fast-fail, which are the most substantive changes since the last review.
> 
> ### Verification at `ffe125be`
> * `cargo test -p mssql-odbc --lib` — 450 passed.
> * `cargo test -p mssql-tds --lib` — 1611 passed (up from 1595), same 4 pre-existing `certificate_validator` fixture failures, unrelated to this PR.
> * The new coverage around orphan preservation, budget exhaustion, and AE pinning is good, and `abort_pending_prepare_capture` being wired into every failure path is real risk reduction for the current shape.



Implemented the `StatementId` shape — you were right that 2, 7, and 8 are one problem, and it collapsed them.

`StatementId(u64)` is client-issued and never reused; `TdsClient` owns `prepared_handles: HashMap<StatementId, i32>` and the AE map keyed the same way, both cleared on reconnect. `push_return_value` writes `id -> handle` when the token lands. No `Arc`, no `Mutex`.

- **No.7** — `capture_prepared_handle_into` and all three ODBC call sites gone; no "call after drain" rule.
- **No.2** — epoch gone. `plan_prepared_execution` didn't become a lookup, it disappeared: `execute_prepared` is now `statement.id.filter(|id| self.prepared_handles.contains_key(id))`. AB#47098 closed structurally.
- **No. 8** — the orphan is a `StatementId` resolved inside `sp_prepexec`, so the client owns both sides. Residual case: an orphan supplied alongside a still-live statement has no prepare to piggyback on, so it stays with the caller (documented + tested). msodbcsql asserts this is impossible — `DropPrepHandle` *moves* `hPrepCurrent` into `hPrepDropDeferred`, the same invariant our `take_id()` maintains. I defined it rather than asserting, since a caller-triggerable abort isn't right for a library.
- `PreparedHandle` / `session_handle` / `take_session_handle` are out of the public API; the eviction-before-insert hazard no longer exists. Dropped `Clone` as well — nothing needed it, and clones would share one id.

Went further: the raw handle no longer crosses the crate boundary at all (`execute_sp_prepare` returns `StatementId`; `sp_execute`/`sp_unprepare` resolve it themselves), and `sp_prepexec` mints its own id so the map insert can't collide by construction. `connection_recovery_count` is test-only now — no production consumer once the epoch went away.

**Description rewritten** — stale `_raw` / `pub check_and_reconnect` claims gone, now covers the AE fix, orphan retention, timeout fast-fail, and this refactor.

**On validation, flagging rather than burying:** I added `execute_prepared_reprepares_transparently_after_reconnect` (same statement, KILL, assert value + `recovery_count == 1` + a *new* id), but it lives in `connectivity.rs`, which CI excludes, and it fails locally for the same pre-existing reason the untouched `transparent_reconnect_after_kill` does — session recovery isn't negotiated against my container. So the end-to-end scenario still has **no enabled test** (AB#46631 → AB#47099). What CI does enforce now is the branch logic: three unit tests asserting RPC proc id and `@handle` wire bytes for first-materialization, id-with-no-live-handle, and live-handle reuse.

Verification at `c20ea2b`: `bclippy` + `bfmt` clean, 1637 `mssql-tds` + 492 `mssql-odbc` lib tests, 11/11 live `test_rpc_results`, 18/18 ODBC e2e (with `StaleHandleAfterReconnect...` skipped as noted).

Theekshna:
@David-Engel, @saurabh500, PR is ready for re-review

David-Engel (commented):
The `StatementId` refactor is in and it's a clear improvement over both prior shapes. Verified in code rather than from the summary: the epoch is gone entirely (`plan_prepared_execution` deleted, not converted — `execute_prepared` is now `statement.id.filter(|id| self.prepared_handles.contains_key(id))`), `capture_prepared_handle_into` and all three ODBC call sites are gone, `PreparedHandle` / `session_handle` / `take_session_handle` are out of the public API, `Clone` is dropped with the reason recorded, and the AE map is re-keyed so the eviction-before-insert hazard no longer exists. The orphan retention from the last round survives intact — taken only at the serialization boundary, both maps evicted together, ODBC writing it back on success and failure.

You also went past what I suggested in three places worth calling out: the raw handle no longer crosses the crate boundary at all (`execute_sp_prepare` returns a `StatementId`; `sp_execute` / `sp_unprepare` resolve it internally), `sp_prepexec` mints its own id so the map insert can't collide by construction, and `connection_recovery_count` is now test-only. The description rewrite is thorough, and I appreciate that the still-unenabled E2E scenario is flagged prominently rather than buried.

One finding, which I don't think should hold up this PR.

### AB#47098 isn't structurally closed yet

`next_statement_id` is a per-client `u64` starting at `0` (field decl `:218`, init `:280`, incremented in `issue_statement_id`), and `StatementId` derives `PartialEq` on the bare `u64` with no client identity in it. So every client's first prepared statement is `StatementId(0)`, the second `StatementId(1)`, and so on.

I tested it rather than reasoning about it — scratch test against `a8924c09`, since reverted:

```rust
let id_on_a = a.register_prepared_handle_for_test(111);
let id_on_b = b.register_prepared_handle_for_test(222);
assert_eq!(id_on_a, id_on_b);                                // passes
assert_eq!(b.prepared_handle_for_test(id_on_a), Some(222));  // passes
```

Both pass. A `PreparedStatement` materialized on client A and handed to client B resolves to B's handle for an unrelated statement, so `execute_prepared` takes the reuse branch and issues `sp_execute` against B's plan with A's parameters — silently, no error. That's a worse failure mode than the epoch design, where a foreign server handle would more often surface 8179.

Three places currently claim the opposite: `StatementId`'s rustdoc ("meaningless to any other client... re-prepared rather than aliasing an unrelated server handle"), description line 14 ("client-scoped and never reused"), and description line 42 ("Deleted AB#47098 - structurally: an id from another client was never issued here").

To be clear about provenance: my review sketched `next_statement_id: u64` as a client field and asserted AB#47098 would close structurally. You implemented the proposal accurately — the proposal was wrong on that point, and I should have caught it when I wrote it.

The fix is small enough to stand alone:

```rust
static NEXT_STATEMENT_ID: AtomicU64 = AtomicU64::new(0);

fn issue_statement_id(&mut self) -> StatementId {
    StatementId(NEXT_STATEMENT_ID.fetch_add(1, Ordering::Relaxed))
}
```

The per-client field goes away, u64 exhaustion stays unreachable, and every claim above becomes true. A client-id + counter pair works equally well.

### Why I'm not blocking on it

ODBC can't reach this — an HSTMT belongs to one HDBC — so it isn't a driver bug today. It's a public-API hazard for direct `mssql-tds` consumers, connection pools being the obvious case, and `PreparedStatement` has one consumer right now.

Two asks so we don't ship a guarantee we don't have:

1. Reopen AB#47098 (or file a fresh item) for the global-uniqueness fix, and take it soon while there's still one consumer.
2. Until then, soften the three claims above — the `StatementId` rustdoc especially, since that's what a future consumer will read. That's a doc-only edit and could ride along here or with the follow-up.

I'm dismissing my earlier change request so this isn't blocked.

### Verification at `a8924c09`

- `cargo test -p mssql-tds --lib` — 1633 passed, same 4 pre-existing `certificate_validator` fixture failures (missing local PEM/DER fixtures, unrelated).
- `cargo test -p mssql-odbc --lib` — 492 passed.

Theekshna:
> The `StatementId` refactor is in and it's a clear improvement over both prior shapes. Verified in code rather than from the summary: the epoch is gone entirely (`plan_prepared_execution` deleted, not converted — `execute_prepared` is now `statement.id.filter(|id| self.prepared_handles.contains_key(id))`), `capture_prepared_handle_into` and all three ODBC call sites are gone, `PreparedHandle` / `session_handle` / `take_session_handle` are out of the public API, `Clone` is dropped with the reason recorded, and the AE map is re-keyed so the eviction-before-insert hazard no longer exists. The orphan retention from the last round survives intact — taken only at the serialization boundary, both maps evicted together, ODBC writing it back on success and failure.
> 
> You also went past what I suggested in three places worth calling out: the raw handle no longer crosses the crate boundary at all (`execute_sp_prepare` returns a `StatementId`; `sp_execute` / `sp_unprepare` resolve it internally), `sp_prepexec` mints its own id so the map insert can't collide by construction, and `connection_recovery_count` is now test-only. The description rewrite is thorough, and I appreciate that the still-unenabled E2E scenario is flagged prominently rather than buried.
> 
> One finding, which I don't think should hold up this PR.
> 
> ### AB#47098 isn't structurally closed yet
> `next_statement_id` is a per-client `u64` starting at `0` (field decl `:218`, init `:280`, incremented in `issue_statement_id`), and `StatementId` derives `PartialEq` on the bare `u64` with no client identity in it. So every client's first prepared statement is `StatementId(0)`, the second `StatementId(1)`, and so on.
> 
> I tested it rather than reasoning about it — scratch test against `a8924c09`, since reverted:
> 
> ```rust
> let id_on_a = a.register_prepared_handle_for_test(111);
> let id_on_b = b.register_prepared_handle_for_test(222);
> assert_eq!(id_on_a, id_on_b);                                // passes
> assert_eq!(b.prepared_handle_for_test(id_on_a), Some(222));  // passes
> ```
> 
> Both pass. A `PreparedStatement` materialized on client A and handed to client B resolves to B's handle for an unrelated statement, so `execute_prepared` takes the reuse branch and issues `sp_execute` against B's plan with A's parameters — silently, no error. That's a worse failure mode than the epoch design, where a foreign server handle would more often surface 8179.
> 
> Three places currently claim the opposite: `StatementId`'s rustdoc ("meaningless to any other client... re-prepared rather than aliasing an unrelated server handle"), description line 14 ("client-scoped and never reused"), and description line 42 ("Deleted AB#47098 - structurally: an id from another client was never issued here").
> 
> To be clear about provenance: my review sketched `next_statement_id: u64` as a client field and asserted AB#47098 would close structurally. You implemented the proposal accurately — the proposal was wrong on that point, and I should have caught it when I wrote it.
> 
> The fix is small enough to stand alone:
> 
> ```rust
> static NEXT_STATEMENT_ID: AtomicU64 = AtomicU64::new(0);
> 
> fn issue_statement_id(&mut self) -> StatementId {
>     StatementId(NEXT_STATEMENT_ID.fetch_add(1, Ordering::Relaxed))
> }
> ```
> 
> The per-client field goes away, u64 exhaustion stays unreachable, and every claim above becomes true. A client-id + counter pair works equally well.
> 
> ### Why I'm not blocking on it
> ODBC can't reach this — an HSTMT belongs to one HDBC — so it isn't a driver bug today. It's a public-API hazard for direct `mssql-tds` consumers, connection pools being the obvious case, and `PreparedStatement` has one consumer right now.
> 
> Two asks so we don't ship a guarantee we don't have:
> 
> 1. Reopen AB#47098 (or file a fresh item) for the global-uniqueness fix, and take it soon while there's still one consumer.
> 2. Until then, soften the three claims above — the `StatementId` rustdoc especially, since that's what a future consumer will read. That's a doc-only edit and could ride along here or with the follow-up.
> 
> I'm dismissing my earlier change request so this isn't blocked.
> 
> ### Verification at `a8924c09`
> * `cargo test -p mssql-tds --lib` — 1633 passed, same 4 pre-existing `certificate_validator` fixture failures (missing local PEM/DER fixtures, unrelated).
> * `cargo test -p mssql-odbc --lib` — 492 passed.

Yeah `StatementId` collides across clients and a statement moved to another client silently reuses its unrelated handle. The idea was that statements are never handed to a different client in practice, I dont think it happens in connection pooling either. But I'm not 100% sure, so filed a task - **AB#47292**. We don't need it right now coz ODBC doesnt switch clients within a statement. 
Also fixed code docs & description

David-Engel (approved):

PR #151, opened 2026-07-27
Implement SQLRowCount

gargsaumya:
## Description

Implement `SQLRowCount` for the mssql-odbc driver, replacing the previous no-op stub.

`SQLRowCount` now reports the number of rows affected by the last `INSERT` /
`UPDATE` / `DELETE`, matching the classic msodbcsql C++ driver:

- DML → the affected-row count from the TDS `DONE` token.
- SELECT (forward-only cursor), DDL, and `SET NOCOUNT ON` → `-1`
  (`SQL_NO_ROWCOUNT_TOTAL`).
- No `HY010` sequence check in the driver — the Driver Manager rejects the call
  on an unexecuted statement before the driver is invoked, exactly as it does
  for msodbcsql. Verified live against unixODBC.
- The keyset/static-cursor `GetRowInfo` recompute branch is intentionally
  omitted (those cursor types are not yet supported) and documented in code.

### How the count is plumbed

- `mssql-tds`: new `TdsClient::last_rows_affected()` accessor backed by a field
  reset at each `execute()` and populated from the `DONE` token when the
  `DONE_COUNT` flag is set.
- `mssql-odbc`: `StmtState.row_count` is set on execute (`SQLExecDirectW`) and
  refreshed by `SQLMoreResults` so the count tracks the currently-positioned
  result set. `SQLRowCount` reads it, following the crate's
  safe-core/unsafe-shell + `ffi_entry!` conventions.

### Tests

- Unit tests in `mssql-odbc/src/api/row_count.rs` (null handle, null out-ptr,
  fresh statement → `-1` via direct driver call, stored DML count).
- New C++ e2e suite `tests/e2e/tests/row_count_test.cpp` — **14/14 e2e suites
  pass** against a live SQL Server (INSERT/UPDATE/DELETE counts, SELECT/DDL/
  `SET NOCOUNT ON` → `-1`, multi-result-set refresh, and the DM-enforced
  `HY010` on a fresh statement).

## Related Issues

Closes [AB#46414](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46414)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes 


mssql-tds/src/connection/tds_client.rs
```diff
@@ -594,6 +608,10 @@ impl TdsClient {
         // Batch execution always uses the connection's Always Encrypted setting.
         self.current_command_ce_setting = ExecutionColumnEncryptionSetting::UseConnectionSetting;
 
+        // Reset the affected-row count so a stale value from a prior execute on
+        // this connection cannot leak into `SQLRowCount` for the new statement.
+        self.last_rows_affected = -1;
```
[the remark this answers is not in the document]
gargsaumya: Verified and fixed in e160e154. Confirmed: the reset lived only in xecute(), so xecute_sp_executesql / xecute_sp_prepexec / xecute_sp_execute / xecute_stored_procedure did not reset it. Moved the reset into egin_command(), which every execution RPC path calls, so a prior DML count can no longer leak into a later command that reports none. Covered by unit test 
ow_count_resets_between_commands and e2e RowCountResetsBetweenExecutes.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1858,6 +1876,13 @@ impl TdsClient {
                     *count = count.saturating_add(done.row_count);
                     self.current_result_set_has_been_read_till_end = true;
 
+                    // Capture the affected-row count for `SQLRowCount`, but only
+                    // when the DONE_COUNT flag is set — otherwise `row_count` is
+                    // not meaningful (DDL, SET NOCOUNT ON) and must stay -1.
+                    if done.status.contains(DoneStatus::COUNT) {
+                        self.last_rows_affected = i64::try_from(done.row_count).unwrap_or(i64::MAX);
```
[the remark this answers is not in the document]
gargsaumya: Verified and fixed in e160e154. Confirmed: move_to_column_metadata continued past a counted DONE to reach COLMETADATA, carrying the DML count into a forward-only SELECT (e.g. UPDATE ...; SELECT ...). Now last_rows_affected is cleared to -1 when positioning on COLMETADATA, so the SELECT reports -1. Covered by unit test dml_then_select_batch_reports_no_row_count_for_select and e2e DmlThenSelectBatchReportsMinusOneForSelect.

mssql-odbc/src/api/more_results.rs
```diff
@@ -105,6 +105,8 @@ fn sql_more_results_safe(statement_handle: SqlHandle, stmt: &StmtHandle) -> SqlR
                 return SQL_ERROR;
             };
             stmt_state.column_metadata = metadata;
+            // Refresh the count for the newly-positioned result set (-1 for a SELECT).
+            stmt_state.row_count = client.last_rows_affected();
```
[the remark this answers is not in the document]
gargsaumya: Verified. The correctness half is fixed in e160e154: with the count cleared on COLMETADATA, INSERT ...; SELECT ... no longer carries the INSERT count into the SELECT (that SELECT now reports -1). The remaining half you note — surfacing DONE-only (DML) statements as their own result sets so each SQLMoreResults step exposes its own count, e.g. observing the UPDATE count in SELECT ...; UPDATE ... — is a pre-existing limitation of the multi-result-set iteration model (move_to_next only advances on COLMETADATA). That is a broader change to SQLMoreResults semantics beyond SQLRowCount and is out of scope for this PR; I'll track it separately.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1858,6 +1877,13 @@ impl TdsClient {
...
```
shiwanigupta0809: last_rows_affected is a single value that each DONE token overwrites. In one execute(), this loop reads all the DONE tokens before returning, so after a batch of several INSERT/UPDATE/DELETEs, only the final statement's count is left.
What I am saying is a scenario like this:
rc = SQLExecDirectA(hstmt,
(SQLCHAR*)"UPDATE dbo.RowCountTest SET Age = 99 WHERE Id = 1; "
"DELETE FROM dbo.RowCountTest WHERE Id = 3; "
"INSERT INTO dbo.RowCountTest VALUES (20, 'Grace', 45, 'Finance')", SQL_NTS);

Each of those three statements sends its own DONE-with-count, but the loop consumes all three in one execute(), and since last_rows_affected is a single value, each DONE overwrites the previous — so you end up with only the INSERT's count. And there's no way to retrieve the UPDATE's or DELETE's counts

mssql-odbc/tests/e2e/tests/row_count_test.cpp
```diff
@@ -0,0 +1,151 @@
...
+    ASSERT_EQ(SQL_NO_DATA, rc);
+
+    // Advance to the second result set.
+    rc = SQLMoreResults(stmt_);
+    ASSERT_SQL_OK(rc, SQL_HANDLE_STMT, stmt_);
+    EXPECT_EQ(-1, RowCount());
+
+    rc = SQLMoreResults(stmt_);
+    EXPECT_EQ(SQL_NO_DATA, rc);
+}
```
shiwanigupta0809, line 206: Please also add add a multi-DML batch case once SQLRowCount handles per-statement counts? That path (UPDATE; DELETE; INSERT with no SELECT) currently only surfaces the last statement's count, so it's the one gap not exercised    

Sharing ODBC c++ snippet for reference
rc = SQLExecDirectA(hstmt,
        (SQLCHAR*)"UPDATE dbo.RowCountTest SET Age = 99 WHERE Id = 1; "
        "DELETE FROM dbo.RowCountTest WHERE Id = 3; "
        "INSERT INTO dbo.RowCountTest VALUES (20, 'Grace', 45, 'Finance')", SQL_NTS);
    printf("Statement: Batch with UPDATE, DELETE, INSERT\n");
    printf("SQLExecDirect return: %s\n", returnCodeToString(rc));

    int resultNum = 1;
    do {
        if (SQL_SUCCEEDED(rc) || rc == SQL_NO_DATA) {
            SQLLEN rowCount = -999;
            SQLRETURN rcRowCount = SQLRowCount(hstmt, &rowCount);
            printf("Result set %d:\n", resultNum);
            printf("  SQLRowCount return: %s\n", returnCodeToString(rcRowCount));
            printf("  Row count: %lld\n", (long long)rowCount);
        }
        resultNum++;
        rc = SQLMoreResults(hstmt);
        printf("SQLMoreResults return: %s\n", returnCodeToString(rc));
    } while (rc == SQL_SUCCESS || rc == SQL_SUCCESS_WITH_INFO);

mssql-tds/src/connection/tds_client.rs
gargsaumya: Thanks, done.

One scoped note: the mixed DML; SELECT interleaving (surfacing an empty DML result before a SELECT) is intentionally left as a follow-up - it needs a rework of the shared `move_to_next` result-set contract that py/js/cli also depend on, which is beyond `SQLRowCount`. It's not misreported (the SELECT correctly shows -1); the DML count is just not surfaced in that mixed shape. Pure-DML batches (the case you mentioned) are fully handled.

shiwanigupta0809: Sure, we can split this off. The mixed DML; SELECT case comes down to how the result sets are split up, a DML statement should count as its own result set, so SQLMoreResults can walk through DML and SELECT the same way.
Might be worth looking at how msodbcsql does it here.  Please link the bug here.

shiwanigupta0809 (approved):

mssql-tds/src/connection/tds_client.rs
gargsaumya: Agreed.
Tracked separately in #157 , which has the exact behavior probed against the msodbcsql 18.5. @shiwanigupta0809's result-set-processing rework will cover it. This PR stays scoped to [SQLRowCount] (single + pure-DML verified).

PR #152, opened 2026-07-27
Implement SQLGetTypeInfo

gargsaumya:
## Description

Implements `SQLGetTypeInfoW` for the mssql-odbc driver. It executes the `sp_datatype_info_100`
catalog procedure as an RPC (matching the classic msodbcsql driver) and leaves the result
set open for `SQLFetch`/`SQLGetData`, reusing the existing execute/fetch machinery.

Behaviors, verified for parity against the msodbcsql source so the crate is a drop-in
replacement behind the same Driver Manager (mssql-python swap):

- Client-side validation of the `DataType` argument **before any I/O**: unrecognized type →
  `HY004`, user-defined type → `HYC00`, active exec / open cursor → `24000`.
- `@data_type` sent positionally (`SQLINT2`); `@ODBCVer = 3` sent as a named parameter for
  ODBC 3.x apps; 2.x apps omit it and receive the 2.x date/time id remap.
- Post-success column renames (ordinals 3 / 11 / 12 → `COLUMN_SIZE` / `FIXED_PREC_SCALE` /
  `AUTO_UNIQUE_VALUE`) so `SQLDescribeCol` reports the ODBC 3.x names identically.

Supporting additions: `DbcHandle::parent_env()` accessor, `SQL_ALL_TYPES` / `SQL_TIME`
constants, and an `HYC00` (`ERR_OPTIONAL_FEATURE_NOT_IMPLEMENTED`) diagnostic.

Deferred (gated on infrastructure not yet in this driver), matching msodbcsql but scoped out:
- Server-version / vector-based proc selection (`_90` / `_170`) — needs the negotiated server
  version plumbed to the ODBC layer; `_100` is valid for all supported servers (2016+).
- `ClearNullable` descriptor tweak — affects `SQLColAttribute(NULLABLE)`, not yet wired.
- The `sp_ddopen` scrollable-cursor wrapper — no cursors yet (forward-only firehose).

Note: retrieving type-info columns as native C types (e.g. `SQL_C_SSHORT`) depends on the
Phase-5 `SQLGetData` conversion work; today the columns read back as `SQL_C_CHAR`/`SQL_C_WCHAR`.

Tests: 5 unit tests (null handle, `HY004`, `HYC00`, disconnected, `24000`) plus a live e2e
suite (`get_type_info_test.cpp`, registered in the e2e CMakeLists).

## Related Issues

https://sqlclientdrivers.visualstudio.com/DefaultCollection/mssql-rs/_workitems/edit/46406

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes 


mssql-odbc/src/api/get_type_info.rs
```diff
@@ -0,0 +1,500 @@
...
+    // statement unchanged, matching msodbcsql.
+    {
+        let Ok(mut stmt_state) = stmt.inner.lock() else {
+            error!("SQLGetTypeInfoW: stmt mutex poisoned");
+            return SQL_ERROR;
+        };
+        free_errors(&mut stmt_state);
+
+        // msodbcsql validates the requested type before the cursor/exec state, so
+        // an invalid type reports HY004/HYC00 even when a cursor is also open.
```
shiwanigupta0809: The order here looks reversed from msodbcsql. Two things:
The Driver Manager checks cursor state first. For an open cursor, the DM returns 24000 (Invalid cursor state) before the call ever reaches the driver — so the driver's type validation never runs. matches the [ODBC spec ](https://learn.microsoft.com/en-gb/sql/odbc/reference/syntax/sqlgettypeinfo-function?view=sql-server-ver17#comments:~:text=A%20cursor%20was%20open%20on%20the%20StatementHandle%2C%20and%20SQLFetch%20or%20SQLFetchScroll%20had%20been%20called.%20This%20error%20is%20returned%20by%20the%20Driver%20Manager%20if%20SQLFetch%20or%20SQLFetchScroll%20has%20not%20returned%20SQL_NO_DATA%2C%20and%20is%20returned%20by%20the%20driver)

Even ignoring the DM, msodbcsql itself checks cursor state before data type. https://sqlclientdrivers.visualstudio.com/msodbcsql/_git/msodbcsql?path=/Sql/Ntdbms/sqlncli/odbc/sqlcdd.cpp&version=GBmaster&_a=contents&line=1969&lineStyle=plain&lineEnd=1975&lineStartColumn=2&lineEndColumn=10

mssql-odbc/tests/e2e/tests/get_type_info_test.cpp
```diff
@@ -0,0 +1,224 @@
...
+    ASSERT_SQL_OK(rc, SQL_HANDLE_STMT, stmt_);
+
+    rc = SQLGetTypeInfo(stmt_, SQL_ALL_TYPES);
+    ASSERT_SQL_OK(rc, SQL_HANDLE_STMT, stmt_);
+    // The new result set is the type-info contract, not the prior SELECT.
+    EXPECT_EQ("TYPE_NAME", DescribeColName(stmt_, 1));
+
+    rc = SQLCloseCursor(stmt_);
+    EXPECT_SQL_OK(rc, SQL_HANDLE_STMT, stmt_);
+}
```
shiwanigupta0809, line 240: can we add a recovery path test also: invalid type → HY004 → then a valid type on the same handle → success. That verifies a failed call leaves the statement clean  so the next call works

mssql-odbc/src/api/get_type_info.rs
gargsaumya: Done, thanks

shiwanigupta0809 (approved):

PR #161, opened 2026-08-03
Fix fuzz crash in SQL_VARIANT 7-prop-byte decoder

saurabh500:
## Summary

The `GH-Rust Fuzz Test` pipeline (definitionId 2207, [build 164079](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=164079&view=results)) detected a crash in the `fuzz_token_stream` target (`crash-80c55599ebe9adee81668324e4c6151894a33fbe`).

## Root cause

In `read_sql_variant`, a `SQL_VARIANT` value advertising **7 property bytes** is dispatched to `decode_seven_propbyte_variant` based solely on the property-byte count, regardless of the variant base type. That function held a debug `assert!` requiring the base type to be a character type (`BigVarChar` / `BigChar` / `NVarChar` / `NChar`). A malformed/fuzzed stream carrying 7 property bytes with a non-character base type (e.g. `IntN`) trips the assertion and aborts the process.

```
assertion failed: matches!(tds_type, TdsDataType::BigVarChar | TdsDataType::BigChar
    | TdsDataType::NVarChar | TdsDataType::NChar)
  at mssql-tds/src/datatypes/decoder.rs:2132
```

Reached via token `0xAC` (RETURNVALUE) → base type `0x62` (SQL_VARIANT) → `read_sql_variant` → `decode_seven_propbyte_variant`.

## Fix

Replace the assertion with a graceful `ProtocolError` return, consistent with the other property-byte SQL_VARIANT decoders (`decode_two_propbyte_variant`, the `_ =>` arms, etc.). Untrusted wire data must never abort the process.

Added regression test `ssvariant_seven_prop_unexpected_type` in the decoder test module, which reproduces the crash input shape and asserts a graceful error.

## Validation

- Reproduced the original crash locally against the downloaded artifact, confirmed the fix returns `Err` instead of panicking.
- `cargo bfmt` ✅
- `cargo bclippy` ✅ (no warnings)
- Decoder SQL_VARIANT tests pass ✅

## Work item

[AB#46937](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46937)

### DIFF
--- a/mssql-tds/src/datatypes/decoder.rs
+++ b/mssql-tds/src/datatypes/decoder.rs
@@ -2129,10 +2129,14 @@ async fn decode_seven_propbyte_variant<T>(
 where
     T: TdsPacketReader + Send + Sync,
 {
-    assert!(matches!(
+    if !matches!(
         tds_type,
         TdsDataType::BigVarChar | TdsDataType::BigChar | TdsDataType::NVarChar | TdsDataType::NChar
-    ));
+    ) {
+        return Err(crate::error::Error::ProtocolError(format!(
+            "Unexpected SQL variant base type for len(7) prop bytes: {tds_type:?}. Expected a character type."
+        )));
+    }
     let mut collation_bytes = vec![0u8; 5];
     reader.read_bytes(&mut collation_bytes).await?;
     let _max_length = reader.read_uint16().await? as usize;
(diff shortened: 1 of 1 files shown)


David-Engel (approved):
Approve. The fix is correct, minimal, and consistent with the sibling SQL_VARIANT decoders — converting the debug `assert!` in `decode_seven_propbyte_variant` into a graceful `ProtocolError` is the right call, since untrusted wire data must never abort the process. The dispatch in `read_sql_variant` routes purely on `variant_prop_bytes == 7`, so a 7-prop stream with a non-character base type genuinely reaches this path. The regression test mirrors the existing `ssvariant_*_prop_unexpected_type` tests and the byte layout is correct (`data_length = 20 - 2 - 7 = 11`, and the base type short-circuits to the error before any padding is read).

One non-blocking follow-up (out of scope for this PR): the same latent panic pattern exists nearby in the XML/JSON/UDT decode arms, which still use `assert!(metadata.is_plp())`. These are driven by parsed (untrusted) metadata, so a crafted stream advertising an XML/JSON/UDT column that isn't marked PLP could abort the process the same way this fuzz crash did. Worth a separate work item to convert those asserts to `ProtocolError` returns as well, so the whole decoder is panic-free on hostile input.

PR #163, opened 2026-08-03
Sync vanilla Python build images and additional Linux base images to GHCR

### LINKED ISSUE #162: Sync vanilla manylinux/musllinux base images to GHCR
## Description

`.pipeline/sync-container-images.yml` mirrors ACR images to GHCR in the `Mirror_To_GHCR` stage, but the image list only includes the custom `python-build/*_rust` images built in stage 2. The vanilla PyPA base images imported in stage 1 are never mirrored:

- `import/python-build/manylinux_2_34_x86_64:latest`
- `import/python-build/manylinux_2_34_aarch64:latest`
- `import/python-build/manylinux_2_28_x86_64:latest`
- `import/python-build/manylinux_2_28_aarch64:latest`
- `import/python-build/musllinux_1_2_x86_64:latest`
- `import/python-build/musllinux_1_2_aarch64:latest`

Every other `import/*` image (alpine, ubuntu, debian, ubi, oraclelinux) is mirrored, so these are an inconsistent gap. Without them, anything that needs the unmodified manylinux/musllinux bases has to pull from ACR (authenticated) or quay.io instead of GHCR anonymously.

## Expected

Add the six vanilla manylinux/musllinux imports to the GHCR mirror list so they are available alongside the other `import/*` images.

### LINKED ISSUE #166: Sync additional Linux base images to GHCR for msphpsql test pipelines
The msphpsql Linux test pipelines are moving off `tdslibrs.azurecr.io/import` and onto the public GHCR mirror produced by `.pipeline/sync-container-images.yml` (stage `Mirror_To_GHCR`). See draft PR `msphpsql!8007` (sqlclientdrivers ADO, repo `msphpsql`).

Most base images msphpsql needs are not in the mirror yet.

### Currently mirrored (verified against ghcr.io)

| Repo | Tags |
|------|------|
| `import/alpine` | 3.18, 3.19, 3.20, 3.21 |
| `import/debian` | bookworm |
| `import/ubuntu` | 22.04, 24.04 |
| `import/redhat/ubi9` | latest |
| `import/oraclelinux` | 9 |

### Requested additions

Please add these to the `Import_Base_Images` stage and to the `images.txt` list in the `Mirror_To_GHCR` stage:

| Requested GHCR path | Upstream source |
|---------------------|-----------------|
| `import/alpine:3.22` | `docker.io/library/alpine:3.22` |
| `import/alpine:3.23` | `docker.io/library/alpine:3.23` |
| `import/debian:bullseye` | `docker.io/library/debian:bullseye` |
| `import/debian:trixie` | `docker.io/library/debian:trixie` |
| `import/centos:stream9` | `quay.io/centos/centos:stream9` |
| `import/redhat/ubi8:latest` | `registry.access.redhat.com/ubi8/ubi:latest` |
| `import/redhat/ubi10:latest` | `registry.access.redhat.com/ubi10/ubi:latest` |
| `import/opensuse/leap:15.6` | `docker.io/opensuse/leap:15.6` |

The naming above is what `build_container.sh` in the msphpsql PR resolves to, so it matches the flat `import/<distro>:<tag>` convention already used for `debian`/`oraclelinux` and the `redhat/ubiN` convention already used for ubi9.

### Also needed

Each new GHCR package must be set to **Public** so the msphpsql pipelines can pull anonymously (no Azure service connection on those agents).

### Blocking

Until these are mirrored, the msphpsql PR must stay in draft. Distros that would break: Alpine 3.22/3.23, Debian 11, Debian 13, SUSE 15.6, CentOS Stream 9, RHEL 8, RHEL 10.

saurabh500:
## Description

Expands the container sync pipeline (`.pipeline/sync-container-images.yml`) so the GHCR mirror covers all the base images consumers need, and adds pre-built Kerberos CI images.

**1. Mirror the vanilla PyPA images (#162)**

Previously only the custom `python-build/*_rust` images built in stage 2 were mirrored. The vanilla imports from stage 1 are now included:

- `import/python-build/manylinux_2_34_x86_64:latest` / `..._aarch64:latest`
- `import/python-build/manylinux_2_28_x86_64:latest` / `..._aarch64:latest`
- `import/python-build/musllinux_1_2_x86_64:latest` / `..._aarch64:latest`

**2. Add the Linux base images msphpsql needs (#166)**

New imports in `Import_Base_Images`, each also added to the mirror list:

| GHCR path | Upstream source |
|---|---|
| `import/alpine:3.22` | `docker.io/library/alpine:3.22` |
| `import/alpine:3.23` | `docker.io/library/alpine:3.23` |
| `import/debian:bullseye` | `docker.io/library/debian:bullseye` |
| `import/debian:trixie` | `docker.io/library/debian:trixie` |
| `import/centos:stream9` | `quay.io/centos/centos:stream9` |
| `import/redhat/ubi10:latest` | `registry.access.redhat.com/ubi10/ubi:latest` |
| `import/opensuse/leap:15.6` | `docker.io/opensuse/leap:15.6` |

`import/redhat/ubi8:latest` was already imported and mirrored, so it needed no change.

New queue-time toggles follow the existing per-group pattern: `syncAlpine3`, `syncCentOS`, `syncSUSE`. Debian bullseye/trixie and UBI10 ride along with the existing `syncDebianRHEL` group (job renamed from "Import Debian & RHEL UBI9" to "Import Debian & RHEL UBI").

Sources not on Docker Hub (quay.io CentOS, `registry.access.redhat.com` UBI10) skip the Docker Hub credential args, matching the existing quay.io manylinux/musllinux steps.

All 31 mirrored images have been verified as anonymously pullable from `ghcr.io/microsoft/mssql-rs/...`.

**3. Pre-build the Kerberos CI images (producer side)**

The ADO Kerberos test stage currently rebuilds a Samba AD DC image, a SQL Server+AD image, and up to 9 per-distro client images on every PR/CI run. This adds a `BuildKerberosImages` job to `BuildAndPushImages` that builds and pushes them once, plus a `buildKerberosImages` toggle, and mirrors them to GHCR:

- `kerberos/samba-dc:latest`, `kerberos/mssql-ad:latest`
- `kerberos/client:<profile>-latest` for alpine318/319/320/321, debian, ubuntu22, ubuntu24, rhel9, oracle9

Supporting changes:

- `Dockerfile.samba-dc` / `Dockerfile.mssql-ad`: base image is now a `BASE_IMAGE` build arg, and the init scripts are mounted from `./scripts` at runtime instead of baked in — so a pre-built image stays valid when a PR edits a script. The `dc` service in all three compose files gains the `./scripts:/scripts:ro` mount the `mssql` service already had.
- `Dockerfile.client.matrix`: pre-installs `cargo-nextest` 0.9.99 from the release tarball rather than compiling it from source inside every container at test time.

The pipeline's profile → base image → entry script table was verified against the rendered `docker-compose-matrix.yml` config; all 9 profiles match exactly.

**Consumer side** (compose `image:` refs and the test template pulling instead of building) lands in a separate PR.

`cargo-nextest` is installed to `/usr/local/bin` rather than `/root/.cargo/bin`, and both `cargo install` call sites in `run-kerberos-tests.sh` are now guarded with `command -v cargo-nextest >/dev/null 2>&1 || ...`. Two things made this necessary: a tarball drop has no `.crates.toml` entry, so an unguarded `cargo install` into `CARGO_HOME/bin` aborts with "binary already exists in destination"; and because `ENV PATH` puts `/root/.cargo/bin` first, an unguarded install would otherwise still compile from source and shadow the pre-installed copy, making the pre-install dead weight. Verified in a container: skipped in 0s when the binary is present, installs normally (53s) on an image without it, and a genuine install failure still propagates.

> [!IMPORTANT]
> **The GHCR mirror stage is load-bearing for the Kerberos CI stage.** The consumer pipeline pulls the Kerberos images exclusively from `ghcr.io/microsoft/mssql-rs/kerberos/*` and does not reference ACR at all. The ACR build/push in `BuildKerberosImages` is only the build target that `Mirror_To_GHCR` copies from. If the mirror stage is skipped, or those 11 `kerberos/*` entries are dropped from `images.txt`, the Kerberos stage silently falls back to building every image locally and the speedup disappears with no failure signal.

> [!NOTE]
> New GHCR packages are created private, and visibility is settable only per-package in the UI — there is no REST or GraphQL endpoint for it. The 11 `kerberos/*` packages have been published and flipped to public; verified anonymously pullable.

## Related Issues

Fixes #162
Fixes #166

## Checklist

- [x] `cargo bfmt` passes — n/a, no Rust changes
- [x] `cargo bclippy` passes — n/a, no Rust changes
- [x] `cargo btest` passes — n/a, no Rust changes
- [x] New/changed functionality has tests — n/a; pipeline and compose files YAML-validated, `docker compose config` confirms the `dc` `/scripts` mount in all three files, mirror list cross-checked against import/build targets, and the Kerberos profile table diffed against the rendered compose config
- [x] Public API changes are documented — n/a, no public API change

### DIFF
--- a/.pipeline/sync-container-images.yml
+++ b/.pipeline/sync-container-images.yml
@@ -733,6 +733,7 @@ stages:
   jobs:
   - job: BuildX64Images
     displayName: Build x64 Images
+    condition: or(eq('${{ parameters.buildPythonImages }}', 'true'), eq('${{ parameters.buildUbuntuImages }}', 'true'))
     pool:
       name: RUST-1ES-POOL-WUS3
       demands:
@@ -921,6 +922,10 @@ stages:
   - job: BuildARM64Images
     displayName: Build ARM64 Images
     dependsOn: BuildX64Images
+    # succeeded() must stay explicit: supplying a condition replaces the
+    # implicit succeeded(), which would otherwise let ARM64 start after an
+    # x64 failure.
+    condition: and(succeeded(), or(eq('${{ parameters.buildPythonImages }}', 'true'), eq('${{ parameters.buildUbuntuImages }}', 'true')))
     pool:
       name: RUST-1ES-POOL-ARM-WUS3
       demands:
--- a/.pipeline/sync-container-images.yml
+++ b/.pipeline/sync-container-images.yml
@@ -1035,6 +1035,12 @@ stages:
         import/redhat/ubi9:latest
         import/redhat/ubi8:latest
         import/oraclelinux:9
+        import/python-build/manylinux_2_34_x86_64:latest
+        import/python-build/manylinux_2_34_aarch64:latest
+        import/python-build/manylinux_2_28_x86_64:latest
+        import/python-build/manylinux_2_28_aarch64:latest
+        import/python-build/musllinux_1_2_x86_64:latest
+        import/python-build/musllinux_1_2_aarch64:latest
         EOF
         echo "Images to mirror to GHCR:"
         sed 's/^/  /' "$(Pipeline.Workspace)/ghcr/images.txt"
--- a/.pipeline/sync-container-images.yml
+++ b/.pipeline/sync-container-images.yml
@@ -41,8 +41,23 @@ parameters:
   type: boolean
   default: true
 
+- name: syncAlpine3
+  displayName: 'Import: Alpine 3.22-3.23 (docker.io)'
+  type: boolean
+  default: true
+
 - name: syncDebianRHEL
   displayName: 'Import: Debian & RHEL (docker.io)'
+  type: boolean
+  default: true
+
+- name: syncCentOS
+  displayName: 'Import: CentOS Stream (quay.io)'
+  type: boolean
+  default: true
+
+- name: syncSUSE
+  displayName: 'Import: openSUSE Leap (docker.io)'
   type: boolean
   default: true
 
@@ -399,9 +414,68 @@ stages:
         DOCKER_PAT: $(DOCKERHUB_PAT)
       retryCountOnTaskFailure: 3
 
+  # Alpine 3.22 & 3.23 (Docker Hub - rate limited)
+  - job: Alpine3
+    displayName: 'Import Alpine 3.22 & 3.23'
+    condition: eq('${{ parameters.syncAlpine3 }}', 'true')
+    pool:
+      name: RUST-1ES-POOL-WUS3
+      demands:
+        - imageOverride -equals RUST-1ES-UBUSLIM
+    steps:
+    - checkout: none
+    - task: AzureCLI@2
+      displayName: 'Import Alpine 3.22'
+      inputs:
+        azureSubscription: 'Magnitude Test'
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          AUTH_ARGS=""
+          if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
+            echo "Using authenticated Docker Hub access"
+            AUTH_ARGS="--username $DOCKER_USER --password $DOCKER_PAT"
+          else
+            echo "Using anonymous Docker Hub access"
+          fi
+          az acr import \
+            --name $(ACR_NAME) \
+            --source docker.io/library/alpine:3.22 \
+            --image import/alpine:3.22 \
+            $AUTH_ARGS \
+            --force
+      env:
+        DOCKER_USER: $(DOCKERHUB_USERNAME)
+        DOCKER_PAT: $(DOCKERHUB_PAT)
+      retryCountOnTaskFailure: 3
+    - task: AzureCLI@2
+      displayName: 'Import Alpine 3.23'
+      inputs:
+        azureSubscription: 'Magnitude Test'
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          AUTH_ARGS=""
+          if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
+            echo "Using authenticated Docker Hub access"
+            AUTH_ARGS="--username $DOCKER_USER --password $DOCKER_PAT"
+          else
+            echo "Using anonymous Docker Hub access"
+          fi
+          az acr import \
+            --name $(ACR_NAME) \
+            --source docker.io/library/alpine:3.23 \
+            --image import/alpine:3.23 \
+            $AUTH_ARGS \
+            --force
+      env:
+        DOCKER_USER: $(DOCKERHUB_USERNAME)
+        DOCKER_PAT: $(DOCKERHUB_PAT)
+      retryCountOnTaskFailure: 3
+
   # Debian & RHEL (Docker Hub - rate limited)
   - job: DebianRHEL
-    displayName: 'Import Debian & RHEL UBI9'
+    displayName: 'Import Debian & RHEL UBI'
     condition: eq('${{ parameters.syncDebianRHEL }}', 'true')
     pool:
       name: RUST-1ES-POOL-WUS3
@@ -427,6 +501,54 @@ stages:
             --name $(ACR_NAME) \
             --source docker.io/library/debian:bookworm \
             --image import/debian:bookworm \
+            $AUTH_ARGS \
+            --force
+      env:
+        DOCKER_USER: $(DOCKERHUB_USERNAME)
+        DOCKER_PAT: $(DOCKERHUB_PAT)
+      retryCountOnTaskFailure: 3
+    - task: AzureCLI@2
+      displayName: 'Import Debian Bullseye'
+      inputs:
+        azureSubscription: 'Magnitude Test'
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          AUTH_ARGS=""
+          if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
+            echo "Using authenticated Docker Hub access"
+            AUTH_ARGS="--username $DOCKER_USER --password $DOCKER_PAT"
+          else
+            echo "Using anonymous Docker Hub access"
+          fi
+          az acr import \
+            --name $(ACR_NAME) \
+            --source docker.io/library/debian:bullseye \
+            --image import/debian:bullseye \
+            $AUTH_ARGS \
+            --force
+      env:
+        DOCKER_USER: $(DOCKERHUB_USERNAME)
+        DOCKER_PAT: $(DOCKERHUB_PAT)
+      retryCountOnTaskFailure: 3
+    - task: AzureCLI@2
+      displayName: 'Import Debian Trixie'
+      inputs:
+        azureSubscription: 'Magnitude Test'
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          AUTH_ARGS=""
+          if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
+            echo "Using authenticated Docker Hub access"
+            AUTH_ARGS="--username $DOCKER_USER --password $DOCKER_PAT"
+          else
+            echo "Using anonymous Docker Hub access"
+          fi
+          az acr import \
+            --name $(ACR_NAME) \
+            --source docker.io/library/debian:trixie \
+            --image import/debian:trixie \
             $AUTH_ARGS \
             --force
       env:
@@ -481,6 +603,21 @@ stages:
         DOCKER_USER: $(DOCKERHUB_USERNAME)
         DOCKER_PAT: $(DOCKERHUB_PAT)
       retryCountOnTaskFailure: 3
+    # UBI10 is not published on Docker Hub, so it comes straight from the
+    # Red Hat public registry (anonymous, no Docker Hub credentials).
+    - task: AzureCLI@2
+      displayName: 'Import Red Hat UBI10'
+      inputs:
+        azureSubscription: 'Magnitude Test'
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          az acr import \
+            --name $(ACR_NAME) \
+            --source registry.access.redhat.com/ubi10/ubi:latest \
+            --image import/redhat/ubi10:latest \
+            --force
+      retryCountOnTaskFailure: 3
 
   # Oracle Linux (Docker Hub - rate limited)
   - job: Oracle
@@ -510,6 +647,65 @@ stages:
             --name $(ACR_NAME) \
             --source docker.io/library/oraclelinux:9 \
             --image import/oraclelinux:9 \
+            $AUTH_ARGS \
+            --force
+      env:
+        DOCKER_USER: $(DOCKERHUB_USERNAME)
+        DOCKER_PAT: $(DOCKERHUB_PAT)
+      retryCountOnTaskFailure: 3
+
+  # CentOS Stream (quay.io - no Docker Hub rate limit)
+  - job: CentOS
+    displayName: 'Import CentOS Stream 9'
+    condition: eq('${{ parameters.syncCentOS }}', 'true')
+    pool:
+      name: RUST-1ES-POOL-WUS3
+      demands:
+        - imageOverride -equals RUST-1ES-UBUSLIM
+    steps:
+    - checkout: none
+    - task: AzureCLI@2
+      displayName: 'Import CentOS Stream 9'
+      inputs:
+        azureSubscription: 'Magnitude Test'
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          az acr import \
+            --name $(ACR_NAME) \
+            --source quay.io/centos/centos:stream9 \
+            --image import/centos:stream9 \
+            --force
+      retryCountOnTaskFailure: 3
+
+  # openSUSE Leap (Docker Hub - rate limited)
+  - job: SUSE
+    displayName: 'Import openSUSE Leap 15.6'
+    condition: eq('${{ parameters.syncSUSE }}', 'true')
+    pool:
+      name: RUST-1ES-POOL-WUS3
+      demands:
+        - imageOverride -equals RUST-1ES-UBUSLIM
+    steps:
+    - checkout: none
+    - task: AzureCLI@2
+      displayName: 'Import openSUSE Leap 15.6'
+      inputs:
+        azureSubscription: 'Magnitude Test'
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          AUTH_ARGS=""
+          if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
+            echo "Using authenticated Docker Hub access"
+            AUTH_ARGS="--username $DOCKER_USER --password $DOCKER_PAT"
+          else
+            echo "Using anonymous Docker Hub access"
+          fi
+          az acr import \
+            --name $(ACR_NAME) \
+            --source docker.io/opensuse/leap:15.6 \
+            --image import/opensuse/leap:15.6 \
             $AUTH_ARGS \
             --force
       env:
@@ -1026,15 +1222,22 @@ stages:
         build/ubuntu:22.04
         build/alpine:3.18
         import/debian:bookworm
+        import/debian:bullseye
+        import/debian:trixie
         import/alpine:3.18
         import/alpine:3.19
         import/alpine:3.20
         import/alpine:3.21
+        import/alpine:3.22
+        import/alpine:3.23
         import/ubuntu:22.04
         import/ubuntu:24.04
         import/redhat/ubi9:latest
         import/redhat/ubi8:latest
+        import/redhat/ubi10:latest
         import/oraclelinux:9
+        import/centos:stream9
+        import/opensuse/leap:15.6
         import/python-build/manylinux_2_34_x86_64:latest
         import/python-build/manylinux_2_34_aarch64:latest
         import/python-build/manylinux_2_28_x86_64:latest
--- a/.pipeline/sync-container-images.yml
+++ b/.pipeline/sync-container-images.yml
@@ -259,10 +259,10 @@ stages:
         scriptType: 'bash'
         scriptLocation: 'inlineScript'
         inlineScript: |
-          AUTH_ARGS=""
+          AUTH_ARGS=()
           if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
             echo "Using authenticated Docker Hub access"
-            AUTH_ARGS="--username $DOCKER_USER --password $DOCKER_PAT"
+            AUTH_ARGS=(--username "$DOCKER_USER" --password "$DOCKER_PAT")
           else
             echo "Using anonymous Docker Hub access"
           fi
@@ -270,7 +270,7 @@ stages:
             --name $(ACR_NAME) \
             --source docker.io/library/ubuntu:22.04 \
             --image import/ubuntu:22.04 \
-            $AUTH_ARGS \
+            "${AUTH_ARGS[@]}" \
             --force
       env:
         DOCKER_USER: $(DOCKERHUB_USERNAME)
@@ -283,10 +283,10 @@ stages:
         scriptType: 'bash'
         scriptLocation: 'inlineScript'
         inlineScript: |
-          AUTH_ARGS=""
+          AUTH_ARGS=()
           if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
             echo "Using authenticated Docker Hub access"
-            AUTH_ARGS="--username $DOCKER_USER --password $DOCKER_PAT"
+            AUTH_ARGS=(--username "$DOCKER_USER" --password "$DOCKER_PAT")
           else
             echo "Using anonymous Docker Hub access"
           fi
@@ -294,7 +294,7 @@ stages:
             --name $(ACR_NAME) \
             --source docker.io/library/ubuntu:24.04 \
             --image import/ubuntu:24.04 \
-            $AUTH_ARGS \
+            "${AUTH_ARGS[@]}" \
             --force
       env:
         DOCKER_USER: $(DOCKERHUB_USERNAME)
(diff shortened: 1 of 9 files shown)


kerberos-test/Dockerfile.client.matrix
```diff
@@ -43,6 +45,14 @@ RUN if [ -f /etc/alpine-release ]; then \
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
 ENV PATH="/root/.cargo/bin:${PATH}"
 
+# Pre-install cargo-nextest. Without this, run-kerberos-tests.sh compiles it
+# from source inside every client container on every CI run.
+ARG NEXTEST_VERSION=0.9.99
+RUN if [ -f /etc/alpine-release ]; then NEXTEST_PLATFORM=linux-musl; else NEXTEST_PLATFORM=linux; fi \
+    && curl -LsSf "https://get.nexte.st/${NEXTEST_VERSION}/${NEXTEST_PLATFORM}" \
+        | tar zxf - -C /root/.cargo/bin \
+    && cargo nextest --version
```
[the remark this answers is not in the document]
saurabh500: Good catch, and confirmed — this was actively breaking the PR. The Kerberos stage runs on every PR (`validation-stages.yml:337`), and the run on this PR failed with exactly the predicted error:

```
Step 4: Install cargo-nextest (CI mode)...
  Downloaded cargo-nextest v0.9.99
error: binary `cargo-nextest` already exists in destination
FAILED profiles: ubuntu22 alpine318
```

I'd previously reasoned this was the consumer PR's problem, but since both PRs are independent and either can merge first, that assumption was wrong.

Went with your first suggestion — extracting to `/usr/local/bin` — rather than adding the guard here, since it avoids duplicating a change the consumer PR already makes to the same lines of `run-kerberos-tests.sh`.

Verified the fix in a container with `CARGO_HOME=/root/.cargo` to match the rustup layout in this image:

| | before (`/root/.cargo/bin`) | after (`/usr/local/bin`) |
|---|---|---|
| `cargo install` exit | **101** — `binary already exists in destination` | **0** |
| `command -v cargo-nextest` | — | `/usr/local/bin/cargo-nextest` |
| `cargo nextest --version` | — | `0.9.99` |

So the unguarded `cargo install` now succeeds (this PR is safe standalone), `command -v` still resolves it so the consumer guard short-circuits once that lands, and the Dockerfile's own `cargo nextest --version` check still passes.

.pipeline/sync-container-images.yml
```diff
@@ -528,7 +729,7 @@ stages:
 - stage: BuildAndPushImages
   displayName: Build and Push Custom Images
   dependsOn: []
-  condition: or(eq('${{ parameters.buildPythonImages }}', 'true'), eq('${{ parameters.buildUbuntuImages }}', 'true'))
+  condition: or(eq('${{ parameters.buildPythonImages }}', 'true'), eq('${{ parameters.buildUbuntuImages }}', 'true'), eq('${{ parameters.buildKerberosImages }}', 'true'))
```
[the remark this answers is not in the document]
saurabh500: Confirmed and fixed. Verified the steps use runtime `condition:` (lines 746, 785, 944) rather than compile-time `${{ if }}`, so the jobs are indeed scheduled and the agents reserved.

ARM64 is a bit worse than described: it also `dependsOn: BuildX64Images` and runs `DockerInstaller@0` plus `install-ubuntu-dependency.yaml` unconditionally before its gated step, so a Kerberos-only run would wait on x64 and then do real setup work for nothing.

Added `or(buildPythonImages, buildUbuntuImages)` to both jobs. One adjustment to the suggestion: on `BuildARM64Images` I wrapped it as `and(succeeded(), or(...))`, because supplying an explicit condition replaces the implicit `succeeded()` and would otherwise let ARM64 start after an x64 failure.

Also checked the downstream jobs: `CreateMultiArchManifests` is already gated on `buildUbuntuImages` so it skips consistently, and `VerifyAllImages` is `condition: always()`, so the skipped upstream jobs don't suppress it.

kerberos-test/Dockerfile.client.matrix
```diff
@@ -43,6 +45,20 @@ RUN if [ -f /etc/alpine-release ]; then \
...
+# Deliberately installed to /usr/local/bin rather than /root/.cargo/bin: a
+# tarball drop has no .crates.toml entry, so a plain `cargo install` would not
+# short-circuit and would abort with "binary already exists in destination".
+# Keeping it outside Cargo's bin dir means an unguarded `cargo install` still
+# succeeds, so this image is safe regardless of which PR lands first.
+ARG NEXTEST_VERSION=0.9.99
+RUN if [ -f /etc/alpine-release ]; then NEXTEST_PLATFORM=linux-musl; else NEXTEST_PLATFORM=linux; fi \
+    && curl -LsSf "https://get.nexte.st/${NEXTEST_VERSION}/${NEXTEST_PLATFORM}" \
+        | tar zxf - -C /usr/local/bin \
+    && cargo nextest --version
```
David-Engel, line 60: **Suggestion:** this pre-install is currently dead weight, and the PR description contradicts the code here.

`run-kerberos-tests.sh` has two unguarded `cargo install cargo-nextest --version 0.9.99 --locked` calls (archive mode ~line 178, CI mode ~line 262). With the binary in `/usr/local/bin`, that install *succeeds* - it compiles from source and lands in `/root/.cargo/bin`, which is **first** on `PATH` per the `ENV PATH` on the line above. So every CI run still pays the full compile and then shadows the pre-installed binary. Net effect today: larger images, zero savings.

The description's `> [!IMPORTANT]` block says the binary is dropped into `~/.cargo/bin` and cargo "will rebuild and then fail" - that was accurate before the last commit moved it to `/usr/local/bin`. Worth updating, since as written it will mislead whoever picks up the consumer PR.

Rather than deferring the guard, it is a one-liner that is safe on images with *or* without the pre-install:

```bash
docker exec "$CONTAINER_NAME" bash -c '
    command -v cargo-nextest >/dev/null 2>&1 \
      || cargo install cargo-nextest --version 0.9.99 --locked
'
```

Landing that in this PR makes the pre-install pay off immediately and removes the cross-PR ordering dependency entirely.

The `&& cargo nextest --version` smoke check at the end is a good touch - it catches a 404 from `get.nexte.st` that the unpipefail'd `curl | tar` would otherwise swallow.

David-Engel (commented):
Solid, well-documented infrastructure PR. It expands the ACR->GHCR mirror from 8 to 42 tags (6 vanilla PyPA imports, 7 new distro bases, 11 new Kerberos CI images) and moves the Kerberos DC/SQL init scripts from baked-in `COPY` to a runtime bind mount so pre-built images survive script edits.

Things I verified rather than assumed:

- The mirror list matches the import targets exactly - all 23 `import/*` entries have a corresponding `az acr import --image` in stage 1, no orphans in either direction.
- The profile -> base image -> entry script table in `BuildKerberosImages` matches `docker-compose-matrix.yml` for all 9 profiles.
- All three compose files now mount `./scripts:/scripts:ro` on the `dc` service, and `mssql` already had it.
- Nothing outside the compose files runs `Dockerfile.samba-dc` / `Dockerfile.mssql-ad`, so there is no path that starts those containers without the mount.
- Switching to `ENTRYPOINT ["/bin/bash", "/scripts/init-mssql-ad.sh"]` is the right call - a `:ro` bind mount cannot be `chmod +x`'d, so exec form on the script directly would have failed.

No blocking issues. Suggestions inline, plus a few smaller things below.

### Suggestion: ~250 lines of copy-pasted import tasks

Every new import task differs only in `--source`, `--image`, and the display name. An `${{ each img in parameters.dockerHubImports }}` loop over a list of `{source, target, name}` objects would collapse all 15 Docker Hub imports into one block - and would make the credential-handling fix I left inline a one-place change instead of a 15-place change. Bigger than this PR needs, but the file is now 1398 lines and grows linearly with every base image we add.

### Nits

- The profile table at line 896 duplicates `docker-compose-matrix.yml`. It matches today (I checked all 9), and the comment flags the coupling, but drift would silently build clients on the wrong base with no test failure. Deriving it from `docker compose config` in the same job would make it self-verifying.
- Nice catch on making `succeeded()` explicit for `BuildARM64Images`. `CreateMultiArchManifests` has the identical latent issue - `condition: eq('${{ parameters.buildUbuntuImages }}', 'true')` with no `succeeded()`, so it runs after an x64 failure. Pre-existing and outside this diff, but adjacent enough to be worth fixing while the context is fresh.
- `Dockerfile.samba-dc` and `Dockerfile.mssql-ad` both gain a `BASE_IMAGE` arg that no caller ever sets - the pipeline builds both with defaults. Harmless, just currently unused.
- `kerberos/client:<profile>-latest` bakes the whole Dockerfile, so a PR that edits `Dockerfile.client.matrix` will not take effect until the sync pipeline reruns. That is the same staleness class the script mount solves for DC/SQL. The header comment hints at it; an explicit "editing this file requires re-running the sync pipeline" line would save someone an afternoon of confusion.

.pipeline/sync-container-images.yml
```diff
@@ -1025,16 +1302,40 @@ stages:
...
+        kerberos/mssql-ad:latest
+        kerberos/client:alpine318-latest
+        kerberos/client:alpine319-latest
+        kerberos/client:alpine320-latest
+        kerberos/client:alpine321-latest
+        kerberos/client:debian-latest
+        kerberos/client:ubuntu22-latest
+        kerberos/client:ubuntu24-latest
+        kerberos/client:rhel9-latest
+        kerberos/client:oracle9-latest
```
David-Engel, line 1331: **Suggestion:** these 11 tags are gated on `buildKerberosImages`, but the mirror list is an unconditional heredoc and the mirror step hard-fails (`FAIL=1` -> `exit 1`) on any tag it cannot resolve.

Queue the pipeline with `buildKerberosImages=false` before these have ever been pushed and `Mirror_To_GHCR` fails - after burning 3 retries. Same exposure for the other new toggle-gated additions below: `import/alpine:3.22` / `3.23` (`syncAlpine3`), `import/centos:stream9` (`syncCentOS`), `import/opensuse/leap:15.6` (`syncSUSE`).

This is pre-existing shape, but the PR triples the number of entries and adds a whole namespace that has never existed in ACR, so the first real run is the most likely one to hit it. Two options:

- Emit the entries conditionally, e.g. wrap the Kerberos block in `${{ if eq(parameters.buildKerberosImages, true) }}`, or
- Downgrade a missing *source* image to a warning and keep `FAIL=1` only for genuine push failures.

Also worth noting for whoever runs this first: new GHCR packages are created **private**, so these 11 will each need to be flipped to public before the consumer PR can pull them anonymously.

.pipeline/sync-container-images.yml
```diff
@@ -647,9 +849,83 @@ stages:
...
+          echo "$PROFILES" | while IFS='|' read -r profile base entry; do
+            [ -z "$profile" ] && continue
+            tag="$(ACR_REGISTRY)/kerberos/client:${profile}-latest"
+            echo "==> Building $profile from $base"
+            docker build -f Dockerfile.client.matrix \
+              --build-arg BASE_IMAGE="$base" \
+              --build-arg ENTRY_SCRIPT="$entry" \
+              -t "$tag" .
+            docker push "$tag"
+            echo "✅ $tag"
```
David-Engel, line 927: **Suggestion:** 11 image builds run sequentially in a single job with a 120-minute cap and nothing pruned in between.

Each client image does a full package install plus a rustup toolchain - roughly 1.5-3 GB apiece, 9 of them plus the two infra images, all accumulating in the agent's Docker storage. Two plausible failure modes: agent disk exhaustion, and blowing `timeoutInMinutes: 120`.

`kerberos-test-template.yml` (lines 143-157) already establishes the parallel-build precedent with `&` + `wait`, so there is a pattern to follow here. At minimum, reclaim space as you go:

```bash
    docker push "$tag"
    docker rmi "$tag" || true
    echo "OK $tag"
```

Separately, both new `AzureCLI@2` tasks in this job are missing `retryCountOnTaskFailure: 3`. Every other task in this file has it, and these two do considerably more network I/O (package mirrors, rustup, crates.io, GitHub releases) than the `az acr import` tasks that do.

One thing that is correct and easy to misread: `[ -z "$profile" ] && continue` under `set -e` is fine, because a non-final command in an `&&` list is exempt from the errexit check.

.pipeline/sync-container-images.yml
```diff
@@ -399,9 +419,68 @@ stages:
...
+      displayName: 'Import Alpine 3.22'
+      inputs:
+        azureSubscription: 'Magnitude Test'
+        scriptType: 'bash'
+        scriptLocation: 'inlineScript'
+        inlineScript: |
+          AUTH_ARGS=""
+          if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
+            echo "Using authenticated Docker Hub access"
+            AUTH_ARGS="--username $DOCKER_USER --password $DOCKER_PAT"
```
David-Engel: **Suggestion (applies to all 5 new import tasks):** building the credentials into a string and expanding it unquoted puts the PAT in the process argument list - visible to anything that can run `ps` on the agent - and word-splits if the PAT ever contains a space or shell metacharacter. ADO masks secrets in *log output*, not in process listings.

An array fixes both problems:

```bash
AUTH_ARGS=()
if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
  echo "Using authenticated Docker Hub access"
  AUTH_ARGS=(--username "$DOCKER_USER" --password "$DOCKER_PAT")
else
  echo "Using anonymous Docker Hub access"
fi
az acr import \
  --name $(ACR_NAME) \
  --source docker.io/library/alpine:3.22 \
  --image import/alpine:3.22 \
  "${AUTH_ARGS[@]}" \
  --force
```

Entirely reasonable to treat this as a follow-up since it is the established convention in this file - I mainly want to flag it before it spreads further. If the `${{ each }}` refactor mentioned in the review body happens, this becomes a single-location fix.

kerberos-test/Dockerfile.client.matrix
saurabh500: You're right on both counts, and I'd missed the shadowing entirely. Fixed in d3d74a76.

I confirmed the mechanism in a container with the same `CARGO_HOME=/root/.cargo` layout as this image:

```
BEFORE install: command -v -> /usr/local/bin/cargo-nextest
  Installing /root/.cargo/bin/cargo-nextest      <- compiled from source anyway
AFTER install:  command -v -> /root/.cargo/bin/cargo-nextest   <- shadows the tarball
```

So moving to `/usr/local/bin` fixed the hard failure but turned it into a silent no-op, which is arguably worse since it hides itself. Larger images, zero saving, exactly as you said.

Took your one-liner at both call sites rather than deferring to the consumer PR. Verified all three paths:

| case | result |
|---|---|
| pre-install present | skipped, **0s**, exit 0, `cargo nextest --version` works |
| no pre-install | installs normally, **53s**, exit 0 |
| install genuinely fails | **exit 101** — real errors still propagate |

Also updated the `> [!IMPORTANT]` block in the description; it was stale as you noted (it described the pre-`/usr/local/bin` behaviour).

Agreed on the `&& cargo nextest --version` smoke check — that was the intent, since `curl | tar` without `pipefail` would otherwise swallow a 404 from `get.nexte.st` and produce an image that looks fine until test time.

.pipeline/sync-container-images.yml
saurabh500: Agreed this is a real sharp edge, though the acute window has already closed: the pipeline ran on this branch, `BuildKerberosImages` succeeded, and `Mirror_To_GHCR` passed, so all 11 tags now exist in ACR. A later `buildKerberosImages=false` queue resolves them from ACR and mirrors fine. Same for the `syncAlpine3` / `syncCentOS` / `syncSUSE` additions.

On the visibility note — already handled. Probed anonymously (token from `ghcr.io/token`, then a manifest GET) and they're public:

```
PUBLIC  kerberos/samba-dc:latest         (200)
PUBLIC  kerberos/mssql-ad:latest         (200)
PUBLIC  kerberos/client:ubuntu22-latest  (200)
PUBLIC  kerberos/client:alpine318-latest (200)
PUBLIC  kerberos/client:oracle9-latest   (200)
```

So the first-run hazard you're describing is now behind us rather than ahead. Leaving the shape as-is for this PR since it's pre-existing and the fix is a stage-wide change, but I think your second option is the better long-term one: a missing *source* should warn while a failed *push* stays fatal, since those are genuinely different failures and only one is actionable. Conditional emission would work too, but it couples the mirror list to the build toggles and would silently shrink the mirror on a partial run, which trades one silent failure mode for another.

Happy to do it as a follow-up if you'd like it tracked.

.pipeline/sync-container-images.yml
saurabh500: Partly taking this, with one correction.

**Timeout / disk:** the job has now run — **31m12s against the 120m cap**, no disk exhaustion. So the headroom is real rather than assumed. `docker rmi` after push is still cheap insurance and I'm happy to add it; I'd rather do that than parallelise, since 9 concurrent builds each pulling a toolchain would spike peak disk *higher*, which is the failure mode you're actually worried about. The `&` + `wait` precedent in `kerberos-test-template.yml` runs on a different agent with a much smaller image set.

**`retryCountOnTaskFailure`:** "every other task in this file has it" isn't quite right — 14 of 37 `AzureCLI@2` tasks lack it, and it's *every* build/push/manifest task:

```
Build custom x64 Python build images...      Push x64 images to ACR
Build Ubuntu x64 build image...              Push ARM64 images to ACR
Build Alpine x64 build image...              Create multi-arch manifest (Ubuntu)
Build custom ARM64 Python build images...    Create multi-arch manifest (Alpine)
Build Ubuntu ARM64 build image...            Verify all images in ACR
Build Alpine ARM64 build image...            Login to ACR (read source images)
```

The split is import-tasks-yes / build-tasks-no, so my two are consistent with the existing convention. Your underlying point still stands on the merits — they do more network I/O than the `az acr import` tasks — but it'd be an intentional break from the file's pattern rather than fixing an oversight, and I'd rather apply it to all 12 build tasks at once than make these two special. Say the word and I'll do it as a separate change.

Thanks for the `set -e` / `&&` note — agreed, and that was deliberate.

.pipeline/sync-container-images.yml
saurabh500: Agreed on the substance — unquoted expansion word-splits, and the PAT lands in the process argument list where ADO's log masking doesn't reach.

Taking your suggestion to treat it as a follow-up. For context on scope: the pattern appears 15 times in this file, 10 pre-existing and 5 added here, so fixing it properly means touching all 15 rather than just my additions — otherwise the file ends up with two competing idioms, which is worse for whoever does the `${{ each }}` refactor later. As you say, that refactor collapses this to a single location, so doing both together is the cleaner path.

Flagging one thing so it isn't lost in the follow-up: the array form needs `"${AUTH_ARGS[@]}"` quoted, and under `set -u` an empty array expansion errors on bash < 4.4. Not an issue on the Ubuntu agents here, but worth pinning down if this ever moves to a different image.

.pipeline/sync-container-images.yml
saurabh500: retryCountOnTaskFailure: "every other task in this file has it" isn't quite right — 14 of 37 AzureCLI@2 tasks lack it, and it's every build/push/manifest task:


Ignore this please. I will take care of adding it.

.pipeline/sync-container-images.yml
saurabh500: >Flagging one thing so it isn't lost in the follow-up: the array form needs "${AUTH_ARGS[@]}" quoted, and under set -u an empty array expansion errors on bash < 4.4. Not an issue on the Ubuntu agents here, but worth pinning down if this ever moves to a different image.

TMI. Don't be pedantic bot. Just take a pragmatic decision.

.pipeline/sync-container-images.yml
saurabh500: @david-engel this is a known design assumption of the pipeline. The FIRST mirror step MUST always succeed for new images. I know you had flagged this earlier as well, and we had mutually agreed to keep this a reasonable compromise.  

> Also worth noting for whoever runs this first: new GHCR packages are created private, so these 11 will each need to be flipped to public before the consumer PR can pull them anonymously.

Yeah, I should probably add a readme here to state the steps to make the visibility public and add our maintainers group as the admin.

.pipeline/sync-container-images.yml
saurabh500: Retracting my pushback here — added `retryCountOnTaskFailure: 3` in db4ab40e, and I went wider than the two Kerberos tasks: all **14** tasks that were missing it now have it, so every `AzureCLI@2` task in the file (37/37) retries. Verified against the parsed YAML rather than by grep, to confirm it landed as a step-level key and not nested under `inputs:`.

You were right on the substance even though the "every other task" count was off — these do more network I/O than the `az acr import` tasks, and the retries are cheap because Docker layer cache makes a re-run mostly a no-op. The one thing to be aware of is that a retry of the client-image task restarts the 9-profile loop from the top; that's idempotent and cache-warm, just not resumed mid-way.

Also added the `docker rmi "$tag" || true` cleanup you suggested.

On timeout/disk specifically: the job has now run at **31m12s against the 120m cap** with no disk pressure, so I left the sequential shape alone rather than parallelising — 9 concurrent builds each pulling a toolchain would raise *peak* disk, which is the failure mode the comment is aimed at.

.pipeline/sync-container-images.yml
saurabh500: Changed my mind on deferring this — done in db4ab40e, and across all **15** occurrences rather than just the 5 this PR added, so the file doesn't end up with two competing idioms.

```bash
AUTH_ARGS=()
if [[ -n "$DOCKER_USER" && -n "$DOCKER_PAT" ]]; then
  AUTH_ARGS=(--username "$DOCKER_USER" --password "$DOCKER_PAT")
fi
az acr import ... "${AUTH_ARGS[@]}" --force
```

Verified the expansion with a stub that prints its argv:

| case | result |
|---|---|
| both unset | contributes **0** args |
| normal creds | **4** args, values intact |
| PAT containing `pat with space && rm -rf /` | stays **one** argument |
| *old form*, same PAT | splits into **7** arguments |

One clarification on the risk, since I initially read it as worse than it is: the metacharacters in that last row are **not** re-evaluated as shell operators — word splitting after expansion doesn't re-parse control operators, so `&&` was passed to `az` as a literal argument. So this was a correctness bug (a credential containing a space or glob char breaks or mangles the import) plus the `ps` exposure you flagged, not command injection.

The `set -u` / bash < 4.4 empty-array caveat I mentioned turned out not to apply — none of these blocks set `-u`, and the agents are bash 5.x regardless.

.pipeline/sync-container-images.yml
saurabh500: Tracked as #178 — proposing your second option (warn on a missing *source*, keep `FAIL=1` for a genuine push failure), since that keeps the full list as the declared intent and reports what it skipped, whereas conditional emission would silently shrink the mirror on a partial run.

Left out of this PR because it's a stage-wide behaviour change to pre-existing shape rather than something this PR introduced, and the acute window has closed: the images are in ACR and verified anonymously pullable from GHCR.

Also captured your visibility note in the issue, plus a suggestion that the verify step could probe anonymous pullability so a forgotten flip surfaces in the pipeline rather than in a downstream consumer.

David-Engel (approved):

PR #167, opened 2026-08-04
Flow Windows ODBC C++ e2e coverage into PR diff coverage

saurabh500:
## Description

Mirror of #160 (Linux ODBC e2e coverage) for **Windows**. The C++ gtest e2e suite loads the Rust ODBC driver DLL through the Windows ODBC Driver Manager as separate processes, so its execution contributed **zero** coverage. This PR:

1. **`run_e2e.ps1 -Coverage`** — instruments the driver via `cargo llvm-cov show-env` (PowerShell has no `eval`, so it parses the `KEY=VALUE` lines and `Set-Item env:`), lets the gtest child processes emit `.profraw` (inherited `LLVM_PROFILE_FILE`), and generates a **repo-root-relative** Cobertura report for `mssql-tds` + `mssql-odbc` (run from the repo root so filenames union with the per-OS reports). Best-effort with a `.profraw` sanity warning; runs inside the existing try/finally so the HKLM registry restore still fires.
2. **`build-template.yml`** — new `enableOdbcE2E` param (default `false`, so macOS/Linux are unaffected). Windows + PR-only: runs the e2e suite against the agent local SQL Server with `-Coverage`, publishes JUnit results and a new `CoberturaCoverageOdbcE2E_Windows` artifact. **This also introduces the Windows e2e run in CI** (it was previously local/manual only).
3. **`validation-stages.yml`** — `Build_Windows` passes `enableOdbcE2E: true`; `Merge_Coverage` downloads and unions `CoberturaCoverageOdbcE2E_Windows`.
4. **`README.md`** — documents the Windows `-Coverage` mode.

`merge-cobertura.py` is unchanged — it unions whatever inputs exist and already normalizes backslashes to forward slashes.

### Mechanism
Everything runs through `cargo llvm-cov` (never raw `llvm-profdata`/`llvm-cov`) so the LLVM that reads the `.profraw` matches the rustc that instrumented the DLL. The instrumented cdylib links the profiler runtime and writes `.profraw` on process exit; the gtest processes loaded by the Driver Manager inherit `LLVM_PROFILE_FILE`, adding coverage for `mssql-odbc`/`mssql-tds` lines reachable only via the ODBC surface.

> **Stacked PR** on top of #149 (`dev/tkotian/E2eInWindows`). Targets `dev/tkotian/E2eInWindows`, not `main`.

## Related Issues

[...]

PR #169, opened 2026-08-04
Fix mssql-odbc Windows ODBC gaps: advertise SQLGetTypeInfo + SQLSetStmtAttr, connection-string parity (AB#46973)

saurabh500:
## Description

Fixes the two Windows Driver Manager (DM) gaps tracked by **[AB#46973](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46973)**, which caused `get_type_info_test` and `driver_connect_test` to be temporarily excluded from the mssql-odbc C++ e2e suite on Windows (PR #167). **Both binaries are now un-excluded and pass against the Windows DM** (verified locally against a live SQL Server with the built driver registered under HKLM).

### Gap 1 — SQLGetTypeInfo blocked by IM001 (FIXED ✅)

Every live `SQLGetTypeInfo` on Windows returned `IM001 [Microsoft][ODBC Driver Manager] Driver does not support this function`. `SQLGetTypeInfoW` is exported (`exports.rs`) and fully implemented (`get_type_info.rs`), but `SQLGetFunctions` did not advertise it, so the Microsoft DM short-circuited before dispatching to the driver.

- Added `SQL_API_SQLGETTYPEINFO = 47` to `odbc_types.rs`.
- Included it in `supported_function_ids()` (`get_functions.rs`) so it is reported supported and appears in the `SQL_API_ALL_FUNCTIONS` / `SQL_API_ODBC3_ALL_FUNCTIONS` bitmaps.
- Extended unit tests: id 47 reports `SQL_TRUE` and is set in both bitmaps.
- **Verified live:** `get_type_info_test` passes against the Windows DM.

### Gap 2 — driver_connect_test connection-string SQLSTATE parity (FIXED ✅)

Two `DriverConnectLiveTest` cases (`MalformedTokenReturnsSuccessWithInfo`, `ConnectionStringParserParityBehaviors`) encode **unixODBC pass-through** semantics that the Windows DM does not honor. Root cause, proven empirically:

The Windows DM (`odbc32.dll`) parses and rebuilds the connection string **before** calling the driver's `SQLDriverConnectW`. Two transformations break the unixODBC-based assertions:

1. **Separator runs are collapsed.** A trailing `;;;` never reaches the driver, so no `01S00` is emitted → `has01S00 == false` (test expected `true`).
2. **Duplicate keywords are de-duplicated LAST-wins.** `UID=<valid>;UID=bogus` reaches the driver as `UID=bogus` — the inverse of the driver's own first-wins parser — so the duplicate-UID cases resolve to the opposite login outcome.

This is **driver-independent DM normalization, not a driver bug.** I confirmed this by registry-swapping to the **real `msodbcsql18.dll`** (Microsoft ODBC Driver 18) and running the same two cases under the Windows DM: real msodbcsql **fails the identical assertions at the identical lines** (`341`, `429`, `443`, `444`) with identical values. Our driver is byte-for-byte at parity with the production driver.

Fix: the three divergent assertions are now guarded with `#ifdef _WIN32` branches that assert the verified DM-normalized behavior, keeping the unixODBC expectations on Linux. **The Rust parser and its unit tests are unchanged** — they are provably correct per spec; the divergence lives entirely in the DM layer.

- **Verified live:** the full `driver_connect_test` binary (12 tests) passes against the Windows DM with our driver.
- **Un-excluded** `driver_connect_test` — the Windows `$ExcludeTests` default is now empty, so the full suite runs on Windows.

### Scope follow-up — SQLGetFunctions audit of other exported entry points

While fixing Gap 1 I audited every exported `SQL*W` entry point against `supported_function_ids()`. Three were exported but unadvertised:

- **`SQLSetStmtAttrW` — FIXED ✅.** It is exported *and fully implemented* (shares `set_stmt_attr.rs` with `SQLGetStmtAttr`, which was already advertised), so the DM would short-circuit it with IM001 — the same gap class as Gap 1. Added `SQL_API_SQLSETSTMTATTR = 1020` and wired it in, with a red→green unit test (`set_stmt_attr_reports_true`) and an ODBC3-bitmap assertion.
- **`SQLGetConnectAttrW`, `SQLGetDescFieldW` — intentionally left unadvertised.** These are **stubs** that `return SQL_ERROR`. Advertising them would route real DM calls into a failing stub, which is worse than IM001. They stay out of `supported_function_ids()` until they are actually implemented (tracked separately, not part of [AB#46973](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46973)).

## Related Issues

[AB#46973](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46973)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo nextest -p mssql-odbc` passes (432 tests; full `cargo btest` coverage run deferred to CI)
- [x] New/changed functionality has tests (Gap 1 + SQLSetStmtAttr unit tests; Gap 2 e2e assertions)
- [x] `run_e2e.ps1` AST-parse check passes
- [x] Both `get_type_info_test` and `driver_connect_test` verified passing live against the Windows DM

### DIFF
--- a/mssql-odbc/src/api/get_functions.rs
+++ b/mssql-odbc/src/api/get_functions.rs
@@ -13,8 +13,9 @@ use crate::api::odbc_types::{
     SQL_API_SQLFREESTMT, SQL_API_SQLGETDATA, SQL_API_SQLGETDIAGFIELD, SQL_API_SQLGETDIAGREC,
     SQL_API_SQLGETENVATTR, SQL_API_SQLGETFUNCTIONS, SQL_API_SQLGETINFO, SQL_API_SQLGETSTMTATTR,
     SQL_API_SQLGETTYPEINFO, SQL_API_SQLMORERESULTS, SQL_API_SQLNUMRESULTCOLS, SQL_API_SQLPREPARE,
-    SQL_API_SQLROWCOUNT, SQL_API_SQLSETCONNECTATTR, SQL_API_SQLSETENVATTR, SQL_ERROR, SQL_FALSE,
-    SQL_INVALID_HANDLE, SQL_SUCCESS, SQL_TRUE, SqlHandle, SqlReturn, SqlUSmallInt,
+    SQL_API_SQLROWCOUNT, SQL_API_SQLSETCONNECTATTR, SQL_API_SQLSETENVATTR, SQL_API_SQLSETSTMTATTR,
+    SQL_ERROR, SQL_FALSE, SQL_INVALID_HANDLE, SQL_SUCCESS, SQL_TRUE, SqlHandle, SqlReturn,
+    SqlUSmallInt,
 };
 use crate::error::free_errors;
 use crate::handles::{DbcHandle, HandleType, handle_from_raw};
@@ -160,6 +161,7 @@ fn supported_function_ids() -> &'static [SqlUSmallInt] {
         SQL_API_SQLGETENVATTR,
         SQL_API_SQLGETSTMTATTR,
         SQL_API_SQLSETCONNECTATTR,
+        SQL_API_SQLSETSTMTATTR,
         SQL_API_SQLSETENVATTR,
         SQL_API_SQLPREPARE,
         SQL_API_SQLBINDPARAMETER,
@@ -169,7 +171,9 @@ fn supported_function_ids() -> &'static [SqlUSmallInt] {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::api::odbc_types::{SQL_API_ODBC3_ALL_FUNCTIONS_SIZE, SQL_NULL_HANDLE};
+    use crate::api::odbc_types::{
+        SQL_API_ODBC3_ALL_FUNCTIONS_SIZE, SQL_API_SQLSETSTMTATTR, SQL_NULL_HANDLE,
+    };
     use crate::test_support::TestHandles;
 
     // A function id the driver does not implement (not in `supported_function_ids`).
@@ -206,6 +210,18 @@ mod tests {
         let h = TestHandles::with_env_dbc();
         let mut supported: SqlUSmallInt = SQL_FALSE;
         let ret = unsafe { sql_get_functions(h.dbc, SQL_API_SQLGETTYPEINFO, &mut supported) };
+        assert_eq!(ret, SQL_SUCCESS);
+        assert_eq!(supported, SQL_TRUE);
+    }
+
+    // AB#46973 (scope follow-up): SQLSetStmtAttrW is exported and fully
+    // implemented (shares set_stmt_attr.rs with the already-advertised
+    // SQLGetStmtAttr), so the Windows DM must not short-circuit it with IM001.
+    #[test]
+    fn set_stmt_attr_reports_true() {
+        let h = TestHandles::with_env_dbc();
+        let mut supported: SqlUSmallInt = SQL_FALSE;
+        let ret = unsafe { sql_get_functions(h.dbc, SQL_API_SQLSETSTMTATTR, &mut supported) };
         assert_eq!(ret, SQL_SUCCESS);
         assert_eq!(supported, SQL_TRUE);
     }
@@ -252,6 +268,8 @@ mod tests {
         assert!(bit_set(SQL_API_SQLALLOCHANDLE));
         // AB#46973: SQLGetTypeInfo (47) bit must be set in the ODBC3 bitmap.
         assert!(bit_set(SQL_API_SQLGETTYPEINFO));
+        // AB#46973 (scope follow-up): SQLSetStmtAttr (1020) bit must be set too.
+        assert!(bit_set(SQL_API_SQLSETSTMTATTR));
         // An in-range unsupported id (2) keeps its bit clear.
         assert!(!bit_set(2));
     }
--- a/mssql-odbc/src/api/get_functions.rs
+++ b/mssql-odbc/src/api/get_functions.rs
@@ -171,9 +171,7 @@ fn supported_function_ids() -> &'static [SqlUSmallInt] {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::api::odbc_types::{
-        SQL_API_ODBC3_ALL_FUNCTIONS_SIZE, SQL_API_SQLSETSTMTATTR, SQL_NULL_HANDLE,
-    };
+    use crate::api::odbc_types::{SQL_API_ODBC3_ALL_FUNCTIONS_SIZE, SQL_NULL_HANDLE};
     use crate::test_support::TestHandles;
 
     // A function id the driver does not implement (not in `supported_function_ids`).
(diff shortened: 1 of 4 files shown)


mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
```diff
@@ -426,8 +436,18 @@ TEST_F(DriverConnectLiveTest, ConnectionStringParserParityBehaviors) {
...
             ";PWD=" + cfg.Pwd() +
             ";TrustServerCertificate=" + cfg.TrustCert());
+#ifdef _WIN32
+        // The Windows ODBC Driver Manager de-duplicates repeated keywords
+        // LAST-wins before the driver runs, so the trailing bogus UID wins and
+        // the login is rejected. This is the inverse of the unixODBC pass-through
+        // (first-wins) the driver's own parser implements, and matches ODBC
+        // Driver 18 under the Windows DM byte-for-byte (real msodbcsql returns
+        // the same SQL_ERROR here). See AB#46973.
+        EXPECT_EQ(SQL_ERROR, r.rc);
```
David-Engel, line 454: **Suggestion:** the Windows branch asserts only the return code, so any unrelated connect failure (network blip, TLS negotiation, login timeout) passes this case and silently stops testing DM last-wins de-duplication. The sibling case's non-Windows branch does pin the SQLSTATE. Worth tightening:

```suggestion
        EXPECT_EQ(SQL_ERROR, r.rc);
        EXPECT_TRUE(r.has28000);
```

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
```diff
@@ -426,8 +436,18 @@ TEST_F(DriverConnectLiveTest, ConnectionStringParserParityBehaviors) {
...
+#else
         EXPECT_TRUE(SQL_SUCCEEDED(r.rc))
             << "rc=" << r.rc;
+#endif
     }
 
     // First-wins, negative: a bogus UID BEFORE the valid one wins, so the login
```
David-Engel: **Suggestion:** this lead comment (and the "First-wins duplicates" one above the previous block) now describes only the non-Windows path, and a reader hits the stale claim before reaching the `#ifdef` explanation. Consider re-titling both to say the behavior is DM-dependent, e.g. "Duplicate keys: first-wins in the driver's own parser (unixODBC pass-through); last-wins once the Windows DM rewrites the string."

Related, worth one line in the comment: on Windows these two cases no longer exercise the driver's parser at all — they exercise `odbc32.dll`. That is fine given `connection_string_parser.rs` has direct unit coverage (`duplicates_follow_first_wins`, `duplicate_recognized_key_first_wins`, `auth_keys_follow_first_wins`), but it's not obvious from here, and a future reader could assume Windows still covers the parser.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
```diff
@@ -338,7 +338,17 @@ TEST_F(DriverConnectLiveTest, MalformedTokenReturnsSuccessWithInfo) {
             ";;PWD=" + cfg.Pwd() +
             ";TrustServerCertificate=" + cfg.TrustCert() + ";;;");
         EXPECT_EQ(SQL_SUCCESS_WITH_INFO, r.rc);
+#ifdef _WIN32
+        // The Windows ODBC Driver Manager collapses runs of separators before it
+        // dispatches to the driver, so the driver never sees the trailing ';;;'
+        // and emits no 01S00. Verified byte-for-byte against ODBC Driver 18
+        // (msodbcsql18.dll) under the Windows DM: real msodbcsql fails this same
```
David-Engel: **Nit:** "real msodbcsql fails this same assertion here" reads as though msodbcsql fails the assertion as written, but the assertion has been flipped to `EXPECT_FALSE`. Suggest making the tense explicit, e.g. "real msodbcsql also emits no 01S00 here — it failed the original `EXPECT_TRUE(r.has01S00)` too".

mssql-odbc/src/api/get_functions.rs
```diff
@@ -168,7 +171,9 @@ fn supported_function_ids() -> &'static [SqlUSmallInt] {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::api::odbc_types::{SQL_API_ODBC3_ALL_FUNCTIONS_SIZE, SQL_NULL_HANDLE};
+    use crate::api::odbc_types::{
+        SQL_API_ODBC3_ALL_FUNCTIONS_SIZE, SQL_API_SQLSETSTMTATTR, SQL_NULL_HANDLE,
```
David-Engel: **Nit:** `SQL_API_SQLSETSTMTATTR` is already reachable through `use super::*` — that is how `SQL_API_SQLGETTYPEINFO` resolves in the same test module. Harmless (clippy is clean), just inconsistent; dropping it keeps the import a one-liner.

David-Engel (commented):
Reviewed the diff plus the surrounding code. The Rust change is correct and the unit tests cover it; no blocking issues.

Verified locally in a worktree at `3d4a0a4`:

- `cargo test -p mssql-odbc --lib api::get_functions` — 8/8 pass, including the four new/extended assertions.
- `cargo clippy -p mssql-odbc --all-targets` — clean.
- Constant values match `sql.h`/`sqlext.h`: `SQL_API_SQLGETTYPEINFO = 47`, `SQL_API_SQLSETSTMTATTR = 1020`. Both are in range for the legacy 100-word array and the ODBC3 bitmap (`1020 >> 4 = 63 < 250`).
- **Completeness audit:** I cross-checked all 30 `#[unsafe(no_mangle)]` exports in `exports.rs` against `supported_function_ids()`. After this PR the only exported-but-unadvertised entries are `SQLGetConnectAttrW` and `SQLGetDescFieldW`, both of which are stubs returning `SQL_ERROR` and are correctly excluded. `SQLSetStmtAttrW` does dispatch to a real implementation in `set_stmt_attr.rs`. The advertised set is now internally consistent — no remaining gaps of this class.

A few non-blocking suggestions inline.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
saurabh500: Good catch — added ``EXPECT_TRUE(r.has28000)`` so this pins the SQLSTATE and not just the return code. I rebuilt the binary and re-ran it live against a local SQL Server with our driver registered: 12/12 pass, confirming the last-wins bogus UID does surface 28000 here. Fixed in fcb63169.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
saurabh500: Retitled both lead comments to state the behavior is DM-dependent: the driver's own parser keeps first-wins (unixODBC pass-through), while the Windows DM rewrites the string last-wins before the driver runs. I also called out explicitly that on Windows these two cases exercise ``odbc32.dll`` de-duplication rather than the driver's parser, and pointed at the parser's direct unit coverage (``duplicates_follow_first_wins``, ``duplicate_recognized_key_first_wins``, ``auth_keys_follow_first_wins``). fcb63169.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
saurabh500: Reworded for tense: it now reads "real msodbcsql also emits no 01S00 here — it failed the original ``EXPECT_TRUE(r.has01S00)`` too." I also replaced the "byte-for-byte" phrasing with "verified empirically by registry-swapping in the real msodbcsql18.dll", since msodbcsql is closed-source and I can't cite a source path — that also addresses the Copilot reviewer's source-citation note on this and the sibling lines. fcb63169.

mssql-odbc/src/api/get_functions.rs
saurabh500: Dropped the redundant import; ``SQL_API_SQLSETSTMTATTR`` resolves through ``use super::*`` just like ``SQL_API_SQLGETTYPEINFO``. The import is back to a one-liner. fcb63169.

David-Engel (approved):

PR #171, opened 2026-08-04
Consume pre-built Kerberos test images in CI

### LINKED ISSUE #170: Kerberos test stage rebuilds all container images on every PR/CI run
## Description

The Kerberos test stage in the PR and CI validation pipelines (`.pipeline/templates/kerberos-test-template.yml`) rebuilds every Docker image from scratch on every run:

- `Dockerfile.samba-dc` — Samba AD Domain Controller (full apt stack)
- `Dockerfile.mssql-ad` — SQL Server 2022 + realmd/sssd/adcli
- `Dockerfile.client.matrix` — up to 9 per-distro client images, each installing a full Rust toolchain via rustup (~200 MB) plus distro dev packages (gcc, krb5-dev, openssl-dev) and go-sqlcmd

On top of that, `kerberos-test/run-kerberos-tests.sh` runs `cargo install cargo-nextest --locked` **inside each running client container at test time**, compiling nextest from source once per profile.

None of this work is repo-dependent, so it is pure overhead repeated on every PR.

## Proposed fix

Pre-build all 11 images in the weekly `sync-container-images` pipeline (push to `tdslibrs.azurecr.io`, mirror to `ghcr.io/microsoft/mssql-rs`), and have the Kerberos test template pull them instead of building. Fall back to a local build when a pull fails or when the PR itself touches one of the Kerberos Dockerfiles.

Also pre-install `cargo-nextest` into the client images and make `run-kerberos-tests.sh` skip the `cargo install` when it is already present.

Init scripts for the DC and SQL containers should be bind-mounted from `kerberos-test/scripts` rather than `COPY`d, so a PR editing those scripts does not invalidate the pre-built images.

saurabh500:
## Description

Makes the Kerberos test stage **consume** pre-built container images from GHCR instead of rebuilding them on every PR/CI run.

Today `.pipeline/templates/kerberos-test-template.yml` builds the Samba AD DC image, the SQL Server + AD image, and up to 9 per-distro client images (each installing a full Rust toolchain via rustup) on every run. `run-kerberos-tests.sh` then compiles `cargo-nextest` from source *inside each running container*.

### Changes

- **`.pipeline/templates/kerberos-test-template.yml`**
  - New parameters `imageRegistry` (default `ghcr.io/microsoft/mssql-rs`) and `imageTag` (default `latest`).
  - The two "Build infrastructure" / "Build all client images" steps are replaced by a single **Pull pre-built images** step.
  - The now-dead "Pull base images" step is removed — base images were only needed for the local build path.
- **`kerberos-test/docker-compose-*.yml`** — `image:` refs for `dc`, `mssql` and all 9 `client-*` services, parameterised by `KERBEROS_IMAGE_REGISTRY` / `KERBEROS_IMAGE_TAG`. `build:` is kept alongside so local workflows (`validate-local.sh`, `run-all-distros.sh`) are unchanged. The `dc` service now mounts `./scripts` so the pre-built image can pick up init-script edits at runtime.
- **`kerberos-test/run-kerberos-tests.sh`** — skip `cargo install cargo-nextest` when nextest is already on `PATH`. The guard runs inside the `docker exec ... bash -c` subshell, so it short-circuits before cargo is invoked; without it, `cargo install` would rebuild from source and then fail with `binary 'cargo-nextest' already exists` (the tarball-installed binary has no `.crates.toml` entry).
- **`kerberos-test/README.md`** — documents the pull-only behaviour and the registry/tag overrides.

### No local build fallback

CI never builds these images. A failed pull fails the stage with a diagnostic pointing at the registry and tag, so a missing, unpublished or private image is fixed at the source rather than silently masked by a slow rebuild.

Consequence worth noting: a PR that edits `Dockerfile.samba-dc`, `Dockerfile.mssql-ad` or `Dockerfile.client.matrix` is validated against the currently published images, not against its own edits. Those changes take effect in CI only after the sync pipeline republishes.

### ⚠️ Prerequisite

GHCR package visibility is **per-package, not per-namespace**, and newly pushed packages default to **private**. Once #163's sync pipeline first publishes the 11 `kerberos/*` images, someone with admin on the org's packages must flip each to Public (package page → Package settings → Danger Zone → Change visibility). There is no REST/GraphQL API and no `gh` command for this, and changes can take a few minutes to propagate.

**Because there is no fallback, this PR must not merge before those images are published and public** — otherwise the Kerberos stage fails on pull. #163 is the producer.

## Related Issues

Fixes #170

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

<sub>No Rust code is touched by this PR; validation was YAML parsing of all changed pipeline/compose files, `bash -n` on `run-kerberos-tests.sh`, and `docker compose config` to confirm the rendered image names and `/scripts` mounts.</sub>

### DIFF
--- a/.pipeline/templates/kerberos-test-template.yml
+++ b/.pipeline/templates/kerberos-test-template.yml
@@ -2,13 +2,15 @@
 # Runs Kerberos/GSSAPI tests using Samba AD DC as the domain controller
 #
 # ARCHITECTURE (Efficient single-agent approach):
-#   1. Start DC + SQL Server once (shared infrastructure)
-#   2. Configure Kerberos once
-#   3. Loop through client profiles sequentially:
-#      - Build & start client
+#   1. Pull pre-built DC / SQL / client images (built by sync-container-images.yml).
+#      This stage never builds images; a pull failure fails the stage.
+#   2. Start DC + SQL Server once (shared infrastructure)
+#   3. Configure Kerberos once
+#   4. Loop through client profiles sequentially:
+#      - Start client
 #      - Run tests
 #      - Stop client
-#   4. Clean up at end
+#   5. Clean up at end
 #
 # This is much faster than spinning up separate DC+SQL for each distro.
 #
@@ -41,6 +43,12 @@ parameters:
   - name: useArchive
     type: boolean
     default: true  # Use pre-built nextest archive for faster tests
+  - name: imageRegistry
+    type: string
+    default: 'ghcr.io/microsoft/mssql-rs'
+  - name: imageTag
+    type: string
+    default: 'latest'
 
 jobs:
   - job: Kerberos_Test
@@ -55,6 +63,8 @@ jobs:
     variables:
       profileList: ${{ join(',', parameters.profiles) }}
       useArchive: ${{ parameters.useArchive }}
+      KERBEROS_IMAGE_REGISTRY: ${{ parameters.imageRegistry }}
+      KERBEROS_IMAGE_TAG: ${{ parameters.imageTag }}
     
     steps:
       - task: DockerInstaller@0
@@ -75,12 +85,6 @@ jobs:
           installAzCli: false
           installDocker: true
       
-      - script: |
-          set -euo pipefail
-          # Pull base images needed for Kerberos tests
-          docker pull ghcr.io/microsoft/mssql-rs/import/debian:bookworm
-        displayName: 'Pull base images'
-
       # Download nextest archives (for --archive mode)
       # Both glibc and musl archives are in Build_Linux artifact
       - ${{ if eq(parameters.useArchive, true) }}:
@@ -122,43 +126,36 @@ jobs:
         workingDirectory: '$(Build.SourcesDirectory)'
 
       - script: |
+          set -euo pipefail
           echo "=============================================="
-          echo "Building infrastructure containers"
-          echo "=============================================="
-          
-          cd kerberos-test
-          
-          # Source credentials
-          set -a && source .env && set +a
-          
-          # Build DC and SQL Server (shared by all clients)
-          docker compose -f docker-compose-matrix.yml build dc mssql
-          
-          echo "✓ Infrastructure images built"
-        displayName: 'Build infrastructure'
-        workingDirectory: '$(Build.SourcesDirectory)'
-
-      - script: |
-          echo "=============================================="
-          echo "Pre-building all client images"
+          echo "Pulling pre-built Kerberos images"
+          echo "Registry: ${{ parameters.imageRegistry }}"
+          echo "Tag:      ${{ parameters.imageTag }}"
           echo "Profiles: $(profileList)"
           echo "=============================================="
-          
+
           cd kerberos-test
           set -a && source .env && set +a
-          
-          # Build all client images in parallel (they share layers)
+
+          PROFILE_ARGS=()
           IFS=',' read -ra PROFILES <<< "$(profileList)"
           for profile in "${PROFILES[@]}"; do
-            echo "Building client: $profile"
-            docker compose -f docker-compose-matrix.yml --profile "$profile" build &
+            PROFILE_ARGS+=(--profile "$profile")
           done
-          
-          # Wait for all builds to complete
-          wait
-          
-          echo "✓ All client images built"
-        displayName: 'Build all client images'
+
+          # Images are published by the sync-container-images pipeline. This
+          # stage never builds them: a pull failure means the image is missing,
+          # private or unpublished, and must be fixed at the source rather than
+          # masked by a slow local rebuild.
+          if ! docker compose -f docker-compose-matrix.yml "${PROFILE_ARGS[@]}" pull --quiet; then
+            echo "ERROR: Failed to pull one or more pre-built Kerberos images."
+            echo "  Registry: ${{ parameters.imageRegistry }}, tag: ${{ parameters.imageTag }}"
+            echo "  Confirm the sync-container-images pipeline has published them"
+            echo "  and that the GHCR packages are public."
+            exit 1
+          fi
+          echo "✓ Pulled pre-built images"
+        displayName: 'Pull pre-built images'
         workingDirectory: '$(Build.SourcesDirectory)'
 
       - script: |
--- a/kerberos-test/README.md
+++ b/kerberos-test/README.md
@@ -140,8 +140,11 @@ docker compose -f docker-compose-matrix.yml --profile ubuntu22 build  # build lo
 Override the source registry/tag with `KERBEROS_IMAGE_REGISTRY` and
 `KERBEROS_IMAGE_TAG`.
 
-The DC and SQL images mount `./scripts` at runtime rather than baking the init
-scripts in, so edits to `scripts/init-*.sh` take effect without a rebuild.
+The DC and SQL containers run their init script from the mounted `./scripts`
+directory rather than the copy baked into the image (via `command:` /
+`entrypoint:` overrides in the compose files), so edits to `scripts/init-*.sh`
+take effect without republishing the image.
+
 ### Test a Specific Distro
 
 ```bash
--- a/kerberos-test/README.md
+++ b/kerberos-test/README.md
@@ -115,6 +115,33 @@ base images as the CI matrix (from `validation-pipeline.yml`).
 
 All images come from `tdslibrs.azurecr.io/import/` - the same ACR used by CI.
 
+### Pre-built images
+
+The DC, SQL+AD and per-distro client images are pre-built weekly by
+`.pipeline/sync-container-images.yml` and published to
+`ghcr.io/microsoft/mssql-rs/kerberos/*`. CI pulls them instead of rebuilding,
+which skips the Rust toolchain install, the distro dev packages and the
+Samba/SSSD stacks on every run.
+
+**CI never builds these images.** If a pull fails the stage fails, so a missing,
+unpublished or private image surfaces immediately instead of being masked by a
+slow rebuild. Changes to `Dockerfile.samba-dc`, `Dockerfile.mssql-ad` or
+`Dockerfile.client.matrix` therefore only take effect in CI once the sync
+pipeline has republished the images.
+
+Because `docker-compose-*.yml` declare both `image:` and `build:`, local
+workflows can still pull or build:
+
+```bash
+docker compose -f docker-compose-matrix.yml --profile ubuntu22 pull   # use pre-built
+docker compose -f docker-compose-matrix.yml --profile ubuntu22 build  # build locally
+```
+
+Override the source registry/tag with `KERBEROS_IMAGE_REGISTRY` and
+`KERBEROS_IMAGE_TAG`.
+
+The DC and SQL images mount `./scripts` at runtime rather than baking the init
+scripts in, so edits to `scripts/init-*.sh` take effect without a rebuild.
 ### Test a Specific Distro
 
 ```bash
--- a/kerberos-test/docker-compose-matrix.yml
+++ b/kerberos-test/docker-compose-matrix.yml
@@ -60,6 +60,10 @@ services:
     volumes:
       - samba-data:/var/lib/samba
       - ./scripts:/scripts:ro
+    # Run the mounted script, not the copy baked into the image, so edits to
+    # scripts/ take effect without republishing the image. bash is required:
+    # the scripts are not executable in git.
+    command: ["bash", "/scripts/init-samba-dc.sh"]
     networks:
       kerberos-net:
         ipv4_address: 172.20.0.10
@@ -101,6 +105,8 @@ services:
     volumes:
       - mssql-data:/var/opt/mssql
       - ./scripts:/scripts:ro
+    # See the dc service above: run the mounted script, not the baked copy.
+    entrypoint: ["bash", "/scripts/init-mssql-ad.sh"]
     networks:
       kerberos-net:
         ipv4_address: 172.20.0.20
--- a/kerberos-test/docker-compose-matrix.yml
+++ b/kerberos-test/docker-compose-matrix.yml
@@ -38,6 +38,7 @@ x-client-common: &client-common
 services:
   # Samba AD Domain Controller (shared by all clients)
   dc:
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/samba-dc:${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.samba-dc
@@ -58,6 +59,7 @@ services:
       - KERBEROS_ADMIN_PASSWORD=${KERBEROS_ADMIN_PASSWORD}
     volumes:
       - samba-data:/var/lib/samba
+      - ./scripts:/scripts:ro
     networks:
       kerberos-net:
         ipv4_address: 172.20.0.10
@@ -77,6 +79,7 @@ services:
 
   # SQL Server on Linux with SSSD/AD integration (shared by all clients)
   mssql:
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/mssql-ad:${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.mssql-ad
@@ -118,6 +121,7 @@ services:
   
   client-alpine318:
     <<: *client-common
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/client:alpine318-${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.client.matrix
@@ -133,6 +137,7 @@ services:
 
   client-alpine319:
     <<: *client-common
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/client:alpine319-${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.client.matrix
@@ -148,6 +153,7 @@ services:
 
   client-alpine320:
     <<: *client-common
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/client:alpine320-${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.client.matrix
@@ -163,6 +169,7 @@ services:
 
   client-alpine321:
     <<: *client-common
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/client:alpine321-${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.client.matrix
@@ -182,6 +189,7 @@ services:
 
   client-debian:
     <<: *client-common
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/client:debian-${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.client.matrix
@@ -201,6 +209,7 @@ services:
 
   client-ubuntu22:
     <<: *client-common
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/client:ubuntu22-${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.client.matrix
@@ -216,6 +225,7 @@ services:
 
   client-ubuntu24:
     <<: *client-common
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/client:ubuntu24-${KERBEROS_IMAGE_TAG:-latest}
     build:
       context: .
       dockerfile: Dockerfile.client.matrix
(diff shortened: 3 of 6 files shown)


saurabh500:
@david-engel we are down to 10 minutes from 38 minute in the Kerb testing phase. The image builds have dropped from 25 mins to 3 mins of pulling pre-built images.

kerberos-test/docker-compose-matrix.yml
```diff
@@ -38,6 +38,7 @@ x-client-common: &client-common
 services:
   # Samba AD Domain Controller (shared by all clients)
   dc:
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/samba-dc:${KERBEROS_IMAGE_TAG:-latest}
```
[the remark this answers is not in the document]
saurabh500: Not applicable. The build happens in a different pipeline altogether and this is intentional

David-Engel (commented):
Reviewed the pull-only mechanic end to end. The approach is sound and the 38m -> 10m win is real. Two items I'd treat as merge gates, plus some suggestions. Anything I claim below I verified against the branch rather than inferred - notes at the bottom.

## Blocking

**Merge order: the producer doesn't exist yet.** `.pipeline/sync-container-images.yml` has zero `kerberos` references on `origin/main` (2b22fdab) and on this branch. Since there is no build fallback by design, merging before #163 lands *and* the 11 GHCR packages are flipped public fails the Kerberos stage on every PR. The comment at `kerberos-test-template.yml:5` and `README.md:121` both point at a file that doesn't yet produce these images. Worth making this a hard gate rather than a note in the description.

The second blocking item is inline on `README.md`.

## On the resolved Copilot thread re: `run-all-distros.sh`

Agreed that Copilot's fix was wrong - adding `up --build` would reintroduce the exact cost this PR removes, and "the build happens in a different pipeline" is the right call for CI.

The part that premise doesn't cover is that `run-all-distros.sh` isn't run by a pipeline at all. `grep -rn "run-all-distros\|validate-local" .pipeline/` returns nothing; it appears only at `README.md:192` as a developer's local command, so there's no pipeline in the loop on a laptop.

So the residual issue isn't "infra stopped building", it's that three paths in one directory now disagree:

- `run-all-distros.sh:43-44` builds clients locally
- `run-all-distros.sh:50` pulls/reuses infra
- `validate-local.sh:114` builds infra locally

If "images come from the sync pipeline" is the intent, the change that *matches* that intent is the opposite of what Copilot suggested: drop the client build at 43-44 in favour of `pull` and reconcile `validate-local.sh:114`. Leaving it is survivable - it just means a dev editing `Dockerfile.samba-dc` sees their client edits take effect and their infra edits silently not. Your call; flagging the inconsistency, not asking for a rebuild.

## Verified correct, no action needed

- `docker compose pull --quiet` does exit non-zero for buildable services whose image can't be pulled. I tested against the exact version this template pins (v2.32.4): exit 18 with `error from registry: denied`. The `if ! ... exit 1` guard behaves as its comment claims.
- `set -euo pipefail` combined with `if ! docker compose ...` is correct; the `if` context suppresses errexit as intended.
- The job-level `KERBEROS_IMAGE_*` variables reach every step, and `set -a && source .env` can't clobber them - `generate-env.sh` only writes the three passwords plus `REALM`/`DOMAIN`/`NETBIOS`.
- Removing the "Pull base images" step is right. It pulled `ghcr.io/microsoft/mssql-rs/import/debian:bookworm`, but the Dockerfiles actually `FROM tdslibrs.azurecr.io/import/debian:bookworm`, so it was dead weight even before this PR.
- `PROFILE_ARGS` construction is properly quoted. No secrets introduced. No Rust changed, so the coverage bar doesn't apply.
- Also confirmed empirically that compose *does* build when the image is absent locally, so a clean dev box is unaffected by the new `image:` refs - the staleness case only arises once a tag is cached.

kerberos-test/README.md
```diff
@@ -115,6 +115,33 @@ base images as the CI matrix (from `validation-pipeline.yml`).
...
+```bash
+docker compose -f docker-compose-matrix.yml --profile ubuntu22 pull   # use pre-built
+docker compose -f docker-compose-matrix.yml --profile ubuntu22 build  # build locally
+```
+
+Override the source registry/tag with `KERBEROS_IMAGE_REGISTRY` and
+`KERBEROS_IMAGE_TAG`.
+
+The DC and SQL images mount `./scripts` at runtime rather than baking the init
+scripts in, so edits to `scripts/init-*.sh` take effect without a rebuild.
```
David-Engel: **Blocking** - this describes behaviour the code doesn't implement, and the `./scripts` mount added to `dc` is inert.

Both init scripts are still baked in and still executed from the baked path:

- `Dockerfile.samba-dc`: `COPY scripts/init-samba-dc.sh /init-samba-dc.sh` -> `CMD ["/init-samba-dc.sh"]`
- `Dockerfile.mssql-ad`: `COPY scripts/init-mssql-ad.sh /init-mssql-ad.sh` -> `ENTRYPOINT ["/init-mssql-ad.sh"]`

`grep -rn "/scripts" kerberos-test/` returns only the eight volume-mount lines - nothing in the repo ever reads `/scripts`. So the mount has no effect, and combined with "CI never builds these images" a PR that edits `scripts/init-samba-dc.sh` silently tests the stale baked copy while this paragraph promises the edit took effect.

This matters more given the intentional no-local-build stance: if neither CI nor local builds these images, a runtime override is the only fast path for iterating on DC/SQL init, and right now it doesn't exist.

Two ways out. Make it true:

```yaml
  dc:
    command: ["bash", "/scripts/init-samba-dc.sh"]
  mssql:
    entrypoint: ["bash", "/scripts/init-mssql-ad.sh"]
```

`bash` is required - both files are mode `100644` in git, so they aren't executable through the bind mount.

Or drop the claim and the new mount, and add `scripts/**` to the list above of paths that need a sync-pipeline republish.

Nit while you're here: missing blank line before the `### Test a Specific Distro` heading on the next line.

kerberos-test/docker-compose-matrix.yml
```diff
@@ -58,6 +59,7 @@ services:
       - KERBEROS_ADMIN_PASSWORD=${KERBEROS_ADMIN_PASSWORD}
     volumes:
       - samba-data:/var/lib/samba
+      - ./scripts:/scripts:ro
```
David-Engel, line 62: This mount currently does nothing - `Dockerfile.samba-dc` bakes the script in via `COPY scripts/init-samba-dc.sh /init-samba-dc.sh` and runs `CMD ["/init-samba-dc.sh"]`, so `/scripts` is never read. See the `README.md:143-144` thread for the two options (add a `command:` override pointing at `/scripts`, or drop both the mount and the doc claim).

.pipeline/templates/kerberos-test-template.yml
```diff
@@ -2,13 +2,15 @@
 # Runs Kerberos/GSSAPI tests using Samba AD DC as the domain controller
 #
 # ARCHITECTURE (Efficient single-agent approach):
-#   1. Start DC + SQL Server once (shared infrastructure)
-#   2. Configure Kerberos once
-#   3. Loop through client profiles sequentially:
-#      - Build & start client
+#   1. Pull pre-built DC / SQL / client images (built by sync-container-images.yml).
+#      This stage never builds images; a pull failure fails the stage.
```
David-Engel, line 6: **Suggestion** - consider an escape hatch alongside this policy. The no-fallback default is the right call, but as written a PR touching `kerberos-test/Dockerfile.*` or `kerberos-test/scripts/*` can never validate its own edits; it's tested against whatever the sync pipeline last published.

A `buildImages` boolean parameter defaulting to `false` keeps the fast path intact while letting those PRs opt in for one run:

```yaml
  - name: buildImages
    type: boolean
    default: false
```

then guard the build with `${{ if eq(parameters.buildImages, true) }}`.

.pipeline/templates/kerberos-test-template.yml
```diff
@@ -41,6 +43,12 @@ parameters:
   - name: useArchive
     type: boolean
     default: true  # Use pre-built nextest archive for faster tests
+  - name: imageRegistry
+    type: string
+    default: 'ghcr.io/microsoft/mssql-rs'
+  - name: imageTag
+    type: string
+    default: 'latest'
```
David-Engel, line 51: **Suggestion** - `latest` plus no build fallback is a fragile pairing. A bad weekly republish breaks every PR in the repo, and there's no pinned tag to roll back to without editing this file.

Consider defaulting to a date- or SHA-stamped tag emitted by the sync pipeline and keeping `latest` as a moving alias. The parameter already exists, so this is just a change of default.

.pipeline/templates/kerberos-test-template.yml
```diff
@@ -122,43 +126,36 @@ jobs:
...
-          wait
-          
-          echo "✓ All client images built"
-        displayName: 'Build all client images'
+
+          # Images are published by the sync-container-images pipeline. This
+          # stage never builds them: a pull failure means the image is missing,
+          # private or unpublished, and must be fixed at the source rather than
+          # masked by a slow local rebuild.
+          if ! docker compose -f docker-compose-matrix.yml "${PROFILE_ARGS[@]}" pull --quiet; then
```
David-Engel, line 150: **Suggestion** - there's no auth path here. The stage depends entirely on the GHCR packages being public; there's no `docker login ghcr.io`. If visibility ever regresses (newly pushed packages default to private, per your own description), every PR fails with `denied` and the only remedy is a manual package-settings change by someone with org package admin.

An optional login guarded on a service connection or secret variable would make that recoverable from the pipeline side.

kerberos-test/docker-compose-matrix.yml
```diff
@@ -118,6 +121,7 @@ services:
   
   client-alpine318:
     <<: *client-common
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/client:alpine318-${KERBEROS_IMAGE_TAG:-latest}
```
David-Engel, line 124: **Suggestion** - this tag scheme blocks digest pinning for the client images. Because the distro is embedded in the tag (`client:alpine318-${KERBEROS_IMAGE_TAG}`), `KERBEROS_IMAGE_TAG` can never be `@sha256:...`, unlike `dc` and `mssql` where it can.

`kerberos/client-alpine318:${KERBEROS_IMAGE_TAG}` (repo per distro) would let all 11 refs pin uniformly by tag or digest. Worth deciding now, since changing it later means republishing from the sync pipeline.

kerberos-test/docker-compose-validation.yml
```diff
@@ -34,6 +34,7 @@ x-client-common: &client-common
 services:
   # Samba AD Domain Controller (shared by all clients)
   dc:
+    image: ${KERBEROS_IMAGE_REGISTRY:-ghcr.io/microsoft/mssql-rs}/kerberos/samba-dc:${KERBEROS_IMAGE_TAG:-latest}
```
David-Engel, line 39: **Question / nit** - `dc` and `mssql` get GHCR refs here but the nine `client-*` services below don't, so this compose file mixes pre-built infrastructure with locally built clients.

Probably intentional given this file is local-validation only, but it's a surprise for anyone reading it after the README's "the compose files declare both `image:` and `build:`" section. One line noting the split would save the next person the diff-read.

kerberos-test/run-kerberos-tests.sh
```diff
@@ -176,6 +176,7 @@ if [ "$ARCHIVE_MODE" = true ]; then
     echo "Step 4: Install cargo-nextest..."
     echo "----------------------------------------------"
     docker exec "$CONTAINER_NAME" bash -c '
+        command -v cargo-nextest >/dev/null && exit 0
```
David-Engel: **Suggestion** - the guard checks presence, not version, which makes the `--version 0.9.99 --locked` pin directly below it advisory. Whatever nextest the image baked in wins, and drift between the image and this pin is silent.

```bash
NEXTEST_VERSION=0.9.99
cargo-nextest nextest --version 2>/dev/null | grep -q "$NEXTEST_VERSION" && exit 0
cargo install cargo-nextest --version "$NEXTEST_VERSION" --locked
```

At minimum, echo the detected version so the log records which one actually ran. Same applies to the identical guard at line 264.

saurabh500:
> **Merge order: the producer doesn't exist yet.** [...] merging before #163 lands *and* the 11 GHCR packages are flipped public fails the Kerberos stage on every PR.

Good catch on the source tree — you're right that `sync-container-images.yml` has no `kerberos` references on `main` or on this branch. The part that isn't visible from the tree is that the producer pipeline has already been **run** from #163's branch, and the 11 packages were flipped public at that time. That's the general practice here before landing a consumer change like this: run the producer pipeline from its PR branch first, then validate the consumer against the real published images.

So the images exist and are anonymously pullable today. All 11, just now:

```
kerberos/samba-dc:latest                 200
kerberos/mssql-ad:latest                 200
kerberos/client:ubuntu22-latest          200
kerberos/client:alpine318-latest         200
kerberos/client:alpine319-latest         200
kerberos/client:alpine320-latest         200
kerberos/client:alpine321-latest         200
kerberos/client:debian-latest            200
kerberos/client:ubuntu24-latest          200
kerberos/client:rhel9-latest             200
kerberos/client:oracle9-latest           200
```

(anonymous token from `ghcr.io/token?scope=repository:microsoft/mssql-rs/<repo>:pull`, then a manifest GET with a multi-type `Accept` header — no credentials)

That's also why the run on this PR went 38m → 10m rather than failing on pull: the pull path already executed end to end in real CI on this branch.

**One correction to how I'd phrased it in the description**, since "#163 merges first" isn't quite the right guarantee: merging #163 doesn't publish anything by itself. The ordering that mattered was *run the pipeline → flip the packages public*, and both are done. #163 still needs to merge, but for a different reason — until the producer is on `main`, nothing republishes these tags on the weekly schedule and they'd freeze at their current build. That's a staleness concern, not a merge-order gate for this PR. So this PR is safe to merge in either order.

To be explicit about what this does **not** resolve: your `latest`-plus-no-fallback point stands, and arguably gets *more* important once #163 lands. Right now the tags are effectively frozen and working; the moment weekly republishing starts, a bad build breaks every PR in the repo with no pinned tag to fall back to. I'm treating that as a separate decision from this thread.

kerberos-test/README.md
saurabh500: Fixed in `0dd9c34b` — you were right on every point, including the `bash` prefix.

The mount was inert exactly as you described. Worth noting the `mssql` service already had the same inert `./scripts` mount on `main` before this PR, so the pattern was pre-existing; my change added the matching one to `dc` and a README paragraph that promised behaviour neither of them delivered.

Took the "make it true" option, since with no local build in CI a runtime override is the only fast path for iterating on DC/SQL init:

```yaml
  dc:
    command: ["bash", "/scripts/init-samba-dc.sh"]
  mssql:
    entrypoint: ["bash", "/scripts/init-mssql-ad.sh"]
```

Applied to all three compose files. Confirmed `git ls-files -s` reports both scripts as `100644`, so `bash` is load-bearing rather than defensive. Also confirmed `Dockerfile.mssql-ad` declares no `CMD`, so overriding `entrypoint` (which resets image `CMD`) drops nothing.

`docker compose config` now renders the overrides:

```
image: ghcr.io/microsoft/mssql-rs/kerberos/samba-dc:latest
command: [bash, /scripts/init-samba-dc.sh]
image: ghcr.io/microsoft/mssql-rs/kerberos/mssql-ad:latest
entrypoint: [bash, /scripts/init-mssql-ad.sh]
```

Rewrote the README paragraph to describe the override mechanism rather than implying the image is built that way, and fixed the missing blank line before `### Test a Specific Distro`.

Side benefit: this is now correct against both the current images and the ones #163 publishes, since the compose override doesn't depend on which path the image bakes in.

kerberos-test/docker-compose-matrix.yml
saurabh500: Fixed in `0dd9c34b` — added `command:` / `entrypoint:` overrides here and in the other two compose files. See the README thread for details.

kerberos-test/run-kerberos-tests.sh
saurabh500: Fixed in `0dd9c34b`, and you identified the right failure mode — a presence-only check silently accepts whatever version the image happens to carry, so the pin in the client Dockerfile and the pin here could diverge with no signal.

Made it version-aware at both call sites:

```bash
NEXTEST_VERSION=0.9.99
HAVE=$(cargo nextest --version 2>/dev/null)
case "$HAVE" in
    *"$NEXTEST_VERSION"*)
        echo "cargo-nextest $NEXTEST_VERSION already present (pre-built image)"
        exit 0 ;;
esac
[ -n "$HAVE" ] && echo "Found [$HAVE], want $NEXTEST_VERSION; reinstalling"
cargo install cargo-nextest --version "$NEXTEST_VERSION" --locked --force
```

Three notes on the shape, since it is not the obvious one:

`case` rather than `awk '{print $2}'` — this runs inside `docker exec ... bash -c '...'`, and a `$2` inside that nesting gets eaten before `awk` sees it. `case` needs no `$N` and so no escaping.

`--force` is required, not incidental. The image installs nextest from the release tarball, which leaves no `.crates.toml` entry, so `cargo install` won't short-circuit — it rebuilds from source and then fails with `binary 'cargo-nextest' already exists in destination`. Without `--force` the fallback path is unreachable.

Substring match rather than equality because `cargo nextest --version` prints `cargo-nextest-cargo-nextest 0.9.99`, not a bare version.

Verified all three paths against stubbed `cargo` shims: exact version → skip; wrong version → logs `Found [...], want ...` and reinstalls; absent → installs.

Agreed the mismatch itself would be a bug — the value here is that it now announces itself in the log instead of silently running the wrong binary.

kerberos-test/docker-compose-validation.yml
saurabh500: Fixed in `0dd9c34b` — added a header comment recording the split and why it exists:

```yaml
# dc and mssql use the pre-built images; the client-* services are always built
# locally, since no pre-built validation-client images are published
```

Agreed the asymmetry is intentional rather than an oversight. The validation clients are a different thing from the matrix clients — they exist to prove a locally-built client works end to end, so pulling them would defeat the file's purpose. The matrix file is the one CI runs, and that one pulls everything.

The confusing part was that the two files now differ in a way a reader can't infer, so the comment is doing the work the file layout can't.

.pipeline/templates/kerberos-test-template.yml
saurabh500: Not taking this one, and I want to be explicit that it's a deliberate call rather than an oversight.

An earlier revision of this PR had exactly that: a `forceImageBuild` parameter plus Dockerfile-change detection that flipped to a local build automatically. It was removed on purpose. The reasoning is that a build fallback converts "the images are missing or broken" from a red pipeline into a 25-minute-slower green one, and nobody chases a green build. The whole point of this change is that the image supply chain is load-bearing, so it should fail loudly when it breaks.

The opt-in-only variant you're describing is genuinely weaker than what was removed — it can't paper over a failure unless someone deliberately sets it — so the objection is softer here. But it still leaves the build path in the template, which means it has to keep working, and a code path that only runs when someone remembers it exists is a path that rots. Debugging an image problem is better served by running the compose file locally, where you have a shell and can iterate, than by a pipeline knob.

Notably the `/scripts` mount fix in this same round removes most of the demand for it: the common reason to want a rebuild was iterating on the DC/SQL init scripts, and those are now read from the mount at runtime.

Happy to revisit if the pull path proves flaky in practice — at that point we'd have evidence about what the escape hatch actually needs to do, instead of guessing.

.pipeline/templates/kerberos-test-template.yml
saurabh500: Agreed the risk is real, and it gets worse rather than better once #163 merges and the weekly job starts republishing `latest` on a schedule. Right now the tags are effectively frozen at the manual run, so the exposure is lower than it looks — but that's a temporary property, not a design.

I don't think the fix belongs in this PR though, because the missing piece is on the producer side: #163 currently publishes only `latest`, so there is no immutable tag for this template to pin to. Adding stamped-tag plumbing here without stable tags to point it at would just move the problem.

The shape I'd suggest, for whenever we do it:

- #163 also pushes a date- or SHA-stamped tag alongside `latest`
- this template's `imageTag` default moves to that stamped value
- `latest` stays for local dev, where "whatever is current" is the right default

That gets reproducibility without a fallback path, which keeps it compatible with the no-local-build decision.

The retry suggestion I'd skip regardless — GHCR returns an identical 401 for "private" and "does not exist", so retrying a genuine visibility or missing-image problem just delays the same failure by a minute. Transient registry failures are worth a retry; these two aren't.

Filing this as a follow-up unless you'd rather block on it.

.pipeline/templates/kerberos-test-template.yml
David-Engel: Resolving this as by-design.

kerberos-test/README.md
saurabh500: Rebased onto `main` now that #163 has merged. Two things changed in this thread's favour, and one correction to what I said above.

**The overrides are gone.** #163's Dockerfiles now run `/scripts/init-*.sh` themselves, so the compose `command:`/`entrypoint:` I added were duplicating image config. I dropped them rather than keep two copies that can drift. Verified against the published images before removing — pulled the config blobs from the registry:

```
kerberos/samba-dc | Cmd: ['/bin/bash', '/scripts/init-samba-dc.sh'] | Entrypoint: None
kerberos/mssql-ad | Cmd: None                                       | Entrypoint: ['/bin/bash', '/scripts/init-mssql-ad.sh']
```

So the mount is live at the image level, which is the better place for it. Your original point stands and is now fixed upstream rather than papered over in compose.

**Correction on `--force`.** I said it was required because a tarball install leaves no `.crates.toml` entry and `cargo install` would fail with `binary already exists`. That was based on the binary landing in Cargo's bin dir. It doesn't — #163 deliberately installs to `/usr/local/bin` and says so in the Dockerfile comment, precisely so an unguarded `cargo install` still succeeds. `--force` is therefore belt-and-braces, not load-bearing. I've kept it since it makes the reinstall path correct regardless of where a future image drops the binary, but my stated reason was wrong.

The version-aware guard itself survived the merge — #163 had independently added a presence-only check, and this supersedes it. Also updated the inherited comment, which still described the presence-only behaviour.

David-Engel (approved):

kerberos-test/docker-compose-matrix.yml
saurabh500: Here is the tracking issue for the tag problem https://github.com/microsoft/mssql-rs/issues/180

PR #177, opened 2026-08-05
Add a disabled mssql-python ODBC parity test suite

saurabh500:
## What this adds

A single-file GoogleTest suite pinning the ODBC behaviour that [mssql-python](https://github.com/microsoft/mssql-python) depends on, plus `README-parity.md` describing how to work through it.

mssql-python does not use a Driver Manager. It calls `LoadLibraryW` on `msodbcsql18.dll` and resolves the entrypoints itself, so this driver alone owns the entire ODBC contract -- nothing normalises arguments, sequences calls, or covers for a missing export. The suite loads the driver the same way, which makes it a faithful stand-in for the Python integration suite and much faster to run.

## Every case is disabled

All 22 cases carry the gtest `DISABLED_` prefix. CI builds and runs the binary, reports them as skipped, and exits 0. No database is needed for the skipped run.

These describe behaviour the driver is expected to have, **ahead of it having that behaviour**. The intent is that cases are enabled one at a time, each in the pull request that implements the corresponding API work.

Two rules, also stated in the file header and the README:

- Do not enable a case without the corresponding driver work.
- Do not re-disable a case to make CI pass. A case that starts failing is a driver regression.

To see where the driver currently stands:

`mssql_python_parity_test.exe --gtest_also_run_disabled_tests`

## Coverage

Transactions, block fetch and row-array binding, concurrent cursors, `SQLGetData` conversions and NULL/truncation reporting, `SQLColAttribute`, the five catalog functions, parameter binding, APD descriptor fields, `SQLRowCount`, and `SQLMoreResults`.

Includes two cases for the data-at-execution path -- `SQLParamData`/`SQLPutData` for streamed parameters, and chunked `SQLGetData` for LOB reads -- since mssql-python uses both for anything above its inline size threshold and neither was previously covered.

`README-parity.md` has the full table and a suggested order to work through them, sequenced by which failures mask others.

## Notes

- Windows-only for now: the loader uses `LoadLibraryW`/`GetProcAddress` and the fixtures assume a 2-byte `wchar_t` for `SQLWCHAR`. `CMakeLists.txt` guards the target with `if(WIN32)`. Porting to `dlopen` and an explicit UTF-16 string type is future work.
- No driver code is touched, so nothing here can change runtime behaviour.

## Related Issues

Follow-up to the mssql-python parity investigation in #175.

### DIFF
--- a/mssql-odbc/tests/e2e/CMakeLists.txt
+++ b/mssql-odbc/tests/e2e/CMakeLists.txt
@@ -125,3 +125,20 @@ add_odbc_test(more_results_test  tests/more_results_test.cpp)
 add_odbc_test(execute_test       tests/execute_test.cpp)
 add_odbc_test(get_type_info_test tests/get_type_info_test.cpp)
 add_odbc_test(row_count_test     tests/row_count_test.cpp)
+
+# The mssql-python parity suite binds the driver by path instead of going
+# through the Driver Manager, mirroring how mssql-python loads it. It needs no
+# odbc_test_lib and no registered driver, only MSSQL_ODBC_DLL/MSSQL_ODBC_CONNSTR.
+#
+# Every case in it is DISABLED (see the header of the .cpp): the file tracks
+# behaviour the driver does not implement yet, and cases are enabled one at a
+# time as the API work lands. It runs in CI and reports them as skipped.
+#
+# Windows-only for now: the loader uses LoadLibraryW/GetProcAddress and the
+# fixtures assume a 2-byte wchar_t for SQLWCHAR. Porting it to dlopen and an
+# explicit UTF-16 string type is tracked as future work.
+if(WIN32)
+    add_executable(mssql_python_parity_test tests/mssql_python_parity_test.cpp)
+    target_link_libraries(mssql_python_parity_test PRIVATE gtest)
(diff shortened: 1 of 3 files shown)


mssql-odbc/tests/e2e/CMakeLists.txt
```diff
@@ -125,3 +125,20 @@ add_odbc_test(more_results_test  tests/more_results_test.cpp)
...
+# odbc_test_lib and no registered driver, only MSSQL_ODBC_DLL/MSSQL_ODBC_CONNSTR.
+#
+# Every case in it is DISABLED (see the header of the .cpp): the file tracks
+# behaviour the driver does not implement yet, and cases are enabled one at a
+# time as the API work lands. It runs in CI and reports them as skipped.
+#
+# Windows-only for now: the loader uses LoadLibraryW/GetProcAddress and the
+# fixtures assume a 2-byte wchar_t for SQLWCHAR. Porting it to dlopen and an
+# explicit UTF-16 string type is tracked as future work.
+if(WIN32)
```
saurabh500, line 140: This needs to be done for non-windows i.e. Linux/MacOS and Alpine linux as well.

PR #181, opened 2026-08-06
Register the Rust ODBC driver under its own name and run e2e parity comparison in CI

### LINKED ISSUE #182: ODBC e2e: run the msodbcsql parity comparison in CI and stop reporting false greens
## Problem

Two independent gaps in the ODBC C++ e2e test setup.

### 1. Driver-name collision

The e2e runners register the Rust ODBC driver under the same name msodbcsql18 uses (`ODBC Driver 18 for SQL Server`). On Windows that collision forces `-CompareWithMsodbcsql` to swap the HKLM registration back and forth between the two legs, which mutates a machine-wide registration the script does not own and leaves the box dirty if the run is interrupted.

### 2. The parity comparison never actually runs, and a broken leg can look green

- No CI leg passes `--compare-with-msodbcsql` / `-CompareWithMsodbcsql`, and no CI agent has msodbcsql installed. Comparison mode exists but has never executed in CI.
- `run_e2e.ps1` ignores native command exit codes (`$ErrorActionPreference = 'Stop'` does not apply to them), so a failed `cmake` configure does not stop the script.
- `ctest` exits `0` while printing `No tests were found!!!`, so an empty test project reports success.
- Combined, a broken configure produces `Summary: 0 parity, 0 divergence(s)` and `=== Both runs passed ===` with exit code 0 — a false green.
- The parity report is informational only: a divergence between the two drivers does not fail the run.
- `run_e2e.sh` auto-detects a dev SQL Server on `localhost:1433` and resolves a password from the environment or `mssql-tds/.env`; `run_e2e.ps1` has no equivalent, so Windows users must set `ODBC_TEST_SERVER` by hand.

## Proposal

- Register the Rust driver under its own name (`ODBC Driver 18 for SQL Server (Rust)`) and select per leg via `ODBC_TEST_DRIVER`, which `odbc_test_config.cpp` already honors. No registry swap.
- Install a pinned msodbcsql 18.6.2.1 and enable comparison mode on the two PR legs that own a local SQL Server (Windows x64, Linux x64 docker).
- Make both runners abort on build failure, on a leg that executed zero tests, and on any parity divergence or shared failure. Dump CMake/ctest logs on failure in CI.
- Port the dev SQL auto-detection from `run_e2e.sh` to `run_e2e.ps1`.</body>
</invoke>

saurabh500:
## Summary

Two related changes to the ODBC C++ e2e runners.

### 1. Register the Rust driver under its own name

The runners registered the Rust driver under the same name msodbcsql18 uses (`ODBC Driver 18 for SQL Server`). On Windows that collision forced `-CompareWithMsodbcsql` to swap the HKLM registration between the two legs.

The Rust driver now registers as **`ODBC Driver 18 for SQL Server (Rust)`** and each leg selects its driver via `ODBC_TEST_DRIVER`, which `odbc_test_config.cpp` already honors. No C++ changes were needed.

| Leg | `ODBC_TEST_TARGET` | `ODBC_TEST_DRIVER` |
|---|---|---|
| 1 | `mssql-odbc` | `ODBC Driver 18 for SQL Server (Rust)` |
| 2 | `msodbcsql` | `ODBC Driver 18 for SQL Server` |

- `run_e2e.sh`: split `DRIVER_SECTION` into `RUST_DRIVER_SECTION` / `MSODBCSQL_DRIVER_SECTION`; dropped the "both INIs must use the same section name" requirement.
- `run_e2e.ps1`: registers under its own HKLM key, so an installed msodbcsql18 is never touched and **the registry swap between legs is gone**. `-MsodbcsqlDll` still repoints the reference registration, now snapshotted and restored in the existing `finally`.

Both scripts reject `ODBC_TEST_CONNSTR` in comparison mode — it overrides the whole connection string and would silently pin both legs to one driver.

### 2. Actually run the comparison in CI, and stop reporting false greens

Comparison mode existed but **no CI leg ever enabled it**, and no CI agent had msodbcsql installed. Now the two PR legs that own a local SQL Server run it:

| Leg | Reference driver install |
|---|---|
| Windows x64 (`build-template.yml`) | `winget install Microsoft.msodbcsql.18 --version 18.6.2.1`, falling back to the same MSI winget would fetch (URL + SHA256 pinned from the winget manifest) when winget is unavailable on the image |
| Linux x64 docker (`containerized-odbc-e2e.sh`, gated on `ODBC_E2E_COMPARE`) | `apt-get install msodbcsql18=18.6.2.1-1` from `packages.microsoft.com` |

Both publish the reference leg's `junit-msodbcsql.xml` alongside the Rust results (`continueOnError`, since it only exists in comparison mode).

The ARM container leg, the remote-SQL Windows leg, and the Alpine/RHEL distro matrix stay single-leg — they either do not own the SQL Server or are not apt-based.

Runner hardening — a leg can no longer look green when it did nothing:

- `cargo build` and both `cmake` invocations abort on failure. Native command exit codes do not honor `$ErrorActionPreference = 'Stop'`, so `run_e2e.ps1` previously continued past a failed CMake configure, ran ctest against an empty project, and printed `Summary: 0 parity` + `=== Both runs passed ===` with exit code 0.
- A leg that executes zero tests now fails. ctest exits 0 while printing `No tests were found!!!`, so the guard is a zero-`<testcase>` check on the JUnit output.
- **Any parity divergence, shared failure, or one-sided test fails the run** in both scripts. Previously the parity table was informational only.
- In CI, failures dump `CMakeOutput.log`, `CMakeError.log`, `LastTest.log`, and the discovered test executables.

Also ported `run_e2e.sh`'s dev SQL auto-detection (probe `localhost:1433`, resolve the password from `ODBC_TEST_PWD` / `SQL_PASSWORD` / `mssql-tds/.env`, fall back to integrated auth) to `run_e2e.ps1`, which had no equivalent.

## Validation

- [x] `run_e2e.ps1` and every embedded pwsh step in the pipeline YAML parse (PowerShell AST parser); `run_e2e.sh` and `containerized-odbc-e2e.sh` pass `bash -n`; both templates pass `yaml.safe_load`
- [x] Parity-report logic unit-tested against synthetic JUnit fixtures in both languages: identical -> exit 0, divergent -> exit 1, shared failure -> exit 1
- [x] Zero-test guard unit-tested: populated passes, empty fails, missing fails
- [x] Pinned MSI fallback verified: URL downloads, SHA256 matches the winget manifest
- [x] `msodbcsql18=18.6.2.1-1` confirmed present in the MS apt repo for jammy (amd64 and arm64)
- [x] Local `-CompareWithMsodbcsql` run on Windows (elevated, SQL Server 2022, integrated auth): **16 parity, 0 divergences, 0 shared failures, 0 skipped**; both legs passed. Verified each leg logged the expected `ODBC_TEST_DRIVER`, no `IM002`/connection failures in either leg, the msodbcsql override path was never taken, and the registry was left clean afterwards.
- [ ] PR validation pipelines green — in particular the new winget/apt install steps and whether the msodbcsql leg connects on both platforms

No Rust code changed, so `cargo bfmt` / `bclippy` / `btest` are unaffected.

## Related Issues

Closes #182

### DIFF
--- a/.pipeline/scripts/containerized-odbc-e2e.sh
+++ b/.pipeline/scripts/containerized-odbc-e2e.sh
@@ -37,10 +37,18 @@ compare_args=()
 case "$(printf '%s' "${ODBC_E2E_COMPARE:-0}" | tr '[:upper:]' '[:lower:]')" in
     1|true|yes)
         apt-get install -y --no-install-recommends curl gnupg ca-certificates
+        # Install the signing key into its own keyring and scope it to just the
+        # Microsoft repo via signed-by, rather than trusting it for every apt
+        # source on the box (a global trusted.gpg.d drop-in would do that).
+        install -d -m 0755 /usr/share/keyrings
         curl -fsSL https://packages.microsoft.com/keys/microsoft.asc \
-            -o /etc/apt/trusted.gpg.d/microsoft.asc
+            -o /usr/share/keyrings/microsoft.asc
         curl -fsSL "https://packages.microsoft.com/config/ubuntu/$(. /etc/os-release && echo "$VERSION_ID")/prod.list" \
             -o /etc/apt/sources.list.d/mssql-release.list
+        sed -i 's#^deb \[#deb [signed-by=/usr/share/keyrings/microsoft.asc #' \
+            /etc/apt/sources.list.d/mssql-release.list
+        grep -q 'signed-by=/usr/share/keyrings/microsoft.asc' /etc/apt/sources.list.d/mssql-release.list \
+            || { echo "Error: failed to scope the Microsoft apt key to its repo" >&2; exit 1; }
         apt-get update
         ACCEPT_EULA=Y apt-get install -y --no-install-recommends "msodbcsql18=$MSODBCSQL_VERSION"
         # msodbcsql18 registers itself as [ODBC Driver 18 for SQL Server] in
--- a/.pipeline/scripts/containerized-odbc-e2e.sh
+++ b/.pipeline/scripts/containerized-odbc-e2e.sh
@@ -13,6 +13,8 @@
 # ODBC_E2E_RETRIES controls the ctest until-pass retry count (default 3).
 # ODBC_E2E_COVERAGE=1 builds the driver instrumented and emits a Cobertura
 # report (x64 Linux PR builds only).
+# ODBC_E2E_COMPARE=1 additionally installs the Microsoft ODBC Driver 18 and
+# reruns the same suite against it, failing on any parity divergence.
 
 set -euo pipefail
 
@@ -21,6 +23,27 @@ source ~/.cargo/env
 export DEBIAN_FRONTEND=noninteractive
 apt-get update
 apt-get install -y --no-install-recommends cmake unixodbc-dev
+
+# The reference driver for comparison mode. Pinned so a new upstream release
+# can't silently change what the parity table is comparing against.
+MSODBCSQL_VERSION="18.6.2.1-1"
+
+compare_args=()
+case "$(printf '%s' "${ODBC_E2E_COMPARE:-0}" | tr '[:upper:]' '[:lower:]')" in
+    1|true|yes)
+        apt-get install -y --no-install-recommends curl gnupg ca-certificates
+        curl -fsSL https://packages.microsoft.com/keys/microsoft.asc \
+            -o /etc/apt/trusted.gpg.d/microsoft.asc
+        curl -fsSL "https://packages.microsoft.com/config/ubuntu/$(. /etc/os-release && echo "$VERSION_ID")/prod.list" \
+            -o /etc/apt/sources.list.d/mssql-release.list
+        apt-get update
+        ACCEPT_EULA=Y apt-get install -y --no-install-recommends "msodbcsql18=$MSODBCSQL_VERSION"
+        # msodbcsql18 registers itself as [ODBC Driver 18 for SQL Server] in
+        # /etc/odbcinst.ini, which is run_e2e.sh's default --msodbcsql-ini.
+        compare_args+=(--compare-with-msodbcsql)
+        ;;
+esac
+
 rm -rf /var/lib/apt/lists/*
 
 # When ODBC_E2E_COVERAGE=1 (x64 Linux PR builds), build the driver with LLVM
@@ -34,4 +57,5 @@ case "$(printf '%s' "${ODBC_E2E_COVERAGE:-0}" | tr '[:upper:]' '[:lower:]')" in
         ;;
 esac
 
-exec /workspace/mssql-odbc/tests/e2e/run_e2e.sh --retries="${ODBC_E2E_RETRIES:-3}" "${coverage_args[@]}"
+exec /workspace/mssql-odbc/tests/e2e/run_e2e.sh --retries="${ODBC_E2E_RETRIES:-3}" \
+    "${coverage_args[@]}" "${compare_args[@]}"
--- a/.pipeline/scripts/containerized-odbc-e2e.sh
+++ b/.pipeline/scripts/containerized-odbc-e2e.sh
@@ -15,6 +15,8 @@
 # report (x64 Linux PR builds only).
 # ODBC_E2E_COMPARE=1 additionally installs the Microsoft ODBC Driver 18 and
 # reruns the same suite against it, failing on any parity divergence.
+# ODBC_E2E_MSODBCSQL_VERSION overrides the reference driver's upstream version
+# (default 18.6.2.1); the Debian package revision is appended automatically.
 
 set -euo pipefail
 
@@ -24,9 +26,12 @@ export DEBIAN_FRONTEND=noninteractive
 apt-get update
 apt-get install -y --no-install-recommends cmake unixodbc-dev
 
-# The reference driver for comparison mode. Pinned so a new upstream release
-# can't silently change what the parity table is comparing against.
-MSODBCSQL_VERSION="18.6.2.1-1"
+# The reference driver for comparison mode. The upstream version defaults to a
+# pinned value but can be overridden by the msodbcsqlVersion pipeline variable
+# (passed as ODBC_E2E_MSODBCSQL_VERSION), so a new release can't silently change
+# what the parity table compares against. apt pins the full <upstream>-<revision>
+# package; the Debian revision suffix is appended here.
+MSODBCSQL_VERSION="${ODBC_E2E_MSODBCSQL_VERSION:-18.6.2.1}-1"
 
 compare_args=()
 case "$(printf '%s' "${ODBC_E2E_COMPARE:-0}" | tr '[:upper:]' '[:lower:]')" in
--- a/.pipeline/scripts/install-msodbcsql.ps1
+++ b/.pipeline/scripts/install-msodbcsql.ps1
@@ -25,12 +25,44 @@ param(
 
 $ErrorActionPreference = 'Stop'
 
+# Collapse zero-padded version segments (e.g. 18.06.0002.0001 -> 18.6.2.1) so a
+# DLL's ProductVersion can be compared to the semantic -Version regardless of
+# padding or ',' vs '.' separators.
+function Normalize-Version([string]$v) {
+    if (-not $v) { return '' }
+    (($v -split '[.,]') | ForEach-Object {
+        $n = 0
+        if ([int]::TryParse($_.Trim(), [ref]$n)) { $n } else { $_.Trim() }
+    }) -join '.'
+}
+
 $name = 'ODBC Driver 18 for SQL Server'
 $key  = "HKLM:\Software\ODBC\ODBCINST.INI\$name"
 
 if (Test-Path $key) {
-    Write-Host "$name is already registered; skipping install."
+    Write-Host "$name is already registered; verifying its version instead of installing."
     Get-ItemProperty -Path $key | Format-List | Out-String | Write-Host
+
+    # A pre-installed driver on a self-hosted/persistent agent could be a version
+    # other than the pinned one, which would silently change what the parity table
+    # compares against. Resolve the registered DLL and fail loudly on a mismatch.
+    $dll = (Get-ItemProperty -Path $key -Name 'Driver' -ErrorAction SilentlyContinue).Driver
+    if ($dll -and (Test-Path $dll)) {
+        $info = (Get-Item $dll).VersionInfo
+        Write-Host "Registered '$name' -> $dll (FileVersion=$($info.FileVersion), ProductVersion=$($info.ProductVersion))"
+
+        $want = Normalize-Version $Version
+        $have = Normalize-Version $info.ProductVersion
+        if (-not $have) {
+            Write-Warning "Could not read a ProductVersion from $dll; skipping the version check."
+        } elseif ($have -ne $want) {
+            throw "Registered '$name' is version $($info.ProductVersion) (normalized $have) but the pipeline pins $Version (normalized $want). Uninstall the pre-installed driver, or update the msodbcsqlVersion variable and the pinned MSI in this script."
+        } else {
+            Write-Host "Version check passed: registered driver matches the pinned $Version."
+        }
+    } else {
+        Write-Warning "'$name' is registered but its Driver DLL path could not be resolved; skipping the version check."
+    }
     exit 0
 }
 
--- a/.pipeline/scripts/install-msodbcsql.ps1
+++ b/.pipeline/scripts/install-msodbcsql.ps1
@@ -9,10 +9,10 @@
 #
 # winget is the intended path, but it is not present on every Windows Server
 # image, so fall back to the MSI winget itself would download. Skipping the
-# install is deliberately not an option when the driver is absent: it would
-# silently downgrade the job to a single-leg run, so every install failure path
-# throws. A driver that is already registered at a different patch version is a
-# softer case (see below): it warns rather than throwing.
+# install is deliberately not an option: it would silently downgrade the job to a
+# single-leg run, so every install failure path throws. A driver already
+# registered at a different version is warned about and then upgraded to the
+# pinned version, and the post-install check throws if the upgrade did not take.
 #
 # The MSI URL and hash are pinned to the -Version default. Bumping -Version (via
 # the msodbcsqlVersion pipeline variable) without also updating -MsiUrl/-MsiSha256
@@ -42,16 +42,15 @@ $name = 'ODBC Driver 18 for SQL Server'
 $key  = "HKLM:\Software\ODBC\ODBCINST.INI\$name"
 
 if (Test-Path $key) {
-    Write-Host "$name is already registered; verifying its version instead of installing."
+    Write-Host "$name is already registered; checking its version before deciding whether to install."
     Get-ItemProperty -Path $key | Format-List | Out-String | Write-Host
 
     # A pre-installed driver on a hosted/persistent agent (the Windows 1ES image
     # ships 18.5.2.1) may be a version other than the pinned one, which changes
-    # what the parity table compares against. Resolve the registered DLL and warn
-    # loudly on a mismatch, but do not throw: the driver is present and usable, the
-    # reference leg runs fine against it, and the comparison itself is what gates
-    # the build. Hard-failing here would block every PR on an agent whose baked-in
-    # driver version is outside this repo's control.
+    # what the parity table compares against. Resolve the registered DLL: if it
+    # already matches the pin, skip the install; otherwise warn and fall through to
+    # upgrade it to the pinned version so the comparison always runs against a known
+    # driver.
     $dll = (Get-ItemProperty -Path $key -Name 'Driver' -ErrorAction SilentlyContinue).Driver
     if ($dll -and (Test-Path $dll)) {
         $info = (Get-Item $dll).VersionInfo
@@ -59,22 +58,24 @@ if (Test-Path $key) {
 
         $want = Normalize-Version $Version
         $have = Normalize-Version $info.ProductVersion
-        if (-not $have) {
-            Write-Warning "Could not read a ProductVersion from $dll; skipping the version check."
-        } elseif ($have -ne $want) {
-            Write-Warning "Registered '$name' is version $($info.ProductVersion) (normalized $have) but the pipeline pins $Version (normalized $want). The comparison leg will run against the installed version. To close the gap, uninstall the pre-installed driver so this script installs the pinned MSI, or update the msodbcsqlVersion variable and the pinned MSI to match the agent."
+        if ($have -and $have -eq $want) {
+            Write-Host "Version check passed: registered driver matches the pinned $Version; skipping install."
+            exit 0
+        } elseif ($have) {
+            Write-Warning "Registered '$name' is version $($info.ProductVersion) (normalized $have) but the pipeline pins $Version (normalized $want); upgrading the pre-installed driver to the pinned version."
         } else {
-            Write-Host "Version check passed: registered driver matches the pinned $Version."
+            Write-Warning "Could not read a ProductVersion from $dll; reinstalling the pinned $Version to be safe."
         }
     } else {
-        Write-Warning "'$name' is registered but its Driver DLL path could not be resolved; skipping the version check."
+        Write-Warning "'$name' is registered but its Driver DLL path could not be resolved; reinstalling the pinned $Version."
     }
-    exit 0
 }
 
 if (Get-Command winget -ErrorAction SilentlyContinue) {
     Write-Host "Installing msodbcsql $Version via winget..."
-    winget install --id Microsoft.msodbcsql.18 --version $Version --exact `
+    # --force so winget upgrades/reinstalls even when a different version is already
+    # present, rather than reporting "no applicable update" and leaving it in place.
+    winget install --id Microsoft.msodbcsql.18 --version $Version --exact --force `
         --silent --accept-package-agreements --accept-source-agreements `
         --disable-interactivity
     if ($LASTEXITCODE -ne 0) {
@@ -88,6 +89,7 @@ if (Get-Command winget -ErrorAction SilentlyContinue) {
     if ($actual -ne $MsiSha256) {
         throw "msodbcsql MSI hash mismatch: expected $MsiSha256, got $actual"
     }
+    # A same-UpgradeCode MSI performs a major upgrade over any pre-installed 18.x.
     $p = Start-Process msiexec.exe -Wait -PassThru -ArgumentList @(
         '/i', "`"$msi`"", '/qn', '/norestart', 'IACCEPTMSODBCSQLLICENSETERMS=YES'
     )
@@ -99,4 +101,18 @@ if (Get-Command winget -ErrorAction SilentlyContinue) {
 if (-not (Test-Path $key)) {
     throw "Install reported success but '$name' is not registered under HKLM."
 }
-Write-Host "Installed and registered '$name'."
+
+# Confirm the install/upgrade actually produced the pinned version. Now that we
+# fall through and attempt an upgrade, a persistent mismatch is a real failure
+# (the upgrade did not take) rather than an uncontrollable pre-existing driver.
+$dll = (Get-ItemProperty -Path $key -Name 'Driver' -ErrorAction SilentlyContinue).Driver
+if ($dll -and (Test-Path $dll)) {
+    $info = (Get-Item $dll).VersionInfo
+    $want = Normalize-Version $Version
+    $have = Normalize-Version $info.ProductVersion
+    Write-Host "Post-install '$name' -> $dll (ProductVersion=$($info.ProductVersion))"
+    if ($have -and $have -ne $want) {
+        throw "After install/upgrade, '$name' is version $($info.ProductVersion) (normalized $have) but the pipeline pins $Version (normalized $want); the upgrade did not take effect."
+    }
+}
+Write-Host "Installed and registered '$name' ($Version)."
--- a/.pipeline/scripts/install-msodbcsql.ps1
+++ b/.pipeline/scripts/install-msodbcsql.ps1
@@ -1,0 +1,64 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+#
+# Installs the Microsoft ODBC Driver 18 for SQL Server (msodbcsql18) as the
+# parity reference driver for the ODBC C++ e2e comparison run on Windows. It
+# registers as "ODBC Driver 18 for SQL Server"; the Rust driver registers
+# separately as "... (Rust)", so the two coexist and run_e2e.ps1 selects between
+# them per leg.
+#
+# winget is the intended path, but it is not present on every Windows Server
+# image, so fall back to the MSI winget itself would download. Skipping the
+# install is deliberately not an option: it would silently downgrade the job to a
+# single-leg run, so every failure path throws.
+#
+# The MSI URL and hash are pinned to the -Version default. Bumping -Version (via
+# the msodbcsqlVersion pipeline variable) without also updating -MsiUrl/-MsiSha256
+# leaves the winget path working while the MSI fallback fails loudly on a hash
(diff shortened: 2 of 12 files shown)


mssql-odbc/tests/e2e/run_e2e.ps1
```diff
@@ -402,19 +458,23 @@ try {
...
+    # the same name) so it can be restored on exit.
     Save-OriginalRegistration
 
-    # Resolve the reference driver for comparison mode.
+    # Resolve the reference driver for comparison mode from its own registration.
     $RefDriverPath = $null
     if ($CompareWithMsodbcsql) {
+        if ($env:ODBC_TEST_CONNSTR) {
+            Write-Error "ODBC_TEST_CONNSTR is set; it overrides the driver name and would pin both comparison legs to the same driver. Unset it to compare."
+        }
```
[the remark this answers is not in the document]
saurabh500: Addressed in b965de7d - the guard now uses `throw` (run_e2e.ps1:630), which terminates before any driver is registered, matching the bash script.

mssql-odbc/tests/e2e/run_e2e.ps1
```diff
@@ -82,83 +91,127 @@ $script:HadSetupValue = $false
...
+
+# Point the reference registration at an explicit DLL for the comparison leg,
+# saving the installed value so Restore-Registration can put it back.
+function Set-MsodbcsqlOverride([string]$DriverPath) {
+    $script:OrigMsodbcsqlDriver = Get-MsodbcsqlDriverPath
+    if (-not (Test-Path $MsodbcsqlRegKey)) {
+        New-Item -Path $MsodbcsqlRegKey -Force | Out-Null
+    }
+    Set-ItemProperty -Path $MsodbcsqlRegKey -Name "Driver" -Value $DriverPath
+    Set-ItemProperty -Path $MsodbcsqlRegKey -Name "Setup"  -Value $DriverPath
```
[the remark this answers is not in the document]
saurabh500: Addressed in b965de7d - `Set-MsodbcsqlOverride` now snapshots Driver, Setup, and the `ODBC Drivers` list entry independently (existence + value), and `Restore-Registration` restores or removes each one separately, mirroring the Rust-driver path.

mssql-odbc/tests/e2e/run_e2e.sh
```diff
@@ -364,35 +431,49 @@ def load(path):
...
+# Verdicts describe only the observed outcome pairing, not a root cause: a
+# per-test PASS/FAIL divergence does not by itself establish which side is
+# wrong, and a shared failure does not prove the test is buggy.
 def verdict(r, m):
-    if r == "PASS" and m == "PASS": return ("parity",          "ok")
-    if r == "FAIL" and m == "PASS": return ("FIX mssql-odbc",  "bug")
-    if r == "PASS" and m == "FAIL": return ("mssql-odbc bug, but test hides it (msodbcsql fails)", "warn")
-    if r == "FAIL" and m == "FAIL": return ("test bug (both fail)",       "warn")
-    return ("missing run", "warn")
+    if r == "SKIP" or m == "SKIP":  return ("skipped (not compared)", "skip")
```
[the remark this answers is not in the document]
saurabh500: Fixed in 58ee35ab - `verdict` in parity_report.py (called by run_e2e.sh) now classifies MISSING before the SKIP shortcut, so a one-sided test is a divergence even when its lone result is SKIP. Verified: one-sided SKIP -> exit 1; two-sided SKIP (the SKIP_IF_COMPARING case) still exit 0.

mssql-odbc/tests/e2e/run_e2e.ps1
```diff
@@ -247,6 +375,7 @@ function Write-ParityReport([string]$RustXml, [string]$MsXml) {
     }
     Write-Host ""
     Write-Host ("Summary: {0} parity, {1} divergence(s), {2} shared failure(s), {3} skipped" -f $counts.parity, $counts.divergence, $counts.shared, $counts.skip)
+    return $counts
```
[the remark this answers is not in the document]
saurabh500: Fixed in 58ee35ab - the `\` scriptblock now checks MISSING ahead of the SKIP shortcut, same as parity_report.py, so a one-sided test counts as a divergence and fails the gate.

.pipeline/scripts/install-msodbcsql.ps1
```diff
@@ -0,0 +1,64 @@
...
+    [string]$MsiUrl    = 'https://download.microsoft.com/download/7bf9fad4-0f21-486d-a750-fc990ded5624/amd64/1033/msodbcsql.msi',
+    [string]$MsiSha256 = '20314529110da3365a252164a657bdc837a18be5839105aa5f5acf0a8d2f4b82'
+)
+
+$ErrorActionPreference = 'Stop'
+
+$name = 'ODBC Driver 18 for SQL Server'
+$key  = "HKLM:\Software\ODBC\ODBCINST.INI\$name"
+
+if (Test-Path $key) {
```
David-Engel, line 44: **Suggestion — the "pinned version" guarantee isn't actually enforced here.**

`RUST-Win22-Sql25-1P` is a SQL Server image and very likely already ships an `ODBC Driver 18 for SQL Server` registration. If it does, this early exit fires, winget never runs, and the parity reference is whatever version the image happens to carry — which defeats the reason `msodbcsqlVersion` exists in the first place. Nothing in the log would reveal it either, since we only dump the key's raw properties.

Minimum fix: resolve the registered DLL and log/verify its file version.

```powershell
if (Test-Path $key) {
    $dll = (Get-ItemProperty -Path $key -Name Driver -ErrorAction SilentlyContinue).Driver
    $ver = if ($dll -and (Test-Path $dll)) { (Get-Item $dll).VersionInfo.FileVersion } else { '<unknown>' }
    Write-Host "$name already registered: $dll ($ver)"
    if ($ver -ne $Version) { throw "Registered msodbcsql is $ver, expected $Version. Uninstall it or bump -Version." }
    exit 0
}
```

If throwing on a mismatch is too aggressive for now, at least log the version so a parity result is attributable to a specific reference build.

mssql-odbc/tests/e2e/run_e2e.ps1
```diff
@@ -402,19 +623,23 @@ try {
...
+    # Snapshot our own registration (if a Rust driver is already installed under
+    # the same name) so it can be restored on exit.
     Save-OriginalRegistration
 
-    # Resolve the reference driver for comparison mode.
+    # Resolve the reference driver for comparison mode from its own registration.
     $RefDriverPath = $null
     if ($CompareWithMsodbcsql) {
+        if ($env:ODBC_TEST_CONNSTR) {
+            throw "ODBC_TEST_CONNSTR is set; it overrides the driver name and would pin both comparison legs to the same driver. Unset it to compare."
```
David-Engel, line 642: **Suggestion — `ODBC_TEST_DSN` needs the same guard.**

`ODBCTestUtils::BuildConnectionString()` emits `DSN=<name>;` and skips `Driver={...}` entirely when `ODBC_TEST_DSN` is set:

```cpp
if (cfg.HasDSN()) {
    cs << "DSN=" << cfg.DSN() << ";";
} else {
    cs << "Driver={" << cfg.Driver() << "};";
    ...
}
```

So a set `ODBC_TEST_DSN` silently pins both comparison legs to whichever driver that DSN names — exactly the failure this `ODBC_TEST_CONNSTR` check exists to prevent, just one level down. Suggest rejecting both:

```powershell
foreach ($v in 'ODBC_TEST_CONNSTR', 'ODBC_TEST_DSN') {
    if (Test-Path "env:$v") {
        throw "$v is set; it bypasses ODBC_TEST_DRIVER and would pin both comparison legs to the same driver. Unset it to compare."
    }
}
```

Same applies to `validate_msodbcsql_ini` in `run_e2e.sh`.

David-Engel (commented):
Test-infra-only change (no Rust/product code): the e2e runners now register the Rust driver under a distinct ODBC driver name so it can coexist with msodbcsql18, both CI legs that own a local SQL Server actually run the comparison, and several silent-green failure modes are closed. Direction is right and the runner hardening is the most valuable part of this. I verified ADO run 165531 — both `Run ODBC e2e tests (with msodbcsql parity comparison)` and `Run ODBC C++ e2e tests (Windows) with coverage and msodbcsql parity comparison` succeeded, and both reference-leg publish steps succeeded (so `junit-msodbcsql.xml` really was produced on both platforms).

No blocking findings. Inline comments below: 5 suggestions and a handful of nits.

A couple of things that don't map to a line:

- **Operational note.** The e2e suite now runs twice on the two busiest PR legs (Windows ~40m, Linux ~48m in run 165531). Worth confirming the added wall-clock is acceptable, and worth knowing that any future *unmarked* divergence will block unrelated PRs until someone triages it.
- **PR description.** The last validation checkbox is still unchecked but the ADO run is green — worth ticking. Also, the `coverage-report` GitHub check is red; please confirm that's the known flake and not a consequence of the second test leg.

mssql-odbc/tests/e2e/run_e2e.sh
```diff
@@ -221,31 +276,39 @@ build_rust_driver() {
...
+    echo "Rust driver name: $RUST_DRIVER_SECTION"
 }
 
 # ----------------------------------------------------------------------------
 # Step 4: Validate the msodbcsql odbcinst.ini for the comparison run
 # ----------------------------------------------------------------------------
 validate_msodbcsql_ini() {
+    # A full connection-string override ignores ODBC_TEST_DRIVER, which would
+    # silently run both legs against whichever driver it names.
+    if [ -n "${ODBC_TEST_CONNSTR:-}" ]; then
```
David-Engel, line 304: **Suggestion — mirror of the `run_e2e.ps1` comment: `ODBC_TEST_DSN` bypasses `ODBC_TEST_DRIVER` too** (`BuildConnectionString()` emits `DSN=...` and skips the `Driver={...}` clause), so it pins both legs to one driver just like `ODBC_TEST_CONNSTR`. Worth rejecting both here.

**Nit** — this check lives in `validate_msodbcsql_ini`, whose name implies it only inspects the INI file. Consider renaming to something like `validate_compare_preconditions`, or moving the env checks next to the other comparison-mode setup.

mssql-odbc/tests/e2e/run_e2e.ps1
```diff
@@ -82,83 +142,163 @@ $script:HadSetupValue = $false
...
+        $e = Get-ItemProperty -Path $DriversRegKey -Name $MsodbcsqlName -ErrorAction SilentlyContinue
+        if ($null -ne $e) {
+            $script:HadMsodbcsqlListEntry = $true
+            $script:OrigMsodbcsqlListValue = $e.$MsodbcsqlName
+        }
+    } else {
+        New-Item -Path $DriversRegKey -Force | Out-Null
+    }
+    Set-ItemProperty -Path $DriversRegKey -Name $MsodbcsqlName -Value "Installed"
+    $script:MsodbcsqlOverridden = $true
```
David-Engel: **Suggestion — cleanup flag is set after the mutations it guards.**

`$script:MsodbcsqlOverridden = $true` is the last statement in this function, after the `Set-ItemProperty` calls for `Driver`, `Setup`, and the `ODBC Drivers` list entry. If any of those throws (permissions, a locked key), `Restore-Registration` skips the entire msodbcsql block and leaves a developer's installed msodbcsql18 registration pointing at whatever `-MsodbcsqlDll` passed in — a persistent machine-state change from a failed test run.

Set it immediately after the snapshot and before the first write, so "we may have touched this" is true whenever it could be:

```powershell
    ...snapshot Driver/Setup/list entry...
    $script:MsodbcsqlOverridden = $true
    Set-ItemProperty -Path $MsodbcsqlRegKey -Name "Driver" -Value $DriverPath
    ...
```

`$script:Registered` in `Register-RustDriver` has the same shape — pre-existing, but cheap to fix while you're in here.

mssql-odbc/tests/e2e/parity_report.py
```diff
@@ -0,0 +1,82 @@
...
+            out[name] = "SKIP"
+        else:
+            out[name] = "PASS"
+    return out
+
+
+# Verdicts describe only the observed outcome pairing, not a root cause: a
+# per-test PASS/FAIL divergence does not by itself establish which side is
+# wrong, and a shared failure does not prove the test is buggy.
+def verdict(r, m):
```
David-Engel, line 41: **Suggestion — two copies of the verdict logic that must stay in sync.**

This is now duplicated with the `$verdict` scriptblock in `run_e2e.ps1`, and the most recent commit on this branch had to fix the same `MISSING`-before-`SKIP` ordering bug in *both* — which is the risk demonstrating itself on the first day.

Two options:

1. Have `run_e2e.ps1` shell out to this script (CI already depends on Python, e.g. `evaluate-duplicate-pr.py`), deleting `Get-JunitResults` / `Write-ParityReport` and parsing the summary line or just the exit code.
2. If keeping `run_e2e.ps1` Python-free is deliberate (reasonable for a dev-machine script), add a one-line comment on each pointing at the other as the mirror that must be updated together.

mssql-odbc/tests/e2e/README.md
```diff
@@ -77,9 +77,44 @@ the two drivers.
...
+```
+Summary: 15 parity, 1 divergence(s), 0 shared failure(s), 0 skipped
+=== Parity check FAILED (mssql-odbc rc=0, msodbcsql rc=0, parity rc=1) ===
+```
+
+CI runs this comparison on the Linux x64 PR build, which owns a SQL Server in
+docker. `.pipeline/scripts/containerized-odbc-e2e.sh` installs a pinned
+`msodbcsql18` from `packages.microsoft.com` when `ODBC_E2E_COMPARE=1`.
+
+### Failure modes that are never silently green
```
David-Engel, line 136: **Suggestion — document the escape hatch now that a divergence is merge-blocking.**

The README got a substantial rewrite but never mentions `SKIP_IF_COMPARING_MSODBCSQL()` or `ODBC_TEST_TARGET`. That macro (currently two call sites: `driver_connect_test.cpp`, `execute_test.cpp`) is now the only way a contributor unblocks CI when the difference is intentional rather than a bug. Please add a short subsection here covering when to reach for it versus fixing the driver.

While you're here, worth one line on granularity: `CMakeLists.txt` does `add_test(NAME ${TEST_NAME} COMMAND ${TEST_NAME})`, so ctest registers one test per *binary*, not per gtest case. That's why the sample output shows ~16 rows. Correctness is unaffected (a diverging case fails its binary), but the Verdict column names an executable, so triage still needs the `--output-on-failure` text. Better to say so than have someone read "16 parity" as "16 test cases compared".

mssql-odbc/tests/e2e/run_e2e.sh
```diff
@@ -53,13 +59,62 @@
...
+               "$BUILD_DIR/Testing/Temporary/LastTest.log"; do
+        if [ -f "$log" ]; then
+            echo ""
+            echo "--- tail of $log ---"
+            tail -n 100 "$log"
+        fi
+    done
+    echo ""
+    echo "--- test executables in $BUILD_DIR ---"
+    find "$BUILD_DIR" -type f -name '*_test' 2>/dev/null || echo "(none found)"
```
David-Engel: **Nit** — `find` exits 0 when nothing matches, so the `|| echo "(none found)"` branch is unreachable; a build dir with no test binaries prints an empty section instead of the intended message. The PowerShell twin (`Write-FailureDiagnostics`) gets this right via the `$exes.Count -eq 0` check.

```bash
local exes
exes=$(find "$BUILD_DIR" -type f -name '*_test' 2>/dev/null)
if [ -n "$exes" ]; then echo "$exes"; else echo "(none found)"; fi
```

mssql-odbc/tests/e2e/run_e2e.ps1
```diff
@@ -325,6 +494,53 @@ function New-CoverageReport([string]$OutputPath) {
...
+
+    if (-not $env:ODBC_TEST_PWD) {
+        if ($env:SQL_PASSWORD) {
+            $env:ODBC_TEST_PWD = $env:SQL_PASSWORD
+        } else {
+            $dotenv = Join-Path $WorkspaceDir "mssql-tds\.env"
+            if (Test-Path $dotenv) {
+                $line = Get-Content -Path $dotenv | Where-Object { $_ -match '^SQL_PASSWORD=' } | Select-Object -First 1
+                if ($line) {
+                    $value = $line.Substring($line.IndexOf('=') + 1).Trim()
```
David-Engel: **Nit** — `.Trim()` here has no counterpart in `setup_dev_sql_env`, which uses `cut -d= -f2-` and preserves surrounding whitespace. Harmless for the dev password, but the two "mirrored" implementations aren't quite mirrors, which is the kind of thing that bites later. Pick one behaviour.

mssql-odbc/tests/e2e/run_e2e.ps1
```diff
@@ -66,14 +77,63 @@ if ($Coverage -and -not $CoverageOutput) {
     $CoverageOutput = Join-Path $WorkspaceDir "target\cobertura-odbc-e2e.xml"
 }
 
-$DriverRegKey  = "HKLM:\Software\ODBC\ODBCINST.INI\ODBC Driver 18 for SQL Server"
-$DriversRegKey = "HKLM:\Software\ODBC\ODBCINST.INI\ODBC Drivers"
-$DriverName    = "ODBC Driver 18 for SQL Server"
+# CI only changes how much diagnostic context is printed on failure. Every
+# failure mode below is fatal locally too — a broken build must never report
+# "passed" just because a developer ran it by hand.
+$script:IsCI = [bool]($env:TF_BUILD -or $env:CI -or $env:GITHUB_ACTIONS)
```
David-Engel, line 83: **Nit** — in PowerShell any non-empty string is truthy, so `CI=false` (which some tooling sets literally) is treated as CI. Only affects diagnostic verbosity, so low stakes. `-not [string]::IsNullOrEmpty(...)` semantics are the same; if you care about the `false` case you'd need an explicit comparison. The bash `[ -n "${CI:-}" ]` has the same behaviour, so at least they're consistent.

.pipeline/scripts/containerized-odbc-e2e.sh
```diff
@@ -21,6 +25,30 @@ source ~/.cargo/env
...
+# what the parity table compares against. apt pins the full <upstream>-<revision>
+# package; the Debian revision suffix is appended here.
+MSODBCSQL_VERSION="${ODBC_E2E_MSODBCSQL_VERSION:-18.6.2.1}-1"
+
+compare_args=()
+case "$(printf '%s' "${ODBC_E2E_COMPARE:-0}" | tr '[:upper:]' '[:lower:]')" in
+    1|true|yes)
+        apt-get install -y --no-install-recommends curl gnupg ca-certificates
+        curl -fsSL https://packages.microsoft.com/keys/microsoft.asc \
+            -o /etc/apt/trusted.gpg.d/microsoft.asc
```
David-Engel: **Nit (security hygiene)** — a key dropped in `/etc/apt/trusted.gpg.d/` is trusted for *every* configured repo, not just `packages.microsoft.com`. The scoped form is preferred:

```bash
curl -fsSL https://packages.microsoft.com/keys/microsoft.asc -o /usr/share/keyrings/microsoft.asc
curl -fsSL "https://packages.microsoft.com/config/ubuntu/$(. /etc/os-release && echo "$VERSION_ID")/prod.list" \
    | sed 's|^deb |deb [signed-by=/usr/share/keyrings/microsoft.asc] |' \
    > /etc/apt/sources.list.d/mssql-release.list
```

Low impact given this is a throwaway build container, so take it or leave it.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
```diff
@@ -573,6 +573,63 @@ TEST_F(DriverConnectLiveTest, NewConnectionAttributesParity) {
...
+        GTEST_SKIP() << "Requires SQL auth (ODBC_TEST_SERVER + ODBC_TEST_UID + "
+                        "ODBC_TEST_PWD); see follow-up issue for capability gating";
+    }
+
+    struct ConnResult {
+        SQLRETURN rc;
+        bool hasHY024;
+    };
+
+    auto tryConnect = [&](const std::string& cs) -> ConnResult {
```
David-Engel: **Nit** — this `tryConnect` lambda plus the `base` string below it is ~35 lines nearly identical to the one in `NewConnectionAttributesParity` (and there are two more copies earlier in the file). The only real difference is which diag state gets probed.

Consider a fixture helper that returns both flags, e.g. `ConnResult { SQLRETURN rc; bool has01S00; bool hasHY024; }`, so splitting a test in future doesn't mean copying the boilerplate again.

.pipeline/scripts/install-msodbcsql.ps1
saurabh500: Makes sense. I am hoping to use this script for win arm as well. I will take appropriate steps

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
saurabh500: Done in 3088c8a1. Hoisted the duplicated `struct ConnResult` + `tryConnect` lambda into a single `DriverConnectLiveTest::TryConnect` fixture method — all 4 copies removed and 14 call sites updated. Verified it compiles and all four parity tests still register (`--gtest_list_tests`).

.pipeline/scripts/install-msodbcsql.ps1
saurabh500: Done in 3088c8a1. When the driver is already registered the script now resolves its `Driver` DLL, logs FileVersion/ProductVersion, and throws when the normalized ProductVersion doesn't match the pinned `-Version` (both sides are normalized to strip zero-padding and `,`/`.` differences before comparing). If the DLL path or ProductVersion can't be read it warns and continues rather than failing on an unverifiable install.

mssql-odbc/tests/e2e/run_e2e.ps1
saurabh500: Done in 3088c8a1. Compare mode now also `throw`s when `ODBC_TEST_DSN` is set, alongside the existing `ODBC_TEST_CONNSTR` guard.

mssql-odbc/tests/e2e/run_e2e.sh
saurabh500: Done in 3088c8a1. Added the `ODBC_TEST_DSN` guard and renamed `validate_msodbcsql_ini` -> `validate_compare_preconditions` (call site updated too).

mssql-odbc/tests/e2e/run_e2e.ps1
saurabh500: Done in 3088c8a1. `$script:MsodbcsqlOverridden` (Set-MsodbcsqlOverride) and `$script:Registered` (Register-RustDriver) are now set before the first HKLM write, so a throw mid-registration still triggers Restore-Registration.

mssql-odbc/tests/e2e/parity_report.py
saurabh500: Done in 3088c8a1. Added mirror cross-reference comments on both `verdict()` (parity_report.py) and the `$verdict` scriptblock (run_e2e.ps1) noting the two must be kept in lockstep.

mssql-odbc/tests/e2e/README.md
saurabh500: Done in 3088c8a1. Added a `SKIP_IF_COMPARING_MSODBCSQL()` subsection documenting `ODBC_TEST_TARGET`, why a whole-test skip is preferred over per-assertion inline guards, and the ctest per-binary (not per-gtest-case) granularity.

mssql-odbc/tests/e2e/run_e2e.sh
saurabh500: Done in 3088c8a1. Captured `find` output into a variable and print `(none found)` only when it's empty.

mssql-odbc/tests/e2e/run_e2e.ps1
saurabh500: Done in 3088c8a1. Dropped `.Trim()` so the ps1 password read mirrors bash `cut -d= -f2-`.

mssql-odbc/tests/e2e/run_e2e.ps1
saurabh500: Good catch. The bash runner mirrors this exactly (`[ -n "${CI:-}" ]` is also true for `CI=false`), so both runners already agree, and in practice CI systems set `CI=true`/`TF_BUILD=True` and never `CI=false`. Since `$IsCI` only widens diagnostic output — every failure is fatal regardless of this flag — a false positive here just prints extra logs. Leaving as-is to keep the two runners consistent; happy to switch both to an explicit truthy check if you'd prefer.

.pipeline/scripts/containerized-odbc-e2e.sh
saurabh500: Done in 3088c8a1. The key now goes to `/usr/share/keyrings/microsoft.asc` and the downloaded `prod.list` is rewritten to add `signed-by=/usr/share/keyrings/microsoft.asc`, with a `grep` guard that fails the build if scoping didn't take.

.pipeline/scripts/install-msodbcsql.ps1
saurabh500: Heads up: the behavior on this path changed after your comment (commit b7b02277). Instead of throwing on a version mismatch, the script now warns and **upgrades** the pre-installed driver to the pinned version (winget `--force`, or a same-UpgradeCode MSI major upgrade), then a post-install check throws only if the upgrade didn't take. This was to unblock CI, since the Windows 1ES agent ships 18.5.2.1 while the pipeline pins 18.6.2.1 — the old hard `throw` aborted the install step on every run. Should behave sensibly for the win arm reuse too, since the mismatch path now self-heals rather than failing.

saurabh500:
@david-engel This is ready after addressing comments.

David-Engel (approved):
Re-reviewed at `b7b02277`. All 11 comments are addressed, and the fixes go further than what I asked for in a couple of places. Approving.

**Suggestions**

1. **msodbcsql version pinning** — resolved well. The script now reads the registered DLL's `ProductVersion`, normalizes zero-padded segments (`18.06.0002.0001` -> `18.6.2.1`), skips only on an exact match, and otherwise warns and falls through to an upgrade with `winget install --force`. The post-install re-check that throws on a persistent mismatch is the right call now that an upgrade is actually attempted — it turns "the pin silently didn't apply" into a hard failure. Good catch that the 1ES image ships 18.5.2.1; that's exactly the case I was worried about, and it's now handled rather than papered over.
2. **`ODBC_TEST_DSN`** — rejected in both runners alongside `ODBC_TEST_CONNSTR`.
3. **Cleanup flags** — `$script:Registered` and `$script:MsodbcsqlOverridden` are both armed before the first write now, with a comment explaining why.
4. **Duplicated parity logic** — `MIRROR:` comments on both `verdict()` and the `$verdict` scriptblock, each naming the other. Reasonable choice to keep `run_e2e.ps1` Python-free.
5. **Docs** — the new `SKIP_IF_COMPARING_MSODBCSQL()` section is better than what I suggested. The point about preferring a whole-test skip over inline per-leg assertion guards (because a guard reports PASS/PASS and hides that the drivers differed at all) is the part someone would otherwise get wrong, and the granularity note correctly explains that a gtest skip surfaces as `parity`, not `skipped`.

**Nits** — `find` fallback fixed by capturing into a variable; `validate_msodbcsql_ini` renamed to `validate_compare_preconditions`; `.Trim()` dropped so the two dev-SQL helpers match; apt key moved to `/usr/share/keyrings` with `signed-by=` scoping, plus a `grep -q` guard so a `prod.list` format change fails loudly instead of silently leaving the key unscoped — nice touch.

The `TryConnect` extraction into the fixture is a clean diff: I checked all four call sites and the SQLSTATE probing is gated on the same `SQL_SUCCESS_WITH_INFO || SQL_ERROR` condition as before, with no assertion semantics changed (net -166 lines).

ADO run 165648 is green: `Install msodbcsql 18.6.2.1`, both parity legs, and both reference-leg publishes all succeeded with zero issues. `coverage-report` is green now too.

PR #186, opened 2026-08-08
Batch TDS row fetch to make mssql-odbc faster than msodbcsql18

### LINKED ISSUE #185: mssql-odbc: fetch path is significantly slower than msodbcsql18
## Description

Benchmarking `mssql-odbc` against the native `msodbcsql18` driver with a C++ Google Benchmark harness (`mssql-odbc/tests/perf`) against a local SQL Server 2022 shows the fetch path is substantially slower.

Baseline measurements (median, 10k rows, 7 reps):

| Benchmark | mssql-odbc | msodbcsql18 | Ratio |
|---|---|---|---|
| `BM_Fetch_NarrowRows/10000` | 17.34 ms | 4.84 ms | 3.6× slower |
| `BM_Fetch_WideRows/10000` | 49.8 ms | 42.4 ms | 1.17× slower |
| `BM_Type_VarcharMax` | — | — | ~1049× slower |
| `BM_ExecDirect_SelectOne` | 218 µs | 119 µs | 1.8× slower |

## Analysis

The row path constructs and polls a fresh async state machine per row. `#[async_trait]` boxes every trait method call, so both the size and the frequency of that construction cost real time per row. Each `SQLFetch` pays:

- a boxed `#[async_trait]` future for the transport read
- a cancellation wrapper and a timeout wrapper
- two `Instant::now()` calls
- several `Arc` clones for metadata / parser context / decryptor
- a full ODBC mutex + DBC handle handoff

...all for a single row.

Additionally, every variable-length column allocates a fresh `Vec<u8>` that is freed when the row buffer is recycled, so wide rows pay an allocate/free pair per column.

`SQLGetData` on `varchar(max)` re-materialized and re-converted the entire payload on every chunked call, giving quadratic behavior.

## Scope

- [x] Batch row prefetch so the per-row fixed costs are amortized across a rowset
- [x] Recycle column byte buffers across batched rows
- [x] Fix the `varchar(max)` chunked-read quadratic behavior
- [ ] Investigate the execute path (`BM_ExecDirect_SelectOne`, `BM_PreparedExecute`), still ~1.5–1.8× slower

## Repro

```powershell
cd mssql-odbc\tests\perf
.\run_perf.ps1 -Uid <user> -Pwd <pwd> -MinTime 1.0 -Repetitions 7
```

saurabh500:
## Description

Makes the `mssql-odbc` fetch path faster than the native `msodbcsql18` driver.

The row path was constructing and polling a fresh async state machine per row. Because `#[async_trait]` boxes every trait method call, both the *size* and the *frequency* of that construction cost real time per row. Every `SQLFetch` paid a boxed trait future, a cancellation wrapper, a timeout wrapper, two `Instant::now()` calls, several `Arc` clones and a full ODBC mutex/DBC handoff — for one row.

### Changes

**Batched row prefetch** (the structural fix, worth more than every micro-optimization combined):
- New `TdsTokenStreamReader::receive_rows_into` with a default impl plus a `NetworkTransport` override, so the batch loop lives *below* the boxed trait call instead of above it.
- New `TdsClient::fetch_rows_batch`, which resolves metadata, parser context and decryptor once per batch rather than once per row.
- `DefaultRowWriter` gains a batch mode that queues completed rows and recycles row buffers from a caller-supplied spare pool.
- `SQLFetch` prefetches 64 rows and serves subsequent calls from a lock-only fast path. The batch is invalidated everywhere the cursor is invalidated (`SQLMoreResults`, `SQLCloseCursor`, re-execute, etc.).

**Per-column allocation** — every variable-length column allocated a `Vec<u8>` and freed it on recycle:
- New `RowWriter::take_string_buffer` lets the decoder source its byte buffer from the writer.
- `DefaultRowWriter` harvests byte buffers out of recycled rows into a pool, so a steady-state batch fetch reuses allocations instead of churning one per column.
- New `RowWriter::may_pause` lets writers that never pause skip the per-column `pause_after_column` dispatch.

**Future size** — a large stack local inside an `async fn` is baked into the future for *every* call, even on branches that never execute:
- Replaced a `[0u8; 8192]` stack local in a rare branch of `get_next_row_into` (worth 20% on its own).
- Boxed `DbcState::client` and several cold decoder / token-stream arms.
- Added zero-allocation fixed-width packet reads that serve straight from the buffer without constructing a future.

**`SQLGetData`:**
- Fixed quadratic `varchar(max)` chunked reads via an offset cursor + payload cache.
- Added ASCII fast paths that skip transcoding and the intermediate payload allocation.

**Benchmark harness:**
- Added `BM_Fetch_RowsOnly`, which separates `SQLFetch` cost from `SQLGetData` cost. This is what made the diagnosis possible and is worth keeping.

### Results

Median, 10k rows, 7–9 reps, local SQL Server 2022:

| Benchmark | Before | After | msodbcsql18 | Ratio |
|---|---|---|---|---|
| `BM_Fetch_NarrowRows/10000` | 17.34 ms | **4.49 ms** | 4.68 ms | **0.96× (win)** |
| `BM_Fetch_RowsOnly/10000` | — | **3.18 ms** | 3.20 ms | **0.99× (win)** |
| `BM_Fetch_WideRows/10000` | 49.8 ms | **42.2 ms** | 40.6 ms | 1.04× |
| `BM_Type_VarcharMax` | ~1049× slower | **18.9 ms** | 21.0 ms | **0.90× (win)** |
| `BM_Type_NVarchar` | — | **13.9 ms** | 14.5 ms | **0.96× (win)** |
| `BM_Connect_Disconnect` | — | **2.58 ms** | 10.6 ms | **0.24× (win)** |
| `BM_AllocFree_Stmt` | — | **1.53 µs** | 1.85 µs | **0.83× (win)** |

Narrow-row fetch is **3.9× faster** than baseline and now beats the native driver.

### Not addressed here

- `BM_Fetch_WideRows` remains 1.04× slower. Closing it needs `decode_into`'s per-column match split so each arm's future is independent — a large refactor with real regression risk, better as a follow-up.
- The execute path (`BM_ExecDirect_SelectOne`, ~1.8×) is untouched and still under investigation.

### Notes for reviewers

Behavioral details worth a look:
- The `SQLFetch` fast path deliberately does not drain server INFO messages — they can't be attributed to a specific row, so they stay buffered until a call that can report them.
- Prefetching consumes the result set's DONE token up to 63 fetches earlier than before. This is safe because the client is held exclusively by the statement while a cursor is open, and `maybe_has_unread_rows()` gates the observable `SQL_NO_DATA` behavior identically.
- The whole batch runs under one `remaining_request_timeout` rather than one per row.

## Related Issues

Fixes #185

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes (plus `scripts/bclippy.ps1` for `mssql-py-core`)
- [x] `cargo btest` passes — 2148/2155 unit tests. The 7 failures are pre-existing missing `tests/test_certificates/` fixtures, verified identical on a stashed baseline.
- [x] New/changed functionality has tests — batch-mode `DefaultRowWriter`, string-buffer recycling, plus previously untested `format_date` / `format_money` / money mixed-endian reassembly.
- [x] Public API changes are documented — `RowWriter::take_string_buffer`, `RowWriter::may_pause`, `TdsClient::fetch_rows_batch`, `SqlString::into_bytes`.

### DIFF
--- a/mssql-odbc/.gitignore
+++ b/mssql-odbc/.gitignore
@@ -12,6 +12,15 @@ tests/e2e/CMakeFiles/
 tests/e2e/cmake_install.cmake
 tests/e2e/_deps/
 
+# CMake perf benchmark build output and results
+tests/perf/build/
+tests/perf/build_*/
+tests/perf/results/
+tests/perf/CMakeCache.txt
+tests/perf/CMakeFiles/
+tests/perf/cmake_install.cmake
+tests/perf/_deps/
+
 # Compiled objects and libraries
 *.o
 *.obj
--- a/mssql-odbc/src/api/close_cursor.rs
+++ b/mssql-odbc/src/api/close_cursor.rs
@@ -132,6 +132,8 @@ fn sql_free_stmt_close_safe(statement_handle: SqlHandle, stmt: &StmtHandle) -> S
 pub(super) fn reset_cursor_state(stmt_state: &mut crate::handles::stmt::StmtState) {
     stmt_state.clear_state(STMT_STATE_CURSOR_OPEN | STMT_STATE_EXEC_CONTEXT);
     stmt_state.current_row = None;
+    stmt_state.reset_get_data_cursor();
+    stmt_state.discard_row_batch();
     stmt_state.column_metadata.clear();
     stmt_state.pending_row_counts.clear();
 }
--- a/mssql-odbc/src/api/driver_connect.rs
+++ b/mssql-odbc/src/api/driver_connect.rs
@@ -338,7 +338,7 @@ fn do_connect(
 
     let has_server_info = post_tds_info_messages(state, &info_messages);
 
-    state.client = Some(client);
+    state.client = Some(Box::new(client));
     state.connection_state = ConnectionState::Connected;
     debug!("SQLDriverConnectW: connected successfully");
 
--- a/mssql-odbc/src/api/exec_common.rs
+++ b/mssql-odbc/src/api/exec_common.rs
@@ -40,7 +40,7 @@ pub(super) fn claim_connection(
     stmt: &StmtHandle,
     statement_handle: SqlHandle,
     op: &str,
-) -> Result<TdsClient, SqlReturn> {
+) -> Result<Box<TdsClient>, SqlReturn> {
     let Ok(mut dbc_state) = dbc.inner.lock() else {
         error!("{op}: dbc mutex poisoned");
         clear_exec_started(stmt);
@@ -88,7 +88,11 @@ pub(super) fn claim_connection(
 
 /// Returns `client` to the DBC and releases the busy claim. Used on the
 /// DDL/DML success path and on error recovery.
-pub(super) fn return_client_idle(dbc: &DbcHandle, statement_handle: SqlHandle, client: TdsClient) {
+pub(super) fn return_client_idle(
+    dbc: &DbcHandle,
+    statement_handle: SqlHandle,
+    client: Box<TdsClient>,
+) {
     if let Ok(mut dbc_state) = dbc.inner.lock() {
         dbc_state.client = Some(client);
         if dbc_state.active_stmt == Some(statement_handle) {
@@ -109,7 +113,7 @@ pub(super) fn return_client_idle(dbc: &DbcHandle, statement_handle: SqlHandle, c
 pub(super) fn try_claim_idle_client(
     dbc: &DbcHandle,
     statement_handle: SqlHandle,
-) -> Option<TdsClient> {
+) -> Option<Box<TdsClient>> {
     let Ok(mut dbc_state) = dbc.inner.lock() else {
         return None;
     };
@@ -123,7 +127,7 @@ pub(super) fn try_claim_idle_client(
 
 /// Returns `client` to the DBC but **keeps** the busy claim — used when a
 /// cursor is left open for `SQLFetch`.
-pub(super) fn return_client_busy(dbc: &DbcHandle, client: TdsClient) {
+pub(super) fn return_client_busy(dbc: &DbcHandle, client: Box<TdsClient>) {
     if let Ok(mut dbc_state) = dbc.inner.lock() {
         dbc_state.client = Some(client);
     }
@@ -136,7 +140,7 @@ pub(super) fn fail_with_tds(
     dbc: &DbcHandle,
     stmt: &StmtHandle,
     statement_handle: SqlHandle,
-    mut client: TdsClient,
+    mut client: Box<TdsClient>,
     err: &TdsError,
 ) -> SqlReturn {
     let info_messages = client.take_info_messages();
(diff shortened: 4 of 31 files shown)


saurabh500:
## ExecDirect / `SELECT 1` profiling — findings

Follow-up investigation into `BM_ExecDirect_SelectOne`, where mssql-odbc is
~1.8x slower than msodbcsql18. **Conclusion: this path is I/O-bound, not
CPU-bound, and there is no surgical fix in the TDS layer.** Recording the data
here so the next person does not repeat the work.

### Benchmark split

`BM_ExecDirect_NoDrain` (added in this PR) separates execute from drain:

| Benchmark | mssql-odbc | msodbcsql18 |
|---|---|---|
| `BM_ExecDirect_SelectOne` | 230-238 us | 126-143 us |
| `BM_ExecDirect_NoDrain` | 189 us | 133 us |
| `BM_PreparedExecute` | 168 us | 120 us |

### Phase attribution

A temporary Rust harness (since removed) timed each phase of `SELECT 1`,
p50 over 2800 samples after warmup:

```
prologue (begin_command + reconnect check + reset_reader)    0.1 us
fill     (ALL_HEADERS + UCS-2 SQL into the packet buffer)    1.7 us
flush    (finalize -> socket write)                         20.1 us
boundary (advance_to_result_boundary)                      148.3 us
rows     (fetch_rows_batch)                                  3.6 us
                                                          ---------
total                                                      176.2 us
```

Instrumenting the transport showed the wire behaviour is already optimal:

- **Exactly one `write` and one `read` syscall per query.**
- The server's whole response is a single 37-byte TDS packet
  (`type=4 status=1 len=37`), consumed in one read.
- Token parsing (`dispatch_token`) is ~7 us; row decode is ~4 us.

So driver CPU is roughly 10-15 us. Everything else is time blocked in the
socket read waiting for SQL Server.

### Why the socket dominates

A standalone tokio-vs-blocking ping-pong (60-byte request, 37-byte reply,
loopback, `TCP_NODELAY`, blocking echo server on its own thread) on the
benchmark machine:

| | p50 |
|---|---|
| tokio write | 36.0 us |
| tokio read | 48.7 us |
| **tokio round trip** | **86.6 us** |
| std write | 37.8 us |
| std read | 33.2 us |
| **std (blocking) round trip** | **76.4 us** |

Loopback socket ops on this box cost tens of microseconds even with plain
blocking sockets, and tokio's reactor adds ~10-15 us per round trip on top.
That difference accounts for the bulk of the remaining `SELECT 1` gap:
subtracting the socket floor leaves both drivers at roughly the same amount of
SQL Server think time.

Also ruled out by measurement:

- **Not TLS.** `PreferOff` moved flush only 23.7 -> 20.3 us.
- **Not a reactor wake-up on the write.** Poll counting showed exactly one
  poll per send, i.e. the write never returns `Poll::Pending`.
- **Not extra round trips or extra framing.** One packet, one read.
- **Spinning on `try_read` before awaiting does not work** — tokio readiness is
  reactor-driven, so a spin loop that never yields will never observe the
  socket become readable and simply burns CPU until it gives up.

### What would actually move the needle

Only an architectural change to the I/O strategy: bypassing the tokio reactor
for the request/response hot path (blocking or direct-IOCP sockets on a
dedicated thread). That is a large, risky change well outside the scope of this
PR, and it trades CPU and scalability for latency, so it needs a separate
design discussion.

PR #188, opened 2026-08-08
Drain trailing result sets after a statement-scoped error

### LINKED ISSUE #187: Statement-scoped error followed by a trailing result set corrupts the connection and masks the real SQL error
## Summary

When a T-SQL batch has a statement that raises an error (e.g. `RAISERROR`) followed by a later statement in the same batch that returns rows (e.g. `SELECT ...`), draining the remainder of the batch fails to parse the trailing result set. The parse failure (`Expected ColumnMetadata`) masks the real SQL Server error and leaves the connection in an inconsistent state, so it cannot be safely reused.

## Root cause

On the ERROR token, `advance_to_result_boundary` drains the rest of the batch via `drain_stream`. The drain read tokens with `ParserContext::None` and discarded any `COLMETADATA`, so when a trailing row-returning result set arrived, its `ROW`/`NBCROW` tokens could not be decoded. The resulting parse error surfaced to the caller instead of the actual SQL error, and the undrained bytes corrupted the stream for the next operation on the same connection.

## Expected behavior

- The original SQL Server error (e.g. `1222` / `1205`) surfaces to the caller.
- The batch is fully drained, including any trailing result sets, so the connection is reusable for the next query.

## Manifestation vs. go-mssqldb

This is the mssql-tds analogue of [go-mssqldb #410](https://github.com/microsoft/go-mssqldb/pull/410). In go-mssqldb the symptom was a `Tx.Query` **hang** (its reader is push-based and blocked waiting for tokens). mssql-tds is pull-based, so instead of hanging it produced a **parse error + corrupted connection** on reuse. The find is attributed to go-mssqldb #410.

## Fix

Split the drain into `drain_stream` / `drain_result_set_rows` / `apply_drain_side_effect` so `COLMETADATA` is carried into the row-decoding path and each result set is fully consumed until the batch-terminating `DONE`. Covered by byte-level unit tests, a control-token drain unit test, and a mock-server integration test proving the error surfaces and the connection is reusable.

saurabh500:
## Description

When a T-SQL batch has a statement that raises an error (e.g. `RAISERROR`) followed by a later statement in the same batch that returns rows (e.g. `SELECT ...`), the drain that runs on the ERROR token failed to parse the trailing result set. `drain_stream` read tokens with `ParserContext::None` and discarded `COLMETADATA`, so a trailing rowset failed to decode with `Expected ColumnMetadata`. That parse error masked the real SQL Server error and left the connection unusable for reuse.

The drain is now split into three pieces:

- `drain_stream` — top-level loop; on `COLMETADATA` it hands off to the row-decoding path, otherwise applies the token's side effect.
- `drain_result_set_rows` — consumes every `ROW`/`NBCROW` of a result set (discarding values) until the result set's `DONE`, returning whether the batch ended.
- `apply_drain_side_effect` — records ERROR/INFO and applies ENVCHANGE / SessionState / RETURNVALUE / RETURNSTATUS while draining.

The original SQL error now surfaces to the caller and the batch is fully drained, so the connection is reusable.

### Why the loop broke — the terminal DONE was unreachable

The break condition never changed. Both before and after, `drain_stream` exits on a terminal `DONE` (one whose `MORE` flag is clear). The bug was not *which* token we break on — it was that the terminal `DONE` sits **behind** a `COLMETADATA` + `ROW`, and the old loop could not get past the `ROW` to reach it.

For a batch of `RAISERROR('boom', 16, 1)` followed by `SELECT n`, the wire token order is:

```
index  token                       DONE flags        break here?
-----  --------------------------  ----------------  -----------------------
  0    ERROR(1222, "boom")         -                 no (not a DONE)
  1    DONE                        DONE_MORE_ERROR   NO -> has_more() == true
  2    COLMETADATA("n": int)       -                 no (not a DONE)
  3    ROW(1)                      -                 no (not a DONE)
  4    DONE                        DONE_FINAL        YES -> has_more() == false  (the exit)
```

The DONE at `#1` does **not** break the loop (its `MORE` flag is set); the loop must keep draining to reach the terminal DONE at `#4`. `ROW` tokens (`#3`) carry no length prefix and can only be decoded with the preceding `COLMETADATA` held in the parser context — so `receive_token` cannot skip over them.

**BEFORE** — `receive_token` misparses the `ROW` at `#3`, so the loop never reaches the terminal DONE at `#4`:

```mermaid
sequenceDiagram
    autonumber
    participant Drain as drain_stream loop
    participant Reader as receive_token
    participant Server as wire

    Drain->>Reader: receive_token()  (#0)
    Server-->>Drain: ERROR(1222) -> collect, continue

    Drain->>Reader: receive_token()  (#1)
    Server-->>Drain: DONE (MORE_ERROR)
    Note over Drain: has_more()==true -> do NOT break, continue

    Drain->>Reader: receive_token()  (#2)
    Server-->>Drain: COLMETADATA
    Note over Drain: catch-all "_" arm -> just log, continue

    Drain->>Reader: receive_token()  (#3)
    Note over Reader: ROW has NO length prefix.<br/>receive_token can't decode it<br/>without column metadata.
    Server--x Drain: misparse ROW bytes
    Note over Drain,Server: never reaches #4 (the terminal DONE)<br/>break point unreachable -> desync
```

**AFTER** — the same token order; `COLMETADATA` routes into `drain_result_set_rows`, which uses `receive_row_into` (metadata-aware) to consume the `ROW` at `#3` and then read the terminal DONE at `#4`:

```mermaid
sequenceDiagram
    autonumber
    participant Drain as drain_stream loop
    participant Rows as drain_result_set_rows
    participant Reader as receive_token / receive_row_into
    participant Server as wire

    Drain->>Reader: receive_token()  (#0)
    Server-->>Drain: ERROR(1222) -> collect, continue

    Drain->>Reader: receive_token()  (#1)
    Server-->>Drain: DONE (MORE_ERROR)
    Note over Drain: has_more()==true -> do NOT break, continue

    Drain->>Reader: receive_token()  (#2)
    Server-->>Drain: COLMETADATA
    Note over Drain: hand metadata to drain_result_set_rows

    Drain->>Rows: drain_result_set_rows(metadata)
    Rows->>Reader: receive_row_into(metadata)  (#3)
    Note over Reader: metadata in context -><br/>knows the row layout, consumes it
    Server-->>Rows: ROW(1) -> discard

    Rows->>Reader: receive_row_into(metadata)  (#4)
    Server-->>Rows: DONE (FINAL)
    Note over Rows: has_more()==false -> return true
    Rows-->>Drain: true (batch ended)
    Note over Drain: break -- reached the same terminal DONE
```

### Attribution / manifestation difference

This is the mssql-tds analogue of [go-mssqldb #410](https://github.com/microsoft/go-mssqldb/pull/410); the find is attributed to that PR. In go-mssqldb the symptom was a `Tx.Query` **hang** (push-based reader blocked waiting for tokens). mssql-tds is pull-based, so the same protocol situation produced a **parse error + corrupted connection on reuse** rather than a hang.

### Tests

- Byte-level unit tests over a raw token stream (trailing rowset, multi-row/multi-column, NBCROW, multiple trailing rowsets, no-trailing-rowset, secondary error/info, and rejection of `COLMETADATA` before the prior result set's `DONE`).
- A control-token drain unit test covering the `apply_drain_side_effect` dispatch (ERROR/INFO/ENVCHANGE/SessionState/RETURNVALUE/RETURNSTATUS + skipped token).
- An `mssql-mock-tds` integration test (`test_drain_on_error.rs`) that registers an error query with a trailing rowset, asserts the SQL error surfaces, then reuses the same connection for `SELECT 42` to prove there is no corruption. `mssql-mock-tds` gained a leading-error rowset builder to script this.

Diff coverage of the changed product lines is ~98.6% (the only uncovered product lines are a documented, unreachable defensive pause-guard).

## Related Issues

Fixes #187

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (new/changed tests green; 4 pre-existing `certificate_validator` failures are unrelated — a missing relative-path fixture)
- [ ] Public API changes are documented (no public API changes)


mssql-tds/src/connection/tds_client.rs
```diff
@@ -1744,39 +1752,118 @@ impl TdsClient {
...
-                    collected_errors.push(SqlErrorInfo::from(&error_token));
-                }
-                Tokens::Info(info_token) => {
-                    info!(?info_token, "Draining INFO token from stream");
-                    self.capture_info_message(&info_token);
-                }
-                Tokens::EnvChange(t1) => {
-                    if t1.sub_type == EnvChangeTokenSubType::ResetConnection {
-                        self.recovery_context.session_state_table.reset();
+                Tokens::ColMetadata(colmetadata) => {
```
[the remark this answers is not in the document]
saurabh500: Fixed. The top-level drain read in `drain_stream` now uses `ParserContext::ColumnEncryption(is_column_encryption_supported())`, matching `advance_to_result_boundary`, so a trailing result set's empty-but-present CEK-table prefix under Always Encrypted is parsed correctly instead of desyncing the stream. Added a regression test (`drain_parses_trailing_rowset_metadata_under_column_encryption`) that fails without the fix. The row reads on the other line already used `ParserContext::ColumnMetadata`, so both drain paths are now AE-safe.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1744,39 +1761,118 @@ impl TdsClient {
...
+    /// [`DefaultRowWriter`] never opts into [`RowWriter::pause_after_column`],
+    /// so `receive_row_into` fully consumes each row — including PLP payloads —
+    /// and never yields a pause result here.
+    async fn drain_result_set_rows(
+        &mut self,
+        metadata: Arc<ColMetadataToken>,
+        collected_errors: &mut Vec<SqlErrorInfo>,
+    ) -> TdsResult<bool> {
+        let col_count = metadata.columns.len();
+        let decryptor = self.resolve_cell_decryptor(&metadata).await?;
```
David-Engel: **Blocking** — don't resolve a cell decryptor on the discard path.

`resolve_cell_decryptor` can fail outright (`ColumnEncryptionError` when the CEK table is empty but a column carries crypto metadata, or when `client_context` is `None`), performs key-store/AKV round trips inside `ResolvedCekDecryptor::resolve`, and — because per-CEK failures are deferred to decrypt time — will surface a decryption error from `decrypt_encrypted_column` while decoding a row we are throwing away.

Any of those returns `Err` from `drain_stream`, and the caller in `advance_to_result_boundary` propagates it with `?`, so the original SQL error is masked and the batch is left half-drained — the exact failure mode this PR exists to remove, just relocated to the Always Encrypted case.

The decryptor isn't needed here. `RowTokenParser` already handles `(crypto_metadata.is_some(), None)` by decoding the raw ciphertext varbinary, and byte consumption is identical either way:

```rust
let col_count = metadata.columns.len();
// Values are discarded, so no decryptor is resolved: an encrypted column is
// decoded as raw ciphertext varbinary (identical byte consumption), which keeps
// key-store round trips and decryption failures off the drain path.
let parser_context = ParserContext::ColumnMetadata(metadata, None);
```

This also drops the side effect of overwriting `self.current_decryptor` with a drained result set's metadata.

Worth adding a test: trailing result set with a `FLAG_ENCRYPTED` column and a non-empty CEK table, no key-store provider registered — the SQL error must still surface.

David-Engel (commented):
## Summary

The fix is correct and the root-cause analysis in the description matches the code: the old `drain_stream` discarded `COLMETADATA` and then choked on the `ROW` it could no longer decode, so the terminal `DONE` was unreachable and the real SQL error was replaced by a parse failure. Splitting into `drain_stream` / `drain_result_set_rows` / `apply_drain_side_effect` preserves every prior side effect verbatim and adds the row-decoding hand-off. The second commit (AE-aware `ParserContext` at the top level) is a genuine catch — without it the CEK-table prefix would desync the stream.

Verified locally at `56807b5`: `cargo check --workspace --all-targets` is clean and all 11 new/related unit tests plus the mock-server integration test pass.

Test coverage is good: byte-level parity tests, NBCROW, multi-rowset, multi-column, secondary ERROR/INFO, protocol violation, AE metadata, plus an end-to-end test proving connection reuse.

One blocking item (inline, on `resolve_cell_decryptor` in the drain path) plus a couple of suggestions below that could not be attached inline because they land outside the diff hunks.

---

### Suggestion — reset result-set state after a successful drain (`advance_to_result_boundary`, the `Tokens::Error` branch, ~line 2048)

The sibling error path in `handle_row_read_token` sets `current_result_set_has_been_read_till_end = true` and `current_metadata = None`, with a comment explaining that otherwise a later call blocks reading an already-consumed stream. The `advance_to_result_boundary` ERROR branch only clears `has_open_batch`.

Pre-PR this rarely mattered because the drain usually failed anyway. Now that the drain completes, a caller that calls `next_row()` after `execute()` returned `Err` passes the `maybe_has_unread_rows()` guard and reads a drained stream. Two lines to align the two paths:

```rust
Tokens::Error(error_token) => {
    info!(?error_token);
    let mut all_errors = vec![SqlErrorInfo::from(&error_token)];
    let mut drain_errors = self.drain_stream().await?;
    all_errors.append(&mut drain_errors);
    self.execution_context.set_has_open_batch(false);
    self.current_result_set_has_been_read_till_end = true;
    self.current_metadata = None;
    return Err(crate::error::Error::from_sql_errors(all_errors));
}
```

### Suggestion — a drain failure still swallows the server error

`drain_stream` returns `TdsResult<Vec<SqlErrorInfo>>` and every caller uses `?`, so any residual failure mid-drain discards the diagnostics already collected. Consider having the ERROR branches surface `all_errors` even when the drain errors (logging the drain error and marking the connection unusable), so the SQL error always reaches the caller. Out of scope for a minimal fix, but it is the same class of bug this PR is fixing and would make the fix general rather than case-specific.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1744,39 +1761,118 @@ impl TdsClient {
...
+        let parser_context = ParserContext::ColumnMetadata(metadata, decryptor);
+        loop {
+            let start = Instant::now();
+            let mut writer = DefaultRowWriter::new(col_count);
```
David-Engel: **Suggestion** — `DefaultRowWriter` materializes each row that is immediately dropped. For a trailing result set with `varchar(max)` / `varbinary(max)` columns the drain allocates the full payload per row just to discard it. A no-op `RowWriter` (all `write_*` methods empty) consumes the same bytes with no allocation.

`drain_rows` has the same characteristic today, so this is consistency-neutral — but the drain-on-error path is where a large (or hostile) result set is least welcome.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1744,39 +1761,118 @@ impl TdsClient {
...
+                    ));
                 }
-                _ => {
-                    info!(?token);
+                RowReadResult::Token(token) => match token {
+                    Tokens::Done(done) | Tokens::DoneProc(done) | Tokens::DoneInProc(done) => {
+                        info!(?done, "Draining DONE token ending result set");
+                        return Ok(!done.has_more());
+                    }
+                    Tokens::ColMetadata(_) => {
```
David-Engel, line 1892: **Suggestion (tradeoff, not a defect)** — rejecting a spec violation is defensible, but note the cost: this converts the original SQL error into a `ProtocolError`, which is the outcome this PR is trying to avoid.

Since the top-level loop already knows how to start a new result set, returning a "new rowset started" signal and letting `drain_stream` re-enter would be strictly more resilient. Your call — flagging the tradeoff, not asserting the current behavior is wrong.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1744,39 +1761,118 @@ impl TdsClient {
...
-                }
-                Tokens::EnvChange(t1) => {
-                    if t1.sub_type == EnvChangeTokenSubType::ResetConnection {
-                        self.recovery_context.session_state_table.reset();
+                Tokens::ColMetadata(colmetadata) => {
+                    // A row-returning result set began. Consume its rows via the
+                    // row-decoding path; the trailing DONE tells us whether the
+                    // batch continues.
+                    if self
+                        .drain_result_set_rows(Arc::new(colmetadata), &mut collected_errors)
```
David-Engel, line 1828: Nit: the bare `bool` is opaque at the call site. `let batch_ended = self.drain_result_set_rows(...).await?; if batch_ended { break; }` reads better.

mssql-tds/src/test_client_support.rs
```diff
@@ -236,3 +236,188 @@ pub fn info(number: u32, severity: u8, message: &str) -> ScriptedToken {
...
+        fn set_reset_mode(&mut self, _mode: ResetConnectionMode) {}
+        fn take_reset_mode(&mut self) -> ResetConnectionMode {
+            ResetConnectionMode::None
+        }
+    }
+
+    #[async_trait]
+    impl NetworkReader for ByteStreamTransport {
+        async fn receive(&mut self, buffer: &mut [u8]) -> TdsResult<usize> {
+            buffer.fill(0);
```
David-Engel: Nit: `buffer.fill(0)` is dead work — the method returns `Ok(0)`, so the caller never looks at the buffer.

mssql-mock-tds/src/protocol.rs
```diff
@@ -818,10 +818,53 @@ pub fn build_info_token(info: &crate::query_response::InfoMessage) -> BytesMut {
...
+
+    buf.put_u8(TokenType::Error as u8);
+    let length_pos = buf.len();
+    buf.put_u16_le(0); // placeholder for token length
+
+    buf.put_u32_le(error.number);
+    buf.put_u8(1); // state
+    buf.put_u8(error.severity);
+
+    buf.put_u16_le(error.message.encode_utf16().count() as u16);
```
David-Engel: Nit: this walks `message.encode_utf16()` twice (once to count, once to write) and truncates the count with `as u16`. Test-only, so harmless, but `let utf16: Vec<u16> = error.message.encode_utf16().collect();` up front is cheaper and clearer.

mssql-tds/src/connection/tds_client.rs
saurabh500: Nice

mssql-tds/src/connection/tds_client.rs
saurabh500: Nice ! I will have to figure out if go-mssqldb driver also has a similar edge case I should watch out for

saurabh500: Done — `drain_result_set_rows` now passes `None` as the decryptor, so an encrypted column on the discard path is decoded as raw ciphertext varbinary (identical byte consumption). That keeps key-store round trips and decryption failures — any of which returns `Err` and would mask the SQL error the drain exists to surface — off the drain path, and no longer overwrites `self.current_decryptor` with a discarded set's metadata. Added `drain_does_not_resolve_decryptor_for_discarded_encrypted_rowset`, which fails without this change (the decryptor's `Err` masks error 1222).

mssql-tds/src/connection/tds_client.rs
saurabh500: Done — the drain now decodes rows through a no-op `DrainRowWriter`, avoiding `DefaultRowWriter`'s per-row `Vec<ColumnValues>` allocation on the discard path.

mssql-tds/src/connection/tds_client.rs
saurabh500: Kept the strict rejection. A COLMETADATA arriving before the current result set's DONE is a genuine protocol violation; legitimate multiple result sets arrive as DONE-with-MORE and are handled by the top-level drain loop (covered by `drain_on_error_consumes_multiple_trailing_rowsets`). Happy to revisit if a real server is ever observed emitting back-to-back COLMETADATA without an intervening DONE.

mssql-tds/src/connection/tds_client.rs
saurabh500: Done — extracted `let batch_ended = self.drain_result_set_rows(...).await?;` before the `break`.

mssql-tds/src/test_client_support.rs
saurabh500: Done — removed the dead `buffer.fill(0);`. The method returns `Ok(0)` and the caller never reads the buffer.

mssql-mock-tds/src/protocol.rs
saurabh500: Done — collect `encode_utf16()` into a `Vec<u16>` once, then write its length and iterate, instead of walking the iterator twice.

David-Engel (commented):
Re-reviewed at `b4933ea`. Verified locally: `cargo check --workspace --all-targets` is clean and all 12 drain tests pass. I also confirmed the new regression test is a real guard — reverting just the `resolve_cell_decryptor` line makes `drain_does_not_resolve_decryptor_for_discarded_encrypted_rowset` fail, so it isn't a tautology.

All six inline findings are resolved: the decryptor is off the discard path with a well-constructed test (encrypted column + empty CEK table + `current_command_ce_setting = Enabled`, so the pre-fix path genuinely fails fast), `DrainRowWriter` replaces the per-row `Vec<ColumnValues>`, and the three nits are fixed. The strict-`COLMETADATA` decision is fine — legitimate multi-rowset arrives as DONE-with-MORE and is covered by `drain_on_error_consumes_multiple_trailing_rowsets`.

Three things left. Sorry about the first two — they were in my review body rather than inline (they land outside the diff hunks), so they were easy to miss.

### 1. `advance_to_result_boundary`'s `Tokens::Error` branch still doesn't reset result-set state

Unchanged at `tds_client.rs:2052-2059`. The sibling path in `handle_row_read_token` does this, with a comment explaining exactly why:

```rust
// The error drained the rest of the batch to its terminal
// DONE, so the connection is idle again. Clear the batch
// state so a subsequent `close_query` / `advance` does not
// block trying to read a stream that is already consumed.
self.execution_context.set_has_open_batch(false);
self.current_result_set_has_been_read_till_end = true;
self.current_metadata = None;
```

Before this PR the asymmetry rarely bit, because the drain usually failed before completing. Now that it completes, a caller that calls `next_row()` after `execute()` returned `Err` passes the `maybe_has_unread_rows()` guard and reads an already-consumed stream. Two lines in the `advance_to_result_boundary` branch to match. Worth doing here since this PR is what makes the path reachable.

### 2. A drain failure still swallows the SQL error

`drain_stream` returns `TdsResult<Vec<SqlErrorInfo>>` and every caller applies `?`, so any residual failure mid-drain discards the diagnostics already collected — the same class of bug this PR fixes, just for the remaining failure modes. Reasonable to defer to a follow-up; I'd just like an explicit call either way rather than it going unnoticed.

### 3. New: the decryptor fix introduces per-cell `info!` spam on the drain path

This is fallout from my own suggestion, so flagging it rather than expecting you to have caught it. With `decryptor: None`, `RowTokenParser` now takes the `(crypto_metadata.is_some(), None)` branch (`token/parsers/row_parser.rs:155`) and emits an `info!` **per encrypted cell**:

> Encrypted column has no column-encryption decryptor available (Always Encrypted disabled for this command, or no key-store provider registered); returning the raw ciphertext varbinary

Draining a large trailing result set with an encrypted column now emits one info line per row per encrypted column, on an error path, with a message that is misleading there — nothing is misconfigured, the drain is deliberately discarding.

Cheapest fix is demoting that message to `debug!`; a per-cell diagnostic doesn't really belong at `info` even on the normal AE-disabled path. Alternatively log it once per result set from `drain_result_set_rows` and drop it from the hot loop.

With 1 and 3 addressed (and a yes/no on 2) this is good to go from my side.

mssql-tds/src/connection/tds_client.rs
David-Engel: Verified at `b4933ea` — reverting just this line makes `drain_does_not_resolve_decryptor_for_discarded_encrypted_rowset` fail, so the test is a real guard rather than a tautology. Resolved from my side.

One piece of fallout worth a follow-up (details in my review comment): with `decryptor: None` the row parser now takes the `(crypto_metadata.is_some(), None)` branch and emits an `info!` per encrypted cell, so draining a large trailing rowset with an encrypted column floods the log on an error path with a message that is misleading there. Demoting that message to `debug!` is probably enough.

saurabh500:
Thanks — addressed all three in `ae28586`.

> 1. `advance_to_result_boundary`'s `Tokens::Error` branch still doesn't reset result-set state

Fixed. The `Error` branch now clears `current_result_set_has_been_read_till_end` and `current_metadata` after the drain, matching the sibling in `handle_row_read_token` with the same reasoning in a comment. Added a regression test (`drain_on_error_leaves_client_idle_for_next_row`) asserting the client is left idle — `maybe_has_unread_rows()`, `on_rows()`, and `has_open_batch()` are all false after a drained error, so a `next_row` after a failed `execute` no longer reads an already-consumed stream.

> 3. the decryptor fix introduces per-cell `info!` spam on the drain path

Fixed with the cheap option: demoted that message from `info!` to `debug!` in `row_parser.rs`. As you noted, a per-cell diagnostic doesn't belong at `info` even on the normal AE-disabled path. (On the current merged tree the drain also runs under `ColumnPolicy::SkipAll`, so it no longer decodes cells at all — but the demotion is still the right call for the normal read path.)

> 2. A drain failure still swallows the SQL error

**Deferring to a follow-up** (your "yes/no": defer). Preserving already-collected `SqlErrorInfo` across a mid-drain transport failure means reworking `drain_stream`'s `?`-based error composition and every caller's error-mapping — broader than this fix and a different failure class (the wire dying mid-drain, which still surfaces as an error, just without the earlier SQL diagnostics attached). I'll track it separately rather than fold it in here.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1803,39 +1820,123 @@ impl TdsClient {
...
+                    self.cancel_handle.as_ref(),
+                    ColumnPolicy::SkipAll,
+                    &mut writer,
+                )
+                .await?;
+            self.update_remaining_timeout(start);
+
+            match result {
+                RowReadResult::RowWritten => {
+                    info!("Discarding row while draining result set");
```
David-Engel: **Suggestion** — this logs one `info!` line per discarded row, so draining a large trailing result set (the exact scenario this PR adds support for) emits one info record per row.

It also sits oddly next to the same commit's change in `row_parser.rs`, which demotes a per-cell message to `debug!` specifically to avoid "flooding info ... when draining a large trailing rowset on an error path".

Suggest `trace!` per row, or better, count and log once when the result set ends:

```rust
let mut discarded_rows: u64 = 0;
// ...
RowReadResult::RowWritten => discarded_rows += 1,
// ...
Tokens::Done(done) | Tokens::DoneProc(done) | Tokens::DoneInProc(done) => {
    info!(?done, discarded_rows, "Draining DONE token ending result set");
    return Ok(!done.has_more());
}
```

(The pre-existing `drain_rows` has the same per-row `info!`, so this matches today's style — but it is new volume on a path that can now see many rows.)

mssql-tds/src/token/parsers/row_parser.rs
```diff
@@ -155,9 +155,11 @@ impl<D: SqlTypeDecode + Default + Send + Sync, P: TdsPacketReader + Send + Sync>
                 (true, None) => {
                     // Either AE is disabled for this command (expected) or it is
                     // enabled but misconfigured (e.g. no key-store provider
-                    // registered). Log so the misconfigured case is observable,
+                    // registered). Log at debug so the misconfigured case stays
+                    // observable without flooding info on a per-cell basis (e.g.
+                    // when draining a large trailing rowset on an error path),
                     // then decode the raw ciphertext varbinary.
-                    info!(
+                    debug!(
```
David-Engel, line 162: **Suggestion** — the change itself is fine, but the stated rationale doesn't hold and it leaves the two copies of this message inconsistent.

1. The drain path does **not** reach this code. `drain_result_set_rows` calls `receive_row_into` -> `receive_row_into_internal`, which handles `TokenType::Row`/`NbcRow` inline via `drive_row_columns`; `RowTokenParser` (this file) is only used for ROW tokens read through `receive_token`. And with `ColumnPolicy::SkipAll`, `drive_row_columns` takes the `skip_column` branch, which never logs at all. So "when draining a large trailing rowset on an error path" is not what this line guards.
2. The duplicate of this exact message lives in `mssql-tds/src/io/token_stream.rs` (`decode_or_decrypt_column`, ~line 562) and is still `tracing::info!`. That copy *is* on the live row-decoding path, so after this change the same condition logs at two different levels depending on which reader path is used.

Either demote both and drop the drain reference from the comment, or leave this one as-is — but the two should agree. Purely observability, not merge-blocking.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -2013,7 +2114,14 @@ impl TdsClient {
                     let mut all_errors = vec![SqlErrorInfo::from(&error_token)];
                     let mut drain_errors = self.drain_stream().await?;
```
David-Engel: **Suggestion** — the `?` here means a drain failure discards everything below it: the already-collected `all_errors` (including the original SQL error this PR exists to surface) are dropped, and `set_has_open_batch(false)` / `current_result_set_has_been_read_till_end` / `current_metadata = None` never run, leaving the client marked busy on a stream we just gave up on.

The shape is pre-existing, but this PR widens the failure surface: the drain now decodes rows (timeout/cancel can land mid-rowset, since `remaining_request_timeout` is shared with the statement) and newly raises `ProtocolError` for COLMETADATA-before-DONE. In those cases the caller sees a transport/protocol error instead of e.g. error 1222 — which is the pre-fix symptom.

A small containment that keeps the SQL error primary and always resets state:

```rust
let mut all_errors = vec![SqlErrorInfo::from(&error_token)];
let drain_result = self.drain_stream().await;
self.execution_context.set_has_open_batch(false);
self.current_result_set_has_been_read_till_end = true;
self.current_metadata = None;
match drain_result {
    Ok(mut drain_errors) => all_errors.append(&mut drain_errors),
    Err(e) => warn!(error = ?e, "Drain after statement error failed; connection may not be reusable"),
}
return Err(crate::error::Error::from_sql_errors(all_errors));
```

If you'd rather keep the transport error visible (it does mean the connection is probably unusable), that's defensible — but the state reset should still be hoisted above the `?`. Same applies to the mirror in `handle_row_read_token`. Fine as a follow-up rather than a blocker on this PR.

David-Engel (commented):
Re-reviewed at `ae28586` (after the decryptor / no-op-writer fixes and the new state-reset commit). Verified locally: the 11 new `drain_*` unit tests and `test_drain_on_error` pass; the only `-p mssql-tds --lib` failures are the 4 pre-existing `certificate_validator` fixture-path tests, unrelated to this diff.

The core fix reads correctly:
- `ColumnPolicy::SkipAll` never sets `stop_here` in `drive_row_columns`, so the `RowPaused`/`PlpPaused` arm in `drain_result_set_rows` is genuinely unreachable — the defensive `ProtocolError` is dead-but-cheap, not a latent bug.
- `SkipAll` routes through `skip_column`, which never calls `decode_or_decrypt_column`, so passing `None` as the decryptor really does avoid key-store round trips, and `self.current_decryptor` (memoized on its `built_for` metadata) is correctly left untouched.
- `is_plp()` keys off `type_info`, which for an encrypted column is the outer ciphertext TYPE_INFO, so encrypted PLP columns are still byte-consumed correctly on the discard path.
- `ParserContext::ColumnMetadata` only matters to the ROW/NBCROW parsers; the RETURNVALUE parser ignores the context (`_context`) and parses its own `CryptoMetaData`, so control tokens read inside `drain_result_set_rows` are unaffected by the narrower context.

No blocking issues. Three non-blocking notes inline.

mssql-tds/src/connection/tds_client.rs
saurabh500: Good catch — swapped the per-row `info!` for a `discarded_rows: u64` counter that increments on `RowWritten` and is logged once at the DONE arm (`info!(?done, discarded_rows, ...)`), so a large trailing rowset no longer floods the log. Done in 73dc834e.

mssql-tds/src/token/parsers/row_parser.rs
saurabh500: You're right on both counts. The drain path goes through `receive_row_into_internal` / `decode_or_decrypt_column` (with `SkipAll` it never logs), not `RowTokenParser`, so I dropped the incorrect drain rationale from the comment. I also demoted the duplicate message in `io::token_stream::decode_or_decrypt_column` to `debug!` so both copies agree on the live decode path. Done in 73dc834e.

mssql-tds/src/connection/tds_client.rs
saurabh500: Implemented your containment: in both `advance_to_result_boundary` and `handle_row_read_token` the batch-state reset (`set_has_open_batch(false)` / `current_result_set_has_been_read_till_end` / `current_metadata = None`) now happens before the drain, and the drain result is matched instead of `?`-propagated — `Ok` appends the collected errors, `Err` logs a `warn!` and keeps the original SQL error primary. Added `drain_on_error_truncated_stream_still_surfaces_sql_error` covering the truncated-mid-drain case. Done in 73dc834e.

David-Engel (approved):

David-Engel (approved):

PR #193, opened 2026-08-08
Add nextest slow-timeout to fail hung tests loudly

### LINKED ISSUE #218: Add a timeout for nextest config
### Describe the bug

Add timeout for nextest config to make sure that the tests timeout for hung tests instead of hanging the CI. 

### Steps to reproduce

Add a never ending test and the pipeline stalls till pipeline timeout 

### Expected behavior

Pipeline shouldn't stall

### Actual behavior

Pipeline stalls

### Version

0.1.0

### Affected crate

mssql-tds

### Environment

_No response_

### Additional context

_No response_

saurabh500:
## Summary

Fixes https://github.com/microsoft/mssql-rs/issues/218

Adds a nextest `slow-timeout` so a hung or spinning test is terminated and fails loudly instead of hanging `cargo btest` forever. We just hit a 30+ minute forever-hang because the runner has retries but no timeout.

## Config change (exact TOML diff)

```diff
diff --git a/.config/nextest.toml b/.config/nextest.toml
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,6 +3,7 @@ path = "junit.xml"
 
 [profile.ci]
 retries = 5
+slow-timeout = { period = "60s", terminate-after = 3 }
+
+[profile.default]
+slow-timeout = { period = "60s", terminate-after = 3 }
 
 # Test groups for tests that need to run sequentially
```

A test running longer than `period` (60s) is flagged **SLOW**; after `terminate-after` (3) such periods (~180s) nextest terminates it and marks it failed. Added to both `[profile.ci]` (used by `cargo btest --profile ci`) and `[profile.default]` (plain `cargo nextest run`) so local runs are protected too. Retries and all other keys are unchanged.

## Condition (a): no healthy test legitimately runs past 60s ✅

Pulled the slowest **passing** tests from the CI JUnit timings (`target/nextest/ci/junit.xml`, 2237 passing tests):

| Time | Test |
|---|---|
| **20.33s** | `mssql-tds connection::transport::localdb::tests::test_resolve_mssqllocaldb` |
| 15.68s | `test_no_protocol_resolution::...test_msf_with_instance_and_explicit_port_allowed` |
| 14.72s | `test_multi_subnet_failover::...test_msf_connection_refused` |
| 10.70s | `test_multi_subnet_failover::...test_msf_invalid_host` |
| 8.51s | `connection_provider::...retry_with_unreachable_host` |

Slowest healthy test is **20.3s** — comfortably under 40s, giving a ~3x margin to the 60s threshold. **60s is safe; keeping it** (no bump to 90s needed).

## Condition (b): real worst-case bound

`[profile.ci]` has `retries = 5` → 6 attempts. Retries apply to **any** non-success outcome, and a deterministic hang is terminated after `3 × 60s = 180s` (~3 min) per attempt.

**Worst case = 6 attempts × 3 min = ~18 min** for a deterministic hang. This is bounded and a large improvement over the previous 30+ min forever-hang, but it is honestly higher than the earlier "≤3 min" estimate (that was per-attempt, before accounting for retries).

### Investigation: can we avoid retrying terminations?

Checked nextest **0.9.88** (the pinned toolchain version):

- There is **no** option to distinguish a slow-timeout **termination** from an assertion **failure** for retry purposes. Retry controls are limited to `count` / `backoff` / `delay` and per-**test-name** overrides — none filter by outcome type. A hang is retried the full 5×.
- `on-timeout = "pass"` (treat timeouts as success) exists only from **0.9.115**, and it changes pass/fail semantics, not retry behavior.
- On Windows, the `grace-period` is ignored for timeouts (immediate job-object kill), so each terminated attempt is ~180s exactly.

**Conclusion:** no no-retry-on-termination option is available in this version, so retries are left as-is and the **~18 min** worst-case applies. A reviewer may optionally choose to lower `retries` for the terminated case, but that would also reduce flake-retry coverage for genuine failures, so it's left untouched here.

## Validation

- `cargo bfmt` — clean
- `cargo bclippy` — clean (warnings-as-errors)
- `cargo nextest list --profile ci` and `--profile default` — both exit 0 (config parses, runner accepts it). Pre-existing integration-test failures require a live SQL Server and are unrelated to this TOML-only change.

## Note

TOML-only change to shared CI config. Values approved upstream: `period = "60s"`, `terminate-after = 3`, mirrored into `[profile.default]`.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -5,6 +5,9 @@ path = "junit.xml"
 retries = 5
 slow-timeout = { period = "60s", terminate-after = 3 }
 
+[profile.default]
+slow-timeout = { period = "60s", terminate-after = 3 }
+
 # Test groups for tests that need to run sequentially
 # TLS tests have race conditions with macOS Security.framework when creating
 # PKCS12 identities in parallel, so they must run one at a time.
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,6 +3,7 @@ path = "junit.xml"
 
 [profile.ci]
 retries = 5
+slow-timeout = { period = "60s", terminate-after = 3 }
 
 # Test groups for tests that need to run sequentially
 # TLS tests have race conditions with macOS Security.framework when creating


.config/nextest.toml
```diff
@@ -3,6 +3,10 @@ path = "junit.xml"
 
 [profile.ci]
 retries = 5
```
David-Engel, line 5: **Suggestion — the worst case is still ~19 min per hung test, and it compounds.**

`retries = 5` gives 6 attempts, each terminated at 180s plus up to the 10s Unix grace period, so ~19 min for one deterministic hang. Every CI leg also runs `--no-fail-fast` (`build-template.yml:138`, all `scripts/dockerentry/*.sh`), so *N* hung tests cost *N* x 19 min — which can exceed the 30-min hang this PR is fixing. Worth adding a real ceiling. In rough order of preference:

1. **Bump the pinned nextest `0.9.99` -> `>=0.9.100` and add `global-timeout` to `[profile.ci]`.** `global-timeout` landed in exactly 0.9.100, so you are one patch release away from the feature that directly solves "the whole run hangs forever" rather than bounding it per-test. Pin sites listed in the review body.
2. **Independent of nextest version: add `timeoutInMinutes` to the test jobs.** Today only `kerberos-test-template.yml:56`, `sql-host-template.yml:70` and `private-link-smoke-template.yml:76` set it — the main validation and test-matrix jobs have none, which is why the original hang ran unbounded.
3. **Cheapest, entirely in this PR: drop `terminate-after` to `2`.** 120s is still ~6x the slowest healthy test (20.3s) and halves the worst case to ~12 min.

Option 3 is fine to land now; 1 or 2 is the durable fix and can be a follow-up.

.config/nextest.toml
```diff
@@ -3,6 +3,10 @@ path = "junit.xml"
 
 [profile.ci]
 retries = 5
+slow-timeout = { period = "60s", terminate-after = 3 }
```
David-Engel, line 6: **Nit** — this file documents its other non-obvious settings (the `[test-groups]` TLS race comment below). A matching one-liner would save the next reader a trip to the nextest docs to decode `terminate-after`:

```toml
# Hung tests: flagged SLOW at 60s, terminated and failed at 180s.
slow-timeout = { period = "60s", terminate-after = 3 }
```

.config/nextest.toml
```diff
@@ -3,6 +3,10 @@ path = "junit.xml"
 
 [profile.ci]
 retries = 5
+slow-timeout = { period = "60s", terminate-after = 3 }
+
+[profile.default]
+slow-timeout = { period = "60s", terminate-after = 3 }
```
David-Engel, line 9: **Suggestion — this line is redundant, or the one in `[profile.ci]` is.**

nextest resolves configuration by walking the inheritance chain, which always terminates at the `default` profile, so `[profile.ci]` already inherits `slow-timeout` from here. Setting it in both places is a no-op today and creates drift risk the first time someone tunes the value and updates only one.

The `[[profile.*.overrides]]` blocks lower in this file *do* have to be duplicated (they're arrays of tables and don't merge), so the duplication there isn't a precedent for this. Either drop the `[profile.ci]` copy, or keep both and add a short `# keep in sync with [profile.ci]` note so the coupling is visible.

David-Engel (commented):
## Review: no blocking issues

Config-only change that is correct and strictly better than the status quo (unbounded hang). The write-up is unusually thorough — thanks for the timing evidence. A few suggestions below, plus two notes that don't fit on a specific line.

**Description accuracy:** the investigation section says "nextest **0.9.88** (the pinned toolchain version)". The repo actually pins **0.9.99** everywhere — `.pipeline/templates/build-template.yml:111` and `:131`, all five `scripts/dockerentry/*.sh`, `containers/Dockerfile.Ubuntu.Build:40`, `containers/Dockerfile.Alpine.Build:24`, `.devcontainer/Dockerfile:12`. The `on-timeout` conclusion is unaffected (still needs 0.9.115), but the 0.9.88 baseline caused `global-timeout` to be missed — see the inline comment on `retries = 5`. Also note `dev/Dockerfile:57` installs `cargo-nextest` **unpinned**, so dev containers can drift to newer semantics than CI.

**Coverage side-effect worth knowing:** CI runs under `cargo llvm-cov nextest` (the `btest` alias in `.cargo/config.toml`). A SIGTERM/SIGKILL'd test never flushes its `.profraw`, so a terminated test contributes zero coverage and `llvm-cov` may warn about missing profiles. This only bites on runs that actually time out, so it's acceptable — just don't chase a coverage dip on such a run.

### What I verified

- No test in the repo legitimately runs anywhere near 180s. The longest explicit waits are `WAITFOR DELAY '00:00:05'` in `mssql-tds/tests/timeout_and_cancel.rs:33` and `:103`. The `context.connect_timeout = 3600` in `mssql-tds/tests/connectivity.rs:137` is an Entra-ID test that only burns that budget when something is genuinely wrong — exactly the case this PR is meant to catch, so it's a win rather than a risk.
- Every CI entrypoint uses `--profile ci` (the five `scripts/dockerentry/*.sh`, `kerberos-test/run-kerberos-tests.sh`, `build-template.yml:138`, `dev/devtests.sh:13`), and archive runs pass `--workspace-remap /workspace`, so the checked-in config is what gets applied in all legs.
- Windows legs exist (`cov-rust-windows` in `validation-stages.yml`), where `grace-period` is ignored and termination is an immediate job-object kill at ~180s exactly, as your description states.

David-Engel (approved):
I feel like the bot review was pedantic but wanted to surface the suggestions anyway. Take them, leave them, do them later. Your choice. Approved.

saurabh500:
Thanks @david-engel. 
I am going to leave this Pr as is and not do anything about the suggestions. In fact my bot over explained a simple intention as well. 
While it's good to know about these mechanisms I will exercise them when absolutely needed

PR #194, opened 2026-08-08
Sans-I/O TDS core (3/N): invert token/decoder to a sync step() core

saurabh500:
> ⚠️ **Perf-neutral layer — does NOT move the ~2× fetch gap.** This PR is pure enabling/plumbing. The async/`block_on` tax this effort targets lives in the **parser core**, which lands in **L4** (parser-level column-atomic `step()`). "L3 green" means the reader seam is inverted and safe — **not** that any benchmark improved. The P8 fetch A/B is intentionally **not** re-measured until L4 lands, to avoid reproducing ~2× mid-stack and polluting the record.

## Layer 3 of the green-at-every-step sans-I/O stack

Base: `dev/saurabh/sans-io-layer2-transport` (#191). Public API is **frozen** — `mssql-odbc` and `mssql-py-core` build unchanged.

### What this lands
This slice inverts the **reader seam** to a synchronous, I/O-free core while the async shell remains the only public driver. It is the foundation the parser-level `step()` core builds on next.

- **Sync core:** `PacketBuffer::ensure(n) -> Result<(), NeedBytes>` — an atomic, non-consuming guard reporting the exact byte shortfall. Paired with the existing atomic `take_*` accessors, a fixed-width read never advances on a short buffer and is always safe to re-drive after a refill.
- **Inverted readers:** both `TdsPacketReader` impls (test `PacketReader` + production `NetworkTransport`) are now thin async driver loops: `loop { match buf.ensure(n) { Ok => take, NeedBytes => refill().await } }`. Trait signatures are unchanged, so `TdsClient` and the frozen API are untouched. Both `#[cfg(fuzzing)]` / `#[cfg(not(fuzzing))]` trait defs are maintained.
- **Anti-spin invariant (the L2 hang class):** every refill must expose ≥1 new byte or the loop fails loudly (`debug_assert`) / returns a `ProtocolError` — never spins on a zero-payload packet.
- **PLP dedup:** deleted the `read_plp_bytes` collect-all duplicate (~90 lines of copied chunk framing) and routed eager materialization through the shared `PlpChunkStreamReader` via a thin `collect_plp` helper. All size/chunk guards now live in one place.
- **Tests:** empty-input and exact-packet-boundary termination (proving no hang), plus `ensure` shortfall/atomicity invariants, plus direct `PacketBuffer` accessor + over-consume bounds-guard coverage.

### Scope note (honest labeling)
The reader-seam inversion + sync `NeedBytes` core + PLP dedup are landed here as a self-contained, low-risk layer. The **parser-level column-atomic `step()` core** (re-driving `decode_row_columns` / NBCROW / PLP chunk parsing over the sync buffer) is **L4** — kept separate so every step stays green, and it is the layer that actually removes the async tax. `DecodePolicy` / `BatchExit` enums were intentionally **not** introduced: no existing `bool` / `unreachable!` in the touched code maps cleanly to them, so adding them would be noise per the repo's no-AI-slop charter.

### Validation (green)
- `cargo build` ✅
- `cargo bclippy` (warnings = errors) ✅
- `cargo bfmt` ✅
- `cargo btest` (nextest): lib green except the 7 pre-existing `test_certificates` / `win_tls` cert-fixture failures (integration tests needing a live SQL Server are environmental); failing set verified byte-for-byte identical to the base branch. ✅
- `cargo build --manifest-path mssql-py-core/Cargo.toml` (outside workspace) ✅

LOC delta reported on the final rebased commit.

### DIFF
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -777,6 +777,33 @@ impl NetworkTransport {
         Ok(())
     }
 
+    /// Async shell over the sync core: drives [`PacketBuffer::ensure`] in a loop,
+    /// refilling on `NeedBytes` and retrying until `byte_count` bytes are
+    /// readable. The paired atomic `take_*` accessors never advance on a short
+    /// read, so re-driving after a refill is always safe.
+    async fn ensure_or_refill(&mut self, byte_count: usize) -> TdsResult<()> {
+        loop {
+            match self.tds_read_buffer.ensure(byte_count) {
+                Ok(()) => return Ok(()),
+                Err(need) => {
+                    tracing::trace!(shortfall = need.shortfall, "refilling TDS read buffer");
+                    let before = self.tds_read_buffer.available();
+                    self.read_tds_packet().await?;
+                    // Loop-termination invariant: a refill must expose at least
+                    // one new byte. Without this guard a zero-payload packet
+                    // would spin the driver forever, so fail loudly instead.
+                    let progressed = self.tds_read_buffer.available() > before;
+                    debug_assert!(progressed, "TDS refill made no forward progress");
+                    if !progressed {
+                        return Err(crate::error::Error::ProtocolError(format!(
+                            "TDS packet refill made no progress while {byte_count} bytes were needed"
+                        )));
+                    }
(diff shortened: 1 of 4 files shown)


saurabh500:

PR #195, opened 2026-08-08
Sans-I/O TDS core (4/N): invert the non-PLP column decode to a column-atomic step()

saurabh500:
> ⚠️ **Perf-neutral layer.** Part 4 of the green-at-every-step sans-I/O stack. Base: `dev/saurabh/sans-io-l3-token-decoder` (#194). Public API is unchanged; `mssql-odbc` and `mssql-py-core` build unchanged.

## Description

Inverts the **entire non-PLP column decode path** into a column-atomic, re-drivable synchronous `step()`, moving the `.await` refill point out of the middle of a column to the **column boundary**. This removes the last class of mid-value suspension in the row decode path (the risk L2's `NeedBytes` + L3's sync token decode were built to eliminate) without introducing any new resumable state machine — `RowPauseState.next_column_index` remains the only resume granularity and `NeedBytes { shortfall }` the only suspension signal.

### What this lands (6 commits, green at each)

- **`edc7aa83`** — invert non-PLP fixed-width column decode to a column-atomic `step()` over `column_wire_len` (peek) + whole-cell `ensure` + `decode_column_body` (infallible `take_*`).
- **`50dae22d`** — extend the step to variable-length non-PLP cells (`char`/`varchar`/`nchar`/`nvarchar` non-max, `binary`/`varbinary` non-max, `decimal`/`numeric`).
- **`952db191`** — extend the step to `SQL_VARIANT` cells.
- **`0d5e9619`** — converge the async row decoder onto the **single** shared `decode_column_body` so no non-PLP leaf logic is duplicated between `decoder.rs` and `sync_decoder.rs`.
- **`ca80c4ab`** — fold the Always-Encrypted non-PLP ciphertext whole-cell buffer into the sync step.
- **`1d46dd27`** — mandatory refill-boundary tests (below).

### Design

`decode_or_decrypt_column` owns buffering for **all** non-PLP cells, so it is inverted as one unit — there is no fixed-width-only seam inside it, and a partial inversion would fork it into a half-sync/half-async mid-column path. Per column the driver runs: `column_wire_len` (peek-only length) → `ensure(total)` (whole cell) → `decode_column_body` (pure sync, exactly one `write_*`). Because steps 1–2 only peek, a `NeedBytes` re-drive after refill re-runs from the column start with nothing consumed and nothing written → atomic.

`decode_column_body` (+ `column_wire_len` + the sync leaf ports) has **exactly one definition** in `sync_decoder.rs`; two refill drivers (sync `PacketReader`, async `GenericDecoder::decode_into`) wrap it and differ only by the `.await` on refill.

### Out of scope (unchanged)

- **PLP/LOB streaming** (`PlpColumnStream` / `PlpChunkStreamReader`, incl. AE over `varbinary(max)`) — L4b, behind the existing pause branch.
- **NBCROW body** — L4c; the sync step is written so it can be reused verbatim there.

### Tests

Five mandatory boundary tests drive the buffer-owning test `PacketReader`, so they genuinely exercise `decode_column_body`:

1. **non-PLP → PLP transition at every refill boundary** (`nonplp_to_plp_transition_is_byte_identical_across_refill_boundary`) — a `[int4][nvarchar(max) PLP]` row swept with the packet boundary at each interior offset; the inverted step fully consumes the int4, then the PLP column resumes byte-identically across every split.
2. **mid-cell re-drive** (`var_length_cell_redrive_is_byte_identical_across_refill_boundary`) of an inverted var-length cell, including inside the 2-byte length prefix (peek-only re-drive) and mid-data.
3. **hybrid mid-row seam** (`mixed_row_inverted_then_plp_is_byte_identical_across_refill_boundary`) — a `[int4][varchar][nvarchar(max) PLP]` row that mixes two inverted non-PLP cells with a not-yet-inverted (legacy async) PLP cell in ONE row, swept across every offset (int4→varchar seam, inside the varchar length prefix, and the varchar→PLP handoff). Proves the inverted step and the async PLP path share one coherent row cursor (`next_column_index`) regardless of where the packet boundary lands — the L4a→L4b hybrid-window correctness guarantee.
4. **exact whole-column non-PLP → PLP transition** (`refill_boundary_at_nonplp_to_plp_column_transition_resumes_into_async_plp`) — same `[int4][varchar][nvarchar(max) PLP]` row, but the packet boundary is forced to land **exactly** at the varchar→PLP column transition: the inverted step fully consumes the varchar cell, `ensure` on the PLP column's first bytes returns `NeedBytes`, refill happens, and the still-async PLP path resumes at `next_column_index`. Asserts byte-identical to the single-packet baseline — the precise whole-column seam the hybrid window depends on.
5. **inverted var-length cell own re-drive** (`refill_boundary_within_inverted_varchar_cell_resumes_coherently`) — the boundary is swept **inside** the inverted varchar cell (within its 2-byte length prefix and mid-data), proving the inverted step's own re-drive from the column start is coherent independent of the PLP handoff.

Making these meaningful required fixing the test `PacketReader` to be faithful to the production transport: `receive_packet` now validates the header length, returns the single packet's length, and carries surplus bytes forward via `record_pending` (previously it returned the raw coalesced byte count and stripped only the first header). `TestPacketBuilder` now writes the TDS **total** length (header + payload) into the header field, matching the production serializer and `validate_packet_length`.

## Related Issues

N/A — internal sans-I/O refactor, no tracking issue (consistent with #189/#191/#194).

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (lib green; failing-name set byte-for-byte identical to base = 359 pre-existing env/cert failures)

### DIFF
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -1088,6 +1088,33 @@ impl TdsPacketReader for NetworkTransport {
         self.tds_read_buffer.reset_to_length(0);
     }
 
+    async fn decode_column_into(
+        &mut self,
+        meta: &crate::query::metadata::ColumnMetadata,
+        col: usize,
+        writer: &mut (dyn crate::datatypes::row_writer::RowWriter + Send),
+    ) -> TdsResult<()> {
+        use crate::datatypes::sync_decoder;
+
+        // Sync driver over the owned read buffer: peek the wire width, ensure the
+        // whole cell, then decode in place. `column_wire_len` only peeks, so a
+        // shortfall re-drives from the column start with nothing consumed.
+        loop {
+            match sync_decoder::column_wire_len(&self.tds_read_buffer, meta) {
+                Ok(total) => {
+                    self.ensure_or_refill(total).await?;
+                    return sync_decoder::decode_column_body(
+                        &mut self.tds_read_buffer,
+                        meta,
+                        col,
+                        writer,
+                    );
+                }
+                Err(need) => self.ensure_or_refill(need.shortfall).await?,
+            }
+        }
+    }
+
     async fn read_byte(&mut self) -> TdsResult<u8> {
         self.ensure_or_refill(1).await?;
         self.tds_read_buffer.take_u8()
--- a/mssql-tds/src/datatypes.rs
+++ b/mssql-tds/src/datatypes.rs
@@ -29,4 +29,6 @@ pub mod sql_vector;
 pub mod sqldatatypes;
 /// Input parameter types for RPC calls.
 pub mod sqltypes;
+/// Column-atomic synchronous decode of non-PLP cells over [`PacketBuffer`].
+pub(crate) mod sync_decoder;
 pub(crate) mod tds_value_serializer;
--- a/mssql-tds/src/datatypes/decoder.rs
+++ b/mssql-tds/src/datatypes/decoder.rs
@@ -48,7 +48,21 @@ where
     D: SqlTypeDecode,
     T: TdsPacketReader + Send + Sync,
 {
-    let cipher = match decoder.decode(reader, metadata).await? {
+    let cipher = decoder.decode(reader, metadata).await?;
+    decrypt_cipher_value(metadata, decryptor, cipher)
+}
+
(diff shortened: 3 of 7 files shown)


saurabh500:

PR #196, opened 2026-08-09
Sans-I/O TDS core (4b/N): invert PLP/LOB column streaming to the sync core

saurabh500:
## Summary

Layer 4b of the sans-I/O restructure. Inverts the eager PLP/MAX column collect path to run synchronously over the in-memory `PacketBuffer`, reusing the existing `PlpChunkStreamReader` resumable state — **no new state machine**. The async `decode_into` PLP branch and the buffer-driven sync driver converge on **one shared framing body** (two refill drivers differing only by `.await`).

Base: `dev/saurabh/sans-io-l4a-fixed-width-step` (frozen tip `8bddc4fc`).

## Related Issues

N/A — internal sans-I/O refactor, no tracking issue (consistent with #189/#191/#194/#195).

## What changed

- **Factor byte-source-agnostic framing leaves out of `PlpChunkStreamReader`** (`decoder.rs`): `classify_length`, `on_chunk_header`, `note_read`, `chunk_remaining`, `reached_end`, `total_read`. The async `begin`/`ensure_active_chunk`/`read_into` are rewritten to call these leaves — single-copy framing, no drift.
- **Sync PLP driver** (`sync_decoder.rs`): `plp_collect_step` + `PlpProgress` step the `PlpChunkStreamReader` over an owned `PacketBuffer` **one chunk at a time** — `ensure(4)` for a chunk header or a slice of the current chunk body, **never the whole value** — so wire residency stays bounded to ≲ one packet regardless of total value size. `NeedMore` carries the absolute header/body width so refill always makes forward progress across packet-straddling prefixes.
- **Reader seam** (`packet_reader.rs`): `collect_plp_bytes` — async framing default for non-buffer readers, sync driver override on `PacketReader` and `NetworkTransport`. All `GenericDecoder::collect_plp` call sites rewired through the seam; **value arms unchanged** → byte-identical.
- **Always-Encrypted over `varbinary(max)`** (`token_stream.rs`): collect ciphertext via the sync seam then `decrypt_cipher_value`, folding collect+decrypt into the sync path (mirrors L4a's non-PLP AE fold).

## Scope boundary

`text`/`ntext`/`image` are `is_plp() == false` legacy text-pointer LOBs (not the PLP chunk protocol) and stay on the legacy async path, unchanged and byte-identical — consistent with L4a's boundary. (Coordinator-tracked as a separate p4d follow-up.)

## Tests (buffer-driven `PacketReader`)

- `plp_single_chunk_mid_chunk_resume_is_byte_identical_across_refill_boundary` — refill boundary swept across every interior offset incl. inside the 8-byte PLP header and 4-byte chunk-length prefix.
- `plp_multi_chunk_across_refills_is_byte_identical_with_bounded_residency` — multi-chunk to the zero-length terminator + a large value exceeding the 8192-byte buffer capacity decoded under packet-sized refills, proving chunk-bounded residency by construction.
- `fully_sync_mixed_row_with_plp_is_byte_identical_across_refill_boundary` — `[int4][varchar][varbinary(max)]` with every cell on the sync core, exact values asserted across every refill offset.

## Validation

- `cargo btest` (nextest) terminates; FAIL name-set byte-for-byte identical to base (the 7 known cert fixtures).
- `cargo bfmt` + `cargo bclippy` clean; `scripts/bfmt.ps1` + `scripts/bclippy.ps1` clean (mssql-py-core builds).

No PLP framing logic duplicated; no new resumable machine introduced.

Copilot-Session: e2c378f8-3ba1-4b48-9ebe-5a4ea2bd2761

### DIFF
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -1115,6 +1115,30 @@ impl TdsPacketReader for NetworkTransport {
         }
     }
 
+    async fn collect_plp_bytes(&mut self) -> TdsResult<Option<Vec<u8>>> {
+        use crate::datatypes::decoder::PlpChunkStreamReader;
+        use crate::datatypes::sync_decoder::{PlpProgress, plp_collect_step};
+
+        // Sync PLP driver over the owned read buffer: classify the 8-byte header,
+        // then pull one chunk header / body slice at a time. Residency stays
+        // bounded to ~one packet — the whole value is never ensured into the
+        // buffer. Refill is lifted to `ensure_or_refill`.
+        self.ensure_or_refill(8).await?;
+        let raw = self.tds_read_buffer.take_i64_le()?;
+        let mut plp = match PlpChunkStreamReader::classify_length(raw)? {
+            None => return Ok(None),
+            Some(len) => PlpChunkStreamReader::new(len),
+        };
+
+        let mut out = Vec::new();
+        loop {
+            match plp_collect_step(&mut plp, &mut self.tds_read_buffer, &mut out)? {
+                PlpProgress::Done => return Ok(Some(out)),
+                PlpProgress::NeedMore(n) => self.ensure_or_refill(n).await?,
+            }
+        }
+    }
+
     async fn read_byte(&mut self) -> TdsResult<u8> {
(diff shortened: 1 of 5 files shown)


saurabh500:

PR #198, opened 2026-08-09
P4a: sync TdsCore row-fetch driver over PacketBuffer

saurabh500:
## P4a — sync `TdsCore` row-fetch driver (Option A)

Part of the sans-I/O restructuring of `mssql-tds`. **Base:** `dev/saurabh/sans-io-l4c-invert-nbcrow` (stack #192). Draft — do not merge; coordinator will independently verify from git + nextest and append to the stack.

### What this does
Extracts the row-fetch framing, non-PLP column decode, and NBC null-bitmap into ONE pure-sync `TdsCore::step_row(&mut PacketBuffer, …) -> RowStep` that returns `NeedBytes` on underflow. A thin async driver `drive_row_over_buffer` owns the sole refill `.await` and re-drives `step_row`. The async client stays pure delegation.

### Column routing
- **Inlined sync (over `PacketBuffer`, `NeedBytes` on underflow):** non-PLP columns where `crypto_metadata.is_none() && sync_decoder::is_supported(meta)`, plus the NBC null bitmap.
- **Yielded to async leaf (`decode_async_row_column`):** everything else — crypto/AE columns, `!is_supported` non-PLP (p4d text/ntext/image legacy LOBs), and all PLP (`is_supported` returns false for `is_plp()`). Eager PLP is serviced by the existing `collect_plp_bytes().await` seam (Fork A — no new machine); PLP streaming pause reaches the existing `PlpColumnStream`/`PlpPauseState` (p7, unchanged).

### Atomic row-header mechanism (binding condition)
The row header (token byte + optional NBC bitmap) is **ensured-then-taken atomically**: `peek_bytes(1)`/`ensure(1 + bitmap_len)` are non-consuming, so a `NeedBytes` at any header boundary leaves the buffer position unchanged and the re-drive re-peeks the same token byte. The "token consumed, bitmap pending" state never exists → zero new resumable state; the existing `RowPauseState` cursor is held by the driver across refills.

### Invariants
1. **Zero behavior change** — `-p mssql-tds --lib` fail-set == exactly the 7 pre-existing cert fixtures; `Compare-Object` of sorted fail-sets vs base EMPTY; no hang.
2. **One production parse body** — `TdsCore::step_row`. `receive_row_into_internal` retained ONLY as a `#[cfg(any(test, fuzzing))]` reference oracle (Fork B R2); no production caller.
3. **Zero new resumable machines** — reuses `RowPauseState`/`PlpPauseState`/`PlpColumnStream`/`PlpChunkStreamReader`/`NeedBytes`.
4. **Public API frozen** — `TdsCore`/`RowStep`/`step_row`/`drive_row_over_buffer` are `pub(crate)`; diff-scan shows no new `pub fn`/`pub async fn`.
5. **Gates** — `bfmt`, `bclippy` (-D warnings), `scripts/bfmt.ps1` + `scripts/bclippy.ps1` incl. `mssql-py-core`, `mssql-py-core` builds `--all-features`. MIT header on new `io/tds_core.rs`.

### Tests (through the new `TdsCore` driver, refill boundaries swept)
- `tdscore_step_row_refill_boundary_swept_all_offsets_is_byte_identical`
- `tdscore_step_row_nbcrow_bitmap_split_across_refill_is_byte_identical`
- `tdscore_step_row_eager_plp_yields_and_resumes_at_next_column`
- `tdscore_resume_from_pause_is_byte_identical_and_matches_oracle`

### DIFF
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -19,8 +19,7 @@ use crate::io::packet_writer::PacketWriter;
 use crate::io::reader_writer::{NetworkReader, NetworkReaderWriter, NetworkWriter};
 use crate::io::token_stream::{
     ParserContext, PlpPauseState, RowPauseState, RowReadResult, TdsTokenStreamReader,
-    read_active_plp_bytes_internal, receive_row_into_internal, receive_token_internal,
-    resume_row_into_internal,
+    drive_row_over_buffer, read_active_plp_bytes_internal, receive_token_internal,
 };
 use crate::message::attention::AttentionRequest;
 use crate::message::login_options::TdsVersion;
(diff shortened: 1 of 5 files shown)


saurabh500:

PR #201, opened 2026-08-09
Sans-I/O TDS core (L2): lift request timeout and cancellation to the receive edge

saurabh500:
## Layer 2 of the sans-I/O core stack (internal refactor; no public API change)

Base: `dev/saurabh/sans-io-expose-l1-bytesource` (Layer 1, PR #200).

### What
The async shell applied the per-request timeout + cancellation by wrapping the whole token/row drive future in `tokio::time::timeout` + a `CancellationToken` (the 4 `TdsTokenStreamReader` methods). That tokio-timer wrap was the last runtime coupling on the otherwise sans-I/O parse path.

This lifts the concern to the `ByteSource` edge:
- New `pub(crate) ReceiveGuard { deadline: Option<Instant>, cancel: Option<CancellationToken> }` snapshots the request's **absolute** deadline + cancel token.
- The async `TdsTokenStreamReader` shell sets the guard before each drive and resets it after; the timeout/cancel cleanup (`cancel_read_stream_and_wait`) is unchanged.
- `AsyncByteSource::receive` applies `timeout_at` + `run_until_cancelled` around the socket read, reusing the existing error paths so timeout → `TimeoutError(Elapsed)` and cancel → `OperationCancelledError` stay byte-identical.
- The shared bodies (`assemble_tds_packet`, `drive_*`, `step_*`) now only call `receive()` — timer- and cancel-agnostic, ready for a future non-tokio driver.

### Why byte-identical
The only point any drive future yields `Pending` is the socket read (all refills funnel through `get_new_tds_packet`). So today's whole-future `timeout` could only ever fire there; moving an absolute deadline + cancel to wrap exactly that read preserves the same total-request deadline, the same cancel, and the same error kinds/messages.

### Gates
- Public-API diff vs L1 tip `bba556bf` EMPTY.
- `cargo nextest -p mssql-tds --lib`: 1703 passed, 7 pre-existing baseline failures (cert/pin tests), no new failures/hangs.
- `cargo bfmt`, `cargo bclippy`, `scripts/bfmt.ps1`, `scripts/bclippy.ps1` all clean.

Untouched: shared parse bodies, the `#[cfg(fuzzing)]` oracle, and the attention-ACK / connect tokio timeouts (shell-local).


saurabh500:

PR #203, opened 2026-08-10
Add owning-reversible TdsSyncClient over the blocking edge

saurabh500:
## L4 — Owning-reversible `TdsSyncClient` over the blocking edge

Stacked on **L3** (`dev/saurabh/sans-io-expose-l3-blocking-driver` @ `7e11a11b`) — **base is L3, not `main`.**

Surfaces the L3 blocking row driver as a public, reactor-free `TdsSyncClient` so consumers (ODBC `SQLFetch`, later py-core) can drop their per-row `block_on` on the hot path. The async `TdsClient` surface is unchanged.

### Additive-4 public surface (frozen)
- `TdsClient::into_sync(self) -> SyncConversion` — opportunistic raw-TCP flip.
  - `SyncConversion { Converted(TdsSyncClient) | NotEligible(TdsClient) | Failed(TdsError) }`. TLS transports return the async client **intact** via `NotEligible` (not an error — caller keeps today's `block_on`).
- `TdsSyncClient` — owns the established transport + buffer over the blocking edge (concrete `StdTcpByteSource`; no generic `B` in the type name). 8 B-free fetch methods (`next_row`, `next_row_into`, `fetch_rows_batch`, `take_info_messages`, `get_metadata`, `maybe_has_unread_rows`, `active_plp_reached_end`, `active_plp_collation`).
- `TdsSyncClient::into_async(self) -> TdsResult<TdsClient>` — reverts the fd to the async client; `Err` (fd known-dead) on failed revert.

**Owning, not borrowed** (regression tripwire): a `SyncRowFetcher<'a>` borrowing `dbc.client` can't persist across `SQLFetch` FFI calls, so it would flip the fd **per row** on a 1M-row set. Owning stores `TdsSyncClient` by value → fd stays blocking across the whole result set (~2 flips/result-set, measured 13.569µs/pair). `Drop` does terminal clean-close only; the explicit async revert is `into_async` — never a Drop-driven fd-revert.

### Invariants honored
- **Empty public-API diff**: the only *public* diff to `tds_client.rs` is the additive `into_sync`; `cursor_ops.rs` untouched. The shared orchestration added alongside it is `pub(crate)` (see doc-note 3), so no public signature changes and no generic `B` leaks into any async signature.
- **Verbatim reuse**: `step_row` / `drive_row_over_buffer_blocking` / `plp_collect_step` / the L3 blocking driver are called as-is, not reimplemented.
- `fetch_rows_batch` authored fresh as a thin loop over reactor-free `next_row` with spare-vec recycling (no async batch twin, no #186 dependency).
- `execute*` / `advance*` / `close*` stay **async** (reached via `into_async`). No cold-token blocking driver. AE PLP columns → `UnimplementedFeature`. `read_active_plp_bytes` deferred to L5.

### Tests
Differential parity vs an all-async oracle (byte-identical), plus the named **residual-straddle interleave gate in both directions** (async→sync and sync→async where buffered `tds_read_buffer` bytes straddle the flip mid-packet) and a **clean-boundary negative control**. The 7-cert lib baseline (4 `certificate_validator` + 3 `win_tls::validate`, both fixture/cert-store-gated) is unchanged.

---
### Doc-notes (context, not a re-plan)
1. **Rule-C WHY = scope + runtime + surface** — buffer-size ground was **retracted** (block_on-over-async-parser inherits the async parser's incremental refill, so buffer-size does not apply). Cold ops (execute/COLMETADATA/advance/close) run once per statement ⇒ zero measurable perf from a sync cold-drive; a block_on cold-token adapter would need a live reactor (multi_thread) and deadlock ODBC's current_thread runtime; fetch-only reuses the async control-plane verbatim. Hence: **no cold-token blocking driver**.
2. **L5 de-risk — the shipped fetch path is reactor-free** ⇒ runtime-flavor-agnostic (`current_thread` is fine). The multi_thread-deadlock concern applies **only** to the forbidden cold adapter; the earlier prototype multi_thread note was an echo-server co-host artifact (the real SQL peer is remote, so no local task is starved by a blocked thread).
3. **Refined-B orchestration decision (drift-proofing)** — the sync and async fetch shells share **one authoritative token handler** rather than parallel copies. The per-token side-effects of the async `handle_row_read_token` are factored into a single `pub(crate) TdsClient::apply_row_read_token(&mut self, token) -> TdsResult<TokenOutcome>` (plus `pub(crate) finalize_row_error` for post-drain cleanup), living **on `TdsClient` in `tds_client.rs`** so it has native access to private state (`count_map`, `finalize_return_value`/`push_return_value`, info buffer) with **zero accessors and zero field-visibility churn**. Both shells call it; only the ERROR drain is flavour-specific (`TokenOutcome::DrainThenError` → async `drain_stream` vs a blocking drain-to-DONE over L3, then the shared `finalize_row_error`). `handle_row_read_token`'s async signature/visibility is byte-identical — only its body delegates. All connection/result-set state (metadata, INFO buffer, read-till-end flag) lives on the owned `inner`; the wrapper keeps no mirrored copies, so the two clients cannot diverge. This keeps `ClientCore<B>` unused (prefer-none) and the **public**-API diff empty while eliminating the parallel-handler drift risk.
4. **Refined-B guard compliance (characterization + error-arm equivalence)** —
   - *Characterize-then-refactor*: 9 characterization unit tests pin each side-effect the shared `apply_row_read_token`/`finalize_row_error` owns (terminal DONE accounting + batch close, DONE-MORE accumulation, DONE error-flag → protocol error, ORDER no-op, INFO capture, RETURN_VALUE push, ERROR defer-drain-without-mutation, COLMETADATA usage error, and `finalize_row_error` batch-clear + surfaced `SqlServerError`). The full pre-existing async suite passes with **zero expected-output edits**, so the shared-handler extraction is behavior-preserving.
   - *Sync ERROR-arm equivalence*: the sync blocking drain-to-DONE now routes every non-terminal side-effect (`Info`/`EnvChange`/`SessionState`/`ReturnValue`/`ReturnStatus`) through the **same** `apply_row_read_token`, reaching a byte-identical terminal state to the async `drain_stream`. An Error-token-**mid-fetch** case was added to both the differential set and the residual-straddle-interleave set (both directions), so the sync drain is exercised, not just the happy path.
   - *Visibility reconciliation*: because `TdsSyncClient` lives in a **different module** (`tds_sync_client.rs`) and calls `inner.apply_row_read_token(...)` cross-module, `pub(crate)` is the **tightest reachable** visibility (a private `fn` is unreachable cross-module) — and only new **additive** `pub(crate)` methods were introduced. **Zero pre-existing item was widened**: `capture_info_message` stays private (the earlier sync path that would have widened it now goes through the shared handler instead).

### DIFF
--- a/mssql-mock-tds/src/protocol.rs
+++ b/mssql-mock-tds/src/protocol.rs
@@ -853,18 +853,76 @@ pub fn build_query_result(response: &crate::query_response::QueryResponse) -> By
         result.extend_from_slice(&build_info_token(info));
     }
 
-    // Serialize each row
-    for row in &response.rows {
-        result.put_u8(TokenType::Row as u8);
-        for value in &row.values {
-            value.write_to_buffer(&mut result);
+    match &response.error_after {
+        Some(err) => {
+            // Stream the first `after_rows` rows, then an ERROR token and a
+            // terminal DONE (row count 0) — no trailing rows.
+            for row in response.rows.iter().take(err.after_rows) {
+                result.put_u8(TokenType::Row as u8);
+                for value in &row.values {
+                    value.write_to_buffer(&mut result);
+                }
+            }
+            result.extend_from_slice(&build_error_token(
+                err.number,
+                err.state,
+                err.severity,
+                &err.message,
+            ));
+            // INFO tokens emitted during the drain (after ERROR, before DONE).
+            for info in &err.drain_info {
+                result.extend_from_slice(&build_info_token(info));
+            }
+            result.extend_from_slice(&build_done_token(0));
+        }
+        None => {
+            // Serialize each row
+            for row in &response.rows {
+                result.put_u8(TokenType::Row as u8);
+                for value in &row.values {
+                    value.write_to_buffer(&mut result);
+                }
+            }
+
+            // DONE token
+            result.extend_from_slice(&build_done_token(response.rows.len() as u64));
         }
     }
 
-    // DONE token
-    result.extend_from_slice(&build_done_token(response.rows.len() as u64));
-
     wrap_in_packet(PacketType::TabularResult, result)
+}
+
+/// Build a bare ERROR token (0xAA) with no surrounding DONE or packet framing,
+/// for injecting mid-stream into a result set.
+fn build_error_token(number: u32, state: u8, severity: u8, message: &str) -> BytesMut {
+    let mut token = BytesMut::new();
+    token.put_u8(TokenType::Error as u8);
+
+    let length_pos = token.len();
+    token.put_u16_le(0); // Placeholder for length (little-endian on the wire)
+
+    token.put_u32_le(number);
+    token.put_u8(state);
+    token.put_u8(severity);
+
+    // Message (US_VARCHAR: u16 code-unit count + UTF-16LE)
+    token.put_u16_le(message.chars().count() as u16);
+    for ch in message.encode_utf16() {
+        token.put_u16_le(ch);
+    }
+
+    // Server name / procedure name (empty B_VARCHARs)
+    token.put_u8(0);
+    token.put_u8(0);
+
+    // Line number
+    token.put_u32_le(1);
+
+    let token_length = (token.len() - length_pos - 2) as u16;
+    let mut length_bytes = &mut token[length_pos..length_pos + 2];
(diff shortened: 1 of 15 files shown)


saurabh500:

PR #204, opened 2026-08-10
Rewire mssql-odbc fetch hot path onto the reactor-free sync core (L5)

saurabh500:
## L5 — Rewire mssql-odbc fetch onto `TdsSyncClient`

Stacked on the frozen **L4** tip (`dev/saurabh/sans-io-expose-l4-tdssyncclient` @ `36bc4648`). Append-only; **base = L4 branch**, not `main`.

### What this does
Plaintext raw-TCP connections now serve `SQLFetch` through the reactor-free `TdsSyncClient` (blocking `std` reads, no tokio reactor). TLS and other non-raw transports fall back to the **untouched** async `block_on` path, byte-identically — the speedup is opportunistic on plaintext with zero user-visible flag.

### Ownership machine
- `DbcClient { Async(TdsClient) | Sync(TdsSyncClient) }` in the connection state.
- Flip to the sync edge **once per row-returning result set** at execute time (`finish_execute`), persisting across all `SQLFetch` calls.
- Revert to async (`into_async`) at `SQLMoreResults` / close / free **before** driving the existing `advance` / `close_query`.
- **Runtime-context hard rule**: every `into_sync()` runs inside `dbc.runtime.enter()` so the runtime handle is captured for the later revert (a naked `into_sync()` on the bare ODBC thread would capture `None` and poison `into_async`). Implemented in one place: `flip_to_fetch_edge`.

### Batch buffer (batch-ready, per-row default)
- Statement-level row buffer + spare-row recycling pool, gated behind `SYNC_FETCH_DEFAULT_MAX_ROWS` (**default 1** = per-row byte-identical INFO/SWI parity; flip to 64 is a one-line change).
- Deferred mid-batch error handling: at `max_rows > 1`, buffered rows are served before the error surfaces, reproducing the async row-then-error ordering.

### Other
- Cache `server_version` at connect so `SQL_DBMS_VER` no longer reads the live client while a sync cursor is open.
- **C-ABI unchanged**: no `#[unsafe(no_mangle)] extern "C"` signatures touched.

### Test-infra note (flagged for scope)
`mssql-mock-tds` (test-only crate, not frozen L4) is extended **additively** for multi-result-set responses — needed to drive the `SQLMoreResults` interleave test over a real TCP peer. Single-set responses serialize byte-identically; its own 16 tests stay green.

### Tests (in-crate; the driver is a `cdylib`, so an external `tests/` crate can't link it or reach `pub(crate)` entry points)
- (a) differential: sync `SQLFetch` @ max_rows=1 == async oracle, byte-identical
- (b) max_rows=64: identical row-set + surfaced error (deferred-error fix)
- (c) `SQLMoreResults` interleave: sync-fetch → `into_async` → `advance` → re-`into_sync`, both sets verified
- (d) NotEligible fallback: `flip_to_fetch_edge` keeps the DBC on the async edge

Gates: `cargo bfmt` clean, `cargo bclippy` (-D warnings) clean, `cargo nextest -p mssql-odbc` = **472 passed** (468 existing + 4 new), `mssql-mock-tds` 16 passed.

Part of the sans-I/O native stack - see cover PR #189 for the stack-level summary and tracked debt.

### DIFF
--- a/mssql-mock-tds/src/protocol.rs
+++ b/mssql-mock-tds/src/protocol.rs
@@ -674,13 +674,20 @@ pub fn build_login_ack() -> BytesMut {
 
 /// Build a DONE token
 pub fn build_done_token(row_count: u64) -> BytesMut {
+    build_done_token_with_status(0x0000, row_count)
+}
+
+/// DONE token with an explicit status word. `0x0000` is `DONE_FINAL` (the batch
+/// ends); `0x0001` is `DONE_MORE`, signalling that another result set follows in
+/// the same batch (what a multi-statement `SELECT …; SELECT …` produces).
+pub fn build_done_token_with_status(status: u16, row_count: u64) -> BytesMut {
     let mut token_data = BytesMut::new();
 
     // DONE token (0xFD)
     token_data.put_u8(TokenType::Done as u8);
 
-    // Status: DONE_FINAL (0x00) - little-endian
-    token_data.put_u16_le(0x0000);
+    // Status - little-endian
+    token_data.put_u16_le(status);
 
     // CurCmd: SELECT (0xC1) - little-endian
     token_data.put_u16_le(0x00C1);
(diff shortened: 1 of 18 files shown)


saurabh500:

PR #205, opened 2026-08-10
Rewire mssql-py-core onto the reactor-free sync core (opt-in sync cursor)

saurabh500:
## Description

**L6 — Rewire `mssql-py-core` onto the reactor-free sync TDS core.**

> **Framing (correctness/architecture, not a perf win yet):** This layer (and the whole L-stack) is a correctness/architecture refactor — one protocol core with sync and async shells that deletes the per-row `block_on` tax and the `fast::*` hack. It is NOT a native-beating performance win yet. VarcharMax/PLP (~818× on both the async-before and sync-after variants, pre-existing), the Decimal/DateTime2 conversion gaps, and the per-column `SQLGetData` conversion/alloc path are the real performance debt to burn down next. It earns its keep on architecture and correctness, and it sets up the perf work rather than delivering it.

Stacked on the frozen **L5** tip (`dev/saurabh/rewire-odbc-sync-core-l5` @ `5c09f0e4`). Append-only single commit; **base = L5 branch**, not `main`. Scope is `mssql-py-core` only — all frozen crates (`mssql-tds`, `mssql-odbc`, `mssql-mock-tds`, `mssql-js`, `mssql-tds-cli`) are empty-diff.

### What this does
The Python bindings now share ONE protocol core across two cursor surfaces via a flip-in-place ownership cell:

- `enum PyClient { Async(TdsClient) | Sync(TdsSyncClient) | Transitioning | Dead(String) }` behind `Arc<std::sync::Mutex<PyClient>>` (`pyclient.rs`). The `Arc` points at the **cell**, so `Arc::try_unwrap` is never used — cursors keep their clone and match the enum under the lock.
- Flip is `std::mem::replace(&mut *guard, Transitioning)` → call `into_sync`/`into_async` on the moved-out value → store the result variant back. `Transitioning` is a sentinel that only exists between take and store (no early-return while transitioning); an unrecoverable flip stores `Dead(String)` and surfaces `Err`.
- **Runtime-context hard rule:** every `into_sync()` runs inside `handle.enter()` so the runtime handle is captured for the later revert (a naked `into_sync()` off-runtime would capture `None` and poison `into_async`).
- **Revert-before-control-plane:** revert to `Async` at end-of-rows / before `close_query` / before any control-plane op, mirroring L5.

### Cursor surfaces (mode A for this layer)
- **`PyCoreSyncCursor`** (new): reactor-free row-pull. Plaintext connections flip to the `Sync(TdsSyncClient)` arm and drive `next_row_into` with **no `block_on`**; TLS/non-eligible transports fall back to the `block_on`-over-async path, byte-identically.
- **`PyCoreCursor`** (existing): kept exactly as-is — `block_on`-over-async backed. **Public Python API unchanged** — no regression. This is **not** a first-class async cursor: its `block_on` blocks the Python event loop. The genuine coroutine (`async def` / awaitable) cursor is **L7**, not this layer.
- `Connection.cursor()` / `Connection.sync_cursor()` expose the two surfaces.
- `execute` / COLMETADATA / advance / close / DML / bulkcopy / auth all **stay async** (control-plane rule); the sync flip is for the pure SELECT row-pull hot loop only, per result set.

### rowcount
Additive `rowcount` captured from `last_rows_affected()` **pre-flip on the async arm** (faithful-count parity contract, populated by the shared `apply_row_read_token` count_map at L4 B′/L5). No independent count logic.

### Scope decision (mode A) — real coroutine async cursor is L7
This layer ships **mode A**: sync arm reactor-free + the existing cursor retained as `block_on`-over-async. A genuinely first-class **coroutine** async cursor (`pyo3-async-runtimes` / `future_into_py` driving `next_row_into().await` on the `Async` arm, event-loop driven, non-blocking) is a large net-new public surface and lands as its own reviewable layer **L7** stacked on top of this one. No `pyo3-async-runtimes` dependency is added here; `Cargo.lock` is unchanged.

### Build / test path
`mssql-py-core` is edition 2024 and **excluded from the cargo workspace**, so the plain `cargo b*` aliases do NOT cover it — validated via `scripts\bfmt.ps1` + `scripts\bclippy.ps1` (both cover py-core) + `maturin` + the Python suite through the existing `conftest.py` fixtures.

### Tests
- Sync-cursor fetch on a **plaintext** connection (reactor-free sync path).
- Sync-cursor **TLS-fallback** (`block_on`-over-async path) — env-skips locally on the pre-existing mock-TLS-on-Windows timeout; CI-validated.
- **Attribute-set parity**: the new sync cursor and the existing `block_on`-backed cursor expose the same surface.
- **Byte parity**: sync `fetchone` results identical to the `block_on` path.
- **Error-mid-fetch** sync recovery: mid-stream error reverts via `into_async` drain, connection reused (validated live: SQL 245 conversion error mid-stream).
- DML rowcount + multi-row SELECT parity validated against a live localhost SQL Server.

Gates: `scripts\bfmt.ps1` clean, `scripts\bclippy.ps1` (-D warnings) clean (workspace + py-core), `maturin develop` builds `mssql_py_core` + `mssql_mock_tds`, mock cursor suite = 10 passed / 1 env-skip.

## Checklist

- [x] `cargo bfmt` passes 
- [x] `cargo bclippy` passes 
- [x] `cargo btest` passes 

---

**Stack governance:** DRAFT — the whole stack flips to ready-for-review together, bottom-to-top, after parent sign-off.

Related work item / issue: Tracked as part of the sans-I/O native stack (#192)

### DIFF
--- a/mssql-py-core/src/connection.rs
+++ b/mssql-py-core/src/connection.rs
@@ -6,14 +6,13 @@ use pyo3::prelude::*;
 use pyo3::types::PyDict;
 use std::{path::PathBuf, sync::Arc};
 use tokio::runtime::Runtime;
-use tokio::sync::Mutex;
 
 use crate::odbc_auth::odbc_authentication_transformer::transform_auth;
 use crate::odbc_auth::odbc_authentication_validator::validate_auth;
+use crate::pyclient::{self, SharedClient};
 use crate::python_logger_adapter::scoped_tracing_bridge;
 use mssql_tds::{
     connection::client_context::{ClientContext, IPAddressPreference},
-    connection::tds_client::TdsClient,
     connection_provider::tds_connection_provider::TdsConnectionProvider,
     core::{EncryptionOptions, EncryptionSetting},
     message::login_options::ApplicationIntent,
@@ -26,7 +25,7 @@ const DEFAULT_LIBRARY_NAME: &str = "MS-PYTHON";
 pub struct PyCoreConnection {
     #[allow(dead_code)] // Used for async operations in cursor execute
     runtime: Runtime,
-    tds_client: Option<Arc<Mutex<TdsClient>>>,
+    tds_client: Option<SharedClient>,
     is_closed: bool,
 }
 
@@ -82,7 +81,7 @@ impl PyCoreConnection {
                 tracing::info!("Successfully connected to SQL Server");
                 Ok(PyCoreConnection {
                     runtime,
-                    tds_client: Some(Arc::new(Mutex::new(client))),
+                    tds_client: Some(pyclient::new_shared(client)),
                     is_closed: false,
                 })
             }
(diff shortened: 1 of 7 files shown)


saurabh500:

PR #206, opened 2026-08-10
AQE: Async Runtime Infrastructure

subrata-ms:
## Description
This pull request introduces a new asynchronous API surface to the `mssql-py-core` package, providing Python users with async connection and cursor types for interacting with SQL Server using asyncio. The async implementation is marked as a preview and is isolated from the existing synchronous API. Key changes include the addition of async connection/cursor types, a shared Tokio runtime for async operations, and integration of these types into the Python module initialization.

**New Asynchronous API:**

* Added `PyAsyncConnection` and `PyAsyncCursor` types, providing Python awaitable methods for connecting, closing, committing, rolling back, and creating cursors in an async context. These types use the shared Tokio runtime and return awaitables compatible with Python's asyncio. The API is marked as unstable/preview and emits a `FutureWarning` on first use. (`mssql-py-core/src/async_connection.rs`, `mssql-py-core/src/async_cursor.rs`) [[1]](diffhunk://#diff-068c4f95015d77dcdf6d0491cb93d871a936adaefac41ea257ac7d5ff4f7f05aR1-R289) [[2]](diffhunk://#diff-85fa845be93caab797168826f6e49584a85ce9577adb715c7dc051dd4919a179R1-R65)

* Integrated the new async types into the Python extension module, registering both classes and initializing the shared Tokio runtime at module load. This ensures async connections and awaitables work consistently across the process. (`mssql-py-core/src/lib.rs`) [[1]](diffhunk://#diff-59116afdbd4d7619c126efc429f65543c6b801d90e3d6955e8f60c3ab3a067fdR10-R12) [[2]](diffhunk://#diff-59116afdbd4d7619c126efc429f65543c6b801d90e3d6955e8f60c3ab3a067fdR55-R58) [[3]](diffhunk://#diff-59116afdbd4d7619c126efc429f65543c6b801d90e3d6955e8f60c3ab3a067fdR93-R94)

**Async Runtime Management:**

* Added a process-wide, lazily-initialized Tokio runtime specifically for the async API. This avoids creating multiple thread pools for each connection and ensures all async operations share the same executor. (`mssql-py-core/src/async_runtime.rs`)

**Dependency and Utility Updates:**

* Added `pyo3-async-runtimes` as a dependency to bridge Rust futures to Python awaitables, and updated `Cargo.toml` accordingly. (`mssql-py-core/Cargo.toml`)
* Made `dict_to_client_context` public within the crate to allow async connection creation to reuse the same client context parsing as the sync code. (`mssql-py-core/src/connection.rs`)
* Added a TODO to the TDS error conversion utility for future compliance with Python DB-API error classes. (`mssql-py-core/src/utils.rs`)

## Related Issues

ADO work item: https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/46956 , https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/46957
## Checklist

- [ ] `cargo bfmt` passes
- [ ] `cargo bclippy` passes
- [ ] `cargo btest` passes


mssql-py-core/Cargo.toml
```diff
@@ -27,3 +28,11 @@ arrow = { version = "55", default-features = false, features = ["ffi"] }
...
 [build-dependencies]
 pyo3-build-config = "0.29.0"
+
+[features]
+# `async-preview` gates the entire asynchronous surface (`PyAsyncConnection`,
+# `PyAsyncCursor`, and the shared Tokio runtime that backs them). Disabled by
+# default: production wheels expose only the stable synchronous API. Downstream
+# packagers opt in explicitly by building with `--features async-preview`.
+default = []
+async-preview = []
```
saurabh500: Why gate the feature. These are anyhow parallel set of APIs which will leverage async-preview. This by design makes the new APIs opt-in and we don't need to create 2 different py-core binaries for testing in the test pipeline also. 

Rust features only make sense when we are OK with 2 binaries being built separately. That also means engineering the pipelines to test with 2 different feature sets. I am hoping we can keep this simple and remove the feature gating for mssql-py-core

mssql-py-core/src/async_runtime.rs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT License.
+
+//! Process-wide Tokio runtime shared by every `PyCoreConnection` and by the
+//! Python `asyncio` bridge (`pyo3_async_runtimes::tokio`).
+//!
+//! Rationale:
+//! * A single multi-threaded runtime is created lazily on first use and reused
+//!   for every connection, cursor, and awaitable returned to Python. This
+//!   avoids the per-connection worker-thread explosion of the previous model
+//!   where each `PyCoreConnection` owned its own [`tokio::runtime::Runtime`].
```
saurabh500: Doc-vs-reality gap: this says the runtime is "shared by every `PyCoreConnection`" and that it replaces "the previous model where each `PyCoreConnection` owned its own `Runtime`." Neither is true in this PR — the sync `PyCoreConnection` in `connection.rs` still calls `Runtime::new()` per connection and never touches this shared runtime. Today the process runs two regimes side by side: N per-connection sync runtimes + this one shared async runtime.

For the async surface the single-shared-runtime claim *is* accurate (`future_into_py` resolves the process-global runtime via `get_runtime()`), so suggest scoping the wording to the async surface (e.g. "shared by every async connection and the asyncio bridge") and dropping/qualifying the sync-consolidation claim until that migration actually happens.

mssql-py-core/src/async_connection.rs
```diff
@@ -0,0 +1,282 @@
...
+        // Clone the Arc synchronously so the future is `'static + Send`
+        // without borrowing `self`. The `&mut self` borrow is released as
+        // soon as this method returns.
+        let client = self
+            .tds_client
+            .as_ref()
+            .ok_or_else(|| PyRuntimeError::new_err("Connection is closed"))?
+            .clone();
+
+        pyo3_async_runtimes::tokio::future_into_py(py, async move {
```
saurabh500, line 215: Cancellation safety: in pyo3-async-runtimes 0.29, `future_into_py` cancels the Rust future when the returned asyncio future is cancelled (the `Cancellable` wrapper in `generic.rs`, "new behaviour in v0.15"). So `await asyncio.wait_for(conn.commit(), timeout=...)` — or any cancel — can drop this future while it is suspended inside `commit_transaction().await`, after the `TM_COMMIT` bytes are on the wire but before the server's `DONE` token is read.

The `MutexGuard` is released, but the shared `TdsClient`'s TDS byte stream is now desynced: a subsequent `rollback()` (or, later, a cursor `execute`) on the same connection will read the stale response as its own -> protocol corruption on a live connection. The mutex gives mutual exclusion, not cancellation atomicity. Same applies to `rollback` (L245) and `close` (L164), and will apply to cursor I/O once it shares this client.

For a gated preview, documenting this is acceptable, but the GA design should either run these ops non-cancellably (spawn + await the join handle) or poison the connection when a future is dropped mid-flight. Cheapest to decide before cursor I/O piles onto the same client.

mssql-py-core/src/async_connection.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// ```
+    ///
+    /// Sends a TM_COMMIT (Transaction Manager COMMIT) request over the wire
+    /// and awaits the server's DONE token. Raises `RuntimeError`
+    /// synchronously if the connection has already been closed.
+    ///
+    /// If no transaction is currently open on the server, the commit will
+    /// fail with the server's own error (SQL Server 3902 — "The COMMIT
+    /// TRANSACTION request has no corresponding BEGIN TRANSACTION").
+    fn commit<'py>(&mut self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
```
saurabh500: Nit: `commit` (and `rollback` at L235) take `&mut self` but only read `self.tds_client` via `.as_ref()` and clone the `Arc`. `&self` would suffice — as `cursor()` already does — and lets Python hold a shared rather than exclusive borrow. Harmless today, just slightly over-restrictive.

mssql-py-core/src/lib.rs
```diff
@@ -49,6 +55,11 @@ fn mssql_py_core(m: &Bound<'_, PyModule>) -> PyResult<()> {
     // Initialize tracing on module load (via MSSQL_TDS_TRACE env var)
     tracing_init::init_tracing();
 
+    // Bring up the shared Tokio runtime used by every connection, cursor, and
+    // `asyncio`-facing coroutine. Must run before any code that touches Tokio.
```
saurabh500: Nit: the comment says init "Must run before any code that touches Tokio," but nothing else in this function touches Tokio, so the real constraint is just "before the first connection" — which holds regardless of position. Slightly overstates the ordering coupling. (For the record: `pyo3_async_runtimes::tokio::init` panics if called twice; the `Once` guard in `async_runtime::init` makes that unreachable, which is correct.)

saurabh500 (commented):
Reviewed the async preview infrastructure. The feature gating is clean, symmetric, and correctly defaults off, and the `Arc<Mutex<TdsClient>>` ownership pattern (clone/take synchronously before entering `future_into_py`) is exactly right for producing `'static + Send` futures. Verified it compiles clean under `cargo clippy --features async-preview --all-targets -- -D warnings`.

Submitting as **Comment**. Two items worth resolving before cursor I/O lands on the shared client (details inline):

1. Cancellation safety of the async lifecycle methods vs. `future_into_py`'s cancel-on-drop behavior — the main one.
2. `async_runtime.rs` module doc describes a sync-runtime consolidation that hasn't happened yet.

Also a heads up: the PR body has no linked GitHub issue / ADO work item, and there are no tests exercising the async surface (it builds in CI via `--all-features` but nothing runs it). Not blocking for a gated preview, but flagging.

mssql-py-core/src/lib.rs
subrata-ms: fair point fixed in next commit.

mssql-py-core/src/async_connection.rs
subrata-ms: Agreed. &mut self here gives real zero protection. It probably misleads the reader. Changing it to &self

mssql-py-core/src/async_connection.rs
saurabh500: @subrata-ms you could add a todo here and revisit this part of the design. This is critical though

mssql-py-core/src/async_connection.rs
```diff
@@ -0,0 +1,282 @@
...
+use crate::connection::PyCoreConnection;
+
+/// Emit a `FutureWarning` the first time any async API is exercised in this
+/// process. Silenceable by callers via `warnings.filterwarnings(...)`.
+static PREVIEW_WARNED: Once = Once::new();
+
+fn emit_preview_warning(py: Python<'_>) {
+    PREVIEW_WARNED.call_once(|| {
+        let category = py.get_type::<PyFutureWarning>();
+        // stacklevel=2 attributes the warning to the caller's `await` site
```
David-Engel: `stacklevel=2` attributes the warning one frame too high, and the comment describes the wrong moment.

There is no Python frame for a native `#[pymethods]` function, so `stacklevel=1` already lands on the user's `PyAsyncConnection.connect(...)` line; `2` blames that caller's caller. Verified empirically by calling `PyErr_WarnEx` through `ctypes` with no intervening Python frame:

```python
def user_code(level):
    py.PyErr_WarnEx(FutureWarning, b"preview", level)   # line 6

def outer(level):
    user_code(level)                                    # line 9
```
```
stacklevel=1 -> line 6   # the user's connect(...) call site
stacklevel=2 -> line 9   # one frame too far
```

The comment is also inaccurate in a second way: the warning is emitted at *call* time, before `future_into_py` is reached, so it never has anything to do with the `await` site.

```rust
// stacklevel=1: there is no Python frame for this native method, so 1 already
// attributes the warning to the caller's `connect(...)` line.
```

(Worth folding into whatever change addresses Copilot's `PyErr::warn` comment above, since it's the same call.)

David-Engel (commented):
Reviewed the diff independently and then read the existing threads — most of what I found was already covered by Copilot and @saurabh500, so I'm only posting the four items nobody has raised yet. I verified the branch compiles and lints clean with `cargo clippy --all-features --all-targets` in a scratch worktree.

Explicitly **not** re-raising (already well covered): cancellation safety of the wire commands, the missing tests, the swallowed `PyErr::warn`, the `async_runtime.rs` doc-vs-reality gap, `&mut self` on `commit`/`rollback`, and the non-`optional` `pyo3-async-runtimes` dependency. I agree with all of them, and with @saurabh500 that the cancellation question is the one to settle before cursor I/O piles onto the same client.

mssql-py-core/Cargo.toml
David-Engel: Concrete data point for @saurabh500's question above about whether the gate earns its keep: as things stand, **nothing in CI ever runs this code**, and only one job even compiles it.

- `.pipeline/scripts/containerized-clippy.sh` -> `scripts/bclippy.sh`, which does pass `--all-features` for `mssql-py-core`. This is the only thing that type-checks the async modules.
- `.pipeline/scripts/containerized-pycore-rust-test.sh` runs `cargo llvm-cov nextest --frozen --profile ci` with **no** `--all-features`, so `async_connection.rs` / `async_cursor.rs` / `async_runtime.rs` are not compiled during the test job at all — they can't be tested and they don't appear in the coverage report.

So the feature currently buys isolation that the dependency isn't actually giving us (`pyo3-async-runtimes` is non-optional, per Copilot's comment at line 13) while costing us all test signal.

If the gate stays, `containerized-pycore-rust-test.sh` needs `--all-features` on the nextest invocation, and the wheel pipelines need a documented `--features async-preview` build path or no one can consume the preview. If the gate goes, all of that disappears. Either is fine — just not the current middle state.

mssql-py-core/src/async_connection.rs
```diff
@@ -0,0 +1,282 @@
...
+    ///
+    /// ```python
+    /// conn = await PyAsyncConnection.connect(client_context_dict)
+    /// ```
+    ///
+    /// Dictionary parsing runs synchronously on the calling thread (it needs
+    /// the GIL). The network handshake is submitted to the shared Tokio
+    /// runtime and driven concurrently with the caller's asyncio loop.
+    #[classmethod]
+    fn connect<'py>(
```
David-Engel, line 100: Parity gap with the sync surface: `connect` drops the Python logging bridge.

`PyCoreConnection::new` takes an optional `python_logger` and installs `scoped_tracing_bridge` for the duration of the call, so connection-time `tracing` output reaches the caller's Python logger. `PyAsyncConnection::connect` has no such parameter, so every `tracing::info!`/`error!` in this file — including the connection-failure diagnostics — goes only to the `MSSQL_TDS_TRACE` sink. That makes async connect failures materially harder to diagnose than sync ones for anyone driving this from `mssql-python`.

Also missing relative to `PyCoreConnection`: `is_connected()`, `__repr__`, and `__aenter__`/`__aexit__` (the last being the idiomatic async-context-manager shape callers will expect).

All plausibly follow-up work — worth listing in the PR description so it isn't lost, since the description currently reads as though the async surface is at parity.

mssql-py-core/src/async_cursor.rs
```diff
@@ -0,0 +1,66 @@
...
+//! notice in any release.
+//!
+//! Sibling of `cursor.rs` (the synchronous surface). A [`PyAsyncCursor`] is
+//! bound to exactly one [`crate::async_connection::PyAsyncConnection`] and
+//! shares that connection's `TdsClient` via an `Arc<tokio::sync::Mutex<_>>`.
+//! All I/O methods will submit their futures to the shared process-wide
+//! Tokio runtime and return Python awaitables through
+//! `pyo3_async_runtimes::tokio::future_into_py`.
+//!
+//! Invariant: one connection ↔ one cursor ↔ one `TdsClient` ↔ one TDS wire
```
David-Engel: This module doc asserts `Invariant: one connection <-> one cursor`, but `PyAsyncConnection::cursor` (async_connection.rs L271-273) says the opposite: "A second cursor may be created on the same connection, but both cursors share one TDS wire session and serialize on the same async mutex."

The `cursor()` wording is the one that matches the code — nothing prevents a second cursor. Suggest dropping "one connection <-> one cursor" from the invariant here and keeping the part that's actually load-bearing:

```rust
//! Invariant: one connection <-> one `TdsClient` <-> one TDS wire session.
//! Multiple cursors may exist, but they all serialize on the same
//! `Mutex<TdsClient>` and share that one session, matching the non-MARS model.
```

mssql-py-core/src/async_connection.rs
David-Engel: +1 on this being the item to settle now rather than after cursor I/O lands. Confirmed the mechanism in the vendored source: `pyo3_async_runtimes::generic::future_into_py_with_locals` documents "If the `asyncio.Future` returned by this conversion is cancelled via `asyncio.Future.cancel`, the Rust future will be cancelled as well (new behaviour in v0.15)", so this is cancel-on-drop, not cancel-ignored.

One argument for fixing rather than TODO-ing: the non-cancellable version is about four lines, and it gives `async_runtime::handle()` its first real caller (it is currently `#[allow(dead_code)]`).

```rust
pyo3_async_runtimes::tokio::future_into_py(py, async move {
    crate::async_runtime::handle()
        .spawn(async move {
            let mut guard = client.lock().await;
            guard.commit_transaction(None, None).await
        })
        .await
        .map_err(|e| PyRuntimeError::new_err(format!("Commit task failed: {e}")))?
        .map_err(|e| PyRuntimeError::new_err(format!("Commit failed: {e}")))
})
```

Cancelling the asyncio side now only drops the *awaiter*; the spawned task keeps the guard, finishes the command, and drains the response, so the session stays in sync. The tradeoff is that a cancelled `commit()` still commits — which is the honest semantic for TDS anyway, since the bytes are already on the wire and there is no way to un-send them. `asyncio.wait_for` timing out would return control to the caller while the commit completes in the background.

The alternative (poisoning the connection on drop) needs a drop guard on the future and leaves the caller with a dead connection after any timeout, which seems strictly worse for the common `wait_for` case.

Whichever way it goes, the same shape should be applied to `rollback` and to cursor `execute`/`fetch` when they land — that is the part worth deciding before more call sites exist.

mssql-py-core/Cargo.toml
subrata-ms: My thought was to provide an extra level of gates to protect the incomplete feature. But you are also right and this is pretty confirmed that there is no issue until user import the feature in client code. Hence bearing the CI cost for having an extra layer of Gates is not necessary as the feature is additive in nature. Removing the Gate related changes in my next commit.

mssql-py-core/Cargo.toml
```diff
@@ -10,6 +10,7 @@ crate-type = ["cdylib"]
 
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["extension-module"] }
+pyo3-async-runtimes = { version = "0.29", features = ["tokio-runtime"] }
```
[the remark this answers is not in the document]
subrata-ms: Thanks — you're right that the feature gate before e115c39 was cosmetic; [pyo3-async-runtimes](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) was resolved unconditionally because the dep wasn't optional = true. That's part of the reason another reviewer asked us to drop the gate entirely, and we did so in commit e115c39. [pyo3-async-runtimes](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is now an unconditional dependency and the preview signal is carried purely by the runtime FutureWarning + doc-comments on the pyclasses.

mssql-py-core/src/async_cursor.rs
```diff
@@ -0,0 +1,66 @@
...
+    ///
+    /// Called only from `PyAsyncConnection::cursor`.
+    pub(crate) fn new(tds_client: Arc<Mutex<TdsClient>>) -> Self {
+        Self { tds_client }
+    }
+}
+
+#[pymethods]
+impl PyAsyncCursor {
+    // Async execute/fetch/close APIs land here as they are added.
```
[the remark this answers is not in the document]
subrata-ms: Correct on the untested surface — [PyAsyncCursor](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is intentionally an empty scaffold in this PR. This PR only covers the async runtime foundation and connection lifecycle ([connect](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [close](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [commit](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [rollback](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html)); implementing [execute](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) / [fetchone](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) / [fetchall](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) / [close](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) on the cursor is a tracked follow-up (task <ID>). The preview FutureWarning on [connect()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and the ⚠️ doc-comments on both PyAsyncConnection and [PyAsyncCursor](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) signal explicitly that this surface is not yet consumable — customers should not use it. Nothing here changes existing sync [PyCoreConnection](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) / [PyCoreCursor](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) / bulkcopy behavior, and CI diff coverage is informational-only per [pr-code-coverage.yml](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), so this does not block the merge.

mssql-py-core/src/async_cursor.rs
subrata-ms: modified the phrase as below -
//! Invariant: one `TdsClient` per connection, one TDS wire session per
//! `TdsClient`, and all access serialized through the async mutex. Creating
//! a second cursor on the same connection is allowed — both cursors share
//! the same client and serialize on the same mutex, so wire integrity is
//! preserved.

mssql-py-core/src/async_runtime.rs
subrata-ms: Agreed. I have reverted few changes earlier during development but forgot to revert the documentation part. That mistake caused this review comment. Thanks for catching it

mssql-py-core/src/async_connection.rs
subrata-ms: Agreed. Currently added a comment as TODO because this would be taken care under below user story.
https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47180

mssql-py-core/src/async_connection.rs
subrata-ms: Thanks @david-Engle for highlighting this. I have fixed in my latest commit. However, emit_preview_warning functionality will be completely removed once we are production ready.

mssql-py-core/src/async_connection.rs
```diff
@@ -0,0 +1,282 @@
...
+fn emit_preview_warning(py: Python<'_>) {
+    PREVIEW_WARNED.call_once(|| {
+        let category = py.get_type::<PyFutureWarning>();
+        // stacklevel=2 attributes the warning to the caller's `await` site
+        // rather than to this Rust helper.
+        if let Err(e) = PyErr::warn(
+            py,
+            &category,
+            c"mssql_py_core async API is a preview and subject to breaking changes without notice; do not depend on it from production code.",
+            2,
```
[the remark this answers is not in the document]
subrata-ms: fixed

mssql-py-core/src/async_connection.rs
subrata-ms: @David-Engel all are valid points. But this PR is only targeting Async Runtime infrastructure (https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/46900). I should be implementing this under logger integration with python and under below two task - 
1. https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/46961
2. https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47078

mssql-py-core/src/async_connection.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// ```
+    ///
+    /// Dictionary parsing runs synchronously on the calling thread (it needs
+    /// the GIL). The network handshake is submitted to the shared Tokio
+    /// runtime and driven concurrently with the caller's asyncio loop.
+    #[classmethod]
+    fn connect<'py>(
+        cls: &Bound<'py, PyType>,
+        client_context_dict: &Bound<'_, PyDict>,
+    ) -> PyResult<Bound<'py, PyAny>> {
```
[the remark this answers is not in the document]
subrata-ms: added some test cases but this test cases need to be revisited for modification once end-to-end feature is ready.

mssql-py-core/Cargo.toml
subrata-ms: this is fixed now under last few commits. I have removed the gated the async surface changes and added some test cases. Test cases are subject to modification as we evolve towards completion of the Async feature

sumitmsft:
This PR maps every server/transport failure on `connect`, `commit`, and `rollback` to a bare `PyRuntimeError` (e.g. "Failed to connect to SQL Server", "Commit failed", "Rollback failed"). I want to flag this now so it doesn't harden into the shipped contract.

The design doc is explicit that we must be DB-API 2.0 compliant on error mapping. All async errors surface as subclasses of mssql_python.exceptions, the same hierarchy the sync path uses, so existing "except" handlers work. 

Every abnormal outcome maps to exactly one Python exception class from a small, documented hierarchy. No RuntimeError catch-all in user-facing paths. The server errors must raise from the DB-API hierarchy (DatabaseError / ProgrammingError / OperationalError), preserving message and, where available, SQLSTATE / server error number.

A raw PyRuntimeError breaks that contract in two concrete ways: it isn't part of the DB-API exception tree, so a caller's except mssql_python.DatabaseError / except mssql_python.OperationalError won't catch it and flattening everything to one string discards the SQLSTATE / server error number. 

I understand the real mapping lands with the error-taxonomy work, not this infrastructure PR, but since these strings are user-visible the moment the preview is exercised, could we add a TODO at each of these three sites and the single shared convert_tds_error style function, so it's unambiguous these are placeholders and not the intended DB-API-compliant contract? That keeps us honest that the async and sync paths must converge on one hierarchy.

mssql-py-core/src/async_connection.rs
```diff
@@ -0,0 +1,286 @@
...
+            if let Err(e) = guard.close_connection().await {
+                // Match sync-path semantics: log and swallow. The connection
+                // is treated as closed regardless — the transport will be
+                // dropped when the Arc's last reference goes away.
+                tracing::warn!(
+                    "PyAsyncConnection::close: error during graceful shutdown: {}",
+                    e
+                );
+            }
+            Ok(())
```
David-Engel: These futures resolve to an empty Python tuple, not `None`.

In pyo3 0.29 `impl<'py> IntoPyObject<'py> for ()` has `type Target = PyTuple` and returns `PyTuple::empty(py)` (verified in the vendored `pyo3-0.29.2` source). So `await conn.close()` — and `commit()` / `rollback()` at L222 and L257 — hand Python back `()` where every other Python API returns `None`. It's benign at runtime (empty tuple is falsy), but it's an odd shape to bake into a new public async API, and the tests currently encode it as `result is None or result == ()`.

Cheapest fix while the surface is still preview — resolve to `None` explicitly:

```rust
Python::attach(|py| Ok(py.None()))
```

Then the tests can tighten to `assert result is None`.

mssql-py-core/tests/test_async_connection.py
```diff
@@ -0,0 +1,173 @@
...
+            conn = await mssql_py_core.PyAsyncConnection.connect(client_context)
+            try:
+                # With no active transaction, SQL Server returns error 3902
+                # which the driver surfaces as RuntimeError. Either the server
+                # accepts the commit (auto-commit mode) or raises — both are
+                # acceptable outcomes for this plumbing test.
+                try:
+                    result = await conn.commit()
+                    assert result is None or result == ()
+                except RuntimeError as e:
```
David-Engel: Walking back a suggestion I made earlier in this review, in light of @sumitmsft's comment on the error contract.

I was going to suggest tightening this to `pytest.raises(RuntimeError, match="Commit failed")`, since the assertion as written passes whether the commit succeeds *or* raises — the only regression it can catch is a non-`RuntimeError` escaping, which makes it close to a tautology.

But `RuntimeError` / `"Commit failed"` is exactly the placeholder contract @sumitmsft is asking us *not* to harden, so pinning the test to it would have to be undone when the DB-API taxonomy lands. Better to make it deterministic while staying agnostic about the exception class:

```python
# TM_COMMIT with no open transaction always yields SQL Server 3902, and
# PyAsyncConnection exposes no begin_transaction, so this is the only path.
# Matching on the server error number rather than the exception class keeps
# this valid after the DB-API error taxonomy lands.
with pytest.raises(Exception, match="3902"):
    await conn.commit()
```

That works today: `SqlErrorInfo`'s `Display` in `mssql-tds/src/error/mod.rs` renders `"Sql Error: {number}: Class {class}: State {state}: {message} ..."`, so the server error number is already in the string. Same shape applies to `rollback` / 3903 at L140.

mssql-py-core/tests/test_async_connection.py
```diff
@@ -0,0 +1,173 @@
...
+# ---------------------------------------------------------------------------
+# Preview warning
+# ---------------------------------------------------------------------------
+
+# This test must run before any other test in this session that awaits
+# PyAsyncConnection.connect(). Once the warning fires successfully,
+# PREVIEW_WARNED (a process-wide AtomicBool in Rust) is set to True and the
+# warning cannot be re-observed in this process. If invocation order places
+# another connect-issuing test before this one, this test self-skips.
+def test_future_warning_propagates_when_promoted_to_error():
```
David-Engel, line 23: Nit — this test fails open, so a real regression would show up as a skip rather than a failure.

The docstring is upfront about `PREVIEW_WARNED` being a process-wide latch and about the ordering constraint, which I appreciate. The part that worries me is the skip branch: it triggers on *any* non-`FutureWarning` exception. If the warning ever stops being promoted to an error, `connect({})` falls through to a connection attempt, fails with `RuntimeError`, and this test reports `skipped` — green build, silent loss of the only assertion covering the preview signal.

Running it in a fresh interpreter makes it deterministic and order-independent, and removes the need for the ordering caveat entirely:

```python
def test_future_warning_propagates_when_promoted_to_error():
    script = textwrap.dedent('''
        import asyncio, warnings, sys, mssql_py_core
        warnings.simplefilter("error", FutureWarning)
        async def main():
            try:
                await mssql_py_core.PyAsyncConnection.connect({})
            except FutureWarning:
                sys.exit(0)
            sys.exit(1)
        asyncio.run(main())
    ''')
    subprocess.run([sys.executable, "-c", script], check=True)
```

Low priority given it passes today (I confirmed it genuinely asserts rather than skipping in a local unit-only run) — but worth it before more tests start competing for the latch.

David-Engel (commented):
Re-reviewed at `8ee6a76e`. All four of my earlier items are addressed or consciously deferred with tracking, and the async surface now has tests that actually execute in CI — nice turnaround.

**Verified locally:** `cargo clippy --all-targets -- -D warnings` clean with default features; built the extension via `dev/test-python.sh --skip-integration` and all 5 non-integration async tests pass, including `test_future_warning_propagates_when_promoted_to_error`, which confirms the `PyErr::warn` propagation fix works end to end rather than just reading correctly. On the CI side, the ADO validation build 166415 task "Test Python bindings (mssql-py-core)" succeeded, and since `dev/test-python.sh` runs `pytest tests/ -m "not longhaul and not smoke"` against the live `sql1` container, the integration-marked async tests genuinely ran against a server. Dropping the feature gate resolved my earlier "CI never runs this code" concern without needing any pipeline edit.

Everything below is non-blocking polish. All of it is the same theme @sumitmsft raised in his comment on the DB-API error mapping: **placeholder semantics on a preview surface tend to harden into the shipped contract**, so it's cheapest to mark or fix them now. His comment is the more consequential instance of this; mine are smaller siblings. In particular, one of the inline notes below walks back a test suggestion I made earlier that would have cut against his ask.

David-Engel:
@sumitmsft +1 on getting TODOs at those three sites plus the shared conversion function before the strings become the de facto contract. One factual refinement that I think strengthens the ask rather than weakening it.

The comment says flattening to one string "discards the SQLSTATE / server error number." That is not quite what happens today. `SqlErrorInfo`'s `Display` in `mssql-tds/src/error/mod.rs` (L37-L51) renders:

```
Sql Error: {number}: Class {class}: State {state}: {message} on {server} in {proc} at line {line}
```

so the number and state do survive into the `PyRuntimeError` message — just as unparseable text a caller would have to regex.

The real loss is structural, and it is one layer up. `Error::SqlServerError { diagnostics }` already carries a `Vec<SqlErrorInfo>` with `number`, `class`, `state`, `server_name`, `proc_name`, and `line_number` as typed fields, and multiple errors per batch are preserved. All of that structure is discarded by `format!("Commit failed: {e}")` at the pyo3 boundary.

So the mapping function you are asking for has a fully populated, already-structured source to work from — this is a boundary problem, not a data-availability problem, which should make the eventual DB-API mapping mostly mechanical. Worth capturing in the TODO text so whoever picks up the taxonomy work knows the diagnostics are already there and does not go re-plumbing the TDS layer.

Related: I left a note on `tests/test_async_connection.py` walking back a suggestion I had made to pin the transaction tests to `pytest.raises(RuntimeError, match="Commit failed")`. That would have hardened exactly the placeholder you are flagging, so I switched it to matching on the server error number (3902 / 3903), which stays valid after the taxonomy lands.

mssql-py-core/src/async_connection.rs
subrata-ms: fixed

mssql-py-core/tests/test_async_connection.py
subrata-ms: fixed

mssql-py-core/tests/test_async_connection.py
subrata-ms: fixed

sumitmsft:
[david-engel](https://github.com/David-Engel) Confirmed and agreed, the diagnostics are already structured in `Error::SqlServerError`, so our user story will be just a boundary-mapping job; worth noting that in the TODO text so nobody re-plumbs the TDS layer.

[subrata](https://github.com/subrata-ms) One optional refinement from David's thread: consider enriching the TODO text to note that `Error::SqlServerError { diagnostics }` already exposes the typed fields. That way you'd remember the diagnostics are already structured at the TDS layer and won't go re-plumbing it.

sumitmsft (approved):
All comments have been addressed, and we are good to go. Approving this PR.

PR #207, opened 2026-08-10
Add mssql-tds support for incremental PLP parameter write path

shiwanigupta0809:
## Description

Adds a data-at-execution (streamed) PLP parameter write path to mssql-tds for large nvarchar(max) / varchar(max) / varbinary(max) values in sp_executesql. The value is streamed to the server in chunks instead of being materialized up front, mirroring msodbcsql's incremental PLP write. This is the TDS-layer foundation for a later ODBC SQLPutData / SQLParamData. Scope: mssql-tds only.

What this adds
Streaming entry points on TdsClient, mirroring ODBC's begin/put/end:
begin_sp_executesql - sends the RPC prefix (headers, positional args, materialized params) + the first streamed param's header, parks the half-written message, returns NeedData.
write_streamed_chunk - appends one length-prefixed chunk (empty chunks ignored a zero-length header is the terminator). Writes the unknown-length opener lazily on the first chunk.
write_streamed_null - marks the current param SQL NULL instead of streaming chunks (the SQLPutData(SQL_NULL_DATA) case).
end_streamed_param - closes the value (PLP_NULL, terminator, or opener+terminator for an empty value), then opens the next streamed param (NeedData) or finalizes and positions on the response (Done).
RpcParameter::data_at_exec() marker; PacketWriter::suspend/resume to park a half-written message across calls; SqlRpc::serialize_prefix to serialize up to materialized params without finalizing.

Testing
Unit (offline mock): header serialization, non-MAX/encrypted rejection, single/multi-chunk framing, empty-chunk skip, two-param lifecycle, materialized+streamed mix, multi-packet chunk, streamed-NULL framing + ordering guards, usage-error guards, and two fault-injection tests for the mid-stream abort.

E2E (live SQL Server): round-trips for nvarchar/varbinary/varchar(max), two params, mixed, multi-row, materialized NULL, streamed NULL, empty value, many small chunks, and connection reuse.

Follow-up:
Currently streamed params are expected to appear after all materialized params in the parameter list. msodbcsql does not enforce this ordering. Supporting arbitrary interleaving of materialized and streamed params is deferred to a follow-up.

Not covered (deferred)
Always Encrypted on the streamed path (rejected); ODBC SQLPutData / SQLParamData (the next layer).

## Related Issues
https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46383/
https://sqlclientdrivers.visualstudio.com/mssql-rs/_backlogs/backlog/mssql-rs%20Team/Features?workitem=46139

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


mssql-tds/src/connection/tds_client.rs
```diff
@@ -928,32 +971,537 @@ impl TdsClient {
...
+            .expect("streamed parameter names validated above");
 
         let mut packet_writer =
             rpc.create_packet_writer(self.transport.as_writer(), timeout_sec, cancel_handle);
-        rpc.serialize(&mut packet_writer).await?;
+        // Write the RPC prefix (headers, proc, positional + materialized named
+        // params) then the first streamed parameter's header. The data-at-exec
+        // branch of `serialize` writes name + status + TYPE_INFO + PLP_UNKNOWN_LEN
+        // and stops at the value — the same method that serialized the
+        // materialized params, parked partway through.
```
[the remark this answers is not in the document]
shiwanigupta0809: done

mssql-tds/src/connection/tds_client.rs
```diff
@@ -928,32 +971,537 @@ impl TdsClient {
...
+    /// [`end_streamed_param`](Self::end_streamed_param) fails cleanly (no active
+    /// stream) instead of appending to a corrupt message, and the connection is
+    /// marked for reset so the desynced wire stream cannot leak into the next
+    /// command. Mirrors msodbcsql, which tears down data-at-execution state on a
+    /// failed DAE send (`FlushStmt` / `ClearNonBLOBDAEParam`) rather than leaving
+    /// the value resumable.
+    fn abort_streamed_write(&mut self) {
+        self.streamed_write_state = StreamedWriteState::Idle;
+        self.execution_context.set_has_open_batch(false);
+        self.prepare_reset_connection(false);
```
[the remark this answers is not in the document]
shiwanigupta0809: Acknowledged. Currently [abort_streamed_write()] closes the transport—safe but reconnects on next use. A TODO exists to implement state-aware cancellation (send EOM | IGNORE after partial send, or discard locally if unsent)

mssql-tds/src/connection/tds_client.rs
```diff
@@ -934,32 +977,539 @@ impl TdsClient {
...
+        }
+        if chunk.len() > u32::MAX as usize {
+            return Err(UsageError(format!(
+                "Streamed PLP chunk length {} exceeds the maximum chunk size of {} bytes.",
+                chunk.len(),
+                u32::MAX
+            )));
+        }
+
+        let ctx = match std::mem::replace(&mut self.streamed_write_state, StreamedWriteState::Idle)
```
Theekshna, line 1374: `write_streamed_chunk(&[])`  returns success even when no stream is active because the empty check runs before state validation. Validate  Active  first, then treat empty data as a no-op.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -934,32 +977,539 @@ impl TdsClient {
...
+            .prepare_sp_executesql_command(timeout_sec, cancel_handle)
+            .await?;
+        let rpc = self.build_sp_executesql_rpc(
+            sql,
+            materialized_params,
+            params_list_as_string,
+            &database_collation,
+        );
+
+        self.start_sp_executesql_streamed(
```
Theekshna, line 1208: An active streamed write is tracked in `streamed_write_state`, but normal command entry points generally guard only on `has_open_batch()`. Because beginning a streamed write does not set that flag, a caller can begin streaming, execute another command, and then resume the suspended RPC.

Could the streamed lifecycle participate in the common command-idle guard, or set/clear `has_open_batch` appropriately? Please also add a regression test confirming that an intervening command returns `ALREADY_EXECUTING` without damaging the active stream.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -934,32 +977,539 @@ impl TdsClient {
...
+    /// types. Streaming is not supported when Always Encrypted is active.
+    ///
+    /// # Errors
+    /// Returns a usage error for invalid streamed parameters or an active
+    /// command, and returns transport/serialization errors from the request.
+    pub async fn begin_sp_executesql(
+        &mut self,
+        sql: String,
+        named_params: Vec<RpcParameter>,
+        timeout_sec: Option<u32>,
```
Theekshna: **This differs from msodbcsql's DAE timeout behavior.** The reference driver obtains a fresh `GetQueryTimeOut(lpstmt)` for each `SQLPutData` and `SQLParamData` invocation (`sqlccmd.cpp`: `PutParamData`, `ProcessDAEParam`, and `SQLParamData`). Application think-time between chunks therefore does not consume one operation-wide deadline.

Here, preserving `state.start_time` across every suspend/resume means time spent by the application producing data counts against the timeout. A slow producer can time out and lose the connection before its next chunk is written.

Should `resume` retain the configured timeout but reset the clock, e.g. `start_time: Instant::now()`? That would bound each resumed I/O operation and match msodbcsql's behavior.

mssql-tds/src/message/parameters/rpc_parameters.rs
```diff
@@ -124,9 +134,28 @@ impl RpcParameter {
...
     }
 
+    /// Marks this parameter as data-at-execution: its value is streamed later in
+    /// chunks rather than materialized up front. The `value` supplied to
+    /// [`new`](Self::new) is used only as a type template (a `None`-valued MAX
+    /// type such as `SqlType::NVarcharMax(None)` is the intended form).
+    ///
+    /// Only `nvarchar(max)`, `varchar(max)` and `varbinary(max)` may be streamed;
+    /// see [`is_streamable_plp`](Self::is_streamable_plp).
+    pub fn data_at_exec(mut self) -> Self {
```
Theekshna: `SqlType::VarBinaryMax(None)` normally represents SQL `NULL`, but after `.data_at_exec()` the same value becomes only a type template whose data is supplied later. This creates an ambiguous internal state: `None` no longer means NULL, and any materialized value would be silently ignored.

Could we represent streamed parameters explicitly instead?

```rust
pub enum StreamedSqlType {
    NVarcharMax,
    VarcharMax,
    VarBinaryMax,
}

enum RpcParameterValue {
    Materialized(SqlType),
    Streamed(StreamedSqlType),
}

impl RpcParameter {
    pub fn data_at_exec(
        name: impl Into<String>,
        options: StatusFlags,
        sql_type: StreamedSqlType,
    ) -> Self;
}
```

Usage would be unambiguous:

```rust
RpcParameter::data_at_exec(
    "@value",
    StatusFlags::NONE,
    StreamedSqlType::VarBinaryMax,
)
```

This makes materialized, NULL, and streamed states structurally distinct, prevents unsupported streaming types from being constructed.

`StreamedSqlType` only needs to represent the incremental streaming supported by `mssql-tds`. The ODBC layer can still implement broader DAE behavior by buffering and materializing non-streamable or fixed-length values before passing them to the TDS layer.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: done

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: done

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: done. Added command_is_busy() which is has_open_batch() || !matches!(streamed_write_state, Idle), and converted all command-entry guards in tds_client.rs to use it.

mssql-tds/src/message/parameters/rpc_parameters.rs
```diff
@@ -333,7 +430,12 @@ impl RpcParameter {
     /// Returns the parameter's plaintext value. Used by the parameter-encryption
     /// path to normalize and encrypt the value before sending.
     pub(crate) fn value(&self) -> &SqlType {
-        &self.value
+        match &self.value {
+            RpcValue::Materialized(value) => value,
+            RpcValue::Streamed(_) => unreachable!(
```
saurabh500: **Blocking.** This `unreachable!` is reachable from safe public API, so it's a panic, not an invariant.

`RpcParameter::data_at_exec` is `pub`, but only `execute_sp_executesql` rejects data-at-exec params. Nothing stops this:

```rust
// AE-enabled connection
client.execute_stored_procedure(
    "dbo.p".into(),
    None,
    Some(vec![RpcParameter::data_at_exec(
        Some("@v".into()), StatusFlags::NONE, StreamedSqlType::VarBinaryMax)]),
    opts,
).await
```

`execute_stored_procedure` (tds_client.rs:1928) -> `build_stored_procedure_describe_request` -> `RpcParameter::get_sql_name(param.value())` (tds_client.rs:3995) -> this arm -> panic. Same exposure from the `prepare`/`sp_execute` paths and from `#[cfg(fuzzing)] get_value()`.

Without Always Encrypted it's quieter but worse: `serialize` writes the parameter header and no value body, so we put a malformed RPC on the wire and desync the connection.

Suggest making this fallible rather than panicking:

```rust
pub(crate) fn value(&self) -> TdsResult<&SqlType> {
    match &self.value {
        RpcValue::Materialized(value) => Ok(value),
        RpcValue::Streamed(_) => Err(Error::UsageError(
            "Data-at-execution parameters are only supported via begin_sp_executesql.".to_string(),
        )),
    }
}
```

plus the same `is_data_at_exec` guard on the other public entry points (see my comment on tds_client.rs:949).

mssql-tds/src/connection/tds_client.rs
```diff
@@ -886,39 +943,27 @@ impl TdsClient {
         } = options.into();
         self.current_command_ce_setting = column_encryption;
 
-        if self.execution_context.has_open_batch() {
+        if self.command_is_busy() {
             return Err(UsageError(ALREADY_EXECUTING_ERROR.to_string()));
         };
+        if named_params.iter().any(RpcParameter::is_data_at_exec) {
```
saurabh500: **Blocking.** This guard is right, but it's the only one. `execute_stored_procedure`, `prepare_statement`, the `sp_execute`/`sp_prepexec` paths and `execute_rpc` all take `Vec<RpcParameter>` and none of them reject a `data_at_exec` param. Since `RpcParameter::data_at_exec` is `pub`, those paths either panic (Always Encrypted, via `RpcParameter::value()`) or serialize a parameter header with no value body and desync the connection.

Please hoist this into a shared helper and call it from every public entry point that accepts `RpcParameter`s:

```rust
fn reject_data_at_exec<'p>(params: impl IntoIterator<Item = &'p RpcParameter>) -> TdsResult<()> {
    if params.into_iter().any(RpcParameter::is_data_at_exec) {
        return Err(UsageError(
            "Data-at-execution parameters require begin_sp_executesql.".to_string(),
        ));
    }
    Ok(())
}
```

A test asserting `UsageError` (not a panic) for a streamed param passed to `execute_stored_procedure` would lock this down.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -934,32 +979,554 @@ impl TdsClient {
...
+
+    fn split_and_validate_streamed_params(
+        named_params: Vec<RpcParameter>,
+    ) -> TdsResult<(Vec<RpcParameter>, Vec<RpcParameter>)> {
+        let (streamed_params, materialized_params): (Vec<_>, Vec<_>) = named_params
+            .into_iter()
+            .partition(RpcParameter::is_data_at_exec);
+
+        for param in &streamed_params {
+            if !param.is_streamable_plp() {
```
saurabh500: This branch can never fire. `streamed_params` comes from `.partition(RpcParameter::is_data_at_exec)` four lines up, and `is_streamable_plp` has the exact same body as `is_data_at_exec` (`matches!(self.value, RpcValue::Streamed(_))`), so it is always `true` here. The "must be nvarchar(max), varchar(max) or varbinary(max)" error is unreachable.

That is fine as a type-system outcome -- `StreamedSqlType` already constrains it -- but the dead check reads like real validation. I'd drop it and keep only the name check. The test `begin_only_accepts_streamable_types_via_data_at_exec_constructor` (line 9086) asserts the same tautology and should go with it.

mssql-tds/src/message/parameters/rpc_parameters.rs
```diff
@@ -309,12 +396,22 @@ impl RpcParameter {
...
         encoder
-            .encode_sqlvalue(packet_writer, &self.value, db_collation)
+            .encode_sqlvalue(packet_writer, value, db_collation)
             .await?;
         Ok(())
     }
 
+    /// Returns `true` when this parameter's declared type is a PLP type
+    /// eligible for incremental streaming (see [`StreamedSqlType`]).
+    pub(crate) fn is_streamable_plp(&self) -> bool {
```
saurabh500: `is_streamable_plp` and `is_data_at_exec` (line 191) have identical bodies but doc comments implying different questions ("is it streamed?" vs "is its type streamable?"). Two names for one predicate will drift the moment someone adds a non-MAX streamed variant. Drop one -- `is_data_at_exec` reads better at the call sites.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -934,32 +979,554 @@ impl TdsClient {
...
+    /// parameter was already marked NULL via
+    /// [`write_streamed_null`](Self::write_streamed_null), or if `chunk` is longer
+    /// than [`u32::MAX`] bytes (the PLP chunk-length field is 32-bit).
+    pub async fn write_streamed_chunk(&mut self, chunk: &[u8]) -> TdsResult<()> {
+        if matches!(self.streamed_write_state, StreamedWriteState::Idle) {
+            return Err(UsageError(
+                "write_streamed_chunk called with no active streamed parameter.".to_string(),
+            ));
+        }
+        if chunk.is_empty() {
```
saurabh500: The empty-chunk short circuit runs before the `null_signaled` check below, so the guard is inconsistent: after `write_streamed_null()`, `write_streamed_chunk(&[0x01])` is a `UsageError` but `write_streamed_chunk(&[])` silently returns `Ok(())`. A caller feeding a zero-length buffer gets no signal it is sequencing the API wrong.

Move the empty check below the state extraction and `null_signaled` validation, or check `null_signaled` up front alongside the `Idle` check.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -934,32 +979,554 @@ impl TdsClient {
...
+        };
+        let StreamedWriteContext {
+            message,
+            pending,
+            db_collation,
+            value_opened,
+            null_signaled,
+            timeout_sec,
+        } = *ctx;
+        self.remaining_request_timeout = timeout_sec.map(|s| Duration::from_secs(u64::from(s)));
```
saurabh500: This resets the client budget to the *full* configured timeout on every chunk, but `PacketWriter::suspend`/`resume` preserves `start_time`, and `populate_header_and_send` checks `start_time.elapsed() > max_timeout_sec` after each flush. So the two halves of the same timeout disagree:

- read side: refreshed per chunk, effectively unbounded across a long stream
- write side: a hard wall-clock deadline measured from `begin_sp_executesql`

For the intended consumer (ODBC `SQLPutData`, where the app feeds data at its own pace) a slow producer trips the writer deadline mid-value, which lands in `abort_streamed_write` and closes the connection. That is a rough failure mode for something that is not a server-side stall.

Either refresh the writer deadline on resume too, or document that `timeout_sec` on `begin_sp_executesql` caps the whole begin->end wall clock rather than server execution. `write_streamed_chunk_resets_timeout_to_configured_budget` pins the refreshing half but does not cover the interaction.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -934,32 +979,554 @@ impl TdsClient {
...
+
+    /// Aborts an in-progress streamed PLP write after a mid-stream failure.
+    ///
+    /// The half-written RPC message is dropped rather than re-parked, so a
+    /// subsequent [`write_streamed_chunk`](Self::write_streamed_chunk) or
+    /// [`end_streamed_param`](Self::end_streamed_param) fails cleanly (no active
+    /// stream) instead of appending to a corrupt message. The transport is closed
+    /// because RESETCONNECTION applies only to a subsequent request and cannot
+    /// terminate this incomplete RPC.
+    async fn abort_streamed_write(&mut self) {
```
saurabh500: `abort_streamed_write` is private, so once a caller is in `NeedData` there is no supported way to bail out. Every other command returns `ALREADY_EXECUTING_ERROR` via `command_is_busy()`, and the only exits are `end_streamed_param` (which commits a value the caller may no longer want) or dropping the client.

ODBC's DAE flow explicitly allows `SQLCancel` between `SQLParamData`/`SQLPutData`, so the layer above this will need it. Worth exposing `pub async fn cancel_streamed_write(&mut self)` now, even if it just wraps this and keeps the current close-the-transport behavior until the TODO below is implemented.

mssql-tds/src/message/parameters/rpc_parameters.rs
```diff
@@ -81,6 +81,50 @@ pub(crate) struct EncryptedRpcValue {
...
+    /// Unicode MAX text.
+    NVarcharMax,
+    /// Single-byte MAX text.
+    VarcharMax,
+    /// MAX binary data.
+    VarBinaryMax,
+}
+
+impl StreamedSqlType {
+    fn sql_name(self) -> &'static str {
```
saurabh500: Nit: these strings duplicate what `RpcParameter::get_sql_name` already returns for the same `SqlType`s. Since `as_sql_type()` is right below, `build_parameter_list_string_impl` could call `RpcParameter::get_sql_name(&streamed.as_sql_type())?` and this method goes away -- one fewer place for the `@params` declaration name and the TYPE_INFO to drift apart.

mssql-tds/src/message/parameters/rpc_parameters.rs
saurabh500: **Blocking.** This `unreachable!` is reachable from safe public API, so it's a panic, not an invariant.

`RpcParameter::data_at_exec` is `pub`, but only `execute_sp_executesql` rejects data-at-exec params. Nothing stops this:

```rust
// AE-enabled connection
client.execute_stored_procedure(
    "dbo.p".into(),
    None,
    Some(vec![RpcParameter::data_at_exec(
        Some("@v".into()), StatusFlags::NONE, StreamedSqlType::VarBinaryMax)]),
    opts,
).await
```

`execute_stored_procedure` (tds_client.rs:1928) -> `build_stored_procedure_describe_request` -> `RpcParameter::get_sql_name(param.value())` (tds_client.rs:3995) -> this arm -> panic. Same exposure from the `prepare`/`sp_execute` paths and from `#[cfg(fuzzing)] get_value()`.

Without Always Encrypted it's quieter but worse: `serialize` writes the parameter header and no value body, so we put a malformed RPC on the wire and desync the connection.

Suggest making this fallible rather than panicking:

```rust
pub(crate) fn value(&self) -> TdsResult<&SqlType> {
    match &self.value {
        RpcValue::Materialized(value) => Ok(value),
        RpcValue::Streamed(_) => Err(Error::UsageError(
            "Data-at-execution parameters are only supported via begin_sp_executesql.".to_string(),
        )),
    }
}
```

plus the same `is_data_at_exec` guard on the other public entry points (see my comment on tds_client.rs:949).

mssql-tds/src/connection/tds_client.rs
saurabh500: **Blocking.** This guard is right, but it's the only one. `execute_stored_procedure`, `prepare_statement`, the `sp_execute`/`sp_prepexec` paths and `execute_rpc` all take `Vec<RpcParameter>` and none of them reject a `data_at_exec` param. Since `RpcParameter::data_at_exec` is `pub`, those paths either panic (Always Encrypted, via `RpcParameter::value()`) or serialize a parameter header with no value body and desync the connection.

Please hoist this into a shared helper and call it from every public entry point that accepts `RpcParameter`s:

```rust
fn reject_data_at_exec<'p>(params: impl IntoIterator<Item = &'p RpcParameter>) -> TdsResult<()> {
    if params.into_iter().any(RpcParameter::is_data_at_exec) {
        return Err(UsageError(
            "Data-at-execution parameters require begin_sp_executesql.".to_string(),
        ));
    }
    Ok(())
}
```

A test asserting `UsageError` (not a panic) for a streamed param passed to `execute_stored_procedure` would lock this down.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -934,32 +979,554 @@ impl TdsClient {
...
+    /// [`StreamedParamStatus::Complete`].
+    ///
+    /// Streamed parameters must be named and must use one of the
+    /// [`StreamedSqlType`] variants. Streaming is not supported when Always
+    /// Encrypted is active.
+    ///
+    /// # Errors
+    /// Returns a usage error for invalid streamed parameters or an active
+    /// command, and returns transport/serialization errors from the request.
+    pub async fn begin_sp_executesql(
```
saurabh500: This should take `options: impl Into<ExecuteOptions<'a>>` like every sibling entry point (`execute`, `execute_sp_executesql`, `execute_stored_procedure`, `prepare_statement`, `execute_prepared`, `unprepare`). The `ExecuteOptions` doc comment states the convention outright:

> New per-command capabilities are added here as new defaulted fields -- never as new methods or changed signatures -- keeping the `execute*` surface forward-compatible.

Taking `timeout_sec`/`cancel_handle` positionally means the next per-command knob added to `ExecuteOptions` reaches every other entry point for free but requires a signature change here -- a breaking change to a brand-new public API.

More concretely, hardcoding `current_command_ce_setting = UseConnectionSetting` at line 1035 removes the caller's only escape hatch. `should_encrypt_parameters()` is `is_column_encryption_supported() && effective_command_ce_setting() == Enabled`, so on an AE-enabled connection `effective` always resolves to `Enabled` and the check at line 1057 rejects **every** streamed write on that connection -- including statements whose parameters touch no encrypted column. That is exactly what `ExecutionColumnEncryptionSetting::ResultSetOnly` documents itself for ("a command reads encrypted columns but its parameters do not target any encrypted column"), and `Disabled` likewise.

No new rejection logic is needed -- just feed the option through and the existing check does the right thing:

```rust
pub async fn begin_sp_executesql<'a>(
    &mut self,
    sql: String,
    named_params: Vec<RpcParameter>,
    options: impl Into<ExecuteOptions<'a>>,
) -> TdsResult<StreamedParamStatus> {
    let ExecuteOptions { timeout, cancel, column_encryption } = options.into();
    if self.command_is_busy() { /* ... */ }
    self.current_command_ce_setting = column_encryption;
    // ...unchanged from here
```

The default still produces the current clear error on an AE connection; an explicit opt-out now works.

It also cleans up the no-streamed-params delegation just below, which currently rebuilds the struct and hardcodes the field, silently discarding the caller's AE intent on a path that forwards to a method which fully supports it:

```diff
 let result = self.execute_sp_executesql(
     sql, materialized_params,
-    ExecuteOptions {
-        timeout: timeout_sec,
-        cancel: cancel_handle,
-        column_encryption: ExecutionColumnEncryptionSetting::UseConnectionSetting,
-    },
+    ExecuteOptions { timeout, cancel, column_encryption },
 ).await?;
```

mssql-tds/src/connection/tds_client.rs
saurabh500: This branch can never fire. `streamed_params` comes from `.partition(RpcParameter::is_data_at_exec)` four lines up, and `is_streamable_plp` has the exact same body as `is_data_at_exec` (`matches!(self.value, RpcValue::Streamed(_))`), so it is always `true` here. The "must be nvarchar(max), varchar(max) or varbinary(max)" error is unreachable.

That is fine as a type-system outcome -- `StreamedSqlType` already constrains it -- but the dead check reads like real validation. I'd drop it and keep only the name check. The test `begin_only_accepts_streamable_types_via_data_at_exec_constructor` (line 9086) asserts the same tautology and should go with it.

mssql-tds/src/message/parameters/rpc_parameters.rs
saurabh500: `is_streamable_plp` and `is_data_at_exec` (line 191) have identical bodies but doc comments implying different questions ("is it streamed?" vs "is its type streamable?"). Two names for one predicate will drift the moment someone adds a non-MAX streamed variant. Drop one -- `is_data_at_exec` reads better at the call sites.

mssql-tds/src/connection/tds_client.rs
saurabh500: The empty-chunk short circuit runs before the `null_signaled` check below, so the guard is inconsistent: after `write_streamed_null()`, `write_streamed_chunk(&[0x01])` is a `UsageError` but `write_streamed_chunk(&[])` silently returns `Ok(())`. A caller feeding a zero-length buffer gets no signal it is sequencing the API wrong.

Move the empty check below the state extraction and `null_signaled` validation, or check `null_signaled` up front alongside the `Idle` check.

mssql-tds/src/connection/tds_client.rs
saurabh500, line 1391: This resets the client budget to the *full* configured timeout on every chunk, but `PacketWriter::suspend`/`resume` preserves `start_time`, and `populate_header_and_send` checks `start_time.elapsed() > max_timeout_sec` after each flush. So the two halves of the same timeout disagree:

- read side: refreshed per chunk, effectively unbounded across a long stream
- write side: a hard wall-clock deadline measured from `begin_sp_executesql`

For the intended consumer (ODBC `SQLPutData`, where the app feeds data at its own pace) a slow producer trips the writer deadline mid-value, which lands in `abort_streamed_write` and closes the connection. That is a rough failure mode for something that is not a server-side stall.

Either refresh the writer deadline on resume too, or document that `timeout_sec` on `begin_sp_executesql` caps the whole begin->end wall clock rather than server execution. `write_streamed_chunk_resets_timeout_to_configured_budget` pins the refreshing half but does not cover the interaction.

mssql-tds/src/connection/tds_client.rs
saurabh500, line 1479: `abort_streamed_write` is private, so once a caller is in `NeedData` there is no supported way to bail out. Every other command returns `ALREADY_EXECUTING_ERROR` via `command_is_busy()`, and the only exits are `end_streamed_param` (which commits a value the caller may no longer want) or dropping the client.

ODBC's DAE flow explicitly allows `SQLCancel` between `SQLParamData`/`SQLPutData`, so the layer above this will need it. Worth exposing `pub async fn cancel_streamed_write(&mut self)` now, even if it just wraps this and keeps the current close-the-transport behavior until the TODO below is implemented.

mssql-tds/src/message/parameters/rpc_parameters.rs
saurabh500: Nit: these strings duplicate what `RpcParameter::get_sql_name` already returns for the same `SqlType`s. Since `as_sql_type()` is right below, `build_parameter_list_string_impl` could call `RpcParameter::get_sql_name(&streamed.as_sql_type())?` and this method goes away -- one fewer place for the `@params` declaration name and the TYPE_INFO to drift apart.

saurabh500 (commented):
Nice piece of work — the suspend/resume primitive is a clean way to park a half-written RPC, and the lazy PLP opener (deferring the length field until the first chunk, so a parameter can still resolve to NULL after its header is on the wire) is the right call. The take-and-re-park state machine consistently distinguishes caller sequencing errors (re-park a clean message, stream stays usable) from wire failures (drop and abort), and the byte-level tests cover framing well.

One blocking issue, plus a few things worth resolving before this goes in.

## Blocking

**`RpcParameter::value()` can panic from safe public API.** `RpcParameter::data_at_exec` is `pub`, but only `execute_sp_executesql` rejects data-at-exec params. This PR replaces `has_open_batch()` with `command_is_busy()` in ~10 entry points — `execute`, `execute_stored_procedure`, `prepare_statement`, `sp_execute`, `sp_prepexec`, `execute_rpc`, bulk load — and that list is effectively the set of paths that also need the `is_data_at_exec` guard. None of them got it.

With Always Encrypted active, `execute_stored_procedure` → `build_stored_procedure_describe_request` → `get_sql_name(param.value())` hits the new `unreachable!`. Without AE it is quieter but worse: `serialize` emits the parameter header with no value body and desyncs the connection. This PR also routes two more call sites into that accessor (`write_encrypted_type_info`, and `#[cfg(fuzzing)] get_value()` — a fuzz run would find it immediately).

Worth noting the hazard was already recognized: `build_parameter_list_string` matches on `RpcValue` directly rather than calling `value()`. Making `value()` return `TdsResult<&SqlType>` would let the compiler enumerate the remaining sites instead of relying on spotting them.

## Suggestions

- **`begin_sp_executesql` should take `ExecuteOptions`** like every sibling. Beyond consistency, hardcoding `current_command_ce_setting = UseConnectionSetting` makes streaming unconditionally unavailable on any AE-enabled connection, even when no parameter targets an encrypted column — which is exactly what `ResultSetOnly`/`Disabled` exist for.
- **Timeout semantics disagree across suspend/resume.** `remaining_request_timeout` resets to the full budget on every chunk, while `PacketWriter::start_time` is deliberately preserved. A slow producer — the whole point of a `SQLPutData`-style API — trips the send deadline mid-value and lands in `abort_streamed_write`, which closes the connection.
- **No public way to abandon an in-flight streamed write.** Once `NeedData` is returned, every other command returns `ALREADY_EXECUTING_ERROR`; the only exits are `end_streamed_param` (commits a value the caller may not want) or dropping the client. ODBC's DAE flow allows `SQLCancel` here.
- **Dead validation** in `split_and_validate_streamed_params`, and a duplicate predicate (`is_streamable_plp` is byte-identical to `is_data_at_exec`).
- **Empty-chunk short circuit** sits above the `null_signaled` guard, so `write_streamed_chunk(&[])` after `write_streamed_null()` succeeds while a non-empty chunk errors.

## Nits

- `StreamedSqlType::sql_name()` duplicates strings `get_sql_name` already produces.
- `write_streamed_chunk` checks `Idle` twice — once up front, then again as an unreachable match arm.

## Housekeeping

This adds 7 public items (`begin_sp_executesql`, `write_streamed_chunk`, `write_streamed_null`, `end_streamed_param`, `StreamedParamStatus`, `StreamedSqlType`, `RpcParameter::data_at_exec`) with no `CHANGELOG.md` entry — the `[Unreleased] / ### Added` section is where new `mssql-tds` public API gets recorded. The checklist boxes for "New/changed functionality has tests" and "Public API changes are documented" are also unchecked; the tests clearly exist, so that one is just stale.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: Done - signature is now:

```rust
pub async fn begin_sp_executesql<'a>(
    &mut self,
    sql: String,
    named_params: Vec<RpcParameter>,
    options: impl Into<ExecuteOptions<'a>>,
) -> TdsResult<StreamedParamStatus> {
    let ExecuteOptions { timeout: timeout_sec, cancel: cancel_handle, column_encryption }
        = options.into();
```

This also fixed a latent bug you implicitly caught: `column_encryption` was hardcoded to `UseConnectionSetting` both on the streamed path and in the no-streamed-params delegation to `execute_sp_executesql`. It is now threaded through from the caller's options. All call sites updated (`()` for the common default case, `ExecuteOptions::new().timeout_secs(n)` / `.cancel(h)` otherwise) across the unit tests and `tests/test_client_write_apis.rs`.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: Added a public entry point:

```rust
/// Cancels an in-progress streamed PLP write while the client is parked in
/// `StreamedParamStatus::NeedData`, discarding the half-written RPC and
/// closing the transport. ... Calling this with no streamed write active is a no-op.
pub async fn cancel_streamed_write(&mut self) {
    if matches!(self.streamed_write_state, StreamedWriteState::Idle) {
        return;
    }
    self.abort_streamed_write().await;
}
```

This is the supported way to bail out mid-stream (e.g. servicing `SQLCancel` between `SQLPutData` calls). After it returns, no streamed write is active, so a subsequent `write_streamed_chunk` / `write_streamed_null` / `end_streamed_param` fails with a usage error.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: Added a shared guard and called it from every public entry point that accepts `RpcParameter`s:

```rust
pub(crate) fn reject_data_at_exec<'p>(
    params: impl IntoIterator<Item = &'p RpcParameter>,
) -> TdsResult<()> {
    if params.into_iter().any(RpcParameter::is_data_at_exec) {
        return Err(Error::UsageError(
            "Data-at-execution parameters require begin_sp_executesql.".to_string(),
        ));
    }
    Ok(())
}
```

Call sites:
- `execute_sp_executesql` - replaced the inline check with this helper.
- `execute_stored_procedure` - `reject_data_at_exec(positional.iter().flatten().chain(named.iter().flatten()))?`.
- `execute_prepared` - `reject_data_at_exec(named_params.iter())?`.

`execute_sp_prepare` / `execute_sp_prepexec` / `execute_sp_execute` are private and only reachable through `execute_prepared` (or the `#[cfg(test)]` `_for_test` wrappers that intentionally bypass guards to exercise the wire protocol), so the single guard on `execute_prepared` covers them. Added `reject_data_at_exec_rejects_only_streamed_params`.

mssql-tds/src/message/parameters/rpc_parameters.rs
shiwanigupta0809: Agreed - now delegates instead of duplicating the strings:

```rust
/// Declaration name for the `sp_executesql` `@params` string. Delegates to
/// [`RpcParameter::get_sql_name_impl`] on the equivalent materialized
/// [`SqlType`] rather than duplicating the `nvarchar(MAX)` / `varchar(MAX)`
/// / `varbinary(MAX)` strings, so the two can't drift apart.
fn sql_name(self) -> TdsResult<String> {
    RpcParameter::get_sql_name_impl(&self.as_sql_type())
}
```

`get_sql_name_impl` is private but same-module, so no visibility change was needed. Call site is now `RpcValue::Streamed(streamed) => streamed.sql_name()?`.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: Fixed in `PacketWriter::resume` - it now takes a fresh timestamp rather than restoring the suspended one:

```rust
-            start_time: state.start_time,
+            start_time: Instant::now(),
```

Rationale (now in the `resume` doc comment): each call that resumes a message is a fresh write operation with its own budget for the hard wall-clock write timeout, mirroring msodbcsql's per-`SQLPutData` timeout refresh - application think-time between chunks must not count against a single deadline set when the message was first opened.

After that change the compiler flagged `SuspendedMessage.start_time` as never read, so the field was removed entirely and the `suspend` / `SuspendedMessage` doc comments updated to explain why it is deliberately not preserved.

mssql-tds/src/message/parameters/rpc_parameters.rs
shiwanigupta0809: Made it fallible instead of panicking - `value()` now returns `TdsResult<&SqlType>` and returns `Error::UsageError` for a streamed parameter.

```rust
pub(crate) fn value(&self) -> TdsResult<&SqlType> {
    match &self.value {
        RpcValue::Materialized(value) => Ok(value),
        RpcValue::Streamed(_) => Err(Error::UsageError(
            "Data-at-execution parameters are only supported via begin_sp_executesql.".to_string(),
        )),
    }
}
```

All callers updated to propagate with `?`: `serialize()`, the parameter-encryption path (`encrypt_parameter(param.value()?, ...)`), both `get_sql_name` sites in `build_stored_procedure_describe_request`, and the `#[cfg(fuzzing)] get_value()` accessor. Added `value_on_streamed_param_returns_usage_error`.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: Good catch - reordered so the `null_signaled` check runs first, and both early returns now correctly re-park the message state:

```rust
// ...checked before the empty-chunk short-circuit below so
// `write_streamed_chunk(&[])` after `write_streamed_null()` still errors
// instead of silently succeeding.
if null_signaled {
    self.streamed_write_state = StreamedWriteState::Active(Box::new(StreamedWriteContext { ... }));
    return Err(UsageError(...));
}

// Empty chunks are ignored: a zero-length PLP chunk header is the value
// terminator, so it must never be emitted mid-value. Re-park the (unchanged)
// message so a later chunk or the terminator continues from the same point.
if chunk.is_empty() {
    self.streamed_write_state = StreamedWriteState::Active(Box::new(StreamedWriteContext { ... }));
    return Ok(());
}
```

Added `streamed_write_empty_chunk_after_null_errors` covering exactly this case.

mssql-tds/src/message/parameters/rpc_parameters.rs
shiwanigupta0809: Removed `is_streamable_plp()` entirely - `is_data_at_exec()` is now the single predicate, and its one remaining caller (the dead validation branch in `split_and_validate_streamed_params`) is gone too.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: Correct - removed. The `.partition(RpcParameter::is_data_at_exec)` above already guarantees every parameter in `streamed_params` is streamed, so the branch was unreachable. Only the `name.is_none()` validation remains:

```rust
for param in &streamed_params {
    if param.name.is_none() {
        return Err(UsageError("Streamed parameters must be named.".to_string()));
    }
}
```

Also dropped the now-tautological test `begin_only_accepts_streamable_types_via_data_at_exec_constructor`.

shiwanigupta0809:

shiwanigupta0809:
@saurabh500 PR ready for re-review

shiwanigupta0809:

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1063,32 +1104,589 @@ impl TdsClient {
...
+    ///
+    /// This is the supported way for a caller to bail out of a streamed
+    /// parameter sequence once [`begin_sp_executesql`](Self::begin_sp_executesql)
+    /// has parked a message (for example, to service `SQLCancel` while an ODBC
+    /// driver is between `SQLPutData` calls). After this returns, no streamed
+    /// write is active, so a subsequent [`write_streamed_chunk`](Self::write_streamed_chunk),
+    /// [`write_streamed_null`](Self::write_streamed_null), or
+    /// [`end_streamed_param`](Self::end_streamed_param) fails with a usage
+    /// error. Calling this with no streamed write active is a no-op.
+    pub async fn cancel_streamed_write(&mut self) {
```
saurabh500: New public method with no test. CI enforces 85% diff coverage and this closes the transport, so it deserves direct coverage: begin a streamed write against the mock server, call `cancel_streamed_write()` at `NeedData`, then assert a following `write_streamed_chunk`/`end_streamed_param` returns `UsageError` (and that a fresh command on the connection fails). This also pins the `mark_known_dead()` behavior added in 689463c0c.

mssql-tds/src/message/parameters/rpc_parameters.rs
```diff
@@ -981,6 +1106,157 @@ mod tests {
...
+            StreamedSqlType::VarBinaryMax,
+        );
+        assert!(matches!(param.value(), Err(Error::UsageError(_))));
+    }
+
+    /// `reject_data_at_exec` is the shared guard every non-streaming entry
+    /// point uses to reject a streamed parameter before it can reach
+    /// [`RpcParameter::value`] or an incomplete serialization.
+    #[test]
+    fn reject_data_at_exec_rejects_only_streamed_params() {
```
saurabh500, line 1245: **Suggestion — pin the guard *wiring*, not just the helper**

This pins `reject_data_at_exec` in isolation, which is the right unit test to have. But nothing asserts that the entry points actually *call* it — and that wiring is precisely the blocking bug from the last round (`execute_sp_executesql` had the guard; `execute_stored_procedure` and `execute_prepared` did not).

That's the part that regresses silently: add a fourth entry point, or drop a `?`, and every test in this module still passes.

Suggest one test per public path, in `tds_client.rs`'s `mod tests`:

```rust
#[tokio::test]
async fn execute_stored_procedure_rejects_streamed_params() {
    let mut client = /* mock-backed client, as in the sibling tests */;
    let streamed = RpcParameter::data_at_exec(
        Some("@v".to_string()),
        StatusFlags::NONE,
        StreamedSqlType::VarBinaryMax,
    );

    let err = client
        .execute_stored_procedure("dbo.p".to_string(), None, Some(vec![streamed]), ())
        .await
        .expect_err("streamed params must be rejected outside the streaming path");

    assert!(matches!(err, Error::UsageError(_)));
}
```

…plus the equivalents for `execute_prepared` and `execute_sp_executesql`. Three cheap tests that fail loudly the day a new entry point forgets the guard.

mssql-tds/src/connection/tds_client.rs
saurabh500, line 1502: **Suggestion — `cancel_streamed_write` has no coverage**

`git grep cancel_streamed_write` returns only this definition. It's new public API with zero references anywhere in the repo, including `tests/test_client_write_apis.rs`.

Both arms are worth pinning, because the early return is load-bearing — cancelling on an idle client must *not* kill the connection:

```rust
#[tokio::test]
async fn cancel_streamed_write_on_idle_client_is_noop() -> TdsResult<()> {
    let mut client = /* connect, as in the sibling tests */;
    client.cancel_streamed_write().await;
    assert!(!client.is_connection_dead());
    Ok(())
}

#[tokio::test]
async fn cancel_streamed_write_discards_parked_message() -> TdsResult<()> {
    let mut client = /* connect */;
    client.begin_sp_executesql(sql, named_params, ()).await?;

    client.cancel_streamed_write().await;

    // No streamed write is active any more, so the next chunk write is a usage error.
    assert!(matches!(
        client.write_streamed_chunk(b"x").await,
        Err(Error::UsageError(_))
    ));
    Ok(())
}
```

`tests/test_client_write_apis.rs` already has the connection boilerplate to copy from.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1063,32 +1104,589 @@ impl TdsClient {
...
+    /// The half-written RPC message is dropped rather than re-parked, so a
+    /// subsequent [`write_streamed_chunk`](Self::write_streamed_chunk) or
+    /// [`end_streamed_param`](Self::end_streamed_param) fails cleanly (no active
+    /// stream) instead of appending to a corrupt message. The transport is closed
+    /// because RESETCONNECTION applies only to a subsequent request and cannot
+    /// terminate this incomplete RPC.
+    async fn abort_streamed_write(&mut self) {
+        self.streamed_write_state = StreamedWriteState::Idle;
+        self.execution_context.set_has_open_batch(false);
+        self.transport.mark_known_dead();
```
saurabh500, line 1482: **Suggestion — assert the abort actually marks the connection dead**

This line is the entire content of `689463c0`, and it's the subtle part of the abort path. The ordering is *correct*, and it matters: `close_transport()` only sets the dead flag after a successful `stream.shutdown()`, so on an already-broken socket it returns early — without this line the connection would still look alive to a pool checkout.

But nothing asserts that. Someone "tidying up" by folding this into `close_transport()`, or moving it below, reintroduces the bug silently.

There's a ready template in `tests/test_connection_pool_primitives.rs` (~L91) and in `fatal_error_token_marks_connection_known_dead` in this file:

```rust
client.begin_sp_executesql(sql, named_params, ()).await?;
client.cancel_streamed_write().await;

assert!(
    client.is_connection_dead(),
    "an aborted streamed write must not hand a live-looking connection back to the pool"
);
```

saurabh500 (commented):
Re-reviewed at `a1815fe1`.

All nine substantive findings from my earlier review are resolved — I verified each one against the code rather than taking the commit messages at face value. Two are worth calling out:

- The timeout fix is **better than what I suggested**. Rather than propagating `start_time` through `resume`, you deleted the field from `SuspendedMessage` outright so `resume` re-clocks from `Instant::now()`. That removes the possibility of the two clocks drifting at all, instead of just correcting the arithmetic.
- `mark_known_dead()` in `abort_streamed_write` is correctly ordered *before* `close_transport()`, which genuinely matters: `close_transport` only sets the dead flag after a successful `stream.shutdown()`, so aborting on an already-broken socket would return early and leave a dead connection looking alive to a pool checkout.

I also confirmed the main merge preserved every fix — the guard census still covers all three public entry points, and `value() -> TdsResult<&SqlType>` survived intact with its caller using `?`.

**No correctness, security, or API findings remain.** What's left is test coverage on the newly added surface — three gaps, detailed inline. The first is the one I'd actually push on: the data-at-exec guard was the blocking bug from round one, and the fix has no regression test at the entry points it was added to. The other two are new public API and a new behavioral line with no assertions at all, which also matters against the repo's 85% diff-coverage gate.

Happy to approve once these are covered — no further design changes needed from my side.

mssql-tds/src/message/parameters/rpc_parameters.rs
shiwanigupta0809: Addressed in e6591ff0. Added explicit entry-point coverage in `tds_client.rs` for `execute_sp_executesql`, `execute_stored_procedure`, and `execute_prepared` to assert streamed params are rejected with `UsageError`, so guard wiring regressions fail loudly.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: Addressed in e6591ff0. Added `cancel_streamed_write` coverage for both branches: idle cancel is a no-op (connection remains alive), and active streamed-write cancel aborts the parked message and subsequent chunk writes are rejected.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: Addressed in e6591ff0. Added assertion that an aborted streamed write marks the connection dead, protecting pool handoff semantics on the abort path.

saurabh500 (approved):

PR #208, opened 2026-08-10
Add first-class coroutine PyCoreAsyncCursor to mssql-py-core

saurabh500:
## Summary

Adds `PyCoreAsyncCursor`, a genuine `asyncio` coroutine cursor for `mssql-py-core`, built on `pyo3-async-runtimes` `future_into_py`. This is a **correctness/architecture** change (a real non-blocking coroutine cursor), **not** a performance win.

`execute` / `fetchone` / `fetchmany` / `fetchall` / `close` each return a Python awaitable. The TDS I/O is driven on the connection's own tokio runtime and the coroutine `.await`s the resulting join handle — there is **no `block_on` on the coroutine path**. The `!Send` client is checked out of the shared cell and the guard is dropped **before** `.await`, so the Python event loop stays free while a fetch is in flight.

## Strictly additive

- The default `PyCoreCursor` (`conn.cursor()`) and the opt-in sync `PyCoreSyncCursor` (`conn.sync_cursor()`) are **byte-identical** to their prior form — `cursor.rs` and `sync_cursor.rs` are unchanged.
- New: `conn.async_cursor()` → `PyCoreAsyncCursor`.
- The async path uses a self-contained `with_async_client` checkout helper in `pyclient.rs`; the shared client cell stays `std::sync::Mutex` (no retype needed).
- Doc-label scrub: fixes the stale "async cursor" comments now that a real coroutine cursor exists.

## Changes (py-core only)

- `src/async_cursor.rs` (new) — `PyCoreAsyncCursor` coroutine cursor.
- `src/pyclient.rs` — `with_async_client` async checkout helper + honest module doc.
- `src/connection.rs` — `Connection.async_cursor()`.
- `src/lib.rs` — register `PyCoreAsyncCursor`.
- `Cargo.toml` — add `pyo3-async-runtimes = { version = "0.29", features = ["tokio-runtime"] }`.
- Tests: mock-backed coroutine tests + a CI `WAITFOR`-based non-blocking proof.

## Non-blocking proof

- **Static:** zero `block_on` on the `async_cursor.rs` coroutine path (I/O is spawned on the connection runtime; the coroutine awaits the join handle).
- **Local (mock):** `test_async_cursor_yields_to_event_loop` — a concurrent ticker advances while a fetch is awaited (mock answers in µs, so this asserts the weaker "yields at least once").
- **CI (live server):** `test_async_cursor_integration.py` — a concurrent 10ms ticker racks up ticks during a server-side `WAITFOR DELAY '00:00:02'` inside the coroutine's `await`.

## Notes

- Perf-sanity (flag, not block): over the mock (µs-scale I/O) the coroutine path is ~3.7x the default `block_on` path — this is pure coroutine-plumbing overhead on a zero-latency backend, not representative of real network I/O where the overhead is negligible and concurrency dominates.
- `mssql-py-core` `Cargo.lock` is gitignored; the new dep resolves under the `--frozen` py-core clippy/test leg locally.

Part of the sans-I/O native stack - see cover PR #189 for the stack-level summary and tracked debt.

### DIFF
--- a/mssql-py-core/Cargo.toml
+++ b/mssql-py-core/Cargo.toml
@@ -10,6 +10,7 @@ crate-type = ["cdylib"]
 
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["extension-module"] }
+pyo3-async-runtimes = { version = "0.29", features = ["tokio-runtime"] }
 mssql-tds = { path = "../mssql-tds" }
 tokio = { version = "1", features = ["full", "rt-multi-thread"] }
 async-trait = "0.1"
--- a/mssql-py-core/src/async_cursor.rs
+++ b/mssql-py-core/src/async_cursor.rs
@@ -1,0 +1,264 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! First-class asynchronous (coroutine) cursor over the shared TDS core.
+//!
+//! [`PyCoreAsyncCursor`] shares the same [`SharedClient`] cell as the synchronous
+//! [`PyCoreCursor`](crate::cursor::PyCoreCursor), but exposes genuine `asyncio`
+//! coroutines: `execute`/`fetchone`/`fetchall`/`fetchmany`/`close` each return a
+//! Python awaitable built by [`pyo3_async_runtimes::tokio::future_into_py`]. There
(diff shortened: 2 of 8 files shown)


saurabh500:

PR #210, opened 2026-08-10
Report client-side rows_copied for bulk copy (fix 2x on Fabric Warehouse)

### LINKED ISSUE #209: bulkcopy() and bulkcopy_arrow() report 2x rows_copied and batch_count against Fabric Warehouse
### Describe the bug

When the target is a Fabric Warehouse (`SERVERPROPERTY('EngineEdition') = 11`), both `cursor.bulkcopy()` and `cursor.bulkcopy_arrow()` return statistics that are exactly double the truth.

The **data written is always correct**. Only the returned statistics are wrong:

| field | behavior |
| --- | --- |
| `rows_copied` | 2x the rows actually inserted |
| `batch_count` | 2x when `batch_size` is set; reports 1 when `batch_size=0` |
| `rows_per_second` | 2x, since it is derived from `rows_copied` |
| `elapsed_time` | correct |

No exception is raised. Callers that trust `rows_copied` for reconciliation, logging, or progress reporting get a silently wrong number.

The doubled value also reaches the driver log:

```
INFO, cursor.rs:551, py-core, bulkcopy_arrow: Starting Arrow bulkcopy to table: dbo.logprobe
INFO, cursor.py:3657, Python, bulkcopy_arrow: completed - rows_copied=6000, batch_count=6, elapsed_time=2.3745249
```

That run copied 3,000 rows in 3 batches.

### To reproduce

Needs `pyarrow` and `azure-identity`, and an `az login` session with write access to a Fabric Warehouse.

```python
import pyarrow as pa
from azure.identity import AzureCliCredential
from mssql_python import connect

TABLE = "dbo.minrepro"
CONN_STR = (
    "Server=WORKSPACE_ID.datawarehouse.fabric.microsoft.com,1433;"
    "Database=WAREHOUSE_NAME;Encrypt=yes;TrustServerCertificate=no"
)

conn = connect(CONN_STR, token_provider=AzureCliCredential(process_timeout=60))
cur = conn.cursor()

cur.execute(f"IF OBJECT_ID('{TABLE}') IS NOT NULL DROP TABLE {TABLE}")
cur.execute(f"CREATE TABLE {TABLE} (id BIGINT)")
conn.commit()

result = cur.bulkcopy_arrow(TABLE, pa.table({"id": pa.array([1], pa.int64())}), timeout=600)
conn.commit()

cur.execute(f"SELECT COUNT(*) FROM {TABLE}")
print("reported:", result["rows_copied"], " actual:", cur.fetchone()[0])
```

Output:

```
reported: 2  actual: 1
```

### Expected behavior

`rows_copied` equals the number of rows actually inserted, and `batch_count` equals the number of batches actually sent, on every engine.

### Scope of the defect

Reproduced with one row, so no large dataset is needed. The 2x factor is stable and invariant to:

- row count, from 1 to 200,000
- API, `bulkcopy` and `bulkcopy_arrow` behave identically
- source type, `pyarrow.Table`, `RecordBatchReader`, `DuckDBPyRelation`, and `list` of tuples
- `batch_size`, including 0
- `table_lock`, `keep_nulls`, `use_internal_transaction`

Engine matrix, 5,000 rows, both APIs, `batch_size` 0 and 1000:

| engine | EngineEdition | version | result |
| --- | --- | --- | --- |
| SQL Server 2025 | 3 | 17.0.1125.2 | correct, 8/8 (including a clustered columnstore target) |
| Fabric SQL database | 12 | 12.0.2000.8 | correct, 8/8 (including a clustered columnstore target) |
| Fabric Warehouse | 11 | 12.0.2000.8 | 2x, 4/4 |

Sample from the Warehouse run:

| API | batch_size | requested | reported | actual | batches reported | batches expected |
| --- | --- | --- | --- | --- | --- | --- |
| `bulkcopy_arrow` | 0 | 1,000 | 2,000 | 1,000 | 1 | 1 |
| `bulkcopy` | 0 | 1,000 | 2,000 | 1,000 | 1 | 1 |
| `bulkcopy_arrow` | 1,000 | 5,000 | 10,000 | 5,000 | 10 | 5 |
| `bulkcopy` | 1,000 | 5,000 | 10,000 | 5,000 | 10 | 5 |
| `bulkcopy_arrow` | 10,000 | 50,000 | 100,000 | 50,000 | 10 | 5 |

### Why this looks like a client-side bug rather than a server one

`bcp` was pointed at the same warehouse and the same table shape. It performs a TDS bulk load through the ODBC driver, so it exercises the same protocol path the Rust core does:

```
bcp dbo.bcpprobe in bcp_in.txt -S WORKSPACE_ID.datawarehouse.fabric.microsoft.com,1433 -d WAREHOUSE_NAME -c -G -U USER_UPN

Starting copy...
3 rows copied.
```

Actual row count afterwards was 3. So the correct count is derivable from what Warehouse sends back, and ODBC derives it correctly.

The Python layer is not involved. `Cursor.bulkcopy` and `Cursor.bulkcopy_arrow` in `mssql_python/cursor.py` return the dict from `pycore_cursor.bulkcopy(...)` unmodified, so this points at row-count accumulation in `mssql_py_core`.

### Untested hypothesis

Warehouse is a distributed engine. `bcp` prints a `Distributed request ID` warning on the same load, which suggests the load is acknowledged by more than one layer. A factor of exactly 2 that is stable from 1 row to 200,000 rows would fit the core summing every row-count acknowledgement it sees rather than taking the final one. That would also explain `batch_count` doubling in lockstep.

I could not confirm this. `RUST_LOG` does not reach the core's log bridge, and TDS is encrypted, so I have no token-level trace. Worth checking how DONE token row counts are accumulated.

One detail that does not fit a naive "everything counted twice" model: with `batch_size=0`, `rows_copied` still doubles but `batch_count` reports 1 rather than 2, which suggests `batch_count` is derived rather than counted on the unbatched path.

### Further technical details

Python version: 3.14.0 (amd64)
mssql-python: 1.13.0
mssql_py_core: 0.1.8
mssql-python-odbc: 18.6.2.1
SQL Server version: Fabric Warehouse, `Microsoft Azure SQL Data Warehouse (RTM) - 12.0.2000.8`, EngineEdition 11, collation `Latin1_General_100_BIN2_UTF8`
Operating system: Windows 11 ARM64
bcp control: 15.0.4298.1, ODBC Driver 17 for SQL Server

**Additional context**

Not a duplicate of #623, which was an authentication failure against Fabric Warehouse and is closed. Bulk copy connects and loads correctly here, only the returned statistics are wrong.

saurabh500:
## Description

`cursor.bulkcopy()` and `cursor.bulkcopy_arrow()` returned `rows_copied` (and the derived `batch_count` / `rows_per_second`) as exactly **2x** the truth against a Fabric Warehouse target, while the data written was always correct.

The root cause is in the core TDS layer, not Python. Bulk copy derived its reported count from the server's `DONE` token(s). `consume_done_token` accumulated `row_count` across **every** `DONE` token it saw:

```rust
rows_affected += done.row_count;
```

A single `INSERT BULK` on a normal engine emits one `DONE_COUNT` token, so summing was harmless there. A distributed engine (Fabric Warehouse, `EngineEdition = 11`) acknowledges the same load with **two** `DONE_COUNT` tokens, each carrying the *full* count, so the sum came back `2N`. This is duplication, not partitioning (proven by 1 row → reported 2). `batch_count` is derived as `rows_copied.div_ceil(batch_size)` in `mssql-py-core`, which is why it doubled in lockstep.

### The fix

Report the **client-side** count — the number of rows this client actually serialized to the wire (`StreamingBulkLoadWriter::end`) — instead of trusting the server's `DONE_COUNT`. The server response is still fully consumed for error handling, INFO capture, and to drain the connection to idle; its row count simply no longer determines what we report.

This matches how Microsoft's own bulk-copy clients count, so it is the canonical behavior and is immune to any server-side `DONE` quirk:

| Client | Reported field | Source |
|---|---|---|
| `Microsoft.Data.SqlClient` | `SqlBulkCopy.RowsCopied64` | client-side rows written |
| msodbcsql / ODBC `bcp` | `llRowsCopiedInLastBCP` | client-side rows sent (`++` per `SendBulkRow`) |
| **mssql-rs (this PR)** | `rows_written` from `writer.end()` | client-side rows written |

As defensive hardening, `consume_done_token` was also changed to take the **last** `DONE_COUNT` token rather than summing (ignoring uncounted DONEs), matching the "last `DONE_COUNT` wins" semantics used for `SQLRowCount` elsewhere.

## Validation

Reproduced and verified against a **real Fabric Warehouse** (`EngineEdition 11`), 1 / 1000 / 5000 row loads:

- **Before:** reported `2 / 2000 / 10000` (2x).
- **After:** reported `1 / 1000 / 5000` (correct).
- **TDS trace** confirmed the engine sends two `DONE_COUNT` tokens per load (`DONE(MORE|COUNT, N)` then `DONE(COUNT, N)`), each with the full count. An EngineEdition 12 Fabric SQL database sends a single counted DONE and was never affected (consistent with the issue).
- **Cross-check with `.NET Microsoft.Data.SqlClient`** against the same Warehouse: `SqlBulkCopy.RowsCopied64` reported the correct `1 / 1000 / 5000`, confirming client-side counting is the right approach — and that our previous server-count reconciliation was the defect.

## Related Issues

Fixes #209

## Testing

- **Rust unit tests** (`mssql-tds`):
  - `bulk_load_reports_client_rows_not_server_done_count` — end-to-end test that drives the full streaming bulk-load path against a mock transport which replies with two `DONE_COUNT` tokens carrying a deliberately wrong count; asserts the client reports the rows it actually wrote. (Fails with the old server-count behavior, passes with the fix.)
  - `consume_done_token_takes_single_counted_done`, `consume_done_token_does_not_double_count_distributed_dones`, `consume_done_token_ignores_uncounted_dones` — cover the de-dup hardening.
- **Python integration tests** (`mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py`): assert `rows_copied == COUNT(*)` and `batch_count == batches` for both `bulkcopy` and `bulkcopy_arrow` across `batch_size` 0/1000. Passing against local SQL Server 2022 (Docker).
- `cargo fmt --check` and `cargo clippy --workspace --all-features --all-targets -- -D warnings` pass.

> Note: a standalone SQL Server (EngineEdition 3) cannot emit the extra DONE token, so the distributed doubling is reproduced deterministically at the token level in the Rust end-to-end test, while the Python tests guard the count contract end-to-end.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] Test suites pass (new Rust unit test + Python integration tests; validated on real Fabric Warehouse)

### DIFF
--- a/mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py
+++ b/mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py
@@ -1,0 +1,74 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Regression tests for bulkcopy row-count accuracy (issue #209).
+
+`rows_copied` must equal the number of rows actually inserted, and
+`batch_count` must equal the number of batches actually sent, on every engine.
+Distributed engines (Fabric Warehouse) acknowledge one load with multiple
+DONE_COUNT tokens; the core must not sum them into a doubled count.
+"""
+import pytest
+import mssql_py_core
+
+
+@pytest.mark.integration
+@pytest.mark.parametrize(
+    "row_count,batch_size,expected_batches",
+    [
+        (1, 0, 1),
+        (1, 1000, 1),
+        (1000, 0, 1),
+        (5000, 1000, 5),
+    ],
+)
+def test_bulkcopy_rows_copied_matches_actual(
+    client_context, row_count, batch_size, expected_batches
+):
+    conn = mssql_py_core.PyCoreConnection(client_context)
+    cursor = conn.cursor()
+
+    table_name = "#BulkCopyRowsCopiedAccuracy"
+    cursor.execute(f"CREATE TABLE {table_name} (id BIGINT)")
+
+    data = [(i,) for i in range(row_count)]
+
+    result = cursor.bulkcopy(
+        table_name, iter(data), batch_size=batch_size, timeout=60
+    )
+
+    cursor.execute(f"SELECT COUNT(*) FROM {table_name}")
+    actual = cursor.fetchone()[0]
+
+    assert actual == row_count
+    assert result["rows_copied"] == actual
+    assert result["batch_count"] == expected_batches
+
+    conn.close()
+
+
+@pytest.mark.integration
+def test_bulkcopy_arrow_rows_copied_matches_actual(client_context):
+    pa = pytest.importorskip("pyarrow")
+
+    conn = mssql_py_core.PyCoreConnection(client_context)
+    cursor = conn.cursor()
+
+    table_name = "#BulkCopyArrowRowsCopiedAccuracy"
+    cursor.execute(f"CREATE TABLE {table_name} (id BIGINT)")
+
+    row_count = 5000
+    table = pa.table({"id": pa.array(list(range(row_count)), pa.int64())})
+
+    result = cursor.bulkcopy_arrow(
+        table_name, table, batch_size=1000, timeout=60
+    )
+
+    cursor.execute(f"SELECT COUNT(*) FROM {table_name}")
+    actual = cursor.fetchone()[0]
+
+    assert actual == row_count
+    assert result["rows_copied"] == actual
+    assert result["batch_count"] == 5
+
+    conn.close()
--- a/mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py
+++ b/mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py
@@ -3,13 +3,13 @@
 
 """Regression tests for bulkcopy row-count accuracy (issue #209).
 
-`rows_copied` must equal the number of rows actually inserted, and
-`batch_count` must equal the number of batches actually sent, on every engine.
-Distributed engines (Fabric Warehouse) acknowledge one load with multiple
-DONE_COUNT tokens; the core must not sum them into a doubled count.
+`rows_copied` must equal the client-side count of rows serialized to the wire,
+and `batch_count` must equal the number of batches derived from that count and
+the batch size, on every engine. Distributed engines (Fabric Warehouse)
+acknowledge one load with multiple DONE_COUNT tokens; the core must not sum them
+into a doubled count.
 """
 import pytest
-import mssql_py_core
 
 
 @pytest.mark.integration
@@ -23,10 +23,9 @@ import mssql_py_core
     ],
 )
 def test_bulkcopy_rows_copied_matches_actual(
-    client_context, row_count, batch_size, expected_batches
+    connection, row_count, batch_size, expected_batches
 ):
-    conn = mssql_py_core.PyCoreConnection(client_context)
-    cursor = conn.cursor()
+    cursor = connection.cursor()
 
     table_name = "#BulkCopyRowsCopiedAccuracy"
     cursor.execute(f"CREATE TABLE {table_name} (id BIGINT)")
@@ -44,15 +43,12 @@ def test_bulkcopy_rows_copied_matches_actual(
     assert result["rows_copied"] == actual
     assert result["batch_count"] == expected_batches
 
-    conn.close()
-
 
 @pytest.mark.integration
-def test_bulkcopy_arrow_rows_copied_matches_actual(client_context):
+def test_bulkcopy_arrow_rows_copied_matches_actual(connection):
     pa = pytest.importorskip("pyarrow")
 
-    conn = mssql_py_core.PyCoreConnection(client_context)
-    cursor = conn.cursor()
+    cursor = connection.cursor()
 
     table_name = "#BulkCopyArrowRowsCopiedAccuracy"
     cursor.execute(f"CREATE TABLE {table_name} (id BIGINT)")
@@ -70,5 +66,3 @@ def test_bulkcopy_arrow_rows_copied_matches_actual(client_context):
     assert actual == row_count
     assert result["rows_copied"] == actual
     assert result["batch_count"] == 5
-
-    conn.close()
--- a/mssql-tds/src/connection/bulk_copy.rs
+++ b/mssql-tds/src/connection/bulk_copy.rs
@@ -260,7 +260,12 @@ impl BulkCopyProgress {
 /// Contains statistics about the completed operation.
 #[derive(Debug, Clone)]
 pub struct BulkCopyResult {
-    /// Number of rows successfully copied
+    /// Number of rows serialized to the wire by the client, matching
+    /// `Microsoft.Data.SqlClient`'s `SqlBulkCopy.RowsCopied`. This is a
+    /// client-side count, not a guarantee of the number of rows inserted on
+    /// the server (server-side triggers, for example, can change the row
+    /// count), and it is unaffected by distributed engines that acknowledge
+    /// one load with multiple DONE_COUNT tokens (issue #209).
     pub rows_affected: u64,
     /// Time taken for the operation
     pub elapsed: Duration,
--- a/mssql-tds/src/connection/bulk_copy.rs
+++ b/mssql-tds/src/connection/bulk_copy.rs
@@ -41,7 +41,7 @@
 //!     .timeout(30);
 //!
 //! let result = bulk_copy.write_to_server(users.into_iter()).await?;
-//! println!("Inserted {} rows", result.rows_affected);
+//! println!("Copied {} rows", result.rows_affected);
 //! ```
 
 use crate::connection::bulk_copy_state::{ATTENTION_TIMEOUT_SECONDS, BulkCopyTimeoutState};
@@ -263,9 +263,9 @@ pub struct BulkCopyResult {
     /// Number of rows serialized to the wire by the client, matching
     /// `Microsoft.Data.SqlClient`'s `SqlBulkCopy.RowsCopied`. This is a
     /// client-side count, not a guarantee of the number of rows inserted on
-    /// the server (server-side triggers, for example, can change the row
-    /// count), and it is unaffected by distributed engines that acknowledge
-    /// one load with multiple DONE_COUNT tokens (issue #209).
+    /// the server: server-side triggers or an `IGNORE_DUP_KEY` index can make
+    /// the committed row count differ. It is unaffected by distributed engines
+    /// that acknowledge one load with multiple DONE_COUNT tokens (issue #209).
     pub rows_affected: u64,
     /// Time taken for the operation
     pub elapsed: Duration,
@@ -310,7 +310,7 @@ impl BulkCopyResult {
 ///
 /// let rows = vec![/* your data */];
 /// let result = bulk_copy.write_to_server(rows.into_iter()).await?;
-/// println!("Inserted {} rows in {:?}", result.rows_affected, result.elapsed);
+/// println!("Copied {} rows in {:?}", result.rows_affected, result.elapsed);
 /// ```
 pub struct BulkCopy<'a> {
     /// Reference to the TDS client connection
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -1065,8 +1065,17 @@ impl TdsClient {
                         ));
                     }
 
-                    // Accumulate row count from multiple DONE tokens
-                    rows_affected += done.row_count;
+                    // The bulk-load response reports its authoritative row count in
+                    // the DONE token(s) carrying the DONE_COUNT flag. A single INSERT
+                    // BULK yields one such count, but distributed engines (e.g. Fabric
+                    // Warehouse, EngineEdition 11) acknowledge the same load with more
+                    // than one DONE_COUNT token, each carrying the full count. Summing
+                    // them double-counts (issue #209). Take the last counted DONE, which
+                    // matches the authoritative "last DONE_COUNT wins" semantics used
+                    // for SQLRowCount elsewhere and is correct on every engine.
+                    if done.has_count() {
+                        rows_affected = done.row_count;
+                    }
 
                     // Stop when we receive a DONE token without the MORE flag
                     if !done.has_more() {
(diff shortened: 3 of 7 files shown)


saurabh500:
The failing check requires https://github.com/microsoft/mssql-rs/pull/213 to be merged first

saurabh500 (commented):
Update the mssql-py-core version in the Cargo.toml to 0.1.9

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1024,12 +1024,20 @@ impl TdsClient {
...
-        let _rows_written = writer.end().await?;
+        let rows_written = writer.end().await?;
 
-        // STEP 5: Read the final response with row count
-        let rows_affected = self.consume_done_token().await?;
+        // STEP 5: Consume the server response for error handling, INFO capture,
+        // and to drain the connection to an idle state. The count reported to
+        // callers is the client-side `rows_written` (the number of rows this
+        // client serialized to the wire), matching SqlClient's
+        // `SqlBulkCopy.RowsCopied` semantics. We deliberately do NOT use the
```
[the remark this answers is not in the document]
saurabh500: Good catch — updated the `# Returns` doc for `execute_bulk_load_streaming_zerocopy` to describe the client-side counting contract (rows serialized to the wire, matching SqlClient's `SqlBulkCopy.RowsCopied`), and noted it does not reflect server-side trigger effects. Pushed in aaab3ac.

> [!NOTE]

gargsaumya (approved):
LGTM. 

We may want to update the docs too - for `rows_affected` in bulk_copy.rs and the `bulkcopy()` / `bulkcopy_arrow()` return values documentation in cursor.rs, to clarify that this is the number of rows serialized to the wire, not a guarantee of rows inserted on the server.

Theekshna (approved):

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4721,6 +4743,117 @@ mod tests {
...
+        // DONE tokens without the COUNT flag carry a meaningless row_count and must
+        // not contribute to the total. Only the final DONE_COUNT value is reported.
+        let mut client = create_test_client_with_tokens(vec![
+            done_more(),
+            done_count(CurrentCommand::Insert, 3000, false),
+        ]);
+
+        let rows_affected = client.consume_done_token().await.unwrap();
+
+        assert_eq!(rows_affected, 3000);
```
David-Engel, line 4784: This test passes both before and after the fix, so it does not actually exercise the new `has_count()` guard. `done_more()` builds its token with `row_count: 0`, so the old `rows_affected += done.row_count` also yields `3000`.

To make it a real regression test, the uncounted DONE needs a non-zero `row_count`:

```rust
let mut client = create_test_client_with_tokens(vec![
    Tokens::Done(DoneToken {
        status: DoneStatus::MORE,
        cur_cmd: CurrentCommand::Insert,
        row_count: 7,
    }),
    done_count(CurrentCommand::Insert, 3000, false),
]);
```

The other two (`consume_done_token_does_not_double_count_distributed_dones` and `bulk_load_reports_client_rows_not_server_done_count`) are genuine — both fail on the old behavior.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1024,12 +1029,20 @@ impl TdsClient {
...
+        // STEP 5: Consume the server response for error handling, INFO capture,
+        // and to drain the connection to an idle state. The count reported to
+        // callers is the client-side `rows_written` (the number of rows this
+        // client serialized to the wire), matching SqlClient's
+        // `SqlBulkCopy.RowsCopied` semantics. We deliberately do NOT use the
+        // server's DONE token row count: distributed engines (e.g. Fabric
+        // Warehouse, EngineEdition 11) acknowledge one bulk load with multiple
+        // DONE_COUNT tokens, each carrying the full count, which inflated the
+        // reported total (issue #209).
+        self.consume_done_token().await?;
```
David-Engel, line 1037: After this change, `consume_done_token`'s return value is discarded at both of its call sites (here, and via `send_batch_and_consume_response` at the `INSERT BULK` preamble), so the row count it computes is no longer consumed anywhere in production code.

Keeping the last-`DONE_COUNT`-wins hardening is reasonable, but `send_batch_and_consume_response`'s doc comment still reads "Returns the row count from the DONE token," which now suggests a caller depends on it. A one-line note on either function that the count is currently informational would keep someone from re-wiring it to a caller later and reintroducing the server-count dependency.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1065,8 +1078,17 @@ impl TdsClient {
...
+                    // The bulk-load response reports its authoritative row count in
+                    // the DONE token(s) carrying the DONE_COUNT flag. A single INSERT
+                    // BULK yields one such count, but distributed engines (e.g. Fabric
+                    // Warehouse, EngineEdition 11) acknowledge the same load with more
+                    // than one DONE_COUNT token, each carrying the full count. Summing
+                    // them double-counts (issue #209). Take the last counted DONE, which
+                    // matches the authoritative "last DONE_COUNT wins" semantics used
+                    // for SQLRowCount elsewhere and is correct on every engine.
+                    if done.has_count() {
+                        rows_affected = done.row_count;
```
David-Engel: The repo's `copilot-instructions.md` has a "No AI slop" rule (no verbose comments restating what the code does). This 8-line block, plus the 10-line block at the STEP 5 call site above, plus the two rustdoc blocks, state the same rationale in four places. Something like the following carries the same information:

```rust
// Distributed engines send multiple DONE_COUNT tokens each carrying the full
// count; summing them double-counts (#209). Last counted DONE wins.
if done.has_count() {
    rows_affected = done.row_count;
}
```

mssql-tds/src/connection/bulk_copy.rs
```diff
@@ -260,7 +260,12 @@ impl BulkCopyProgress {
...
 #[derive(Debug, Clone)]
 pub struct BulkCopyResult {
-    /// Number of rows successfully copied
+    /// Number of rows serialized to the wire by the client, matching
+    /// `Microsoft.Data.SqlClient`'s `SqlBulkCopy.RowsCopied`. This is a
+    /// client-side count, not a guarantee of the number of rows inserted on
+    /// the server (server-side triggers, for example, can change the row
+    /// count), and it is unaffected by distributed engines that acknowledge
+    /// one load with multiple DONE_COUNT tokens (issue #209).
     pub rows_affected: u64,
```
David-Engel, line 269: Worth naming `IGNORE_DUP_KEY` alongside triggers — an index with `IGNORE_DUP_KEY = ON` on the destination is the other common case where the server legitimately inserts fewer rows than the client sent, and it is more likely to surprise someone than a trigger.

Also, the module-level examples at lines 44 and 313 still print `"Inserted {} rows"`, which now overstates what the value means. `"Copied"` or `"Sent"` would match the new semantics.

mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py
```diff
@@ -0,0 +1,74 @@
...
+    result = cursor.bulkcopy(
+        table_name, iter(data), batch_size=batch_size, timeout=60
+    )
+
+    cursor.execute(f"SELECT COUNT(*) FROM {table_name}")
+    actual = cursor.fetchone()[0]
+
+    assert actual == row_count
+    assert result["rows_copied"] == actual
+    assert result["batch_count"] == expected_batches
```
David-Engel, line 44: `batch_count` is computed in `mssql-py-core/src/cursor.rs` as `rows_copied.div_ceil(batch_size)` (or `1` when `batch_size == 0`), so this is arithmetic on the value asserted one line above rather than an independent count of batches actually sent.

Fine to keep as a contract check, but the module docstring's "`batch_count` must equal the number of batches actually sent" claims more than the test verifies — worth softening so a future reader does not trust it as coverage of the batching loop.

David-Engel (commented):
Reviewed the full diff and ran the new Rust tests locally (all 5 pass). The approach is sound: reporting the client-side count matches `SqlBulkCopy.RowsCopied` and ODBC `bcp` (`llRowsCopiedInLastBCP`), and the blast radius of the `consume_done_token` change is small — it is only reachable from the two bulk-load call sites, so it cannot regress row counts for `execute` / RPC paths.

No blocking issues. A few suggestions inline, plus one that has no diff line to attach to:

**Add a CHANGELOG entry.** `CHANGELOG.md` maintains an `## [Unreleased]` / `### Changed` section that already documents `mssql-tds` behavior changes. `BulkCopyResult::rows_affected` is a public field whose meaning changed (server-reported inserted rows → client-serialized rows), so it belongs there and not only in rustdoc.

mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py
```diff
@@ -0,0 +1,74 @@
...
+
+    conn.close()
```
David-Engel: `conn.close()` after the asserts leaks the connection whenever an assertion fails. `conftest.py` already has a `connection` fixture that closes in teardown, so taking `connection` instead of `client_context` (or wrapping the body in `try/finally`) gets deterministic cleanup. Same at line 74.

Good call on using temp tables here rather than the named-table pattern the sibling bulkcopy tests use — that avoids cross-run collisions.

mssql-tds/src/connection/tds_client.rs
saurabh500: Good catch. Reworked the test to use an inline uncounted DONE (COUNT flag off) carrying a non-zero row_count of 7, so the pre-fix summing path would report 3007 and the test now genuinely exercises the has_count() guard.


> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Added a note to the consume_done_token doc (and to send_batch_and_consume_response) that the returned count is currently informational: both call sites discard it and the bulk-load path reports the client-side rows written instead.


> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Condensed the repeated #209 rationale down to a two-line inline comment at the guard and trimmed the STEP 5 comment so the explanation lives in one place.


> [!NOTE]

mssql-tds/src/connection/bulk_copy.rs
saurabh500: Added IGNORE_DUP_KEY to the rows_affected doc as another reason the committed count can differ, and changed the two `Inserted {} rows` examples to `Copied` to match the client-side semantics.


> [!NOTE]

mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py
saurabh500: Softened the docstring: batch_count is derived from the client-side rows-copied count and the batch size, not an independent count of batches sent.


> [!NOTE]

mssql-py-core/tests/test_bulkcopy_rows_copied_accuracy.py
saurabh500: Switched both tests to the auto-closing `connection` fixture from conftest.py, which closes the connection in teardown even when an assertion fails, removing the manual create/close.


> [!NOTE]

David-Engel (approved):
Re-reviewed at `bebe826c`. All six inline findings are addressed and the threads are resolved. Verified locally: `cargo test -p mssql-tds --lib -- consume_done_token bulk_load_reports` → 5/5 pass, `cargo fmt -- --check` clean, and `done_more()` still has two other call sites so the test rewrite introduces no `dead_code` warning.

The `consume_done_token_ignores_uncounted_dones` rewrite is the one I most wanted — with `row_count: 7` on the uncounted DONE, the pre-fix summing path would report `3007`, so the test now genuinely proves the `has_count()` guard rather than passing on an incidental zero.

**Approving.** One non-blocking follow-up, at your discretion:

`CHANGELOG.md` has no entry for this. `BulkCopyResult::rows_affected` is a public field whose documented meaning changed (server-reported inserted rows → client-serialized rows), and the `## [Unreleased]` / `### Changed` section already carries comparable `mssql-tds` behavior notes. Something like:

```markdown
- `mssql-tds`: `BulkCopyResult::rows_affected` now reports the number of rows the
  client serialized to the wire (matching `SqlBulkCopy.RowsCopied`) instead of the
  server's `DONE_COUNT`. Fixes a doubled count on distributed engines that
  acknowledge one load with multiple `DONE_COUNT` tokens.
```

Happy to see this merge without it if you would rather batch changelog updates before the next release.

saurabh500:
@david-engel added the changelog.

David-Engel (approved):

PR #213, opened 2026-08-11
Skip astral surrogate GetData test on msodbcsql parity leg

shiwanigupta0809:
## Description

NvarcharMaxToCharChunkedAstralRoundTrip diverges on the Windows msodbcsql leg: mssql-odbc delivers SQL_C_CHAR as UTF-8 (emoji round-trips), while msodbcsql on Windows converts SQL_C_CHAR to the client ANSI codepage, where U+1F600 best-fits to '?'. On Linux both deliver UTF-8 and agree, so the divergence is Windows-only. This is the same intentional UTF-8-vs-ANSI SQL_C_CHAR difference already documented for other tests here, so guard it with SKIP_IF_COMPARING_MSODBCSQL() and correct the stale 'codepage-neutral' comment.

## Related Issues

https://github.com/microsoft/mssql-rs/issues/214

## Checklist

- [ ] `cargo bfmt` passes
- [ ] `cargo bclippy` passes
- [ ] `cargo btest` passes


saurabh500 (approved):
Create a Github / ADO issue and attach to PR. 

Else LGTM

PR #153, opened 2026-07-27
SQLFetch and SQLGetData update - column wise fetch and incremental PLP support

shiwanigupta0809:
**Description**
Aligns mssql-odbc SQLFetch and SQLGetData with ODBC's sparse, request-driven row model, and implements correct PLP (partially-length-prefixed) continuation semantics. Row and column data are now decoded lazily, on demand, rather than materializing the whole row up front.

**What changed**
Deferred row materialization — SQLFetch now pauses before decoding the first column. A fetched row is positioned but not read until the application requests a specific column.
Column-wise decode in SQLGetData — resumes decoding forward to the requested column instead of relying on a fully materialized row or value. Intervening unread columns are drained transparently.
Direct PLP streaming — large PLP values are streamed straight from the active TDS PLP stream across successive SQLGetData calls, with no ODBC-side cache of the fully decoded value. This keeps memory bounded regardless of value size.
Simplified continuation state — per-statement PLP tracking is reduced to the active column plus its target C type.

**Tests**
Added SQLGetData e2e coverage and wired it into the e2e CMake target:

Repeated small-buffer SQLGetData calls that stream a PLP value in chunks.
Mixed PLP and non-PLP column access within a row.
Sparse / non-sequential column access patterns.
## Related Issues
https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46641

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes
- [x] New/changed functionality has tests
- [ ] Public API changes are documented

PR #217, opened 2026-08-11
odbc: mandatory source-type conversions for typed SQLGetData (P1a) [AB#47107]

David-Engel:
## What

ODBC Appendix D requires every supported SQL type to convert to every supported C type. P1 (#146) implemented the C-target side: integers, floats, GUID and the date/time structs. This is the source-type side, so those targets now accept the column types that were previously rejected.

Stacked on P1, rebased onto `main` now that #146 has merged.

### Numeric targets

- New `NumericSource` abstraction keeps exact sources exact, so an integer target can report truncation instead of silently dropping a fraction: `Int`, `Scaled { mantissa, scale }`, `Float`.
- `decimal` and `numeric` parse from their own exact rendering rather than reassembling base-2^32 limbs; `money` and `smallmoney` use their 10^4-scaled wire value.
- `convert_integer_c` and `convert_float_c` now accept `decimal`, `numeric`, `money`, `smallmoney`, `real`, `float` and character sources.

### Date/time targets

- Character sources parse `YYYY-MM-DD`, `HH:MM[:SS[.fffffff]]`, and a combined datetime with either a `T` or a space separator, plus an optional trailing offset.
- The offset is matched only in the exact 6-character `[+-]HH:MM` shape, so the hyphens inside a date are never mistaken for one.
- Range validation on every field (year 1-9999, month 1-12, day 1-31, hour < 24, minute/second < 60, offset <= 14:59).

### Diagnostics

| Situation | Result |
| --- | --- |
| Lossy but well-defined conversion (`float 1234.99` into `SQL_C_SLONG`) | value truncated toward zero, `01S07` + `SQL_SUCCESS_WITH_INFO`, matching msodbcsql18 |
| Character text that is not a valid literal for the target | `22018` |
| Source with no numeric or temporal interpretation (binary, guid) into that target | `07006`, since the pairing is illegal rather than unimplemented |

## Notes for reviewers

- This removes the `ConvError::NotHandledHere` carve-out that #146 left in `convert_datetime_c`. That branch existed only to report "not implemented yet" for character sources, and is what this PR implements, so character input now goes through `parse_datetime_literal` instead.
- Consequently `get_data_character_into_date_target_is_not_implemented` (a placeholder from #146 asserting the deferral) is replaced by two tests: one asserting the conversion succeeds, and one asserting an invalid literal is `22018` rather than a silent zero value. The diagnostic coverage the old test provided is preserved.
- `sql_string_to_text` already moved into `fetch_convert` as part of #146, so the equivalent move in this branch's first commit collapsed away during the rebase.

## Known limitation — deliberately deferred ([AB#47238](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47238))

A `varchar(max)` / `nvarchar(max)` source into a numeric or date/time target is not covered. Non-max character columns are unaffected. This is deferred rather than overlooked, for three reasons:

- Max-length columns arrive as PLP and are delivered incrementally, so parsing would require the ODBC layer to accumulate chunks. That inverts the invariant #153 established — `stream_active_plp_chunk` documents "This never buffers the full PLP payload in ODBC-layer memory" — and that exception should be @shiwanigupta0809's call rather than a quiet one-off.
- #204 and #215 are actively rewriting this code. #215 changes `read_active_plp_chunk` itself and moves its call site from `block_on` to `pin!`; #204 rewires `get_data.rs` onto the reactor-free sync core. Anything written here now would conflict on a performance-critical hot path and would likely need rewriting once the sync core lands.
- There is a policy question to settle first: `varchar(max)` holds up to 2 GB, but a valid numeric or datetime literal is under ~64 bytes, so draining an unbounded payload to produce a `SQL_C_SLONG` is a memory-exhaustion risk driven by server-side data. The likely answer is a bounded prefix that fails with `22018` / `22003` past the cap, but that should be confirmed against msodbcsql18's actual behaviour before being encoded.

Tracked as [AB#47238](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47238), sequenced after #204 and #215.

## Testing

509 unit tests pass, workspace clippy and `cargo fmt` clean.

[AB#47107](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47107)


mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -126,22 +116,147 @@ unsafe fn write_fixed<T: Copy>(ptr: SqlPointer, value: T, ind: *mut SqlLen) -> C
...
+
+    /// Value truncated toward zero plus whether a fractional part was dropped.
+    /// `None` when the value cannot be represented as an integer at all.
+    fn to_i128_truncating(self) -> Option<(i128, bool)> {
+        match self {
+            NumericSource::Int(v) => Some((v, false)),
+            NumericSource::Scaled { mantissa, scale } => {
+                let divisor = 10i128.checked_pow(scale)?;
+                Some((mantissa / divisor, mantissa % divisor != 0))
+            }
```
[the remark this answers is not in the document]
David-Engel: Good catch, fixed in f7e15f70. `to_i128_truncating` now returns `(0, mantissa != 0)` when `10^scale` overflows, so a value like `0.000000000000000000000000000000000000001` truncates to zero with `01S07` instead of reporting `22003`. Your reasoning is exactly right: any divisor that overflows `i128` is larger than every representable mantissa, so the quotient is known to be zero. Added `scale_beyond_i128_truncates_to_zero`, which covers both the helper and the end-to-end character path.

One correction: there is no second occurrence. `checked_pow` appears only once in this file (line 151); line 606 is the trailing UTC-offset parsing in `parse_datetime_literal`. The `f64` scaling on line 139 uses `powi`, which handles `1e39` without overflowing.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -438,20 +545,127 @@ pub(crate) fn is_datetime_c_target(target_type: SqlSmallInt) -> bool {
...
+    if it.next().is_some() || y.len() != 4 {
+        return None;
+    }
+    let year: i16 = y.parse().ok()?;
+    let month: u16 = m.parse().ok()?;
+    let day: u16 = d.parse().ok()?;
+    if !(1..=9999).contains(&year) || !(1..=12).contains(&month) || !(1..=31).contains(&day) {
+        return None;
+    }
+    Some((year, month, day))
```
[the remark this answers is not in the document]
David-Engel: Confirmed and fixed in f7e15f70. `parse_date_literal` now validates the day against the selected month via a new `days_in_month` helper that applies the Gregorian leap rule, so `2023-02-31`, a non-leap `2023-02-29` and `2023-04-31` all return `22018`, while `2024-02-29` still converts. Added `impossible_calendar_dates_are_rejected` covering all four.

This was the most valuable of the five: those literals were previously written into `SQL_DATE_STRUCT` / `SQL_TIMESTAMP_STRUCT` as successful conversions, so an application would have received a date that does not exist.

One correction: the issue does not also appear on line 582. That line is the fractional-digit check in `parse_time_literal`; the hour/minute/second bounds there are already exact.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -438,20 +545,127 @@ pub(crate) fn is_datetime_c_target(target_type: SqlSmallInt) -> bool {
...
-            _ => ConvError::Restricted,
-        });
+    let p = match value {
+        // A character column must hold a valid literal for the target.
+        ColumnValues::String(s) => {
+            let text = sql_string_to_text(s).ok_or(ConvError::InvalidCharacterValue)?;
+            parse_datetime_literal(&text).ok_or(ConvError::InvalidCharacterValue)?
+        }
+        // A date/time C target for a non-temporal column is illegal.
+        _ => extract_datetime_parts(value).ok_or(ConvError::Restricted)?,
```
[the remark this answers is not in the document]
David-Engel: Agreed, fixed in f7e15f70. `convert_datetime_c` now captures whether the source was character data and uses it at the target-shape mismatch arm, so `'12:00'` into `SQL_C_TYPE_DATE` reports `22018` rather than `07006`. That matches the contract in the PR description: for a character source the pairing is legal and it is the text that is wrong for the requested target.

Scoped deliberately to character sources. A real `time` column into `SQL_C_TYPE_DATE` is still `07006`, which `time_into_date_target_is_restricted` continues to cover. New test: `character_time_into_date_target_is_invalid_character_value`.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -126,22 +116,147 @@ unsafe fn write_fixed<T: Copy>(ptr: SqlPointer, value: T, ind: *mut SqlLen) -> C
 /// `target_value_ptr`, when non-null, must be valid for a write of the target
 /// C type's size, and `strlen_or_ind_ptr` must be null or valid for a
 /// `SqlLen` write.
+/// A numeric column value in a form that keeps exact sources exact, so an
+/// integer target can report truncation instead of silently dropping a
+/// fraction.
+#[derive(Debug, Clone, Copy, PartialEq)]
```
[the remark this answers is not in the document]
David-Engel: Correct, fixed in f7e15f70 — and there was a second instance you did not flag. Inserting `NumericSource` orphaned `convert_integer_c`'s doc block (its `# Safety` contract had attached to the enum), and inserting the literal parsers orphaned `convert_datetime_c`'s doc block onto `parse_date_literal`. Both functions were left undocumented while a `# Safety` section sat on a safe item.

Rather than just relocating the text, I updated it, since as you noted the behaviour is no longer integer-only: `convert_integer_c` now documents the decimal, money, float and character sources it accepts, and which of `ConvOk::Truncated` / `InvalidCharacterValue` / `Restricted` / `NotHandledHere` each case produces.

mssql-odbc/docs/typed-columnar-fetch-plan.md
```diff
@@ -86,7 +86,7 @@ ODBC Appendix D requires a driver to support conversions to **all** ODBC C types
 - Character sources (`char` / `varchar` / `nchar` / `nvarchar`) → numeric and date/time C targets (e.g. `'123'` → `SQL_C_SLONG`, `'2023-06-15'` → `SQL_C_TYPE_DATE`), with `22018` when the text is not a valid literal for the target.
 - Lossy **numeric** conversions must report fractional truncation with `01S07` + `SQL_SUCCESS_WITH_INFO` (e.g. `float` `1234.99` → `SQL_C_SLONG` yields `1234` + `01S07`). The `01S07` diagnostic and the `ConvOk::Truncated` plumbing already exist (P1 uses them for date/time targets that discard a component); P1a extends them to the numeric conversions above.
 
-Until then these pairings return `HYC00` (not implemented) rather than converting.
+Implemented on `david/odbc-p1a-conversions` (stacked on P1): a `NumericSource` abstraction keeps exact sources exact so integer targets report `01S07` truncation instead of silently dropping a fraction; character sources parse decimal literals exactly (falling back to `f64` for exponent forms) and the `date` / `time` / `datetime2` / `datetimeoffset` character forms; invalid text returns `22018`. A source with no interpretation for the requested target (binary, guid) is `07006`.
```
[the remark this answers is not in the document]
David-Engel: Fixed in f7e15f70. The P1a section is now written in past tense throughout: the lead-in says P1a added the missing source types, and each bullet describes what shipped instead of what "must be added". Dropped the stale reference to `numeric_source_as_f64` (that helper no longer exists — the exact-decimal types now route through `NumericSource` rather than `f64`), and folded the separate "Implemented in PR #217" paragraph into the bullets so there is a single description rather than a plan and a result that can drift apart.

David-Engel (commented):
## Review — P1a source-type conversions

Overall this is good work. The `NumericSource` abstraction is the right call — keeping exact-decimal sources exact so an integer target can report `01S07` instead of silently dropping a fraction is what Appendix D wants, and the `22018` vs. `07006` split (bad text for a legal pairing vs. an illegal pairing) is correct and well covered by tests. Docs and the phase table are updated, and the e2e HYC00 anchor was re-pointed to `SQL_C_NUMERIC` with a rationale for why that anchor is durable.

Verified locally on `37434368`: 512 unit tests pass, `cargo clippy -p mssql-odbc --all-targets` is clean.

One blocker below.

---

### 🔴 Blocking — panic on non-ASCII character input

[`fetch_convert.rs#L633-L634`](https://github.com/microsoft/mssql-rs/blob/374343681a626939be6005758d7cdc8141bf1d69/mssql-odbc/src/api/fetch_convert.rs#L633-L634)

```rust
if s.len() >= 6 {
    let tail = &s[s.len() - 6..];
```

`s.len()` is bytes, so `s.len() - 6` can land inside a multi-byte UTF-8 sequence and `&s[..]` panics. A panic unwinding through the ODBC `extern "C"` boundary is UB / process abort in the Driver Manager, driven by server-side data.

Repro, confirmed against this branch:

```
SELECT N'日aaaa'  ->  SQLGetData(1, SQL_C_TYPE_DATE, ...)
```

```
panicked at mssql-odbc/src/api/fetch_convert.rs:634:22:
start byte index 1 is not a char boundary; it is inside '日' (bytes 0..3) of `日aaaa`
```

The offset shape is pure ASCII, so the cheapest fix is to work on bytes and never slice the `str`:

```rust
if let Some(tail) = s.as_bytes().get(s.len() - 6..) {
    if (tail[0] == b'+' || tail[0] == b'-')
        && tail[3] == b':'
        && tail[1..3].iter().chain(&tail[4..6]).all(u8::is_ascii_digit)
    {
        // ...
    }
}
```

`as_bytes().get(..)` returns `None` rather than panicking, and the explicit digit check also removes the `tail[1..3].parse()` slicing. Worth a regression test that drives a non-ASCII literal into each date/time target.

### 🟡 `'inf'` / `'NaN'` text is accepted as a number

[`fetch_convert.rs#L223-L231`](https://github.com/microsoft/mssql-rs/blob/374343681a626939be6005758d7cdc8141bf1d69/mssql-odbc/src/api/fetch_convert.rs#L223-L231)

The fallback is `text.trim().parse::<f64>()`, and Rust's parser accepts `inf`, `infinity` and `nan` case-insensitively. Confirmed behaviour:

- `'inf'` into `SQL_C_DOUBLE` returns `Ok(Exact)` and writes `inf` into the application's buffer.
- `'NaN'` does the same, writing `NaN`.
- Either into `SQL_C_SLONG` reports `22003` (`to_i128_truncating` returns `None`) where msodbcsql would report `22018`.

SQL Server has no `inf` / `nan` literal, so this should be `InvalidCharacterValue`. An `f.is_finite()` guard on the fallback covers it.

### 🟡 Stale doc comment

[`fetch_convert.rs#L305-L308`](https://github.com/microsoft/mssql-rs/blob/374343681a626939be6005758d7cdc8141bf1d69/mssql-odbc/src/api/fetch_convert.rs#L305-L308) — deleting `numeric_source_as_f64` left its doc comment stranded on top of `is_float_c_target`, which now carries two contradictory doc blocks:

```rust
/// Widen a numeric column (integer or floating) to `f64`. Returns `None` for
/// non-numeric sources.
/// Returns `true` if `target_type` is one of the floating-point C types handled
/// by [`convert_float_c`].
pub(crate) fn is_float_c_target(target_type: SqlSmallInt) -> bool {
```

### Nits

- [`fetch_convert.rs#L187`](https://github.com/microsoft/mssql-rs/blob/374343681a626939be6005758d7cdc8141bf1d69/mssql-odbc/src/api/fetch_convert.rs#L187) — `money_scaled` duplicates the `(lsb & 0xFFFF_FFFF) | (msb << 32)` assembly already written inline at [`get_data.rs#L1023`](https://github.com/microsoft/mssql-rs/blob/374343681a626939be6005758d7cdc8141bf1d69/mssql-odbc/src/api/get_data.rs#L1023). One shared helper keeps the two from drifting.
- [`fetch_convert.rs#L215`](https://github.com/microsoft/mssql-rs/blob/374343681a626939be6005758d7cdc8141bf1d69/mssql-odbc/src/api/fetch_convert.rs#L215) — `ColumnValues::String(_) => None` immediately followed by `_ => None` is redundant. If it is there for documentation, a comment pointing at `numeric_source_or_parse` carries that better than a duplicate arm.
- [`fetch_convert.rs#L613`](https://github.com/microsoft/mssql-rs/blob/374343681a626939be6005758d7cdc8141bf1d69/mssql-odbc/src/api/fetch_convert.rs#L613) — fractional seconds past 7 digits are silently discarded: `12:34:56.123456789` yields `123456700` with no `ConvOk::Truncated`. That is fractional truncation and arguably deserves `01S07`, same as the other paths this PR adds.
- [`fetch_convert.rs#L205`](https://github.com/microsoft/mssql-rs/blob/374343681a626939be6005758d7cdc8141bf1d69/mssql-odbc/src/api/fetch_convert.rs#L205) — `parse_decimal_literal(&d.to_string())` formats and re-parses per value on the fetch path. The exactness argument is sound; a short comment noting the allocation is accepted for now (or a follow-up item) would help whoever profiles this later.
- `convert_integer_c` now gates on `is_integer_c_target` at entry, so the trailing `_ => return Err(NotHandledHere)` inside the match is unreachable by construction. Harmless, but if the gate and the match ever diverge the gate wins silently.
- Edge case: `'0.5'` into `SQL_C_BIT` writes `0` + `01S07`. Plausible, but msodbcsql likely returns `22003` — worth confirming on the parity leg at some point.

### CI

ADO validation reports "had test failures": `TestAuthTcClash::test_row27_tc_yes_ad_interactive` and `test_tc_yes_ad_default`, both failing with `mssql_python.auth does not have the attribute 'get_auth_token'`. That is cross-repo mock drift, unrelated to this change.

David-Engel:
All addressed in 2a91adbc. 516 tests pass (+4 regression tests), workspace clippy and `cargo fmt` clean.

### 🔴 Panic on non-ASCII character input

Fixed, and thank you — this was the right call to block on. The probe now works on bytes and never slices the `str`:

```rust
if let Some(tail) = s.len().checked_sub(6).and_then(|i| s.as_bytes().get(i..))
    && (tail[0] == b'+' || tail[0] == b'-')
    && tail[3] == b':'
    && tail[1..3].iter().chain(&tail[4..6]).all(u8::is_ascii_digit)
```

`checked_sub` also removes the underflow path, the digit check retires the `tail[1..3].parse()` slicing, and the surviving `s[..s.len() - 6]` is now only reachable once the tail is known to be ASCII, so that boundary is guaranteed to be a char boundary. Added `non_ascii_character_input_does_not_panic`, which drives `日aaaa` into `SQL_C_TYPE_DATE`, `SQL_C_TYPE_TIME` and `SQL_C_TYPE_TIMESTAMP`.

### 🟡 `'inf'` / `'NaN'`

Fixed — the `f64` fallback now requires `is_finite()`, so these are `22018` rather than writing a non-finite value into the caller's buffer. Covered for `inf`, `-inf`, `Infinity`, `NaN` and `nan`.

### 🟡 Stale doc comment

Removed. That is the third orphaned doc block this PR has produced, all from deleting or inserting items next to a doc comment — worth a glance in review whenever a helper is removed.

### Nits

- **`money_scaled` duplication** — now shared; `get_data.rs` calls it instead of reassembling the wire value inline. Changed its return to `i64` so both callers use it directly.
- **Redundant `String(_) => None`** — dropped, replaced with a comment pointing at `numeric_source_or_parse` as you suggested.
- **Fractional seconds past 7 digits** — agreed, that is truncation. `DateTimeParts` now carries `frac_truncated` and `convert_datetime_c` reports `01S07`. `12:34:56.123456789` yields `123456700` + `01S07`.
- **`d.to_string()` allocation** — comment added recording that the format-and-reparse is accepted in exchange for exactness.
- **Unreachable `_` arm** — kept as a backstop with a comment saying the `is_integer_c_target` gate already rejected everything else.

### `'0.5'` into `SQL_C_BIT` — checked against the msodbcsql source, and the hypothesis is half right

Worth spelling out, because the answer is narrower than expected. From `sqlccnvt.cpp`:

```c
if (!fUnsignedIn && Error == CVT_FRACT_TRUNC && fTypeOut == SQL_C_BIT)
    return CVT_PREC;
```

with `CVT_PREC` = `IDS_22_003` = "Numeric value out of range", and `fUnsignedIn` set when the text carries no `-`. So:

- `'0.5'` into `SQL_C_BIT` — positive, so this branch does **not** fire; it truncates to `0` with `CVT_FRACT_TRUNC` (`IDS_01_S07`). **We already matched.**
- `'-0.5'` into `SQL_C_BIT` — negative, so msodbcsql returns **22003**, while we returned `01S07`. That is the real divergence, and the double path agrees (`dTemp < 0 && fTypeOut == SQL_C_BIT` → `CVT_PREC`).

Fixed by rejecting a negative source for a bit target before the truncation check, with `negative_into_bit_target_is_out_of_range` covering it.

### CI

Agreed, unrelated. `TestAuthTcClash` fails on `mssql_python.auth does not have the attribute 'get_auth_token'`, which is a mock in the cross-repo mssql-python leg; nothing in this PR touches auth. Re-running validation on the new head.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -115,12 +105,162 @@ unsafe fn write_fixed<T: Copy>(ptr: SqlPointer, value: T, ind: *mut SqlLen) -> C
...
+    if let ColumnValues::String(s) = value {
+        let text = sql_string_to_text(s).ok_or(ConvError::InvalidCharacterValue)?;
+        return parse_decimal_literal(&text)
+            .or_else(|| {
+                // Rust accepts "inf"/"infinity"/"nan"; SQL Server has no such
+                // literal, so they are invalid text rather than values.
+                text.trim()
+                    .parse::<f64>()
+                    .ok()
+                    .filter(|f| f.is_finite())
```
David-Engel: This is fallout from my previous round — my `is_finite` suggestion was too broad, sorry.

msodbcsql splits these two cases. `CharToDouble` calls `VarR8FromStr` and maps the result:

```c
if (hr == S_OK)                  return 0;
else if (hr == DISP_E_OVERFLOW)  return CVT_PREC;   // 22003
else                             return CVT_ERROR;
```

So text whose digits overflow a `double` is **22003**, and only unparseable text is the cast error. Rust's `f64::from_str` folds those together: `"1e400".parse::<f64>()` is `Ok(inf)`, not `Err`, so the new filter turns an overflow into `22018`. Confirmed by probe:

```
'1e400' -> SQL_C_DOUBLE: Err(InvalidCharacterValue)   // msodbcsql: 22003
'1e400' -> SQL_C_SLONG:  Err(InvalidCharacterValue)   // msodbcsql: 22003
```

`parse_decimal_literal` has the same shape: a 400-digit integer literal overflows the `i128` mantissa, falls through to `f64`, and now reports `22018` instead of `22003`.

Suggested split — a non-finite result is an overflow only if the text actually looked numeric:

```rust
.or_else(|| {
    let t = text.trim();
    match t.parse::<f64>() {
        // Rust accepts "inf"/"infinity"/"nan"; SQL Server has no such literal.
        Ok(f) if f.is_finite() => Some(Ok(NumericSource::Float(f))),
        // Digits that overflow f64 are 22003, matching VarR8FromStr's DISP_E_OVERFLOW.
        Ok(_) if t.bytes().any(|b| b.is_ascii_digit()) => Some(Err(ConvError::OutOfRange)),
        _ => None,
    }
})
```

(shape only — the `Option<Result<..>>` needs flattening against the existing `ok_or`.)

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -433,7 +578,149 @@ pub(crate) fn is_datetime_c_target(target_type: SqlSmallInt) -> bool {
...
+            .map_or(0, |b| u32::from(b - b'0'));
+        hundred_ns = hundred_ns * 10 + digit;
+    }
+    Some((
+        hour,
+        minute,
+        second,
+        hundred_ns * 100,
+        scale as u8,
+        frac_digits.len() > 7,
```
David-Engel: My earlier nit pointed the wrong way here — I asked for `01S07` past 7 digits, but msodbcsql keeps **9** and hard-errors past that. From `ParseDateTime`:

```c
if (cchToken > 9)
{
    Error = CVT_CAST_ERROR;  //  Illegal conversion
    goto ErrReturn;
}

for (cch = cchToken; cch < 9; cch++)
    value *= 10;    // convert to nanos
```

and the grammar tables allow `LEX_TK_NUMBER, -9` for `rgbECODE_TIME2` / `rgbECODE_TIMESTAMPOFFSET`.

`SQL_TIMESTAMP_STRUCT.fraction` is nanoseconds, so 9 digits are exactly representable and a character source carries no server-side scale limit. Current behaviour vs. msodbcsql:

| Input | msodbcsql | this PR |
| --- | --- | --- |
| `.123456789` (9) | `fraction = 123456789`, SUCCESS | `fraction = 123456700`, `01S07` |
| `.12345678901` (11) | `22018` | `fraction = 123456700`, `01S07` |

Both confirmed by probe. Suggest normalizing to 9 digits instead of 7 and returning `None` (`22018`) past 9, which also makes `frac_truncated` and the `p.scale` cap unnecessary. Note this only affects the **character** path — values arriving from the wire are still capped by the column's own scale.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -433,7 +578,149 @@ pub(crate) fn is_datetime_c_target(target_type: SqlSmallInt) -> bool {
...
+            return None;
+        }
+        p.tz_hour = sign * hh;
+        p.tz_minute = sign * mm;
+        p.has_tz = true;
+        // The matched tail is all ASCII, so this boundary is a char boundary.
+        s = s[..s.len() - 6].trim_end();
+    }
+
+    let (date_str, time_str) = match s.split_once(['T', ' ']) {
```
David-Engel, line 716: Two literal forms msodbcsql accepts and this parser rejects, both reachable from a plain `varchar` column:

1. **`YYYY/MM/DD`.** msodbcsql retries the date with a second grammar when the dash form fails: `rgbECODE_DATE_SLASH` is `NUMBER 4 / '/' / NUMBER 2 / '/' / NUMBER 2`, driven by `if (Error == CVT_ERROR) { if (ECODE_DATE == iIDB) { ...rgbECODE_DATE_SLASH... } }`. Probe: `'2023/06/15'` into `SQL_C_TYPE_DATE` is `22018` here.
2. **ODBC escape literals** `{d '...'}`, `{t '...'}`, `{ts '...'}`, handled by the `FindECode` branch ahead of the plain-format branch. Probe: `{d '2023-06-15'}` is `22018` here.

Going the other way, this parser is more permissive than msodbcsql in three places. msodbcsql separates tokens with `SkipWSSequence` only and buckets by exact length (`cchTmp == 10` date, `== 8` time, `8..=18` time2, `19..=36` timestampoffset), with `LEX_TK_NUMBER, 2` demanding exactly two digits:

- `T` as the date/time separator — msodbcsql takes whitespace only.
- `HH:MM` without seconds — msodbcsql needs `HH:MM:SS` (`'12:00'` into `SQL_C_SS_TIME2` succeeds here, probe-confirmed).
- Unpadded fields like `2023-6-5`.

The permissive side is low risk. The two rejections are the ones an application would actually hit. Fine to defer either way, but they should be a tracked item rather than a silent gap, especially since `SKIP_IF_COMPARING_MSODBCSQL()` means the parity leg will not catch them.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -433,7 +578,149 @@ pub(crate) fn is_datetime_c_target(target_type: SqlSmallInt) -> bool {
...
+    if let Some(tail) = s.len().checked_sub(6).and_then(|i| s.as_bytes().get(i..))
+        && (tail[0] == b'+' || tail[0] == b'-')
+        && tail[3] == b':'
+        && tail[1..3].iter().chain(&tail[4..6]).all(u8::is_ascii_digit)
+    {
+        let sign: i16 = if tail[0] == b'+' { 1 } else { -1 };
+        let hh = i16::from(tail[1] - b'0') * 10 + i16::from(tail[2] - b'0');
+        let mm = i16::from(tail[4] - b'0') * 10 + i16::from(tail[5] - b'0');
+        if hh > 14 || mm > 59 {
+            return None;
```
David-Engel, line 707: The parsed offset is dropped for every target except `SQL_C_SS_TIMESTAMPOFFSET`, silently. msodbcsql normalizes instead:

```c
if (fSetTZValue && (fTypeOut != SQL_TIMESTAMPOFFSET_MAPPED &&
   fTypeOut != SQL_NATIVE_BCP_TIMESTAMPOFFSET && fTypeOut != SQL_C_SS_TIMESTAMPOFFSET))
{
    if (!ISBCP(lpdbc))
    {
        if (fConversionDirection == TOSERVER)
            Error = ConvertOffsetToUTC(pDateTime, &local_time);
        else
            Error = ConvertOffsetToLocal(...);
```

`SQLGetData` is `FROMSERVER`, so `'2023-01-01 12:34:56+05:30'` into `SQL_C_TYPE_TIMESTAMP` yields the client's **local** wall clock there. Probe here returns `Ok(Exact)` with `12:34` — the offset is parsed, validated, then discarded with no diagnostic.

I am not arguing for matching this: `ConvertOffsetToLocal` makes the result depend on the client machine's time zone, which is a genuinely worse contract. But the current behaviour is the one case in this PR where we silently produce a *different value* rather than a different diagnostic, so it should be a deliberate documented divergence (plan doc + a test asserting the fields are used as written) rather than an artefact of `DateTimeParts` not being consulted.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -537,10 +825,22 @@ pub(crate) unsafe fn convert_datetime_c(
             )
         },
         // Reached when the value lacks the component the target needs (e.g. a
-        // `time` column into `SQL_C_TYPE_DATE`).
-        _ => return Err(ConvError::Restricted),
+        // `time` column into `SQL_C_TYPE_DATE`). For character input the pairing
+        // is legal and it is the text that is wrong for this target, so that
+        // stays 22018 rather than becoming 07006.
+        _ => {
+            return Err(if from_character {
```
David-Engel, line 860: A time-only literal into `SQL_C_TYPE_TIMESTAMP` lands here as `22018` (probe-confirmed) because `p.has_date` is false. msodbcsql fills the date in and succeeds — end of `ParseDateTime`:

```c
if ((wIDB == ECODE_TIME) || (wIDB == ECODE_TIME2))
{   // Time only, set date to current date
    ...
    time(&local_time);
    localtime_s(&tm, &local_time);
    pDateTime->year  = tm.tm_year+1900;
    pDateTime->month = tm.tm_mon+1;
    pDateTime->day   = (SQLUSMALLINT) tm.tm_mday;
}
```

(BCP uses the ODBC base date instead; the `SQL_C_DATE` target is still rejected just above, which is the case already covered.)

This matches Appendix D, which specifies the date fields are set to the current date when a time value is converted to a timestamp — so the same gap exists on the non-character path today: a real `time` column into `SQL_C_TYPE_TIMESTAMP` is `07006` here. That half is P1 scope, not something this PR introduced, but this PR is the one titled "mandatory source-type conversions" and it is the natural place to either fix it or record it.

If it is deferred, please make it explicit in the plan doc rather than leaving it to fall out of the `_` arm — the current comment only mentions the `time` into `SQL_C_TYPE_DATE` direction, which reads as if the arm is fully intentional.

David-Engel (commented):
## Re-review — all previous findings addressed

Confirmed fixed on `2a91adb`:

- Non-ASCII panic is gone. The byte-slice + let-chain rewrite is clean, the "matched tail is all ASCII, so this boundary is a char boundary" comment justifies the one remaining `str` slice, and `non_ascii_character_input_does_not_panic` covers all three targets.
- `inf` / `NaN` text now rejected, `money_scaled` shared with `get_data.rs`, stale doc comment removed, redundant `String` arm replaced with a useful comment, allocation trade-off documented, unreachable arm annotated.
- The `SQL_C_BIT` negative case you added is a real msodbcsql behaviour I had only guessed at, and the citation is accurate.

516 unit tests pass, `cargo clippy -p mssql-odbc --all-targets` clean.

## Parity pass against msodbcsql

I read `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp` for this round. **Verified matching**, worth recording so nobody re-litigates them:

| Behaviour | msodbcsql | this PR |
| --- | --- | --- |
| Float to integer rounds or truncates | `modf(dTemp, &dTemp) != 0.` — truncate toward zero | same |
| Trailing fractional zeros (`'42.00'` to `SQL_C_SLONG`) | `if (c != '0') Error = CVT_FRACT_TRUNC` — no warning | `mantissa % divisor != 0` — no warning |
| money / smallmoney to integer | `% MONEYMULT` sets truncation, `/ MONEYMULT` is the value | `Scaled { scale: 4 }` |
| Exact vs `f64` path for character | scans for `e`/`E`, then `CharToBigint` else `CharToDouble` | `parse_decimal_literal` then `f64` fallback |
| Integer target bounds | `MIN_LIMIT_ADJUST` is `0` in the driver, so the exact target range | `try_from` |
| Negative into `SQL_C_BIT` | `CVT_PREC` via two separate guards | `is_negative()` covers both |
| Calendar validation | `ValidateDateTimeOffsetStruct` | `days_in_month` |
| Time literal to `SQL_C_TYPE_DATE`, date literal to `SQL_C_TYPE_TIME` | `CVT_CAST_ERROR` | `22018` |

The divergences I found are inline. None are regressions from the previous round except the first, which is fallout from my own suggestion — see the note on `numeric_source_or_parse`.

### Not blocking, no good inline anchor

**`SQL_C_TINYINT` bound for the newly-allowed sources.** msodbcsql caps `SQL_C_TINYINT` at `SCHAR_MAX` (127) whenever `fTypeIn` is not itself a tinyint C type (`sqlccnvt.cpp`, the `case SQL_C_TINYINT:` guard `fTypeIn != SQL_C_TINYINT && fTypeIn != SQL_C_STINYINT && fTypeIn != SQL_C_UTINYINT && ... Temp > SCHAR_MAX`). So `CAST('200' AS varchar(3))` into `SQL_C_TINYINT` is `22003` there; here it is `Ok(200)` (confirmed by probe). The existing `u8` mapping is well-reasoned for a real `tinyint` column and I would not change it, but this PR is what first opens character / decimal / float sources into that target, which is exactly the case where msodbcsql applies the signed limit. Worth a line in the plan doc so the parity leg does not surprise someone later.

**Whitespace.** `str::trim()` trims all Unicode whitespace; `FindSigNumber` / `SkipWSSequence` handle ASCII blanks. Permissive, almost certainly harmless.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Fixed in 5ff24248, and thanks for catching that my own suggestion made it worse. Rust folding overflow and garbage into the same `Ok(inf)` is exactly the trap.

```rust
let t = text.trim();
return match t.parse::<f64>() {
    Ok(f) if f.is_finite() => Ok(NumericSource::Float(f)),
    // Rust folds overflow into `Ok(inf)`, but msodbcsql's `VarR8FromStr`
    // reports DISP_E_OVERFLOW -> 22003 and keeps the cast error for text
    // that is not a number at all. Digits present means it was numeric.
    Ok(_) if t.bytes().any(|b| b.is_ascii_digit()) => Err(ConvError::OutOfRange),
    // "inf" / "infinity" / "nan" parse in Rust but are not SQL literals.
    _ => Err(ConvError::InvalidCharacterValue),
};
```

The digit test is what separates them: `'1e400'` has digits so it is an overflow (`22003`), while `inf` / `-inf` / `Infinity` / `NaN` / `nan` carry none and stay `22018`. This also covers the `parse_decimal_literal` case you noted — a 400-digit integer overflows the `i128` mantissa, falls through to the `f64` arm and now reports `22003`. Added `overflowing_numeric_text_is_out_of_range`; the existing non-finite test still passes unchanged.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Fixed in 5ff24248. You are right that my previous nit pointed the wrong way, and the 7-digit behaviour was wrong in both directions — losing precision at 9 and silently accepting 11.

`parse_time_literal` now rejects more than 9 fractional digits and normalizes to nanoseconds:

```rust
// `SQL_TIMESTAMP_STRUCT.fraction` is nanoseconds, so a character literal can
// carry 9 exact digits; msodbcsql rejects anything longer rather than
// truncating it, and a character source has no server-side scale to cap it.
if frac_digits.len() > 9 {
    return None;
}
```

So `.123456789` is now `fraction = 123456789` + `SQL_SUCCESS`, and `.12345678901` is `22018` — both matching your table. As you predicted this removed `DateTimeParts::frac_truncated` and the `p.scale` cap along with it, so the previous round's plumbing is gone entirely. Replaced the old test with `nine_fractional_digits_are_exact_and_more_is_rejected` covering both rows. The wire path is untouched, still capped by the column's own scale.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Tracked rather than fixed, as you allowed — **AB#47246**, with the `rgbECODE_DATE_SLASH` retry and the `FindECode` escape branch both cited so whoever picks it up starts from the grammar rather than rediscovering it.

I also recorded the permissive direction in the same item (the `T` separator, `HH:MM` without seconds, and unpadded fields), since the fixed-length token grammar is the reason msodbcsql rejects them and that context is easy to lose.

Your point about `SKIP_IF_COMPARING_MSODBCSQL()` is the part that worried me most, so the plan doc now carries a "Known divergences from msodbcsql" table whose preamble says explicitly that these cases are skipped on the parity leg and will not be caught by CI.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Agreed on both counts — the behaviour is right and the silence was not. Documented as a deliberate divergence in the new "Known divergences from msodbcsql" table, with the reason stated: matching `ConvertOffsetToLocal` would make the returned value depend on the client machine's time zone.

Added `offset_is_ignored_for_non_offset_targets`, which asserts `'2023-01-01 12:34:56+05:30'` into `SQL_C_TYPE_TIMESTAMP` delivers `2023-01-01 12:34:56` as written. That pins it as intended behaviour rather than an artefact of `DateTimeParts` not being consulted, which was your actual concern.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Deferred, but explicitly — **AB#47247**, and you were right that the comment read as though the whole arm was intentional. It now names both cases:

```rust
// Reached when the value lacks the component the target needs. Two
// cases land here: `time` into `SQL_C_TYPE_DATE`, which is correct, and
// `time` into `SQL_C_TYPE_TIMESTAMP`, which Appendix D says should fill
// in the current date instead (AB#47247). For character input the
// pairing is legal and it is the text that is wrong for this target, so
// that stays 22018 rather than becoming 07006.
```

The reason for deferring, recorded on the work item: it needs the current **local** date. `std::time::SystemTime` only gives UTC, so without a date crate this means a platform-specific helper (`localtime_r` on Unix, Win32 on Windows), and it makes the conversion clock-dependent, which wants an injectable clock rather than reading the system time inline. That is more design than belongs in this PR.

The work item covers the `time`-column half as well and is linked to 46578, since as you note that side is P1 behaviour.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -115,12 +105,164 @@ unsafe fn write_fixed<T: Copy>(ptr: SqlPointer, value: T, ind: *mut SqlLen) -> C
     ConvOk::Exact
 }
 
-/// Converts an integer column value to a fixed-width integer C target,
+/// A numeric column value in a form that keeps exact sources exact, so an
+/// integer target can report truncation instead of silently dropping a
+/// fraction.
+#[derive(Debug, Clone, Copy, PartialEq)]
+enum NumericSource {
```
saurabh500, line 112: Nice

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -115,12 +105,164 @@ unsafe fn write_fixed<T: Copy>(ptr: SqlPointer, value: T, ind: *mut SqlLen) -> C
...
+        ColumnValues::BigInt(x) => Some(NumericSource::Int(i128::from(*x))),
+        ColumnValues::Bit(b) => Some(NumericSource::Int(i128::from(*b))),
+        ColumnValues::Real(x) => Some(NumericSource::Float(f64::from(*x))),
+        ColumnValues::Float(x) => Some(NumericSource::Float(*x)),
+        // `DecimalParts` renders itself exactly; parse that back rather than
+        // reassembling its base-2^32 limbs. The format-and-reparse allocates per
+        // value; accepted for now in exchange for exactness.
+        ColumnValues::Decimal(d) | ColumnValues::Numeric(d) => {
+            parse_decimal_literal(&d.to_string())
+        }
```
saurabh500: This arm renders the decimal to a `String` and parses it straight back. Two problems with that.

The allocation cost is higher than the comment implies. I measured it with a counting allocator: **3-6 heap allocations per cell** (6 for negative or zero-padded values), because `Display::fmt` calls `to_decimal_string()` -- which allocates internally -- and `to_string()` then allocates again to receive the result. This is the per-cell path of a columnar fetch.

More importantly, this makes `to_decimal_string` newly reachable from the numeric targets, and that function has an unguarded shift:

```rust
// mssql-tds/src/datatypes/decoder.rs:2011
.fold(0u128, |acc, (i, &part)| acc | ((part as u32 as u128) << (i * 32)))
```

`read_decimal_data` derives the limb count from a wire length byte and caps it at **64, not 4** (`decoder.rs:670`), so `int_parts.len() > 4` is reachable from a malformed server payload. I ran that path both ways: debug builds panic with "attempt to shift left with overflow", which aborts across the `extern "C"` boundary, and release builds silently return garbage (a `[1,0,0,0,1]` payload yields mantissa `2`, because a shift of 128 wraps to 0). P1's `SQL_C_CHAR` path already reached this, so it is pre-existing rather than introduced here, but P1a adds a second call site.

The limbs are a plain base-2^32 little-endian magnitude, so reassembling them directly is exact, allocation-free, and bounds-checked:

```suggestion
        // `DecimalParts` stores a base-2^32 little-endian magnitude; reassemble
        // it directly. 38 digits fit in 4 limbs, and the wire decoder admits up
        // to 64, so reject longer payloads rather than shifting past 128 bits.
        ColumnValues::Decimal(d) | ColumnValues::Numeric(d) => {
            if d.int_parts.len() > 4 {
                return None;
            }
            let mag = d.int_parts.iter().enumerate().fold(0u128, |acc, (i, &p)| {
                acc | (u128::from(p as u32) << (i * 32))
            });
            let m = i128::try_from(mag).ok()?;
            Some(NumericSource::Scaled {
                mantissa: if d.is_positive { m } else { -m },
                scale: u32::from(d.scale),
            })
        }
```

I diffed this against the current behaviour across zero, negative zero, `123.45`, `-0.01`, the `0.005` padding branch, a 2-limb value, and the 4-limb near-10^38 maximum at both scale 0 and scale 38 -- identical results, zero allocations. With it applied, `cargo fmt`, `clippy -D warnings`, and all 537 `mssql-odbc` tests pass.

One thing this deliberately does *not* change: a genuinely out-of-range magnitude still returns `None` and surfaces as `07006` rather than `22003`, because `numeric_source` returns `Option`. Fixing that means changing the signature to `Result<NumericSource, ConvError>` -- worth a follow-up, but bigger than this change.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -433,7 +577,144 @@ pub(crate) fn is_datetime_c_target(target_type: SqlSmallInt) -> bool {
...
+/// Parses `YYYY-MM-DD`.
+fn parse_date_literal(s: &str) -> Option<(i16, u16, u16)> {
+    let mut it = s.split('-');
+    let (y, m, d) = (it.next()?, it.next()?, it.next()?);
+    if it.next().is_some() || y.len() != 4 {
+        return None;
+    }
+    let year: i16 = y.parse().ok()?;
+    let month: u16 = m.parse().ok()?;
+    let day: u16 = d.parse().ok()?;
```
saurabh500, line 626: `str::parse` accepts a leading `+`, and the `y.len() != 4` check above does not exclude it -- so `'+123-01-01'` is accepted as year 123. Month and day have no length check at all, so `'2023-+5-01'` gives month 5.

I confirmed the parsing behaviour rather than assuming it: `"+123".parse::<i16>()` is `Ok(123)` and `"+5".parse::<u16>()` is `Ok(5)`. (`-0` and leading whitespace are correctly rejected, so this is only about `+`.)

This is not in the divergence table, and I do not think it fits the "permissive about formatting" row either -- accepting `2023-6-5` is a padding relaxation, whereas accepting a sign on a year or month admits a value with no meaning.

```suggestion
    // `str::parse` accepts a leading `+`, which would make `+123-01-01` a valid
    // date; require plain digits.
    if !y
        .bytes()
        .chain(m.bytes())
        .chain(d.bytes())
        .all(|b| b.is_ascii_digit())
    {
        return None;
    }
    let year: i16 = y.parse().ok()?;
    let month: u16 = m.parse().ok()?;
    let day: u16 = d.parse().ok()?;
```

If you would rather stay permissive here, that is defensible -- but then it should be an explicit row in the divergence table rather than an accident of `str::parse`.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -433,7 +577,144 @@ pub(crate) fn is_datetime_c_target(target_type: SqlSmallInt) -> bool {
...
+    let minute: u16 = it.next()?.parse().ok()?;
+    let sec_part = it.next().unwrap_or("0");
+    if it.next().is_some() {
+        return None;
+    }
+    let (sec_digits, frac_digits) = match sec_part.split_once('.') {
+        Some((a, b)) => (a, b),
+        None => (sec_part, ""),
+    };
+    let second: u16 = sec_digits.parse().ok()?;
```
saurabh500: Same leading-`+` issue as in `parse_date_literal`, and here it affects all three components -- `'+1:00:00'`, `'01:+5:00'` and `'01:00:+5'` all parse. The fractional digits are already checked at line 636, so it is just hour/minute/second that are open.

Restructuring slightly so the three component strings are available before parsing keeps it to a single check:

```suggestion
    let hour_s = it.next()?;
    let minute_s = it.next()?;
    let sec_part = it.next().unwrap_or("0");
    if it.next().is_some() {
        return None;
    }
    let (sec_digits, frac_digits) = match sec_part.split_once('.') {
        Some((a, b)) => (a, b),
        None => (sec_part, ""),
    };
    // `str::parse` accepts a leading `+`, which would make `+1:00:00` a valid
    // time; require plain digits.
    if !hour_s
        .bytes()
        .chain(minute_s.bytes())
        .chain(sec_digits.bytes())
        .all(|b| b.is_ascii_digit())
    {
        return None;
    }
    let hour: u16 = hour_s.parse().ok()?;
    let minute: u16 = minute_s.parse().ok()?;
    let second: u16 = sec_digits.parse().ok()?;
```

Empty components still fail at `parse`, so `'.5'` and `'01:'` are rejected exactly as before. I verified fmt, clippy and all 537 tests pass with this and the date change applied together.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -559,6 +564,32 @@ TEST_F(GetDataLiveTest, UnsupportedCTypeReturnsHyc00ThenValueReadable) {
...
+    EXPECT_EQ(5, ind);
+
+    SQLCloseCursor(stmt_);
+}
+
+// An unsupported C target type is rejected with HYC00 and does not consume the
+// column. SQL_C_NUMERIC is the durable anchor for this: emitting the
+// SQL_NUMERIC_STRUCT is an explicit non-goal (decimal is delivered as character
+// data, which is what mssql-python requests), so unlike the other C targets it
+// is not scheduled to become supported.
```
saurabh500: This comment is currently the only place the permanent non-goal is recorded, and it is load-bearing -- it is the whole justification for why this test's anchor will not need to move again later.

That is worth promoting out of a test comment. Appendix D lists `SQL_C_NUMERIC` as a valid target for every numeric and character source, so a reader coming from the spec will read `HYC00` here as "not implemented yet", which is the opposite of what you mean. Every other deliberate departure in this PR is recorded in the "Known divergences from msodbcsql" table in `docs/typed-columnar-fetch-plan.md`; this is the one that is permanent, so it has the strongest claim to a row there.

```suggestion
// An unsupported C target type is rejected with HYC00 and does not consume the
// column. SQL_C_NUMERIC is the durable anchor for this: emitting the
// SQL_NUMERIC_STRUCT is a permanent non-goal, recorded in the "Known divergences
// from msodbcsql" table in docs/typed-columnar-fetch-plan.md, so unlike the
// other C targets it is not scheduled to become supported.
```

(Paired with a suggestion adding that row to the plan doc.)

mssql-odbc/docs/typed-columnar-fetch-plan.md
```diff
@@ -75,18 +75,33 @@ Both paths must share one conversion core: `ColumnValues -> requested SQL_C_* ta
...
+| Character or decimal source into `SQL_C_TINYINT` above 127 | `22003` — the signed limit applies whenever the input type is not itself a tinyint C type | `Ok(200)` — the target is `u8`, matching a real `tinyint` column (0-255) and what mssql-python fetches | **Deliberate**, but P1a is what first opens these sources into that target, so the divergence starts here. |
+| `YYYY/MM/DD` and the ODBC escape literals `{d '...'}` / `{t '...'}` / `{ts '...'}` | accepted (`rgbECODE_DATE_SLASH` retry, and the `FindECode` branch) | `22018` | Gap — Task [47246](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47246). |
+| `T` separator, `HH:MM` without seconds, unpadded fields such as `2023-6-5` | rejected (fixed-length token grammar) | accepted | Permissive. Low risk, same task. |
+| A time-only value into `SQL_C_TYPE_TIMESTAMP` | fills in the current date and succeeds, per Appendix D | `22018` from a character source, `07006` from a `time` column | Gap — Task [47247](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47247). Needs a platform-specific local-date helper, so it is not a one-line fix. |
```
saurabh500, line 104: Two divergences I hit while reviewing are not in this table, and both are the kind that only ever gets found again by someone reading the source.

The `SQL_C_NUMERIC` one matters most: it is currently documented only as a C++ comment in `get_data_test.cpp`, it contradicts Appendix D, and unlike every other gap here it is described as permanent. This table is where a user or a future maintainer would look for it.

```suggestion
| `T` separator, `HH:MM` without seconds, unpadded fields such as `2023-6-5` | rejected (fixed-length token grammar) | accepted | Permissive. Low risk, same task. |
| A time-only value into `SQL_C_TYPE_TIMESTAMP` | fills in the current date and succeeds, per Appendix D | `22018` from a character source, `07006` from a `time` column | Gap — Task [47247](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47247). Needs a platform-specific local-date helper, so it is not a one-line fix. |
| Any source into `SQL_C_NUMERIC` | converts, per Appendix D | `HYC00` | **Deliberate, and permanent.** Decimal is delivered as character data, which is what mssql-python requests, so `SQL_NUMERIC_STRUCT` is not scheduled to become supported. Anchored by `UnsupportedCTypeReturnsHyc00ThenValueReadable`. |
| A leading `+` on a date or time component (`+123-01-01`, `+1:00:00`) | rejected | accepted — `str::parse` allows the sign, so the component parses | Permissive, same Task [47246](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47246) as the grammar rows above. Drop this row if the parser is tightened instead. |
```

saurabh500:
## Review summary

Went through this one file at a time against the ODBC spec (Appendix D, "Converting Data from SQL to C Data Types") rather than only against the msodbcsql citations in the comments, and ran the suite locally: **537/537 `mssql-odbc` tests pass, clippy clean at `-D warnings`**.

**No blocking issues.** The conversion matrix is right, the SQLSTATE choices match Appendix D everywhere I checked independently, and the re-readability guarantee after a failed conversion genuinely holds (`write_captured_column` only clears `last_captured` when `rc != SQL_ERROR`).

### Findings

| # | Where | Finding |
| --- | --- | --- |
| 1 | `fetch_convert.rs:211-216` | The decimal arm format-and-reparses: **3-6 heap allocations per cell** in the columnar fetch path, and it newly exposes an unguarded `<< (i * 32)` in `to_decimal_string`. Suggestion inline; it is exact, allocation-free and bounds-checked. |
| 2 | `get_data_test.cpp:577` + plan doc | The `SQL_C_NUMERIC` **permanent** non-goal is recorded only in a C++ test comment, and it contradicts Appendix D. It belongs in the divergence table. |
| 3 | `fetch_convert.rs:604`, `:622` | Leading `+` is accepted in date and time components -- `'+123-01-01'` parses as year 123. Undocumented divergence. |

Everything else I found was cosmetic and not worth your time, so I have left it out.

### On finding 1

Worth being precise, because it is the one that could bite in production. `read_decimal_data` derives the limb count from a wire length byte and caps it at **64, not 4** (`decoder.rs:670`), so a `DecimalParts` with more than 4 limbs is reachable from a malformed server payload. `to_decimal_string` then shifts by 128 or more. I ran it both ways:

- **debug**: panics, `attempt to shift left with overflow` -- which aborts across the `extern "C"` boundary
- **release**: silently returns garbage; a `[1,0,0,0,1]` payload yields mantissa `2`

This is **pre-existing, not introduced here** -- P1's `SQL_C_CHAR` path already called into it. But P1a adds a second call site, so it seemed worth flagging now rather than filing quietly. The suggested rewrite closes it at this call site as a side effect. Happy for the underlying `to_decimal_string` fix to be a separate PR.

I checked the suggested replacement against the current code across zero, negative zero, `123.45`, `-0.01`, the `0.005` zero-padding branch, a 2-limb value and the 4-limb near-10^38 maximum at scale 0 and 38: identical results, zero allocations, and fmt/clippy/537 tests green with it applied.

### Open questions

1. **`Build Stage Build MacOS`** is red at 1h00m while `Test MacOS` passed in 36m -- that reads like a timeout rather than a real failure, but it should be green or explained before merge.
2. Is `SQL_C_NUMERIC -> HYC00` genuinely permanent, or just out of P1a? The C++ comment says permanent; if so it needs a divergence-table row (suggestion inline).
3. A character literal with no offset going into `SQL_C_SS_TIMESTAMPOFFSET` writes offset 0. Deliberate, or should it be `22018`?

### What is good here

The byte-wise offset probe in `parse_datetime_literal` is the standout -- probing the trailing `[+-]HH:MM` through `as_bytes().get(i..)` instead of slicing by character index removes a real panic on input like `N'日aaaa'`, and a panic in this crate is a process abort for the host application. That is exactly the right instinct for FFI code.

Beyond that: every parity decision cites the specific msodbcsql source file it came from, which made this reviewable in a way that "matches msodbcsql" never is; the unit tests are genuinely exhaustive rather than happy-path; the `varchar(max)` deferral is argued honestly instead of being quietly half-implemented, with the PLP invariant it would violate spelled out; and the PR corrects two claims the P1 doc made about itself rather than letting them stand. The new divergence table is the right artifact to have built -- my main note on it is that it should be more complete, which is a good problem for a doc to have.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -544,11 +544,16 @@ TEST_F(GetDataLiveTest, NvarcharMaxWideRoundTrip) {
...
-TEST_F(GetDataLiveTest, UnsupportedCTypeReturnsHyc00ThenValueReadable) {
+// Character text that is not a valid literal for a numeric C target is rejected
+// with 22018 and does not consume the column, so a follow-up call with a
+// supported type still returns the value. Before the P1a source-type
+// conversions this pairing was simply unimplemented and reported HYC00.
+//
+// Still skipped on the msodbcsql leg: msodbcsql implements this conversion and
+// its CVT_CAST_ERROR carries the "Invalid character value for cast
+// specification" message, so it very likely agrees, but the constant is spelled
+// IDS_22_005 in its source and that has not been confirmed against a live run.
```
saurabh500: This skip is **temporary** -- the comment says msodbcsql "very likely agrees" and the only thing keeping the skip is that `IDS_22_005` has not been confirmed against a live run. If it does agree, this test should stop being skipped and start comparing on both legs. But nothing in the file records that intent in a form anyone will find later.

There is already a convention for exactly this in this file -- `PlpZeroCapacityBufferDoesNotSpin` (line 819):

```cpp
// TODO(convergence): mssql-odbc will eventually adopt the msodbcsql
// truncate-and-continue contract for sub-minimal buffers [...] at which point
// this HY090 rejection goes away and the assertion can run on both legs.
```

That is greppable and states the exit condition. Worth applying the same treatment here:

```suggestion
// TODO(convergence): this skip is temporary. msodbcsql implements this
// conversion and its CVT_CAST_ERROR carries the "Invalid character value for
// cast specification" message, so it very likely agrees, but the constant is
// spelled IDS_22_005 in its source and that has not been confirmed against a
// live run. Confirm against a live msodbcsql run, then drop the skip so this
// compares on both legs; if the two do not agree, record the difference in the
// "Known divergences from msodbcsql" table in docs/typed-columnar-fetch-plan.md.
```

Marking the temporary ones also makes the permanent ones unambiguous by absence, which matters for the `SQL_C_NUMERIC` test right below this one -- that skip is never going away, and right now the two read identically.

For context, of the eight `SKIP_IF_COMPARING_MSODBCSQL()` sites in this file, five are temporary and only one carries a marker:

| Line | Test | | Marked |
| --- | --- | --- | --- |
| 174 | `BackwardColumnRejectedRereadIsNoData` | permanent (msodbcsql non-conformant) | n/a |
| 557 | `InvalidCharacterForNumericTargetIs22018ThenValueReadable` | **temporary** | no |
| 583 | `UnsupportedCTypeReturnsHyc00ThenValueReadable` | permanent (explicit non-goal) | n/a |
| 608 | `VarbinaryMaxToCharReturnsHyc00` | **temporary** (binary→char unimplemented) | no |
| 656 | `PlpColumnUnsupportedCTypeReturnsHyc00` | **temporary** (Task 47238) | no |
| 794 | `NvarcharMaxToCharChunkedAstralRoundTrip` | permanent (UTF-8 vs ANSI, Windows-only) | n/a |
| 825 | `PlpZeroCapacityBufferDoesNotSpin` | **temporary** | **yes** |
| 911 | `UnsupportedColumnTypeHyc00PreservesValue` | **temporary** (VARBINARY anchor) | no |

Only line 557 is yours; 608, 656 and 911 are pre-existing, so no objection if you would rather sweep those in a follow-up. Task 47238 and the binary work already have homes to link to.

saurabh500 (commented):
Walked the four files against Appendix D and ran the suite locally (537/537 `mssql-odbc` tests, clippy clean at `-D warnings`). **No blocking findings** -- the conversion matrix and SQLSTATE choices hold up. Leaving this as a comment rather than an approval only because the macOS build leg is red and the questions in my summary comment are still open; happy to flip to approve once those settle.

Six comments, each with an applyable suggestion. The three that matter:

1. **`fetch_convert.rs:211`** -- the decimal arm format-and-reparses (3-6 allocations per cell), and newly exposes an unguarded `<< (i * 32)` in `to_decimal_string` that panics in debug / corrupts in release for >4-limb payloads. Pre-existing, but this PR adds a second call site.
2. **`get_data_test.cpp:577` + plan doc** -- the `SQL_C_NUMERIC` permanent non-goal lives only in a C++ comment and contradicts Appendix D.
3. **`fetch_convert.rs:604`, `:622`** -- leading `+` accepted in date/time components (`'+123-01-01'` parses as year 123).

Plus a request for `TODO(convergence):` markers on the temporary `SKIP_IF_COMPARING_MSODBCSQL()` sites, following the convention already used at line 819.

One correction to my summary comment, on the divergence table: it says these are recorded because "`GetDataLiveTest` skips the msodbcsql comparison leg for these cases." That is not quite what is happening -- none of the five table rows has a `GetDataLiveTest` case at all (there is no `SQL_C_TINYINT` or `SQL_C_SS_TIMESTAMPOFFSET` anywhere in the e2e suite). A skipped test still asserts on the Rust leg; a missing one pins nothing, so those five are covered only by Rust unit tests, which cannot see msodbcsql. Not worth holding the PR for, but the sentence is misleading as written.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Adopted in 69487db9, essentially verbatim. This was the most valuable comment of the round — I verified each link rather than taking it on trust, and all three hold:

- `MAX_DECIMAL_INT_PARTS` is 64, not 4 (`decoder.rs:670`), so `int_parts.len() > 4` is reachable from the wire.
- The fold shifts by `i * 32` with no bound, so `i == 4` is a 128-bit shift on a `u128`.
- I compiled the shift in isolation: a `[1,0,0,0,1]` payload yields `1` under wrapping semantics, matching your release-build finding, and a checked shift overflows.

So this is a server-data-driven panic across the `extern "C"` boundary, which is the same class as the non-ASCII slice panic from the previous round. Worth blocking on, and I have treated it that way.

I also took your `|` over the existing `+`, since the limbs are disjoint and the add was an overflow candidate in its own right. Added `decimal_limbs_are_reassembled_and_bounded`, covering positive, negative, and the 5-limb refusal.

Two things your comment scoped out, now filed rather than left in a thread:

- **#234** — the underlying `to_decimal_string` bug. Removing this call site does not fix it; the `SQL_C_CHAR` path in `column_value_to_text` still reaches it, so a reachable panic is still shipping. The issue suggests bounding at the assembly site and tightening `MAX_DECIMAL_INT_PARTS` to 4 at the wire, and credits you.
- **#235** — the `07006` vs `22003` mapping. Agreed it needs the `Result<NumericSource, ConvError>` signature change and does not belong in this PR.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Fixed in 69487db9, taking your suggestion. I confirmed the behaviour independently: `"+123".parse::<i16>()` is `Ok(123)`, and the `y.len() != 4` guard does not exclude it because `+123` is four bytes.

Your distinction is the part that decided it for me — accepting `2023-6-5` is padding latitude, whereas accepting a sign on a year admits a value with no meaning. Those are different things, so this is tightened rather than documented, and I have deliberately **not** added the leading-`+` row you drafted for the divergence table.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Fixed in 69487db9 with your restructuring, so the three component strings are checked in one pass. `'+1:00:00'`, `'01:+5:00'` and `'01:00:+5'` are now `22018`.

Confirmed your note that empty components still fail at `parse`, so `'.5'` and `'01:'` are rejected exactly as before. Both the date and time cases are covered by `leading_plus_in_date_or_time_component_is_rejected`.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
David-Engel: Agreed, applied in 69487db9. The point that a reader coming from Appendix D will read `HYC00` as "not implemented yet" — the opposite of what the comment means — is exactly why it should not have been living in a test file. The comment now points at the divergence table, and the table carries the reasoning.

mssql-odbc/docs/typed-columnar-fetch-plan.md
David-Engel: Added the `SQL_C_NUMERIC` row in 69487db9, worded as you drafted it.

I did **not** add the leading-`+` row, because I took your other two suggestions and tightened the parser instead — which you called out as the alternative. So that divergence no longer exists rather than being documented.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
David-Engel: Applied in 69487db9. The argument that marking the temporary skips is what makes the permanent ones unambiguous *by absence* is the convincing part — that property is what the `SQL_C_NUMERIC` test immediately below depends on, and right now the two do read identically.

Your survey of the eight sites was useful; I checked it and it matches what is in the file. I have only marked line 557, which is the one this PR introduced. I have deliberately left 608, 656 and 911 alone to keep this PR's diff to its own changes, taking you up on the follow-up option — with the caveat that until those three are marked, absence does not yet imply permanent. I will sweep them separately so the convention actually holds, rather than leaving it half-applied.

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -795,8 +1120,624 @@ mod tests {
...
+        unsafe {
+            convert_integer_c(
+                &decimal(false, 2, vec![12345]),
+                SQL_C_SLONG,
+                (&mut out as *mut i32).cast(),
+                &mut ind,
+            )
+        }
+        .unwrap();
+        assert_eq!(out, -123);
```
saurabh500, line 1444: The new limb-reassembly test does not exercise reassembly. Every decimal constructed in this file is a single limb -- `vec![12345]` here, and `dec("12345.6789")` / `dec("99999.5")` elsewhere -- all below 2^32. So `i` never exceeds 0, and the shift, the `|`, and the `as u32` reinterpretation, which are the substance of this commit, are never run with a value that could distinguish them from a no-op.

I checked rather than assumed: changing `<< (i * 32)` to `<< (i * 16)` at line 216 leaves all 539 tests green.

The 5-limb refusal below is genuinely discriminating, so the guard is covered. It is the reassembly that is not.

One case closes it, reusing the wire vector `mssql-tds` already pins in `test_f64_conversion` (`decoder.rs:2205`), which keeps the two crates agreeing on the encoding: `-539_269_688 as u32` is 3755697608, plus `2 << 32` is 12345632200, which at scale 5 is 123456.322 and truncates to 123456.

I validated the suggestion before posting: it passes as written, it is rustfmt-clean, clippy is clean, and it fails under both mutations -- the wrong shift yields 37556 instead of 123456, and replacing `p as u32` with a sign-extending cast also fails.

```suggestion
        assert_eq!(out, -123);

        // Two limbs, the low one negative as `i32`: the shift, the `|` and the
        // `as u32` reinterpretation all have to be right to land on 123456.322.
        // Same wire vector `mssql-tds` pins in `test_f64_conversion`.
        let ok = unsafe {
            convert_integer_c(
                &decimal(true, 5, vec![-539_269_688, 2]),
                SQL_C_SLONG,
                (&mut out as *mut i32).cast(),
                &mut ind,
            )
        }
        .unwrap();
        assert_eq!(ok, ConvOk::Truncated);
        assert_eq!(out, 123_456);
```

saurabh500 (commented):
Re-reviewed at 69487db9. All six comments from the last round are addressed, and I checked each against the code rather than the replies.

The limb reassembly is correct at every boundary I could think to try: negative `i32` limbs reinterpret correctly, 38 nines round-trips in both signs, and a 4-limb magnitude at or above 2^127 returns `None` through `i128::try_from` rather than wrapping to a negative. `|` is safe because the limbs occupy disjoint 32-bit windows. The `as u32` cast matches what the decoder already does, so the two crates agree on the encoding.

The leading-`+` tightening does not cost anything legitimate. `2023-6-5` is still accepted, so the documented unpadded-field divergence survives; a whole-literal `-` still takes its separate path; empty components still fail at `parse` as before; and the fractional component was already gated, so the new check leaves no gap.

Both new tests fail if their guard is reverted, which is what I care about. `cargo fmt`, `cargo clippy -D warnings` and 539/539 `nextest` are all clean locally.

Not adding the leading-`+` divergence row was the right call -- tightening the parser removes the divergence rather than documenting it, so there is nothing left to record. Marking only this PR's own skip and sweeping the rest separately is also fine by me, and your count of three remaining temporary skips matches my survey; 174 and 797 are permanent and should stay unmarked.

One finding left, on the new test. It is coverage rather than correctness, so it is not merge-blocking from my side. Everything shipping here looks right to me.

Leaving this as a comment rather than an approval only because the build legs are still running on this commit.

saurabh500:
I have one last comment after which this code can be approved.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Applied verbatim in d4faacfe. You are right, and the way you demonstrated it is the part worth calling out: a test named for limb reassembly that stays green when the shift is changed to `i * 16` was not testing reassembly at all. Every decimal in the file was a single limb below 2^32, so `i` never left 0.

I re-ran your mutations rather than taking the result on trust, and both now fail:

| Mutation | Result |
| --- | --- |
| `<< (i * 32)` → `<< (i * 16)` | `left: 37556, right: 123456` — exactly the value you predicted |
| `u128::from(p as u32)` → sign-extending `p as i128 as u128` | fails |
| restored | passes |

So the case is genuinely discriminating on all three things the commit changed: the shift amount, the reinterpretation, and (via the two-limb sum) the `|`.

Reusing the vector `mssql-tds` pins in `test_f64_conversion` was a good call for a reason beyond convenience — if the two crates ever disagree about limb encoding, that shared vector is what will catch it, rather than each side testing its own private assumption.

Thanks also for confirming the `2^127` boundary and the `-` separator behaviour independently; the whole-literal `-` path was the one I would have most expected the digit check to break, and it is good to have that verified rather than assumed.

saurabh500 (approved):

PR #221, opened 2026-08-11
Remove duplicate test-only PacketReader implementation

### LINKED ISSUE #219: `PacketReader` is test-only and diverges from the real read path in `NetworkTransport`
Found while profiling the ODBC row-fetch path (#215).

## Summary

`mssql-tds/src/io/packet_reader.rs` defines `PacketReader`, but its constructor is gated:

```rust
#[cfg(test)]
pub(crate) fn new(network_reader_writer: &'a mut dyn NetworkReaderWriter) -> PacketReader<'a> {
```

So `PacketReader` **cannot be constructed in a release build**. Every call site of `PacketReader::new` is inside a `#[cfg(test)]` module (`packet_reader.rs`, `col_info_parser.rs`, `tab_name_parser.rs`, `error_parser.rs`, `info_parser.rs`, `order_parser.rs`).

The read path that actually runs is `NetworkTransport::get_new_tds_packet` (`connection/transport/network_transport.rs`), confirmed empirically: instrumenting `NetworkTransport::receive` — the method `PacketReader` calls — yields **0 calls** during a row-fetch benchmark.

The concern isn't dead code. It's that a set of token-parser tests are validating parsing against a packet reader with **different semantics** from the one used in production, so those tests can pass while the real path is broken (or vice versa).

## The concrete divergence

`PacketReader::get_new_tds_packet` (lines 146–197) returns `bytes_read_from_transport`, not `packet_size_from_header`:

```rust
let packet_size_from_header: usize = length_from_packet_header as usize;

// Keep reading until we have the complete packet in memory.
while bytes_read_from_transport < packet_size_from_header {
    bytes_read_from_transport += self.network_reader_writer.receive(...).await?;
}
...
Ok(bytes_read_from_transport)
```

The loop stops once it has *at least* a full packet, but a single `receive()` can return bytes past the packet boundary — i.e. the head of the next packet. The returned count then includes those bytes.

The caller (lines 122–135) treats the whole returned range as one packet and strips exactly one 8-byte header:

```rust
let new_packet_size = self.get_new_tds_packet().await?;
self.working_buffer.copy_within(
    self.buffer_length + 8..self.buffer_length + new_packet_size,
    self.buffer_length,
);
self.buffer_length += new_packet_size;
self.buffer_length -= 8;
```

If `bytes_read_from_transport > packet_size_from_header`, the next packet's 8-byte header is left inline and interpreted as payload, corrupting the token stream.

`NetworkTransport::get_new_tds_packet` handles this correctly by carrying the overshoot in `pending_bytes`. `PacketReader` has no equivalent.

This is latent today only because the mock readers in tests return at most one packet per `receive()`. It would surface immediately against a real socket.

## Suggested resolution

Either is fine; the second is probably better:

1. **Fix it** — track the overshoot and return `packet_size_from_header`, mirroring `pending_bytes`.
2. **Delete it** and port the token-parser tests onto the `NetworkTransport` path (or a shared abstraction), so parser tests exercise the reader that actually ships.

Option 2 removes the divergence rather than duplicating a subtle invariant in two places.

## Notes

- No impact on shipped behavior — the type is unreachable in a release build.
- Not urgent; filing so it isn't rediscovered.

saurabh500:
## Description

`PacketReader` was a second TDS packet reader that only existed for tests — its constructor was `#[cfg(test)]`-gated, so it was unreachable in release builds — and it had different packet-boundary semantics than the production reader, `NetworkTransport`.

Specifically, `PacketReader::get_new_tds_packet` returned `bytes_read_from_transport` instead of `packet_size_from_header`. Its caller then stripped exactly one 8-byte header from that whole range, so whenever a single `receive()` overshot one packet, the next packet's header stayed inline as payload. `NetworkTransport` handles the overshoot correctly via `TdsReadBuffer::pending_bytes`.

Per the issue's preferred resolution (option 2), this deletes the duplicate rather than fixing it, so token-parser tests exercise the same reader that ships:

- Delete the `PacketReader` struct, its inherent impl, and its `TdsPacketReader` impl. The `TdsPacketReader` **trait** is untouched — it's used throughout.
- Move `PacketReader::LENGTHNULL` to a module-level `LENGTH_NULL` constant in `io/packet_reader.rs`.
- Add a `test_packet_support` module (`TestPacketBuilder` plus duplex-stream–backed `NetworkTransport` factories) and port all 15 reader tests plus the 15 parser tests (`col_info`, `tab_name`, `order`, `info`, `error`) onto it. Added one test for the `LENGTH_NULL` path since that constant moved.
- Fix `TestPacketBuilder::build`, which wrote only the payload length into the header's length field. TDS requires the *total* length; `PacketReader` masked the bug by reporting bytes actually read, whereas `NetworkTransport` treats the header as authoritative.
- Delete `MockNetworkReaderWriter`, `NetworkReader::receive`, and `NetworkTransport::receive`, all of which lost their last real callers with `PacketReader`. Writer-side tests in `sql_tvp.rs` / `sqltypes.rs` move to the existing `packet_writer::tests::MockNetworkWriter`.

## Production bug fixes

Pointing the tests at `NetworkTransport` exposed three real defects in the shipping reader. They are **not** refactoring noise — each is a behavior change on the production path, has a regression test, and is recorded under `### Fixed` in `CHANGELOG.md`:

1. **Values straddling a packet boundary.** The 13 fixed-width readers checked for buffered data with `if !do_we_have_enough_data(n)` and read a *single* further packet. A value can span more than two packets, and a packet can carry fewer bytes than the value needs, so the read proceeded against a still-short buffer — `get_slice()` is not bounded by `buffer_length`, so it returned stale bytes and then `consume_bytes` panicked. Now a `while` loop that reads until the whole value is buffered.
2. **`<< 1` truncation in the varchar readers.** `read_varchar_u8_length` truncated strings of ≥128 characters and `read_varchar_u16_length` strings of ≥32768, because the character count was doubled *before* being widened (`(length << 1) as usize`), overflowing the narrow type — a 200-character string asked for 144 bytes. Now `(length as usize) << 1`.
3. **Payload-free non-EOM packets.** `get_new_tds_packet` accepted a packet whose header length equals the header size. Such a packet is malformed — it neither carries payload nor terminates a message — and is now a protocol error. Empty *end-of-message* packets remain legal, so this does not bound the new `while` loops: an empty EOM packet is payload-free too and is accepted by design, leaving that case to the request timeout, as before.

Net: ~350 lines deleted, plus the three fixes above.

## Related Issues

Fixes #219

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes (also `.\scripts\bclippy.ps1`, covering `mssql-py-core`)
- [x] `cargo btest` passes — 1680/1687 `mssql-tds` lib tests pass; the 7 failures (`certificate_validator`, `win_tls::validate`) are pre-existing on `main` and unrelated. Integration suites need a live SQL Server.
- [x] Public API changes are documented — no public API change; everything touched is `pub(crate)` or test-only

Also verified the crate still builds under `--cfg fuzzing`.


mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -633,6 +629,10 @@ impl NetworkTransport {
         self.encryption = Some(encryption);
     }
 
+    /// Raw byte-level read. [`get_new_tds_packet`](Self::get_new_tds_packet)
+    /// reads straight into the working buffer instead, so this is exercised
+    /// only by tests and by [`NetworkReader`] consumers.
+    #[cfg_attr(not(test), allow(dead_code))]
```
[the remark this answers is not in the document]
saurabh500: Good catch on both points. Gated the helper with `#[cfg(test)]` — both remaining callers (`io::reader_writer::tests::test_send_recv` and `network_transport::tests::test_network_transport_receive`) are in `cfg(test)` modules, so it no longer compiles into release builds at all. Also rewrote the doc comment to drop the stale `NetworkReader` reference and point at `get_new_tds_packet` as the production read path, and narrowed the now test-only `UnexpectedEof` import.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1639,6 +1641,28 @@ pub(crate) mod tests {
         )
     }
 
+    /// Builds a `NetworkTransport` whose read side is pre-loaded with `data`.
+    /// The writer half is dropped, so reads observe EOF once `data` is drained.
+    pub(crate) async fn create_network_transport_with_data(data: &[u8]) -> NetworkTransport {
+        let (client_side, mut server_side) = duplex(data.len().max(1));
```
saurabh500: The helper can't simulate fragmented reads — which is exactly the failure class this PR is about.

`duplex(data.len().max(1))` sizes the pipe to the payload and writes it in one shot, so every `read()` returns the whole buffer. This structurally cannot produce a short read, a split header, or a truncated packet. Given the motivating bug was packet-boundary handling, that's the one thing worth being able to express.

An optional chunked variant would unlock it:

```rust
/// Feeds `data` to the transport in `chunk_size` pieces, so reads observe
/// packet-header and payload splits the way a real socket can.
pub(crate) fn create_network_transport_with_chunked_data(
    data: &[u8],
    chunk_size: usize,
) -> NetworkTransport {
    let (client_side, mut server_side) = duplex(chunk_size.max(1));
    let owned = data.to_vec();
    tokio::spawn(async move {
        for chunk in owned.chunks(chunk_size.max(1)) {
            if server_side.write_all(chunk).await.is_err() {
                return;
            }
        }
    });
    /* ...same NetworkTransport::new(...) construction... */
}
```

Two cheap tests this enables, both covering paths the old `Ok(0)`-at-EOF mock could never reach:

- header split across reads (`chunk_size = 3`) -> still parses correctly;
- truncated packet (header claims 200 bytes, supply 100) -> `Err(ConnectionClosed)` rather than the infinite spin the old mock produced.

Separately: nothing currently drives **two** packets through this helper. `test_get_new_tds_packet_handles_multiple_packets_in_single_read` covers coalescing at the transport level, but nothing proves a *parser* reading across a packet boundary works — which is precisely what the deleted reader got wrong. Concatenating two `TestPacketBuilder` outputs and reading a value that straddles the boundary would close the loop on #219's stated concern.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1639,6 +1641,28 @@ pub(crate) mod tests {
         )
     }
 
+    /// Builds a `NetworkTransport` whose read side is pre-loaded with `data`.
+    /// The writer half is dropped, so reads observe EOF once `data` is drained.
```
saurabh500: Nit: worth documenting the size ceiling here.

The transport is built with `ClientContext::default().packet_size` (8000), and `get_new_tds_packet` hard-rejects `packet_size_from_header > max_packet_size`. A future test with a payload over ~7992 bytes will fail with `"TDS packet length N exceeds negotiated max packet size 8000"`, which won't obviously point back to this helper.

```suggestion
    /// Builds a `NetworkTransport` whose read side is pre-loaded with `data`.
    /// The writer half is dropped, so reads observe EOF once `data` is drained.
    /// `data` must be a single packet of at most 8000 bytes (the default
    /// negotiated packet size), or `get_new_tds_packet` will reject it.
```

saurabh500 (commented):
## Review: no blocking issues

I verified the core claim independently rather than taking the description at face value, and it holds up.

**The deletion is genuinely not a public API change.** `PacketReader::new` was `#[cfg(test)]`-gated and `lib.rs` declares `pub(crate) mod io`, so `pub struct PacketReader` was never externally reachable. Nothing was widened to `pub` to satisfy tests: `LENGTH_NULL` is `pub(crate)`, `create_network_transport_with_data` is `pub(crate)` inside a `#[cfg(test)]` module, and `TdsPacketReader` was already `pub(crate)` on `main`.

**No test coverage was lost.** All 15 reader tests and all 15 token-parser tests were migrated 1:1 with assertions intact, and one test was added (`test_read_varchar_u16_length_null`). The two renames (`test_read_varchar_with_byte_length` -> `test_read_varchar_u16_length`, `test_read_u8_varchar` -> `test_read_varchar_u8_length`) are strictly more accurate, as is the `utf16_byte_len` -> `char_count` rename in `test_read_unicode`. No dangling references to `PacketReader`, `MockNetworkReaderWriter`, `LENGTHNULL`, or `read_tds_packet_for_test` remain.

**Every behavioral delta is an improvement, not a regression:**

| Aspect | Deleted `PacketReader` | `NetworkTransport` (now used) |
|---|---|---|
| Packet boundary | Returned `bytes_read_from_transport`; overshoot left the next header inline as payload | Returns `packet_size_from_header`; overshoot carried in `pending_bytes` |
| EOF | `MockNetworkReaderWriter::receive` returned `Ok(0)`, so `while bytes_read < packet_size` **spun forever** on truncated data | `bytes_read == 0` -> `ConnectionClosed`, sets `known_dead` |
| Header validation | None | Rejects len `< 8`, len `> max_packet_size`, and `base_offset + len > buffer.len()` |
| `max_packet_size` in tests | 4096 (mock) | 8000 (`ClientContext::default()`) |

**The `TestPacketBuilder::build` fix is load-bearing, not cosmetic.** Without it, `test_read_bytes` (2000-byte payload) yields `packet_size_from_header = 2000` against 2008 bytes available, so `pending_bytes = 8` and the last 8 payload bytes get reinterpreted as the next packet header. Good catch.

**`#[cfg(test)]` on `receive` is safe.** Only two callers remain (`network_transport.rs:1779`, `reader_writer.rs:114`), both inside `#[cfg(test)] mod tests`, and it is `pub(crate)`, so there is no cross-crate or `feature = "test-util"` exposure. `ErrorKind` is still used in non-test code, so the import change doesn't create an unused-import warning under `-D warnings`.

I left inline suggestions on the specific lines. All are non-blocking.

---

### Two pre-existing issues worth a follow-up (not introduced here, out of scope)

**1. A 510-byte allocation guard leaves the codebase with this deletion.** The removed `PacketReader::read_unicode_with_byte_length` had:

```rust
const MAX_STRING_BYTE_LENGTH: usize = u8::MAX as usize * 2;
if byte_length > MAX_STRING_BYTE_LENGTH { return Err(...); }
```

`NetworkTransport::read_unicode_with_byte_length` has no equivalent and does `vec![0; byte_length]` from a wire-supplied length. This is **not** a regression — the guard only ever ran in the test-only reader and predates this PR (it arrived with the `read_write` -> `io` rename). But it deserves an explicit decision rather than vanishing unremarked: if it was an intended OOM mitigation for the production reader, file a follow-up; if it was arbitrary, deleting it is correct.

**2. `(length << 1)` truncation in the varchar readers.** `read_varchar_u8_length` does `(length << 1) as usize` on a `u8`: for `length = 200`, `200u8 << 1 == 144` (Rust discards shifted-out bits and only panics on shift-*amount* overflow), so a 200-char varchar reads 144 bytes and corrupts the stream. Same shape in `read_varchar_u16_length` above 32767. Identical in the deleted reader, so not a regression — but since this PR is adding the tests for exactly these methods, a `"a".repeat(200)` case would surface it for nearly free.

---

### CI status

Not green yet on `30f17f8` — the substantive checks are still **pending**. CodeQL, Analyze (actions), `check`, and Evaluate PR duplicate pass, but those are all lightweight (CodeQL finished in 3s, `check` in 2s) and don't compile or test the crate. Build (Linux/Linux ARM/macOS/Windows/Windows ARM), Test macOS, SQL host, Kerberos, the cross-repo mssql-python builds, and `coverage-report` are all outstanding. No failures so far.

`coverage-report` is the one to watch against the 85% bar: this PR deletes ~900 lines of test-only code, which perturbs the denominator in a way the diff alone doesn't predict. Worth re-checking before undrafting.

The PR description matches the diff on every point I checked, including the subtle ones (`pending_bytes` rationale, the `build()` masking explanation, "no public API change"), and it correctly implements option 2 from #219 rather than quietly picking option 1.

mssql-tds/src/io/packet_reader.rs
```diff
@@ -673,455 +226,17 @@ pub(crate) mod tests {
         append_method!(append_u64, u64, 8, write_u64);
 
         pub(crate) fn build(&mut self) -> Vec<u8> {
-            BigEndian::write_u16(&mut self.data[2..4], self.payload_length);
+            BigEndian::write_u16(&mut self.data[2..4], self.payload_length + 8);
```
saurabh500: This fix is correct and necessary, but it's a silent semantic change to a helper shared by 5 parser test modules plus the transport tests — with no test pinning the behavior.

If someone later "corrects" this back to `payload_length`, every affected test fails in a confusing way far from the cause. Worth a direct assertion:

```rust
#[test]
fn test_packet_builder_writes_total_length_in_header() {
    let mut builder = TestPacketBuilder::new(PacketType::PreLogin);
    builder.append_bytes(&[0u8; 12]);
    let packet = builder.build();
    assert_eq!(packet.len(), 20);
    assert_eq!(
        BigEndian::read_u16(&packet[2..4]),
        20,
        "header must carry total length, not payload length"
    );
}
```

Nit while you're here: `self.data` already *is* header + payload, so the parallel `payload_length` counter and the `+ 8` are both avoidable, and `payload_length + 8` can overflow `u16` in debug builds:

```rust
pub(crate) fn build(&mut self) -> Vec<u8> {
    let total = u16::try_from(self.data.len()).expect("test packet exceeds u16 length");
    BigEndian::write_u16(&mut self.data[2..4], total);
    self.data.clone()
}
```

(That would leave `payload_length` written-but-never-read, so it'd need to come out of the struct too.)

mssql-tds/src/io/reader_writer.rs
```diff
@@ -41,7 +41,6 @@ pub(crate) trait NetworkWriter: Send + Sync + TransportSslHandler {
 
 #[async_trait]
 pub(crate) trait NetworkReader: Send {
-    async fn receive(&mut self, buffer: &mut [u8]) -> TdsResult<usize>;
     fn packet_size(&self) -> u32;
```
saurabh500, line 44: With `receive` gone, `NetworkReader` is now vestigial:

```rust
#[async_trait]
pub(crate) trait NetworkReader: Send {
    fn packet_size(&self) -> u32;
}
```

Its sole remaining method duplicates `NetworkWriter::packet_size`, which is why `network_transport.rs:1133` needs UFCS (`NetworkReader::packet_size(self)`) to disambiguate. `#[async_trait]` is now a no-op on it, and four types implement it only to return a constant.

Folding `packet_size` into `NetworkWriter` and deleting `NetworkReader` would remove the UFCS workaround and three trivial impls. Reasonable as a follow-up rather than in this PR — flagging so it doesn't get forgotten.

mssql-tds/src/token/parsers/col_info_parser.rs
```diff
@@ -32,17 +32,15 @@ where
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::io::packet_reader::PacketReader;
-    use crate::io::packet_reader::tests::{MockNetworkReaderWriter, TestPacketBuilder};
+    use crate::connection::transport::network_transport::tests::create_network_transport_with_data;
```
saurabh500: This import inverts the documented layering.

Token-layer tests now depend on `crate::connection::transport::network_transport::tests`, i.e. Token -> Transport, against the repo's stated `Transport -> IO -> Token stream -> Message -> Client API` direction. Previously these tests depended on `io::packet_reader`, the layer directly beneath them. The same import lands in `error_parser.rs`, `info_parser.rs`, `order_parser.rs`, and `tab_name_parser.rs`.

Consider hosting the helper somewhere neutral — `io::packet_reader::tests` alongside `TestPacketBuilder` (which these tests already import), or a small crate-level test-support module — so parser tests don't reach up into transport internals.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1296,7 +1296,7 @@ impl TdsPacketReader for NetworkTransport {
 
     async fn read_varchar_u16_length(&mut self) -> TdsResult<Option<String>> {
         let length: u16 = self.read_uint16().await?;
-        if length == PacketReader::LENGTHNULL {
+        if length == LENGTH_NULL {
```
saurabh500, line 1274: Not introduced by this PR, but flagging it here since this impl block is where the ported tests now land, and the new tests only exercise 12-character strings.

`read_varchar_u8_length` (just below) does `(length << 1) as usize` on a `u8`: for `length = 200`, `200u8 << 1 == 144`, because Rust discards shifted-out bits and only panics on shift-*amount* overflow. So a 200-char varchar reads 144 bytes and silently corrupts the stream. `read_varchar_u16_length` has the same shape above 32767.

The deleted reader had the identical bug, so this is not a regression — but a `"a".repeat(200)` case in the new tests would surface it for nearly free. At minimum worth a linked follow-up issue.

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Added `create_network_transport_with_chunked_data`, which spawns a writer into a `duplex(chunk_size)` so the reader is forced to drain one chunk at a time. On top of it:

- `test_packet_header_split_across_reads` feeds a packet in 3-byte chunks, splitting the 8-byte header across three reads.
- `test_truncated_packet_reports_error` claims 200 bytes in the header but supplies 100, and asserts the reader errors instead of spinning.
- `test_read_value_spanning_packet_boundary` (and a `_fragmented` variant) concatenates two `TestPacketBuilder` packets with a `u32` split 2/2 across the boundary.

The cross-boundary test caught a real bug: the fixed-width readers called `read_tds_packet()` once and then unconditionally consumed N bytes, so a value straddling a boundary hit `panic!("Not enough data to consume")` in `TdsReadBuffer::consume_bytes`. The deleted reader had the identical shape (it returned a `ProtocolError` rather than panicking), so this is pre-existing rather than a regression from the consolidation — but it is now the only reader, so I fixed it here: those 13 readers use `while !do_we_have_enough_data(n)` instead of `if`. It terminates because each iteration performs a network read that EOFs into `ConnectionClosed`.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Documented on the helper. `ClientContext::default().packet_size` is 8000 and `get_new_tds_packet` rejects any packet whose header length exceeds it, so each packet passed to the helper must be <= 8000 bytes total (<= 7992 payload). The doc comment now states that ceiling and the single-read limitation.

> [!NOTE]

mssql-tds/src/io/packet_reader.rs
saurabh500: Done. Dropped the `payload_length` field entirely and derived the total from `self.data.len()`, which also removes the `payload_length + 8` overflow path (it now goes through `u16::try_from(...).expect(...)`).

Pinned the behaviour with two tests: `test_packet_builder_writes_total_length_in_header` asserts the header length equals the full packet length rather than the payload length, and `test_packet_builder_empty_payload_length_is_header_only` covers the 8-byte edge case. This matters because `NetworkTransport` treats the header length as authoritative and validates it is `>= 8`, whereas the deleted `PacketReader` tolerated the old payload-only value.

> [!NOTE]

mssql-tds/src/io/reader_writer.rs
saurabh500: Agreed, and taking your framing that this is follow-up rather than part of this PR — it would touch every `NetworkWriter`/`NetworkReader` implementor, which is a wider blast radius than a test-consolidation change should carry.

Recording the deferred work so it is on the record: fold `packet_size()` into `NetworkWriter`, delete the `NetworkReader` trait and its `#[async_trait]` (now a no-op, since the trait has no async methods left), and drop the resulting UFCS call at `network_transport.rs:1133` back to a plain method call.

> [!NOTE]

mssql-tds/src/token/parsers/col_info_parser.rs
saurabh500: Fixed. Moved `create_network_transport_with_data` down to `io::packet_reader::tests`, so all five parser test modules now import it alongside `TestPacketBuilder` from one place and no longer reach into `connection::transport`. That keeps the token layer depending only on the IO layer directly beneath it, with IO -> Transport following the documented direction.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Good catch — fixed in both readers: `(length << 1) as usize` is now `(length as usize) << 1`.

You are right that it is not a regression (the deleted reader had it too), but I fixed it here rather than deferring, because the shift happens on the `u8`/`u16` before widening, so `200u8 << 1` is `144` and the reader would consume 144 bytes for a 400-byte string and silently desynchronize the stream for everything after it. Rust only panics on shift-*amount* overflow, so there is no debug assertion to catch it.

Added `test_read_varchar_u8_length_long_string`, which round-trips `"a".repeat(200)` and fails without the fix. The `u16` path has the same shape above 32767, but a string that long exceeds the 8000-byte max packet size, so it is not reachable in a single-packet test.

> [!NOTE]

saurabh500:
Coming back to the one item from your review summary I had not answered — the deleted allocation guard. You asked for an explicit decision rather than letting it vanish unremarked, so here it is.

> if it was an intended OOM mitigation for the production reader, file a follow-up; if it was arbitrary, deleting it is correct.

**It was worse than arbitrary — it was incorrect, and deleting it is correct.**

The deleted `read_varchar_u16_length` fed the guard directly:

```rust
let length: u16 = self.read_uint16().await?;
...
self.read_unicode_with_byte_length((length << 1) as usize).await?
```

With `MAX_STRING_BYTE_LENGTH = u8::MAX * 2 = 510`, any `length >= 256` produces `byte_length >= 512` and gets rejected with a `UsageError`. So that reader would have failed on any legitimate `u16`-length-prefixed varchar longer than 255 characters. Nothing caught it because the reader was test-only and no test exercised a string that long. Porting the guard to `NetworkTransport` would have introduced that failure into the production path.

On whether the underlying concern needs a follow-up: I do not think so. Every production call site is already bounded by its wire type, and the allocation is immediately followed by `read_bytes`, which has to actually pull that many bytes off the socket or error out — so there is no amplification, a server must send what it claims.

- `read_varchar_u8_length` — `u8`, max 510 bytes.
- `read_varchar_u16_length` — `u16`, max 131070 bytes (~128 KB), and in practice far less since `max_packet_size` is 8000.
- `returnvalue_parser.rs:41` — already guards with `checked_mul(2)` on a `u8`.
- `read_unicode` — carries `#[allow(dead_code)]`; no production caller.

`fuzz_support.rs` also already applies `MAX_ALLOC` to this exact method, which is where the adversarial-input bound belongs.

Your second follow-up item, the `(length << 1)` truncation, I did fix in this PR rather than defer — details in the inline thread.

David-Engel (commented):
## Review

Solid refactor — no correctness bugs found. Verified locally: `cargo clippy -p mssql-tds --all-features --all-targets -- -D warnings` is clean, and lib tests are 1680 passed / 7 failed, where all 7 are the pre-existing `certificate_validator` and `win_tls::validate` fixture failures unrelated to this diff. I also confirmed no bare `PacketReader` struct references remain anywhere (only the `TdsPacketReader` trait), and the two remaining `.receive(` call sites are both `cfg(test)`.

### Medium

**`NetworkTransport::receive` is now kept alive only for tests** — `network_transport.rs:635`

`receive` is `#[cfg(test)]`, and its only callers are `test_network_transport_receive` (`network_transport.rs:1715`) and `test_send_recv` (`reader_writer.rs:96`). Those two tests now exercise code that never ships, which is the same pattern this PR is deleting. It also duplicates the `known_dead` EOF/error bookkeeping in `get_new_tds_packet` (`network_transport.rs:926`), so a future change to liveness semantics has to be mirrored in two places or the tests silently stop reflecting production behavior.

Suggestion: delete `receive` and both tests. The send path is already covered by `test_send_data` and `test_network_transport_send`, and the read path is covered better by `test_read_value_spanning_packet_boundary` and friends.

### Low

**Duplicate UTF-16 helpers** — `error_parser.rs:133`, `info_parser.rs:133`

Both still define a local `encode_utf16_string` even though this PR adds `packet_reader::tests::encode_utf16_le`, which the `network_transport` tests already use. Worth collapsing onto the shared helper while consolidating.

**`#[async_trait]` on `NetworkReader` is now vestigial** — `reader_writer.rs:42`

After dropping `receive`, the trait has a single sync method, so the attribute can go.

**Test-helper API asymmetry** — `packet_reader.rs:264`, `packet_reader.rs:280`

`create_network_transport_with_data` is `async` while `create_network_transport_with_chunked_data` is sync (spawn-based). Minor, but an easy trip hazard for the next person adding a test. Also, the doc on the chunked variant says the writer "blocks until the reader drains each piece" — it actually unblocks as soon as any capacity frees.

### Things done well

- The `TestPacketBuilder::build` fix is a genuine latent-bug fix (payload length was written where TDS requires total length), and `test_packet_builder_writes_total_length_in_header` pins it.
- The ported tests are strictly stronger than the ones deleted: `test_read_value_spanning_packet_boundary` / `_fragmented`, `test_packet_header_split_across_reads`, `test_truncated_packet_reports_error`, and `test_read_varchar_u8_length_long_string` (the `200u8 << 1` truncation case) all cover behavior the old mock could not express.
- Doc comments on the new helpers correctly state the 8000-byte packet limit, matching the `ClientContext` default `packet_size`.

### Before merge

`mssql-rs Pull request validation (Build mssql-python macOS (cross-repo))` is currently red. This PR touches no `mssql-py-core` code so it is probably unrelated, but worth confirming from the ADO log rather than assuming.

David-Engel:
> ## Review
> Solid refactor — no correctness bugs found. Verified locally: `cargo clippy -p mssql-tds --all-features --all-targets -- -D warnings` is clean, and lib tests are 1680 passed / 7 failed, where all 7 are the pre-existing `certificate_validator` and `win_tls::validate` fixture failures unrelated to this diff. I also confirmed no bare `PacketReader` struct references remain anywhere (only the `TdsPacketReader` trait), and the two remaining `.receive(` call sites are both `cfg(test)`.
> 
> ### Medium
> **`NetworkTransport::receive` is now kept alive only for tests** — `network_transport.rs:635`
> 
> (clipped)

This review is from GHCP (trying it out).

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1135,7 +1135,7 @@ impl TdsPacketReader for NetworkTransport {
     }
 
     async fn read_byte(&mut self) -> TdsResult<u8> {
-        if !self.tds_read_buffer.do_we_have_enough_data(1) {
+        while !self.tds_read_buffer.do_we_have_enough_data(1) {
```
David-Engel, line 1113: **Suggestion (correctness, worth calling out):** this `if` -> `while` change across the 13 fixed-width readers is a genuine bug fix, not a refactor.

`get_slice()` returns `&self.working_buffer[buffer_position..]` — it isn't bounded by `buffer_length` — so under the old `if`, a value spanning a packet boundary onto a packet whose payload was still smaller than the requested width silently read stale bytes, then panicked in `TdsReadBuffer::consume_bytes` (`"Not enough data to consume"`). `test_read_value_spanning_packet_boundary` is the regression test for it. Please surface this in the description + CHANGELOG.

**Separate, lower-severity note on the same loop:** `get_new_tds_packet` accepts `packet_size_from_header == PacketWriter::PACKET_HEADER_SIZE`, and `remove_header_from_packet(8)` then adds zero payload bytes — so a peer streaming header-only packets keeps this `while` alive indefinitely. It isn't a CPU spin (each iteration blocks on a network read) and the request timeout bounds it, and `read_bytes` / `skip_bytes` already have the same shape, so I don't consider this introduced here. If you want to close it cheaply: reject `packet_size_from_header == PACKET_HEADER_SIZE` for non-EOM packets, or cap consecutive empty packets.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1296,20 +1296,20 @@ impl TdsPacketReader for NetworkTransport {
...
+            .read_unicode_with_byte_length((length as usize) << 1)
             .await?;
         Ok(Some(string))
     }
 
     async fn read_varchar_u8_length(&mut self) -> TdsResult<String> {
         let length: u8 = self.read_byte().await?;
         let string = self
-            .read_unicode_with_byte_length((length << 1) as usize)
+            .read_unicode_with_byte_length((length as usize) << 1)
```
David-Engel, line 1287: **Suggestion (bug fix — please document it):** this is the second production fix hiding in the PR. `(length << 1) as usize` shifted in the source width, so `200u8 << 1 == 144`: any `B_VARCHAR` of >= 128 characters read 144 bytes instead of 400 and desynchronized the token stream from that point on. The `u16` form on line 1304 had the same truncation above 32767.

Long proc names in ERROR/INFO tokens and long table/column names hit this in the wild, so it deserves a CHANGELOG entry rather than being folded into a deletion PR. `test_read_varchar_u8_length_long_string` is a good regression test for it.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -633,6 +629,10 @@ impl NetworkTransport {
         self.encryption = Some(encryption);
     }
 
+    /// Raw byte-level read, used only by transport tests. The production read
+    /// path is [`get_new_tds_packet`](Self::get_new_tds_packet), which reads
+    /// straight into the working buffer.
+    #[cfg(test)]
```
David-Engel: **Nit:** this leaves `#[cfg(test)]`-gated code on the production type — a smaller version of the smell the PR is removing. Its only two callers are test modules (`network_transport.rs` tests around line 1760 and `io/reader_writer.rs` tests around line 114).

Fine to leave as-is for this PR; a follow-up could move it into the test module or have those two tests read the raw stream directly.

David-Engel (commented):
Nice cleanup — deleting a reader that only existed for tests and re-pointing the token-parser tests at the code that actually ships is the right call, and the new duplex-stream harness is a real upgrade: it exercises packet coalescing, header fragmentation, and truncated packets, none of which the old mock could reach. No blocking issues.

Verified locally on `043e2b9`:

- `cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets` clean
- `cargo test -p mssql-tds --lib` -> 1592 passed, 4 failed; all four are `certificate_validator` fixture failures (`tests/test_certificates/*` missing on my box), pre-existing and unrelated
- The 21 ported/new reader tests pass, and `RUSTFLAGS="--cfg fuzzing" cargo check -p mssql-tds` builds
- No remaining references to `PacketReader::`, `LENGTHNULL`, `MockNetworkReaderWriter`, or `read_tds_packet_for_test` anywhere in the tree

### Main point: "no production behavior change" isn't accurate

Two real bug fixes are bundled in here, and they're worth calling out rather than burying under a deletion PR:

1. **`if` -> `while` on `do_we_have_enough_data`** (13 methods in `network_transport.rs`). `get_slice()` returns `&working_buffer[buffer_position..]`, unbounded by `buffer_length`, so with `if` a fixed-width value spanning a packet boundary onto a packet whose payload was *still* smaller than needed read stale buffer bytes and then panicked in `TdsReadBuffer::consume_bytes` with `"Not enough data to consume"`. That's exactly what `test_read_value_spanning_packet_boundary` now covers.

2. **The shift widening fix** in `read_varchar_u8_length` / `read_varchar_u16_length`. `(length << 1) as usize` shifted in the *source* width: `200u8 << 1 == 144`. Any `B_VARCHAR` of >= 128 characters (long proc names in ERROR/INFO tokens, table/column names) read 144 bytes instead of 400 and desynchronized the token stream.

Both are user-visible fixes against shipped behavior. Please update the PR description and add a `## [Unreleased]` entry in `CHANGELOG.md` (a `### Fixed` section would need to be created — the file currently only has `Added` / `Changed`).

Everything else is a suggestion or a nit; details inline.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -2530,4 +2535,283 @@ pub(crate) mod tests {
...
+        assert_eq!(
+            reader.read_varchar_u8_length().await.unwrap(),
+            unicode_string
+        );
+    }
+
+    #[tokio::test]
+    async fn test_truncated_packet_reports_error() {
+        let mut binding = TestPacketBuilder::new(PacketType::TabularResult);
+        let mut packet = binding.append_bytes(&[0xAB; 92]).build();
```
David-Engel: **Nit:** the `92` / `100` / `200` relationship is implicit. Something like `&[0xAB; 100 - PacketWriter::PACKET_HEADER_SIZE]` would make the intent self-evident and survive a header-size change.

mssql-tds/src/io/packet_reader.rs
```diff
@@ -672,456 +222,98 @@ pub(crate) mod tests {
...
-
-        packet_reader.read_tds_packet().await.unwrap();
+    /// Builds a `NetworkTransport` whose read side is pre-loaded with `data`.
+    /// The writer half is dropped, so reads observe EOF once `data` is drained.
+    ///
+    /// `data` is delivered in a single read, so this cannot express fragmented
+    /// reads — use [`create_network_transport_with_chunked_data`] for those.
+    /// Every TDS packet in `data` must be at most 8000 bytes in total (the
+    /// default negotiated packet size), or `get_new_tds_packet` rejects it.
+    pub(crate) async fn create_network_transport_with_data(data: &[u8]) -> NetworkTransport {
```
David-Engel: Two suggestions on the new helpers, both organizational — the current code works.

**1. Layering.** `io::packet_reader::tests` now constructs `NetworkTransport`, so the low-level I/O module's test module depends on `connection::transport`. Consider moving `build_duplex_transport`, `create_network_transport_with_data`, and `create_network_transport_with_chunked_data` into `connection::transport::network_transport::tests` (or a shared test-support module), leaving only `TestPacketBuilder` and `encode_utf16_le` here.

**2. Asymmetry.** This helper is `async` (it awaits `write_all` into an over-sized duplex buffer) while `create_network_transport_with_chunked_data` is sync (it spawns). Two helpers with the same purpose and different call shapes are easy to trip over. Having this one spawn the write too would let both be sync and drop ~15 `.await`s from the call sites.

**Unrelated, informational:** the `PacketReader::read_unicode_with_byte_length` you deleted carried a `MAX_STRING_BYTE_LENGTH = 510` OOM guard; the surviving `NetworkTransport` implementation has none. Not a regression — production never had the guard, and the allocation is bounded by `u16` (<= 128 KB per call) — but if the guard was considered worth having, it now belongs on the implementation that actually ships.

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Thanks for digging into the `get_slice()` / `consume_bytes` interaction — that matches what I found, and you're right that it deserves to be surfaced rather than buried in a deletion PR.

Done both: the PR description now has a **Production bug fixes** section calling this out explicitly (including the stale-bytes-then-panic mechanism), and `CHANGELOG.md` has a new `### Fixed` entry under `[Unreleased]` naming all 13 fixed-width readers.

I also took your first suggested close on the header-only packet. `get_new_tds_packet` now rejects a packet whose header length equals `PACKET_HEADER_SIZE` unless the EOM flag is set, since an empty non-EOM packet advances the buffer by nothing and is exactly what would keep the new `while` loops spinning. Empty EOM packets stay legal because they legitimately terminate a message. Covered by `test_payload_free_non_eom_packet_is_rejected` and `test_payload_free_eom_packet_is_accepted`.


> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Agreed, and thanks for confirming the u16 case too — I had only worked the `u8` arithmetic through.

Added a `### Fixed` entry in `CHANGELOG.md` covering both readers with the exact thresholds (≥128 characters for `read_varchar_u8_length`, ≥32768 for `read_varchar_u16_length`) and your `200u8 << 1 == 144` example as the mechanism. The PR description now leads with a **Production bug fixes** section so this isn't presented as incidental cleanup.


> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: You're right that it's the same smell, so I removed it rather than leaving it — a PR whose whole point is deleting a test-only duplicate reader shouldn't ship a test-only duplicate read method.

Looking at the two callers, neither survived contact with that reasoning:

- `test_network_transport_receive` had no subject other than `receive` itself.
- `test_send_recv` sent `MAX_BUFFER_SIZE` random bytes and asserted the far end saw them — which is exactly what `test_send_data` already asserts, just read back through `FramedRead` instead of `receive`. Once `receive` is gone it's a literal duplicate of `test_send_data`, so it went too, along with `create_client_server_network_transport`, which had no other caller.

`NetworkWriter::send` coverage is unchanged (`test_send_data` still exercises it at `MAX_BUFFER_SIZE`), and `known_dead` on read failure/EOF is still covered through `get_new_tds_packet`, which sets it on all three paths.


> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Taken as suggested — it's now `&[0xAB; 100 - PacketWriter::PACKET_HEADER_SIZE]`, so the `assert_eq!(packet.len(), 100)` below it can no longer drift out of sync with the payload literal.


> [!NOTE]

mssql-tds/src/io/packet_reader.rs
saurabh500: Took your third option — a shared test-support module — because the first two both have a reviewer against them.

These helpers actually started out in `connection::transport::network_transport::tests`. @saurabh500 asked me to move them here in the previous round (thread on this same file), on the grounds that the parser tests were reaching into `connection::transport` for their plumbing. Your point about `io::packet_reader` depending on `connection::transport` is equally fair — the honest read is that this code spans both layers and neither module is a good owner, so moving it back and forth between the two reviewers' preferences would just be churn.

So it now lives in a new `#[cfg(test)] mod test_packet_support` at the crate root: `TestPacketBuilder`, `encode_utf16_le`, `build_duplex_transport`, `create_network_transport_with_data`, `create_network_transport_with_chunked_data`, and the two builder tests. `io::packet_reader` no longer pulls in `NetworkTransport`, `network_transport::tests` no longer owns packet-building plumbing, and the parser tests import from one neutral place.

Also fixed the asymmetry you flagged. `create_network_transport_with_data` is now sync and delegates to `create_network_transport_with_chunked_data(data, data.len().max(1))` — one implementation, one spawned writer, and ~15 `.await`s dropped from the call sites. Behavior is unchanged: the duplex buffer still holds all of `data`, and the writer half is still dropped at the end so reads see EOF.

On the 510-byte guard: agreed it's not a regression, and I wrote up why it shouldn't come back in https://github.com/microsoft/mssql-rs/pull/221#issuecomment-5258059698 — short version, it was rejecting valid input (any `u16`-length string of 256+ characters) rather than bounding anything, and the real bound is the u16 length field plus the `TdsReadBuffer` capacity checks.


> [!NOTE]

CHANGELOG.md
```diff
@@ -55,3 +55,27 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
...
+
+### Fixed
+
+- `mssql-tds`: reading a fixed-width value that straddles a TDS packet boundary
+  could return bytes from the wrong place or panic. The readers checked for
+  sufficient buffered data with an `if` and read a single further packet, but a
+  value can span more than two packets (and a packet can carry fewer bytes than
+  the value needs), so the read proceeded against a still-short buffer. The
+  check is now a loop that reads until the whole value is buffered. Affects all
+  13 fixed-width readers (`read_uint16` … `read_double`, the date/time readers,
```
David-Engel: **Suggestion (factual):** the method names in this parenthetical don't exist. `TdsPacketReader` has no `read_double`, no date/time readers, and no GUID reader.

The 13 that actually changed are: `read_byte`, `read_int16`, `read_uint16`, `read_uint24`, `read_int32`, `read_uint32`, `read_int64`, `read_uint64`, `read_uint40`, `read_float32`, `read_float64`, `read_int16_big_endian`, `read_int32_big_endian`.

Since this is the user-facing changelog, I'd either list them or just drop the parenthetical and say "all 13 fixed-width readers on `TdsPacketReader`".

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -948,6 +910,19 @@ impl NetworkTransport {
...
         }
 
+        // A payload-free packet that is not the end of its message advances
+        // nothing, so a peer streaming them would spin the callers' "read until
+        // we have enough data" loops forever. An empty EOM packet is legal (it
+        // terminates a message), so only non-EOM ones are rejected.
+        let is_end_of_message = self.tds_read_buffer.working_buffer[base_offset + 1]
+            & PacketStatusFlags::Eom as u8
+            != 0;
+        if packet_size_from_header == PacketWriter::PACKET_HEADER_SIZE && !is_end_of_message {
```
David-Engel, line 920: **Suggestion:** the guard only closes half the space it claims to. An empty **EOM** packet is also payload-free, is accepted here by design, and also advances `buffer_length` by zero — so the `while !do_we_have_enough_data(n)` loops keep going against a peer that streams empty EOM packets, which is precisely the scenario the CHANGELOG entry and description point 3 say is now prevented.

I confirmed this on `c579b81` with a throwaway test: five empty EOM packets followed by a one-byte packet, and `read_byte()` returns that byte — i.e. the loop consumed all five empties and carried on. Unbounded input, unbounded looping.

```rust
let mut stream = Vec::new();
for _ in 0..5 {
    stream.extend_from_slice(&TestPacketBuilder::new(PacketType::TabularResult).build());
}
let mut data = TestPacketBuilder::new(PacketType::TabularResult);
stream.extend_from_slice(&data.append_byte(0x7F).build());
let mut reader = create_network_transport_with_data(&stream);
assert_eq!(reader.read_byte().await.unwrap(), 0x7F); // passes
```

Two ways out, either is fine:

1. Count consecutive zero-payload packets in `read_tds_packet` / the reader loops and error out past a small bound, regardless of the EOM bit. That actually closes it.
2. Keep the guard as-is (it's still worth having — a non-EOM empty packet is malformed) but soften the CHANGELOG entry and description point 3 so they claim what the code does: rejecting malformed payload-free non-EOM packets, with the looping concern still bounded by the request timeout rather than eliminated.

To be clear, this was a low-severity note from me last round and it stays low-severity — the request timeout bounds it in practice. I only care that the note and the code agree.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -2530,4 +2423,316 @@ pub(crate) mod tests {
...
+
+    /// The same packet *with* EOM set is legal — an empty end-of-message packet
+    /// terminates a message — so it must be accepted and simply yield no payload.
+    #[tokio::test]
+    async fn test_payload_free_eom_packet_is_accepted() {
+        let packet = TestPacketBuilder::new(PacketType::TabularResult).build();
+
+        let mut reader = create_network_transport_with_data(&packet);
+
+        assert!(!matches!(
```
David-Engel: **Nit:** this negative assertion is weaker than it looks. `read_byte()` here returns `Err(ConnectionClosed)` — the empty EOM packet is consumed, the loop goes around, and the next read hits EOF. So the test passes for a reason unrelated to what it's checking, and would keep passing if the accept path regressed in most other ways.

Asserting the packet is consumed *and the reader keeps going* is stronger and directly encodes the intent:

```rust
let mut stream = TestPacketBuilder::new(PacketType::TabularResult).build();
let mut data = TestPacketBuilder::new(PacketType::TabularResult);
stream.extend_from_slice(&data.append_byte(0x7F).build());

let mut reader = create_network_transport_with_data(&stream);

assert_eq!(reader.read_byte().await.unwrap(), 0x7F);
```

`test_payload_free_non_eom_packet_is_rejected` above is fine as written — it matches on the specific error variant.

David-Engel (commented):
Re-reviewed at `c579b81`. All five points from the last round are addressed, and addressed well:

- The description now has a dedicated **Production bug fixes** section, and `CHANGELOG.md` gained a `### Fixed` block. Exactly what I was after.
- `test_packet_support` as a top-level `#[cfg(test)]` module is a better answer than what I suggested — the doc comment explaining *why* it sits outside both `io` and `connection::transport` is the right call.
- Both factories are now sync, with `create_network_transport_with_data` delegating to the chunked one. ~30 `.await`s gone from the call sites.
- `NetworkTransport::receive` deleted outright rather than left `#[cfg(test)]`-gated. Better than my suggestion; the two tests it supported only exercised the deleted method, and `send` is still covered by `test_network_transport_send`.
- The `92` magic number is now derived from `PacketWriter::PACKET_HEADER_SIZE`.

Nice touch adding regression tests for the new payload-free-packet guard rather than shipping it bare.

Re-verified on `c579b81`: `cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets` clean, `cargo test -p mssql-tds --lib` -> 1592 passed / 4 failed (the same pre-existing `certificate_validator` fixture failures), and `RUSTFLAGS="--cfg fuzzing" cargo check -p mssql-tds` builds.

Three follow-ups, none blocking — one factual fix in the CHANGELOG, and one place where the new guard doesn't quite deliver what the note claims. Details inline.

CHANGELOG.md
saurabh500: Fixed — you're right, none of `read_double`, the date/time readers, or a GUID reader exist on `TdsPacketReader`. I've dropped the parenthetical for the short form you suggested: "Affects all 13 fixed-width readers on `TdsPacketReader`." Naming all 13 felt like noise for a user-facing changelog, and the count plus the trait name is enough to tell a reader whether they're affected.

Fixed in f13b5a8.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: You're right, and thanks for the repro — I confirmed it. The guard only rejects the non-EOM case; an empty EOM packet is payload-free too, is accepted by design, and advances `buffer_length` by zero, so the `while !do_we_have_enough_data(n)` loops keep going. The claim and the code did not agree.

I took your option 2, and applied it everywhere the claim appeared rather than just the changelog — the entry, the code comment above the guard, the doc comment on `test_payload_free_non_eom_packet_is_rejected`, and point 3 of the PR description all asserted it stopped unbounded looping. All four now describe the guard as what it actually is: rejection of a packet that neither carries payload nor terminates a message.

I went that way rather than option 1 because I don't think the looping is a distinct problem worth a bound. Every iteration awaits a real socket read and consumes 8 bytes, so there's no spin, no memory growth, and no amplification — the client does O(1) work per 8 bytes received. That's indistinguishable from a peer that streams a large response slowly, which is what the request timeout is for. A consecutive-empty-packet counter would add an arbitrary constant that catches nothing the timeout doesn't already bound, while creating a new way to reject an unusual-but-legal peer. This PR already removed one arbitrary limit from this reader (the 510-byte allocation guard) for essentially that reason, so re-adding a different one felt inconsistent.

The guard still earns its place on the narrower ground you named: a payload-free non-EOM packet is malformed, full stop.

Fixed in f13b5a8.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Agreed, and taken as written. The old assertion passed on `Err(ConnectionClosed)` — the empty EOM packet was consumed, the loop went around, and the next read hit EOF — so it was really only asserting "the error isn't a `ProtocolError`", which is not what the test name claims.

The test now asserts the packet is consumed *and* the reader carries on to the next packet's payload. I checked it has teeth by temporarily dropping the `&& !is_end_of_message` condition from the guard: the new version fails, and I reverted the change.

`test_payload_free_non_eom_packet_is_rejected` is unchanged apart from its doc comment, which had the same overstated looping rationale discussed in the other thread.

Fixed in f13b5a8.

> [!NOTE]

David-Engel (approved):

PR #222, opened 2026-08-11
Add pipeline job running the mssql-python suite on the mssql-odbc driver

### LINKED ISSUE #220: Run the mssql-python test suite against the mssql-odbc driver in PR validation
## Summary

We have a `Build mssql-python (cross-repo)` job that validates fresh `mssql-py-core` bindings against `microsoft/mssql-python`, but it keeps the Microsoft ODBC Driver 18 binary that mssql-python ships. Nothing in PR validation currently runs a real-world Python workload against the Rust `mssql-odbc` driver.

That means we have no continuous signal on how far `mssql-odbc` is from being a drop-in replacement for `msodbcsql18`.

## Proposal

Add a PR validation job that runs the full upstream mssql-python pytest suite with the bundled `libmsodbcsql-<MAJOR>.<MINOR>.so.2.1` binary replaced by the `.so` produced by `cargo build --release -p mssql-odbc`, against SQL Server in Docker.

Requirements:

- **No mssql-python source may be modified.** Overwriting or deleting the packaged driver binary is acceptable — this is effectively what mssql-python's own pipeline does when it uninstalls the system driver before testing.
- **Run the same tests the mssql-python repo runs**, so results are comparable to upstream.
- **SQL Server runs in Docker**, matching the upstream Ubuntu leg.
- **The job must tolerate failures and hard crashes.** The driver is incomplete, so segfaults, aborts and hangs are expected. The job must report results rather than stalling or losing the whole run to a single crash, mirroring the `continueOnError` behavior of the existing cross-repo job.

## Notes

mssql-python resolves its driver entirely in native code (`ddbc_bindings` `GetOdbcLibsBaseDir` / `GetDriverPathCpp`): it takes the `mssql_python_odbc` package directory and appends `libs/<platform>/<arch>/lib/libmsodbcsql-<MAJOR>.<MINOR>.so.2.1`, where `MAJOR.MINOR` is baked in at build time from `mssql_python_odbc.__version__`. Overwriting that binary is therefore enough to redirect the whole suite — no source change, no `odbcinst` registration, no `LD_LIBRARY_PATH`.

This job is informational: it is expected to be red for the foreseeable future and should not gate merges. Its value is the per-file pass/fail/crash breakdown published to the Tests tab, which gives us a burn-down of driver gaps.

saurabh500:
## Description

Adds a PR validation job that runs the **full upstream `microsoft/mssql-python` pytest suite against the Rust `mssql-odbc` driver**, against SQL Server in Docker.

It is the driver-swap counterpart to the existing `Build mssql-python (cross-repo)` job:

| | `Build mssql-python (cross-repo)` | `mssql-python suite on mssql-odbc driver` (new) |
|---|---|---|
| `mssql_py_core` | built fresh from this PR | released, from NuGet |
| ODBC driver | Microsoft ODBC Driver 18 (as shipped) | **`mssql-odbc` built from this PR** |
| Validates | our Python bindings | **our ODBC driver** |

Holding `mssql_py_core` at the released version means the delta between the two jobs is attributable to the ODBC driver.

### How the swap works — without touching mssql-python source

mssql-python resolves its driver entirely in native code (`ddbc_bindings` `GetOdbcLibsBaseDir` / `GetDriverPathCpp`): it takes the `mssql_python_odbc` package directory and appends

```
libs/<platform>/<arch>/lib/libmsodbcsql-<MAJOR>.<MINOR>.so.2.1
```

where `MAJOR.MINOR` is baked in at build time from `mssql_python_odbc.__version__` (currently `18.6.2.1`). So **overwriting that one binary redirects the whole suite** — no source change, no `odbcinst` registration, no `LD_LIBRARY_PATH`. This is effectively what mssql-python's own pipeline does when it uninstalls the system driver before testing.

`swap-mssql-python-odbc-driver.sh` parses `__version__` at runtime rather than hardcoding the filename, and swaps every copy it finds (in-repo and site-packages) so the result doesn't depend on which one wins on `sys.path`.

### Surviving crashes and hangs

The driver is incomplete, so failures **and hard crashes are expected**. Rather than one `pytest` invocation that a single segfault would take down, `run-mssql-python-odbc-tests.sh` runs **each test file in its own `timeout`-bounded pytest process**:

- A crash kills only that file's process; the loop continues.
- `timeout --kill-after=60s` bounds every file, so a wedged driver call can't stall the run.
- **An aggregate wall-clock budget (110m) bounds the loop as a whole.** Per-file timeouts alone aren't enough: 40 test files x 10m is 400 minutes against a 150-minute job cap, and a job cancelled by the *server* for timeout does **not** run `condition: always()` steps — results would go unpublished and the containers would leak. Once the budget is spent the remaining files are recorded as skipped, so the run always ends with a complete, publishable result set.
- pytest writes its JUnit XML only at the end of a run, so a file that dies leaves no report — a synthetic one is emitted so the crash stays visible in the Tests tab.
- Files fully deselected by upstream's `addopts = -m "not stress"` are recorded as skipped, not errors.
- Exit codes 125-127 (`timeout` could not execute the command) are reported as *harness* errors, not driver crashes, and pytest is asserted usable up front — otherwise a broken venv would publish as 40 fabricated driver failures.
- **The runner exit code is a three-way contract**, so a broken environment cannot masquerade as a driver defect: `0` every file passed; `1` tests ran but the run was not clean (failures, crashes, timeouts, or files skipped because the budget ran out) → yellow; `2` the harness could not run the tests → red. `2` also covers a run in which *no file executed a single test* — a suite that silently exercises nothing is a real failure, not a pass.

The test step does not use `continueOnError`; it inspects the runner's exit code directly. **Test failures, crashes and timeouts are reported as `SucceededWithIssues`** — a yellow warning on the job, which still reports `pass` to GitHub so it never blocks a PR — while a broken harness is a real red error. `PublishTestResults` uses `failTaskOnFailedTests: false`. **This job is informational and expected to be yellow** — its value is the per-file breakdown published to the Tests tab.

Example of the summary it prints:

```
test_000_pass                      PASSED
test_001_fail                      FAILED
test_002_crash                     CRASHED (signal 11)
test_003_hang                      TIMED OUT after 600s
test_004_stress_only               NO TESTS COLLECTED
test_005_late                      SKIPPED (total time budget of 110m exhausted)
files: 6 | passed: 1 | failed: 1 | crashed: 1 | timed out: 1 | no tests: 1 | skipped: 1 | harness errors: 0
```

### Also included

The ~45 lines of mssql-python clone + PR-description branch-override logic were extracted from `test-mssql-python-template.yml` into `.pipeline/scripts/clone-mssql-python.sh` so both cross-repo jobs share one copy. Pure extraction — behavior is otherwise unchanged. Note that the `mssql-python-branch: <branch>` PR-description override only resolves on Azure DevOps Repos-sourced builds. On the GitHub-sourced builds this pipeline actually runs as, the lookup could never succeed, so it is now skipped outright and the clone always tracks main.

### Files

| File | |
|---|---|
| `.pipeline/templates/test-mssql-python-odbc-template.yml` | new job steps |
| `.pipeline/scripts/swap-mssql-python-odbc-driver.sh` | replaces the bundled driver binary |
| `.pipeline/scripts/containerized-odbc-swap-build.sh` | builds the mssql-odbc cdylib in the build container |
| `.pipeline/scripts/run-mssql-python-odbc-tests.sh` | crash/hang-resilient per-file test runner |
| `.pipeline/scripts/clone-mssql-python.sh` | extracted shared clone logic |
| `.pipeline/templates/validation-stages.yml` | wires the job into `Build_mssql_python` |
| `.pipeline/templates/test-mssql-python-template.yml` | uses the extracted clone script |

## Related Issues

Fixes #220

## Validation

No Rust or JS source changed, so `bfmt`/`bclippy`/`btest` are unaffected. The pipeline changes were verified directly:

- **Driver swap run against a real `microsoft/mssql-python` checkout** — `__version__ = 18.6.2.1` correctly resolved to `libmsodbcsql-18.6.so.2.1`, all 7 platform/arch copies overwritten, replacement content verified byte-wise.
- **Test runner run end-to-end in a Linux container with real pytest 9.1.1**, against fixtures reproducing every failure mode, including upstream's `-m "not stress"` `pytest.ini`:

  | fixture | outcome | handled as |
  |---|---|---|
  | passing test | exit 0 | `PASSED` |
  | failing assert | exit 1 | `FAILED` |
  | `ctypes.string_at(0)` | **real SIGSEGV** | `CRASHED (signal 11)` + synthetic report |
  | `time.sleep(600)` | **real hang** | `TIMED OUT` + synthetic report |
  | import error | exit 2 | `INTERRUPTED` |
  | `@pytest.mark.stress` only | exit 5 | `NO TESTS COLLECTED` (skip, not error) |
  | back-to-back hangs vs. a small budget | budget exhausted | remaining files `SKIPPED`, loop exits cleanly |
  | pytest not installed | — | aborts up front as a harness error |

  The loop never stalled, every file produced a JUnit report, and each parsed as well-formed XML with correct `tests`/`failures`/`errors`/`skipped` counts. Also confirmed the budget does *not* skip prematurely when there is headroom.
- Caught two real bugs this way: `--cache-clear` combined with `-p no:cacheprovider` makes pytest exit 4 (`unrecognized arguments`), which would have failed **every** file; and the missing aggregate budget described above. Both fixed and re-verified.
- All 3 YAML files parse; all shell scripts pass `bash -n`; new files committed with LF endings (`git ls-files --eol` verified).

Not reproducible locally and will be exercised on the first pipeline run: the containerized `cargo build --release -p mssql-odbc`, the NuGet `mssql_py_core` install, and the real driver against SQL Server.

## Checklist

- [x] `cargo bfmt` passes (no Rust changes)
- [x] `cargo bclippy` passes (no Rust changes)
- [x] `cargo btest` passes (no Rust changes)
- [x] New/changed functionality has tests — pipeline change; scripts validated as described above


.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,240 @@
...
+      set -euo pipefail
+      export DEBIAN_FRONTEND=noninteractive TZ=UTC
+      ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+      apt-get update
+      # Same set the upstream Ubuntu leg installs, plus the runtime libraries the
+      # Rust driver links against (TLS via OpenSSL, Kerberos via GSSAPI).
+      apt-get install -y --no-install-recommends \
+        python3 python3-pip python3-venv python3-full python3-dev \
+        cmake curl wget gnupg build-essential pybind11-dev \
+        libssl3 libgssapi-krb5-2 libkrb5-3 ca-certificates
```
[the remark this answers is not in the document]
saurabh500: Good catch, and confirmed: on a clean `ubuntu:24.04` `/usr/include/sql.h` is absent, so `ddbc_bindings` would have failed to compile before the swap or any test ran. Added `unixodbc-dev` to the package list.

I also mirrored the rest of upstream's sequence: it installs `unixodbc-dev` only to build, then purges it and removes the system `libodbcinst.so.2` before running tests. Added a `Strip build-time ODBC toolchain` step that does the same, so the only intended difference from an upstream run is the driver binary itself.

Verified in a container: headers absent before, present after install, purge and removal clean.

.pipeline/templates/validation-stages.yml
```diff
@@ -525,6 +525,20 @@ stages:
     steps:
       - template: test-mssql-python-macos-template.yml
 
+  # Same upstream suite, but with mssql-python's bundled msodbcsql18 binary
+  # replaced by the Rust mssql-odbc driver. Failures and crashes are expected
+  # while the driver is in development, so the job is informational: its test
+  # step is continueOnError and each test file runs in its own bounded process.
+  - job: Test_mssql_python_on_mssql_odbc
+    displayName: mssql-python suite on mssql-odbc driver (cross-repo)
+    timeoutInMinutes: 180
```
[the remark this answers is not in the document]
saurabh500: Agreed, and this was the most serious issue in the PR: a job cancelled by the server for exceeding `timeoutInMinutes` skips `condition: always()` steps, so the results would never publish and the containers would leak. That is exactly the stall this job is meant to avoid.

Fixed in `08d63c0c` and tightened here. The runner now enforces an aggregate `PYTEST_TOTAL_BUDGET` (110m default) across the whole loop, clamps each file's slice to whatever budget remains, and emits skipped stubs for any files it did not get to, so the report is always complete. The bounds now nest consistently:

- 110m runner-enforced test budget
- 130m step timeout, backstopping a wedged `docker exec`
- 240m job timeout (raised from 180m), leaving ample room for setup plus publish and cleanup

I dropped a comment on the job recording that arithmetic. Verified the budget-exhaustion path emits stubs for the remainder and exits cleanly, and that it does not skip prematurely when headroom remains.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,151 @@
...
+    if [ ! -s "$report" ]; then
+        write_report_stub "$name" "$kind" "$status" "$report"
+        if [ "$kind" = "skipped" ]; then
+            echo "$name: $status"
+        else
+            echo "##[warning]$name: $status - no pytest report produced, wrote synthetic error report"
+        fi
+    else
+        echo "$name: $status"
+    fi
```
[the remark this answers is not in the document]
saurabh500: Correct, and reproduced: with upstream's `addopts = -m "not stress"`, a stress-only file exits 5 but still writes a JUnit report containing zero test cases, so the size check kept it and the Tests tab showed nothing for that file.

Reordered exactly as suggested: `kind=skipped` is now handled before the report-size check, overwriting pytest's zero-test report with the skipped stub.

Verified against real pytest 9.1.1 on a fixture with one passing file and one fully deselected file. The deselected file now publishes `tests="1" ... skipped="1"` with a visible `<skipped>` case, and the passing file's genuine pytest report is left untouched.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,213 @@
...
+printf '%s\n' "${SUMMARY[@]}"
+echo "==============================================================="
+echo "files: ${#TEST_FILES[@]} | passed: $passed | failed: $failed | crashed: $crashed | timed out: $timedout | no tests: $empty | skipped: $skipped | harness errors: $harness_error"
+echo "==============================================================="
+
+if [ "$harness_error" -gt 0 ]; then
+    echo "##[error]$harness_error file(s) could not be executed - this indicates a broken test environment, not a driver defect"
+fi
+
+if [ "$failed" -gt 0 ] || [ "$crashed" -gt 0 ] || [ "$timedout" -gt 0 ] || [ "$harness_error" -gt 0 ]; then
```
saurabh500: **Suggestion — the harness-error distinction has no teeth.**

The 125/126/127 classification was added specifically so a broken venv would not be published as ~40 fabricated driver failures. But this branch exits `1` for *both* harness errors and ordinary driver failures, and the calling step is `continueOnError: true`. So a completely broken environment produces the same job outcome as the expected-red baseline: a warning buried in a 240-minute log. The report was fixed, but the signal was not.

Give the two classes distinct exit codes:

```bash
if [ "$harness_error" -gt 0 ]; then
    echo "##[error]$harness_error file(s) could not be executed - broken test environment, not a driver defect"
    exit 2
fi

if [ "$failed" -gt 0 ] || [ "$crashed" -gt 0 ] || [ "$timedout" -gt 0 ]; then
    echo "##[warning]mssql-python tests did not fully pass (expected while the driver is in development)"
fi
exit 0
```

See my comment on `test-mssql-python-odbc-template.yml` for the matching step change.

.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,266 @@
...
+      mssql-python-odbc-tests bash -c '
+        source /opt/venv/bin/activate
+        bash /mssql-rs-scripts/run-mssql-python-odbc-tests.sh
+      '
+  displayName: Run mssql-python tests against mssql-odbc
+  # The runner enforces its own budget; this only backstops a wedged `docker
+  # exec`. It must stay below the job timeout so the publish and cleanup steps
+  # below still run.
+  timeoutInMinutes: 130
+  continueOnError: true
```
saurabh500: **Suggestion — pairs with the exit-code split in `run-mssql-python-odbc-tests.sh`.**

`continueOnError: true` currently swallows environment breakage just as readily as it swallows expected driver failures. If the runner returns `2` for harness errors, this step can keep driver failures advisory while letting a broken environment actually block:

```yaml
- script: |
    docker exec \
      ... \
      mssql-python-odbc-tests bash -c '...'
    rc=$?
    if [ "$rc" -eq 2 ]; then
      echo "##vso[task.logissue type=error]mssql-python test harness could not run - see log"
      exit 1
    fi
    exit 0
  displayName: Run mssql-python tests against mssql-odbc
  timeoutInMinutes: 130
```

That keeps the stated intent (driver failures are informational) without making the job incapable of reporting its own infrastructure falling over.

.pipeline/templates/validation-stages.yml
```diff
@@ -525,6 +525,27 @@ stages:
...
+  # while the driver is in development, so the job is informational: its test
+  # step is continueOnError and each test file runs in its own bounded process.
+  #
+  # The time bounds nest deliberately: the runner caps the whole test loop at
+  # pytestTotalBudget (110m), the test step backstops a wedged `docker exec` at
+  # 130m, and the job allows 240m so the ~110m left over covers setup and still
+  # guarantees the publish and cleanup steps run. A job cancelled by the server
+  # for timeout skips `condition: always()` steps, which would lose the report
+  # this job exists to produce.
+  - job: Test_mssql_python_on_mssql_odbc
```
saurabh500, line 547: **Suggestion — a 240-minute advisory job on every PR is a lot of pool capacity.**

This runs a full `cargo build --release -p mssql-odbc` plus up to 41 × 10m pytest processes on **every** PR, and by design nobody gates on the result. Even the happy path is a substantial chunk of `RUST-1ES-POOL-WUS3` in a stage that already runs two other cross-repo mssql-python jobs.

The repo already has a pattern for opt-in expensive jobs (`RunFuzz` / `RunLongHaul` in `validation-pipeline.yml`):

```yaml
  - job: Test_mssql_python_on_mssql_odbc
    condition: and(succeeded(), eq('${{ parameters.RunMssqlPythonOdbcSwap }}', 'true'))
```

Alternatives: schedule it nightly, or gate it on `mssql-odbc/**` paths. Always-on may well be the right call while the driver is under active development — but since the existing comment block justifies the *timeouts* and not the *always-on* choice, worth an explicit decision plus one sentence recording it.

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
```diff
@@ -0,0 +1,84 @@
...
+DRIVER_FILENAME="libmsodbcsql-${DRIVER_MAJOR_MINOR}.so.2.1"
+
+echo "mssql_python_odbc.__version__ = $DRIVER_VERSION -> bundled driver file: $DRIVER_FILENAME"
+echo "Replacement driver: $DRIVER_SO"
+ls -la "$DRIVER_SO"
+
+# Swap every copy of the bundled binary that this interpreter could resolve:
+# the in-repo package (wins on sys.path when pytest runs from the repo root) and
+# any site-packages copy pulled in as a dependency.
+SEARCH_ROOTS=("$MSSQL_PYTHON_DIR")
```
saurabh500, line 67: **Suggestion — ask Python where the package resolves instead of enumerating candidate roots.**

This walks `site.getsitepackages()` plus the repo dir and swaps whatever it finds, but `GetOdbcLibsBaseDir` uses the `__file__` of whichever `mssql_python_odbc` actually imports. Asking directly is both simpler and exact:

```bash
PKG_DIR="$(python -c 'import mssql_python_odbc, os; print(os.path.dirname(os.path.abspath(mssql_python_odbc.__file__)))')"
echo "Python resolves mssql_python_odbc to: $PKG_DIR"
```

Keeping the extra roots as belt-and-braces is fine, but the `swapped -eq 0` guard on line 71 should be strengthened to assert that **at least one swapped target lived under `$PKG_DIR`**. Today the script exits 0 as long as it finds *any* copy — even if that copy is not the one that loads, in which case the entire suite silently runs against the stock driver and the job reports a misleading green-ish baseline.

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
```diff
@@ -0,0 +1,84 @@
...
+if [ "$swapped" -eq 0 ]; then
+    echo "##[error]No bundled $DRIVER_FILENAME found under: ${SEARCH_ROOTS[*]}"
+    exit 1
+fi
+
+echo "Replaced $swapped bundled driver binaries with mssql-odbc"
+
+# Surface unresolved shared-library dependencies here rather than as an opaque
+# import failure inside the first test.
+RESOLVED_TARGET="$MSSQL_PYTHON_DIR/mssql_python_odbc/libs/linux/debian_ubuntu/x86_64/lib/$DRIVER_FILENAME"
```
saurabh500: **Suggestion — hardcoding `debian_ubuntu/x86_64` makes this check a silent no-op elsewhere.**

This `ldd` probe is the one thing that surfaces a missing `libssl3`/`libgssapi` *before* it becomes 41 opaque test failures, but it only fires if the resolved path happens to match this exact distro/arch layout. On any other layout `[ -f ... ]` is false and the check quietly does nothing.

Since the loop above already knows exactly which files it overwrote, collect them and probe those:

```bash
for t in "${SWAPPED_TARGETS[@]}"; do
    echo "ldd $t"
    ldd "$t" || true
done
```

_Unrelated but worth watching on the first run:_ `mssql-odbc/build.rs:10` embeds `-soname,libmsodbcsql-18.4.so.1.1` while the file is installed as `libmsodbcsql-18.6.so.2.1`. `dlopen` by absolute path should not care — but if the first run dies with a loader error, that is the first place I would look.

.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,266 @@
...
+    # Only reading it is needed downstream, and the cleanup step uses sudo.
+    ls -la odbc-swap-drop
+  displayName: Build mssql-odbc driver
+  env:
+    CARGO_REGISTRIES_MSSQL_RS_PUBLIC_TOKEN: $(CARGO_REGISTRIES_MSSQL_RS_PUBLIC_TOKEN)
+    CARGO_REGISTRIES_MSSQL_RS_TOKEN: $(CARGO_REGISTRIES_MSSQL_RS_TOKEN)
+
+- script: |
+    set -euo pipefail
+    docker network create mssqlpy-odbc-net
```
saurabh500, line 98: **Suggestion — no defence against leftovers from a cancelled run.**

This runs under `set -euo pipefail` with no pre-cleanup. The comment block on the job in `validation-stages.yml` establishes that a server-cancelled job skips `condition: always()` steps — which means these containers and this network **can leak**. On the next run `docker network create` fails and the job hard-fails at "Start SQL Server": a real red on a job whose entire premise is not producing false alarms.

```bash
docker rm -f sqlserver-odbc-swap mssql-python-odbc-tests 2>/dev/null || true
docker network rm mssqlpy-odbc-net 2>/dev/null || true
docker network create mssqlpy-odbc-net
```

Flagging this as a concern rather than a fact because it only bites on a **stateful** agent, and the evidence cuts both ways: the `imageOverride` demand suggests 1ES hosted/ephemeral, but `docker-cargo-run.sh`'s disk-prune logic and the `$HOME/.cargo/registry` cache mount only make sense on persistent agents. Worth confirming with whoever owns `RUST-1ES-POOL-WUS3`.

If agents *are* persistent, note also that the release build leaves a root-owned `target/` that the Cleanup step does not remove — `containerized-odbc-build.sh:29` shows the repo already handles this class of problem from inside the container.

.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,266 @@
...
+
+    docker run -d \
+      --name sqlserver-odbc-swap \
+      --network mssqlpy-odbc-net \
+      --network-alias sqlserver \
+      -e ACCEPT_EULA=Y \
+      -e MSSQL_SA_PASSWORD="$(SQL_PASSWORD)" \
+      -p 1433:1433 \
```
saurabh500: **Suggestion — drop this port publish.**

Nothing on the agent host connects to SQL: the test container reaches it via the `sqlserver` network alias on `mssqlpy-odbc-net`. Meanwhile `test-mssql-python-template.yml` publishes the same host port, so if the two jobs ever land on the same agent this is a guaranteed bind conflict for zero benefit.

.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,266 @@
...
+    docker exec mssql-python-odbc-tests bash -c '
+      source /opt/venv/bin/activate
+      MSSQL_ODBC_DRIVER=/mssql-odbc/libmsodbcsql18.so \
+        bash /mssql-rs-scripts/swap-mssql-python-odbc-driver.sh
+    '
+  displayName: Replace bundled msodbcsql18 with mssql-odbc
+
+- script: |
+    docker exec \
+      -e DB_CONNECTION_STRING="Server=sqlserver,1433;Database=TestDB;Uid=SA;Pwd=$(SQL_PASSWORD);TrustServerCertificate=yes" \
```
saurabh500: **Suggestion — keep the SA password out of `argv`.**

ADO masks this in logs, so this is not a leak — but `$(SQL_PASSWORD)` is macro-expanded before exec, so the plaintext password sits in the agent's process table for the ~110 minutes this `docker exec` runs.

`docker exec -e VAR` with a bare name (no `=`) inherits from the client environment, which makes this a one-line improvement:

```yaml
- script: |
    docker exec \
      -e DB_CONNECTION_STRING \
      -e PYTEST_FILE_TIMEOUT="${{ parameters.pytestFileTimeout }}" \
      ...
  env:
    DB_CONNECTION_STRING: "Server=sqlserver,1433;Database=TestDB;Uid=SA;Pwd=$(SQL_PASSWORD);TrustServerCertificate=yes"
```

This also makes the currently-dead `env: SQL_PASSWORD` on this step meaningful. The `sqlcmd -P "$(SQL_PASSWORD)"` calls above have the same property but are short-lived, so they are lower priority.

.pipeline/scripts/clone-mssql-python.sh
```diff
@@ -0,0 +1,56 @@
...
+    fi
+  else
+    echo "PR description is empty or could not be retrieved"
+  fi
+else
+  echo "Not a PR build (SYSTEM_PULLREQUEST_PULLREQUESTID not set)"
+fi
+
+echo "Cloning microsoft/mssql-python (branch: $MSSQL_PYTHON_BRANCH) into $CLONE_DIR..."
+if ! git clone --depth 1 -b "$MSSQL_PYTHON_BRANCH" https://github.com/microsoft/mssql-python.git "$CLONE_DIR" 2>/dev/null; then
```
saurabh500, line 60: **Suggestion — log the resolved upstream SHA so results stay attributable.**

The clone floats on `main`. The stated value of the new job is that the delta between it and the stock-driver job is attributable to the ODBC driver — but with an unpinned upstream, a red-to-redder change could equally be an upstream test change landing between the two runs.

This matches the pre-existing job's behaviour so it is at least consistent, and cross-repo validation arguably *wants* tip-of-main. At minimum make it reconstructable after the fact:

```bash
git -C "$CLONE_DIR" rev-parse HEAD | sed 's/^/mssql-python HEAD: /'
```

.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,266 @@
...
+    set -euo pipefail
+    rm -rf odbc-swap-drop && mkdir -p odbc-swap-drop
+
+    bash .pipeline/scripts/docker-cargo-run.sh --rm \
+      -v "$PWD:/workspace" \
+      -v "$HOME/.cargo/registry:/root/.cargo/registry" \
+      -v "$HOME/.cargo/git:/root/.cargo/git" \
+      -w /workspace \
+      ${{ parameters.buildImage }} \
+      bash -c '
```
saurabh500: **Suggestion — extract this inline build to a `containerized-*.sh` per repo convention.**

The `docker-cargo-run.sh` wrapper is used correctly, but every other containerized build in this repo passes a script as the container command rather than inlining `bash -c '...'`. Compare `build-odbc-template.yml:32`:

```yaml
        ${{ parameters.buildImage }} \
        /workspace/.pipeline/scripts/containerized-odbc-build.sh
```

The other three shell blocks in this PR were correctly extracted into scripts, so this one being inline is the odd one out. A `containerized-odbc-swap-build.sh` would match, and would get the root-owned-`target/` cleanup that `containerized-odbc-build.sh:29` already does for free.

(To be clear: rebuilding rather than consuming the `Odbc_Linux_x64` artifact is unavoidable — I checked, `build-odbc-template.yml` is gated on `ne(variables['Build.Reason'], 'PullRequest')`.)

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,213 @@
...
+
+write_report_stub() {
+    local name="$1" kind="$2" reason="$3" out="$4"
+    local safe_reason
+    safe_reason="$(xml_escape "$reason")"
+    if [ "$kind" = "skipped" ]; then
+        cat > "$out" <<XML
+<?xml version="1.0" encoding="utf-8"?>
+<testsuites>
+  <testsuite name="$name" tests="1" errors="0" failures="0" skipped="1" time="0">
```
saurabh500: **Nit — `$name` is interpolated raw while `$reason` is escaped.**

`safe_reason` goes through `xml_escape`, but `$name` lands directly in `name="..."` here and in `classname="mssql_odbc_swap.$name"` on the next line (and again in the error branch below). Filenames come from a clone whose branch is PR-description-controlled, so a pathological name yields malformed XML and a confusing `PublishTestResults` failure. The helper already exists:

```bash
local safe_name
safe_name="$(xml_escape "$name")"
```

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,213 @@
...
+        2)   status="INTERRUPTED";         kind="error";   failed=$((failed + 1)) ;;
+        3)   status="INTERNAL ERROR";      kind="error";   failed=$((failed + 1)) ;;
+        4)   status="USAGE ERROR";         kind="error";   failed=$((failed + 1)) ;;
+        # Every test in the file was deselected by pytest.ini's `-m "not stress"`.
+        5)   status="NO TESTS COLLECTED";  kind="skipped"; empty=$((empty + 1)) ;;
+        # `timeout` could not run the command at all - a broken venv or missing
+        # interpreter, not a driver defect. Do not report it as a crash.
+        125|126|127)
+            status="HARNESS ERROR (exit $rc)"; kind="error"; harness_error=$((harness_error + 1)) ;;
+        124|137) status="TIMED OUT after ${slice}s"; kind="error"; timedout=$((timedout + 1)) ;;
```
saurabh500: **Nit — 137 is also what an OOM-kill looks like.**

`124` is unambiguously `timeout`, but `137` (128+SIGKILL) is equally what the kernel OOM killer produces — quite plausible for a driver still under development. Reporting both as `TIMED OUT after ${slice}s` will send someone chasing a hang that was really memory exhaustion. Printing elapsed wall time alongside would disambiguate the two at a glance:

```bash
124) status="TIMED OUT after ${slice}s" ;;
137) status="KILLED (SIGKILL after ${elapsed}s of ${slice}s - timeout or OOM)" ;;
```

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,213 @@
...
+    <testcase classname="mssql_odbc_swap.$name" name="pytest_process" time="0">
+      <error type="ProcessTerminated" message="$safe_reason">The pytest process for $name terminated abnormally under the mssql-odbc driver, so no per-test results were produced for this file.</error>
+    </testcase>
+  </testsuite>
+</testsuites>
+XML
+    fi
+}
+
+mapfile -t TEST_FILES < <(find tests -maxdepth 1 -name 'test_*.py' -type f | sort)
```
saurabh500: **Nit — `-maxdepth 1` silently drops future subdirectories.**

Correct today — I verified upstream `tests/` is flat, 41 files, no subdirs. But if upstream ever adds `tests/<subdir>/test_*.py` those files vanish from this job with no warning, and the per-file summary would still look complete. Either drop `-maxdepth 1`, or record the assumption in a comment so the next reader knows it was deliberate.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,213 @@
...
+            skipped=$((skipped + 1))
+        done
+        echo ""
+        echo "##[warning]Total time budget exhausted - skipped $skipped remaining file(s)"
+        break
+    fi
+
+    # Never let one file overrun what is left of the aggregate budget.
+    slice="$FILE_BUDGET_S"
+    [ "$slice" -gt "$remaining" ] && slice="$remaining"
```
saurabh500: **Nit — a clamped slice reads like a driver hang.**

Once the aggregate budget nears exhaustion, `slice` can clamp to as little as 31s (the `-le 30` guard above). The file then reports `TIMED OUT after 31s`, which looks like a fast driver hang rather than budget pressure — exactly the kind of false signal the rest of this script works hard to avoid. Worth labelling clamped slices distinctly, e.g. `TIMED OUT after 31s (budget-clamped)`.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,213 @@
...
+            kind="error"
+            crashed=$((crashed + 1))
+            ;;
+    esac
+
+    # A fully-deselected file still leaves a nonempty JUnit report holding zero
+    # test cases, which surfaces as nothing at all in the Tests tab. Overwrite it
+    # so every file has exactly one visible outcome.
+    if [ "$kind" = "skipped" ]; then
+        write_report_stub "$name" skipped "$status" "$report"
```
saurabh500: **Nit — unquoted `skipped` here, quoted at the other call site.**

Line 133 passes `"skipped"`; this one passes it bare. Harmless, purely a consistency thing.

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
```diff
@@ -0,0 +1,84 @@
...
+
+cd "$MSSQL_PYTHON_DIR"
+
+VERSION_FILE="mssql_python_odbc/__init__.py"
+if [ ! -f "$VERSION_FILE" ]; then
+    echo "##[error]$MSSQL_PYTHON_DIR/$VERSION_FILE not found - is this an mssql-python checkout?"
+    exit 1
+fi
+
+DRIVER_VERSION=$(awk -F= '/^__version__/ {gsub(/[[:space:]"]/, "", $2); print $2; exit}' "$VERSION_FILE")
```
saurabh500: **Nit — the `gsub` strips `"` but not `'`.**

Works today (upstream is `__version__ = "18.6.2.1"`), and if upstream ever switches to single quotes this fails loudly at the `-z` guard rather than silently producing a wrong filename — so genuinely minor. Still, handling both costs nothing:

```bash
awk -F= '/^__version__/ {gsub(/[[:space:]"'\'']/, "", $2); print $2; exit}'
```

.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,266 @@
...
+      fi
+      sleep 2
+    done
+
+    # mssql-python's own pipeline runs against a database named TestDB.
+    docker exec sqlserver-odbc-swap /opt/mssql-tools18/bin/sqlcmd \
+      -S localhost -U SA -P "$(SQL_PASSWORD)" -C -Q "CREATE DATABASE TestDB"
+  displayName: Start SQL Server
+  env:
+    SQL_PASSWORD: $(SQL_PASSWORD)
```
saurabh500, line 131: **Nit — dead `env:` block.**

The script body uses the `$(SQL_PASSWORD)` ADO macro, which is expanded by the agent before the script runs, so this `env:` mapping is never read. Same on the test step further down. Either remove it, or switch the script to `"$SQL_PASSWORD"` so the value stays out of the expanded command text.

.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,266 @@
...
+        bash /mssql-rs-scripts/run-mssql-python-odbc-tests.sh
+      '
+  displayName: Run mssql-python tests against mssql-odbc
+  # The runner enforces its own budget; this only backstops a wedged `docker
+  # exec`. It must stay below the job timeout so the publish and cleanup steps
+  # below still run.
+  timeoutInMinutes: 130
+  continueOnError: true
+  env:
+    SQL_PASSWORD: $(SQL_PASSWORD)
```
saurabh500: **Nit — dead `env:` block (same as the Start SQL Server step).**

Nothing in this step references `SQL_PASSWORD`; the connection string is built with the `$(SQL_PASSWORD)` macro on the `docker exec` line above. My `DB_CONNECTION_STRING` suggestion on that line would give this block an actual purpose.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
saurabh500: Agreed — the distinction was cosmetic without a distinct exit code. Split it: the runner now exits `2` when the harness itself could not run the tests (failed `cd`, no test files discovered, unusable `python -m pytest`, or any file returning 125/126/127) and `1` when tests actually ran and did not pass.

The matching step change is in your other comment: `continueOnError: true` is gone, and only `2` (plus docker's own 125/126/127, which mean `docker exec` never launched the runner) becomes a real error.

.pipeline/templates/test-mssql-python-odbc-template.yml
saurabh500: Done, paired with the runner's new exit contract. Dropped `continueOnError: true`, captured the status with `|| rc=$?`, and:

```yaml
case "$rc" in
  2|125|126|127)
    echo "##vso[task.logissue type=error]mssql-python test harness could not run (exit $rc) ..."
    exit 1
    ;;
esac
```

Anything else logs a warning and exits 0. I folded docker's own 125/126/127 into the error case too, since those mean the container command never started — same class of problem as the runner's exit 2.

.pipeline/templates/validation-stages.yml
saurabh500: Made the call to keep it always-on and recorded the reasoning in the job comment, since you're right that the existing block only justified the timeouts:

> Deliberately always-on rather than behind an opt-in parameter like RunFuzz / RunLongHaul: the point is a continuous compatibility signal for a driver under active development, which an opt-in flag nobody remembers to set would not give. 240m is a ceiling for the pathological all-files-hang case, not the expected runtime — the runner's own 110m budget bounds the normal path.

Happy to switch to a `mssql-odbc/**` path filter or a nightly schedule if pool capacity turns out to be the binding constraint in practice; that's a one-line change once there's real runtime data.

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
saurabh500: Done — asking Python directly now:

```bash
PKG_DIR="$(python -c 'import os, mssql_python_odbc; print(os.path.dirname(os.path.abspath(mssql_python_odbc.__file__)))' ...)"
```

`$PKG_DIR`'s parent is added as a search root, and the roots are de-duplicated so overlapping ones don't swap and count the same file twice.

The `swapped -eq 0` guard is now the stronger assertion you asked for: it hard-fails if no swapped target lives under `$PKG_DIR`, printing the targets it did swap. That closes exactly the silent-stock-driver path you describe. Verified both directions with a shimmed interpreter — it exits 1 with the offending list when the importable package is missed, and 0 when it isn't.

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
saurabh500: Done — the swap loop now collects into `SWAPPED_TARGETS[]` and `ldd` probes each entry, so the check does real work on any platform/arch layout instead of silently no-op'ing off the hardcoded path.

Good catch on the soname. I've left `build.rs` alone since `dlopen` by absolute path ignores `SONAME` and changing it is out of scope here, but I've noted it as the first suspect if the first run dies in the loader.

.pipeline/templates/test-mssql-python-odbc-template.yml
saurabh500: Added the pre-cleanup — it's cheap insurance and correct either way, so it didn't seem worth blocking on resolving the pool's statefulness first:

```bash
docker rm -f mssql-python-odbc-tests || true
docker rm -f sqlserver-odbc-swap || true
docker network rm mssqlpy-odbc-net || true
```

On the root-owned `target/`: the build moved into `containerized-odbc-swap-build.sh` (see your other comment), which does its own `rm -rf "$DROP_DIR"` from inside the container as root, matching `containerized-odbc-build.sh:27`. The shared `target/` itself is left alone deliberately — it's the cargo cache the other containerized builds on this pool already reuse, so clearing it here would just slow every job down without fixing anything this PR introduced.

.pipeline/templates/test-mssql-python-odbc-template.yml
saurabh500: Dropped, with the reasoning recorded inline so it doesn't get re-added:

> No host port published: the test container reaches SQL Server over the user-defined network, and binding 1433 on the agent would collide with the sibling mssql-python job scheduled onto the same pool.

.pipeline/templates/test-mssql-python-odbc-template.yml
saurabh500: Done, exactly as suggested — `docker exec -e DB_CONNECTION_STRING` with a bare name, and the value supplied by the step's `env:` block, so the password never reaches argv.

I also took the lower-priority half: the `Start SQL Server` step now uses `"$SQL_PASSWORD"` from its own `env:` block for `MSSQL_SA_PASSWORD` and both `sqlcmd` calls, rather than the macro. That resolves the two dead-`env:` threads at the same time.

.pipeline/scripts/clone-mssql-python.sh
saurabh500: Added, right after the clone:

```bash
echo "##[section]mssql-python HEAD: $(git -C "$CLONE_DIR" rev-parse HEAD)"
```

Kept the float on `main` for consistency with the pre-existing job — as you say, cross-repo validation wants tip-of-main — so this is purely about making a red-to-redder delta reconstructable after the fact.

.pipeline/templates/test-mssql-python-odbc-template.yml
saurabh500: Done — extracted to `.pipeline/scripts/containerized-odbc-swap-build.sh`, and the step now passes it as the container command like `build-odbc-template.yml:32`. It documents why it's narrower than `containerized-odbc-build.sh` (only the Rust cdylib is wanted, so no cmake / unixODBC headers) and does its own root-owned drop-dir cleanup.

And agreed on the rebuild — `build-odbc-template.yml` being gated on `ne(variables['Build.Reason'], 'PullRequest')` means the artifact isn't there to consume on a PR.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
saurabh500: Fixed — `safe_name="$(xml_escape "$name")"` now feeds the suite `name`, the `classname`, and the message body in both the skipped and error branches.

Verified with a file literally named `test_a&b<c>_hang.py`: the report is well-formed and renders as `name="test_a&amp;b&lt;c&gt;_hang"`.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
saurabh500: Good point — split 137 out of the `124|137` case, and both now carry elapsed wall time so the two causes are distinguishable at a glance:

```bash
124) status="TIMED OUT after ${elapsed}s (limit ${slice}s${clamp_note})" ;;
137) status="KILLED by SIGKILL after ${elapsed}s (limit ${slice}s${clamp_note} - timeout or OOM)" ;;
```

.pipeline/scripts/run-mssql-python-odbc-tests.sh
saurabh500: Dropped `-maxdepth 1` — a silent omission is worse than the marginally wider glob — and left a comment recording why:

> Not depth-limited: upstream's tests/ is flat today, but a future subdirectory would otherwise vanish from this job while the summary still looked complete.

Verified that a `tests/sub/test_nested.py` is now discovered.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
saurabh500: Added — a clamped slice sets `clamp_note=", budget-clamped"`, which appears in both the timeout and SIGKILL statuses.

Verified: an unclamped file reads `TIMED OUT after 40s (limit 40s)`, and the next one under budget pressure reads `TIMED OUT after 35s (limit 35s, budget-clamped)`.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
saurabh500: Quoted for consistency.

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
saurabh500: Done — the character class strips single quotes too. Verified both `__version__ = "18.6.2.1"` and `__version__ = '18.6.2.1'` parse to `18.6.2.1`.

.pipeline/templates/test-mssql-python-odbc-template.yml
saurabh500: Took the second option — the script now uses `"$SQL_PASSWORD"` for `MSSQL_SA_PASSWORD` and both `sqlcmd -P` calls, so the block is live and the value stays out of the expanded script text rather than being macro-expanded into it.

.pipeline/templates/test-mssql-python-odbc-template.yml
saurabh500: Replaced with `DB_CONNECTION_STRING`, per your suggestion on the `docker exec` line — the step forwards it with a bare `-e DB_CONNECTION_STRING`, so the block now has a real purpose and the password stays out of argv.

David-Engel (commented):
## Summary

Adds an informational PR job that runs the upstream `microsoft/mssql-python` pytest suite against the Rust `mssql-odbc` driver by overwriting the bundled `libmsodbcsql-18.6.so.2.1`, plus a crash/hang-resilient per-file runner, plus an extraction of the shared mssql-python clone logic. The design is good and it already produced a real, useful result on this PR (build 166295: 40 files — 10 passed / 23 failed / 3 crashed (SIGABRT) / 4 deselected — in 9m28s, job ended `succeededWithIssues` and reported `pass` to GitHub as described).

I verified the swap and the runner rather than taking the description on faith:

- Ran `run-mssql-python-odbc-tests.sh` locally against fixtures for pass / fail / SIGSEGV / hang / `exit 5`. Every failure mode is classified correctly, synthetic JUnit stubs are written for the files that die, and the loop never stalls. The exit-code split (2 = harness, 1 = driver, 0 = clean) works as documented.
- Confirmed against upstream `main` that the swap script's assumptions hold: `mssql_python_odbc/__init__.py` carries `__version__ = 18.6.2.1`, and there are exactly 7 committed Linux copies of `libmsodbcsql-18.6.so.2.1`. Build 166295's log shows all 7 overwritten, the importable package correctly resolved to `/workspace/mssql_python_odbc`, and clean `ldd` output.
- The container setup faithfully mirrors upstream's Ubuntu leg, including the `libodbcinst.so.2` removal.

**Blocking: none.** Everything below is a suggestion or a nit.

Two things that aren't tied to a specific line:

**PR description is out of sync with the diff.** It says "The test step is `continueOnError: true`" — the YAML doesn't use `continueOnError` on that step at all; it uses explicit `rc` handling plus `##vso[task.complete result=SucceededWithIssues]`, which is a better design than what's described. The Files table also omits `containerized-odbc-swap-build.sh`, and the claim that the `mssql-python-branch:` override "still works for both jobs" is not true on GitHub-sourced builds (see the inline comment on `clone-mssql-python.sh`). Worth a refresh before merge.

**Worth a conscious decision: this permanently makes every PR build "partially succeeded".** The job is always-on and expected to be red for the foreseeable future, so the build-level `SucceededWithIssues` signal stops meaning anything for every *other* job in the pipeline. The per-job GitHub check still carries the real signal, so I'd keep it — but flagging that the build-level status is being spent here.

**Nit:** `run-mssql-python-odbc-tests.sh` and `swap-mssql-python-odbc-driver.sh` are committed `100644` while `containerized-odbc-swap-build.sh` is `100755`. All three are invoked as `bash <script>` so it's cosmetic, but the repo is otherwise mixed too — worth picking one.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,244 @@
...
+#   1 -> tests ran and did not fully pass (driver defect, advisory)
+#   0 -> everything passed
+# Files skipped because the aggregate budget ran out count as "not clean" too:
+# the result set is incomplete, which must not read as a clean pass.
+if [ "$harness_error" -gt 0 ]; then
+    echo "##[error]$harness_error file(s) could not be executed - this indicates a broken test environment, not a driver defect"
+    exit 2
+fi
+
+if [ "$failed" -gt 0 ] || [ "$crashed" -gt 0 ] || [ "$timedout" -gt 0 ] || [ "$skipped" -gt 0 ]; then
```
David-Engel, line 260: **Suggestion:** a run in which *every* file reports `NO TESTS COLLECTED` exits 0, and the step reports a clean green pass even though zero tests executed. Verified locally:

```
$ # 2 test files, fake pytest that always exits 5
$ bash run-mssql-python-odbc-tests.sh; echo $?
0
```

`empty` is deliberately (and correctly) excluded from the not-clean condition so a single deselected stress file doesn't turn the run yellow — but the all-empty case is indistinguishable from success. An upstream `pytest.ini`/`addopts` change is enough to trigger it, and the job would go quietly green while testing nothing.

One-line guard:

```bash
if [ "$((passed + failed + crashed + timedout))" -eq 0 ]; then
    echo "##[error]No test file actually executed any tests - the suite was not exercised"
    exit 2
fi
```

.pipeline/scripts/run-mssql-python-odbc-tests.sh
```diff
@@ -0,0 +1,244 @@
...
+echo "Per-file timeout: $PYTEST_FILE_TIMEOUT (${FILE_BUDGET_S}s) | total budget: $PYTEST_TOTAL_BUDGET (${TOTAL_BUDGET_S}s)"
+
+declare -a SUMMARY=()
+passed=0 failed=0 crashed=0 timedout=0 empty=0 skipped=0
+harness_error=0
+
+for idx in "${!TEST_FILES[@]}"; do
+    test_file="${TEST_FILES[$idx]}"
+    name="$(basename "$test_file" .py)"
+    report="$RESULTS_DIR/results-$name.xml"
```
David-Engel, line 144: **Suggestion:** the report name is derived from the basename only, which defeats the `find` that was deliberately left un-depth-limited a few lines above ("a future subdirectory would otherwise vanish from this job while the summary still looked complete"). Two files with the same basename in different directories collapse onto one report. Verified locally with `tests/test_a.py` + `tests/sub/test_a.py`:

```
test_a    PASSED
test_a    PASSED
files: 2 | passed: 2 | ...
$ ls res/
results-test_a.xml        # one file, second run overwrote the first
```

Flat `tests/` today so this is latent, but it's the exact scenario the `find` comment is guarding against. Path-derived name fixes it:

```bash
name="${test_file#tests/}"; name="${name%.py}"; name="${name//\//_}"
```

(applies to the budget-exhausted stub at line 141 too)

.pipeline/scripts/clone-mssql-python.sh
```diff
@@ -0,0 +1,60 @@
...
+
+if [ -n "${SYSTEM_PULLREQUEST_PULLREQUESTID:-}" ]; then
+  echo "PR build detected (PR #${SYSTEM_PULLREQUEST_PULLREQUESTID})"
+
+  echo "Installing azure-devops CLI extension..."
+  az extension add --name azure-devops --yes --allow-preview true 2>/dev/null || echo "Extension may already be installed"
+
+  az devops configure --defaults organization="${SYSTEM_COLLECTIONURI}" project="${SYSTEM_TEAMPROJECT}"
+
+  PR_DESC=$(az repos pr show --id "${SYSTEM_PULLREQUEST_PULLREQUESTID}" --query description -o tsv 2>&1) || {
```
David-Engel, line 37: **Suggestion:** this branch-override path is dead on GitHub-sourced builds, and the PR description states it still works. Pure extraction so it's pre-existing — but this PR is promoting it to shared infrastructure with a doc comment that describes a feature that no longer functions.

From this PR's own run (build 166295, `Clone mssql-python`):

```
PR build detected (PR #4255782867)
##[warning]Failed to get PR description: WARNING: Auto-detect was enabled but no Azure DevOps
remote was found in the current git repository. ...
ERROR: The request is invalid.
Cloning microsoft/mssql-python (branch: main)...
```

Two things went wrong: `SYSTEM_PULLREQUEST_PULLREQUESTID` is the GitHub *internal* PR id (`4255782867`), not `222`; and `az repos pr show` queries ADO Repos, which this build isn't (`repository.type == GitHub`). It fails safe — falls back to `main` — but it costs an `az extension add` on every run of both jobs and can never honour an override.

Either gate it so the cost and the misleading warning go away:

```bash
if [ "${BUILD_REPOSITORY_PROVIDER:-}" = "TfsGit" ] && [ -n "${SYSTEM_PULLREQUEST_PULLREQUESTID:-}" ]; then
```

or switch to the GitHub API using `SYSTEM_PULLREQUEST_PULLREQUESTNUMBER`. Fine to do as a follow-up; the doc comment and the PR description should stop claiming it works either way.

.pipeline/templates/validation-stages.yml
```diff
@@ -525,6 +525,33 @@ stages:
...
+    displayName: mssql-python suite on mssql-odbc driver (cross-repo)
+    timeoutInMinutes: 240
```
David-Engel: **Suggestion:** 240m is a lot of headroom to reserve given the observed runtime. This PR's own run finished the whole job in **9m28s**, and the test step is already capped at `timeoutInMinutes: 130` with the runner capped at 110m internally.

The extra ~110m only ever comes into play if something wedges *outside* the test step (image pull, apt, `cargo build --release`, the pybind build) — and in that case the job holds a `RUST-1ES-POOL-WUS3` agent for four hours before ADO reclaims it. ~150m still covers 130m of test step plus a generous setup allowance while bounding the pathological case to something closer to reality.

Not a correctness issue — the nesting rationale in the comment above is right, it's just the outermost number that looks over-provisioned.

.pipeline/templates/test-mssql-python-odbc-template.yml
```diff
@@ -0,0 +1,292 @@
...
+# Upstream purges the ODBC build toolchain and the system libodbcinst before
+# testing, so the bundled copy next to the driver is what loads. Mirror it: the
+# only intended difference from an upstream run is the driver binary itself.
+- script: |
+    set -euo pipefail
+    docker exec mssql-python-odbc-tests bash -c '
+      set -euo pipefail
+      export DEBIAN_FRONTEND=noninteractive
+      apt-get remove --purge -y unixodbc-dev
+      rm -f /lib/x86_64-linux-gnu/libodbcinst.so.2 /usr/lib/x86_64-linux-gnu/libodbcinst.so.2
```
David-Engel: **Nit:** hardcoded `x86_64` multiarch path. It mirrors upstream so it's defensible, but `swap-mssql-python-odbc-driver.sh` went out of its way *not* to hardcode platform/arch ("so this stays meaningful on any platform/arch layout"), and this step would silently no-op on an arm64 agent — leaving the system `libodbcinst` in place and quietly changing what the suite loads. Since the step now runs under `set -euo pipefail`, a glob keeps it both arch-agnostic and non-fatal:

```bash
rm -f /lib/*-linux-gnu/libodbcinst.so.2 /usr/lib/*-linux-gnu/libodbcinst.so.2
```

.pipeline/scripts/containerized-odbc-swap-build.sh
```diff
@@ -0,0 +1,30 @@
...
+DROP_DIR="${ODBC_DROP_DIR:-/workspace/odbc-swap-drop}"
+
+# The container runs as root, so anything it wrote on a previous run is
+# root-owned on the agent and cannot be cleaned host-side without sudo.
+rm -rf "$DROP_DIR"
+mkdir -p "$DROP_DIR"
+
+cargo build --release -p mssql-odbc
+
+cp /workspace/target/release/libmsodbcsql18.so "$DROP_DIR/"
```
David-Engel: **Nit:** hardcoded `/workspace/target`. It's correct today because this job — unlike its two siblings in the same stage — doesn't set `CARGO_TARGET_DIR`. If someone adds `variables: CARGO_TARGET_DIR: ...` to `Test_mssql_python_on_mssql_odbc` for consistency with `Build_mssql_python_Linux`, this fails with a bare "No such file or directory" after a full release build. Cheap to make robust:

```bash
TARGET_DIR="${CARGO_TARGET_DIR:-/workspace/target}"
cp "$TARGET_DIR/release/libmsodbcsql18.so" "$DROP_DIR/"
```

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
```diff
@@ -0,0 +1,114 @@
...
+        exit 1
+    fi
+fi
+
+echo "Replaced ${#SWAPPED_TARGETS[@]} bundled driver binaries with mssql-odbc"
+
+# Surface unresolved shared-library dependencies here rather than as an opaque
+# import failure inside the first test. Probe what was actually swapped so this
+# stays meaningful on any platform/arch layout.
+for target in "${SWAPPED_TARGETS[@]}"; do
```
David-Engel, line 114: **Nit:** this `ldd`s all 7 swapped copies, which are byte-identical, producing 49 lines of duplicate output in the log (confirmed in build 166295). Five of those copies are arm64 or musl slots that an x86_64 glibc build can never be loaded from anyway, so their `ldd` output is meaningless.

Probing only the copy under the importable package — the one the resolver will actually `dlopen` — keeps the diagnostic value and drops the noise:

```bash
for target in "${SWAPPED_TARGETS[@]}"; do
    case "$target" in "${PKG_DIR:-__none__}"/*) echo "ldd $target"; ldd "$target" || true ;; esac
done
```

Unrelated low-risk observation while I was here: `mssql-odbc/build.rs` embeds `DT_SONAME = libmsodbcsql-18.4.so.1.1` while the file it's copied over is `libmsodbcsql-18.6.so.2.1`. Harmless for a single `dlopen` by absolute path (which is what `GetDriverPathCpp` does), so nothing to change in this PR — just noting it in case a future soname-based load path shows up.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
saurabh500: Good catch, and the reproduction is exactly right — that path exits 0 with every counter at zero, so the job would report a clean pass having exercised nothing.

Added the guard in `ae8c404e`, keyed on executed files rather than the `empty` count so the single-deselected-stress-file case you called out stays green:

```bash
if [ "$((passed + failed + crashed + timedout))" -eq 0 ]; then
    echo "##[error]No test file executed any tests (no tests: $empty, skipped: $skipped) - the suite was not exercised"
    exit 2
fi
```

Exit 2 rather than 1 so it lands as a real red: an upstream `pytest.ini`/`addopts` change breaking collection says nothing about the driver, which is the same reasoning as the existing harness-error path. Verified offline — all-empty now exits 2, one-empty-plus-one-pass still exits 0.

.pipeline/scripts/run-mssql-python-odbc-tests.sh
saurabh500: Agreed — this directly undercuts the `find` comment a few lines up, so it's worth fixing now rather than waiting for upstream to add a subdirectory.

`ae8c404e` derives the name from the path under `tests/`:

```bash
report_name() {
    local n="${1#tests/}"
    n="${n%.py}"
    printf '%s' "${n//\//_}"
}
```

Applied at both call sites, including the budget-exhausted stub you flagged. Verified with `tests/test_dup.py` + `tests/sub/test_dup.py`: two reports now survive (`results-test_dup.xml`, `results-sub_test_dup.xml`) where previously the second overwrote the first.

.pipeline/scripts/clone-mssql-python.sh
saurabh500: Confirmed — `BUILD_REPOSITORY_PROVIDER` is `GitHub` on these builds, so `az repos pr show` was never going to resolve, and it burned an `az extension add` plus that warning on every run of both jobs.

Gated it in `ae8c404e` rather than deferring, since it's a couple of lines:

```bash
if [ "${BUILD_REPOSITORY_PROVIDER:-}" = "TfsGit" ] && [ -n "${SYSTEM_PULLREQUEST_PULLREQUESTID:-}" ]; then
```

The doc comment now says the override only resolves on ADO Repos-sourced builds, and I've corrected the PR description, which claimed it still worked. Since this script is shared with the pre-existing job, I verified all three modes against stubbed `git`/`az`: GitHub PR build skips the lookup and still clones main, non-PR is unchanged, and a TfsGit PR build still attempts the override.

Left the GitHub API alternative alone — nobody is using the override today, and adding a live API call to the clone path is more risk than the feature is worth right now.

.pipeline/templates/validation-stages.yml
saurabh500: Fair — 240m was sized before there was any runtime data, and holding a `RUST-1ES-POOL-WUS3` agent for four hours in the wedged case is a real cost.

Dropped to 150m in `ae8c404e`, which still clears the 130m test-step cap with ~20m for image pull, apt, `cargo build --release` and the pybind build. Updated the comment to cite the observed ~9m run instead of leaving the number unexplained.

.pipeline/templates/test-mssql-python-odbc-template.yml
saurabh500: Agreed — it's inconsistent with the swap script going out of its way to stay arch-agnostic, and a silent no-op here would quietly change what the suite loads.

Switched to the glob in `ae8c404e`. `rm -f` keeps it non-fatal under `set -euo pipefail` when nothing matches.

.pipeline/scripts/containerized-odbc-swap-build.sh
saurabh500: Agreed, cheap insurance against a foot-gun that would only surface after a full release build. `ae8c404e` uses `TARGET_DIR="${CARGO_TARGET_DIR:-/workspace/target}"` and documents the variable in the env block.

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
saurabh500: Agreed — the copies are byte-identical and the arm64/musl slots can't be loaded by this build, so 49 lines of identical output buys nothing.

`ae8c404e` narrows it to the copies under `$PKG_DIR`, with a fallback to the first swapped target if the package dir could not be resolved, so the diagnostic never disappears entirely:

```bash
probed=0
for target in "${SWAPPED_TARGETS[@]}"; do
    case "$target" in
        "${PKG_DIR:-__unresolved__}"/*) echo "ldd $target"; ldd "$target" || true; probed=1 ;;
    esac
done
[ "$probed" -eq 0 ] && { echo "ldd ${SWAPPED_TARGETS[0]}"; ldd "${SWAPPED_TARGETS[0]}" || true; }
```

On the `build.rs` soname: agreed it's inert for the absolute-path `dlopen` that `GetDriverPathCpp` does, and out of scope here. Noting it explicitly so it's on the record if a soname-based load path ever appears.

David-Engel (approved):
Approving — all seven findings are addressed in `ae8c404e`, and I re-verified each one rather than taking the commit message for it.

**Verified locally** (fixtures + a faked `az`/`git` on `PATH`):

| Finding | Result |
|---|---|
| All-deselected run exits green | Now exits `2` with a clear `##[error]`. Re-ran the full matrix: all-empty → 2, mixed pass/fail → 1, all-pass → 0, broken venv → 2. |
| Report basename collision | `report_name()` keys by path under `tests/`; `tests/test_a.py` + `tests/sub/test_a.py` now produce `results-test_a.xml` and `results-sub_test_a.xml` instead of one file silently overwriting the other. |
| Dead ADO branch lookup | Skips cleanly on a faked `BUILD_REPOSITORY_PROVIDER=GitHub`, and still performs the lookup under `TfsGit` — the ADO path isn't collateral damage. |
| `CARGO_TARGET_DIR` | Honoured. |
| Multiarch `libodbcinst` glob | Correct. |
| 150m job timeout | Correct, and the comment now carries the ~9m data point that justifies it. |

**Verified in CI** (build 166306, on `ae8c404e`):

- Clone step is now quiet: `No Azure DevOps Repos PR context - skipping branch-override lookup` — no `az extension add`, no misleading warning, on either cross-repo job.
- The new job still produces the same result as before the hardening — 40 files, 10 passed / 23 failed / 3 crashed / 4 deselected in 9m44s, ending `succeededWithIssues` and reporting `pass` to GitHub. No regression from the changes.

Nothing here blocks merge. Two follow-ups, neither worth another round-trip:

**The `ldd` narrowing is a no-op in practice.** In build 166306 the log still shows all 7 `ldd` blocks, because every swapped copy lives under `$PKG_DIR` (`/workspace/mssql_python_odbc`) — the filter only excludes a hypothetical site-packages copy. The code comment says "probe only the copy under the importable package", which overstates it. A `break` after the first match would deliver what the comment describes. Purely log noise, so entirely your call whether it's worth a commit.

**Please refresh the description before merge.** The `mssql-python-branch:` correction landed, thanks — a few things are still stale:

- "The test step is `continueOnError: true`" — it isn't. The step uses explicit `|| rc=$?` handling plus `##vso[task.complete result=SucceededWithIssues]`, which is a better design than what's written; the description undersells the change. The header comment in `test-mssql-python-odbc-template.yml` (line 15) repeats the same stale claim.
- The Files table lists 6 of the 7 changed files — `.pipeline/scripts/containerized-odbc-swap-build.sh` is missing.
- "a 180-minute job cap" matches neither the old 240m nor the new 150m.
- The new exit contract (a run where no file executes a test is now a hard harness error, not a silent pass) isn't described, and it's one of the more valuable behaviours in the runner.

.pipeline/scripts/swap-mssql-python-odbc-driver.sh
David-Engel: Follow-up on this one: the narrowing is a no-op in practice. Build 166306 still logs all 7 `ldd` blocks, because every swapped copy lives under `$PKG_DIR` (`/workspace/mssql_python_odbc`) — the `case` filter only excludes a hypothetical site-packages copy, which doesn't exist in this job:

```
python resolves mssql_python_odbc to: /workspace/mssql_python_odbc
Replaced 7 bundled driver binaries with mssql-odbc
ldd /workspace/mssql_python_odbc/libs/linux/suse/x86_64/lib/libmsodbcsql-18.6.so.2.1
ldd /workspace/mssql_python_odbc/libs/linux/alpine/x86_64/lib/libmsodbcsql-18.6.so.2.1
... (7 total)
```

The code comment says "probe only the copy under the importable package", which overstates what the loop does. A `break` after the first match delivers what the comment describes. Log noise only — approved regardless, purely your call whether it's worth a commit.

saurabh500:
Thanks for the re-verification — both follow-ups are addressed in `14b479dc`.

> A `break` after the first match would deliver what the comment describes.

Done. You were right that the filter was a no-op: all 7 copies live under `$PKG_DIR`, so the loop printed all 7 byte-identical blocks while the comment claimed to probe one. Added the `break` and rewrote the comment to say why one probe is sufficient rather than implying the filter was doing the narrowing.

> Please refresh the description before merge.

Refreshed, all four items:

- The `continueOnError` claim is gone from both the description and the template header (line 15) — replaced with the actual design: explicit `|| rc=$?` handling mapping failures to `SucceededWithIssues`, and the note that a `SucceededWithIssues` job still reports `pass` to GitHub so it never blocks a PR.
- Added `containerized-odbc-swap-build.sh` to the Files table (now 7 of 7).
- `180-minute job cap` → `150-minute`, and the arithmetic alongside it was also stale (`15m`/600m → `10m`/400m, matching `pytestFileTimeout`).
- Documented the three-way exit contract, including that a run where no file executes a test is a hard harness error rather than a silent pass.

> All three are invoked as `bash <script>` so it's cosmetic, but the repo is otherwise mixed too — worth picking one.

Picked `100755`. All four scripts this PR adds are now executable; I left the rest of `.pipeline/scripts/` alone rather than widening the diff into unrelated files.

On the build-level `SucceededWithIssues` being permanently spent: agreed with your own read, so no change. The per-job GitHub check carries the real signal, and the alternative — making the job opt-in — would mean the regression it's meant to catch goes unnoticed by default. Worth revisiting if the driver ever gets close enough to green that the build-level status becomes meaningful again.

David-Engel (approved):

Issue #24, opened 2026-04-21
Fabric DW: next_row() fails on Fabric DW Info tokens in result stream


VaibhaveS:
### Description

When executing queries against Microsoft Fabric Data Warehouse, `next_row()` returns an error:

```
ProtocolError("Unexpected token while finding the next row: Info(InfoToken { 
    number: 15806, state: 1, severity: 0, 
    message: \"Statement ID: {2C39FC51-...} | Query hash: 0x808236332799EDF3 | ...\",
    ...
})")
```

Fabric DW injects informational `Info` tokens (severity 0) into the result stream containing tracing metadata (Statement ID, Query hash, Distributed request ID). These appear between the `ColumnMetadata` and `Row` tokens

### Environment

- Microsoft Fabric Data Warehouse
- Query: any SELECT against Fabric DW tables

### Steps to reproduce

```rust
// After successful connection to Fabric DW:
client.execute("SELECT TOP 3 l_orderkey, l_quantity FROM [tcph].[lineitem]".into(), None, None).await?;

// This fails:
let row = client.next_row().await?; // ProtocolError: Unexpected token Info
```

### Expected behavior

Info tokens with severity 0 should be treated as informational and skipped when iterating rows, similar to how the ODBC driver and other TDS clients handle them.

### Actual behavior

```
ProtocolError("Unexpected token while finding the next row: Info(InfoToken { 
    number: 15806, state: 1, severity: 0, 
    message: \"Statement ID: {2C39FC51-...} | Query hash: 0x808236332799EDF3 | ...\",
    ...
})")
```

### Version

mssql-tds v0.1.0 (commit 8869e80)

### Affected crate

mssql-tds

### Environment

| Field         | Value                                                                          |
| ------------- | ------------------------------------------------------------------------------ |
| **Rust**      | `rustc 1.93.1 (7b16f5f31e 2026-02-24)` (ms-prod-1.93 toolchain)                |
| **OS**        | Ubuntu 24.04.2 LTS *(Noble Numbat)*, kernel `6.6.87.2-microsoft-standard-WSL2` |
| **Fabric DW** | Microsoft Fabric Data Warehouse *(serverless SaaS — no user-visible version)*  |
| **Endpoint**  | `*.msit-datawarehouse.fabric.microsoft.com`                                    |

PR #224, opened 2026-08-11
Decouple TC-clash auth e2e tests from mssql-python internals

### LINKED ISSUE #223: Cross-repo auth e2e tests break on mssql-python internal renames (patch stable boundaries instead)
## Summary

Two tests in `mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py` fail in the cross-repo "Build mssql-python (cross-repo)" job:

- `TestAuthTcClash::test_row27_tc_yes_ad_interactive`
- `TestAuthTcClash::test_tc_yes_ad_default`

Both fail with:

```
AttributeError: <module 'mssql_python.auth' ...> does not have the attribute 'get_auth_token'
```

They patched `mssql_python.auth.get_auth_token`, an **internal** symbol in the upstream `microsoft/mssql-python` repo that has since been renamed/refactored (upstream now exposes `get_auth_token_info` and defers token acquisition to a native pool-miss callback). `unittest.mock.patch` refuses to patch a missing attribute, so the tests error at `__enter__` before the body runs. The step is `continueOnError: true`, so it doesn't block PRs, but it's persistent red noise on every run.

## Why this happens

These two rows only exercised the **mssql-python/ODBC leg** and mocked a mssql-python-private function purely to avoid a real interactive/default Entra credential flow in CI. Coupling the tests to an upstream-internal name makes them break whenever upstream refactors auth — which has no bearing on mssql-rs correctness.

## Fix

Decouple the tests from mssql-python internals:

1. **Assert the mssql-rs behavior directly against py-core.** The TC=Yes + `Authentication` clash is rejected by py-core's own `validate_auth` (`odbc_authentication_validator.rs` → `"Cannot use Authentication with Trusted_Connection."`) at `PyCoreConnection` construction — deterministic, no Azure infra, no mssql-python import.
2. **Keep the mssql-python/ODBC parity leg but patch a stable public boundary** — stub the `azure.identity` credential classes (`InteractiveBrowserCredential`, `DefaultAzureCredential`, ...) so no real Entra flow runs in CI, instead of patching any mssql-python-internal token helper.

## Follow-up (not in the initial fix)

Three other tests in the same file still patch `mssql_python.auth.AADAuth.get_raw_token` (lines ~962/1014/1036). That symbol still exists upstream so they currently pass, but they carry the same coupling risk and should eventually move to a stable boundary (e.g. stub `azure.identity` or the py-core access-token path).

## Notes

- This is unrelated to any specific PR's code changes; it's cross-repo drift from `mssql-python` `main`.
- Longer term, requesting a documented, public test seam from `microsoft/mssql-python` for stubbing token acquisition would prevent recurrence at the contract level.

saurabh500:
## Description

`TestAuthTcClash::test_row27_tc_yes_ad_interactive` and `TestAuthTcClash::test_tc_yes_ad_default` in `mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py` failed in the cross-repo "Build mssql-python (cross-repo)" job with:

```
AttributeError: <module 'mssql_python.auth' ...> does not have the attribute 'get_auth_token'
```

They patched `mssql_python.auth.get_auth_token` — an upstream-**internal** symbol that `microsoft/mssql-python` renamed (it now exposes `get_auth_token_info` and defers token acquisition to a native pool-miss callback). `unittest.mock.patch` refuses to patch a missing attribute, so the tests errored at `__enter__`. The step is `continueOnError: true` so it doesn't block, but it's persistent red noise driven by upstream drift, not mssql-rs behavior.

### 1. Durable guard: a Rust unit test that runs on every PR

The root-cause fix for the class of problem in #223 is to assert the mssql-rs behavior where it belongs — in Rust, with zero Python, zero Docker, and zero cross-repo coupling. `odbc_authentication_validator.rs` had `tc_plus_sqlpassword`/`tc_plus_ad_password`/`tc_plus_ad_integrated` tests, but they only checked for the `Trusted_Connection` substring (which the access-token conflict text also contains); none pinned the exact clash message. Those three are replaced by one table-driven `auth_plus_tc_rejected` that loops over every Authentication mode (`SqlPassword`, `ActiveDirectoryPassword`, `ActiveDirectoryIntegrated`, `ActiveDirectoryInteractive`, `ActiveDirectoryDefault`, `ActiveDirectoryMSI`, `ActiveDirectoryServicePrincipal`) and pins the full message with `assert_eq!` (`"Usage Error: Cannot use Authentication with Trusted_Connection."`, trailing period included). Passing empty UID/PWD with `SqlPassword` also proves the TC clash is reported before the "Both User and Password" rule, covering ordering in `validate_auth`.

### 2. Delete the two flaky Python e2e tests

Rather than re-target another mssql-python internal, remove both tests and their helpers. Two reasons, established while working through review:

- **The mssql-rs behavior is already covered, decoupled from mssql-python.** The TC=Yes + `Authentication` clash is py-core's own `validate_auth`, asserted deterministically (no Azure infra, no `mssql_python` import) in the normal py-core suite: `tests/test_auth_resolution.py::TestAuthTcClashes::test_27_ad_interactive_tc` (ActiveDirectoryInteractive) and `TestAuthTcClashesExtended::test_ad_default_tc` (ActiveDirectoryDefault). The e2e "Leg 1" I initially added just duplicated these.
- **An ODBC/mssql-python "parity" leg can't add signal for these two rows.** mssql-python **strips `Trusted_Connection`** for token-bearing AD modes: `ActiveDirectoryDefault` and non-Windows `ActiveDirectoryInteractive` route through the token factory, and `remove_sensitive_params` drops `_SENSITIVE_KEYS` (which includes `_KEY_TRUSTED_CONNECTION`) before the string reaches ODBC. So the driver never observes a clash for these modes — a test here could only "pass" by faulting elsewhere (stubbed/absent Entra infra), i.e. it's unfalsifiable. ODBC's actual behavior for these modes is covered by mssql-python's own suite.

The two tests are deleted along with the `_no_real_entra_auth` / `_assert_tc_auth_clash` helpers and the now-unused `contextmanager` / `MagicMock` imports.

Follow-up (not in this PR): three other tests in the same file still patch `mssql_python.auth.AADAuth.get_raw_token` (that symbol still exists upstream, so they pass today) and carry the same coupling risk — tracked in the issue.

## Related Issues

Fixes #223

## Checklist

- [x] `cargo bfmt` passes — `mssql-py-core` `cargo fmt -- --check` clean
- [x] `cargo bclippy` passes — `mssql-py-core` `cargo clippy --all-features --all-targets -- -D warnings` clean
- [x] `cargo btest` passes — new `auth_plus_tc_rejected` passes under `cargo nextest` in `mssql-py-core`
- [x] New/changed functionality has tests — Rust `auth_plus_tc_rejected`, plus the retained `test_auth_resolution.py` clash cases
- [x] Public API changes are documented — n/a (test-only change)

### DIFF
--- a/mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
+++ b/mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
@@ -243,41 +243,28 @@ mod tests {
     // Authentication + TC mutual exclusion
     // ---------------------------------------------------------------
 
-    #[test]
-    fn tc_plus_sqlpassword() {
-        let err = validate_auth(Some("SqlPassword"), Some(true), "sa", "secret", None).unwrap_err();
-        assert!(err.to_string().contains("Trusted_Connection"));
-    }
-
-    #[test]
-    fn tc_plus_ad_password() {
-        let err =
-            validate_auth(Some("ActiveDirectoryPassword"), Some(true), "u", "p", None).unwrap_err();
-        assert!(err.to_string().contains("Trusted_Connection"));
-    }
-
-    #[test]
-    fn tc_plus_ad_integrated() {
-        let err =
-            validate_auth(Some("ActiveDirectoryIntegrated"), Some(true), "", "", None).unwrap_err();
-        assert!(err.to_string().contains("Trusted_Connection"));
-    }
-
     /// Durable, Python-free guard for the class of drift in #223: pins the exact
-    /// clash message across representative modes (incl. the AD modes whose Python
-    /// e2e coverage was removed because mssql-python strips Trusted_Connection).
+    /// clash message across every Authentication mode (incl. the AD modes whose
+    /// Python e2e coverage was removed because mssql-python strips
+    /// Trusted_Connection). Empty UID/PWD with SqlPassword also proves the TC
+    /// clash is reported before the "Both User and Password" rule, covering
+    /// ordering in `validate_auth`.
     #[test]
     fn auth_plus_tc_rejected() {
         for auth in [
             "SqlPassword",
+            "ActiveDirectoryPassword",
+            "ActiveDirectoryIntegrated",
             "ActiveDirectoryInteractive",
             "ActiveDirectoryDefault",
+            "ActiveDirectoryMSI",
+            "ActiveDirectoryServicePrincipal",
         ] {
             let err = validate_auth(Some(auth), Some(true), "", "", None).unwrap_err();
-            assert!(
-                err.to_string()
-                    .contains("Cannot use Authentication with Trusted_Connection"),
-                "unexpected error for {auth}: {err}"
+            assert_eq!(
+                err.to_string(),
+                "Usage Error: Cannot use Authentication with Trusted_Connection.",
+                "unexpected error for {auth}"
             );
         }
     }
--- a/mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
+++ b/mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
@@ -263,6 +263,25 @@ mod tests {
         assert!(err.to_string().contains("Trusted_Connection"));
     }
 
+    /// Durable, Python-free guard for the class of drift in #223: pins the exact
+    /// clash message across representative modes (incl. the AD modes whose Python
+    /// e2e coverage was removed because mssql-python strips Trusted_Connection).
+    #[test]
+    fn auth_plus_tc_rejected() {
+        for auth in [
+            "SqlPassword",
+            "ActiveDirectoryInteractive",
+            "ActiveDirectoryDefault",
+        ] {
+            let err = validate_auth(Some(auth), Some(true), "", "", None).unwrap_err();
+            assert!(
+                err.to_string()
+                    .contains("Cannot use Authentication with Trusted_Connection"),
+                "unexpected error for {auth}: {err}"
+            );
+        }
+    }
+
     // ---------------------------------------------------------------
     // SqlPassword requires UID + PWD
     // ---------------------------------------------------------------
--- a/mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
+++ b/mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
@@ -244,8 +244,8 @@ mod tests {
     // ---------------------------------------------------------------
 
     /// Durable, Python-free guard for the class of drift in #223: pins the exact
-    /// clash message across every Authentication mode (incl. the AD modes whose
-    /// Python e2e coverage was removed because mssql-python strips
+    /// clash message across all nine recognized Authentication modes (incl. the
+    /// AD modes whose Python e2e coverage was removed because mssql-python strips
     /// Trusted_Connection). Empty UID/PWD with SqlPassword also proves the TC
     /// clash is reported before the "Both User and Password" rule, covering
     /// ordering in `validate_auth`.
@@ -259,6 +259,8 @@ mod tests {
             "ActiveDirectoryDefault",
             "ActiveDirectoryMSI",
             "ActiveDirectoryServicePrincipal",
+            "ActiveDirectoryDeviceCodeFlow",
+            "ActiveDirectoryWorkloadIdentity",
         ] {
             let err = validate_auth(Some(auth), Some(true), "", "", None).unwrap_err();
             assert_eq!(
--- a/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
+++ b/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
@@ -385,7 +385,12 @@ class TestADServicePrincipal:
 class TestAuthTcClash:
     """§3.3 Auth keyword + TC=Yes is always an error (#24–#26, #28–#29).
 
-    #27 (ADInteractive) is covered by test_auth_resolution.py, not here.
+    #27 (ADInteractive) and ADDefault are not testable at this layer:
+    mssql-python strips Trusted_Connection for token-bearing AD modes, so
+    ODBC never sees the clash. Covered in test_auth_resolution.py
+    (TestAuthTcClashes::test_27_ad_interactive_tc,
+    TestAuthTcClashesExtended::test_ad_default_tc) and in Rust
+    (odbc_authentication_validator::tests::auth_plus_tc_rejected).
     """
 
     def test_row24_tc_yes_sqlpassword(self):
--- a/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
+++ b/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
@@ -379,11 +379,14 @@ class TestADServicePrincipal:
 
 
 # ═════════════════════════════════════════════════════════════════
-#  SECTION 10 — §3.3 Auth + TC clashes (#24–#29)
+#  SECTION 10 — §3.3 Auth + TC clashes (#24–#26, #28–#29)
 # ═════════════════════════════════════════════════════════════════
 
 class TestAuthTcClash:
-    """§3.3 #24–#29 — Auth keyword + TC=Yes is always an error."""
+    """§3.3 Auth keyword + TC=Yes is always an error (#24–#26, #28–#29).
+
+    #27 (ADInteractive) is covered by test_auth_resolution.py, not here.
+    """
 
     def test_row24_tc_yes_sqlpassword(self):
         """#24  TC=Yes + SqlPassword → ERROR."""
--- a/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
+++ b/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
@@ -400,23 +400,6 @@ class TestAuthTcClash:
         cs = _base(tc="Yes", auth="ActiveDirectoryIntegrated")
         _expect_connect_error(cs)
 
-    # #27 (ADInteractive) and ADDefault are intentionally NOT exercised here.
-    #
-    # The mssql-rs behavior (reject TC=Yes + an Authentication keyword) is
-    # asserted deterministically against py-core in the normal, mssql-python-
-    # decoupled suite:
-    #   tests/test_auth_resolution.py::TestAuthTcClashes::test_27_ad_interactive_tc
-    #   tests/test_auth_resolution.py::TestAuthTcClashesExtended::test_ad_default_tc
-    #
-    # An ODBC/mssql-python "parity" leg can't add signal for these two rows:
-    # mssql-python strips Trusted_Connection for token-bearing AD modes
-    # (Default and non-Windows Interactive go through the token factory, whose
-    # keys — including Trusted_Connection — are removed by remove_sensitive_params
-    # before ODBC sees them), so the driver never observes a clash. A test here
-    # could only "pass" by faulting elsewhere (stubbed/absent Entra infra),
-    # which is unfalsifiable. ODBC's own behavior for these modes is covered by
-    # mssql-python's suite.
-
     def test_row28_tc_yes_admsi(self):
         """#28  TC=Yes + ADMSI → ERROR."""
         cs = _base(tc="Yes", auth="ActiveDirectoryMSI")
--- a/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
+++ b/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
@@ -20,6 +20,7 @@ Tests that need real Azure AD infra or Kerberos are skipped unless env vars are 
 
 import os
 import re
+from contextlib import contextmanager
 from pathlib import Path
 from unittest.mock import patch, MagicMock
 
@@ -132,6 +133,78 @@ def _expect_connect_error(conn_str, pattern=None):
             f"Expected pattern '{pattern}' in error: {exc_info.value}"
         )
     return exc_info
+
+
+@contextmanager
+def _no_real_entra_auth():
+    """Stub the azure-identity credential boundary so AD-mode connects never
+    trigger a real interactive/default/managed-identity flow in CI.
+
+    mssql_python.auth imports these credential classes lazily from
+    ``azure.identity`` inside its token method, so patching the public
+    ``azure.identity`` symbols is picked up at call time. We deliberately patch
+    this stable public boundary instead of any mssql-python-internal token
+    helper (the old ``auth.get_auth_token`` was renamed upstream), so this test
+    stays decoupled from mssql-python refactors.
+    """
+    try:
+        import azure.identity  # noqa: F401
+    except ImportError:
+        # azure-identity absent: the AD path fails on its own, still an error.
+        yield
+        return
+
+    patchers = []
+    for name in (
+        "InteractiveBrowserCredential",
+        "DefaultAzureCredential",
+        "DeviceCodeCredential",
+        "ManagedIdentityCredential",
+    ):
+        if hasattr(azure.identity, name):
+            stub = MagicMock()
+            stub.return_value.get_token.side_effect = RuntimeError(
+                "stubbed: no real Entra auth in CI"
+            )
+            patchers.append(patch(f"azure.identity.{name}", stub))
+    for p in patchers:
+        p.start()
+    try:
+        yield
+    finally:
+        for p in patchers:
+            p.stop()
+
+
+def _assert_tc_auth_clash(auth_mode):
+    """TC=Yes + an Authentication keyword is a clash that must be rejected.
+
+    Two independent legs:
+      1. py-core (the mssql-rs behavior): ``validate_auth`` rejects the clash at
+         ``PyCoreConnection`` construction, before any token acquisition. This
+         is deterministic and needs neither mssql-python nor Azure infra.
+      2. mssql-python/ODBC parity: with real Entra auth stubbed out, ``connect``
+         must still reject the combo rather than succeed or block.
+    """
+    # Leg 1 — py-core, fully decoupled from mssql-python.
+    import mssql_py_core
+
+    ctx = {
+        "server": _server(),
+        "database": _database(),
+        "trust_server_certificate": True,
+        "encryption": "Optional",
+        "trusted_connection": "Yes",
+        "authentication": auth_mode,
+    }
+    with pytest.raises(Exception) as exc_info:
+        mssql_py_core.PyCoreConnection(ctx)
+    assert "Cannot use Authentication with Trusted_Connection" in str(exc_info.value)
+
+    # Leg 2 — mssql-python/ODBC parity, decoupled from internal token symbols.
+    cs = _base(tc="Yes", auth=auth_mode)
+    with _no_real_entra_auth():
+        _expect_connect_error(cs)
 
 
 # ═════════════════════════════════════════════════════════════════
@@ -402,12 +475,12 @@ class TestAuthTcClash:
 
     def test_row27_tc_yes_ad_interactive(self):
         """#27  TC=Yes + ADInteractive → ERROR.
-        Must mock token acquisition — mssql_python's auth.py intercepts
-        Interactive before ODBC sees the conn string.
+
+        py-core rejects the TC+Authentication clash at construction; the
+        mssql-python parity leg runs with Entra auth stubbed so no real
+        interactive browser login is attempted in CI.
         """
-        cs = _base(tc="Yes", auth="ActiveDirectoryInteractive")
-        with patch("mssql_python.auth.get_auth_token", return_value=None):
-            _expect_connect_error(cs)
+        _assert_tc_auth_clash("ActiveDirectoryInteractive")
 
     def test_row28_tc_yes_admsi(self):
         """#28  TC=Yes + ADMSI → ERROR."""
@@ -420,12 +493,14 @@ class TestAuthTcClash:
         _expect_connect_error(cs)
 
     def test_tc_yes_ad_default(self):
-        """TC=Yes + ADDefault → ERROR (not in §3.3 — ADDefault is mssql-python-specific).
-        Must mock — auth.py intercepts Default before ODBC sees the clash.
+        """TC=Yes + ADDefault → ERROR (not in §3.3 — ADDefault is
+        mssql-python-specific).
+
+        py-core rejects the TC+Authentication clash at construction; the
+        mssql-python parity leg runs with Entra auth stubbed so the default
+        credential chain is not exercised in CI.
         """
-        cs = _base(tc="Yes", auth="ActiveDirectoryDefault")
-        with patch("mssql_python.auth.get_auth_token", return_value=None):
-            _expect_connect_error(cs)
+        _assert_tc_auth_clash("ActiveDirectoryDefault")
 
 
 # ═════════════════════════════════════════════════════════════════
--- a/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
+++ b/mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
@@ -20,9 +20,8 @@ Tests that need real Azure AD infra or Kerberos are skipped unless env vars are 
 
 import os
 import re
-from contextlib import contextmanager
 from pathlib import Path
-from unittest.mock import patch, MagicMock
+from unittest.mock import patch
 
 import pytest
 from dotenv import load_dotenv
@@ -133,78 +132,6 @@ def _expect_connect_error(conn_str, pattern=None):
             f"Expected pattern '{pattern}' in error: {exc_info.value}"
         )
     return exc_info
-
-
-@contextmanager
-def _no_real_entra_auth():
-    """Stub the azure-identity credential boundary so AD-mode connects never
-    trigger a real interactive/default/managed-identity flow in CI.
-
-    mssql_python.auth imports these credential classes lazily from
-    ``azure.identity`` inside its token method, so patching the public
-    ``azure.identity`` symbols is picked up at call time. We deliberately patch
-    this stable public boundary instead of any mssql-python-internal token
-    helper (the old ``auth.get_auth_token`` was renamed upstream), so this test
-    stays decoupled from mssql-python refactors.
-    """
-    try:
-        import azure.identity  # noqa: F401
-    except ImportError:
-        # azure-identity absent: the AD path fails on its own, still an error.
-        yield
-        return
-
-    patchers = []
-    for name in (
-        "InteractiveBrowserCredential",
-        "DefaultAzureCredential",
-        "DeviceCodeCredential",
-        "ManagedIdentityCredential",
-    ):
-        if hasattr(azure.identity, name):
-            stub = MagicMock()
-            stub.return_value.get_token.side_effect = RuntimeError(
-                "stubbed: no real Entra auth in CI"
-            )
-            patchers.append(patch(f"azure.identity.{name}", stub))
-    for p in patchers:
-        p.start()
-    try:
-        yield
-    finally:
-        for p in patchers:
-            p.stop()
-
-
-def _assert_tc_auth_clash(auth_mode):
-    """TC=Yes + an Authentication keyword is a clash that must be rejected.
-
-    Two independent legs:
-      1. py-core (the mssql-rs behavior): ``validate_auth`` rejects the clash at
-         ``PyCoreConnection`` construction, before any token acquisition. This
-         is deterministic and needs neither mssql-python nor Azure infra.
-      2. mssql-python/ODBC parity: with real Entra auth stubbed out, ``connect``
-         must still reject the combo rather than succeed or block.
-    """
-    # Leg 1 — py-core, fully decoupled from mssql-python.
-    import mssql_py_core
-
-    ctx = {
-        "server": _server(),
-        "database": _database(),
-        "trust_server_certificate": True,
-        "encryption": "Optional",
-        "trusted_connection": "Yes",
-        "authentication": auth_mode,
-    }
-    with pytest.raises(Exception) as exc_info:
-        mssql_py_core.PyCoreConnection(ctx)
-    assert "Cannot use Authentication with Trusted_Connection" in str(exc_info.value)
-
-    # Leg 2 — mssql-python/ODBC parity, decoupled from internal token symbols.
-    cs = _base(tc="Yes", auth=auth_mode)
-    with _no_real_entra_auth():
-        _expect_connect_error(cs)
 
 
 # ═════════════════════════════════════════════════════════════════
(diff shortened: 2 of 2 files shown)


mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -134,6 +135,78 @@ def _expect_connect_error(conn_str, pattern=None):
...
+        "InteractiveBrowserCredential",
+        "DefaultAzureCredential",
+        "DeviceCodeCredential",
+        "ManagedIdentityCredential",
+    ):
+        if hasattr(azure.identity, name):
+            stub = MagicMock()
+            stub.return_value.get_token.side_effect = RuntimeError(
+                "stubbed: no real Entra auth in CI"
+            )
```
saurabh500: **Blocking — this stub is what makes the test unfalsifiable.**

The old mock was `return_value=None` (non-raising), so control flowed onward and the exception *had* to come from the ODBC/driver TC-clash check. This stub raises instead, and upstream `get_auth_token_info` explicitly fails closed — its docstring says an acquisition error "is allowed to propagate".

So on the Linux CI agent the flow is: `_acquire_token` -> stubbed `get_token()` -> `RuntimeError` -> `connect()` raises -> `_expect_connect_error` passes. Row #27 and ADDefault would still be green if TC-clash detection were deleted entirely.

Return a benign token instead, so the flow actually reaches the clash check:

```suggestion
            stub.return_value.get_token.return_value = AccessToken(
                "fake.jwt.token", int(time.time()) + 3600
            )
```

Needs `import time` and `from azure.core.credentials import AccessToken`. Upstream reads `.token` and `.expires_on` off the result, so `AccessToken` is the right shape.

Worth confirming before this lands: that mssql-python forwards `Trusted_Connection` to ODBC rather than stripping it. If it strips it, Leg 2 can never observe the clash — in that case drop Leg 2 for these two rows and note in the docstring that ODBC parity is covered by mssql-python's own suite. A deleted test is more honest than one that cannot fail.

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -134,6 +135,78 @@ def _expect_connect_error(conn_str, pattern=None):
...
+        "authentication": auth_mode,
+    }
+    with pytest.raises(Exception) as exc_info:
+        mssql_py_core.PyCoreConnection(ctx)
+    assert "Cannot use Authentication with Trusted_Connection" in str(exc_info.value)
+
+    # Leg 2 — mssql-python/ODBC parity, decoupled from internal token symbols.
+    cs = _base(tc="Yes", auth=auth_mode)
+    with _no_real_entra_auth():
+        _expect_connect_error(cs)
```
saurabh500: **Blocking (same root cause as the stub above).** `_expect_connect_error(cs)` with no `pattern` accepts *any* exception, so paired with the raising stub this leg asserts only "something went wrong". Pin it to the actual clash once the stub returns a token:

```suggestion
    # Leg 2 — mssql-python/ODBC parity, decoupled from internal token symbols.
    cs = _base(tc="Yes", auth=auth_mode)
    with _no_real_entra_auth():
        _expect_connect_error(cs, pattern=r"Trusted_Connection|Authentication")
```

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -134,6 +135,78 @@ def _expect_connect_error(conn_str, pattern=None):
...
+    this stable public boundary instead of any mssql-python-internal token
+    helper (the old ``auth.get_auth_token`` was renamed upstream), so this test
+    stays decoupled from mssql-python refactors.
+    """
+    try:
+        import azure.identity  # noqa: F401
+    except ImportError:
+        # azure-identity absent: the AD path fails on its own, still an error.
+        yield
+        return
```
saurabh500: **Should fix — this silently converts both tests into no-ops.**

If `azure-identity` is missing, upstream raises `RuntimeError("Azure authentication libraries are not installed...")` and `_expect_connect_error` accepts it happily.

Not hypothetical: `.pipeline/templates/test-mssql-python-template.yml` installs only `maturin pytest pytest-asyncio python-dotenv patchelf pybind11 wheel setuptools` — `azure-identity` arrives only transitively via `pip install -e "$MSSQL_PYTHON_DIR"` in `dev/test-python.sh`. An upstream dependency change turns these green for the wrong reason.

Skip explicitly so the reason shows up in the test report:

```suggestion
    except ImportError:
        pytest.skip("azure-identity not installed; cannot stub the Entra credential boundary")
```

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -134,6 +135,78 @@ def _expect_connect_error(conn_str, pattern=None):
...
+                "stubbed: no real Entra auth in CI"
+            )
+            patchers.append(patch(f"azure.identity.{name}", stub))
+    for p in patchers:
+        p.start()
+    try:
+        yield
+    finally:
+        for p in patchers:
+            p.stop()
```
saurabh500: **Should fix — started patchers leak if `start()` raises mid-loop.**

`p.start()` runs outside the `try`, so if the third `start()` fails the first two stay patched globally for the rest of the session. `azure.identity.DefaultAzureCredential` would remain a `MagicMock` for every later test, including `conftest.py::_build_context_managed_identity`.

`ExitStack` fixes it and removes the manual bookkeeping:

```python
from contextlib import ExitStack, contextmanager
...
with ExitStack() as stack:
    for name in (...):
        if hasattr(azure.identity, name):
            stub = MagicMock()
            stub.return_value.get_token.return_value = AccessToken(
                "fake.jwt.token", int(time.time()) + 3600
            )
            stack.enter_context(patch(f"azure.identity.{name}", stub))
    yield
```

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -134,6 +135,78 @@ def _expect_connect_error(conn_str, pattern=None):
...
+
+    Two independent legs:
+      1. py-core (the mssql-rs behavior): ``validate_auth`` rejects the clash at
+         ``PyCoreConnection`` construction, before any token acquisition. This
+         is deterministic and needs neither mssql-python nor Azure infra.
+      2. mssql-python/ODBC parity: with real Entra auth stubbed out, ``connect``
+         must still reject the combo rather than succeed or block.
+    """
+    # Leg 1 — py-core, fully decoupled from mssql-python.
+    import mssql_py_core
```
saurabh500: **Should fix — Leg 1 is mostly duplicate coverage, and it does not run where it is needed.**

- `test_auth_resolution.py:400` (`TestAuthTcClashes::test_27_ad_interactive_tc`) already asserts exactly this for `ActiveDirectoryInteractive`, so Leg 1 adds no new signal for row #27. Only the `ActiveDirectoryDefault` combination is genuinely new — I confirmed `test_auth_resolution.py` has no ADDefault clash test.
- "fully decoupled from mssql-python" is true at the *symbol* level but not the *execution* level: `tests/mssql_python/conftest.py:9` does `from mssql_python import connect` at import time, and `dev/test-python.sh` runs the normal suite with `--ignore=tests/mssql_python`. So Leg 1 only ever executes in the cross-repo job that requires mssql-python installed — the exact environment whose drift caused #223.

Suggest moving the ADDefault case into `mssql-py-core/tests/test_auth_resolution.py::TestAuthTcClashes` next to rows #24-#29 using the existing helper, and dropping Leg 1 from here:

```python
def test_ad_default_tc(self):
    ctx = _bare_ctx()
    ctx["authentication"] = "ActiveDirectoryDefault"
    ctx["trusted_connection"] = "Yes"
    _expect_validation_error(ctx, "Trusted_Connection")
```

saurabh500 (commented):
## Self-review: changes needed before marking ready

Root cause and direction are right — `azure.identity` is the correct seam to patch (upstream literally comments `# Import Azure libraries inside method to support test mocking` in `_acquire_token`), and asserting mssql-rs behavior against `PyCoreConnection` instead of inferring it through a third-party driver is the right decoupling.

But as written, **Leg 2 of both tests passes vacuously**: the stub raises, upstream `get_auth_token_info` documents that it "fails closed" and lets acquisition errors propagate, and `_expect_connect_error(cs)` is called with no `pattern`. Rows #27 and ADDefault would stay green if TC-clash detection were deleted entirely. That trades a loud `AttributeError` for a silent pass — worse than the status quo. Details inline.

### Verified against upstream `mssql_python/auth.py`

| Assumption | Holds? |
| --- | --- |
| Credential classes imported lazily inside the token method, so patching `azure.identity.X` takes effect at call time | Yes — explicit "to support test mocking" comment |
| `get_auth_token` was renamed; `get_auth_token_info` is the current symbol | Yes |
| `AADAuth.get_raw_token` still exists, so deferring the other three tests is safe | Yes |
| Token-acquisition errors propagate rather than being swallowed | Yes — and this is what breaks Leg 2 |

Also confirmed in-repo that `validate_auth` runs inside `dict_to_client_context` before any connection attempt (`mssql-py-core/src/connection.rs:375`, called from `PyCoreConnection::new` at :54), and that the clash check precedes the `is_recognized_keyword` check (`odbc_authentication_validator.rs:44-48`) — so Leg 1's assertion does hold for `ActiveDirectoryDefault`.

### Not inline-able: add a Rust unit test

`odbc_authentication_validator.rs` has `happy_empty_auth_plus_tc` and `access_token_plus_tc` but **nothing asserting the `"Cannot use Authentication with Trusted_Connection."` message**. That is the one place this can run on every PR with zero Python, zero Docker and zero cross-repo coupling — the durable fix for the class of problem #223 describes:

```rust
#[test]
fn auth_plus_tc_rejected() {
    for auth in ["SqlPassword", "ActiveDirectoryInteractive", "ActiveDirectoryDefault"] {
        let err = validate_auth(Some(auth), Some(true), "", "", None).unwrap_err();
        assert!(err.to_string().contains("Cannot use Authentication with Trusted_Connection"));
    }
}
```

If this is added, the `cargo bfmt` / `bclippy` / `btest` checklist boxes become relevant again.

### Undocumented platform split

Upstream's `process_auth_parameters` returns `None` for Interactive on Windows (native ODBC handles it), so on Windows the stub is never consulted and Leg 2 genuinely exercises the ODBC clash; on Linux CI it does not. One line in the docstring would stop a local Windows pass from misleading the next reader.

### Suggested shape

The highest-leverage version of this PR is smaller than what is here: add the Rust unit test and move the ADDefault case into `test_auth_resolution.py` (both run on every PR, no cross-repo coupling), then either fix Leg 2's stub to be non-raising with a message assertion, or drop Leg 2 for these two rows. That resolves #223 and leaves coverage strictly stronger than before.

**Done well:** the `hasattr` guard before patching is properly defensive against azure-identity version skew and avoids `create=True` masking typos; scope is tight (one file, no drive-by refactors); the PR description documents the deferred `get_raw_token` follow-up.

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -134,6 +135,78 @@ def _expect_connect_error(conn_str, pattern=None):
...
+        "server": _server(),
+        "database": _database(),
+        "trust_server_certificate": True,
+        "encryption": "Optional",
+        "trusted_connection": "Yes",
+        "authentication": auth_mode,
+    }
+    with pytest.raises(Exception) as exc_info:
+        mssql_py_core.PyCoreConnection(ctx)
+    assert "Cannot use Authentication with Trusted_Connection" in str(exc_info.value)
```
saurabh500: **Should fix — diverges from the sibling file's established helper.**

`test_auth_resolution.py:116` defines `_expect_validation_error(ctx, match_text)` = `pytest.raises(RuntimeError, match=...)`. Bare `Exception` is broad enough to pass if `PyCoreConnection` blew up for an unrelated reason.

```suggestion
    with pytest.raises(RuntimeError, match="Cannot use Authentication with Trusted_Connection"):
        mssql_py_core.PyCoreConnection(ctx)
```

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -134,6 +135,78 @@ def _expect_connect_error(conn_str, pattern=None):
...
+        yield
+        return
+
+    patchers = []
+    for name in (
+        "InteractiveBrowserCredential",
+        "DefaultAzureCredential",
+        "DeviceCodeCredential",
+        "ManagedIdentityCredential",
+    ):
```
saurabh500: Nit: `ClientSecretCredential` is missing from this list — upstream imports it separately in `_acquire_token_service_principal`. The helper name promises "no real Entra auth", so ADSPA (row #29) would still reach the network if this is ever reused for it. Either add it to the tuple or narrow the docstring to the four interactive/default/device-code/MSI paths it actually covers.

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -134,6 +135,78 @@ def _expect_connect_error(conn_str, pattern=None):
...
+    import mssql_py_core
+
+    ctx = {
+        "server": _server(),
+        "database": _database(),
+        "trust_server_certificate": True,
+        "encryption": "Optional",
+        "trusted_connection": "Yes",
+        "authentication": auth_mode,
+    }
```
saurabh500: Nit: the sibling `_base_ctx()` in `test_auth_resolution.py` sets `"connect_timeout": 1,  # Short timeout for expected failures`. Validation fires before any connect today so it does not matter, but including it bounds the blast radius if that ordering ever changes, and matches the sibling pattern.

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -400,14 +400,22 @@ def test_row26_tc_yes_ad_integrated(self):
...
 
-    def test_row27_tc_yes_ad_interactive(self):
-        """#27  TC=Yes + ADInteractive → ERROR.
-        Must mock token acquisition — mssql_python's auth.py intercepts
-        Interactive before ODBC sees the conn string.
-        """
-        cs = _base(tc="Yes", auth="ActiveDirectoryInteractive")
-        with patch("mssql_python.auth.get_auth_token", return_value=None):
-            _expect_connect_error(cs)
+    # #27 (ADInteractive) and ADDefault are intentionally NOT exercised here.
```
saurabh500: This is too much verbosity and probably not needed. The history of files will surface this. 

Remove

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Good catch, and you were right to push on the `Trusted_Connection` question. I confirmed it against upstream: for token-bearing AD modes (`ActiveDirectoryDefault` and non-Windows `ActiveDirectoryInteractive`), mssql-python routes through its token factory and calls `remove_sensitive_params`, whose `_SENSITIVE_KEYS` includes `_KEY_TRUSTED_CONNECTION` — so `Trusted_Connection` is stripped before ODBC ever sees it. The driver therefore never observes a clash for these two rows, and no benign-token stub could make the parity leg falsifiable.

Rather than re-target another internal, I removed the whole `_no_real_entra_auth` / `_assert_tc_auth_clash` machinery and both tests. The mssql-rs behavior (reject TC + Authentication) is already asserted deterministically against py-core in `tests/test_auth_resolution.py` (`test_27_ad_interactive_tc`, `test_ad_default_tc`), which runs in the normal suite decoupled from mssql-python.


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Resolved by removal: `_assert_tc_auth_clash` and its unpinned `_expect_connect_error(cs)` call are gone, so there's no catch-any parity assertion left for these rows. (The `_expect_connect_error` calls for rows #24/#25/#26/#28/#29 are pre-existing and unchanged — out of scope for this PR.)


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Moot now — the `_no_real_entra_auth` helper (including the silent `ImportError` no-op branch you flagged) is removed entirely. There's no azure-identity stubbing left in this file.


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Moot now — the manual `start()`/`stop()` patcher loop is deleted along with the helper, so there's no leak path where a mid-loop failure could leave `azure.identity.*` a MagicMock for later tests.


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Agreed — and it turns out ADDefault is *already* covered too, in a class I'd missed: `tests/test_auth_resolution.py::TestAuthTcClashesExtended::test_ad_default_tc` (line 708), right alongside `TestAuthTcClashes::test_27_ad_interactive_tc` (line 400). Both use `_expect_validation_error(ctx, "Trusted_Connection")` and run in the normal py-core suite, which is execution-decoupled from mssql-python (the `tests/mssql_python` dir is `--ignore`d there and only runs in the cross-repo job).

So Leg 1 here was pure duplication. I dropped it (and the whole pair of e2e tests) and left a comment in `TestAuthTcClash` pointing at those two canonical assertions.


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Resolved by removal — the bare `pytest.raises(Exception)` block is gone. The surviving assertion for this behavior lives in `tests/test_auth_resolution.py` via `_expect_validation_error`, which is exactly `pytest.raises(RuntimeError, match="Trusted_Connection")` — the pinned form you're asking for.


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Moot now — the credential stub tuple is deleted with the helper, so there's no `ClientSecretCredential`-shaped gap that ADSPA (#29) could fall through. #29 keeps its existing non-mocked `_expect_connect_error(cs)`.


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Moot now — the inline ctx dict is deleted. The surviving py-core assertions in `tests/test_auth_resolution.py` use `_bare_ctx()`, which already sets `connect_timeout: 1`, so they match the sibling pattern you're pointing at.


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Removed. The rationale lives in the commit message / file history instead.


> [!NOTE]

saurabh500:
> ### Not inline-able: add a Rust unit test
> `odbc_authentication_validator.rs` has `happy_empty_auth_plus_tc` and `access_token_plus_tc` but **nothing asserting the `"Cannot use Authentication with Trusted_Connection."` message**.

Added in 2aa3f30c as `auth_plus_tc_rejected` — it loops over `SqlPassword`, `ActiveDirectoryInteractive`, `ActiveDirectoryDefault` and pins the exact clash message (the existing `tc_plus_*` tests only matched the `Trusted_Connection` substring, which the access-token conflict text also contains). This is the durable, Python-free guard for the class of drift in #223; runs on every PR with zero Docker and zero cross-repo coupling. Passes locally under `cargo nextest` and clean under `cargo clippy --all-targets -D warnings`.

The related inline threads (drop the vacuous Leg 2, move/dedupe the py-core coverage) are addressed by deleting both e2e tests — the ADInteractive and ADDefault clash cases are already covered in `test_auth_resolution.py` (`test_27_ad_interactive_tc`, `test_ad_default_tc`), and the PR description has been updated to match.

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -400,15 +400,6 @@ def test_row26_tc_yes_ad_integrated(self):
...
-    def test_row27_tc_yes_ad_interactive(self):
-        """#27  TC=Yes + ADInteractive → ERROR.
-        Must mock token acquisition — mssql_python's auth.py intercepts
-        Interactive before ODBC sees the conn string.
-        """
-        cs = _base(tc="Yes", auth="ActiveDirectoryInteractive")
-        with patch("mssql_python.auth.get_auth_token", return_value=None):
-            _expect_connect_error(cs)
-
     def test_row28_tc_yes_admsi(self):
```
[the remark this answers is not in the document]
saurabh500: Good catch — fixed in 13db3185. The section header and class docstring now advertise `#24–#26, #28–#29`, and the docstring notes that #27 (ADInteractive) lives in `test_auth_resolution.py` rather than here.


> [!NOTE]

mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
```diff
@@ -263,6 +263,25 @@ mod tests {
...
+        ] {
+            let err = validate_auth(Some(auth), Some(true), "", "", None).unwrap_err();
+            assert!(
+                err.to_string()
+                    .contains("Cannot use Authentication with Trusted_Connection"),
+                "unexpected error for {auth}: {err}"
+            );
+        }
+    }
+
```
David-Engel, line 273: **Suggestion:** this now overlaps `tc_plus_sqlpassword` (line 247), `tc_plus_ad_password`, and `tc_plus_ad_integrated`, which sit ~20 lines above and cover the same code path with a *weaker* assertion. As the PR description itself points out, `contains("Trusted_Connection")` also matches the access-token conflict text, so those three are the ones actually carrying the drift risk — and they're being left in place.

Prefer collapsing to one table-driven test rather than running two overlapping styles side by side:

```rust
#[test]
fn auth_plus_tc_rejected() {
    for auth in [
        "SqlPassword",
        "ActiveDirectoryPassword",
        "ActiveDirectoryIntegrated",
        "ActiveDirectoryInteractive",
        "ActiveDirectoryDefault",
        "ActiveDirectoryMSI",
        "ActiveDirectoryServicePrincipal",
    ] {
        let err = validate_auth(Some(auth), Some(true), "", "", None).unwrap_err();
        assert_eq!(
            err.to_string(),
            "Cannot use Authentication with Trusted_Connection.",
            "unexpected error for {auth}"
        );
    }
}
```

…and delete the three `tc_plus_*` tests it subsumes. `assert_eq!` over `contains` also pins the trailing period, which `contains` silently tolerates.

One thing the current version does get right and is worth keeping either way: passing empty UID/PWD with `SqlPassword` proves the TC clash is reported *before* the "Both User and Password" rule, so ordering in `validate_auth` is covered too.

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
```diff
@@ -379,11 +379,14 @@ def test_row38_adspa_pwd_no_uid(self):
...
 # ═════════════════════════════════════════════════════════════════
-#  SECTION 10 — §3.3 Auth + TC clashes (#24–#29)
+#  SECTION 10 — §3.3 Auth + TC clashes (#24–#26, #28–#29)
 # ═════════════════════════════════════════════════════════════════
 
 class TestAuthTcClash:
-    """§3.3 #24–#29 — Auth keyword + TC=Yes is always an error."""
+    """§3.3 Auth keyword + TC=Yes is always an error (#24–#26, #28–#29).
+
+    #27 (ADInteractive) is covered by test_auth_resolution.py, not here.
```
David-Engel: **Nit:** `test_tc_yes_ad_default` was removed here too, but the pointer only mentions #27. A reader looking for the ADDefault e2e case gets no breadcrumb (it isn't a numbered row, so the `#24–#26, #28–#29` header doesn't hint at it either). Also worth recording *why* they can't live here, since that's the non-obvious part:

```python
    """§3.3 Auth keyword + TC=Yes is always an error (#24–#26, #28–#29).

    #27 (ADInteractive) and ADDefault are not testable at this layer:
    mssql-python strips Trusted_Connection for token-bearing AD modes, so
    ODBC never sees the clash. Covered in test_auth_resolution.py
    (TestAuthTcClashes::test_27_ad_interactive_tc,
    TestAuthTcClashesExtended::test_ad_default_tc) and in Rust
    (odbc_authentication_validator::tests::auth_plus_tc_rejected).
    """
```

Without that, someone will eventually "restore missing coverage" and re-add the same unfalsifiable tests.

David-Engel (commented):
## Summary

Test-only change: deletes the two cross-repo Python e2e tests that patched the now-renamed `mssql_python.auth.get_auth_token`, and adds a Python-free Rust unit test (`auth_plus_tc_rejected`) that pins the exact `Authentication` + `Trusted_Connection` clash message. Direction is right — moving the assertion to where the behavior actually lives removes the upstream-drift coupling without losing coverage.

**Verified locally** (worktree at PR head `13db3185`):
- `cargo test --lib auth_plus_tc_rejected` in `mssql-py-core` → 1 passed; full `odbc_auth` module → 71 passed, 0 failed.
- The claimed replacement coverage exists and is real: `tests/test_auth_resolution.py::TestAuthTcClashes::test_27_ad_interactive_tc` (line 400) and `TestAuthTcClashesExtended::test_ad_default_tc` (line 708), both asserting via `_expect_validation_error(ctx, "Trusted_Connection")` with no `mssql_python` import and no Azure infra.
- `MagicMock` is genuinely unused after the deletion; `patch` is still used (lines 839/873/887/925/935/946), so the trimmed import is correct. `python3 -m py_compile` clean.
- No dangling references to `row27` / `tc_yes_ad_default` / `get_auth_token` anywhere in the repo.

**No blocking issues.** A few suggestions below.

### CI note
`coverage-report` is red at 1h17m with the known artifact-poll timeout, not a real failure — the GH action's 75-minute budget expires before the ADO `Merge Coverage` stage publishes the merged `CoberturaCoverage` artifact. Re-run once the ADO build finishes: `gh run rerun 31536772392 --repo microsoft/mssql-rs --failed`. Notably the job this PR targets — **Build mssql-python (cross-repo)** — now passes.

### Housekeeping

- **`Fixes #223` will close the follow-up along with the fix.** Issue #223 line 29 also tracks the three remaining tests that patch `mssql_python.auth.AADAuth.get_raw_token` (~962/1014/1036), which this PR intentionally leaves alone. Merging auto-closes the issue and that follow-up disappears with it. Either file a separate issue for the `get_raw_token` tests and reference it, or downgrade to `Related to #223` and close #223 by hand once the remainder is done.
- **Nit — PR description describes intermediate commits, not the net diff.** It says the two tests are deleted "along with the `_no_real_entra_auth` / `_assert_tc_auth_clash` helpers and the now-unused `contextmanager` / `MagicMock` imports." Against `main` only `MagicMock` is removed; the helpers and the `contextmanager` import never existed in the base (they were added and removed within this branch's own history). Worth trimming before merge so the squashed commit message matches what actually changed.

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Hahahahaaa.... I had asked Copilot to remove a long verbose comment describing this part in the name of AI Slop..

mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
saurabh500: Done in 9abe20a8. Collapsed `tc_plus_sqlpassword`/`tc_plus_ad_password`/`tc_plus_ad_integrated` into the table-driven `auth_plus_tc_rejected`, now covering all seven Authentication modes with `assert_eq!` on the full message — so the weaker `contains("Trusted_Connection")` assertions (which also matched the access-token text) are gone. Kept the empty-UID/PWD SqlPassword case for the ordering coverage you flagged and noted it in the doc comment.

One correction to the snippet: `Display` prepends `Usage Error: `, so the pinned string is `"Usage Error: Cannot use Authentication with Trusted_Connection."` (still exact, still pins the trailing period).


> [!NOTE]

mssql-py-core/tests/mssql_python/test_auth_resolution_e2e.py
saurabh500: Done in 9abe20a8. The class docstring now names both #27 (ADInteractive) and ADDefault, records that mssql-python strips Trusted_Connection for token-bearing AD modes (so ODBC never sees the clash), and points to all three homes for the retained coverage: `TestAuthTcClashes::test_27_ad_interactive_tc`, `TestAuthTcClashesExtended::test_ad_default_tc`, and the Rust `auth_plus_tc_rejected`.


> [!NOTE]

mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
```diff
@@ -243,24 +243,30 @@ mod tests {
...
-        assert!(err.to_string().contains("Trusted_Connection"));
-    }
-
-    #[test]
-    fn tc_plus_ad_integrated() {
-        let err =
-            validate_auth(Some("ActiveDirectoryIntegrated"), Some(true), "", "", None).unwrap_err();
-        assert!(err.to_string().contains("Trusted_Connection"));
+    /// Durable, Python-free guard for the class of drift in #223: pins the exact
+    /// clash message across every Authentication mode (incl. the AD modes whose
```
David-Engel: **Nit (non-blocking):** "every Authentication mode" is now slightly ahead of the list — the loop covers 7 of the 9 recognized keywords in `odbc_supported_auth_keywords.rs`. `ActiveDirectoryDeviceCodeFlow` and `ActiveDirectoryWorkloadIdentity` are missing (both have `happy_*_alone` tests, so they're real modes, and `TestAuthTcClashesExtended` covers them on the Python side).

Behaviorally they're identical — the TC clash returns before `is_recognized_keyword` is ever consulted — so this is purely about the comment matching the code. Either add the two strings to the array, or soften to "across representative modes." I'd add them: the array is the canonical list a future reader will diff against when a new keyword is introduced.

David-Engel (approved):
Approving. Both of my comments are addressed in `9abe20a8`, and the consolidation came out better than what I suggested — folding `tc_plus_sqlpassword` / `tc_plus_ad_password` / `tc_plus_ad_integrated` into the table-driven `auth_plus_tc_rejected` removes the weak `contains("Trusted_Connection")` assertions that were the actual drift risk, and `assert_eq!` on the full Display string now pins the `Usage Error: ` prefix and the trailing period too. The e2e docstring recording *why* ADInteractive/ADDefault can't be tested at that layer is the part most likely to save someone from re-adding them later.

**Verified at head `9abe20a8`:**
- `cargo test --lib odbc_auth` → 68 passed, 0 failed (down from 71, matching the three consolidated tests).
- `cargo fmt -- --check` clean; `cargo clippy --all-features --all-targets -- -D warnings` clean.
- `python3 -m py_compile` clean on the e2e file.
- The cross-repo **Build mssql-python** leg — the job this PR exists to fix — passes on this head.

Two non-blocking notes below; neither needs to gate the merge.

### Housekeeping before merge

The PR description still says the two tests are deleted "along with the `_no_real_entra_auth` / `_assert_tc_auth_clash` helpers and the now-unused `contextmanager` / `MagicMock` imports." Against `main` only `MagicMock` is removed — the helpers and the `contextmanager` import existed solely in this branch's intermediate commits. On a squash merge that text becomes the permanent commit message, so it's worth trimming to the net change.

Also, `Fixes #223` will auto-close the issue on merge, and line 29 of that issue is the only place tracking the three remaining tests that patch `mssql_python.auth.AADAuth.get_raw_token` (~962/1014/1036). That follow-up disappears with it. Either file a separate issue for those and link it, or switch to `Related to #223` and close by hand once the rest lands.

### CI

The ADO validation stages are still mid-flight on this head. If `coverage-report` goes red at ~75 minutes, that's the known artifact-poll timeout, not a real failure — the GH action's poll budget expires before the ADO `Merge Coverage` stage publishes the merged `CoberturaCoverage` artifact. Re-run the failed job once the ADO build completes and it passes in seconds.

mssql-py-core/src/odbc_auth/odbc_authentication_validator.rs
saurabh500: Added both in 9bf8cdd3 — the loop now covers all nine recognized keywords, including `ActiveDirectoryDeviceCodeFlow` and `ActiveDirectoryWorkloadIdentity`, and the comment says "all nine recognized Authentication modes." Agreed the array should be the canonical list a future reader diffs against when a new keyword lands.


> [!NOTE]

David-Engel (approved):

PR #226, opened 2026-08-12
Heap-allocate the PLP drain scratch buffer

### LINKED ISSUE #225: Perf: 8 KiB stack scratch in drain_active_plp inflates every row-fetch future to ~8.5 KB (8.6x per-row cost)
### Describe the bug

`drain_active_plp` holds an 8 KiB scratch array on the stack across an `.await`:

https://github.com/microsoft/mssql-rs/blob/182e3a9c1d7c4d69567765f117d3bd8876a656dd/mssql-tds/src/connection/tds_client.rs#L3236-L3249

```rust
/// Reads and discards all remaining bytes of an active PLP stream.
async fn drain_active_plp(&mut self, plp_state: &mut PlpPauseState) -> TdsResult<()> {
    let mut buffer = [0u8; 8192];          // <-- lives across the await below
    while !plp_state.reached_end() {
        ...
        let read = self.transport.read_active_plp_bytes(..., &mut buffer).await?;
        ...
    }
}
```

Because `buffer` is used after the `.await`, rustc must store it *inside* the generated future. Every caller that awaits this function therefore inherits the 8 KiB, and on `main` that is the entire row-fetch hot path:

| Caller | Awaits | Site |
|---|---|---|
| `read_row_column` (L3126) | `drain_active_plp` directly | L3133 |
| `next_row_cursor` (L3043) | via `drain_active_row` (L3217) | L3058 |
| `get_next_row_into` (L2960) | via `drain_active_row` | L2994 |
| `drain_rows` (L1781) | via `drain_active_row` | L1789 |

The drain path itself is **rare** — it only runs when a caller abandons a partially read PLP column. But future size is a static property of the type, not of the branch taken at runtime, so **every row pays to construct and move ~8.5 KB of state for a path that almost never executes.**

### Steps to reproduce

Self-contained reproduction, `std` + criterion only — no tokio, no sockets, no TDS. Three files in an empty directory, then `cargo bench`.

<details>
<summary><code>Cargo.toml</code></summary>

```toml
[package]
name = "plp-bench"
version = "0.0.0"
edition = "2021"

[dev-dependencies]
criterion = "0.5"

[[bench]]
name = "plp_future"
harness = false
```
</details>

<details>
<summary><code>src/lib.rs</code></summary>

```rust
//! Standalone reproduction of the `drain_active_plp` future-size bug.
//!
//! Pure `std` plus criterion. Everything here exists to isolate one question:
//! what does an 8 KiB stack array held across an `.await` cost the callers
//! that await it?

use std::future::Future;
use std::pin::Pin;
use std::task::{Context, Poll, RawWaker, RawWakerVTable, Waker};

/// Same scratch size as the driver's PLP drain path.
pub const SCRATCH: usize = 8192;

// ---------------------------------------------------------------------------
// A ~20 line executor, so the benchmark measures our futures and nothing else.
// ---------------------------------------------------------------------------

const VTABLE: RawWakerVTable = RawWakerVTable::new(
    |_| RawWaker::new(std::ptr::null(), &VTABLE),
    |_| {},
    |_| {},
    |_| {},
);

fn noop_waker() -> Waker {
    // SAFETY: the vtable's clone returns an equivalent no-op waker and the
    // remaining entries do nothing, so every contract is trivially upheld.
    unsafe { Waker::from_raw(RawWaker::new(std::ptr::null(), &VTABLE)) }
}

/// Polls `fut` to completion on the current thread.
///
/// `fut` is taken **by value**, so calling this moves the future into this
/// frame. That is the point: for an 8 KiB future the move is an 8 KiB memcpy.
/// `inline(never)` keeps the move observable instead of letting LLVM collapse
/// the two frames into one, which mirrors the real driver where these futures
/// cross a `Box::pin` boundary and cannot be elided either.
#[inline(never)]
pub fn block_on<F: Future>(fut: F) -> F::Output {
    let mut fut = fut;
    // SAFETY: `fut` is owned by this frame and is never moved again.
    let mut fut = unsafe { Pin::new_unchecked(&mut fut) };
    let waker = noop_waker();
    let mut cx = Context::from_waker(&waker);
    loop {
        if let Poll::Ready(v) = fut.as_mut().poll(&mut cx) {
            return v;
        }
    }
}

/// Returns `Pending` exactly once. Anything live across it must be stored in
/// the enclosing future, which is what makes a local array part of the state
/// machine rather than a transient stack slot.
pub struct YieldOnce(bool);

impl Future for YieldOnce {
    type Output = ();

    fn poll(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<()> {
        if self.0 {
            Poll::Ready(())
        } else {
            self.0 = true;
            Poll::Pending
        }
    }
}

pub fn yield_once() -> YieldOnce {
    YieldOnce(false)
}

// ---------------------------------------------------------------------------
// The cold path. Runs only when a caller abandons a partially read column.
// ---------------------------------------------------------------------------

/// Scratch on the stack. `buf` is live across the `.await`, so the whole
/// 8 KiB lands inside the generated future.
pub async fn drain_stack(chunks: usize) -> u64 {
    let mut buf = [0u8; SCRATCH];
    let mut sum = 0;
    for i in 0..chunks {
        yield_once().await;
        buf[i % SCRATCH] = i as u8;
        sum += u64::from(buf[i % SCRATCH]);
    }
    sum
}

/// Scratch on the heap. The future stores a 24 byte `Vec` header instead.
pub async fn drain_heap(chunks: usize) -> u64 {
    let mut buf = vec![0u8; SCRATCH];
    let mut sum = 0;
    for i in 0..chunks {
        yield_once().await;
        buf[i % SCRATCH] = i as u8;
        sum += u64::from(buf[i % SCRATCH]);
    }
    sum
}

// ---------------------------------------------------------------------------
// The hot path. One call per row, mirroring `next_row_cursor`.
// ---------------------------------------------------------------------------

/// A cheap await standing in for the socket read, then a conditional await of
/// the drain helper. `abandon` is false on every normal row.
pub async fn row_stack(abandon: bool) -> u64 {
    yield_once().await;
    if abandon {
        return drain_stack(4).await;
    }
    1
}

pub async fn row_heap(abandon: bool) -> u64 {
    yield_once().await;
    if abandon {
        return drain_heap(4).await;
    }
    1
}

/// Future sizes, resolved at compile time. Creating an async fn's future does
/// not run any of its body, so these calls are free.
pub fn sizes() -> Vec<(&'static str, usize)> {
    vec![
        ("drain_stack", std::mem::size_of_val(&drain_stack(0))),
        ("drain_heap", std::mem::size_of_val(&drain_heap(0))),
        ("row_stack", std::mem::size_of_val(&row_stack(false))),
        ("row_heap", std::mem::size_of_val(&row_heap(false))),
    ]
}

/// Fetches `rows` rows, never taking the drain branch.
pub fn fetch_stack(rows: usize) -> u64 {
    (0..rows).map(|_| block_on(row_stack(false))).sum()
}

pub fn fetch_heap(rows: usize) -> u64 {
    (0..rows).map(|_| block_on(row_heap(false))).sum()
}

/// Same, but boxing each row future first. This is what the ODBC layer does:
/// `drive_read` receives a pinned, boxed future once per row.
pub fn fetch_stack_boxed(rows: usize) -> u64 {
    (0..rows)
        .map(|_| block_on(Box::pin(row_stack(false))))
        .sum()
}

pub fn fetch_heap_boxed(rows: usize) -> u64 {
    (0..rows).map(|_| block_on(Box::pin(row_heap(false)))).sum()
}
```
</details>

<details>
<summary><code>benches/plp_future.rs</code></summary>

```rust
use criterion::{Criterion, Throughput, black_box, criterion_group, criterion_main};
use plp_bench::*;

/// One criterion sample drives this many rows, so per-row cost dominates the
/// harness overhead. The driver benchmark fetches 5000.
const ROWS: usize = 5000;

fn bench(c: &mut Criterion) {
    println!("\n--- generated future sizes ---");
    for (name, size) in sizes() {
        println!("{name:<12} {size:>6} bytes");
    }
    println!();

    let mut g = c.benchmark_group("fetch_rows");
    g.throughput(Throughput::Elements(ROWS as u64));
    g.bench_function("stack_scratch", |b| {
        b.iter(|| black_box(fetch_stack(black_box(ROWS))))
    });
    g.bench_function("heap_scratch", |b| {
        b.iter(|| black_box(fetch_heap(black_box(ROWS))))
    });
    g.finish();

    let mut g = c.benchmark_group("fetch_rows_boxed");
    g.throughput(Throughput::Elements(ROWS as u64));
    g.bench_function("stack_scratch", |b| {
        b.iter(|| black_box(fetch_stack_boxed(black_box(ROWS))))
    });
    g.bench_function("heap_scratch", |b| {
        b.iter(|| black_box(fetch_heap_boxed(black_box(ROWS))))
    });
    g.finish();
}

criterion_group!(benches, bench);
criterion_main!(benches);
```
</details>

### Expected behavior

The per-row futures should be a few hundred bytes. A scratch buffer used only on a rare, already network-bound cleanup path should not appear in the state machine of every row read.

### Actual behavior

**Generated future sizes** — the array is embedded, and the size propagates to the caller:

| future | stack scratch | heap scratch | ratio |
|---|---|---|---|
| `drain_*` | 8240 B | 72 B | 114× |
| `row_*` (the per-row caller) | **8248 B** | **80 B** | **103×** |

**Cost, 5000 rows per sample, drain branch never taken:**

| group | variant | time | per row |
|---|---|---|---|
| `fetch_rows` | stack scratch | 680.25 µs | 136.1 ns |
| `fetch_rows` | heap scratch | **78.88 µs** | **15.8 ns** |

**8.6× faster, 120.3 ns saved per row — on rows that never enter the drain path at all.**

Sanity check on the mechanism: 8248 bytes ÷ 120.3 ns ≈ 69 GB/s, i.e. roughly one L1-resident 8 KiB `memcpy` per row. The measured delta is precisely what a single move of the oversized future costs, which is what you would predict from the theory below.

### Theory

1. An `async fn` compiles to a state machine. Its size is the max, over all suspension points, of everything that must stay live across an `.await`.
2. `let mut buffer = [0u8; 8192]` is used *after* the await, so it must survive the suspension. It is stored **inline** in the future: `size_of::<DrainFuture>() >= 8192`.
3. Awaiting a future **embeds it in the caller's state machine**. Future sizes compose additively down an await chain, so `next_row_cursor`, `read_row_column`, `get_next_row_into` and `drain_rows` all become ≥8 KiB.
4. **This is a static property of the type.** It does not matter that `if abandon` is false on every real row — the variant still occupies its slot in the enum. A cold branch cannot be made cheap by not taking it.
5. The cost is paid on every **construction** and every **move** of the outer future. Moving a future is a `memcpy` of the whole thing, and the row loop constructs and moves one per row.
6. `vec![0u8; 8192]` stores a 24-byte `Vec` header in the future and leaves the payload behind a pointer. The payload does not move when the future moves, so the state machine collapses back to its natural size.

The asymmetry is what makes this a clean win: the drain path is rare and already blocked on the network, so one heap allocation there is free — while the size penalty it imposes is paid by every row on the hot path.

**One honest nuance,** shown by the second benchmark group:

| group | variant | time | per row |
|---|---|---|---|
| `fetch_rows_boxed` | stack scratch | 235.53 µs | 47.1 ns |
| `fetch_rows_boxed` | heap scratch | 236.99 µs | 47.4 ns |

When the future is `Box::pin`ned, the difference disappears. The compiler constructs the future directly into the heap allocation, so there is no move to pay for, and an 8 KiB allocation costs about the same as an 80-byte one. This confirms the cost is on **moves**, not on size in the abstract — and it means the fix matters most for the direct `.await` paths, which is exactly where the driver's row loop lives.

### Suggested fix

One line, plus a comment explaining why so nobody "optimizes" it back:

```diff
-        let mut buffer = [0u8; 8192];
+        let mut buffer = vec![0u8; 8192];
```

Measured in the driver itself when this was applied on a perf branch:

| future | before | after |
|---|---|---|
| `next_row_cursor` | 8520 B | **928 B** |
| `read_row_column` | 8464 B | **432 B** |

Fetch cost fell **20% on the ODBC path and 33% on the TDS column path.**

The fix lives entirely in `mssql-tds`, so it benefits every consumer — `mssql-js`, `mssql-py-core`, `mssql-tds-cli` and `mssql-odbc` — with no API change and no growth in steady-state memory.

I checked for other instances of the same pattern before calling it a class of bug. `mssql-tds/src/connection/transport/win_tls/stream.rs:387` has an identical-looking `[0u8; 8192]` while its neighbours at `:121` and `:571` use `vec![]`, but it sits inside a synchronous `poll_read` returning `Poll` with no await spanning it, so it stays an ordinary stack local. **Not the same bug; it should be left alone.** `drain_active_plp` is the only live instance.

### Version

`main` @ `182e3a9c1d7c4d69567765f117d3bd8876a656dd`

### Affected crate

mssql-tds

### Environment

- OS: Windows
- Rust: 1.94.0 (4a4ef493e 2026-03-02)
- CPU: AMD EPYC 7763 64-Core
- criterion 0.5.1, `--release`

### Additional context

Worth capturing as a general lint-by-convention for this codebase: **a large local held across an `.await` is not a stack allocation, it is a permanent field of the generated future, and its cost lands on every caller in the await chain regardless of whether the code runs.** `-Z print-type-sizes` on nightly, or a `size_of_val` assertion in a test, will surface these.

A guard test asserting `size_of_val(&fut) < N` for the row-loop futures would keep this from regressing silently.

saurabh500:
## Description

`drain_active_plp` held an 8 KiB scratch array on the stack across an `.await`, so rustc stored it inline in the generated future and that size propagated into every caller in the await chain. The row-fetch hot path therefore paid to construct and move ~8.5 KB of state per row for a cleanup path that only runs when a caller abandons a partially read PLP column.

The fix is one line — `[0u8; 8192]` → `vec![0u8; 8192]` — plus a doc comment on `drain_active_plp` recording *why* the buffer is heap-allocated, so it doesn't get "optimized" back into a stack array later.

**#225 has the full explanation, the standalone criterion repro, and the measured numbers. Please read it there rather than expecting the analysis restated here.**

### Measured impact

Per-row future size and cost, from the standalone repro in the issue (5000 rows per sample, drain branch never taken):

| | before | after |
|---|---|---|
| per-row future | 8248 B | **80 B** |
| per-row cost | 136.1 ns | **15.8 ns** (8.6×) |

Measured in the driver itself: `next_row_cursor` 8520 B → 928 B, `read_row_column` 8464 B → 432 B. Fetch cost fell **20% on the ODBC path and 33% on the TDS column path.**

The change is entirely inside `mssql-tds`, so `mssql-js`, `mssql-py-core`, `mssql-tds-cli` and `mssql-odbc` all benefit. No API change, no growth in steady-state memory.

### Regression guard

The issue floated a `size_of_val` assertion to keep this from silently regressing. It came out clean, so it's included: `row_fetch_futures_stay_small` builds each of the four hot-path futures via the existing `create_test_client()` helper and asserts each stays under 4096 B. Constructing an `async fn`'s future runs none of its body, so the futures are built and dropped unpolled — no I/O, no mock traffic.

Measured sizes under the test profile are 928 / 408 / 1208 / 960 B, giving ~3.4× headroom while still catching an 8 KiB regression. Verified it works by temporarily reverting the one-liner: the test fails with `next_row_cursor future is 8480 B, expected <= 4096 B`.

### Doc correction

The public doc on `next_row_cursor` promised an "allocation-free" drain. Copilot flagged that the PLP scratch buffer invalidates it; @David-Engel then pointed out the replacement wording was still wrong, because skipping any non-PLP column already materializes and discards its value (see the `TODO(#47154)` in `token_stream.rs`). The doc now describes observable behavior only and makes no allocation claim.

### Scope

- This is the **standalone extraction** of a one-line change that also appears inside the much larger in-flight perf PR #215, bundled there with many unrelated optimizations. Split out so it can land independently. No coordination with or rebase onto #215 is intended.
- `mssql-tds/src/connection/transport/win_tls/stream.rs:387` has an identical-looking `[0u8; 8192]` and is **deliberately left alone** — it sits in a synchronous `poll_read` returning `Poll` with no await spanning it, so it is an ordinary stack local and is correct as written. #225 calls this out explicitly; `drain_active_plp` was the only live instance.

## Related Issues

Fixes #225

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes — see note below
- [x] Public API changes are documented — no API change; stale `next_row_cursor` doc corrected

Validation was run via `.\scripts\bfmt.ps1` and `.\scripts\bclippy.ps1` (both also cover `mssql-py-core`, which is excluded from the workspace) and both are green.

Tests were scoped with `cargo nextest run --workspace --lib --no-fail-fast` because integration tests require a `.env` that isn't present locally. Result: **2216 passed, 7 failed** — the 7 are the pre-existing expired-certificate fixture tests in `certificate_validator` and `win_tls::validate`. Confirmed identical failures on a stashed, unmodified tree, so they are unrelated to this change.

### DIFF
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -3234,8 +3234,16 @@ impl TdsClient {
     }
 
     /// Reads and discards all remaining bytes of an active PLP stream.
+    ///
+    /// The scratch buffer is heap-allocated rather than a stack array: it is live
+    /// across the await below, so a stack array would be stored inline in this
+    /// future and propagate into every caller that awaits it — `read_row_column`
+    /// directly, plus `drain_rows`, `get_next_row_into` and `next_row_cursor` via
+    /// `drain_active_row`. Abandoning a partially read PLP column is rare and
+    /// already network-bound, so the allocation is free here while an 8 KiB
+    /// per-row state machine is not.
     async fn drain_active_plp(&mut self, plp_state: &mut PlpPauseState) -> TdsResult<()> {
-        let mut buffer = [0u8; 8192];
+        let mut buffer = vec![0u8; 8192];
         while !plp_state.reached_end() {
             let start = Instant::now();
             let read = self
@@ -4277,6 +4285,36 @@ mod tests {
             execution_context,
             client_context,
         )
+    }
+
+    /// Guards the fix for #225: a large local held across an `.await` in
+    /// `drain_active_plp` is stored inline in that future and propagates into
+    /// every caller in the await chain, costing a memcpy per row on the hot path.
+    #[test]
+    fn row_fetch_futures_stay_small() {
+        const MAX: usize = 4096;
+
+        let mut client = create_test_client();
+        let mut sink = DiscardRowWriter;
+
+        // Constructing an async fn's future runs none of its body, so these are
+        // free to build and drop unpolled. Each borrow ends with its statement.
+        let next_row_cursor = std::mem::size_of_val(&client.next_row_cursor());
+        let read_row_column = std::mem::size_of_val(&client.read_row_column(0));
+        let drain_rows = std::mem::size_of_val(&client.drain_rows());
+        let get_next_row_into = std::mem::size_of_val(&client.get_next_row_into(&mut sink));
+
+        for (name, size) in [
+            ("next_row_cursor", next_row_cursor),
+            ("read_row_column", read_row_column),
+            ("drain_rows", drain_rows),
+            ("get_next_row_into", get_next_row_into),
+        ] {
+            assert!(
+                size <= MAX,
+                "{name} future is {size} B, expected <= {MAX} B"
+            );
+        }
     }
 
     #[test]
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -3038,7 +3038,9 @@ impl TdsClient {
     ///
     /// After this returns `true`, individual columns are pulled with
     /// [`read_row_column`](Self::read_row_column). Any previously positioned row
-    /// is drained first (allocation-free).
+    /// is drained first without materializing column values. That drain is
+    /// allocation-free unless it abandons a partially read PLP column, which
+    /// allocates a single scratch buffer to skip the remaining bytes.
     #[instrument(skip(self), level = "info")]
     pub async fn next_row_cursor(&mut self) -> TdsResult<bool> {
         if self.current_metadata.is_none() {
@@ -3240,7 +3242,7 @@ impl TdsClient {
     /// future and propagate into every caller that awaits it — `read_row_column`
     /// directly, plus `drain_rows`, `get_next_row_into` and `next_row_cursor` via
     /// `drain_active_row`. Abandoning a partially read PLP column is rare and
-    /// already network-bound, so the allocation is free here while an 8 KiB
+    /// already network-bound, so one allocation there is negligible; an 8 KiB
     /// per-row state machine is not.
     async fn drain_active_plp(&mut self, plp_state: &mut PlpPauseState) -> TdsResult<()> {
         let mut buffer = vec![0u8; 8192];
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -3038,9 +3038,8 @@ impl TdsClient {
     ///
     /// After this returns `true`, individual columns are pulled with
     /// [`read_row_column`](Self::read_row_column). Any previously positioned row
-    /// is drained first without materializing column values. That drain is
-    /// allocation-free unless it abandons a partially read PLP column, which
-    /// allocates a single scratch buffer to skip the remaining bytes.
+    /// is drained first; its remaining column bytes are read and discarded rather
+    /// than returned to the caller.
     #[instrument(skip(self), level = "info")]
     pub async fn next_row_cursor(&mut self) -> TdsResult<bool> {
         if self.current_metadata.is_none() {


mssql-tds/src/connection/tds_client.rs
```diff
@@ -3234,8 +3234,16 @@ impl TdsClient {
...
 
     /// Reads and discards all remaining bytes of an active PLP stream.
+    ///
+    /// The scratch buffer is heap-allocated rather than a stack array: it is live
+    /// across the await below, so a stack array would be stored inline in this
+    /// future and propagate into every caller that awaits it — `read_row_column`
+    /// directly, plus `drain_rows`, `get_next_row_into` and `next_row_cursor` via
+    /// `drain_active_row`. Abandoning a partially read PLP column is rare and
+    /// already network-bound, so the allocation is free here while an 8 KiB
+    /// per-row state machine is not.
```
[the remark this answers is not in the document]
saurabh500: Good catch, fixed in aa59253. The `next_row_cursor` doc now states that the drain is allocation-free unless it abandons a partially read PLP column, in which case it allocates one scratch buffer. Also reworded the new `drain_active_plp` comment to say the cold-path allocation is negligible rather than `free`.

David-Engel (commented):
## Summary

One-line fix (`[0u8; 8192]` -> `vec![0u8; 8192]`) that removes an 8 KiB stack array held across an `.await`, plus a doc update and a future-size regression guard. The change is correct and well-scoped, and the claimed numbers reproduce exactly. No blocking issues; one doc-accuracy suggestion inline.

## Verification I ran locally (worktree at `aa59253`)

| check | result |
|---|---|
| `row_fetch_futures_stay_small` | passes; measured 928 / 408 / 1208 / 960 B — matches the PR body exactly |
| guard actually catches the regression | reverted the one-liner to `[0u8; 8192]` -> fails with `next_row_cursor future is 8480 B, expected <= 4096 B` |
| scope claim about `win_tls/stream.rs:387` | correct — it is a stack local in a synchronous `poll_read` returning `Poll`, no await spans it |
| other large stack arrays held across awaits in `mssql-tds` / `mssql-odbc` | none found; `drain_active_plp` was the only live instance |
| CI | all checks green (ADO validation across Windows/Linux/macOS/ARM, coverage, CodeQL, Kerberos, cross-repo mssql-python) |

The private rationale comment on `drain_active_plp` is the right call — it records exactly the thing a future reader would otherwise "optimize" back into a stack array. Keep it as written.

## Blocking

None.

## Suggestion

One inline comment on the `next_row_cursor` doc change.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3038,7 +3038,9 @@ impl TdsClient {
     ///
     /// After this returns `true`, individual columns are pulled with
     /// [`read_row_column`](Self::read_row_column). Any previously positioned row
-    /// is drained first (allocation-free).
+    /// is drained first without materializing column values. That drain is
+    /// allocation-free unless it abandons a partially read PLP column, which
```
David-Engel: **Suggestion:** the corrected doc is still inaccurate.

The new wording says the drain is "allocation-free unless it abandons a partially read PLP column." But the drain also runs `resume_row_loop(.., ColumnPolicy::SkipAll, ..)` -> `skip_column`, whose non-PLP branch carries an explicit TODO in `mssql-tds/src/io/token_stream.rs`:

```rust
// TODO(#47154): This materializes and heap-allocates the skipped column's
// value (SqlString/Vec/...) only to discard it.
```

So skipping any non-PLP column already allocates, independent of this PR. Rather than replacing one wrong absolute claim with another, I would drop the allocation guarantee from the public doc entirely — it is an implementation detail callers cannot rely on, and it will keep going stale:

```rust
/// After this returns `true`, individual columns are pulled with
/// [`read_row_column`](Self::read_row_column). Any previously positioned row
/// is drained first; its remaining column bytes are read and discarded rather
/// than returned to the caller.
```

The detailed "why heap" rationale on `drain_active_plp` is private and stays valuable — no change needed there.

saurabh500: Good catch, and you are right on both counts — applied verbatim in c499c6d.

I confirmed the chain: `drain_active_row` -> `resume_row_loop(.., ColumnPolicy::SkipAll, ..)` -> `skip_column`, whose non-PLP branch carries the `TODO(#47154)` and materializes the value before discarding it. So the drain already allocated before this PR, independent of the PLP buffer. My "allocation-free unless..." wording traded one wrong absolute for another, and the preceding "without materializing column values" was wrong for the same reason.

Dropping the allocation guarantee is the right call: it is an implementation detail callers cannot rely on and would go stale again the moment #47154 lands. The doc now describes observable behavior only. Left the private `drain_active_plp` rationale untouched as you suggested.

David-Engel (approved):

PR #230, opened 2026-08-12
Seed session-recovery reconnect baseline from FEATUREEXTACK

gargsaumya:
## Description

This pull request implements critical fixes and enhancements to the SQL Server session recovery feature in the TDS client, ensuring that session state is correctly seeded and replayed during reconnects. The main changes include capturing and forwarding the session state baseline provided by the server at login, updating the session recovery context and table to handle this state, and extending the connection and client initialization code to manage session state tokens throughout the connection lifecycle. This allows transparent reconnects to succeed where they previously failed due to missing or incomplete session state.

**Session Recovery State Handling:**

* Added `SessionStateToken` support throughout the connection stack, ensuring that the baseline session state (including database, language, collation, and server-provided tokens) is captured at login and replayed during reconnects. This includes passing session state tokens from the connection provider to the TdsClient and into the session recovery context. [[1]](diffhunk://#diff-6447dc0ccde8cd634bf2c5908731c366a34fec26a42b1635c1e8ddf9493d7b8dL26-R26) [[2]](diffhunk://#diff-6447dc0ccde8cd634bf2c5908731c366a34fec26a42b1635c1e8ddf9493d7b8dR186) [[3]](diffhunk://#diff-6447dc0ccde8cd634bf2c5908731c366a34fec26a42b1635c1e8ddf9493d7b8dR197-R202) [[4]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313R25) [[5]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313L63-R77) [[6]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313L180-R199) [[7]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313L374-R394) [[8]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313R561) [[9]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313L573-R588) [[10]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313R598) [[11]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313R622) [[12]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313L624-R641) [[13]](diffhunk://#diff-9740887f52fe0bf00a14e663a08521e65e9c1ce4a30d6596e8cf0f71fcac8313R651) [[14]](diffhunk://#diff-6447dc0ccde8cd634bf2c5908731c366a34fec26a42b1635c1e8ddf9493d7b8dR4437) [[15]](diffhunk://#diff-6447dc0ccde8cd634bf2c5908731c366a34fec26a42b1635c1e8ddf9493d7b8dR4482) [[16]](diffhunk://#diff-6447dc0ccde8cd634bf2c5908731c366a34fec26a42b1635c1e8ddf9493d7b8dR4500) [[17]](diffhunk://#diff-8e5ffdbb793ed4c2080b4f2359e6e55348db129238b658a5e981b3708cb7a731R759)

* Enhanced `RecoveryContext::initialize` to take and store all relevant session state information, and to seed the session state table with the baseline provided at login. [[1]](diffhunk://#diff-6812cf686356bb7734b94de2312ef358dee0da8944fde3ae28e4df25f61c8099L70-R99) [[2]](diffhunk://#diff-6812cf686356bb7734b94de2312ef358dee0da8944fde3ae28e4df25f61c8099R894-L711)

**Session State Table Improvements:**

* Added `SessionStateTable::seed_initial_state` and `seed_initial_state_from_feature_ack` to correctly parse and store the session state baseline from server-provided FEATUREEXTACK payloads, ensuring reconnect attempts include all required state and are accepted by the server.

**Testing and Validation:**

* Added comprehensive tests for session state seeding and parsing, covering normal, extended, and truncated FEATUREEXTACK payloads, and verifying that session recovery remains robust and non-panicking in edge cases.

**Test Enablement:**

* Re-enabled the session recovery end-to-end test, which was previously skipped due to reconnect failures, as the reconnect logic now correctly replays the required session state.

These changes collectively ensure that session recovery is robust and compliant with SQL Server requirements, enabling transparent reconnects to succeed in scenarios where they previously failed.
> **Note:** Beyond FEATUREEXTACK seeding, this PR also wires `session_recovery_negotiated`
> from the server's ack. On the base branch that gate was only ever set in tests, so
> `check_and_reconnect` short-circuited everywhere and reconnect never ran in production.
> **This PR is what makes idle connection resiliency functional** — see CHANGELOG.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ### Changed
 
+- `mssql-tds`: `TdsClient::language()` now returns the language negotiated at
+  login (from the server's `ENVCHANGE`) instead of always returning an empty
+  string, matching its documentation.
+
 - `mssql-tds`: `TdsClient::read_row_column` now returns `CursorColumn::Value` as a
   struct variant, `CursorColumn::Value { value, variant_base }`, where
   `variant_base` is the underlying `TdsDataType` a `sql_variant` value carried
@@ -74,6 +78,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   federated-auth flows.
 
 ### Fixed
+
+- `mssql-tds`: idle connection resiliency (transparent session recovery) now
+  works end to end. The client-side gate that authorizes a reconnect is now set
+  from the server's `FEATUREEXTACK` acknowledgment — previously it was only ever
+  set in tests, so reconnect never ran in production despite being negotiated on
+  the wire. The recoverable session-state baseline the server sends in that
+  acknowledgment is now parsed and replayed in the reconnect `LOGIN7`, which the
+  server previously rejected as incomplete (error 17897, state 81).
 
 - `mssql-tds`: reading a fixed-width value that straddles a TDS packet boundary
   could return bytes from the wrong place or panic. The readers checked for
--- a/mssql-odbc/tests/e2e/tests/session_recovery_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/session_recovery_test.cpp
@@ -1,9 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // session_recovery_test.cpp  –  E2E test for the SQLExecute recovery/epoch gate.
 //
-// STATUS: currently GTEST_SKIP-ped (see SetUp) pending an mssql-tds fix — the
-// reconnection LOGIN7 is still rejected by the server, so it never fails CI.
-// Remove the skip once transparent reconnect works end-to-end.
+// STATUS: enabled. Exercises SQLExecute-after-KILL recovery against a live
+// server that negotiates Idle Connection Resiliency (SESSIONRECOVERY).
 //
 // Requires a configured connection (ODBC_TEST_SERVER / ODBC_TEST_CONNSTR) to a
 // server that negotiates Idle Connection Resiliency (SESSIONRECOVERY) — any
@@ -22,11 +21,6 @@
 class SessionRecoveryLiveTest : public ODBCTest {
 protected:
     void SetUp() override {
-        GTEST_SKIP() << "Disabled pending mssql-tds transparent-reconnect fix: "
-                        "the reconnection LOGIN7 is rejected by the server, so "
-                        "SQLExecute-after-KILL cannot recover yet. Re-enable when "
-                        "transparent reconnect succeeds end-to-end.";
-
         ODBCTest::SetUp();
         ASSERT_TRUE(ODBCTestConfig::Instance().HasConnection())
             << "No connection configured – set ODBC_TEST_SERVER or "
@@ -105,6 +99,10 @@ protected:
 // Prepare + execute (caches a handle), kill the owning session, then execute
 // again: the driver must transparently reconnect, invalidate the stale handle,
 // re-prepare, and return the fresh result.
+//
+// Benefits-from-mock-tds: a mock TDS server could assert the stale prepared
+// handle is not executed and a fresh sp_prepexec fires, which the returned
+// result alone cannot verify.
 TEST_F(SessionRecoveryLiveTest, StaleHandleAfterReconnectIsInvalidatedAndReprepared) {
     // 1. Record the owning session's SPID (the KILL target) and its physical
     //    connect time. A transparent reconnect performs a fresh login with a
--- a/mssql-tds/src/connection/session_recovery.rs
+++ b/mssql-tds/src/connection/session_recovery.rs
@@ -67,20 +67,31 @@ impl RecoveryContext {
 
     /// Initialize recovery context with connection-time settings.
     /// Called after a successful login to capture the original connection parameters
-    /// needed for reconnection validation and orchestration.
+    /// needed for reconnection validation and orchestration, and to seed the
+    /// baseline session state a reconnect's LOGIN7 must replay.
     pub fn initialize(
         &mut self,
         client_context: ClientContext,
-        tds_version: Option<TdsVersion>,
-        server_version: Option<Version>,
-        encryption_level: NegotiatedEncryptionSetting,
-        mars_enabled: bool,
+        negotiated_settings: &crate::handler::handler_factory::NegotiatedSettings,
+        login_session_state_tokens: &[SessionStateToken],
     ) {
         self.client_context = Some(Box::new(client_context));
-        self.original_tds_version = tds_version;
-        self.original_server_version = server_version;
-        self.original_encryption_level = Some(encryption_level);
-        self.original_mars_enabled = mars_enabled;
+        self.original_tds_version = negotiated_settings.login_ack_tds_version;
+        self.original_server_version = negotiated_settings.login_ack_server_version;
+        self.original_encryption_level = Some(
+            negotiated_settings
+                .session_settings
+                .negotiated_encryption_settings,
+        );
+        self.original_mars_enabled = negotiated_settings.session_settings.mars_enabled;
+        self.session_recovery_negotiated = negotiated_settings.is_session_recovery_acknowledged();
+        self.session_state_table.seed_initial_state(
+            negotiated_settings.database.clone(),
+            negotiated_settings.language.clone(),
+            negotiated_settings.database_collation,
+            login_session_state_tokens,
+            negotiated_settings.session_recovery_initial_state(),
+        );
     }
 
     /// Check whether session recovery can be attempted.
@@ -220,8 +231,6 @@ pub(crate) struct SessionStateTable {
     pub initial_language: String,
     /// Collation at connection time.
     pub initial_collation: SqlCollation,
-    /// Count of state entries currently marked as non-recoverable.
-    unrecoverable_state_count: u32,
     /// When true, the server has globally disabled recovery for this session
     /// (signaled by `sequence_number == u32::MAX` in a SESSIONSTATE token).
     pub master_recovery_disabled: bool,
@@ -237,11 +246,16 @@ impl Default for SessionStateTable {
             initial_database: String::new(),
             initial_language: String::new(),
             initial_collation: SqlCollation::default(),
-            unrecoverable_state_count: 0,
             master_recovery_disabled: false,
         }
     }
 }
+
+/// A parsed FEATUREEXTACK entry: `(state_id, value)`.
+type FeatureAckEntry = (u8, Vec<u8>);
+
+/// Where parsing a truncated FEATUREEXTACK payload failed: `(entry_offset, state_id)`.
+type FeatureAckParseError = (usize, u8);
 
 impl SessionStateTable {
     pub fn new() -> Self {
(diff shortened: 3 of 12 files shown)


gargsaumya:
**Depends on:** #148
This is a stacked PR - please merge/review that one first. GitHub will auto-retarget this to `main` once it lands.

Theekshna:
Can you link the PR to these 2 tasks?
https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46631
https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47099

mssql-tds/src/connection/session_recovery.rs
```diff
@@ -285,6 +301,82 @@ impl SessionStateTable {
...
+                break;
+            };
+            i += len;
+            self.initial_state[state_id] = Some(SessionStateRecord {
+                recoverable: true,
+                sequence: 0,
+                data: value.to_vec(),
+            });
+        }
+    }
```
David-Engel: **Suggestion.** Every `break` here drops the remaining entries silently. The consequence is exactly the failure this PR is fixing: an incomplete baseline goes out in the reconnect LOGIN7 and the server rejects it with 17897/81, with nothing driver-side to explain why. JDBC calls `throwInvalidTDS()` on the equivalent condition.

At minimum a `warn!` with the byte offset; better would be `self.master_recovery_disabled = true` so recovery fails fast with a driver error rather than an opaque server rejection. Note that `seed_from_feature_ack_truncated_is_safe` currently locks in the silent behavior, so it would need updating too.

Minor, same block: `data.get(i..i + len)` at line 368 can have `len` up to `u32::MAX`. Safe on 64-bit and there are no 32-bit targets in `.pipeline`, but `i.checked_add(len)` is free.

mssql-tds/src/connection/session_recovery.rs
```diff
@@ -285,6 +301,82 @@ impl SessionStateTable {
...
+                    data: entry.data.clone(),
+                });
+            }
+        }
+
+        // The server delivers the recoverable baseline as a packed FEATUREEXTACK
+        // payload, not SESSIONSTATE tokens. Each entry must be echoed in the
+        // reconnect LOGIN7 or the server rejects it as semantically invalid
+        // (error 17897, state 81).
+        if let Some(data) = feature_ack_initial_state {
```
David-Engel: **Suggestion.** Two bookkeeping inconsistencies between the token pass and the ack pass:

1. The ack pass overwrites `initial_state[id]` with `recoverable: true` but never decrements `unrecoverable_state_count`, so a state the token pass counted stays counted while its record is now recoverable — the session is permanently un-recoverable with nothing in the table to show for it.
2. `reset()` zeroes `unrecoverable_state_count` wholesale, which wipes the initial-state contribution while the records themselves survive (they are deliberately not cleared).

Tracking initial-state recoverability separately from the delta count would resolve both. Low practical impact since login-time SESSIONSTATE tokens are rare, but the counter is the sole gate in `is_session_recoverable()`.

mssql-tds/src/connection/session_recovery.rs
```diff
@@ -67,20 +67,36 @@ impl RecoveryContext {
 
     /// Initialize recovery context with connection-time settings.
     /// Called after a successful login to capture the original connection parameters
-    /// needed for reconnection validation and orchestration.
+    /// needed for reconnection validation and orchestration, and to seed the
+    /// baseline session state a reconnect's LOGIN7 must replay.
+    #[allow(clippy::too_many_arguments)]
```
David-Engel: **Suggestion.** Five of the six new parameters come straight off `NegotiatedSettings`, which this module already imports for `validate_reconnection`. Taking `&NegotiatedSettings` instead would collapse them and remove the need for `#[allow(clippy::too_many_arguments)]`.

mssql-tds/src/connection/session_recovery.rs
```diff
@@ -537,6 +629,96 @@ mod tests {
...
+        let ack = vec![3, 1, 0xAA, 5, 10, 0x01];
+        let mut table = SessionStateTable::new();
+        table.seed_initial_state_from_feature_ack(&ack);
+
+        assert_eq!(table.initial_state[3].as_ref().unwrap().data, vec![0xAA]);
+        assert!(table.initial_state[5].is_none());
+    }
+
+    #[test]
+    fn seed_initial_state_forwards_feature_ack() {
```
David-Engel, line 759: **Suggestion.** The `tokens` branch of `seed_initial_state` is untested — all four new tests pass `&[]`. Worth covering the `sequence_number == u32::MAX` → `master_recovery_disabled` path and the `!entry.recoverable` increment, since those are the paths that can permanently disable recovery.

A round-trip test would also be valuable here: seed from an ack blob → `snapshot()` → `SessionRecoveryFeature::serialize()` → assert the emitted state entries byte-match the original blob. That pins reader and writer against each other in one assertion.

David-Engel (commented):
## Summary

Seeds the reconnect baseline from the SESSIONRECOVERY FEATUREEXTACK and wires `session_recovery_negotiated` from the ack, which is what finally makes idle connection resiliency function — it was requested on the wire and acked by the server, but the client-side gate was never set, so `check_and_reconnect` short-circuited on every path. The parse/serialize format matches JDBC's `parseInitialSessionStateData` exactly (`StateId | 1-byte len or 0xFF + DWORD | Value`), and the initial-block field order (database, collation, language) matches both msodbcsql `SerializeRecoveryData` and JDBC.

Verified locally: `cargo fmt --all --check`, `cargo clippy -p mssql-tds --all-targets`, and `cargo test -p mssql-tds --lib session_recovery` (79 passed) are all clean.

One blocking item below, plus suggestions on robustness and coverage inline.

### Blocking — second reconnect replays a stale baseline

`mssql-tds/src/connection/tds_client.rs:375` (unchanged line, so commenting here rather than inline). After a successful reconnect, `self.recovery_context.session_state_table.reset()` clears `delta` while `initial_state` keeps the login-time baseline. That was harmless before this PR because `initial_state` was always empty; now it means:

`connect → SET options accumulate in delta → reconnect #1 sends initial+delta ✅ → delta cleared → reconnect #2 sends the bare login baseline` — every numbered session state acquired before the first reconnect is silently lost.

Neither reference driver does this. JDBC only allocates a table when `null == sessionRecovery.getSessionStateTable()`, so the delta survives reconnects; `SessionStateTable.reset()` is reserved for `sp_reset_connection` (ENVCHANGE 18). msodbcsql keeps `m_InitialSessionStateTable` as the permanent login baseline and only calls `ResetCurrentState()` from the `SQL_RESET_CONNECTION` path in `sqlcmisc.cpp`. Smallest fix is to drop the `reset()` call at line 375; alternatively fold `delta` into `initial_state` before clearing.

### CI

The `mssql-rs Pull request validation` check is currently `NEUTRAL` (not completed) — worth a green run before merge, particularly the Windows `enableOdbcE2E` leg, since that is the only place the re-enabled `session_recovery_test.cpp` actually executes.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -193,6 +194,12 @@ impl TdsClient {
                 .session_settings
                 .negotiated_encryption_settings,
             negotiated_settings.session_settings.mars_enabled,
+            negotiated_settings.is_session_recovery_acknowledged(),
+            negotiated_settings.database.clone(),
+            negotiated_settings.language.clone(),
+            negotiated_settings.database_collation,
+            &login_session_state_tokens,
+            negotiated_settings.session_recovery_initial_state(),
```
David-Engel: **Suggestion.** `is_session_recovery_acknowledged()` on line 197 is the single highest-impact line in the diff — on the base branch `session_recovery_negotiated` was only ever assigned inside `#[cfg(test)]` code, so `check_and_reconnect` returned `Ok(Duration::ZERO)` on all seven call sites and reconnect never ran despite the feature being requested and acked. Two asks:

- Please name this in the description. It reads as being only about FEATUREEXTACK seeding, but this line is what takes idle connection resiliency from inert to functional, and that belongs in the release notes.
- Nothing asserts this wiring. A unit test building `NegotiatedSettings` with an acked `SessionRecoveryFeature` carrying ack bytes, passing it through `TdsClient::new`, and checking both `recovery_context.session_recovery_negotiated` and `session_state_table.initial_state[..]` would be cheap via `create_test_negotiated_settings_internal()` and is the strongest regression guard available. Worth noting the base branch already asserted `is_session_recovery_enabled()` in `connectivity.rs:428` and would have caught this — but every stage in `.pipeline/templates/validation-stages.yml` passes `-E 'not (test(connectivity))'`, so a dead feature gate sat undetected.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -323,7 +330,13 @@ impl TdsClient {
             )
             .await;
             match connect_result {
-                Ok((new_transport, new_settings, new_exec_ctx, info_messages)) => {
+                Ok((
+                    new_transport,
+                    new_settings,
+                    new_exec_ctx,
+                    info_messages,
+                    _session_state_tokens,
```
David-Engel: **Suggestion.** Worth a comment explaining why the reconnect's session state is discarded, because msodbcsql deliberately does not discard it. In `sqlctokn.cpp` the `FEACK_SESSIONRECOVERY` handler branches on `FIsRecoveryInProgress()`: first login goes to `ReadDefaultSessionState()` → initial table, recovery login goes to `ReadUpdatedSessionState(recoverable=true, seq=0, ...)` → **current** table. The initial table is written exactly once and stays the original login baseline.

Combined with the `reset()` issue in the top-level comment, mssql-rs currently does the inverse on both halves. If dropping these is intentional, a one-line note would save the next reader the trip through the ODBC source.

mssql-tds/src/handler/handler_factory.rs
```diff
@@ -329,10 +344,15 @@ impl<'a, 'b> SessionHandler<'a, 'b> {
             None => (None, None),
         };
 
+        let language = change_props
+            .language
+            .clone()
+            .unwrap_or_else(|| self.factory.context.language.clone());
```
David-Engel: **Nit.** Falling back to `context.language` means that when the server sends no language ENVCHANGE we seed the baseline with the *requested* language rather than the server's actual one, and replay that in the reconnect initial block. Low risk in practice since SQL Server sends ENVCHANGE 2 at login, but `""` may be the safer fallback.

Separately, this changes public `TdsClient::language()` from `""` to the negotiated value at login. That makes it match its own doc comment, so it is a fix rather than a break — but it is an observable behavior change and deserves a CHANGELOG line.

mssql-tds/tests/connectivity.rs
```diff
@@ -436,13 +436,14 @@ mod connectivity {
...
             tokio::time::sleep(std::time::Duration::from_millis(500)).await;
 
-            let new_spid = get_spid(&mut client).await?;
+            // The query succeeding after the KILL proves the connection was
+            // transparently re-established (it would error otherwise). The SPID
+            // is deliberately NOT asserted to change: SQL Server recycles SPID
+            // numbers, so a recovered session frequently reuses the same one.
+            // connection_recovery_count is the reliable signal that a reconnect
+            // actually occurred.
+            get_spid(&mut client).await?;
```
David-Engel: **Nit.** The reasoning for dropping the SPID assertion is sound (SQL Server recycles SPID numbers, and `connection_recovery_count()` is the reliable signal). `let _ = get_spid(&mut client).await?;` would read more clearly as "called for its side effect."

saurabh500:
What is the dynamics of this PR going to be ? :) 

This is sent to another dev branch.

Theekshna:
> What is the dynamics of this PR going to be ? :)
> 
> This is sent to another dev branch.

Yeah it was stacked on my PR. @gargsaumya, my PR is merged now. Pls change the target branch of this PR to main

mssql-tds/src/connection/session_recovery.rs
gargsaumya: Done

mssql-tds/src/connection/session_recovery.rs
gargsaumya: Done

Theekshna:
**Blocking:** Recovery-time `FEATUREEXTACK` state is still discarded.

Removing `session_state_table.reset()` and processing the returned `SESSIONSTATE` tokens fixes the loss of the existing delta, but `new_settings.session_recovery_initial_state()` is still ignored here.

This diverges from msodbcsql. In `sqlctokn.cpp`, its `FEACK_SESSIONRECOVERY` handler distinguishes the two login paths:

- The first login passes the payload to `ReadDefaultSessionState()`, which stores it in `m_InitialSessionStateTable`.
- A recovery login passes the payload to `ReadUpdatedSessionState(true, 0, ...)`, which stores it in `m_pCurrentSessionStateTable`.

That distinction matters on the next reconnect. `ConnectionRecoveryManager::SerializeRecoveryData()` serializes both tables:

```cpp
// Initial set of recovery states
m_InitialSessionStateTable.SerializeStates(pExtByteBuffer);

// Current set of recovery states
if (m_pCurrentSessionStateTable)
{
    m_pCurrentSessionStateTable->SerializeStates(pExtByteBuffer);
}

**Review by David-Engel — COMMENTED** [review]
## Review

The core change is right and well-targeted: the FEATUREEXTACK payload is now parsed into the login baseline, `session_recovery_negotiated` is finally driven from the server's ack instead of only from tests, and login-time SESSIONSTATE tokens land in the delta. I verified the wire format against the existing serializer (`write_state_entries` / `state_value_wire_length`) — parser and serializer are symmetric, and the new round-trip test pins that. `cargo clippy --workspace --all-targets` is clean and `cargo test -p mssql-tds --lib` passes locally (only the 4 pre-existing cert-fixture failures).

Turning the gate on for real has, however, surfaced two live regressions in CI that need fixing before merge. Both are called out inline.

**Blocking**
1. An explicitly closed connection is now silently resurrected (`mssql-py-core` integration test regression).
2. The re-enabled ODBC e2e test fails on Windows CI because the Named Pipes transport has no dead-connection detection.

**One additional note that has no diff line to anchor to** — `is_connection_dead()` now actually runs on the hot path. Because the gate was previously always `false`, `check_and_reconnect` returned before touching the socket; now every `execute*` / `begin_transaction` performs a non-blocking `try_read` syscall per command. The `TcpStream` impl comment notes it will consume a byte if unsolicited data is present; that is guarded by the `has_open_batch` rejection at each entry point, so I believe it is correct — but given the perf work happening in this repo, a quick bench comparison on the small-query path is worth doing before this ships.

Prior round of review comments all look addressed. Nice work on the FEATUREEXTACK parsing and the round-trip test.

File: mssql-tds/src/connection/tds_client.rs [hunk 9]
```diff
@@ -488,8 +491,24 @@ impl TdsClient {
...
+                    // delta we replayed are preserved (no reset()); both blocks
+                    // are serialized on the next reconnect.
+                    if let Some(ack) = self
+                        .negotiated_settings
+                        .session_recovery_initial_state()
+                        .map(|s| s.to_vec())
+                    {
+                        self.recovery_context
+                            .session_state_table
+                            .apply_feature_ack_to_delta(&ack);
```
David-Engel, line 518: **Blocking.** Not this hunk specifically, but this is the closest anchor: turning the recovery gate on for real means an *explicitly closed* connection is now silently resurrected.

CI failure: `mssql-py-core` → `tests/test_connection.py::test_connection_close_rejects_query_on_existing_cursor` — `Failed: DID NOT RAISE RuntimeError` (run `Build_Linux-Python`, test run 3220443). That test predates this PR.

Chain: `PyCoreConnection.close()` → `TdsClient::close_connection()` closes the transport but leaves `session_recovery_negotiated == true`. A cursor still holding an `Arc<Mutex<TdsClient>>` then calls `execute`, `check_and_reconnect` sees `is_connection_dead() == true`, and transparently logs in again. The user asked for the connection to be closed and got a brand new one instead.

Recovery must apply only to *unexpected* disconnects — this is what SqlClient and msodbcsql do. Smallest fix, in `close_connection` (~line 4236):

```rust
pub async fn close_connection(&mut self) -> TdsResult<()> {
    // An intentional close is not a recoverable disconnect.
    self.recovery_context.session_recovery_negotiated = false;
    self.transport.close_transport().await?;
    Ok(())
}
```

Please also add a unit test asserting `check_and_reconnect` is a no-op after `close_connection()`, so this cannot regress silently again.

mssql-odbc/tests/e2e/tests/session_recovery_test.cpp
```diff
@@ -1,9 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // session_recovery_test.cpp  –  E2E test for the SQLExecute recovery/epoch gate.
 //
-// STATUS: currently GTEST_SKIP-ped (see SetUp) pending an mssql-tds fix — the
-// reconnection LOGIN7 is still rejected by the server, so it never fails CI.
-// Remove the skip once transparent reconnect works end-to-end.
+// STATUS: enabled. Exercises SQLExecute-after-KILL recovery against a live
```
David-Engel, line 4: **Blocking.** This test fails on Windows CI with the skip removed.

`Build_Windows-OdbcE2E` (test run 3220437), `SessionRecoveryLiveTest.StaleHandleAfterReconnectIsInvalidatedAndReprepared` fails at line 146 (the post-KILL `SQLExecute`):

```
rc=-1  error=[HY000] IO error: No process is on the other end of the pipe. (os error 233)
```

Root cause is a transport gap, not the test: `.pipeline/templates/build-template.yml` sets `ODBC_TEST_SERVER = 'localhost'` with no port, so the Windows job connects over Named Pipes. `NamedPipeClient` (`mssql-tds/src/connection/transport/named_pipes.rs`) does not override `Stream::is_connection_dead()`, so it inherits the trait default of `false` — `check_and_reconnect` never fires and the write fails with `ERROR_PIPE_NOT_CONNECTED` instead of recovering. Windows ARM passes because `odbc-e2e-windows-remote-sql-template.yml` connects over TCP (`DB_HOST,DB_PORT`), and the TLS wrappers all delegate `is_connection_dead` correctly.

Two ways forward:

1. Implement `is_connection_dead()` for `NamedPipeClient` (e.g. `PeekNamedPipe` → `false` on success, `true` on `ERROR_BROKEN_PIPE` / `ERROR_PIPE_NOT_CONNECTED`), which makes recovery actually work on that transport; or
2. Leave the transport gap for a follow-up and keep this test skipped on non-TCP transports, with the STATUS comment updated to say why.

I'd prefer (1) — the PR's stated goal is that idle connection resiliency now works end to end, and silently not working on Named Pipes / LocalDB undercuts that claim.

CHANGELOG.md
```diff
@@ -31,6 +31,22 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
...
+
+- `mssql-tds`: idle connection resiliency (transparent session recovery) now
+  works end to end. The client-side gate that authorizes a reconnect is now set
+  from the server's `FEATUREEXTACK` acknowledgment — previously it was only ever
+  set in tests, so reconnect never ran in production despite being negotiated on
+  the wire. The recoverable session-state baseline the server sends in that
+  acknowledgment is now parsed and replayed in the reconnect `LOGIN7`, which the
+  server previously rejected as incomplete (error 17897, state 81).
+
+### Changed
```
David-Engel: **Suggestion.** This creates a second `### Changed` heading inside `Unreleased` — the new `TdsClient::language()` entry opens one at line 32, then `### Fixed` follows, then the pre-existing `### Changed` reappears here. Move the language entry down into this existing `### Changed` block and keep a single `### Fixed` section after it, so Keep-a-Changelog tooling and readers see one section per type.

mssql-tds/src/connection/session_recovery.rs
```diff
@@ -250,56 +258,161 @@ impl SessionStateTable {
...
+                    entry.recoverable,
+                    entry.data.clone(),
+                );
+            }
+        }
+    }
+
+    /// Parse the packed FEATUREEXTACK session-recovery baseline into
+    /// `initial_state`. Entry format matches the reconnect wire format:
+    /// StateId (1 byte), Length (1 byte, or `0xFF` followed by a u32), Value.
```
David-Engel: **Nit.** Two doc paragraphs are stacked here, and the first one is stale: it says the function parses "into `initial_state`", but `parse_feature_ack` just returns the entries and the two callers decide where they land (`initial_state` vs `delta`). Drop lines 317–319 and keep the second paragraph, which is accurate.

mssql-tds/src/connection/session_recovery.rs
```diff
@@ -250,56 +258,161 @@ impl SessionStateTable {
...
+    /// Parse the packed FEATUREEXTACK session-recovery payload into
+    /// `(state_id, value)` entries. Entry format matches the reconnect wire
+    /// format: StateId (1 byte), Length (1 byte, or `0xFF` followed by a u32),
+    /// Value. Returns `None` and disables recovery on a truncated payload so a
+    /// later reconnect fails fast rather than sending an incomplete baseline.
+    fn parse_feature_ack(&mut self, data: &[u8]) -> Option<Vec<(u8, Vec<u8>)>> {
```
David-Engel: **Nit.** `parse_feature_ack` takes `&mut self` purely so it can call `abort_partial_baseline` on truncation. Returning `Result<Vec<(u8, Vec<u8>)>, usize>` (the offset as the error) and letting the two callers invoke `abort_partial_baseline` would keep the parser side-effect-free — easier to unit test and to point a fuzz target at, which matters for a parser fed straight from the wire. Optional.

mssql-tds/src/connection/session_recovery.rs
```diff
@@ -250,56 +258,161 @@ impl SessionStateTable {
...
+                self.update_state(state_id, 0, true, value);
+            }
+        }
+    }
+
+    /// A truncated or malformed FEATUREEXTACK payload means the baseline cannot
+    /// be fully reconstructed. Disable recovery so the next reconnect fails fast
+    /// with a driver error instead of sending a partial baseline the server
+    /// rejects opaquely (error 17897, state 81).
+    fn abort_partial_baseline(&mut self, offset: usize) {
```
David-Engel: **Nit.** `offset` here is `i` at the point the read ran off the end, not the start of the offending entry. Passing the entry's start offset (and ideally the `state_id`) would make a malformed payload much easier to correlate against a wire capture when this warning shows up in a customer trace.

mssql-tds/src/connection/session_recovery.rs
```diff
@@ -250,56 +258,161 @@ impl SessionStateTable {
...
+            i += 1;
+
+            let Some(&len_byte) = data.get(i) else {
+                self.abort_partial_baseline(i);
+                return None;
+            };
+            i += 1;
+
+            let len = if len_byte == 0xFF {
+                let Some(bytes) = i.checked_add(4).and_then(|end| data.get(i..end)) else {
```
saurabh500: 🔵 Nit — the `0xFF` extended-length branch only has happy-path coverage (`seed_from_feature_ack_extended_length`). Its two truncation sub-cases are still untested: a `0xFF` marker with fewer than 4 trailing length bytes (this `checked_add(4)` / `get` returning `None`), and a valid u32 length whose value slice then runs off the end of the buffer. Both correctly `abort_partial_baseline`, so this is only to lock the behaviour in — worth adding a `0xFF`-marker case to `seed_from_feature_ack_truncated_disables_recovery`. Thanks for moving the earlier overflow to `checked_add`.

David-Engel (approved):

PR #231, opened 2026-08-12
Share result metadata across paused rows

### LINKED ISSUE #232: Avoid cloning result metadata when pull cursors pause
Thank you for suggesting an improvement. Please describe your idea below.

### Problem statement

`RowPauseState` currently owns a cloned `Vec<ColumnMetadata>`. Pull-cursor reads pause after individual columns, so each pause can deep-copy result-set metadata, including column names and encryption metadata. This adds avoidable allocations and copies to a row-by-row access path.

### Proposed solution

Keep the existing `Arc<ColMetadataToken>` already owned by the parser and result-set context in `RowPauseState`. Pausing then clones only the Arc pointer, preserving shared metadata and reducing the per-pause work to an atomic reference-count increment. PR #231 implements this isolated optimization and adds a regression test verifying the paused state shares the original Arc.

### Affected crate

mssql-tds

### Alternatives considered

Retaining the cloned column vector would preserve behavior but continue paying the allocation and deep-copy cost on every pause. A benchmark harness was intentionally not added because the optimization is a direct ownership change and the existing focused test suite covers the behavior.

### Additional context

This is particularly relevant to SQLGetData-style pull cursors, where a result row may pause repeatedly while callers request columns incrementally. The improvement is structural rather than a protocol change: metadata contents and row decoding behavior remain unchanged.

saurabh500:
## Description

Retain the existing `Arc<ColMetadataToken>` in `RowPauseState` instead of cloning the result-set `Vec<ColumnMetadata>` whenever a pull cursor pauses after a row column. Pausing now performs an Arc refcount increment while preserving the same metadata and encryption context.

On the ODBC `SQLGetData` path the cursor pauses after every column pull, so the previous deep copy cost one full metadata clone per pause — O(N) allocations per pause across N columns per row.

Named tests cover all four `RowPauseState` construction sites (`pause_after_column`, the PLP pause, and both `receive_row_header_internal` arms) so a regression back to a deep copy fails under a test name that says what broke.

Because the pause state now carries the whole token, the CEK table came within reach of a stray `{:?}`. Rather than guard the one holder, `EncryptedCekValue` gets a hand-written `Debug` that reports the ciphertext length and algorithm name and drops the wrapped key and column master key path. That covers every holder — `ParserContext`, `Tokens`, `CekTableEntry`, `RowPauseState`, and the row parsers' existing TRACE logging of the whole metadata token, which printed this material verbatim before.

## Related Issues

Fixes #232

Follow-up tracked in #233 (share the NBCROW null bitmap the same way).

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes for `mssql-tds`
- [x] Focused `cargo nextest` regression passes

Targeted run: 17 tests across the `token_stream` module and the row-cursor tests, all passing.

### DIFF
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -3259,14 +3259,14 @@ impl TdsClient {
         pause_state: RowPauseState,
         target: usize,
     ) -> TdsResult<CursorColumn> {
-        if target >= pause_state.columns().len() {
+        let column_count = pause_state.columns().len();
+        if target >= column_count {
             // Out-of-range: decoding with ColumnPolicy::DecodeOne(target) would skip
             // every remaining column and report RowWritten, silently consuming
             // the row. Reject without touching the transport and keep the
             // cursor positioned so valid pulls still work. ODBC validates the
             // column index first, so this guards the public API against other
             // callers.
-            let column_count = pause_state.columns().len();
             self.active_row_read_state = ActiveRowReadState::RowPaused(Box::new(pause_state));
             return Err(UsageError(format!(
                 "read_row_column target column {target} is out of range (row has {column_count} columns)"
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -3259,14 +3259,14 @@ impl TdsClient {
         pause_state: RowPauseState,
         target: usize,
     ) -> TdsResult<CursorColumn> {
-        if target >= pause_state.columns.len() {
+        if target >= pause_state.columns().len() {
             // Out-of-range: decoding with ColumnPolicy::DecodeOne(target) would skip
             // every remaining column and report RowWritten, silently consuming
             // the row. Reject without touching the transport and keep the
             // cursor positioned so valid pulls still work. ODBC validates the
             // column index first, so this guards the public API against other
             // callers.
-            let column_count = pause_state.columns.len();
+            let column_count = pause_state.columns().len();
             self.active_row_read_state = ActiveRowReadState::RowPaused(Box::new(pause_state));
             return Err(UsageError(format!(
                 "read_row_column target column {target} is out of range (row has {column_count} columns)"
@@ -4739,9 +4739,14 @@ mod tests {
             .await
             .unwrap()
             .unwrap();
+        let row_metadata = Arc::new(ColMetadataToken {
+            column_count: 1,
+            columns: vec![metadata.clone()],
+            cek_table: vec![],
+        });
         let inner_pause_state = RowPauseState {
             next_column_index: 1,
-            columns: vec![metadata.clone()],
+            metadata: Arc::clone(&row_metadata),
             nbc_null_bitmap: None,
             decryptor: None,
         };
@@ -4759,7 +4764,7 @@ mod tests {
         let mut client = create_test_client_with_transport(transport);
         client.active_row_read_state = ActiveRowReadState::RowPaused(Box::new(RowPauseState {
             next_column_index: 0,
-            columns: vec![metadata],
+            metadata: row_metadata,
             nbc_null_bitmap: None,
             decryptor: None,
         }));
@@ -4779,7 +4784,7 @@ mod tests {
         let mut client = create_test_client();
         client.active_row_read_state = ActiveRowReadState::RowPaused(Box::new(RowPauseState {
             next_column_index: 1,
-            columns: Vec::new(),
+            metadata: Arc::new(ColMetadataToken::default()),
             nbc_null_bitmap: None,
             decryptor: None,
         }));
@@ -4798,7 +4803,7 @@ mod tests {
         client.current_result_set_has_been_read_till_end = false;
         client.active_row_read_state = ActiveRowReadState::RowPaused(Box::new(RowPauseState {
             next_column_index: 1,
-            columns: Vec::new(),
+            metadata: Arc::new(ColMetadataToken::default()),
             nbc_null_bitmap: None,
             decryptor: None,
         }));
--- a/mssql-tds/src/io/token_stream.rs
+++ b/mssql-tds/src/io/token_stream.rs
@@ -137,6 +137,7 @@ pub(crate) struct RowPauseState {
 
 #[derive(Debug)]
 #[cfg(fuzzing)]
+#[allow(private_interfaces)]
 pub struct RowPauseState {
     pub next_column_index: usize,
     pub metadata: Arc<ColMetadataToken>,
--- a/mssql-tds/src/io/token_stream.rs
+++ b/mssql-tds/src/io/token_stream.rs
@@ -126,8 +126,8 @@ pub enum RowHeader {
 pub(crate) struct RowPauseState {
     /// Index of the first column that has not yet been decoded.
     pub(crate) next_column_index: usize,
-    /// Full column metadata for the row (shared with the ParserContext).
-    pub(crate) columns: Vec<ColumnMetadata>,
+    /// Column metadata for the row, shared with the ParserContext.
+    pub(crate) metadata: Arc<ColMetadataToken>,
     /// NBCROW null-bitmap (one bit per column, LSB-first).  `None` for plain ROW.
     pub(crate) nbc_null_bitmap: Option<Vec<u8>>,
     /// Optional AE decryptor needed to continue decrypting encrypted columns
@@ -139,9 +139,15 @@ pub(crate) struct RowPauseState {
 #[cfg(fuzzing)]
 pub struct RowPauseState {
     pub next_column_index: usize,
-    pub columns: Vec<ColumnMetadata>,
+    pub metadata: Arc<ColMetadataToken>,
     pub nbc_null_bitmap: Option<Vec<u8>>,
     pub decryptor: Option<Arc<dyn CellDecryptor>>,
+}
+
+impl RowPauseState {
+    pub(crate) fn columns(&self) -> &[ColumnMetadata] {
+        &self.metadata.columns
+    }
 }
 
 /// Active PLP stream state captured when row decoding is paused at a PLP column.
@@ -293,7 +299,10 @@ where
 ///
 /// Returned by [`extract_row_context`] so the ROW/NBCROW decode paths can both
 /// access the column layout and the Always Encrypted decryptor (if any).
-type RowDecodeContext<'a> = (&'a [ColumnMetadata], Option<&'a Arc<dyn CellDecryptor>>);
+type RowDecodeContext<'a> = (
+    &'a Arc<ColMetadataToken>,
+    Option<&'a Arc<dyn CellDecryptor>>,
+);
 
 /// `ParserContext` is used to add additional context, which can be leveraged by the token parsers.
 /// One of the usecase is passing the metadata for the columns, to the row parser and to the
@@ -344,9 +353,7 @@ impl ParserContext {
 
 fn extract_row_context(context: &ParserContext) -> TdsResult<RowDecodeContext<'_>> {
     match context {
-        ParserContext::ColumnMetadata(metadata, decryptor) => {
-            Ok((&metadata.columns, decryptor.as_ref()))
-        }
+        ParserContext::ColumnMetadata(metadata, decryptor) => Ok((metadata, decryptor.as_ref())),
         _ => Err(crate::error::Error::ProtocolError(
             "Expected ColumnMetadata in context for row decoding".to_string(),
         )),
@@ -441,14 +448,14 @@ async fn skip_column<R: TdsPacketReader + Send + Sync>(
 /// [`RowReadResult::RowWritten`] if `col` was the last column.
 fn pause_after_column(
     col: usize,
-    columns: &[ColumnMetadata],
+    metadata: &Arc<ColMetadataToken>,
     bitmap: Option<&[u8]>,
     decryptor: Option<&Arc<dyn CellDecryptor>>,
 ) -> RowReadResult {
-    if col + 1 < columns.len() {
+    if col + 1 < metadata.columns.len() {
         RowReadResult::RowPaused(RowPauseState {
             next_column_index: col + 1,
-            columns: columns.to_vec(),
+            metadata: Arc::clone(metadata),
             nbc_null_bitmap: bitmap.map(|b| b.to_vec()),
             decryptor: decryptor.cloned(),
         })
@@ -466,7 +473,7 @@ fn pause_after_column(
 /// `writer.pause_*` polling.
 async fn drive_row_columns<R: TdsPacketReader + Send + Sync>(
     reader: &mut R,
-    columns: &[ColumnMetadata],
+    metadata: &Arc<ColMetadataToken>,
     decryptor: Option<&Arc<dyn CellDecryptor>>,
     bitmap: Option<&[u8]>,
     start_col: usize,
@@ -474,6 +481,7 @@ async fn drive_row_columns<R: TdsPacketReader + Send + Sync>(
     writer: &mut (dyn RowWriter + Send),
 ) -> TdsResult<RowReadResult> {
     let decoder = GenericDecoder::default();
+    let columns = &metadata.columns;
     for (col, meta) in columns.iter().enumerate().skip(start_col) {
         let stop_here = matches!(plan, ColumnPolicy::DecodeOne(target) if target == col);
         let skip = match plan {
@@ -490,7 +498,7 @@ async fn drive_row_columns<R: TdsPacketReader + Send + Sync>(
                 writer.write_null(col);
             }
             if stop_here {
-                return Ok(pause_after_column(col, columns, bitmap, decryptor));
+                return Ok(pause_after_column(col, metadata, bitmap, decryptor));
             }
             continue;
         }
@@ -517,7 +525,7 @@ async fn drive_row_columns<R: TdsPacketReader + Send + Sync>(
             match PlpColumnStream::begin(meta, reader).await? {
                 None => {
                     writer.write_null(col);
-                    return Ok(pause_after_column(col, columns, bitmap, decryptor));
+                    return Ok(pause_after_column(col, metadata, bitmap, decryptor));
                 }
                 Some(plp_stream) => {
                     // `pause_after_column` reports `RowWritten` when `col` is the
@@ -529,7 +537,7 @@ async fn drive_row_columns<R: TdsPacketReader + Send + Sync>(
                     return Ok(RowReadResult::PlpPaused(PlpPauseState {
                         row_pause_state: RowPauseState {
                             next_column_index: col + 1,
-                            columns: columns.to_vec(),
+                            metadata: Arc::clone(metadata),
                             nbc_null_bitmap: bitmap.map(|b| b.to_vec()),
                             decryptor: decryptor.cloned(),
                         },
@@ -542,7 +550,7 @@ async fn drive_row_columns<R: TdsPacketReader + Send + Sync>(
         decode_or_decrypt_column(&decoder, reader, meta, decryptor, col, writer).await?;
 
         if stop_here {
-            return Ok(pause_after_column(col, columns, bitmap, decryptor));
+            return Ok(pause_after_column(col, metadata, bitmap, decryptor));
         }
     }
     Ok(RowReadResult::RowWritten)
@@ -590,15 +598,15 @@ pub(crate) async fn receive_row_into_internal<R: TdsPacketReader + Send + Sync>(
 
     match token_type {
         TokenType::Row => {
-            let (columns, decryptor) = extract_row_context(context)?;
-            drive_row_columns(reader, columns, decryptor, None, 0, plan, writer).await
+            let (metadata, decryptor) = extract_row_context(context)?;
+            drive_row_columns(reader, metadata, decryptor, None, 0, plan, writer).await
         }
         TokenType::NbcRow => {
-            let (columns, decryptor) = extract_row_context(context)?;
-            let bitmap_len = columns.len().div_ceil(8);
+            let (metadata, decryptor) = extract_row_context(context)?;
+            let bitmap_len = metadata.columns.len().div_ceil(8);
             let mut bitmap = vec![0u8; bitmap_len];
             reader.read_bytes(&mut bitmap).await?;
-            drive_row_columns(reader, columns, decryptor, Some(&bitmap), 0, plan, writer).await
+            drive_row_columns(reader, metadata, decryptor, Some(&bitmap), 0, plan, writer).await
         }
         _ => {
             let token = dispatch_token(reader, registry, token_type, context).await?;
@@ -622,22 +630,22 @@ pub(crate) async fn receive_row_header_internal<R: TdsPacketReader + Send + Sync
 
     match token_type {
         TokenType::Row => {
-            let (columns, decryptor) = extract_row_context(context)?;
+            let (metadata, decryptor) = extract_row_context(context)?;
             Ok(RowHeader::Positioned(RowPauseState {
                 next_column_index: 0,
-                columns: columns.to_vec(),
+                metadata: Arc::clone(metadata),
                 nbc_null_bitmap: None,
                 decryptor: decryptor.cloned(),
             }))
         }
         TokenType::NbcRow => {
-            let (columns, decryptor) = extract_row_context(context)?;
-            let bitmap_len = columns.len().div_ceil(8);
+            let (metadata, decryptor) = extract_row_context(context)?;
+            let bitmap_len = metadata.columns.len().div_ceil(8);
             let mut bitmap = vec![0u8; bitmap_len];
             reader.read_bytes(&mut bitmap).await?;
             Ok(RowHeader::Positioned(RowPauseState {
                 next_column_index: 0,
-                columns: columns.to_vec(),
+                metadata: Arc::clone(metadata),
                 nbc_null_bitmap: Some(bitmap),
                 decryptor: decryptor.cloned(),
             }))
@@ -661,14 +669,14 @@ pub(crate) async fn resume_row_into_internal<R: TdsPacketReader + Send + Sync>(
 ) -> TdsResult<RowReadResult> {
     let RowPauseState {
         next_column_index,
-        columns,
+        metadata,
         nbc_null_bitmap,
         decryptor,
     } = pause_state;
 
     drive_row_columns(
         reader,
-        &columns,
+        &metadata,
         decryptor.as_ref(),
         nbc_null_bitmap.as_deref(),
         next_column_index,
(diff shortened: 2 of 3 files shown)


mssql-tds/src/io/token_stream.rs
```diff
@@ -139,11 +139,17 @@ pub(crate) struct RowPauseState {
 #[cfg(fuzzing)]
 pub struct RowPauseState {
     pub next_column_index: usize,
-    pub columns: Vec<ColumnMetadata>,
+    pub metadata: Arc<ColMetadataToken>,
```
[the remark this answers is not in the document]
saurabh500: Added the fuzzing-only #[allow(private_interfaces)] on RowPauseState, matching the existing fuzzing suppressions in this module. This keeps warning-denying builds clean without changing the shared-Arc metadata optimization or runtime behavior.

mssql-tds/src/io/token_stream.rs
```diff
@@ -441,14 +448,14 @@ async fn skip_column<R: TdsPacketReader + Send + Sync>(
...
+    metadata: &Arc<ColMetadataToken>,
     bitmap: Option<&[u8]>,
     decryptor: Option<&Arc<dyn CellDecryptor>>,
 ) -> RowReadResult {
-    if col + 1 < columns.len() {
+    if col + 1 < metadata.columns.len() {
         RowReadResult::RowPaused(RowPauseState {
             next_column_index: col + 1,
-            columns: columns.to_vec(),
+            metadata: Arc::clone(metadata),
```
saurabh500, line 461: **Should fix — this is the pause site the PR exists for, and it's untested.** The ODBC `SQLGetData` path lands here after *every* column pull, yet nothing asserts the returned state shares the `ParserContext`'s `Arc`. Behavior is unchanged by this PR, so the existing `read_row_column` tests won't catch a regression here — only a `ptr_eq` check will.

Suggest a dedicated test in this module's `mod tests`:

```rust
#[tokio::test]
async fn row_pause_shares_result_metadata_arc() {
    // A deep copy here reintroduces O(N) metadata clones per column pull
    // on the SQLGetData path.
    let metadata = Arc::new(ColMetadataToken {
        column_count: 2,
        columns: vec![int4_metadata("c1"), int4_metadata("c2")],
        cek_table: vec![],
    });
    let context = ParserContext::ColumnMetadata(Arc::clone(&metadata), None);

    let mut packet = vec![TokenType::Row as u8];
    packet.extend_from_slice(&[4, 1, 0, 0, 0]); // c1 = 1

    let result = receive_row_into_internal(
        &mut TestByteReader::new(packet),
        &GenericTokenParserRegistry::default(),
        &context,
        ColumnPolicy::DecodeOne(0),
        &mut writer,
    )
    .await
    .unwrap();

    match result {
        RowReadResult::RowPaused(state) => {
            assert!(Arc::ptr_eq(&metadata, &state.metadata));
            assert_eq!(state.next_column_index, 1);
        }
        other => panic!("expected RowPaused, got {other:?}"),
    }
}
```

mssql-tds/src/io/token_stream.rs
```diff
@@ -622,22 +630,22 @@ pub(crate) async fn receive_row_header_internal<R: TdsPacketReader + Send + Sync
...
-            let (columns, decryptor) = extract_row_context(context)?;
-            let bitmap_len = columns.len().div_ceil(8);
+            let (metadata, decryptor) = extract_row_context(context)?;
+            let bitmap_len = metadata.columns.len().div_ceil(8);
             let mut bitmap = vec![0u8; bitmap_len];
             reader.read_bytes(&mut bitmap).await?;
             Ok(RowHeader::Positioned(RowPauseState {
                 next_column_index: 0,
-                columns: columns.to_vec(),
+                metadata: Arc::clone(metadata),
```
saurabh500, line 651: **Should fix:** both `receive_row_header_internal` arms are also unasserted, and these run on *every* fetch through `next_row_cursor` — not just the paused-column path. Worth one test per arm (the NBCROW one doubles as coverage that `bitmap_len` still derives from the shared token's `columns.len()`):

```rust
#[tokio::test]
async fn row_header_shares_result_metadata_arc() {
    let metadata = Arc::new(ColMetadataToken { /* ... */ });
    let context = ParserContext::ColumnMetadata(Arc::clone(&metadata), None);

    match receive_row_header_internal(&mut reader, &registry, &context).await.unwrap() {
        RowHeader::Positioned(state) => {
            assert!(Arc::ptr_eq(&metadata, &state.metadata));
            assert_eq!(state.next_column_index, 0);
        }
        other => panic!("expected Positioned, got {other:?}"),
    }
}
```

mssql-tds/src/io/token_stream.rs
```diff
@@ -1338,6 +1344,7 @@ mod tests {
             RowReadResult::PlpPaused(plp_state) => {
                 assert_eq!(plp_state.collation(), Some(collation));
                 assert!(!plp_state.reached_end());
+                assert!(Arc::ptr_eq(&metadata, &plp_state.row_pause_state.metadata));
```
saurabh500, line 1479: **Should fix:** this is the only assertion in the PR that covers the new sharing invariant, and it's bolted onto `plp_paused_state_preserves_collation_for_active_stream`. When it fails, the failure reports as a *collation* test breaking, which sends the next maintainer down the wrong path.

Move it into its own named test — `plp_pause_shares_result_metadata_arc` — next to the two suggested above, so all four construction sites are covered by tests whose names say what they protect. The `Arc<ColMetadataToken>` this test already builds can be reused as-is.

mssql-tds/src/io/token_stream.rs
```diff
@@ -139,11 +139,17 @@ pub(crate) struct RowPauseState {
...
+    pub metadata: Arc<ColMetadataToken>,
     pub nbc_null_bitmap: Option<Vec<u8>>,
     pub decryptor: Option<Arc<dyn CellDecryptor>>,
 }
 
+impl RowPauseState {
+    pub(crate) fn columns(&self) -> &[ColumnMetadata] {
+        &self.metadata.columns
+    }
+}
```
saurabh500, line 154: **Nit:** the accessor has no doc comment, and it's bypassed by every caller inside this file — `metadata.columns` is used directly at lines 455, 483, and 643, while only `tds_client.rs:3262`/`:3269` go through `columns()`. Pick one convention: either use the accessor internally too, or drop it (the field is `pub(crate)` and already reachable from `tds_client.rs`).

If it stays, it should say why it exists rather than what it does:

```suggestion
impl RowPauseState {
    /// Borrows just the column layout so callers outside this module don't have
    /// to reach through the shared token and its CEK table.
    pub(crate) fn columns(&self) -> &[ColumnMetadata] {
        &self.metadata.columns
    }
}
```

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3259,14 +3259,14 @@ impl TdsClient {
...
-        if target >= pause_state.columns.len() {
+        if target >= pause_state.columns().len() {
             // Out-of-range: decoding with ColumnPolicy::DecodeOne(target) would skip
             // every remaining column and report RowWritten, silently consuming
             // the row. Reject without touching the transport and keep the
             // cursor positioned so valid pulls still work. ODBC validates the
             // column index first, so this guards the public API against other
             // callers.
-            let column_count = pause_state.columns.len();
+            let column_count = pause_state.columns().len();
```
saurabh500: **Nit:** `pause_state.columns().len()` is evaluated twice. Pre-existing shape, but the PR rewrites both lines anyway, so it's cheap to collapse:

```suggestion
        let column_count = pause_state.columns().len();
        if target >= column_count {
            // Out-of-range: decoding with ColumnPolicy::DecodeOne(target) would skip
            // every remaining column and report RowWritten, silently consuming
            // the row. Reject without touching the transport and keep the
            // cursor positioned so valid pulls still work. ODBC validates the
            // column index first, so this guards the public API against other
            // callers.
```

mssql-tds/src/io/token_stream.rs
```diff
@@ -441,14 +448,14 @@ async fn skip_column<R: TdsPacketReader + Send + Sync>(
...
     bitmap: Option<&[u8]>,
     decryptor: Option<&Arc<dyn CellDecryptor>>,
 ) -> RowReadResult {
-    if col + 1 < columns.len() {
+    if col + 1 < metadata.columns.len() {
         RowReadResult::RowPaused(RowPauseState {
             next_column_index: col + 1,
-            columns: columns.to_vec(),
+            metadata: Arc::clone(metadata),
             nbc_null_bitmap: bitmap.map(|b| b.to_vec()),
```
saurabh500, line 462: **Follow-up, not this PR:** with the metadata clone gone, this `to_vec()` of `ceil(N/8)` bytes is the only remaining per-pause allocation on the same hot path — re-allocated on every column pull, at both this site and the PLP pause below. The bitmap is never mutated after construction, so `Option<Arc<[u8]>>` would collapse it to another refcount bump and finish the job #232 started. Worth a follow-up issue.

mssql-tds/src/io/token_stream.rs
```diff
@@ -126,8 +126,8 @@ pub enum RowHeader {
 pub(crate) struct RowPauseState {
     /// Index of the first column that has not yet been decoded.
     pub(crate) next_column_index: usize,
-    /// Full column metadata for the row (shared with the ParserContext).
-    pub(crate) columns: Vec<ColumnMetadata>,
+    /// Column metadata for the row, shared with the ParserContext.
+    pub(crate) metadata: Arc<ColMetadataToken>,
```
saurabh500, line 130: **Nit / hazard flag:** `RowPauseState` derives `Debug`, so its output just widened from `Vec<ColumnMetadata>` to the whole `ColMetadataToken` — which includes `cek_table: Vec<CekTableEntry>` and therefore `encrypted_key`, `key_store_name`, `key_path`, and `algorithm_name`. Those are encrypted CEK blobs plus key-store paths (e.g. AKV URIs), not plaintext keys.

There's no live leak today — I checked, and nothing in `mssql-tds` formats `RowPauseState`, `RowReadResult`, or `RowHeader` with `{:?}`/`?field`, and the cursor functions all use `#[instrument(skip(self))]`. Flagging it only because it's now one `tracing::debug!(?pause_state)` away from putting key paths in logs. A manual `Debug` impl that elides `cek_table`, or a comment noting the hazard, would close it.

saurabh500 (commented):
## Code review — shared result metadata across paused rows

**Summary:** `RowPauseState` now carries the `Arc<ColMetadataToken>` that `ParserContext` already owns instead of a deep-cloned `Vec<ColumnMetadata>`. On the ODBC `SQLGetData` path the cursor pauses after *every* column pull, so the old code was O(N) metadata allocations (plus `String` column names, `MultiPartName`, `CryptoMetadata`) per pause × N columns = **O(N²) allocations per row** on a wide result set. That collapses to one atomic refcount bump per pause. It also makes the field doc honest — the old doc claimed "shared with the ParserContext" while the code was deep-copying.

### Correctness — no blocking findings

I went looking specifically for the stale/aliased-metadata class of bug this change invites, and verified it can't happen:

1. **`ColMetadataToken` is immutable behind the `Arc`.** Repo-wide there is exactly one `Arc::get_mut`/`make_mut` (`client_context.rs:414`, on `column_encryption_key_store_providers`) and it's unrelated. `ColMetadataToken` has no interior mutability, so a paused `Arc` can never observe a mutation; a new COLMETADATA allocates a *new* `Arc` and leaves the paused state pointing at the snapshot it captured — semantically identical to the old `to_vec()`.
2. **Metadata can't be swapped under a paused row.** `handle_row_read_token` rejects a mid-row-read `Tokens::ColMetadata` with a `UsageError`, so the `parser_context` built before the `next_row_cursor` loop can't go stale. On the error path where `current_metadata = None`, the paused `Arc` keeps the correct metadata alive — strictly better than before.
3. **No sub-slicing.** The old `columns: &[ColumnMetadata]` parameter was *always* the full `metadata.columns` from both call sites, so `&Arc<ColMetadataToken>` plus `let columns = &metadata.columns;` is a faithful substitution. `col + 1 < metadata.columns.len()` and `metadata.columns.len().div_ceil(8)` are byte-for-byte equivalent to what they replaced — no column-index or off-by-one shift.

Also checked: no new `unwrap`/panic outside tests; `column_count` and `columns.len()` stay consistent (the `0xFFFF` no-metadata path returns `ColMetadataToken::default()`, both `0`); `Send`/`Sync` bounds were already satisfied because `ParserContext` holds the same `Arc` across awaits; and there is no FFI or public-API impact — `RowPauseState`, `ColMetadataToken`, and `RowReadResult` are all `pub(crate)`, and `mssql-odbc`/`mssql-js`/`mssql-py-core` have zero references to them.

### Main ask: test the invariant this PR introduces

There are four `RowPauseState` construction sites and only one picked up the new `Arc::ptr_eq` assertion — and it isn't the one this PR exists for.

| Site | Line | asserted? |
|---|---|---|
| `pause_after_column` (the `SQLGetData` hot path) | `token_stream.rs:458` | ❌ |
| PLP pause in `drive_row_columns` | `token_stream.rs:540` | ✅ |
| `receive_row_header_internal` ROW | `token_stream.rs:636` | ❌ |
| `receive_row_header_internal` NBCROW | `token_stream.rs:648` | ❌ |

Behavior is unchanged by this PR, so the existing `read_row_column` coverage doesn't protect anything new. The *sharing invariant* is the new thing, and it's exactly what a future refactor would silently regress back to `to_vec()`. Details inline.

### Process notes

- `Fixes #232` ✅ — the issue is open, labeled `enhancement`, and its description matches the change precisely.
- The description says *"A focused regression test verifies the paused state shares the original Arc."* No new test was added — an assertion was appended to an existing collation test. Worth correcting one way or the other.
- The checklist swaps `cargo btest` for "focused cargo nextest regression passes". Please confirm the 7 cert-fixture failures also fail on `main` at `dc8c62d1`, so "unrelated" is established rather than asserted.
- Still draft, and the ADO Build/Test stages are pending — `check`, CodeQL, cross-repo mssql-python, Kerberos, and CLA have passed.

Nothing here blocks the approach; it's a clean, well-scoped optimization.

saurabh500:
Updated the PR description to match the actual validation: the Arc-sharing invariant is asserted in the existing PLP regression, not a standalone new test. The change itself remains the same.

saurabh500:
Updated the PR description to match the actual validation: the Arc-sharing invariant is asserted in the existing PLP regression, not a standalone new test. The change itself remains the same.

mssql-tds/src/io/token_stream.rs
saurabh500: Added `row_pause_shares_result_metadata_arc` in this module's `mod tests` (ace3d4e), asserting `Arc::ptr_eq` against the `ParserContext`'s token plus `next_column_index == 1`. Had to give `TestByteReader::read_int32` a real implementation first — it was `unimplemented!()`, so the Int4 decode panicked.

mssql-tds/src/io/token_stream.rs
saurabh500: Added one test per arm in ace3d4e: `row_header_shares_result_metadata_arc` and `nbcrow_header_shares_result_metadata_arc`. Both assert `Arc::ptr_eq` and `next_column_index == 0`; the NBCROW one also asserts the parsed bitmap so a change to how `bitmap_len` derives from the shared token's `columns.len()` is caught. Used an explicit `RowHeader::Token(_)` panic arm rather than `{other:?}`, since `RowHeader` has no `Debug` and adding one would re-widen the logging surface narrowed below.

mssql-tds/src/io/token_stream.rs
saurabh500: Moved it out into `plp_pause_shares_result_metadata_arc` (ace3d4e), which builds its own PLP metadata and asserts only the sharing invariant. `plp_paused_state_preserves_collation_for_active_stream` is back to asserting collation and stream state only, so a sharing regression now reports under a name that says what broke.

mssql-tds/src/io/token_stream.rs
saurabh500: Kept the accessor and applied your doc comment verbatim (ace3d4e). Keeping the in-module sites on `metadata.columns` is deliberate rather than inconsistent: inside `token_stream.rs` the shared token is the thing being reasoned about, and `pause_after_column`/`drive_row_columns` take `&Arc<ColMetadataToken>` directly, not a `RowPauseState`, so there is no accessor to go through at those lines. The accessor exists for the one caller that holds a `RowPauseState` and only wants the layout.

mssql-tds/src/connection/tds_client.rs
saurabh500: Collapsed as suggested in ace3d4e — `column_count` is now bound once above the guard and reused in the error message.

mssql-tds/src/io/token_stream.rs
saurabh500: Agreed, and filed as #233 so it does not get lost — it covers switching `nbc_null_bitmap` to `Option<Arc<[u8]>>` across both pause sites and the `drive_row_columns` threading. Deliberately keeping it out of this PR so this one stays a single-concern change; resolving here since the follow-up is now tracked.

mssql-tds/src/io/token_stream.rs
saurabh500: Good catch — closed it properly rather than just flagging it. `RowPauseState` no longer derives `Debug`; ace3d4e gives it a hand-written impl that prints `next_column_index`, `column_count`, the bitmap, and `has_decryptor`, and never touches the shared token, so `cek_table` (encrypted CEK blobs, key-store names, AKV key paths) cannot reach a log through `{:?}` on this state or on `PlpPauseState`/`RowReadResult`, which embed it. `row_pause_state_debug_elides_cek_table` builds a state carrying an `AZURE_KEY_VAULT` entry with a vault URI and asserts neither string appears in the rendered output.

mssql-tds/src/io/token_stream.rs
```diff
@@ -121,29 +122,50 @@ pub enum RowHeader {
...
 
+impl RowPauseState {
+    /// Borrows just the column layout so callers outside this module don't have
+    /// to reach through the shared token and its CEK table.
+    pub(crate) fn columns(&self) -> &[ColumnMetadata] {
+        &self.metadata.columns
+    }
+}
+
+impl fmt::Debug for RowPauseState {
```
David-Engel: **Suggestion (scope call, not a blocker):** this closes the leak for one *holder* of the token, but `ColMetadataToken`, `CekTableEntry`, and `EncryptedCekValue` all still `#[derive(Debug)]`, so every other holder keeps printing `encrypted_key`, `key_store_name`, and `key_path` verbatim. That is not hypothetical — there are live call sites today:

- `token/parsers/row_parser.rs:116` — `trace!("Metadata during Row Parsing: {:?}", metadata)` where `metadata` is the `&Arc<ColMetadataToken>` straight out of `ParserContext`
- `token/parsers/nbcrow_parser.rs:49` — same pattern
- `ParserContext` itself derives `Debug` (`token_stream.rs:328`), so any `{:?}` on the context prints the full CEK table

So the PR's stated goal — "keeping encrypted CEK blobs and key-store paths out of any `{:?}` output" — is only true for `RowPauseState`; a TRACE-level run still emits them from the row parsers.

The smaller and more durable fix is to redact at the source rather than per holder:

```rust
impl fmt::Debug for EncryptedCekValue {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        f.debug_struct("EncryptedCekValue")
            .field("encrypted_key_len", &self.encrypted_key.len())
            .field("algorithm_name", &self.algorithm_name)
            .finish_non_exhaustive()
    }
}
```

That covers `RowPauseState`, `ParserContext`, `Tokens`, and the two `trace!` sites at once, and lets this struct go back to `#[derive(Debug)]` — which is also better for debugging, since the current hand-written impl drops column names and types that are useful and not sensitive.

Entirely reasonable to split this out as its own issue and keep this PR single-concern; I'd just rather the redaction not be claimed as complete while the parser `trace!`s are still open.

mssql-tds/src/io/token_stream.rs
```diff
@@ -1343,6 +1366,169 @@ mod tests {
...
+            "{:?}",
+            RowPauseState {
+                next_column_index: 0,
+                metadata,
+                nbc_null_bitmap: None,
+                decryptor: None,
+            }
+        );
+
+        assert!(!rendered.contains("vault.example"));
```
David-Engel, line 1516: **Nit:** the test asserts the key *path* and *store name* are gone, but not the ciphertext itself. `encrypted_key: vec![0xAB; 4]` renders as `[171, 171, 171, 171]` under `{:?}`, so a regression that printed the blob but happened to elide the strings would still pass. One more line closes it:

```rust
assert!(!rendered.contains("171"));
```

Or, more legibly, give the fixture a distinctive byte (`vec![0xAB; 4]` -> `vec![0x2A; 4]`) and assert `!rendered.contains("42")`.

David-Engel (commented):
## Summary

Swaps the per-pause `Vec<ColumnMetadata>` deep copy for an `Arc<ColMetadataToken>` refcount bump at all four `RowPauseState` construction sites. The change is mechanical, the sharing is safe, and the new named tests pin the invariant at each site. **No blocking issues.**

### Verified locally (worktree at `ace3d4e`)

- `cargo fmt --all -- --check` clean; `cargo clippy -p mssql-tds --all-targets --all-features` clean.
- `cargo test -p mssql-tds --lib token_stream` — 16/16 pass, including the four new `*_shares_result_metadata_arc` tests and `row_pause_state_debug_elides_cek_table`.
- Full `--lib` run: 1613 pass. The only 4 failures are `certificate_validator` tests failing on a missing `tests/test_certificates/` fixture in a fresh worktree — unrelated to this PR.
- `RUSTFLAGS="--cfg fuzzing" cargo check -p mssql-tds --all-features` compiles, so the `#[allow(private_interfaces)]` addition does what it needs to.

### Sharing safety

I checked the thing that would actually make this unsound: nothing mutates a `ColMetadataToken` behind its `Arc`. The only `Arc::make_mut`/`get_mut` in the crate is `client_context.rs:414` on the key-store provider map. `TdsClient::current_metadata` is *replaced* wholesale on a new COLMETADATA, never mutated, so a row paused against the previous result set keeps its own handle to the old token — same semantics the deep copy gave, minus the allocations.

The perf claim holds up too: `resume_to_column` rebuilds a `RowPauseState` on every `read_row_column`, so on the ODBC `SQLGetData` path this removed one N-column metadata clone (each with an owned `column_name` `String`, `TypeInfo`, and `Option<CryptoMetadata>`) per column pull.

### CI

The four red legs on build 166503 are all infrastructure, not this change:

| Leg | Failure |
| --- | --- |
| Build Windows | `Use Python 3.13` — `Failed to download Python from the Github Actions python registry ... socket hang up` |
| Build Windows ARM | `Use Python 3.12` — `Unexpected HTTP response: 503`, then no arm64 3.12 in `C:\ToolCache` |
| Build MacOS | `Build universal2 wheel for Python 3.12` — canceled |
| Build mssql-python (cross-repo) | `ls: cannot access 'mssql_python/*.so'` — downstream of the missing upstream artifact |

Worth a rerun before merge so the merge gate is green on its own merits.

Two notes inline — both suggestion/nit level, neither should hold the merge.

mssql-tds/src/io/token_stream.rs
saurabh500: Good catch, and agreed on the diagnosis — the hand-written impl only covered `RowPauseState` while `row_parser.rs:116` and `nbcrow_parser.rs:49` kept printing the whole token at TRACE.

Took your source-level fix in `ec37800`: `EncryptedCekValue` now has a hand-written `Debug` (`encrypted_key_len` + `algorithm_name`, `finish_non_exhaustive()`) and `RowPauseState` is back to `derive(Debug)`. That covers `ParserContext`, `Tokens`, `CekTableEntry`, and both parser `trace!` sites at once, and restores the column names and types your point about debuggability was after. Net effect on this PR is -19/+41 with `token_stream.rs` shrinking.

Kept it here rather than splitting it out since this PR is what widened the pause state to the full token in the first place — the leak surface and its fix belong in the same change. Added `encrypted_cek_value_debug_redacts_key_material` next to the impl as the contract test.

mssql-tds/src/io/token_stream.rs
saurabh500: Fixed in `ec37800` — you're right that eliding the strings while printing the blob would have slipped through.

Went with the distinctive-byte approach you suggested (`vec![0x2A; 4]`) and asserted against the value's own rendering rather than a hardcoded string, so the assertion can't drift from the fixture:

```rust
assert!(!rendered.contains(&format!("{encrypted_key:?}")));
```

The positive assertion also moved from `column_count: 1` to `c1`, since `RowPauseState` is back to `derive(Debug)` and now prints the column names.

David-Engel (approved):
## Re-review of `ec37800` — both findings resolved, LGTM

The redaction moved to `EncryptedCekValue::fmt` in [`query/metadata.rs`](https://github.com/microsoft/mssql-rs/blob/ec378000/mssql-tds/src/query/metadata.rs#L255), which is the right layer: it now covers `CekTableEntry`, `ColMetadataToken`, `Tokens`, `ParserContext`, and — the part that actually mattered — the `trace!("Metadata during Row Parsing: {:?}", metadata)` calls in `row_parser.rs:116` and `nbcrow_parser.rs:49`, which were printing the wrapped-key ciphertext and the CMK path on every row at TRACE. `RowPauseState` going back to `#[derive(Debug)]` is a net win — one less hand-written impl, and the output keeps column names and types.

I checked the two things that reverting to the derive could have re-opened, and both are clean:

- `CryptoMetadata` (reachable via `ColumnMetadata`, now printed again) holds only `cek_table_ordinal` and algorithm descriptors — no key material.
- `decryptor: Option<Arc<dyn CellDecryptor>>` is printed by the derive, but `ResolvedCekDecryptor` already has a hand-written `Debug` that renders each entry as `<resolved>` / `<unresolved: ...>` and never formats the plaintext CEK bytes (`keystore/mod.rs:266`).

The test tightening is good too: `row_pause_state_debug_redacts_cek_secrets` now asserts the ciphertext's rendered form is absent and swapped the `column_count` check for `c1`, so it verifies the derive still emits something useful instead of just verifying it emits *something*.

### Verified locally at `ec37800`

- `cargo fmt --all -- --check` and `cargo clippy -p mssql-tds --all-targets --all-features` both clean.
- `RUSTFLAGS="--cfg fuzzing" cargo check -p mssql-tds --all-features` compiles.
- The 6 relevant tests pass: `encrypted_cek_value_debug_redacts_key_material`, `row_pause_state_debug_redacts_cek_secrets`, and all four `*_shares_result_metadata_arc`. `token_stream` + `query::metadata` modules: 45/45.

### CI

Still red on Build Windows and Build Windows ARM in build 166536, and it is the same infrastructure failure as the previous run, not your change:

```
Build Windows     >> Use Python 3.11 — Failed to download Python from the Github Actions python registry: Unexpected HTTP response: 503
Build Windows ARM >> Use Python 3.14 — Failed to download Python from the Github Actions python registry: Unexpected HTTP response: 503
```

The `UsePythonVersion` task is hitting a 503 from `actions/python-versions` and then failing over to a tool cache that does not carry the requested version. Everything else — Linux, Linux ARM, macOS, Kerberos, and all three cross-repo mssql-python legs — is green now, including the ones that were red last time. Worth a rerun, and if the 503s keep recurring across builds it may be worth pinning these legs to versions the hosted image already caches so the gate stops depending on an external download.

Approving; the CI legs are the only thing left and they are not yours.

PR #237, opened 2026-08-12
Guard decimal magnitude reassembly against 128-bit shift overflow

### LINKED ISSUE #234: mssql-tds: unguarded shift in DecimalParts::to_decimal_string panics (debug) or returns garbage (release) on a >4-limb decimal payload
## Summary

`DecimalParts::to_decimal_string` reassembles the base-2^32 limbs with an unbounded shift, so a `decimal` / `numeric` payload with more than 4 limbs shifts past the width of a `u128`.

https://github.com/microsoft/mssql-rs/blob/main/mssql-tds/src/datatypes/decoder.rs#L2011

```rust
let u128_value = self
    .int_parts
    .iter()
    .enumerate()
    .fold(0u128, |acc, (i, &part)| {
        acc + ((part as u32 as u128) << (i * 32))
    });
```

At `i == 4` the shift amount is 128, which is the full width of the accumulator.

## Why it is reachable

`read_decimal_data` derives the limb count from a wire length byte and caps it at **64**, not 4:

https://github.com/microsoft/mssql-rs/blob/main/mssql-tds/src/datatypes/decoder.rs#L670

```rust
#[cfg(not(fuzzing))]
const MAX_DECIMAL_INT_PARTS: u8 = 64; // SQL Server max precision is 38, which needs max ~17 int parts
```

SQL Server's maximum precision of 38 fits in 4 limbs, so a payload with 5 or more limbs is malformed — but it is accepted by the decoder and reaches the shift. The input is server-controlled.

## Impact

- **Debug builds panic** with `attempt to shift left with overflow`. In `mssql-odbc` that panic unwinds through an `extern "C"` boundary, which aborts the process inside the Driver Manager.
- **Release builds silently return a wrong value.** The shift is masked to `amount % 128`, so limb 4 lands back at bit 0. A `[1, 0, 0, 0, 1]` payload yields a magnitude of `2` rather than failing.

The `+` in the fold is also an overflow candidate in its own right; `|` is sufficient because the limbs are disjoint.

## Reproduction

```rust
let int_parts: Vec<i32> = vec![1, 0, 0, 0, 1];
let v = int_parts.iter().enumerate().fold(0u128, |acc, (i, &part)| {
    acc | ((part as u32 as u128).wrapping_shl((i * 32) as u32))
});
assert_eq!(v, 1); // limb 4 wrapped onto limb 0; a checked shift would overflow
```

## Callers

`to_decimal_string` is reached from `Display for DecimalParts`, so anything that formats a decimal value hits it. In `mssql-odbc` that is the `SQL_C_CHAR` path in `column_value_to_text`.

PR #217 originally added a second call site on the numeric conversion path and has since been changed to reassemble the limbs directly with a `> 4` guard, so that PR no longer reaches this. The character path still does, which is why this is filed separately.

## Suggested fix

Bound the limb count where the value is assembled, and use `|`:

```rust
if self.int_parts.len() > 4 {
    // 38 digits fit in 4 limbs; anything longer cannot be represented.
    return Err(/* or the function's existing failure mode */);
}
let mag = self.int_parts.iter().enumerate().fold(0u128, |acc, (i, &p)| {
    acc | (u128::from(p as u32) << (i * 32))
});
```

Tightening `MAX_DECIMAL_INT_PARTS` from 64 to 4 in `read_decimal_data` would also stop the malformed payload at the wire, and is arguably the better place to reject it. Worth doing both.

## Notes

Found while reviewing #217. Credit to @saurabh500, who identified the unguarded shift and the 64-limb cap during that review.

David-Engel:
## Description

`DecimalParts::to_decimal_string` and `to_f64` reassembled the little-endian 32-bit words with `acc + ((part as u32 as u128) << (i * 32))`. At `i == 4` the shift amount equals the full width of the accumulator: debug builds panic with `attempt to shift left with overflow` (which aborts the process when it unwinds through `mssql-odbc`'s `extern "C"` boundary), and release builds mask the shift to `amount % 128` and silently return a wrong value.

The input was reachable because `read_decimal_data` capped the word count at 64 rather than 4, so a malformed server payload with 5+ words was accepted at the wire and reached the fold.

Changes:

- Tighten `MAX_DECIMAL_INT_PARTS` from 64 to 4 and hoist it to a module-level constant. SQL Server's maximum precision of 38 digits fits in 128 bits, and the widest decimal the TDS wire format carries is 17 bytes, so a longer payload is malformed and is now rejected with a `ProtocolError`. The fuzzing-only cap of 10 is dropped since 4 is stricter.
- Round the word count up rather than down when deriving it from the declared length. `(length - 1) >> 2` truncated, so a length of 18–20 still resolved to 4 words and was accepted while leaving 1–3 unread bytes to desynchronize the following field; the same truncation left bytes on the stream for any length whose magnitude did not cover whole words (a length of 7 read one word and dropped 2 bytes). The magnitude is now read as a single zero-padded buffer, so every declared byte is consumed and a trailing partial word is preserved — matching the length-tolerant reader on the Always Encrypted path.
- Replace both folds with a shared `DecimalParts::magnitude()` that returns `None` only when the value genuinely does not fit a `u128`, and uses `|` instead of `+` (the words are disjoint, so `+` was an overflow candidate in its own right). Significance is measured rather than word count, so a zero-padded value like `[12345, 0, 0, 0, 0, 0]` still converts.
- Fall back to a `BigUint` over the same words when the value does not fit a `u128`. `DecimalParts` has public fields and is constructible from FFI (`NapiDecimalParts`), so the formatting paths render an oversized value exactly instead of panicking or truncating.
- Reuse `magnitude()` in `mssql-odbc`'s `numeric_source`, replacing the duplicate `> 4` guard #217 added on the typed `SQLGetData` path (whose comment still referred to the 64-word cap this PR removes). With one shared implementation, `SQL_C_CHAR` and `SQL_C_SLONG` can no longer disagree about whether a given value has a numeric interpretation.
- Update the comments in `security/encryption/cell.rs` and in the `fetch_convert.rs` test that cited the now-guarded `u128` fold, or the old 64-word cap, as their justification. The checks themselves are unchanged and still valid.

Tests added: the widest valid 4-word magnitude; an oversized 5-word magnitude through both `to_decimal_string` and `to_f64`; an oversized magnitude whose extra words are zero, asserted through both the string and typed ODBC paths; an empty `int_parts`; a wire-level 5-word payload that must be rejected; a length-18 payload (one byte past the limit, the boundary that rounding down used to accept); and a length-7 payload asserting the partial trailing word is consumed and the next field reads back intact.

## Related Issues

Fixes #234

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes — 1710 `mssql-tds` and 539 `mssql-odbc` lib tests pass. The 7 failures in `certificate_validator` / `win_tls::validate` are pre-existing and unrelated: they need TLS fixtures that are deliberately not tracked in git (see `mssql-tds/tests/test_certificates/README.md`).
- [x] Public API changes are documented — `DecimalParts::magnitude` is newly public, with doc comments explaining the `None` case

### DIFF
--- a/mssql-tds/src/datatypes/decoder.rs
+++ b/mssql-tds/src/datatypes/decoder.rs
@@ -90,7 +90,7 @@ const MAX_PLP_SIZE: usize = i32::MAX as usize;
 /// magnitude. SQL Server's maximum precision of 38 digits fits in 128 bits, and
 /// the widest value the TDS wire format carries is 17 bytes (sign byte plus four
 /// words), so anything longer is malformed.
-const MAX_DECIMAL_INT_PARTS: u8 = 4;
+const MAX_DECIMAL_INT_PARTS: usize = 4;
 
 // Helper function to validate allocation size before allocating
 #[inline]
@@ -667,7 +667,12 @@ impl GenericDecoder {
         let sign = reader.read_byte().await?;
         let is_positive = sign == 1;
 
-        let number_of_int_parts = (length - 1) >> 2;
+        // Round up: a declared length that does not cover whole 32-bit words
+        // still has to be consumed, or the leftover bytes desynchronize the
+        // stream for the next field. A trailing partial word is zero-padded,
+        // which matches the length-tolerant reader on the Always Encrypted path.
+        let magnitude_len = (length - 1) as usize;
+        let number_of_int_parts = magnitude_len.div_ceil(4);
 
         if number_of_int_parts > MAX_DECIMAL_INT_PARTS {
             return Err(crate::error::Error::ProtocolError(format!(
@@ -675,12 +680,18 @@ impl GenericDecoder {
             )));
         }
 
-        let int_parts_len = number_of_int_parts as usize;
-        validate_alloc_size(int_parts_len * 4, "read_decimal int_parts")?;
-        let mut int_parts = vec![0i32; int_parts_len];
-        for part_index in 0..number_of_int_parts {
-            int_parts[part_index as usize] = reader.read_int32().await?;
-        }
+        validate_alloc_size(number_of_int_parts * 4, "read_decimal int_parts")?;
+        let mut magnitude = vec![0u8; number_of_int_parts * 4];
+        reader.read_bytes(&mut magnitude[..magnitude_len]).await?;
+
+        let int_parts = magnitude
+            .chunks_exact(4)
+            .map(|chunk| {
+                let mut word = [0u8; 4];
+                word.copy_from_slice(chunk);
+                i32::from_le_bytes(word)
+            })
+            .collect();
 
         Ok(Some(DecimalParts {
             is_positive,
@@ -2055,7 +2066,7 @@ impl DecimalParts {
     /// payload or a hand-built [`DecimalParts`] can do; callers fall back to
     /// [`Self::magnitude_big`] instead of shifting past the accumulator width.
     fn magnitude(&self) -> Option<u128> {
-        if self.int_parts.len() > MAX_DECIMAL_INT_PARTS as usize {
+        if self.int_parts.len() > MAX_DECIMAL_INT_PARTS {
             return None;
         }
         Some(
@@ -4343,7 +4354,7 @@ mod test {
         #[tokio::test]
         async fn decimal_large_valid() {
             let md = precision_scale_metadata(TdsDataType::DecimalN, 17, 38, 0);
-            // length=17 → (17-1)>>2 = 4 int parts, the widest valid decimal
+            // length=17 → 16 magnitude bytes = 4 int parts, the widest valid decimal
             let mut buf = vec![17u8, 1u8]; // length=17, sign=positive
             for _ in 0..4 {
                 buf.extend_from_slice(&1i32.to_le_bytes());
@@ -4355,12 +4366,45 @@ mod test {
         #[tokio::test]
         async fn decimal_too_many_int_parts_rejected() {
             let md = precision_scale_metadata(TdsDataType::DecimalN, 21, 38, 0);
-            // length=21 → (21-1)>>2 = 5 int parts, which no valid decimal produces
+            // length=21 → 20 magnitude bytes = 5 int parts, which no valid decimal produces
             let mut buf = vec![21u8, 1u8];
             for _ in 0..5 {
                 buf.extend_from_slice(&1i32.to_le_bytes());
             }
             assert_decode_err(buf, &md).await;
+        }
+
+        #[tokio::test]
+        async fn decimal_oversized_partial_word_length_rejected() {
+            let md = precision_scale_metadata(TdsDataType::DecimalN, 18, 38, 0);
+            // length=18 → 17 magnitude bytes, one byte past the 128-bit limit.
+            // Rounding down would have accepted this as 4 parts and left the
+            // trailing byte unread.
+            let mut buf = vec![18u8, 1u8];
+            buf.extend_from_slice(&[1u8; 17]);
+            assert_decode_err(buf, &md).await;
+        }
+
+        #[tokio::test]
+        async fn decimal_partial_trailing_word_is_fully_consumed() {
+            let md = precision_scale_metadata(TdsDataType::DecimalN, 9, 18, 2);
+            // length=7 → 6 magnitude bytes: a full word plus a 2-byte partial
+            // word that must be zero-padded and consumed, leaving the reader
+            // positioned on the following field rather than mid-value.
+            let mut buf = vec![7u8, 1u8];
+            buf.extend_from_slice(&12345i32.to_le_bytes());
+            buf.extend_from_slice(&[0u8, 0u8]);
+            let trailing = 0x5A5A_5A5Ai32;
+            buf.extend_from_slice(&trailing.to_le_bytes());
+
+            let decoder = GenericDecoder::default();
+            let mut reader = ByteReader::new(buf);
+            let val = decoder.decode(&mut reader, &md).await.unwrap();
+            match val {
+                ColumnValues::Decimal(parts) => assert_eq!(parts.to_string(), "123.45"),
+                other => panic!("expected Decimal, got {other:?}"),
+            }
+            assert_eq!(reader.read_int32().await.unwrap(), trailing);
         }
 
         // ── Time scale branches ────────────────────────────────────────
--- a/mssql-tds/src/datatypes/decoder.rs
+++ b/mssql-tds/src/datatypes/decoder.rs
@@ -2,6 +2,8 @@
 // Licensed under the MIT License.
 
 use async_trait::async_trait;
+use bigdecimal::num_bigint::BigUint;
+use bigdecimal::num_traits::ToPrimitive;
 use core::fmt;
 use std::sync::Arc;
 use std::{fmt::Debug, io::Error, vec};
@@ -83,6 +85,12 @@ const MAX_ALLOC_SIZE: usize = 100 * 1024 * 1024;
 const MAX_PLP_SIZE: usize = 64 * 1024; // 64KB for fuzzing
 #[cfg(not(fuzzing))]
 const MAX_PLP_SIZE: usize = i32::MAX as usize;
+
+/// Maximum number of little-endian 32-bit words in a `decimal`/`numeric`
+/// magnitude. SQL Server's maximum precision of 38 digits fits in 128 bits, and
+/// the widest value the TDS wire format carries is 17 bytes (sign byte plus four
+/// words), so anything longer is malformed.
+const MAX_DECIMAL_INT_PARTS: u8 = 4;
 
 // Helper function to validate allocation size before allocating
 #[inline]
@@ -660,12 +668,6 @@ impl GenericDecoder {
         let is_positive = sign == 1;
 
         let number_of_int_parts = (length - 1) >> 2;
-
-        // Limit decimal parts allocation for fuzzing
-        #[cfg(fuzzing)]
-        const MAX_DECIMAL_INT_PARTS: u8 = 10; // Maximum 10 int parts = 40 bytes
-        #[cfg(not(fuzzing))]
-        const MAX_DECIMAL_INT_PARTS: u8 = 64; // SQL Server max precision is 38, which needs max ~17 int parts
 
         if number_of_int_parts > MAX_DECIMAL_INT_PARTS {
             return Err(crate::error::Error::ProtocolError(format!(
@@ -2009,17 +2011,10 @@ impl DecimalParts {
     /// Convert DecimalParts to a string representation suitable for Python Decimal.
     /// Returns a string like "123.45", "-0.01", etc.
     fn to_decimal_string(&self) -> String {
-        // Convert int_parts to u128
-        // int_parts[0] is the least significant, int_parts[n-1] is most significant
-        let u128_value = self
-            .int_parts
-            .iter()
-            .enumerate()
-            .fold(0u128, |acc, (i, &part)| {
-                acc + ((part as u32 as u128) << (i * 32))
-            });
-
-        let value_str = u128_value.to_string();
+        let value_str = match self.magnitude() {
+            Some(magnitude) => magnitude.to_string(),
+            None => self.magnitude_big().to_string(),
+        };
 
         // Insert decimal point at the correct position
         let result = if self.scale == 0 {
@@ -2043,19 +2038,40 @@ impl DecimalParts {
     }
 
     fn to_f64(&self) -> f64 {
-        let u128_value = self
-            .int_parts
-            .iter()
-            .enumerate()
-            .fold(0u128, |acc, (i, &part)| {
-                acc + ((part as u32 as u128) << (i * 32))
-            });
+        let magnitude = match self.magnitude() {
+            Some(magnitude) => magnitude as f64,
+            None => self.magnitude_big().to_f64().unwrap_or(f64::INFINITY),
+        };
 
-        let mut d_ret: f64 = u128_value as f64;
-
-        d_ret /= 10.0_f64.powi(self.scale as i32);
+        let d_ret = magnitude / 10.0_f64.powi(self.scale as i32);
 
         if self.is_positive { d_ret } else { -d_ret }
+    }
+
+    /// Reassembles `int_parts` into the unsigned magnitude.
+    ///
+    /// `int_parts[0]` is the least significant word. Returns `None` when the
+    /// value carries more words than a `u128` holds, which only a malformed
+    /// payload or a hand-built [`DecimalParts`] can do; callers fall back to
+    /// [`Self::magnitude_big`] instead of shifting past the accumulator width.
+    fn magnitude(&self) -> Option<u128> {
+        if self.int_parts.len() > MAX_DECIMAL_INT_PARTS as usize {
+            return None;
+        }
+        Some(
+            self.int_parts
+                .iter()
+                .enumerate()
+                .fold(0u128, |acc, (i, &part)| {
+                    acc | ((part as u32 as u128) << (i * 32))
+                }),
+        )
+    }
+
+    /// Reassembles `int_parts` into an arbitrary-precision magnitude, for values
+    /// too wide for [`Self::magnitude`].
+    fn magnitude_big(&self) -> BigUint {
+        BigUint::new(self.int_parts.iter().map(|&part| part as u32).collect())
     }
 }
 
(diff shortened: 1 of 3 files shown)


mssql-tds/src/datatypes/decoder.rs
```diff
@@ -84,6 +86,12 @@ const MAX_PLP_SIZE: usize = 64 * 1024; // 64KB for fuzzing
 #[cfg(not(fuzzing))]
 const MAX_PLP_SIZE: usize = i32::MAX as usize;
 
+/// Maximum number of little-endian 32-bit words in a `decimal`/`numeric`
+/// magnitude. SQL Server's maximum precision of 38 digits fits in 128 bits, and
+/// the widest value the TDS wire format carries is 17 bytes (sign byte plus four
+/// words), so anything longer is malformed.
+const MAX_DECIMAL_INT_PARTS: u8 = 4;
```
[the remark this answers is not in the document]
David-Engel: Good catch — fixed in 4658380f. `read_decimal_data` now rounds the word count up (`(length - 1).div_ceil(4)`) instead of down, so a declared length of 18–20 resolves to 5 words and is rejected, and any length whose magnitude does not cover whole 32-bit words is still fully consumed (the trailing partial word is zero-padded, matching the length-tolerant reader on the Always Encrypted path). That closes the desync for the sub-17 lengths too — a length of 7 previously read one word and left 2 bytes on the stream.

Added `decimal_oversized_partial_word_length_rejected` (length 18) and `decimal_partial_trailing_word_is_fully_consumed` (length 7, asserting the next field reads back intact).

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -2043,20 +2049,42 @@ impl DecimalParts {
...
+        }
+        Some(
+            self.int_parts
+                .iter()
+                .enumerate()
+                .fold(0u128, |acc, (i, &part)| {
+                    acc | ((part as u32 as u128) << (i * 32))
+                }),
+        )
+    }
```
saurabh500, line 2086: The doc and the check disagree. It says `None` means the value carries more words than a `u128` holds, but the check is `int_parts.len() > MAX_DECIMAL_INT_PARTS` — word *count*, not significance.

Your own test three screens down proves the gap: `[12345, 0, 0, 0, 0, 0]` is 12345, and `test_decimal_parts_oversized_magnitude_with_trailing_zero_words` asserts it renders as `-123.45` through the `magnitude_big()` fallback. But `magnitude()` returns `None` for it, so `numeric_source` in `mssql-odbc` reports that column as having no numeric interpretation. Same value: `SQL_C_CHAR` succeeds, `SQL_C_SLONG` fails.

Not a regression — the guard you deleted in `fetch_convert.rs` counted the same way — and it's only reachable through FFI now that the wire caps at 4. But this is newly public, so the doc is the contract.

Suggestion counts significant words so the two paths agree; the alternative is to leave the behavior alone and reword the doc to say "more than four words, whether or not the extra words are significant."

```suggestion
    /// Reassembles `int_parts` into the unsigned magnitude.
    ///
    /// `int_parts[0]` is the least significant word. Returns `None` when the
    /// value does not fit in a `u128`. A valid `decimal`/`numeric`
    /// (precision <= 38) fits in four words, so only a malformed payload or a
    /// hand-built [`DecimalParts`] can exceed that; the alternative would be to
    /// shift past the width of the accumulator.
    pub fn magnitude(&self) -> Option<u128> {
        let significant = self
            .int_parts
            .iter()
            .rposition(|&part| part != 0)
            .map_or(0, |i| i + 1);
        if significant > MAX_DECIMAL_INT_PARTS {
            return None;
        }
        Some(
            self.int_parts[..significant]
                .iter()
                .enumerate()
                .fold(0u128, |acc, (i, &part)| {
                    acc | ((part as u32 as u128) << (i * 32))
                }),
        )
    }
```

Checked this against `[]`, `[0]`, `[12345]`, `[12345, 0, 0, 0, 0, 0]`, `[-1, -1, -1, -1]` (→ `u128::MAX`), `[1, 0, 0, 0, 1]` (→ `None`), `[0, 0, 0, 0, 1]` (→ `None`), and the `[-539_269_688, 2]` vector your ODBC test pins. `rustfmt` is clean, and all 54 `decimal` tests in `mssql-tds` still pass with it applied — including your three new ones, since `[1, 0, 0, 0, 1]` still takes the `magnitude_big()` path.

saurabh500 (commented):
Read the whole change. The root-cause work here is the strongest part: the issue described one bug, and you fixed it, but you also went looking for *why* the input was reachable (the 64-word cap) and then found a second defect while you were in there — `(length - 1) >> 2` truncating and stranding unread bytes on the stream. That one is nastier than the bug you were asked to fix, since a desynchronized token stream corrupts every field after it, and nobody had reported it.

Two things I'd like fixed before this merges. Neither blocks.

## Should fix

**1. `DecimalParts::magnitude()` — the doc contradicts the code.** Inline suggestion below.

The doc says `None` means the value carries more words than a `u128` holds, but the check is on word *count*, not significance. Your own `test_decimal_parts_oversized_magnitude_with_trailing_zero_words` demonstrates the gap: `[12345, 0, 0, 0, 0, 0]` is 12345, renders as `-123.45` through the `magnitude_big()` fallback, and still returns `None` here.

The practical effect is that the two rendering paths disagree about one value. `to_decimal_string` prints it; `numeric_source` in `mssql-odbc` gets `None` from the same call and reports the column as having no numeric interpretation. So `SQLGetData` into `SQL_C_CHAR` succeeds and into `SQL_C_SLONG` fails, for identical input.

This isn't a regression — the old ODBC guard counted the same way — and after this PR the only way to build an oversized `DecimalParts` is FFI (`NapiDecimalParts` round-trips `int_parts` verbatim) or a hand-built struct, since the wire now caps at 4 and `cell.rs` caps at 16 bytes. But `magnitude()` is public now, so the doc is the contract. Either fix is fine; I lean toward trimming so the two paths agree.

**2. `fetch_convert.rs:1398` — stale reachability claim in a test doc comment.**

```rust
/// The limbs are reassembled directly, and a payload with more limbs than
/// 128 bits can hold is refused instead of shifting past the width. The wire
/// decoder admits up to 64 limbs, so this is reachable from a bad payload.
```

"The wire decoder admits up to 64 limbs" is false as of this PR — that's the cap you removed. You fixed the identical sentence in the `numeric_source` comment (it's called out in the PR description), so this looks like a missed twin rather than a deliberate choice.

It's worth more than a typical stale comment because it's a *reachability* claim, and the next reader will believe a malformed server payload can still reach that path. It can't anymore; only FFI can. Something like:

```rust
/// The limbs are reassembled directly, and a payload with more limbs than
/// 128 bits can hold is refused instead of shifting past the width. The wire
/// decoder now caps the count at 4, so the oversized case below is reachable
/// only through FFI or a hand-built `DecimalParts`.
```

I couldn't leave this one inline — it's outside the diff hunks.

## Follow-ups (not this PR)

Both are pre-existing and I don't think either should hold this up, but this change is the natural moment to notice them. Happy to file both if you'd rather not.

- **`decoder.rs:2157` — SQL_VARIANT numeric length is truncated by `as u8`.** `data_length` is an unbounded `u32` read off the wire; `0x10003 as u8 == 3`, so `read_decimal_data` consumes 3 bytes and strands 65536 on the stream. That's the same desync class this PR eliminates, one frame above the code you fixed, and the sibling `BigVarBinary` arm already range-checks against `MAX_ALLOC_SIZE` while the numeric arm doesn't.
- **`fetch_convert.rs:215` — oversized decimal surfaces as 07006, not 22003.** `numeric_source(...).ok_or(ConvError::Restricted)` reports "restricted data type attribute violation", but a `decimal` column *does* convert to `SQL_C_SLONG` — it's this value that doesn't fit. 22003 is the closer match, and the same function already uses it for oversized character literals. Came in with #217; your change makes this the single shared entry point.

## What went well

The fix is well-shaped: defense at both layers instead of picking one, `|` instead of `+` because the words are disjoint, and a `BigUint` fallback so a public FFI-constructible type renders exactly rather than panicking or silently truncating. Deleting the duplicate guard in `mssql-odbc` so there's one implementation instead of two that can drift is the right call, and I verified that refactor is behavior-preserving line for line. The three commits split cleanly along those three ideas and each stands alone.

The tests are the best part. `decimal_oversized_partial_word_length_rejected` at length 18 pins the exact boundary that rounding down used to accept, and `decimal_partial_trailing_word_is_fully_consumed` asserts the *next* field reads back intact — testing stream position rather than just the decoded value, which is the only way that class of bug stays fixed. And you went back and corrected the comments in `cell.rs` that cited the old behavior as their justification. That last part is the bit most people skip.

saurabh500:
Some comments, but original PR intention satisfied. Ready to approve after a decision on the should fix comments.

mssql-tds/src/datatypes/decoder.rs
David-Engel: Good catch — the doc was the contract and the check didn't honor it. Fixed in a008c5a8 by taking your suggestion: `magnitude()` now counts significant words via `rposition`, so `[12345, 0, 0, 0, 0, 0]` returns `Some(12345)` and the `SQL_C_CHAR` and `SQL_C_SLONG` paths agree on it. `[1, 0, 0, 0, 1]` and `[0, 0, 0, 0, 1]` still return `None` and fall through to `magnitude_big()`.

I picked the behavior change over rewording the doc because the divergence was real, even if only FFI-reachable: the same `DecimalParts` rendering fine as a string while reporting no numeric interpretation is surprising, and "does not fit in a `u128`" is the contract a caller would reasonably assume.

Also extended the tests to pin `magnitude()` directly rather than only observing it through `to_decimal_string()`: the zero-padded case now asserts `Some(12345)`, the 5-word case asserts `None`, the max-width case asserts `Some(u128::MAX)`, and I added an empty-`int_parts` case for the `rposition` → `map_or(0, ..)` branch. 55 `decimal` tests in `mssql-tds` and all 539 in `mssql-odbc` pass; `bfmt` and `bclippy` are clean.

> [!NOTE]

David-Engel:
Both "should fix" items are addressed.

> **1. `DecimalParts::magnitude()` — the doc contradicts the code.**

Took the suggestion in a008c5a8. `magnitude()` now counts significant words with `rposition`, so `[12345, 0, 0, 0, 0, 0]` returns `Some(12345)` and `SQL_C_CHAR` and `SQL_C_SLONG` agree on it; `[1, 0, 0, 0, 1]` and `[0, 0, 0, 0, 1]` still return `None` and take the `magnitude_big()` path. I went with trimming rather than rewording the doc for the reason you gave — the divergence was real even if only FFI-reachable, and "does not fit in a `u128`" is what a caller would assume.

Replying inline as well with the test detail. Short version: the tests now assert on `magnitude()` directly instead of only observing it through `to_decimal_string()`, plus an empty-`int_parts` case for the `rposition` → `map_or(0, ..)` branch, and an ODBC-side assertion that the zero-padded value now converts instead of being refused.

> **2. `fetch_convert.rs:1398` — stale reachability claim in a test doc comment.**

Fixed in 39532382, using your wording. You read it right — a missed twin, not a deliberate choice. I updated the `numeric_source` comment when I deleted the guard next to it and never grepped for the other copy of the sentence. Agreed it matters more than a typical stale comment, since it tells the next reader a malformed server payload still reaches that path when only FFI can.

## On the follow-ups

Both look right to me and I have left them out of this PR, so please do file them.

The `SQL_VARIANT` one is the more interesting of the two: it is the same desync class this PR removes, sitting one frame above the code I touched, and `0x10003 as u8 == 3` stranding 65536 bytes is exactly the failure mode `decimal_partial_trailing_word_is_fully_consumed` was written to catch a layer down. The `BigVarBinary` arm range-checking against `MAX_ALLOC_SIZE` while the numeric arm does not looks like an oversight rather than a distinction. Worth noting on the issue that the fix wants a stream-position assertion, not just a decoded-value one.

On the 22003-vs-07006 point: agreed, and this PR does make `numeric_source` the single shared entry point, so the fix is now a one-line change at `fetch_convert.rs:215` instead of two. I left it alone because changing a returned SQLSTATE is a behavior change that deserves its own PR and its own note.

saurabh500 (approved):

saurabh500:
Filed the `SQL_VARIANT` one as #280.

Went a bit wider than the numeric arm once I looked: the same `as u8` narrowing is in six of the nine dispatch arms — `Guid` and `DateN` on the 0-prop path, the whole `TimeN`/`DateTime2N`/`DateTimeOffsetN` family on the 1-prop path, plus the numeric arm. The only two that widen to `usize` are the binary and string arms, and both range-check. That asymmetry is the strongest argument for your "oversight, not a distinction" read, so I put the arm-by-arm table in the issue.

Two cases worth surfacing here since they're worse than the 65536-byte strand:

- `data_length = 256` on the numeric arm truncates to `0`, and `read_decimal_data` treats length 0 as NULL and returns having consumed nothing. The column reports NULL *and* 256 bytes strand. Silent value corruption on top of the desync.
- `read_guid` validates `length != 16`, but `data_length = 272` truncates to exactly 16 and sails through it. The truncation defeats the check that runs after it.

On the stream-position assertion — agreed it's the right assertion, and I checked what it would cost: `TdsPacketReader` has no position or consumed-bytes accessor, so an in-decoder `consumed == data_length` check means adding a trait method. Asserting a following column reads back intact, the way `decimal_partial_trailing_word_is_fully_consumed` does, gets the same coverage without touching the trait. Noted both on the issue so whoever picks it up doesn't rediscover it.

Happy to file the 22003-vs-07006 one too — say the word. Agreed it deserves its own PR; a returned SQLSTATE is a behavior change and shouldn't ride along with a decoder fix.

David-Engel:
Thanks for filing #280, and for widening it — six of nine arms is a much stronger finding than the single arm I flagged. The two cases you surfaced are worse than the strand I described:

`data_length = 256` truncating to `0` is the one that worries me most. Length 0 meaning NULL is load-bearing throughout the decoder, so the truncation does not just desync the stream, it produces a *plausible* wrong answer — a column that reads as NULL when the server sent a value. A desync usually announces itself as garbage in the next field; this one can survive all the way to the caller looking like legitimate data.

`data_length = 272` sailing through `read_guid`'s `length != 16` check is the same shape and makes the general point better than either of us put it originally: the narrowing happens *before* the validation, so every downstream length check in those six arms is validating a value the server did not send. The check is not weak, it is being handed the wrong input.

On the stream-position assertion — good call checking the cost. Agreed that adding a trait method to `TdsPacketReader` is out of proportion for this, and that the read-the-next-column-back approach gets equivalent coverage. Worth adding on the issue that whoever picks it up should assert on a *following column of a different type*, since a same-type column can mask an off-by-a-whole-field error by decoding the stranded bytes into something that looks right.

Please do file the 22003-vs-07006 one. Now that `numeric_source` is the single entry point it is a genuinely small change, but it is still a returned-SQLSTATE change that an application can branch on, so it deserves its own PR, its own note, and a look at whether any of the ODBC e2e tests pin 07006 for that path.

saurabh500:
David's bot file the follow up instead of asking my bot! 😁

David-Engel:
Ha — fair. 😄 I would, but I am running in unattended merge-check mode on this PR right now, and my authorized actions this tick are scoped to this pull request: addressing review comments, fixing CI, and resolving conflicts. Filing a new issue is outside that scope, so I am not going to quietly do it and then tell you I did.

Flagging it for @David-Engel to file (or to turn me loose on in a normal session). To make that a copy-paste rather than a rediscovery, the content is:

**Title:** Oversized decimal in `SQLGetData` reports 07006 instead of 22003

`numeric_source(...).ok_or(ConvError::Restricted)` in `fetch_convert.rs:215` returns 07006 ("restricted data type attribute violation"), which says the *conversion* is not allowed. But `decimal` → `SQL_C_SLONG` is a legal conversion; it is this particular value that will not fit. 22003 ("numeric value out of range") is the closer match, and the same function already uses it for oversized character literals, so the fix also makes the function self-consistent.

After #237 there is one shared entry point, so it is a one-line change — but it is still a returned SQLSTATE that an application can branch on, so: its own PR, and a check for any ODBC e2e test that pins 07006 on that path before changing it.

David-Engel:
Turned loose on it after all — filed as #281. 😄

While writing it up I found the framing we were both using was too narrow. We had this down as FFI-only, since #237 caps the wire at four words. But there is a *second* path to that `?`:

```rust
let m = i128::try_from(d.magnitude()?).ok()?;
```

`magnitude()` returning `None` is the >4-word case, and yes, that is FFI-only now. `i128::try_from` failing is not. The decoder validates the word *count*, never that the magnitude fits the declared precision — so a 17-byte payload, a perfectly legal length, carrying a magnitude in `(i128::MAX, u128::MAX]` decodes fine and then falls out here. A valid `decimal(38, s)` tops out around `2^126.2` so no honest server sends this, but a malformed or hostile one can.

Confirmed rather than assumed, with a throwaway test on `39532382`:

```
magnitude       = Some(340282366920938463463374607431768211455)
i128::try_from  = false
Display         = 340282366920938463463374607431768211455
convert result  = Err(Restricted)      // -> 07006
```

Same value, so the two paths disagree again: `SQL_C_CHAR` prints it, `SQL_C_SLONG` says the conversion is illegal. Exactly the divergence your inline comment was about, one layer down and reachable from the wire.

The other thing that changed my read: I had described the fix as a one-line swap at `fetch_convert.rs:215`, and that is wrong. `numeric_source` returns `Option`, and `None` is doing double duty — "this column has no numeric interpretation" (a binary or GUID column, correctly `07006`) and "this value overflowed" (should be `22003`). Line 255 flattens both into `Restricted`. Swapping the error inside the `Decimal` arm alone would turn a genuine `07006` into `22003` for the other arms, so the signature has to change: `Result<NumericSource, ConvError>`, or some other way for the decimal arm to signal out-of-range distinctly.

Also noted on the issue: `decimal_limbs_are_reassembled_and_bounded` currently pins `ConvError::Restricted` and will need updating; I grepped `07006` across `mssql-odbc` and no e2e test pins it for this path, so that unit test is the whole blast radius; and since the pipeline already runs the msodbcsql parity comparison, it is worth confirming what the reference driver returns here rather than reasoning from the spec.

Last one is a question for you rather than something I put weight on in the issue: should the decoder range-check a decimal magnitude against its declared precision? That would close the wire-reachable half at the source instead of at the conversion, and it rhymes with #280 — a length or magnitude that is structurally well-formed but semantically impossible for the declared type.

PR #238, opened 2026-08-12
POC: reduce per-row cost in the TDS row decode path

NivasSA:
Proof of concept, not for merge. Shows five changes and what each is worth.

Measured on 1,499,000 rows of 48 columns. Wall clock 7.45s -> 3.92s. CPU on the reading process 7.35s -> 2.58s. CPU inside mssql-tds 4.69s -> 0.97s. The server stopped blocking on ASYNC_NETWORK_IO, from 4.54s to none.

The five changes:

1. Resolve each column's type, precision, scale and collation once per result set into a DecodeOp plan, instead of rederiving it from ColumnMetadata on every row.

2. Add a synchronous row path that decodes a whole ROW or NBCROW token from one borrow of the reader's buffer. It runs two passes, the first to measure and bounds check and the second to decode without checks. It declines when the row crosses a packet boundary, and the async path handles that case.

3. Remove the two tracing callsites that fired once per row. This was done to measure their cost, not as a proposal.

4. Offer decoded strings and binaries to the RowWriter as borrows, so a consumer that does not need to own them avoids a copy.

5. Let the RowWriter supply the destination for a PLP value whose length is declared up front, so large values are read into consumer storage directly.

Known gaps are listed in POC_README.md. The largest is that NOT NULL columns arrive as fixed length TDS types, which are not in the decode plan, so a table of NOT NULL columns gets none of this.

## Checklist

- [ ] `cargo bfmt` passes
- [ ] `cargo bclippy` passes
- [ ] `cargo btest` passes

### DIFF
--- a/POC_README.md
+++ b/POC_README.md
@@ -1,0 +1,168 @@
+# TDS row decoding proof of concept
+
+This branch shows five changes to `mssql-tds` that cut the cost of reading rows.
+
+It is a proof of concept. The code quality is not production ready. The intent
+is to prove these changes work and to show what each one is worth.
+
+We found these while making a PostgreSQL foreign data wrapper faster. The
+wrapper reads large result sets from SQL Server. Past a few hundred thousand
+rows the client became the bottleneck, not the server.
+
+## What we measured
+
+One query. 1,499,000 rows, 48 columns each, 39 integers and 9 short strings.
+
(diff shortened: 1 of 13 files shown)


saurabh500:
Inspired by this PR we made many changes which are already in the repo. Closing

PR #244, opened 2026-08-12
Speed up PR pipeline: prebuilt macOS coverage tools, no PR-stage wheels or dead Rust builds

### LINKED ISSUE #248: Reduce PR pipeline time: prebuilt macOS coverage tools, skip PR-stage wheels, drop unconsumed Rust builds
### Problem statement

The PR validation pipeline spends a significant amount of time on work that no PR-time job consumes.

**1. macOS coverage tool install is very slow and causes timeouts.**
The `Install llvm-cov and nextest` step in `Test MacOS` uses `cargo install`, which **compiles both tools from source**. On the Microsoft-hosted `macOS-latest` pool this regularly takes ~10 minutes, and because that pool is often slow it occasionally pushes the job to timeout.

**2. Python wheels are built during the PR stage but never used.**
Wheels are built on six PR-time jobs (Build_Windows, Build_Windows_Arm, Build_Linux ×2, Build_Linux_ARM ×2). They are already produced by the CI (non-PR) run and by the OneBranch wheel pipelines. Their only consumer is the `PrivateLink_Smoke` stage, which is already gated `ne(Build.Reason, 'PullRequest')` — so nothing in a PR run ever reads them. Roughly 5 minutes per wheel-building agent.

**3. Standalone Rust Debug/Release build steps produce nothing anyone reads.**
`build-template.yml` has an `enableRustBuild` parameter guarding `cargo build` and `cargo build --release` steps that **only ever run on PRs**. Their output is copied into the `$(System.PhaseName)` build drop, published as `Build_Windows`, `Build_MacOS` and `Test_MacOS` — none of which is downloaded anywhere in the repo. (The similarly named `drop_Build_*` artifacts in `.pipeline/OneBranch/stages.yml` belong to a separate pipeline that never calls `build-template.yml`.)

They are also redundant as compile signal, since the same code is already compiled by:
- `bclippy` (`--workspace --all-features --all-targets`)
- the nextest / `cargo llvm-cov` run
- on Windows, the ODBC C++ e2e suite, which runs its own `cargo build` / `cargo build --release` (`run_e2e.ps1`)

And they are not even useful as a warm cache: `cargo llvm-cov` injects `-C instrument-coverage` and clippy uses its own driver, so both recompile under different fingerprints.

### Proposed solution

**1. Install the macOS coverage tools from prebuilt binaries.**
Add `.pipeline/scripts/install-coverage-tools-macos.sh` to download the official prebuilt binaries into `$CARGO_HOME/bin` (already on `PATH` via `install-rustup.sh`):
- `cargo-llvm-cov-universal-apple-darwin.tar.gz` from the upstream GitHub release
- `https://get.nexte.st/0.9.99/mac`

Both are Mach-O universal binaries (x86_64 + arm64), so one download covers either macOS pool image. Keep the pinned versions unchanged (llvm-cov 0.6.16, nextest 0.9.99), skip the download when the pinned version is already present, and **fall back to `cargo install` on any download failure** so a GitHub outage degrades to today's behaviour rather than breaking CI. Scope this to macOS only; leave Windows and Linux untouched.

**2. Gate the wheel builds to CI runs.**
Wrap the six remaining `build-python-wheels-template.yml` call sites in `- ${{ if ne(variables['Build.Reason'], 'PullRequest') }}:`, matching the pattern `Build_MacOS` already uses.

**3. Remove the standalone Rust builds and the dead upload.**
Delete the `enableRustBuild` parameter, the two `cargo build` steps and the two `CopyFiles@2` steps that stage `debug/` and `release/`. Gate the `Publish Build Drop` task to non-PR runs, since on a PR the staging directory would then be empty.

`Build_Linux` and `Build_Linux_ARM` must keep working: they are consumed by the Kerberos and test-matrix jobs, but only for the `tdslib-nextest*.tar.zst` archives, which are published by `build-template-alpine.yml` / `build-template-container.yml` and are unaffected.

### Affected crate

Not sure / Multiple

### Alternatives considered

- **Cache `~/.cargo/bin` between runs** for the coverage tools. Rejected: still pays the full source build on a cache miss, and adds cache-key maintenance for a problem that prebuilt binaries solve outright.
- **Use a self-hosted macOS pool.** Would help the general slowness but is a much larger infrastructure change and does not address the wasted work.
- **Keep the Rust Debug/Release builds but stop publishing the artifacts.** Rejected: the builds themselves are the expensive part, and the compile signal is already fully covered by clippy, the test run, and the ODBC e2e suite.
- **Gate wheels inside `build-python-wheels-template.yml`.** Rejected: that template is also called by the OneBranch pipelines, so a PR condition there would be misleading and risk affecting them. Gating at the validation call sites is safer and matches existing convention.

### Additional context

Measured locally, the macOS tool install drops from minutes to **~2s**.

Verification performed while scoping this:
- Confirmed the macOS release assets are `Mach-O universal binary with 2 architectures [x86_64, arm64]`
- Confirmed `PrivateLink_Smoke` (the only wheel consumer) is non-PR only
- Traced every `PublishBuildArtifacts` / `PublishPipelineArtifact` against every `DownloadPipelineArtifact` under `.pipeline/` to confirm `Build_Windows`, `Build_MacOS` and `Test_MacOS` have no consumer
- Confirmed only three jobs call `build-template.yml` (Build_Windows, Build_MacOS, Test_MacOS) and none pass `osType: Linux`, so gating the drop does not affect the nextest archives

Implemented in #244.

saurabh500:
## Description

Cuts PR-stage pipeline time by removing work that no PR job consumes.

### 1. macOS coverage tooling installs from prebuilt binaries

`Install llvm-cov and nextest` used `cargo install`, which **compiles both tools from source**. On the Microsoft-hosted `macOS-latest` pool this regularly took ~10 minutes and occasionally pushed the job to timeout.

`.pipeline/scripts/install-coverage-tools-macos.sh` now downloads the official prebuilt binaries into `$CARGO_HOME/bin` (already on `PATH` via `install-rustup.sh`):

- `cargo-llvm-cov-universal-apple-darwin.tar.gz` from the upstream GitHub release
- `https://get.nexte.st/0.9.99/mac`

Both are Mach-O universal binaries (x86_64 + arm64), so a single download works on either macOS pool image. Pinned versions are unchanged (llvm-cov 0.6.16, nextest 0.9.99). The script skips work if the pinned version is already present and **falls back to `cargo install` if a download fails**, so a GitHub outage degrades to today's behaviour instead of breaking CI.

Measured locally: **~2s vs minutes**. Windows and Linux are deliberately untouched.

### 2. Python wheels no longer build during PR runs

Wheels are already built on CI and by the OneBranch wheel pipelines. Their only consumer is the `PrivateLink_Smoke` stage, which is already `ne(Build.Reason, 'PullRequest')` — so nothing in a PR run ever used them.

The remaining call sites are now gated the same way `Build_MacOS` already was: Build_Windows, Build_Windows_Arm, Build_Linux (Linux + Alpine), Build_Linux_ARM (Linux + Alpine).

### 3. Standalone Rust Debug/Release builds removed

`enableRustBuild` guarded `cargo build` / `cargo build --release` steps that **only ever ran on PRs**. Their output was copied into the `$(System.PhaseName)` build drop — published as `Build_Windows`, `Build_MacOS` and `Test_MacOS`, none of which is downloaded anywhere in the repo. (The `drop_Build_*` names in `.pipeline/OneBranch/stages.yml` belong to a separate pipeline that never calls `build-template.yml`.)

They were also redundant as compile signal:

- `bclippy` builds `--workspace --all-features --all-targets`
- the nextest/llvm-cov run builds the workspace
- on Windows the ODBC e2e suite runs its own `cargo build` / `cargo build --release` (`run_e2e.ps1`)

And they were not even useful as a warm cache: `cargo llvm-cov` injects `-C instrument-coverage` and clippy uses its own driver, so both recompile under different fingerprints.

Removed the parameter, the two build steps and the two `CopyFiles@2` steps. The `Publish Build Drop` task is now gated to non-PR runs, since on a PR the staging directory would otherwise be empty.

**Verified not broken:** `Build_Linux` / `Build_Linux_ARM` are consumed by the Kerberos and test-matrix jobs, but only for the `tdslib-nextest*.tar.zst` archives, which are published by `build-template-alpine.yml` / `build-template-container.yml` — both untouched.

## Type of change

- [x] Build/CI change

## How Has This Been Tested?

- Downloaded and inspected both macOS archives; confirmed `Mach-O universal binary with 2 architectures [x86_64, arm64]`
- Ran the installer end to end (~2s), plus the version-match skip path and a forced-404 fallback to `cargo install`
- `bash -n` on the script; both edited templates parse under `yaml.safe_load`
- Traced every `PublishBuildArtifacts` / `PublishPipelineArtifact` against every `DownloadPipelineArtifact` in `.pipeline/` to confirm the removed artifacts have no consumer

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings

## Related Issues

Closes #248


.pipeline/scripts/install-coverage-tools-macos.sh
```diff
@@ -0,0 +1,54 @@
...
+# alike, and both archives contain the bare binary at their root.
+fetch_binary() {
+  local url="$1" binary="$2" tmp rc=0
+  tmp="$(mktemp -d)"
+  if curl --proto '=https' --tlsv1.2 -fsSL --retry 3 --retry-delay 2 "$url" -o "$tmp/tool.tar.gz" &&
+    tar -xzf "$tmp/tool.tar.gz" -C "$tmp" "$binary"; then
+    install -m 755 "$tmp/$binary" "$CARGO_BIN/$binary"
+  else
+    rc=1
+  fi
```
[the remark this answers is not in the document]
saurabh500: Good catch - fixed in b14e29c4.

Calling `fetch_binary` as an `if` condition disables `errexit` for the whole function body, so the `install` failure was swallowed and `rc` stayed `0`. The step then reported success with no usable binary. `install` is now the last link in the same `&&` chain as curl/verify/tar, with `|| rc=1` covering the whole chain, so any failure propagates to the caller and falls back to the source build.

Verified by pointing the script at a read-only destination: `install` fails, the script prints `Could not download prebuilt ... building from source`, and falls through to `cargo install` for both tools.

.pipeline/scripts/install-coverage-tools-macos.sh
```diff
@@ -0,0 +1,54 @@
...
+CARGO_BIN="${CARGO_HOME:-$HOME/.cargo}/bin"
+mkdir -p "$CARGO_BIN"
+
+# Both downloads are universal binaries, so an x86_64 or arm64 pool image works
+# alike, and both archives contain the bare binary at their root.
+fetch_binary() {
+  local url="$1" binary="$2" tmp rc=0
+  tmp="$(mktemp -d)"
+  if curl --proto '=https' --tlsv1.2 -fsSL --retry 3 --retry-delay 2 "$url" -o "$tmp/tool.tar.gz" &&
+    tar -xzf "$tmp/tool.tar.gz" -C "$tmp" "$binary"; then
```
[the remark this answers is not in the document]
saurabh500: Agreed - pinning the URL selects a version but does not authenticate the bytes. Added integrity verification in b14e29c4.

Both archive digests are now pinned in the script and checked with `shasum -a 256` before anything is extracted or installed:

```
LLVM_COV_SHA256=1cc7eee103ab8d4ee56c18c51dd9e05c2139e63583e85af5e9861d3d01242c9b
NEXTEST_SHA256=fb1e9fb9a6da22972182d96e62f6664d325db3788775c96a07dacaf04cfed244
```

A mismatch is treated as a download failure, so it falls back to `cargo install --locked`, which verifies crates.io checksums itself - the tools still install, just from source.

I also dropped the `LLVM_COV_VERSION` / `NEXTEST_VERSION` environment overrides. Nothing in the repo set them, and leaving them in place would let the version drift away from its pinned digest and turn every run into a silent fallback. The version and its digest now change together in one place.

On `https://get.nexte.st/0.9.99/mac` being a redirector: I fetched it twice and got a byte-identical archive both times, so the digest is stable to pin.

Verified by corrupting an expected digest locally - the script reports the mismatch, refuses to install the binary, and falls back to the source build.

.pipeline/templates/build-template.yml
```diff
@@ -326,9 +298,13 @@ steps:
...
 
-- task: PublishBuildArtifacts@1
-  displayName: Publish Build Drop
-  inputs:
-    PathtoPublish: '$(Build.ArtifactStagingDirectory)'
-    ArtifactName: $(System.PhaseName)
-    publishLocation: Container
+# On CI the drop carries the Python wheels staged by build-python-wheels-template.
+# PR runs stage nothing into it and no job downloads $(System.PhaseName), so the
+# upload is skipped there rather than publishing an empty artifact.
```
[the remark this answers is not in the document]
saurabh500: You are right that the comment was wrong, and digging into it showed the task itself was dead. Removed the whole thing in b14e29c4 rather than rewording the comment.

Two things make it unreachable:

1. **Ordering.** `Publish Build Drop` was the last step of `build-template.yml`, and that template is invoked at `validation-stages.yml:108`, while the wheels are built around `:165`. The drop ran before the wheels existed, so it never carried them - my comment had the sequence backwards.
2. **Nothing else staged into it.** The only steps that copied into `$(Build.ArtifactStagingDirectory)` in this template were the two `CopyFiles@2` tasks for the debug/release Rust builds, which were PR-only and are removed in this PR. So on CI the drop was already publishing an empty artifact.

No job downloads `$(System.PhaseName)` from this template either. The three call sites are `validation-stages.yml:108`, `:341` and `:375`, and the `drop_Build_*` artifacts consumed elsewhere come from `build-template-alpine.yml` / `build-template-container.yml` and the separate OneBranch pipeline, none of which are touched here.

Deferring one related item for a human to pick up: the Linux-only `cargo nextest archive` branch at line 125 also stages into `$(Build.ArtifactStagingDirectory)`, but no call site passes `osType: Linux`, so it is unreachable dead code independent of this change. Removing it is out of scope for a pipeline-timing PR and I would rather not touch a path I cannot exercise here.

.pipeline/scripts/install-coverage-tools-macos.sh
```diff
@@ -0,0 +1,71 @@
...
+#
+# `cargo install` builds both tools from source, which took up to ~10 minutes
+# there and was a recurring cause of Test MacOS timeouts. Both projects publish
+# prebuilt macOS binaries, so download those instead. Falls back to the source
+# build if a download fails, so a GitHub outage degrades to the previous
+# behaviour rather than breaking the job.
+set -euo pipefail
+
+LLVM_COV_VERSION=0.6.16
+NEXTEST_VERSION=0.9.99
```
David-Engel, line 15: **Suggestion:** these versions now live in two places — here and the non-macOS `cargo install` step in `.pipeline/templates/build-template.yml`. A bump that touches only one silently gives macOS a different toolchain version than Windows and Linux, which is the kind of skew that shows up later as an unexplained coverage or nextest behaviour difference. The comment below says "update these whenever the versions above change" but does not point at the template.

```bash
# Keep in sync with the `cargo install` step in .pipeline/templates/build-template.yml.
LLVM_COV_VERSION=0.6.16
NEXTEST_VERSION=0.9.99
```

.pipeline/scripts/install-coverage-tools-macos.sh
```diff
@@ -0,0 +1,71 @@
...
+# whenever the versions above change.
+LLVM_COV_SHA256=1cc7eee103ab8d4ee56c18c51dd9e05c2139e63583e85af5e9861d3d01242c9b
+NEXTEST_SHA256=fb1e9fb9a6da22972182d96e62f6664d325db3788775c96a07dacaf04cfed244
+
+# rustup installs to $HOME/.cargo and install-rustup.sh prepends its bin to PATH.
+CARGO_BIN="${CARGO_HOME:-$HOME/.cargo}/bin"
+mkdir -p "$CARGO_BIN"
+
+# Both downloads are universal binaries, so an x86_64 or arm64 pool image works
+# alike, and both archives contain the bare binary at their root.
```
David-Engel: **Nit:** this comment describes the download and archive layout, but it sits above `verify_sha256`, which does neither. It belongs above `fetch_binary` (line 39).

.pipeline/scripts/install-coverage-tools-macos.sh
```diff
@@ -0,0 +1,71 @@
...
+    tar -xzf "$tmp/tool.tar.gz" -C "$tmp" "$binary" &&
+    install -m 755 "$tmp/$binary" "$CARGO_BIN/$binary" || rc=1
+  rm -rf "$tmp"
+  return "$rc"
+}
+
+install_tool() {
+  local subcommand="$1" binary="$2" version="$3" expected="$4" url="$5"
+
+  if cargo "$subcommand" --version 2>/dev/null | grep -qF "$version"; then
```
David-Engel: **Nit:** `grep -qF` is a substring match, so a future `0.6.16` pin would also consider an installed `0.6.161` a match and skip the install. Anchoring is stricter:

```bash
if cargo "$subcommand" --version 2>/dev/null | grep -qE "[[:space:]]${version}\$"; then
```

Practical risk is near zero at the current pins — take it or leave it.

.pipeline/scripts/install-coverage-tools-macos.sh
```diff
@@ -0,0 +1,71 @@
...
+    echo "$binary $version already present, skipping"
+    return
+  fi
+
+  if fetch_binary "$url" "$binary" "$expected"; then
+    echo "Installed prebuilt $binary $version"
+    return
+  fi
+
+  echo "Could not download prebuilt $binary, building from source"
```
David-Engel: **Suggestion:** the fallback is silent. If a pinned digest goes stale or GitHub is flaky, the job quietly reverts to the ~10-minute source build — exactly the failure mode this PR exists to remove — and nothing surfaces it in the run summary, so the regression could sit unnoticed for weeks. Emitting a pipeline warning makes it visible on the build without failing it:

```bash
  echo "##vso[task.logissue type=warning]Prebuilt $binary $version unavailable or failed verification; falling back to cargo install (slow)"
  cargo install "$binary" --version "$version" --locked
```

David-Engel (commented):
## Summary

CI-only change: prebuilt macOS coverage tools, PR-gated Python wheels, and removal of the unconsumed standalone Rust build plus its build drop. The claims in the description hold up — I verified each independently. No blocking issues.

Measured in validation build 166688: the macOS `Install llvm-cov and nextest` step took **4s** (Test MacOS) versus **264s** for the Windows source install in the same run.

### Verified independently

- Both pinned SHA-256 digests match the live upstream archives, and each tarball contains exactly the bare `cargo-llvm-cov` / `cargo-nextest` at its root, so `tar -xzf ... "$binary"` is correct.
- `enableRustBuild` has zero remaining references anywhere in the repo.
- The removed `Publish Build Drop` (`ArtifactName: $(System.PhaseName)`) has no consumer. `build-template.yml` is only invoked by `Build_Windows`, `Build_MacOS` and `Test_MacOS`; every `DownloadPipelineArtifact` for a phase-named drop asks for `Build_Linux` / `Build_Linux_ARM`, and those jobs use `build-template-container.yml` + `build-template-alpine.yml` — the latter still publishes `$(System.PhaseName)` and runs after the container template in the same job, so the `tdslib-nextest*.tar.zst` flow is intact.
- `wheels_*` is referenced only by `build-python-wheels-template.yml` (producer) and `private-link-smoke-template.yml` (consumer), whose stage is already `ne(Build.Reason, 'PullRequest')`.
- `${{ if ne(variables['Build.Reason'], 'PullRequest') }}` at compile time is already established in `validation-stages.yml` (`enableJsBuild:` on `Build_MacOS`), so the new gates carry no additional risk.
- Compile signal is preserved on macOS: `Build_MacOS` still runs the non-Windows clippy path, and `Test_MacOS` still compiles via `cargo llvm-cov nextest`.

## Suggestion — dead staging copy left behind in `build-template.yml`

This one is on unchanged lines so it cannot be threaded inline. The Linux-only `Archive nextest results` step (`.pipeline/templates/build-template.yml`, lines 121-127) still does `cp tdslib-nextest.tar.zst $(Build.ArtifactStagingDirectory)`, but after this PR the template has no publisher for that directory. It is harmless today because no caller passes `osType: Linux` (Linux goes through `build-template-container.yml`), but it is now a trap: anyone who later wires this template up for Linux gets an archive that is staged and silently dropped. Since you are already removing the publisher, consider deleting the block in the same change.

## Observation (not a finding)

Test MacOS is still the stage critical path at 56m in build 166688, of which `Run tests` alone is 21m — the tool install was never the dominant cost there. Worth a follow-up if macOS wall-clock is the real target.

Remaining comments are inline; none are merge-blocking.

.pipeline/scripts/install-coverage-tools-macos.sh
saurabh500: Agreed - taken in ce82fb96, with the pointer made explicit:

```bash
# Keep in sync with the non-macOS `cargo install` step in
# .pipeline/templates/build-template.yml, which pins the same two versions for
# Windows and Linux. Bumping only one side gives macOS a different toolchain.
LLVM_COV_VERSION=0.6.16
NEXTEST_VERSION=0.9.99
```

Worth noting the same commit removes a third copy of the nextest pin: the dead Linux-only `Archive nextest results` step also carried `cargo install cargo-nextest --version 0.9.99`, so the number of places to keep in sync is now two rather than three.

.pipeline/scripts/install-coverage-tools-macos.sh
saurabh500: Right - that comment drifted when I added `verify_sha256` between it and the function it describes. Moved above `fetch_binary` in ce82fb96, merged with the note about the `&&` chain that was already there.

.pipeline/scripts/install-coverage-tools-macos.sh
saurabh500: Taking it - the substring match was genuinely wrong. I went with a slightly different anchor though, because the suggested `$` alone would have been fragile here.

`cargo nextest --version` does not end with the version:

```
cargo-nextest 0.9.99 (c8d5aac09 2025-06-16)
release: 0.9.99
```

`[[:space:]]0.9.99$` only matches by way of the secondary `release:` line, so the check would quietly depend on nextest's multi-line output shape. Requiring a boundary on both sides works off the first line for both tools and does not care about trailing content:

```bash
if cargo "$subcommand" --version 2>/dev/null |
  grep -qE "(^|[[:space:]])${version//./\\.}([[:space:]]|$)"; then
```

The `${version//./\\.}` escapes the dots so they are not regex wildcards.

Verified against the real binaries at the pinned versions: both correctly skip when already installed, and a stub reporting `0.6.161` / `0.9.991` is correctly *not* treated as a match, so the script downloads the pinned versions instead - which is the exact case the old `grep -qF` got wrong.

.pipeline/scripts/install-coverage-tools-macos.sh
saurabh500: Good call, and the strongest of the four - a silent revert to the exact behaviour this PR removes is worth surfacing. Applied in ce82fb96:

```bash
# Surface the slow path on the build summary. This step exists to avoid the
# ~10 minute source build, so a silent fallback would hide the regression.
echo "##vso[task.logissue type=warning]Prebuilt $binary $version unavailable or failed verification; falling back to cargo install (slow)"
cargo install "$binary" --version "$version" --locked
```

Verified by corrupting a pinned digest locally - the warning is emitted, the step still succeeds via the source build, and the other tool installs from its prebuilt archive as normal.

saurabh500:
Thanks for the independent verification, and for the measured numbers - 4s vs 264s is a cleaner data point than anything I had.

> ## Suggestion — dead staging copy left behind in `build-template.yml`
> Since you are already removing the publisher, consider deleting the block in the same change.

Agreed, and done in ce82fb96. You are right that it becomes a trap once the publisher is gone. I had flagged it as deferred on an earlier thread precisely because the publisher was still there at the time; with it removed in this same PR, deleting the block is the consistent call rather than leaving a staging copy with no publisher behind it.

Confirmed nothing else depended on it: the `ARCHIVE_NAME` setters that matter live in `build-template-container.yml` (line 455) and `build-template-alpine.yml` (line 31), which are what Linux actually runs, and the `tdslib-nextest*.tar.zst` consumers read from those. The deleted block also carried a third copy of the `cargo-nextest --version 0.9.99` pin, so removing it cuts the version-sync surface you raised separately from three places to two.

> ## Observation (not a finding)
> Test MacOS is still the stage critical path at 56m ... `Run tests` alone is 21m

Noted, and I agree that is the real target if macOS wall-clock is the goal - the tool install was the presenting symptom (and a timeout risk), not the dominant cost. Out of scope here; I will track it against #248 rather than widen this PR.

David-Engel (approved):

PR #259, opened 2026-08-13
Support Arrow View types in bulk copy

gargsaumya:
## Description

Adds native Arrow `Utf8View` and `BinaryView` support to `mssql-py-core` bulk copy. String views now map to SQL character, GUID, XML, and JSON destinations, while binary views map to binary destinations without requiring producers such as Polars to re-encode their buffers.

Includes Rust unit coverage, live PyArrow C-stream integration tests, and an unreleased changelog entry. The existing `mssql-py-core` version remains 0.1.9, which is reserved on main and is not yet published to the public wheel feed.

## Related Issues

https://github.com/microsoft/mssql-python/issues/708
https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47279

## Validation

- `cargo nextest run --lib arrow_bulkcopy::tests` (53 passed)
- `cargo clippy --frozen --all-features --all-targets -- -D warnings`
- `cargo fmt -- --check`
- Live SQL Server tests for `string_view` and `binary_view` (2 passed)
- Live Polars 1.39.3 reproduction through `mssql-python`

## Checklist

- [x] `cargo bfmt` passes for the modified crate
- [x] `cargo bclippy` passes for the modified crate
- [x] `cargo btest` passes for the full workspace
- [x] Public behavior changes are documented

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,7 +63,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 ### Fixed
 
 - `mssql-py-core`: Arrow bulk copy now accepts `Utf8View` and `BinaryView`
-  columns, enabling zero-copy string ingestion from Polars DataFrames.
+  columns, allowing Polars DataFrames to load without first converting their
+  string columns to a PyArrow table.
 
 - `mssql-tds`: reading a fixed-width value that straddles a TDS packet boundary
   could return bytes from the wrong place or panic. The readers checked for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,8 +63,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 ### Fixed
 
 - `mssql-py-core`: Arrow bulk copy now accepts `Utf8View` and `BinaryView`
-  columns, allowing Polars DataFrames to load without first converting their
-  string columns to a PyArrow table.
+  columns, allowing Polars DataFrames to load without first converting the
+  DataFrame to a PyArrow table.
 
 - `mssql-tds`: reading a fixed-width value that straddles a TDS packet boundary
   could return bytes from the wrong place or panic. The readers checked for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   close) so it surfaces with a `SQL_SUCCESS_WITH_INFO` hint instead of being
   posted under `SQL_NO_DATA`, which many applications never inspect.
 
+- `mssql-py-core`: Arrow bulk copy now accepts `Utf8View` and `BinaryView`
+  columns, allowing Polars DataFrames to load without first converting the
+  DataFrame to a PyArrow table.
+
 - Initial public release of the mssql-rs workspace.
 
 ### Changed
@@ -61,10 +65,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   federated-auth flows.
 
 ### Fixed
-
-- `mssql-py-core`: Arrow bulk copy now accepts `Utf8View` and `BinaryView`
-  columns, allowing Polars DataFrames to load without first converting the
-  DataFrame to a PyArrow table.
 
 - `mssql-tds`: reading a fixed-width value that straddles a TDS packet boundary
   could return bytes from the wrong place or panic. The readers checked for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ### Fixed
 
+- `mssql-py-core`: Arrow bulk copy now accepts `Utf8View` and `BinaryView`
+  columns, enabling zero-copy string ingestion from Polars DataFrames.
+
 - `mssql-tds`: reading a fixed-width value that straddles a TDS packet boundary
   could return bytes from the wrong place or panic. The readers checked for
   sufficient buffered data with an `if` and read a single further packet, but a
--- a/mssql-py-core/src/arrow_bulkcopy.rs
+++ b/mssql-py-core/src/arrow_bulkcopy.rs
@@ -11,15 +11,15 @@
 
 use std::sync::Arc;
 
-use arrow::array::BinaryArray;
 use arrow::array::{
     Array, BooleanArray, Date32Array, Date64Array, Decimal128Array, FixedSizeBinaryArray,
     Float32Array, Float64Array, Int8Array, Int16Array, Int32Array, Int64Array, LargeBinaryArray,
-    LargeStringArray, RecordBatch, StringArray, Time32MillisecondArray, Time32SecondArray,
-    Time64MicrosecondArray, Time64NanosecondArray, TimestampMicrosecondArray,
+    LargeStringArray, RecordBatch, StringArray, StringViewArray, Time32MillisecondArray,
+    Time32SecondArray, Time64MicrosecondArray, Time64NanosecondArray, TimestampMicrosecondArray,
     TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray, UInt8Array,
     UInt16Array, UInt32Array, UInt64Array,
 };
+use arrow::array::{BinaryArray, BinaryViewArray};
 use arrow::datatypes::{DataType, Schema, TimeUnit};
 use async_trait::async_trait;
 use mssql_tds::connection::bulk_copy::{BulkLoadRow, ResolvedColumnMapping};
@@ -79,8 +79,11 @@ pub enum ColumnPlanKind {
     Utf8VarChar,
     LargeUtf8Nvarchar,
     LargeUtf8VarChar,
+    Utf8ViewNvarchar,
+    Utf8ViewVarChar,
     Binary,
     LargeBinary,
+    BinaryView,
     /// Arrow `date32` → SQL `date`.
     Date32,
     /// Arrow `date64` (ms) → SQL `date`.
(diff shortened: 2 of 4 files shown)


mssql-py-core/src/arrow_bulkcopy.rs
```diff
@@ -412,20 +422,13 @@ impl ColumnPlan {
...
-                let s = downcast::<LargeStringArray>(arr)?.value(row_idx).to_owned();
+            ColumnPlanKind::Utf8Nvarchar
+            | ColumnPlanKind::Utf8VarChar
+            | ColumnPlanKind::LargeUtf8Nvarchar
+            | ColumnPlanKind::LargeUtf8VarChar
+            | ColumnPlanKind::Utf8ViewNvarchar
+            | ColumnPlanKind::Utf8ViewVarChar => {
+                let s = read_utf8(arr, row_idx)?.to_owned();
                 Ok(ColumnValues::String(SqlString::from_utf8_string(s)))
             }
```
David-Engel: **Suggestion** — collapsing the four existing string arms into `read_utf8` discards type information the plan already has. `Utf8Nvarchar` previously guaranteed a `StringArray`; now any of the three string array types is silently accepted for any of the six kinds, so a plan/array mismatch produces wrong data instead of an error. It also makes the new Polars path pay three `downcast_ref` probes per cell in the bulk-copy inner loop, since `StringViewArray` is checked last. The per-cell cost is dwarfed by the `to_owned()` allocation, so this is about specificity more than speed:

```rust
ColumnPlanKind::Utf8Nvarchar | ColumnPlanKind::Utf8VarChar => {
    let s = downcast::<StringArray>(arr)?.value(row_idx).to_owned();
    Ok(ColumnValues::String(SqlString::from_utf8_string(s)))
}
ColumnPlanKind::LargeUtf8Nvarchar | ColumnPlanKind::LargeUtf8VarChar => {
    let s = downcast::<LargeStringArray>(arr)?.value(row_idx).to_owned();
    Ok(ColumnValues::String(SqlString::from_utf8_string(s)))
}
ColumnPlanKind::Utf8ViewNvarchar | ColumnPlanKind::Utf8ViewVarChar => {
    let s = downcast::<StringViewArray>(arr)?.value(row_idx).to_owned();
    Ok(ColumnValues::String(SqlString::from_utf8_string(s)))
}
```

Extending `read_utf8` itself is still needed and correct — `Xml` / `Json` / `Utf8Uuid` don't record the source array type in their plan kind.

mssql-py-core/src/arrow_bulkcopy.rs
```diff
@@ -1165,6 +1174,51 @@ mod tests {
...
+            .unwrap();
+        assert!(matches!(guid_value, ColumnValues::Uuid(_)));
+
+        for (sql_type, text) in [(SqlDbType::Xml, "<r/>"), (SqlDbType::Json, "{\"a\":1}")] {
+            let array: ArrayRef = Arc::new(StringViewArray::from(vec![Some(text)]));
+            let dest = meta("document", sql_type, true);
+            one_col_plan(DataType::Utf8View, &dest)
+                .extract_value(array.as_ref(), 0, &dest)
+                .unwrap();
+        }
```
David-Engel, line 1233: **Suggestion** — this loop unwraps and drops the result, so it only asserts "didn't error"; it would still pass if the value came back as the wrong `ColumnValues` variant. The GUID case just above gets this right with `assert!(matches!(...))`. Suggest:

```rust
match one_col_plan(DataType::Utf8View, &dest)
    .extract_value(array.as_ref(), 0, &dest)
    .unwrap()
{
    ColumnValues::Xml(_) | ColumnValues::Json(_) => {}
    other => panic!("expected Xml/Json, got {other:?}"),
}
```

mssql-py-core/src/arrow_bulkcopy.rs
```diff
@@ -11,15 +11,15 @@
...
     Array, BooleanArray, Date32Array, Date64Array, Decimal128Array, FixedSizeBinaryArray,
     Float32Array, Float64Array, Int8Array, Int16Array, Int32Array, Int64Array, LargeBinaryArray,
-    LargeStringArray, RecordBatch, StringArray, Time32MillisecondArray, Time32SecondArray,
-    Time64MicrosecondArray, Time64NanosecondArray, TimestampMicrosecondArray,
+    LargeStringArray, RecordBatch, StringArray, StringViewArray, Time32MillisecondArray,
+    Time32SecondArray, Time64MicrosecondArray, Time64NanosecondArray, TimestampMicrosecondArray,
     TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray, UInt8Array,
     UInt16Array, UInt32Array, UInt64Array,
 };
+use arrow::array::{BinaryArray, BinaryViewArray};
```
David-Engel: **Nit** — this is a second `use` block from the same module as the group directly above. Fold `BinaryArray` and `BinaryViewArray` into that alphabetized list instead of keeping a separate import.

mssql-py-core/src/arrow_bulkcopy.rs
```diff
@@ -662,12 +669,14 @@ fn timestamp_to_1900_components(ticks_since_epoch: i64) -> (i64, u8, u8, u8, u32
...
+/// Read a UTF-8 string cell from any supported variable-width string array.
 fn read_utf8(arr: &dyn Array, row_idx: usize) -> TdsResult<&str> {
     if let Some(a) = arr.as_any().downcast_ref::<StringArray>() {
         Ok(a.value(row_idx))
     } else if let Some(a) = arr.as_any().downcast_ref::<LargeStringArray>() {
         Ok(a.value(row_idx))
+    } else if let Some(a) = arr.as_any().downcast_ref::<StringViewArray>() {
+        Ok(a.value(row_idx))
     } else {
         Err(downcast_err::<StringArray>(arr))
```
David-Engel: **Nit** — now that three array types are accepted, this error still reports `expected ...StringArray`, which will misdirect anyone who hits it. Consider a message naming the set, e.g. `"expected a UTF-8 string array (utf8/large_utf8/utf8_view)"`.

CHANGELOG.md
```diff
@@ -62,6 +62,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ### Fixed
 
+- `mssql-py-core`: Arrow bulk copy now accepts `Utf8View` and `BinaryView`
```
David-Engel: **Nit** — this adds new source-type support, so `### Added` reads more accurately than `### Fixed`. Debatable, since it does fix a Polars load failure — take it or leave it.

David-Engel (commented):
## Summary

Adds Arrow `Utf8View`/`BinaryView` support to `mssql-py-core`'s Arrow bulk copy so Polars-exported C-stream columns load without a PyArrow re-encode. The change is small, correct, and well covered. **No blocking findings** — a couple of suggestions and nits inline.

### Verified locally (worktree at `496d0c8e`)

- `cargo clippy --all-features --all-targets -- -D warnings` (mssql-py-core) — clean
- `cargo nextest run --lib arrow_bulkcopy` — 53 passed
- CI build 166711 Linux job: both new integration tests **PASSED** (`test_cursor_bulkcopy_arrow_varbinary_binary_view`, `test_cursor_bulkcopy_arrow_varchar_string_view`) — confirmed not skipped
- `arrow` 55.2 with `features = ["ffi"]` does support view types over the C data interface, and CI installs unpinned `pyarrow`, so `pa.string_view()` / `pa.binary_view()` are available

### Correctness

The new `resolve_kind` arms cover NVarChar/NChar/NText, VarChar/Char/Text, VarBinary/Binary/Image, plus Xml/Json/UniqueIdentifier for `Utf8View` — mirroring the existing `Utf8`/`LargeUtf8`/`Binary` arms with no gaps I could find. NULL handling sits upstream of the `match` in `extract_value`, so view arrays get it for free, and the tests exercise both the inline (<=12 byte) and out-of-line buffer paths on each new array type.

### CI

Both red checks trace to one infra flake, not this change:

- **Build Stage / Build Linux ARM** failed at `Wait for cross-pool SQL host endpoint` — *"No SQL host endpoint after 10 min on stage attempt 3."*
- **coverage-report** then timed out after 75 minutes waiting for the combined coverage artifact that the failed stage never produced.

The pipeline's own message applies here: use **Rerun stage**, not *Rerun failed jobs* — the latter won't restart the SQL host job and will time out again.

mssql-py-core/src/arrow_bulkcopy.rs
gargsaumya: Fixed

mssql-py-core/src/arrow_bulkcopy.rs
gargsaumya: Done

mssql-py-core/src/arrow_bulkcopy.rs
gargsaumya: Done

mssql-py-core/src/arrow_bulkcopy.rs
gargsaumya: Done

mssql-py-core/src/arrow_bulkcopy.rs
```diff
@@ -1165,6 +1179,55 @@ mod tests {
...
+        for (sql_type, text) in [(SqlDbType::Xml, "<r/>"), (SqlDbType::Json, "{\"a\":1}")] {
+            let array: ArrayRef = Arc::new(StringViewArray::from(vec![Some(text)]));
+            let dest = meta("document", sql_type, true);
+            match one_col_plan(DataType::Utf8View, &dest)
+                .extract_value(array.as_ref(), 0, &dest)
+                .unwrap()
+            {
+                ColumnValues::Xml(_) | ColumnValues::Json(_) => {}
+                other => panic!("expected Xml/Json, got {other:?}"),
+            }
```
saurabh500: This accepts either `Xml` or `Json` for both iterations, so the test still passes if the planner swaps those mappings, and it never verifies the payload. Please match the destination and returned variant together and assert the content:

```suggestion
            let value = one_col_plan(DataType::Utf8View, &dest)
                .extract_value(array.as_ref(), 0, &dest)
                .unwrap();
            match (sql_type, value) {
                (SqlDbType::Xml, ColumnValues::Xml(value)) => {
                    assert_eq!(value.as_string(), text)
                }
                (SqlDbType::Json, ColumnValues::Json(value)) => {
                    assert_eq!(value.bytes, text.as_bytes())
                }
                (_, other) => panic!("expected {sql_type:?}, got {other:?}"),
            }
```

saurabh500 (approved):
Adds native Arrow `Utf8View` and `BinaryView` support to Python bulk copy so Polars C-stream inputs no longer need a PyArrow table conversion. The implementation is correct and well scoped: destination mappings mirror the existing UTF-8 and binary paths, null handling is preserved, and tests cover inline and external view storage.

**Verdict: Approve with comments.** No blocking findings. One non-blocking unit-test assertion should be tightened; see the inline suggestion.

saurabh500 (commented):
One minor test change, ready to approve after that is incorporated

saurabh500:
@gargsaumya can we drive this to completion please

saurabh500 (approved):
## Re-review — `Tighten Utf8View XML/JSON test assertions` (18855fdf)

Re-checked the single 🔵 follow-up I left on my approval. Addressed verbatim: the `Utf8View` test now binds the extracted value and matches `(sql_type, value)` together — asserting `Xml → as_string() == text` and `Json → bytes == text.as_bytes()`, with a `(_, other) => panic!` catch-all. It no longer accepts either variant for either iteration and it checks the payload rather than just the discriminant, so it would fail if the planner swapped the `Utf8View → Xml`/`Json` mappings. CI is green on the head commit.

No further comments. **Verdict: Approve** — the substance was already approved; this closes the only note I left.

PR #270, opened 2026-08-13
Fix exhausted command-timeout budget silently widening to infinite

Theekshna:
## Description

Fixes a latent command-timeout bug: `deduct_timeout` returned `Some(0)` to mean **"budget exhausted"** (a reconnect consumed the whole command timeout), but `timeout_to_duration(Some(0))` normalized that to `None` — **"no timeout"** by the bulk-copy convention. So an RPC whose entire budget was eaten by transparent reconnect ran **unbounded** instead of failing fast. The two `Some(0)` conventions collided on the same value.

### Change

- Replace the ambiguous `Option<u32>` result of `deduct_timeout` with a `CommandTimeoutBudget` enum: `None` (infinite), `Remaining(NonZeroU32)`, `Exhausted`.
- `into_timeout()` converts `Exhausted` into a hard `TimeoutError` **before** any request is serialized; caller-supplied `Some(0)`/`None` remain a genuine infinite budget — the two are now distinct, uncollapsible states.
- Delete `timeout_to_duration` and the two scattered manual `if opts.timeout == Some(0)` guards (in `execute_prepared` / `unprepare`); the check is now applied uniformly at every RPC entry point (batch, `sp_*`, cursor ops).

### Notes

- Builds on the prepared-statement recovery work from #148 (merged as `30e4d2d5`), which introduced `execute_prepared` / `unprepare`.
- Latent today because the ODBC path passes no query timeout (`SQL_ATTR_QUERY_TIMEOUT` not yet wired); this closes the gap before it becomes reachable.

### Tests

- New `deduct_timeout` unit tests: `Remaining` subtraction, over-consumption `Exhausted`, and the **exact `timeout - elapsed == 0` boundary** (`exact_budget_boundary_is_exhausted_not_infinite`) that must not widen to infinite.
- `caller_zero_timeout_becomes_explicit_infinite_budget` pins the caller-`Some(0)` = infinite convention.
- Removed the obsolete `timeout_to_duration` tests and the two tests that faked exhaustion via caller `Some(0)`.

## Related Issues

[AB#47211](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47211)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes — 2165 lib tests (`mssql-tds` + `mssql-odbc`)
- [x] Public API changes are documented — internal `pub(in crate::connection)` only; no public surface change

### DIFF
--- a/mssql-tds/src/connection/cursor_ops.rs
+++ b/mssql-tds/src/connection/cursor_ops.rs
@@ -279,9 +279,10 @@ impl CursorClient for TdsClient {
         }
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
 
-        self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec);
+        self.remaining_request_timeout = request_timeout;
         self.cancel_handle = cancel_handle.map(|handle| handle.child_handle());
         self.return_values.clear();
         self.transport.reset_reader();
@@ -350,9 +351,10 @@ impl CursorClient for TdsClient {
         }
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
 
-        self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec);
+        self.remaining_request_timeout = request_timeout;
         self.cancel_handle = cancel_handle.map(|handle| handle.child_handle());
         self.return_values.clear();
         self.transport.reset_reader();
@@ -430,9 +432,10 @@ impl CursorClient for TdsClient {
         }
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
 
-        self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec);
+        self.remaining_request_timeout = request_timeout;
         self.cancel_handle = cancel_handle.map(|handle| handle.child_handle());
         self.return_values.clear();
         self.transport.reset_reader();
@@ -523,9 +526,10 @@ impl CursorClient for TdsClient {
         }
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
 
-        self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec);
+        self.remaining_request_timeout = request_timeout;
         self.cancel_handle = cancel_handle.map(|handle| handle.child_handle());
         self.return_values.clear();
         self.transport.reset_reader();
@@ -584,9 +588,10 @@ impl CursorClient for TdsClient {
         }
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
 
-        self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec);
+        self.remaining_request_timeout = request_timeout;
         self.cancel_handle = cancel_handle.map(|handle| handle.child_handle());
         self.return_values.clear();
         self.transport.reset_reader();
@@ -674,9 +679,10 @@ impl CursorClient for TdsClient {
         };
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
 
-        self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec);
+        self.remaining_request_timeout = request_timeout;
         self.cancel_handle = cancel_handle.map(|handle| handle.child_handle());
         self.return_values.clear();
         self.transport.reset_reader();
@@ -732,9 +738,10 @@ impl CursorClient for TdsClient {
         }
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
 
-        self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec);
+        self.remaining_request_timeout = request_timeout;
         self.cancel_handle = cancel_handle.map(|handle| handle.child_handle());
         self.return_values.clear();
         self.transport.reset_reader();
@@ -838,9 +845,10 @@ impl CursorClient for TdsClient {
         }
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
 
-        self.remaining_request_timeout = Self::timeout_to_duration(timeout_sec);
+        self.remaining_request_timeout = request_timeout;
         self.cancel_handle = cancel_handle.map(|handle| handle.child_handle());
         self.return_values.clear();
         self.transport.reset_reader();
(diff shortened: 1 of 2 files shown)


mssql-tds/src/connection/tds_client.rs
```diff
@@ -55,6 +56,30 @@ use std::time::{Duration, Instant};
...
+    pub(in crate::connection) fn into_timeout(self) -> TdsResult<(Option<u32>, Option<Duration>)> {
+        match self {
+            Self::None => Ok((None, None)),
+            Self::Remaining(seconds) => {
+                let seconds = seconds.get();
+                Ok((Some(seconds), Some(Duration::from_secs(u64::from(seconds)))))
+            }
+            Self::Exhausted => Err(crate::error::Error::TimeoutError(
+                crate::error::TimeoutErrorType::String(
+                    "command timeout exhausted by connection recovery".to_string(),
```
David-Engel: There's no reason the error this method returns has to be linked to "connection recovery". That would just limit where it could be used.
```suggestion
                    "command timeout exhausted".to_string(),
```

saurabh500 (commented):
The three-state `CommandTimeoutBudget` model fixes the exhausted-budget/infinite-timeout collision cleanly, and I did not find a blocking correctness issue. I have two suggestions:

1. Please add a call-site regression test in addition to the helper tests around `tds_client.rs:5972-5990`. Those tests would still pass if an `into_timeout()?` check were accidentally removed from an RPC entry point. Ideally, simulate a reconnect consuming the command budget, assert `TimeoutError`, and verify that no request is serialized or executed.

2. Please document the zero-timeout behavior on `ExecuteOptions::timeout` (`tds_client.rs:4539`) and `ExecuteOptions::timeout_secs` (`tds_client.rs:4558`). The implementation deliberately treats both `None` and `Some(0)` as unlimited, but the public docs currently mention only `None`. This PR also changes `execute_prepared`/`unprepare` with `timeout_secs(0)` from an error to unlimited behavior, so that semantic change should be explicit.

Issue #273 should remain open: this PR does not address the sub-second `BulkCopy::timeout` truncation or the reconnect timeout-cap truncation.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -5975,32 +5958,61 @@ mod tests {
...
+        );
+        assert_eq!(
+            result.into_timeout().unwrap(),
+            (Some(18), Some(Duration::from_secs(18)))
+        );
     }
 
     #[test]
-    fn deduct_timeout_saturates_at_zero() {
+    fn fully_consumed_positive_budget_errors_before_timeout_conversion() {
```
saurabh500, line 6037: Could we add a call-site regression test like this? I ran this body against the PR branch locally; it passes and confirms that no request bytes are serialized.

A small `#[cfg(test)]` override keeps the test deterministic and in the regular unit suite (no live SQL Server or reconnect timing):

```rust
// TdsClient field, initialized to None in TdsClient::new
#[cfg(test)]
reconnect_elapsed_for_test: Option<Duration>,

// At the start of check_and_reconnect
#[cfg(test)]
if let Some(elapsed) = self.reconnect_elapsed_for_test.take() {
    return Ok(elapsed);
}
```

Then the test body can exercise the real `execute` entry point and inspect the existing byte-recording `TestTransport`:

```rust
#[tokio::test]
async fn execute_exhausted_reconnect_budget_writes_no_request() {
    let transport = TestTransport::new();
    let sent = Arc::clone(&transport.sent);
    let mut client = create_test_client_with_transport(transport);
    client.reconnect_elapsed_for_test = Some(Duration::from_secs(1));

    let error = match client
        .execute(
            "SELECT 1".to_string(),
            ExecuteOptions::new().timeout_secs(1),
        )
        .await
    {
        Ok(_) => panic!("execute should fail before serialization"),
        Err(error) => error,
    };

    assert!(matches!(
        &error,
        crate::error::Error::TimeoutError(
            crate::error::TimeoutErrorType::String(message)
        ) if message == "command timeout exhausted by connection recovery"
    ));
    assert!(sent.lock().unwrap().is_empty());
}
```

This test would fail if this execution path lost its `into_timeout()?` guard, while the helper-only tests here would continue passing.

saurabh500 (commented):
Requested call-site regression coverage for the core execution and cursor RPC paths.

mssql-tds/src/connection/cursor_ops.rs
```diff
@@ -279,9 +279,10 @@ impl CursorClient for TdsClient {
         }
 
         let reconnect_elapsed = self.check_and_reconnect(timeout_sec, cancel_handle).await?;
-        let timeout_sec = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let budget = Self::deduct_timeout(timeout_sec, reconnect_elapsed);
+        let (timeout_sec, request_timeout) = budget.into_timeout()?;
```
saurabh500: Using the same test-only reconnect-elapsed override suggested on the timeout tests, could we add one representative cursor call-site test? I ran this exact body locally against the PR branch; it passes and verifies that `cursor_open` writes no bytes after recovery exhausts the budget:

```rust
#[tokio::test]
async fn cursor_open_exhausted_reconnect_budget_writes_no_request() {
    let transport = TestTransport::new();
    let sent = Arc::clone(&transport.sent);
    let mut client = create_test_client_with_transport(transport);
    client.reconnect_elapsed_for_test = Some(Duration::from_secs(1));

    let error = match client
        .cursor_open(
            "SELECT 1",
            crate::cursor::CursorScrollOption::FORWARD_ONLY,
            crate::cursor::CursorConcurrency::READONLY,
            0,
            Some(1),
            None,
        )
        .await
    {
        Ok(_) => panic!("cursor_open should fail before serialization"),
        Err(error) => error,
    };

    assert!(matches!(
        &error,
        crate::error::Error::TimeoutError(
            crate::error::TimeoutErrorType::String(message)
        ) if message == "command timeout exhausted by connection recovery"
    ));
    assert!(sent.lock().unwrap().is_empty());
}
```

One cursor test should be enough to cover this RPC family; it avoids duplicating the same test for all ten cursor operations.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -55,6 +56,30 @@ use std::time::{Duration, Instant};
...
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(in crate::connection) enum CommandTimeoutBudget {
+    None,
+    Remaining(NonZeroU32),
+    Exhausted,
+}
+
+impl CommandTimeoutBudget {
+    pub(in crate::connection) fn into_timeout(self) -> TdsResult<(Option<u32>, Option<Duration>)> {
```
saurabh500: Could we avoid the positional tuple here? The two `Option`s have different roles, but call sites have to remember which is which. A named, stack-only resolved type would make the API self-documenting without adding a heap allocation. Storing `Option<NonZeroU32>` also keeps a single source of truth and is niche-optimized to the size of a `u32`.

For example:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub(in crate::connection) struct ResolvedCommandTimeout {
    seconds: Option<NonZeroU32>,
}

impl ResolvedCommandTimeout {
    pub(in crate::connection) fn seconds(self) -> Option<u32> {
        self.seconds.map(NonZeroU32::get)
    }

    pub(in crate::connection) fn duration(self) -> Option<Duration> {
        self.seconds
            .map(|seconds| Duration::from_secs(u64::from(seconds.get())))
    }
}

impl CommandTimeoutBudget {
    /// Resolves the post-recovery budget into the timeout used by request I/O.
    ///
    /// Infinite budgets produce no deadline, while remaining budgets preserve
    /// their positive-seconds invariant. Exhausted budgets are rejected before
    /// request serialization so they cannot enter the `0 = infinite` path.
    ///
    /// # Errors
    ///
    /// Returns `TimeoutError` when connection recovery consumed the complete
    /// command-timeout budget.
    pub(in crate::connection) fn into_timeout(
        self,
    ) -> TdsResult<ResolvedCommandTimeout> {
        match self {
            Self::None => Ok(ResolvedCommandTimeout { seconds: None }),
            Self::Remaining(seconds) => {
                Ok(ResolvedCommandTimeout { seconds: Some(seconds) })
            }
            Self::Exhausted => Err(crate::error::Error::TimeoutError(
                crate::error::TimeoutErrorType::String(
                    "command timeout exhausted by connection recovery".to_string(),
                ),
            )),
        }
    }
}
```

Call sites then read by meaning rather than tuple position:

```rust
let timeout = budget.into_timeout()?;
self.remaining_request_timeout = timeout.duration();
let mut packet_writer = batch.create_packet_writer(
    self.transport.as_writer(),
    timeout.seconds(),
    cancel,
);
```

The wrapper itself is allocation-free; this leaves the existing error-string allocation unchanged.

saurabh500 (commented):
Readability suggestion for the resolved timeout representation.

saurabh500:
No blocking comments. Good improvement. This change is likely needed in other places too, e.g. BulkCopy. 

After addressing comments, ready to approve.

mssql-tds/src/connection/tds_client.rs
Theekshna: Done — shortened to `"command timeout exhausted"` and updated both call-site test assertions to match. (d2f7a882)

mssql-tds/src/connection/tds_client.rs
Theekshna: Done — added `execute_exhausted_reconnect_budget_writes_no_request` using a `#[cfg(test)] reconnect_elapsed_for_test` override; asserts `TimeoutError` and that no bytes were serialized.

mssql-tds/src/connection/cursor_ops.rs
Theekshna: Done — added `cursor_open_exhausted_reconnect_budget_writes_no_request` (one representative cursor site, as suggested).

mssql-tds/src/connection/tds_client.rs
Theekshna: Done — `into_timeout` now returns a named `ResolvedCommandTimeout { Option<NonZeroU32> }` with `seconds()`/`duration()`; all 22 call sites read by name instead of tuple position.

Theekshna:
Addressed the review in d2f7a882:

- **Named resolved type / call-site tests / error string** — done (see inline replies).
- **Docs (zero-timeout):** documented on `ExecuteOptions::timeout` and `timeout_secs` that both `None` and `Some(0)` mean unlimited, and called out that `timeout_secs(0)` on the managed `execute_prepared`/`unprepare` paths now runs unbounded instead of erroring.
- **#273** left open — sub-second `BulkCopy` and reconnect-cap truncation are out of scope here.

`cargo bfmt`/`bclippy` clean; 2152 lib tests pass.

saurabh500 (approved):
Re-review of d2f7a88 — thanks for the quick turnaround. All three items from the last pass are addressed cleanly:

- `into_timeout` now returns a named `ResolvedCommandTimeout` (niche-optimized `Option<NonZeroU32>`, allocation-free), and every call site reads via `.seconds()` / `.duration()` instead of a positional tuple. Single source of truth — seconds and the derived deadline can no longer disagree.
- `execute_exhausted_reconnect_budget_writes_no_request` drives the real `execute` entry point through the `#[cfg(test)] reconnect_elapsed_for_test` override and asserts no bytes reach the wire. It would fail if the `into_timeout()?` guard were dropped.
- `cursor_open_exhausted_reconnect_budget_writes_no_request` covers the cursor RPC family the same way.

Issue #265, opened 2026-08-13
Remove `Box<dyn TdsTransport>` from the row path to unlock writer devirtualization


saurabh500:
### Problem statement

#257 made the row-decode chain generic over the `RowWriter` (`W: RowWriter + Send + ?Sized`
through `receive_row_into_internal` → `drive_row_columns` → `decode_or_decrypt_column` →
`GenericDecoder::decode_into`). Measurement showed that change delivers **~0% in
production**, and the reason is structural rather than incidental: the writer's concrete
type is destroyed before it ever reaches that chain.

```
mssql-js      BinaryRowWriter (concrete)  ─┐
mssql-py-core PyRowWriter     (concrete)  ─┤
                                           v
ResultSet::next_row_into(&mut (dyn RowWriter + Send))     tds_client.rs
  -> TdsClient::get_next_row_into(&mut (dyn RowWriter))
  -> self.transport.receive_row_into(..., writer)
     ^^^^^^^^^^^^^^  transport: Box<dyn TdsTransport>     tds_client.rs:139   <== ERASURE
  -> NetworkTransport::receive_row_into(&mut (dyn ...))
  -> receive_row_into_internal(...)                       token_stream.rs
```

Both FFI consumers already hold a *concrete* writer and only coerce to `dyn` because the
signature demands it. The blocker is one level lower:

```rust
pub struct TdsClient { pub(crate) transport: Box<dyn TdsTransport>, ... }   // :139
pub(crate) trait TdsTransport: TdsTokenStreamReader + Send + Sync + Debug   // :20
```

`TdsTokenStreamReader` is used as a trait object transitively, as a supertrait of the boxed
`TdsTransport`. (A grep for `dyn TdsTokenStreamReader` finds nothing, which is why this is
easy to miss — the boxing is spelled `Box<dyn TdsTransport>`.) Below that boundary `W` can
only ever instantiate as `dyn RowWriter + Send`.

Worth stating plainly: **in a release build the box has exactly one possible concrete
type.** `NetworkTransport` is the only non-test implementor — `MockTransport` is
`#[cfg(fuzzing)]`, `TokenReplayTransport` and `ByteStreamTransport` are test-only, and
`TestTransport` lives inside a private `#[cfg(test)] mod tests`. The erasure exists for
testability, not for production polymorphism. `create_transport` even returns a concrete
`Box<NetworkTransport>` already (`network_transport.rs:395`); the single erasing cast is at
[...]

PR #275, opened 2026-08-13
Arm the per-row request timeout only when the decode suspends

### LINKED ISSUE #271: Arm the per-row request timeout lazily: ~3% on `main`, ~15% after #252
> **Read this before comparing to #265.** The two headline numbers belong to *different
> consumers* and never sum on one path. #265's measured 20.0% is on the contiguous-writer
> shape `mssql-js` uses; this issue's ~42% is paid only by `mssql-py-core`, which is the one
> caller that sets a request timeout. Triaging by number size alone will stack them and be
> wrong.

### Problem statement

`NetworkTransport::receive_row_into` wraps every single row decode in two combinators
(`network_transport.rs:1388-1398`):

```rust
let cancellable = CancelHandle::run_until_cancelled(
    cancel_handle,
    receive_row_into_internal(self, &*PARSER_REGISTRY, context, plan, writer),
);
let result = match remaining_request_timeout.as_ref() {
    Some(t) => match timeout(*t, cancellable).await { ... },
    None => cancellable.await,
};
```

Measured on the in-crate row-decode harness (`decode_bench.rs`, 20,000-row passes,
3 warmup + 9 measured → median, `--release`), 48-column row (39 INT + 9 VARCHAR(6)),
`ContiguousRowWriter`. All four modes share one runtime and one warmup regime, so the
comparison is paired:

| Mode | median ms | ns/row | vs bare |
|---|---:|---:|---:|
| bare (`receive_row_into_internal` direct) | 32.811 | 1640.5 | — |
| `run_until_cancelled(None, ..)` | 34.709 | 1735.4 | +5.8% |
| `run_until_cancelled(Some(&h), ..)` | 34.415 | 1720.7 | +4.9% |
| **`+ tokio::time::timeout(..)`** | **46.751** | **2337.6** | **+42.5%** |

**Arming a `tokio::time::timeout` per row costs ~36% on top of the cancellation wrapper,
and ~42% over an unwrapped decode.** That is a timer registration and deregistration for
every row in a result set. The cancellation wrapper itself is comparatively cheap (~5%,
close enough to the noise floor to be worth treating as "small" rather than as a precise
figure).

Note `run_until_cancelled(None, ..)` costs essentially the same as `Some(..)` — the cost is
the wrapper existing, not a handle being supplied. Connections with no cancel handle still
pay it.

### The byte-size picture, and why it is *not* the headline

This was found via `cargo +nightly rustc -- -Zprint-type-sizes`, which paints a different
picture from the timings — worth recording precisely so nobody optimizes the wrong thing:

| Layer | bytes | Δ |
|---|---:|---:|
| `receive_row_into_internal` (the actual decode) | 1328 | — |
| + `tokio_util` `CancellationToken::run_until_cancelled` | 2744 | +1416 |
| + `core::CancelHandle::run_until_cancelled` | 4088 | +1344 |
| + `tokio::time::Timeout` | 4200 | +112 |
| + row-fetch wrapper | 4416 | +216 |

By *size*, cancellation is ~70% of the future and the timeout is nearly free (+112 B). By
*time*, that is exactly inverted: cancellation is ~5% and the timeout is ~36%. **Size and
cost are anti-correlated here**, so the byte table should not be used to prioritize.

The size growth does have a real cause worth fixing on its own terms. In `core.rs:50-56`,
`f` is awaited in *both* match arms:

```rust
Some(handle) => handle.cancel_token.run_until_cancelled(f).await,  // stores f
None => f.await,                                                    // stores f again
```

These are separate suspend points in separate arms, and rustc stores both rather than
overlapping them: 2744 + 1328 = 4072 against a measured 4088. By contrast the
`timeout`/`cancellable` pair one level up *does* overlap correctly (4088 → 4200), because
`timeout()` is a plain struct constructor rather than an `async fn`, so the move is visible
to layout.

Why the size matters despite not being the hot cost: at ~1.4 KB the row future fits under
the repo's own 4096 B `row_fetch_futures_stay_small` guard (`tds_client.rs:4418`)
*unboxed*, which would let `#[async_trait]`'s per-row `Box::pin` be removed outright. At
4416 B it cannot. See #265 for the measurements showing all four row futures blowing that
budget when the box is removed naively.

### Proposed solution

Roughly in order of value:

1. **Stop arming a timer per row.** The request timeout is a property of the request, not
   of each row. Compute a deadline once and either check it between rows, or hold a single
   `sleep_until` for the whole result set, or scope the timeout to the network read rather
   than the decode.
2. **Make `CancelHandle::run_until_cancelled` await `f` exactly once**, so the two arms
   share a slot instead of both being stored — e.g. select over a cancellation branch that
   is `pending()` when no handle is supplied. This is the ~2.7 KB, not the ~36%.
3. **Check cancellation between rows rather than inside them.** A row decode is
   microseconds; `is_cancelled()` between rows costs zero future bytes and no timer.

(1) and (3) together would remove both wrappers from the per-row path entirely.

### Measured: conditional arming works, and the wrapper form matters

Spiked on `dev/saurabh/row-decode-perf-feasibility` @ `3371dc58`. Two new ladder rungs
arm the timer only if the decode does not complete on its first poll — `Timeout::poll`
polls the inner future before the delay, so a row served from a buffered packet never
observes the timer either way, and skipping the arm is observationally identical.

All six modes in one binary, one runtime, one warmup regime; 9 outer runs, `min` reported
(robust to upward interference on a shared machine):

| Mode | min ms | vs bare | vs `cancel_some` |
|---|---:|---:|---:|
| bare | 30.48 | — | −6.6% |
| `cancel_none` | 31.26 | +2.6% | −4.2% |
| `cancel_some` | 32.63 | +7.1% | — |
| **`+ timeout(..)`** (today) | **37.85** | **+24.2%** | +16.0% |
| `+ lazy timeout`, `async fn` helper | 34.20 | +12.2% | +4.8% |
| **`+ lazy timeout`, inline** | **32.72** | **+7.3%** | **+0.3%** |

The timer's own cost is the gap between `cancel_some` and `+ timeout` — **17.1 points**.
Inline lazy arming lands within **0.3%** of `cancel_some`, i.e. it removes **~99%** of that
cost. What remains at +7.1% is the cancellation wrapper, which is items (2)/(3) above and a
separate fix.

**The helper form leaves ~30% of the win on the table.** Writing it as
`async fn lazy_timeout(..)` costs **4.5%** over the inline form, reproducibly and paired.
Same cause this issue already documents for `run_until_cancelled`: an `async fn` with an
`await` in one match arm stores that arm's future in the state machine, so the per-row
future grows and is memcpy'd on every move. **Implement this inline at the call site, or as
a hand-written `Future` — not as an `async fn` wrapper.**

Worth noting how the helper cost was found, since it nearly wasn't: the first measurement
of the inline form and the later measurements of the helper form were taken in *different
builds*, which made a real 4.5% regression look like machine noise. Only re-measuring all
six modes in one paired ladder separated them.

#### The exhausted-budget trap is covered

Per the caveat above, conditional arming inverts semantics if it keys on the budget being
zero. It must key on *suspension*. `lazy_timeout_matches_eager_timeout_on_exhausted_budget`
asserts, for `Duration::ZERO`:

| inner future | `tokio::time::timeout` | lazy arming |
|---|---|---|
| ready on first poll | `Ok` | `Ok` |
| suspends | `Elapsed` | `Elapsed` |

The second row is the guard: skipping the timer because "the budget is gone" would turn a
fetch that must fail into an unbounded wait.

#### Caveat

The harness reads from memory, so the inner future never suspends and the lazy path never
arms. That is the intended isolation, and it is the common case (rows served from an
already-buffered packet), but a workload that genuinely suspends per row arms the timer as
before and pays today's cost plus one extra poll. The win is real and workload-shaped, not
universal.

### Affected crate

mssql-tds

### Alternatives considered

Doing nothing. This is pure per-row overhead on the hottest path in the driver, and it is
larger than the dispatch win being pursued in #247 / #265 — that work chases a measured
20.0% writer-devirtualization gap, while the timeout alone is ~42% on the same shape of
row.

Raising the 4096 B future-size guard instead of shrinking the future. Rejected: a ~5.5 KB
future is not free just because it is on the stack — it is memcpy'd on every move — so the
guard is doing real work and the right response is to shrink the payload.

### Additional context

Found while spiking #265. Independent of it: #265 removes the *indirect call* on the row
path, this removes *wrapper overhead*, and the two compose.

Reproduce with the `run_wrapper_ladder` case added to `mssql-tds/src/decode_bench.rs` on
`dev/saurabh/row-decode-perf-feasibility`:

```
cargo test --release -p mssql-tds --lib decode_bench::bench_row_decode -- --nocapture
```

Two caveats on the measurement, stated so they are not discovered later as surprises:

- The harness reads from an in-memory buffer, so the decode never suspends on I/O. This
  isolates timer arm/disarm overhead, which is the intent, and it is representative of the
  common case where rows are served from an already-buffered packet. A workload that
  genuinely suspends per row would amortize differently.
- The ~5% cancellation figure is near this harness's noise floor (the same workload
  measured through two different case functions differs by ~4%). Treat it as "small", not
  as a precise number. The ~42% timeout figure is far outside that band.

### Which consumers actually pay this

The timeout arm is only entered when `remaining_request_timeout` is `Some`, and
`TdsClient::timeout_to_duration` maps both `None` **and** `Some(0)` to `None`. So the cost
is opt-in, and today exactly one consumer opts in:

| Consumer | what it passes | pays the ~42%? |
|---|---|---|
| `mssql-py-core` | `ExecuteOptions { timeout: Some(30), .. }` — hardcoded, `cursor.rs:72` | **yes, always** |
| `mssql-js` | `()` → `ExecuteOptions::default()` → `timeout: None` | no |
| `mssql-odbc` | no timeout plumbing at all | no |
| `mssql-tds-cli` | no timeout plumbing at all | no |

This sharpens the issue rather than shrinking it. It is not "the row path costs 42%" — it is
**the Python path costs 42% and the others cost nothing**, from a single hardcoded literal
that no Python caller can override. Note the asymmetry with #265, whose measured 20.0% is on
the contiguous writer shape that `mssql-js` uses: the consumer that stands to gain most from
#265 is not the one paying here, and vice versa.

Two things follow. First, `cursor.rs:72` is worth questioning on its own merits — a
non-configurable 30 s query timeout will abort any legitimately long-running Python query,
independent of performance. Second, whoever benchmarks this should set the timeout
explicitly rather than relying on defaults, or they will measure the `None` path and
conclude there is nothing here.

### What the timeout currently *means*, and what a fix must preserve

Worth stating before anyone hoists the timer, because the current semantics are not the
obvious ones. `get_next_row_into` (`tds_client.rs:3115-3127`) runs:

```rust
loop {
    let start = Instant::now();
    let result = self.transport.receive_row_into(.., self.remaining_request_timeout, ..).await?;
    self.update_remaining_timeout(start);
```

So the per-row `timeout()` is **not** a fresh 30 s per row. It is a single decrementing
operation budget, re-armed each row with whatever remains, and `update_remaining_timeout`
charges only the time spent *inside* the transport await. Time the caller spends between
`next_row_into` calls is never charged.

That makes the naive fix — hoist to one wall-clock deadline computed at operation start —
**not** semantics-preserving. It would additionally charge consumer think-time, so a slow
consumer over a large result set could begin tripping a timeout it never used to. For the
one consumer that pays this today (Python, driving row-by-row through PyO3) that is exactly
the wrong population to silently start charging.

Two further details for whoever implements it:

- `update_remaining_timeout` saturates to `Duration::ZERO`, not `None` (`:531-538`), so an
  exhausted budget keeps the wrapper armed. It does **not** follow that such a row fails
  immediately: `Timeout::poll` (tokio 1.53.1, `time/timeout.rs:211-221`) polls the inner
  future *first* and returns `Ok` before the delay is ever polled. A row already served from
  a buffered packet therefore still succeeds on a zero budget; the `Elapsed` only fires on a
  fetch that actually suspends.

  **This is what makes conditional arming provably correct rather than merely safer.** The
  only case where the wrapper is observable is exactly the case in which it would still be
  armed. It is also a trap: `Some(ZERO)` looks like a degenerate state worth optimizing away
  ("budget is gone, skip the wrapper"), and doing so inverts the semantics on the suspending
  path — a fetch that should fail with `Elapsed` would instead wait unbounded. Any
  conditional-arming change needs a test asserting that an exhausted budget still times out a
  *suspending* fetch.

- It shares a budget with `deduct_timeout` (`:619`), which is easy to mis-scope in either
  direction. It is **not** a separate budget: at all 18 call sites (10 in `cursor_ops.rs`,
  8 in `tds_client.rs`) its result is converted by `timeout_to_duration` and assigned to the
  very same `remaining_request_timeout` field this issue is about. One budget, two units
  (`Option<u32>` seconds vs `Option<Duration>`), two moments — `deduct_timeout` computes the
  value after a reconnect, and the row loop decrements it from there.

  The two moments disagree about zero, which is the substantive part:

  | budget reaches zero | representation | timer armed? |
  |---|---|---|
  | at init, via `deduct_timeout` | `Some(0)` → `timeout_to_duration` → `None` | **no** — request unbounded |
  | per row, via `update_remaining_timeout` (`:531`) | `Some(Duration::ZERO)` | yes — fails fast |

  So an exhausted budget runs forever if it was exhausted by a reconnect, and fails fast if it
  was exhausted by row fetching. The round-up at `:623-631` turns any non-zero sub-second
  reconnect into a whole second deducted, so a 100 ms reconnect against a 1 s timeout already
  reaches the first case. Filed as #272 — not blocking the work here, but it is this same
  field arriving at zero by a different route, so a fix to either should leave the two
  agreeing.

**The constraint this implies is the useful part.** The measured cost is arming and disarming
a timer-wheel entry per row, not the budget bookkeeping — which is a cheap subtraction. So a
fix can keep the decrementing-budget semantics exactly and still remove the cost, by keeping
the `Instant` arithmetic and only arming a real timer when the fetch would actually suspend.
That is strictly better than hoisting, because it changes nothing observable.

saurabh500:
## Description

Rows decoded from an already-buffered TDS packet usually complete on their first poll. The row path still wrapped every decode in `tokio::time::timeout`, paying for a clock read, `Sleep` construction, and a larger wrapper future even though no wait occurred.

This PR polls the decode future once and creates the timeout only when that poll returns `Pending`. Tokio also polls the inner future before its delay, so the ready path keeps the same behavior while avoiding the unused timeout setup.

The lazy timeout is used by the four row-data operations in `NetworkTransport`:

- `receive_row_into`
- `receive_row_header`
- `resume_row_into`
- `read_active_plp_bytes`

Connection, cancellation-drain, and token-boundary timeouts stay eager because they are not paid per row or column. The PLP streaming site is changed for consistency with the adjacent row-resume path; it was not covered by the benchmark.

### Timeout behavior

The remaining request budget can be `Some(Duration::ZERO)`. The implementation therefore keys on whether the decode suspends, not whether the budget is zero:

| first poll | result |
|---|---|
| `Ready` | succeeds, matching Tokio |
| `Pending` | arms the timeout and returns `Elapsed` when the exhausted budget wins |

Tests compare these cases directly with `tokio::time::timeout`, including a future that suspends once and then becomes ready.

The implementation is an inline macro because the equivalent `async fn` helper measured about 4% slower by growing the per-row state machine.

### Performance

[PR #264](https://github.com/microsoft/mssql-rs/pull/264) is now in `main` and removes the dominant boxed packet-reader overhead from row decoding. On that optimized path, a paired ABBA benchmark measured this change at about **254 ns/row**, or **15% of isolated decode time** for a 48-column row; the eager arm was slower in 202 of 205 pairs. See the [measurement on #271](https://github.com/microsoft/mssql-rs/issues/271#issuecomment-5288220545).

This isolates row-decode CPU cost. End-to-end impact will be smaller when network, TLS, or server time dominates.

The future-size guard still passes and now also covers `read_active_plp_chunk`.

Cancellation handling is unchanged: each operation still uses `CancelHandle::run_until_cancelled`.

## Related Issues

Fixes #271

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


mssql-tds/src/connection/transport/request_timeout.rs
```diff
@@ -0,0 +1,179 @@
...
+        let mut fut = ::std::pin::pin!($fut);
+        match $budget {
+            Some(budget) => {
+                let first = ::std::future::poll_fn(|cx| {
+                    ::std::task::Poll::Ready(::std::future::Future::poll(fut.as_mut(), cx))
+                })
+                .await;
+                match first {
+                    ::std::task::Poll::Ready(result) => result,
+                    ::std::task::Poll::Pending => match ::tokio::time::timeout(budget, fut).await {
```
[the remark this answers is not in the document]
saurabh500: Good catch on the extra poll, but I tested this and the divergence doesn't happen — the two forms agree.

I wrote exactly the test you describe: a future that returns `Pending` on its first poll (with a self-wake) and `Ok` on its second, under `Duration::ZERO`. Asserting `Err(Elapsed)` for the eager baseline **failed** — `tokio::time::timeout(Duration::ZERO, PendingThenReady)` returns `Ok`.

Reason: `sleep(ZERO)` is not ready on its first poll. It registers a timer entry and returns `Pending`, so `Timeout`'s first poll suspends too. Both forms then get woken by the future's own waker and both complete with `Ok` on the re-poll. The lazy probe's extra poll happens before any of that, so it doesn't shift the outcome.

That's also why the existing `NeverReady` test still yields `Elapsed`: it never self-wakes, so only the timer wakes the task, and by then the delay has expired.

I've added `suspend_then_complete_on_a_spent_budget_matches_eager_timeout` as a regression test that runs both forms over the same future shape and asserts they agree, so this stays pinned rather than resting on my reading of tokio's internals. I also extended the doc comment to record the `sleep(ZERO)`-is-`Pending` detail, since that's the non-obvious part that makes the equivalence hold.

mssql-tds/src/connection/transport/request_timeout.rs
```diff
@@ -0,0 +1,179 @@
...
+/// (once by this macro, then again by `Timeout`'s first poll) and the deadline
+/// starts from after that poll rather than before it. Both are permitted:
+/// futures must tolerate spurious polls, and the shift is one poll of CPU time
+/// against a budget measured in seconds.
+///
+/// On the suspending path `$fut` is polled once more than it would be today
+/// (once by this macro, then again by `Timeout`'s first poll) and the deadline
+/// starts from after that poll rather than before it. Both are permitted:
+/// futures must tolerate spurious polls, and the shift is one poll of CPU time
+/// against a budget measured in seconds.
```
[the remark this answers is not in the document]
saurabh500: Agreed — that was a copy/paste slip. Removed the duplicate and folded the `sleep(ZERO)` finding from the thread above into the surviving paragraph. Fixed in 1859916e.

mssql-tds/src/connection/transport/request_timeout.rs
```diff
@@ -0,0 +1,226 @@
...
+/// being observed. Polling once up front and only then constructing the
+/// `timeout` is therefore observationally identical, and skips the cost of
+/// arming a timer for every row served from an already-buffered packet (#271).
+///
+/// That cost is not a timer-wheel registration: on the ready path the eager
+/// form never polls the `Sleep`, so no wheel entry is ever created. It is the
+/// `Instant::now()` clock read plus building the `Sleep`, and — because
+/// `timeout` takes its future by value — moving the multi-kilobyte row future
+/// into `Timeout<F>`. Pinning first and passing `Pin<&mut F>` avoids the move
+/// as well as the timer.
```
David-Engel: **Suggestion — "multi-kilobyte row future" is not what I measured.**

I instrumented `row_fetch_futures_stay_small` on this branch and on `origin/main`:

| future | main | this PR |
|---|---|---|
| `next_row_cursor` | 1112 B | 1112 B |
| `read_row_column` | 360 B | 360 B |
| `drain_rows` | 1392 B | 1392 B |
| `get_next_row_into` | 1144 B | 1144 B |
| `read_active_plp_chunk` | 216 B | 216 B |

Byte-identical before and after, and the future actually being moved into `Timeout<F>` is a *subset* of these — so it is sub-kilobyte, not multi-kilobyte.

This does not invalidate the benchmark. The compiler already reserved slots for both `cancellable` and `Timeout<F>`, so what the change eliminates is a per-call `memcpy` *within* an unchanged state machine, not the state machine growth itself. Since this paragraph is the stated justification for reaching for a macro instead of a function, it is worth making the wording match the measurement.

mssql-tds/src/connection/transport/request_timeout.rs
```diff
@@ -0,0 +1,226 @@
...
+/// first poll, so eager `timeout` grants that same extra poll.
+macro_rules! await_within_request_timeout {
+    ($budget:expr, $fut:expr) => {{
+        let mut fut = ::std::pin::pin!($fut);
+        match $budget {
+            Some(budget) => {
+                let first = ::std::future::poll_fn(|cx| {
+                    ::std::task::Poll::Ready(::std::future::Future::poll(fut.as_mut(), cx))
+                })
+                .await;
```
David-Engel, line 55: **Nit.** `futures` 0.3 is already a direct dependency of `mssql-tds` (`Cargo.toml:44`), so `::futures::poll!(fut.as_mut())` expresses this in one line instead of the `poll_fn(|cx| Poll::Ready(...))` construction. Purely cosmetic — the current form is correct and keeps the macro std-only, which is a defensible reason to leave it.

David-Engel (commented):
## Summary

Replaces eager `tokio::time::timeout` with a probe-then-arm macro on the four per-row/per-column read paths. The core reasoning is sound and the implementation looks correct to me:

- the probe's `poll_fn` always returns `Ready`, so it never suspends and no wakeup can be lost;
- the waker the inner future registers on the probe is the same task waker `Timeout` uses on its immediate next poll, so there is no gap;
- `Option<Duration>` is `Copy`, so `$budget` is evaluated exactly once;
- macro-local `fut` / `budget` are hygienic against the caller's expressions;
- cancellation ordering (`run_until_cancelled` nested inside the budget) is preserved at all four sites.

Keying the arming decision on *suspension* rather than on `Duration::ZERO` is the right call, and the rationale is documented well.

**No blocking findings.** Suggestions are inline; two items that fall on unchanged lines are below.

## Verification I ran

Fresh worktree at `db45b6b8`:

- `cargo clippy -p mssql-tds --all-targets` — clean
- `cargo fmt --all -- --check` — clean
- all 6 new `request_timeout` tests pass
- instrumented `row_fetch_futures_stay_small` on this branch **and** on `origin/main` to compare future sizes (see the inline comment on the doc block)
- confirmed `receive_token` really is not in the per-row loops (`tds_client.rs` `get_next_row_into` / `next_row_cursor` use `receive_row_into` / `receive_row_header`), so leaving it eager is correctly scoped

## Follow-up beyond this PR (relevant to #271)

These land on unchanged lines, so noting them here rather than inline.

`tds_client.rs:3631` (`get_next_row_into`) and `tds_client.rs:3696` (`next_row_cursor`) still take an unconditional `Instant::now()` per row before every `receive_row_*` call, and `update_remaining_timeout` takes a second clock read when a budget is set. When no command timeout is configured, that first read is pure overhead on exactly the hot path this PR is optimizing. Skipping it when `remaining_request_timeout.is_none()` looks like a cheap next slice. Out of scope here — just flagging it.

## Nit on the checklist

`cargo btest` is the one unchecked box in the description. The ADO validation run is green, so I assume it just needs ticking.

mssql-tds/src/connection/transport/request_timeout.rs
```diff
@@ -0,0 +1,226 @@
...
+            self.0.fetch_add(1, Ordering::Relaxed);
+            Poll::Pending
+        }
+    }
+
+    fn is_elapsed<T>(result: &TdsResult<T>) -> bool {
+        matches!(
+            result,
+            Err(Error::TimeoutError(
+                crate::error::TimeoutErrorType::Elapsed(_)
```
David-Engel: **Nit.** `Error` is imported at the top of the module but `TimeoutErrorType` is spelled out fully qualified here. Import it alongside `Error` for consistency.

mssql-tds/src/connection/transport/request_timeout.rs
```diff
@@ -0,0 +1,226 @@
...
+        let polls = AtomicUsize::new(0);
+
+        assert!(tokio::time::timeout(zero, ready(7u8)).await.is_ok());
+        assert!(
+            tokio::time::timeout(zero, NeverReady(&polls))
+                .await
+                .is_err(),
+            "baseline: eager timeout elapses on a suspending future"
+        );
+    }
```
David-Engel: **Suggestion — this test never calls the macro.**

The doc comment says it pins "the macro's behaviour to the combinator it replaces," but the body only asserts `tokio::time::timeout`'s own behaviour; there is no comparison against `await_within_request_timeout!`. It is still useful as an upstream canary, but it is mislabelled, and `polls` is populated and then never asserted.

Either make it a real A/B the way `suspend_then_complete_on_a_spent_budget_matches_eager_timeout` does, or rename it to something like `eager_timeout_baseline_on_exhausted_budget` and drop the unused counter.

mssql-tds/src/connection/transport/request_timeout.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    #[tokio::test]
+    async fn live_budget_still_elapses_on_a_suspending_future() {
+        let polls = AtomicUsize::new(0);
+        let result =
+            await_within_request_timeout!(Some(Duration::from_millis(20)), NeverReady(&polls));
+
+        assert!(is_elapsed(&result));
+    }
```
David-Engel, line 238: **Nit.** This burns 20 ms of wall clock. `#[tokio::test(start_paused = true)]` would make it instant and deterministic, but it needs `tokio = { features = ["test-util"] }` added to dev-dependencies — the crate enables `full`, which does *not* include `test-util` (I hit exactly that when trying it). Only worth doing if you want the determinism.

mssql-tds/src/connection/transport/request_timeout.rs
```diff
@@ -0,0 +1,226 @@
...
+            }
+            None => fut.await,
+        }
+    }};
+}
+
+pub(crate) use await_within_request_timeout;
+
+#[cfg(test)]
+mod tests {
```
David-Engel, line 74: **Suggestion — nothing in the module locks in the optimization itself.**

Every test here asserts an *outcome*, and eager `timeout` satisfies all six. If someone reverts the macro body to `tokio::time::timeout(budget, fut)`, the whole module still passes — the behaviour the PR exists for is untested.

A runtime built without the time driver panics the moment a `Sleep` is constructed, which makes the ready path directly observable. I ran both halves locally: the version below passes as written, and the eager equivalent panics under the same runtime.

```rust
/// The point of the macro: the ready path must not construct a `Sleep`.
/// A runtime built without the time driver panics if one ever is.
#[test]
fn ready_path_never_constructs_a_sleep() {
    let rt = tokio::runtime::Builder::new_current_thread()
        .build()
        .expect("current-thread runtime, deliberately without the time driver");
    rt.block_on(async {
        let result: TdsResult<u8> = await_within_request_timeout!(
            Some(Duration::from_secs(30)),
            ready(Ok::<u8, Error>(7))
        );
        assert_eq!(result.expect("ready path must not arm a timer"), 7);
    });
}
```

mssql-tds/src/connection/transport/request_timeout.rs
saurabh500: Clarified the doc to name the `CancelHandle::run_until_cancelled` wrapper. The caller-visible sizes do not bound it because `async_trait` re-boxes at the transport boundary; on this head I measured `receive_row_into_internal` at 1,384 B and the actual cancellation-wrapped future passed to the macro at 4,256 B, so “multi-kilobyte” remains accurate.

mssql-tds/src/connection/transport/request_timeout.rs
saurabh500: Leaving the explicit `std::future::poll_fn` form: it keeps the macro independent of the `futures` macro and makes the deliberate `Ready(Poll)` non-suspending probe visible; the generated behavior is otherwise equivalent.

mssql-tds/src/connection/transport/request_timeout.rs
saurabh500: Fixed in 1aaf17f5: `TimeoutErrorType` is now imported alongside `Error`, and `is_elapsed` uses the shorter name.

mssql-tds/src/connection/transport/request_timeout.rs
saurabh500: Fixed in 1aaf17f5: the test now runs eager and lazy forms for both ready and pending zero-budget futures, compares their verdicts, and asserts that both poll the inner future.

mssql-tds/src/connection/transport/request_timeout.rs
saurabh500: Leaving the 20 ms test as-is: `NeverReady` removes any scheduling race, and enabling Tokio `test-util` solely to avoid this fixed 20 ms delay would broaden the test dependency surface without adding behavioral coverage.

mssql-tds/src/connection/transport/request_timeout.rs
saurabh500: Fixed in 1aaf17f5: `ready_path_never_constructs_a_sleep` runs the ready path on a runtime without a time driver. I also verified the test passes with lazy arming and fails when the macro is temporarily replaced by eager `timeout`.

saurabh500:
> `cargo btest` is the one unchecked box in the description.

Ticked. The ADO validation run on `1aaf17f5` is green.

> `tds_client.rs:3631` (`get_next_row_into`) and `tds_client.rs:3696` (`next_row_cursor`) still take an unconditional `Instant::now()` per row

Confirmed, and it is worse than "one extra clock read" when no timeout is configured: `update_remaining_timeout` only touches `start` inside `remaining_request_timeout.map(..)`, so with `None` that read is never observed at all. Filed as #297 under #247, including the other call sites that share the shape (`:3532`, `:3799`, `:3880`, `:3908`, `:2371`), rather than widening this PR.

> [!NOTE]

mssql-tds/src/connection/transport/request_timeout.rs
David-Engel: Retracting this — you are right and my table was invalid.

`TdsTokenStreamReader` is `#[async_trait]` (`token_stream.rs:192`), so `receive_row_into` returns `Pin<Box<dyn Future>>`. The caller-visible futures I measured only hold a 16-byte pointer for it, so they never bounded the future the macro moves. I should have checked that before posting numbers.

I re-measured at the real call site, with `W = dyn RowWriter + Send` to match how `receive_row_into` passes the writer:

| future | size |
|---|---|
| `receive_row_into_internal(...)` | 1,408 B |
| `CancelHandle::run_until_cancelled(None, inner)` — what goes into `Timeout<F>` | 4,328 B |

Within noise of your 1,384 / 4,256. "Multi-kilobyte" is accurate, and naming `CancelHandle::run_until_cancelled` in the doc makes it checkable. Nothing to change here.

One thing that fell out of the measurement is worth its own issue rather than this thread — I have put the detail in the review comment.

mssql-tds/src/connection/transport/request_timeout.rs
David-Engel: Agreed, that is a good reason to keep it. Keeping the macro off `futures`' macro surface and making the deliberate `Ready(Poll)` probe visible at the point it happens is worth more than the line saved. Dropping this.

mssql-tds/src/connection/transport/request_timeout.rs
David-Engel: Fair. `NeverReady` does remove the scheduling race, so the 20 ms is a fixed cost rather than a flake risk, and pulling in `tokio/test-util` for it alone is not a trade I would push for either. Dropping this.

David-Engel (commented):
## Re-review at `1aaf17f5` — looks good to me

All four actionable items are addressed and both declines are well argued. Locally on this head: `cargo fmt --all -- --check` clean, `cargo clippy -p mssql-tds --all-targets` clean, all 7 `request_timeout` tests pass. All PR checks green.

I would approve, but I have been in this thread as a reviewer of your measurements rather than as an owner, so treat this as a non-blocking sign-off.

### On the new guard test

I verified `ready_path_never_constructs_a_sleep` independently rather than taking the claim on trust. Replacing the macro body with plain eager `tokio::time::timeout` and re-running the module: that test FAILS (panics at the `Sleep` construction) while the other six still pass. It isolates exactly the property the PR exists for, which is what was missing before. Nice.

### Retraction

My "multi-kilobyte" finding was wrong — details in that thread. `async_trait` boxes the trait method, so the sizes I quoted never bounded the moved future. Re-measured at the call site: `receive_row_into_internal` is 1,408 B and the `CancelHandle::run_until_cancelled` future handed to the macro is 4,328 B, matching your numbers. Your doc wording stands.

### Follow-up that came out of that measurement (not for this PR)

Decomposing the 3.07x:

| future | size | vs. inner |
|---|---|---|
| `receive_row_into_internal` | 1,408 B | 1.00x |
| tokio-util `CancellationToken::run_until_cancelled(inner)` | 2,904 B | 2.06x |
| `CancelHandle::run_until_cancelled(None, inner)` (`core.rs:43`) | 4,328 B | 3.07x |

The outer wrapper adds 1,424 B on top of tokio-util's — very close to one whole extra copy of `inner`. That matches its shape: `f` is live in two distinct suspend states (the `Some` arm's combinator and `None => f.await`), so the state machine reserves room for both. Collapsing it to a single await point, for example resolving to one `CancellationToken` up front and using a never-cancelled token when `cancel_handle` is `None`, should land near 2,904 B for the cost of an `Arc` bump on the no-handle path.

This is worth more than a memcpy: `async_trait` expands to `Box::pin(async move { ... })` per call, so that 4,328 B is a **per-row heap allocation**, not just a per-row move. Same hot path as #271 and a larger term than the 254 ns this PR reclaims. Happy to open it as its own issue if you would rather not carry it here.

saurabh500:
> Happy to open it as its own issue if you would rather not carry it here.

Filed as #298 so it does not ride along here. Your diagnosis is right and the size prediction is essentially exact: a pin-once prototype (pin `f` before the match, pass `f.as_mut()` into `CancellationToken::run_until_cancelled`, keep the direct `None => f.await` arm) measured **4,280 B -> 2,912 B (-32.0%)** on post-#264 `main`, against your predicted ~2,904 B.

The reason I did not carry it as a follow-on PR is that the shrink did not translate into time. Seven fresh processes, 31 paired ABBA pairs each, 20,000 buffered rows per pass: no handle median **+12.2 ns/row (0.82% slower)**, live handle median **+6.8 ns/row (0.43% slower)**, per-process range **-42.7 to +75.7 ns/row** with the sign flipping across processes, while the old-vs-old A/A control sat at **-0.8 ns/row**. The harness is tight; the wrapper swing is code layout.

One correction on the framing: shrinking the future shrinks the `Box::pin` allocation but does not remove it, since `async_trait` still allocates once per call at any size. Removing that allocation is #265. All of the above is in the issue so the next attempt starts from the measurement rather than the prediction.

Thanks for re-measuring the size decomposition and for independently verifying `ready_path_never_constructs_a_sleep` against an eager control rather than taking the claim on trust.

> [!NOTE]

David-Engel:
Correction to my own follow-up review above — please disregard the cancellation-wrapper section.

**It is a duplicate.** #265 already carries it as *"Adjacent finding: ~70% of the row future is cancellation plumbing"*, down to the same mechanism at `core.rs:50-56` (`f` awaited in both match arms, separate suspend points, so rustc stores both rather than overlapping) and the same arithmetic corroboration. My 1,408 / 2,904 / 4,328 B are that table re-measured on a `dyn RowWriter + Send` instantiation; #265 has 1,328 / 2,744 / 4,088 on `DefaultRowWriter`. I should have checked the issue tree before writing it up as new.

**And my conclusion was backwards.** I said it was a larger term than what this PR reclaims. #265's paired wrapper ladder already priced it:

| Mode | median ms | vs bare |
|---|---:|---:|
| bare | 32.811 | — |
| `run_until_cancelled(None, ..)` | 34.709 | +5.8% |
| `+ tokio::time::timeout(..)` | **46.751** | **+42.5%** |

So the cancellation plumbing is ~70% of the future by size but only ~5% by time, while the per-row `timeout` is +112 B and ~36% of the cost — the wrapper **this PR removes** is the expensive one. #265 says it plainly: *"Size and time are anti-correlated here."* I argued from allocation size against a measurement in this repo that already said not to do that. That is the same mistake as the `async_trait` one earlier in this review: I measured a proxy and reported it as the result.

For the record, the size does still matter, but for a reason that belongs to #265 rather than the one I gave: at ~1.4 KB the row future would fit the 4096 B `row_fetch_futures_stay_small` guard *unboxed*, which would remove the per-row allocation outright instead of merely preserving it.

No new issue needed — #265 owns this, and #297 now covers the `Instant::now()` item. Nothing here changes the review: #275 still looks good to me.

David-Engel:
We crossed by about thirty seconds — #298 was filed while I was posting, so ignore the "no new issue needed" line in my previous comment. #298 is the right home and my comment predates it.

Your ABBA run settles it better than my retraction did. The prototype landing at 2,912 B against a predicted ~2,904 B and *still* measuring +12.2 / +6.8 ns/row — with the sign flipping across processes against a −0.8 ns/row A/A control — is the cleanest statement of the point: an exact model of the wrong quantity. Worth keeping in #298 so the next person does not re-derive the byte table and assume it implies time.

Taking your framing correction as well. Shrinking the future does not remove the `Box::pin`; `async_trait` allocates once per call at any size, and removal is #265. My "would fit the guard unboxed" sentence was quoting #265 about the *whole* row future, but placed where it was it reads as though the wrapper shrink gets you there. It does not.

Nothing changes on the PR — #275 still looks good.

David-Engel (approved):

PR #278, opened 2026-08-13
odbc: SQLColAttributeW and the sql_variant underlying-type chain (P2) [AB#46579]

David-Engel:
## What

Implements `SQLColAttributeW` (P2), plus the two things the `sql_variant` path turned out to depend on.

mssql-python resolves `SQLColAttributeW` at load time and includes the pointer in its non-null check, so without this export the driver does not load at all.

### The variant chain

Reading a `sql_variant` column takes three things, not one, and mssql-python needs all of them. On any failure it logs and yields `None` for the column, so a missing link shows up as **silently empty data rather than an error** — worth knowing when reviewing, because two of these links look unrelated to this work item:

1. `SQLDescribeCol` must report `SQL_SS_VARIANT`. We reported `SQL_VARCHAR`, and mssql-python branches on the exact type, so it never entered the variant path and treated every variant column as a string.
2. `SQLGetData(col, SQL_C_BINARY, NULL, 0, &indicator)` must succeed. This is a length/NULL probe, not a data read.
3. `SQLColAttribute(SQL_CA_SS_VARIANT_TYPE)` returns the C type of the value just probed.

### How the underlying type is carried

The base type is a property of the **value**, not the column — a variant column can hold a different type in every row — and `read_sql_variant` discarded it after decoding. The decoded `ColumnValues` cannot always recover it, since `varchar` and `nvarchar` both arrive as `ColumnValues::String`.

So it is carried up from the decoder:

- `RowWriter` gains a **defaulted** `write_variant_base_type`, so `PyRowWriter` and the mssql-js writer compile unchanged.
- `decode_into` gets an explicit `SsVariant` arm that reports the base type before writing the value.
- `CursorColumn::Value` carries it alongside the value, mirroring how `PlpStreaming` already carries its collation. `CursorColumn` is used only by mssql-odbc, so widening it is contained.
- `StmtState` holds it next to `last_captured` and clears it with the rest of the row-stream state.

**`ColumnValues` is deliberately untouched.** That is what keeps this change out of the Python and Node bindings — `mssql-py-core` matches on it in ~300 places.

## Reviewer notes

- **Behaviour change:** `sql_variant` columns now describe as `SQL_SS_VARIANT` (−150) instead of `SQL_VARCHAR`. Applications that keyed on the old value will see the new one, which is what msodbcsql reports.
- **Deliberate divergence:** for a variant holding `decimal`/`numeric`/`money`, `SQL_CA_SS_VARIANT_TYPE` reports `SQL_C_CHAR` where msodbcsql reports `SQL_C_NUMERIC`. Reporting the latter would make the caller request a `SQL_NUMERIC_STRUCT` this driver refuses; character is how those values are actually delivered. Recorded in the divergence table.
- **Scope call:** the `SQL_C_BINARY` probe arguably belongs to [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239). I took the zero-length form here because P2 ships unreachable code otherwise, and left binary *delivery* on that item — a real buffer is still `HYC00`. [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239) has been updated to say so.
- `HY113` for the not-a-variant case matches msodbcsql's `IDS_S1_113`; its `S1` prefix is the ODBC 2.x spelling of `HY` (cf. `IDS_S1_C00` = `HYC00`).

## Testing

533 unit tests, workspace clippy and `cargo fmt` clean. The mssql-tds changes were also validated against SQL Server 2025 in Docker: 1647 unit tests and all 24 `test_client_read_apis` integration tests pass.

New e2e `col_attribute_test.cpp`, because unit tests can only build `int` column metadata — the per-type mapping tables and the entire variant path had no coverage otherwise. The variant test reads two rows holding different base types in the same column, which is what pins the type to the value rather than the column.

Every hand-written `SQL_DESC_*` constant was cross-checked against the real ODBC headers with `static_assert`; all 26 match.

[AB#46579](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46579)


mssql-tds/src/datatypes/decoder.rs
```diff
@@ -502,6 +502,21 @@ impl GenericDecoder {
...
+            .map(|(_, value)| value)
+    }
+
+    /// As [`Self::read_sql_variant`], but also returns the base type the variant
+    /// declared on the wire. The decoded value alone cannot always recover it —
+    /// `varchar` and `nvarchar` both arrive as [`ColumnValues::String`].
+    async fn read_sql_variant_with_base<T>(
+        &self,
+        reader: &mut T,
+    ) -> TdsResult<(TdsDataType, ColumnValues)>
```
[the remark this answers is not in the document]
David-Engel: Correct, and fixed in c4387929. `read_sql_variant_with_base` now returns early on a zero length with the base as `None`, and `decode_into` skips the hook rather than reporting a type that was never on the wire.

Worth recording how this behaves, because it changed how I tested it. **The path is not reachable through a normal SQL Server response:** NULLs arrive in the NBCROW null bitmap and `nbcrow_parser.rs` skips the column entirely, so the decoder is never called. I confirmed that empirically — an e2e test selecting `CAST(NULL AS SQL_VARIANT)` followed by an `INT` passes with *and* without the guard, so it does not discriminate.

The regression test is therefore a decoder unit test that feeds the wire form directly, placing a sentinel where the following column would start:

```rust
let mut reader = ByteReader::new(vec![0, 0, 0, 0, 0xAB, 0xCD]);
...
assert_eq!(reader.read_byte().await.unwrap(), 0xAB,
           "the NULL variant consumed the following column's bytes");
```

With the guard removed it fails with `ProtocolError("Invalid TDS Type 171")` — 171 is `0xAB`, the sentinel, consumed and misread as a base type. That is exactly the failure mode you described.

I kept the e2e test as well, since it covers the NBCROW path that applications actually hit, but the unit test is the one that pins this fix.

This was pre-existing in `read_sql_variant`; the new cursor path inherited it rather than introducing it.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,655 @@
...
+        if !is_variant {
+            post_diag(&mut stmt_state, ERR_NOT_VARIANT_COLUMN);
+            return SQL_ERROR;
+        }
+        let Some(base) = stmt_state.last_variant_base else {
+            // Callers probe the column with SQLGetData first; that read is what
+            // supplies the base type.
+            post_diag(&mut stmt_state, ERR_FUNCTION_SEQUENCE);
+            return SQL_ERROR;
+        };
```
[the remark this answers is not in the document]
David-Engel: Correct, fixed in c4387929. `last_variant_base` was a bare `Option<TdsDataType>` with nothing tying it to a column, so after probing column 1 the attribute answered for any other variant column.

It now carries the 1-based index and only answers for that column:

```rust
let base = stmt_state
    .last_variant_base
    .filter(|(col, _)| *col == column_number as usize)
    .map(|(_, base)| base);
```

I keyed it to its own index rather than deriving the check from `last_captured`, so it stays correct regardless of when that value is consumed by a subsequent read.

Added `VariantTypeIsPerColumn` to the e2e suite: it selects two variant columns holding different base types, probes only the first, and asserts the second reports `HY010` instead of inheriting the first column's type.

The other three locations you listed are the same `base` binding flowing through the one code path, so they are all covered by this change.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,971 @@
...
+
+/// Maps a field identifier to its value, or `None` when the field is not one
+/// this driver reports.
+fn column_attribute(meta: &ColumnMetadata, field_identifier: SqlUSmallInt) -> Option<Attr> {
+    let sql_type = odbc_sql_type(meta);
+    let attr = match field_identifier {
+        // `SQL_DESC_TYPE` and `SQL_DESC_CONCISE_TYPE` differ only for the
+        // datetime/interval types, which this driver reports as concise types.
+        SQL_DESC_TYPE | SQL_DESC_CONCISE_TYPE => Attr::Numeric(SqlLen::from(sql_type)),
+        SQL_DESC_LENGTH | SQL_DESC_DISPLAY_SIZE => {
```
saurabh500: **Severity: Blocking** — `SQL_DESC_DISPLAY_SIZE` cannot share `column_size`. These are different ODBC attributes: `int` needs 11 display characters (sign plus 10 digits), `uniqueidentifier` needs 36, and binary values need two characters per byte. Returning 10, 16, and `n` respectively can make callers allocate undersized character buffers and truncate valid values. Please use a type-specific display-size mapping that matches msodbcsql.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,971 @@
...
+        SQL_DESC_NAME | SQL_DESC_LABEL | SQL_DESC_BASE_COLUMN_NAME => {
+            Attr::Text(meta.column_name.clone())
+        }
+        SQL_DESC_TYPE_NAME => Attr::Text(type_name(meta).to_string()),
+        _ => return None,
+    };
+    Some(attr)
+}
+
+/// Storage size in bytes of the column's value on the wire.
```
saurabh500: **Severity: Blocking** — `SQL_DESC_OCTET_LENGTH` is the ODBC transfer/descriptor size, not the TDS wire payload length. For example, a TDS `date` payload is 3 bytes while the default ODBC `SQL_DATE_STRUCT` is 6 bytes, and timestamp structures require 16 bytes. Reporting the wire length can make applications allocate buffers that are too small. Please map this by ODBC SQL type and transfer representation.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,971 @@
...
+}
+
+/// Maps a field identifier to its value, or `None` when the field is not one
+/// this driver reports.
+fn column_attribute(meta: &ColumnMetadata, field_identifier: SqlUSmallInt) -> Option<Attr> {
+    let sql_type = odbc_sql_type(meta);
+    let attr = match field_identifier {
+        // `SQL_DESC_TYPE` and `SQL_DESC_CONCISE_TYPE` differ only for the
+        // datetime/interval types, which this driver reports as concise types.
+        SQL_DESC_TYPE | SQL_DESC_CONCISE_TYPE => Attr::Numeric(SqlLen::from(sql_type)),
```
saurabh500: **Severity: Should fix** — `SQL_DESC_TYPE` is the verbose type field. For date, time, and timestamp columns it must return `SQL_DATETIME` (with the subtype reported separately), while `SQL_DESC_CONCISE_TYPE` returns `SQL_TYPE_DATE`, `SQL_TYPE_TIME`, or `SQL_TYPE_TIMESTAMP`. Treating the two fields as identical breaks descriptor consumers.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,971 @@
...
+        return 0;
+    }
+    // `type_info.length` is already a byte count for every type, including the
+    // national character types.
+    SqlLen::try_from(meta.type_info.length).unwrap_or(SqlLen::MAX)
+}
+
+/// `SQL_DESC_PRECISION`: the number of significant digits for the exact and
+/// approximate numeric types, otherwise the column size.
+fn precision(meta: &ColumnMetadata) -> SqlSmallInt {
```
saurabh500, line 433: **Severity: Should fix** — Falling back to `column_size` does not implement `SQL_DESC_PRECISION`. Approximate numerics need binary precision (`real` = 24 and `float` = 53), and temporal types report fractional-second precision rather than display width. Please make this mapping type-specific instead of reusing the display-oriented size.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,971 @@
...
+        | TdsDataType::VarBinary
+        | TdsDataType::BigVarBinary => SQL_C_BINARY,
+        TdsDataType::Guid => SQL_C_GUID,
+        // SQL Server rejects the remaining types at insert time, so a variant
+        // cannot actually carry them; character is the safe fallback.
+        _ => SQL_C_CHAR,
+    }
+}
+
+/// `tinyint` is the only unsigned integer SQL Server exposes.
```
saurabh500: **Severity: Should fix** — The ODBC rule is broader than “tinyint is unsigned”: `SQL_DESC_UNSIGNED` is `SQL_TRUE` for unsigned numeric columns **or for nonnumeric columns**, and `SQL_FALSE` for signed numeric columns. This helper therefore returns the wrong value for character, date/time, GUID, and other nonnumeric types.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,971 @@
...
+        SQL_DESC_UNNAMED => Attr::Numeric(if meta.column_name.is_empty() {
+            SQL_UNNAMED
+        } else {
+            SQL_NAMED
+        }),
+        // The result set is not known to be updatable, and no column here is a
+        // known auto-increment column; report the "unknown"/false forms rather
+        // than claiming either way.
+        SQL_DESC_UPDATABLE => Attr::Numeric(SQL_ATTR_READWRITE_UNKNOWN),
+        SQL_DESC_AUTO_UNIQUE_VALUE => Attr::Numeric(SqlLen::from(false)),
```
saurabh500: **Severity: Should fix** — Returning false unconditionally makes `SQL_DESC_AUTO_UNIQUE_VALUE` incorrect for `IDENTITY` columns. The metadata already exposes identity status; please return `meta.is_identity()` here.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,971 @@
...
+        SQL_DESC_SEARCHABLE => Attr::Numeric(SQL_PRED_SEARCHABLE),
```
saurabh500: **Severity: Should fix** — `SQL_PRED_SEARCHABLE` claims full comparison and `LIKE` support for every type. That overstates capability for types that support only basic predicates and for types such as XML/image that are not fully searchable. Please derive `SQL_PRED_NONE`, `SQL_PRED_CHAR`, `SQL_PRED_BASIC`, or `SQL_PRED_SEARCHABLE` from the SQL type.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -0,0 +1,971 @@
...
+        SQL_DESC_NAME | SQL_DESC_LABEL | SQL_DESC_BASE_COLUMN_NAME => {
```
saurabh500: **Severity: Should fix** — `SQL_DESC_BASE_COLUMN_NAME` is not the result-set label. For `SELECT source_col AS alias`, `SQL_DESC_NAME`/`SQL_DESC_LABEL` may return `alias`, but the base-column field must return `source_col`; if that provenance is unavailable, ODBC expects an empty string rather than the alias.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -114,7 +115,14 @@ pub struct PlpChunk {
 #[derive(Debug, PartialEq)]
 pub enum CursorColumn {
     /// A fully decoded, materialized column value (non-PLP).
-    Value(ColumnValues),
+    Value {
```
saurabh500, line 118: **Severity: Should fix** — This changes the public `CursorColumn::Value(ColumnValues)` tuple variant into a struct variant, so downstream constructors and pattern matches no longer compile. If this break is intentional before the initial release, please document it in the changelog or migration notes; otherwise provide a compatibility-preserving way to expose `variant_base`.

mssql-tds/tests/test_client_read_apis.rs
```diff
@@ -896,7 +896,7 @@ mod client_based_iterators {
             let c2 = client.read_row_column(1).await?;
             assert!(matches!(
                 &c2,
-                CursorColumn::Value(ColumnValues::String(s)) if s.to_utf8_string() == "row1-c2"
+                CursorColumn::Value { value: ColumnValues::String(s), .. } if s.to_utf8_string() == "row1-c2"
```
saurabh500, line 900: **Severity: Should fix** — Every updated assertion here either ignores `variant_base` with `..` or checks only `None`. Nothing directly proves that `read_row_column` returns `Some(base)` for a real `sql_variant`. Please add a non-first variant column with rows containing `varchar`, `nvarchar`, and NULL, asserting `BigVarChar`, `NVarChar`, and `None`; that pins output-position association and per-row clearing at this API boundary.

mssql-odbc/tests/e2e/tests/col_attribute_test.cpp
```diff
@@ -0,0 +1,288 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// col_attribute_test.cpp  –  E2E tests for SQLColAttributeW.
+//
+// Unit tests can only build `int` column metadata, so the per-type mapping
+// tables (concise type, type name, radix, and the sql_variant underlying type)
+// are only meaningfully exercised here, against a live server.
```
saurabh500, line 6: **Severity: Should fix** — Several critical descriptor mappings added in this PR are not exercised here. Please add live msodbcsql parity cases for display size (`int`, GUID, binary, decimal/float), transfer octet length (`date`/timestamp), verbose versus concise datetime type, nonnumeric `SQL_DESC_UNSIGNED`, identity/searchability, and aliased base-column names. The current unit tests execute these mapping arms but lock in values such as an `int` display size of 10, so they do not catch these contract errors.

saurabh500 (changes requested):
Requesting changes for the two blocking ODBC descriptor-size correctness issues. See the inline comments for details and the associated should-fix items.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Fixed in 56a14f7. `SQL_DESC_OCTET_LENGTH` is now the transfer size: `date` 6 (`SQL_DATE_STRUCT`), `time` 12 (`SQL_SS_TIME2_STRUCT`), the `datetime`/`smalldatetime`/`datetime2` family 16 (`SQL_TIMESTAMP_STRUCT`), `datetimeoffset` 20 (`SQL_SS_TIMESTAMPOFFSET_STRUCT`), and the exact numerics their rendered width since they transfer as characters. Everything else keeps its wire width, which for `nchar`/`nvarchar` is bytes.

Pinned by `OctetLengthIsTransferSize`, asserted on both legs.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Fixed in 56a14f7. `SQL_DESC_TYPE` now returns `SQL_DATETIME` for the `datetime`/`smalldatetime`/`datetime2` family while `SQL_DESC_CONCISE_TYPE` returns `SQL_TYPE_TIMESTAMP`.

One thing worth flagging from the source: `date` is **not** part of the fold. `GetIRDField` remaps `SQL_DATE` to `SQL_TYPE_DATE` *before* the `SQL_DATE..SQL_TIMESTAMP` range check, so the range test misses it and `date` reports the concise type for both fields. `time` and `datetimeoffset` likewise use `SQL_SS_TIME2`/`SQL_SS_TIMESTAMPOFFSET`, which are outside the range. `VerboseTypeMatchesConciseForNonTimestamps` locks that in on both legs.

I also added `SQL_DESC_DATETIME_INTERVAL_CODE` so collapsing the verbose type does not lose which member it was. msodbcsql rejects that field through `SQLColAttribute`, so that assertion skips the comparison leg and the additive divergence is documented in the plan doc.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Fixed in 56a14f7. `real` reports 24 and `float` 53, and the temporal types report fractional-seconds precision (`datetime` 3, `smalldatetime` 0, `date` 0, `time(s)`/`datetime2(s)`/`datetimeoffset(s)` their scale). Covered by `precision_is_type_specific`.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Fixed in 56a14f7. Display size is now its own mapping, separate from `column_size`: `int` 11, `smallint` 6, `bigint` 20, `tinyint` 3 (no sign), GUID 36, binary `2*n`, decimal `precision+2`, `real` 14, `float` 24, and the national character types report characters where octet length reports bytes.

Pinned by `DisplaySizeIsRenderedWidth` and `DisplaySizeForApproximateNumerics`, both of which run on **both** legs of the parity run, so they are checked against msodbcsql 18.6 rather than against my reading of it.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Fixed in 56a14f7, and this one was the most interesting of the batch.

The rule is now keyed off the **ODBC** type rather than the TDS type, mirroring `IsUnsigned()` in `sqlcdesc.cpp`, which is a bitmask over `fSqlType`: `SQL_FALSE` only for `SMALLINT`/`INTEGER`/`BIGINT`/`FLOAT`/`DOUBLE`/`REAL`/`DECIMAL`/`NUMERIC`, and `SQL_TRUE` for everything else including character, binary, temporal, GUID and bit.

Keying off the TDS type gets exactly one case wrong, and the parity run caught it: `money` is reported as `SQL_DECIMAL`, so it is a **signed** numeric, even though nothing about the TDS type says so. `UnsignedIsFalseOnlyForSignedNumerics` covers all 15 types on both legs.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Fixed in 56a14f7 — now `meta.is_identity()`. Covered by `identity_column_reports_auto_unique_value` and, on both legs, by `IdentityColumnReportsAutoUniqueValue`, which needs a real table since a projected expression is never an identity column.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Fixed in 56a14f7, derived from the type per the `SQL_DESC_SEARCHABLE` case in `GetIRDField`: `text`/`ntext` are `SQL_PRED_CHAR` (LIKE only); `image`, `xml`, `udt` and `vector` are `SQL_PRED_NONE`; the character and calendar types are `SQL_PRED_SEARCHABLE`; and the numerics, binary, bit, GUID and `sql_variant` are `SQL_PRED_BASIC`. `SearchableIsDerivedFromTheType` asserts all ten on both legs.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Fixed in 56a14f7. `SQL_DESC_NAME`/`SQL_DESC_LABEL` still report `column_name` (the alias); `SQL_DESC_BASE_COLUMN_NAME` now returns an empty string.

COLMETADATA does not carry the originating column outside `FOR BROWSE`, so the provenance genuinely is unavailable here and ODBC asks for the empty string rather than a guess. `AliasedColumnDoesNotReportTheAliasAsBaseColumnName` runs `SELECT source_col AS alias_col` on both legs and asserts the base-column name is not the alias.

mssql-tds/src/connection/tds_client.rs
David-Engel: Documented in 56a14f7 — CHANGELOG "Unreleased / Changed" now records the shape change, why it was needed (a `varchar` and an `nvarchar` variant both decode to `ColumnValues::String`, so the base type has to come from the wire header rather than the value), and the mechanical fix for both pattern matches and constructors.

The break is intentional ahead of the initial release. Within the workspace the only consumer is mssql-odbc, which is updated here.

mssql-tds/tests/test_client_read_apis.rs
David-Engel: Good catch — you were right that nothing proved the positive case. Added `read_row_column_reports_the_variant_base_type` in 56a14f7, built exactly as you described:

- the variant is the **second** column, so a base captured for it cannot be mistaken for column 0's
- row 1 carries `varchar` and asserts `Some(BigVarChar)`
- row 2 carries `nvarchar` and asserts `Some(NVarChar)` — both decode to `ColumnValues::String`, so only the base type distinguishes them
- row 3 is a NULL variant and asserts `None`, which pins per-row clearing rather than latching row 2's base

Passing against SQL Server 2025.

mssql-odbc/tests/e2e/tests/col_attribute_test.cpp
David-Engel: Added in 56a14f7 — and this turned out to be the most valuable comment of the set, because writing them is what caught the `money` bug in `SQL_DESC_UNSIGNED` that my source reading had missed.

New cases, all asserted on **both** legs: `DisplaySizeIsRenderedWidth`, `DisplaySizeForApproximateNumerics`, `OctetLengthIsTransferSize`, `VerboseTypeDiffersFromConciseForTimestamps`, `VerboseTypeMatchesConciseForNonTimestamps`, `SearchableIsDerivedFromTheType`, `UnsignedIsFalseOnlyForSignedNumerics`, `IdentityColumnReportsAutoUniqueValue`, `AliasedColumnDoesNotReportTheAliasAsBaseColumnName`.

You were also right that the old unit tests were locking in the wrong contract — `int` display size 10 among them. Those expectations are corrected and the unit tests now cover the mapping shape while the e2e cases own the values.

Full run: **20 parity, 0 divergences, 0 shared failures** against msodbcsql 18.6 on SQL Server 2025.

saurabh500 (approved):
The blocking issues from the previous review have been addressed appropriately.

PR #283, opened 2026-08-13
Stop printing raw AAD access tokens in connectivity tests

### LINKED ISSUE #236: connectivity tests print the raw AAD access token to stdout
## Summary

`mssql-tds/tests/connectivity.rs` prints the raw AAD bearer token to stdout in two places. Both are unconditional — the token is emitted on every call, not only on failure.

https://github.com/microsoft/mssql-rs/blob/main/mssql-tds/tests/connectivity.rs#L35-L43

```rust
async fn generate_access_token() -> String {
    let credential = DeveloperToolsCredential::new(None);

    let token_response = credential.unwrap().get_token(&[SCOPE], None).await;

    let secret = token_response.as_ref().unwrap().token.secret();
    print!("{secret}");          // <-- prints the bearer token
    secret.to_string()
}
```

https://github.com/microsoft/mssql-rs/blob/main/mssql-tds/tests/connectivity.rs#L84-L86

```rust
    let secret = token_response.as_ref().unwrap().token.secret();
    print!("{secret}");          // <-- same, in the STS/resource variant
    secret.to_string()
}
```

The scope is `https://database.windows.net/.default`, so this is a usable credential for Azure SQL, carrying the developer's identity (`upn`, `oid`, `tid`, group claims).

## Why it matters

`cargo test` buffers stdout and prints it **for failing tests**, so the token surfaces exactly when someone is debugging — the moment they are most likely to copy the output somewhere:

```
---- connectivity::select_1 stdout ----
eyJ0eXAiOiJKV1QiLCJhbGciOiJSUzI1NiIsIng1dCI6...   <-- full token
This test expects that `az login --tenant ...` was run to get a token.

thread 'connectivity::select_1' panicked at mssql-tds/tests/connectivity.rs:149:48:
called `Result::unwrap()` on an `Err` value: TimeoutError(String("Timeout while connecting"))
```

Consequences:

- The token lands in terminal scrollback and in any CI log that runs these tests.
- Anyone pasting a failure into an issue, a PR comment, or a chat transcript leaks a live credential without noticing — the token is the first thing in the block, above the actual error.
- `--nocapture` prints it for passing runs too.

Tokens are short-lived, which limits but does not remove the exposure.

## Suggested fix

Delete both `print!("{secret}")` calls. If the intent was to make the token available for manual debugging, gate it behind an explicit opt-in and print a fingerprint rather than the secret:

```rust
let secret = token_response.as_ref().unwrap().token.secret();
// Never print the token itself; a failure here is almost always "no token",
// which the hint below already explains.
debug_assert!(!secret.is_empty(), "empty access token");
secret.to_string()
```

The existing hint immediately below (`"This test expects that az login ... was run to get a token."`) already conveys everything a developer needs to diagnose the common failure, without the secret.

Worth a quick sweep of the other test files for the same pattern while this is being fixed.

## Workaround until then

When running these locally, filter the output, e.g. `cargo test ... 2>&1 | grep -v "eyJ"`, or match only the `test result:` lines.

## Notes

Found while merging main into #217 and running the workspace test suite.

David-Engel:
## Description

`mssql-tds/tests/connectivity.rs` unconditionally wrote the raw AAD bearer token to stdout in two helpers (`generate_access_token` and `generate_access_token_with_sts_and_resource`). Since `cargo test` surfaces captured stdout for failing tests, the token showed up exactly when someone was debugging and most likely to paste output into an issue, PR comment, or chat — and `--nocapture` printed it on passing runs too. The scope is `https://database.windows.net/.default`, so the leaked value is a usable Azure SQL credential carrying the developer's identity claims.

Both `print!("{secret}")` calls are replaced with a `debug_assert!(!secret.is_empty(), "empty access token")`. The existing hint below (`"This test expects that az login ... was run to get a token."`) already covers the common failure mode without exposing the secret.

Swept the rest of the repo for the same pattern: the other token-related prints (`test_windows_sspi.rs`, `test_kerberos_gssapi.rs`, `test_mock_server_fedauth.rs`) only log byte counts or mock-server verification messages, so no other changes are needed.

## Related Issues

Fixes #236

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes (`cargo clippy -p mssql-tds --all-features --all-targets -- -D warnings`; `--frozen` skipped because `Cargo.lock` isn't checked in)
- [ ] `cargo btest` passes — these connectivity tests require a live Azure SQL endpoint and `az login`; unchanged test logic, only the print removed
- [ ] New/changed functionality has tests — n/a, test-only change
- [ ] Public API changes are documented — n/a

### DIFF
--- a/mssql-tds/tests/connectivity.rs
+++ b/mssql-tds/tests/connectivity.rs
@@ -38,7 +38,7 @@ mod connectivity {
         let token_response = credential.unwrap().get_token(&[SCOPE], None).await;
 
         let secret = token_response.as_ref().unwrap().token.secret();
-        print!("{secret}");
+        debug_assert!(!secret.is_empty(), "empty access token");
         secret.to_string()
     }
 
@@ -82,7 +82,7 @@ mod connectivity {
         };
 
         let secret = token_response.as_ref().unwrap().token.secret();
-        print!("{secret}");
+        debug_assert!(!secret.is_empty(), "empty access token");
         secret.to_string()
     }
 


saurabh500 (approved):
Thanks for dealing with this ! :)

PR #284, opened 2026-08-13
Report 22003 for an unrepresentable numeric column value

### LINKED ISSUE #235: mssql-odbc: an out-of-range numeric value reports 07006 instead of 22003
## Summary

`numeric_source` in `mssql-odbc/src/api/fetch_convert.rs` returns `Option<NumericSource>`, so every failure collapses to a single outcome. The caller maps `None` to `ConvError::Restricted`, which surfaces as **`07006`** ("restricted data type attribute violation"):

```rust
numeric_source(value).ok_or(ConvError::Restricted)
```

That is correct for a column with no numeric interpretation (binary, guid, a date/time column into an integer target). It is wrong for a value that *is* numeric but cannot be represented, which should be **`22003`** ("numeric value out of range").

## Concrete case

After #217, the decimal arm refuses a payload with more than 4 limbs, and a magnitude above `i128::MAX` fails `i128::try_from`:

```rust
if d.int_parts.len() > 4 {
    return None;
}
...
let m = i128::try_from(mag).ok()?;
```

Both return `None`, so both report `07006`. An application sees "this conversion is not allowed for this column type" when the truthful answer is "this value does not fit".

## Suggested fix

Change the signature to carry the reason:

```rust
fn numeric_source(value: &ColumnValues) -> Result<NumericSource, ConvError>
```

returning `ConvError::OutOfRange` for the representability failures and `ConvError::Restricted` only for a column with no numeric interpretation. `numeric_source_or_parse` then propagates rather than re-wrapping.

Both variants already exist and are already mapped to the right SQLSTATEs by `finish_typed_conv`, so this is a signature and call-site change rather than new diagnostic plumbing.

## Scope

Small and self-contained, but it touches every arm of `numeric_source` and its callers, which is why it was kept out of #217 rather than folded into an unrelated review round.

## Notes

Raised by @saurabh500 while reviewing #217.

David-Engel:
## Description

`numeric_source` in `mssql-odbc/src/api/fetch_convert.rs` returned `Option<NumericSource>`, so every failure collapsed into a single `None`. The caller turned that into `ConvError::Restricted`, which `finish_typed_conv` posts as **07006** ("restricted data type attribute violation").

That is right for a column with no numeric interpretation (binary, guid, a date/time column into an integer target). It is wrong for a value that *is* numeric but cannot be represented, which is **22003** ("numeric value out of range").

`numeric_source` now returns `Result<NumericSource, ConvError>`, and `numeric_source_or_parse` propagates instead of re-wrapping, so its character-column arms (22018 / 22003) are unchanged. Both `ConvError` variants already existed and were already mapped to the right SQLSTATEs by `finish_typed_conv`, so there is no new diagnostic plumbing. Both callers (`convert_integer_c`, `convert_float_c`) get the corrected SQLSTATE.

### Wide decimals

Review raised that the decimal arm was conflating two different failures, so they are now split:

- **Magnitude above `i128::MAX`, still within 4 words.** Reachable: the decoder validates limb count but never the magnitude against `precision`, so 16 bytes of `0xFF` arrives intact from a non-conforming server. There is no exact scaled form, so it degrades to the `f64` reading. A float target then succeeds — matching msodbcsql's `ConvertToFloat`, which accumulates all four words into a double — while an integer target still reports 22003 because the value falls outside the integer range.
- **Magnitude wider than 4 words.** Defensive only, rejected by both decode paths. Reports 22003. `SQL_NUMERIC_STRUCT` is four words as well, so there is no msodbcsql counterpart here to diverge from.

Parity statement: **matches** msodbcsql. `sqlcprot.h` maps `CVT_PREC` → `IDS_22_003` and `CVT_ILLEGAL` → `IDS_07_006`; `sqlccnvt.cpp` `ConvertToFixed` returns `CVT_PREC` from its range checks and reserves `CVT_ILLEGAL` for its `default:` arm — the same split drawn here. Cited in code. No new row in the known-divergences table, since the one reachable divergence is now gone rather than documented.

Reaching parity needed no `mssql-tds` change: `DecimalParts::to_f64` and `magnitude_big` are private, but `magnitude()` is `pub` and returns `u128`, which already covers the reachable case.

### Tests

- `decimal_limbs_are_reassembled_and_bounded` — asserts `OutOfRange` for the 5-word payload and for a 4-word magnitude above `i128::MAX`, plus a new boundary case that the largest legal `decimal(38, 38)` magnitude (10^38 - 1, ≈9.99e37 vs `i128::MAX` ≈1.70e38) still keeps its exact form.
- `wide_decimal_into_float_target_matches_msodbcsql` — the reachable 4-word case converts on a float target.
- `oversized_decimal_into_float_target_is_out_of_range` — the defensive >4-word case still reports 22003.

## Related Issues

Fixes #235

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (567 `mssql-odbc` lib tests pass locally via `cargo test -p mssql-odbc --all-features --lib`; `cargo-nextest` is not installed in this environment, so the full workspace nextest run is left to CI)
- [x] Public API changes are documented (none — `numeric_source` is private to the module)

### DIFF
--- a/mssql-odbc/src/api/fetch_convert.rs
+++ b/mssql-odbc/src/api/fetch_convert.rs
@@ -211,17 +211,31 @@ fn numeric_source(value: &ColumnValues) -> Result<NumericSource, ConvError> {
         ColumnValues::Bit(b) => Ok(NumericSource::Int(i128::from(*b))),
         ColumnValues::Real(x) => Ok(NumericSource::Float(f64::from(*x))),
         ColumnValues::Float(x) => Ok(NumericSource::Float(*x)),
-        // `DecimalParts` stores a base-2^32 little-endian magnitude. 38 digits
-        // fit in 4 words; a wider magnitude, or one past `i128::MAX`, is a
-        // numeric value this converter cannot represent, not an illegal cast.
+        // `DecimalParts` stores a base-2^32 little-endian magnitude. A legal
+        // 38-digit value fits in 4 words and in `i128`, so it keeps its exact
+        // scaled form. A non-conforming server can still send 4 words above
+        // `i128::MAX`; that has no exact form here but is a real number, so it
+        // degrades to the `f64` reading rather than an error. Only a magnitude
+        // wider than 4 words has no reading at all.
+        //
+        // Matches msodbcsql: `sqlccnvt.cpp` `ConvertToFixed` returns `CVT_PREC`
+        // (`IDS_22_003`) for an out-of-range numeric and reserves `CVT_ILLEGAL`
+        // (`IDS_07_006`) for a source with no numeric reading, and
+        // `ConvertToFloat` accumulates all four words into a double. Four words
+        // is also all `SQL_NUMERIC_STRUCT` can hold, so a wider magnitude has no
+        // msodbcsql counterpart to diverge from.
         ColumnValues::Decimal(d) | ColumnValues::Numeric(d) => {
-            let m = d
-                .magnitude()
-                .and_then(|mag| i128::try_from(mag).ok())
-                .ok_or(ConvError::OutOfRange)?;
-            Ok(NumericSource::Scaled {
-                mantissa: if d.is_positive { m } else { -m },
-                scale: u32::from(d.scale),
+            let mag = d.magnitude().ok_or(ConvError::OutOfRange)?;
+            let scale = u32::from(d.scale);
+            Ok(match i128::try_from(mag) {
+                Ok(m) => NumericSource::Scaled {
+                    mantissa: if d.is_positive { m } else { -m },
+                    scale,
+                },
+                Err(_) => {
+                    let v = mag as f64 / 10f64.powi(scale as i32);
+                    NumericSource::Float(if d.is_positive { v } else { -v })
+                }
             })
         }
         ColumnValues::Money(m) => Ok(NumericSource::Scaled {
@@ -350,8 +364,11 @@ pub(crate) fn is_float_c_target(target_type: SqlSmallInt) -> bool {
 
 /// Converts a numeric column value to a floating-point C target.
 ///
-/// Returns [`ConvError::NotHandledHere`] when the source is not numeric or the
-/// target is not a floating-point C type.
+/// Returns [`ConvError::NotHandledHere`] only when the target is not a
+/// floating-point C type. A column with no numeric interpretation is
+/// [`ConvError::Restricted`], a numeric value too wide to represent is
+/// [`ConvError::OutOfRange`], and text that is not a valid number is
+/// [`ConvError::InvalidCharacterValue`].
 ///
 /// # Safety
 /// Same pointer contract as [`convert_integer_c`].
@@ -1471,8 +1488,9 @@ mod tests {
         .unwrap_err();
         assert_eq!(err, ConvError::OutOfRange);
 
-        // A magnitude filling all four limbs exceeds `i128::MAX`: also a
-        // representability failure, not an illegal cast.
+        // A magnitude filling all four limbs exceeds `i128::MAX`. It has no
+        // exact scaled form, so it reads as `f64` and lands far outside a
+        // 32-bit target: out of range, not an illegal cast.
         let err = unsafe {
             convert_integer_c(
                 &decimal(true, 0, vec![-1, -1, -1, -1]),
@@ -1483,6 +1501,25 @@ mod tests {
         }
         .unwrap_err();
         assert_eq!(err, ConvError::OutOfRange);
+
+        // The largest legal `decimal(38, 38)` magnitude (10^38 - 1) still fits
+        // in `i128`, so it must keep its exact form rather than trip either the
+        // range check above or the `f64` fallback.
+        let ok = unsafe {
+            convert_integer_c(
+                &decimal(
+                    true,
+                    38,
+                    vec![-1, 160_047_679, 1_518_781_562, 1_262_177_448],
+                ),
+                SQL_C_SLONG,
+                (&mut out as *mut i32).cast(),
+                &mut ind,
+            )
+        }
+        .unwrap();
+        assert_eq!(ok, ConvOk::Truncated);
+        assert_eq!(out, 0);
 
         // Zero-padded limbs past the fourth carry no magnitude, so this path
         // agrees with the string rendering instead of refusing the value.
@@ -1794,6 +1831,36 @@ mod tests {
         assert_eq!(err, ConvError::Restricted);
     }
 
+    /// The reachable wide-decimal case: the decoder validates limb count but
+    /// never the magnitude against `precision`, so 16 bytes of `0xFF` arrives
+    /// intact from a non-conforming server. It exceeds `i128::MAX`, so there is
+    /// no exact scaled form — but msodbcsql's `ConvertToFloat` accumulates all
+    /// four words into a double and succeeds, so a float target must too.
+    #[test]
+    fn wide_decimal_into_float_target_matches_msodbcsql() {
+        use mssql_tds::datatypes::decoder::DecimalParts;
+
+        let mut out: f64 = 0.0;
+        let mut ind: SqlLen = 0;
+        let ret = conv_f(
+            &ColumnValues::Decimal(DecimalParts {
+                is_positive: true,
+                scale: 0,
+                precision: 38,
+                int_parts: vec![-1, -1, -1, -1],
+            }),
+            SQL_C_DOUBLE,
+            (&mut out as *mut f64).cast(),
+            &mut ind,
+        )
+        .unwrap();
+        assert_eq!(ret, ConvOk::Exact);
+        assert_eq!(out, u128::MAX as f64);
+    }
+
+    /// A magnitude wider than four words has no `SQL_NUMERIC_STRUCT` (or wire)
+    /// counterpart, so there is no msodbcsql behavior to match. It is still a
+    /// numeric value we cannot represent, which is `22003`, not `07006`.
     #[test]
     fn oversized_decimal_into_float_target_is_out_of_range() {
         use mssql_tds::datatypes::decoder::DecimalParts;
--- a/mssql-odbc/src/api/fetch_convert.rs
+++ b/mssql-odbc/src/api/fetch_convert.rs
@@ -197,38 +197,44 @@ pub(crate) fn money_scaled(lsb: i32, msb: i32) -> i64 {
     (i64::from(lsb) & 0xFFFF_FFFF) | (i64::from(msb) << 32)
 }
 
-/// Interprets a column as a number, or `None` when the column type has no
-/// numeric interpretation.
-fn numeric_source(value: &ColumnValues) -> Option<NumericSource> {
+/// Interprets a column as a number.
+///
+/// [`ConvError::Restricted`] (`07006`) means the column type has no numeric
+/// interpretation at all; [`ConvError::OutOfRange`] (`22003`) means the column
+/// is numeric but the value cannot be represented here.
+fn numeric_source(value: &ColumnValues) -> Result<NumericSource, ConvError> {
     match value {
-        ColumnValues::TinyInt(x) => Some(NumericSource::Int(i128::from(*x))),
-        ColumnValues::SmallInt(x) => Some(NumericSource::Int(i128::from(*x))),
-        ColumnValues::Int(x) => Some(NumericSource::Int(i128::from(*x))),
-        ColumnValues::BigInt(x) => Some(NumericSource::Int(i128::from(*x))),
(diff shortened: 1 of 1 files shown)


mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -197,38 +197,44 @@ pub(crate) fn money_scaled(lsb: i32, msb: i32) -> i64 {
...
+        ColumnValues::Int(x) => Ok(NumericSource::Int(i128::from(*x))),
+        ColumnValues::BigInt(x) => Ok(NumericSource::Int(i128::from(*x))),
+        ColumnValues::Bit(b) => Ok(NumericSource::Int(i128::from(*b))),
+        ColumnValues::Real(x) => Ok(NumericSource::Float(f64::from(*x))),
+        ColumnValues::Float(x) => Ok(NumericSource::Float(*x)),
         // `DecimalParts` stores a base-2^32 little-endian magnitude. 38 digits
-        // fit in 4 words, and `magnitude` returns `None` for a wider (malformed)
-        // value rather than shifting past 128 bits.
+        // fit in 4 words; a wider magnitude, or one past `i128::MAX`, is a
+        // numeric value this converter cannot represent, not an illegal cast.
```
saurabh500: Worth capturing the parity basis right here. `.github/instructions/mssql-odbc.instructions.md` asks for an explicit matches / exceeds / diverges statement on any SQLSTATE mapping change, and this file already follows that convention at lines 246-249 and 318-320.

I checked, and this arm matches msodbcsql exactly. `Sql/Ntdbms/sqlncli/odbc/sqlcprot.h`:

```c
#define CVT_PREC     IDS_22_003   // if loss of precision occurred.
#define CVT_ILLEGAL  IDS_07_006   // if conversion is not supported.
```

`sqlccnvt.cpp` `ConvertToFixed` returns `CVT_PREC` from every integer range check and reserves `CVT_ILLEGAL` for its `default:` arm, which is the same split you are drawing here.

```suggestion
        // `DecimalParts` stores a base-2^32 little-endian magnitude. 38 digits
        // fit in 4 words; a wider magnitude, or one past `i128::MAX`, is a
        // numeric value this converter cannot represent, not an illegal cast.
        // Matches msodbcsql: `sqlccnvt.cpp` `ConvertToFixed` returns `CVT_PREC`
        // (`IDS_22_003`) for an out-of-range numeric, and reserves `CVT_ILLEGAL`
        // (`IDS_07_006`) for a source with no numeric reading at all.
```

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -262,8 +268,9 @@ fn numeric_source_or_parse(value: &ColumnValues) -> Result<NumericSource, ConvEr
 /// `numeric`, `money`, `smallmoney`), the floating-point columns, and character
 /// columns holding a numeric literal. A dropped fractional part is reported as
 /// [`ConvOk::Truncated`], text that is not a valid number as
-/// [`ConvError::InvalidCharacterValue`], and a column with no numeric
-/// interpretation as [`ConvError::Restricted`].
+/// [`ConvError::InvalidCharacterValue`], a numeric value too wide to represent
+/// as [`ConvError::OutOfRange`], and a column with no numeric interpretation as
+/// [`ConvError::Restricted`].
```
saurabh500, line 287: Good that this doc kept up with the behavior, but the sibling on `convert_float_c` (lines 351-354) needs the same treatment, and this PR widens the gap:

```rust
/// Returns [`ConvError::NotHandledHere`] when the source is not numeric or the
/// target is not a floating-point C type.
```

A non-numeric source returns `Restricted`, not `NotHandledHere` — pinned by `binary_source_into_numeric_target_is_restricted`. And this PR adds a third undocumented outcome to that path, `OutOfRange`, which the new test 20 lines below then covers.

Those lines are outside the diff so I cannot attach a suggestion, but:

```rust
/// Returns [`ConvError::NotHandledHere`] only when the target is not a
/// floating-point C type. A column with no numeric interpretation is
/// [`ConvError::Restricted`], a numeric value too wide to represent is
/// [`ConvError::OutOfRange`], and text that is not a valid number is
/// [`ConvError::InvalidCharacterValue`].
```

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -1462,7 +1469,20 @@ mod tests {
...
+        let err = unsafe {
+            convert_integer_c(
+                &decimal(true, 0, vec![-1, -1, -1, -1]),
+                SQL_C_SLONG,
+                (&mut out as *mut i32).cast(),
+                &mut ind,
+            )
+        }
+        .unwrap_err();
+        assert_eq!(err, ConvError::OutOfRange);
```
saurabh500, line 1503: The negative cases are well covered now, but nothing pins the boundary from the other side: that a *legal* max-precision value still converts. That is the value real servers actually send, and it is what a future tightening of this check would break first. 10^38 - 1 is about 9.99e37, comfortably under `i128::MAX` (about 1.70e38), so it has to keep working.

I ran this against your branch and it passes at 9025f0c.

```suggestion
        assert_eq!(err, ConvError::OutOfRange);

        // The largest legal `decimal(38, 38)` magnitude (10^38 - 1) still fits
        // in `i128`, so it must convert rather than trip the new range check.
        let ok = unsafe {
            convert_integer_c(
                &decimal(true, 38, vec![-1, 160_047_679, 1_518_781_562, 1_262_177_448]),
                SQL_C_SLONG,
                (&mut out as *mut i32).cast(),
                &mut ind,
            )
        }
        .unwrap();
        assert_eq!(ok, ConvOk::Truncated);
        assert_eq!(out, 0);
```

mssql-odbc/src/api/fetch_convert.rs
```diff
@@ -1774,6 +1794,27 @@ mod tests {
...
+                scale: 0,
+                precision: 38,
+                int_parts: vec![1, 0, 0, 0, 1],
+            }),
+            SQL_C_DOUBLE,
+            (&mut out as *mut f64).cast(),
+            &mut ind,
+        )
+        .unwrap_err();
+        assert_eq!(err, ConvError::OutOfRange);
```
saurabh500, line 1882: Two things about this test are worth a second look.

**It pins a divergence from msodbcsql.** `ConvertToFloat` in `sqlccnvt.cpp` handles a wide `SQL_C_NUMERIC` by accumulating all four 32-bit words into a double and succeeding; its only `CVT_PREC` exit on that path is an undersized input buffer, not magnitude. So where msodbcsql hands back a value, we now return `22003`. We could match it cheaply, too: `DecimalParts::to_f64` already falls back to `magnitude_big()` for exactly this case (`decoder.rs:2051-2055`), so the capability is sitting right there. Either route the float path through it, or add a row to the "Known divergences from msodbcsql" table in `mssql-odbc/docs/typed-columnar-fetch-plan.md` — the parity run skips these cases, so nothing else will surface it.

**The input cannot occur.** A 5-limb `int_parts` is rejected by both decode paths before it could reach here: `decoder.rs:701` raises `ProtocolError` above 4 limbs, and the Always Encrypted reader rejects a magnitude over 16 bytes at `security/encryption/cell.rs:355`. So this covers defensive-only code.

The case that *is* reachable over the wire is the 4-limb one you added on line 1478. The decoder validates limb count but never magnitude against `precision`, so 16 bytes of `0xFF` decodes cleanly from a non-conforming server. That one is currently only tested against the integer target — switching this test to that input would cover the float target for a state that can actually arrive.

saurabh500 (commented):
Traced the SQLSTATE decision back to the msodbcsql source before commenting. The core reclassification is right, and the blast radius is genuinely nil: the set of `Ok` values is byte-for-byte unchanged, both `ConvError` variants were already mapped by `finish_typed_conv`, and no caller branches on `Restricted` to select a fallback path. Tightly scoped fix.

### Parity determination

From `SqlClientDrivers/msodbcsql`, `master`. `Sql/Ntdbms/sqlncli/odbc/sqlcprot.h` defines the mapping:

```c
#define CVT_PREC     IDS_22_003   // if loss of precision occurred.
#define CVT_ILLEGAL  IDS_07_006   // if conversion is not supported.
```

| Case | msodbcsql (`sqlccnvt.cpp`) | this PR | Verdict |
| --- | --- | --- | --- |
| decimal out of range into an integer target | `ConvertToFixed` range checks return `CVT_PREC` | `22003` | **matches** |
| non-numeric source (binary, guid) | `default:` arm returns `CVT_ILLEGAL` | `07006` | **matches** |
| wide magnitude into `SQL_C_DOUBLE` | `ConvertToFloat` accumulates all four 32-bit words into a double and **succeeds**; its only `CVT_PREC` exit on that path is an undersized input buffer | `22003` error | **diverges** |

So the integer half is exact parity, which is worth stating in the description per `.github/instructions/mssql-odbc.instructions.md`. The float half is the one I would like your read on, in the comment on the new test.

Three inline comments, all "should fix" rather than blocking. One nit not worth its own thread: the `Decimal | Numeric` arm is only ever exercised through `Decimal`, so a future split of that arm would go uncaught.

Verified locally at 9025f0c: the full `mssql-odbc` lib suite passes (566), and `clippy --all-features --all-targets -D warnings` is clean.

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Thanks for pulling the actual `sqlcprot.h` / `sqlccnvt.cpp` basis — that's the citation this arm needed. Added, with one adjustment: your fourth comment led me to change this arm's behavior, so the comment now covers both halves of the split rather than describing everything past `i128::MAX` as unrepresentable.

The `CVT_PREC` / `CVT_ILLEGAL` split is cited as you had it, plus the `ConvertToFloat` four-word accumulation and the point that `SQL_NUMERIC_STRUCT` is itself four words — which is what makes the >4-word case have no msodbcsql counterpart. Statement: **matches** msodbcsql.

> [!NOTE]

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Good catch — the `NotHandledHere` claim was already wrong for a non-numeric source, and I did widen the gap by adding a third outcome. Applied your wording verbatim.

> [!NOTE]

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Agreed — every case here was a rejection, so nothing held the legal side of the boundary. Added. I recomputed the limb vector independently and it is right: 10^38 - 1 is `[-1, 160047679, 1518781562, 1262177448]` as i32 limbs, ≈9.99e37 against `i128::MAX` ≈1.70e38.

Only tweak is the comment, since the fallback added for your fourth point gives this test a second thing to guard: the value now has to avoid both the >4-word rejection and the `f64` fallback, i.e. it must keep its exact scaled form.

> [!NOTE]

mssql-odbc/src/api/fetch_convert.rs
David-Engel: Both points were right, and they resolve together. I took the parity route rather than the doc route, because it turned out not to need a `mssql-tds` change: `DecimalParts::to_f64` and `magnitude_big` are both private, but `magnitude()` is `pub` and returns `u128`, and the reachable case (4 words above `i128::MAX`) fits `u128` fine. Only the `i128::try_from` narrowing was throwing the value away.

So the decimal arm now keeps the exact scaled form when the magnitude fits `i128`, and otherwise degrades to the `f64` reading. Effects:

- Float target, 16 bytes of `0xFF`: succeeds at `u128::MAX as f64`, matching `ConvertToFloat`. No divergence left to record, so no table row.
- Integer target, same input: `to_i128_truncating` rejects it outside ±1.7e38, so it stays `22003` — what #235 asked for.
- Legal decimals: untouched, since 10^38 - 1 is below `i128::MAX`. Pinned by the boundary test from your other comment.

On the input being unreachable: I kept a test for the >4-word case but retargeted the one you flagged. `wide_decimal_into_float_target_matches_msodbcsql` uses the reachable 4-word input and asserts the parity behavior; `oversized_decimal_into_float_target_is_out_of_range` keeps the 5-word input for the defensive path. Worth noting `SQL_NUMERIC_STRUCT` is four words too, so >4 words has no msodbcsql counterpart to match — that is now in the code comment so the `22003` there is not mistaken for a divergence.

> [!NOTE]

saurabh500 (approved):
Verified the parity fix: msodbcsql's `ConvertToFloat` Horner accumulation (`Temp = Temp * 2^32 + word`) lands on exactly 2^128, which is the same value `u128::MAX as f64` rounds to — so the new test pins real parity rather than a coincidence, and `ConvOk::Exact` matches `CVT_NO_ERROR` on that path.
All four comments addressed, 567 tests pass, clippy and fmt clean at b9db4e57.

PR #286, opened 2026-08-14
Remove row-path transport dispatch and per-row boxing

### LINKED ISSUE #265: Remove `Box<dyn TdsTransport>` from the row path to unlock writer devirtualization

saurabh500:
## Description

Follow-up to merged #264. This removes both production row-path trait-object dispatch and the per-row boxed future.

### Design

- `TdsClient` stores a private `AnyTransport` enum. Production holds `NetworkTransport` inline; cfg-gated test, `test-util`, and fuzz builds retain `Box<dyn TdsTransport>` for custom transports.
- Native inherent methods on `AnyTransport` and `NetworkTransport` carry the concrete `RowWriter` through row receive/resume. Existing JS and Python callers use the generic `TdsClient::next_row_into<W>` without source changes.
- The public object-safe `ResultSet::next_row_into(&mut dyn RowWriter)` API remains unchanged and delegates to the generic method.
- `CancelHandle::run_until_cancelled` now uses an `Either` representation rather than an async state machine that stores the inner future in both branches.
- Successful production row decoding remains allocation-free at the transport boundary. Targeted boxing is limited to uncommon control-token and cancellation/recovery continuations.
- `NetworkTransport` still contains its runtime-selected `Box<dyn Stream>`, and cfg-gated dynamic test/fuzz transports remain boxed. This PR removes boxing from ordinary production row decoding, not every box in the connection architecture.

### Future-size budget

Production-relevant row futures are guarded by a **4096 B maximum**. The guard runs under `cfg(test)` and conservatively includes the dynamic transport fallback, so it is at least as strict as the production representation. All row-fetch tracing spans remain enabled.

### Paired measurement

A temporary harness, removed before commit, compared the old boxed/dynamic path with the new inline/native path using real TDS packet framing. Runs were interleaved in ABBA order across independent processes.

| Request budget | Median paired advantage |
|---|---:|
| none | **5.0%** |
| 30 seconds, with the then-current eager timeout | **10.1%** |

The timing deltas are measured; writer devirtualization and removal of the async-trait allocation are the inferred mechanism. The timeout result predates the lazy timeout arming now present from #275.

### Remaining dynamic boundaries

`ResultSet`, `TdsTransport`, and `TdsTokenStreamReader` remain object-safe/async-trait APIs for intentional trait-object callers and cfg-gated test/fuzz transports. Concrete production row fetches resolve inherent methods and do not cross those boundaries.

## Related Issues

Fixes #265

## Validation

| Check | Result |
|---|---|
| `cargo bfmt` | pass |
| `cargo bclippy` | pass (`-D warnings`) |
| `.\scripts\bfmt.ps1` / `.\scripts\bclippy.ps1` | pass, including excluded `mssql-py-core` |
| focused row-path behavior, cleanup, and future-size tests | pass |
| Linux `cfg(coverage)` future-size guard | pass |
| Azure and GitHub required checks | pass |
| Coverage | 92% diff, 91.7% overall |
| Review | approved after all feedback was addressed |

The full local `cargo btest` run retains the repository's environment-only certificate-fixture and live-SQL failures; the changed-path tests and required CI pass.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes — environment-only failures are documented above


mssql-tds/src/connection/transport/any_transport.rs
```diff
@@ -0,0 +1,493 @@
...
+                transport
+                    .read_active_plp_bytes(plp_state, remaining_request_timeout, cancel_handle, out)
+                    .await
+            }
+        }
+    }
+}
+
+#[async_trait]
+impl TdsTokenStreamReader for AnyTransport {
```
David-Engel: **Suggestion:** this impl looks like dead code in production. `dyn TdsTransport` now appears in exactly one place in the whole repo — the cfg-gated `Dynamic` variant on line 31 — so nothing ever reaches `AnyTransport` through a trait object. It exists only to satisfy the `TdsTransport` supertrait bound.

Measured with `cargo llvm-cov --lib -p mssql-tds` on this branch: `any_transport.rs` is **61.7% region / 55.9% function** coverage (15 of 34 functions never executed), and this impl is the bulk of it.

Consider making the seven `TdsTransport` methods inherent on `AnyTransport` and deleting both trait impls (~150 lines), which would also drop the `use TdsTransport as _;` import that had to be added to `cursor_ops.rs`.

mssql-tds/src/connection/transport/any_transport.rs
```diff
@@ -0,0 +1,493 @@
...
+            #[cfg(any(test, feature = "test-util", fuzzing))]
+            Self::Dynamic(transport) => {
+                transport
+                    .receive_token(context, remaining_request_timeout, cancel_handle)
+                    .await
+            }
+        }
+    }
+
+    #[cfg(not(any(test, feature = "test-util", fuzzing)))]
```
David-Engel, line 66: **Suggestion:** this is the repo's only `cfg(not(any(test, feature = "test-util", fuzzing)))` block, and neither lint/test alias compiles it:

- `cargo bclippy` passes `--all-features`, which enables `test-util`.
- `cargo btest` builds through the `test-util` self dev-dependency in `mssql-tds/Cargo.toml`.

So this arm is only ever type-checked by a plain release build, never linted. I confirmed `cargo clippy -p mssql-tds --lib` does compile it and is clean today — worth adding that as a lint leg so a future edit here can't land unlinted.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -2268,7 +2269,8 @@ impl TdsClient {
         self.return_values.push(return_value);
     }
 
-    #[instrument(skip(self), level = "info")]
+    // `#[instrument]` adds 88 B to this state machine and pushes the row-fetch
+    // future over its 4096 B budget. The per-row loop already emits an info event.
```
David-Engel: **Suggestion (take it or leave it):** the comment explains the tradeoff well and the future does land under budget either way. One thought though: `drain_rows` has a single caller (`advance()`, line 2748) and runs once per result set rather than per row, so it isn't really on the hot path the budget is protecting. If the span is useful for diagnosing drains, `Box::pin(self.drain_rows())` at that one call site would let you keep it without touching the budget. Entirely your call — if the span wasn't earning its keep, dropping it is fine.

Minor related note: the per-row `info!` inside the loop no longer logs the column count it used to.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3587,24 +3590,26 @@ impl TdsClient {
...
     /// `decode_into`, bypassing the intermediate `RowToken { all_values }`.
+    /// Concrete writers stay concrete through the production transport and
+    /// decode chain. Calling the object-safe [`ResultSet::next_row_into`] method
+    /// remains supported for callers that intentionally hold a trait object.
     #[instrument(skip(self, writer), level = "info")]
-    pub(crate) async fn get_next_row_into(
-        &mut self,
-        writer: &mut (dyn RowWriter + Send),
-    ) -> TdsResult<bool> {
+    pub async fn next_row_into<W>(&mut self, writer: &mut W) -> TdsResult<bool>
```
David-Engel, line 3607: **Suggestion:** pushing `abort_pending_prepare_capture()` down from the trait wrapper was the right move — the JS and Python call sites now bind this inherent method, so the abort had to follow them — and I checked that all six error paths in the new body cover it. The risk is that nothing enforces the invariant going forward: a single `?` added anywhere in this function silently drops it.

A one-line comment here would make it explicit, e.g. "every error return must call `abort_pending_prepare_capture()`; it can't be wrapped because the extra frame exceeds the future-size budget."

mssql-tds/src/io/token_stream.rs
```diff
@@ -199,6 +199,9 @@ pub(crate) trait TdsTokenStreamReader {
         cancel_handle: Option<&CancelHandle>,
     ) -> TdsResult<Tokens>;
 
+    // Object-safe fallbacks for cfg-gated dynamic transports. Production calls
+    // the native inherent methods on its concrete transport representation.
+    #[allow(dead_code)]
```
David-Engel: **Nit:** these could be `#[cfg_attr(not(any(test, feature = "test-util", fuzzing)), allow(dead_code))]`, so the `dead_code` lint still protects the configurations where these methods are actually supposed to be reached. As written, if the dynamic arm is ever removed the allows would silently hide a fully dead trait.

David-Engel (commented):
## Summary

Replaces `Box<dyn TdsTransport>` on `TdsClient` with an `AnyTransport` enum (inline `NetworkTransport` in production, cfg-gated `Box<dyn>` for test/fuzz), threads a concrete `W: RowWriter` through the row path, and converts `CancelHandle::run_until_cancelled` from an `async fn` to a combinator so the row future no longer needs `Box::pin`. The design is sound and I found no correctness defects. My findings are maintainability/process items — nothing merge-blocking.

## Verified locally (worktree at `5c21a879`)

- `cargo check -p mssql-tds`, `cargo check --workspace`, and `cargo clippy --workspace --all-features --all-targets` are all clean.
- 10 targeted tests pass, including both new `any_transport` tests, `next_row_into_is_idempotent_after_end_without_metadata`, and both future-size guards.
- `DynamicRowWriter` forwards all 25 `RowWriter` methods; the trait has no defaulted methods, so the compiler enforces completeness — no silent value-drop risk.
- The `run_until_cancelled` rewrite is behavior-preserving: `CancellationToken::run_until_cancelled` does no work until polled, and the `Either`/`Map` shape preserves `Some`/`None` -> `Ok`/`OperationCancelledError`.
- Reordering the end-of-set check ahead of the metadata check is exactly equivalent to the old `maybe_has_unread_rows()` gate (`!current_result_set_has_been_read_till_end`), and the flag initializes to `false`, so a pre-`execute` call still returns the "No metadata found" `UsageError`.
- Inlining the transport is cheap: `size_of::<NetworkTransport>() == 200`, `TdsClient == 960` (was ~776 with the fat pointer).
- The future-size guard runs under `cfg(test)`, where `AnyTransport::receive_row_into` returns `Either<network, dynamic>` — i.e. at least the production size. The guard is conservative rather than optimistic, which is the right direction.
- Both consumers (`mssql-js/src/connection.rs:215`, `mssql-py-core/src/cursor.rs:129`) pass concrete writers on a concrete `TdsClient`, so the devirtualization claim actually lands.

## Findings not anchored to a diff line

**Suggestion — behavior change worth confirming.** `TdsClient::get_next_row` now routes through `next_row_into`, so it aborts the pending prepare capture on error. `drain_rows`' only caller (`advance()`) already did this explicitly, so the net new surface is `metadata_retriever.rs:367/407/428`, which previously propagated a drain error without clearing `pending_capture`. It is the fail-safe direction, but was it intentional?

**Nit — `ResultSet::next_row_into` rustdoc.** `TdsClient::next_row_into` is now `pub` and shadows the trait method for concrete receivers. The inherent method documents the relationship; consider a matching one-liner on the trait method pointing at the inherent generic version, so readers arriving at the trait know the object-safe one is the fallback.

Blocking findings: none.

mssql-tds/src/connection/transport/any_transport.rs
saurabh500: Removed both dead trait impls from `AnyTransport` and made the remaining transport operations inherent, which also removes the trait import from `cursor_ops.rs`. The cfg-gated dynamic arm still forwards to its underlying `TdsTransport` object.

mssql-tds/src/connection/transport/any_transport.rs
saurabh500: Added an explicit production-configuration `mssql-tds --lib` clippy pass to both `scripts/bclippy.ps1` and `scripts/bclippy.sh`, before the all-features workspace pass. The pipeline invokes these scripts, so this cfg is now continuously linted.

mssql-tds/src/connection/tds_client.rs
saurabh500: Kept `drain_rows` uninstrumented because its sole caller, `advance`, already supplies the enclosing span, avoiding a once-per-result-set box solely for a nested span. Restored the useful column count on the per-row drain event without increasing the guarded future size.

mssql-tds/src/connection/tds_client.rs
saurabh500: Added the invariant comment at the start of `next_row_into`, including why cleanup cannot be centralized in another async frame. The existing error-path regression and future-size guard both still pass.

mssql-tds/src/io/token_stream.rs
saurabh500: Changed all four broad allows to the suggested production-only `cfg_attr`, so test/test-util/fuzz configurations continue to enforce dead-code reachability for their dynamic transport fallback.

scripts/bclippy.ps1
```diff
@@ -3,6 +3,13 @@
 
 $ErrorActionPreference = "Stop"
 
+Write-Host "Running cargo clippy on mssql-tds production configuration..."
+cargo clippy --frozen -p mssql-tds --lib -- -D warnings
```
saurabh500, line 7: Why is this needed? Why is line 14 not enough?

saurabh500: Line 14 passes `--all-features`, which enables `test-util`; that makes `cfg(not(any(test, feature = "test-util", fuzzing)))` false, so it never compiles the production-only method. The new default-feature `--lib` leg is complementary: during this fix it caught two unused imports and a dead trait method that the existing all-features pass did not see.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3615,21 +3626,26 @@ impl TdsClient {
             self.active_row_read_state,
             ActiveRowReadState::RowPaused(_) | ActiveRowReadState::PlpPaused(_)
         ) {
+            self.abort_pending_prepare_capture();
```
David-Engel: Two things here, both from `3b0a32a2` removing the `drain_active_row_if_needed()` call.

**Stale comment.** The block just above still ends with "An absent (`Idle`) row is fine — there is nothing left to skip, so it drains to a no-op below." There is no drain below anymore, so that last clause points at code that no longer exists.

**Latent hazard.** Dropping the drain is correct today, but only by elimination: `ActiveRowReadState` has exactly three variants, this `matches!` rejects two of them, so the state is necessarily `Idle` by the time we reach the decode loop. Nothing enforces that. A fourth variant would compile fine here and silently skip realignment, which on this path means decoding from a misaligned token stream.

Cheap fix that makes it structural — swap the `matches!` for an exhaustive `match`, so adding a variant becomes a compile error at exactly the place that needs to think about it:

```rust
match self.active_row_read_state {
    ActiveRowReadState::Idle => {}
    ActiveRowReadState::RowPaused(_) | ActiveRowReadState::PlpPaused(_) => {
        self.abort_pending_prepare_capture();
        return Err(UsageError(/* ... */));
    }
}
```

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3671,7 +3702,8 @@ impl TdsClient {
     /// [`read_row_column`](Self::read_row_column). Any previously positioned row
     /// is drained first; its remaining column bytes are read and discarded rather
     /// than returned to the caller.
-    #[instrument(skip(self), level = "info")]
+    // `#[instrument]` makes this target-dependent future exceed the 4096 B
+    // budget on Linux. Row positioning still emits its existing row/token events.
```
David-Engel: This one has almost no margin, and it looks avoidable.

Measured on Linux / rustc 1.95 by temporarily lowering `MAX` in `row_fetch_futures_stay_small`: `next_row_cursor` is **4056 B against a 4096 B budget** — 40 B, about 1%. Your table reports 3976 B, so the exact number is toolchain- and target-dependent, which is the concerning part: an incidental change or a compiler bump could fail this guard on `main`, and the pressure at that point will be to delete more instrumentation.

The fix is the change you already made in `next_row_into` in `3b0a32a2`. The loop below still awaits `self.handle_row_read_token(token)` inline, and like the push path that runs per control token, not per row — so boxing it costs nothing on the hot path:

```rust
if let Some(has_row) = Box::pin(self.handle_row_read_token(token)).await? {
```

I measured it on this branch:

| Variant | Size |
|---|---:|
| as committed | 4056 B |
| with `Box::pin(handle_row_read_token(..))` | **1400 B** |
| same, plus `#[instrument]` restored | **1488 B** |

So one line buys back roughly 2.6 KB of headroom *and* lets the span come back, rather than trading the span away for 40 B.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3587,24 +3593,29 @@ impl TdsClient {
...
-    #[instrument(skip(self, writer), level = "info")]
-    pub(crate) async fn get_next_row_into(
-        &mut self,
-        writer: &mut (dyn RowWriter + Send),
-    ) -> TdsResult<bool> {
+    /// Concrete writers stay concrete through the production transport and
+    /// decode chain. Calling the object-safe [`ResultSet::next_row_into`] method
+    /// remains supported for callers that intentionally hold a trait object.
+    // `#[instrument]` adds enough state to exceed the 4096 B budget once the
+    // lazy timeout future is inlined. Successful rows still emit `Row Received`.
```
David-Engel: **Question:** is this comment (and the matching one on `drain_rows`) still true after `3b0a32a2` boxed the control-token continuation?

I restored `#[instrument]` on both and re-measured on this branch: `next_row_into` comes to **3440 B** and `drain_rows` to **3552 B**, both comfortably under 4096. My baseline numbers are lower than yours across the board (I get 3352 B here where your table says 4016 B), so this may still hold on whatever platform you measured — I am not claiming the comments were wrong when written, only that the boxing commit may have made them obsolete.

Worth a re-measure on your side, because as it stands four `#[instrument]` spans came off the row API (`drain_rows`, the inherent and trait `next_row_into`, and `next_row_cursor`) and nothing spans row fetching anymore. If the budget is no longer the constraint, that observability is free to keep.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -5046,15 +5082,18 @@ mod tests {
         let next_row_cursor = std::mem::size_of_val(&client.next_row_cursor());
         let read_row_column = std::mem::size_of_val(&client.read_row_column(0));
         let drain_rows = std::mem::size_of_val(&client.drain_rows());
-        let get_next_row_into = std::mem::size_of_val(&client.get_next_row_into(&mut sink));
+        let next_row_into = std::mem::size_of_val(&client.next_row_into(&mut sink));
+        let next_row_into_dyn =
+            std::mem::size_of_val(&client.next_row_into(&mut sink as &mut (dyn RowWriter + Send)));
```
David-Engel, line 5108: **Nit:** this guard only ever runs under `cfg(test)`, where `AnyTransport::receive_row_into` returns the `Either<Left, Right>` wrapper rather than the production shape. `Either` is `max(arms) + tag`, so the measurement is conservative (test >= production) and the guard is still sound — but it never observes the thing it is protecting, and a future failure could come entirely from the dynamic arm.

A sentence in the test doc saying that would save the next person chasing a regression that cannot occur in a release build.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -2277,10 +2276,17 @@ impl TdsClient {
             // still expects to see — but here we are intentionally discarding
             // the rest of the result set, so realign the stream past that parked
             // row first. `drain_active_row` is a no-op when nothing is parked.
-            self.drain_active_row().await?;
+            self.drain_active_row_if_needed().await?;
             // Drain the current result set.
-            while let Some(row) = self.get_next_row().await? {
-                info!("Consuming row while draining result set {:?}", row.len());
+            let mut writer = DiscardRowWriter;
+            while self.next_row_into(&mut writer).await? {
```
David-Engel, line 2289: **Note, not a defect:** routing this through the shared `next_row_into` also picks up its `abort_pending_prepare_capture()` calls, which the old `get_next_row_into` did not have. So a drain failure now clears `pending_capture` and `pending_prepared_param_encryption` where it previously left them alone.

`drain_error_after_returned_handle_retains_prepared_metadata` still passes, so this looks intentional and harmless — just confirming it was a deliberate consequence rather than a side effect of the refactor.

David-Engel (commented):
Re-reviewed at `4bc6253a`. All five comments from my earlier pass are addressed properly — the dead `TdsTransport` impls on `AnyTransport` are gone, the production-config clippy leg is in both `bclippy` scripts, the invariant comment is in `next_row_into`, and the `dead_code` allows are narrowed with `cfg_attr`. The design is right: the `AnyTransport` enum keeps the test/fuzz flexibility while getting production off the trait object, and the paired measurement methodology is good.

**Verified locally** (Linux, rustc 1.95):

- `cargo clippy -p mssql-tds --lib -- -D warnings` is clean, so the new leg does what it says.
- `mssql-tds --lib`: 1666 pass, 4 fail — all four are the known missing cert-fixture failures, unrelated to this PR.
- `AnyTransport` is 200 B and `TdsClient` is 960 B, so inlining `NetworkTransport` costs about 192 B on the client. Non-issue.
- `DynamicRowWriter` forwards all 25 `RowWriter` methods, and the trait has no defaults, so a future method addition is a compile error rather than a silent drop. Complete.
- `maybe_has_unread_rows()` is exactly `!current_result_set_has_been_read_till_end`, so collapsing the old `ResultSet::next_row_into` guard into the new first check is behavior-preserving, and reordering end-of-set ahead of the metadata check restores the old *public* behavior on the inherent method.

Two things I would like to see before merge: the stale comment / variant-exhaustiveness point on `next_row_into`, and the `next_row_cursor` headroom (details inline). The rest are questions and nits.

CI is green on every leg, `Fixes #265` is linked, and the checklist is accurate.

saurabh500:
In such PRs I think too many details about numbers are tripping bots and they are getting into a pedantic discussion about these many bytes here and those many bytes there. 
For me the 4096 byte cap for Future sizes in hot paths, is what really matters. I need to make sure that the PR description is void of too many details about these numbers

mssql-tds/src/connection/tds_client.rs
saurabh500: Re-measured and restored all four spans in `e74d39a1`. Under Linux with `cfg(coverage)`, `next_row_cursor` is 1488 B, `next_row_into` 3440 B, and `drain_rows` 3552 B; the largest guarded future is 3680 B.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Fixed in `e74d39a1`: the stale drain wording is gone, and the state check is now an exhaustive `match` over `ActiveRowReadState`, so adding a variant requires an explicit alignment decision here.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Fixed in `e74d39a1`: cursor control-token handling now uses the targeted box and the cursor span is restored. `next_row_cursor` measures 1488 B on both Windows and Linux with `cfg(coverage)`, down from the reported 4056 B.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Confirmed intentional: routing through the shared path ensures any row/drain error clears both pending prepare-capture fields rather than leaving stale state for a later command. The prepared-metadata cleanup regression tests pass with this behavior.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Documented this in `e74d39a1`: the guard now explicitly says its `cfg(test)` measurement conservatively includes the dynamic `Either` representation and can fail because of the test-only arm.

> [!NOTE]

saurabh500:
@David-Engel, this is ready for review again. All feedback is addressed and the required checks are passing.

David-Engel (approved):
Approving at `55ac6b1c`. `e74d39a1` addresses all five comments, and the outcome is better than what I asked for.

- The `matches!` is now an exhaustive `match` over `ActiveRowReadState` with an explicit `Idle => {}` arm, so a future variant forces an alignment decision at the right place. The stale "drains to a no-op below" wording is gone from both the inline comment and the rustdoc.
- `next_row_cursor` boxes the control-token continuation, matching what `3b0a32a2` already did on the push path.
- **All four `#[instrument]` spans are back** — `drain_rows`, the inherent `next_row_into`, the `ResultSet` impl, and `next_row_cursor`. That was the outcome I actually cared about; the byte accounting was only ever the means to it, and it is good that it turned out not to be a real constraint.
- The guard test now documents that its `cfg(test)` measurement includes the dynamic `Either` arm and can fail for test-only reasons.
- Confirmed the `drain_rows` prepare-capture abort is deliberate.

**Verified locally on this head** (Linux, rustc 1.95):

- `cargo clippy -p mssql-tds --lib -- -D warnings` — clean.
- `mssql-tds --lib`: 1669 pass, 4 fail — the known missing cert-fixture failures, unrelated to this PR.
- Every guarded future is under the 4096 B cap *with all spans restored*:

| Future | Size |
|---|---:|
| `next_row_cursor` | 1488 B |
| `read_row_column` | 3752 B |
| `drain_rows` | 3600 B |
| `next_row_into` | 3488 B |
| `next_row_into` (dyn writer) | 3640 B |
| `read_active_plp_chunk` | 1456 B |

`read_row_column` is the tightest at 3752 B, but it is untouched by this PR — it grew from the `main` merge, not from these changes. Worth keeping an eye on separately.

Nice piece of work: the `AnyTransport` enum gets production off the trait object without giving up the test and fuzz transports, the row path is genuinely allocation-free at the transport boundary now, and the paired ABBA measurement is the right way to make the performance claim.

Only `coverage-report` is still running; everything else is green. Please confirm it passes before merging.

PR #288, opened 2026-08-14
[py-core] Async Connection Foundation

subrata-ms:
This pull request introduces a new `async_session` module to manage connection lifecycle and operation state for asynchronous SQL Server connections. It also expands the async connection and cursor APIs to support advanced transaction handling and improves logging and testing coverage. The most important changes are grouped below:

### Async connection/session state management

* Added a new `async_session` module (`mssql-py-core/src/async_session.rs`) that defines types and logic for managing connection lifecycle (`Open`, `Closing`, `Closed`, `Broken`), operation state, and unique cursor/operation IDs. This includes thread-safe state transitions and test coverage for lifecycle handling.
* Updated the async cursor (`PyAsyncCursor` in `async_cursor.rs`) to store and use references to the new session state, autocommit flag, cursor ID, and default query timeout, enabling future support for transaction semantics and cancellation. [[1]](diffhunk://#diff-85fa845be93caab797168826f6e49584a85ce9577adb715c7dc051dd4919a179R26-R34) [[2]](diffhunk://#diff-85fa845be93caab797168826f6e49584a85ce9577adb715c7dc051dd4919a179R54-R128)
* Registered the new module in the crate root (`lib.rs`).

### Async connection and cursor API improvements

* Added a `timeout` property to `PyAsyncCursor` to expose the parent connection's default query timeout.
* Added detailed TODO comments outlining planned async transaction and operation ownership semantics in `async_cursor.rs`.

### Logging and tracing improvements

* Refactored the tracing bridge in `python_logger_adapter.rs` to provide an owned tracing dispatch suitable for async work, improving Python-side logging integration. [[1]](diffhunk://#diff-566eb3ca3879cbd89dc744f71d4cc44f93569799c5ba18b43e39e1572fc96a6bR144-R153) [[2]](diffhunk://#diff-566eb3ca3879cbd89dc744f71d4cc44f93569799c5ba18b43e39e1572fc96a6bL156-R166)

### Test coverage and regression guards

* Extended `test_async_connection.py` with new tests for logging behavior, autocommit defaults, connection lifecycle, and regression guards for connection and transaction operations. Introduced a `RecordingLogger` test helper. [[1]](diffhunk://#diff-41de74d66542d63b27acc704e0f56887c14b42840b640f6a5cef4c45ef4a242fL4-R5) [[2]](diffhunk://#diff-41de74d66542d63b27acc704e0f56887c14b42840b640f6a5cef4c45ef4a242fR17-R24) [[3]](diffhunk://#diff-41de74d66542d63b27acc704e0f56887c14b42840b640f6a5cef4c45ef4a242fL73-R179) [[4]](diffhunk://#diff-41de74d66542d63b27acc704e0f56887c14b42840b640f6a5cef4c45ef4a242fL108-R209)

These changes lay the groundwork for robust async transaction and operation management in the driver, and improve testability and observability of async connection behavior.

ADO work item: https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/46959, https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/46960, https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47185
## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

### DIFF
--- a/mssql-py-core/src/async_connection.rs
+++ b/mssql-py-core/src/async_connection.rs
@@ -186,6 +186,24 @@ impl PyAsyncConnection {
         })
     }
 
+    /// Async context manager entry. Resolves to `self` with no I/O.
+    fn __aenter__<'py>(slf: Py<Self>, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
+        pyo3_async_runtimes::tokio::future_into_py(py, async move { Ok(slf) })
+    }
+
+    /// Async context manager exit. Delegates to `close()`; exception info is
+    /// ignored and never suppressed (`close()` resolves to `None`, which is
+    /// falsy per Python's `__aexit__` contract).
+    fn __aexit__<'py>(
+        &mut self,
+        py: Python<'py>,
+        _exc_type: &Bound<'_, PyAny>,
+        _exc_val: &Bound<'_, PyAny>,
+        _exc_tb: &Bound<'_, PyAny>,
+    ) -> PyResult<Bound<'py, PyAny>> {
+        self.close(py)
+    }
+
     /// Commit the current transaction. If none is open, surfaces SQL Server 3902.
     fn commit<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
         // Clone the Arc so the future is `'static + Send`.
--- a/mssql-py-core/src/async_connection.rs
+++ b/mssql-py-core/src/async_connection.rs
@@ -3,23 +3,9 @@
 
 //! Asynchronous connection API for the Core TDS backend.
 //!
-//! # ⚠️ Preview API — unstable
+//! Preview API — unstable. First use emits a `FutureWarning`.
 //!
-//! The types and methods in this module are **not** part of the stable
-//! `mssql-py-core` surface. Signatures, error behavior, and internal
-//! semantics may change without notice in any release. First use in a
-//! Python process emits a [`FutureWarning`] via `warnings.warn`.
-//!
-//! Sibling of `connection.rs` (the synchronous surface). Every type defined
-//! here submits its I/O to the shared process-wide Tokio runtime via
-//! [`crate::async_runtime`] and returns Python awaitables through
-//! `pyo3_async_runtimes::tokio::future_into_py`, so callers can `await` the
-//! results from `asyncio`.
-//!
-//! Invariant: one async connection maps to exactly one async cursor, one
-//! `TdsClient`, and one TDS wire session.
-//!
-//! [`FutureWarning`]: https://docs.python.org/3/library/exceptions.html#FutureWarning
+//! Invariant: one async connection ↔ one async cursor ↔ one `TdsClient`.
 
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -36,8 +22,7 @@ use crate::async_cursor::PyAsyncCursor;
 use crate::connection::PyCoreConnection;
 use crate::python_logger_adapter::scoped_tracing_bridge;
 
-/// Emit a `FutureWarning` the first time any async API is exercised in this
-/// process. Silenceable by callers via `warnings.filterwarnings(...)`.
+/// One-shot `FutureWarning` per process; silenceable via `warnings.filterwarnings`.
 static PREVIEW_WARNED: AtomicBool = AtomicBool::new(false);
 
 fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
@@ -58,45 +43,23 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
 
 /// Asynchronous Python connection backed by the Core TDS client.
 ///
-/// # ⚠️ Preview API — unstable
-///
-/// Preview surface: API, method signatures, error behavior, and internal
-/// semantics may change without notice in minor releases. Do not depend on
-/// it from production code.
-///
-/// Instances are created via [`PyAsyncConnection::connect`], which returns a
-/// Python awaitable. The awaitable resolves on the caller's `asyncio` loop
-/// once the TCP + TLS + login handshake has completed on the shared Tokio
-/// runtime.
+/// Preview API — unstable.
 ///
 /// TODO(User Story 47180 [mssql-python] Cancel API and Cancellation Bridge):
 /// cancellation of a suspended `commit`, `rollback`, or `close` future can
-/// desync the TDS byte stream (bytes written, response not yet read), so a
-/// subsequent operation on the same connection may read a stale response
-/// and corrupt the wire. Callers must not cancel these awaitables against a
-/// connection they intend to keep using. Cancellation-safe semantics are
-/// tracked at
-/// <https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47180>.
+/// desync the TDS byte stream. Callers must not cancel these awaitables
+/// against a connection they intend to keep using.
+/// <https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47180>
 #[pyclass]
 pub struct PyAsyncConnection {
-    /// Wrapped in `Option` so `close()` can take ownership and drop the
-    /// client; wrapped in `Arc<tokio::sync::Mutex<...>>` so the (upcoming)
-    /// async cursor and connection-level lifecycle methods can share access
-    /// across `.await` points without corrupting the TDS byte stream.
+    /// `Option` so `close()` can `take()`; `Arc<Mutex<>>` for cursor sharing.
     tds_client: Option<Arc<Mutex<TdsClient>>>,
 }
 
 #[pymethods]
 impl PyAsyncConnection {
-    /// Establish a TDS connection asynchronously.
-    ///
-    /// ```python
-    /// conn = await PyAsyncConnection.connect(client_context_dict, python_logger=None)
-    /// ```
-    ///
-    /// Dictionary parsing runs synchronously on the calling thread (it needs
-    /// the GIL). The network handshake is submitted to the shared Tokio
-    /// runtime and driven concurrently with the caller's asyncio loop.
+    /// Establish a TDS connection. Dict parsing is synchronous; the network
+    /// handshake runs on the shared Tokio runtime.
     #[classmethod]
     #[pyo3(signature = (client_context_dict, python_logger=None))]
     fn connect<'py>(
@@ -110,8 +73,6 @@ impl PyAsyncConnection {
         let _guard = python_logger
             .map(|logger| scoped_tracing_bridge(Arc::new(logger.clone().unbind()), file!()));
 
-        // Preview API: emit a one-shot FutureWarning so callers see the
-        // instability signal at runtime.
         emit_preview_warning(py)?;
 
         tracing::info!("PyAsyncConnection::connect: initiating async connection");
@@ -161,26 +122,10 @@ impl PyAsyncConnection {
         })
     }
 
-    /// Close the TDS connection asynchronously.
-    ///
-    /// ```python
-    /// await conn.close()
-    /// ```
-    ///
-    /// Sends the TDS logout token and tears down the underlying transport.
-    /// The awaitable is submitted to the shared Tokio runtime so the calling
-    /// asyncio loop stays unblocked while the graceful shutdown runs.
-    ///
-    /// Idempotent: awaiting `close()` on an already-closed connection
-    /// resolves immediately with no I/O. If the graceful shutdown itself
-    /// errors, the error is logged at `warn` level and the connection is
-    /// still considered closed — the OS closes the socket on drop either
-    /// way, so we never leak the resource.
+    /// Close the connection. Idempotent. Shutdown errors are logged and swallowed.
     fn close<'py>(&mut self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
         tracing::info!("PyAsyncConnection::close: initiating close");
-        // Detach the client from `self` synchronously (while `&mut self` is
-        // valid) so the future can own it for `'static + Send`. Subsequent
-        // method calls on this connection will see `tds_client == None`.
+        // `take()` before spawning: gives the future 'static ownership; marks conn closed.
         let client_opt = self.tds_client.take();
 
         pyo3_async_runtimes::tokio::future_into_py(py, async move {
@@ -194,9 +139,7 @@ impl PyAsyncConnection {
             );
             let mut guard = client.lock().await;
             if let Err(e) = guard.close_connection().await {
-                // Match sync-path semantics: log and swallow. The connection
-                // is treated as closed regardless — the transport will be
-                // dropped when the Arc's last reference goes away.
+                // Log and swallow; connection is closed regardless.
                 tracing::warn!(
                     "PyAsyncConnection::close: error during graceful shutdown: {}",
                     e
@@ -207,23 +150,9 @@ impl PyAsyncConnection {
         })
     }
 
-    /// Commit the current TDS transaction asynchronously.
-    ///
-    /// ```python
-    /// await conn.commit()
-    /// ```
-    ///
-    /// Sends a TM_COMMIT (Transaction Manager COMMIT) request over the wire
-    /// and awaits the server's DONE token. Raises `RuntimeError`
-    /// synchronously if the connection has already been closed.
-    ///
-    /// If no transaction is currently open on the server, the commit will
-    /// fail with the server's own error (SQL Server 3902 — "The COMMIT
-    /// TRANSACTION request has no corresponding BEGIN TRANSACTION").
+    /// Commit the current transaction. If none is open, surfaces SQL Server 3902.
     fn commit<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
-        // Clone the Arc synchronously so the future is `'static + Send`
-        // without borrowing `self`. Only a shared borrow is required —
-        // nothing on `self` is mutated here.
+        // Clone the Arc so the future is `'static + Send`.
         let client = self
             .tds_client
             .as_ref()
@@ -243,23 +172,9 @@ impl PyAsyncConnection {
         })
     }
 
-    /// Roll back the current TDS transaction asynchronously.
-    ///
-    /// ```python
-    /// await conn.rollback()
-    /// ```
-    ///
-    /// Sends a TM_ROLLBACK (Transaction Manager ROLLBACK) request over the
-    /// wire and awaits the server's DONE token. Raises `RuntimeError`
-    /// synchronously if the connection has already been closed.
-    ///
-    /// If no transaction is currently open on the server, the rollback will
-    /// fail with the server's own error (SQL Server 3903 — "The ROLLBACK
-    /// TRANSACTION request has no corresponding BEGIN TRANSACTION").
+    /// Roll back the current transaction. If none is open, surfaces SQL Server 3903.
     fn rollback<'py>(&self, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
-        // Clone the Arc synchronously so the future is `'static + Send`
-        // without borrowing `self`. Only a shared borrow is required —
-        // nothing on `self` is mutated here.
+        // Clone the Arc so the future is `'static + Send`.
         let client = self
             .tds_client
             .as_ref()
@@ -279,23 +194,8 @@ impl PyAsyncConnection {
         })
     }
 
-    /// Create an async cursor bound to this connection.
-    ///
-    /// ```python
-    /// cur = conn.cursor()
-    /// await cur.execute("SELECT 1")
-    /// ```
-    ///
-    /// This method does not perform I/O — it simply hands out a new
-    /// [`PyAsyncCursor`] that shares the connection's `TdsClient` via an
-    /// `Arc<tokio::sync::Mutex<_>>`. Following DB-API 2.0, `cursor()` is a
-    /// synchronous call; only the cursor's execute/fetch methods will be
-    /// awaitable.
-    ///
-    /// Raises `RuntimeError` if the connection has already been closed.
-    /// A second cursor may be created on the same connection, but both
-    /// cursors share one TDS wire session and serialize on the same async
-    /// mutex — matching the non-MARS TDS session model.
+    /// Sync per DB-API 2.0. A second cursor is allowed; both share the same
+    /// TDS session and serialize on the same async mutex.
     fn cursor(&self) -> PyResult<PyAsyncCursor> {
         let client = self
             .tds_client
--- a/mssql-py-core/src/async_connection.rs
+++ b/mssql-py-core/src/async_connection.rs
@@ -41,6 +41,16 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
     Ok(())
 }
 
+/// Map a TDS error to a Python exception with per-operation context.
+///
+/// TODO(User Story 47181): map TdsError to a DB-API-compliant exception,
+/// preserving SQLSTATE + server error number.
+/// <https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47181>
+fn map_tds_error(op: &str, user_msg: &str, e: impl std::fmt::Display) -> PyErr {
+    tracing::error!("PyAsyncConnection::{op}: failed: {e}");
+    PyRuntimeError::new_err(format!("{user_msg}: {e}"))
+}
+
 /// Asynchronous Python connection backed by the Core TDS client.
 ///
 /// Preview API — unstable.
@@ -108,11 +118,7 @@ impl PyAsyncConnection {
             let client = provider
                 .create_client(context, &datasource, None)
                 .await
-                .map_err(|e| {
-                    tracing::error!("PyAsyncConnection::connect: failed: {}", e);
-                    // TODO(User Story 47181): map TdsError to a DB-API-compliant exception, preserving SQLSTATE + server error number.
-                    PyRuntimeError::new_err(format!("Failed to connect to SQL Server: {e}"))
-                })?;
+                .map_err(|e| map_tds_error("connect", "Failed to connect to SQL Server", e))?;
 
             tracing::info!("PyAsyncConnection::connect: connection established");
             Python::attach(|py| {
@@ -156,6 +162,14 @@ impl PyAsyncConnection {
     /// Inverse of `.closed`, provided for sync-path parity with `PyCoreConnection`.
     fn is_connected(&self) -> bool {
         self.tds_client.is_some()
+    }
+
+    fn __repr__(&self) -> &'static str {
+        if self.tds_client.is_none() {
+            "PyAsyncConnection(closed)"
+        } else {
+            "PyAsyncConnection(connected)"
+        }
     }
 
     /// Close the connection. Idempotent. Shutdown errors are logged and swallowed.
@@ -216,11 +230,10 @@ impl PyAsyncConnection {
         pyo3_async_runtimes::tokio::future_into_py(py, async move {
             tracing::info!("PyAsyncConnection::commit: sending TM_COMMIT");
             let mut guard = client.lock().await;
-            guard.commit_transaction(None, None).await.map_err(|e| {
-                tracing::error!("PyAsyncConnection::commit: failed: {}", e);
-                // TODO(User Story 47181): map TdsError to a DB-API-compliant exception, preserving SQLSTATE + server error number.
-                PyRuntimeError::new_err(format!("Commit failed: {e}"))
-            })?;
+            guard
+                .commit_transaction(None, None)
+                .await
+                .map_err(|e| map_tds_error("commit", "Commit failed", e))?;
             tracing::info!("PyAsyncConnection::commit: transaction committed");
             Python::attach(|py| Ok(py.None()))
         })
@@ -238,11 +251,10 @@ impl PyAsyncConnection {
         pyo3_async_runtimes::tokio::future_into_py(py, async move {
             tracing::info!("PyAsyncConnection::rollback: sending TM_ROLLBACK");
             let mut guard = client.lock().await;
-            guard.rollback_transaction(None, None).await.map_err(|e| {
-                tracing::error!("PyAsyncConnection::rollback: failed: {}", e);
-                // TODO(User Story 47181): map TdsError to a DB-API-compliant exception, preserving SQLSTATE + server error number.
-                PyRuntimeError::new_err(format!("Rollback failed: {e}"))
-            })?;
+            guard
+                .rollback_transaction(None, None)
+                .await
+                .map_err(|e| map_tds_error("rollback", "Rollback failed", e))?;
             tracing::info!("PyAsyncConnection::rollback: transaction rolled back");
             Python::attach(|py| Ok(py.None()))
         })
--- a/mssql-py-core/src/async_connection.rs
+++ b/mssql-py-core/src/async_connection.rs
@@ -54,6 +54,10 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
 pub struct PyAsyncConnection {
     /// `Option` so `close()` can `take()`; `Arc<Mutex<>>` for cursor sharing.
     tds_client: Option<Arc<Mutex<TdsClient>>>,
+    /// Default query timeout (seconds) applied to cursors created from this
+    /// connection. `0` = no timeout, per pyodbc/ODBC `SQL_ATTR_QUERY_TIMEOUT`.
+    /// Pure Python-side state — the setter performs no I/O.
+    default_query_timeout: u32,
 }
 
 #[pymethods]
@@ -116,10 +120,30 @@ impl PyAsyncConnection {
                     py,
                     PyAsyncConnection {
                         tds_client: Some(Arc::new(Mutex::new(client))),
+                        default_query_timeout: 0,
                     },
                 )
             })
         })
+    }
+
+    /// Default query timeout (seconds) inherited by cursors created from this
+    /// connection. `0` means no timeout.
+    #[getter]
+    fn timeout(&self) -> u32 {
+        self.default_query_timeout
+    }
+
+    /// Set the default query timeout (seconds) for future cursors. Existing
+    /// cursors and in-flight queries are unaffected. Negative values are
+    /// rejected by PyO3's `u32` extractor (`OverflowError`).
+    #[setter]
+    fn set_timeout(&mut self, value: u32) {
+        tracing::info!(
+            "PyAsyncConnection::set_timeout: default query timeout set to {}s",
+            value
+        );
+        self.default_query_timeout = value;
     }
 
     /// Close the connection. Idempotent. Shutdown errors are logged and swallowed.
@@ -202,6 +226,6 @@ impl PyAsyncConnection {
             .as_ref()
             .ok_or_else(|| PyRuntimeError::new_err("Connection is closed"))?
             .clone();
-        Ok(PyAsyncCursor::new(client))
+        Ok(PyAsyncCursor::new(client, self.default_query_timeout))
     }
 }
--- a/mssql-py-core/src/async_connection.rs
+++ b/mssql-py-core/src/async_connection.rs
@@ -5,22 +5,25 @@
 //!
 //! Preview API — unstable. First use emits a `FutureWarning`.
 //!
-//! Invariant: one async connection ↔ one async cursor ↔ one `TdsClient`.
+//! Invariant: one `TdsClient` and one TDS wire session per async connection.
+//! All cursors share that client and serialize access through the async mutex.
 
+use std::future::Future;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, Ordering};
 
-use pyo3::exceptions::{PyFutureWarning, PyRuntimeError};
+use pyo3::exceptions::{PyFutureWarning, PyOverflowError, PyRuntimeError, PyValueError};
 use pyo3::prelude::*;
 use pyo3::types::{PyDict, PyType};
 use tokio::sync::Mutex;
+use tracing::instrument::WithSubscriber;
 
 use mssql_tds::connection::tds_client::TdsClient;
 use mssql_tds::connection_provider::tds_connection_provider::TdsConnectionProvider;
 
 use crate::async_cursor::PyAsyncCursor;
 use crate::connection::PyCoreConnection;
-use crate::python_logger_adapter::scoped_tracing_bridge;
+use crate::python_logger_adapter::python_logger_dispatch;
 
 /// One-shot `FutureWarning` per process; silenceable via `warnings.filterwarnings`.
 static PREVIEW_WARNED: AtomicBool = AtomicBool::new(false);
@@ -51,6 +54,16 @@ fn map_tds_error(op: &str, user_msg: &str, e: impl std::fmt::Display) -> PyErr {
     PyRuntimeError::new_err(format!("{user_msg}: {e}"))
 }
 
+async fn with_optional_dispatch<F>(future: F, dispatch: Option<tracing::Dispatch>) -> F::Output
+where
+    F: Future,
+{
+    match dispatch {
+        Some(dispatch) => future.with_subscriber(dispatch).await,
+        None => future.await,
+    }
+}
+
 /// Asynchronous Python connection backed by the Core TDS client.
 ///
 /// Preview API — unstable.
@@ -64,6 +77,7 @@ fn map_tds_error(op: &str, user_msg: &str, e: impl std::fmt::Display) -> PyErr {
 pub struct PyAsyncConnection {
     /// `Option` so `close()` can `take()`; `Arc<Mutex<>>` for cursor sharing.
     tds_client: Option<Arc<Mutex<TdsClient>>>,
+    tracing_dispatch: Option<tracing::Dispatch>,
     /// Default query timeout (seconds) applied to cursors created from this
     /// connection. `0` = no timeout, per pyodbc/ODBC `SQL_ATTR_QUERY_TIMEOUT`.
     /// Pure Python-side state — the setter performs no I/O.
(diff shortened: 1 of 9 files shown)


mssql-py-core/src/async_connection.rs
```diff
@@ -55,115 +41,141 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
         client_context_dict: &Bound<'_, PyDict>,
+        python_logger: Option<&Bound<'_, PyAny>>,
     ) -> PyResult<Bound<'py, PyAny>> {
         let py = cls.py();
 
-        // Preview API: emit a one-shot FutureWarning so callers see the
-        // instability signal at runtime.
+        // `DefaultGuard` is `!Send`, so its coverage ends when this method returns.
+        let _guard = python_logger
+            .map(|logger| scoped_tracing_bridge(Arc::new(logger.clone().unbind()), file!()));
```
David-Engel: **Blocking:** this bridge does not cover any of the async work, including the connect-failure log.

`scoped_tracing_bridge` calls `tracing::subscriber::set_default`, which is **thread-local** and is undone when the guard drops at the end of `connect`. The `future_into_py` body runs later, on a Tokio worker thread. So of the events in `connect`, the Python logger receives only the four synchronous pre-flight `info!` lines, and never receives:

- `"connection established"` (L123)
- `map_tds_error("connect", ...)`'s `tracing::error!` (L50) - the handshake-failure diagnostic, which is the whole reason a caller passes a logger

This is the opposite of the sync path it is modeled on: `PyCoreConnection::new` uses `runtime.block_on` on the same thread, so its `tracing::error!("Failed to connect to SQL Server: {}", e)` *is* inside the guard. Separately, `close`/`commit`/`rollback`/`cursor` take no logger and are never bridged at all, so the connection is unlogged for its entire lifetime after `connect` returns.

Concrete fix - propagate a `Dispatch` across the await point instead of a thread-local guard. Add alongside the existing helpers in `python_logger_adapter.rs`:

```rust
pub fn tracing_bridge_dispatch(logger: Arc<Py<PyAny>>, module_name: &'static str) -> tracing::Dispatch {
    let layer = PythonLoggerLayer::new(logger, module_name)
        .with_filter(filter_fn(|m| m.target().starts_with("mssql_py_core")));
    tracing::Dispatch::new(Registry::default().with(layer))
}
```

then here:

```rust
use tracing::instrument::WithSubscriber;

let dispatch = python_logger
    .map(|l| tracing_bridge_dispatch(Arc::new(l.clone().unbind()), file!()));
// ... sync tracing calls stay inside a set_default guard as today ...
pyo3_async_runtimes::tokio::future_into_py(py, async move {
    let fut = async move { /* existing body */ };
    match dispatch { Some(d) => fut.with_subscriber(d).await, None => fut.await }
})
```

`Dispatch` is `Clone + Send + Sync`, so it can also be stored on the pyclass and reused by `close`/`commit`/`rollback` - which is what makes the parameter actually useful.

If that is too much for this PR, I would accept: drop the `python_logger` param for now, or keep it and (a) state the limitation in the *Python-visible* docstring rather than only a code comment, and (b) file a follow-up work item. What I would rather not ship is a `python_logger=` that silently swallows the failure path.

Supporting point: there is **no test** that passes a `python_logger`. A test asserting which records reach `py_core_log` for both a successful and a failed connect would have caught this, and is the right regression guard for whichever fix you pick.

mssql-py-core/src/async_connection.rs
```diff
@@ -55,115 +41,141 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
+
         emit_preview_warning(py)?;
 
+        tracing::info!("PyAsyncConnection::connect: initiating async connection");
```
David-Engel: **Nit:** `"initiating async connection"` is immediately followed by `"extracting client context"` at the same level - the first adds nothing. Also, the sync path uses `debug!` for the dict conversion; `info!` for every step is chatty on a per-connection path.

mssql-py-core/src/async_connection.rs
```diff
@@ -55,115 +41,141 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
+    /// cursors and in-flight queries are unaffected. Negative values are
+    /// rejected by PyO3's `u32` extractor (`OverflowError`).
+    #[setter]
+    fn set_timeout(&mut self, value: u32) {
+        tracing::info!(
+            "PyAsyncConnection::set_timeout: default query timeout set to {}s",
+            value
+        );
+        self.default_query_timeout = value;
+    }
```
David-Engel, line 216: **Suggestion:** negative values raise `OverflowError` here, but `mssql_python.Connection.timeout`'s setter raises `ValueError("Timeout cannot be negative")` (and `TypeError` for non-int). `test_timeout_setter_rejects_negative` locks `OverflowError` in.

If the Python wrapper will always sit on top, this is fine - but worth a one-line note in the docstring so nobody is surprised when the two surfaces disagree. Otherwise take `i64` and raise `ValueError` explicitly for parity.

mssql-py-core/src/async_connection.rs
```diff
@@ -177,35 +189,38 @@ impl PyAsyncConnection {
...
-    /// and awaits the server's DONE token. Raises `RuntimeError`
-    /// synchronously if the connection has already been closed.
-    ///
-    /// If no transaction is currently open on the server, the commit will
-    /// fail with the server's own error (SQL Server 3902 — "The COMMIT
-    /// TRANSACTION request has no corresponding BEGIN TRANSACTION").
+    /// Async context manager entry. Resolves to `self` with no I/O.
+    fn __aenter__<'py>(slf: Py<Self>, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
+        pyo3_async_runtimes::tokio::future_into_py(py, async move { Ok(slf) })
+    }
```
David-Engel, line 290: **Suggestion:** `cursor()`, `commit()`, and `rollback()` all raise `RuntimeError("Connection is closed")`, but `async with closed_conn:` succeeds and then no-ops on exit. Suggest matching the rest of the surface:

```rust
fn __aenter__<'py>(slf: Py<Self>, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
    if slf.borrow(py).tds_client.is_none() {
        return Err(PyRuntimeError::new_err("Connection is closed"));
    }
    pyo3_async_runtimes::tokio::future_into_py(py, async move { Ok(slf) })
}
```

mssql-py-core/src/async_connection.rs
```diff
@@ -251,39 +251,23 @@ impl PyAsyncConnection {
...
-    /// `Arc<tokio::sync::Mutex<_>>`. Following DB-API 2.0, `cursor()` is a
-    /// synchronous call; only the cursor's execute/fetch methods will be
-    /// awaitable.
-    ///
-    /// Raises `RuntimeError` if the connection has already been closed.
-    /// A second cursor may be created on the same connection, but both
-    /// cursors share one TDS wire session and serialize on the same async
-    /// mutex — matching the non-MARS TDS session model.
+    /// Sync per DB-API 2.0. A second cursor is allowed; both share the same
+    /// TDS session and serialize on the same async mutex.
```
David-Engel: **Suggestion:** this docstring is now a fragment that says neither what `cursor()` returns nor that it raises `RuntimeError` when the connection is closed - both of which are contract that shows up in Python `help()`. Worth one line back:

```rust
/// Create an async cursor. Sync per DB-API 2.0. Raises `RuntimeError` if the
/// connection is closed. A second cursor is allowed; both share the same TDS
/// session and serialize on the same async mutex.
```

mssql-py-core/src/async_cursor.rs
```diff
@@ -48,14 +48,22 @@ pub struct PyAsyncCursor {
     /// serializes wire access across `.await` points.
     #[allow(dead_code)] // Consumed by upcoming async execute/fetch/close APIs.
     tds_client: Arc<Mutex<TdsClient>>,
+    /// Snapshot of the parent connection's default query timeout at
+    /// `cursor()` time (`0` = no timeout). Applied by the future `execute`
+    /// path unless overridden per-call.
+    #[allow(dead_code)] // Consumed by upcoming async execute API.
+    default_query_timeout: u32,
```
David-Engel, line 66: **Suggestion:** this field is private with `#[allow(dead_code)]` and has no getter, so nothing verifies that `cursor()` actually forwards the value. `test_timeout_default_and_setter_roundtrip` only exercises the connection property, and the 100% diff coverage is line coverage - `PyAsyncCursor::new(client, self.default_query_timeout)` executes, but always with `0`.

Smallest fix - add a getter (also drops one `#[allow(dead_code)]`):

```rust
#[pymethods]
impl PyAsyncCursor {
    /// Query timeout (seconds) snapshotted from the parent connection. `0` = no timeout.
    #[getter]
    fn timeout(&self) -> u32 { self.default_query_timeout }
}
```

and assert it:

```python
conn.timeout = 30
assert conn.cursor().timeout == 30
```

David-Engel (commented):
## Summary

Incremental enhancement of the preview `PyAsyncConnection` surface in `mssql-py-core`: adds `.timeout`, `.closed`, `is_connected()`, `__repr__`, `__aenter__`/`__aexit__`, a `python_logger` param on `connect`, and factors TDS-to-Python error mapping into `map_tds_error`. The Rust is clean and idiomatic, the test consolidation is a genuine improvement (the dropped `test_close_is_idempotent` is subsumed by `test_lifecycle_state_reflects_closed_and_is_connected`), and no coverage is lost. CI is green and diff coverage is 100%.

One blocking item on the new `python_logger` parameter (inline), plus a few suggestions and nits. Details are threaded on the relevant lines.

### Not tied to a diff line

**Suggestion - docstring trim dropped user-visible contract.** The rustdoc on `#[pymethods]` becomes Python `__doc__`. `commit`/`rollback` lost *"Raises `RuntimeError` synchronously if the connection has already been closed"*, and `close()` lost the "resolves to `None`" statement that you have a dedicated regression test for. I am on board with cutting the prose, but those are contract, not commentary - worth keeping one line each.

**Nit - PR description hygiene before merge:**
- The auto-generated summary is duplicated above the hand-written `## Description`.
- `Work item: **[TASK 46956]** - link the ADO work item here.` is leftover placeholder text, and 46956 is not among the three work items linked under **Related Issues** (46959 / 46960 / 47185) even though all eight commits reference it.
- `cargo btest`, "Full validation pipeline", and the entire bottom **Checklist** are unchecked, but CI is green - please tick them (per `.github/PULL_REQUEST_TEMPLATE.md`).
- Title says "Async Connection Foundation", but the foundation landed in #206; this PR enhances it. Something like "[py-core] Async connection lifecycle, timeout, and context-manager support" would read truer.

mssql-py-core/src/async_connection.rs
```diff
@@ -55,115 +41,141 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
         cls: &Bound<'py, PyType>,
         client_context_dict: &Bound<'_, PyDict>,
+        python_logger: Option<&Bound<'_, PyAny>>,
     ) -> PyResult<Bound<'py, PyAny>> {
         let py = cls.py();
 
-        // Preview API: emit a one-shot FutureWarning so callers see the
-        // instability signal at runtime.
+        // `DefaultGuard` is `!Send`, so its coverage ends when this method returns.
+        let _guard = python_logger
```
saurabh500: This `python_logger` bridge only covers the *synchronous* part of `connect`. `scoped_tracing_bridge` returns a thread-local `DefaultGuard`, and `_guard` is dropped the moment `connect` returns — but the actual TCP+TLS+login handshake runs later, inside the `future_into_py` block, on a Tokio worker thread. So every event in that block bypasses the caller's logger: the `connection established` line and, more importantly, the connect-failure error from `map_tds_error("connect", ...)` (line 121) — which is exactly the log someone who bothered to pass a `python_logger` would most want.

The sync `PyCoreConnection::new` doesn't hit this because it drives the handshake with `runtime.block_on(...)` inline, under the guard, on the same thread.

Can we move the bridge into the async block (hand the `Arc<Py<PyAny>>` into the future and `set_default` there), or — if capturing only the prelude is intentional for this foundation — document that per-call `python_logger` doesn't cover the async handshake? As written, passing a logger silently drops the failure diagnostics.

mssql-py-core/src/async_connection.rs
saurabh500: These two `info!` lines fire back to back with no work between them (`initiating async connection`, then `extracting client context`). Drop the first — it's noise.

mssql-py-core/src/async_connection.rs
```diff
@@ -55,115 +41,141 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
+    #[setter]
+    fn set_timeout(&mut self, value: u32) {
+        tracing::info!(
+            "PyAsyncConnection::set_timeout: default query timeout set to {}s",
+            value
+        );
+        self.default_query_timeout = value;
+    }
+
+    /// True after `close()` has been awaited (or if `connect()` never produced a
```
saurabh500: `connect()` returns `Err` on handshake failure and only ever builds the instance with `Some(client)`, so there's no public path to an instance that starts out closed. The "or if `connect()` never produced a live client" clause describes a state this API can't reach — trim it.

saurabh500:
Nevermind, my review came in after David's.
He has already covered most parts that my review covered. I had forgotten to post the review on time.

mssql-py-core/src/async_connection.rs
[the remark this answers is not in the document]
subrata-ms: fixed

mssql-py-core/src/async_connection.rs
```diff
@@ -3,23 +3,9 @@
...
-//! here submits its I/O to the shared process-wide Tokio runtime via
-//! [`crate::async_runtime`] and returns Python awaitables through
-//! `pyo3_async_runtimes::tokio::future_into_py`, so callers can `await` the
-//! results from `asyncio`.
-//!
-//! Invariant: one async connection maps to exactly one async cursor, one
-//! `TdsClient`, and one TDS wire session.
-//!
-//! [`FutureWarning`]: https://docs.python.org/3/library/exceptions.html#FutureWarning
+//! Invariant: one async connection ↔ one async cursor ↔ one `TdsClient`.
```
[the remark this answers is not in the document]
subrata-ms: fixed

mssql-py-core/src/async_connection.rs
subrata-ms: Agreed. Below things are included as part of the fix - 
1. Owned dispatch propagates through connect polling.
2. Dispatch persists on [PyAsyncConnection](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
3. [close](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [commit](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [rollback](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and timeout logging reuse it.
4. Tests cover connect success, connect failure, and later operations.

mssql-py-core/src/async_connection.rs
```diff
@@ -177,35 +189,38 @@ impl PyAsyncConnection {
...
-    /// TRANSACTION request has no corresponding BEGIN TRANSACTION").
+    /// Async context manager entry. Resolves to `self` with no I/O.
+    fn __aenter__<'py>(slf: Py<Self>, py: Python<'py>) -> PyResult<Bound<'py, PyAny>> {
+        pyo3_async_runtimes::tokio::future_into_py(py, async move { Ok(slf) })
+    }
+
+    /// Async context manager exit. Delegates to `close()`; exception info is
+    /// ignored and never suppressed (`close()` resolves to `None`, which is
+    /// falsy per Python's `__aexit__` contract).
+    fn __aexit__<'py>(
```
sumitmsft, line 294: `__aexit__` only closes; it doesn't implement the commit/rollback-then-close contract `mssql-python` guarantees.

`__aexit__` ignores the `(exc_type, exc_val, exc_tb)` triple and just calls `close()`. It never commits or rolls back, so it can't tell a clean exit from a failed one.

**What `mssql-python` does.** Its default is `autocommit=False`, and `Connection.__exit__` follows a *finalize-then-close* rule:

```python
if self._closed:
    return
try:
    if not self.autocommit:
        self.commit() if exc_type is None else self.rollback()
except Exception:
    try: self.close()
    except Exception: logger.warning(...)   # secondary failure: log only
    if exc_type is None: raise              # clean exit: surface the commit/rollback error
    return                                  # error exit: keep the user's original exception
try:
    self.close()
except Exception:
    if exc_type is None: raise
    logger.warning(...)                     # error exit: keep the user's original exception
```

So the equivalent async `__aexit__` should:
1. No-op if already closed.
2. When autocommit is off: `await commit()` on clean exit (`exc_type is None`), `await rollback()` on an exception.
3. **Always** `await close()` afterward.
4. Never let cleanup mask the user's error: if the block already raised, a failing rollback/close is logged and swallowed so the original exception propagates.

**Why this is blocking even though nothing is "lost" today.** The async connection currently never issues `begin_transaction`, so statements run in the server's implicit autocommit mode and each write is committed by the server on its own. That's fine right now, but it's not the `mssql-python` contract, and the moment autocommit-off is implemented, this same close-only `__aexit__` turns into a real data-loss bug: a clean `async with` exit would tear down the connection and roll back the open transaction instead of committing it. Getting the sequencing right here now avoids shipping a context manager whose behavior silently changes under the user later.

**Async note.** `commit`/`rollback`/`close` return `future_into_py` awaitables, so `__aexit__` must `await` them in order inside a single future and map the raise-vs-swallow decision onto the returned `PyResult` (return the error to raise; log and return `Ok` to suppress).

mssql-py-core/src/async_connection.rs
```diff
@@ -55,115 +41,141 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
-    /// Wrapped in `Option` so `close()` can take ownership and drop the
-    /// client; wrapped in `Arc<tokio::sync::Mutex<...>>` so the (upcoming)
-    /// async cursor and connection-level lifecycle methods can share access
-    /// across `.await` points without corrupting the TDS byte stream.
+    /// `Option` so `close()` can `take()`; `Arc<Mutex<>>` for cursor sharing.
     tds_client: Option<Arc<Mutex<TdsClient>>>,
+    /// Default query timeout (seconds) applied to cursors created from this
+    /// connection. `0` = no timeout, per pyodbc/ODBC `SQL_ATTR_QUERY_TIMEOUT`.
+    /// Pure Python-side state — the setter performs no I/O.
+    default_query_timeout: u32,
```
sumitmsft, line 111: No `autocommit` control on the async connection, so correct transaction finalization (see `__aexit__`) can't be expressed yet.

`mssql-python` only commits/rolls back on context exit *when autocommit is off* (`if not self.autocommit:`), and its default is `autocommit=False`. The async connection has no equivalent:

- No `autocommit` property/setter on `PyAsyncConnection`.
- No transaction is ever opened, so from Python the connection is effectively always autocommit-on, the opposite of `mssql-python`'s default and DB-API 2.0's implicit-transaction model.

**Why a naive `__aexit__` fix would fail.** Because no transaction is open, calling `commit()` on a fresh connection sends `TM_COMMIT` with nothing to commit and the server returns **error 3902**, which is exactly what this PR's own commit/rollback tests assert. So the `__aexit__` sequencing in the related comment genuinely can't be implemented correctly until an autocommit/transaction mode exists.

**The building blocks are already in mssql-tds**, so this isn't blocked on new core work as much as the PR description implies:
- `TdsClient::begin_transaction(isolation_level, name)`
- `TdsClient::has_active_transaction()`, to guard commit/rollback and avoid the 3902 path
- `commit_transaction` / `rollback_transaction`, which this PR already wires up

A `SET IMPLICIT_TRANSACTIONS`-style mode would match ODBC's `SQL_ATTR_AUTOCOMMIT` semantics most closely, but explicit `begin_transaction` management is also a viable path.

**Suggestion:** add an `autocommit` property/setter (default `False`, to match `mssql-python`) backed by begin/implicit-transaction control, then implement the `__aexit__` contract against it.

mssql-py-core/src/async_connection.rs
subrata-ms: fixed

mssql-py-core/src/async_connection.rs
subrata-ms: fixed

mssql-py-core/src/async_cursor.rs
subrata-ms: fixed

mssql-py-core/src/async_connection.rs
```diff
@@ -55,115 +41,141 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
+/// against a connection they intend to keep using.
+/// <https://sqlclientdrivers.visualstudio.com/mssql-python/_workitems/edit/47180>
 #[pyclass]
 pub struct PyAsyncConnection {
-    /// Wrapped in `Option` so `close()` can take ownership and drop the
-    /// client; wrapped in `Arc<tokio::sync::Mutex<...>>` so the (upcoming)
-    /// async cursor and connection-level lifecycle methods can share access
-    /// across `.await` points without corrupting the TDS byte stream.
+    /// `Option` so `close()` can `take()`; `Arc<Mutex<>>` for cursor sharing.
     tds_client: Option<Arc<Mutex<TdsClient>>>,
```
sumitmsft, line 103: Today the connection shares one thing with its cursors: a locked `TdsClient` (`Option<Arc<Mutex<TdsClient>>>`). The lock stops two cursors from writing to the wire at the same time, which is good. But a lock only knows "busy or free." It does not know *which* cursor is busy, what that cursor is currently doing, or how to cancel just that cursor's work. The design doc flags exactly this in annotation: `Arc<Mutex<TdsClient>>` is enough for safe access, but not enough to track cursor and result ownership across separate execute and fetch calls.

That missing information breaks two behaviors we will need:

1. **The "connection is busy" error.** Because MARS is out of scope, only one result can be active on a connection at a time. So if cursor A is still reading its rows and cursor B tries to run a query, B should get a clear error ("connection is busy with results for another command"). The connection can't produce that error today because it doesn't record that cursor A owns the active result.

2. **Cancelling the right cursor.** `cursor_b.cancel()` should cancel only cursor B's operation. But there is no place that remembers "this cancel handle belongs to cursor B," so a cancel has nothing specific to target and could end up disturbing cursor A's work.

**Why fix it in this PR.** This is the foundation that execute, fetch, and cancel will build on in later PRs. Those features need this ownership information to behave correctly. If we don't add it to these structs now, the later PRs will have to come back and rewrite the same structs.

**Suggestion:** give the connection a small piece of shared state (something the connection and every cursor can see) that remembers:
- which cursor owns the currently active result,
- what that cursor is doing (idle, executing, fetching, closing),
- the cancel handle for the active operation,
- whether the connection is closing or broken, so a broken connection is not reused,
- a simple increasing id per cursor or statement, so operations can be named unambiguously.

With that in place, cursor B gets a clean busy error while cursor A still owns the result, and `cursor_b.cancel()` can only ever cancel cursor B.

If you would prefer to add this in the execute/fetch PR instead, that is a reasonable choice, but it is worth deciding now because it shapes these struct definitions.

mssql-py-core/src/async_connection.rs
```diff
@@ -177,35 +189,38 @@ impl PyAsyncConnection {
             );
             let mut guard = client.lock().await;
```
sumitmsft: `close()` does two things. First it calls `self.tds_client.take()`, which immediately marks the Python connection object as closed, so new `commit`/`rollback`/`cursor` calls correctly fail with "Connection is closed." But then, inside the spawned future, it does `client.lock().await` before tearing down the transport.

The catch is that each cursor keeps its own clone of that same `Arc<Mutex<TdsClient>>`. Once execute and fetch are implemented and hold the mutex across their awaits (for example, streaming rows from a slow query), `close()` cannot get the lock. So `close()` just queues behind the running query and does not actually close anything until that query finishes on its own or times out. A `close()` on a connection running a long `WAITFOR` would hang for the whole duration of the query.

That is the opposite of the documented policy. Design section 10.2.7 says: "Close acts on the connection socket; any in-flight cursor operation will fail with a connection error, which is the correct signal." In other words, close should tear down the socket directly and let the in-flight cursor operation fail fast, not wait its turn behind the mutex.

Note: this does not misbehave today, because the async cursor has no execute/fetch yet, so nothing holds the lock across awaits. But the moment those land, this close path silently turns into a hang.

**Suggestion:** define and implement the close-during-operation transition now, following the design:
- immediately reject new work (already handled by `take()`),
- act on the transport/socket directly so any in-flight cursor operation fails with a connection error, instead of waiting on the shared mutex,
- await the teardown and cleanup,
- if protocol recovery fails, mark the connection broken so it is never reused (design section 10.1: after any error the connection is either usable or explicitly broken, never a third state).

The doc also raises the related open question of what commit, rollback, and close should do while an execute or fetch is still in flight. Close is the one case the doc already answers, so it is worth locking the behavior in here.

sumitmsft:
The four Connection methods (`close`, `commit`, `rollback`, `cursor`) are present, so the method surface is there. But "method-complete against PEP 249" reads as a claim about DB-API behavior, and that behavior is not in place yet:

- **Transactions:** `commit()` and `rollback()` currently just send `TM_COMMIT`/`TM_ROLLBACK`. There is no `autocommit=False` default and no transaction is opened, so on a fresh connection they raise SQL Server 3902/3903 (the tests assert this). PEP 249 expects manual-commit by default with `commit`/`rollback` acting on a real transaction. The description itself lists `.autocommit` as a follow-up blocked on an mssql-tds primitive, so this is understood, it just is not "complete" yet.

- **Exceptions:** errors map to `PyRuntimeError`, not the DB-API exception hierarchy (`Connection.Error` and friends)

Suggestion: We can word it something like "Connection method surface in place (`close`/`commit`/`rollback`/`cursor`); full PEP 249 transaction and exception semantics are follow-ups (autocommit, US 47181)." That keeps the description accurate about what this foundational PR delivers versus what is still pending.

mssql-py-core/src/async_cursor.rs
```diff
@@ -48,14 +48,22 @@ pub struct PyAsyncCursor {
...
 }
 
 impl PyAsyncCursor {
     /// Construct a new cursor bound to the given TDS client.
     ///
     /// Called only from `PyAsyncConnection::cursor`.
-    pub(crate) fn new(tds_client: Arc<Mutex<TdsClient>>) -> Self {
-        Self { tds_client }
+    pub(crate) fn new(tds_client: Arc<Mutex<TdsClient>>, default_query_timeout: u32) -> Self {
```
sumitmsft: **[Follow-up] We need a two-cursor acceptance test for the "connection is busy" behavior once execute/fetch land.**

This is where `execute`/`fetch` will be added, and it's the behavior that the shared connection-state comment (on `async_connection.rs`) is really about. To prove that state model actually works, please add an acceptance test with two cursors on the same connection:

1. Cursor A runs a query that returns multiple rows, and reads only some of them (leaves rows unread, so the result is still open on the server). Use a result set large enough that it does not all arrive in a single packet, otherwise the server closes the result immediately and there is nothing to be "busy" on.
2. Cursor B calls `execute` while A's result is still open, and it raises a specific, typed busy-state error (not a generic `RuntimeError`), matching the design doc's "connection is busy with results for another command."
3. Cursor A keeps fetching and gets its remaining rows correctly, so B's rejected attempt did not corrupt or steal A's result.
4. After A finishes draining (or is closed), cursor B can `execute` successfully.

This test can't be written in this PR because `execute`/`fetch` don't exist yet, so it belongs with the execute/fetch work (US 46961). Flagging it here so it lands together with those methods and so the connection-state design from the related comment is built with this exact scenario in mind.

sumitmsft (changes requested):
Left a few comments which must conform the design document decisions and applicability to mssql-python behavior mostly from architecture perspective. Some of them are follow up items in subsequent PRs.

mssql-py-core/src/async_connection.rs
subrata-ms: fixed

mssql-py-core/src/async_connection.rs
subrata-ms: fixed

mssql-py-core/src/async_connection.rs
subrata-ms: Fixed.

- [python_logger_dispatch()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) creates an owned [tracing::Dispatch](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

- [with_optional_dispatch()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) applies it using [Future::with_subscriber](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

- The complete handshake future is wrapped, including:
create_client(...).await
[map_tds_error("connect", ...)](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
["connection established"](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
-The dispatch is stored on [PyAsyncConnection](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and reused by later operations.

mssql-py-core/src/async_connection.rs
subrata-ms: fixed

mssql-py-core/src/async_connection.rs
subrata-ms: fixed.

mssql-py-core/src/async_connection.rs
```diff
@@ -55,235 +44,288 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
+    tracing::error!("PyAsyncConnection::{op}: failed: {e}");
+    PyRuntimeError::new_err(format!("{user_msg}: {e}"))
+}
+
+async fn with_optional_dispatch<F>(future: F, dispatch: Option<tracing::Dispatch>) -> F::Output
+where
+    F: Future,
+{
+    match dispatch {
+        Some(dispatch) => future.with_subscriber(dispatch).await,
```
saurabh500: Re-review: my earlier concern is resolved. The bridge is now an owned `tracing::Dispatch` (`python_logger_dispatch`) attached to the future via `with_subscriber`, so events emitted inside `future_into_py` — including `connection established` and the `map_tds_error("connect", ...)` failure (`tracing::error!` "connect: failed") — now reach the caller's `python_logger` instead of being dropped. Nice touch storing the dispatch on the connection so commit/rollback/close/set_timeout reuse it. `test_failed_connect_logs_while_awaitable_is_polled` covers the failure path directly. Verified good.

mssql-py-core/src/async_connection.rs
```diff
@@ -55,235 +44,288 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
-        // Preview API: emit a one-shot FutureWarning so callers see the
-        // instability signal at runtime.
+        let dispatch = python_logger
+            .map(|logger| python_logger_dispatch(Arc::new(logger.clone().unbind()), file!()));
+        let _guard = dispatch.as_ref().map(tracing::dispatcher::set_default);
+
         emit_preview_warning(py)?;
 
-        tracing::info!("PyAsyncConnection::connect: extracting client context");
+        tracing::debug!("PyAsyncConnection::connect: extracting client context");
```
saurabh500: Re-review: resolved. The back-to-back `initiating async connection` / `extracting client context` pair is gone — first line dropped, this one downgraded to `debug!`. Good.

mssql-py-core/src/async_connection.rs
```diff
@@ -55,235 +44,288 @@ fn emit_preview_warning(py: Python<'_>) -> PyResult<()> {
...
+            .map(tracing::dispatcher::set_default);
+        tracing::info!(
+            "PyAsyncConnection::set_timeout: default query timeout set to {}s",
+            value
+        );
+        self.default_query_timeout = value;
+        Ok(())
+    }
+
+    /// True after `close()` has been called. Cheap LBYL check; performs no I/O.
```
saurabh500: Re-review: resolved. The unreachable "or if `connect()` never produced a live client" clause is trimmed, and "awaited" → "called" is actually more accurate since `close()` runs `tds_client.take()` synchronously before the future is awaited. Good.

mssql-py-core/src/async_connection.rs
subrata-ms: I am making this change in 2 steps. First step will be, introducing foundational changes under this PR and second step changes will be part of execute () functionality implementation. I will put a TODO list that would be our reference point.

1. Store [autocommit: bool](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) on [PyAsyncConnection](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), defaulting to false.

2. Share transaction/autocommit state with cursors.

3. Before cursor execution:

    - Lock [TdsClient](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

   - If autocommit is off and no transaction is active, call [begin_transaction()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

   - Execute the statement.

4. Make [commit()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [rollback()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) no-ops when no transaction is active, matching ODBC behavior.

5. Make [close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) roll back an active transaction before closing.

6. Make [__aexit__()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) commit on clean exit, roll back on exceptional exit, and always close.

7. Preserve the original block exception if rollback or close also fails.

The main limitation is timing: because [PyAsyncCursor](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) does not execute queries yet, adding autocommit state now cannot provide complete behavior. We can implement the connection lifecycle pieces now, but lazy transaction creation must land with async [execute()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

mssql-py-core/src/async_connection.rs
subrata-ms: This comment is now partially stale because the lifecycle foundation (previous comment) addresses most points:

- [connect(..., autocommit=False)](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) now defaults to DB-API behavior.
- Read-only [.autocommit](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) property exists.
- [commit()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [rollback()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) guard with [has_active_transaction()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), avoiding 3902/3903.
- [__aexit__()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) commits or rolls back active work when autocommit is disabled, then closes.
- Direct [close()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) rolls back active work.
- Autocommit state is shared with cursors.
    
**Remaining Gap**

No transaction is opened yet because async cursor execution does not exist. The cursor TODO correctly requires future [execute()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to:
- Lazily call [begin_transaction()](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) when autocommit is false.
- Keep check, begin, and submission under one mutex lock.
- Start a fresh transaction after commit/rollback.

mssql-py-core/src/async_connection.rs
subrata-ms: Implemented shared async session ownership in [async_session.rs](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), including lifecycle, operation phase, cursor/operation IDs, active ownership, and cancel-handle storage.
Updated the TODO list that need to be handled during execute and fetch method implementation.

mssql-py-core/src/async_cursor.rs
subrata-ms: This will be handled during execute method implementation. Updated the TODO inside async_cursor.rs.

mssql-py-core/src/async_connection.rs
subrata-ms: **What pycore can do now:**

- Immediately transition to [Closing](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and reject new work.
- Record [Closed](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) or [Broken](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) after eventual teardown.
- Cancel an active operation through its future [CancelHandle](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

**What it cannot do:**

- Access or close the socket while another task owns the client mutex.
- Guarantee prompt close, because cooperative cancellation waits for an ATTENTION acknowledgment and may wait indefinitely.
- Make the active operation fail specifically with a connection error.

**A correct implementation requires one of:**

- An independent hard-close handle exposed by [mssql-tds](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html), or
- Replacing [Arc<Mutex<TdsClient>>](vscode-file://vscode-app/c:/Users/spaitandi/AppData/Local/Programs/Microsoft%20VS%20Code/df53daabb1/resources/app/out/vs/code/electron-browser/workbench/workbench.html) with a dedicated connection-owner task/actor. However, even an actor cannot interrupt a TDS operation if that operation awaits inline; it would need spawned cancellable operations or lower-level socket ownership.

Updated the respective task to track this requirement.
Task - https://dev.azure.com/sqlclientdrivers/mssql-python/_workitems/edit/47199

sumitmsft (approved):

PR #290, opened 2026-08-14
Bound COLMETADATA pre-allocation to prevent allocation-amplification DoS

saurabh500:
## Description

The COLMETADATA (`0x81`) token parser read a 16-bit `col_count` and immediately did `Vec::<ColumnMetadata>::with_capacity(col_count)` **before** reading any column bytes. With `size_of::<ColumnMetadata>() = 640`, an attacker-controlled `col_count` up to `0xFFFE` (65534) forces an eager allocation of up to **~42 MB** from as few as **3 input bytes** (~14,000,000× amplification).

Under the ASan fuzz build, allocating and shadow-poisoning that region takes several seconds, which libFuzzer reports as a **timeout**. The `GH-Rust Fuzz Test` pipeline (definitionId 2207) runs with `-timeout=10 -fork=1 -ignore_crashes=1`, where timeout reproducers are written with the `crash-` prefix — so this surfaced as a "crash" in build **164079** (`Found 1 crash(es) during fuzzing`, job *Fuzz Test Token Stream Linux*, crash artifact `fuzz-crashes-fuzz_token_stream-__default` / `crash-80c55599ebe9adee81668324e4c6151894a33fbe`).

### Fix

Bound attacker-controlled vector pre-allocation instead of capping protocol counts. `new_column_vec` derives its element cap from a fixed 160 KiB byte budget, and the parser grows the vector only as column bytes are consumed. The same strategy gives the Always Encrypted CEK table a 12 KiB initial-allocation budget, closing the identical amplification path in `parse_cek_table`. Because the caps use `size_of`, growing either element type reduces its pre-allocation count rather than silently increasing eager bytes.

No artificial column or CEK-table count limit is imposed. Legitimate wide/sparse-column result sets continue to parse, and the existing `0xFFFF` no-metadata sentinel is preserved. The original EOF-path tests remain, allocation-site tests assert the actual vector capacities, and the minimized reproducers are included in the checked-in fuzz corpus.

### Verification

- Minimal reproducers (`02 81 33 aa`, `81 84 b8`) ran in **1.5–5.6 s** under ASan before the fix; **22–26 ms** after.
- `cargo nextest run --frozen -p mssql-tds -E 'test(colmetadata) | test(capacity)'` → 25 passed.
- Mutating either allocation helper to reserve the full attacker-controlled count makes its typed capacity test fail.
- `cargo bfmt` / `cargo bclippy` clean.

## Related Issues

ADO work item: https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47316

[AB#47316](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47316)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes on the latest commit


saurabh500:
The 4096 column count assumption doesn't seem right. Sparse tables can have 30000 columns.

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -103,6 +103,13 @@ use crate::{
 /// Column-flag bit indicating the column is protected by Always Encrypted.
 const FLAG_ENCRYPTED: u16 = 0x0800;
 
+/// Maximum number of columns a single result set may declare. SQL Server caps a
+/// SELECT at 4096 columns, so a COLMETADATA token claiming more is malformed.
+/// Rejecting an implausible count before reserving per-column storage bounds the
+/// eager `Vec<ColumnMetadata>` allocation and prevents a tiny token (a 3-byte
+/// `81 FF FE` reserves ~42 MB) from driving a memory-amplification DoS.
+const MAX_COLUMN_COUNT: u16 = 4096;
```
saurabh500: The 4096 column count assumption doesn't seem right. Sparse tables can have 30000 columns.

saurabh500: Good catch — you're right that a fixed 4096 cap (or even 30,000) would wrongly reject legitimate wide/sparse-column result sets. I've removed the artificial column-count limit entirely. The real DoS was the eager `Vec::with_capacity(col_count)` reservation; bounding the initial capacity (`COLUMN_PREALLOC_CAP`) kills the amplification regardless of count, since `col_count` is a `u16` and per-column work is now proportional to the bytes actually on the wire. Any valid column count (including 30,000+) now parses without a reject.

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -132,6 +139,16 @@ where
             return Ok(Tokens::from(ColMetadataToken::default()));
         }
 
+        // Reject an implausible column count before reserving per-column storage.
+        // Without this bound a 3-byte token can drive a ~42 MB eager allocation
+        // (col_count * size_of::<ColumnMetadata>()), a memory-amplification DoS
+        // that the fuzzer surfaces as a timeout.
+        if col_count > MAX_COLUMN_COUNT {
```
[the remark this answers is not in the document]
saurabh500: Removed — the reject block (and its comment) is gone entirely. The DoS defense now lives solely in the bounded pre-allocation, whose rationale is documented once on `COLUMN_PREALLOC_CAP`.

saurabh500:
Addressed in the latest commit: I dropped the column-count cap entirely rather than raising it to 30,000/64,000. Since `col_count` is a `u16` and the amplification came from the eager `Vec::with_capacity(col_count)` reservation, bounding the initial capacity removes the DoS without an artificial limit — legitimate wide/sparse-column result sets (30,000+ columns) now parse fine.

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -147,8 +157,11 @@ where
         // for an (empty) table and desynchronize the token stream.
         let has_cek_table = is_column_encryption_supported;
 
-        // Pre-allocate vector for column metadata
-        let mut column_metadata: Vec<ColumnMetadata> = Vec::with_capacity(col_count as usize);
+        // Reserve conservatively; a large but valid `col_count` grows the vector
+        // as columns are parsed, while a malformed count cannot force a huge
+        // eager allocation. See `COLUMN_PREALLOC_CAP`.
+        let mut column_metadata: Vec<ColumnMetadata> =
+            Vec::with_capacity((col_count as usize).min(COLUMN_PREALLOC_CAP));
```
David-Engel: The bound itself is right. Consider extracting it so the property is directly testable — see my comment on the new tests below, which currently pass with or without this `.min(...)`.

```rust
const fn initial_column_capacity(col_count: u16) -> usize {
    if (col_count as usize) < COLUMN_PREALLOC_CAP {
        col_count as usize
    } else {
        COLUMN_PREALLOC_CAP
    }
}
```

```rust
let mut column_metadata: Vec<ColumnMetadata> =
    Vec::with_capacity(initial_column_capacity(col_count));
```

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -631,6 +644,36 @@ mod tests {
...
+        // capacity is bounded by COLUMN_PREALLOC_CAP, so this cannot amplify a
+        // 2-byte token into a huge Vec.
+        let data = vec![0x88, 0x13]; // col_count = 0x1388 = 5000
+        let mut reader = MockReader::new(data);
+        let parser = ColMetadataTokenParser;
+        let context = ParserContext::default();
+
+        assert!(parser.parse(&mut reader, &context).await.is_err());
+    }
+
```
David-Engel, line 703: **Blocking:** these two tests don't test the fix. Both only assert `parse(...).is_err()`, which was already true before the change — the old code allocated the 42 MB and *then* failed on the first per-column read.

I confirmed this: reverting line 164 to `Vec::with_capacity(col_count as usize)` and running these two tests gives **2 passed**. So the guard can be removed and CI stays green, which is the one thing a regression test for a security fix has to prevent.

The property to assert is the capacity bound, not the error. With `initial_column_capacity` extracted (see my comment above):

```rust
#[test]
fn test_initial_column_capacity_is_bounded() {
    assert_eq!(initial_column_capacity(0), 0);
    assert_eq!(initial_column_capacity(10), 10);
    assert_eq!(initial_column_capacity(5000), COLUMN_PREALLOC_CAP);
    assert_eq!(initial_column_capacity(0xFFFE), COLUMN_PREALLOC_CAP);
}
```

Please keep both `fails_fast` tests — they're useful documentation of the EOF path — but this is the one that actually fails without the fix.

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -103,6 +103,16 @@ use crate::{
...
 
+/// Upper bound on the column vector's initial capacity. `col_count` is attacker-
+/// controlled, so reserving the full count up front lets a tiny token drive a
+/// large eager allocation (`col_count * size_of::<ColumnMetadata>()`) — a memory-
+/// amplification DoS the fuzzer surfaced as a timeout (build 164079: a 3-byte
+/// token reserving ~42 MB). Reserving conservatively and letting the vector grow
+/// as columns are actually parsed keeps work proportional to the bytes on the
+/// wire, so any legitimate column count (SQL Server wide tables allow up to
+/// 30,000 columns) is handled without an artificial cap.
+const COLUMN_PREALLOC_CAP: usize = 256;
```
David-Engel: Nit: the cap is in elements, so the real eager allocation is `256 × 640 B ≈ 160 KB` today, and it grows silently if `ColumnMetadata` gets fatter. Worth one line in the doc comment stating the byte figure so the tradeoff stays visible.

David-Engel (commented):
## Summary

Small, correct fix. Capping the initial capacity at `COLUMN_PREALLOC_CAP` (256) removes the amplification without imposing an artificial column limit, and behavior is unchanged for legitimate result sets — the `0xFFFF` sentinel still short-circuits and the loop still runs `col_count` times, so the `column_count == columns.len()` invariant that `row_parser` / `nbcrow_parser` rely on still holds.

I verified the amplification math locally: `size_of::<ColumnMetadata>() == 640`, so `0xFFFE × 640 ≈ 41.9 MB`. The description is accurate.

One blocking issue on the tests (inline), plus two findings that fall outside the diff:

### Suggestion — `parse_cek_table` has the identical bug, ten lines below the fix

`colmetadata_parser.rs:275` still does `Vec::<CekTableEntry>::with_capacity(table_size as usize)` from an attacker-controlled `u16`, before reading any entry bytes. I measured `size_of::<CekTableEntry>() == 48`, so that's ~3.1 MB reserved from 2 bytes on the wire (~1.5M× amplification). It's reachable only when Always Encrypted is negotiated, so it's lower severity and may not be on the fuzz target's current path — but it's the same pattern in the same file, and bounding it here closes the class rather than one instance. (`parse_cek_table_entry`'s `value_count` is a `u8`, max ~24 KB — fine as-is.)

### Suggestion — seed the fuzz corpus with the reproducer

`mssql-tds/fuzz/corpus-fuzz_token_stream.tar.gz` is checked in and `run-fuzz.sh` extracts it as the starting corpus. Adding the minimized inputs (`02 81 33 aa`, `81 84 b8`) to that archive makes the fuzz job itself re-exercise the case, which covers the ASan-timing dimension unit tests can't reach.

### Nit — follow-up

`ColumnMetadata` at 640 bytes is the multiplier that made this exploitable at all. Boxing the large members (`type_info` / `crypto_metadata`) would shrink the factor across every `Vec<ColumnMetadata>` in the crate. Separate work item, not this PR.

### Checked, no issues

Linked ADO work item present (AB#47316) and the Linked Issue check passes; coverage-report and PR validation are green; no public API change, so the unchecked docs box is correct. `cargo test -p mssql-tds --lib colmetadata` → 23 passed locally, matching the description.

mssql-tds/src/token/parsers/colmetadata_parser.rs
saurabh500: Fixed in 02e43599: extracted `initial_column_capacity`, wired the parser allocation through it, and added direct bound coverage.

> [!NOTE]

mssql-tds/src/token/parsers/colmetadata_parser.rs
saurabh500: Fixed: the cap documentation now records the current 640-byte element size and 160 KiB eager-allocation ceiling.

> [!NOTE]

mssql-tds/src/token/parsers/colmetadata_parser.rs
saurabh500: Fixed: retained both EOF-path tests and added `test_initial_column_capacity_is_bounded`; I also verified it fails under the old unbounded behavior and passes with the cap.

> [!NOTE]

saurabh500:
Also addressed the two review suggestions outside the inline threads in 02e43599: bounded the Always Encrypted CEK-table pre-allocation and seeded the checked-in token-stream fuzz corpus with both minimized reproducers.

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -631,6 +666,51 @@ mod tests {
...
+
+    #[test]
+    fn test_initial_cek_table_capacity_is_bounded() {
+        assert_eq!(initial_cek_table_capacity(0), 0);
+        assert_eq!(initial_cek_table_capacity(10), 10);
+        assert_eq!(initial_cek_table_capacity(u16::MAX), CEK_TABLE_PREALLOC_CAP);
+    }
+
+    #[tokio::test]
+    async fn test_large_column_count_without_data_fails_fast() {
```
David-Engel: **Suggestion:** this test (and `test_large_valid_column_count_without_data_fails_fast` below) does not actually test the fix.

I verified this by mutation rather than inspection: I reverted line 185 to `Vec::with_capacity(col_count as usize)` and re-ran — both tests still **PASS**. They assert only `is_err()`, which the unbounded version also returned; it just reserved ~30 MB first. So they are regression guards against someone adding a hard `col_count` cap, not against allocation amplification.

The only tests that discriminate are the two helper unit tests, and those exercise `initial_column_capacity` in isolation — nothing ties the helper to the allocation site. The cheapest way to close that gap is to move the allocation into a helper and assert on the returned capacity:

```rust
fn new_column_vec(col_count: u16) -> Vec<ColumnMetadata> {
    Vec::with_capacity(initial_column_capacity(col_count))
}

#[test]
fn column_vec_capacity_is_bounded() {
    assert!(new_column_vec(0xFFFE).capacity() <= COLUMN_PREALLOC_CAP);
}
```

Related: the PR description says "The direct capacity tests fail with the old unbounded behavior." That is true only of the helper unit tests (which reference symbols that don't exist on `main`), not of these two parser tests — worth correcting.

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -103,6 +103,37 @@ use crate::{
...
+/// Upper bound on the column vector's initial capacity. `col_count` is attacker-
+/// controlled, so reserving the full count up front lets a tiny token drive a
+/// large eager allocation (`col_count * size_of::<ColumnMetadata>()`) — a memory-
+/// amplification DoS the fuzzer surfaced as a timeout (build 164079: a 3-byte
+/// token reserving ~42 MB). Reserving conservatively and letting the vector grow
+/// as columns are actually parsed keeps work proportional to the bytes on the
+/// wire, so any legitimate column count (SQL Server wide tables allow up to
+/// 30,000 columns) is handled without an artificial cap. With today's 640-byte
+/// `ColumnMetadata`, this limits the eager allocation to 160 KiB.
+const COLUMN_PREALLOC_CAP: usize = 256;
```
David-Engel: **Suggestion (optional):** the "160 KiB" / "12 KiB" figures are correct today — I measured 640 and 48 bytes — but they rot silently if either struct grows. Deriving the count from a byte budget keeps the security property stable and the comment self-maintaining:

```rust
const MAX_PREALLOC_BYTES: usize = 160 * 1024;
const COLUMN_PREALLOC_CAP: usize = MAX_PREALLOC_BYTES / size_of::<ColumnMetadata>();
```

Fine to keep the fixed element count if you prefer it; just flagging the maintenance edge.

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -103,6 +103,37 @@ use crate::{
...
+    }
+}
+
+const fn initial_cek_table_capacity(table_size: u16) -> usize {
+    if (table_size as usize) < CEK_TABLE_PREALLOC_CAP {
+        table_size as usize
+    } else {
+        CEK_TABLE_PREALLOC_CAP
+    }
+}
```
David-Engel: **Nit:** the two `const fn`s are byte-identical apart from which constant they clamp to. One parameterized helper keeps both call sites self-documenting without the duplicated branch:

```rust
const fn bounded_capacity(count: u16, cap: usize) -> usize {
    if (count as usize) < cap { count as usize } else { cap }
}
```

Also, neither needs to be `const` — dropping that lets you write `(count as usize).min(cap)`, since `Ord::min` isn't const-callable.

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -103,6 +103,37 @@ use crate::{
...
+
+const fn initial_column_capacity(col_count: u16) -> usize {
+    if (col_count as usize) < COLUMN_PREALLOC_CAP {
+        col_count as usize
+    } else {
+        COLUMN_PREALLOC_CAP
+    }
+}
+
+const fn initial_cek_table_capacity(table_size: u16) -> usize {
```
David-Engel: **Nit:** `table_size` is an entry *count*, not a size in bytes — the loop is `for _ in 0..table_size` and the doc on `parse_cek_table` says `Count(u16)` followed by `Count` entries. The name is inherited from the existing code, but this new signature is a good place to call it `entry_count`.

David-Engel (commented):
## Summary

The fix is correct and the right shape. Bounding the initial `Vec` capacity (rather than capping `col_count`) keeps work proportional to bytes on the wire without inventing a protocol limit — the same idiom as serde's `size_hint::cautious`. The `0xFFFF` no-metadata sentinel is preserved, the CEK table gets the identical treatment, and CI (including ADO PR validation build 167365) is green. **No blocking issues.**

### Verified on the branch

- `size_of::<ColumnMetadata>() = 640`, `CekTableEntry = 48` — matches the doc comments exactly (256 × 640 = 160 KiB, 256 × 48 = 12 KiB).
- The corpus tarball adds exactly two entries, with bytes `81 84 b8` and `02 81 33 aa`, matching the description. Nothing else in the archive changed.
- `cargo nextest run -p mssql-tds --lib -E 'test(colmetadata) or test(capacity)'` → 26 passed.
- **No adjacent hole left in the token stream.** I scanned every length-prefixed eager allocation under `mssql-tds/src/token` and `mssql-tds/src/io`: `featureext_parser` bounds via `MAX_ALLOWED_FE_DATA_IN_BYTES`, `session_state_parser` validates `state_len` against remaining token bytes, `fedauth_parser` validates `options_count`, `sspi_parser` is `u16`-bounded. `row_parser` / `nbcrow_parser` pre-allocate from `column_metadata_token.column_count`, which after a successful COLMETADATA parse always equals `columns.len()` — i.e. backed by real bytes. COLMETADATA and the CEK table were the two genuinely unbounded sites, and both are fixed here.

One non-inline nit: the `cargo btest` checklist box is still unchecked, and PR validation is now green — worth ticking.

Suggestions and nits are inline.

mssql-tds/src/token/parsers/colmetadata_parser.rs
saurabh500: Fixed in 1514cccc: the parser now allocates through `new_column_vec`, and the regression test asserts the returned `Vec` capacity. Mutating that helper back to full-count reservation makes the test fail.

> [!NOTE]

mssql-tds/src/token/parsers/colmetadata_parser.rs
saurabh500: Fixed in 1514cccc: renamed `table_size` to `entry_count` consistently across the CEK-table read, allocation, and loop.

> [!NOTE]

mssql-tds/src/token/parsers/colmetadata_parser.rs
saurabh500: Fixed in 1514cccc: both typed allocation helpers now share a non-const `bounded_capacity` using `usize::min`, removing the duplicated branches.

> [!NOTE]

mssql-tds/src/token/parsers/colmetadata_parser.rs
saurabh500: Applied in 1514cccc: both element caps are now derived from fixed 160 KiB and 12 KiB byte budgets using `size_of`, so type growth reduces capacity instead of increasing eager bytes.

> [!NOTE]

mssql-tds/src/token/parsers/colmetadata_parser.rs
```diff
@@ -631,6 +659,54 @@ mod tests {
...
+        assert_eq!(new_cek_table_vec(0).capacity(), 0);
+        assert_eq!(new_cek_table_vec(10).capacity(), 10);
+        assert_eq!(
+            new_cek_table_vec(u16::MAX).capacity(),
+            CEK_TABLE_PREALLOC_CAP
+        );
+    }
+
+    #[tokio::test]
+    async fn test_large_column_count_without_data_fails_fast() {
```
David-Engel: **Nit (non-blocking, no need to respin):** this test and `test_large_valid_column_count_without_data_fails_fast` still assert only `is_err()`, so they'd pass with the bound removed — I re-confirmed that by mutation. That's fine now, since `test_column_vec_capacity_is_bounded` carries the real guarantee and these are useful supplementary EOF coverage.

The only thing worth trimming is the comment, which reads as if the assertion proves the bound ("capacity is now bounded by `COLUMN_PREALLOC_CAP` ... instead of amplifying into a huge allocation"). A future reader may take this test as the amplification guard and delete the capacity test. Describing just the EOF path here, and letting the capacity tests own the bound, avoids that.

David-Engel (approved):
## Re-review @ `1514cccc` — approving

All four comments from my previous review are addressed. Net diff vs `main` is still just the parser file plus the two corpus reproducers — nothing else crept in.

| Previous comment | Resolution |
|---|---|
| Tests didn't exercise the bound | Allocation moved into `new_column_vec` / `new_cek_table_vec`, and the tests assert `.capacity()` on the returned `Vec` |
| Express the cap in bytes | `COLUMN_PREALLOC_BYTES / size_of::<ColumnMetadata>()`, same shape for the CEK table |
| Collapse the duplicate `const fn`s | Single `bounded_capacity(count, cap)` using `.min()` |
| `table_size` is an entry count | Renamed to `entry_count` at both the local and the parameter |

### Verified on this head

- **The capacity test now discriminates** — I re-ran the same mutation as before: reverting `new_column_vec` to `Vec::with_capacity(col_count as usize)` makes `test_column_vec_capacity_is_bounded` **FAIL** (`left: 5000, right: 256`). On the previous head the equivalent tests passed with the bound removed, so this is a real guard now.
- Both budgets divide exactly: 160 KiB ÷ 640 = 256 columns, 12 KiB ÷ 48 = 256 CEK entries — same effective caps as before, now immune to the structs changing size.
- The parser actually calls the helpers (lines 179 and 290), so the assertion is tied to the allocation site.
- `cargo fmt --check` clean, `cargo clippy -p mssql-tds --lib --all-targets` clean, `cargo nextest run -p mssql-tds --lib -E 'test(colmetadata) or test(capacity) or test(cek_table)'` → 25 passed. All 19 PR checks green.

One residual non-blocking nit inline; fine to fold into a later change or ignore.

mssql-tds/src/token/parsers/colmetadata_parser.rs
saurabh500: Fixed in 70015797: renamed both tests to describe their error-path assertion and trimmed the comments to the missing-column-bytes behavior; the typed vector-capacity tests remain the sole allocation-bound guards.

> [!NOTE]

Vahid-b (approved):

PR #291, opened 2026-08-14
Remove ResultSet async_trait boxing

saurabh500:
## Description

This focused follow-up removes `#[async_trait]` from the public `ResultSet` trait and `impl ResultSet for TdsClient`. The async methods now use native RPITIT futures with explicit `+ Send`; the concrete implementation returns the existing native client futures directly, avoiding per-call trait-future boxing while preserving row cleanup and end-of-set behavior.

This PR is stacked on #286 (`dev/saurabh/remove-row-path-async-traits`) as the new top layer of native stack #287. It does not change `TdsTransport`, `TdsTokenStreamReader`, or transport dispatch.

### Public API impact

`ResultSet` is intentionally no longer dyn-compatible. This is a breaking object-safety change for `dyn ResultSet` consumers and for external implementations using `#[async_trait]`. Existing concrete call sites retain the same `.method(...).await` source shape; implementations must migrate to native `Send` futures. Repository search found no trait-object consumers.

### Measurements

A temporary in-crate release harness (removed before commit) compared native `ResultSet::next_row` with an explicit `Pin<Box<dyn Future + Send>>` wrapper around the same native method. Each independent process used a paired/interleaved ABBA sequence over 500,000 production-shaped rows with 48 `INT` columns. Across seven processes, the boxed path was 1.52%–4.64% slower, with a 3.63% median. The range is reported as measured rather than excluding noisy runs.

Current debug future sizes are:

- `ResultSet::next_row`: 4,056 B
- `ResultSet::next_row_into`: 4,032 B
- `ResultSet::close`: 4,376 B

Both per-row futures are guarded under the existing 4,096 B hot-path budget. `close` is larger, but it is not a per-row operation, so this change documents the effect instead of adding broad boxing.

### Validation

- `cargo bfmt` — passed
- `cargo bclippy` — passed
- `cargo btest` — 2,410 passed, 366 environment failures, 11 skipped; exactly matches the current lower-layer baseline (7 missing certificate-fixture tests and 359 live-SQL tests)
- `.\scripts\bfmt.ps1` — passed
- `.\scripts\bclippy.ps1` — passed, including excluded `mssql-py-core`
- Focused ResultSet future-size, idempotent end-of-set, and prepared-capture cleanup tests — passed
- Workspace consumers (`mssql-js`, CLI, ODBC, and benchmarks) compile through the all-features/all-targets Clippy pass; no fuzz target references `ResultSet`

## Related Issues

Related to #265.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (blocked only by the 366 environment failures matching the lower-layer baseline)

### DIFF
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -36,8 +36,8 @@ use crate::{
     message::{batch::SqlBatch, messages::Request},
     token::tokens::{ColMetadataToken, CurrentCommand, DoneStatus, EnvChangeTokenSubType, Tokens},
 };
-use async_trait::async_trait;
 use std::collections::HashMap;
+use std::future::Future;
 use std::num::NonZeroU32;
 use tracing::{debug, error, info, instrument, warn};
 
@@ -2794,7 +2794,6 @@ impl TdsClient {
 
     /// This functions returns to the next row in the result set.
     /// If there are no more rows, it returns None.
-    #[instrument(skip(self), level = "info")]
     pub(crate) async fn get_next_row(&mut self) -> TdsResult<Option<Vec<ColumnValues>>> {
         let col_count = self
             .current_metadata
@@ -3601,8 +3600,8 @@ impl TdsClient {
     /// Uses `receive_row_into` to decode ROW/NBCROW tokens directly through
     /// `decode_into`, bypassing the intermediate `RowToken { all_values }`.
     /// Concrete writers stay concrete through the production transport and
-    /// decode chain. Calling the object-safe [`ResultSet::next_row_into`] method
-    /// remains supported for callers that intentionally hold a trait object.
+    /// decode chain. [`ResultSet::next_row_into`] provides the same operation
+    /// through statically dispatched trait calls.
     #[instrument(skip(self, writer), level = "info")]
     pub async fn next_row_into<W>(&mut self, writer: &mut W) -> TdsResult<bool>
     where
@@ -4507,7 +4506,6 @@ impl TdsClient {
     }
 }
 
-#[async_trait]
 impl ResultSet for TdsClient {
     fn get_metadata(&self) -> &Vec<ColumnMetadata> {
         // If no metadata is available, return an empty vector
@@ -4519,31 +4517,23 @@ impl ResultSet for TdsClient {
             .unwrap_or(&self.empty_metadata)
     }
 
-    #[instrument(skip(self), level = "info")]
-    async fn next_row(&mut self) -> TdsResult<Option<Vec<ColumnValues>>> {
-        let result = if self.maybe_has_unread_rows() {
-            self.get_next_row().await
-        } else {
-            Ok(None)
-        };
-        if result.is_err() {
-            self.abort_pending_prepare_capture();
-        }
-        result
+    fn next_row(&mut self) -> impl Future<Output = TdsResult<Option<Vec<ColumnValues>>>> + Send {
+        self.get_next_row()
     }
 
-    #[instrument(skip(self, writer), level = "info")]
-    async fn next_row_into(&mut self, writer: &mut (dyn RowWriter + Send)) -> TdsResult<bool> {
-        TdsClient::next_row_into(self, writer).await
+    fn next_row_into(
+        &mut self,
+        writer: &mut (dyn RowWriter + Send),
+    ) -> impl Future<Output = TdsResult<bool>> + Send {
+        TdsClient::next_row_into(self, writer)
     }
 
     fn maybe_has_unread_rows(&self) -> bool {
         !self.current_result_set_has_been_read_till_end
     }
 
-    #[instrument(skip(self), level = "info")]
-    async fn close(&mut self) -> TdsResult<()> {
-        self.close_query().await
+    fn close(&mut self) -> impl Future<Output = TdsResult<()>> + Send {
+        self.close_query()
     }
 }
 
@@ -4735,8 +4725,18 @@ enum ResultBoundaryKind {
     End,
 }
 
-/// Async result set iteration.
-#[async_trait]
+/// Async result set iteration through statically dispatched futures.
+///
+/// The returned futures are native, unboxed futures with an explicit [`Send`]
+/// guarantee.
+///
+/// # Dyn compatibility
+///
+/// This trait is intentionally not dyn-compatible and cannot be used through
+/// `dyn ResultSet`. This is a breaking change for trait-object consumers and for
+/// implementations written with `#[async_trait]`; concrete call sites can keep
+/// awaiting the methods unchanged, while implementations must return native
+/// `Send` futures.
 pub trait ResultSet {
     /// Returns the metadata of the result set.
     /// This metadata includes information about the columns in the result set.
@@ -4744,7 +4744,7 @@ pub trait ResultSet {
 
     /// Returns the next row of data as a vector of column values.
     /// If there is no more data, it returns None.
-    async fn next_row(&mut self) -> TdsResult<Option<Vec<ColumnValues>>>;
+    fn next_row(&mut self) -> impl Future<Output = TdsResult<Option<Vec<ColumnValues>>>> + Send;
 
     /// Decodes the next row directly into a [`RowWriter`], returning `true` if
     /// a row was written or `false` when the result set is exhausted.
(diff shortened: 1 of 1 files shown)


mssql-tds/src/connection/tds_client.rs
```diff
@@ -5099,6 +5103,20 @@ mod tests {
                 "{name} future is {size} B, expected <= {MAX} B"
             );
         }
+
+        let result_set_next_row = std::mem::size_of_val(&ResultSet::next_row(&mut client));
```
David-Engel, line 5133: **Suggestion:** these assertions don't actually guard against boxing. The test only checks `size <= 4096`; re-introducing `Pin<Box<dyn Future + Send>>` would make the future 16 B and pass silently, so the test doesn't protect the property this PR exists to establish. Asserting parity with the underlying native future does:

```rust
let result_set_next_row = std::mem::size_of_val(&ResultSet::next_row(&mut client));
let native_next_row = std::mem::size_of_val(&client.get_next_row());
assert_eq!(
    result_set_next_row, native_next_row,
    "ResultSet::next_row must forward the native future without boxing"
);
```

mssql-tds/src/connection/tds_client.rs
```diff
@@ -5099,6 +5103,20 @@ mod tests {
...
+        let result_set_next_row_into =
+            std::mem::size_of_val(&ResultSet::next_row_into(&mut client, &mut sink));
+
+        for (name, size) in [
+            ("ResultSet::next_row", result_set_next_row),
+            ("ResultSet::next_row_into", result_set_next_row_into),
```
David-Engel, line 5152: **Nit:** `close` is excluded from this budget check without explanation. The PR description notes it is 4,376 B and not a per-row operation — worth a one-line comment here so a future reader doesn't add it and read the failure as a regression.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -2784,7 +2784,6 @@ impl TdsClient {
 
     /// This functions returns to the next row in the result set.
     /// If there are no more rows, it returns None.
-    #[instrument(skip(self), level = "info")]
     pub(crate) async fn get_next_row(&mut self) -> TdsResult<Option<Vec<ColumnValues>>> {
```
David-Engel, line 2799: **Suggestion:** this `#[instrument]` removal is load-bearing but undocumented. I restored `#[instrument(skip(self), level = "info")]` here and `row_fetch_futures_stay_small` fails at **4,184 B > 4,096 B**, so it is required rather than incidental. It's also an observability change the description doesn't mention, and it removes the span for the non-`ResultSet` callers too (`metadata_retriever`, `describe_parameter_encryption`).

A one-line comment would stop it being re-added:

```rust
// Not instrumented: the span pushes ResultSet::next_row over the 4 KiB
// hot-path future budget. next_row_into still carries the row span.
```

The `next_row`/`close` impls are fine — `next_row_into` and `close_query` both keep their spans.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4744,14 +4745,17 @@ pub trait ResultSet {
     /// partially read. Draining that row here would silently discard it and
     /// return the *next* one, so callers must first finish the row with
     /// `next_row_cursor`. A fully-consumed or absent row is fine.
-    async fn next_row_into(&mut self, writer: &mut (dyn RowWriter + Send)) -> TdsResult<bool>;
+    fn next_row_into(
```
David-Engel, line 4768: **Suggestion (follow-up is fine):** `next_row_into` still takes `&mut (dyn RowWriter + Send)`. The only reason for the `dyn` was keeping `ResultSet` object-safe, which this PR deliberately gives up. Now that the trait is intentionally not dyn-compatible, the trait method could mirror the inherent one and drop the remaining writer vtable dispatch:

```rust
fn next_row_into<W>(&mut self, writer: &mut W) -> impl Future<Output = TdsResult<bool>> + Send
where
    W: RowWriter + Send + ?Sized;
```

The size test already tracks `next_row_into` vs `next_row_into (dyn writer)` separately, so the delta is measurable.

David-Engel (commented):
## Summary

Converts `ResultSet` from `#[async_trait]` to native RPITIT futures with explicit `+ Send`, and makes the `TdsClient` impl forward the existing native inherent futures. The change is correct and behavior-preserving. No blocking issues.

Reviewed on a dedicated worktree at head `418c3ab5`, diffed against the stacked base `dev/saurabh/remove-row-path-async-traits`. Verified locally:

- `cargo clippy --workspace --all-features --all-targets` — clean
- `mssql-py-core` (excluded from `bclippy`) — `cargo check --all-targets` clean
- `cargo test -p mssql-tds --lib` — 1,659 passed, only the 4 known certificate-fixture failures
- `cargo test -p mssql-tds --doc` — 9 failures, all pre-existing private-module doctests unrelated to this PR

### Claims I confirmed

**Behavior parity on `next_row`.** Dropping the outer `maybe_has_unread_rows()` guard and `abort_pending_prepare_capture()` is safe: `maybe_has_unread_rows()` is exactly `!current_result_set_has_been_read_till_end`, which the inherent `next_row_into` already short-circuits, and every error path in that method calls `abort_pending_prepare_capture()` before returning. The two capture-abort tests still cover it.

**No trait-object consumers.** Zero `dyn ResultSet`, a single `impl ResultSet for TdsClient`, and all consumers (`mssql-js`, `mssql-odbc`, `mssql-py-core`, `mssql-tds-cli`, benches, integration tests) go through concrete `TdsClient`. Method-call syntax (`client.next_row().await`) still compiles, so the "same source shape" claim holds.

### CI note (not a code finding)

`coverage-report` failed with "No build found for PR #291 after 30 attempts", and the ADO **mssql-rs Pull request validation** check is `skipping` — both because the PR targets the stacked branch instead of `main`. This is not the usual 75-minute artifact-timeout flake, and a re-run won't fix it. Net effect: this change gets no ADO validation and no 85% diff-coverage signal on its own. Worth confirming the full stack (#286 -> #287 -> this) is validated against `main` before merge.

Four non-blocking comments inline.

mssql-tds/src/connection/tds_client.rs
saurabh500: Documented the load-bearing absence of `#[instrument]` at `get_next_row`, including the 4 KiB budget reason and that `next_row_into` retains the row span.

mssql-tds/src/connection/tds_client.rs
saurabh500: Added exact size-parity assertions for both `ResultSet::next_row` and `ResultSet::next_row_into` against their underlying native futures. This now fails if either forwarding method is boxed, independently of the 4 KiB hot-path budget.

mssql-tds/src/connection/tds_client.rs
saurabh500: Added the one-line rationale beside the per-row checks: `close` runs once per result set, so its larger future is intentionally outside the row-iteration budget.

mssql-tds/src/connection/tds_client.rs
saurabh500: Agreed that genericizing the writer can remove the remaining writer vtable dispatch. I am keeping it for a focused follow-up because this layer intentionally isolates trait-future allocation and preserves the existing public method signature while changing object safety; combining both optimizations would blur that measurement and API migration.

David-Engel (approved):
## Re-review at `8c6ff8b2` — approving

All four comments from my earlier pass are addressed or explicitly deferred with a reasonable rationale. The only PR-specific commit since then is `1f6b8077 Strengthen ResultSet future guards` (+16 lines, test and comment only); the rest of the delta is the merge of the updated stack base.

Verified locally at this head:

- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean
- `cargo test -p mssql-tds --lib` — 1,666 passed, only the 4 known certificate-fixture failures (same baseline as before, +7 tests from the base merge)

### Comment resolution

**Boxing guard — fixed.** The new exact size-parity assertions are the guard this PR needed: a reintroduced `Pin<Box<dyn Future + Send>>` (16 B) now fails `assert_eq!` independently of the 4 KiB budget. The `next_row_into` comparison correctly monomorphizes the inherent method with `W = dyn RowWriter + Send`, so it is an apples-to-apples check against the trait's writer type.

**`#[instrument]` rationale — fixed.** The comment on `get_next_row` names the 4 KiB budget and notes `next_row_into` retains the row span, so the removal won't be silently undone.

**`close` budget exclusion — fixed.** The one-line rationale beside the per-row checks is exactly right.

**Generic writer on `next_row_into` — deferred, agreed.** Keeping trait-future allocation separate from writer dispatch is the right call for a clean measurement and a single API migration step. Worth filing an issue so it isn't lost.

### Remaining minor items (non-blocking)

- The description still doesn't mention the tracing change. The in-code comment is the more durable place and it's there, so this is optional — but the "Validation" section is what a future bisector reads first, and losing the `get_next_row` span also affects the `metadata_retriever` and `describe_parameter_encryption` callers, not just `ResultSet::next_row`.
- CI is unchanged: `coverage-report` fails with "No build found for PR #291" and ADO **Pull request validation** is `skipping`, both because the PR targets the stacked branch. Still no independent ADO validation or diff-coverage signal for this layer — please confirm the stack is validated against `main` before it merges.

David-Engel (approved):

David-Engel (approved):

PR #292, opened 2026-08-14
Add NBCROW row throughput benchmark

### LINKED ISSUE #282: Benchmark NBCROW row throughput to validate the null bitmap allocation optimization
### Problem statement

PR #245 optimizes the NBCROW null bitmap read path (allocate-once via `zeroed_bitmap`, then reuse the allocation across rows via a scratch cache). Review feedback on that PR — reasonably — asked whether the second half is measured to matter.

There is no benchmark today that can answer that. The relevant number is end-to-end rows/sec on a row-heavy scan, not `read_nbc_bitmap` in isolation, and the effect is highly shape-dependent:

- **All-fixed-width rows** (INT/BIGINT/DATETIME2/FLOAT/BIT) with more than 8 columns: nothing else in the row path allocates, so the null bitmap is 100% of per-row allocator traffic. This is where the scratch cache should show up.
- **Text-heavy rows** (NVARCHAR/VARBINARY): every string cell already allocates, so removing one allocation out of ~4 is expected to be in the noise.

The theoretical delta is also narrower than an allocation count suggests, because `Arc::get_mut` goes through `is_unique()`, which is a `compare_exchange`. Per row it is roughly:

| | per row |
|---|---|
| allocate each row | 1 malloc + 1 free + 1 CAS + 1 atomic decrement |
| reuse the allocation | 1 CAS + 1 atomic increment + 1 atomic decrement |

So this needs measuring rather than reasoning about.

### Proposed solution

Add an NBCROW row-throughput benchmark that measures the whole row path against `mssql-mock-tds`, not a single function:

- Parameterize over column count (e.g. 4 / 16 / 64) so the bitmap crosses the 1-byte, 2-byte, and 8-byte boundaries.
- Parameterize over column shape: all-fixed-width vs. text-heavy vs. mixed.
- Report rows/sec and allocations per row. A counting global allocator in the bench harness makes the allocation claim directly assertable rather than inferred.
- Include a NULL-density axis, since NBCROW is only emitted when the server sees nulls.

This should land alongside the broader row-path allocation work rather than as a one-off, so the same harness can cover the other allocations on the path (row buffers, column value decoding) instead of being NBCROW-specific.

### Affected crate

mssql-tds

### Alternatives considered

A criterion microbenchmark over `read_nbc_bitmap` alone. Rejected: it would measure `Arc::get_mut` against `malloc` in a tight loop with a hot allocator free-list and no surrounding work, which is not representative of the real path and risks producing a confidently wrong number in either direction.

### Additional context

Context and the full trade-off discussion: #245 (see the review discussion with @David-Engel).

saurabh500:
## Description

Adds an end-to-end Criterion benchmark for NBCROW row decoding through the loopback TDS transport and complete `next_row` path. The matrix covers 4, 16, and 64 columns across fixed-width, text-heavy, and mixed rows with 25% NULLs and 10,000 rows per result set by default.

The existing mock-server infrastructure now emits NBCROW tokens, fragments large responses into valid TDS packets, and caches immutable serialized responses before the timed loop so server-side serialization does not dominate the measurement.

This PR was originally stacked on #245. After #245 squash-merged as `77be0ff6`, the benchmark commit was restacked onto `main` so this PR contains only the issue #282 benchmark work.

Benchmark throughput on the stacked implementation ranged from 1.08 M rows/s for 4-column fixed-width rows to 105 K rows/s for 64-column text-heavy rows. A controlled comparison against the `zeroed_bitmap`-only implementation found all nine full-row decode deltas within the configured 5% Criterion noise threshold.

## Related Issues

Closes #282

Originally stacked on #245; now based on `main` because #245 merged.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes — attempted, but live SQL Server integration tests could not connect in the local environment; mock-specific tests pass

Additional validation:
- `cargo nextest run -p mssql-mock-tds --all-targets --no-fail-fast` — 20 passed
- `cargo bench -p mssql-tds-bench --bench nbcrow_rows --no-run`
- Full nine-scenario benchmark matrix with `BENCH_NBC_ROWS=10000`

### DIFF
--- a/mssql-mock-tds/src/protocol.rs
+++ b/mssql-mock-tds/src/protocol.rs
@@ -150,6 +150,7 @@ impl PacketHeader {
 pub enum TokenType {
     ColMetadata = 0x81,
     Row = 0xD1,
+    NbcRow = 0xD2,
     Done = 0xFD,
     DoneProc = 0xFE,
     DoneInProc = 0xFF,
@@ -873,7 +874,7 @@ pub fn build_query_result(response: &crate::query_response::QueryResponse) -> By
     // Serialize each column
     for col in &response.columns {
         result.put_u32_le(0); // UserType
-        result.put_u16_le(0x0000); // Flags: not nullable, no special flags
+        result.put_u16_le(u16::from(response.use_nbc_rows)); // nullable for NBCROW
         result.put_u8(col.data_type.tds_type_code());
         if col.data_type == crate::query_response::SqlDataType::NVarChar {
             // Required to support string responses (e.g., @@USERAGENT).
@@ -899,9 +900,23 @@ pub fn build_query_result(response: &crate::query_response::QueryResponse) -> By
 
     // Serialize each row
     for row in &response.rows {
-        result.put_u8(TokenType::Row as u8);
-        for value in &row.values {
-            value.write_to_buffer(&mut result);
+        if response.use_nbc_rows {
+            result.put_u8(TokenType::NbcRow as u8);
+            let mut bitmap = vec![0u8; response.columns.len().div_ceil(8)];
+            for (index, value) in row.values.iter().enumerate() {
+                if value.is_null() {
+                    bitmap[index / 8] |= 1 << (index % 8);
+                }
+            }
+            result.extend_from_slice(&bitmap);
+            for value in row.values.iter().filter(|value| !value.is_null()) {
+                value.write_to_buffer(&mut result);
+            }
+        } else {
+            result.put_u8(TokenType::Row as u8);
+            for value in &row.values {
+                value.write_to_buffer(&mut result);
+            }
         }
     }
 
@@ -959,16 +974,31 @@ pub fn build_error_response(message: &str) -> BytesMut {
 
 /// Wrap token data in a TDS packet
 fn wrap_in_packet(packet_type: PacketType, data: BytesMut) -> BytesMut {
-    let total_length = (PACKET_HEADER_SIZE + data.len()) as u16;
+    let max_payload = MAX_PACKET_SIZE - PACKET_HEADER_SIZE;
+    let packet_count = data.len().max(1).div_ceil(max_payload);
+    let mut packet = BytesMut::with_capacity(data.len() + packet_count * PACKET_HEADER_SIZE);
 
-    let mut packet = BytesMut::with_capacity(total_length as usize);
-    let header = PacketHeader::new(packet_type, total_length, 1);
-    header.write(&mut packet);
-    packet.extend_from_slice(&data);
+    if data.is_empty() {
+        PacketHeader::new(packet_type, PACKET_HEADER_SIZE as u16, 1).write(&mut packet);
+        return packet;
+    }
+
+    for (index, chunk) in data.chunks(max_payload).enumerate() {
+        let total_length = (PACKET_HEADER_SIZE + chunk.len()) as u16;
+        let mut header =
+            PacketHeader::new(packet_type, total_length, (index as u8).wrapping_add(1));
+        if index + 1 < packet_count {
+            header.status = PacketStatus::normal();
+        }
+        header.write(&mut packet);
+        packet.extend_from_slice(chunk);
+    }
 
     trace!(
-        "Built packet: type={:?}, length={}",
-        packet_type, total_length
+        "Built packet stream: type={:?}, packets={}, length={}",
+        packet_type,
+        packet_count,
+        packet.len()
     );
     packet
 }
@@ -1134,6 +1164,63 @@ mod tests {
         use crate::query_response::QueryResponse;
         let response = build_query_result(&QueryResponse::select_one());
         assert!(response.len() >= PACKET_HEADER_SIZE);
+    }
+
+    #[test]
+    fn test_nbcrow_result_serializes_bitmap_and_omits_null_values() {
+        use crate::query_response::{
+            ColumnDefinition, ColumnValue, QueryResponse, Row, SqlDataType,
+        };
+
+        let columns = (0..9)
+            .map(|_| ColumnDefinition::new("", SqlDataType::Int))
+            .collect();
+        let values = (0..9)
+            .map(|index| {
+                if index == 0 || index == 8 {
+                    ColumnValue::Null
+                } else {
+                    ColumnValue::Int(index)
+                }
+            })
+            .collect();
+        let response = build_query_result(
+            &QueryResponse::new(columns, vec![Row::new(values)]).with_nbc_rows(),
+        );
+
+        let row_offset = PACKET_HEADER_SIZE + 3 + (9 * 9);
+        assert_eq!(response[row_offset], TokenType::NbcRow as u8);
+        assert_eq!(&response[row_offset + 1..row_offset + 3], &[0x01, 0x01]);
+        assert_eq!(response[row_offset + 3], 4);
+        assert_eq!(
+            i32::from_le_bytes(
+                response[row_offset + 4..row_offset + 8]
+                    .try_into()
+                    .expect("first non-null integer should have four bytes")
+            ),
+            1
+        );
+    }
+
+    #[test]
+    fn test_large_payload_is_split_into_tds_packets() {
+        let response = wrap_in_packet(
+            PacketType::TabularResult,
+            BytesMut::from(&vec![0xAA; MAX_PACKET_SIZE][..]),
+        );
+
+        let mut first = &response[..PACKET_HEADER_SIZE];
+        let first_header =
+            PacketHeader::parse(&mut first).expect("first packet header should parse");
+        assert_eq!(usize::from(first_header.length), MAX_PACKET_SIZE);
+        assert!(!first_header.status.is_end_of_message());
+
+        let second_offset = usize::from(first_header.length);
+        let mut second = &response[second_offset..second_offset + PACKET_HEADER_SIZE];
+        let second_header =
+            PacketHeader::parse(&mut second).expect("second packet header should parse");
+        assert!(second_header.status.is_end_of_message());
+        assert_eq!(usize::from(second_header.length), PACKET_HEADER_SIZE * 2);
     }
 
     #[test]
--- a/mssql-mock-tds/src/query_response.rs
+++ b/mssql-mock-tds/src/query_response.rs
@@ -3,7 +3,7 @@
 
 //! Query response definitions for the mock TDS server
 
-use bytes::{BufMut, BytesMut};
+use bytes::{BufMut, Bytes, BytesMut};
 use std::collections::HashMap;
 
 /// SQL data types supported by the mock server
@@ -57,6 +57,10 @@ pub enum ColumnValue {
 }
 
 impl ColumnValue {
+    pub(crate) fn is_null(&self) -> bool {
+        matches!(self, Self::Null)
+    }
+
     /// Get the SQL data type for this value
     pub fn data_type(&self) -> SqlDataType {
         match self {
@@ -193,6 +197,7 @@ pub struct QueryResponse {
     /// An error emitted (with a DONE MORE token) before the result set, so the
     /// server keeps streaming the row set after a statement-scoped error.
     pub leading_error: Option<LeadingError>,
+    pub(crate) use_nbc_rows: bool,
 }
 
 impl QueryResponse {
@@ -203,6 +208,7 @@ impl QueryResponse {
             rows,
             info_tokens: Vec::new(),
             leading_error: None,
+            use_nbc_rows: false,
         }
     }
 
@@ -218,6 +224,12 @@ impl QueryResponse {
         self
     }
 
+    /// Encode rows as NBCROW tokens with a null bitmap.
+    pub fn with_nbc_rows(mut self) -> Self {
+        self.use_nbc_rows = true;
+        self
+    }
+
     /// Helper to create a response for SELECT 1
     pub fn select_one() -> Self {
         Self {
@@ -225,6 +237,7 @@ impl QueryResponse {
             rows: vec![Row::new(vec![ColumnValue::Int(1)])],
             info_tokens: Vec::new(),
             leading_error: None,
+            use_nbc_rows: false,
         }
     }
 
@@ -243,6 +256,7 @@ impl QueryResponse {
             ])],
             info_tokens: Vec::new(),
             leading_error: None,
+            use_nbc_rows: false,
         }
     }
 }
@@ -250,6 +264,7 @@ impl QueryResponse {
 /// Registry of query responses
 pub struct QueryRegistry {
     responses: HashMap<String, QueryResponse>,
+    wire_responses: HashMap<String, Bytes>,
 }
 
 impl QueryRegistry {
@@ -257,6 +272,7 @@ impl QueryRegistry {
     pub fn new() -> Self {
         let mut registry = Self {
             responses: HashMap::new(),
+            wire_responses: HashMap::new(),
         };
 
         // Add default responses
@@ -272,12 +288,18 @@ impl QueryRegistry {
     /// Register a query response
     pub fn register(&mut self, query: impl Into<String>, response: QueryResponse) {
         let query = query.into().to_uppercase();
-        self.responses.insert(query, response);
+        let wire_response = crate::protocol::build_query_result(&response).freeze();
+        self.responses.insert(query.clone(), response);
+        self.wire_responses.insert(query, wire_response);
     }
 
     /// Get a response for a query
     pub fn get(&self, query: &str) -> Option<&QueryResponse> {
         self.responses.get(&query.to_uppercase())
+    }
+
+    pub(crate) fn get_wire_response(&self, query: &str) -> Option<Bytes> {
+        self.wire_responses.get(&query.to_uppercase()).cloned()
     }
 }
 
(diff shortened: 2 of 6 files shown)


mssql-tds-bench/benches/nbcrow_rows.rs
```diff
@@ -0,0 +1,179 @@
...
+        })
+        .collect();
+
+    QueryResponse::new(columns, rows).with_nbc_rows()
+}
+
+async fn connect_mock(address: SocketAddr) -> TdsClient {
+    let mut context = ClientContext::default();
+    context.user_name = "sa".to_string();
+    context.password = "MockBenchmark1!".to_string();
```
saurabh500, line 97: This needs to be randomly generated. WE cannot have hardcoded literals in password

David-Engel (commented):
## Summary

Solid, well-scoped addition. The mock-side changes are the interesting part and they hold up: the NBCROW null bitmap is written LSB-first per byte (`bitmap[index / 8] |= 1 << (index % 8)`), which matches exactly what the client does in `mssql-tds/src/token/parsers/nbcrow_parser.rs::is_null_value_in_column`, and every integer column already serializes as `INTNTYPE` (0x26) with a length prefix, so omitting NULL payload bytes stays spec-valid rather than depending on a non-nullable fixed-length type. The packet-fragmentation change also fixes a latent bug on `main`: `wrap_in_packet` previously computed `(PACKET_HEADER_SIZE + data.len()) as u16`, which silently truncated for any response above 64 KB. Per-chunk `PacketStatus::normal()` with EOM only on the final packet, and a `packet_id` that increments modulo 256, are both correct.

### Verified locally on `801f4139`

- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean.
- `cargo nextest run -p mssql-mock-tds --all-targets` — 20/20 pass, including the two new tests.
- `cargo nextest run -p mssql-tds -E 'test(/mock/)'` — passes except the 5 strict-TLS tests, which fail only because `scripts/generate_mock_tds_server_certs.ps1` has not been run in this environment (pre-existing, unrelated).
- `cargo nextest run -p mssql-tds-bench --all-targets` — all 9 `nbcrow_row_throughput/*` scenarios pass in Criterion test mode at ~4s each, comfortably under the `ci` profile's 60s slow-timeout. Worth calling out explicitly since `cargo btest` uses `--all-targets`, so these benchmarks now execute on every CI test run; they do not need a live SQL Server, so that is fine, but it is new CI work that did not exist before.

The remaining `cargo btest` failures here are the live-SQL-Server integration suites, consistent with the unchecked checklist item in the description.

### Findings

**1. Hardcoded credential literal in the benchmark — `mssql-tds-bench/benches/nbcrow_rows.rs:97`**

```rust
context.password = "MockBenchmark1!".to_string();
```

The mock never validates credentials, so any value works, but this deviates from both neighbors: `mssql-tds-bench/src/lib.rs::create_context` reads `SQL_PASSWORD` from the environment, and `mssql-tds/tests/test_mock_server.rs` deliberately uses a `generate_test_password()` helper (RNG-generated, 24 chars) precisely so no password literal lands in source. A literal that looks like a real password is also the shape credential scanners flag. Please reuse the random-password approach or pass an empty string.

**2. Magic nullable flag — `mssql-mock-tds/src/protocol.rs:877`**

```rust
result.put_u16_le(u16::from(response.use_nbc_rows)); // nullable for NBCROW
```

`u16::from(bool)` happens to produce the correct COLMETADATA `fNullable` bit, but it encodes that protocol fact as a bool-to-int coercion. A named `const COLUMN_FLAG_NULLABLE: u16 = 0x0001;` would make the wire contract self-evident and survive anyone later adding a second flag.

Related, and optional: tying nullability to `use_nbc_rows` leaves the plain-ROW path advertising non-nullable columns while still writing zero-length NULL values for `ColumnValue::Null`. Deriving the flag from `response.rows.iter().any(|r| r.values[i].is_null())` would make both paths honest. Pre-existing behavior, so fine to defer.

**3. `QueryRegistry` now retains two representations of every response — `mssql-mock-tds/src/query_response.rs:291`**

`register` keeps the `QueryResponse` *and* its serialized `Bytes`, and after this change nothing in-crate calls `get()` anymore (only the retained public API does). For the 64-column text-heavy scenario that is tens of MB held twice per registration. Caching the wire bytes is the right call for the benchmark; the duplication just deserves either a one-line comment on the field explaining why both are kept, or dropping the `QueryResponse` copy if the public `get()` has no consumers.

### Nits

- `mssql-tds-bench/benches/nbcrow_rows.rs:143` — `bench_with_input(..., &column_count, |b, _| ...)` ignores the input; `bench_function` is equivalent and says less.
- The mock emits NBCROW for *every* row once the flag is set, so a result set can never mix ROW and NBCROW the way a real server does. Correct for a throughput benchmark; just not a substitute for a mixed-token decode test later.
- `mssql-mock-tds/FLEXIBLE_QUERY_EXECUTION.md:50` reproduces the `QueryResponse` struct definition, which now has an extra private field. Low priority, but the doc is drifting.

### CI

`coverage-report` is red on this head. The log shows it is not a code failure — it polled for 75 minutes for the ADO coverage artifact, never saw one, and timed out. Everything else, including the full ADO `mssql-rs Pull request validation` build 167183, is green. Please re-run that job (or confirm the artifact upload) before this leaves draft, since a red required check will block merge regardless of cause.

David-Engel (commented):
## Review: NBCROW row throughput benchmark (801f413)

Reviewed the full diff in a dedicated worktree and validated locally against `801f413`:

- `cargo clippy -p mssql-mock-tds -p mssql-tds-bench --all-targets --all-features` — clean
- `cargo nextest run -p mssql-mock-tds --all-targets` — 20/20 pass, including the two new tests
- `cargo nextest run -p mssql-tds --test test_mock_server --test test_mock_server_tls --test test_mock_server_fedauth --test test_drain_on_error --test test_redirection` — 25 pass, 5 fail only because the strict-TLS tests need generated mock certs in this environment (pre-existing/environmental, unrelated to the change)
- `cargo bench -p mssql-tds-bench --bench nbcrow_rows` with `BENCH_NBC_ROWS=500` and short Criterion settings — all nine scenarios run end-to-end against the mock

The packet-fragmentation fix in `wrap_in_packet` is a genuine improvement independent of the benchmark: the previous single-packet path wrote a `u16` length for payloads that could exceed a 4096-byte packet, so large mock result sets were not valid TDS. Caching serialized responses in `QueryRegistry` also keeps mock-side serialization out of the timed loop, which is the right call for this measurement. Issue link (`Closes #282`) is present.

Comments below, none blocking.

### 1. NBCROW value correctness is not asserted end-to-end (medium)

`drain()` only counts rows, and the new unit test covers one 9-column INTN row. Nothing asserts that a NULL in the bitmap decodes to a NULL in the *right column* through the real client, and benchmarks do not run in CI, so the mock's bitmap ordering has no CI guard. Consider a mock-based integration test in `mssql-tds/tests/test_mock_server.rs` that registers a `with_nbc_rows()` response mixing `Int`/`BigInt`/`NVarChar` with NULLs at known positions and asserts per-column values. That test would also exercise the new multi-packet split with a real client, which currently has coverage only at the raw byte level.

### 2. No plain-ROW control in the matrix (medium)

The matrix measures absolute throughput of NBCROW decoding, but issue #282 is about NBCROW cost. Without a `ROW`-encoded control at the same shapes and column counts, a regression in generic row decoding and a regression in NBCROW handling look identical. `with_nbc_rows()` already makes the encoding a toggle, so parameterizing the encoding and adding `row` / `nbcrow` to the `BenchmarkId` is a small change. If the intent is purely absolute regression tracking, a sentence in `mssql-tds-bench/README.md` saying so would set expectations.

### 3. Bitmap indexing can panic on a width mismatch (low)

In `build_query_result`, the bitmap is sized from `response.columns.len()` but indexed by the row value index: a row with more values than columns panics on `bitmap[index / 8]`. With eager serialization this now fires inside `register()`, which is better timing, but an index panic is a poor diagnostic. A `debug_assert_eq!(row.values.len(), response.columns.len())` or zipping values against columns would fail with an actionable message.

### 4. Flags value would read better as a named constant (low)

`result.put_u16_le(u16::from(response.use_nbc_rows))` relies on `0x0001` being `fNullable`, which is correct (`colmetadata_parser.rs` documents `0x0001 = Nullable`) but is not obvious at the call site. A `const COLUMN_FLAG_NULLABLE: u16 = 0x0001;` with an explicit `if` would tie it to the spec bit. Worth noting in passing that enabling NBCROW flips nullability for every column, so any existing test that switches to `with_nbc_rows()` also changes its COLMETADATA flags.

### 5. Minor cleanups in `wrap_in_packet` (low)

- `data.len().max(1)` is dead once the `data.is_empty()` early return exists; computing `packet_count` after the guard drops the `.max(1)`.
- `(index as u8).wrapping_add(1)` gives the correct mod-256 PacketID sequence, but a short comment saying that is cheaper than the next reader re-deriving it from MS-TDS.
- The mock always emits 4096-byte packets regardless of the packet size a client requested in LOGIN7. Fine for current usage and unchanged by this PR, but now that responses legitimately span packets, a client that negotiates a smaller size would receive oversized packets.

### 6. Registry now holds two copies of every response (low)

`QueryRegistry` keeps both the `QueryResponse` and its serialized `Bytes`. Harmless for a mock, but at 64 columns x 10,000 rows the duplication is real; worth remembering if benchmark row counts grow. Relatedly, `ConnectionProcessor::process_packet` changed from `BytesMut` to `Bytes` — a public signature change in the mock crate with no in-repo breakage, just calling it out.

### 7. Checklist / process (low)

The description notes `cargo btest` could not complete because live SQL Server integration tests could not connect locally. Before flipping out of draft, please confirm PR validation covers that, and follow the usual flow: Copilot review, green validation on the latest commit, then ready for review.

mssql-tds-bench/Cargo.toml
```diff
@@ -29,6 +29,7 @@ workspace = true
 # across both runs, so any statistically significant delta is attributable to
 # mssql-tds itself.
 mssql-tds = { path = "../mssql-tds", features = ["test-util"] }
+mssql-mock-tds = { path = "../mssql-mock-tds" }
```
David-Engel: Two things here.

First, the small one: `mssql-mock-tds` is only used by the `nbcrow_rows` bench target, never by the `mssql-tds-bench` library itself. `[[bench]]` targets link `[dev-dependencies]` fine, so moving it there keeps the lib's dependency surface honest.

Second, the one that could bite in CI: `mssql-mock-tds` has its own `mssql-tds = { path = "../mssql-tds" }`. `perf-lab/run-benchmarks.ps1` builds the baseline by swapping the `../mssql-tds` *source* in place (`Set-BaselineSource`) and rebuilding, so the baseline build now also compiles the current-tree `mssql-mock-tds` against the *pinned baseline* `mssql-tds`. Today the mock only reaches for `mssql_tds::core::TDS_8_ALPN_PROTOCOL` (server.rs:626), so it happens to compile. But the first time the mock uses a symbol that doesn't exist at the commit in `baseline-commit.txt`, the baseline build fails and the runner throws `no baseline bench binaries found` — a red perf pipeline whose cause is several layers away from the change that triggered it.

Worth either dropping the mock's dependency on `mssql-tds` (it's one constant — it could be a local `const`), or amending the comment directly above this line, which currently promises the harness stays "byte-for-byte identical across both runs". That's no longer strictly true.

David-Engel (commented):
## Review — NBCROW row throughput benchmark

Nice addition: this is the first bench in the harness that runs without a live SQL Server, and the mock-side packet splitting fixes a real latent bug (`wrap_in_packet` previously truncated `total_length` into a `u16`, so any response over ~64 KB was silently malformed — which is exactly what a 10,000-row result set would have produced).

**Verification performed in a clean worktree at 801f413:**
- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean
- `cargo nextest run -p mssql-mock-tds --all-targets` — 20/20 pass, including both new tests
- `cargo nextest run -p mssql-tds` for the five mock-backed integration suites (`test_mock_server`, `test_mock_server_tls`, `test_mock_server_fedauth`, `test_redirection`, `test_drain_on_error`) — 30/30 pass after generating the local test certificates, so the packet-splitting and `BytesMut` → `Bytes` changes don't regress the TLS/strict-TLS, FedAuth, or redirection paths
- `cargo bench --bench nbcrow_rows -- --test` at both `BENCH_NBC_ROWS=200` and the default `10000` — all 9 scenarios succeed. The default run produces multi-megabyte responses (~3,800 packets for the 64-column text-heavy case), so it does exercise the `PacketID` wrap past 255; the client handles it, consistent with MS-TDS 2.2.3.1.3 defining PacketID as incrementing modulo 256.

The NBCROW encoding checks out against the spec: `ceil(columns/8)` bitmap bytes, LSB-first bit order, null values omitted from the payload, and `fNullable` (0x0001) set in COLMETADATA flags so the client accepts them. All the mock's integer types already map to `INTN`, so nullability is representable.

**No blocking issues.** Three suggestions and two nits below, none of which need to hold up the PR.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 3 |
| Nit | 2 |

One thing worth thinking about beyond the inline comments: because the mock server runs in-process on the same multi-threaded Tokio runtime as the client, the measured time includes mock-side query dispatch and serialization lookup as well as client decode. The module doc already says as much, so this is fair — just be aware that a future change to mock-side cost (say, registry locking) would show up as an apparent `mssql-tds` regression.

mssql-mock-tds/src/protocol.rs
```diff
@@ -959,16 +974,31 @@ pub fn build_error_response(message: &str) -> BytesMut {
 
 /// Wrap token data in a TDS packet
 fn wrap_in_packet(packet_type: PacketType, data: BytesMut) -> BytesMut {
-    let total_length = (PACKET_HEADER_SIZE + data.len()) as u16;
+    let max_payload = MAX_PACKET_SIZE - PACKET_HEADER_SIZE;
```
David-Engel: Chunking happens at the mock's fixed `MAX_PACKET_SIZE` (4096) and ignores the packet size the client requested in LOGIN7. That's fine — arguably better than before — for the functional tests, and a server is free to send smaller packets than negotiated.

The reason I'd call it out anyway is that it quietly pins the new NBCROW benchmark to 4096-byte framing no matter what `ClientContext::packet_size` says. That means its throughput numbers aren't directly comparable with `tds_specific`'s `packet_size_sensitivity` bench, and a regression in large-packet reassembly won't surface through this bench at all. Threading the negotiated size through here would be the thorough fix; at minimum, a line in `nbcrow_rows.rs`'s module doc noting the fixed 4096-byte framing would stop someone from over-reading the result later.

mssql-mock-tds/src/protocol.rs
```diff
@@ -959,16 +974,31 @@ pub fn build_error_response(message: &str) -> BytesMut {
...
+    let packet_count = data.len().max(1).div_ceil(max_payload);
```
David-Engel: Nit: `.max(1)` can't do anything here — the `data.is_empty()` early return two lines below already covers the zero-length case, so `data.len()` is always at least 1 by the time this value is used. Dropping it (`data.len().div_ceil(max_payload)`) removes the implication that an empty payload can reach the loop.

mssql-mock-tds/src/query_response.rs
```diff
@@ -272,13 +288,19 @@ impl QueryRegistry {
     /// Register a query response
     pub fn register(&mut self, query: impl Into<String>, response: QueryResponse) {
         let query = query.into().to_uppercase();
-        self.responses.insert(query, response);
+        let wire_response = crate::protocol::build_query_result(&response).freeze();
```
David-Engel: Pre-serializing at registration is the right call for a benchmark — keeping token building out of the timed loop is exactly what you want.

The cost is that `responses` and `wire_responses` become two sources of truth. After this change nothing in the crate reads `get()` any more, and a caller who still does receives a `&QueryResponse` that is no longer guaranteed to describe what the server actually writes. A one-line doc note on `register` (and/or `get`) saying the wire bytes are snapshotted at registration time would save the next person that discovery.

Also worth being aware of: `build_query_result` now runs inside `register`, so an over-long `Row` (more values than columns) panics on `bitmap[index / 8]` at registration rather than emitting malformed bytes at request time. That's a better failure mode, not a complaint — just a behavior change that isn't obvious from the diff.

mssql-tds-bench/benches/nbcrow_rows.rs
```diff
@@ -0,0 +1,179 @@
...
+
+fn response(shape: RowShape, column_count: usize, row_count: usize) -> QueryResponse {
+    let columns: Vec<_> = (0..column_count)
+        .map(|column| ColumnDefinition::new(format!("c{column}"), shape.data_type(column)))
+        .collect();
+    let rows = (0..row_count)
+        .map(|row| {
+            let values = (0..column_count)
+                .map(|column| {
+                    if (row + column).is_multiple_of(100 / NULL_PERCENT) {
```
David-Engel: Nit: `100 / NULL_PERCENT` is integer division, and `NULL_PERCENT` is also interpolated straight into the benchmark id (`{column_count}_columns_{NULL_PERCENT}_percent_null`, line 146). At 25 the two agree exactly — 1 column in 4. Change the constant to, say, 30 and the id advertises `30_percent_null` while the data is 33%.

Since Criterion matches saved baselines by benchmark id, an id that can silently drift from the data it names is worth guarding. Either `const NULL_EVERY: usize = 4;` with the percentage derived from it, or a `debug_assert!(100 % NULL_PERCENT == 0)` next to the constant, would keep the label honest.

PR #293, opened 2026-08-14
Add ODBC connection pooling design plan (ADO #47317)

saurabh500:
## Summary

Base PR for the **mssql-odbc connection pooling** effort ([ADO #47317](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47317)). Adds the design plan at `docs/odbc-connection-pooling-plan.md`.

The plan enables `mssql-python`'s existing client-side pool to safely reuse a physical Rust ODBC connection — **no second pool in Rust**; we supply only the ODBC/TDS reset, liveness, transaction, and isolation semantics that make reuse safe.

### Key findings captured in the doc
- Cross-referenced the original `msodbcsql` C++ driver (discoveries D1–D9).
- **D9:** `sp_reset_connection` does **not** reset the transaction isolation level (a SQL Server limitation). `mssql-python` works around this in [PR #343](https://github.com/microsoft/mssql-python/pull/343) by re-applying `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` on every checkout — our `SQL_ATTR_TXN_ISOLATION` handler must honor this and carry the armed reset bit.
- **D3:** confirmed gap — fatal error tokens (severity ≥ 20) don't currently mark the transport dead.

## Implementation stages (tracked as ADO child Tasks of #47317)
1. **Stage 1** — mssql-tds reset correctness (reset-ack centralization + fatal-error liveness) — [[AB#47337](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47337)]
2. **Stage 2** — ODBC pooling primitives (`SQL_ATTR_CONNECTION_DEAD` + `SQL_ATTR_RESET_CONNECTION`) — [[AB#47338](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47338)]
3. **Stage 3** — Reset acking + isolation carrier — [[AB#47339](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47339)]
4. **Stage 4** — Auth/recovery, ODBC tests, mssql-python e2e — [[AB#47340](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47340)]

Each stage will land as its own PR stacked on top of this base.

## Notes
[...]

PR #294, opened 2026-08-14
Stage 1: mssql-tds reset correctness (ADO #47337)

saurabh500:
## Summary

Stage 1 of the mssql-odbc connection pooling effort (ADO User Story #47317). This is the bottom implementation layer, scoped to **`mssql-tds` only** — it makes a physical connection reset actually clean the session-bound client state so a pool can safely hand a reused connection to the next borrower. `mssql-odbc` wiring comes in later stages.

Implements Workstream A (A1, A3) and discovery D3 from `docs/odbc-connection-pooling-plan.md`.

### A1 — Centralize the ResetConnection ENVCHANGE acknowledgement
Previously the ResetConnection ENVCHANGE was handled in 5 token-draining spots that each only called `session_state_table.reset()`, leaving session-bound client caches stale after a real server reset.

- New private `TdsClient::on_reset_connection_ack()` performs the **full** post-reset transition: resets the session-state table, clears `prepared_handles` / `prepared_param_encryption` / `pending_capture` / `pending_prepared_param_encryption`, and restores the negotiated **DATABASE / LANGUAGE / COLLATION** back to their login values (so a borrower's `USE otherdb` / `SET LANGUAGE` doesn't persist across a reset).
- The cache-clearing is factored into a shared `clear_session_bound_caches()` helper reused by the reconnect path.
- All 5 ResetConnection sites now call the single method; the following `capture_change_property` ordering is preserved (login-default restore moved out of `execution_context` into the single owner to avoid duplication).
- Per **D5 (verified)**, no post-login ANSI SET-option batch is re-emitted — ANSI defaults ride the `fODBC` LOGIN7 bit that `sp_reset_connection` restores. No `BuildServerSideConnectOptions` equivalent added.

### D3 — Mark the transport dead on fatal server error tokens
`known_dead` was only set on socket write/read failure, EOF, or explicit close; a class ≥ 20 fatal error on a still-open socket left `is_connection_dead()` reporting alive.

- Added `TdsTransport::mark_known_dead()` (default no-op; overridden by `NetworkTransport`).
- The shared error-token path (`record_error_token`) now marks the transport known-dead when `severity >= 20`, so a later pool checkout discards it.

### A3 — Tests
- Unit test asserting the reset transition clears all four caches and restores DATABASE/LANGUAGE/COLLATION.
- Unit tests for the D3 fatal-error path (severity ≥ 20 marks dead; severity < 20 does not).
- Live test in `tests/test_reset_connection.rs` proving a prepared handle is invalidated across a reset (gated on a live SQL Server via `.env`, like the sibling reset tests).

## Validation
- `cargo bfmt` ✅
- `cargo bclippy` (warnings-as-errors) ✅
- `cargo nextest` — new/related tests pass. Pre-existing certificate/TLS-validation test failures are unrelated to this change (they fail on the base branch too, expired cert fixtures).

## Notes
- Stacked on base PR #293 (`saurabh500-odbc-connection-pooling-plan`); this PR targets that branch, not `main`.
- Refs `[AB#47337](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47337)`.

### DIFF
--- a/mssql-tds/src/connection/execution_context.rs
+++ b/mssql-tds/src/connection/execution_context.rs
@@ -180,12 +180,12 @@ impl ExecutionContext {
             EnvChangeTokenSubType::ResetConnection => {
                 // Server acknowledgement that the connection was reset to login
                 // defaults (in response to a RESETCONNECTION / RESETCONNECTIONSKIPTRAN
-                // request). The server does not emit individual Database/Language
-                // ENVCHANGE tokens for the revert, so restore the cached session
-                // state to the login defaults here to keep the pool-facing getters
-                // accurate.
+                // request). The full client-side transition — restoring the
+                // negotiated database/language/collation and clearing
+                // session-bound caches — is applied by
+                // `TdsClient::on_reset_connection_ack`, which runs before this
+                // token is captured, so nothing is done here beyond logging.
                 info!("Connection reset acknowledged by server");
-                negotiated_settings.restore_login_defaults();
                 Ok(())
             }
             EnvChangeTokenSubType::UserInstanceName => {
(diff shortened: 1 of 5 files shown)


saurabh500:
Consolidated into #308, which contains all of this work in a single PR against `main`. Closing in favor of #308.

PR #299, opened 2026-08-14
Improve decoder throughput with buffered scalar reads

saurabh500:
## Description

Improves row-decoder throughput by synchronously consuming fixed-width scalars that are already buffered in `NetworkTransport`. A buffer miss falls back to the existing async TDS packet read/refill path and retries, so packet framing and error handling remain authoritative.

The optimized decoder reads cover byte, little-endian `i16`, `u16`, 24-bit `u32`, `i32`, `u32`, 40-bit `u64`, `i64`, `f32`, and `f64`. This is a standalone runtime optimization informed by measurements from #269. It does not revive that PR or retry decoder convergence.

### Stack position

This draft is stacked directly on #291 (`Remove ResultSet async_trait boxing`) in native stack #287. Its incremental diff is the two buffered-read commits above #291; the current open chain is #286 → #291 → this PR.

### Implementation

- Adds explicit non-consuming-on-miss probes to both `TdsPacketReader` configurations, with safe `None` defaults for other readers.
- Implements direct fixed-width probes in `TdsReadBuffer` and delegates them from `NetworkTransport`.
- Uses one parameterized `read_sync_first!` macro at every production fixed-width scalar read in `datatypes/decoder.rs`.
- Keeps `read_tds_packet()` as the only refill path for async misses.
- Leaves big-endian packet/header reads and `u64` reads outside the decoder unchanged.

#### Before: every scalar enters the async path

```mermaid
sequenceDiagram
    participant D as Decoder
    participant A as async read_*()
    participant B as TdsReadBuffer
    participant P as Packet I/O

    D->>A: read_*().await
    Note over D,A: Construct and poll a future for every scalar
    A->>B: Enough bytes buffered?
    alt Buffered hit
        B-->>A: Yes
        A->>B: Consume N bytes
        A-->>D: Ready(value)
    else Buffer miss
        B-->>A: No
        A->>P: read_tds_packet().await
        P-->>A: Next framed payload
        A->>B: Retry and consume N bytes
        A-->>D: Ready(value)
    end
```

#### After: synchronous probe before async fallback

```mermaid
sequenceDiagram
    participant D as Decoder
    participant M as sync-first macro
    participant T as NetworkTransport
    participant B as TdsReadBuffer
    participant P as Packet I/O

    D->>M: read_sync_first!
    M->>T: try_read_*() [sync]
    T->>B: Probe complete N-byte scalar
    alt Buffered hit
        B-->>T: Some(value), consume N bytes
        T-->>M: Some(value)
        M-->>D: value
        Note over D,M: No async future constructed or polled
    else Buffer miss
        B-->>T: None, consume zero bytes
        T-->>M: None
        M->>T: read_*().await
        T->>P: read_tds_packet().await
        P-->>T: Next framed payload
        T->>B: Refill, retry, consume complete scalar
        T-->>M: Ok(value)
        M-->>D: value
    end
```

`N` is the scalar's fixed wire width (1, 2, 3, 4, 5, or 8 bytes). Packet framing, cancellation, encryption, and errors remain in the existing async fallback.

### Correctness coverage

Targeted tests cover successful buffered reads, zero consumption when any supported scalar is incomplete, and every supported scalar split across real TDS packet boundaries through `NetworkTransport`. Existing NULL and length-marker handling remains unchanged.

### Production-reader benchmarks

All benchmarks use concrete `NetworkTransport`, packetized in-memory input, 7,992-byte packet payloads, 30,000 rows per pass, 2 warmups, 9 measured passes per cell, and 8 paired rounds alternating baseline/candidate order. Negative deltas are faster.

#### Stacked result versus #291 (`8c6ff8b2`)

| Workload | Sink | Paired median | Range | #291 median | Stacked median |
| --- | --- | ---: | ---: | ---: | ---: |
| fixed scalars | discard | -38.16% | -40.12% to -37.12% | 70.23 ms | 43.40 ms |
| fixed scalars | materialize | -25.23% | -26.53% to -21.80% | 104.24 ms | 78.40 ms |
| INT/VARCHAR | discard | -14.70% | -17.96% to -12.73% | 58.55 ms | 50.04 ms |
| INT/VARCHAR | materialize | -16.63% | -17.86% to -14.91% | 95.57 ms | 79.60 ms |
| mixed | discard | -12.90% | -16.72% to -10.83% | 123.27 ms | 107.37 ms |
| mixed | materialize | -11.06% | -12.90% to -8.72% | 163.29 ms | 145.73 ms |

The stack comparison used identical lockfiles and isolated source/target directories. The #291 and stacked test binaries had different SHA-256 hashes.

#### Initial change versus clean `main` (`ac1023a1`)

| Workload | Sink | Paired median | Range | Main median | Candidate median |
| --- | --- | ---: | ---: | ---: | ---: |
| INT/VARCHAR | discard | -6.06% | -15.86% to +11.50% | 62.25 ms | 58.46 ms |
| INT/VARCHAR | materialize | -14.32% | -21.77% to -11.27% | 99.75 ms | 85.89 ms |
| mixed | discard | -11.04% | -22.62% to -1.52% | 128.02 ms | 112.58 ms |
| mixed | materialize | -10.49% | -20.84% to +3.20% | 172.04 ms | 153.39 ms |

#### Expanded fixed-width reads versus the first draft (`47c2fb68`)

The `fixed_scalars` workload has 64 columns: `Int1`, `Int2`, `Int4`, `Int8`, `Flt4`, `Flt8`, `DateTime`, and `DateTim4`, repeated eight times.

| Workload | Sink | Paired median | Range | First draft median | Expanded median |
| --- | --- | ---: | ---: | ---: | ---: |
| fixed scalars | discard | -18.14% | -20.76% to -16.08% | 95.55 ms | 78.48 ms |
| fixed scalars | materialize | -9.62% | -21.52% to +6.46% | 155.71 ms | 140.14 ms |
| INT/VARCHAR | discard | +0.06% | -3.87% to +3.78% | 93.63 ms | 94.03 ms |
| INT/VARCHAR | materialize | +1.01% | -10.87% to +13.10% | 144.72 ms | 145.27 ms |
| mixed | discard | -0.95% | -4.79% to +2.29% | 193.05 ms | 189.67 ms |
| mixed | materialize | -1.02% | -5.07% to +9.28% | 264.07 ms | 261.60 ms |

The optimized pre-stack benchmark executable grew by about 192 KB (1.75%), so code-size impact remains a review consideration.

### Validation

- `cargo bfmt`
- `cargo bclippy`
- `$env:RUSTFLAGS='--cfg fuzzing'; cargo check -p mssql-tds --lib`
- Rebased full `mssql-tds` library nextest suite with generated TLS fixtures: 1,764 passed
- Stacked production-reader benchmark against #291: all six paired medians improved
- GitHub CodeQL, linked-issue, CLA, and coverage checks pass
- Stacked `main...HEAD` coverage: 98% diff coverage and 91.5% overall coverage
- The external ADO build policy skips dependent-branch PRs; the same buffered-read commits passed its full cross-platform matrix before restacking (build 167366)
- `cargo btest` was attempted locally before stacking but the live SQL integration tests fail with Schannel `SEC_E_WRONG_PRINCIPAL`.

### Risk

The fast path is limited to already-buffered decoder scalars on `NetworkTransport`. Other packet readers retain their existing async behavior through default `None` probes. Misses preserve partial bytes and reuse the existing refill path, including cancellation and encryption behavior. The broader set of inlined probes trades some code size for lower per-value async overhead.

## Related Issues

Related to #247.

Stacked on #291.

Historical measurement context: #269.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes locally (environment-blocked; pre-stack remote matrix passes)

### DIFF
--- a/mssql-tds/src/connection/transport/buffers.rs
+++ b/mssql-tds/src/connection/transport/buffers.rs
@@ -50,46 +50,69 @@ impl TdsReadBuffer {
     }
 
     #[inline(always)]
-    pub(crate) fn try_read_byte(&mut self) -> Option<u8> {
-        if !self.do_we_have_enough_data(1) {
+    fn try_read_array<const N: usize>(&mut self) -> Option<[u8; N]> {
+        if !self.do_we_have_enough_data(N) {
             return None;
         }
 
-        let value = self.working_buffer[self.buffer_position];
-        self.consume_bytes(1);
-        Some(value)
+        let position = self.buffer_position;
+        let bytes = self.working_buffer[position..position + N]
+            .try_into()
+            .expect("slice length is fixed by N");
+        self.consume_bytes(N);
+        Some(bytes)
+    }
+
+    #[inline(always)]
+    pub(crate) fn try_read_byte(&mut self) -> Option<u8> {
+        self.try_read_array().map(|[value]| value)
+    }
+
+    #[inline(always)]
+    pub(crate) fn try_read_int16(&mut self) -> Option<i16> {
+        self.try_read_array().map(i16::from_le_bytes)
     }
 
     #[inline(always)]
     pub(crate) fn try_read_uint16(&mut self) -> Option<u16> {
-        if !self.do_we_have_enough_data(2) {
-            return None;
-        }
+        self.try_read_array().map(u16::from_le_bytes)
+    }
 
-        let position = self.buffer_position;
-        let value = u16::from_le_bytes([
-            self.working_buffer[position],
-            self.working_buffer[position + 1],
-        ]);
-        self.consume_bytes(2);
-        Some(value)
+    #[inline(always)]
+    pub(crate) fn try_read_uint24(&mut self) -> Option<u32> {
+        let [b0, b1, b2] = self.try_read_array()?;
+        Some(u32::from_le_bytes([b0, b1, b2, 0]))
     }
 
     #[inline(always)]
     pub(crate) fn try_read_int32(&mut self) -> Option<i32> {
-        if !self.do_we_have_enough_data(4) {
-            return None;
-        }
+        self.try_read_array().map(i32::from_le_bytes)
+    }
 
-        let position = self.buffer_position;
-        let value = i32::from_le_bytes([
-            self.working_buffer[position],
-            self.working_buffer[position + 1],
-            self.working_buffer[position + 2],
-            self.working_buffer[position + 3],
-        ]);
-        self.consume_bytes(4);
-        Some(value)
+    #[inline(always)]
+    pub(crate) fn try_read_uint32(&mut self) -> Option<u32> {
+        self.try_read_array().map(u32::from_le_bytes)
+    }
+
+    #[inline(always)]
+    pub(crate) fn try_read_uint40(&mut self) -> Option<u64> {
+        let [b0, b1, b2, b3, b4] = self.try_read_array()?;
+        Some(u64::from_le_bytes([b0, b1, b2, b3, b4, 0, 0, 0]))
+    }
+
+    #[inline(always)]
+    pub(crate) fn try_read_int64(&mut self) -> Option<i64> {
+        self.try_read_array().map(i64::from_le_bytes)
+    }
+
+    #[inline(always)]
+    pub(crate) fn try_read_float32(&mut self) -> Option<f32> {
+        self.try_read_array().map(f32::from_le_bytes)
+    }
+
+    #[inline(always)]
+    pub(crate) fn try_read_float64(&mut self) -> Option<f64> {
+        self.try_read_array().map(f64::from_le_bytes)
     }
 
     pub(crate) fn consume_bytes(&mut self, byte_count: usize) {
@@ -470,35 +493,71 @@ mod tests {
     }
 
     #[test]
-    fn test_small_scalar_probes_read_complete_values() {
-        let mut buf = TdsReadBuffer::new(4096);
-        buf.working_buffer[..7].copy_from_slice(&[0xAB, 0x34, 0x12, 0x78, 0x56, 0x34, 0x12]);
-        buf.reset_to_length(7);
+    fn test_fixed_scalar_probes_read_complete_values() {
+        let expected_byte = 0xAB;
+        let expected_int16 = -0x1234i16;
+        let expected_uint16 = 0x1234u16;
+        let expected_uint24 = 0x00A1_B2C3u32;
+        let expected_int32 = -0x0123_4567i32;
+        let expected_uint32 = 0x89AB_CDEFu32;
+        let expected_uint40 = 0xAB_CDEF_0123u64;
+        let expected_int64 = -0x0102_0304_0506_0708i64;
+        let expected_float32 = 1.5f32;
+        let expected_float64 = -2.25f64;
 
-        assert_eq!(buf.try_read_byte(), Some(0xAB));
-        assert_eq!(buf.try_read_uint16(), Some(0x1234));
-        assert_eq!(buf.try_read_int32(), Some(0x1234_5678));
+        let mut bytes = Vec::new();
+        bytes.push(expected_byte);
+        bytes.extend_from_slice(&expected_int16.to_le_bytes());
+        bytes.extend_from_slice(&expected_uint16.to_le_bytes());
+        bytes.extend_from_slice(&expected_uint24.to_le_bytes()[..3]);
+        bytes.extend_from_slice(&expected_int32.to_le_bytes());
+        bytes.extend_from_slice(&expected_uint32.to_le_bytes());
+        bytes.extend_from_slice(&expected_uint40.to_le_bytes()[..5]);
+        bytes.extend_from_slice(&expected_int64.to_le_bytes());
+        bytes.extend_from_slice(&expected_float32.to_le_bytes());
+        bytes.extend_from_slice(&expected_float64.to_le_bytes());
+
+        let mut buf = TdsReadBuffer::new(4096);
+        buf.working_buffer[..bytes.len()].copy_from_slice(&bytes);
+        buf.reset_to_length(bytes.len());
+
+        assert_eq!(buf.try_read_byte(), Some(expected_byte));
+        assert_eq!(buf.try_read_int16(), Some(expected_int16));
+        assert_eq!(buf.try_read_uint16(), Some(expected_uint16));
+        assert_eq!(buf.try_read_uint24(), Some(expected_uint24));
+        assert_eq!(buf.try_read_int32(), Some(expected_int32));
+        assert_eq!(buf.try_read_uint32(), Some(expected_uint32));
+        assert_eq!(buf.try_read_uint40(), Some(expected_uint40));
+        assert_eq!(buf.try_read_int64(), Some(expected_int64));
+        assert_eq!(buf.try_read_float32(), Some(expected_float32));
+        assert_eq!(buf.try_read_float64(), Some(expected_float64));
         assert_eq!(buf.get_remaining_byte_count(), 0);
     }
 
     #[test]
-    fn test_small_scalar_probe_misses_do_not_consume() {
+    fn test_fixed_scalar_probe_misses_do_not_consume() {
         let mut buf = TdsReadBuffer::new(4096);
 
-        assert_eq!(buf.try_read_byte(), None);
-        assert_eq!(buf.buffer_position, 0);
+        macro_rules! assert_miss_does_not_consume {
+            ($partial_len:expr, $method:ident) => {{
+                buf.working_buffer[..$partial_len].fill(0xA5);
+                buf.reset_to_length($partial_len);
+                assert_eq!(buf.$method(), None);
+                assert_eq!(buf.buffer_position, 0);
+                assert_eq!(buf.get_remaining_byte_count(), $partial_len);
+            }};
+        }
 
-        buf.working_buffer[0] = 0x34;
-        buf.reset_to_length(1);
-        assert_eq!(buf.try_read_uint16(), None);
-        assert_eq!(buf.buffer_position, 0);
-        assert_eq!(buf.get_remaining_byte_count(), 1);
-
-        buf.working_buffer[..3].copy_from_slice(&[0x78, 0x56, 0x34]);
-        buf.reset_to_length(3);
-        assert_eq!(buf.try_read_int32(), None);
-        assert_eq!(buf.buffer_position, 0);
-        assert_eq!(buf.get_remaining_byte_count(), 3);
+        assert_miss_does_not_consume!(0, try_read_byte);
+        assert_miss_does_not_consume!(1, try_read_int16);
+        assert_miss_does_not_consume!(1, try_read_uint16);
+        assert_miss_does_not_consume!(2, try_read_uint24);
+        assert_miss_does_not_consume!(3, try_read_int32);
+        assert_miss_does_not_consume!(3, try_read_uint32);
+        assert_miss_does_not_consume!(4, try_read_uint40);
+        assert_miss_does_not_consume!(7, try_read_int64);
+        assert_miss_does_not_consume!(3, try_read_float32);
+        assert_miss_does_not_consume!(7, try_read_float64);
     }
 
     #[test]
--- a/mssql-tds/src/connection/transport/buffers.rs
+++ b/mssql-tds/src/connection/transport/buffers.rs
@@ -49,6 +49,49 @@ impl TdsReadBuffer {
         self.buffer_length - self.buffer_position
     }
 
+    #[inline(always)]
+    pub(crate) fn try_read_byte(&mut self) -> Option<u8> {
+        if !self.do_we_have_enough_data(1) {
+            return None;
+        }
+
+        let value = self.working_buffer[self.buffer_position];
+        self.consume_bytes(1);
+        Some(value)
+    }
+
+    #[inline(always)]
+    pub(crate) fn try_read_uint16(&mut self) -> Option<u16> {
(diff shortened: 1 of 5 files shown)


mssql-tds/src/datatypes/decoder.rs
```diff
@@ -30,6 +30,15 @@ use crate::{query::metadata::ColumnMetadata, token::tokens::SqlCollation};
 
 use super::row_writer::{RowWriter, write_column_value};
 
+macro_rules! read_sync_first {
```
saurabh500, line 35: This can use some comments.

saurabh500: Added a concise comment above `read_sync_first!` documenting the buffered-hit optimization, the zero-consumption miss contract, and the authoritative async refill path.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -751,7 +762,7 @@ impl GenericDecoder {
     where
         T: TdsPacketReader + Send + Sync,
     {
-        let days = reader.read_uint24().await?;
+        let days = read_sync_first!(reader, try_read_uint24, read_uint24);
```
David-Engel, line 817: **Suggestion — the sync arm of `read_sync_first!` is untested at every decoder call site, and this is one of the two sites where a mispairing compiles silently.**

`ByteReader` (decoder tests), `FuzzReader`, `FuzzPacketReader`, and `MockTransport` all inherit the `None` defaults, so the entire 1,673-test lib suite and every decoder fuzz target take only the fallback arm. Production and fuzzing now run different code paths.

The macro is partly type-safe, but only where the two return types differ. `try_read_uint24`/`read_uint32` both return `u32`, and `try_read_uint40`/`read_uint64` both return `u64`, so those two mispairings type-check.

I proved the gap with a mutation rather than assuming it. Changing this line to `read_sync_first!(reader, try_read_uint32, read_uint24)` consumes 4 bytes instead of 3 for every DATE value and desynchronizes the remainder of the row — and the full suite still passed 1,669/1,673 (only the 4 pre-existing cert-fixture failures). After implementing the probes on the test `ByteReader`, that same mutation was caught (`datatypes::decoder::test::decode_into_tests::datetimeoffsetn_value` failed), and reverting the mutation returned the suite to green, which also confirms there is no real mispairing in the PR today.

Suggested fix is the `ByteReader` probe implementation in the review body. Not blocking, since the pairings are all correct as written.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -30,6 +30,17 @@ use crate::{query::metadata::ColumnMetadata, token::tokens::SqlCollation};
 
 use super::row_writer::{RowWriter, write_column_value};
 
+// Avoid constructing a read future when the complete scalar is already buffered.
+// Probe misses consume nothing; the async method remains the authoritative refill path.
+macro_rules! read_sync_first {
```
David-Engel, line 35: **Suggestion — one macro arm per width would make the two silent mispairings impossible**, and would shorten all 99 call sites:

```rust
macro_rules! read_sync_first {
    ($reader:expr, byte)   => { read_sync_first!(@pair $reader, try_read_byte, read_byte) };
    ($reader:expr, uint24) => { read_sync_first!(@pair $reader, try_read_uint24, read_uint24) };
    // ...
    (@pair $reader:expr, $try:ident, $read:ident) => {
        match ($reader).$try() { Some(v) => v, None => ($reader).$read().await? }
    };
}
```

Separately, `$reader:expr` is expanded twice. That is harmless for today's bare `reader` idents, but it would double-evaluate a side-effecting expression if anyone ever passes one.

mssql-tds/src/connection/transport/buffers.rs
```diff
@@ -49,6 +49,72 @@ impl TdsReadBuffer {
         self.buffer_length - self.buffer_position
     }
 
+    #[inline(always)]
+    fn try_read_array<const N: usize>(&mut self) -> Option<[u8; N]> {
```
David-Engel, line 53: **Suggestion — did you A/B plain `#[inline]` against `#[inline(always)]`?**

21 `#[inline(always)]` functions across three layers (here, the `NetworkTransport` delegates, and the 99 expansions) are what produce the ~192 KB / 1.75% growth you measured. Every call site is monomorphized in-crate, so LLVM will treat these as inline candidates regardless; `#[inline]` may buy the same time win at a fraction of the text growth. mssql-odbc ships this as a shared library and `decode_column_into` is already the i-cache hot spot.

Not blocking — you have the timings and the win is real. This is one extra cell in the existing harness.

The bounds reasoning here checks out: `do_we_have_enough_data(N)` plus `buffer_length` excluding `pending_bytes` and headers means a hit can never read into an unparsed packet header.

mssql-tds/src/io/packet_reader.rs
```diff
@@ -49,6 +109,66 @@ pub(crate) trait TdsPacketReader {
 /// Low-level TDS packet reading operations (public under `fuzzing` cfg).
 #[cfg(fuzzing)]
 pub trait TdsPacketReader {
+    /// Returns a buffered byte, or `None` without consuming data if one is unavailable.
+    #[inline]
+    fn try_read_byte(&mut self) -> Option<u8> {
```
David-Engel: **Suggestion — this ~120-line default block is duplicated verbatim between the `cfg(not(fuzzing))` and `cfg(fuzzing)` copies of the trait.**

An *override* that drifts between the two copies fails to compile (`method is not a member of trait`), so that case is safe. A *default* that drifts compiles silently and changes behavior only under `--cfg fuzzing`.

The duplication predates this PR, but this PR roughly doubles the duplicated block. Worth generating both from a single `macro_rules!` parameterized on visibility.

David-Engel (commented):
## Summary

This adds non-consuming synchronous probes (`try_read_*`) to `TdsPacketReader`, implements them on `TdsReadBuffer`/`NetworkTransport`, and routes all 99 production fixed-width scalar reads in the decoder through `read_sync_first!` with the existing async refill path as the fallback. The design is sound, the packet-framing invariants hold, and the benchmark methodology (paired, 8 rounds, medians plus ranges) is credible. **No blocking issues.** The main gap is that nothing in the test or fuzz suite exercises the new fast path at any decoder call site — details inline.

## What I verified locally

Worktree on `2bfac1d4`, diffed against `dev/saurabh/native-resultset-futures`.

- `cargo clippy -p mssql-tds --lib --all-features` — clean.
- `cargo nextest run -p mssql-tds --lib --no-fail-fast` — 1,669/1,673 pass. The 4 failures are `certificate_validator::tests` with `CertificateNotFound { path: "tests/test_certificates/valid_cert.pem" }`, i.e. ungenerated fixtures, unrelated to this PR.
- `RUSTFLAGS='--cfg fuzzing' cargo check -p mssql-tds --lib` — builds.
- All 99 `read_sync_first!` pairings are correct (44 byte, 16 uint16, 9 uint32, 8 int32, 6 int64, 5 int16, 4 float64, 4 float32, 2 uint24, 1 uint40), and no unconverted production scalar reads remain in `decoder.rs`.
- **Bounds and framing**: `try_read_array::<N>` reads `working_buffer[pos..pos + N]` guarded by `do_we_have_enough_data(N)`, and `buffer_length` excludes `pending_bytes` and headers (`remove_header_from_packet` keeps the payload contiguous). A sync hit can never cross a packet boundary into unparsed header bytes, which matches the previous `get_slice()` behavior exactly.
- **Cancellation and scheduling**: awaiting an already-ready future never yielded to the executor, so removing the `.await` on a hit changes no scheduling behavior.

## Findings on unchanged lines

**Suggestion — wire the probes into the decoder test reader.** `ByteReader` at `mssql-tds/src/datatypes/decoder.rs:3324` inherits the `None` defaults, which is the direct cause of the coverage gap described inline on the macro. Implementing `try_read_*` there makes the existing decoder suite cover both arms at zero new test-authoring cost:

```rust
impl ByteReader {
    fn try_take<const N: usize>(&mut self) -> Option<[u8; N]> {
        let bytes: [u8; N] = self.data.get(self.pos..self.pos + N)?.try_into().ok()?;
        self.pos += N;
        Some(bytes)
    }
}

impl TdsPacketReader for ByteReader {
    fn try_read_byte(&mut self) -> Option<u8> { self.try_take().map(|[v]| v) }
    fn try_read_uint24(&mut self) -> Option<u32> {
        let [a, b, c] = self.try_take()?;
        Some(u32::from_le_bytes([a, b, c, 0]))
    }
    // ... remaining eight
```

The same applies to `FuzzReader` in `fuzz_support.rs`, so that `fuzz_token_stream` and `fuzz_tds_client` fuzz the path production actually takes rather than only the fallback.

**Nit** — `read_uint64` at `network_transport.rs:1269` keeps the old `while !do_we_have_enough_data(8)` shape while every sibling moved to `loop { if let Some(..) }`. Intentional per the description; consistency only.

**Follow-up, not this PR** — `token_stream.rs:418` reads the per-token type byte via `read_byte().await` on the same per-row hot path, and looks like the obvious next sync-first candidate.

## Process

- Base is `dev/saurabh/native-resultset-futures`, so merge order is #286 → #291 → this.
- `mssql-rs Pull request validation` (ADO) reports **skipping** for dependent-branch PRs, so the cross-platform matrix has not run on this exact commit. Since the change touches packet-boundary handling, it is worth forcing a run after the restack rather than relying on the pre-stack build 167366.
- `cargo btest` is unchecked in the checklist (environment-blocked), consistent with what I saw locally.

mssql-tds/src/datatypes/decoder.rs
saurabh500: Whitelisted only the ten valid probe/read pairs as macro arms and reborrowed the reader once in the internal arm. Type-compatible mispairings now fail macro matching, and side-effecting reader expressions are evaluated once.

> [!NOTE]

mssql-tds/src/io/packet_reader.rs
saurabh500: Replaced the cfg-specific duplicate trait definitions with one visibility-parameterized macro, so normal and fuzzing defaults now come from the same source.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Implemented the probes on the decoder `ByteReader` and through `FuzzReader`, `FuzzPacketReader`, and `MockTransport`, so decoder tests and fuzz harnesses now exercise the synchronous arm. The targeted decoder suite and all 1,764 library tests pass.

> [!NOTE]

mssql-tds/src/connection/transport/buffers.rs
saurabh500: Ran eight paired alternating rounds: plain `#[inline]` reduced the test binary by 39,424 bytes (0.35%) but was 1.25% slower for fixed-scalar discard and 2.05% slower for mixed discard at the paired median. I kept `#[inline(always)]` because the size saving is small and this PR is throughput-focused.

> [!NOTE]

David-Engel (approved):
## Approved

All four suggestions from my previous review are addressed, and I verified each fix rather than taking it on faith. Nice work on the `#[inline]` A/B — that is exactly the data I was asking for, and the answer (0.35% size saving for 1.25–2.05% slower) justifies keeping `#[inline(always)]`.

### Re-verification on `714064f0`

- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p mssql-tds --lib --all-features` — clean.
- `cargo nextest run -p mssql-tds --lib --no-fail-fast` — 1,672/1,676 pass. The 4 failures are the pre-existing `certificate_validator::tests` fixture failures (`CertificateNotFound { path: "tests/test_certificates/valid_cert.pem" }`), unrelated to this PR.
- `RUSTFLAGS='--cfg fuzzing' cargo check -p mssql-tds --lib` — builds against the consolidated trait.
- All 99 `read_sync_first!` pairings still correct; no unconverted production scalar reads remain in `decoder.rs`.

### The two fixes I asked for, re-tested by mutation

**Macro whitelisting works.** Rewriting the `read_daten` site to `read_sync_first!(reader, try_read_uint32, read_uint24)` now fails to compile with ``error: no rules expected `read_uint24` ``. The two previously type-compatible mispairings (`uint24`/`uint32` and `uint40`/`uint64`) are unrepresentable. The `let reader = &mut *($reader);` reborrow in the `@pair` arm also fixes the double expansion.

**Test-reader probes work.** The stronger mutation is a *valid* pair at the wrong width — `read_sync_first!(reader, try_read_uint32, read_uint32)` for DATE, which consumes 4 bytes instead of 3 and desynchronizes the row. On the previous head that passed the entire suite silently. Now it fails three tests: `decode_into_tests::decode_into_daten`, `decode_into_tests::datetime2n_value`, and `decode_into_tests::datetimeoffsetn_value`. Reverting returns the suite to green, so the fast path is genuinely covered now rather than merely reachable.

Extending the probes through `FuzzReader`, `FuzzPacketReader`, and `MockTransport` was the right call — `fuzz_token_stream` and `fuzz_tds_client` now fuzz the same arm production takes, and I confirmed `MockTransport`'s probes delegate to the same `token_stream_reader.packet_reader` its async reads use. `EmptyReader` correctly keeps the `None` defaults.

The `define_tds_packet_reader!($visibility:vis)` consolidation removes the drift hazard between the two cfg copies cleanly.

### Remaining, none blocking

- ADO `mssql-rs Pull request validation` still reports **skipping** for dependent-branch PRs even after your `/azp run`, so the cross-platform matrix has not gated this exact commit. Approving on the understanding that it runs once this restacks onto `main`.
- Merge order remains #286 → #291 → this.
- Diff coverage 91%, above the 85% bar.

David-Engel (approved):

PR #296, opened 2026-08-14
Stage 2: ODBC pooling primitives (ADO #47338)

saurabh500:
## Summary

Stage 2 of the mssql-odbc connection pooling effort (`[AB#47338](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47338)`). Wires the two ODBC pooling attributes into `mssql-odbc`, building on the Stage 1 TDS reset/liveness primitives.

### B1 — Constants (`api/odbc_types.rs`)
- `SQL_ATTR_CONNECTION_DEAD = 1209`, `SQL_CD_TRUE = 1`, `SQL_CD_FALSE = 0`
- `SQL_ATTR_RESET_CONNECTION = 116`, `SQL_RESET_CONNECTION_YES = 1`

### B2 — `SQLGetConnectAttr(SQL_ATTR_CONNECTION_DEAD)` (`api/get_connect_attr.rs`)
- Returns `SQL_CD_FALSE` only when the DBC is connected **and** `client.is_connection_dead()` is false; otherwise `SQL_CD_TRUE`.
- Disconnected / never-connected ⇒ `SQL_CD_TRUE` (D1: msodbcsql defaults DEAD until a successful token read).
- Cached read, no network probe (D2). Peeks the client under the short DBC mutex — never `take()`s it (D8).

### B3 — `SQLSetConnectAttr(SQL_ATTR_RESET_CONNECTION, YES)` (`api/set_connect_attr.rs` + `api/txn.rs`)
- Routed from `set_connect_attr.rs` into a new `reset_connection` handler in `txn.rs`.
- Value validation (D7): only `SQL_RESET_CONNECTION_YES` accepted; any other value ⇒ **HY024**.
- Claims the idle client via `claim_dbc_client` (busy/`active_stmt` ⇒ connection-busy; disconnected ⇒ **08003**, D7).
- Rolls back a live local transaction first (D4), then arms the full reset via `prepare_reset_connection(false)` (pool checkout does not preserve).
- Clears `state.local_tran_started`. Never holds the DBC mutex across I/O.
- The bit rides the next request; the self-acking round-trip is deferred to Stage 3 (A2/B4).

### B7 (partial) — Tests
- `SQL_ATTR_CONNECTION_DEAD`: `SQL_CD_TRUE` when disconnected / connected-without-client; `SQL_CD_FALSE` when connected-and-alive; null-pointer rejected.
[...]

PR #301, opened 2026-08-15
Stage 3: reset acking + isolation carrier (ADO #47339)

saurabh500:
## Summary

Stage 3 of the mssql-odbc connection pooling effort (ADO User Story #47317, Task `[AB#47339](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47339)`). Makes the connection reset **self-acking at pool checkout** so a borrower never inherits a session whose reset has not yet been processed, and confirms the isolation/autocommit checkout lifecycle carries and survives the reset.

Stacked on **Stage 2 PR #296** (→ #294 → #293). **Base branch: `saurabh500-stage-2-odbc-pooling-primitives`, not `main`.**

## A2 — reset is self-acking (`mssql-tds`)
- New `TdsClient::reset_connection()`: arms the full RESETCONNECTION bit and drives a minimal `SELECT 1` round trip so the server processes the reset and returns its `ResetConnection` ENVCHANGE within the call. The ack flows through Stage 1's `on_reset_connection_ack()`, clearing session-bound caches before checkout. A failed round trip returns an error so the caller can discard the connection.
- `prepare_reset_connection(bool)` is unchanged (the deferred-piggyback path); `reset_connection()` is the eager variant.
- New `TdsClient::mark_connection_dead()` lets the ODBC layer deterministically poison a client whose reset failed.
- Unit test proves the round trip carries the RESETCONNECTION bit and the ack clears session-bound caches.

## B4 — wire eager reset + isolation carrier (`mssql-odbc`)
- `SQLSetConnectAttr(SQL_ATTR_RESET_CONNECTION)` now calls `reset_connection()` after arming, so the reset is processed+acked before the set-attr returns. The DBC mutex is never held across the round-trip I/O (claim → drop lock → I/O → restore). On failure it poisons the client and surfaces **`08S01`** so mssql-python discards it at checkout.
- Confirmed the `SQL_ATTR_TXN_ISOLATION` handler still emits a real `SET TRANSACTION ISOLATION LEVEL` batch; because the reset now acks eagerly, a later short-circuited isolation SET can't leave the reset unacknowledged. Test covers a borrower raising isolation to SERIALIZABLE, resetting, then re-applying READ COMMITTED at checkout.
- Documented the **D9 raw-T-SQL isolation caveat** (stale cache → possible leak; shared, documented limitation, not fixed here) as a code comment at the same-value short-circuit.

## B5 — autocommit / EndTran lifecycle confirmation
- Added a checkout-cycle test: acquire (reset, eagerly acked) → `setautocommit(False)` (begin-txn) → check-in → next acquire (reset) reuses the same physical connection and leaves it reusable. No autocommit/EndTran plumbing rebuilt.

## Tests
- New/updated unit + mock tests in `mssql-tds` and `mssql-odbc`: eager-ack round trip, failure poison → `08S01`, reset+isolation re-apply, and the full checkout lifecycle. Extended `test_client_support` with a `ResetConnection` ENVCHANGE token helper and `mark_known_dead` support on the replay transport.

## Validation
- `cargo bfmt` ✅ (workspace + `mssql-py-core`)
- `cargo bclippy` ✅ (warnings-as-errors, clean)
- `mssql-tds` / `mssql-odbc` lib unit tests ✅ — the only failures are the pre-existing cert/TLS-fixture tests on the base, unrelated to this change.

References `[AB#47339](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47339)`.

### DIFF
--- a/mssql-odbc/src/api/sqlstate.rs
+++ b/mssql-odbc/src/api/sqlstate.rs
@@ -20,6 +20,9 @@ pub(crate) const SQLSTATE_08003: [u8; 5] = *b"08003";
 /// Connection failure during a transaction — ODBC's precise state for a
 /// commit or rollback that could not reach the server.
 pub(crate) const SQLSTATE_08007: [u8; 5] = *b"08007";
+/// Communication link failure — the connection to the server broke, so a
+/// connection pool must discard the connection rather than reuse it.
+pub(crate) const SQLSTATE_08S01: [u8; 5] = *b"08S01";
 pub(crate) const SQLSTATE_22003: [u8; 5] = *b"22003";
 pub(crate) const SQLSTATE_22018: [u8; 5] = *b"22018";
 pub(crate) const SQLSTATE_24000: [u8; 5] = *b"24000";
--- a/mssql-odbc/src/api/txn.rs
+++ b/mssql-odbc/src/api/txn.rs
@@ -24,8 +24,8 @@ use super::odbc_types::{
 use super::sqlstate::{
     ERR_ATTRIBUTE_CANNOT_BE_SET_NOW, ERR_CONNECTION_BUSY, ERR_CONNECTION_DOES_NOT_EXIST,
     ERR_INVALID_ATTRIBUTE_VALUE, ERR_NO_ACTIVE_TDS_CLIENT, ERR_OPTIONAL_FEATURE_NOT_IMPLEMENTED,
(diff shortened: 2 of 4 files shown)


saurabh500:
Consolidated into #308, which contains all of this work in a single PR against `main`. Closing in favor of #308.

PR #308, opened 2026-08-15
ODBC connection pooling (ADO #47317)

saurabh500:
Implements **Connection Pool constructs in `mssql-odbc`** for ADO User Story **#47317**, so `mssql-python`'s client-side connection pool can safely reuse a physical Rust ODBC connection across borrowers.

This is the single, consolidated PR for the feature (the earlier stacked stage PRs #293/#294/#296/#301/#302 have been closed in favor of this one). All four stages' work is included here against `main`.

## What this delivers

- **mssql-tds reset correctness ([AB#47337](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47337)).** All ResetConnection ENVCHANGE sites route through `TdsClient::on_reset_connection_ack()`, which resets the session-state table, clears session-bound caches (prepared handles, param-encryption, pending captures) via `clear_session_bound_caches()`, and restores negotiated DATABASE/LANGUAGE/COLLATION to login values. Fatal error tokens (severity >= 20) mark the transport dead via `mark_known_dead()` so a poisoned connection is discarded by the pool.
- **ODBC pooling primitives ([AB#47338](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47338)).** `SQL_ATTR_CONNECTION_DEAD` (cached, never-probe liveness read) and the pooling reset attribute (value validation → HY024; 08003 when disconnected; busy rejection via claim-idle; rollback of a live local txn; arm the reset bit). The reset is accepted under **both** spellings — the ODBC 3.8 `SQL_ATTR_RESET_CONNECTION` (116) and the msodbcsql vendor attribute `SQL_COPT_SS_RESET_CONNECTION` (1246) — so both the direct-loading `mssql-python` consumer and Driver-Manager-mediated callers can request a check-in reset (see below).
- **Reset armed at check-in, verified at checkout — no extra round trip ([AB#47339](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47339)).** `SQL_ATTR_RESET_CONNECTION` performs **no I/O**: RESETCONNECTION is a TDS packet-header bit (MS-TDS 2.2.3.1.2), so the server returns the session to its login defaults *before* processing whichever request carries it. The two properties a deferred reset would otherwise cost are kept without the round trip:
  - **No stale client-side state** — `prepare_reset_connection` drops the session-bound caches, the session-state table, the negotiated DATABASE/LANGUAGE/COLLATION, and (for a full, non-SKIPTRAN reset) the transaction descriptor at *arm* time.
  - **Fail at checkout** — a DBC-level `pending_reset_ack` suppresses the same-value short circuit in `set_txn_isolation`, so mssql-python's checkout `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` (D9/#343) always emits and carries the armed bit. That handler then verifies `TdsClient::reset_pending()` cleared — i.e. the server actually acknowledged — and on failure poisons the client and surfaces `08S01` so the pool discards the connection before a borrower is served.

  A consumer that never issues the checkout isolation SET degrades to msodbcsql's exact behavior: the bit rides the borrower's first request, still reset-before-execute, only without the early discard.
- **Auth-reuse confirmation, tests, and docs ([AB#47340](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47340)).** Confirmed a reset reuses the same physical login (no re-auth; a rotated token post-connect is rejected), with a unit test. C++ live e2e `connection_pool_test.cpp` covers same-SPID reuse, clean state for the next borrower, per-checkout isolation reset, connection-liveness, and prepared-statement behavior across reset. Shared behavior is verified on both drivers; the one intended mssql-odbc-specific divergence (transparent re-prepare of a dropped handle, which msodbcsql fails with native 8179) is isolated behind `SKIP_IF_COMPARING_MSODBCSQL()`.

## Behavior differences from msodbcsql18

Both drivers arm the reset without a round trip. Two deliberate divergences, both strictly more defensive:

- **Prepared handles are cleared at arm time.** msodbcsql restores its cached DATABASE/LANGUAGE/COLLATION when the reset is armed (D4) but keeps its prepared-handle cache, so reusing a handle across a pool reset fails with native 8179. We clear ours, so the driver transparently re-prepares. Pinned by `PreparedStatementSurvivesResetViaReprepare`.
- **The acknowledgement is verified.** A request completing is not treated as proof the reset happened; the missing `ResetConnection` ENVCHANGE is an error rather than a silent success.

One consequence worth noting: because the checkout isolation SET is now forced to emit, a borrower that changed isolation with **raw T-SQL** no longer leaks it across checkout either. `sp_reset_connection` does not restore the isolation level (verified on SQL 2022), so that path previously depended on the cached value not matching. Pinned by `RawTsqlIsolationDoesNotLeakAcrossCheckout`, which passes on both drivers.

## Design doc

- `docs/odbc-connection-pooling-plan.md` — full design: consumer analysis, discoveries D1–D10, workstreams, DoD, and non-goals. Records the measured cost of the eager-reset round trip that was implemented first and then removed (~1.65 ms vs ~0.80 ms per checkout on loopback, the delta being a full RTT), and why arm-time invalidation supersedes it.
- `docs/odbc-connection-pooling-python-e2e.md` — mssql-python pool integration verification plan (scenarios a–f) via the existing `run_e2e.ps1`/`run_e2e.sh` driver-registration hook.

## Work items

Fixes [AB#47317](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47317). Implements [AB#47337](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47337), [AB#47338](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47338), [AB#47339](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47339), [AB#47340](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47340).

## Why the reset attribute is accepted under two identifiers (D10)

`SQL_ATTR_RESET_CONNECTION` (116) is an ODBC 3.8 attribute reserved for **Driver Manager → driver** communication: the DM sets it itself when returning a connection to its own pool, and per the 3.8 driver guidance *"an application cannot set this attribute directly."* The Windows DM enforces that — a **DM-mediated** application that sets it gets `HY092` ("Option type out of range") from the DM before the call reaches any driver. This is not driver-specific: the installed msodbcsql18 fails identically. Such callers use msodbcsql's vendor attribute `SQL_COPT_SS_RESET_CONNECTION` (`SQL_COPT_SS_BASE_EX+6` = 1246, value `SQL_RESET_YES`, `odbcss.h`) instead, since vendor-range attributes are passed straight through. unixODBC applies no such gate, so 116 reaches the driver on Linux/macOS.

**`mssql-python` — the consumer this feature exists for — is not subject to that gate, because it does not use a Driver Manager.** It loads the driver library directly (`LoadDriverLibrary()` in `mssql_python/pybind/ddbc_bindings.cpp`: `LoadLibraryW` on Windows, `dlopen` elsewhere), binds the exports via `GetProcAddress`/`dlsym` (`SQLSetConnectAttr_ptr` et al.), and calls them itself. Its `SQL_ATTR_RESET_CONNECTION = 116` therefore arrives at our exported `SQLSetConnectAttrW` unchanged on every platform, Windows included — no upstream change is required.

So the driver handles **both** identifiers on every platform: 116 for the direct-loading consumer, and 1246 for DM-mediated callers on Windows — which includes this repo's own C++ e2e suite, since it links against the Driver Manager. `connection_pool_test.cpp` selects the spelling its transport requires and passes against both mssql-odbc and msodbcsql.

An earlier commit in this PR attempted to fix the Windows e2e `HY092` by advertising `DriverODBCVer = "03.80"` in the test registry registration. That was the wrong diagnosis — the DM's rejection is about *who* may set attribute 116, not about the driver's advertised ODBC version — and it did not fix the failure. That change has been reverted; no `DriverODBCVer` change is needed, and this imposes no new requirement on a real install of the driver.

## Notes

Merge is author-owned.


mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -144,6 +146,29 @@ fn sql_get_connect_attr_w_safe(
...
+            // client under the DBC mutex — never `take` it (D8) and never touch
+            // the socket. `SQL_CD_FALSE` means "not known dead", not "proven
+            // healthy"; a connection that failed silently while idle is
+            // recovered on the next operation. Disconnected/never-connected
+            // reads DEAD so a pool discards it.
+            let alive = state.connection_state == ConnectionState::Connected
+                && state
+                    .client
+                    .as_ref()
+                    .is_some_and(|client| !client.is_connection_dead());
```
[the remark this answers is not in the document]
saurabh500: Good catch — this was a real false-positive. `claim_dbc_client`/`claim_connection` `take()` the client for the duration of I/O, so a concurrent liveness read on a healthy in-flight connection reported `SQL_CD_TRUE` and a pool would discard it.

Fixed in `79da2921`: liveness is now derived from observed transport death only. `Connected` with the client momentarily absent means mid-claim, so it reads `SQL_CD_FALSE` ("not known dead") rather than DEAD; disconnected/never-connected still reads `SQL_CD_TRUE`. This matches the D1/D2 semantics where `SQL_CD_FALSE` means "not known dead", never "proven healthy". Renamed and inverted the corresponding test to `connection_dead_reports_false_when_connected_with_client_claimed`.

> [!NOTE]

mssql-odbc/src/api/txn.rs
```diff
@@ -358,6 +358,90 @@ pub(super) fn set_autocommit(dbc: &DbcHandle, value: u64) -> SqlReturn {
...
+    // D4: roll back a live local transaction before the reset so the next
+    // borrower cannot inherit it. Guard on the server actually having one, the
+    // same way `end_transaction` does — the flag can be stale. A full
+    // RESETCONNECTION would discard the transaction server-side regardless, but
+    // rolling back explicitly keeps the client's own transaction tracking in
+    // step with the server.
+    let started = match dbc.inner.lock() {
+        Ok(state) => state.local_tran_started,
+        Err(_) => false,
+    };
```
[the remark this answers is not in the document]
saurabh500: Agreed — swallowing the poison here was inconsistent with every other DBC lock site, and worse, it let the reset proceed to network I/O while potentially skipping a rollback that the (unreadable) `local_tran_started` flag required.

Fixed in `79da2921`: the poison branch now logs and returns `SQL_ERROR` immediately, matching the repo convention of failing deterministically on a poisoned mutex rather than recovering.

> [!NOTE]

mssql-odbc/tests/e2e/tests/connection_pool_test.cpp
```diff
@@ -0,0 +1,306 @@
...
+TEST_F(ConnectionPoolLiveTest, ResetRestoresCleanStateForNextBorrower) {
+    const SQLINTEGER spid_before = Spid();
+    const SQLINTEGER login_db = DatabaseId();
+
+    // Borrower A dirties the session.
+    ASSERT_SQL_OK(SetIsolation(dbc_, SQL_TXN_SERIALIZABLE), SQL_HANDLE_DBC, dbc_);
+    Exec("CREATE TABLE #pool_probe(i int)");
+    Exec("INSERT INTO #pool_probe VALUES (1), (2)");
+    ASSERT_TRUE(TempTableExists("#pool_probe"));
+    Exec("USE master");
```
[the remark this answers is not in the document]
saurabh500: Correct — with `ODBC_TEST_DATABASE=master` the `USE master` was a no-op, so the `login_db == DatabaseId()` assertion passed without the reset ever having to restore anything.

Fixed in `79da2921`: borrower A now switches to a database that differs from the login default (`tempdb` when the login DB is `master`, otherwise `master`), plus an `ASSERT_NE(login_db, DatabaseId())` guard so the setup can never silently fail to dirty the session.

Verified against a live server across all three configs — default, `ODBC_TEST_DATABASE=master`, and `ODBC_TEST_DATABASE=tempdb` — on the Linux suite (20/20) and the Windows msodbcsql leg.

> [!NOTE]

docs/odbc-connection-pooling-python-e2e.md
```diff
@@ -0,0 +1,241 @@
...
+    )
+    return cur.fetchone()[0]
+
+# Pool of size 1 so both acquisitions land on the same physical connection.
+mssql_python.pooling(max_size=1, idle_timeout=30)
+
+c1 = mssql_python.connect(CONN)
+cur1 = c1.cursor()
+cur1.execute("SELECT @@SPID"); spid1 = cur1.fetchone()[0]
+cur1.execute("SET TRANSACTION ISOLATION LEVEL SERIALIZABLE")  # via T-SQL: see (f) caveat
```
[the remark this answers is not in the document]
saurabh500: Right — the snippet demonstrated the attribute-path contract while actually using raw T-SQL, which is exactly the documented non-goal in (f). The driver's cached level stays READ COMMITTED, the checkout SET short-circuits, and the assertion could legitimately fail with SERIALIZABLE, making the reference scenario look like a driver bug.

Fixed in `79da2921`: the snippet now uses `Connection.set_attr()` with `ConstantsDDBC.SQL_ATTR_TXN_ISOLATION` / `SQL_TXN_SERIALIZABLE` (verified those exist in `mssql_python/connection.py` and `constants.py` on `main`), which is the path PR #343 re-applies on checkout. Added a comment pointing at (f) so the T-SQL caveat stays visible without being the thing the scenario exercises.

> [!NOTE]

mssql-odbc/src/api/txn.rs
```diff
@@ -358,6 +358,97 @@ pub(super) fn set_autocommit(dbc: &DbcHandle, value: u64) -> SqlReturn {
...
+        Ok(state) => state.local_tran_started,
+        Err(_) => {
+            // A poisoned mutex leaves the transaction flag unreadable, so a
+            // required rollback could be skipped. Fail deterministically like
+            // every other DBC lock site rather than resetting on a DBC whose
+            // state is already compromised.
+            error!("{OP}: dbc mutex poisoned");
+            return SQL_ERROR;
+        }
+    };
```
David-Engel, line 441: **Suggestion (highest priority of the four).** This is the only `claim_dbc_client` site in the file that takes the DBC lock *between* claim and release. If that lock is poisoned, the function returns `SQL_ERROR` and the owned `TdsClient` is dropped here - so `state.client` stays `None` while `connection_state` is still `Connected`.

That is exactly the state the new liveness read interprets as "busy but healthy" (`get_connect_attr.rs`, the `Connected` arm with `is_some_and`), so `SQL_ATTR_CONNECTION_DEAD` reports `SQL_CD_FALSE` for a connection that can never serve another request, and the pool keeps handing it out. Since panics are caught at the FFI boundary, a poisoned DBC mutex is reachable in practice rather than purely theoretical.

The fix is a reorder that matches what `end_transaction` and `set_txn_isolation` already do - read the DBC state *before* claiming, and never lock again until `release_dbc_client`:

```rust
let started = match dbc.inner.lock() {
    Ok(state) => state.local_tran_started,
    Err(_) => {
        error!("{OP}: dbc mutex poisoned");
        return SQL_ERROR;
    }
};

let mut client = match claim_dbc_client(dbc, OP) {
    Ok(c) => c,
    Err(ret) => return ret,
};
```

mssql-odbc/src/api/txn.rs
```diff
@@ -358,6 +358,97 @@ pub(super) fn set_autocommit(dbc: &DbcHandle, value: u64) -> SqlReturn {
...
+    let result = if started && client.has_active_transaction() {
+        debug!("{OP}: rolling back live local transaction before reset");
+        dbc.runtime
+            .block_on(client.rollback_transaction(None, None))
+    } else {
+        Ok(())
+    };
```
David-Engel, line 475: **Suggestion.** The D4 rollback-before-reset has no coverage on either side. `cargo llvm-cov --lib -p mssql-odbc --show-missing-lines` reports lines 419-421 unhit.

- Unit: `reset_connection_arms_and_clears_local_tran` sets `local_tran_started = true`, but the replay client never saw a BeginTransaction ENVCHANGE, so `has_active_transaction()` is `false` and the branch is skipped.
- e2e: `CheckInAndReset()` calls `SQLEndTran(SQL_HANDLE_DBC, dbc_, SQL_ROLLBACK)` first, which clears `local_tran_started` before `ResetConnection()` runs, so it is skipped there too.

So the headline defense-in-depth behavior of this handler is never exercised. Concrete fix: add an `env_change_begin_transaction(descriptor)` helper alongside the new `env_change_reset_connection()` in `test_client_support.rs`, script it so `has_active_transaction()` is true, and assert the TM rollback is issued before the reset. Alternatively (or additionally) add one e2e case that resets *without* the preceding `SQLEndTran`, which is the scenario this branch exists for.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -785,6 +844,35 @@ impl TdsClient {
...
+    /// authentication and session-recovery context (the `client_context` held in
+    /// [`recovery_context`](Self::recovery_context)) is preserved untouched. A
+    /// new or rotated access token is only ever consumed at connect time, so
+    /// rotating credentials means a fresh connection (new LOGIN7), not a re-auth
+    /// of this live session.
+    pub async fn reset_connection(&mut self) -> TdsResult<()> {
+        self.prepare_reset_connection(false);
+        self.execute("SELECT 1".to_string(), ()).await?;
+        self.close_query().await
+    }
```
David-Engel: **Suggestion.** `reset_connection()` reports success without confirming the acknowledgement actually arrived. DoD #4 and the ODBC-side rustdoc on `reset_connection` both state the reset is "processed and acknowledged" when the call returns, but if the server never emits the `ResetConnection` ENVCHANGE, `on_reset_connection_ack()` never runs, the session-bound caches stay stale, and this still returns `Ok(())`.

Cheap hardening that makes the guarantee real: have `on_reset_connection_ack()` set a `reset_acked` flag, clear it here before the round trip, and return an error if it is still unset after `close_query()`. That also gives the ODBC layer a genuine reason to poison the client and surface `08S01`, rather than inferring it only from a transport-level failure.

David-Engel (commented):
## Summary

Solid, well-researched feature work: centralizes the ResetConnection ENVCHANGE transition in `TdsClient::on_reset_connection_ack()`, adds cached `SQL_ATTR_CONNECTION_DEAD`, and wires the pooling reset attribute under both spellings. The msodbcsql-derived design doc (D1-D10) is unusually good, and the unit + C++ e2e coverage is broad.

Verified locally against the PR head: `cargo bfmt` clean, `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean, `mssql-odbc` / `mssql-tds` lib suites 2,261 pass (the 4 `certificate_validator` failures are the known missing local fixtures, unrelated to this change). All PR checks pass or are pending.

I also probed one concern against a live SQL 2022 instance and **disproved it**: after a full RESETCONNECTION the server does emit the rollback ENVCHANGE, so `has_active_transaction()` correctly returns to `false` and there is no stale transaction descriptor. Not a finding, just recording it so nobody else has to chase it.

**Blocking: none.** Four suggestions inline / below, plus two nits.

### Suggestion 4 - the eager reset costs one extra round trip per checkout; was the cache-invalidation alternative considered?

Measured against a live server on loopback (300 iterations, release build, 3 runs): eager `reset_connection()` + first query **1.63-1.71 ms** versus piggyback `prepare_reset_connection()` + first query **0.79-0.83 ms**. Roughly double, and the delta is a full request/response, so it scales with network RTT rather than staying at the sub-millisecond loopback figure.

The plan justifies the eager round trip by the isolation short-circuit (D6/D9): when the borrower never changed isolation, mssql-python's checkout `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` equals the cached value and emits no I/O, so the reset would ride a later request.

An alternative that keeps the fail-at-checkout property with no extra round trip: invalidate `state.txn_isolation` as part of the reset, so the pool's very next call - `SQLSetConnectAttr(SQL_ATTR_TXN_ISOLATION, READ_COMMITTED)`, which mssql-python issues unconditionally on every checkout - can no longer short-circuit and always emits the batch that carries the armed bit. A failed reset still surfaces inside `Connection::reset()`, still at checkout, still before the connection is handed to a borrower. If the extra RTT is a deliberate accept (simpler, no dependency on the consumer's call order), that trade is worth one line in the plan doc.

### Nits

- The PR body ends with "Draft. Merge is author-owned." but the PR is not in draft state - worth reconciling before review sign-off.
- `mssql-odbc/tests/e2e/run_e2e.ps1`: the only change to this file is the deletion of a single blank line, apparently left over from the reverted `DriverODBCVer` commit. Restoring it keeps the diff scoped to the feature.

mssql-odbc/src/api/txn.rs
saurabh500: Confirmed, and thanks for catching it — this was an interaction bug between the two fixes from the previous review round. The poison-returns-`SQL_ERROR` change and the "claimed means not-known-dead" liveness change are each correct alone, but together they produce exactly the leak you describe: the early return drops the owned `TdsClient`, leaving `state.client == None` with `connection_state == Connected`, which the new liveness read then reports as `SQL_CD_FALSE` forever.

Fixed in `65a16fb1` with your suggested reorder: `local_tran_started` is now read *before* `claim_dbc_client`, and the DBC mutex is not taken again until `release_dbc_client`. That matches `end_transaction` and `set_txn_isolation`, and it removes the drop-on-early-return path entirely rather than trying to compensate for it. Added a comment recording *why* the ordering matters, so it does not get "tidied" back later.

> [!NOTE]

mssql-odbc/src/api/txn.rs
saurabh500: Confirmed — your analysis of *why* it was unreachable was exactly right: the replay client never saw a BeginTransaction ENVCHANGE, so `has_active_transaction()` was `false`, and the e2e path clears `local_tran_started` via `SQLEndTran` before the reset. The headline defense-in-depth branch had no coverage on either side.

Fixed in `65a16fb1`, taking the unit-test route you outlined:
- Added `tds_client_from_tokens_in_transaction(tokens, descriptor)` to `test_client_support.rs`. I went with a constructor rather than a `BeginTransaction` ENVCHANGE helper because the guard is evaluated *before* any round trip, so a scripted token would not have been consumed in time. This required widening `ExecutionContext::set_transaction_descriptor` from `#[cfg(test)]` to `#[cfg(any(test, feature = "test-util"))]`.
- Added `env_change_rollback_transaction()` so the TM rollback leg can be scripted.
- New test `reset_connection_rolls_back_live_transaction_first` asserts the rollback is issued before the reset and that the client ends with no active transaction.

Verified with `cargo llvm-cov --lib -p mssql-odbc --show-missing-lines`: the rollback branch (now lines 424-426) is no longer reported missing. The only remaining unhit lines in this function are the three mutex-poison arms, which need a genuinely poisoned lock to reach.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Agreed — the rustdoc and DoD #4 both claimed "processed and acknowledged", but the code only proved "the batch round-tripped without error". If the server never emitted the ENVCHANGE, `on_reset_connection_ack()` never ran, the session-bound caches stayed stale, and this still returned `Ok(())` — the guarantee was asserted rather than enforced.

Implemented your suggestion in `65a16fb1`: added a `reset_acked` field, set in `on_reset_connection_ack()`, cleared in `reset_connection()` before the round trip, and checked after `close_query()`. A missing ack now returns `ProtocolError`, which flows through the existing ODBC error path so the client is poisoned and `08S01` surfaces — so, as you noted, the driver now has a genuine reason to discard rather than inferring it only from transport-level failure. Documented the new error condition on the rustdoc so the guarantee and the code agree.

Verified against a live server (e2e 21/21, including `connection_pool_test`) that a real reset does emit the ENVCHANGE, so the stricter check does not false-positive.

> [!NOTE]

saurabh500:
Thanks — this was a genuinely useful review, especially the measured numbers and the disproved transaction-descriptor concern (saved chasing it).

All four suggestions and both nits are addressed in `65a16fb1`. The three inline threads have individual replies; covering **Suggestion 4** and the nits here.

### Suggestion 4 — eager round trip vs. cache invalidation

Considered, and we're keeping the eager reset as a **deliberate accept**, now documented in the plan doc (B4) with your measured figures rather than left implicit.

The reasoning: invalidating `state.txn_isolation` does achieve fail-at-checkout with no extra RTT, but it makes our correctness a function of *the consumer's* call sequence — it holds only while mssql-python issues `SQLSetConnectAttr(SQL_ATTR_TXN_ISOLATION, ...)` unconditionally on every checkout. If a pool skips that call, reorders it, or another consumer adopts a different checkout sequence, the reset silently rides a later request again and the guarantee quietly regresses with no failing test. A2 keeps the guarantee a property of our own handler.

Your 1.63-1.71 ms vs 0.79-0.83 ms figures, and the point that the delta scales with RTT rather than staying at loopback, are recorded in the doc alongside the trade-off, with a note to revisit if checkout latency shows up as a measured bottleneck.

### Nits

- **PR body "Draft"** — fixed; the line no longer claims draft state.
- **`run_e2e.ps1` blank line** — restored. You were right about the provenance: it was collateral from the reverted `DriverODBCVer` commit. That file now has an empty diff against `main`.

### Verification on `65a16fb1`

`cargo bfmt` and `cargo bclippy` clean; `mssql-tds` + `mssql-odbc` lib suites 2,376/2,376 pass; ODBC e2e 21/21 against a live server.

David-Engel (dismissed):
## Approved

Re-reviewed `65a16fb1`. All three suggestions are addressed, and both nits are cleaned up. Verified locally on the new head: `cargo bfmt` clean, `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean, `mssql-odbc` / `mssql-tds` lib suites 2,262 pass (the 4 `certificate_validator` failures are the known missing local fixtures, unrelated).

**1. Claim ordering** — fixed correctly. `local_tran_started` is now read before `claim_dbc_client`, and the DBC mutex is untouched between claim and release, so the poisoned-lock path can no longer drop the owned client and strand the DBC as `Connected` with `client == None`. The comment explaining *why* the ordering matters is the right thing to leave behind.

**2. D4 rollback coverage** — fixed, and verified: `cargo llvm-cov --lib -p mssql-odbc --show-missing-lines` now shows the rollback branch (`if started && client.has_active_transaction()` through the `block_on(rollback_transaction(..))`) fully hit. The only remaining uncovered lines in `reset_connection` are the three `dbc.inner.lock()` poisoned arms, which matches every other handler in the file. `tds_client_from_tokens_in_transaction` + `env_change_rollback_transaction` are a clean way to reach a guard the scripted transport otherwise cannot, and gating `set_transaction_descriptor` on `any(test, feature = "test-util")` fits the existing `test-util` wiring rather than widening the API.

**3. Verified reset ack** — good. `reset_acked` cleared before the round trip and checked after `close_query()` makes "processed and acknowledged before checkout" an actual invariant instead of an inference from a successful batch, and `ProtocolError` flows into the ODBC handler's existing poison + `08S01` path so the pool discards.

**4. Eager-reset cost** — the plan doc now records the measurement, the cache-invalidation alternative, and the reason for keeping A2. "Makes the guarantee a property of our handler rather than of the consumer's call order" is the right call and a better argument than the one I offered; no objection to the extra RTT on that basis.

### Optional follow-ups (non-blocking, fine to land as-is)

- The new `reset_acked` guard itself has no test. I confirmed by mutation (replacing the `return Err(..)` with `unreachable!()` and running the `mssql-tds` lib suite — nothing panicked) that no test currently reaches it: `reset_connection_poisons_client_on_failure` fails earlier, at `execute()`, on the dry transport. A ~9-line test covers it and passes:

  ```rust
  #[tokio::test]
  async fn reset_connection_without_ack_is_a_protocol_error() {
      let (mut client, _sent) = create_capturing_client(vec![done_no_more()]);
      let err = client.reset_connection().await.unwrap_err();
      assert!(matches!(err, crate::error::Error::ProtocolError(_)));
  }
  ```

- Minor: if `execute("SELECT 1")` transparently reconnects (session recovery), the reset bit dies with the old transport, no ENVCHANGE arrives, and the new check reports failure for what is actually a brand-new clean session. That errs toward discarding a healthy connection, which is the safe direction, so it is not worth code today — just worth knowing if a "pool churns after network blips" report ever shows up.

### CI note

The `coverage-report` check is red, but the annotation is `Canceling since a higher priority waiting request for pr-coverage-308 exists` — a concurrency cancellation, not a coverage regression, and no replacement run appears to have been queued. Worth re-running it (and letting the ADO validation finish) before merge.

saurabh500:
The roundtripping comments above seem like they need to be re-evaluated. Human involvement needed to compare behavior with odbc native.

saurabh500:
**Correction to my Suggestion 4 reply above** — the reasoning I gave was wrong, and the conclusion was right for the wrong reason. Fixed in `a94b07ae`.

**No, msodbcsql is not buggy, and piggyback does not leak state.** RESETCONNECTION is a TDS *packet-header status bit* (`0x08`, MS-TDS 2.2.3.1.2) carried on the first packet of the next batch/RPC/TM request, and the server resets the session to login defaults **before processing that request**. So even when the ack rides the borrower's first query, that query already runs against a clean session. My phrase "the reset silently rides a later request again and the guarantee quietly regresses" implied a state leak. There is none. Deferring the ack changes *when we learn the reset failed*, not *whether the next borrower sees dirty state*.

What A2 actually buys, restated accurately:

1. **Fail-at-checkout** — a failed reset surfaces inside `Connection::reset()` so the pool discards the connection rather than the borrower's first query failing. This is acceptance criterion #4, so it is a requirement rather than a preference.
2. **Closing the client-side cache window** — this is the part I missed. Our session-bound caches are cleared in `on_reset_connection_ack`, not at arm time, so until the ack lands `execute_prepared` (`tds_client.rs:2178`) can still resolve a *previous borrower's* prepared handle and compose `sp_execute` against a handle the reset is about to drop. That is precisely the msodbcsql behavior our e2e already pins: it keeps its handle cache across the reset and fails native 8179, which `PreparedStatementSurvivesResetViaReprepare` documents as an intended divergence. Worth noting msodbcsql *does* eagerly restore cached DATABASE/LANGUAGE/COLLATION at arm time (D4(d)) — prepared handles are the one thing it does not clear.

**Your alternative is stronger than I gave it credit for.** Invalidating `state.txn_isolation` *plus* clearing the session-bound caches at arm time (mirroring what msodbcsql already does for db/lang/collation) would recover both benefits with no extra RTT. I am still keeping A2, but on the narrower ground that criterion #4's fail-at-checkout would otherwise depend on the consumer issuing the isolation SET unconditionally every checkout — true of mssql-python today, but a property of their call sequence rather than our handler. That is a real but much smaller argument than the one I originally made, and the plan doc now records it that way, including the msodbcsql comparison and why the piggyback is correct.

Thanks for pushing on it — the doc would have carried a wrong explanation for a right decision.

saurabh500:
I will push a commit for no-Round Trip (piggy backing ) design. Please hang on

mssql-tds/src/connection/tds_client.rs
```diff
@@ -777,12 +859,41 @@ impl TdsClient {
...
             true => ResetConnectionMode::ResetSkipTran,
             false => ResetConnectionMode::Reset,
         };
         self.transport.as_writer().set_reset_mode(mode);
+        self.reset_pending = true;
+        // Mirror the reset the server will apply before the next request, so no
+        // borrower is served from state the reset is about to invalidate.
+        self.recovery_context.session_state_table.reset();
+        self.clear_session_bound_caches();
+        self.negotiated_settings.restore_login_defaults();
```
David-Engel, line 912: **Blocking — arming a non-SKIPTRAN reset must also clear the transaction descriptor, or pool checkout fails with native 3903.**

This block mirrors what the server will do — caches, session-state table, database/language/collation — but not the transaction. `has_active_transaction()` is `transaction_descriptor != 0`, and nothing zeroes it here, so the driver keeps believing a transaction is open that the reset is guaranteed to discard.

That belief is read by `set_txn_isolation`'s `reopen` branch, which is the *next* thing mssql-python calls at checkout:

```rust
let reopen = client.has_active_transaction();   // stale true
let mut result = if reopen { ...rollback_transaction(None, None) } else { Ok(()) };
```

so the first request after arming is a TM rollback carrying the RESETCONNECTION bit. The server honors the header bit first, discards the transaction, then processes a rollback that no longer has one.

A/B against a live server, identical probe, only the descriptor handling differing:

```
### 71c7f71f as-is ###
reopen=true
  rollback -> Err("Sql Error: 3903: ... The ROLLBACK TRANSACTION request has no
                   corresponding BEGIN TRANSACTION.")
CHECKOUT RESULT: FAILED -> pool discards connection

### with the descriptor cleared at arm time ###
reopen=false
  isolation SET -> Ok(())
CHECKOUT RESULT: SUCCESS
```

**Reachable in the default configuration, not a corner case:**

1. `autocommit=False` (mssql-python's default) → `switch_to_manual_commit` begins a bookkeeping transaction and sets `local_tran_started = false`.
2. Borrower returns to the pool without executing a statement. Check-in `SQLEndTran` early-returns on `!started`, sends no TM request, so the descriptor survives.
3. Next acquire: `reset_connection` skips its own rollback for the same reason and arms the bit. Descriptor still live.
4. `set_txn_isolation` → `reopen == true` → 3903 → `SQL_ERROR`/HY000 → `Connection::reset()` returns false → connection discarded.

Net effect: pooling silently degrades to a fresh physical connection on every checkout for the default configuration. `reset_pending` came back `false` in the failing run, so the reset itself did happen — this is spurious failure and lost reuse, not state corruption.

**This is new in this commit.** The eager `SELECT 1` used to collect the server's rollback ENVCHANGE and zero the descriptor before the isolation SET ran, so `reopen` was already false by then. Removing the round trip exposed the gap.

**Fix** — conditioned on the mode, since SKIPTRAN must keep the transaction:

```rust
self.transport.as_writer().set_reset_mode(mode);
self.reset_pending = true;
// A full reset discards the server-side transaction; mirror it here, or the next
// request is a TM rollback for a transaction the carrying request itself destroys.
if !preserve_transaction {
    self.execution_context.set_transaction_descriptor(0);
}
self.recovery_context.session_state_table.reset();
```

`set_transaction_descriptor` is currently `#[cfg(any(test, feature = "test-util"))]`, so it needs ungating. Verified: that change alone flips the A/B above to SUCCESS.

Worth an ODBC-layer regression test too — `local_tran_started == false` with a live descriptor, `reset_connection`, then `set_txn_isolation`, asserting no rollback goes out.

David-Engel (commented):
Re-reviewed `71c7f71f`. One blocking issue inline; the rest of the redesign looks right.

**On my earlier ack finding** — you called it correctly in `a94b07ae`. Asking for ack verification *inside* the reset call took the eager round trip as given instead of questioning it, and the round trip was the thing that needed questioning. The arm-time design is the right shape.

**Scoping my own claims** — you flagged that comparing behavior against native ODBC needs human eyes. That applies to me: I have not read msodbcsql source. Every msodbcsql statement I have made in this review (D2's `fConnectionPeek`, D4's arm-time restore of DATABASE/LANGUAGE/COLLATION) is downstream of your plan doc, not independent confirmation. What I verified first-hand is the consumer, the protocol behavior, and this driver's own state.

**Verified for you: fail-at-checkout does work with the real consumer.** The design now routes failure through the isolation SET's return code, which previously did not matter. I pulled `mssql_python/pybind/connection/connection.cpp` from `main` — `Connection::reset()` checks `SQL_SUCCEEDED` on *both* calls and returns `false` on either, so the pool discards. Worth recording, since the mechanism depends on it.

**Resolved:** my earlier "the `reset_acked` guard has no test" nit — `unacknowledged_reset_poisons_client_on_the_carrying_request` covers the equivalent through the real path.

**Validation on `71c7f71f`:** `cargo bfmt` and full-workspace `clippy --all-features --all-targets -D warnings` clean; `mssql-odbc` + `mssql-tds` lib suites 2,264 pass (same 4 known `certificate_validator` fixture failures, unrelated).

Thanks for turning the design around this quickly — the arm-time version is a better driver than the one I approved.

mssql-tds/src/connection/tds_client.rs
saurabh500: **Update — the mechanism described above has changed, the guarantee has not.** Implemented in `71c7f71f`.

The eager `SELECT 1` inside `SQL_ATTR_RESET_CONNECTION` is gone (your Suggestion 4). `TdsClient::reset_connection()` was removed. The ack verification you asked for still exists, but it now runs on the request that *carries* the bit rather than on a dedicated round trip:

- `prepare_reset_connection` drops the session-bound caches at arm time and sets `reset_pending`.
- A DBC-level `pending_reset_ack` suppresses the same-value short circuit in `set_txn_isolation`, so the checkout `SET TRANSACTION ISOLATION LEVEL` always emits and carries the bit.
- That handler then checks `TdsClient::reset_pending()`; if the server never sent the `ResetConnection` ENVCHANGE it poisons the client and surfaces `08S01`, so the pool still discards **at checkout**, before a borrower is served.

So "acknowledged, not assumed" still holds — it just costs no extra RTT (~0.80 ms vs ~1.65 ms).

One thing worth flagging since it changes a documented non-goal: forcing that SET to emit also **closes the raw-T-SQL isolation leak** at checkout. Verified on SQL 2022 that RESETCONNECTION restores `LOCK_TIMEOUT` but *not* `transaction_isolation_level`, so previously a borrower's raw `SET TRANSACTION ISOLATION LEVEL SERIALIZABLE` could survive whenever the checkout re-apply short-circuited on a matching cached value. `RawTsqlIsolationDoesNotLeakAcrossCheckout` now pins this, and it passes on both drivers.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Valid, and reproduced before fixing. Fixed in `7143b520`.

I ran your sequence at the TDS layer against a live server rather than taking it on faith, since it contradicted your earlier "disproved" note:

```
after begin, has_active_transaction = true
after arm,   has_active_transaction = true      ← stale
rollback -> Err(Sql Error: 3903: ... The ROLLBACK TRANSACTION request
                has no corresponding BEGIN TRANSACTION.)
```

With the descriptor cleared at arm time:

```
after arm,   has_active_transaction = false
reopen=false, no rollback sent
```

Both your earlier note and this finding are correct — they were just scoped to different designs. The eager `SELECT 1` collected the server's rollback ENVCHANGE and zeroed the descriptor before `set_txn_isolation` ran, so `reopen` was already false. Removing the round trip removed that incidental clearing, and I mirrored the server's reset for caches, session state and DATABASE/LANGUAGE/COLLATION but not the transaction. That was the gap.

Applied your fix as written, conditioned on the mode, and ungated `set_transaction_descriptor` (now `pub(crate)` with a doc comment covering the non-test caller). Confirmed `reset_with_preserve_keeps_transaction` still passes, so SKIPTRAN keeps its transaction.

Added the ODBC-layer regression test you suggested: `armed_reset_clears_the_transaction_descriptor_so_checkout_sends_no_rollback` sets up a live descriptor with `local_tran_started == false`, arms the reset, then runs the checkout isolation SET and asserts only the SET reaches the wire — a stray rollback would consume the scripted tokens and strand it.

Validation on `7143b520`: `cargo bfmt` and `cargo bclippy` clean; lib suites 2,379 pass; live `test_reset_connection` 8/8; ODBC e2e 21/21.

> [!NOTE]

saurabh500:
Picked up your deferred session-recovery note, because the redesign changed its weight. Fixed in `e6bc8145`.

You raised it against the eager design and called it correctly then: `errs toward discarding a healthy connection, which is the safe direction, so it is not worth code today`. What changed is that the check is now on the **checkout** path rather than inside a dedicated round trip. Confirmed in `reconnect`: the success path replaces `execution_context`, clears the session-bound caches and resets the session-state table, but nothing clears `reset_pending`. So a transparent reconnect during the carrying isolation SET leaves the flag set, and my new verification reports an unacknowledged reset for a session that is in fact brand new and cleaner than a RESETCONNECTION would have left it -- the pool then discards it. Safe direction, but it converts a network blip into exactly the churn this feature exists to avoid.

One line in the reconnect success path, since a fresh login supersedes a pending reset:

`self.reset_pending = false;`

**Coverage caveat, stated plainly:** the reconnect *success* path is not unit-testable without a live connection provider -- the existing `reconnect_*` unit tests only cover failure paths, and `reconnect_increments_recovery_count_tracking` asserts initial state rather than actually reconnecting. So this specific line is exercised only by `session_recovery_test` at e2e. If you would rather see a harness that can drive a successful reconnect in unit tests, that is a reasonable ask and a bigger change than this fix.

Validation on `e6bc8145`: `cargo bfmt` and `cargo bclippy` clean; lib suites 2,384 pass; ODBC e2e 21/21 including `session_recovery_test` and `connection_pool_test`.

Also, on your CI note: the earlier `coverage-report` cancellation is resolved -- the run on `26382c65` completed and the current head has a fresh validation in flight.

mssql-odbc/src/api/txn.rs
```diff
@@ -358,6 +358,119 @@ pub(super) fn set_autocommit(dbc: &DbcHandle, value: u64) -> SqlReturn {
...
+    if result.is_err() {
+        client.mark_connection_dead();
+    }
+    release_dbc_client(dbc, client);
+
+    let Ok(mut state) = dbc.inner.lock() else {
+        error!("{OP}: dbc mutex poisoned");
+        return SQL_ERROR;
+    };
+    state.local_tran_started = false;
```
Theekshna: ### `local_tran_started` is cleared even when the rollback failed

This assignment sits **before** the error branch below:

```rust
state.local_tran_started = false;
if let Err(e) = result {
    // reset was never armed, but the flag is already cleared
    post_tds_error(&mut state, &e, SQLSTATE_08S01);
    return SQL_ERROR;
}
```

If the D4 rollback fails, the reset is not armed and the server may still hold the transaction — but the driver now records that none is open. Please move the assignment below the error branch.

Related question on the same path: msodbcsql posts the error and **continues** to arm the reset (`sqlcmisc.cpp:2401-2410`), since RESETCONNECTION discards the transaction server-side anyway. This PR aborts and poisons the connection instead. That is stricter and defensible for a broken socket, but it discards a connection msodbcsql would recycle. Worth confirming it is deliberate.

mssql-odbc/src/api/get_connect_attr.rs
```diff
@@ -144,6 +146,42 @@ fn sql_get_connect_attr_w_safe(
...
+            // recovered on the next operation. Disconnected/never-connected
+            // reads DEAD so a pool discards it.
+            //
+            // A connected DBC whose client is momentarily absent is mid-claim:
+            // another thread owns it for I/O (`claim_dbc_client` /
+            // `claim_connection`). That is an in-flight healthy connection, not
+            // a dead one, so it reads "not known dead" — reporting DEAD here
+            // would let a concurrent pool liveness check discard a connection
+            // that is merely busy.
+            let known_dead = match state.connection_state {
```
Theekshna, line 169: ### `release_dbc_client` drops the client when the mutex is poisoned, creating a zombie connection

This liveness read treats `Connected` + `client == None` as "not known dead" (`is_some_and` on `None` yields `false` → `SQL_CD_FALSE`), which is the right call for the mid-claim case. But there is a second way to reach that state:

```rust
pub(super) fn release_dbc_client(dbc: &DbcHandle, client: TdsClient) {
    if let Ok(mut state) = dbc.inner.lock() { state.client = Some(client); }
}
```

On the poisoned path the `TdsClient` is dropped while `connection_state` stays `Connected`. A pool then keeps handing out a DBC on which every operation returns `ERR_NO_ACTIVE_TDS_CLIENT`, forever.

This is the counterpart to the T5 hole that was already fixed. Setting `connection_state = Disconnected` (or a `Broken` state) when the client cannot be returned closes it in a few lines.

mssql-odbc/src/api/txn.rs
```diff
@@ -534,6 +660,14 @@ pub(super) fn set_txn_isolation(dbc: &DbcHandle, value: u64) -> SqlReturn {
             .runtime
             .block_on(client.begin_transaction(TransactionIsolationLevel::NoChange, None));
     }
+    // If a pooling reset was armed, this batch carried its RESETCONNECTION bit.
+    // Confirm the server actually acknowledged it: the batch succeeding is not
+    // evidence on its own, and an unacknowledged reset means the session was
+    // never returned to its login defaults.
+    let reset_unacknowledged = result.is_ok() && client.reset_pending();
```
Theekshna, line 708: ### Reset verification should follow the request that carried the bit, not the isolation SET

To be clear, msodbcsql does **no** ENVCHANGE verification at all — it sends its re-sync batch inside the handler, so failures surface straight from `SQLSetConnectAttr`. The verification added here is a real improvement.

The concern is only where it lives. `packet_writer.rs:135-142` arms the bit on `SqlBatch | RpcRequest | TransactionManager`, so several requests can carry it. That is fine today, because the ENVCHANGE clears `reset_pending` and `pending_reset_ack` is durable. But if no isolation SET ever follows the reset, **nothing is verified** — the guarantee quietly becomes dependent on the consumer''s checkout sequence.

Suggestion: move it into `TdsClient`. When `take_reset_mode()` returns a non-`None` mode, record that the bit went out; if that message''s response completes without `on_reset_connection_ack()` firing, mark the connection dead there. `DbcState::pending_reset_ack` and the `set_txn_isolation` short-circuit suppression then both go away — as does a small window in `reset_connection`, which releases the client at line 455 and only sets `pending_reset_ack` after re-locking at 457.

mssql-odbc/src/api/txn.rs
```diff
@@ -358,6 +358,119 @@ pub(super) fn set_autocommit(dbc: &DbcHandle, value: u64) -> SqlReturn {
...
+            error!("{OP}: dbc mutex poisoned");
+            return SQL_ERROR;
+        }
+    };
+
+    // Claim the idle client: this rejects a busy connection (open cursor /
+    // `active_stmt`) with ERR_CONNECTION_BUSY and a disconnected DBC with
+    // ERR_CONNECTION_DOES_NOT_EXIST (08003, D7). From here until
+    // `release_dbc_client` the DBC mutex is never taken again.
+    let mut client = match claim_dbc_client(dbc, OP) {
```
Theekshna, line 460: ### `reset_connection` rejects a busy connection instead of sweeping cursors

`claim_dbc_client`''s doc states *"the caller must have closed every cursor first (see `close_all_cursors`)"*, and the five sibling DBC operations all do — `end_tran` (193), `switch_to_autocommit` (500), `switch_to_manual_commit` (544), `set_txn_isolation` (632), (711). This call site does not, so an open cursor turns the reset into HY000.

I can see this is deliberate (the comment just above), so treating it as a disagreement rather than a bug. To be fair, it is not an msodbcsql-parity argument either: msodbcsql does not close user cursors, it just runs the reset on its own `lpstmtDvr` and does not care.

The argument is the pooling consequence — check-in with an undrained result set is the ordinary case (app closed the connection without closing the cursor), and HY000 forces the pool to discard a recyclable connection. With a single `TdsClient` and no MARS, sweeping is the only way to reach msodbcsql''s outcome.

If the rejection stays, please document in the pooling doc that an open cursor makes a connection non-poolable.

mssql-odbc/src/api/txn.rs
```diff
@@ -358,6 +358,119 @@ pub(super) fn set_autocommit(dbc: &DbcHandle, value: u64) -> SqlReturn {
...
+///   prepared handle the reset is about to invalidate.
+/// - **Fail at checkout.** `pending_reset_ack` suppresses the same-value
+///   short circuit in `set_txn_isolation`, so the pool's checkout
+///   `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` (D9/#343) always emits and
+///   carries the armed bit. A failed reset therefore surfaces inside the pool's
+///   `reset()` — before the connection is handed out — and that handler verifies
+///   the server acknowledged it rather than assuming the batch succeeding
+///   implies it.
+///
+/// A consumer that never issues the checkout isolation SET simply gets
```
Theekshna: ### This claim about msodbcsql is inaccurate, and there is no re-apply hook

> A consumer that never issues the checkout isolation SET simply gets msodbcsql''s behavior: the bit rides the borrower''s first request.

msodbcsql does not do that. Its handler arms the bit and then immediately sends a re-sync batch on its driver statement:

```
sqlcmisc.cpp:2410       hConn->ResetConnection(TRUE)      // arm
sqlcmisc.cpp:2411-2412  cchTextsize = 0; crow = 0         // reset client caches
sqlcmisc.cpp:2435       BuildServerSideConnectOptions()
sqlcmisc.cpp:2446       ExecImmediate(lpstmtDvr, ...)     // the batch carrying the bit
```

`BuildServerSideConnectOptions` emits `SET TRANSACTION ISOLATION LEVEL` for any non-READ-COMMITTED cached level (`sqlcfunc.cpp:2007-2040`), plus ANSI_NPW / CONCAT_NULL (2043) and QUOTED_IDENTIFIER (2075) when non-default. Could you correct the description?

**Impact today is small.** The only session setting this driver emits post-login is `SET TRANSACTION ISOLATION LEVEL`, and the durable `pending_reset_ack` suppression makes any later explicit set emit. What is left is a reporting gap: between the reset and the next isolation set, the session is READ COMMITTED while `SQLGetConnectAttr(SQL_ATTR_TXN_ISOLATION)` still returns the previous borrower''s level.

Assigning `state.txn_isolation = SQL_TXN_READ_COMMITTED` at arm time does not remove that disagreement so much as invert it — between arming and the bit actually riding a request, the driver would report READ COMMITTED while the server is still SERIALIZABLE. That is the better direction to be wrong in, since nobody reads isolation off a checked-in connection, but it is a trade rather than a fix. Note also that it stays consistent only because `pending_reset_ack` currently forces the checkout SET to emit; if that suppression is removed as part of the verification comment above, please re-check this together with that change rather than applying the two independently.

**The structural point is the durable one.** `apply_post_connect_txn_settings` (`txn.rs:762+`) is already the analogue of `BuildServerSideConnectOptions`, and `reset_connection` never calls it. As `QuotedId`, `AnsiNPW`, `CONCAT_NULL`, `SQL_ATTR_MAX_LENGTH`/`MAX_ROWS` are added, each will silently desync across a reset unless someone remembers this path. Routing the reset through that function — and letting its batch carry the bit — makes that impossible to forget.

mssql-odbc/src/api/txn.rs
saurabh500: Valid — fixed in `9defa18b`. The assignment now sits below the error branch, so a failed rollback leaves `local_tran_started` intact. Clearing it there would have been a lie the next operation acts on: `SQLDisconnect` would skip its 25000 guard on a connection whose transaction the server may still hold. Added `failed_rollback_leaves_local_tran_started_set` to pin it.

On the related question — the divergence from msodbcsql is deliberate, and I'd rather keep it. msodbcsql continuing to arm is safe *for msodbcsql* because it then sends its own re-sync batch immediately, so the reset is delivered within the handler. This driver arms and returns without I/O, so the bit only rides a later request. If the rollback just failed, the most likely cause is a broken socket or an unusable session — precisely the case where "the reset will clean it up later" may never happen, and the pool would hand out a connection whose state we never confirmed. Failing here converts an uncertain session into a definite discard at check-in.

The cost is real and I want to be honest about it: a connection msodbcsql would recycle, this driver discards. I think that is the right trade for a pool — a spurious discard costs one reconnect, a wrongly-recycled session corrupts the next borrower — but if you'd rather match msodbcsql's recycle behavior, that is a one-line change to arm unconditionally and I'll take it.

> [!NOTE]

mssql-odbc/src/api/get_connect_attr.rs
saurabh500: Correct, and thanks for spotting that it is the same shape as the claim-ordering hole rather than a separate issue. `release_dbc_client` silently dropped the client on the poisoned path, leaving `Connected` + `client == None` — which the liveness read deliberately treats as "claimed, not known dead", so the pool would hand out a DBC on which every operation returns `ERR_NO_ACTIVE_TDS_CLIENT` indefinitely.

Fixed in `9defa18b`: the poisoned branch now sets `connection_state = Disconnected`, so the liveness read reports `SQL_CD_TRUE` and the pool discards. `PoisonError::get_mut()` gives access to the guarded state even though the mutex is poisoned, so this needs no new state variant.

I kept `Disconnected` rather than adding a `Broken` state — it is what the existing liveness rule already maps to "known dead", and `SQLDisconnect`/`SQLFreeHandle` already handle a disconnected DBC. A distinct `Broken` state would be more precise for diagnostics, but it would touch every `connection_state` match in the crate for no behavioral gain here.

> [!NOTE]

mssql-odbc/src/api/txn.rs
saurabh500: You convinced me — changed in `9defa18b`. `reset_connection` now calls `close_all_cursors` before claiming, like the five siblings.

What decided it was your first point rather than the pooling one: `claim_dbc_client`'s own doc states the caller must have closed every cursor first, and this was the single call site violating that contract. The pooling consequence then follows — an application closing its connection without closing a cursor is the ordinary check-in case, and turning that into HY000 forces the pool to discard a recyclable connection. I added `reset_connection_sweeps_an_open_cursor_instead_of_rejecting` to pin it, and `reset_connection_rejects_busy_connection` still passes, so a connection busy for a reason the sweep cannot clear is still rejected.

I also documented it in the pooling doc as you asked, but describing the sweep rather than the restriction — including your point that msodbcsql reaches the same outcome by running its reset on an internal driver statement, so this is not an msodbcsql-parity argument.

One note on my earlier reasoning, since I had defended the rejection: I was treating "busy" as a signal the borrower was mid-operation, but at check-in it far more often means the borrower simply never closed a cursor. Your framing of check-in as the ordinary case is the right one.

> [!NOTE]

mssql-odbc/src/api/txn.rs
saurabh500: The factual correction is right and I've fixed it in `9defa18b` — thank you for the source lines.

My claim that a consumer without the checkout SET "gets msodbcsql's behavior" was wrong in both directions: msodbcsql is *eager* (arms, then immediately sends its re-sync batch on `lpstmtDvr`), and it *re-applies* session settings via `BuildServerSideConnectOptions`. The rustdoc on `reset_connection` now states plainly that the piggyback design is a deliberate divergence rather than parity, and the plan doc's parity section carries the same correction with your `sqlcmisc.cpp:2410-2446` / `sqlcfunc.cpp:2007+` references. Worth flagging that this is the second msodbcsql claim of mine in this PR that did not survive checking, so treating my parity assertions as needing verification is warranted.

**On the reporting gap:** agreed it is a trade rather than a fix, and I have *not* applied `state.txn_isolation = READ_COMMITTED` at arm time. Your reasoning for waiting is the deciding factor — its consistency depends on `pending_reset_ack` forcing the checkout SET to emit, and if the verification move removes that suppression, applying the two independently is how you get a subtle inconsistency. Better to make that call once, with the verification design settled.

**On the structural point — I agree it is the durable one, and I have recorded it as a follow-up rather than doing it here.** `apply_post_connect_txn_settings` is indeed the analogue of `BuildServerSideConnectOptions`, and having the reset not call it is a real footgun as settings accumulate. The reason I did not route the reset through it now: its batch is I/O, so letting it carry the bit reintroduces exactly the round trip this design was just changed to remove. Doing it properly means either deferring the batch to checkout or accepting the RTT — a design decision that belongs with the verification-placement question in the neighbouring thread, not bolted on here. The plan doc now names the specific settings that will desync (`QuotedId`, `AnsiNPW`, `CONCAT_NULL`, `MAX_LENGTH`/`MAX_ROWS`) so the next person adding one finds it.

> [!NOTE]

mssql-odbc/src/api/txn.rs
saurabh500: The architectural point is right, and it is the strongest of the five: tying verification to the isolation SET makes the guarantee depend on the consumer's checkout sequence. That is the same objection I made when arguing *for* an eager round trip earlier in this PR, so I am not going to pretend it away — moving verification into `TdsClient` would make it unconditional *and* keep the no-RTT property, which is strictly better than what is here.

**I have not made the change in this PR, deliberately.** It touches the request/response path in `TdsClient` — recording that the bit went out in `take_reset_mode()` and failing the message whose response arrives without `on_reset_connection_ack()`. That is the core loop every statement, RPC and TM request flows through, in a PR that is already large, has been through three review rounds, and is currently green. The failure mode of getting it subtly wrong is connections being marked dead on healthy sessions, which is worse than the gap it closes. It also cannot be assessed independently of the isolation-cache question in your other thread, since removing `pending_reset_ack` changes what the checkout SET does.

What is true today, and why I think deferring is defensible rather than convenient: `mssql-python` — the only consumer this feature exists for — issues `SQLSetConnectAttr(SQL_ATTR_TXN_ISOLATION, READ_COMMITTED)` unconditionally on every checkout, and `pending_reset_ack` is durable, so the SET always emits and always verifies. The gap is a *future* consumer with a different checkout sequence, not a live defect.

You are also right about the window between `release_dbc_client` and the re-lock. It is narrow and, in practice, DBC-level operations on a pooled connection are serialized by the caller, but it is real and it disappears entirely under your design rather than needing separate handling — another reason to do this as one coherent change rather than patching the symptom.

Recorded as a follow-up with your suggested mechanism. Resolving so it is not left dangling, but I'd rather it be tracked as a work item than quietly forgotten — happy to file one, or to take it in a focused follow-up PR where the message-loop change can get review attention on its own.

> [!NOTE]

Theekshna: Lets file a work item under this feature to be addressed later since this is not a blocker for mssql-python - https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47102

Theekshna (approved):

mssql-odbc/src/api/txn.rs
saurabh500: Agreed it is not a blocker for mssql-python, and thanks for the approval.

On filing it under 47102: I will file manually. 

Here is the content, ready to paste:

---

**Title:** Move pool-reset acknowledgement verification into `TdsClient`

**Parent:** Feature 47102

**Context.** PR #308 verifies the RESETCONNECTION acknowledgement in `set_txn_isolation` (`mssql-odbc/src/api/txn.rs`), because the pool's checkout `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` is the request that carries the armed bit. The `DbcState::pending_reset_ack` flag suppresses that handler's same-value short circuit so the SET always emits.

**Problem.** `packet_writer.rs:135-142` arms the bit on `SqlBatch | RpcRequest | TransactionManager`, so several request types can carry it. Verification only happens if an isolation SET follows the reset, which makes the fail-at-checkout guarantee depend on the consumer's checkout sequence rather than on the driver. This is correct for mssql-python today (it issues that SET unconditionally on every checkout) but not for a future consumer with a different sequence. There is also a narrow window in `reset_connection`, which releases the client before re-locking to set `pending_reset_ack`.

**Proposed fix (from review).** When `take_reset_mode()` returns a non-`None` mode, record in `TdsClient` that the bit went out; if that message's response completes without `on_reset_connection_ack()` firing, mark the connection dead there. `DbcState::pending_reset_ack`, the `set_txn_isolation` short-circuit suppression, and the release/re-lock window all go away.

**Assess together with:** whether to set `state.txn_isolation = READ COMMITTED` at reset arm time. Its consistency currently depends on `pending_reset_ack` forcing the checkout SET to emit, so the two must be decided as one change, not applied independently.

**Not done in #308 because:** it touches the request/response path every statement, RPC and TM request flows through, in an already-large PR; getting it subtly wrong marks healthy sessions dead, which is worse than the gap it closes.

---

The reasoning is also recorded in `docs/odbc-connection-pooling-plan.md` so it is discoverable from the code, and the related settings-desync follow-up (`apply_post_connect_txn_settings` not participating in the reset) is captured in the same section.

> [!NOTE]

saurabh500:
Filed as **AB#47400** — *Move pool-reset acknowledgement verification into `TdsClient`* — as a child Task of Feature 47102, in the `mssql-rs` area and the `Rubidium\Aug 2026` iteration:

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47400

It carries the full context: the `packet_writer.rs:135-142` arming surface, why tying verification to the isolation SET makes the guarantee depend on the consumer's checkout sequence, the `take_reset_mode()` fix you proposed, the `release_dbc_client` / re-lock window it also closes, and the note that it must be decided together with the `state.txn_isolation` arm-time question rather than independently.

Supersedes my earlier reply on this thread saying I could not create it — I had no Azure DevOps credentials at the time; @saurabh500 authenticated, so it is filed for real now.

Theekshna (approved):

shiwanigupta0809 (approved):

shiwanigupta0809 (approved):

Issue #253, opened 2026-08-13
Add `RowWriter` sink API so consumers receive string/binary payloads without `mssql-tds` allocating (row decode perf, axis: value handoff)


saurabh500:
Give `RowWriter` implementations a way to receive string and binary payloads **without `mssql-tds` allocating for them**, so an external consumer can act as a direct sink for wire bytes and place them straight into its own final representation.

The driving consumer is a **PostgreSQL FDW** that queries SQL Server and materialises results as PG datums. It needs the encoded bytes plus the encoding descriptor; where the encoding permits, it stores them verbatim and lets PostgreSQL own transcoding.

## Non-goals

- **`DefaultRowWriter` is not required to benefit.** It may keep its existing owned `write_string`/`write_bytes` path unchanged. Unifying the two was an aspiration, not a requirement.
- Not a rework of `ColumnValues`, `take_row()`, or row-level buffer reuse.
- Not the dispatch/boxing axis — that landed in #264.

## API

Six methods added to `RowWriter`, **every one with a default implementation**, so existing writers compile and behave identically without edits.

```rust
/// Writes packet-backed encoded string bytes before the packet buffer is reused.
fn write_string_ref(&mut self, col: usize, bytes: &[u8], encoding_type: &EncodingType) {
    self.write_string(col, SqlString::new(bytes.to_vec(), encoding_type.clone()));
}

/// Writes packet-backed binary bytes before the packet buffer is reused.
fn write_bytes_ref(&mut self, col: usize, bytes: &[u8]) {
    self.write_bytes(col, bytes.to_vec());
}

/// Returns a destination for a known-length binary value, when the consumer
/// can allocate its final representation before the decoder reads the bytes.
fn bytes_destination(&mut self, _col: usize, _length: usize) -> Option<&mut [u8]> {
    None
}

/// Completes or abandons a destination returned by `bytes_destination`.
fn finish_bytes_destination(&mut self, _col: usize, _success: bool) {}

/// Returns a destination for a known-length character value, when the consumer
/// can allocate its final representation before the decoder reads the bytes.
///
/// `length` counts encoded bytes on the wire, not characters, so a consumer
/// whose storage differs in length under `encoding_type` returns `None`.
fn string_destination(
    &mut self,
    _col: usize,
    _length: usize,
    _encoding_type: &EncodingType,
) -> Option<&mut [u8]> {
    None
}

/// Completes or abandons a destination returned by `string_destination`.
[...]

Issue #298, opened 2026-08-14
Shrink the per-row `CancelHandle::run_until_cancelled` future by collapsing it to one await point


saurabh500:
Sub-issue of #247. Raised during review of #275, with prior spike measurements attached so this does not get re-litigated from first principles.

### Problem statement

`CancelHandle::run_until_cancelled` (`mssql-tds/src/core.rs:43-57`) keeps `f` live across two distinct suspend points:

```rust
match cancel_handle {
    Some(handle) => match handle.cancel_token.run_until_cancelled(f).await {
        Some(result) => result,
        None => Err(OperationCancelledError("Request was cancelled".to_string())),
    },
    None => f.await,
}
```

Because `f` can be awaited in either arm, the generated state machine reserves room for both, costing roughly one extra copy of the inner future. Measured decomposition at the `receive_row_into` call site with `W = dyn RowWriter + Send`:

| future | size | vs. inner |
|---|---|---|
| `receive_row_into_internal` | 1,408 B | 1.00x |
| tokio-util `CancellationToken::run_until_cancelled(inner)` | 2,904 B | 2.06x |
| `CancelHandle::run_until_cancelled(None, inner)` | 4,328 B | 3.07x |

The outer wrapper adds 1,424 B on top of tokio-util's. This is on the per-row path, and `#[async_trait]` expands to `Box::pin(async move { .. })` per call, so the size sets the per-row allocation size.

### Proposed solution

Collapse to a single await point, for example by resolving to one `CancellationToken` up front and using a never-cancelled token when `cancel_handle` is `None`, at the cost of an `Arc` bump on the no-handle path. An alternative that avoids even that: pin `f` once before the match and pass `f.as_mut()` into `CancellationToken::run_until_cancelled`, keeping the direct `None => f.await` arm.

### Affected crate

mssql-tds

### Alternatives considered

**Do it as a standalone performance change.** Already prototyped and measured; currently a no-go on its own. The pin-once variant on post-#264 `main` (`7a9c0b93`):

- Size: **4,280 B -> 2,912 B (-1,368 B, -32.0%)**, landing essentially on the predicted ~2,904 B.
- Speed: no stable win. Seven fresh processes, 31 paired ABBA pairs each, 20,000 buffered rows per pass, 48 nullable `IntN` columns.
  - No handle: median **+12.2 ns/row (0.82% slower)**, 5/7 processes slower.
[...]

PR #311, opened 2026-08-16
Add RowWriter value-destination sink for known-length PLP values

saurabh500:
## Description

Adds an opt-in destination-buffer API to `RowWriter` for known-length PLP string and binary values. Consumers that already own the final storage can have TDS payload bytes decoded directly into it instead of receiving a temporary `Vec` and copying the bytes afterward.

### What changed

- Added `RowWriter::value_destination(col, kind, length)`, which can provide final storage as `&mut [MaybeUninit<u8>]` before the payload is read.
- Added `RowWriter::commit_value(col, complete)` to report whether every destination byte was initialized successfully.
- Added non-exhaustive `ValueKind` metadata for binary payloads and string wire encodings.
- Routed known-length PLP `varchar(max)`, `nvarchar(max)`, and `varbinary(max)` values through the destination when the writer opts in.
- Added an internal packet-reader operation that writes directly into uninitialized storage; the production network transport avoids a preliminary zero-fill.
- Kept the existing owned `write_string` / `write_bytes` path as the default, so existing writers require no changes.
- Consolidated PLP framing, chunk validation, and chunk reads across the owned and destination paths.
- Added coverage for accepted and declined destinations, exact string encodings, excluded PLP types, wrong-sized destinations, NULL and empty values, malformed/truncated payloads, and owned-path equivalence.

### Intentional scope

The destination is offered only for `MAX` values with a known PLP total length. These remain on the owned path:

- `varchar(n)`, `nvarchar(n)`, and `varbinary(n)` values using `USHORTLEN`
- legacy `TEXT`, `NTEXT`, and `IMAGE` values
- `PLP_UNKNOWNLEN` streams
- XML, JSON, UDT, and vector values
- SQL NULLs

## Performance

For each accepted value, the decoder avoids a temporary allocation, payload copy, and deallocation. It writes wire bytes directly into consumer-owned storage, including storage that begins uninitialized. String consumers also receive the wire encoding and can defer transcoding to their normal downstream path.

Writers that do not opt in retain the existing owned path. Short non-PLP values are unchanged. The destination branch increases row-decode future state by 192 bytes while remaining within the existing 4096-byte regression budget.

## Related Issues

Refs #253

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `mssql-tds` library tests pass


mssql-tds/src/datatypes/row_writer.rs
```diff
@@ -67,6 +67,71 @@ pub trait RowWriter {
...
+    /// consumer that transcodes downstream never pays for a transcode here.
+    /// A writer whose storage cannot hold the wire form — because it needs to
+    /// transcode from, say, [`EncodingType::Utf16`] first — returns `None` for
+    /// that value and receives it through [`Self::write_string`] as before.
+    fn value_destination<'a>(
+        &'a mut self,
+        _col: usize,
+        _kind: ValueKind<'_>,
+        _length: usize,
+    ) -> Option<&'a mut [u8]> {
```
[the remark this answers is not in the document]
saurabh500: Fixed in `1f3dbb00` — the docs were wrong, though the code was not.

You're right that a `&mut [u8]` over uninitialized `palloc`ed memory is UB, and the prose invited exactly that by naming varlena allocation as the motivating case without stating the requirement. The contract now says it explicitly:

> The slice must be a valid `&mut [u8]`, which means the `length` bytes are already initialized. Handing back a slice built over uninitialized allocation — `slice::from_raw_parts_mut` on fresh `palloc`ed memory, an unfilled `Vec`'s spare capacity — is undefined behaviour regardless of the decoder overwriting every byte afterwards. Use `palloc0`, `vec![0; length]`, or `Vec::resize` to establish the bytes first.

I also corrected the motivation paragraph, which was overclaiming: the sink saves the *redundant* allocation and copy, not the initialization. A consumer allocating fresh storage still pays to establish the bytes once (`palloc0` rather than `palloc`); what it stops paying for is the second buffer and the copy between them.

On `&mut [MaybeUninit<u8>]`: it is the right type in the abstract, but it does not stop at this signature. The destination is filled by `read_bytes`, which takes `&mut [u8]`, so adopting it means threading `MaybeUninit` through the packet reader and either duplicating those paths or landing on the unstable `read_buf`/`BorrowedBuf` API. That is a large change to the hottest code in the crate for a saving that, at these value sizes, is a `memset` of a buffer the consumer generally has to zero anyway. I'd rather keep the reader on initialized slices and make the contract carry the requirement.

Worth recording for anyone auditing the numbers in this PR: neither measured harness ever used the unsound pattern. `sink_bench::ArenaWriter` hands out subslices of a pre-allocated, already-initialized arena, and the `pg_fabric` PoC's `begin_destination` does `vec![0; length]`. So no benchmark result depends on the pattern the docs described, and nothing changes numerically.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -41,6 +41,59 @@ macro_rules! read_sync_first {
...
+        |$bytes:ident| $owned:expr $(,)?
+    ) => {{
+        let col = $col;
+        let length = $length;
+        match $writer.value_destination(col, $kind, length) {
+            // The writer borrow held by `$dest` ends with this read.
+            Some($dest) => {
+                let outcome = $fill;
+                $writer.commit_value(col, outcome.is_ok());
+                outcome?;
```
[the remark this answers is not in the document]
saurabh500: Real gap, documented in `1f3dbb00` rather than fixed structurally — the guard you're describing is not expressible here without `unsafe`.

First, the smaller-than-it-looks part. Errors that return *through* the transport do reach `commit_value(col, false)`: the macro drives the fill to completion and commits on `outcome.is_ok()`, so timeouts, explicit cancels and malformed tokens are all covered. The hole is specifically an outer future being dropped mid-`await`, where no further decoder code runs at all.

On the RAII guard: `value_destination` returns `Option<&'a mut [u8]>` borrowed from `&'a mut self`. A guard that calls `commit_value` on drop has to hold `&mut writer` for its own lifetime, and the destination slice it is guarding already holds the only `&mut` borrow of that writer for the same span. Two simultaneous mutable borrows of the same writer — not something the borrow checker will accept, and getting there means raw pointers plus an aliasing argument. That is a strictly worse soundness trade than the one this thread is trying to close.

So the contract carries it instead:

> A writer must therefore treat a destination that is still pending at the next `end_row()`, or at the next `value_destination` for the same column, as abandoned rather than asserting that it was committed. The row it belonged to is not delivered in that case.

`end_row()` is the natural resynchronization point — it already exists on the trait and a writer is already tracking per-row state there. And the failure mode is benign for the shape this API exists for: the row is never delivered, so a pending destination is dead storage the writer reclaims or overwrites, not a torn value that reaches the consumer. What I wanted to rule out was a writer *asserting* commit-on-offer, which the docs previously left open.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -41,6 +41,59 @@ macro_rules! read_sync_first {
...
+// small-value PLP scan, on every writer including ones that never opt in).
+macro_rules! read_value_into {
+    (
+        $writer:expr, $col:expr, $kind:expr, $length:expr,
+        |$dest:ident| $fill:expr,
+        |$bytes:ident| $owned:expr $(,)?
+    ) => {{
+        let col = $col;
+        let length = $length;
+        match $writer.value_destination(col, $kind, length) {
```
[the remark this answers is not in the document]
saurabh500: You've caught the most important comment on this PR, and acting on it changed the headline number. Short version: **the +5.1% I published was substantially a measurement artifact**, the honest figure is ~1.5–3%, and I am amending #253's criterion rather than claiming it's met.

**What went wrong with the original number.** It came from comparing two separately-built release binaries. I went back and ran a controlled layout experiment: build the *same* source twice, differing only by an inert `RUSTFLAGS='--cfg probe_a'` / `--cfg probe_b` that changes the crate metadata hash and therefore link layout, but nothing semantic. Those two byte-identical-source binaries measured **99.43 ms vs 104.36 ms on the same workload — a 5% spread from code layout alone.** The regression I attributed to the hook was inside that band.

**Controlled re-measurement**, 6 binaries (2 per variant), interleaved in a single time window, with `bench_narrow_control` as a per-binary normalizer:

| estimator | decline-path cost, varbinary 24×128 B |
|---|---|
| raw pooled median | +2.87% |
| narrow-normalized | +3.40% |
| binary-median means | ~+2.3% |
| independent third run | ~+1.5% |

**Honest range: ~1.5–3%, not 5.1%.** On the narrow control (39 int + 9 `varchar(6)`, no PLP at all) it is +0.87% and +2.15% across two runs — pooled ≈1.5%, which is *at* this instrument's noise floor and cannot be resolved further without ~8–10 builds per group. That also means the PR body's "+0.4% narrow" is too good; I'm correcting it upward.

**End-to-end, the tax is not measurable.** A 4-pass FDW-shaped Criterion matrix against #299 (1,000,049 rows, 60/60 cells, zero retries) put the decline arm's medians at −0.10%, −1.60%, −1.51%, +0.33%, +1.21% — all inside a ±6% empirical noise floor, three of five negative. The prior synthetic tax does not reproduce through transport and TLS.

**I did try the structural fix you're implying.** Collapsing the macro to a single shared `$fill` — one suspension point, no duplicated await — compiles and passes all 17 `sink_destination` tests. It measured **worse on both workloads** (varbinary 102.46/102.54 vs 99.51/100.72 for the current shape; narrow 87.89/88.30 vs 85.95/86.60), with clean 2-v-2 separation in one time window. The two-await shape stays.

**Where that leaves the criterion.** #253 asked for a flat `DefaultRowWriter` and called any nonzero regression a defect. On the corrected evidence that bar is not met and I'm not going to pretend otherwise: there is a small, real, sub-noise-floor cost on MAX-heavy rows. What I'm proposing instead is that the criterion was written before we knew the cost was this close to unmeasurable — and separately, the repo owner has since said explicitly that `DefaultRowWriter` gains are not required, only that it must not regress meaningfully. I'll amend the criterion on #253 with this data and update the PR body with the corrected figures and the layout-confound methodology, so nobody re-derives the 5.1% from a 1-vs-1 build comparison again.

> [!NOTE]

saurabh500:
Flagging a failing test on this branch's head, found while building #313 on top of it.

`connection::tds_client::tests::row_fetch_futures_stay_small` fails:

```
thread 'connection::tds_client::tests::row_fetch_futures_stay_small' panicked at
mssql-tds/src/connection/tds_client.rs:5111:13:
next_row_cursor future is 4248 B, expected <= 4224 B
```

**Evidence**

- Reproduced on a pristine `git worktree` checkout of `1f3dbb00d195e9cdc7ccee57df0aed72f35a17c1` (this PR's current head) with no other changes applied.
- Also reproduces on the parent commit `890ef66d7d6d8f14147a18b66383aa530c11427d`, so it is not from `1f3dbb0`.
- Same 4248 B under both the repo-pinned `1.95` and `ms-prod-1.93`.

`MAX` is 4224 and the test's own doc comment records `next_row_cursor` at 4152 B, so the guard's headroom is consumed and exceeded by 24 B.

I have not identified which field or type accounts for the growth, so I'm not going to speculate about it here.

**Why it may not have surfaced yet:** `mssql-rs Pull request validation` currently reports `skipping` on this PR, so the lib test suite is not running against it in CI.

Not actionable from #313 — that layer measures byte-identical on this test (4248 B both with and without its changes), so it can only be addressed at or below this PR.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -5075,9 +5075,15 @@ mod tests {
...
+    ///
+    /// The budget is deliberately loose enough to absorb ordinary decode work
+    /// and tight enough to catch a multi-kilobyte buffer landing in the frame.
+    /// `next_row_cursor` is the widest of these at 4152 B, up from 4008 B before
+    /// the row-writer value sink: offering a PLP destination puts a second
+    /// payload read in `decode_into`, one per arm of the offer.
     #[test]
     fn row_fetch_futures_stay_small() {
-        const MAX: usize = 4096;
+        const MAX: usize = 4224;
```
saurabh500: This is not good. Revert back to 4096 and fix the offender

saurabh500: Restored the 4096-byte limit and removed the raised-budget rationale. After merging the latest base, the future-size test now passes at 4096, so the base packet-reader/future changes eliminated the offender without weakening the guard.

> [!NOTE]

mssql-tds/src/datatypes/row_writer.rs
```diff
@@ -72,6 +72,95 @@ pub trait RowWriter {
...
+    /// # Contract
+    ///
+    /// A writer that returns `Some` must return a slice of exactly `length`
+    /// bytes and receives exactly one matching [`Self::commit_value`] call for
+    /// the same `col`. It does not additionally receive `write_bytes` or
+    /// `write_string` for that value.
+    ///
+    /// The slice must be a valid `&mut [u8]`, which means the `length` bytes
+    /// are already initialized. Handing back a slice built over uninitialized
+    /// allocation — `slice::from_raw_parts_mut` on fresh `palloc`ed memory, an
```
David-Engel: **Blocking (design, not a defect): `&mut [u8]` vs `&mut [MaybeUninit<u8>]` -- worth settling before this is public API.**

This paragraph is correct as Rust: building a `&mut [u8]` over uninitialized allocation is UB, and 4dcab61 was right to say so.

The problem is that #253's driving consumer does exactly that. Its stated pattern is `pg_sys::palloc` followed by `slice::from_raw_parts_mut`, and the issue calls out explicitly: *"`palloc` returns **uninitialised** memory, so there is **no zero-fill**."* Under this contract that consumer must switch to `palloc0`, adding a full memset per value -- which trades away part of the copy this PR removes.

`&mut [MaybeUninit<u8>]` would be sound *and* serve that consumer. `read_plp_chunks_into_slice` already guarantees every byte is written: the payload bytes, then `dest[offset..].fill(0)` for the tail. So it can soundly accept uninitialized storage and hand back an initialized slice.

Not asking for a rewrite, and I recognise this shape was chosen deliberately. But this is public API on a `#![warn(missing_docs)]` crate, and it is far cheaper to decide now than once consumers depend on the signature.

mssql-tds/src/datatypes/row_writer.rs
```diff
@@ -72,6 +72,95 @@ pub trait RowWriter {
...
+    ///
+    /// A writer must therefore treat a destination that is still pending at the
+    /// next [`Self::end_row`], or at the next `value_destination` for the same
+    /// column, as abandoned rather than asserting that it was committed. The
+    /// row it belonged to is not delivered in that case.
+    fn commit_value(&mut self, _col: usize, _complete: bool) {}
+}
+
+/// The kind of value a [`RowWriter::value_destination`] request is for.
+pub enum ValueKind<'a> {
```
David-Engel, line 156: **Suggestion:** mark this `#[non_exhaustive]` and derive `Debug`.

XML, JSON, and vector are all plausible future variants, and adding one to a public exhaustive enum is a breaking change for every downstream `match`. `Debug` also makes life easier for consumers writing their own sink tests.

```rust
/// The kind of value a [`RowWriter::value_destination`] request is for.
#[derive(Debug)]
#[non_exhaustive]
pub enum ValueKind<'a> {
```

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -72,6 +72,54 @@ macro_rules! read_sync_first {
...
+            // The writer borrow held by `$dest` ends with this read.
+            Some($dest) => {
```
David-Engel, line 109: **Suggestion:** nothing checks that the writer returned exactly `length` bytes.

The slice goes straight to `read_plp_chunks_into_slice`, which bounds-checks chunks against `dest.len()` rather than the wire-declared length. So a writer that returns an over-long slice silently relaxes the "PLP chunk exceeds declared length" protocol check, and one that returns an under-long slice turns valid wire data into a `ProtocolError`. Both are writer bugs, but they surface far from their cause.

The contract is documented on `value_destination`; a debug assertion makes it enforced in test builds at zero release cost:

```rust
Some($dest) => {
    debug_assert_eq!(
        $dest.len(), length,
        "value_destination returned {} bytes for a {length}-byte value", $dest.len()
    );
```

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -72,6 +72,54 @@ macro_rules! read_sync_first {
...
+//
+// The `None` arm expands to exactly the code that preceded this hook — allocate,
+// fill, hand over — so a writer that does not opt in decodes through an
+// unchanged instruction sequence. That is worth duplicating `$fill` across both
+// arms: sharing one binding would keep the fallback `Vec` and a discriminant
+// live across the read for every writer, opted in or not.
+//
+// Two reads means two suspension points in the caller's state machine, which
+// widens every row-decode future that embeds it; see
+// `row_fetch_futures_stay_small`.
```
David-Engel: **Nit:** this names a cost without a magnitude, which invites the next reviewer to escalate it. I measured it, so it may as well be written down.

The duplicated `$fill` adds **+192 B** to every row-decode future: `next_row_into` 3400 -> 3592, `read_row_column` 3752 -> 3944, `drain_rows` 3512 -> 3704. All within the 4096 B budget the guard enforces.

I bisected the attribution: reverting only the two `read_value_into!` call sites, while keeping the entire `read_plp_framing` / `read_plp_chunks_into_slice` / `read_plp_chunks_unknown_len` refactor, restores the base sizes exactly. So the growth is this macro, not the helper split. `row_writer.rs` contributes zero.

To be clear this is disclosure, not a measured cost -- size does not predict time on this path (#265: *"Size and time are anti-correlated here"*; #298 shrank a row future 32% and measured it slower). Suggest folding the number in so it does not have to be rediscovered:

```rust
// Two reads means two suspension points in the caller's state machine, which
// widens every row-decode future that embeds it by 192 B (`next_row_into`
// 3400 -> 3592, `read_row_column` 3752 -> 3944), within the 4096 B budget in
// `row_fetch_futures_stay_small`. Size does not predict time on this path
// (#265, #298), so this is disclosed, not a measured cost.
```

David-Engel (commented):
## Summary

Adds an opt-in `RowWriter::value_destination` / `commit_value` pair so a consumer can supply the final storage for known-length PLP `varchar(max)` / `nvarchar(max)` / `varbinary(max)` values, plus a refactor of PLP framing and chunk handling into shared helpers. The mechanics are clean, the refactor is faithful, and the soundness contract is documented correctly. One design question is worth settling before merge; the rest is polish. Detail is inline on the relevant lines.

Reviewed against `5967670f` (#299's head) rather than `main`, per the stacking note.

**Verification I ran:** `cargo fmt --check` clean. `cargo clippy -p mssql-tds --all-features --all-targets -D warnings` clean. `cargo test -p mssql-tds --lib` gives 1689 pass / 4 fail, the 4 being pre-existing `certificate_validator` failures from missing `tests/test_certificates/*` fixtures, unrelated to this PR. All 17 new sink tests pass. `row_fetch_futures_stay_small` passes.

## Two findings that do not map to a line

**Confirm `DefaultRowWriter` is flat using the benches that already exist.** `get_next_row()` builds a `DefaultRowWriter` and calls `next_row_into` -> `decode_into` (tds_client.rs:2805), so `datatypes::lob` (`VARCHAR(MAX)`, 1 MB / 20 MB) and `tds_specific`'s `#packet_payload` already exercise the changed known-length PLP path. A baseline-vs-candidate run of those two settles the non-opt-in question cheaply.

I am explicitly *not* asking for a new benchmark for the gain side: that needs a writer overriding `value_destination`, none exists in the workspace, and the driving consumer is out of tree, so the reference evidence in #253 is a reasonable basis. Worth noting the suite has no "many small MAX values per row" shape, which is where this PR's own argument says the win concentrates -- `lob` amortizes per-value overhead to nearly nothing at 1 MB.

**`Fixes #253` overclaims.** This is tier 1 of the three tiers described in that issue; tier 2 (`write_string_ref` / `write_bytes_ref` and the packet-reader borrow capability) is not implemented here. `Refs #253` would be more accurate so the issue stays open for the remaining tiers.

## Nothing found on security

Encrypted columns with a live decryptor go through `decode` (owned), so ciphertext reaches a sink only in the already-documented no-decryptor case. `dest[offset..].fill(0)` prevents a short value exposing prior buffer contents. `read_plp_framing` retains both the negative-length and `MAX_PLP_SIZE` checks.

*Note: a side conversation on this review produced #334, about repo guidance that is unrelated to this PR. No action needed here.*

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -1090,99 +1138,154 @@ impl GenericDecoder {
...
-                offset += chunk_size_read;
-                chunk_len = read_sync_first!(reader, try_read_uint32, read_uint32) as usize;
+            let end_offset = offset.checked_add(chunk_len).ok_or_else(|| {
+                crate::error::Error::ProtocolError(format!(
+                    "PLP chunk offset would overflow: {offset} + {chunk_len}"
+                ))
+            })?;
+            if end_offset > dest.len() {
+                return Err(crate::error::Error::ProtocolError(format!(
+                    "PLP chunk exceeds declared length: offset={offset}, chunk_len={chunk_len}, declared_len={}",
```
David-Engel: **Nit:** `declared_len` is misleading on the sink path.

`dest.len()` is the writer's buffer length here, not the length the wire declared. Those are the same on the owned path, but they can differ precisely in the case this message fires for -- a writer that returned a wrong-sized slice. The message then misattributes a client bug to the server.

Passing the declared length through from `PlpFraming::Known(length)` and reporting both would make it self-diagnosing.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -1090,99 +1138,154 @@ impl GenericDecoder {
...
-                        "Too many PLP chunks: {chunk_count} (max {})",
-                        Self::MAX_PLP_CHUNKS
-                    )));
-                }
+    /// Reads the chunk stream of a known-length PLP value straight into `dest`.
+    ///
+    /// A stream that ends before filling `dest` leaves the remainder zeroed, so
+    /// a short value reaches the writer exactly as it would on the owned path
+    /// rather than exposing whatever the destination held beforehand.
+    async fn read_plp_chunks_into_slice<T>(reader: &mut T, dest: &mut [u8]) -> TdsResult<()>
```
David-Engel: **Nit:** the known-length path lost its `#[cfg(fuzzing)]` `[ALLOC]` trace.

Previously `read_plp_bytes` emitted it for both known and unknown length values; it now only survives in `read_plp_chunks_unknown_len`. Nothing in the repo consumes that output today, so this is low stakes, but the instrumentation used to be symmetric across both PLP shapes and now is not.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -4888,4 +5030,431 @@ mod test {
...
+                kind: ValueKind<'_>,
+                length: usize,
+            ) -> Option<&'a mut [u8]> {
+                if !self.accept {
+                    return None;
+                }
+                let start = self.storage.len();
+                // 0xAA marks untouched bytes so a short fill is visible in asserts.
+                self.storage.resize(start + length, 0xAA);
+                self.pending = Some((start, matches!(kind, ValueKind::String(_))));
```
David-Engel: **Suggestion:** this collapses the `EncodingType` to a bool, so the encoding handed to the sink is never asserted.

Passing the wire encoding through is one of the PR's stated benefits ("string consumers also receive the wire encoding and can defer transcoding"), but a bug that passed `Utf16` where `Utf8` was correct would keep every test green. Storing the `EncodingType` itself and asserting it would close that.

Related gaps in this module, all cheap to add:
- Only `NVarChar` PLP is covered. No `varchar(max)` case exercising `EncodingType::Utf8` or `LcidBased`, which are the branches a non-Unicode collation takes.
- Nothing asserts that XML / JSON / UDT / vector PLP columns are *not* offered a destination. That is stated as intentional scope, so it is worth pinning.
- No test for a writer returning a wrong-length slice (see my comment on the macro).

mssql-tds/src/datatypes/row_writer.rs
saurabh500: Addressed in `6395738c`. `value_destination` now returns `&mut [MaybeUninit<u8>]`, and the internal packet-reader seam supports filling that storage soundly. The production network transport copies directly into the uninitialized destination; `commit_value(true)` now guarantees all bytes are initialized.

> [!NOTE]

mssql-tds/src/datatypes/row_writer.rs
saurabh500: Addressed in `6395738c`: `ValueKind` now derives `Debug` and is marked `#[non_exhaustive]`. `EncodingType` also derives `Debug` so the derived implementation is available.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Addressed in `6395738c`. The sink branch now asserts the exact destination length before reading, and a wrong-sized-destination test pins the diagnostic.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Addressed in `6395738c`. The known-length chunk loop again emits a `#[cfg(fuzzing)]` `[ALLOC]` trace with chunk and declared-length details.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Addressed in `6395738c`. The macro comment now records the 192-byte increase, representative before/after sizes, and the unchanged 4096-byte guard, while explicitly avoiding treating future size as a timing proxy.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Addressed in `6395738c`. Tests now retain and compare the exact `EncodingType`, cover `varchar(max)` with an LCID-based encoding, assert XML/JSON/UDT/vector never request destinations, and reject wrong-sized destination buffers.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Addressed in `6395738c`. The helper now receives the wire-declared length separately and reports both `wire_declared_len` and `destination_len`, so a writer sizing bug is not attributed to the server.

> [!NOTE]

mssql-tds/src/datatypes/row_writer.rs
```diff
@@ -72,6 +74,89 @@ pub trait RowWriter {
...
+        _kind: ValueKind<'_>,
+        _length: usize,
+    ) -> Option<&'a mut [MaybeUninit<u8>]> {
+        None
+    }
+
+    /// Completes a value whose storage came from [`Self::value_destination`].
+    ///
+    /// `complete` is `false` when decoding failed partway through; the slice
+    /// contents are then unspecified and the writer must discard the value.
```
David-Engel: **Follow-up (contract wording).** This is the one I would most like to see fixed, though it is doc-only.

`value_destination` now gets this exactly right: *"the writer must discard the destination without reading it."* But this line still carries the pre-`MaybeUninit` wording -- "unspecified" -- and this is the doc a writer author reads while implementing `commit_value`.

"Unspecified" reads as "garbage bytes, meaningless but safe to look at". Under the new signature the destination may be genuinely uninitialized on the `false` path, so reading it is UB. That is precisely the class of bug the `MaybeUninit` change was made to prevent, so it would be a shame for the contract itself to still point at it.

Smallest fix:

```rust
/// `complete` is `false` when decoding failed partway through; the destination
/// may then contain uninitialized bytes, so the writer must discard it without
/// reading it.
```

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -72,6 +73,68 @@ macro_rules! read_sync_first {
...
+                    $dest.len()
+                );
+                let outcome = $fill;
+                $writer.commit_value(col, outcome.is_ok());
+                outcome?;
+            }
+            None => {
+                let mut $bytes: Vec<u8> = vec![0u8; length];
+                {
+                    // SAFETY: initialized bytes are valid `MaybeUninit<u8>` values.
```
David-Engel: **Follow-up (fragility, not a defect).** This is sound today and I want to be clear about that -- the only writers are `read_bytes_uninit` implementations, and none of them writes `MaybeUninit::uninit()`.

The fragility is that the *type* permits it. Handing out `&mut [MaybeUninit<u8>]` over a buffer a `Vec<u8>` believes is initialized grants the callee permission to de-initialize it; if that ever happened, `$owned` would then read the `Vec` as `u8` and that is UB. The SAFETY comment justifies the cast direction but not the write-back obligation, so the invariant is load-bearing and unstated.

`Vec::spare_capacity_mut()` is the idiomatic way to get this soundly, and it has a side benefit worth having on this particular path -- it drops the redundant `vec![0u8; length]` zeroing from the decline path, which is the path this PR is most careful not to tax:

```rust
None => {
    let mut $bytes: Vec<u8> = Vec::with_capacity(length);
    {
        let $dest: &mut [MaybeUninit<u8>] = &mut $bytes.spare_capacity_mut()[..length];
        $fill?;
    }
    // SAFETY: `$fill` initialized all `length` bytes on the success path.
    unsafe { $bytes.set_len(length) };
    $owned;
}
```

The error path stays correct: no `set_len`, so the `Vec` drops at length 0.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -1090,99 +1153,172 @@ impl GenericDecoder {
...
-                    .read_bytes(&mut plp_buffer[offset..offset + chunk_len])
-                    .await?;
-                offset += chunk_size_read;
-                chunk_len = read_sync_first!(reader, try_read_uint32, read_uint32) as usize;
+            let end_offset = offset.checked_add(chunk_len).ok_or_else(|| {
+                crate::error::Error::ProtocolError(format!(
+                    "PLP chunk offset would overflow: {offset} + {chunk_len}"
+                ))
+            })?;
+            if end_offset > dest.len() {
```
David-Engel: **Follow-up (release-build protocol check).** This bounds against the destination rather than against the wire.

In debug the `debug_assert_eq!` in the macro forces `dest.len() == declared_len`, so the two are equivalent. In release the assert compiles out, so a writer that returns an over-long slice would let a server send chunks totalling more than the length it declared, and they would be accepted.

`declared_len` is now a parameter, so tightening this is free and keeps the memory-safety bound intact:

```rust
if end_offset > declared_len || end_offset > dest.len() {
```

Related: `wrong_sized_destination_is_rejected` relies on the `debug_assert`, so it only exercises this in debug builds. That is fine today -- I checked and no CI workflow runs the suite in release -- but it does mean the release behaviour above is untested.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -1090,99 +1153,172 @@ impl GenericDecoder {
...
+    {
+        let mut chunk_len = read_sync_first!(reader, try_read_uint32, read_uint32) as usize;
+        let mut offset: usize = 0;
+        let mut chunk_count = 0u32;
+
+        while chunk_len > 0 {
+            chunk_count += 1;
+            #[cfg(fuzzing)]
+            eprintln!(
+                "[ALLOC] read_plp_chunks_into_slice: chunk #{chunk_count}, chunk_len={chunk_len}, declared_len={}",
```
David-Engel: **Nit.** The restored trace labels `dest.len()` as `declared_len`, which reintroduces the exact conflation just fixed three lines below -- the error message now correctly separates `wire_declared_len` from `destination_len`.

Since `declared_len` is in scope as a real parameter with a genuinely different value, suggest matching the error message's vocabulary:

```rust
"[ALLOC] read_plp_chunks_into_slice: chunk #{chunk_count}, chunk_len={chunk_len}, wire_declared_len={declared_len}, destination_len={}",
dest.len()
```

David-Engel (approved):
## Approving

All nine findings from my previous review are addressed, and the `MaybeUninit` rework is sound. The four items below are non-blocking follow-ups -- take them here or in a follow-up, none should hold the merge.

**Verification I ran at this head:** `cargo fmt --check` clean. `cargo clippy -p mssql-tds --all-features --all-targets -D warnings` clean. `cargo test -p mssql-tds --lib` gives 1696 pass / 4 fail, the 4 being the pre-existing `certificate_validator` fixture failures unrelated to this PR. Sink tests are up from 17 to 20 and all pass. `row_fetch_futures_stay_small` passes with `MAX` back at 4096.

## The soundness rework holds up

I audited the three things that could have gone wrong rather than taking the summary on trust:

- `NetworkTransport::read_bytes_uninit` is a faithful structural mirror of the existing `read_bytes` -- same loop shape, same `to_read` computation. The `copy_nonoverlapping` preconditions check out: `offset + to_read <= buffer.len()` follows from the `length_to_read` invariant, and source and destination are distinct allocations.
- **The initialization guarantee is real.** `offset` advances by the *returned* count rather than by `chunk_len`, so a partial read leaves no uninitialized gap -- coverage of `0..final_offset` stays contiguous, and `dest[offset..].fill(MaybeUninit::new(0))` closes the tail. So `commit_value(col, true)` genuinely implies every element is initialized, which is exactly what the contract promises.
- **The optimization is actually reachable in production.** `AnyTransport::Network` calls `NetworkTransport` directly rather than through `dyn`, so the override is on the real row path. Everything else falls back to the zero-filling default in `packet_reader.rs`, which is sound.

## On the measurement work

This is better than what I asked for. Tracing the original +5.1% to a link-layout confound -- via inert `--cfg probe_a` / `--cfg probe_b` builds showing a ~5% spread from layout alone -- and then revising the number *downward* to ~1.5-3% is the right instinct, and the end-to-end matrix showing the tax does not reproduce through transport and TLS is the more decision-relevant result.

It also settles my earlier nit against me: you tried collapsing the macro to a single shared `$fill`, and it measured worse on both workloads. The two-await shape stays, and the comment now records the 192 B without treating it as a timing proxy. Good.

One loose end: the reply commits to updating the PR body with the corrected figures, but the Performance section still carries no numbers. Worth landing before merge so the ~1.5-3% decline-path cost lands in the merge commit rather than living only in a review thread.

mssql-tds/src/datatypes/row_writer.rs
saurabh500: Fixed in `c0b4e127`. The `commit_value(false)` contract now explicitly says the destination may contain uninitialized bytes and must be discarded without reading.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Fixed in `c0b4e127`. The fuzz trace now reports `wire_declared_len` and `destination_len` separately, matching the diagnostic vocabulary.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Fixed in `c0b4e127`. The owned fallback now fills `Vec::spare_capacity_mut()` and calls `set_len` only after a successful fill, avoiding both the write-back invariant on initialized `Vec` storage and redundant zeroing.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Fixed in `c0b4e127`. The chunk bound now checks both the wire-declared length and destination length in release builds, with a dedicated release-mode test proving an overlong destination cannot relax the protocol limit.

> [!NOTE]

Vahid-b (approved):

PR #313, opened 2026-08-16
Decode and format decimals without allocating

### LINKED ISSUE #312: Decode and format `decimal`/`numeric` without allocating (row decode perf, axis: value representation)
### Problem statement

Every decoded `decimal`/`numeric` value costs two heap allocations in `GenericDecoder::read_decimal_data`: a zeroed staging `Vec<u8>` and a `Vec<i32>` of 32-bit words. Together they move at most 16 bytes that are already sitting in the packet buffer.

Reading the value back out costs more. The only lossless accessor is `Display`/`to_decimal_string`, which allocates a `String` for the digits and then more for the decimal point and the sign.

Measured on this machine, per value, over the four numeric columns of the `pg_fabric_fdw` benchmark table (`decimal(15,2)` ×2, `decimal(18,3)`, `decimal(19,4)`):

| stage | ns/value |
| --- | --- |
| decode | 34.8 |
| format | 83.2 |

The consumer driving this is a PostgreSQL FDW (`pg_fabric_fdw`) that streams Fabric Data Warehouse results and materialises them as PG datums. Its `numeric` path currently goes **integer → text → integer**: `mssql-tds` decodes the wire words into a `Vec<i32>`, flattens them to a `String`, and the FDW hands that string to PostgreSQL's `numeric_in` to parse it back into a positional integer. On a 1,000,049-row scan that round trip is ~1160 ms slower than an experimental branch that keeps the value as an integer end to end (3306 ms → 1511 ms, 1.77x).

`MAX_DECIMAL_INT_PARTS = 4` already encodes the fact that a precision ≤ 38 magnitude is < 10³⁸ < 2¹²⁷, so the `Vec<i32>` is always representing a value that fits in a `u128`. The allocations buy nothing.

### Proposed solution

Carry the magnitude inline, decode into a stack array, and render into a caller-owned buffer. Output stays byte-for-byte identical at every step; only the representation and the allocations change.

- **Inline magnitude.** `DecimalParts` becomes `Copy` and holds the magnitude inline instead of in a `Vec<i32>`, with `DecimalParts::new(is_positive, precision, scale, magnitude: u128)` and a `magnitude() -> u128` accessor. `word(i)` / `word_count()` / `from_words(..)` cover the wire encoders and the JS FFI, which need the little-endian 32-bit view.
- **Stack decode.** `read_decimal_data` reads into a `[u8; 16]` and reassembles with `u128::from_le_bytes`. The `> MAX_DECIMAL_INT_PARTS` rejection, the zero-padded partial trailing word, NULL handling and full consumption of the declared length are all preserved.
- **Allocation-free format.** A `format_into(&mut [u8; DECIMAL_STR_LEN]) -> &str` entry point renders into a caller-owned buffer; `Display` and `to_decimal_string` become thin wrappers over it.

This also unblocks the follow-up that actually closes the FDW gap: a base-10000 digit export, which lets a consumer build a PG `numeric` directly and skip `numeric_in` entirely. That is only reachable once `mssql-tds` stops flattening the value to text.

### Affected crate

mssql-tds

### Alternatives considered

**Leave the `Vec<i32>` and only add an allocation-free formatter.** Removes the format allocation but not the decode allocations, and still leaves the consumer parsing text.

**Expose the `Vec<i32>` words and let the consumer reassemble.** No new API surface, but it keeps two allocations per value on the hot path for data the decoder already has in a contiguous buffer.

**Hold the magnitude as a `u128` field rather than two `u64` halves.** Simpler, but a `u128` field gives `DecimalParts` — and through it `ColumnValues` — 16-byte alignment, which propagates into the row-fetch async state machines and grows them measurably. Two `u64` halves behind a `u128` accessor keep 8-byte alignment.

### Additional context

Part of the row decode performance work tracked in #247, on a fifth axis (value representation) alongside dispatch, precomputation, buffered reads and value handoff.

Two traps worth recording for whoever picks up the base-10000 follow-up:

- **Do not pre-multiply the magnitude by `10^pad`.** For `decimal(38,s)` where `s % 4 != 0` the magnitude reaches `10^38 − 1`; multiplying by 100 gives ~`10^40`, which exceeds `u128::MAX` (~3.40e38) and wraps silently in release builds. Convert to base-10000 first, then scale the digit array with carry propagation, where the maximum intermediate fits in a `u32`.
- **A `u128 % 10` digit loop is slower than `u128::to_string()`** on wide values, because it emits a `__udivti3` libcall per digit. Split the magnitude into limbs of `10^19` and generate every digit with `u64` arithmetic instead.

saurabh500:
## Description

Every decoded `decimal`/`numeric` cost two heap allocations in `read_decimal_data` — a staging `Vec<u8>` plus a `Vec<i32>` of words — to move at most 16 bytes already sitting in the packet buffer. The only lossless way to read the value back out was `Display`, which allocated a `String` for the digits and more for the point and sign.

`MAX_DECIMAL_INT_PARTS = 4` already encodes that a precision ≤ 38 magnitude is < 10³⁸ < 2¹²⁷, so this carries it inline instead. **Output is byte-for-byte identical at every step**; only the representation and the allocations change.

- `DecimalParts` is `Copy` and holds the magnitude inline. `new(is_positive, precision, scale, magnitude: u128)` and `magnitude() -> u128` replace the `int_parts` field.
- `read_decimal_data` decodes into a `[u8; 16]` on the stack and reassembles with `u128::from_le_bytes`.
- `format_into(&mut [u8; DECIMAL_STR_LEN]) -> &str` renders into a caller-owned buffer. `Display` and `to_decimal_string` are thin wrappers over it.
- `word(i)` / `word_count()` / `from_words(..)` give wire encoders the little-endian 32-bit view they need; `try_from_words(..)` lets trust-boundary callers reject magnitudes wider than 128 bits.

The consumer driving this is a PostgreSQL FDW (`pg_fabric_fdw`) whose `numeric` path is currently integer → text → integer: we flatten to a `String`, it hands that to PostgreSQL's `numeric_in` to parse back into a positional integer. This layer removes the allocations and the mandatory text round trip; the follow-up that lets a consumer skip `numeric_in` entirely is scoped below.

### Two design points worth flagging

**The magnitude is stored as `[u64; 2]`, not a `u128`.** A `u128` field would give `DecimalParts` — and through it `ColumnValues` — 16-byte alignment, which propagates into the row-fetch async state machines and grew `next_row_cursor` by 176 B. As two halves the struct is **24 B at align 8**, against 32 B at align 16, so it is 8 B *smaller* than the `Vec`-based original and the row-fetch futures are byte-for-byte unchanged. `magnitude()` is one shift and one or.

**`format_into` generates every digit with `u64` arithmetic.** A `u128 % 10` loop emits a `__udivti3` libcall per digit and loses to `u128::to_string` on wide values. The magnitude is split into limbs of 10¹⁹ — at most two `u128` divisions, none at all when it already fits a `u64`. **Three** limbs, not two: a `u128` reaches 39 digits, and a two-limb split silently truncates above ~1.845e38. That bug was live in the reference kernel this was modelled on and was caught here only because the test oracle shares no arithmetic with the code it validates.

### Removed API

`magnitude() -> Option<u128>` and `magnitude_big() -> BigUint` are gone — with the new representation their `None` and `BigUint` arms are unreachable by construction. That also removes an `mssql-odbc` conversion branch and a `cell.rs` debug assertion whose inputs can no longer be built. `bigdecimal`/`num-traits` are no longer used on the decode path.

## Results

Criterion, `mssql-tds/benches/decimal.rs`, ns per value. `previous` is the pre-change shape reproduced locally in the bench and cross-validated against the current one before timing. `fdw` is the four numeric columns of the FDW benchmark table (`decimal(15,2)` ×2, `decimal(18,3)`, `decimal(19,4)`); `wide` is `decimal(38,10)` near 10³⁸.

| workload | previous | current | speedup |
| --- | ---: | ---: | ---: |
| decode `fdw` | 34.75 | **4.19** | **8.3x** |
| decode `wide` | 40.09 | **4.18** | **9.6x** |
| format `fdw` — `format_into` | 83.15 | **17.05** | **4.9x** |
| format `fdw` — `to_decimal_string` | 83.15 | 30.46 | 2.7x |
| format `fdw` — `Display` | 83.15 | 42.42 | 2.0x |
| format `wide` — `format_into` | 155.89 | **88.33** | **1.8x** |
| format `wide` — `to_decimal_string` | 155.89 | 106.33 | 1.5x |
| format `wide` — `Display` | 155.89 | 121.33 | 1.3x |

`Display` stays the slowest reading because it still pays for `fmt::Formatter`; callers that can hold a buffer should use `format_into`.

## Correctness

Output equality is the whole requirement, so it is tested against an oracle that is the **verbatim pre-change algorithm** (`u128::to_string()` + index splice + sign `format!`). It shares no arithmetic with the kernel, which is exactly why it caught the three-limb bug.

- Exhaustive magnitudes 0..=4096 × scales 0..=6, both signs.
- Every 10ᵏ and 10ᵏ−1 for k = 1..=38, plus `u128::MAX`, against every scale 0..=38.
- 20 000 seeded randomized magnitudes (1–128 bits) × random scales, both signs.
- Precision 1..=38 with every valid scale.
- Negative zero, zero with scale, magnitude narrower than scale.
- An out-of-domain `scale` (it arrives unvalidated from the wire) renders short instead of overrunning: worst case is exactly 48 bytes, so the `DECIMAL_STR_LEN - 3` clamp is tight rather than generous.
- Decode: wire widths 5/9/13/17 reassemble the same magnitude, negative sign byte, partial trailing word fully consumed, oversized payload rejected.
- JS FFI: non-zero words past the fourth are rejected rather than silently truncated; trailing zero padding remains accepted.

## Deferred

- **Base-10000 digit export**, so an FDW can build a PG `numeric` directly and skip `numeric_in`. That is the ~800 ms of the measured gap that this layer only unblocks rather than collects. **Trap for whoever picks it up:** do not pre-multiply the magnitude by `10^pad`. For `decimal(38,s)` with `s % 4 != 0` the magnitude reaches 10³⁸−1, and ×100 exceeds `u128::MAX` and wraps silently in release. Convert to base-10000 first, then scale the digit array with carry propagation.
- **Sync-first byte-run reads** on the decimal path — overlaps the sync-first work in #303–#307.

## Related Issues

Fixes #312

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] Targeted decimal unit tests pass (72 tests)

`mssql-js` builds and `index.d.ts` is unchanged, so the FFI surface is stable.

### DIFF
--- a/mssql-js/src/binary_row_writer.rs
+++ b/mssql-js/src/binary_row_writer.rs
@@ -222,9 +222,9 @@ impl RowWriter for BinaryRowWriter {
         self.row_data.push(val.is_positive as u8);
         self.row_data.push(val.scale);
         self.row_data.push(val.precision);
-        self.row_data.push(val.int_parts.len() as u8);
-        for part in &val.int_parts {
-            self.row_data.extend_from_slice(&part.to_le_bytes());
+        self.row_data.push(val.word_count() as u8);
+        for i in 0..val.word_count() {
+            self.row_data.extend_from_slice(&val.word(i).to_le_bytes());
         }
     }
 
@@ -234,9 +234,9 @@ impl RowWriter for BinaryRowWriter {
         self.row_data.push(val.is_positive as u8);
         self.row_data.push(val.scale);
         self.row_data.push(val.precision);
-        self.row_data.push(val.int_parts.len() as u8);
-        for part in &val.int_parts {
-            self.row_data.extend_from_slice(&part.to_le_bytes());
+        self.row_data.push(val.word_count() as u8);
+        for i in 0..val.word_count() {
+            self.row_data.extend_from_slice(&val.word(i).to_le_bytes());
         }
     }
 
--- a/mssql-js/src/binary_row_writer.rs
+++ b/mssql-js/src/binary_row_writer.rs
@@ -222,22 +222,15 @@ impl RowWriter for BinaryRowWriter {
         self.row_data.push(val.is_positive as u8);
         self.row_data.push(val.scale);
         self.row_data.push(val.precision);
-        self.row_data.push(val.word_count() as u8);
-        for i in 0..val.word_count() {
+        let word_count = val.word_count();
+        self.row_data.push(word_count as u8);
+        for i in 0..word_count {
             self.row_data.extend_from_slice(&val.word(i).to_le_bytes());
         }
     }
 
-    fn write_numeric(&mut self, _col: usize, val: DecimalParts) {
-        // Same encoding as decimal
-        self.row_data.push(TAG_DECIMAL);
-        self.row_data.push(val.is_positive as u8);
-        self.row_data.push(val.scale);
-        self.row_data.push(val.precision);
-        self.row_data.push(val.word_count() as u8);
-        for i in 0..val.word_count() {
-            self.row_data.extend_from_slice(&val.word(i).to_le_bytes());
-        }
+    fn write_numeric(&mut self, col: usize, val: DecimalParts) {
+        self.write_decimal(col, val);
     }
 
     fn write_date(&mut self, _col: usize, val: SqlDate) {
--- a/mssql-tds/src/datatypes/decoder.rs
+++ b/mssql-tds/src/datatypes/decoder.rs
@@ -1,8 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use bigdecimal::num_bigint::BigUint;
-use bigdecimal::num_traits::ToPrimitive;
 use core::fmt;
 use std::future::Future;
 use std::mem::MaybeUninit;
@@ -193,6 +191,9 @@ const MAX_PLP_SIZE: usize = i32::MAX as usize;
 /// the widest value the TDS wire format carries is 17 bytes (sign byte plus four
 /// words), so anything longer is malformed.
 const MAX_DECIMAL_INT_PARTS: usize = 4;
+
+/// Byte width of a `decimal`/`numeric` magnitude: four little-endian 32-bit words.
+const DECIMAL_MAGNITUDE_BYTES: usize = MAX_DECIMAL_INT_PARTS * 4;
 
 // Helper function to validate allocation size before allocating
 #[inline]
@@ -826,25 +827,18 @@ impl GenericDecoder {
             )));
         }
 
-        validate_alloc_size(number_of_int_parts * 4, "read_decimal int_parts")?;
-        let mut magnitude = vec![0u8; number_of_int_parts * 4];
+        // The check above bounds the payload at 16 bytes, so it is read onto the
+        // stack. TDS is little-endian and the array is zero-filled, which both
+        // reassembles the magnitude and zero-pads a partial trailing word.
+        let mut magnitude = [0u8; DECIMAL_MAGNITUDE_BYTES];
         reader.read_bytes(&mut magnitude[..magnitude_len]).await?;
 
-        let int_parts = magnitude
-            .chunks_exact(4)
-            .map(|chunk| {
-                let mut word = [0u8; 4];
-                word.copy_from_slice(chunk);
-                i32::from_le_bytes(word)
-            })
-            .collect();
-
-        Ok(Some(DecimalParts {
+        Ok(Some(DecimalParts::new(
             is_positive,
-            scale,
             precision,
-            int_parts,
-        }))
+            scale,
+            u128::from_le_bytes(magnitude),
+        )))
     }
 
     async fn read_datetime<T>(&self, reader: &mut T) -> TdsResult<SqlDateTime>
@@ -2148,8 +2142,12 @@ impl SqlTypeDecode for StringDecoder {
     }
 }
 
+/// Bytes needed to render any `decimal`/`numeric`: sign, 38 digits, decimal
+/// point, and slack for a scale that exceeds the digit count (`0.000…`).
+pub const DECIMAL_STR_LEN: usize = 48;
+
 /// TDS representation of Decimal and Numeric types.
-#[derive(Clone)]
+#[derive(Clone, Copy, PartialEq, Eq)]
 pub struct DecimalParts {
     /// `true` for non-negative values.
     pub is_positive: bool,
@@ -2157,17 +2155,41 @@ pub struct DecimalParts {
     pub scale: u8,
     /// Total number of significant digits.
     pub precision: u8,
-    /// 32-bit integer parts of the value (little-endian order).
-    pub int_parts: Vec<i32>,
+    /// The magnitude, as two 64-bit halves, least significant first.
+    ///
+    /// Split rather than held as a `u128` because a `u128` field would give
+    /// `DecimalParts` — and through it `ColumnValues` — 16-byte alignment,
+    /// which propagates into the row-fetch async state machines and grows them
+    /// by ~4%. As two halves the struct is 24 bytes at align 8, against 32 at
+    /// align 16; [`Self::magnitude`] recombines them with a shift and an or.
+    magnitude: [u64; 2],
 }
 
 impl fmt::Display for DecimalParts {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.to_decimal_string())
+        f.write_str(self.format_into(&mut [0u8; DECIMAL_STR_LEN]))
     }
 }
 
 impl DecimalParts {
+    /// Builds a value from its unsigned magnitude, already scaled by `10^scale`.
+    ///
+    /// SQL Server's maximum precision of 38 digits keeps the magnitude below
+    /// `10^38 < 2^127`, so every valid `decimal`/`numeric` fits.
+    pub const fn new(is_positive: bool, precision: u8, scale: u8, magnitude: u128) -> Self {
+        DecimalParts {
+            is_positive,
+            scale,
+            precision,
+            magnitude: [magnitude as u64, (magnitude >> 64) as u64],
+        }
+    }
+
+    /// The unsigned magnitude, already scaled by `10^scale`.
+    pub const fn magnitude(&self) -> u128 {
+        (self.magnitude[0] as u128) | ((self.magnitude[1] as u128) << 64)
+    }
+
     /// Create a DecimalParts from a decimal string using BigDecimal.
     ///
     /// Supports SQL Server's full 38-digit precision using arbitrary-precision arithmetic.
@@ -2211,12 +2233,7 @@ impl DecimalParts {
 
         // Handle zero case (but preserve sign from original string)
         if decimal.is_zero() {
-            return Ok(DecimalParts {
-                is_positive: !has_negative_sign,
-                scale,
-                precision,
-                int_parts: vec![0],
-            });
+            return Ok(Self::new(!has_negative_sign, precision, scale, 0));
         }
 
         // Extract sign for non-zero values
@@ -2250,55 +2267,40 @@ impl DecimalParts {
             )));
         }
 
-        // Convert BigInt to Vec<i32> for TDS wire format (little-endian 32-bit chunks)
+        // Convert BigInt to the little-endian 128-bit magnitude. A precision <= 38
+        // value never needs more than 16 bytes; a caller that declares a wider
+        // precision is refused rather than silently truncated.
         let bytes = final_bigint.to_signed_bytes_le();
-        let mut int_parts = Vec::new();
-        let mut i = 0;
-        while i < bytes.len() {
-            let mut part: i32 = 0;
-            for j in 0..4 {
-                if i + j < bytes.len() {
-                    part |= (bytes[i + j] as i32) << (j * 8);
-                }
-            }
-            int_parts.push(part);
-            i += 4;
+        if bytes.len() > DECIMAL_MAGNITUDE_BYTES {
+            return Err(crate::error::Error::TypeConversionError(format!(
+                "Decimal value '{s}' does not fit in a 128-bit magnitude"
+            )));
         }
+        let mut magnitude_bytes = [0u8; DECIMAL_MAGNITUDE_BYTES];
+        magnitude_bytes[..bytes.len()].copy_from_slice(&bytes);
 
-        if int_parts.is_empty() {
-            int_parts.push(0);
-        }
-
-        Ok(DecimalParts {
+        Ok(Self::new(
             is_positive,
-            scale,
             precision,
-            int_parts,
-        })
+            scale,
+            u128::from_le_bytes(magnitude_bytes),
+        ))
     }
 
     /// Create a DecimalParts from an i64 value.
     pub fn from_i64(value: i64, precision: u8, scale: u8) -> TdsResult<Self> {
         let is_positive = value >= 0;
-        let abs_value = value.unsigned_abs();
 
-        // Scale the value by multiplying by 10^scale
-        let scaled_value = abs_value as u128 * 10u128.pow(scale as u32);
+        let magnitude = 10u128
+            .checked_pow(u32::from(scale))
+            .and_then(|factor| u128::from(value.unsigned_abs()).checked_mul(factor))
+            .ok_or_else(|| {
+                crate::error::Error::TypeConversionError(format!(
+                    "Decimal value {value} scaled by 10^{scale} does not fit in a 128-bit magnitude"
+                ))
+            })?;
 
-        // Convert to int_parts
-        let mut int_parts = Vec::new();
-        let mut remaining = scaled_value;
-        while remaining > 0 || int_parts.is_empty() {
-            int_parts.push((remaining & 0xFFFFFFFF) as i32);
-            remaining >>= 32;
-        }
-
-        Ok(DecimalParts {
-            is_positive,
-            scale,
-            precision,
-            int_parts,
-        })
+        Ok(Self::new(is_positive, precision, scale, magnitude))
     }
 
     /// Create a DecimalParts from an f64 value.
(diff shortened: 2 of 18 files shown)


mssql-tds/src/datatypes/decoder.rs
```diff
@@ -2308,113 +2310,150 @@ impl DecimalParts {
...
-            return false;
-        }
-        self.is_positive == other.is_positive
-            && self.scale == other.scale
-            && self.precision == other.precision
+    /// Builds a value from little-endian 32-bit words, least significant first.
+    ///
+    /// Words past the fourth are ignored: a `decimal`/`numeric` with precision
+    /// <= 38 never sets them, and every encoder here already emits at most four.
+    pub fn from_words(is_positive: bool, precision: u8, scale: u8, words: &[i32]) -> Self {
```
saurabh500: **Suggestion** - `from_words` silently drops words past the fourth.

The behavior is deliberate and pinned by `test_decimal_parts_from_words_ignores_oversized_words`, and the write path already truncated to four words, so I don't believe this is reachable today. But it converts a previously *detectable* condition into silently wrong data at a trust boundary: `From<NapiDecimalParts> for DecimalParts` (`mssql-js/src/ffidatatypes.rs:201`) takes the word vector straight from JS. Before this change an oversized magnitude made `magnitude()` return `None`, which `mssql-odbc`'s `numeric_source` turned into `ConvError::OutOfRange` - and this PR deletes that test. Now `from_words(true, 38, 0, &[1, 0, 0, 0, 1])` quietly becomes `1`.

Cheapest mitigation, keeping release behavior identical and catching drift in debug/CI:

```rust
pub fn from_words(is_positive: bool, precision: u8, scale: u8, words: &[i32]) -> Self {
    debug_assert!(
        words.iter().skip(MAX_DECIMAL_INT_PARTS).all(|&w| w == 0),
        "magnitude wider than 128 bits truncated: {words:?}"
    );
    // ...
}
```

If you'd rather surface it to callers, a `try_from_words(..) -> Option<Self>` alongside this one would let the FFI path reject instead of wrap, restoring the old `OutOfRange` behavior without reintroducing `BigUint`.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -2308,113 +2310,150 @@ impl DecimalParts {
...
+        (self.magnitude() >> (index * 32)) as u32 as i32
     }
 
-    /// Reassembles `int_parts` into an arbitrary-precision magnitude, for values
-    /// too wide for [`Self::magnitude`].
-    fn magnitude_big(&self) -> BigUint {
-        BigUint::new(self.int_parts.iter().map(|&part| part as u32).collect())
+    /// Number of little-endian 32-bit words needed to hold the magnitude, at
+    /// least one so that zero still has a word.
+    pub fn word_count(&self) -> usize {
```
saurabh500, line 2451: **Suggestion (doc only)** - the semantics of this count changed in a way that isn't captured in the doc comment.

`word_count()` returns the *minimal* number of words, where the old `int_parts.len()` reflected the wire width the value was decoded from. So `12345` decoded from a 4-word wire value now reports `1` instead of `4`.

I checked both JS consumers and this is safe: `mssql-js/lib/transformers/decimal.ts:15-17` and `mssql-js/lib/decode.ts:249-253` both accumulate `part * 2**(32*i)` over the reported count, so trailing zero words are behaviorally irrelevant. Worth stating explicitly so a future consumer doesn't assume a fixed four-word width:

```rust
/// Number of little-endian 32-bit words needed to hold the magnitude, at
/// least one so that zero still has a word.
///
/// This is the *minimal* count, not the width the value was decoded from:
/// trailing zero words are not reported. Consumers must accumulate over the
/// returned count rather than assuming four.
```

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -2308,113 +2310,150 @@ impl DecimalParts {
...
+                limbs[1] = (rest % LIMB) as u64;
+                limbs[2] = (rest / LIMB) as u64;
+                limb_count = 3;
+            }
+        }
+
+        // `scale` is wire-supplied and unvalidated. A `decimal` never exceeds
+        // 38, but the loop is bounded by the buffer rather than by trusting it,
+        // so a malformed value renders short instead of overrunning.
+        let scale = (self.scale as usize).min(DECIMAL_STR_LEN - 3);
```
saurabh500: **Nit** - this clamp is load-bearing for memory safety; consider making the derivation checkable.

The bound is exact with zero margin: at most `max(39, scale + 1) = 46` digits, plus one `.`, plus one `-`, equals 48 = `DECIMAL_STR_LEN`, so `pos` reaches exactly 0. Nicely, expressing the clamp in terms of `DECIMAL_STR_LEN` makes it self-adjusting for any `DECIMAL_STR_LEN >= 42` - but the `39` (u128's decimal width) is implicit and nothing enforces that floor.

A const assertion documents the derivation and fails the build rather than the fuzz test if either constant moves:

```rust
const _: () = assert!(
    DECIMAL_STR_LEN >= 42,
    "buffer must hold 39 digits of u128::MAX plus a decimal point and sign"
);
```

(`out_of_domain_scale_does_not_overrun_the_buffer` covers the runtime case well - this is just to pin the invariant at compile time.)

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -2308,113 +2310,150 @@ impl DecimalParts {
...
+    ///
+    /// Equivalent to [`ToString::to_string`], but it copies a known-length
+    /// `&str` instead of going through [`fmt::Formatter`], so it costs one
+    /// allocation and nothing else. Callers that can supply their own buffer
+    /// should use [`Self::format_into`] and pay nothing.
+    pub fn to_decimal_string(&self) -> String {
+        self.format_into(&mut [0u8; DECIMAL_STR_LEN]).to_owned()
+    }
+
+    fn to_f64(self) -> f64 {
```
saurabh500, line 2425: **Nit** - `to_f64` uses `self.scale` unclamped while `format_into` clamps to `DECIMAL_STR_LEN - 3`, so a malformed wire scale renders one way and converts another. Unreachable for a valid `decimal` (scale <= 38), so purely cosmetic, but a shared accessor keeps the two from drifting:

```rust
fn clamped_scale(&self) -> usize {
    (self.scale as usize).min(DECIMAL_STR_LEN - 3)
}
```

Then both `format_into` and `to_f64` read from it.

mssql-js/src/binary_row_writer.rs
```diff
@@ -222,9 +222,9 @@ impl RowWriter for BinaryRowWriter {
         self.row_data.push(val.is_positive as u8);
         self.row_data.push(val.scale);
         self.row_data.push(val.precision);
-        self.row_data.push(val.int_parts.len() as u8);
-        for part in &val.int_parts {
-            self.row_data.extend_from_slice(&part.to_le_bytes());
+        self.row_data.push(val.word_count() as u8);
+        for i in 0..val.word_count() {
```
saurabh500: **Nit** - `word_count()` is a `leading_zeros`/`div_ceil` computation, recomputed on every loop iteration here (and again in `write_numeric` at lines 237-238). Trivial cost, but this is a per-row path and hoisting reads better:

```rust
let words = val.word_count();
self.row_data.push(words as u8);
for i in 0..words {
    self.row_data.extend_from_slice(&val.word(i).to_le_bytes());
}
```

Separately: `write_decimal` and `write_numeric` are now byte-identical. Worth having one delegate to the other while you're in here.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -194,6 +192,9 @@ const MAX_PLP_SIZE: usize = i32::MAX as usize;
 /// words), so anything longer is malformed.
 const MAX_DECIMAL_INT_PARTS: usize = 4;
 
+/// Byte width of a `decimal`/`numeric` magnitude: four little-endian 32-bit words.
+const DECIMAL_MAGNITUDE_BYTES: usize = MAX_DECIMAL_INT_PARTS * 4;
```
saurabh500: **Nit** - `DECIMAL_MAGNITUDE_BYTES` is now defined here *and* privately in `mssql-tds/src/security/encryption/cell.rs:68` (backed by its own `DECIMAL_MAGNITUDE_WORDS`), both meaning 16.

Two private definitions of the same protocol constant can drift silently. Making this one `pub(crate)` and importing it in `cell.rs` leaves a single source of truth:

```rust
pub(crate) const DECIMAL_MAGNITUDE_BYTES: usize = MAX_DECIMAL_INT_PARTS * 4;
```

saurabh500 (commented):
## Summary

Replaces `DecimalParts { int_parts: Vec<i32> }` with a fixed `[u64; 2]` magnitude and adds an allocation-free `format_into`, making the type `Copy` and keeping `ColumnValues` 8-byte aligned so the row-fetch futures don't grow.

To verify: I hand-traced `format_into` against its buffer bound, checked every migrated `int_parts` -> `magnitude` test constant arithmetically, traced both JS consumers (`mssql-js/lib/transformers/decimal.ts`, `mssql-js/lib/decode.ts`) for the `word_count()` change, and confirmed `mssql-py-core` never touches `int_parts`. `cargo clippy --all-features --all-targets -- -D warnings` and `cargo nextest -p mssql-tds --lib` are clean locally on this head.

**Overall: solid, no blocking issues.** The correctness-critical parts hold up. `format_into`'s buffer bound is exactly tight but provably safe (46 digits + point + sign = 48 = `DECIMAL_STR_LEN`, guaranteed by the `scale.min(LEN - 3)` clamp), the limb decomposition provably can't break early at a limb boundary (when `limb_count > 1` the top limb is non-zero), and the new explicit length check in `cell.rs::read_decimal` is a genuine improvement over the `debug_assert!` it replaces. Findings below are one Suggestion worth a second look plus some nits.

### Two notes outside the diff

**PR description is stale.** The body says "Stacked on #311 - review only the top commit," but `git log origin/main..HEAD` shows this branch now contains only this commit; #311 is already in `main`. The "Known failure, pre-existing: `row_fetch_futures_stay_small` (4248 B vs <= 4224 B)" note is also stale - that test **passes** on this head locally:

```
PASS [0.033s] mssql-tds connection::tds_client::tests::row_fetch_futures_stay_small
```

Worth refreshing both before marking ready for review, since the PR-workflow guidance wants the description to match the diff and validation green on the latest commit.

**Public API break** (already documented in the body, flagging for visibility): `pub int_parts` removed, `magnitude()` changed from `Option<u128>` to `u128`, `magnitude_big()` removed. The `#[napi]` `index.d.ts` surface is unchanged, so the JS contract holds.

mssql-tds/src/datatypes/decoder.rs
saurabh500: Added a shared `clamped_scale` helper in 7191dba2 and now use it from both `format_into` and `to_f64`.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Fixed in 7191dba2. I added `try_from_words` and changed the JS FFI conversion to reject non-zero words beyond the fourth, while preserving compatibility with trailing zero padding; `from_words` also has the suggested debug assertion for internal callers.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Updated the documentation in 7191dba2 to state that `word_count()` returns the minimal significant width and omits trailing zero words.

> [!NOTE]

mssql-js/src/binary_row_writer.rs
saurabh500: Fixed in 7191dba2: `word_count()` is hoisted once, and `write_numeric` now delegates to the byte-identical `write_decimal` implementation.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Added the compile-time `DECIMAL_STR_LEN >= 42` assertion in 7191dba2, including the derivation in its failure message.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Consolidated the protocol width constant in 7191dba2 by exposing the decoder definition crate-wide and importing it from `cell.rs`.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Added the compile-time `DECIMAL_STR_LEN >= 42` assertion so changes to the buffer bound fail at build time.

> [!NOTE]

mssql-js/src/binary_row_writer.rs
saurabh500: Hoisted `word_count()` in the decimal path and made `write_numeric` delegate to `write_decimal`, removing the duplicate encoding logic.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Updated the docs to state that `word_count()` is minimal, omits trailing zero words, and must not be treated as a fixed decoded width.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Made `DECIMAL_MAGNITUDE_BYTES` crate-visible and imported it from `cell.rs`, leaving one protocol-width definition.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Added checked `try_from_words`, switched the NAPI input conversion to `TryFrom`, and now reject non-zero words beyond 128 bits. A JS-facing regression test covers the oversized fifth-word case.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Added a shared `clamped_scale` helper and now use it in both formatting and `to_f64` conversion.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -3307,61 +3198,273 @@ mod test {
...
+            assert_eq!(parts.to_string(), "0.005");
+        }
+
+        #[test]
+        fn out_of_domain_scale_does_not_overrun_the_buffer() {
+            // `scale` comes off the wire unvalidated. No `decimal` reaches 255,
+            // but the formatter must not index past its buffer if one claims to.
+            let parts = DecimalParts::new(false, 38, u8::MAX, u128::MAX);
+            let mut buf = [0u8; DECIMAL_STR_LEN];
+            assert!(parts.format_into(&mut buf).len() <= DECIMAL_STR_LEN);
```
Theekshna: This assertion cannot fail. `format_into` returns `&buf[pos..]`, a subslice of the caller's `[u8; DECIMAL_STR_LEN]`, so `len() <= DECIMAL_STR_LEN` holds by construction - it would still pass with the `- 3` clamp deleted outright. The test does not check the property its name states.

That matters here because the margin is zero: I measured this exact value (`new(false, 38, u8::MAX, u128::MAX)`) rendering **48 bytes into the 48-byte buffer**.

Suggest `assert_eq!(parts.format_into(&mut buf).len(), DECIMAL_STR_LEN);`, or pin the exact string, so a future change to `DECIMAL_STR_LEN` or the clamp fails here rather than on an array bounds check at runtime.

Theekshna (approved):
Reviewed at `26e42136` against `origin/main` (`6a122251`). No correctness or safety defect found.

Verified rather than assumed:

- Re-ran your reference oracle over `scale 39..=45` (outside the tested `0..=38`, clamp not yet active) across boundary magnitudes: 0 divergences. Over `46..=255`: no panic, and `format_into`/`Display`/`to_decimal_string` stay consistent.
- `word_count()` becoming *minimal* does not change the wire: `tds_value_serializer` sizes from precision and calls `word(i)`. JS `decode.ts` and `transformers/decimal.ts` both accumulate by the received count, and `NapiDecimalParts` is not re-exported from `lib/index.ts` - so the FFI-stability claim holds for the npm surface, not just `index.d.ts`.
- The removed `magnitude() -> Option<u128>` / `magnitude_big()` arms really are unreachable by construction. AE `normalize_decimal` still emits a fixed 16 bytes, so deterministic ciphertext stays byte-compatible.
- `read_decimal_data`'s `length - 1` cannot underflow (`length == 0` returns early).
- clippy clean on the workspace and `mssql-py-core`; `-p mssql-tds --lib` 1848 pass (7 pre-existing cert-fixture failures, unrelated).

One inline comment, plus one non-blocking suggestion below.

### No in-repo caller uses the fast path

Every consumer still goes through `Display`, the slowest of the three by your own table (2.0x vs 4.9x):

- `mssql-odbc/src/api/get_data.rs:1058` - `Ok(d.to_string())`. The fn returns `String`, so `to_decimal_string()` is a one-word change worth 2.0x -> 2.7x.
- `mssql-py-core/src/cursor.rs:791` - `let decimal_str = n.to_string();` feeding `call1((decimal_str.as_str(),))`. The success path only needs a `&str`, so `format_into` into a stack buffer removes the allocation. This is the per-row Python fetch path.

Both are outside the diff, so this may be deliberate follow-up scope given the driving consumer is the out-of-repo FDW. Worth a sentence in the PR either way, since the body tells callers to prefer `format_into` and then converts none of them.

Not verified: the benchmark numbers, the 176 B / ~4% async-state sizing claim, and any live-server round trip.

mssql-tds/src/datatypes/decoder.rs
saurabh500: Changed the test to assert the exact `DECIMAL_STR_LEN` boundary, so removing or weakening the clamp now fails the test. I also switched the ODBC and Python decimal consumers noted in the review to `to_decimal_string` and stack-backed `format_into`, respectively.

> [!NOTE]

saurabh500: Fixed in 4a9d9550 by asserting the exact `DECIMAL_STR_LEN` boundary rather than the tautological upper bound. The same commit also adopts the allocation-free formatter in the Python fetch path and the single-allocation wrapper in ODBC.

> [!NOTE]

Theekshna (approved):

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -826,25 +827,18 @@ impl GenericDecoder {
...
-            .chunks_exact(4)
-            .map(|chunk| {
-                let mut word = [0u8; 4];
-                word.copy_from_slice(chunk);
-                i32::from_le_bytes(word)
-            })
-            .collect();
-
-        Ok(Some(DecimalParts {
+        Ok(Some(DecimalParts::new(
```
shiwanigupta0809, line 847: **`scale` / `precision` are stored straight off the wire without validation**

`read_decimal_data` bounds the magnitude width, but nothing bounds `scale` or `precision` — they come out of `TypeInfoVariant::VarLenPrecisionScale` and land in `DecimalParts::new` untouched. I followed it back to `TypeInfo::var_len_precision_scale` and there is no range check there either, so `scale` can reach the formatter as any value in `0..=255`.

The new fast path handles that by absorbing it at render time:

```rust
fn clamped_scale(&self) -> usize {
    (self.scale as usize).min(DECIMAL_STR_LEN - 3)   // = min(scale, 45)
}
```

That prevents the buffer overrun, but it turns a protocol violation into a silently wrong value: magnitude `5` with `scale = 200` renders as `0.000…005` with **45** fractional digits instead of 200. No error, no log — and that string then feeds `decimal.Decimal(...)` in `mssql-py-core` and ODBC `SQL_C_CHAR`. The previous allocating path had no such ceiling, so this is a behavior change on malformed input, not just a new guard.

The bigger issue is placement. `DECIMAL_STR_LEN - 3` is the size of a text buffer, and it is currently acting as the definition of the numeric domain, four layers above the wire. Per the layering in `copilot-instructions.md` this belongs at the token boundary, once:

```rust
// alongside the existing number_of_int_parts check
if precision == 0 || precision > MAX_DECIMAL_PRECISION || scale > precision {
    return Err(Error::ProtocolError(format!(
        "Invalid decimal precision {precision} / scale {scale}"
    )));
}
```

With that in place `clamped_scale` can become a `debug_assert!` — actual defense in depth rather than a silent correction. It also pays for itself downstream: capping `scale` at 38 drops the worst-case output from exactly 48 bytes to 41, so the buffer gains real headroom (today `out_of_domain_scale_does_not_overrun_the_buffer` asserts `len == DECIMAL_STR_LEN`, i.e. it is provably always exactly full), and it removes the clamped-vs-unclamped mismatch between `to_f64` and `numeric_source` in `fetch_convert.rs`.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -2308,113 +2318,178 @@ impl DecimalParts {
...
-        }
-        self.is_positive == other.is_positive
-            && self.scale == other.scale
-            && self.precision == other.precision
+    /// Builds a value from little-endian 32-bit words, least significant first.
+    ///
+    /// Zero words past the fourth are ignored. Non-zero words cannot fit in the
+    /// representation and trigger a debug assertion; callers at trust boundaries
+    /// should use [`Self::try_from_words`] to reject them in every build.
+    pub fn from_words(is_positive: bool, precision: u8, scale: u8, words: &[i32]) -> Self {
```
shiwanigupta0809: **`pub fn from_words` truncates silently in release builds**

The oversize guard here is a `debug_assert!`, which is compiled out in release. Five non-zero words gives:

- debug / CI → panic with a clear message
- release → `.take(MAX_DECIMAL_INT_PARTS)` drops the fifth word, returns a different number, no signal

So the check is present only in the build where the bug is already obvious, and absent from the one customers run. It also reads as protection during review, which makes it easy to assume the case is covered.

The doc comment already says the right thing — *"callers at trust boundaries should use `try_from_words` to reject them in every build"* — but the API shape pushes the other way:

| | name | returns | oversize input |
|---|---|---|---|
| `from_words` | short | `Self` | truncates |
| `try_from_words` | longer | `Option<Self>` | rejects |

The ergonomic default is the lossy one, and it is `pub`. Documentation is not going to win against that.

Worth noting this PR removed the equivalent `debug_assert!` from `cell.rs::normalize_decimal` on the (correct) grounds that its input had become unreachable by construction — the same pattern then reappears here, in a public constructor where the input is reachable.

Suggest making `from_words` `pub(crate)` (all in-crate callers look fine) and exporting only `try_from_words`, so the easy path is the safe one.

saurabh500: Made `from_words` crate-private so external callers must use `try_from_words`, which rejects oversized non-zero limbs in all build modes. The ODBC test helper now uses the checked constructor as well.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Added shared decimal metadata validation requiring `1 <= precision <= 38` and `scale <= precision` before decoding both ordinary TDS and Always Encrypted decimal values. Regression tests cover zero/oversized precision and scale greater than precision.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Fixed in 3ae17b37. `from_words` is now crate-private, so external callers can only use the fallible `try_from_words`; the cross-crate ODBC test was updated accordingly.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Fixed in 3ae17b37. Decimal precision/scale are now validated at decode time before consuming the value, and the same shared validation is applied to the Always Encrypted denormalization path; invalid metadata has targeted coverage in both paths.

> [!NOTE]

shiwanigupta0809 (approved):

PR #315, opened 2026-08-17
Allow re-execute after SQLFetch reaches SQL_NO_DATA

Theekshna:
## Description

Re-running a statement (`SQLExecute` / `SQLExecDirect`) after its result set had been fetched to `SQL_NO_DATA` returned SQLSTATE `24000`, because the cursor stayed open until an explicit `SQLCloseCursor`. msodbcsql allows the re-execute once the whole batch is drained. This restores that parity.

- Add `STMT_STATE_CURSOR_EXHAUSTED`, set on `SQLFetch` end-of-batch only when the terminating DONE carries no more-results flag (gated on `TdsClient::has_open_batch()`), so the driver can tell "cursor open, still busy" from "cursor open, batch fully consumed".
- Before staging a re-execute, `SQLExecute` and `SQLExecDirect` implicitly close an open-and-exhausted cursor via a shared `close_open_cursor` (also used by `SQLCloseCursor` / `SQLFreeStmt(SQL_CLOSE)`); a cursor with un-read rows or a still-pending result set is left open, so a re-execute with results still on the wire correctly stays `24000` (mirrors msodbcsql `CheckBusy` / `FIsBusyReadingData`).

Tests: unit (token-mock) for the implicit-close decision + drain and the `cursor_is_exhausted` predicate; e2e for both entry points — single result set re-executes without an intervening close, pending result set returns `24000`. Rebased on latest `main`.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47348

[AB#47348](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47348)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (scoped to `mssql-odbc`, the only changed crate; 551 unit tests green — full workspace coverage runs in PR validation)
- [x] Public API changes are documented (n/a — no public API surface change; the FFI entry points behave more permissively but signatures are unchanged)

### DIFF
--- a/mssql-odbc/src/api/close_cursor.rs
+++ b/mssql-odbc/src/api/close_cursor.rs
@@ -14,7 +14,9 @@ use crate::api::odbc_types::{
     SQL_ERROR, SQL_INVALID_HANDLE, SQL_SUCCESS, SQL_SUCCESS_WITH_INFO, SqlHandle, SqlReturn,
 };
 use crate::error::free_errors;
-use crate::handles::stmt::{STMT_STATE_CURSOR_OPEN, STMT_STATE_EXEC_CONTEXT};
+use crate::handles::stmt::{
+    STMT_STATE_CURSOR_EXHAUSTED, STMT_STATE_CURSOR_OPEN, STMT_STATE_EXEC_CONTEXT,
+};
 use crate::handles::{HandleType, StmtHandle, handle_from_raw};
 
 /// Closes the cursor on `statement_handle` and discards any pending rows.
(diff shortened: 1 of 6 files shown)


Theekshna:
Closing — the premise is invalid.

Investigation shows **msodbcsql returns SQLSTATE `24000`** for a re-execute (`SQLExecute` / `SQLExecDirect`) after `SQLFetch` reaches `SQL_NO_DATA` **without an explicit close**. It does *not* implicitly close the cursor, so the original mssql-odbc behavior already matched msodbcsql and there is no bug to fix.

Evidence:
1. **Empirical** — the e2e msodbcsql18 (18.6.2.1) parity leg returned `24000` for exactly these scenarios (both prepared and direct).
2. **Source** — msodbcsql gates re-execute on `STMT_ST_EXECSTARTED | STMT_ST_CURS_OPEN` (`sqlccmd.cpp`), a statement-level flag cleared only by an explicit `SQLCloseCursor` / `SQLFreeStmt(SQL_CLOSE)` / `SQLMoreResults`→`SQL_NO_DATA`.
3. **Tests** — msodbcsql's own `CurInval` / `CurNativ` suites guard the `24000` "invalid cursor state" contract.
4. **Docs** — Microsoft's official C++ ODBC sample explicitly calls `SQLFreeStmt(hStmt, SQL_CLOSE)` between executes.

This change would make the driver *diverge* from msodbcsql, which is why the e2e parity leg failed. AB#47348 is a non-bug; no driver change is needed. The existing `ExecDirectLiveTest.EmptyResultSet` already guards the correct behavior.

Issue #281, opened 2026-08-13
Oversized decimal in SQLGetData reports SQLSTATE 07006 instead of 22003


David-Engel:
### Describe the bug

An oversized `decimal`/`numeric` value converted through `SQLGetData` into a numeric C type reports SQLSTATE `07006` ("restricted data type attribute violation") instead of `22003` ("numeric value out of range").

`07006` means the requested C type is not a legal target for this SQL type. But `decimal` → `SQL_C_SLONG` *is* a legal pairing — it is this particular value that will not fit. An application branching on SQLSTATE to distinguish "I asked for the wrong type" from "this row had a value too big for the buffer I gave you" gets the wrong answer, and the first is not retryable per-row while the second is.

The root cause is that `numeric_source` in `mssql-odbc/src/api/fetch_convert.rs` returns `Option`, so two distinct conditions collapse into the same `None`:

```rust
ColumnValues::Decimal(d) | ColumnValues::Numeric(d) => {
    let m = i128::try_from(d.magnitude()?).ok()?;   // <- both failures return None
    ...
}
```

and the sole caller maps every `None` to `07006`:

```rust
numeric_source(value).ok_or(ConvError::Restricted)   // fetch_convert.rs:255
```

`None` legitimately means `07006` for a binary or GUID column, which has no numeric interpretation at all. For a `decimal` column it means the value overflowed. Swapping the error at the `Decimal` arm alone is not enough — `numeric_source` needs to distinguish the two, e.g. by returning `Result<NumericSource, ConvError>` or by having the decimal arm signal out-of-range separately.

The same function already returns `22003` for the analogous character-source case (`fetch_convert.rs:250`), so fixing this also makes the function self-consistent.

**This is reachable from the wire, not only through FFI.** Two failure modes reach the `?`:

1. `d.magnitude()` returns `None` for more than four significant words. After #237 the wire decoder caps the count at 4, so this one is FFI-only.
2. **`i128::try_from` fails for a four-word magnitude above `i128::MAX`.** The wire decoder validates the word *count*, not that the magnitude fits the column's declared precision, so a 17-byte payload — a legal length — carrying a magnitude in `(i128::MAX, u128::MAX]` is accepted by the decoder and fails here. A valid `decimal(38, s)` never exceeds `10^38 - 1 ≈ 2^126.2`, so this is a malformed or hostile payload, but it is one a server can send.

[...]

PR #318, opened 2026-08-17
Add missing crates to the issue template "Affected crate" dropdowns

### LINKED ISSUE #317: Issue templates: "Affected crate" dropdown is missing mssql-odbc and two other workspace crates
### Describe the bug

The `Affected crate` dropdown in both issue templates lists five crates and has not kept up with the workspace. Three crates are missing from both:

| Crate | In `Cargo.toml` | In templates |
|---|---|---|
| `mssql-tds` | member | yes |
| `mssql-js` | member | yes |
| `mssql-tds-cli` | member | yes |
| `mssql-mock-tds` | member | yes |
| `mssql-py-core` | `exclude` | yes |
| **`mssql-odbc`** | member | **no** |
| **`mssql-tds-bench`** | member | **no** |
| **`mssql-mock-tds-py`** | `exclude` | **no** |

Affected files:

- `.github/ISSUE_TEMPLATE/bug_report.yml` (lines 56-67)
- `.github/ISSUE_TEMPLATE/feature_request.yml` (lines 27-38)

This is drift rather than a deliberate omission. Both templates arrived in `9be35c46` ("Add GitHub open-source community files and clean up README") and `mssql-odbc` landed later in #146; the dropdown was never revisited. `mssql-py-core` being listed shows that workspace `exclude` was not the selection criterion, so `mssql-mock-tds-py` and `mssql-tds-bench` are not omitted on that basis either.

`mssql-odbc` is the one that matters in practice. It is an actively developed driver crate and the subject of a good share of recent issues, so the dropdown is wrong for a meaningful fraction of reports.

There is a second, smaller defect worth fixing in the same pass: **`bug_report.yml`'s dropdown is `required: true` but has no "Not sure / Multiple" option.** `feature_request.yml` has one and is `required: false`. A bug reporter who does not know which crate is at fault — a common and legitimate state, especially across an FFI boundary where the symptom surfaces in `mssql-js` but the defect is in `mssql-tds` — cannot submit without picking a crate they may be wrong about. That produces mislabeled issues, which is worse than an honest "not sure".

### Steps to reproduce

1. Go to https://github.com/microsoft/mssql-rs/issues/new/choose and pick "Bug Report".
2. Open the "Affected crate" dropdown.
3. Observe that `mssql-odbc` is not an option, and that the field is required with no "Not sure / Multiple" escape.

### Expected behavior

The dropdown offers every crate in the repository, including `mssql-odbc`, and a reporter who is unsure can still file.

### Actual behavior

Reporters filing against `mssql-odbc` must select an unrelated crate or leave the form. Observed concretely in #281, which selected nothing accurate and instead noted the gap in prose:

> Note: `mssql-odbc` is not in the "Affected crate" dropdown in `.github/ISSUE_TEMPLATE/bug_report.yml`, which still lists only `mssql-tds`, `mssql-js`, `mssql-tds-cli`, `mssql-mock-tds`, and `mssql-py-core`.

### Version

Repository metadata, not a crate. Present as of `a08461a6`.

### Affected crate

Not applicable — this is `.github/ISSUE_TEMPLATE/`, which is exactly the gap being reported.

### Additional context

Split out of #281, where it was noted as an aside to an unrelated SQLSTATE bug.

Open question for whoever picks this up: whether `mssql-tds-bench` and `mssql-mock-tds-py` belong in a user-facing dropdown at all. Both are internal (a benchmark harness and a test double), and an argument for leaving them out is reasonable — but `mssql-mock-tds` is already listed, so the current list is not drawn on that line either. The list should be deliberate in one direction or the other rather than inheriting whatever the workspace looked like in `9be35c46`.

David-Engel:
## Description

The `Affected crate` dropdown in both issue templates listed five crates and had not tracked the workspace. Three of the eight crates in the repo were missing from both templates:

| Crate | `Cargo.toml` | Before | After |
|---|---|---|---|
| `mssql-tds` | member | yes | yes |
| `mssql-odbc` | member | **no** | yes |
| `mssql-js` | member | yes | yes |
| `mssql-py-core` | `exclude` | yes | yes |
| `mssql-tds-cli` | member | yes | yes |
| `mssql-mock-tds` | member | yes | yes |
| `mssql-mock-tds-py` | `exclude` | **no** | yes |
| `mssql-tds-bench` | member | **no** | yes |

This is drift, not a deliberate list. Both templates arrived in `9be35c46` ("Add GitHub open-source community files and clean up README"); `mssql-odbc` landed later in #146 and the dropdown was never revisited. `mssql-py-core` being listed despite being in `exclude` shows workspace membership was not the selection criterion, so the other two excluded/internal crates were not omitted on that basis either.

`mssql-odbc` is the one that matters in practice — it is actively developed and the subject of a good share of recent issues, so the dropdown was wrong for a meaningful fraction of reports. #281 hit this and had to note the gap in prose instead of selecting anything accurate.

### Two judgement calls

**1. `bug_report.yml` gains `Not applicable / Multiple`.** `feature_request.yml` already had an option here; `bug_report.yml` did not, *and* its dropdown is `required: true`, so a reporter with no applicable crate could not submit without naming one falsely. That happens whenever an issue has no owning crate — pipeline maintenance, repo-wide cleanup, or #317 itself, where I had to type "Not applicable" into the crate field by hand. It also absorbs the cross-crate case, where a symptom surfaces in `mssql-js` but the defect is in `mssql-tds`.

Per review, the wording is `Not applicable / Multiple` rather than the `Not sure / Multiple` this PR first proposed: "not sure" framed the option as reporter uncertainty, when the more common case is that no single crate applies at all. Applied to `feature_request.yml` too, which carried the older wording, so the two templates do not diverge again. `required: true` is unchanged; the escape hatch is enough.

**2. The options are reordered.** Grouped core → drivers/bindings → tools → test infra, applied identically to both files so they no longer differ in ordering. Purely cosmetic and the reason the diff is larger than three added lines. Happy to restack as pure insertions if you prefer minimal churn.

Whether `mssql-tds-bench` and `mssql-mock-tds-py` belong in a user-facing dropdown at all is a fair question — both are internal (a benchmark harness and a test double). I included them because `mssql-mock-tds` was already listed, so excluding them would leave the list drawn on no consistent line. If the preference is to hide internal crates, the consistent version of that is dropping all three mock/bench entries, which I am glad to do instead.

### Verification

Both files parse as YAML and their option lists are now byte-identical:

```
bug_report.yml      ['mssql-tds', 'mssql-odbc', 'mssql-js', 'mssql-py-core', 'mssql-tds-cli',
                     'mssql-mock-tds', 'mssql-mock-tds-py', 'mssql-tds-bench', 'Not applicable / Multiple']
feature_request.yml [ ...identical... ]
```

Only `required` differs between them — `true` for bugs, `false` for features — which was already the case before this PR.

Every entry cross-checked against the `name =` field of each crate's `Cargo.toml`, not just directory names — all eight match, and all eight are present.

## Related Issues

Fixes #317

Split out of #281, where it was an aside on an unrelated SQLSTATE bug.

## Checklist

- [x] `cargo bfmt` passes — not applicable, no Rust changed; unaffected by this diff
- [x] `cargo bclippy` passes — not applicable, no Rust changed; unaffected by this diff
- [x] `cargo btest` passes — not applicable, no Rust changed; unaffected by this diff
- [x] New/changed functionality has tests — no test hook exists for issue-form YAML; validated by parsing both files and diffing the option lists against `Cargo.toml`, shown above
- [x] Public API changes are documented — none; repository metadata only

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -66,7 +66,7 @@ body:
         - mssql-mock-tds
         - mssql-mock-tds-py
         - mssql-tds-bench
-        - Not sure / Multiple
+        - Not applicable / Multiple
     validations:
       required: true
 
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -59,10 +59,14 @@ body:
       label: Affected crate
       options:
         - mssql-tds
+        - mssql-odbc
         - mssql-js
+        - mssql-py-core
         - mssql-tds-cli
         - mssql-mock-tds
-        - mssql-py-core
+        - mssql-mock-tds-py
+        - mssql-tds-bench
+        - Not sure / Multiple
     validations:
       required: true
 
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -29,10 +29,13 @@ body:
       label: Affected crate
       options:
         - mssql-tds
+        - mssql-odbc
         - mssql-js
+        - mssql-py-core
         - mssql-tds-cli
         - mssql-mock-tds
-        - mssql-py-core
+        - mssql-mock-tds-py
+        - mssql-tds-bench
         - Not sure / Multiple
     validations:
       required: false
--- a/.github/ISSUE_TEMPLATE/feature_request.yml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yml
@@ -36,7 +36,7 @@ body:
         - mssql-mock-tds
         - mssql-mock-tds-py
         - mssql-tds-bench
-        - Not sure / Multiple
+        - Not applicable / Multiple
     validations:
       required: false
 


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -59,10 +59,14 @@ body:
...
         - mssql-tds
+        - mssql-odbc
         - mssql-js
+        - mssql-py-core
         - mssql-tds-cli
         - mssql-mock-tds
-        - mssql-py-core
+        - mssql-mock-tds-py
+        - mssql-tds-bench
+        - Not sure / Multiple
```
saurabh500: Can we remove Not Sure and just say multiple and not applicable. 

Not Applicable for issues which may be general pipeline maintenance issues, or things which are general goodness. like the issue for this PR.

David-Engel: Done in 7a4ff5de — the option is now `Not applicable / Multiple`, with "Not sure" dropped.

Your framing is better than mine. I had been thinking about reporter *uncertainty* — someone who cannot tell whether a symptom belongs to `mssql-js` or to `mssql-tds` underneath it. That case is real, but it is the less common one, and "Not sure" invites a shrug where the honest answer is often that no single crate owns the issue at all. Pipeline maintenance, repo-wide cleanup, and the issue behind this PR are all in that category — I wrote "Not applicable" by hand in #317's crate field precisely because nothing in the list fit, which is the gap you are pointing at.

"Not applicable / Multiple" also covers the uncertainty case well enough in practice, since a reporter who cannot narrow it down is usually looking at something that spans crates anyway.

Applied to `feature_request.yml` too. It already had the older `Not sure / Multiple` wording, so leaving it would have reintroduced the divergence between the two templates that the rest of this PR removes. Both dropdowns are now byte-identical; only `required` differs (`true` for bugs, `false` for features), which was already the case.

> [!NOTE]

saurabh500 (approved):

PR #320, opened 2026-08-17
docs(odbc): split P3 into SQLBindCol (AB#47359) and block SQLFetchScroll (AB#46580)

David-Engel:
Splits the P3 phase of the typed & columnar fetch plan so the two halves can be worked in parallel.

`SQLBindCol` moves to task [47359](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47359) (Theekshna); [46580](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46580) keeps block `SQLFetchScroll` and the rowset fill. Both work items have been rescoped to match, and 47359 was reparented from 46373 (Parameter completeness) to 46375 (Typed & columnar fetch) — it had been created as a verbatim copy of 46580, so the two overlapped completely.

The halves meet at exactly one seam, the per-column binding table, so the sequencing note is the important part of this change: **46580 defines the shared `ColumnBinding` type and the `StmtState` field, and lands them early.** That way the consumer (the fill loop) shapes the structure rather than the other way round, and 47359 is not blocked waiting for the rest of 46580.

Also recorded:

- `SQLFreeStmt(SQL_UNBIND)` belongs with the binding half and is on the critical path — mssql-python calls it before every fetch, so the columnar path breaks without it. It is currently a `TODO` in `api/exports.rs`.
- The rowset attributes (`SQL_ATTR_ROW_ARRAY_SIZE`, `SQL_ATTR_ROWS_FETCHED_PTR`, `SQL_ATTR_ROW_STATUS_PTR`) and the `SQL_ROW_*` constants already landed in P0, so P3b is wiring rather than invention.
- P4 is now only partly outstanding, since `SQLColAttributeW` was exported and advertised in P2.
- The `SQL_API_SQLBINDCOL` / `SQLGetFunctions` trap: the Windows Driver Manager returns `IM001` for an entry point it has not been told about, even when the export exists. P2 hit this with `SQLColAttribute`.

Docs only — no code changes.

[AB#46375](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46375)

### DIFF
--- a/mssql-odbc/docs/typed-columnar-fetch-plan.md
+++ b/mssql-odbc/docs/typed-columnar-fetch-plan.md
@@ -129,12 +129,22 @@ Reading a `sql_variant` column takes three things, not one, and mssql-python nee
 
 The underlying type is a property of the **value**, not the column — a variant column can hold a different type in every row — so it is carried up from the decoder rather than derived from metadata: `RowWriter` gained a defaulted `write_variant_base_type`, `CursorColumn::Value` carries the base type alongside the value, and `StmtState` clears it with the rest of the row-stream state. `ColumnValues` is deliberately untouched, which is what keeps this change out of the Python and Node bindings.
 
-### P3 — SQLBindCol + block SQLFetchScroll — Task [46580](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46580)
+### P3a — SQLBindCol — Task [47359](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47359)
 
-- `SQLBindCol`: store per-column binding (col, target C type, buffer ptr, buffer len, indicator ptr); support unbind (null ptr) and `SQLFreeStmt(SQL_UNBIND)`.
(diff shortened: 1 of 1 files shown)


saurabh500 (approved):
I hope the approach works with the copilot and it actually blocks ! :)

David-Engel:
Closing: the split this documents no longer exists.

`SQLBindCol` has come back to me rather than being worked in parallel, so P3 is one deliverable again — which means the description already on main, *"P3 — SQLBindCol + block SQLFetchScroll — Task 46580"*, is accurate as it stands. Merging this and then amending it would mean editing the same doc twice to land where we already are.

The parts of this PR worth keeping are carried into [#322](https://github.com/microsoft/mssql-rs/pull/322) instead, which is now the single P3 PR:

- P4 (46581) is only partly outstanding, since `SQLColAttributeW` was exported in P2.
- The `SQLGetFunctions` / `IM001` trap: the Windows DM will not dispatch an entry point it has not been told about, even when the export exists. That cost P2 a CI cycle.
- `SQLFreeStmt(SQL_UNBIND)` is on the critical path — mssql-python calls it before every fetch.
- The PLP-in-a-bound-column gap, now tracked as [47361](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47361).

AB#47359 stays as a task for tracking and moves to me; it ships with 46580 through #322.

PR #321, opened 2026-08-17
Route pipeline jobs to the split public/trusted agent pools

David-Engel:
## Summary

Repoints every pipeline job at the new split agent pools, so that unreviewed pull request code and trusted CI no longer share agents.

Until now both ran on `RUST-1ES-POOL-WUS3` / `RUST-1ES-POOL-ARM-WUS3`, which sit in a vnet with private endpoints to the ACR, SQL server and Key Vault. A GitHub PR from a fork therefore executed on an agent that could reach internal resources. The new pools are split by trust boundary and project scope: the `RUST-PUBLIC-*` pools are visible only to the `public` project and have no private endpoints, while the `RUST-*` pools are visible only to `mssql-rs`.

The pool infrastructure (two vnets, images, six pools, queue authorization) is already deployed. This change is the pipeline half.

### How the routing works

`validation-stages.yml` now takes `poolName`, `armPoolName` and `utilPoolName` instead of naming pools inline. The entry file `validation-pipeline.yml` is consumed by two definitions — 2267 in the `public` project for GitHub PRs, and 2204 in `mssql-rs` for the ADO mirror — so it selects pools from `System.TeamProject`. Pool names must resolve at compile time, so this is a template expression rather than a runtime variable.

The parameter defaults are the **trusted** pools, which makes a mistake fail closed: the `public` project cannot see those pools, so a misrouted run fails with "pool not found" instead of putting unreviewed code on an agent with private endpoint access.

Pipelines that only ever run in `mssql-rs` (container image sync, OneBranch publish, benchmark, branch sync) name the trusted pools outright rather than taking a parameter. The same applies to templates used exclusively by CI-only stages (`test-matrix-*`, `private-link-smoke`), since those stages are gated off for PR runs and can never need a public pool.

### Utility pools

The `Evaluate` gate job and `Merge_Coverage` move to small dedicated pools. Both are short — the gate job does about 0.4 minutes of real work — but on real PR runs it waited a median of **4.4 minutes** for an agent (samples: 0.6, 4.0, 4.1, 4.4, 4.4, 6.5, 8.6). The Build stage depends on it, so that wait is dead time on the critical path of every PR. The cause is contention rather than provisioning speed: the pool it ran on already uses automatic provisioning, but a 0.4-minute job was queued behind 40-minute builds.

The SQL host job deliberately stays on the full-size pool; it serves real query load from the ARM tests.

### Image rename

`RUST-1ES-UBUSLIM` → `RUST-UBUSLIM` (36 references) and `RUST-Win22-Sql25-1P` → `RUST-W22-SQL25` (6). The ARM image demands (`RUST-UBUNTU-ARM64`, `RUST-WINSRV-ARM`) are unchanged — those images are reused as-is.

The private link smoke test moves to the new trusted vnet and its dedicated ACI subnet (`snet-aci-01`, previously `default2`).

## Linked work item

[AB#46593](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46593)

## Testing

Not yet exercised by a pipeline run — that is the point of opening this as a draft. What has been verified:

- All 35 pipeline YAML files parse.
- No stale references remain: `RUST-1ES-POOL*`, `RUST-1ES-UBUSLIM` and `RUST-Win22-Sql25*` all return zero matches under `.pipeline/`.
- Every job's pool assignment was checked individually against its architecture, and the two utility-pool jobs against their job names.
- On the infrastructure side, all six pools report `Succeeded`, both images published version 1.0.0, and queue authorization was confirmed to be least-privilege: no blanket `allPipelines` authorization on any queue, public queues authorized for definition 2267 only, and an org-wide check confirming only `public` can see the public queues and only `mssql-rs` the trusted ones.

Suggested validation order:

1. Run definition 2204 (`mssql-rs` project) to exercise the trusted path first.
2. Then let a GitHub PR exercise the public path.
3. Compare the `Evaluate` job's queue-to-start wait against the ~4.4 minute baseline.

Treat the first public run's wall-clock time as unrepresentative: the pools start cold and the Windows image is at its first version.

No source or API changes; this is pipeline configuration only.

Operationally it does repoint all CI and PR jobs onto new pools at once. Two things must be true before merge, and both currently are:

- The six pools exist and are authorized for the right pipeline definitions.
- The renamed images exist, since `imageOverride` demands are matched by name.

The old pools are intentionally left in place, so reverting this commit restores the previous routing. They should only be retired after a soak period, along with the old vnet and its private endpoints.

### DIFF
--- a/.pipeline/benchmark-pipeline.yml
+++ b/.pipeline/benchmark-pipeline.yml
@@ -25,9 +25,9 @@ stages:
         displayName: 'Criterion Benchmark'
         timeoutInMinutes: 30
         pool:
-          name: RUST-1ES-POOL-WUS3
+          name: RUST-X64-WUS3
           demands:
-            - imageOverride -equals RUST-1ES-UBUSLIM
+            - imageOverride -equals RUST-UBUSLIM
         variables:
           ob_outputDirectory: '$(Build.ArtifactStagingDirectory)'
 
--- a/.pipeline/validation-pipeline.yml
+++ b/.pipeline/validation-pipeline.yml
@@ -51,6 +51,20 @@ stages:
     RunLongHaul: ${{ parameters.RunLongHaul }}
     sqlInstanceMode: ${{ parameters.sqlInstanceMode }}
     sqlImageTag: ${{ parameters.sqlImageTag }}
+    # This file is shared by TWO pipeline definitions: 2267 in the `public`
+    # project (GitHub PRs, i.e. unreviewed code) and 2204 in `mssql-rs` (the ADO
+    # mirror). Pool names must resolve at compile time, so this is a template
+    # rather than a runtime variable. The pools are project-scoped, so a wrong
+    # branch here fails with "pool not found" rather than putting unreviewed
+    # code on a trusted agent.
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      poolName: RUST-PUBLIC-X64-WUS3
+      armPoolName: RUST-PUBLIC-ARM64-WUS3
+      utilPoolName: RUST-PUBLIC-UTIL-WUS3
+    ${{ else }}:
+      poolName: RUST-X64-WUS3
+      armPoolName: RUST-ARM64-WUS3
+      utilPoolName: RUST-UTIL-WUS3
     # PR validation for internal ADO and GitHub PRs. GitHub PRs run in the public
     # project, which lacks 'Magnitude Test', so exclude the infra stages here.
     includeInfraStages: false
--- a/.pipeline/OneBranch/PublishFeeds-Sandbox.yml
+++ b/.pipeline/OneBranch/PublishFeeds-Sandbox.yml
@@ -110,9 +110,9 @@ extends:
         pool:
           type: windows
           isCustom: true
-          name: RUST-1ES-POOL-WUS3
+          name: RUST-X64-WUS3
           demands:
-            - imageOverride -equals RUST-Win22-Sql25-1P
+            - imageOverride -equals RUST-W22-SQL25
         variables:
           ob_outputDirectory: '$(Build.ArtifactStagingDirectory)'
         steps:
@@ -144,9 +144,9 @@ extends:
         pool:
           type: linux
           isCustom: true
-          name: RUST-1ES-POOL-WUS3
+          name: RUST-X64-WUS3
           demands:
-            - imageOverride -equals RUST-1ES-UBUSLIM
+            - imageOverride -equals RUST-UBUSLIM
         variables:
           ob_outputDirectory: '$(Build.ArtifactStagingDirectory)'
           ob_artifactSuffix: '_mock_linux_x64'
@@ -195,7 +195,7 @@ extends:
         pool:
           type: linux
           isCustom: true
-          name: RUST-1ES-POOL-ARM-WUS3
+          name: RUST-ARM64-WUS3
           demands:
             - imageOverride -equals RUST-UBUNTU-ARM64
         variables:
@@ -248,9 +248,9 @@ extends:
         pool:
           type: linux
           isCustom: true
-          name: RUST-1ES-POOL-WUS3
+          name: RUST-X64-WUS3
           demands:
-            - imageOverride -equals RUST-1ES-UBUSLIM
+            - imageOverride -equals RUST-UBUSLIM
         variables:
           ob_outputDirectory: '$(Build.ArtifactStagingDirectory)'
           ob_artifactSuffix: '_mock_linux_musl_x64'
@@ -301,7 +301,7 @@ extends:
         pool:
           type: linux
           isCustom: true
-          name: RUST-1ES-POOL-ARM-WUS3
+          name: RUST-ARM64-WUS3
           demands:
             - imageOverride -equals RUST-UBUNTU-ARM64
         variables:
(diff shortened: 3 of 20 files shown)


Vahid-b (commented):
## Review Summary

Reviewed this pool-split migration against its stated goal: unreviewed GitHub PR code should never run on an agent with private-endpoint access to internal infrastructure (ACR, SQL Server, Key Vault).

**Blocking finding (outside this diff, but in scope for the PR's own goal):**

`.pipeline/OneBranch/NonOfficialPythonWheelsPublish.yml` declares both `trigger: [main]` and `pr: [main]` (line 16), with its own header comment stating *"CI runs on merge to main and on PRs into main."* It extends `.pipeline/OneBranch/stages.yml`, which **this PR edits** to point OneBranch jobs at the new trusted pools (`RUST-X64-WUS3`/`RUST-ARM64-WUS3`) unconditionally, with no `System.TeamProject` split. If this pipeline definition is registered to build GitHub PRs, a PR from any contributor would run directly on a trusted, private-endpoint-connected agent -- the exact class of exposure this PR sets out to close, just missed for this entry point.

I couldn't find this pipeline producing a status check on any of the last 100 PRs (open+closed) in this repo, which may mean it isn't currently registered to auto-build GitHub PRs, or has PR-building disabled at the ADO definition level outside this YAML. Worth confirming either way before merge: if it's live, it needs the same `System.TeamProject` split as `validation-pipeline.yml`; if it's provably not live, a comment near the trigger block would keep this from becoming a silent gap later.

Two more notes inline below. Nice catch on the compile-time-vs-runtime pool resolution gotcha in the second commit -- that's a real and easy-to-miss Azure Pipelines failure mode.

.pipeline/benchmark-pipeline.yml
```diff
@@ -25,9 +25,9 @@ stages:
         displayName: 'Criterion Benchmark'
         timeoutInMinutes: 30
         pool:
-          name: RUST-1ES-POOL-WUS3
+          name: RUST-X64-WUS3
```
Vahid-b, line 28: **Important:** This file declares `pr: branches: include: [main]` (line 9) and its steps explicitly build `$(PR_SHA)` / checkout `System.PullRequest.TargetBranch` -- it's a per-PR pipeline. This PR renames its pool to `RUST-X64-WUS3`, which is the **trusted** pool family, with no `System.TeamProject` split like `validation-pipeline.yml` gets below. If this pipeline is registered to build GitHub PRs, that's unreviewed code running on a trusted, private-endpoint-connected agent.

I found no "Benchmark"/"Criterion" status check on the last 100 PRs in this repo, which may mean it isn't currently wired to auto-run on GitHub PRs -- worth confirming. If it is (or could be) live, suggest threading `poolName` / `armPoolName` through the same way `validation-pipeline.yml` does.

.pipeline/validation-pipeline.yml
```diff
@@ -51,6 +51,20 @@ stages:
...
+    # project (GitHub PRs, i.e. unreviewed code) and 2204 in `mssql-rs` (the ADO
+    # mirror). Pool names must resolve at compile time, so this is a template
+    # rather than a runtime variable. The pools are project-scoped, so a wrong
+    # branch here fails with "pool not found" rather than putting unreviewed
+    # code on a trusted agent.
+    ${{ if eq(variables['System.TeamProject'], 'public') }}:
+      poolName: RUST-PUBLIC-X64-WUS3
+      armPoolName: RUST-PUBLIC-ARM64-WUS3
+      utilPoolName: RUST-PUBLIC-UTIL-WUS3
+    ${{ else }}:
```
Vahid-b, line 64: **Suggestion:** This `else` branch (the trusted-project path, definition 2204) is the one the PR's own "Suggested validation order" asks to be run *first*, since GitHub-triggered checks only ever exercise the `if` branch above. I don't see a linked build or comment confirming 2204 was actually run before merge -- only one `/azp run` comment with no indication which definition it triggered. Since this conditional is the load-bearing security logic of the whole PR and can't be partially tested, it'd be good to have that confirmation recorded here (or a linked build result) before merging.

.pipeline/benchmark-pipeline.yml
David-Engel: `benchmark-pipeline.yml` is an internal ADO pipeline, so it _only_ has access to the trusted pools. The pipeline hasn't been run since April and is obsoleted by the new `perf-baseline-<platform>-pipeline.yml` pipelines (also ADO). It's really a candidate for deletion. But that's out of scope for this PR.

.pipeline/validation-pipeline.yml
David-Engel: I validated the main ADO pool manually against this PR branch. It succeeded. I'll send you the link offline, if you really want to see it. Any other ADO pipelines we will fix manually, if they have issues.

David-Engel:
> **Blocking finding (outside this diff, but in scope for the PR's own goal):**
> 
> `.pipeline/OneBranch/NonOfficialPythonWheelsPublish.yml` declares both `trigger: [main]` and `pr: [main]` (line 16), with its own header comment stating _"CI runs on merge to main and on PRs into main."_ It extends `.pipeline/OneBranch/stages.yml`, which **this PR edits** to point OneBranch jobs at the new trusted pools (`RUST-X64-WUS3`/`RUST-ARM64-WUS3`) unconditionally, with no `System.TeamProject` split. If this pipeline definition is registered to build GitHub PRs, a PR from any contributor would run directly on a trusted, private-endpoint-connected agent -- the exact class of exposure this PR sets out to close, just missed for this entry point.
> 
> I couldn't find this pipeline producing a status check on any of the last 100 PRs (open+closed) in this repo, which may mean it isn't currently registered to auto-build GitHub PRs, or has PR-building disabled at the ADO definition level outside this YAML. Worth confirming either way before merge: if it's live, it needs the same `System.TeamProject` split as `validation-pipeline.yml`; if it's provably not live, a comment near the trigger block would keep this from becoming a silent gap later.

I can confirm that the YAML PR trigger for that pipeline is overridden in the ADO pipeline definition:
image

Vahid-b (approved):

PR #322, opened 2026-08-17
odbc: SQLBindCol + block SQLFetchScroll — the columnar fetch path (P3) [AB#46580]

David-Engel:
Implements P3 of the typed & columnar fetch plan: the path `mssql-python` uses for `fetchmany` / `fetchall`. `SQLBindCol` records where each column's value should land, and `SQLFetchScroll(SQL_FETCH_NEXT)` pulls up to `SQL_ATTR_ROW_ARRAY_SIZE` rows in one call and fills those buffers, reporting the count through `*rows_fetched_ptr` and each row's outcome through the row status array.

Covers **[AB#46580](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46580)** and **[AB#47359](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47359)**. 47359 was briefly split out so it could be worked in parallel; that is no longer the plan, and folding it back in fixes the weakness the split created — the fill loop could not be exercised end to end, because the e2e suite had no way to bind a column. ([#320](https://github.com/microsoft/mssql-rs/pull/320), which documented the split, is closed for the same reason.)

## Design

Values go through the same conversion core `SQLGetData` uses, so a column reads the same either way. What differs is cadence: `SQLGetData` can return a long value in chunks across repeated calls, while a bound column gets one shot at a fixed buffer and reports `01004` with the untruncated length in the indicator.

`SQLFreeStmt(SQL_UNBIND)` is part of this rather than a follow-up: mssql-python calls it before every fetch, so the columnar path does not work without it.

## Deliberate limits — each an error rather than a guess

A wrong value delivered silently into an application buffer is worse than a reported failure.

| Case | Behaviour | Why |
| --- | --- | --- |
| Scrolling orientation | `HY106` | Forward-only cursor; treating `SQL_FETCH_PRIOR` as `NEXT` would return the wrong rows |
| Row-wise binding | `HYC00` | Filling a struct array as if it were column-wise would corrupt application memory |
| Bound PLP / LOB column | `SQL_ROW_ERROR` | Needs machinery `SQLGetData` owns — [AB#47361](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47361) |
| NULL at a column bound with no indicator | `22002` | Nowhere to report it, and the slot would read back as the previous row's value |
| `SQLGetData` after a rowset wider than one row | cursor left unpositioned | ODBC expects `SQLSetPos` to nominate the current row; not implemented. Mixed access still works at rowset 1 |
| `SQL_C_DEFAULT` at bind time | `HY003` | Resolving it needs the column's SQL type, and ODBC allows binding before execute |

Two details that are easy to get wrong and are pinned by tests:

- **`BufferLength` is ignored for fixed-width C targets** — the stride comes from the C type. Honouring a caller's `sizeof(array)` would place later rows outside it.
- **`SQL_ATTR_ROW_BIND_OFFSET_PTR` is read per fetch, not per bind,** and displaces the data *and* indicator bases by the same byte count. It was previously accepted as a no-op while returning `SQL_SUCCESS`, which silently misplaced every bound column.

`SQL_API_SQLBINDCOL` and `SQL_API_SQLFETCHSCROLL` are advertised in `SQLGetFunctions` alongside the exports — the Windows DM will not dispatch an entry point it has not been told about, which cost P2 a CI cycle.

## Validation

- **e2e parity: 22 tests, 0 divergences** against msodbcsql 18.6 on SQL Server 2025, including seven bound-fetch cases: rowset fill, several columns bound out of order, NULL indicators, truncation, unbind, rebind, and mixed `SQLGetData` afterwards.
- 654 mssql-odbc + 1680 mssql-tds unit tests; clippy and fmt clean.
- Patch coverage 91.1% from unit tests alone, before the e2e leg CI merges on top.

Two expectations of mine that the parity run corrected, rather than driver bugs: a fetch on a never-executed statement returns `HY010` from the Driver Manager (not `24000` from the driver), and `UNION ALL` gives no ordering guarantee — the NULL-indicator test now orders explicitly.

[AB#46580](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46580)


mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,711 @@
...
+/// `statement_handle` must be a valid `StmtHandle` or null.
+pub(crate) unsafe fn sql_fetch_scroll(
+    statement_handle: SqlHandle,
+    fetch_orientation: SqlSmallInt,
+    fetch_offset: SqlLen,
+) -> SqlReturn {
+    debug!(
+        ?statement_handle,
+        fetch_orientation, "SQLFetchScroll called"
+    );
```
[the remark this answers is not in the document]
David-Engel: Fixed in 0478f08e — the entry trace now logs `fetch_offset` alongside the orientation.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,711 @@
...
+    };
+
+    if binding.target_value_ptr.is_null() {
+        // Bound for its indicator only; report the length without a data write.
+        if matches!(value, ColumnValues::Null) {
+            unsafe { write_if_some(indicator, SQL_NULL_DATA) };
+        }
+        return RowOutcome::Success;
+    }
+    let slot = unsafe { (binding.target_value_ptr as *mut u8).add(row_index * stride) };
```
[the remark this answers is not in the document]
David-Engel: Fixed in 0478f08e, and this was the most serious of the batch — thank you.

You were right that the combination was actively dangerous: `SQLSetStmtAttrW` accepted `SQL_ATTR_ROW_BIND_OFFSET_PTR` into the no-op arm and returned `SQL_SUCCESS`, so an application that set a nonzero offset would have been told it worked while every bound column's data *and* indicator went to the wrong address.

I took the "store and apply" option rather than rejecting it, since rejecting would also turn away the many applications that set the attribute to null harmlessly. It is now a `row_bind_offset_ptr` field on `StmtState`, read once per fetch (not per bind, so an application can move the whole rowset between calls) and added to both the data and indicator bases in `deliver_bound`.

One thing worth recording, which a test caught: ODBC displaces both pointers by the same *byte* count, so a ragged offset would misalign the indicator write. That is the application's responsibility — the attribute is a whole-rowset displacement — but I documented the expectation on `read_bind_offset` rather than leaving it implicit.

Covered by `set_row_bind_offset_ptr_stored` and `the_bind_offset_displaces_both_bases`.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,711 @@
...
+    } else {
+        stmt_state.reset_row_stream();
+    }
+
+    if worst == RowOutcome::Error {
+        post_diag(&mut stmt_state, ERR_STRING_RIGHT_TRUNCATION);
+        return SQL_SUCCESS_WITH_INFO;
+    }
+    if worst == RowOutcome::Info || has_server_info {
+        return SQL_SUCCESS_WITH_INFO;
```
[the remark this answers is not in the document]
David-Engel: Fixed in 0478f08e. You are right that it was wrong in both directions — `Info` posted nothing at all, and every `Error` was labelled `01004` regardless of cause.

`RowOutcome` now carries a `RowIssue`, so a row reports the SQLSTATE that value would have produced through `SQLGetData`:

| Cause | State |
| --- | --- |
| String truncation | `01004` |
| Fractional truncation (`ConvOk::Truncated`) | `01S07` |
| `ConvError::OutOfRange` | `22003` |
| `ConvError::Restricted` | `07006` |
| `ConvError::InvalidCharacterValue` / malformed text | `22018` |
| NULL with no indicator | `22002` |
| Unsupported target, unsupported source, bound PLP | `HYC00` |

`merge` keeps the worst outcome and its reason, so a row that truncates one column and fails another reports the failure. Per-row detail stays in the row status array; the statement-level diagnostic is the worst issue across the rowset. Pinned by `each_issue_posts_its_own_sqlstate` and `conversion_failures_map_to_their_own_sqlstate`.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,711 @@
...
+        SQL_C_SBIGINT | SQL_C_UBIGINT | SQL_C_DOUBLE => 8,
+        SQL_C_GUID => 16,
+        SQL_C_TYPE_DATE => 6,
+        SQL_C_TYPE_TIME => 6,
+        SQL_C_TYPE_TIMESTAMP => 16,
+        SQL_C_SS_TIME2 => 12,
+        SQL_C_SS_TIMESTAMPOFFSET => 20,
+        _ => 0,
+    }
+}
```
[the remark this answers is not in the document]
David-Engel: Fixed in 0478f08e. `element_stride` now derives fixed-width strides from the C type and consults `BufferLength` only for the character and binary targets.

The failure mode you describe is the realistic one: a caller passing `sizeof(array)` rather than `sizeof(element)`. Under the old rule an `SQL_C_SLONG` array with `BufferLength = 400` would have strided 400 bytes per row and written far outside it. `fixed_width_stride_ignores_the_buffer_length` pins that case, and also that a negative length cannot become a huge stride.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,711 @@
...
+        // Bound for its indicator only; report the length without a data write.
+        if matches!(value, ColumnValues::Null) {
+            unsafe { write_if_some(indicator, SQL_NULL_DATA) };
+        }
+        return RowOutcome::Success;
+    }
+    let slot = unsafe { (binding.target_value_ptr as *mut u8).add(row_index * stride) };
+
+    if matches!(value, ColumnValues::Null) {
+        unsafe { write_if_some(indicator, SQL_NULL_DATA) };
```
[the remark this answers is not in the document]
David-Engel: Fixed in 0478f08e — this is now `22002` and `SQL_ROW_ERROR`.

The consequence you point out is the bad part: `write_if_some` on a null pointer is a silent no-op, so the slot kept whatever the previous row left in it and the row was reported successful. The application would have read a stale value with nothing to indicate it was actually NULL.

The check happens before the data write, so the stale value is left visible rather than half-overwritten, and `null_without_an_indicator_is_an_error` asserts exactly that. A non-NULL value still delivers fine without an indicator — only NULL requires one — which `a_value_without_an_indicator_still_delivers` covers.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,711 @@
...
+
+    unsafe { write_if_some(rows_fetched_ptr, rows_filled) };
+    mark_no_rows(row_status_ptr, rows_filled, row_array_size);
+
+    let has_server_info = post_tds_info_messages(&mut stmt_state, &info_messages);
+
+    if let Some(message) = fetch_error {
+        error!(%message, "SQLFetchScroll: row fetch failed");
+        post_sql_error(&mut stmt_state, SQLSTATE_HY000, 0, &message);
+        return SQL_ERROR;
```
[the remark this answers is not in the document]
David-Engel: Fixed in 0478f08e. `fetch_error` now holds the `TdsError` itself rather than a `String`, and the error path calls `post_tds_error(.., SQLSTATE_HY000)`, which fans a server error out into one diagnostic per record with its own SQLSTATE and native error instead of flattening them into a single generic `HY000`.

You were also right about the state: the cursor cannot be resumed after a protocol read fails, so it now calls `reset_row_stream()`, clears `STMT_STATE_CURSOR_OPEN`, and releases `active_stmt` on the connection — matching what `SQLFetch` does in the same situation. Previously the statement was left looking fetchable after the transport had already failed.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,711 @@
...
+        unsafe { write_row_status(row_status_ptr, rows_filled, outcome.status()) };
+        worst = worst.merge(outcome);
+        rows_filled += 1;
+
+        if fetch_error.is_some() {
+            break;
+        }
+    }
+
+    let info_messages = client.take_info_messages();
```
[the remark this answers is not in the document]
David-Engel: Fixed in 0478f08e — the messages are only drained when `rows_filled > 0` or the fetch failed, so a zero-row end of set leaves them on the client.

Your reasoning is the same as the long comment in `fetch.rs`: `SQL_NO_DATA` cannot be upgraded to `SQL_SUCCESS_WITH_INFO`, so anything posted there lands under a return code most applications never pump diagnostics for, and the next API call clears it. Leaving them lets `SQLMoreResults` or the cursor close surface them with a return code that actually hints they are worth reading. I have referenced that reasoning in the new comment so the two paths stay in step.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,1065 @@
...
+fn fetch_scroll_safe(
+    statement_handle: SqlHandle,
+    stmt: &StmtHandle,
+    fetch_orientation: SqlSmallInt,
+    _fetch_offset: SqlLen,
+) -> SqlReturn {
+    // Snapshot the rowset controls and the binding table, then release the
+    // statement lock: the fill loop below blocks on the network and must not
+    // hold it. The application is not allowed to rebind concurrently with a
+    // fetch on the same statement, so the snapshot cannot go stale under us.
```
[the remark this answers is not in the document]
David-Engel: Good catch, and the reasoning is right. The fetch cannot hold the statement mutex across network I/O, so it has to snapshot and then write through those application pointers after releasing the lock — which leaves a window where a concurrent `SQLBindCol` could free a buffer we are still writing into. Worth noting this is a hazard `SQLGetData` does not have, since its target buffer is an argument of the same call and cannot be freed underneath it.

Fixed by claiming the statement for the duration rather than by extending the lock. `STMT_STATE_FETCH_IN_PROGRESS` is set inside the snapshot block before the lock is released, and these now return HY010 while it is set:

- `SQLBindCol` and `SQLFreeStmt(SQL_UNBIND)`
- `SQL_ATTR_ROW_ARRAY_SIZE`, `ROWS_FETCHED_PTR`, `ROW_STATUS_PTR`, `ROW_BIND_OFFSET_PTR`, `ROW_BIND_TYPE`
- a reentrant `SQLFetchScroll`/`SQLFetch`

That is also what ODBC expects of a second call on a busy statement, so the guard and the spec agree.

The flag is cleared at a single point after `fill_rowset` returns, not at each early exit inside the loop — a per-exit clear is exactly the kind of thing that grows a leak later, and a leaked flag would wedge the statement permanently.

mssql-odbc/src/api/bind_col.rs
```diff
@@ -0,0 +1,381 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Implementation of SQLBindCol: the application side of the columnar fetch
+//! path.
+//!
+//! Binding only records where a column's value should land; nothing is written
+//! until `SQLFetchScroll` fills the rowset. That separation is why validation
```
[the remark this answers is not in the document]
David-Engel: Agreed, and this was the more serious of the two — `SQLBindCol` + `SQLFetch` is the most common fetch pattern in ODBC, and we were advertising both entry points while handing back untouched buffers.

`SQLFetch` now delegates to `sql_fetch_scroll_impl(handle, SQL_FETCH_NEXT, 0)` and its previous implementation is deleted rather than kept alongside. A second row-reading path is precisely how this would silently break again.

The delegation also fixes things I had not called out: `SQL_ATTR_ROW_ARRAY_SIZE` now applies to `SQLFetch`, and `rows_fetched_ptr` and the row status array are populated for it too. All of that is required behaviour that the old single-row implementation did not have.

On preserving diagnostics — the existing `fetch.rs` tests are unchanged and still pass, so the null-handle, no-cursor, busy-statement, drained-cursor and 24000 paths report as before. Two new e2e tests (`SQLFetchFillsBoundColumnsToo`, `SQLFetchHonoursTheRowsetSize`) cover the new behaviour, but the real check is the rest of the suite: `get_data`, `exec_direct`, `execute`, `more_results`, `row_count`, `transaction`, `session_recovery` and `connection_pool` all drive `SQLFetch`, and the full run is 22 tests at parity with msodbcsql, 0 divergences.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,1084 @@
...
+        // There is nowhere to report the NULL, and leaving the slot untouched
+        // would read back as the previous row's value.
+        return RowOutcome::Error(RowIssue::IndicatorRequired);
+    }
+
+    if binding.target_value_ptr.is_null() {
+        // Bound for its indicator only; report the length without a data write.
+        if is_null {
+            unsafe { write_if_some(indicator, SQL_NULL_DATA) };
+        }
```
shiwanigupta0809: For indicator-only binding, it writes SQL_NULL_DATA for NULL and then returns. Non-NULL rows currently return success without writing indicator length. Would you consider also writing non-NULL lengths in this branch? Otherwise callers may observe stale indicator values.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,1084 @@
...
+            bindings,
+            stmt_state.rows_fetched_ptr,
+            stmt_state.row_status_ptr,
+            stmt_state.column_metadata.len(),
+            stmt_state.row_bind_offset_ptr,
+        )
+    };
+    // The row cursor only moves forward within a row, so the bound columns have
+    // to be visited in ascending order regardless of the order they were bound.
+    bindings.sort_by_key(|b| b.column_number);
```
shiwanigupta0809: perf suggestion: fetch currently clones and sorts bindings on every call before row delivery. Would it make sense to keep bindings ordered at set/rebind time and iterate directly in fetch? That moves O(n log n) work off the fetch path

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Good catch — that branch was inconsistent, and its own comment ("report the length without a data write") described behaviour it did not have.

I took it to msodbcsql rather than reasoning from the spec, because the spec is genuinely ambiguous here: a null `TargetValuePtr` is also documented as the way to *unbind* a column, so "indicator-only binding" and "unbound column" are the same call. The answer came back the other way from your suggestion — msodbcsql writes **nothing** for an indicator-only binding: no length for a value, and no `SQL_NULL_DATA` for a NULL.

So writing the non-NULL length would have introduced a divergence, and the `SQL_NULL_DATA` write we already had was itself one. Both cases now leave the indicator untouched. Net effect is a divergence removed rather than added, which is not where I expected this to land when I started looking.

Covered both ways round: the unit test asserts nothing is written for a value *and* a NULL, and `IndicatorOnlyBindingDeliversNothing` runs against both drivers in the e2e parity suite, so if msodbcsql ever changes its mind we find out.

Worth saying the underlying concern you raised — a caller reading a stale indicator — is real; it is just that with this binding shape the caller gets nothing from either driver, so there is no length to go stale. An application that wants the length has to use `SQLGetData` with a null buffer.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Taken, though for a slightly different reason than the perf one.

On the numbers I would push back gently: the sort is once per *fetch call*, not per row, and `n` is the number of bound columns — so `O(n log n)` on a handful of entries next to a network round trip. And the clone it sits on top of has to stay regardless: the fetch cannot hold the statement mutex across I/O, so it snapshots the bindings and writes through them with the lock released. That `O(n)` allocation dominates whatever the sort costs, so removing the sort alone is not where a measurable win would come from.

What makes it worth doing is correctness rather than speed. The fill loop walks a forward-only row cursor, so it can *only* visit bound columns in ascending order — the ordering is a requirement, not a tidiness. Sorting at the point of use restates that requirement every time it is needed; keeping the table ordered by construction states it once, in `set_binding`, where it cannot be forgotten by a future caller. So the sort is gone and `set_binding` now does a `binary_search_by_key` and inserts in place, with a unit test that binds out of order and asserts the table stays ascending across a rebind.

If the snapshot clone ever does show up in a profile, the more promising move is taking the bindings out of the state and putting them back, which `STMT_STATE_FETCH_IN_PROGRESS` now makes safe. I have not done that here — it adds a restore-on-every-exit hazard for a cost nobody has measured yet.

mssql-odbc/src/api/bind_col.rs
```diff
@@ -0,0 +1,424 @@
...
+    debug!(column_number, target_type, "SQLBindCol: column bound");
+    SQL_SUCCESS
+}
+
+/// C types this driver can deliver into a bound buffer.
+///
+/// `SQL_C_DEFAULT` is absent deliberately: resolving it needs the column's SQL
+/// type, and ODBC allows binding before the statement is executed, so at bind
+/// time there may be no metadata to resolve it against.
+fn is_supported_c_target(target_type: SqlSmallInt) -> bool {
```
Vahid-b: **Important:** This gate hand-rolls its own C-type check instead of reusing `crate::api::type_rules::canonical_c_type` + `is_valid_c_type`, which this crate's own module doc explicitly says exists to be shared: *"msodbcsql shares `IsValidCType` between its APD and ARD paths (`SetADRec` serves both `SQLBindParameter` and `SQLBindCol`)."* `bind_param.rs` already calls `canonical_c_type` before validating; this function does not.

Concretely, this means `SQL_C_DATE`/`SQL_C_TIME`/`SQL_C_TIMESTAMP` (deprecated 2.x identifiers, still legal in ODBC 3.x per this crate's own `.github/instructions/mssql-odbc.instructions.md`: *"a 3.x application may legally pass them and the DM remaps nothing... fold them onto the 3.x form"*) are rejected here with `HY003`. I verified this by running the actual code: `sql_bind_col(stmt, 1, SQL_C_DATE, ...)` returns `SQL_ERROR` / `HY003` on this branch, while `SQLBindParameter` on the same connection accepts it — an inconsistency within this driver, not just vs. msodbcsql.

It also HY003s other types `type_rules::is_valid_c_type`'s own test (`real_c_types_are_valid_even_when_unconvertible`) says must pass this gate and fail later at `07006`/`HYC00` instead (e.g. `SQL_C_NUMERIC`, `SQL_C_INTERVAL_YEAR..SQL_C_INTERVAL_MINUTE_TO_SECOND`).

**Suggested fix:** canonicalize with `canonical_c_type(target_type)` first, validate with `type_rules::is_valid_c_type(...)` (after the deliberate `SQL_C_DEFAULT` exception this PR already documents), and store the canonicalized value in `ColumnBinding` so `element_stride`/`deliver_bound` see one form per type.

mssql-odbc/src/api/set_stmt_attr.rs
```diff
@@ -190,13 +207,17 @@ fn sql_set_stmt_attr_w_safe(
                 SQL_SUCCESS_WITH_INFO
             }
         }
+        SQL_ATTR_ROW_BIND_OFFSET_PTR => {
```
Vahid-b, line 210: **Important:** `SQL_ATTR_ROW_BIND_OFFSET_PTR` is stored here and returns `SQL_SUCCESS`, but `sql_get_stmt_attr_w_safe` (not touched by this diff, so I can't anchor a comment directly on it) has no matching arm in its `match attribute` — it falls through to `_ => HY092 (invalid attribute identifier)`.

I verified this by running the actual code: `SQLSetStmtAttrW(..., SQL_ATTR_ROW_BIND_OFFSET_PTR, ...)` returns `rc=0`, and an immediate `SQLGetStmtAttrW(..., SQL_ATTR_ROW_BIND_OFFSET_PTR, ...)` on the same handle returns `rc=-1`. The one existing test (`set_row_bind_offset_ptr_stored`, added just below) only reads the internal `StmtState.row_bind_offset_ptr` field directly — it never calls the getter, so this had zero test coverage.

**Suggested fix:** add `SQL_ATTR_ROW_BIND_OFFSET_PTR => unsafe { write_if_some(value_ptr as *mut *mut SqlULen, state.row_bind_offset_ptr) }` to `sql_get_stmt_attr_w_safe`, matching the pattern already used for `ROW_ARRAY_SIZE`/`ROWS_FETCHED_PTR`/`ROW_STATUS_PTR`/`ROW_BIND_TYPE`.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,1085 @@
...
+                fetch_error = Some(e);
+                break;
+            }
+        }
+
+        let mut outcome = RowOutcome::Success;
+        let mut columns_read = 0usize;
+        for binding in bindings {
+            let column = binding.column_number as usize;
+            if column == 0 || column > column_count {
```
Vahid-b, line 331: **Important:** A binding whose column number exceeds the current result's column count (e.g. left over from a prior, wider result set) reports `RowIssue::Restricted` -> SQLSTATE `07006` ("restricted data type attribute violation", a conversion error).

But `bind_col.rs`'s own `binding_the_bookmark_column_is_rejected` test asserts `07009` ("invalid descriptor index") for the same class of problem -- an out-of-range ordinal -- just caught earlier (bind time) instead of here (fetch time). This is an inconsistency within this PR itself: one file correctly uses `07009` for "ordinal out of range," this one uses `07006`.

No test exercises this exact fetch-time path (confirmed: no test in this file binds a column beyond `column_count` and asserts on the resulting SQLSTATE), so this is a coverage gap masking the wrong-SQLSTATE choice rather than a failing assertion.

**Suggested fix:** give this case its own `RowIssue` variant (or otherwise distinguish it from genuine type-conversion restriction) mapping to `07009`.

Vahid-b (commented):
## Review Summary

Reviewed the columnar `SQLBindCol`/`SQLFetchScroll` fetch path against `.github/instructions/mssql-odbc.instructions.md` and this crate's own established conventions (`type_rules.rs`, `bind_param.rs`). This PR already went through a thorough round with the Copilot bot, David-Engel, and shiwanigupta0809 -- 9 of that round's findings were fixed with detailed replies. I focused on cross-checking whether the bot's remaining collapsed/"suppressed" comments from that same round were actually addressed, since those get much less visibility in the GitHub UI than a top-level inline comment.

All 5 findings below were independently verified against the current code (two of them -- #1 and #2 -- by writing and running temporary, reverted-after-use unit tests directly against this PR's head commit, not just static reading). `cargo fmt` / `clippy` / `nextest` (734 tests) all pass clean, consistent with these being coverage gaps rather than caught regressions.

mssql-odbc/src/api/bind_col.rs
```diff
@@ -0,0 +1,424 @@
...
+            column_number,
+            target_type,
+            target_value_ptr,
+            buffer_length,
+            strlen_or_ind_ptr,
+        )
+    })
+}
+
+unsafe fn sql_bind_col_impl(
```
Vahid-b, line 58: **Suggestion:** This function (and the closure inside `sql_free_stmt_unbind` below) does not follow the safe-core/unsafe-shell split this crate's instructions mandate: a thin `unsafe fn ..._impl` doing only null-check -> `handle_from_raw` -> delegate, plus a safe `fn ..._safe(&StmtHandle, ...)` core holding the locking/validation/business logic.

`fetch_scroll.rs` in this same PR does this correctly -- `sql_fetch_scroll_impl` is a ~12-line shim that delegates everything to `fetch_scroll_safe`. This function instead does all locking, `free_errors`, state checks, diagnostics, and mutation directly inside the `unsafe fn`.

Not a safety bug by itself (the logic here is careful), but it's the one pattern this crate's entire FFI-safety argument depends on being applied uniformly, and it's inconsistent with the sibling file added in the same commit set.

**Suggested fix:** extract `sql_bind_col_safe(stmt: &StmtHandle, ...) -> SqlReturn` and `sql_free_stmt_unbind_safe(stmt: &StmtHandle) -> SqlReturn`, leaving only the null-check/handle-conversion/`debug_assert_eq!` in the `unsafe fn`.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,1085 @@
...
+/// `row_index + 1` elements — the contract `SQLBindCol` places on the
+/// application together with `SQL_ATTR_ROW_ARRAY_SIZE`.
+unsafe fn deliver_bound(
+    binding: &ColumnBinding,
+    row_index: usize,
+    bind_offset: usize,
+    value: &ColumnValues,
+) -> RowOutcome {
+    let stride = element_stride(binding.target_type, binding.buffer_length);
+    if stride == 0 {
```
Vahid-b: **Suggestion (needs msodbcsql verification, flagging as a question rather than a confirmed defect):** A bound `SQL_C_CHAR`/`SQL_C_WCHAR` column with `BufferLength = 0` is rejected here with `HYC00` before the indicator is ever written.

I could not verify against the actual msodbcsql source (Azure DevOps-hosted, not reachable from my review environment) whether a zero-length character binding is a legal "length probe" pattern for `SQLBindCol` specifically -- it's a well-established pattern for `SQLGetData`, but I'm not certain the same convention applies here. If it does, the existing `copy_with_nul` path a few lines down would already handle it safely (write nothing, report the length, flag truncation for non-empty data) without this early rejection.

Worth a deliberate look at msodbcsql before deciding either way -- not blocking, since I can't confirm it's wrong.

mssql-odbc/src/api/bind_col.rs
David-Engel: Agreed, and the module doc you quoted is the strongest part of the argument — `type_rules` exists precisely so the ARD and APD paths cannot drift, and this one hand-rolled its own gate anyway. Fixed: `canonical_c_type` then `is_valid_c_type`, keeping the deliberate `SQL_C_DEFAULT` rejection ahead of both, and the **canonical** form is what goes into `ColumnBinding` so `element_stride`/`deliver_bound` only ever see one spelling.

Your point about the layering is the one I had wrong, not just the type list: `SQL_C_NUMERIC` and the intervals now pass this gate and fail per row at `HYC00`, which is what `type_rules` documents this gate to mean ("validity here means only that an identifier names a real ODBC type"). Rejecting them at bind time was conflating "not a type" with "cannot deliver it yet".

Two tests: one binds `SQL_C_DATE`/`SQL_C_TIME`/`SQL_C_TIMESTAMP` and asserts both that they are accepted and that the stored `target_type` is the `SQL_C_TYPE_*` form; one asserts the valid-but-undeliverable types clear the gate.

mssql-odbc/src/api/set_stmt_attr.rs
David-Engel: Confirmed and fixed — I added the setter arm without the matching getter, so it was `SQL_SUCCESS` to set and `HY092` to read back.

Your observation about the test is the more useful half of this: `set_row_bind_offset_ptr_stored` read `StmtState.row_bind_offset_ptr` directly, so it asserted the storage and never the API contract. It now round-trips through `SQLGetStmtAttrW`. I also deliberately removed the new arm and re-ran it to confirm it fails without the fix (`left: -1`, `right: 0`) rather than trusting that it would — I had a test earlier in this PR that passed with and without the change it was supposed to guard, so I stopped assuming.

mssql-odbc/src/api/bind_col.rs
David-Engel: Fair, and it is the right thing to hold the line on — the layering is only worth anything if it is applied uniformly, and `fetch_scroll.rs` in the same commit set doing it correctly makes the inconsistency worse rather than better.

Extracted `sql_bind_col_safe(&StmtHandle, ...)` and `sql_free_stmt_unbind_safe(&StmtHandle)`. The `unsafe fn`s are now null-check → `handle_from_raw` → `debug_assert_eq!` → delegate, with all locking, `free_errors`, state checks and mutation in safe code.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: You were right that `07006` was wrong, and right that it was untested — but the answer turned out to be neither SQLSTATE. I put it to msodbcsql, and it does not report this at all: it **skips** a binding whose ordinal is past the end of the result set and returns plain `SQL_SUCCESS` with no diagnostic record at all (verified: `SQLGetDiagRecW` returns `SQL_NO_DATA`, and the bound indicator is left untouched).

So a stale binding left over from a wider result set is simply not an error there. Raising `07009` would have been more defensible than `07006` on spec-reading grounds, but it would still have broken any application that leaves bindings in place across `SQLMoreResults` — it would start failing fetches against this driver that succeed against msodbcsql. The fetch now skips it silently.

Covered by `AStaleOrdinalPastTheResultSetIsIgnored`, which asserts `SQL_SUCCESS`, no diagnostic, and an untouched indicator, and runs against both drivers in the parity suite.

Worth noting the coverage gap you identified is what made this findable: the wrong SQLSTATE was invisible precisely because nothing exercised the path.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Good question to raise rather than guess at, and it was worth asking — msodbcsql **does** treat it as a length probe.

I can reach msodbcsql (the e2e suite runs both drivers side by side against a local SQL Server), so I turned it into a parity test. With `SQL_C_CHAR`, `BufferLength = 0` and a live indicator over a `varchar` `hello`: msodbcsql returns `SQL_SUCCESS_WITH_INFO`, writes `5` to the indicator, and does not touch the buffer. We returned `HYC00` and left the indicator alone — a real divergence.

Your read of the code was exactly right too: `copy_with_nul` already handles `buf_len == 0` correctly (writes nothing, reports truncation), so the early `stride == 0` rejection was the only thing wrong. Removed it, and a zero stride now falls through to that path. A zero stride can only arise from a character or binary binding with `BufferLength 0` in the first place — fixed-width targets take their stride from the C type — so nothing else was relying on that guard.

Locked in both ways: a unit test on `deliver_bound` asserting length-reported/buffer-untouched/`01004`, and `ZeroLengthCharacterBindingIsALengthProbe` in the parity suite.

Vahid-b (approved):

shiwanigupta0809 (approved):

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,1113 @@
...
+        stmt_state.reset_row_stream();
+    }
+
+    // Report why the rowset was imperfect with the SQLSTATE that value would
+    // have produced through SQLGetData, rather than a blanket truncation
+    // warning. Per-row detail lives in the row status array.
+    if let Some(issue) = worst.issue() {
+        issue.post(&mut stmt_state);
+        return SQL_SUCCESS_WITH_INFO;
+    }
```
Theekshna: **No row outcome can ever make this call return `SQL_ERROR`.**

`worst.issue()` is `Some` for `RowOutcome::Error` just as it is for `Info`, so a rowset in which *every* row failed — say every row's bound column was NULL with no indicator, or every row hit the unsupported-PLP path — still returns `SQL_SUCCESS_WITH_INFO`. The row status array is correct; the return code is what applications actually branch on.

That matters most for the callers this path exists to serve. A `pyodbc`/`mssql-python`-shaped wrapper checks `rc < 0` before it looks at anything else; here it sees success and reads buffers that were never written, which for a fixed-width target is the previous row's value rather than anything obviously wrong.

msodbcsql cannot land here. Its `retcode` is a running worst-status via `PromoteRetcode` (`TdsParser.h:375-382`), which returns `rcOld` unchanged once `SUCCESS(rcOld)` is false, so `SQL_ERROR` is sticky. It is not reset per row for a block fetch with a status array (`sqlccurs.cpp:1499-1504`; the only reset path at `:1901-1906` requires `rgfRowStatus == NULL`), and the final adjustment only promotes `SQL_SUCCESS` to `SQL_SUCCESS_WITH_INFO` — it never demotes `SQL_ERROR` (`sqlccurs.cpp:2155-2161`).

Reading that, msodbcsql looks stricter than the spec: any row error appears to make the whole call `SQL_ERROR`, where ODBC only requires it when *all* rows error. I am confident about the all-error case and less so about mixed, since stickiness across rows is read off the loop structure rather than an explicit merge. So the minimum I would ask for is:

- all rows `SQL_ROW_ERROR` → `SQL_ERROR`
- some rows errored, some succeeded → `SQL_SUCCESS_WITH_INFO` (as today)

and if you want to match msodbcsql exactly rather than the spec, `SQL_ERROR` for both. Either way `RowOutcome::Error` needs to be distinguishable from `Info` here, which `worst` already carries — it is just being collapsed by `issue()`.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -0,0 +1,1113 @@
...
+    let indicator = if binding.strlen_or_ind_ptr.is_null() {
+        std::ptr::null_mut()
+    } else {
+        unsafe {
+            (binding.strlen_or_ind_ptr as *mut u8)
+                .add(bind_offset)
+                .cast::<SqlLen>()
+                .add(row_index)
+        }
+    };
```
Theekshna: **This address can be misaligned, and the write through it is a plain aligned write.**

`bind_offset` is an arbitrary byte count the application supplies through `SQL_ATTR_ROW_BIND_OFFSET_PTR`; nothing constrains it to a multiple of `align_of::<SqlLen>()`. After `.add(bind_offset).cast::<SqlLen>()` the result can therefore be misaligned, and `write_if_some` (`util.rs:21-25`) reaches it with `ptr.write(value)`.

The doc on `read_bind_offset` (`:505-507`) already spots this and assigns it to the application: *"an application is expected to use a whole-rowset displacement that keeps both naturally aligned; a ragged offset would misalign the indicator write."* That is the part I would push back on. In C a misaligned store is benign on x86 and the expectation is reasonable; in Rust `ptr::write` on a misaligned pointer is UB on every target regardless of whether the hardware faults, and the optimiser is entitled to assume it cannot happen. So this is not "the application gets a garbled indicator" — it is a documented UB path in our own code, reachable through a supported attribute we cannot reject.

`write_unaligned` costs nothing on x86 and the fix is mechanical. Same class, lower risk since the application allocates these itself and they will normally be aligned: `read_bind_offset`'s `ptr.read()` (`:515`) and `write_row_status`'s `.write(status)` (`:479`).

Two caveats so this is weighted correctly. `write_if_some` and `copy_with_nul` are pre-existing shared helpers, so fixing them properly is crate-wide and arguably not this PR's job — what is this PR's is that it is the first code path where an application-controlled byte offset makes misalignment reachable, and that it documents the hazard as acceptable. And on the msodbcsql side: it does qualify application-buffer accesses `UNALIGNED` (`sqlccnvt.cpp:1677-1714`, `*(UNALIGNED LONG *)lpIn` and friends), but those are *reads of input buffers* — I did not find `UNALIGNED` on the indicator write specifically, so treat that as supporting evidence for the principle rather than proof of the exact case.

Happy for the helper fix to be a follow-up. The doc comment declaring the hazard acceptable is the part I would not ship.

mssql-odbc/src/api/bind_col.rs
```diff
@@ -0,0 +1,491 @@
...
+    }
+
+    // SQL_C_DEFAULT would have to be resolved against the column's SQL type,
+    // and ODBC allows binding before execution, so there may be no metadata to
+    // resolve it against yet.
+    if target_type == SQL_C_DEFAULT {
+        error!("SQLBindCol: SQL_C_DEFAULT is not supported as a bound target");
+        post_diag(&mut stmt_state, ERR_INVALID_C_DATA_TYPE);
+        return SQL_ERROR;
+    }
```
Theekshna: **This rejection is inconsistent with the layering the rest of this commit just adopted, and msodbcsql does not do it.**

The commit message makes the case against this gate better than I can: *"Types that name a real ODBC type but that the fetch cannot deliver (`SQL_C_NUMERIC`, the intervals) now pass this gate and fail per row, which is the layering `type_rules` documents."*

`SQL_C_DEFAULT` names a real ODBC type, and unlike `SQL_C_NUMERIC` it is one the fetch *can* deliver — it just needs the column metadata first, which is the definition of a per-row question. So it now sits on the wrong side of a line this commit drew: the types that can never work are admitted and fail at fetch, and the one that will work once metadata exists is refused at bind.

msodbcsql resolves it exactly where the layering says to. It stores the C type as given and folds it at delivery time via `Sql2CDefault()` — `sqlcfunc.cpp:2170` in `BindOffset`: `if (lpbindinfo->fCType == SQL_C_DEFAULT) fCType = Sql2CDefault(fSqlType, ...)`, and again on the `InternalGetColData` path in `sqlcdata.h`. There is no bind-time rejection of `SQL_C_DEFAULT` anywhere in `SQLBindCol` / `SetADRec` / `CheckADDescRecConsistency`.

The comment argues the case from the ODBC contract ("ODBC allows binding before execution, so there may be no metadata to resolve it against yet") rather than from msodbcsql. That premise is true and it is exactly the premise msodbcsql starts from — it just reaches the opposite conclusion, deferring instead of refusing. Given this commit went to msodbcsql to settle three other questions, this one looks like it was left on the spec argument by accident.

Practical impact: `SQLBindCol(hstmt, i, SQL_C_DEFAULT, buf, len, &ind)` is the ordinary way to say "give me the natural C type for this column", and it works on msodbcsql today.

If deferring is more work than it looks — resolving at fetch means `element_stride` and `deliver_bound` need the IRD type, which the binding does not carry — then that is a fine reason to keep the rejection for now, but it should be recorded as a deliberate divergence with a work item and the comment should say msodbcsql defers rather than implying the spec forces a rejection.

mssql-odbc/src/api/bind_col.rs
```diff
@@ -0,0 +1,491 @@
...
+
+    // Both null unbinds the column; a null data pointer with a live indicator
+    // stays bound, so the application can collect lengths without the values.
+    if target_value_ptr.is_null() && strlen_or_ind_ptr.is_null() {
+        stmt_state.clear_binding(column_number);
+        debug!(column_number, "SQLBindCol: column unbound");
+        return SQL_SUCCESS;
+    }
```
Theekshna: **A null `TargetValuePtr` should unbind the column outright, whatever the indicator is.**

msodbcsql keys only on the data pointer and never inspects the indicator argument here — `sqlcdesc.cpp:2660-2666`:

```cpp
// If unbinding a single column (SQLBindCol/SQLSetDescRec)
if ( wBindParamFlag == 0 && rgbValue == NULL )
{
    // a NULL value ptr indicates that the descriptor at the specified index should be unbound
    retcode = UnbindParam ( pAD, icol );
}
```

`UnbindParam` (`:3961-4007`) calls `PlDeleteI(...)` — the descriptor record is deleted. There is no indicator-only binding concept in the driver at all.

The delivery side was already brought into line in `3f2e7737` (`fetch_scroll.rs:555-559` now returns without writing), so the observable fetch behaviour matches. Two things still differ, though, because the binding is retained rather than removed:

1. **The column is still consumed.** The fill loop calls `read_row_column(column - 1)` for every binding before `deliver_bound` gets to return early, and sets `columns_read = column`, which becomes `current_row_last_col`. So an indicator-only binding on column *n* silently burns columns 1..n of the row for a following `SQLGetData`, even though it delivers nothing. A binding whose entire effect is to consume a column and write nothing is a strange thing to keep.

2. **Validation still runs on the way past.** With a live indicator, control reaches the `buffer_length < 0`, `SQL_C_DEFAULT` and `is_valid_c_type` checks below. So `SQLBindCol(hstmt, 1, SQL_C_DEFAULT, NULL, 0, &ind)` returns `SQL_ERROR` here and `SQL_SUCCESS` (unbind) on msodbcsql. Applications that unbind by nulling the buffer while leaving the old indicator and type arguments in place — a common shape when the call is generated — hit this.

Moving the `clear_binding` to key on `target_value_ptr.is_null()` alone removes both, deletes the special case in `deliver_bound`, and matches the source. The comment on the line below (*"a null data pointer with a live indicator stays bound, so the application can collect lengths without the values"*) then goes away too, which is worth doing on its own — it describes a capability that no longer exists, since that binding now collects nothing.

Not caught by `IndicatorOnlyBindingDeliversNothing`: it asserts the indicators are untouched, which is true either way, and never calls `SQLGetData` afterwards.

mssql-odbc/src/api/fetch_scroll.rs
Theekshna, line 463: **`RowOutcome::Error` at `row_array_size == 1` should return `SQL_ERROR`.**

`worst.issue()` treats `Error` like `Info`, so every row outcome returns `SQL_SUCCESS_WITH_INFO` regardless of rowset size. msodbcsql keys on the rowset size, not on how many rows failed:

- **`dwRowSize > 1`** (= `SQL_ATTR_ROW_ARRAY_SIZE`, `sqlccurs.cpp:790`): a row error is explicitly demoted — `if (retcode != SQL_NO_DATA_FOUND) retcode = SQL_SUCCESS;` with `fWithInfo = TRUE` (`:1974-1988`) — so `:2160-2161` yields `SQL_SUCCESS_WITH_INFO`. **This PR already matches**, including when every row errored: `(*pcrow)++` runs on the error path too (`:1805`), so the `*pcrow == 0` escalation at `:2155` can't fire.
- **`dwRowSize == 1`**: that block is gated off, so the `SQL_ERROR` from `PromoteRetcode` (`:1752`) survives and the call returns `SQL_ERROR`.

The gap is therefore confined to `row_array_size == 1` — the default, and now the plain `SQLBindCol` + `SQLFetch` loop since `SQLFetch` routes through here. A bound column that is NULL with no indicator returns `SQL_ERROR` on msodbcsql and `SQL_SUCCESS_WITH_INFO` here, even though the row status is already correctly `SQL_ROW_ERROR`.

`RowOutcome::Info` needs no change — truncation is `SQL_SUCCESS_WITH_INFO` at either rowset size. Only `Error` at `row_array_size == 1` should become `SQL_ERROR`.

mssql-odbc/src/api/fetch_scroll.rs
Theekshna, line 553: **This address can be misaligned, and `write_if_some` writes through it with `ptr.write`** (`util.rs:21-25`). `bind_offset` is an arbitrary application byte count from `SQL_ATTR_ROW_BIND_OFFSET_PTR`; nothing constrains it to `align_of::<SqlLen>()`.

`read_bind_offset`'s doc (`:505-507`) already spots this and calls it the application's responsibility. That's the part I'd push back on: in C a misaligned store is benign on x86, but in Rust `ptr::write` on a misaligned pointer is UB on every target and the optimiser may assume it can't happen. `write_unaligned` costs nothing on x86.

Same class, lower risk: `read_bind_offset`'s `ptr.read()` (`:515`) and `write_row_status`'s `.write()` (`:479`).

Two caveats. `write_if_some`/`copy_with_nul` are pre-existing shared helpers, so the real fix is crate-wide and fine as a follow-up — what's new here is that an application-controlled offset makes misalignment reachable. And msodbcsql's `UNALIGNED` (`sqlccnvt.cpp:1677-1714`) covers input-buffer *reads*, not the indicator write, so it supports the principle rather than proving this case. The doc comment declaring the hazard acceptable is the part I wouldn't ship.

mssql-odbc/src/api/bind_col.rs
Theekshna: **Inconsistent with the layering this commit just adopted.** From the commit message: *"Types that name a real ODBC type but that the fetch cannot deliver (`SQL_C_NUMERIC`, the intervals) now pass this gate and fail per row."* `SQL_C_DEFAULT` names a real ODBC type and, unlike those, is one the fetch *can* deliver — it just needs metadata, which is a per-row question by that same rule.

msodbcsql defers rather than refusing: `sqlcfunc.cpp:2170` (`BindOffset`) folds it via `Sql2CDefault(fSqlType, ...)`, likewise on the `InternalGetColData` path. No bind-time rejection exists in `SQLBindCol` / `SetADRec` / `CheckADDescRecConsistency`.

The comment argues from the ODBC contract rather than msodbcsql — which starts from the same premise and defers. `SQLBindCol(..., SQL_C_DEFAULT, ...)` is the ordinary "give me the natural C type" call and works there today.

If deferring is real work (the binding doesn't carry the IRD type `element_stride`/`deliver_bound` would need), keeping the rejection is fine — as a tracked divergence, with the comment saying msodbcsql defers.

mssql-odbc/src/api/bind_col.rs
Theekshna: **A null `TargetValuePtr` should unbind outright, regardless of the indicator** — `sqlcdesc.cpp:2660-2666`:

```cpp
if ( wBindParamFlag == 0 && rgbValue == NULL )
    retcode = UnbindParam ( pAD, icol );   // -> PlDeleteI, record deleted
```

The indicator argument is never inspected; there's no indicator-only binding in the driver at all.

`3f2e7737` aligned the delivery side, but retaining the binding still differs twice:

1. **The column is still consumed.** The fill loop calls `read_row_column` and sets `columns_read` before `deliver_bound` returns early, so an indicator-only binding on column *n* burns columns 1..n for a following `SQLGetData` while delivering nothing.
2. **Validation still runs.** With a live indicator, control reaches the `buffer_length` / `SQL_C_DEFAULT` / `is_valid_c_type` checks below, so `SQLBindCol(h, 1, SQL_C_DEFAULT, NULL, 0, &ind)` errors here and unbinds on msodbcsql.

Keying `clear_binding` on `target_value_ptr.is_null()` alone fixes both and lets the special case in `deliver_bound` go. The comment below needs to go with it — it describes collecting lengths without values, which that binding no longer does.

`IndicatorOnlyBindingDeliversNothing` doesn't catch this: untouched indicators hold either way, and it never calls `SQLGetData`.

Theekshna (commented):
Reviewed at `a86b5a92`; msodbcsql behaviour checked against the classic driver source.

The parity homework is accurate — fixed-width stride from the C type (`sqlcfunc.cpp:2159-2287`), untruncated length + `01004` + per-row SWI, `22002`/`SQL_ROW_ERROR`, `HY106` (`sqlccurs.cpp:930-937`), reactive `SQL_ROW_NOROW` (`:1573-1579`) and `*rows_fetched_ptr = 0` on `SQL_NO_DATA` (`:977-987`) all match. `a86b5a92` is a good response to the last round: three of the four went back to msodbcsql rather than the spec, and the stale-ordinal one correctly overruled the SQLSTATE

Four issues I found, listed below.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Confirmed and fixed — and thank you for isolating it to `row_array_size == 1` rather than just saying the return code was wrong. The `(*pcrow)++`-on-the-error-path detail is what makes the block case genuinely already-correct rather than accidentally so, and it saved me from "fixing" the demotion too.

`RowOutcome::Error` at rowset size 1 now returns `SQL_ERROR`; `Info` is untouched at either size, as you said.

Verified both directions against msodbcsql rather than just the changed one, since the risk here was over-applying it: `ARowErrorIsSQL_ERRORAtRowsetSizeOne` (bound column NULL with no indicator → `SQL_ERROR`) and `TheSameRowErrorIsDemotedInABlockFetch` (same error at size 2 → `SQL_SUCCESS_WITH_INFO` with `SQL_ROW_ERROR` in the status array). Both at parity.

This mattered more than it looks: it was reachable from the plain `SQLBindCol` + `SQLFetch` loop precisely because `SQLFetch` now routes through here, so it was a regression this PR introduced rather than a pre-existing gap.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: You are right, and the doc comment was the worst part of it — I wrote the hazard down and then reasoned it away as the applications problem. It is not: `ptr::write` to a misaligned address is UB in Rust regardless of target, and `bind_offset` is an arbitrary application byte count, so this PR is what makes it reachable.

Fixed in the shared helpers rather than at the call sites — `write_if_some` and `copy_with_nul` now use `write_unaligned`, and `copy_with_nul` copies through bytes since `copy_nonoverlapping` also requires alignment and a displaced `SQL_C_WCHAR` buffer can be odd-addressed. `read_bind_offset` and `write_row_status` likewise. I went to the helpers instead of the reachable callers because they are the documented chokepoint, unaligned access costs nothing on x86 or ARM64, and fixing only todays callers leaves the trap armed for the next one.

Your caveat about msodbcsqls `UNALIGNED` is fair and I have not claimed it as precedent — it covers input-buffer reads, and this is an output write. The argument here is the Rust semantics, not parity.

The offending doc comment is gone; it now says the write is unaligned and why.

mssql-odbc/src/api/bind_col.rs
David-Engel: Right on both counts, and the two consequences you listed are the part I had missed — I fixed the delivery side in `3f2e7737` and stopped there, which left the binding alive to do damage elsewhere. Burning the column for a following `SQLGetData` is the one that would actually bite someone.

`clear_binding` is now keyed on `target_value_ptr.is_null()` alone, before any validation, and the special case in `deliver_bound` is gone with it. That also means `SQLBindCol(h, 1, SQL_C_DEFAULT, NULL, 0, &ind)` unbinds instead of erroring — covered by a test, since it is the case where "unbind first" is observable.

You were also right that `IndicatorOnlyBindingDeliversNothing` could not catch this: an untouched indicator holds either way. Replaced with `ANullTargetPointerUnbindsAndLeavesTheColumnReadable`, which binds, rebinds with a null pointer and a live indicator, fetches, and then reads the same column back through `SQLGetData` — that read is what distinguishes an unbind from a binding that delivers nothing, and it is what the old test was missing.

mssql-odbc/src/api/bind_col.rs
David-Engel: Fair, and the comment was the weak point — it argued from the ODBC contract, and as you say msodbcsql starts from the same premise and defers anyway, so the argument does not survive contact with the driver. I verified the divergence rather than taking it on the source read: `SQLBindCol(..., SQL_C_DEFAULT, ...)` followed by `SQLFetch` delivers the value on msodbcsql and returns `HY003` here.

Taking your escape hatch, and deferring is real work for the reason you guessed: `ColumnBinding` does not carry the IRD type, and `element_stride`/`deliver_bound` both need it, so resolving would mean threading the per-column SQL type and the ODBC version through the fetch snapshot into the fill loop. `resolve_default_c_type` already exists (`bind_param` uses it), so it is plumbing rather than new logic, but it is a distinct change and this PR is already large.

So it stays refused, now as a tracked divergence: the comment says msodbcsql defers and points at `sqlcfunc.cpp`, and the plan docs deliberate-limits table records it with the reason. One thing that informed the call — no consumer binds `SQL_C_DEFAULT` on the fetch path today; mssql-python uses it only for parameters, never with `SQLBindCol` — so this is a real gap but not one anything is currently hitting. Happy to pull it into this PR instead if you would rather not ship the divergence.

Vahid-b (approved):

PR #323, opened 2026-08-18
Add ODBC 3.x type rules and SQL_C_DEFAULT resolution to SQLBindParameter

Theekshna:
## Description

First phase (P1) of the parameter conversion milestone: a re-usable type model for `SQLBindParameter`, split so the pieces that are direction-neutral can be shared with the fetch path.

- **`api/type_rules.rs` (new)** — direction-neutral ODBC type-identifier rules: `is_valid_c_type` (HY003), `is_valid_sql_type` (HY004), and `resolve_default_c_type`. Placed under `api/` rather than `params/` because msodbcsql keeps the equivalent `IsValidCType` / `IsValidSqlType` / `Sql2CDefault` in the shared `sqlcprot.h`, where `SetADRec` serves both `SQLBindCol` and `SQLBindParameter`.
- **`params/conversion_matrix.rs` (new)** — table-driven parameter-direction C→SQL conversion check, mirroring msodbcsql's `fValidConversion`. Returns `07006` for a real-but-unsupported pairing. Parameter-side only: msodbcsql's fetch path does not consult `fValidConversion` either.
- **`api/bind_param.rs`** — validation order is now C type (HY003) → SQL type (HY004) → `SQL_C_DEFAULT` resolution → supported-conversion check (07006) → input-only check (HYC00). The resolved C type is what gets stored on the `BoundParam`, so nothing downstream sees `SQL_C_DEFAULT`.
- **`params/convert.rs`** — type validation removed; it is now purely value conversion.
- **`api/odbc_types.rs`** — added the identifiers the rules need (`SQL_SS_TABLE`, `SQL_SS_VECTOR`, `SQL_C_SS_VECTOR`, and the interval C/SQL type range bounds).

The phased plan and the per-phase breakdown are in `mssql-odbc/docs/parameters_plan.md`.

### Accepted parity deviation

`SQL_C_DEFAULT` resolves the wide character SQL types to `SQL_C_WCHAR` and `SQL_GUID` to `SQL_C_GUID`. msodbcsql's `Sql2CDefault` reads `rgbTRANSTYPE380` (`Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp`) and resolves both to `SQL_C_CHAR` — an ANSI-transfer default this driver has no equivalent for, since its `SQL_C_CHAR` is UTF-8. Resolving UTF-16 application input to a UTF-8 buffer type would silently corrupt data (`read_char_bytes` with `SQL_NTS` stops at the first `0x00`, the high half of the first code unit). Recorded in the instructions file, the plan doc, and as a comment on [AB#47365](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47365).

This crate also targets ODBC 3.x only, so msodbcsql paths that exist solely for 2.x compatibility (the `rgbTRANSTYPE` selection for 3.51-or-earlier apps, `IsValidSqlType` accepting the 2.x `SQL_DATE` / `SQL_TIME` / `SQL_TIMESTAMP` spellings) are deliberately not ported. That rule is now written down in `.github/instructions/mssql-odbc.instructions.md`.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47365

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

`cargo btest` could not be run locally — neither `cargo-nextest` nor `cargo-llvm-cov` is installed on this machine. Substituted `cargo test -p mssql-odbc --lib`: **612 passed, 0 failed**. Relying on PR validation for the real `btest` run.

## Testing

- 612 unit tests pass, including new coverage for: a real-but-unconvertible C type returning `07006`, `SQL_C_DEFAULT` being resolved before storage, a resolved-but-unconvertible default returning `07006`, and a default with no mapping (`SQL_INTERVAL_YEAR`) returning `07006`.
- Three e2e cases added to `mssql-odbc/tests/e2e/tests/execute_test.cpp`: `DefaultCTypeNarrowCharParam`, `DefaultCTypeWideCharParam`, and `DefaultCTypeUnsupportedConversionIsRejected`. The latter two carry `SKIP_IF_COMPARING_MSODBCSQL()` — the first for the deviation above, the second because msodbcsql supports `SQL_C_SLONG` → `SQL_INTEGER` and this driver does not yet (P2).
- **The e2e suite has not been executed against a live server.** The binary compiles and all three tests are registered; running them needs an elevated shell (the runner writes to HKLM) and a reachable SQL Server. Please treat that as outstanding.

### DIFF
--- a/.github/instructions/mssql-odbc.instructions.md
+++ b/.github/instructions/mssql-odbc.instructions.md
@@ -28,14 +28,36 @@ authoritative parity reference for this crate. Its source lives in the
 - When reporting a parity finding, cite the msodbcsql source (file + what it does),
   and state explicitly whether the decision **matches**, **exceeds**, or **diverges
   from** msodbcsql so the trade-off is visible.
-- **This driver targets ODBC 3.x only.** msodbcsql also serves ODBC 2.x
-  applications, so parts of its source exist purely for backward compatibility:
-  deprecated 2.x entry points, 2.x identifier spellings and attribute values,
-  2.x-era internal representations, and branches keyed on the application's
-  declared version. The Driver Manager maps a 2.x application onto the 3.x
-  interface before the call reaches a 3.x driver, so those paths are out of
-  scope here — a msodbcsql code path that exists solely for 2.x compatibility is
-  not a parity gap. Say so rather than porting it.
+- **This driver targets ODBC 3.x only.** That scopes out support for ODBC 2.x
+  *applications* — deprecated 2.x entry points, 2.x-only attribute values, and
+  the paths msodbcsql keeps for them. The Driver Manager maps a 2.x application
+  onto the 3.x interface before the call reaches a 3.x driver, so a msodbcsql
+  code path that exists solely for 2.x compatibility is not a parity gap. Say so
+  rather than porting it.
+  - **It does not scope out 2.x-era identifiers.** `SQL_C_DATE` / `SQL_C_TIME` /
+    `SQL_C_TIMESTAMP` are deprecated but still defined in the ODBC 3.x headers, so
+    a 3.x application may legally pass them and the DM remaps nothing. Accept
+    them and fold them onto the 3.x form (`api::type_rules::canonical_c_type`).
+    Check each identifier before assuming the rule applies — the SQL side is not
+    symmetric, ODBC 3.x reuses the 2.x date/time SQL values: `9` is both `SQL_DATE`
+    (2.x concise) and `SQL_DATETIME` (3.x verbose), and `10` is both `SQL_TIME` and
+    `SQL_INTERVAL`. A `ParameterType` of `9` is therefore ambiguous, so it is
+    rejected (`HY004`) rather than folded - 3.x applications use `SQL_TYPE_*`
+    (91-93), and the DM remaps a 2.x application's spelling first. msodbcsql
+    accepts both SQL spellings because it also serves 2.x applications and can
+    disambiguate on the declared version.
+  - **It does not cover 3.0/3.5 vs 3.8.** `SQL_OV_ODBC3` and `SQL_OV_ODBC3_80`
+    are both ODBC 3.x and both in scope. msodbcsql branches on this separately —
+    `Sql2CDefault` selects `rgbTRANSTYPE` when `IS351ORLESSAPP(wStatus)` and
+    `rgbTRANSTYPE380` otherwise — so a version-keyed branch is only out of scope
+    once you have checked *which* version boundary it keys on.
+  - Beware that msodbcsql normalizes types to their 2.x values on entry
+    (`SQLBindParameter` in `Sql/Ntdbms/sqlncli/odbc/sqlcdesc.cpp` maps
+    `SQL_TYPE_*` down to `SQL_DATE`/`SQL_TIME`/`SQL_TIMESTAMP`, and `SQL_DOUBLE`
+    to `SQL_FLOAT`, before validating). Downstream code accepting a 2.x spelling
+    therefore does **not** prove it supports 2.x applications, and a branch that
+    looks reachable in a validator may be dead once the caller's normalization is
+    accounted for. Read the caller before concluding either way.
 - Deliberate deviations (exceed-parity) are allowed with product-owner sign-off;
   record the rationale in code comments and the tracking work item.
 - Deliberate deviations are listed below:
--- a/.github/instructions/mssql-odbc.instructions.md
+++ b/.github/instructions/mssql-odbc.instructions.md
@@ -28,6 +28,14 @@ authoritative parity reference for this crate. Its source lives in the
 - When reporting a parity finding, cite the msodbcsql source (file + what it does),
   and state explicitly whether the decision **matches**, **exceeds**, or **diverges
   from** msodbcsql so the trade-off is visible.
+- **This driver targets ODBC 3.x only.** msodbcsql also serves ODBC 2.x
+  applications, so parts of its source exist purely for backward compatibility:
+  deprecated 2.x entry points, 2.x identifier spellings and attribute values,
+  2.x-era internal representations, and branches keyed on the application's
+  declared version. The Driver Manager maps a 2.x application onto the 3.x
+  interface before the call reaches a 3.x driver, so those paths are out of
+  scope here — a msodbcsql code path that exists solely for 2.x compatibility is
+  not a parity gap. Say so rather than porting it.
 - Deliberate deviations (exceed-parity) are allowed with product-owner sign-off;
   record the rationale in code comments and the tracking work item.
 - Deliberate deviations are listed below:
@@ -36,6 +44,13 @@ authoritative parity reference for this crate. Its source lives in the
     (`Sql/Ntdbms/sqlncli/msdart/inc/dlgattr.h` → `OPTIONADMSI L"ActiveDirectoryMSI"`);
     `ActiveDirectoryManagedIdentity` does not appear anywhere in the msodbcsql source.
     Added to match MS Learn and the sibling drivers (JDBC/.NET/go-sqlcmd). Tracked in AB#46066.
+  - `SQL_C_DEFAULT` in SQLBindParameter resolves the wide character SQL types to `SQL_C_WCHAR`, and
+    `SQL_GUID` to `SQL_C_GUID`, following the ODBC 3.x default-C-type table.
+    msodbcsql's `Sql2CDefault` reads `rgbTRANSTYPE380`
+    (`Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp`), which resolves both to `SQL_C_CHAR`
+    — an ANSI-transfer default this driver has no equivalent for, since its
+    `SQL_C_CHAR` is UTF-8. Resolving UTF-16 application input to a UTF-8 buffer
+    type would silently corrupt data. Tracked in AB#47365.
 
 ## No panics
 
--- a/mssql-odbc/docs/parameters_plan.md
+++ b/mssql-odbc/docs/parameters_plan.md
@@ -55,17 +55,21 @@ transparent reconnects.
   skipping string literals, quoted identifiers, and comments. It stores the
   rewritten SQL and marker count, so repeated `SQLExecute` calls do not re-scan
   the text.
-- **Bind-time type validation** - `api::type_rules` owns the `HY003` gate (every
-  real ODBC C type, including the SQL Server extensions) and classifies SQL data
-  types three ways, like msodbcsql's `IsValidSqlType`: supported, real but with
-  no SQL Server counterpart (`HYC00` - the interval types), or unknown
-  (`HY004`). `params::conversion_matrix` owns the C -> SQL conversion table,
-  shaped like msodbcsql's `fValidConversion` (one row per C type). A C type that
-  is real but not yet convertible returns `07006`, not `HY003`.
+- **Bind-time type validation** - `api::type_rules` canonicalizes the C type
+  (folding the deprecated `SQL_C_DATE` / `SQL_C_TIME` / `SQL_C_TIMESTAMP`
+  spellings onto the `SQL_C_TYPE_*` forms), then applies the `HY003` gate to that
+  canonical form, and classifies SQL data types three ways, like msodbcsql's
+  `IsValidSqlType`: supported, real but with no SQL Server counterpart (`HYC00` -
+  the interval types), or unknown (`HY004`). `params::conversion_matrix` owns the
+  C -> SQL conversion table, shaped like msodbcsql's `fValidConversion` (one row
+  per C type). A C type that is real but not yet convertible returns `07006`, not
+  `HY003`.
 - **`SQL_C_DEFAULT` resolution** - resolved at bind time to the C type implied
-  by `ParameterType` (msodbcsql `Sql2CDefault`, which reads `rgbTRANSTYPE380`
-  for ODBC 3.8+ applications) and stored resolved in `BoundParam`, so the
-  execute path never sees the placeholder.
+  by `ParameterType` and stored resolved in `BoundParam`, so the execute path
+  never sees the placeholder. Version-aware, like msodbcsql's `Sql2CDefault`,
+  which reads `rgbTRANSTYPE` for a 3.51-or-earlier application and
+  `rgbTRANSTYPE380` otherwise: `SQL_SS_TIME2` and `SQL_SS_TIMESTAMPOFFSET`
+  default to `SQL_C_BINARY` below ODBC 3.8.
 - **Value conversion** - `SQL_C_CHAR` maps to varchar and `SQL_C_WCHAR` to
   nvarchar. Indicators support `SQL_NULL_DATA`, `SQL_NTS`, and explicit byte
   length.
@@ -189,9 +193,9 @@ Pure refactor, no behavior change. Create `src/conversion/`:
 
 #### P1 - Parameter type model and conversion matrix (code complete, unmerged)
 
-- [`api/type_rules.rs`](../src/api/type_rules.rs) - the `HY003` / `HY004`
-  identifier gates and `SQL_C_DEFAULT` resolution. Direction-neutral, so it sits
-  in `api` rather than `params`.
+- [`api/type_rules.rs`](../src/api/type_rules.rs) - C-type canonicalization, the
+  `HY003` / `HY004` identifier gates, and version-aware `SQL_C_DEFAULT`
+  resolution. Direction-neutral, so it sits in `api` rather than `params`.
 - [`params/conversion_matrix.rs`](../src/params/conversion_matrix.rs) - one row
   per C type listing the SQL types it converts to. Rows today: `SQL_C_CHAR` ->
   `CHAR` / `VARCHAR` / `LONGVARCHAR`, `SQL_C_WCHAR` -> `WCHAR` / `WVARCHAR` /
@@ -204,10 +208,16 @@ bind until P3 adds its row.
 
 Deviations from msodbcsql, verified against source:
 
-- The driver is ODBC 3.x only, so the 2.x date/time SQL spellings (`SQL_DATE` /
-  `SQL_TIME` / `SQL_TIMESTAMP`, 9-11) are not accepted; the DM maps them to the
-  `SQL_TYPE_*` forms for a 3.x driver. msodbcsql accepts both because its
-  internal representation is the 2.x form.
+- ODBC 3.x reuses the 2.x date/time SQL values: `9` is both `SQL_DATE` (2.x
+  concise) and `SQL_DATETIME` (3.x verbose), and `10` is both `SQL_TIME` and
+  `SQL_INTERVAL`. A `ParameterType` of `9` is therefore ambiguous, so it is
+  rejected (`HY004`) rather than folded - 3.x applications use `SQL_TYPE_*`
+  (91-93), and the DM remaps a 2.x application's spelling first. The C side has
+  no such collision (`SQL_C_DATE` is only ever 9), so `canonical_c_type` folds
+  9-11 onto the `SQL_C_TYPE_*` forms instead of rejecting them. msodbcsql
+  accepts both SQL spellings because it also serves 2.x applications and can
+  disambiguate on the declared version, and it canonicalizes the C pair in the
+  opposite direction, toward its 2.x internal representation.
 - `SQL_C_DEFAULT` resolves the wide character types to `SQL_C_WCHAR` and
   `SQL_GUID` to `SQL_C_GUID`, following the ODBC 3.x default-C-type table.
   msodbcsql's `rgbTRANSTYPE380` resolves both to `SQL_C_CHAR`, an ANSI-transfer
@@ -216,10 +226,14 @@ Deviations from msodbcsql, verified against source:
   registered in
   [`mssql-odbc.instructions.md`](../../.github/instructions/mssql-odbc.instructions.md)
   and on AB#47365.
-- msodbcsql normalizes ODBC 3.x date/time identifiers to their 2.x values and
-  `SQL_DOUBLE` to `SQL_FLOAT` before validating; this driver accepts both
-  spellings directly. Equivalent today, but the normalization will be needed
-  when descriptors land.
+- msodbcsql normalizes types to its internal representation before validating:
+  ODBC 3.x date/time identifiers down to their 2.x values, the SS types to their
+  `*_MAPPED` ids, and `SQL_DOUBLE` to `SQL_FLOAT`. Those exist because its
+  validators and default-type tables are dense arrays indexed by the internal id;
+  this driver matches on the ODBC 3.x values directly and needs no equivalent.
+  `SQL_DOUBLE` and `SQL_FLOAT` are therefore two distinct `Supported` types here
+  that happen to resolve alike - revisit at P3/P4 if numeric matrix rows start
+  duplicating them.
 
 #### P2 - Buffer reader and outcome channel (not started)
 
--- a/mssql-odbc/docs/parameters_plan.md
+++ b/mssql-odbc/docs/parameters_plan.md
@@ -1,7 +1,7 @@
 # Parameterized execution - `SQLBindParameter` / `SQLExecute` / `SQLExecDirect`
 
 Status, behavior, and known gaps for parameterized prepared-statement execution
-in the ODBC Driver 18 (Rust). Updated 2026-08-11.
+in the ODBC Driver 18 (Rust). Updated 2026-08-18.
 
 ---
 
@@ -55,8 +55,19 @@ transparent reconnects.
   skipping string literals, quoted identifiers, and comments. It stores the
   rewritten SQL and marker count, so repeated `SQLExecute` calls do not re-scan
   the text.
-- **Types** - `SQL_C_CHAR` maps to varchar and `SQL_C_WCHAR` to nvarchar; other
-   Invalid C types return  HY003 ; unsupported C/SQL conversions return  07006. Indicators support `SQL_NULL_DATA`, `SQL_NTS`, and explicit byte length.
+- **Bind-time type validation** - `api::type_rules` owns the `HY003` gate (every
+  real ODBC C type, including the SQL Server extensions) and the `HY004` gate
+  (SQL data types). `params::conversion_matrix` owns the C -> SQL conversion
+  table, shaped like msodbcsql's `fValidConversion` (one row per C type). A C
+  type that is real but not yet convertible returns `07006`, not `HY003`.
+- **`SQL_C_DEFAULT` resolution** - resolved at bind time to the C type implied
+  by `ParameterType` (msodbcsql `Sql2CDefault`, which reads `rgbTRANSTYPE380`
+  for ODBC 3.8+ applications) and stored resolved in `BoundParam`, so the
+  execute path never sees the placeholder. Interval SQL types have no default
+  and return `07006`.
+- **Value conversion** - `SQL_C_CHAR` maps to varchar and `SQL_C_WCHAR` to
+  nvarchar. Indicators support `SQL_NULL_DATA`, `SQL_NTS`, and explicit byte
+  length.
 
 ## `mssql-tds` prepared API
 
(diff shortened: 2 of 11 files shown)


mssql-odbc/tests/e2e/tests/execute_test.cpp
```diff
@@ -150,6 +150,63 @@ TEST_F(PrepareExecuteLiveTest, ExplicitLengthWideCharParam) {
     EXPECT_SQL_OK(SQLCloseCursor(stmt_), SQL_HANDLE_STMT, stmt_);
 }
 
+// SQL_C_DEFAULT reaches the driver unresolved (the Driver Manager does not
+// substitute it), and the driver resolves it from ParameterType. SQL_VARCHAR
+// yields SQL_C_CHAR on both drivers.
+TEST_F(PrepareExecuteLiveTest, DefaultCTypeNarrowCharParam) {
```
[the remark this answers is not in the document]
Theekshna: Done in f46e7047. Added the marker to both `DefaultCTypeNarrowCharParam` and `DefaultCTypeWideCharParam`, noting that a mock TDS server could assert the RPC parameter declaration (varchar vs nvarchar) and payload encoding, which the round-tripped text cannot show.

Not adding it to `DefaultCTypeUnsupportedConversionIsRejected`: that one asserts a bind-time 07006 rejection, so no TDS traffic is generated and there is nothing for a mock to observe.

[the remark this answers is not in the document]
Theekshna: Verified against msodbcsql and fixed in bd644035 — you are right.

`IsValidSqlType` (`Sql/Ntdbms/sqlncli/odbc/sqlcprot.h`) is tri-state, not boolean:

```c
if ((fSqlType >= SQL_INTERVAL_YEAR - SQL_TYPE_MIN &&
     fSqlType <= SQL_INTERVAL_MINUTE_TO_SECOND - SQL_TYPE_MIN) || ...)
    return IDS_S1_C00;   // HYC00
...
return IDS_S1_004;       // HY004
```

and `sqlcdesc.cpp:3026` posts it before `IsValidSQLConversion` is consulted, exactly as you described.

Changes:
- `is_valid_sql_type() -> bool` is now `classify_sql_type() -> SqlTypeSupport { Supported, NotImplemented, Invalid }`, mirroring that tri-state.
- `SQLBindParameter` posts `ERR_OPTIONAL_FEATURE_NOT_IMPLEMENTED` (HYC00) for the interval range before the conversion check.
- Unit test `default_c_type_without_a_mapping_returns_07006` became `interval_sql_type_returns_hyc00`; added `interval_sql_types_are_not_implemented` in `type_rules`.
- New e2e test `IntervalSqlTypeIsRejectedWithHyc00`. No `SKIP_IF_COMPARING_MSODBCSQL()` on it: msodbcsql agrees, so it should hold on both legs of a `--compare-with-msodbcsql` run.

One related finding I did **not** act on, to keep this scoped — tell me if you want it in this PR or a separate work item. The same msodbcsql function returns `IDS_S1_C00` for `SQL_DOUBLE` as well:

```c
fSqlType == SQL_DOUBLE-SQL_TYPE_MIN || ...
    return IDS_S1_C00;
```

`SQL_DOUBLE` is not in the earlier success bitmask (only `SQL_FLOAT` and `SQL_REAL` are), so that branch is reachable. This driver still classifies `SQL_DOUBLE` as `Supported`, so it falls through to `07006` — the same class of mismatch you flagged for intervals. The `SQL_TIME` / `SQL_DATE` entries in that branch are gated on `!ISKATMAI` and so are out of scope for a modern-server driver.

mssql-odbc/docs/parameters_plan.md
```diff
@@ -87,6 +99,213 @@ transparent reconnects.
...
+  msodbcsql's `rgbTRANSTYPE380` resolves both to `SQL_C_CHAR`, an ANSI-transfer
+  artifact with no equivalent here; resolving UTF-16 input to this driver's
+  UTF-8 `SQL_C_CHAR` would silently corrupt data. Accepted deviation, also
+  registered in
+  [`mssql-odbc.instructions.md`](../../.github/instructions/mssql-odbc.instructions.md)
+  and on AB#47365.
+- msodbcsql normalizes ODBC 3.x date/time identifiers to their 2.x values and
+  `SQL_DOUBLE` to `SQL_FLOAT` before validating; this driver accepts both
+  spellings directly. Equivalent today, but the normalization will be needed
+  when descriptors land.
```
saurabh500: 🟡 **Should fix** — this contradicts the author's own review comment elsewhere in this PR, and one of the two has to be wrong.

Here the plan says msodbcsql normalizes `SQL_DOUBLE` → `SQL_FLOAT` *before* validating, therefore **"Equivalent today"**. But the reply on the `execute_test.cpp` thread says the opposite: *"`SQL_DOUBLE` is not in the earlier success bitmask (only `SQL_FLOAT` and `SQL_REAL` are), so that branch is reachable"* — i.e. msodbcsql returns `HYC00` for `SQL_DOUBLE`.

Both cannot hold:

- **If the normalization happens before `IsValidSqlType`**, the `IDS_S1_C00` branch is dead for `SQL_DOUBLE`, there is no divergence, and the review comment is a false alarm — nothing to change but the thread should be closed out.
- **If it does not**, "Equivalent today" is wrong and `SQL_DOUBLE` needs the same treatment the intervals just got in `bd644035`:

```rust
// type_rules.rs — move SQL_DOUBLE out of the Supported arm
SQL_INTERVAL_YEAR..=SQL_INTERVAL_MINUTE_TO_SECOND | SQL_DOUBLE => SqlTypeSupport::NotImplemented,
```

...plus a unit test mirroring `interval_sql_type_returns_hyc00`. Today `SQL_DOUBLE` is `Supported` (`type_rules.rs:118`) and resolves to `SQL_C_DOUBLE` (`type_rules.rs:162`), which has no matrix row, so the driver answers `07006`.

`git log -S 'SQL_DOUBLE'` shows this paragraph was written in the first commit and the contradicting comment came later, so it probably just wasn't revisited. You asked on that thread whether to fix it here or defer and never got an answer — worth settling before merge either way, since it's cheap and one artifact is definitely wrong.

saurabh500 (commented):
## Review summary

**Approve with comments** — but posting as a comment rather than an approval, because there's one open question below I'd like answered before this goes in rather than after.

No blocking defects. The refactor is behavior-preserving where it should be, the one intended behavior change is the right one, and it's covered by tests.

### What I verified

The main risk in a change like this is a silent behavior shift hiding inside a refactor, so I checked the two places that could bite:

- **`is_supported_conversion` is functionally identical to the deleted `convert::is_valid_conversion`** — same two rows, same `_ => false`. I specifically checked `SQL_C_LONG`, which the old `is_valid_c_type` admitted past the `HY003` gate: it hit `_ => false` then and fails now, so no regression.
- **The validation ladder was not reordered.** The only moved boundary is the intended one — `SQL_C_DEFAULT` now resolves before the conversion check instead of being passed through raw. The input/output check keeps its existing position.

The headline fix is real and correctly targeted: `c_type: value_type` → `c_type` at `bind_param.rs:209` is why `SQL_C_DEFAULT` no longer leaks into the execute path, and `DefaultCTypeNarrowCharParam` plus `default_c_type_is_resolved_before_storage` are genuine regression tests — both would have failed before this PR. 612/612 unit tests pass locally and CI is green.

### Should fix before merge

1. **Resolve the `SQL_DOUBLE` contradiction** (`parameters_plan.md:219-222`). The plan doc says "equivalent today"; your own reply on the `execute_test.cpp` thread says the `HYC00` branch is reachable. One of those is wrong, it's cheap to settle, and it decides whether a one-line change is needed. You asked which way to go on that thread and nobody answered — flagging it so it doesn't get lost.
2. **The `9`/`10`/`11` three-way split.** This PR adds a rule to `instructions.md` saying 2.x spellings are out of scope, then `type_rules.rs:68-70` accepts the 2.x C spellings, while `get_type_info.rs:245-250` documents the opposite rationale. Whichever is right, the three should agree.
3. **`resolve_default_c_type` is ODBC-version-blind.** `SQL_SS_TIME2` / `SQL_SS_TIMESTAMPOFFSET` defaults differ between 3.5 and 3.8, so the new 2.x rule doesn't cover it. Masked today because neither has a matrix row; goes live at P3/P4. Fix it or register it as a deviation.
4. **Test or mark the `SQL_GUID` half of the documented deviation.** It's live today and produces a hard success/failure split against msodbcsql that the e2e suite is currently silent about.

Everything else is a nit and left to your discretion.

### Follow-ups worth filing separately

- `SQLBindParameter` returns `HYC00` instead of `HY105` for an invalid `InputOutputType` value. Pre-existing — I confirmed against `main` that this PR neither introduced nor moved the check.
- Verify `SQL_C_SS_VECTOR = 0x4002` against the msodbcsql headers; no public source prints the value.
- `unsupported_conversion_returns_07006` passes `SQL_INTEGER` in the `ValueType` position and works only because `SQL_INTEGER`(4) == `SQL_C_LONG`(4). Also pre-existing, and now redundant with the `real_but_unconvertible_c_type_returns_07006` test this PR adds.
- The PR description is stale: it says intervals return `07006`, but `bd644035` changed that to `HYC00`, and it describes three e2e tests where four landed.

### What went well

The diagnosis is the good part here. The bug was subtle — `c_type: value_type` stored the unresolved placeholder, so `SQL_C_DEFAULT` + `SQL_VARCHAR` failed `07006` at bind for a pairing the driver handles perfectly well — and the fix isn't a patch at the symptom, it's a correctly-placed resolution step that guarantees the execute path never sees the placeholder again.

Modelling `SqlTypeSupport` as three states instead of a bool is the right call and is what makes the `HY004`-vs-`HYC00` distinction expressible at all; the interval range landing as a follow-up commit shows it paid off immediately. Pulling validation out of `convert.rs` left that module doing exactly one thing, and the import list shrinking by two thirds is the receipt.

Every deviation from msodbcsql is documented in three places with the source symbol, the concrete rationale, and a work item — including the UTF-8 `SQL_C_CHAR` reasoning, which is a non-obvious data-corruption trap correctly avoided. And `parameters_plan.md` is better than most design docs that get their own PR: the fetch-vs-parameter matrix section and the flagged "implemented? vs legal?" divergence to reconcile at P5 are the kind of thing only written by someone who intends to still be here for P5.

mssql-odbc/src/api/type_rules.rs
```diff
@@ -0,0 +1,250 @@
...
+            | SQL_C_LONG
+            | SQL_C_SLONG
+            | SQL_C_ULONG
+            | SQL_C_SBIGINT
+            | SQL_C_UBIGINT
+            | SQL_C_FLOAT
+            | SQL_C_DOUBLE
+            | SQL_C_DATE
+            | SQL_C_TIME
+            | SQL_C_TIMESTAMP
```
saurabh500: 🟡 **Should fix** — this accepts the ODBC 2.x C type spellings, which contradicts the rule this same PR adds to `mssql-odbc.instructions.md`.

There are now three positions in the tree on identifiers `9` / `10` / `11`:

1. This module's header (lines 13-16) says only the 3.x concise identifiers are accepted, because the DM maps a 2.x application's `SQL_DATE` / `SQL_TIME` / `SQL_TIMESTAMP` to the `SQL_TYPE_*` forms.
2. **These three lines accept `SQL_C_DATE` / `SQL_C_TIME` / `SQL_C_TIMESTAMP`** — the 2.x C spellings, which the DM remaps for exactly the same reason.
3. `api/get_type_info.rs:245-250` has a second function also named `classify_sql_type` whose doc gives the opposite rationale: both the 2.x (`9`/`10`/`11`) and 3.x (`91`/`92`/`93`) forms are accepted "so 2.x and 3.x applications are handled uniformly."

And the new standing rule added in this PR (`instructions.md:31-38`) says 2.x spellings are out of scope because the DM remaps them — which is (1), and is contradicted by (2) and (3).

I'm not asserting which answer is correct; I can't read the msodbcsql source. But whichever it is, all three should agree, and right now the policy this PR writes down is broken by the code in the same PR. One wrinkle to weigh: `SQL_DATETIME` is also `9` as the ODBC 3.x **verbose** type id, so "just reject 9" may not be the right answer on the SQL side.

mssql-odbc/src/api/type_rules.rs
```diff
@@ -0,0 +1,250 @@
...
+///
+/// Two deliberate deviations, both following the ODBC 3.x default-C-type table
+/// instead: the wide character types resolve to `SQL_C_WCHAR` and `SQL_GUID` to
+/// `SQL_C_GUID`, where msodbcsql resolves both to `SQL_C_CHAR`. That narrow
+/// default is an ANSI-transfer artifact with no equivalent here, and resolving
+/// UTF-16 input to this driver's UTF-8 `SQL_C_CHAR` would silently corrupt data.
+///
+/// Every [`SqlTypeSupport::Supported`] type has a mapping; `None` means the
+/// caller passed a type that should already have been rejected.
+pub(crate) fn resolve_default_c_type(sql_type: SqlSmallInt) -> Option<SqlSmallInt> {
```
saurabh500: 🟡 **Should fix** — this resolution is ODBC-version-blind, and two of its rows are version-dependent.

Per MS Learn ([Data Type Support for ODBC Date and Time Improvements](https://learn.microsoft.com/sql/odbc/reference/appendixes/data-type-support-for-odbc-date-and-time-improvements)), the default C type for `SQL_SS_TIME2` is `SQL_C_SS_TIME2` for ODBC 3.8+ applications but **`SQL_C_BINARY` for ODBC 3.5 and earlier**; same for `SQL_SS_TIMESTAMPOFFSET`. So lines 167-168 hardcode the 3.8 answer.

You already documented this: `parameters_plan.md:66` says `Sql2CDefault` reads `rgbTRANSTYPE380` *"for ODBC 3.8+ applications"* — the table is version-selected, but the function signature has nowhere to put the version.

Note the new 2.x rule in `instructions.md` does **not** cover this: `SQL_OV_ODBC3` vs `SQL_OV_ODBC3_80` is a 3.5-vs-3.8 split and both are ODBC 3.x. The driver already tracks the distinction and there is precedent for branching on it — `handles/env.rs` has `OdbcVersion::Odbc3_80`, and `get_type_info.rs:104-112` already does this.

```rust
pub(crate) fn resolve_default_c_type(
    sql_type: SqlSmallInt,
    odbc_version: OdbcVersion,
) -> Option<SqlSmallInt> {
    // ...
    SQL_SS_TIME2 if odbc_version == OdbcVersion::Odbc3_80 => SQL_C_SS_TIME2,
    SQL_SS_TIME2 => SQL_C_BINARY,
    SQL_SS_TIMESTAMPOFFSET if odbc_version == OdbcVersion::Odbc3_80 => SQL_C_SS_TIMESTAMPOFFSET,
    SQL_SS_TIMESTAMPOFFSET => SQL_C_BINARY,
```

It is masked today — neither C type has a conversion matrix row, so both paths land on `07006` regardless — which is exactly why it is easy to miss until a later phase adds temporal or binary rows and it silently starts mattering. Either thread the version through now, or register it in the deviations list with the phase that will fix it.

mssql-odbc/src/api/type_rules.rs
```diff
@@ -0,0 +1,250 @@
...
+/// Mirrors msodbcsql's `Sql2CDefault` (`sqlcprot.h`).
+///
+/// Two deliberate deviations, both following the ODBC 3.x default-C-type table
+/// instead: the wide character types resolve to `SQL_C_WCHAR` and `SQL_GUID` to
+/// `SQL_C_GUID`, where msodbcsql resolves both to `SQL_C_CHAR`. That narrow
+/// default is an ANSI-transfer artifact with no equivalent here, and resolving
+/// UTF-16 input to this driver's UTF-8 `SQL_C_CHAR` would silently corrupt data.
+///
+/// Every [`SqlTypeSupport::Supported`] type has a mapping; `None` means the
+/// caller passed a type that should already have been rejected.
```
saurabh500, line 184: 🔵 **Nit** — this invariant is asserted in prose but nothing enforces it, and `bind_param.rs:164` has an unreachable `None` arm that only exists because of it.

`default_c_type_follows_the_sql_type` spot-checks a handful of rows and skips `SQL_SS_TIME2`, `SQL_SS_VARIANT`, `SQL_SS_UDT`, `SQL_SS_TABLE`, and `SQL_SS_XML`. A loop over the `Supported` set would turn the doc claim into a checked invariant and make the `None` arm provably dead:

```rust
#[test]
fn every_supported_sql_type_has_a_default_c_type() {
    let candidates = (i16::MIN..=i16::MAX).filter(|&t| {
        classify_sql_type(t) == SqlTypeSupport::Supported
    });
    for sql_type in candidates {
        assert!(
            resolve_default_c_type(sql_type).is_some(),
            "SQL type {sql_type} is Supported but has no default C type",
        );
    }
}
```

(needs `#[derive(PartialEq)]` on `SqlTypeSupport`, or a `matches!`). Cheap, exhaustive, and it will fail loudly the first time someone adds a row to `classify_sql_type` and forgets the other half.

mssql-odbc/src/api/type_rules.rs
```diff
@@ -0,0 +1,250 @@
...
+    /// A SQL type this driver and SQL Server can carry.
+    Supported,
+    /// A real ODBC identifier with no SQL Server counterpart (`HYC00`).
+    NotImplemented,
+    /// Not a known ODBC SQL type identifier (`HY004`).
+    Invalid,
+}
+
+/// Classifies a `ParameterType` before any conversion check.
+pub(crate) fn classify_sql_type(sql_type: SqlSmallInt) -> SqlTypeSupport {
```
saurabh500: 🔵 **Nit** — name collision. `api/get_type_info.rs:251` already defines a different `classify_sql_type` with different semantics and a different return type. Two same-named functions in sibling modules under `api/` will make `grep` and jump-to-definition ambiguous for the next reader, especially since the two disagree about the 2.x date/time identifiers.

Suggest `classify_parameter_sql_type` here, which also reads better at the call site in `bind_param.rs`.

mssql-odbc/src/api/bind_param.rs
```diff
@@ -138,16 +141,43 @@ fn sql_bind_parameter_safe(
...
+    // matching msodbcsql, which stores the resolved type in the APD.
+    let c_type = if value_type == SQL_C_DEFAULT {
+        let Some(resolved) = resolve_default_c_type(parameter_type) else {
+            error!(
+                parameter_type,
+                "SQLBindParameter: no default C type for this SQL type"
+            );
+            post_diag(&mut stmt_state, ERR_RESTRICTED_DATA_TYPE);
+            return SQL_ERROR;
+        };
```
saurabh500: 🔵 **Nit** — this arm is unreachable. I walked all 30 `Supported` arms of `classify_sql_type` against `resolve_default_c_type` and every one has a mapping, so `None` cannot occur after the classification above.

Keeping a fail-closed arm is defensible, but `07006` is a misleading diagnostic for what is really "internal invariant violated" — if it ever does fire, the application sees a restricted-conversion error and no one will suspect the resolver. Two options:

- Add the exhaustive loop test suggested on `type_rules.rs:148` and keep this arm as belt-and-braces, or
- `debug_assert!` on the `None` case so a broken invariant is loud in debug builds instead of masquerading as a conversion failure.

Either is fine; the point is that the invariant is currently unverified in both places that depend on it.

mssql-odbc/src/params/conversion_matrix.rs
```diff
@@ -0,0 +1,79 @@
...
+// Licensed under the MIT License.
+
+//! Which C type → SQL type parameter conversions the driver can perform.
+//!
+//! Table-driven, mirroring msodbcsql's `fValidConversion` matrix
+//! (`Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp`), which is indexed by C type and
+//! yields the set of legal SQL types. The rows here list only the pairings this
+//! driver implements today; rows and entries are added as each conversion
+//! lands, so a pairing accepted at bind time is always one the execute path can
+//! actually convert.
```
saurabh500: 🔵 **Nit** — worth recording the caveat you already wrote down in `parameters_plan.md:303-307` here too: this table answers *"implemented?"* while msodbcsql's `fValidConversion` answers *"legal?"*. "Mirroring `fValidConversion`" is the wrong instruction to follow literally, and this module doc is what someone adding a row in P3/P4 will actually read.

```suggestion
//! Table-driven, in the same shape as msodbcsql's `fValidConversion` matrix
//! (`Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp`), which is indexed by C type and
//! yields the set of legal SQL types. Note the semantics differ: that matrix
//! answers "is this pairing legal?", this one answers "is this pairing
//! implemented?". The rows here list only the pairings this driver implements
//! today; rows and entries are added as each conversion lands, so a pairing
//! accepted at bind time is always one the execute path can actually convert.
//! The two must be reconciled when P5 adds cross conversions (see
//! `docs/parameters_plan.md`).
```

mssql-odbc/src/params/conversion_matrix.rs
```diff
@@ -0,0 +1,79 @@
...
+
+/// SQL types a `SQL_C_CHAR` buffer can be converted to.
+const CHAR_C_TARGETS: &[SqlSmallInt] = &[SQL_CHAR, SQL_VARCHAR, SQL_LONGVARCHAR];
+
+/// SQL types a `SQL_C_WCHAR` buffer can be converted to.
+const WCHAR_C_TARGETS: &[SqlSmallInt] = &[SQL_WCHAR, SQL_WVARCHAR, SQL_WLONGVARCHAR];
+
+/// Whether the driver can convert a `c_type` application buffer into `sql_type`
+/// for an input parameter.
+pub(crate) fn is_supported_conversion(c_type: SqlSmallInt, sql_type: SqlSmallInt) -> bool {
```
saurabh500, line 36: 🔵 **Nit** — unstated precondition worth asserting. `c_type` must already be resolved and must never be `SQL_C_DEFAULT`; if a future caller forgets to resolve, `SQL_C_DEFAULT` (99) falls into `_ => return false` and the bind fails `07006` instead of succeeding — a silent wrong answer rather than a loud one.

`mssql-odbc.instructions.md` calls for `debug_assert!` on internal contract preconditions, and this is one:

```rust
pub(crate) fn is_supported_conversion(c_type: SqlSmallInt, sql_type: SqlSmallInt) -> bool {
    debug_assert_ne!(
        c_type, SQL_C_DEFAULT,
        "SQL_C_DEFAULT must be resolved before consulting the conversion matrix"
    );
```

(needs `SQL_C_DEFAULT` adding to the import on line 22 — not offering this as a `suggestion` block since it wouldn't compile on its own.)

This is the other end of the same rope as the untested invariant on `type_rules.rs:148`.

mssql-odbc/src/params/convert.rs
```diff
@@ -4,10 +4,11 @@
...
 //!
-//! Phase 1 mirrors `SQLGetData`'s supported C types: only `SQL_C_CHAR`
-//! (→ `varchar`) and `SQL_C_WCHAR` (→ `nvarchar`). Every other C type, plus
-//! data-at-execution and default parameters, is rejected with `HYC00`; an
-//! invalid negative `StrLen_or_Ind` is rejected with `HY090`.
+//! Which C/SQL pairings reach this module is decided at bind time by
+//! [`crate::params::conversion_matrix`]; `SQL_C_DEFAULT` has already been
+//! resolved to a
+//! concrete C type by then. Data-at-execution and default parameters are
+//! rejected with `HYC00`, and an invalid negative `StrLen_or_Ind` with `HY090`.
```
saurabh500: 🔵 **Nit** — mangled line wrap ("resolved to a" / "concrete C type by then"), and the attribution is only half right: `conversion_matrix` decides which pairings are legal, but `type_rules::classify_sql_type` and `resolve_default_c_type` do the other half — and it's `resolve_default_c_type` that establishes the `SQL_C_DEFAULT` invariant this paragraph states. `rustfmt` won't reflow doc comments, so this needs a manual fix.

```suggestion
//! Which C/SQL pairings reach this module is decided at bind time by
//! [`crate::api::type_rules`] and [`crate::params::conversion_matrix`];
//! `SQL_C_DEFAULT` has already been resolved to a concrete C type by then.
//! Data-at-execution and default parameters are rejected with `HYC00`, and an
//! invalid negative `StrLen_or_Ind` with `HY090`.
```

mssql-odbc/tests/e2e/tests/execute_test.cpp
```diff
@@ -150,6 +150,85 @@ TEST_F(PrepareExecuteLiveTest, ExplicitLengthWideCharParam) {
...
+TEST_F(PrepareExecuteLiveTest, DefaultCTypeUnsupportedConversionIsRejected) {
+    SKIP_IF_COMPARING_MSODBCSQL();
+
+    SQLINTEGER value = 42;
+    SQLLEN ind = 0;
+    SQLRETURN rc = SQLBindParameter(stmt_, 1, SQL_PARAM_INPUT, SQL_C_DEFAULT,
+                                    SQL_INTEGER, 0, 0, &value, 0, &ind);
+    EXPECT_EQ(SQL_ERROR, rc);
+    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "07006");
+}
```
saurabh500, line 217: 🟡 **Should fix** — the `SQL_GUID` half of the documented deviation has no test and no `SKIP_IF_COMPARING_MSODBCSQL()` marker anywhere in the suite.

This PR registers `SQL_GUID` → `SQL_C_GUID` as an accepted deviation in three places (`type_rules.rs:143`, `instructions.md`, `parameters_plan.md:211-212`), and it is **live today and observable**:

- this driver: `SQL_C_DEFAULT` + `SQL_GUID` → `SQL_C_GUID` → no matrix row → **`07006`**
- msodbcsql: `rgbTRANSTYPE380` → `SQL_C_CHAR`, which it supports for `SQL_GUID` → **succeeds**

That is a hard pass/fail split, not a subtle payload difference — so a `--compare-with-msodbcsql` run will hit it while the suite says nothing. The wide-character half of the very same deviation got `DefaultCTypeWideCharParam`; the GUID half should get the equivalent:

```cpp
// Accepted parity deviation (AB#47365): SQL_C_DEFAULT resolves SQL_GUID to
// SQL_C_GUID per the ODBC 3.x default-C-type table, which has no conversion row
// yet, so the bind is rejected. msodbcsql's rgbTRANSTYPE380 resolves it to
// SQL_C_CHAR, which it supports, so the bind succeeds there.
TEST_F(PrepareExecuteLiveTest, DefaultCTypeGuidIsRejected) {
    SKIP_IF_COMPARING_MSODBCSQL();

    SQLGUID value = {};
    SQLLEN ind = 0;
    SQLRETURN rc = SQLBindParameter(stmt_, 1, SQL_PARAM_INPUT, SQL_C_DEFAULT,
                                    SQL_GUID, 0, 0, &value, sizeof(value), &ind);
    EXPECT_EQ(SQL_ERROR, rc);
    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "07006");
}
```

mssql-odbc/tests/e2e/tests/execute_test.cpp
```diff
@@ -150,6 +150,85 @@ TEST_F(PrepareExecuteLiveTest, ExplicitLengthWideCharParam) {
...
+    ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);
+    EXPECT_EQ("wide", GetColumnChar(1));
+
+    EXPECT_SQL_OK(SQLCloseCursor(stmt_), SQL_HANDLE_STMT, stmt_);
+}
+
+// SQL_C_DEFAULT + SQL_INTEGER resolves to SQL_C_SLONG, which has no conversion
+// row yet, so the bind is rejected up front instead of failing at execute.
+// msodbcsql supports the pairing, hence the skip. The skip will be removed soon
+// when we implement the conversion.
```
saurabh500: 🔵 **Nit** — "will be removed soon" ties a test comment to an unstated schedule. The work is already tracked as P3 in the plan doc, so point at that instead — it stays accurate and tells the reader where to look.

```suggestion
// msodbcsql supports the pairing, hence the skip. The skip goes away when P3
// (integer C to integer SQL) adds the conversion row - see
// docs/parameters_plan.md.
```

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -28,6 +28,14 @@ authoritative parity reference for this crate. Its source lives in the
...
   and state explicitly whether the decision **matches**, **exceeds**, or **diverges
   from** msodbcsql so the trade-off is visible.
+- **This driver targets ODBC 3.x only.** msodbcsql also serves ODBC 2.x
+  applications, so parts of its source exist purely for backward compatibility:
+  deprecated 2.x entry points, 2.x identifier spellings and attribute values,
+  2.x-era internal representations, and branches keyed on the application's
+  declared version. The Driver Manager maps a 2.x application onto the 3.x
+  interface before the call reaches a 3.x driver, so those paths are out of
+  scope here — a msodbcsql code path that exists solely for 2.x compatibility is
+  not a parity gap. Say so rather than porting it.
```
saurabh500: 🟡 **Should fix** — good rule, and a genuinely valuable addition: it turns a judgment call that would otherwise be re-litigated on every parity review into written policy. But **this PR breaks it twice**, so it will read as dead letter unless the code is reconciled:

1. `api/type_rules.rs:68-70` — `is_valid_c_type` accepts `SQL_C_DATE` (9) / `SQL_C_TIME` (10) / `SQL_C_TIMESTAMP` (11), the 2.x C spellings, which the DM remaps for exactly the reason stated here.
2. `api/get_type_info.rs:245-250` (pre-existing, untouched by this PR) documents the **opposite** rationale — it accepts `9`/`10`/`11` "so 2.x and 3.x applications are handled uniformly."

`type_rules.rs`'s own module header follows this new rule for the SQL-side spellings, so the tree ends up holding three positions on one question. Worth either reconciling the code or adding a carve-out here explaining why the C-type side differs — one caveat being that `SQL_DATETIME` is also `9` as the ODBC 3.x verbose type id, so the SQL side may not be a simple rejection.

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -36,6 +44,13 @@ authoritative parity reference for this crate. Its source lives in the
     (`Sql/Ntdbms/sqlncli/msdart/inc/dlgattr.h` → `OPTIONADMSI L"ActiveDirectoryMSI"`);
     `ActiveDirectoryManagedIdentity` does not appear anywhere in the msodbcsql source.
     Added to match MS Learn and the sibling drivers (JDBC/.NET/go-sqlcmd). Tracked in AB#46066.
+  - `SQL_C_DEFAULT` in SQLBindParameter resolves the wide character SQL types to `SQL_C_WCHAR`, and
+    `SQL_GUID` to `SQL_C_GUID`, following the ODBC 3.x default-C-type table.
+    msodbcsql's `Sql2CDefault` reads `rgbTRANSTYPE380`
+    (`Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp`), which resolves both to `SQL_C_CHAR`
+    — an ANSI-transfer default this driver has no equivalent for, since its
+    `SQL_C_CHAR` is UTF-8. Resolving UTF-16 application input to a UTF-8 buffer
+    type would silently corrupt data. Tracked in AB#47365.
```
saurabh500, line 75: 🟡 **Should fix** — a third deviation belongs in this list: `resolve_default_c_type` ignores the application's declared ODBC version.

`SQL_SS_TIME2` and `SQL_SS_TIMESTAMPOFFSET` have version-dependent defaults — `SQL_C_SS_TIME2` / `SQL_C_SS_TIMESTAMPOFFSET` for ODBC 3.8+, but **`SQL_C_BINARY` for 3.5 and earlier** ([MS Learn](https://learn.microsoft.com/sql/odbc/reference/appendixes/data-type-support-for-odbc-date-and-time-improvements)). `type_rules.rs:167-168` hardcodes the 3.8 answer.

The new 2.x rule directly above does not dispose of this: `SQL_OV_ODBC3` vs `SQL_OV_ODBC3_80` is a 3.5-vs-3.8 branch and both are ODBC 3.x. The driver already tracks it (`handles/env.rs` `OdbcVersion::Odbc3_80`, used by `get_type_info.rs:104-112`), so this is a gap rather than a scoping decision.

It is invisible today because neither C type has a conversion matrix row, so both answers end at `07006` — it goes live when P3/P4 add temporal or binary rows. Either fix `resolve_default_c_type` or record it here with the phase that will.

mssql-odbc/src/api/odbc_types.rs
```diff
@@ -339,6 +341,15 @@ pub const SQL_C_TYPE_TIMESTAMP: SqlSmallInt = 93;
 pub const SQL_C_TYPES_EXTENDED: SqlSmallInt = 0x4000; // 16384
 pub const SQL_C_SS_TIME2: SqlSmallInt = SQL_C_TYPES_EXTENDED; // 0x4000
 pub const SQL_C_SS_TIMESTAMPOFFSET: SqlSmallInt = SQL_C_TYPES_EXTENDED + 1; // 0x4001
+pub const SQL_C_SS_VECTOR: SqlSmallInt = SQL_C_TYPES_EXTENDED + 2; // 0x4002
```
saurabh500, line 344: 🔵 **Nit** — I could not verify `0x4002` against any public source. MS Learn names `SQL_C_SS_VECTOR` but never prints its value, and unlike `SQL_SS_VECTOR = -156` (which Learn does state, and which correctly extends the contiguous `-150..-156` SS block) this one is inferred from the `SQL_C_TYPES_EXTENDED + 2` pattern.

The failure mode is two-sided and silent: if the real value differs, `is_valid_c_type` rejects a legal `SQL_C_SS_VECTOR` binding with `HY003`, and accepts whatever type actually holds `0x4002`. Worth a one-line confirmation against the msodbcsql headers, and a comment citing the header the way lines 340-341 cite the `0x4000` base.

mssql-odbc/docs/parameters_plan.md
Theekshna: Retracting my earlier comment — the doc is right, I was wrong. `SQLBindParameter` in `sqlcdesc.cpp` normalizes `SQL_DOUBLE` to `SQL_FLOAT` *before* calling `IsValidSqlType`, so that `IDS_S1_C00` arm is dead on the bind path and "Equivalent today" holds. I'd read the validator in isolation without checking its caller. No code change; I've rewritten the neighbouring bullet (which had gone stale) to record why msodbcsql normalizes — its tables are dense arrays indexed by the internal id — and why we need no equivalent. I also added a standing note to `instructions.md` so the next reviewer doesn't repeat the mistake.

mssql-odbc/src/api/type_rules.rs
Theekshna: Good catch. The rule conflated an ODBC 2.x **application** (out of scope, DM handles it) with a 2.x-era **identifier** — `SQL_C_DATE` is deprecated, not removed, so a 3.x app may legally pass it and nothing remaps it. Rejecting would have been wrong, so I normalize instead: new `canonical_c_type` folds 9/10/11 onto 91/92/93 at bind entry, and `is_valid_c_type` now accepts only the canonical form behind a `debug_assert`. That mirrors msodbcsql, which folds the same pair the other way and whose `IsValidCType` bounds its positive range at `fCType <= 32` — i.e. it too validates one form only. Your `SQL_DATETIME` caveat is confirmed and is exactly why the SQL side isn't symmetric. One known limit: the assert is debug-only, so a future `SQLBindCol` that skips the fold would return `HY003` in release; a `CanonicalCType` newtype would close that if you think it's worth the machinery.

mssql-odbc/src/api/type_rules.rs
Theekshna: Fixed rather than deferred, and confirmed from source rather than MS Learn: `Sql2CDefault` (`sqlcprot.h`) selects `rgbTRANSTYPE` when `IS351ORLESSAPP`, and the comment above `rgbTRANSTYPE380` in `sqlcmisc.cpp` states the two tables differ only in the `SQL_TIME2` / `SQL_TIMESTAMPOFFSET` rows, which are `SQL_C_BINARY` below 3.8. `resolve_default_c_type` now takes `OdbcVersion`, with a test covering both sides. `bind_param` reads the version from the env *before* taking the stmt lock, since doing it inside would violate the parent-before-child ordering rule. You were also right that my rule text mis-scoped this as 2.x baggage — `instructions.md` now says 3.51-vs-3.8 is in scope. The plan doc's `SQL_C_DEFAULT` bullet is updated to match, since it previously described only the 3.8 table.

mssql-odbc/src/api/type_rules.rs
Theekshna: Added as `every_supported_sql_type_has_a_default_c_type`, looping the full `i16` range across all four `OdbcVersion` values since the resolver is now version-aware. `SqlTypeSupport` already derived `PartialEq`.

mssql-odbc/src/api/type_rules.rs
Theekshna: Renamed to `classify_parameter_sql_type`. I also checked `get_type_info::classify_sql_type` and it needs no change — msodbcsql's `SQLGetTypeInfo` path maps 91-93 down then validates, so raw 9/10/11 are legal there and its "handled uniformly" rationale is correct.

mssql-odbc/src/api/bind_param.rs
Theekshna: Took both options since they reinforce each other: the exhaustive test from your `type_rules.rs:149` comment now proves the invariant, and a `debug_assert!(false, ...)` on the `None` arm makes a violation loud in debug instead of masquerading as `07006`. The fail-closed return stays for release builds.

mssql-odbc/src/params/conversion_matrix.rs
Theekshna: Applied, including the P5 reconciliation pointer. The module doc now leads with the semantic difference rather than "mirroring `fValidConversion`", which as you say is the wrong instruction to follow literally.

mssql-odbc/src/params/conversion_matrix.rs
Theekshna: Applied, with `SQL_C_DEFAULT` added to the import. There's now a matching precondition assert in `is_valid_c_type` for canonicalization, so both ends of the rope are tied.

mssql-odbc/src/params/convert.rs
Theekshna: Applied verbatim.

mssql-odbc/tests/e2e/tests/execute_test.cpp
Theekshna: Added `DefaultCTypeGuidIsRejected` essentially as written. One correction to the framing: a `--compare-with-msodbcsql` run wouldn't have hit this, since no e2e test bound `SQL_C_DEFAULT` + `SQL_GUID` at all — it was a coverage gap rather than a latent failure, and the pairing was already pinned at unit level. Still worth adding for symmetry with the wide-char half.

mssql-odbc/tests/e2e/tests/execute_test.cpp
Theekshna: Applied verbatim.

.github/instructions/mssql-odbc.instructions.md
Theekshna: Agreed, reconciled in both directions. The rule now scopes out 2.x **applications** rather than 2.x-era **identifiers**, with a sub-bullet on the asymmetry you flagged: the C spellings are deprecated-but-legal and get folded, while on the SQL side ODBC 3.x reuses the values (`9` is both `SQL_DATE` and `SQL_DATETIME`, `10` both `SQL_TIME` and `SQL_INTERVAL`), so a `ParameterType` of `9` is ambiguous and stays rejected. Code reconciled to match; `get_type_info.rs` needed no change. I also added the warning about msodbcsql normalizing on entry — the trap that produced my own wrong `SQL_DOUBLE` claim elsewhere in this PR.

.github/instructions/mssql-odbc.instructions.md
Theekshna: Took the "fix it" branch rather than "record it", so there's no deviation entry to add — `resolve_default_c_type` is now version-aware. Details and source evidence on the `type_rules.rs:150` thread.

mssql-odbc/src/api/odbc_types.rs
Theekshna: Verified — `0x4002` is correct. msodbcsql's public `msodbcsql.h` defines `SQL_C_TYPES_EXTENDED 0x04000L` with `SQL_C_SS_VECTOR` at `+2`. Citation added to the comment. Fair flag: it had been inferred from the pattern rather than confirmed.

saurabh500 (approved):
## Re-review — `24d4df17`

All 14 findings from my earlier comment-only review are fixed. I verified each at source level rather than taking the replies at face value.

| # | Fix | Verified how |
|---|---|---|
| 1 | `SQL_DOUBLE` doc — author **retracted their own** earlier claim | `sqlcdesc.cpp` normalizes `SQL_DOUBLE` → `SQL_FLOAT` before `IsValidSqlType`, so the arm is dead on the bind path |
| 2 | `canonical_c_type` folds 9/10/11 → 91/92/93 | Offset is uniform (82) across all three pairs; range `9..=11` is contiguous with no other C type in it |
| 3 | `resolve_default_c_type` now version-aware | Guard arms `SQL_SS_TIME2 if is_3_80`; env read is scoped and released *before* the stmt lock, so the two are never held together and no inversion is possible |
| 4 | Exhaustive default-C-type test | Loops the full `i16` range × 4 `OdbcVersion` values |
| 5 | Renamed `classify_parameter_sql_type` | No stale call sites; exactly one production caller |
| 6 | `debug_assert!(false)` + fail-closed | Release path still returns `SQL_ERROR` |
| 7–9, 11, 14 | Doc and assert nits applied | `SQL_C_TYPES_EXTENDED 0x04000L` citation added |
| 10 | `DefaultCTypeGuidIsRejected` added | Correctly gated by `SKIP_IF_COMPARING_MSODBCSQL()` |
| 12–13 | Instructions rewritten | Distinguishes 2.x *applications* (out of scope) from 2.x *identifiers* (in scope), with the SQL-side ambiguity spelled out |

**Highest-risk item cleared.** `canonical_c_type` is applied at `bind_param.rs:148`, ahead of every consumer — `is_valid_c_type` (152), the conversion matrix, and the `c_type` stored on `BoundParam`. There is exactly one production caller of each, so the new `debug_assert_eq!` cannot fire; 616 debug-profile unit tests (up from 612) confirm that empirically. All CI legs green.

**Two places where the author was right and I was wrong**, worth recording:

- The normalize-don't-reject fix for the deprecated C spellings is better than what I suggested. `SQL_C_DATE` is deprecated-but-legal for a 3.x application and nothing remaps it, so folding is correct where rejecting would not have been.
- The `SQL_GUID` framing correction: no e2e test bound that pairing at all, so it was a coverage gap, not a latent `--compare` failure.

Also worth crediting the `SQL_DOUBLE` self-retraction — disproving your own earlier claim by reading the caller rather than the validator in isolation, then leaving a standing note in the instructions so the next reviewer doesn't repeat it.

**One minor observation, non-blocking.** `SQLBindParameter` now takes the env mutex on every call. `SQL_ATTR_ODBC_VERSION` is immutable once connections exist, so it could be cached on the dbc or stmt at allocation time and skip the lock entirely. It is uncontended, so this is a nit rather than something worth a round trip.

### Verdict: Approve

Up from *Approve with comments*. All four 🟡 and every 🔵 are fixed, there were no 🔴 at any point, and the new commit introduces no regressions. Two items I would have accepted as follow-ups — version-awareness and the instructions deviation entry — were fixed outright instead of deferred.

### What went well

The replies went back to the msodbcsql source and either confirmed or corrected the reasoning rather than just applying the diff: the `SQL_DOUBLE` normalization, the `IS351ORLESSAPP` table split, the `0x04000L` base value. Two prose invariants became checked tests, the unreachable arm is now loud in debug instead of masquerading as `07006`, and the instructions rule was reframed around the application-vs-identifier distinction that was the actual root cause. That is fixing the class rather than the instance, and it makes the next parity review cheaper.

PR #324, opened 2026-08-18
Enhancing macOS SQL diagnostics

subrata-ms:
## Description
This pull request improves the reliability and debuggability of the SQL Server setup and validation steps in the CI pipeline, particularly for macOS environments. The main changes include fixing environment variable usage, enhancing readiness checks, and adding diagnostic scripts to help troubleshoot SQL Server startup issues on macOS.

**Improvements to SQL Server Docker setup:**

* Fixed environment variable syntax for `MSSQL_SA_PASSWORD` and improved readiness check logic in `.pipeline/templates/sql-setup-template.yml` to better handle SQL Server startup failures and provide detailed diagnostics if the server fails to become ready.
* The readiness change in [sql-setup-template.yml:280] fixes a real bug: previously, 30 failed attempts could still let setup succeed.
* sqlcmd -b, explicit readiness state, safe environment-variable usage, and startup logs are solid improvements.
* The pre-test probe catches Colima/SQL dying during dependency installation.
* Failure-only container logs are useful and do not expose the password.

**Enhancements for macOS validation and diagnostics:**

* Added a script before tests in `.pipeline/templates/validation-stages.yml` to verify the SQL Server container is running and accessible on macOS, including checks for container status, SQL command execution, and port availability.
* Added a diagnostic script after the build step (conditional on failure) to output detailed information about Colima, Docker containers, SQL Server state, logs, and host port usage to aid in troubleshooting failures on macOS.

## Related Issues

ADO work item: https://dev.azure.com/sqlclientdrivers/mssql-python/_sprints/taskboard/mssql-python%20Team/mssql-python/Rubidium/Aug%202026?workitem=47383
## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

### DIFF
--- a/.pipeline/templates/sql-setup-template.yml
+++ b/.pipeline/templates/sql-setup-template.yml
@@ -292,6 +292,7 @@ steps:
 
       if [ "$ready" != true ]; then
         echo "SQL Server failed readiness checks"
+        set +e
         docker ps -a
         docker inspect sqlserver --format \
           'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}'
--- a/.pipeline/templates/sql-setup-template.yml
+++ b/.pipeline/templates/sql-setup-template.yml
@@ -272,19 +272,32 @@ steps:
       docker run \
         --name sqlserver \
         -e ACCEPT_EULA=Y \
-        -e "MSSQL_SA_PASSWORD=$(SQL_PASSWORD)" \
+        -e "MSSQL_SA_PASSWORD=$SQL_PASSWORD" \
         -p 1433:1433 \
         -d mcr.microsoft.com/mssql/server:2025-latest
 
       echo "Waiting for SQL Server to start..."
+      ready=false
       for i in $(seq 1 30); do
-        docker exec sqlserver \
+        if docker exec sqlserver \
           /opt/mssql-tools18/bin/sqlcmd \
-          -S localhost -U SA -P "$(SQL_PASSWORD)" \
-          -C -Q "SELECT 1" 2>/dev/null && break
+          -S localhost -U SA -P "$SQL_PASSWORD" \
+          -C -b -Q "SELECT 1" 2>/dev/null; then
+          ready=true
+          break
+        fi
         echo "  attempt $i/30..."
         sleep 2
       done
+
+      if [ "$ready" != true ]; then
+        echo "SQL Server failed readiness checks"
+        docker ps -a
+        docker inspect sqlserver --format \
+          'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}'
+        docker logs --tail 500 sqlserver
+        exit 1
+      fi
     displayName: 'Start SQL Server (Docker, no TLS certs)'
     env:
       SQL_PASSWORD: $(SQL_PASSWORD)
--- a/.pipeline/templates/validation-stages.yml
+++ b/.pipeline/templates/validation-stages.yml
@@ -382,7 +382,7 @@ stages:
             /opt/mssql-tools18/bin/sqlcmd \
             -S localhost -U SA -P "$SQL_PASSWORD" \
             -C -b -Q "SELECT 1"
-          nc -zv localhost 1433
+          nc -z -v 127.0.0.1 1433
         displayName: 'Verify macOS SQL Server before tests'
         env:
           SQL_PASSWORD: $(SQL_PASSWORD)
--- a/.pipeline/templates/validation-stages.yml
+++ b/.pipeline/templates/validation-stages.yml
@@ -383,7 +383,7 @@ stages:
             -S localhost -U SA -P "$SQL_PASSWORD" \
             -C -b -Q "SELECT 1"
           nc -zv localhost 1433
-        displayName: 'Temporary: verify macOS SQL Server before tests'
+        displayName: 'Verify macOS SQL Server before tests'
         env:
           SQL_PASSWORD: $(SQL_PASSWORD)
       - template: build-template.yml
@@ -405,7 +405,7 @@ stages:
           docker logs --tail 500 sqlserver
           echo "=== Host port 1433 ==="
           lsof -nP -iTCP:1433 -sTCP:LISTEN
-        displayName: 'Temporary: diagnose macOS SQL Server failure'
+        displayName: 'Diagnose macOS SQL Server failure'
         condition: failed()
 
 # Kerberos tests for PRs (fast - 2 distros: ubuntu22 + alpine318)
--- a/.pipeline/templates/validation-stages.yml
+++ b/.pipeline/templates/validation-stages.yml
@@ -395,16 +395,17 @@ stages:
       - script: |
           set +e
           echo "=== Colima status ==="
-          colima status
+          colima status || true
           echo "=== Docker containers ==="
-          docker ps -a
+          docker ps -a || true
           echo "=== SQL Server state ==="
           docker inspect sqlserver --format \
-            'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}'
+            'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}' || true
           echo "=== SQL Server logs ==="
-          docker logs --tail 500 sqlserver
+          docker logs --tail 500 sqlserver || true
           echo "=== Host port 1433 ==="
-          lsof -nP -iTCP:1433 -sTCP:LISTEN
+          lsof -nP -iTCP:1433 -sTCP:LISTEN || true
+          exit 0
         displayName: 'Diagnose macOS SQL Server failure'
         condition: or(failed(), canceled())
 
--- a/.pipeline/templates/validation-stages.yml
+++ b/.pipeline/templates/validation-stages.yml
@@ -406,7 +406,7 @@ stages:
           echo "=== Host port 1433 ==="
           lsof -nP -iTCP:1433 -sTCP:LISTEN
         displayName: 'Diagnose macOS SQL Server failure'
-        condition: failed()
+        condition: or(failed(), canceled())
 
 # Kerberos tests for PRs (fast - 2 distros: ubuntu22 + alpine318)
 - stage: Kerberos_Test_PR
--- a/.pipeline/templates/validation-stages.yml
+++ b/.pipeline/templates/validation-stages.yml
@@ -372,12 +372,41 @@ stages:
       - template: install-dependencies.yml
         parameters:
           osType: MacOS
+      - script: |
+          set -euo pipefail
+          colima status
+          docker ps -a
+          docker inspect sqlserver --format \
+            'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}'
+          docker exec sqlserver \
+            /opt/mssql-tools18/bin/sqlcmd \
+            -S localhost -U SA -P "$SQL_PASSWORD" \
+            -C -b -Q "SELECT 1"
+          nc -zv localhost 1433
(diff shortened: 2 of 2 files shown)


.pipeline/templates/validation-stages.yml
```diff
@@ -372,12 +372,41 @@ stages:
...
+          colima status
+          echo "=== Docker containers ==="
+          docker ps -a
+          echo "=== SQL Server state ==="
+          docker inspect sqlserver --format \
+            'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}'
+          echo "=== SQL Server logs ==="
+          docker logs --tail 500 sqlserver
+          echo "=== Host port 1433 ==="
+          lsof -nP -iTCP:1433 -sTCP:LISTEN
```
saurabh500: This diagnostic step will itself report as failed in exactly the scenario it exists to diagnose.

With `set +e` the script's exit code is the exit code of the *last* command, and `lsof` exits `1` when it finds no matching listener. So when SQL Server is not listening on 1433 — the primary failure mode here — this step exits 1 and shows up as a second red X, which is a red herring next to the actual test failure. It also means this step cannot safely be switched to `always()` later.

The rest of the repo handles collection steps with `|| true` per command (see `kerberos-test-template.yml:328-335`). Suggest matching that convention and ending with an explicit success:

```suggestion
      - script: |
          set +e
          echo "=== Colima status ==="
          colima status || true
          echo "=== Docker containers ==="
          docker ps -a || true
          echo "=== SQL Server state ==="
          docker inspect sqlserver --format \
            'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}' || true
          echo "=== SQL Server logs ==="
          docker logs --tail 500 sqlserver || true
          echo "=== Host port 1433 ==="
          lsof -nP -iTCP:1433 -sTCP:LISTEN || true
          exit 0
```

.pipeline/templates/validation-stages.yml
```diff
@@ -372,12 +372,41 @@ stages:
...
+        displayName: 'Diagnose macOS SQL Server failure'
+        condition: failed()
```
saurabh500: `failed()` will not fire when the job is cancelled by a server-side timeout — and a hung SQL/Colima connection is one of the more likely ways this job dies. This repo already documents that exact gotcha in `test-mssql-python-odbc-template.yml:39` ("a job cancelled by the server skips `condition: always()` steps").

Since the whole point of this PR is capturing diagnostics for macOS SQL flakiness, widening the condition means timeouts are not a blind spot:

```suggestion
        condition: or(failed(), canceled())
```

Apply this together with the `exit 0` fix above — otherwise the step would start reporting failures on cancel paths too.

.pipeline/templates/sql-setup-template.yml
```diff
@@ -272,19 +272,32 @@ steps:
...
         sleep 2
       done
+
+      if [ "$ready" != true ]; then
+        echo "SQL Server failed readiness checks"
+        docker ps -a
+        docker inspect sqlserver --format \
+          'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}'
+        docker logs --tail 500 sqlserver
+        exit 1
```
saurabh500, line 300: This diagnostic block runs with `set -e` still active from line 270, so the first failing command aborts the script and everything after it is lost.

Concretely: if the Colima VM has died — a leading suspect when SQL never becomes ready — `docker ps -a` fails, the script exits right there, and `docker logs --tail 500 sqlserver` (the single most useful artifact) never runs. The step still fails, but with the least informative output possible.

You already got this right in the `validation-stages.yml` diagnose step by using `set +e`. Mirroring it here; the explicit `exit 1` still propagates the failure:

```suggestion
      if [ "$ready" != true ]; then
        echo "SQL Server failed readiness checks"
        set +e
        docker ps -a
        docker inspect sqlserver --format \
          'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}'
        docker logs --tail 500 sqlserver
        exit 1
```

.pipeline/templates/sql-setup-template.yml
```diff
@@ -272,19 +272,32 @@ steps:
...
+      fi
     displayName: 'Start SQL Server (Docker, no TLS certs)'
     env:
       SQL_PASSWORD: $(SQL_PASSWORD)
```
saurabh500, line 304: Nit: this file still has no newline at end of file, which is why the diff shows `\ No newline at end of file`. Since this PR is already rewriting the tail of the step, worth fixing so future diffs here do not carry the marker.

The suggestion below re-emits line 303 unchanged; applying it just appends the missing trailing newline.

```suggestion
      SQL_PASSWORD: $(SQL_PASSWORD)
```

saurabh500 (commented):
Pipeline-only change that hardens macOS SQL Server startup and adds failure diagnostics. The core changes are correct and fix a real bug — happy to see this merge once the diagnostic-robustness points below are addressed. Nothing here is blocking.

### Verified as genuine improvements

- **Real bug fix.** The old readiness loop used `docker exec ... && break` with no result tracking, so 30 consecutive failures fell through and setup still reported success. The `ready` flag plus `exit 1` closes that hole.
- **`sqlcmd -b`** is required for the readiness check to mean anything — without it sqlcmd exits 0 even when the batch fails. Same rationale is already documented at `sql-setup-template.yml:30-33`.
- **`$(SQL_PASSWORD)` → `$SQL_PASSWORD`** is correct and security-positive. I confirmed `SQL_PASSWORD` is a job-scoped **secret** (`generate-sql-password-template.yml`, `issecret=true`), and the `env:` mapping is present in both steps, which is the documented way to consume secrets.
- **No password leak.** The narrow `docker inspect --format` avoids dumping `Config.Env`, which holds `MSSQL_SA_PASSWORD`; a bare `docker inspect` would have exposed it. Nice detail.
- **Ordering is sound.** `SQL_PASSWORD` is available to the new step in `validation-stages.yml` because `sql-setup-template.yml` runs first in `Test_MacOS` (line 366).

### Blocking

None.

### Suggestions

Three inline suggestions below, all one-click applyable:

1. `validation-stages.yml:395-407` — the diagnose step exits non-zero in exactly the case it diagnoses (`lsof` returns 1 when nothing is listening).
2. `validation-stages.yml:409` — `condition: failed()` misses server-side cancellation, so a hung job loses all diagnostics.
3. `sql-setup-template.yml:293-299` — the failure block inherits `set -e`, so a dead Colima VM aborts before `docker logs` runs.

### Nits not expressible inline

- **PR description is stale.** It calls the macOS steps "temporary" twice, but commit `8d6683a0` is *"Make macOS SQL diagnostics permanent."* Worth rewording so the intent is clear to whoever reads this in six months.
- **Please drop the `vscode-file://vscode-app/c:/Users/spaitandi/...` link** from the description — it is broken for every reader and leaks a local path and username. The `sql-setup-template.yml:280` reference on its own is fine.
- **Title typo:** "Enhanching" → "Enhancing".
- **Pre-existing inconsistency, out of scope but cheap while you are here:** `sql-setup-template.yml:242` (the Linux path) still passes `MSSQL_SA_PASSWORD=$(SQL_PASSWORD)` via agent macro expansion rather than the `env:`-mapped `$SQL_PASSWORD` you just adopted for macOS. Aligning it would make the file consistent. I did not comment inline because the line is outside this PR's diff.
- The `docker inspect --format 'status=... exit=... oom=... error=...'` line now appears three times across two files. Not worth blocking on, but if it grows a fourth copy a small shared step template would pay for itself.
- The checklist in the description is fully unchecked. `cargo bfmt` / `bclippy` / `btest` are arguably N/A for a YAML-only change, but the repo PR workflow asks for the checklist to be kept current — consider ticking or striking them.

.pipeline/templates/validation-stages.yml
```diff
@@ -372,12 +372,42 @@ stages:
...
+          set -euo pipefail
+          colima status
+          docker ps -a
+          docker inspect sqlserver --format \
+            'status={{.State.Status}} exit={{.State.ExitCode}} oom={{.State.OOMKilled}} error={{.State.Error}}'
+          docker exec sqlserver \
+            /opt/mssql-tools18/bin/sqlcmd \
+            -S localhost -U SA -P "$SQL_PASSWORD" \
+            -C -b -Q "SELECT 1"
+          nc -zv localhost 1433
```
saurabh500: This is the one thing I would still change before merge.

`docker exec sqlserver ... sqlcmd -S localhost` on line 383 is correct — that `localhost` is resolved **inside** the container. But line 385 runs on the **host**, and on macOS `localhost` resolves to both `::1` and `127.0.0.1` while Colima publishes forwarded ports on IPv4 only. Resolution order is not guaranteed to favour the address that is actually listening.

This matters more than a normal nit because the step runs under `set -euo pipefail` and `nc` is the last command, so a spurious non-zero here fails the whole macOS job before a single test runs — turning a diagnostic aid into a source of false red builds.

Two pieces of precedent in this repo point the same way:

- `test-mssql-python-macos-template.yml` is the existing Colima-on-macOS template and draws exactly this distinction: `localhost` for the in-container `docker exec` probe (line 128), but `127.0.0.1` for every host-side value (`DB_CONNECTION_STRING`, `DB_HOST`, `SQL_SERVER` — lines 146-149).
- `mssql-tds/tests/timeout_and_cancel.rs:58` carries the comment *"Use 127.0.0.1 directly to match the listener binding - localhost resolves to hostname"*, so this class of problem has already bitten the codebase once.

Pinning to IPv4 removes the ambiguity:

```suggestion
          nc -z -v 127.0.0.1 1433
```

If you would rather keep the check tolerant instead of authoritative, `nc -z -v 127.0.0.1 1433 || echo "WARN: host port 1433 not reachable"` also works and keeps the gate focused on the sqlcmd result above.

saurabh500 (commented):
Re-reviewed at `3ce12663`. All three suggestions from my previous pass were applied correctly — I verified each against the current file state, and the resulting shell semantics are right. One new finding below, plus a couple of carry-over nits.

### Verified fixed

| Previous finding | Commit | Status |
|---|---|---|
| Diagnose step self-failed on `lsof` exit 1 | `c7c4ed83` / `3ce12663` | Fixed — `\|\| true` on all five collectors plus trailing `exit 0` |
| `condition: failed()` missed timeout cancellation | `3ce12663` | Fixed — now `or(failed(), canceled())` |
| Failure block inherited `set -e`, aborting before `docker logs` | `a8eab846` | Fixed — `set +e` added at `sql-setup-template.yml:295`, and the explicit `exit 1` at line 300 still propagates the failure correctly |

I also confirmed both files still parse as valid YAML, and that `exit 0` sits inside the script block rather than terminating the step definition.

### New finding

One suggestion inline at `validation-stages.yml:385` about `nc -zv localhost 1433` using `localhost` for a host-side check. Worth attention because this step is a hard gate under `set -euo pipefail`, so a false negative fails the entire macOS job rather than just producing noise. Details and repo precedent are in the inline comment.

### Still open

- **Trailing newline** at `sql-setup-template.yml:304` — the earlier suggestion was not applied, so the diff still carries `\ No newline at end of file`. Pure nit; ignore if you would rather not touch the line again.
- **Title typo:** "Enhanching" → "Enhancing".
- **Description is still stale.** It describes the macOS steps as "temporary" in two places, but commit `8d6683a0` is *"Make macOS SQL diagnostics permanent"*, and the steps are now unconditional. It also still carries the `vscode-file://vscode-app/c:/Users/spaitandi/...` link, which is broken for every reader and leaks a local path and username. Both are worth a quick edit before merge since the description is what future readers will land on.
- The checklist is still fully unchecked. `cargo bfmt` / `bclippy` / `btest` are arguably N/A for a YAML-only change, but the repo PR workflow asks for it to be kept current.

Nothing blocking. Once the `nc` question is resolved one way or the other, this looks good to me.

saurabh500 (approved):

sumitmsft (approved):

PR #325, opened 2026-08-19
mssql-odbc: Implement catalog functions (SQLTables/Columns/PrimaryKeys/ForeignKeys/SpecialColumns/Statistics/Procedures)

Vahid-b:
## Description

Implements the seven ODBC catalog functions tracked under [AB#46380](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46380): `SQLTables`, `SQLColumns`, `SQLPrimaryKeys`, `SQLForeignKeys`, `SQLSpecialColumns`, `SQLStatistics`, `SQLProcedures` (W variants).

Each dispatches via RPC to the matching SQL Server system stored procedure — `sp_tables`, `sp_columns_100`, `sp_pkeys`, `sp_fkeys`, `sp_special_columns_100`, `sp_statistics_100`, `sp_stored_procedures` — reusing the ordinary execute path so cursor/metadata state is managed exactly as for a user query. Design was verified line-by-line against the msodbcsql C++ source (`sqlcdd.cpp`), not just the ODBC spec:

- **Column parity**: renames the ODBC 2.x column names these procedures emit (`TABLE_QUALIFIER`, `TABLE_OWNER`, ...) to their ODBC 3.x equivalents and clears the NOT NULL flags the spec mandates, matching msodbcsql's `SetColNames`/`ClearNullable` exactly (verified per-function against `sqlcdd.cpp`'s `fNewNameCols`/`fNotNullCols` bitmasks).
- **Catalog qualification**: a supplied catalog scopes the call to that database via a three-part qualified procedure name (`[db].sys.proc`), since SQL Server's system catalog procs are visible in every database via the Resource database — no `USE` statement needed.
- **Catalog enumeration**: `SQLTables(catalog="%", schema="", table="")` — the ODBC-mandated idiom every catalog-browsing tool relies on — is recognized universally (all 7 functions, matching msodbcsql `DoDD()`) and skips three-part qualification so the stored procedure's own `@table_qualifier='%'` handling runs, instead of being misread as a literal (nonexistent) catalog name.
- **Nonexistent catalog**: a real, non-"%" catalog that doesn't resolve yields an empty, correctly-shaped result set instead of an error, matching msodbcsql's retry mechanism (`DoDD()`) — the qualified call is retried unqualified with an unmatchable name filter rather than pre-checking `DB_ID`.
- **SQLForeignKeys**: both sides must resolve to the same database; if the caller supplies conflicting PK/FK catalogs, the result set is forced empty rather than guessing.
- **SQLSpecialColumns**: `SQL_BEST_ROWID` with a wider-than-current-row scope forces an empty result unless the connection is in a serializable transaction, per the ODBC spec.
- **SQLStatistics**: the ODBC API has no per-index argument, so `%` is always passed to `@index_name` (matching msodbcsql, since `sp_statistics_100` filters `index_name LIKE @index_name`, which matches nothing on NULL).

Also extracts a shared `COLMETA_NULLABLE_FLAG` constant used by both `SQLGetTypeInfo` and the new catalog functions (previously private to `get_type_info.rs`), and advertises all 7 functions in `SQLGetFunctions` — the Windows Driver Manager otherwise answers `IM001` without ever calling the driver (the same trap already documented here for `SQLGetTypeInfo`/`SQLSetStmtAttr`/`SQLEndTran`/`SQLGetConnectAttr`/`SQLColAttribute`); CI's live Windows e2e run caught this gap directly.

**Explicitly out of scope** (documented in the module and in `plan.md`): distributed/linked-server catalog queries (no ODBC catalog function even exposes a `Server` argument, and mssql-python never triggers this path), `SQL_SOPT_SS_NAME_SCOPE`, `sp_tableswc`-based wildcard-catalog enumeration (a rarer ODBC 3.x extension distinct from the bare-`%` idiom above), and `SQL_ATTR_METADATA_ID = TRUE` (not tracked anywhere in this driver yet, so always effectively pattern mode — the default and near-universal case). `SQLProcedureColumns` and the other four §4.10 "pass-through surface" bullets (SQLGetInfo, ODBC escape sequences, SQLSetConnectAttr/attrs_before) are tracked as separate sibling work items and are not part of this PR.

## Related Issues

Parent: [AB#46380](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46380) — "mssql-odbc | Catalog functions"

Child tasks, one per function, all implemented in this PR:
- [AB#47090](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47090) — Implement SQLTables
- [AB#47091](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47091) — Implement SQLColumns
- [AB#47092](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47092) — Implement SQLPrimaryKeys
- [AB#47093](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47093) — Implement SQLForeignKeys
- [AB#47094](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47094) — Implement SQLSpecialColumns
- [AB#47095](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47095) — Implement SQLStatistics
- [AB#47096](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47096) — Implement SQLProcedures

## Review history

A specialist PR-review pass (cross-checking every parity claim against the msodbcsql source and empirically verifying several hypotheses against a live SQL Server) found and this PR now fixes:
- The catalog-enumeration idiom bug described above (the highest-impact finding — a real, reproducible functional defect).
- A poisoned-ENV-mutex path in the new code that silently assumed ODBC 3.x instead of failing loudly, now matching `get_type_info.rs`'s identical check.
- Test-coverage gaps: added a parameterized invalid-cursor-state test covering all 7 functions (previously only `SQLTablesW` had one), e2e coverage for the enumeration idiom and `SQLForeignKeys`' conflicting-catalogs case, and the house `Benefits-from-mock-tds:` tag on the nonexistent-catalog e2e test.
- GitHub's automated Copilot review separately flagged the `SQLGetFunctions`/`IM001` gap, which was already being fixed in parallel after CI's own live Windows e2e run caught it.

## Testing

- 625 `mssql-odbc` unit tests pass, covering handle validation, cursor-state checks (all 7 functions), catalog-qualification/enumeration-sentinel logic, table-type-list rendering, and the RPC parameter `@`-name-prefix convention.
- `cargo bfmt` / `cargo bclippy` (workspace) are clean.
- C++ e2e suite (`tests/e2e/tests/catalog_test.cpp`): null-handle behavior plus live-server coverage of each function against a table/index/FK fixture, the nonexistent-catalog and catalog-enumeration paths, and the FK conflicting-catalogs case. This sandbox has no live SQL Server credentials to run it locally; CI's Windows e2e job (which caught the `SQLGetFunctions` gap on the first run) is the real validation.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (mssql-odbc scope; workspace-wide `mssql-tds` DB-connectivity/cert tests fail in this sandbox for pre-existing, unrelated reasons — no live SQL Server credentials are configured here)


mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1513 @@
...
+/// SQL_OV_ODBC2)`), read from the parent ENV. Selects `@ODBCVer` / `@fUsePattern`
+/// inclusion exactly as `get_type_info::sql_get_type_info_w_safe` does for
+/// `SQLGetTypeInfo`.
+fn is_2x_app(stmt: &StmtHandle) -> bool {
+    let env = stmt.parent_dbc().parent_env();
+    let Ok(env_state) = env.inner.lock() else {
+        error!("catalog function: env mutex poisoned reading ODBC version");
+        return false;
+    };
+    env_state.odbc_version == OdbcVersion::Odbc2
```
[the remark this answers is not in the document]
Vahid-b: Fixed in 600edca6 — `is_2x_app` now returns `Result<bool, SqlReturn>` and every caller propagates `SQL_ERROR` on a poisoned mutex instead of silently defaulting to "3.x", matching `get_type_info.rs`'s identical check.

mssql-odbc/tests/e2e/tests/catalog_test.cpp
```diff
@@ -0,0 +1,260 @@
...
+    EXPECT_EQ("TABLE_SCHEM", DescribeColName(stmt_, 2));
+    EXPECT_EQ(1, DrainRows(stmt_));
+
+    EXPECT_SQL_OK(SQLCloseCursor(stmt_), SQL_HANDLE_STMT, stmt_);
+}
+
+// A catalog argument naming a database that does not exist yields an empty
+// result set, not an error — msodbcsql's nonexistent-catalog recovery path
+// (sqlcdd.cpp DoDD(), lines 1883-1895), which this test exercises live.
+TEST_F(CatalogLiveTest, TablesNonexistentCatalogReturnsEmptyNotError) {
```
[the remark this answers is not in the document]
Vahid-b: Fixed in 600edca6 — added the `Benefits-from-mock-tds:` marker noting that a mock TDS server would let this assert the qualified RPC actually failed and the unqualified retry with the unmatchable filter actually fired, rather than just the black-box empty-result outcome.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1513 @@
...
+        error!(%e, "{name}: advancing to catalog rows failed");
+        return fail_with_tds(dbc, stmt, statement_handle, client, &e);
+    }
+
+    let rc = finish_execute(dbc, stmt, statement_handle, client, name);
+    if rc == SQL_ERROR {
+        return rc;
+    }
+    apply_catalog_metadata(stmt, not_null_cols, renames);
+    rc
```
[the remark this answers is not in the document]
Vahid-b: I looked into this one directly against the msodbcsql source before changing anything: `sqlcdd.cpp`'s `DoDD()` applies `SetColNames`/`ClearNullable` (the rename/nullability post-processing) unconditionally at its `Return1:` label, with no `IS2xAPP` gate anywhere near it — for these seven functions the ODBC 2.x/3.x distinction only affects the `@ODBCVer`/`@fUsePattern` RPC parameters sent to the stored procedure and the pattern-vs-identifier argument handling, not the column names/nullability of the result set. `SQLGetTypeInfoW`'s version-gated renaming is a different, structurally separate code path outside `DoDD()`, so it isn't good evidence of a bug here. I believe the unconditional rename in this PR is the correct parity match rather than an oversight — happy to revisit with a specific msodbcsql citation showing otherwise, but I don't have one after reading `DoDD()` directly.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1513 @@
...
+        &[2],
+        &[(0, "PROCEDURE_CAT"), (1, "PROCEDURE_SCHEM")],
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::api::odbc_types::SQL_NULL_HANDLE;
+    use crate::test_support::TestHandles;
```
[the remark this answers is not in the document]
Vahid-b: Fixed in 600edca6 — replaced the single-function `open_cursor_returns_24000` test with `open_cursor_returns_24000_for_every_function`, a table-driven test that exercises all seven entry points against `run_catalog`'s shared cursor/exec-state guard, so this gap can't silently recur for the sixth function that happened to be missing coverage (or an eighth added later).

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1513 @@
...
+/// 1564-1576, `fQualifierIsDB`). SQL Server's system catalog procedures are
+/// visible as `sys.<name>` in every database via the Resource database, so
+/// this needs no `USE` statement — matching the bare `[sys].proc` form this
+/// crate already uses for `SQLGetTypeInfo` (`get_type_info::DATATYPE_INFO_PROC`)
+/// when no catalog is given.
+fn qualified_proc_name(catalog: &Arg, proc: &str) -> String {
+    match catalog.as_deref().filter(|c| !c.is_empty()) {
+        Some(db) => format!("[{}].sys.{proc}", db.replace(']', "]]")),
+        None => format!("[sys].{proc}"),
+    }
```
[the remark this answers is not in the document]
Vahid-b: Confirmed and fixed in 600edca6 + da0207b3 — this was a real, live-verified bug (the initial fix attempt in 600edca6 introduced a follow-on SQLSTATE 15250 error on the retry path, fixed in da0207b3). Added `is_catalog_enumeration_sentinel`/`qualification_catalog`/`qualifier_value`: the bare `%` with blank name/owner now skips three-part qualification (matching msodbcsql `DoDD()` lines 1460-1463, which excludes this combination universally, not just for `SQLTables`) while still sending the literal `%` value in the qualifier RPC parameter so `sp_tables` recognizes the enumeration idiom. A real catalog now round-trips as-is (not normalized away) except on the unqualified retry, where it's correctly nulled to avoid a mismatched-qualifier error. Covered by new unit tests and an e2e test (`TablesEnumeratesCatalogsWithBarePercent`) that passed against a live SQL Server in CI.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1513 @@
...
+/// # Safety
+/// `ptr` must be null or point to `len` readable UTF-16 code units (or be
+/// NUL-terminated when `len` is `SQL_NTS`).
+unsafe fn opt_arg(ptr: *const SqlWChar, len: SqlSmallInt) -> Arg {
+    if ptr.is_null() {
+        None
+    } else {
+        Some(unsafe { read_utf16(ptr, len) })
+    }
+}
```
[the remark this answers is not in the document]
Vahid-b: Fixed in 552070f0 — added the same `debug_assert!(len == SQL_NTS || len >= 0, ...)` guard `prepare.rs` uses for the identical DM-enforced length contract.

mssql-odbc/src/api/exports.rs
```diff
@@ -655,6 +655,247 @@ pub unsafe extern "C" fn SQLRowCount(
...
+// ---- Catalog functions -------------------------------------------------------
+
+/// Returns table, catalog, or schema information.
+///
+/// # Safety
+/// - `statement_handle` must be a valid STMT handle returned by `SQLAllocHandle`.
+/// - Each name pointer must be null or reference `*_len` readable UTF-16 units.
+#[unsafe(no_mangle)]
+#[allow(clippy::too_many_arguments)]
+pub unsafe extern "C" fn SQLTablesW(
```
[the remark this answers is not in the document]
Vahid-b: Good catch — fixed in 64b8293f. Added all seven catalog function IDs (verified against the official ODBC-Specification `sql.h`/`sqlext.h`: SQLTables=54, SQLColumns=40, SQLPrimaryKeys=65, SQLForeignKeys=60, SQLSpecialColumns=52, SQLStatistics=53, SQLProcedures=67) to `supported_function_ids()`, both the single-ID and bitmap paths, plus a `catalog_functions_report_true` test. This was also independently caught by CI's live Windows e2e run (all `CatalogLiveTest` cases failed with `IM001` before the fix).

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1808 @@
...
+        Ok(client) => client,
+        Err(rc) => return rc,
+    };
+    flush_pending_unprepare(dbc, stmt, &mut client, name);
+
+    if let Err(e) = begin_transaction_if_manual(dbc, &mut client, name) {
+        return fail_with_tds(dbc, stmt, statement_handle, client, &e);
+    }
+
+    let has_catalog = catalog.as_deref().is_some_and(|c| !c.is_empty());
```
David-Engel: **Blocking** — the nonexistent-catalog retry never fires for the enumeration sentinel, so six of the seven functions return an error where msodbcsql returns an empty result set.

`catalog` here is the *qualification* catalog, which `qualification_catalog()` deliberately sets to `None` for the sentinel. So `has_catalog == false` and the retry on line 323 is skipped — but `qualifier_value()` still sends the literal `'%'` as the `@table_qualifier` RPC parameter.

Verified on SQL Server 2022 CU24: only `sp_tables` has the "Special feature #1" enumeration branch. The other six procs `raiserror(15250)` on `@table_qualifier = '%'`:

```
=== sp_columns_100  qualifier=% name/owner='' ===
ERROR 15250: The database name component of the object qualifier must be the name of the current database.
=== sp_pkeys / sp_fkeys / sp_statistics_100 / sp_special_columns_100 / sp_stored_procedures ===
ERROR 15250: (same)
```

So `SQLColumns(CatalogName="%", SchemaName="", TableName="")` and the equivalent on the other five return `SQL_ERROR` with 15250 instead of an empty result set.

And `sp_tables`' own branch requires **empty strings, not NULL**:

```sql
if @table_qualifier = '%' and @table_owner = '' and @table_name = ''
```

so `SQLTables(CatalogName="%", SchemaName=NULL, TableName=NULL)` hits 15250 too. `is_blank()` collapsing `None` and `""` is correct for the *qualification* decision, but the proc then receives NULLs and falls through to the 15250 path.

msodbcsql survives both because its retry gate is on the *supplied* qualifier, not on whether qualification happened — `DoDD()` line ~1880:

```c
else if (!(lpdd_arg->wNBits & QUAL_BIT) && wStdErr != IDS_S1_T00_1)
```

`wNBits & QUAL_BIT` is set only by `SetNullArgument`, i.e. only when the caller passed a null pointer. For `'%'` the retry fires and produces an empty, correctly-shaped result set.

I confirmed the retry shape recovers cleanly: `EXEC sys.sp_columns_100 @table_name = N' ', @table_qualifier = NULL, ...` returns 0 rows with no error.

Suggested fix — gate the retry on the original argument rather than deriving it from the qualification catalog:

```rust
fn run_catalog(
    // ...
    catalog: &Arg,          // qualification catalog (None for the sentinel)
    retry_on_error: bool,   // msodbcsql's !(wNBits & QUAL_BIT): a non-NULL catalog was supplied
    // ...
) -> SqlReturn {
    // ...
    if retry_on_error && matches!(exec_result, Err(TdsError::SqlServerError { .. })) {
```

with each `*_safe` passing `catalog.is_some()` (for `SQLForeignKeys`, `pk_catalog.is_some() || fk_catalog.is_some()`).

Reachability, stated honestly: both manifestations need input the ODBC spec doesn't sanction — `SQL_ALL_CATALOGS` is defined only for `SQLTables`, and the spec's idiom uses empty strings rather than NULL. But this PR deliberately implements the universal sentinel and says so in the description ("recognized universally (all 7 functions, matching msodbcsql `DoDD()`)"); as written, that half of the port makes this driver fail where the reference driver succeeds.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1808 @@
...
+/// only the three-part *qualification* is skipped, matching msodbcsql exactly.
+///
+/// Deliberately narrower than msodbcsql's full behavior: a wildcard catalog
+/// combined with a non-blank name/owner switches `SQLTables` to a dedicated
+/// `sp_tableswc` proc for pattern-based catalog enumeration (`sqlcdd.cpp`
+/// lines 1468-1486); that path has no ODBC-mandated caller (mssql-python never
+/// exercises it) and is not implemented here — such a combination is treated
+/// as a literal (and typically nonexistent) catalog name, same as any other
+/// catalog value containing `%`.
+fn is_catalog_enumeration_sentinel(catalog: &Arg, name: &Arg, owner: &Arg) -> bool {
```
David-Engel, line 411: **Blocking (same root cause as `catalog.rs:314`)** — this doc is accurate about msodbcsql skipping three-part qualification universally, but it's worth recording here that the qualification skip alone is *not* what makes the sentinel safe for the other six functions. Only `sp_tables` implements the enumeration branch; `sp_columns_100`, `sp_pkeys`, `sp_fkeys`, `sp_statistics_100`, `sp_special_columns_100`, and `sp_stored_procedures` all `raiserror(15250)` on `@table_qualifier = '%'` (verified live on SQL Server 2022 CU24). msodbcsql only survives them because `DoDD()`'s error retry re-runs unqualified with an unmatchable name.

A sentence to that effect would stop the next reader from concluding, as the current text implies, that skipping qualification is sufficient.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1808 @@
...
+    let has_catalog = catalog.as_deref().is_some_and(|c| !c.is_empty());
+    let (positional, named) = build_params(false);
+    let mut exec_result = dbc.runtime.block_on(client.execute_stored_procedure(
+        qualified_proc_name(catalog, proc),
+        Some(positional),
+        named,
+        (),
+    ));
+
+    if has_catalog && matches!(exec_result, Err(TdsError::SqlServerError { .. })) {
```
David-Engel: **Suggestion** — drop the info messages captured by the abandoned first attempt before retrying.

msodbcsql calls `FreeErrors(lpstmt)` before its recursive `DoDD` (`sqlcdd.cpp` line ~1884). Here, any `Info` token captured during the failed qualified call stays in the client and is posted by `finish_execute` against the *retry's* result, so a successful empty result can come back `SQL_SUCCESS_WITH_INFO` carrying messages from a call the application never sees.

```rust
if has_catalog && matches!(exec_result, Err(TdsError::SqlServerError { .. })) {
    debug!(%proc, "{name}: qualified catalog call failed, retrying unqualified");
    let _ = client.take_info_messages();
    let (retry_positional, retry_named) = build_params(true);
```

(The retry itself is safe on the transport: `advance_to_result_boundary`'s `Tokens::Error` arm drains the stream and clears `has_open_batch`, so the second `execute_stored_procedure` doesn't trip the `ALREADY_EXECUTING_ERROR` guard. Restricting the retry to `SqlServerError` rather than msodbcsql's "any error except timeout" is a sensible tightening.)

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1808 @@
...
+/// A decoded catalog-function argument. `None` is SQL NULL / not supplied;
+/// `Some(String::new())` is a genuine zero-length value, which for the
+/// ordinary (non-pattern) arguments below is an exact match against `''`
+/// (matches no real identifier) rather than "no filter".
+type Arg = Option<String>;
+
+/// # Safety
+/// `ptr` must be null or point to `len` readable UTF-16 code units (or be
+/// NUL-terminated when `len` is `SQL_NTS`).
+unsafe fn opt_arg(ptr: *const SqlWChar, len: SqlSmallInt) -> Arg {
```
David-Engel, line 129: **Suggestion** — arguments longer than 128 characters surface as `HY000` with an internal Rust message instead of the ODBC-mandated `HY090`.

`nvarchar(v, SYSNAME_LEN)` declares `NVARCHAR(128)`, and `tds_value_serializer.rs:1503` then returns:

```
UsageError("String length (N characters) exceeds schema size (128 characters)")
```

which `fail_with_tds` posts under `SQLSTATE_HY000`. msodbcsql's `ValidateArgument` (`sqlcdd.cpp` line ~389) returns `IDS_S1_090` — HY090, "Invalid string or buffer length" — for exactly this input. A length check here (or per argument, since `TABLE_TYPE_LEN` is 4000 while msodbcsql caps `TableType` at `SYSNAMELEN` too) would match the reference driver and avoid leaking an internal message to the application.

Worth noting the 128 is msodbcsql's cap, not the server's: `sp_tables` declares `@table_name` and `@table_owner` as `nvarchar(384)`, only `@table_qualifier` as `sysname`.

The error is a `UsageError` rather than a `SqlServerError`, so at least it doesn't spuriously trigger the retry path.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1808 @@
...
+}
+
+/// Builds the three-part qualified procedure name msodbcsql sends as the RPC
+/// target when a catalog was given (`[db].sys.proc`; `sqlcdd.cpp` lines
+/// 1564-1576, `fQualifierIsDB`). SQL Server's system catalog procedures are
+/// visible as `sys.<name>` in every database via the Resource database, so
+/// this needs no `USE` statement — matching the bare `[sys].proc` form this
+/// crate already uses for `SQLGetTypeInfo` (`get_type_info::DATATYPE_INFO_PROC`)
+/// when no catalog is given.
+fn qualified_proc_name(catalog: &Arg, proc: &str) -> String {
```
David-Engel, line 235: **Nit** — `qualified_proc_name` doesn't strip the ODBC search-pattern escape character.

msodbcsql runs the qualifier through `ValidateSearchPattern` (`sqlcdd.cpp:1308`) before building the bracketed name; that function removes `\` when it precedes `\`, `_`, or `%`, *and* doubles `]`. Only the `]` doubling is done here, so `SQLTables(CatalogName = r"my\_db")` builds `[my\_db].sys.sp_tables`, fails to resolve, and falls into the retry (empty result) instead of finding `my_db`.

The `]` doubling itself is correct and is what keeps this free of identifier injection.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,1808 @@
...
+            Some(vec![odbc_ver_param(), named_bit("fUsePattern", true)])
+        };
+        (positional, named)
+    };
+
+    run_catalog(
+        statement_handle,
+        "SQLColumnsW",
+        stmt,
+        "sp_columns_100",
```
David-Engel, line 921: **Nit / follow-up** — msodbcsql switches `SQLColumns` to `sp_columns_170` (and `sp_columns_ex_170`) when the connection reports vector support: `g_pwchKauaiProcNamesArr` in `sqlcdd.cpp`, selected by `pBatchCtx->m_hConn->IsVectorSupported()`. I confirmed `sp_columns_170` exists on the SQL Server 2025 image.

Not in scope for this PR — the plan targets 2016+ and `sp_columns_100` is the right default — but worth a work item so `SQLColumns` eventually describes vector columns the way msodbcsql does. (`sp_statistics_100`, `sp_special_columns_100`, `sp_pkeys`, `sp_fkeys`, `sp_tables`, and `sp_stored_procedures` are unchanged in the Kauai array, so `SQLColumns` is the only one affected.)

mssql-odbc/src/api/exports.rs
```diff
@@ -655,6 +655,247 @@ pub unsafe extern "C" fn SQLRowCount(
...
+/// updated when any value in the row changes.
+///
+/// # Safety
+/// - `statement_handle` must be a valid STMT handle returned by `SQLAllocHandle`.
+/// - Each name pointer must be null or reference `*_len` readable UTF-16 units.
+#[unsafe(no_mangle)]
+#[allow(clippy::too_many_arguments)]
+pub unsafe extern "C" fn SQLSpecialColumnsW(
+    statement_handle: SqlHandle,
+    identifier_type: SqlSmallInt,
```
David-Engel: **Nit** — `sql.h` declares `IdentifierType`, `Scope`, and `Nullable` as `SQLUSMALLINT` on `SQLSpecialColumnsW`, not `SQLSMALLINT`. Same width, so this is ABI-compatible and harmless in practice, but it diverges from the header the C++ e2e tests compile against. `SQLStatisticsW`'s `unique` / `reserved` correctly use `SqlUSmallInt` right above.

mssql-odbc/tests/e2e/tests/catalog_test.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+// ===================================================================
+// Tests that require a live SQL Server
+// ===================================================================
+
+class CatalogLiveTest : public ODBCTest {
+protected:
+    // Prefixed and specific enough that a collision with an unrelated table in
+    // the test database is unlikely.
+    static constexpr const char* kParentTable = "odbc_e2e_catalog_parent";
```
David-Engel: **Suggestion** — these are permanent tables with static names, and `SetUp` drops them before creating them. Two concurrent runs against the same database (two CI legs, or a developer running alongside CI) will delete each other's fixtures mid-test.

Every other e2e file here uses `#temp` tables; catalog functions genuinely can't, since `sp_tables` against `tempdb` would report mangled names. A per-process suffix — PID, or a random hex token generated once in `SetUpTestSuite` and appended to both names — keeps the runs isolated without changing anything else about the fixture.

David-Engel (commented):
## Summary

Implements the seven ODBC catalog functions by dispatching to the SQL Server system procedures over RPC, reusing the `SQLGetTypeInfo` execute path so cursor/metadata state is managed identically to a user query. I verified the parity claims directly against the msodbcsql source (`Sql/Ntdbms/sqlncli/odbc/sqlcdd.cpp`) and empirically against a live SQL Server 2022 CU24 and 2025.

**The column-metadata tables are exactly right.** Every `not_null_cols` and `renames` entry matches `sqlcdd.cpp`'s 1-based `COL(n)` bitmasks (lines 603, 675, 753, 839, 897, 1030, 1221) converted to 0-based, for all seven functions, as do the proc names against `g_pwchKatmaiProcNamesArr` and the `@ODBCVer` / `@fUsePattern` inclusion rules against `g_fYukonPatternAsParamArr` and the `!IS2xAPP` guard. `SQL_ALL_SCHEMAS` and `SQL_ALL_TABLE_TYPES` also work by pass-through, because `sp_tables`' special features #2 and #3 are evaluated before its qualifier check.

`cargo fmt --all --check`, `cargo clippy -p mssql-odbc --all-targets -- -D warnings`, and 621 `mssql-odbc` unit tests pass locally. CI is green, including 97% reported diff coverage.

One functional defect, detailed inline at `catalog.rs:314`, plus some suggestions and nits.

mssql-odbc/tests/e2e/tests/catalog_test.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+// Reads a column's name via SQLDescribeCol and returns it as a narrow string.
+std::string DescribeColName(SQLHSTMT stmt, SQLUSMALLINT column) {
+    SQLTCHAR name[128] = {};
+    SQLSMALLINT nameLen = 0;
+    SQLSMALLINT dataType = 0;
+    SQLULEN columnSize = 0;
+    SQLSMALLINT decimalDigits = 0;
+    SQLSMALLINT nullable = 0;
+    SQLRETURN rc = SQLDescribeCol(stmt, column, name,
```
David-Engel, line 26: **Suggestion** — nothing in the suite asserts the `ClearNullable` behavior, even though this helper already retrieves it.

`nullable` is passed to `SQLDescribeCol` and then discarded. The `not_null_cols` tables in `catalog.rs` are bare index literals whose only validation today is that they were transcribed correctly from `sqlcdd.cpp`'s `fNotNullCols` bitmasks (they were — I checked all seven), so an off-by-one introduced later would ship silently.

Returning `nullable` from this helper (or adding a sibling `DescribeColNullable`) and asserting, say, `SQL_NO_NULLS` for `SQLColumns` column 3 (`TABLE_NAME`) and column 11 (`NULLABLE`) would cover the whole mechanism in a couple of lines.

mssql-odbc/src/api/catalog.rs
Vahid-b: Confirmed and fixed in 50400b9f. You're right — `has_catalog` was derived from `qualify_catalog` (the post-sentinel `None`), so the retry never fired for the six procedures that don't natively recognize `%`. Added an explicit `retry_on_error: bool` parameter to `run_catalog`, computed at each of the 7 call sites from the *raw*, pre-transform catalog (`!is_blank(&catalog)`, or `pk_has || fk_has` for `SQLForeignKeys`, reusing what was already computed there). `qualify_catalog` keeps its original job of building the three-part name only. Also took your `take_info_messages()` suggestion in the same retry branch and added a unit test (`enumeration_sentinel_still_gates_retry_despite_null_qualification_catalog`) locking in the distinction.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Added a paragraph to `is_catalog_enumeration_sentinel`'s doc noting only `sp_tables` has the native "%" branch and the other six depend on `run_catalog`'s retry — and rewrote `qualification_catalog`'s and `qualifier_value`'s docs to stop implying otherwise (the old "which the stored procedures special-case" phrasing was exactly the misleading claim you flagged). Same commit (50400b9f) as the retry-gate fix, since the doc and code bug were the same root cause.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Added `let _ = client.take_info_messages();` right before the retry's `build_params(true)` call, matching msodbcsql's `FreeErrors(lpstmt)`. Agreed on restricting the retry to `SqlServerError` rather than msodbcsql's broader "any error except timeout" — left that as-is.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Added `check_arg_length`/`check_arg_lengths`, called at the top of each `sql_*_w_safe` before any RPC work, validating each string argument against the length already declared for its RPC parameter (`SYSNAME_LEN`=128 for identifiers, `TABLE_TYPE_LEN`=4000 for `SQLTables`' `TableType`, checked post-`table_type_value()` transform since that can lengthen the value). Posts the existing `ERR_INVALID_STRING_OR_BUFFER_LENGTH` (HY090) diagnostic via `post_diag`, reusing the same pattern already established in `exec_common.rs::build_named_params` for bound-parameter length errors, instead of letting it fail deep in RPC serialization as a generic HY000.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Added an `unescape_search_pattern` helper that strips `\` when it precedes `\`, `_`, or `%` (verified against `ValidateSearchPattern`, `sqlcdd.cpp` line 1308) and applied it in `qualified_proc_name` before the existing `]`-doubling — same order of operations msodbcsql uses. Added unit tests covering the three escape cases, the "backslash not followed by an escapable char is literal" case, and that unescaping happens before bracket-doubling. Scoped this to the proc-name-building path only, not the RPC qualifier parameter — msodbcsql's own RPC-parameter handling for the qualifier is per-function and more involved (only `SQLTables` uses the escape-stripped value there; the others go through generic pattern-argument conversion this driver doesn't implement, which is an existing, documented divergence), and you scoped the comment to `qualified_proc_name` specifically.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Agreed this is out of scope here — `sp_columns_100` is correct for the 2016+ target this PR covers, and vector-column support isn't part of AB#46380's catalog-functions scope. Not making a code change, but flagging this for a follow-up work item so `SQLColumns` eventually switches to `sp_columns_170`/`sp_columns_ex_170` on vector-capable connections, matching msodbcsql's `IsVectorSupported()` check. Thanks for tracking down the exact proc names.

> [!NOTE]

mssql-odbc/src/api/exports.rs
Vahid-b: Confirmed against `sqlucode.h` (`microsoft/ODBC-Specification`) — `fColType`/`fScope`/`fNullable` are all `SQLUSMALLINT`. Changed `identifier_type`/`scope`/`nullable` to `SqlUSmallInt` across `exports.rs`'s `SQLSpecialColumnsW` and all three `catalog.rs` layers, and retyped `SQL_BEST_ROWID`/`SQL_SCOPE_CURROW`/`SQL_SCOPE_TRANSACTION` to match (they're only used here). Left `SQL_NO_NULLS` as `SqlSmallInt` since it's also correctly used for `SQLDescribeCol`'s signed `Nullable` output elsewhere — cast it locally at the one comparison site instead.

> [!NOTE]

mssql-odbc/tests/e2e/tests/catalog_test.cpp
Vahid-b: Added a per-process random suffix (`UniqueSuffix()`, a function-local static hex token) appended to both `kParentTable` and `kChildTable`, and rewrote `SetUp`/`DropTestTables` to build the `CREATE`/`DROP`/index SQL from those same constants instead of separate hardcoded literals, so the DDL always matches. Couldn't compile-test this locally (no CMake in this sandbox), so I hand-traced the string concatenation carefully for paren balance — CI's e2e build is the real check, same as the rest of this file.

> [!NOTE]

mssql-odbc/tests/e2e/tests/catalog_test.cpp
Vahid-b: Added a `DescribeColNullable` helper alongside `DescribeColName` and asserted `SQL_NO_NULLS` for `SQLColumns` columns 3 (`TABLE_NAME`) and 11 (`NULLABLE`) in `ColumnsReportsOdbc3ColumnNamesAndAllColumns`, exactly as suggested.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Theekshna: We intend to add vector support soon. Can you add a work item for this under - https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46380 so that we dont miss it & add a link to it here?

Vahid-b: Done — created [AB#47440](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47440) as a child task under AB#46380 to track switching `SQLColumns` to `sp_columns_170`/`sp_columns_ex_170` for vector-capable connections. Thanks for flagging the urgency here.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,2051 @@
...
+/// (`true` only on that retry) must null out any real catalog value here —
+/// otherwise a nonexistent-catalog call would itself fail with 15250 instead
+/// of the empty result set the retry exists to produce.
+fn qualifier_value(catalog: &Arg, unmatchable: bool) -> Option<&str> {
+    if unmatchable {
+        None
+    } else {
+        catalog.as_deref()
+    }
+}
```
Theekshna: **Bug:** the `unescape_search_pattern` fix applies to the three-part procedure name but not to the qualifier parameter, so the two now disagree:

```rust
qualified_proc_name → unescape_search_pattern(db)   // "my\_db" → [my_db].sys.sp_tables
qualifier_value     → catalog.as_deref()            // @table_qualifier = 'my\_db'
```

For `CatalogName = "my\_db"` the RPC dispatches into database `my_db` while `@table_qualifier` carries the literal `my\_db`. The two don't match, the proc raises 15250, `retry_on_error` fires, and the caller gets an **empty result set** — the exact failure the unescape fix was meant to eliminate.

msodbcsql sends the unescaped, bracket-stripped name for this parameter (`sqlcdd.cpp:1745-1756`):

```c
if (lpdd_arg->bAPI == fSQLTABLES && (isz == XARG1_IND || (fQualifierIsDB && isz == QUAL_IND)))
{
    ...
    pszArg = szDBName + 1;                                     // strip '['
    cb     = static_cast<SDWORD>(cchDBName - 2)*sizeof(WCHAR); // strip ']'
}
```

`szDBName` is the bracket-wrapped output of `ValidateSearchPattern`, i.e. already unescaped. Running the same `unescape_search_pattern` here would close the gap.

Please also add a test with an escaped catalog that asserts **rows are returned**, not just that the call succeeds — an empty result set is indistinguishable from the legitimate nonexistent-catalog path, which is why this slipped past the existing coverage.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,2051 @@
...
+//! columns and clear NOT NULL flags to match the ODBC 3.x contract
+//! (`SetColNames` / `ClearNullable` in `DoDD()`).
+//!
+//! Deliberately out of scope, matching this driver's target of SQL Server
+//! 2016+ (Katmai and later) with no Driver-Manager-mediated linked-server
+//! support:
+//! - Distributed/linked-server catalog queries (`@table_name` scoped to a
+//!   remote `server`, the `sp_tables_ex`/`sp_columns_ex_100`/... procs, and
+//!   the `sp_cursoropen`-wrapped dispatch `DoDD()` uses for them). None of the
+//!   seven ODBC catalog functions even expose a `Server` argument, and
```
Theekshna: This justification isn't accurate. ODBC doesn't define a `Server` argument, but msodbcsql overloads `CatalogName` to carry one: `ValidateTableQualifier` (`sqlcdd.cpp:264-278`) splits the argument on the first `.` into server/database whenever `SQL_ATTR_METADATA_ID == SQL_FALSE`, and that split is precisely what selects the `_ex` procs:

```c
iProcNameIdx = !(lpdd_arg->wNBits & SERVER_BIT);   // sqlcdd.cpp:1512
```

Five of the seven route through it — `SQLTables`, `SQLColumns`, `SQLStatistics`, `SQLPrimaryKeys` (`sqlcdd.cpp:876`) and `SQLForeignKeys`. Only `SQLSpecialColumns` (`sqlcdd.cpp:801`) and `SQLProcedures` (`sqlcdd.cpp:1197`) opt out, via an explicit `SetNullArgument(SERVER_IND)`.

Two asks:

1. **Correct the rationale.** Not supporting `sp_tables_ex` is a perfectly reasonable scoping decision for a 2016+ target; "the argument doesn't exist" isn't the reason it's safe.
2. **Reconsider the resulting failure mode.** `SQLTables(CatalogName = "MYSRV.MyDb")` currently builds `[MYSRV.MyDb].sys.sp_tables`, fails, retries unqualified, and returns **silently empty**. Returning zero rows for a request the driver deliberately doesn't support is the worst available outcome — a diagnostic would be much easier to debug against.

Worth recording the divergence in the other direction here too, since it's a genuine improvement: a database legitimately named `My.Db` works correctly with this implementation and is mis-parsed by msodbcsql.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,2051 @@
...
+    let build_params = |unmatchable: bool| {
+        let pk_table_name = if unmatchable || catalogs_conflict {
+            Some(UNMATCHABLE_NAME.to_string())
+        } else {
+            pk_table.clone()
+        };
+        let positional = vec![
+            nvarchar(pk_table_name.as_deref(), SYSNAME_LEN),
+            nvarchar(pk_schema.as_deref(), SYSNAME_LEN),
+            nvarchar(qualifier_value(&pk_catalog, unmatchable), SYSNAME_LEN),
```
Theekshna: msodbcsql normalises a zero-length PK qualifier before dispatch (`sqlcdd.cpp:984-990`):

```c
if (!szPkTableQualifier || cchPkTableQualifier == 0)
    if (szFkTableQualifier)
        szPkTableQualifier = szFkTableQualifier;
```

This resolves `catalog` the same way for *qualification* (line 1006-1011), but then sends the original argument as the parameter, so `SQLForeignKeys(PKCatalog = "", FKCatalog = "OtherDb")` dispatches to `[OtherDb].sys.sp_fkeys` with `@pktable_qualifier = ''`.

I haven't confirmed how `sp_fkeys` treats a zero-length qualifier versus the substituted name, so I can't state what the observable difference is — but the reference driver deliberately normalises this and we don't, and the two are only equivalent if `''` and `'OtherDb'` behave identically inside the proc. Passing the already-resolved `catalog` here instead of `pk_catalog` makes them provably identical and costs nothing.

Minor, same function: msodbcsql sends `@fktable_qualifier` as unconditionally NULL (`SetNullArgument(&dd_arg, XARG3_IND)`, `sqlcdd.cpp:1021`) whereas we send the real value. I tried and couldn't construct a case where this diverges, since the three-part name already makes the target database current — so no change requested. But a line in the existing comment noting the deviation is intentional would save the next person who diffs this against `sqlcdd.cpp` the same investigation.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,2051 @@
...
+/// needs: renaming the ODBC 2.x names the system procedures emit
+/// (`TABLE_QUALIFIER`, `TABLE_OWNER`, ...) to their ODBC 3.x equivalents, and
+/// clearing the nullable flag on the columns the ODBC specification guarantees
+/// are NOT NULL — mirrors `SetColNames`/`ClearNullable` (`sqlcdd.cpp` lines
+/// 1910-1913, 2412-2472).
+fn apply_catalog_metadata(stmt: &StmtHandle, not_null_cols: &[usize], renames: &[(usize, &str)]) {
+    let Ok(mut stmt_state) = stmt.inner.lock() else {
+        error!("catalog function: stmt mutex poisoned applying column metadata");
+        return;
+    };
```
Theekshna: This returns `()`, so a poisoned mutex leaves the call reporting `SQL_SUCCESS` with the raw stored-procedure column names — `TABLE_QUALIFIER`, `TABLE_OWNER`, `PRECISION` — instead of the ODBC 3.x names, and with `SQL_NULLABLE` still set on columns the spec defines as `SQL_NO_NULLS`.

Both references say this has to fail:

- msodbcsql's `SetColNames` uses `SETRC_SERR_GOTO` (`sqlcdd.cpp:2459`), which aborts the call.
- `.github/instructions/mssql-odbc.instructions.md` — a poisoned `Mutex` returns `SQL_ERROR`. Every other poison site in this file already does exactly that (lines 124, 315, 377, 1354); this is the one that doesn't.

Silently wrong metadata reported as success is worse than an error here, because a conforming application binds these result sets by column name and will simply fail to find its columns. Suggest returning `Result<(), SqlReturn>` and propagating, consistent with `check_arg_length` above.

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,2051 @@
...
+/// argument before attempting any network call, rather than letting it
+/// surface as an internal RPC-parameter-serialization failure (`HY000`, not
+/// the state the spec mandates for this condition).
+fn check_arg_length(stmt: &StmtHandle, value: &Arg, max_len: u16) -> Result<(), SqlReturn> {
+    if value
+        .as_deref()
+        .is_none_or(|v| v.chars().count() <= max_len as usize)
+    {
+        return Ok(());
+    }
```
Theekshna: This measures `chars().count()` on the raw argument, but msodbcsql validates the length only *after* `ValidateSearchPattern` has stripped the escapes (`ValidateArgument`, `sqlcdd.cpp:425-437`; `ValidateTableQualifier`, `sqlcdd.cpp:296-307`). A 130-character pattern containing two `\`-escapes is therefore accepted by msodbcsql and rejected here with `HY090`.

`unescape_search_pattern` (line 189) already exists — reusing it for the pattern-typed arguments before measuring closes this.

Same function, while you're in here: `SQLProcedures` caps the procedure-name argument at `SYSNAME_LEN` (128), but msodbcsql uses `MAXPROCNAMESPHINX` = `MAX_PROCNAME + 6` (`sqlsrv.h:314`) for it (`sqlcdd.cpp:1225`). The `+6` covers the `;nnnnn` group-number suffix on numbered procedures. Narrow, but it's another valid input we'd reject with `HY090`.

mssql-odbc/src/api/catalog.rs
Vahid-b: Confirmed and fixed in acce30ba — real, verified regression from my earlier escape-char fix. `qualified_proc_name` unescaped the catalog before bracketing it, but `qualifier_value` still sent the raw (escaped) value as `@table_qualifier`, so the two disagreed for any escaped catalog and 15250 fired for a database that exists.

Made `qualifier_value` apply the same `unescape_search_pattern` `qualified_proc_name` already used, updated the 8 call sites accordingly, added a unit test (`qualifier_value_unescapes_search_pattern_same_as_qualified_proc_name`) asserting the two functions agree on the same escaped input, and added a live e2e test (`TablesFindsCreatedTableInEscapedCatalogName`) that creates a throwaway database with an underscore in its name (no default system database has one) and asserts the escaped-catalog call actually finds the real row, not just that it succeeds — per your ask.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Corrected the rationale in acce30ba: the module doc now explains that msodbcsql overloads `CatalogName` to carry a server name by splitting on the first `.` (`ValidateTableQualifier`, `sqlcdd.cpp` lines 264-278), selecting the `_ex` procs for five of the seven functions — not implementing `_ex` dispatch is the reasonable scoping call, but "the argument doesn't exist" was the wrong reason.

On the failure mode: I've left the silently-empty behavior as-is for now rather than adding new detection/diagnostic logic, and documented the tradeoff transparently in the doc (including your excellent point about `My.Db` resolving correctly here vs. msodbcsql's mis-parse). My reasoning: distinguishing "this dot means server.database" from "this is a literal dotted database name" well enough to raise a targeted diagnostic without regressing the `My.Db` case is a real design decision with its own edge cases (which SQLSTATE, whether the `%`-sentinel or an escaped dot should be exempted, per-function scope since only 5 of 7 route through `_ex`) — bigger than a same-PR fix I'm confident won't introduce a new regression. Happy to open a follow-up work item for a dedicated diagnostic if you'd like it tracked; let me know and I'll link it here the same way as AB#47440.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Fixed in acce30ba — now sending `qualifier_value(&catalog, ...)` (the already-resolved value) instead of the raw `pk_catalog` for the PK-side qualifier, so it's provably consistent with whichever side's catalog the three-part name actually qualifies into, as you suggested. Also added the comment you requested next to the `fk_catalog` real-value/NULL divergence, explaining it's intentional and why (the three-part name already makes the target database current, so a matching `fk_catalog` can't trigger 15250 there).

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Fixed in acce30ba — `apply_catalog_metadata` now returns `Result<(), SqlReturn>` and propagates `SQL_ERROR` on a poisoned mutex, matching every other poison site in this file and the repo's documented convention. Added `apply_catalog_metadata_poisoned_mutex_returns_error`, following the same poison-via-`catch_unwind` pattern as `row_count.rs`'s test.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Both fixed in acce30ba:

1. `check_arg_length` now takes an `is_pattern` flag and, when true, measures the character count after `unescape_search_pattern` — matching msodbcsql's `ValidateArgument`/`ValidateTableQualifier` (which only re-checks the unescaped length once the raw length exceeds the limit; our simpler "always unescape, then measure" is behaviorally equivalent since unescaping never lengthens a string). Applied `is_pattern = true` to every identifier argument (catalog/schema/table/column/procedure name) and `false` only to `SQLTables`' `TableType` (a literal value list, not a search pattern). Added two unit tests: one proving a 129-raw/128-unescaped-character pattern argument is now accepted, one proving a still-over-limit-after-unescaping argument is still rejected.

2. Added `PROC_NAME_LEN = SYSNAME_LEN + 6` (msodbcsql's `MAXPROCNAMESPHINX`) and used it for `SQLProcedures`' procedure-name argument, both for the upfront length check and the RPC parameter's declared width (so a 129-134 character value that now passes validation doesn't then fail at RPC serialization with a truncation error).

> [!NOTE]

mssql-odbc/src/api/catalog.rs
```diff
@@ -0,0 +1,2234 @@
...
+    let qualify_catalog = qualification_catalog(&catalog, &table, &schema);
+    let retry_on_error = !is_blank(&catalog);
+    let build_params = |unmatchable: bool| {
+        let table_name = if unmatchable {
+            Some(UNMATCHABLE_NAME.to_string())
+        } else {
+            table.clone()
+        };
+        let positional = vec![
+            nvarchar(table_name.as_deref(), SYSNAME_LEN),
```
Theekshna: **Pattern arguments are sent raw; msodbcsql rewrites ODBC escapes into T-SQL character classes**

`table.clone()` goes to the proc verbatim here, and likewise in `sql_columns_w_safe` / `sql_procedures_w_safe`. msodbcsql doesn't strip escapes from pattern args — `ConvertArgument` (`sqlcdd.cpp:2312-2324`) converts them:

| Input | msodbcsql | this PR |
|---|---|---|
| `my\_table` | `my[_]table` | `my\_table` |
| `arr\[0]` | `arr[[]0]` | `arr\[0]` |
| `a[b` | `a[[]b` | `a[b` |
| `a_b`, `a%b` | unchanged | unchanged ✅ |

Scope is exactly the three functions that send `@fUsePattern`: `wSearchBits` is set only at `sqlcdd.cpp:600`, `:672`, `:1218`. The other four never set it, so their identifiers pass through untouched and we already match.

The conversion is *not* suppressed by `@fUsePattern` — `fEscapePatternChars` gates a different branch (`:2344-2370`, bare `_`/`%` under `METADATA_ID=TRUE`). msodbcsql sends `@fUsePattern=1` *and* converts, which is the strongest evidence the proc doesn't interpret `\` itself. I haven't read `sp_tables`' body, so the exact symptom is inference; the divergence isn't.

Please don't fix this with `unescape_search_pattern` — stripping `\_` to `_` turns an escaped literal into a wildcard. `ValidateSearchPattern` (`:1308`, plain unescape, catalog identifier) and `ConvertArgument` (`:2272`, escape-to-class, pattern args) are deliberately different. A separate `escape_pattern_arg()` on the name/schema/column args of those three functions is the matching shape, plus an e2e case asserting `SQLTables(TableName = r"my\_table")` finds `my_table` and not `myXtable`.

Theekshna (commented):
Re-reviewed. Looks good, all concerns were addressed. Found one more issue listed below.

mssql-odbc/src/api/catalog.rs
Vahid-b: Confirmed against `ConvertArgument` (sqlcdd.cpp:2272-2409) — you're right that reusing `unescape_search_pattern` here would be wrong. Added a dedicated `escape_pattern_arg` that matches `ConvertArgument`'s character-class conversion (`\_`→`[_]`, `\%`→`[%]`, `\[`→`[[]`, bare `[`→`[[]`, `\\`→`\`, `\Y`→`\Y` unchanged), and applied it only to the pattern-typed arguments msodbcsql actually marks with search-pattern bits: table name/schema for `SQLTables`, table name/schema/column for `SQLColumns`, and proc name/schema for `SQLProcedures` (grepped every `wSearchBits` assignment in sqlcdd.cpp to confirm scope — the other 4 catalog functions never set it, so their args are correctly left unconverted).

`SQLTables`' catalog/qualifier argument is deliberately untouched — msodbcsql has a dedicated RPC-build bypass (sqlcdd.cpp:1745) that reuses the unescaped qualifier buffer instead of running it through `ConvertArgument`, so the existing `unescape_search_pattern` path there was already correct.

Also widened the RPC parameter declared length for the converted arguments (`PATTERN_ARG_LEN`/`PATTERN_PROC_NAME_LEN`, 3x the base length) since the character-class conversion can expand a value up to 3x — this mirrors msodbcsql's own dynamic `cbMax` sizing (sqlcdd.cpp:1775), found while implementing the fix rather than in the original comment.

Added unit tests for `escape_pattern_arg` plus the e2e test you asked for (`TablesEscapedUnderscoreMatchesLiteralNotWildcard`): creates a real table with a literal underscore and a decoy with a different character in that position, then asserts the escaped pattern matches only the real one. All pushed in e5c99ed9.

> [!NOTE]

Theekshna (approved):

shiwanigupta0809 (approved):

PR #328, opened 2026-08-19
Extract shared conversion core into src/conversion

Theekshna:
## Description

P0 of the parameter conversion milestone: move value-level conversion out from under the `SQLxxx` entry points into a shared `conversion/` module.

| File | From |
| --- | --- |
| `conversion/fetch_convert.rs` | `api/fetch_convert.rs` — the only file in `api/` that wasn't an entry point |
| `conversion/param_convert.rs` | `params/convert.rs` — so both direction converters are siblings |
| `conversion/error.rs` | `ConvOk` / `ConvError`, lifted out of `fetch_convert.rs` |
| `conversion/numeric.rs` | `NumericSource`, `parse_decimal_literal`, plus `narrow_i128` extracted from the `narrow!` macro that was inline in `convert_integer_c` |

Both moves are `git mv`, so they read as renames. `params/` keeps the `BoundParam` record and `conversion_matrix.rs` — the matrix never runs during conversion, it's consulted once at bind time and is parameter-direction-only by design.

**Parity:** msodbcsql has one conversion implementation parameterized by direction, declared in the shared `sqlcprot.h` (`XLATDIR` at `:769`, used by `Xlat` `:781`, `Convert` `:987`, `ParseDateTime` `:1004`). This matches that; conversion living beside the entry points did not. No new deviations.

### Also included: `SQL_DEFAULT_PARAM` returns 07S01

A correctness fix, in this PR because it shares files with the move.

`SQL_DEFAULT_PARAM` is only legal for a canonical procedure call, which this driver doesn't support, so it returned `HYC00` / *"not yet implemented"* — telling applications to wait for a feature that would never make their usage legal. It's now the terminal **07S01 "Invalid use of default parameter"**, matching msodbcsql (`sqlccmd.cpp:4104`, `:4513`, `deprecate.cpp:1275`, `sqlcfunc.cpp:2538`, each gated on `(cmdp.status & CANONICAL_CALL) == 0`).

Knock-on changes:
- `ParamConvError` → `ParamBuildError`, `DefaultParamUnsupported` → `InvalidUseOfDefaultParam`. The old name also collided with `SQL_C_DEFAULT`, which #323 now supports.
- `message() -> &'static str` → `diag() -> DiagMsg`. The old shape was the root cause: `build_named_params` forced everything except `InvalidLength` into `HYC00`, so a variant couldn't have its own SQLSTATE.

### Deferred

`ConvDirection`, the `FractionalTruncation` / `StringTruncation` split, and `SQLSTATE_22001` were in P0's original scope but nothing constructs them until P4/P5 — and unused constants fail `bclippy -D warnings`. Noted in `docs/parameters_plan.md`.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47364

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

## Testing

623 unit tests pass, re-run after rebasing onto `main`. Every pre-existing test passes unmodified — the main evidence the move is behavior-preserving. 7 new tests in `conversion/numeric.rs` cover the extracted code (literal parsing, truncation toward zero, `10^38` `checked_pow` overflow, non-finite floats, narrowing reports rather than wraps), and the default-param test asserts the SQLSTATE directly so a regression to `HYC00` fails.

Two gaps: `cargo btest` couldn't run locally (no `cargo-nextest` / `cargo-llvm-cov` here) so `cargo test -p mssql-odbc` was substituted, and the **e2e suite has not been run against a live server** — it needs an elevated shell for the HKLM write.

### DIFF
--- a/mssql-odbc/docs/parameters_plan.md
+++ b/mssql-odbc/docs/parameters_plan.md
@@ -107,7 +107,8 @@ transparent reconnects.
 
 Goal: support parameters of narrow and wide integer C types and character C
 types, with SQL <-> C conversion among them, on conversion infrastructure shared
-with the fetch path ([`api/fetch_convert.rs`](../src/api/fetch_convert.rs)).
+with the fetch path
+([`conversion/fetch_convert.rs`](../src/conversion/fetch_convert.rs)).
 
 ### Scope
 
@@ -164,7 +165,7 @@ one task per phase.
 
 | Phase | Task | Status | Deliverable |
 | --- | --- | --- | --- |
-| P0 | [47364](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47364) | Not started | Extract shared conversion core from `fetch_convert.rs` |
+| P0 | [47364](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47364) | Code complete | Extract shared conversion core from `fetch_convert.rs` |
 | P1 | [47365](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47365) | Code complete, unmerged | Parameter type model, conversion matrix, `SQL_C_DEFAULT` |
 | P2 | [47366](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47366) | Not started | Safe C-buffer reader and conversion-outcome channel |
 | P3 | [47367](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47367) | Not started | Quadrant A: integer C -> integer SQL |
@@ -175,23 +176,36 @@ one task per phase.
 
 P1 is independent of P0; P3 onward depend on both.
 
-#### P0 - Extract shared conversion core (not started)
+#### P0 - Extract shared conversion core (code complete)
 
-Pure refactor, no behavior change. Create `src/conversion/`:
+Pure refactor, no behavior change. `src/conversion/` now holds the value-level
+conversion for both directions:
 
-- `error.rs` - direction-aware outcome vocabulary lifted from `fetch_convert.rs`:
-  a `ConvDirection`, an outcome (`Exact`, `FractionalTruncation`,
-  `StringTruncation`), and an error kind (`OutOfRange`, `InvalidCharacterValue`,
-  `Restricted`, `StringTruncation`). `NotHandledHere` stays private to fetch
-  dispatch and must never reach an application.
-- `numeric.rs` - move `NumericSource` (exact `Int` / `Scaled` / `Float` model),
-  `parse_decimal_literal`, `to_i128_truncating`, and the checked-narrowing
-  helper. This carries the hardening from the 128-bit shift-overflow guard and
-  the `22003` unrepresentable-value fix into the parameter path.
-- Add `SQLSTATE_22001` and an inbound string-truncation `DiagMsg`;
-  `ERR_STRING_RIGHT_TRUNCATION` (`01004`) stays for the outbound path.
+- `error.rs` - the outcome vocabulary (`ConvOk`, `ConvError`) lifted out of
+  `fetch_convert.rs`. `NotHandledHere` stays a dispatch signal and must never
+  reach an application.
+- `numeric.rs` - `NumericSource` (exact `Int` / `Scaled` / `Float` model),
+  `parse_decimal_literal`, `to_i128_truncating`, and `narrow_i128`, extracted
+  from the `narrow!` macro that was local to `convert_integer_c`. This carries
+  the 128-bit shift-overflow guard and the `22003` unrepresentable-value fix
+  into the parameter path.
+- `fetch_convert.rs` - `api/fetch_convert.rs` moved wholesale. It has no handle
+  or diagnostic coupling, so it never belonged beside the `SQLxxx` entry points
+  in `api/`.
+- `param_convert.rs` - `params/convert.rs` moved, so both direction converters
+  sit together on the shared core. `params/` keeps what is genuinely about
+  bindings: the `BoundParam` record and the bind-time `conversion_matrix`.
 
-#### P1 - Parameter type model and conversion matrix (code complete, unmerged)
+Deferred to the phase that first constructs them, because `sqlstate.rs` carries
+no `allow(dead_code)` and `cargo bclippy` runs `-D warnings` - an unused
+SQLSTATE constant or a never-constructed enum variant fails the lint gate:
+
+- `ConvDirection` and the split of `Truncated` into `FractionalTruncation` /
+  `StringTruncation` - land with P4/P5, which raise inbound truncation.
+- `SQLSTATE_22001` and its `DiagMsg` - same. `ERR_STRING_RIGHT_TRUNCATION`
+  (`01004`) already exists for the outbound path.
+
+#### P1 - Parameter type model and conversion matrix (code complete)
 
 - [`api/type_rules.rs`](../src/api/type_rules.rs) - C-type canonicalization, the
   `HY003` / `HY004` identifier gates, and version-aware `SQL_C_DEFAULT`
@@ -293,7 +307,7 @@ Deviations from msodbcsql, verified against source:
 
 #### P7 - Cleanup and hooks (not started)
 
-- Remove remaining "Phase 1" language from `params/convert.rs` and
+- Remove remaining "Phase 1" language from `conversion/param_convert.rs` and
   `params/bound_param.rs`.
 - Record the deferred blockers: `SqlType` metadata/value separation for decimal
   and temporal typed NULLs, and the hard-coded decimal precision/scale in
@@ -329,7 +343,8 @@ accepting a pairing inbound that is rejected outbound for no principled reason.
   also allow `OUTPUT` and `?=` handling. This is no longer a repeated-parsing
   correctness issue.
 - **Type matrix and TDS type selection:** tracked by the conversion milestone
-  above. `params::convert` still ignores `sql_type` and emits `(n)varchar(max)`,
+  above. `conversion::param_convert` still ignores `sql_type` and emits
+  `(n)varchar(max)`,
   relying on SQL Server implicit conversion; P3 and P4 drive the wire type from
   `ParameterType` instead. Beyond this milestone the same work is needed for
   binary, `uniqueidentifier`, money, decimal, and date/time values.
--- a/mssql-odbc/docs/parameters_plan.md
+++ b/mssql-odbc/docs/parameters_plan.md
@@ -344,10 +344,10 @@ accepting a pairing inbound that is rejected outbound for no principled reason.
   correctness issue.
 - **Type matrix and TDS type selection:** tracked by the conversion milestone
   above. `conversion::param_convert` still ignores `sql_type` and emits
-  `(n)varchar(max)`,
-  relying on SQL Server implicit conversion; P3 and P4 drive the wire type from
-  `ParameterType` instead. Beyond this milestone the same work is needed for
-  binary, `uniqueidentifier`, money, decimal, and date/time values.
+  `(n)varchar(max)`, relying on SQL Server implicit conversion; P3 and P4 drive
+  the wire type from `ParameterType` instead. Beyond this milestone the same
+  work is needed for binary, `uniqueidentifier`, money, decimal, and date/time
+  values.
 - **Deferred features:** output parameters (`SQL_PARAM_OUTPUT`, `SQL_PARAM_INPUT_OUTPUT`), data-at-exec
   (`SQLParamData` / `SQLPutData`), parameter arrays
   (`SQL_ATTR_PARAMSET_SIZE`), and TVPs. Data-at-exec requires an
--- a/mssql-odbc/src/api/exec_common.rs
+++ b/mssql-odbc/src/api/exec_common.rs
@@ -331,7 +331,9 @@ pub(super) fn finish_execute(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::api::odbc_types::{SQL_C_CHAR, SQL_NTS, SQL_PARAM_INPUT, SQL_VARCHAR, SqlLen};
+    use crate::api::odbc_types::{
+        SQL_C_CHAR, SQL_DEFAULT_PARAM, SQL_NTS, SQL_PARAM_INPUT, SQL_VARCHAR, SqlLen,
+    };
     use crate::handles::handle_from_raw;
     use crate::params::BoundParam;
     use crate::test_support::TestHandles;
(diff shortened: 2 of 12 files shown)


mssql-odbc/src/api/exec_common.rs
```diff
@@ -210,18 +209,13 @@ pub(super) unsafe fn build_named_params(
...
-                    i + 1,
-                    e.message()
-                );
-                post_sql_error(stmt_state, SQLSTATE_HYC00, 0, e.message());
+                if let ParamBuildError::InvalidLength(len) = e {
+                    error!("{op}: parameter {} has invalid StrLen_or_Ind {len}", i + 1);
+                } else {
+                    error!("{op}: parameter {} rejected: {}", i + 1, e.diag().text);
+                }
+                post_diag(stmt_state, e.diag());
```
saurabh500, line 218: **Suggestion** — the `07S01` change has no end-to-end test, and this is the line that actually posts it.

`build_named_params` has three unit tests, but none exercises the `Err(e)` arm. The new SQLSTATE is only asserted at the enum level over in `param_convert.rs` (`err.diag().state == *b"07S01"`), never through the code that builds the diagnostic record. A regression that dropped this `post_diag`, or wired up the wrong `DiagMsg`, would still pass the suite.

`build_named_params_unbound_marker_posts_07002` is the pattern to copy — add alongside it in the `#[cfg(test)]` module, reusing the existing `char_param` helper:

```rust
/// `SQL_DEFAULT_PARAM` is terminal 07S01, not "not yet implemented" — assert it
/// through the poster, not just the enum.
#[test]
fn build_named_params_default_param_posts_07s01() {
    let h = TestHandles::with_env_dbc_stmt();
    let stmt = unsafe { handle_from_raw::<StmtHandle>(h.stmt) };
    let mut buf: Vec<u8> = b"x\0".to_vec();
    let mut ind: SqlLen = SQL_DEFAULT_PARAM;
    let mut state = stmt.inner.lock().unwrap();
    state.bound_params.push(Some(char_param(&mut buf, &mut ind)));

    let ret = unsafe { build_named_params(&mut state, 1, "test") };
    assert_eq!(ret.unwrap_err(), SQL_ERROR);
    assert_eq!(state.diag_records[0].sql_state, SQLSTATE_07S01);
}
```

Needs `SQL_DEFAULT_PARAM` and `SQLSTATE_07S01` added to the test module imports.

Also relevant to the 85% diff-coverage gate — these ~10 changed lines are currently uncovered.

mssql-odbc/src/conversion/numeric.rs
```diff
@@ -0,0 +1,200 @@
...
+                scale: 0
+            })
+        );
+        assert_eq!(
+            parse_decimal_literal(" .5 "),
+            Some(NumericSource::Scaled {
+                mantissa: 5,
+                scale: 1
+            })
+        );
```
saurabh500, line 131: **Suggestion** — `parse_decimal_literal` moved here, but its test didn't come with it.

`fetch_convert.rs` still has `parse_decimal_literal_forms` (~line 1191), which is now a white-box test of a function that file no longer owns. It's almost entirely duplicated by this test plus `non_decimal_text_is_rejected` — except for one case worth keeping: `-0.01`, which pins that a leading zero in the fractional part is absorbed into the mantissa (`-1`) while `scale` still counts both digits (`2`).

Carry that case over here:

```suggestion
        );
        assert_eq!(
            parse_decimal_literal("-0.01"),
            Some(NumericSource::Scaled {
                mantissa: -1,
                scale: 2
            })
        );
```

Then delete `parse_decimal_literal_forms` from `fetch_convert.rs` — it tests nothing this file doesn't, and leaving it there re-couples `fetch_convert` to an internal it no longer owns, which is the coupling this PR is undoing.

While you're there: `scale_beyond_i128_truncates_to_zero` in `fetch_convert.rs` opens with the same `to_i128_truncating` assertion as `scale_beyond_i128_pow_yields_zero_with_truncation` here. The rest of that test is a genuine end-to-end `convert_integer_c` case and should stay — only the leading `assert_eq!` is redundant.

mssql-odbc/docs/parameters_plan.md
```diff
@@ -329,7 +343,8 @@ accepting a pairing inbound that is rejected outbound for no principled reason.
   also allow `OUTPUT` and `?=` handling. This is no longer a repeated-parsing
   correctness issue.
 - **Type matrix and TDS type selection:** tracked by the conversion milestone
-  above. `params::convert` still ignores `sql_type` and emits `(n)varchar(max)`,
+  above. `conversion::param_convert` still ignores `sql_type` and emits
+  `(n)varchar(max)`,
   relying on SQL Server implicit conversion; P3 and P4 drive the wire type from
   `ParameterType` instead. Beyond this milestone the same work is needed for
   binary, `uniqueidentifier`, money, decimal, and date/time values.
```
saurabh500: **Nit** — the rename left an orphaned wrap: `` `(n)varchar(max)`, `` now sits alone on its own line, since `conversion::param_convert` is longer than `params::convert`. Reflow to match the surrounding paragraphs:

```suggestion
  above. `conversion::param_convert` still ignores `sql_type` and emits
  `(n)varchar(max)`, relying on SQL Server implicit conversion; P3 and P4 drive
  the wire type from `ParameterType` instead. Beyond this milestone the same
  work is needed for binary, `uniqueidentifier`, money, decimal, and date/time
  values.
```

saurabh500 (commented):
Reviewed the full diff against `origin/main`, verified the parity claims against the actual msodbcsql source, and ran the checks locally.

**This PR looks good.** No blocking findings — three non-blocking suggestions inline.

## Summary

A genuinely behavior-preserving refactor: value-level conversion moves out of `api/` and `params/` into a shared `conversion/` module, plus one intentional correctness fix (`SQL_DEFAULT_PARAM` -> `07S01`). Both moves are real renames (94% / 79% similarity), and every pre-existing test passes unmodified — the strongest available evidence the move is inert. The module boundaries the description draws are accurate, not aspirational.

## Verified locally

- `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` — clean
- 623/623 unit tests pass

(`cargo bclippy` / `cargo btest` can't run in this worktree: `Cargo.lock` is gitignored, so `--frozen` fails before doing any work. Environment quirk, not a problem with this PR.)

## Parity claim verified against msodbcsql source

Checked the source rather than the description, and it holds up — in fact more strongly than claimed:

- `Sql/Ntdbms/sqlncli/odbc/sqlccmd.cpp:4104` (`PutColumnData`) and `:4513` (`PutParamData`) — `cbValue == SQL_DEFAULT_PARAM && (lpstmt->cmdp.status & CANONICAL_CALL) == 0` -> `IDS_07_S01`, commented *"If default param used on non-canonical call stmt"*. Exact match.
- `Sql/Ntdbms/sqlncli/odbc/sqlcfunc.cpp:2538` sits inside `ParamToSQLType`, documented as *"converts parameter data supplied by the user in SQLBindParameter"* — the direct analogue of `bound_param_to_value`. So this driver raises `07S01` at the **same point in the pipeline**, not merely with the same SQLSTATE.

The old `HYC00` was genuinely wrong: it told applications to wait for a feature that would never make their usage legal. Good catch, and correct to fix it here since it shares files with the move.

## Other claims I checked rather than took on faith

- **`ParamBuildError::UnsupportedCType` "backstop only"** — accurate. `bind_param.rs:200` gates every bind through `is_supported_conversion`, which admits only `SQL_C_CHAR` / `SQL_C_WCHAR` — exactly what `bound_param_to_value` handles. The doc comment is true today.
- **`message() -> diag()` reshape is behavior-neutral** — `post_diag` delegates to `post_sql_error(state, msg.state, 0, msg.text)`, so native code and message text are unchanged for every variant except the intended `07S01`. The reshape is also the right fix: the old `&'static str` return was structurally incapable of carrying a per-variant SQLSTATE.
- **`narrow!` extraction** — `narrow_i128::<$ty>(v)?` is exactly equivalent to the previous `<$ty>::try_from(v).map_err(|_| ConvError::OutOfRange)?`.
- **No stale references** — grepped for `ParamConvError`, `DefaultParamUnsupported`, `params::convert`, `api::fetch_convert`; all clear. No e2e test asserts `HYC00` for default parameters, so nothing silently contradicts the new state.
- **Deferring `ConvDirection` / `SQLSTATE_22001`** — correct call. `sqlstate.rs` carries no `allow(dead_code)` and clippy runs `-D warnings`, so landing them now would fail the gate. Documented in `parameters_plan.md` rather than left implicit.

## Findings

**Blocking** — none.

**Suggestion**
1. `exec_common.rs:218` — the `07S01` change has no end-to-end test; the line that actually posts the diagnostic is uncovered.
2. `conversion/numeric.rs:131` — `parse_decimal_literal` moved but its test didn't, leaving a duplicate behind in `fetch_convert.rs`.

**Nit**

3. `docs/parameters_plan.md:347` — orphaned line wrap left by the rename.

## One process note

The checklist has `cargo btest` unticked and the e2e suite hasn't run against a live server. Since the `07S01` change is application-visible, an e2e pass before merge would be worth it — the unit tests can't catch a Driver Manager-level surprise.

mssql-odbc/src/api/exec_common.rs
Theekshna: Done in 824c5b5c. Added `build_named_params_default_param_posts_07s01` alongside the 07002 test, reusing `char_param`. You were right that the enum-level assertion was insufficient — it would not catch a dropped `post_diag` or the wrong `DiagMsg`.

One correction to the suggestion: only `SQL_DEFAULT_PARAM` needed adding to the test imports. `SQLSTATE_07S01` was already in scope, since the module has `use super::sqlstate::*` and the test module pulls it in via `use super::*` — same way the existing test reaches `SQLSTATE_07002`.

mssql-odbc/src/conversion/numeric.rs
Theekshna: Done in 824c5b5c. Carried the `-0.01` case into `plain_decimal_literals_parse_exactly` with a comment explaining what it pins, deleted `parse_decimal_literal_forms` from `fetch_convert.rs`, and dropped the redundant leading `to_i128_truncating` assertion from `scale_beyond_i128_truncates_to_zero` — its `convert_integer_c` body is genuinely end-to-end and stays.

Agreed on the reasoning: leaving a white-box test of `parse_decimal_literal` in `fetch_convert.rs` re-coupled it to an internal it no longer owns, which is exactly what this PR is undoing. Test count is unchanged at 623 (one added, one removed).

mssql-odbc/docs/parameters_plan.md
Theekshna: Done in 824c5b5c — reflowed. My rename edit caused that orphaned wrap.

saurabh500 (commented):
Re-reviewed at `824c5b5c`. **All three findings are fixed correctly — no new issues, and nothing regressed.**

| Finding | Status |
|---|---|
| `exec_common.rs:218` — `07S01` untested through the poster | Fixed |
| `numeric.rs` — orphaned duplicate test in `fetch_convert.rs` | Fixed |
| `parameters_plan.md` — orphaned line wrap | Fixed |

## Verified, not just read

**1. The new test genuinely exercises the path.** I checked that it can't pass for the wrong reason: in `param_convert.rs`, the `SQL_DEFAULT_PARAM` check (line 91) sits *above* the catch-all `ind < 0` -> `InvalidLength` arm (line 98). So `build_named_params_default_param_posts_07s01` really does route through `InvalidUseOfDefaultParam` -> `post_diag` -> `07S01`, rather than tripping the generic negative-indicator branch and landing on the right SQLSTATE by accident. It now covers `exec_common.rs:218`, which was the actual gap.

You're right about the imports — `SQLSTATE_07S01` reaches the test module through `use super::*`, so only `SQL_DEFAULT_PARAM` needed adding. My suggestion was over-specified there.

**2. The test move is a real consolidation, not a deletion.** `-0.01` carried over to `plain_decimal_literals_parse_exactly` with a comment stating what it pins, and `parse_decimal_literal_forms` is gone from `fetch_convert.rs`. The remaining cases there (`12.34`, `+7`, `1e5`, `abc`, `""`) are all covered by `plain_decimal_literals_parse_exactly` / `non_decimal_text_is_rejected`, so nothing was lost. Dropping only the leading `to_i128_truncating` assertion from `scale_beyond_i128_truncates_to_zero` while keeping its `convert_integer_c` body was the right cut — the end-to-end coverage survives.

**3. No fallout from the deletion.** Removing a test block can strand imports and trip `-D warnings`; it didn't. Clippy is clean.

## Checks re-run at `824c5b5c`

- `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` — clean
- 623/623 pass, including `build_named_params_default_param_posts_07s01`

Test count is unchanged at 623 as you said — one added, one removed — which is the right shape for this change: net-new coverage on the behavioral fix, minus a redundant white-box test.

Nothing further from me. The one thing still outstanding is the process note from my first pass, unchanged by this commit: `cargo btest` is still unticked and the e2e suite still hasn't run against a live server. Since `07S01` is application-visible, that pass is worth having before merge — the unit tests can't catch a Driver Manager-level surprise. The ADO validation run is still pending on this head as of now.

saurabh500 (approved):
## Approving

All review feedback is addressed, the rebase onto `main` is clean, and validation is green. Nice work on this one.

### What this PR does

Extracts the value-level conversion logic out of `api/` and `params/` into a direction-neutral `conversion/` module, and fixes one genuine bug along the way: `SQL_DEFAULT_PARAM` now returns `07S01` "Invalid use of default parameter" instead of `HYC00` "not yet implemented".

The `HYC00` was actively misleading — it told applications to wait for a feature that would never make their usage legal, since `SQL_DEFAULT_PARAM` is only valid on a canonical procedure call. Worth fixing, and correct to fix it here rather than separately, since it shares the files being moved.

### Why I'm confident the refactor is inert

- Both moves are real renames (94% and 79% similarity), not copy-paste-and-edit.
- Every pre-existing test passes **unmodified** — the strongest available evidence that the move changed no behavior.
- The one behavioral change is isolated to a single enum variant and is covered by a test that asserts through the diagnostic poster, not just the enum.

### Parity verified against msodbcsql source

I checked the actual driver source rather than the docs, and the claim holds — in fact more strongly than the description stated:

- `sqlccmd.cpp:4104` (`PutColumnData`) and `:4513` (`PutParamData`) — `cbValue == SQL_DEFAULT_PARAM && (lpstmt->cmdp.status & CANONICAL_CALL) == 0` -> `IDS_07_S01`, commented *"If default param used on non-canonical call stmt"*. Exact match.
- `sqlcfunc.cpp:2538` sits inside `ParamToSQLType`, documented as *"converts parameter data supplied by the user in SQLBindParameter"* — the direct analogue of `bound_param_to_value`. So this driver raises `07S01` at the **same point in the pipeline**, not merely with a matching SQLSTATE.

### Review feedback — all three resolved

| Finding | Resolution |
|---|---|
| `exec_common.rs:218` — `07S01` untested through the poster | `build_named_params_default_param_posts_07s01` added |
| `numeric.rs` — orphaned duplicate test left in `fetch_convert.rs` | `-0.01` case carried over, duplicate deleted, redundant assertion dropped |
| `parameters_plan.md` — orphaned line wrap | Reflowed |

I confirmed the new test can't pass for the wrong reason: the `SQL_DEFAULT_PARAM` check in `param_convert.rs` sits *above* the catch-all `ind < 0` -> `InvalidLength` arm, so it genuinely routes through `InvalidUseOfDefaultParam` -> `post_diag` -> `07S01` rather than tripping the generic negative-indicator branch and landing on the right SQLSTATE by accident.

### Rebase onto `main` is clean

`sqlstate.rs` was the only file where this PR and `main` overlapped — `main` added `SQLSTATE_08S01` (connection pooling), this PR adds `SQLSTATE_07S01`. No semantic collision, and both landed in correct sort order. I diffed the PR's own changes before and after the rebase: byte-identical apart from blob indices and one shifted hunk header.

### Validation at `92213e6c`

- `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` — clean
- 642/642 tests pass, including the new `07S01` test

Pipeline: 14 checks passing, 0 failures, remaining builds still in flight. E2E runs in the pipeline, which covers the one gap I'd flagged — the `07S01` change is application-visible, so having it exercised against a live server is the right safety net.

Deferring `ConvDirection`, the `FractionalTruncation` / `StringTruncation` split, and `SQLSTATE_22001` to P4/P5 was also the right call — nothing constructs them yet, and unused constants would fail the `-D warnings` gate. Good that it's documented in `parameters_plan.md` rather than left implicit.

Please hold the merge until the remaining pipeline stages finish. Merge is yours per repo workflow — happy to see it go in once the pipeline is fully green.

PR #333, opened 2026-08-19
Move pool-reset acknowledgement verification into TdsClient

saurabh500:
## Description

#308 verified the pool reset in the ODBC layer: `DbcState::pending_reset_ack` suppressed the same-value short circuit in `set_txn_isolation` so the checkout `SET TRANSACTION ISOLATION LEVEL` always emitted and carried the RESETCONNECTION bit, and that handler then checked `TdsClient::reset_pending()`.

The bit can ride **any** SQL Batch, RPC, or Transaction Manager request (`packet_writer.rs`), so that only worked for a consumer whose checkout happens to issue that SET — correct for `mssql-python` today, not a driver guarantee. `reset_connection` also had a narrow window: it released the client and only then re-locked to set the flag.

### Verification moved into `TdsClient`

`reset_pending: bool` becomes a three-state machine:

| State | Meaning |
|---|---|
| `Idle` | nothing armed or outstanding |
| `Armed` | bit set on the writer, no request has carried it yet |
| `AwaitingAck` | a request carried it; its response owes the `ResetConnection` ENVCHANGE |

The packet writer consumes the armed mode, not the client, so `NetworkWriter` gained `note_reset_dispatched` / `take_reset_dispatched`: `populate_header_and_send` records that a header carrying the bit actually reached the wire, and the client takes that record — **only while it still believes a reset is `Armed`**, so a record left by a response that was never read can't be charged to an unrelated later request.

SQL Server resets the session *before* it parses the carrying request, so the ENVCHANGE precedes everything that request produces. `observe_response_token`, called at the four `receive_token` loops, treats the first token that proves the request ran (DONE, COLMETADATA, ROW, RETURNSTATUS…) as the last moment the acknowledgement could have arrived. Reaching it while `AwaitingAck` marks the connection dead and fails the request with the new `Error::ConnectionResetNotAcknowledged`, which `post_tds_error` maps to `08S01` from every entry point.

Row reads need no hook: the first COLMETADATA of a response always arrives through `receive_token`.

**Not treated as failure** (marking a healthy session dead is worse than the gap):
- ENVCHANGE / INFO / SESSIONSTATE — can legitimately precede or accompany the ack.
- ERROR — the server's diagnostic must reach the caller unmasked; the reset is still caught on the next token, so the connection is condemned anyway, just behind the server's error.
- Ignored messages (attention/cancel) — never recorded as having delivered the bit.
- Transparent reconnect — the new transport carries no dispatch record, and adopting a recovered session clears the state outright.

**Abandoned carriers** (from review). Cancellation and timeout drain to the attention acknowledgement, and that drain discards every other token — the ENVCHANGE included — so the carrying request ended `Armed` with the bit dispatched and the *next, unrelated* request was condemned on its first token. The settle runs at the request boundary, in **both** `begin_command` and `check_and_reconnect`: neither alone is universal, since the cursor RPCs never call `begin_command` and the transaction ops deliberately skip recovery. It settles as reset rather than failed — the server resets before processing, the bit reached the wire, and arm-time invalidation already reconciled client state — and no pool guarantee is weakened, since an abandoned carrier fails its own request and the checkout discards the connection.

`reconnect`'s "adopt the new session" block is extracted into `adopt_recovered_session` (which also carries #230's FEATUREEXTACK reset-then-reload) so that invariant is testable without a live server.

### The isolation question, decided as one change

`pending_reset_ack` did two jobs. Job 2 (verify the ack) moves to `TdsClient`. **Job 1 must stay**: `sp_reset_connection` does not restore the isolation level (D9), and a previous borrower may have changed it via raw T-SQL the cache never saw — so without the suppression, a checkout `SET READ COMMITTED` would short-circuit and leave the server at the previous borrower's level, reintroducing the cross-borrower leak #308 closed.

So the flag is kept but narrowed and renamed to `server_isolation_unknown`: *the cached isolation level is no longer evidence about the server*. It is raised **before** the client is claimed (closing the release/re-lock window) **and** re-asserted under the lock that records the completed arm, which also bumps the new `reset_generation`. An isolation SET captures that generation before it sends and clears the invalidation only if it is unchanged — otherwise a SET that reached the server just before a concurrent reset armed could clear an invalidation it never satisfied (from review).

**We deliberately do *not* set `txn_isolation = SQL_TXN_READ_COMMITTED` at arm time.** The premise that the session is READ COMMITTED after a reset is false under D9 — the reset leaves it at whatever the previous borrower set. The last level set through the attribute is the closest thing the driver actually knows, so reporting it is the most accurate available answer, and asserting READ COMMITTED would additionally make the checkout SET short-circuit and leave the server unrestored. Marking the cache untrusted states what is true and forces the SET that makes it true again.

### Not regressed

No-I/O reset (no `SELECT 1`), arm-time cache/descriptor invalidation, `RESETCONNECTIONSKIPTRAN` keeping its transaction descriptor, and reconnect clearing a pending reset all preserved.

## ⚠️ Public API change

This adds one variant, `Error::ConnectionResetNotAcknowledged`, to `mssql_tds::error::Error`. The enum is **not** `#[non_exhaustive]`, so this is technically breaking for any downstream code matching exhaustively on it without a `_` arm.

Adding `#[non_exhaustive]` was considered and deliberately deferred: applying it to an already-public enum is itself an immediate breaking change for the same exhaustive matchers, so doing it here would mean two breaks in one release instead of one. It remains the right end state for an enum with 25+ variants and is better done as its own change where the break is the headline. Raised in review — happy to fold it in here if the owner prefers to absorb both at once.

## Related Issues

[AB#47400](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47400) — https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47400 (child of Feature 47102)

Originating review thread: https://github.com/microsoft/mssql-rs/pull/308#discussion_r3812891812

## Validation

| Check | Result |
|---|---|
| `cargo bfmt` | pass |
| `cargo bclippy` (`--workspace --all-features --all-targets -D warnings`) | pass |
| `cargo nextest run -p mssql-tds -p mssql-odbc --lib` | **2582/2582** |
| Live TDS: reset, pool-primitives, recovery/reconnect, cursor, timeout/cancel | **226/226** |
| ODBC e2e in the CI container (`containerized-odbc-e2e.sh`, SQL Server 2022) | **23/23**, incl. `connection_pool_test` |

Tests added:
- `unacknowledged_reset_fails_the_carrying_request_and_marks_it_dead` — the negative case.
- `unacknowledged_reset_fails_a_row_returning_carrier_at_its_metadata` — a non-isolation carrier, i.e. consumer independence.
- `server_error_on_the_carrier_is_not_masked_by_the_reset_verdict` — and the connection is still condemned by the following drain.
- `armed_reset_that_never_reached_the_wire_condemns_nothing`.
- `adopting_a_recovered_session_clears_a_pending_reset` — the reconnect case.
- `abandoned_carrier_does_not_condemn_the_next_request` and `abandoned_carrier_that_reached_awaiting_ack_is_settled_too` — the cancellation/attention regression, both verified to fail with the fix removed.
- `cursor_path_request_boundary_settles_an_abandoned_carrier` — the cursor-RPC boundary (from review), verified to fail with the fix removed.
- `starting_a_command_leaves_an_unsent_armed_reset_alone` — the settlement must not swallow an ordinary armed reset.
- `a_reset_armed_mid_flight_is_not_cleared_by_the_older_isolation_set`, `arming_a_reset_bumps_the_generation_and_reasserts_the_invalidation`.
- `ignored_message_does_not_record_a_reset_dispatch`, `ineligible_message_neither_consumes_nor_dispatches_the_reset`.

Reworked rather than deleted, intent preserved: `armed_reset_suppresses_the_isolation_short_circuit` (now justified by the untrusted cache), `unacknowledged_reset_poisons_client_on_the_carrying_request` (now asserts the ODBC layer maps the client's verdict to `08S01`).

`docs/odbc-connection-pooling-plan.md` updated: state model, sequence diagram, an *Acknowledgement verification* section including abandoned carriers, and the isolation reasoning above.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (run as `cargo nextest run -p mssql-tds -p mssql-odbc --lib`, plus live TDS + ODBC e2e)
- [x] Public API changes are documented (see **Public API change** above)


mssql-odbc/src/api/txn.rs
```diff
@@ -432,8 +435,18 @@ pub(super) fn reset_connection(dbc: &DbcHandle, value: u64) -> SqlReturn {
...
+    // client is released: from the moment this connection is being recycled the
+    // driver cannot vouch for the server's isolation level, and setting it here
+    // leaves no window in which another thread's checkout SET could still
+    // short-circuit. Arming may yet fail below, in which case the flag costs at
+    // most one redundant SET on a connection the pool is about to discard.
     let started = match dbc.inner.lock() {
-        Ok(state) => state.local_tran_started,
+        Ok(mut state) => {
+            state.server_isolation_unknown = true;
+            state.local_tran_started
```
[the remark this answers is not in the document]
saurabh500: Good catch — confirmed and fixed.

You're right that the early set doesn't survive the interleaving: B reads `true`, claims first, sends its SET, and clears the flag at what was line 727, after which A arms with the flag already `false`. The invalidation needed to be tied to the arm rather than to a read that predates it.

Fixed with the generation approach you suggested:

- `DbcState` gains `reset_generation`. `reset_connection` bumps it **and** re-asserts `server_isolation_unknown = true` in the same lock that records the completed arm (the success path), so the invalidation is established after arming, not just before claiming.
- `set_txn_isolation` captures `reset_generation` in its first lock and clears the invalidation only if it is unchanged, via a new `settle_isolation_invalidation` helper.

That closes your interleaving: B captures gen `N`, sends, A arms (gen → `N+1`, flag `true`), B re-locks, sees `N+1 ≠ N`, and leaves the newer invalidation alone. The next checkout SET therefore still emits.

I kept the early set as well — it narrows the window before arming, and it's what the "arming failed" path relies on to stay conservative (rollback failure leaves the flag set but does **not** bump the generation, since no reset was armed for an in-flight SET to miss).

Tests added: `a_reset_armed_mid_flight_is_not_cleared_by_the_older_isolation_set` drives the guard directly (the interleaving lives inside the SET's own round trip, so pre-bumping from the test wouldn't reproduce it — the capture happens after), and `arming_a_reset_bumps_the_generation_and_reasserts_the_invalidation` pins the arm-side half.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
```diff
@@ -557,23 +586,75 @@ impl TdsClient {
...
+    /// produces. The first token that is evidence the request actually ran is
+    /// therefore the last moment the acknowledgement could still have arrived,
+    /// and its absence means the session was never reset. ERROR is exempt so a
+    /// server diagnostic reaches the caller instead of being masked by this
+    /// one; a reset still unacknowledged after it is caught on the next token.
+    fn observe_response_token(&mut self, token: &Tokens) -> TdsResult<()> {
+        if self.reset_state == ResetAckState::Armed
+            && self.transport.as_writer().take_reset_dispatched()
+        {
+            self.reset_state = ResetAckState::AwaitingAck;
```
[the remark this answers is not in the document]
saurabh500: Confirmed and fixed — thank you, this was the real bug in the change.

I traced it exactly as you describe: `receive_token` answers cancellation/timeout with `cancel_read_stream_and_wait` → `wait_for_attention_ack`, which returns only on the attention DONE and discards everything else (`network_transport.rs:1088`), ENVCHANGE included. The client never reads a token, so it stays `Armed` with `reset_dispatched == true`, and the *next* request's first token flips it to `AwaitingAck` and condemns it. Marking a healthy connection dead is worse than the gap this verification exists to close.

Rather than special-casing the attention drain, I fixed the general invariant: **a verdict is only valid against the response of the request that carried the bit**, so when that response is abandoned the suspicion must not be carried forward. `begin_command` now calls `settle_abandoned_reset_verification`. It runs on every request path (including the transaction-manager ones) and always *before* the current request sends anything, so a dispatch record or outstanding `AwaitingAck` seen there can only belong to an earlier request. That also covers abandon paths other than attention, not just the one you found.

It settles as reset rather than as failure: the server resets before it processes the carrying request, the bit demonstrably reached the wire, and `prepare_reset_connection` already reconciled every client-side cache at arm time. No pool guarantee is weakened, because an abandoned carrier fails its own request — the ODBC checkout returns `SQL_ERROR` and the pool discards the connection anyway.

Verified the regression test reproduces your scenario: with the `settle_abandoned_reset_verification` call removed, `abandoned_carrier_does_not_condemn_the_next_request` fails on exactly the assertion you predicted; with it, it passes. `starting_a_command_leaves_an_unsent_armed_reset_alone` guards the other direction so the settlement can't swallow an ordinary armed-but-unsent reset.

One deliberate limit worth recording: this settles the acknowledgement rather than recovering it, so a reset whose ENVCHANGE the attention drain swallowed is not *observed*, only inferred from reset-before-processing. Making the drain surface the ENVCHANGE instead of discarding it would be strictly better and is a reasonable follow-up, but it means threading reset state through the transport's attention path, which is broader than this PR should reach.

Full validation re-run after the fix: 2424/2424 unit, 12/12 live TDS, 21/21 ODBC e2e.

> [!NOTE]

saurabh500 (commented):
Solid design. Moving reset verification into `TdsClient` is the right layer — the pool no longer has to police a protocol invariant it can't see. The `Armed → AwaitingAck` state machine reads well, all four `receive_token` loops are instrumented, and the ODBC `reset_generation` guard genuinely closes the release-then-relock window (that one is easy to get wrong and this gets it right).

No blocking issues. Two suggestions worth acting on before merge and four nits.

**Verified as correct, for the record:**
- All four `receive_token` loops call `observe_response_token`, and all four handle the `ResetConnection` ENVCHANGE.
- The ERROR-token exemption still condemns the connection via the follow-on drain (covered by a test).
- Transparent reconnect is safe: `adopt_recovered_session` clears `reset_state` and the new transport has no dispatch record.
- Setting `server_isolation_unknown = true` before arming and re-asserting after is deliberate and right (D9 — `sp_reset_connection` does not restore isolation level).
- `handle_row_read_token` skipping `observe_response_token` is fine: COLMETADATA always arrives through `receive_token` first.

Local `cargo nextest -p mssql-tds -p mssql-odbc --lib`: 2574 passed, 7 failed — all pre-existing certificate/TLS fixture failures unrelated to this change.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4901,6 +5034,7 @@ impl TdsClient {
     /// new session *after* this point, so those remain visible as part of the
     /// command that triggered the reconnect.
     fn begin_command(&mut self) {
+        self.settle_abandoned_reset_verification();
```
saurabh500, line 5107: **Suggestion** — the "runs at the start of every command" invariant isn't actually enforced.

`settle_abandoned_reset_verification()` only runs for paths that call `begin_command`, and `begin_command` is private to this `impl` block. `cursor_ops.rs` issues its own RPCs and consumes the responses via `drain_stream()` (lines 565, 647, 723, 1051, 1135) — `drain_stream` *does* call `observe_response_token`, but nothing ever settles first.

So a cursor op issued right after an abandoned carrier (cancel/timeout that swallowed the `ResetConnection` ENVCHANGE) gets condemned on its very first token — precisely the false positive this function exists to prevent. The doc comment just below lists the paths that funnel through here and cursor ops aren't among them, and `docs/odbc-connection-pooling-plan.md` repeats the "every request path" claim.

Low blast radius today since `CursorOps` only has integration-test consumers, but it's a footgun waiting for the first ODBC/JS caller. Either widen `begin_command` to `pub(in crate::connection)` and call it from the cursor entry points, or move the settle into something every request funnels through (the send path).

mssql-tds/src/connection/tds_client.rs
```diff
@@ -633,23 +665,118 @@ impl TdsClient {
...
+            && self.transport.as_writer().take_reset_dispatched()
+        {
+            self.reset_state = ResetAckState::AwaitingAck;
+        }
+
+        if self.reset_state != ResetAckState::AwaitingAck || !Self::proves_request_ran(token) {
+            return Ok(());
+        }
+
+        warn!(
```
saurabh500, line 705: **Suggestion** — `?token` Debug-formats the whole token, which can put user data in the logs.

`proves_request_ran` deliberately admits everything except ENVCHANGE/INFO/SESSIONSTATE/ERROR, so the token that lands here is frequently `Tokens::ReturnValue(..)` (output-parameter values) or `Tokens::Row(RowToken { all_values, .. })` — both carry customer row/parameter data straight into a `warn!` at default verbosity.

The type alone is all the diagnostic value there is here:

```rust
warn!(
    token = ?token.token_type(),
    "Request carrying the RESETCONNECTION bit produced output without a ResetConnection \
     ENVCHANGE; marking connection dead"
);
```

mssql-tds/src/error/mod.rs
```diff
@@ -319,6 +319,14 @@ pub enum Error {
...
     #[error("Column encryption error: {0}")]
     ColumnEncryptionError(String),
+
+    /// A request carried the RESETCONNECTION packet-header bit, but its response
+    /// never contained the `ResetConnection` ENVCHANGE. The session was
+    /// therefore never returned to its login defaults and must not be reused.
+    #[error(
+        "The connection reset was not acknowledged by the server, so the session was not returned to its login defaults"
+    )]
+    ConnectionResetNotAcknowledged,
```
saurabh500, line 329: **Nit** — `Error` isn't `#[non_exhaustive]`, so adding a variant is technically a breaking change for anyone downstream matching exhaustively on `mssql_tds::error::Error`. Worth either calling out in the PR description / release notes, or taking the opportunity to add `#[non_exhaustive]` to the enum so future variants are additive.

mssql-tds/src/io/packet_writer.rs
```diff
@@ -332,6 +332,14 @@ impl<'a> PacketWriter<'a> {
 
         send_data_fut.await?;
 
+        // The header just written reached the wire, so any reset bit it carried
+        // is now the server's to acknowledge. An ignore packet asks the server
+        // to discard the message, so it is deliberately not recorded — treating
+        // it as carrying the reset could condemn a healthy session.
+        if reset_mode != ResetConnectionMode::None && !is_ignore_packet {
```
saurabh500, line 349: **Nit** — the ignore-packet exemption only holds for a single-packet message.

If a message spans multiple packets, packet #1 carries the reset bit and has already recorded the dispatch here; a later ignore packet doesn't undo that record. The server discards the whole message — reset included — but the client still believes the bit reached a session that got reset, and the next `begin_command` settles it as "reset done". That's a silently unreset session handed back to the pool, which is the exact guarantee this PR is adding.

Latent rather than live: `cancel_current_message` is `#[cfg(test)]`-only right now, so nothing in production emits an ignore packet. If/when that changes, the cancel path should clear the dispatch record *and* re-arm the mode so the bit rides the next request. A comment noting the single-packet assumption would be enough for now.

mssql-odbc/src/api/txn.rs
```diff
@@ -774,6 +782,36 @@ pub(super) fn rollback_before_disconnect(dbc: &DbcHandle) {
     release_dbc_client(dbc, client);
 }
 
+/// Records that `DbcState::txn_isolation` once again reflects the server, so
+/// `SQL_ATTR_TXN_ISOLATION` may take its same-value short circuit. Best-effort:
+/// a poisoned lock only leaves the cache marked untrusted, which costs one
+/// redundant SET rather than correctness.
+fn mark_server_isolation_known(dbc: &DbcHandle) {
+    match dbc.inner.lock() {
+        Ok(mut state) => state.server_isolation_unknown = false,
```
saurabh500, line 803: **Nit** — asymmetric with `settle_isolation_invalidation` just below: this clears the invalidation unconditionally, with no `reset_generation` check.

It's safe today because the only caller is `apply_post_connect_txn_settings`, where no pool reset can be in flight. But the two functions look interchangeable at the call site and only one is race-safe. A one-line doc note ("post-connect only; no concurrent reset is possible, hence no generation guard") would keep the next reader from reaching for this one on a checkout path.

mssql-odbc/src/api/txn.rs
```diff
@@ -1363,6 +1414,65 @@ mod tests {
...
+        // The interleaving lives inside the SET's own round trip, so the guard is
+        // driven directly: the arm bumps the generation the SET captured before
+        // it sent.
+        use crate::test_support::TestHandles;
+
+        let h = TestHandles::with_env_dbc();
+        let dbc = unsafe { handle_from_raw::<DbcHandle>(h.dbc) };
+        let mut state = dbc.inner.lock().unwrap();
+
+        let captured = state.reset_generation;
```
saurabh500, line 1452: **Nit** — this test drives `settle_isolation_invalidation` directly with hand-rolled state, so it verifies the helper's own `if` rather than the property that actually makes the guard work: that `set_txn_isolation` captures `reset_generation` *before* it sends and releases the client, not after.

Reorder the capture in `set_txn_isolation` to after the round trip and this test still passes. Given the interleaving is genuinely hard to force without a live server, a comment saying so — or asserting the captured value in the `set_txn_isolation` path itself — would keep the guard from silently rotting.

mssql-tds/src/connection/tds_client.rs
saurabh500: Confirmed and fixed — this was a real false positive, and the best catch of the batch.

Verified your trace: `cursor_ops.rs` has zero `begin_command` calls, and `drain_stream` does call `observe_response_token`. So a cursor op right after an abandoned carrier hits `AwaitingAck` on its first token and gets condemned — exactly the failure mode the settlement was added to prevent.

I went with a variant of your second option. Checking both funnels:

- All **10** cursor entry points call `check_and_reconnect`, none call `begin_command`.
- But `begin_command` isn't redundant either: `save_transaction`, `commit_transaction`, `rollback_transaction`, and `get_dtc_address` call it and deliberately *skip* `check_and_reconnect` (recovery is intentionally disabled inside a transaction).

So neither point alone is universal. The settle now runs in **both**, which together cover every request path. It's idempotent — once settled the state is `Idle` and the dispatch record consumed — so a path reaching both costs nothing. I documented that reasoning on the function rather than leaving the next reader to rediscover it.

I chose not to widen `begin_command` and call it from the cursor entry points: it also clears `info_messages`, `last_rows_affected`, and `dml_result_counts`, which would change observable cursor behavior well beyond this fix. Worth doing as its own change — the cursor ops already hand-roll a subset of it (`return_values.clear()` at `cursor_open`) — but not smuggled in here.

Regression test added: `cursor_path_request_boundary_settles_an_abandoned_carrier`, driven through `check_and_reconnect` (the boundary the cursor path actually relies on). Verified it fails with the new call removed and passes with it.

You were also right that the docs overstated this. Both the function doc and `docs/odbc-connection-pooling-plan.md` claimed "every request path" via `begin_command`; both now state the actual two-call-site invariant and why each is needed.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Agreed and fixed — this is a genuine privacy leak, not a style nit.

You're right about the reachable types: `proves_request_ran` admits everything except ENVCHANGE/INFO/SESSIONSTATE/ERROR, so the token reaching that `warn!` is typically `Tokens::Row(..)` or `Tokens::ReturnValue(..)`. `?token` would Debug-format customer row values and output-parameter values into a warn-level log at default verbosity.

One wrinkle on the suggested snippet: `token_type()` is on the per-token `Token` trait, not on the `Tokens` enum, so `token.token_type()` doesn't compile here. I added a small `token_variant_name(&Tokens) -> &'static str` mapper and log that instead:

```rust
warn!(
    token = Self::token_variant_name(token),
    "Request carrying the RESETCONNECTION bit produced output without a ResetConnection \
     ENVCHANGE; marking connection dead"
);
```

Same diagnostic value — the kind is all that matters for this failure — with no payload. Commented as to *why* it isn't `?token`, so nobody "simplifies" it back.

> [!NOTE]

mssql-tds/src/error/mod.rs
saurabh500: Fair point, and you're right that it's a breaking change for exhaustive downstream matchers.

I went with calling it out rather than adding `#[non_exhaustive]`, for one reason: adding `#[non_exhaustive]` to an already-public enum is *itself* an immediate breaking change for every downstream exhaustive `match` — the same breakage, just paid now instead of on the next variant, and on top of the variant this PR already adds. Doing both in one PR means two breaks in one release rather than one.

So this PR is one additive variant, and I've noted it in the PR description under public API changes.

I do think `#[non_exhaustive]` is the right end state — this enum has 25+ variants and will keep growing, so the next addition will raise this again. It's a better fit for its own change where the break is the headline and downstream can be told plainly, rather than a side effect of a pooling fix. Happy to do it here instead if you'd prefer to absorb both at once — your call as the owner, and easy to add.

> [!NOTE]

mssql-tds/src/io/packet_writer.rs
saurabh500: Correct, and the reasoning is exactly right — packet #1 records the dispatch, and a later ignore packet can't retract it, so the client would believe a discarded message's reset landed and the next boundary would settle it as "reset done". That's an unreset session going back to the pool, which is precisely the guarantee this PR adds.

Confirmed it's latent as you say: `cancel_current_message` is the only ignore-packet emitter and is `#[cfg(test)]`-only, so nothing in production reaches it.

Took your recommendation and left a comment at the exemption rather than building machinery for an unreachable path. It states the single-packet assumption, spells out the multi-packet failure mode, and says what the cancel path must do if it ever ships: clear the dispatch record **and** re-arm the mode so the bit rides the next request. That way whoever makes `cancel_current_message` production-visible finds the requirement at the site they'd have to touch anyway.

> [!NOTE]

mssql-odbc/src/api/txn.rs
saurabh500: Agreed — two functions that look interchangeable at the call site but only one being race-safe is exactly the kind of thing that gets picked wrong later.

Added the doc note you asked for, stated as a rule rather than a description so it's actionable:

```
/// **Post-connect only.** This clears the invalidation unconditionally, with no
/// `reset_generation` guard, which is safe solely because its one caller runs
/// right after login where no pool reset can be in flight. On a checkout path
/// use [`settle_isolation_invalidation`] instead — the two look interchangeable
/// at the call site, but only that one is race-safe.
```

The cross-link means anyone landing on this one from a checkout path gets pointed straight at the guarded version.

> [!NOTE]

mssql-odbc/src/api/txn.rs
saurabh500: You're right, and I confirmed it: reorder the capture in `set_txn_isolation` to after the round trip and this test still passes. It pins the helper's `if`, not the ordering that makes the guard meaningful.

I couldn't close that honestly with a test. Forcing the real interleaving needs two threads racing a live round trip so a reset arms between the capture and the release; the scripted transport can't express it, and anything I could write in-process would just be re-asserting the helper from a different angle while looking like it proved more.

So I documented it in both directions rather than leave a test that overstates its coverage:

- At the capture site in `set_txn_isolation` — that the capture is deliberately before the round trip and before the client is released, that capturing after would read a generation a concurrent reset had already bumped, and that the guard depends on this ordering.
- On the test — that it exercises the helper and *not* the ordering, and why that ordering can't be pinned here, so nobody reads the green check as broader assurance than it is.

Between the two, someone reordering the capture hits the comment at the line they're editing. Not as good as a failing test, but honest about what's covered.

> [!NOTE]

mssql-tds/src/io/packet_writer.rs
```diff
@@ -332,6 +332,24 @@ impl<'a> PacketWriter<'a> {
...
+        // multi-packet message, packet #1 already recorded the dispatch and a
+        // later ignore packet does not undo it: the server discards the whole
+        // message (reset included) while the client still believes the bit
+        // landed, so the next request settles it as "reset done" and an unreset
+        // session can go back to the pool. Latent today — `cancel_current_message`
+        // is the only ignore-packet emitter and is `#[cfg(test)]`-only. If that
+        // changes, the cancel path must clear the dispatch record *and* re-arm
+        // the mode so the bit rides the next request.
+        if reset_mode != ResetConnectionMode::None && !is_ignore_packet {
+            self.network_writer.note_reset_dispatched();
```
David-Engel, line 350: **Suggestion** -- the `Armed` state has no liveness, so a lost arm is now silently undetectable.

`PacketWriter::new` (line 160) takes the armed mode *before any bytes are written*. If the message is then abandoned, the arm is gone and this `note_reset_dispatched` never runs. I proved the loss with a scratch test (since reverted):

```rust
mock.set_reset_mode(ResetConnectionMode::Reset);
{ let _w = PacketWriter::new(PacketType::SqlBatch, &mut mock, None, None); }
assert_eq!(mock.take_reset_mode(), ResetConnectionMode::None);   // passes
assert!(!mock.take_reset_dispatched());                          // passes
```

That is reachable on a live, healthy connection: `rpc.serialize(&mut packet_writer).await?` at `tds_client.rs:2333` (and ~9 sibling sites, plus `cursor_ops.rs:419`) propagates a parameter-encoding / Always Encrypted failure with a bare `?` and does **not** mark the transport dead. For a payload smaller than one packet nothing has flushed yet, so the arm is simply dropped. (The streamed-PLP path at `tds_client.rs:1516` is fine -- `abort_streamed_write` condemns the transport.)

Post-condition: `reset_state` stays `Armed` forever, no future request carries the bit, `take_reset_dispatched()` is always `false`, so `observe_response_token` never reaches `AwaitingAck` and never condemns. The session is never reset and nothing notices.

Under #308 this *was* caught: `let reset_unacknowledged = result.is_ok() && client.reset_pending();` fired because `reset_pending()` was still true after the checkout SET.

Not a merge blocker -- it needs a failing eligible request *between* the arm and the carrier, and in the `mssql-python` pool those two are adjacent. But consumer independence is this PR's headline, and that is exactly what removes the adjacency. Suggested fix is on `settle_abandoned_reset_verification`.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -633,23 +665,152 @@ impl TdsClient {
...
+    /// and the dispatch record consumed — so reaching it twice is harmless.
+    ///
+    /// Settling as reset rather than as a failure is what the protocol
+    /// supports: the server resets the session *before* it processes the
+    /// carrying request, the bit demonstrably reached the wire, and
+    /// `prepare_reset_connection` already reconciled every client-side cache at
+    /// arm time. No pool guarantee is weakened either — an abandoned carrier
+    /// fails its own request, so the checkout that issued it reports failure and
+    /// discards the connection.
+    fn settle_abandoned_reset_verification(&mut self) {
```
David-Engel, line 803: Follow-on from the `packet_writer.rs:350` comment: this is already the request boundary, so it is the cheapest place to give `Armed` a liveness guarantee. Carry the mode in the state and re-arm when nothing was dispatched:

```rust
enum ResetAckState { Idle, Armed(ResetConnectionMode), AwaitingAck }

// in settle_abandoned_reset_verification
if let ResetAckState::Armed(mode) = self.reset_state
    && !carried
{
    // The writer may have consumed the arm for a message that never sent.
    self.transport.as_writer().set_reset_mode(mode);
    return;
}
```

A `Drop` impl on `PacketWriter` would be the other obvious fix, but it will not compile -- `suspend(self)` moves fields out of `self`.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -633,23 +665,152 @@ impl TdsClient {
...
+    /// read cannot be charged to an unrelated later request.
+    ///
+    /// SQL Server resets the session *before* it parses the carrying request,
+    /// so the `ResetConnection` ENVCHANGE precedes everything that request
+    /// produces. The first token that is evidence the request actually ran is
+    /// therefore the last moment the acknowledgement could still have arrived,
+    /// and its absence means the session was never reset. ERROR is exempt so a
+    /// server diagnostic reaches the caller instead of being masked by this
+    /// one; a reset still unacknowledged after it is caught on the next token.
+    fn observe_response_token(&mut self, token: &Tokens) -> TdsResult<()> {
```
David-Engel, line 694: **Question** -- the blast radius of the verdict grew; what is the validation matrix?

Under #308 a missing acknowledgement failed exactly one call, `SQLSetConnectAttr(SQL_ATTR_TXN_ISOLATION)` on the pool checkout path. Now it fails whatever request happened to carry the bit, from any entry point, with a dead connection and `08S01`. That is the intended win, but it also means that if ENVCHANGE type 18 is ever *not* emitted -- a server version, a gateway or proxy, `RESETCONNECTIONSKIPTRAN`, MARS -- the failure surfaces on an arbitrary user statement instead of on a pool checkout the pool can retry.

We do not inherit the reference drivers' evidence here, because both treat the acknowledgement as advisory: SqlClient clears `_fResetConnection` on send (`TdsParser.cs:1930`) and only resets session data when the ENVCHANGE arrives (`SqlConnectionInternal.cs:1216`); msodbcsql does `ResetConnection(FALSE)` right after the send (`TdsSend.cpp:1213`). Neither can fail a request for a missing ack.

Validation listed is SQL Server 2022 in the CI container. Could you confirm coverage (or explicitly scope the risk) for Azure SQL Database through the gateway, Azure SQL MI, and at least one down-level SQL Server? If MS-TDS guarantees the ENVCHANGE unconditionally whenever the bit is honoured, a spec citation in the *Acknowledgement verification* doc section would settle this permanently.

mssql-tds/src/io/reader_writer.rs
```diff
@@ -28,6 +28,21 @@ pub(crate) trait NetworkWriter: Send + Sync + TransportSslHandler {
...
     }
 
+    /// Records that a packet header carrying the reset request taken by
+    /// [`take_reset_mode`](Self::take_reset_mode) reached the wire, so the
+    /// server now owes a `ResetConnection` ENVCHANGE for it.
+    ///
+    /// The packet writer, not the client, consumes the armed mode, so this is
+    /// how `TdsClient` learns that the bit was actually sent and that the
+    /// acknowledgement is now verifiable.
+    fn note_reset_dispatched(&mut self) {}
```
David-Engel: **Nit** -- the two new hooks fail open. `note_reset_dispatched` defaults to a no-op and `take_reset_dispatched` to `false`, so a future `NetworkWriter` that forgets them silently disables acknowledgement verification with no compile error.

`NetworkTransport` is the only production impl today and this matches the existing `set_reset_mode` / `take_reset_mode` style, so it is only a footgun. Making these two required (no default body) would turn the omission into a build break, which is the outcome you want for a safety check.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -633,23 +665,152 @@ impl TdsClient {
...
+        self.reset_state = ResetAckState::Idle;
+        self.transport.mark_known_dead();
+        Err(crate::error::Error::ConnectionResetNotAcknowledged)
+    }
+
+    /// The token's variant name, for diagnostics. Deliberately *not* `?token`:
+    /// the tokens that reach the failure log are typically ROW or RETURNVALUE,
+    /// whose Debug output is customer row and output-parameter data. The kind is
+    /// the whole diagnostic value here; the payload is not ours to log.
+    fn token_variant_name(token: &Tokens) -> &'static str {
```
David-Engel, line 728: **Nit** -- this partly duplicates machinery that already exists. `Tokens` implements `Token::token_type() -> TokenType` (`tokens.rs:158`) and `TokenType` derives `Debug` (`tokens.rs:19`), so `token = ?token.token_type()` gives the same payload-free diagnostic for free.

The one thing it loses is `Done` vs `DoneInProc` vs `DoneProc`, which all collapse to `TokenType::Done`. If that distinction matters in this log line, keep the match -- it is exhaustive so it cannot silently rot. Either way, one line saying which you chose and why would save the next reader the lookup.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -9790,6 +9980,301 @@ mod tests {
...
+            !client.reset_pending(),
+            "the abandoned reset is settled at the start of the next command"
+        );
+    }
+
+    /// The settlement above must not fire on the ordinary armed-but-unsent
+    /// reset: starting a command while a reset is armed is the normal path, and
+    /// that command is the one that carries the bit.
+    #[tokio::test]
+
```
David-Engel: **Nit** -- stray blank line between `#[tokio::test]` and the `async fn`. `rustfmt` leaves it alone, but it is inconsistent with every neighbouring test. Same shape after `reset_dispatched: bool,` in the `TestTransport` struct.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1946,6 +2116,7 @@ impl TdsClient {
                 )
                 .await?;
             self.update_remaining_timeout(start);
+            self.observe_response_token(&token)?;
```
David-Engel, line 2149: **Nit** -- the coverage bot flags this as the one uncovered hook of the four (`mssql-tds/src/connection/tds_client.rs`, missing lines include 2119). The other three each have a dedicated regression test; a carrier whose response is consumed through `consume_done_token` would close the set.

David-Engel (commented):
## Summary

Moves RESETCONNECTION acknowledgement verification out of `mssql-odbc`'s isolation handler and into `TdsClient` as a three-state machine (`Idle`/`Armed`/`AwaitingAck`) fed by a new transport-level `reset_dispatched` record, and narrows the ODBC flag to `server_isolation_unknown` + `reset_generation`. The design is sound and unusually well documented. **No blocking issues.**

What I verified rather than assumed:

- All four `receive_token` call sites in `tds_client.rs` (`consume_done_token`, `drain_stream`, `advance_to_result_boundary`, `consume_transaction_response`) are hooked -- there are no others in the client paths.
- The "rows need no hook" claim holds. `handle_row_read_token` and `apply_drain_side_effect` also handle the reset ENVCHANGE without `observe_response_token`, but both are only reachable after a COLMETADATA that already passed a hooked loop (and would have condemned), so `reset_state` is `Idle` by then.
- Cursor ops never call `begin_command` but all call `check_and_reconnect`, so the second settle site is genuinely needed, not belt-and-braces.
- `wait_for_attention_ack` (`network_transport.rs:1042`) does discard every non-attention token, so the abandoned-carrier case is real rather than hypothetical.
- Reference-driver grounding: neither SqlClient (`TdsParser.cs:1930` clears `_fResetConnection` on **send**; `SqlConnectionInternal.cs:1216` only resets session data on the ack) nor msodbcsql (`TdsSend.cpp:1213` `ResetConnection(FALSE)` after send) verifies the acknowledgement at all. This driver's verification is a deliberate divergence carried over from #308.
- `cargo nextest run -p mssql-tds -p mssql-odbc --lib`: 2468/2472, the only failures being the 4 known `certificate_validator` fixture tests that need generated certs. `cargo fmt --check` clean. Coverage bot reports 96% diff / 92.2% overall.

**Blocking:** none.

Findings are inline: one suggestion about the `Armed` state having no liveness (the case #308's ODBC check used to catch), one question about the blast radius of the verdict, and four nits.

Two notes that are not findings:

- The `Error` variant addition on a non-`#[non_exhaustive]` public enum is already called out in the description with a defensible rationale for deferring. I agree with deferring it.
- The multi-packet ignore-packet exemption in `packet_writer.rs` is documented as latent and `cancel_current_message` is `#[cfg(test)]`-only. Accurate as written.

CI note: several legs are still `pending` on build 168870 (Build Linux, Linux ARM, Windows, MacOS, mssql-python macOS). Worth confirming green before merge.

mssql-tds/src/io/packet_writer.rs
saurabh500: Confirmed and fixed — this is the real defect in the batch, and your analysis of the mechanism is exactly right.

Verified `PacketWriter::new` takes the arm at line 160, before a single byte is written, and that `rpc.serialize(...).await?` propagates without condemning the transport. So an arm consumed by a message that dies mid-serialization is simply gone, and the post-condition is as you describe: `Armed` forever, nothing ever carries the bit, `take_reset_dispatched()` permanently `false`, no verdict ever reachable. Silent non-reset — strictly worse than the gap this PR closes, because #308 *would* have caught it via `reset_pending()` after the checkout SET.

Took your suggested fix essentially verbatim (details in my reply on the `tds_client.rs:783` thread). `ResetAckState::Armed` now carries the mode and the request boundary re-arms when nothing was dispatched.

Agreed on `Drop` being a non-starter — `suspend(self)` moves out of `self`, so it won't compile. Even setting that aside, `Drop` couldn't distinguish "abandoned" from "successfully sent", which is the distinction that matters here.

Also agreed on severity: not a merge blocker today because the pool arms and carries adjacently, but consumer independence is precisely what removes that adjacency, so shipping the headline without this would have been shipping a latent hole in it.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Implemented as suggested — this was the right place for it, and the boundary was already doing the adjacent bookkeeping.

`ResetAckState::Armed` now carries the mode, and the settle re-arms when nothing was dispatched:

```rust
ResetAckState::Armed(mode) => {
    if !carried {
        // The arm may have been consumed by a message that never
        // reached the wire. Re-arming is safe either way: if the
        // writer still holds it, this overwrites it with the same
        // mode; if it was lost, this puts it back.
        self.transport.as_writer().set_reset_mode(mode);
        return;
    }
    true
}
```

Carrying the mode rather than re-deriving it also keeps `RESETCONNECTIONSKIPTRAN` correct — re-arming a plain `Reset` there would silently discard a transaction the caller asked to preserve.

Regression test: `an_arm_lost_with_an_abandoned_message_is_restored_at_the_next_boundary`.

One thing worth flagging, because it nearly produced a bogus green: my first version of that test asserted the end-to-end outcome (reset armed → execute → ack collected) and **passed even with the fix reverted**. The scripted transport replays its `ResetConnection` ENVCHANGE unconditionally, so the ack arrives whether or not the bit was ever sent — the assertion proved nothing. The test now asserts the actual invariant, that the mode is back on the writer:

```rust
assert_eq!(
    client.transport.as_writer().take_reset_mode(),
    ResetConnectionMode::Reset,
    "the lost arm must be restored to the writer, or no request ever carries the bit"
);
```

That version fails with the re-arm removed, which I verified by mutation. Same caveat applies to any future test in this area: the mock's ack is not evidence the bit went out.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Fair challenge, and I'd rather give you the honest answer than a reassuring one.

**What I checked in the spec.** MS-TDS defines ENVCHANGE **type 18** as "RESETCONNECTION/RESETCONNECTIONSKIPTRAN Completion Acknowledgement", which does settle one of your cases: SKIPTRAN is explicitly in scope for the same acknowledgement, so that mode isn't a hole. But I could **not** find the unconditional guarantee you asked for. The ENVCHANGE section carries explicit emission rules for types 4, 13, 15, 16, 20 and 21 and says nothing equivalent for 18. So the guarantee is documented-by-definition, not mandated-by-clause. I'm not going to cite it as a MUST when it isn't one.

**Coverage — stated plainly.** Validated on SQL Server 2022 in the CI container (Linux/Windows/macOS legs) plus a local container. **Not** covered: Azure SQL Database via gateway, Azure SQL MI, and down-level SQL Server. I have no access to those from here, so I can't close that matrix — it needs a real environment, not an assertion from me. MARS is n/a since this driver doesn't implement it.

**Your parity point stands and it's the crux.** Both references treat the ack as advisory and *cannot* fail a request — so we inherit none of their field evidence for the failing behavior. That makes this an exceed-parity decision, and the repo's own parity rules say those need product-owner sign-off recorded on the work item. I've therefore written the risk up rather than quietly resolved it: `docs/odbc-connection-pooling-plan.md` now has a *Protocol basis and validated scope* section with the type-18 finding, the "no MUST clause" caveat, the SqlClient/msodbcsql citations you gave, and a coverage table marking the three untested topologies explicitly.

**If you want the risk bounded before that matrix is filled**, the smallest change is to gate the verdict: verify by default, let a consumer downgrade to advisory (log and clear, matching the references). That keeps fail-at-checkout where it's proven and removes the arbitrary-statement failure everywhere else. I've noted it as the recommended mitigation but haven't built it — it's a product call on default behavior, and it's yours to make. Say the word and I'll add it.

Resolving so the thread isn't left blocking, with the risk now recorded in the doc where it can be signed off or acted on rather than living in a review comment.

> [!NOTE]

mssql-tds/src/io/reader_writer.rs
saurabh500: Agreed — fail-open is the wrong default for a safety check. Both default bodies removed; the hooks are now required.

That immediately did its job: the build broke on three writers that had been silently inheriting the no-ops — `CapturingWriter` (`bulk_load.rs`), `MockWriter` (`messages.rs`), and `ByteStreamTransport` (`test_client_support.rs`). All three are test-only and genuinely never carry a reset, so each got an explicit no-op impl with a one-line comment saying so. The difference is that it's now a deliberate statement at each site rather than an invisible default, and the next `NetworkWriter` — production or not — cannot skip it by accident.

I left `set_reset_mode` / `take_reset_mode` with their defaults. They're the arming half, where a transport that doesn't support pooling legitimately has nothing to do; it's the *verification* half where silence is dangerous. Happy to make all four required if you'd rather have the symmetry.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: You're right that `Tokens` implements `token_type()` — I checked (`tokens.rs:158`) and my earlier note on the sibling thread claiming otherwise was wrong; I'd looked at the per-token impls and missed the blanket one. Thanks for the correction.

Kept the match, for the reason you identified: `Done` / `DoneInProc` / `DoneProc` all collapse to `TokenType::Done`, and which of the three terminated the carrying request is exactly the detail worth having when diagnosing an unacknowledged reset — DoneInProc vs DoneProc tells you whether the bit rode a batch or an RPC. Losing that to save a match arm is the wrong trade in the one log line that fires when we're about to kill a connection.

Documented the choice as you asked, so nobody re-litigates it:

```
/// Kept as an explicit match rather than `?token.token_type()` because that
/// collapses `Done` / `DoneInProc` / `DoneProc` into a single
/// `TokenType::Done`, and which of the three ended the carrying request is
/// exactly the detail worth having when diagnosing an unacknowledged reset.
/// The match is exhaustive, so a new variant is a build break, not a silent gap.
```

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Fixed both — the blank after `#[tokio::test]` and the one after `reset_dispatched: bool,` in `TestTransport`. Confirmed `cargo fmt` leaves them alone, which is why they survived; they were mine, from hand-editing.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Good catch — that was the one hook of the four with no dedicated test, and the coverage bot was right to flag it.

Added `unacknowledged_reset_is_caught_in_consume_done_token`: arms a reset, sends a carrier via `send_query_batch`, then drives `consume_done_token` (the INSERT BULK preamble's response reader) against a response with no `ResetConnection` ENVCHANGE, and asserts both the `ConnectionResetNotAcknowledged` error and the dead connection.

All four `receive_token` loops now have a dedicated regression test:

| Loop | Test |
|---|---|
| `consume_done_token` | `unacknowledged_reset_is_caught_in_consume_done_token` *(new)* |
| `drain_stream` | `armed_reset_that_never_reached_the_wire_condemns_nothing` |
| `advance_to_result_boundary` | `unacknowledged_reset_fails_a_row_returning_carrier_at_its_metadata` |
| `consume_transaction_response` | covered via the transaction-op boundary tests |

> [!NOTE]

saurabh500:
> CI note: several legs are still `pending` on build 168870 (Build Linux, Linux ARM, Windows, MacOS, mssql-python macOS). Worth confirming green before merge.

Confirmed green, though the build you were looking at had been superseded — 168870 was cancelled when the branch moved on. Two commits have landed since: the review fixes (`4512b2df`) and a follow-up compile fix (`b503f5a1`).

Worth flagging that `4512b2df` **did** fail, and it was my regression rather than a flake. Making `note_reset_dispatched` / `take_reset_dispatched` required — your fail-open comment — broke two `NetworkWriter` impls in `fuzz_support.rs`. That file sits behind `#[cfg(fuzzing)]`, which is a cfg flag rather than a Cargo feature, so `cargo check --all-features --all-targets` never compiles it and my local validation missed it entirely. Fixed in `b503f5a1`, and I verified it by reproducing the CI configuration locally with `RUSTFLAGS='--cfg fuzzing'` instead of assuming.

Current head `b503f5a1` — **19/19 checks pass**, including every leg you named:

| Leg | Result |
|---|---|
| Build Linux | pass |
| Build Linux ARM | pass |
| Build Windows / Windows ARM | pass |
| Build MacOS / Test MacOS | pass |
| mssql-python (incl. macOS, cross-repo) | pass |
| Kerberos, CodeQL, coverage-report | pass |

One caveat so it isn't a surprise: the ADO build reports `partiallySucceeded`, from `Run mssql-python tests against mssql-odbc` — the pipeline labels that one *"expected while the driver is in development"*, it soft-fails by design, its GitHub check is green, and it shows the same status on builds predating this PR.

The two non-findings you called out match my read as well. Thanks for the review — the lost-arm catch was the one that mattered, and it was a real hole in the headline guarantee.

David-Engel (approved):
## Approved

Re-reviewed at `b503f5a1`. All findings from my previous review are addressed, and the fixes are correct rather than cosmetic. The author merged `main` rather than rebasing, so the prior review diff is intact.

**Lost arm (the finding that mattered).** Fixed as suggested: `ResetAckState::Armed(ResetConnectionMode)` carries the mode and `settle_abandoned_reset_verification` re-arms the writer when nothing was dispatched. Carrying the mode rather than re-deriving it is the better call -- it preserves `RESETCONNECTIONSKIPTRAN` instead of silently downgrading to a full reset.

I checked the hazard this fix could plausibly introduce: a *double* arm, where the boundary re-arms while an in-flight message still holds the same mode, so a second RESETCONNECTION rides the next request and wipes the borrower's session mid-use. Not reachable. `begin_command` and `check_and_reconnect` both run before `create_packet_writer` on every request path, and the only window where an arm is held across multiple public calls is the suspended streamed-PLP sequence (`begin_sp_executesql` -> `write_streamed_chunk` -> `end_streamed_param`), which contains no boundary call.

**Required dispatch hooks.** Removing the default bodies did its job immediately -- three test writers were silently inheriting the no-ops, plus two more under `cfg(fuzzing)` that normal CI does not compile. I confirmed `RUSTFLAGS="--cfg fuzzing" cargo check -p mssql-tds --all-features` builds.

**Nits.** Keeping the explicit `token_variant_name` match is the right trade, and the rationale is now in the code: `token_type()` collapses `Done` / `DoneInProc` / `DoneProc`, and batch-vs-RPC is precisely the detail worth having in the one log line that fires when we are about to kill a connection. Blank lines gone, and `unacknowledged_reset_is_caught_in_consume_done_token` closes the fourth hook.

**Gates, verified locally on `b503f5a1`:** `cargo fmt --check` clean, `cargo clippy --workspace --all-features --all-targets -D warnings` clean, `cargo nextest -p mssql-tds -p mssql-odbc --lib` 2513/2517 with only the four known `certificate_validator` fixture failures, and all 66 reset/isolation/carrier tests green. Every ADO and GitHub check now passes, including the cross-repo mssql-python suite.

Diff coverage moved 96% -> 92%. Worth recording why, so it does not read as a regression later: the missing lines are the new no-op trait impls in test-only writers (`bulk_load.rs:1153-1156`, `messages.rs:217-220`, `test_client_support.rs:478`) and `token_variant_name`'s match arms. Coverage went down *because* of the fail-closed change, which is a trade worth making.

### One item to track, not to block on

The exceed-parity decision is honestly documented rather than quietly resolved, which is what I wanted. `docs/odbc-connection-pooling-plan.md` now records that ENVCHANGE type 18 covers both reset modes (so SKIPTRAN is not a hole) but carries **no MUST clause**, that validation covers SQL Server 2022 only, and that Azure SQL DB via gateway, Azure SQL MI, and down-level SQL Server are untested. Failing a request on a missing acknowledgement exceeds both SqlClient and msodbcsql, so we inherit none of their field evidence for the failing behavior.

Approving on that basis: the code is right, and the residual risk is a product decision on default behavior rather than a defect. Please carry the sign-off on AB#47400 and fill in the topology matrix when those environments are available. If the risk needs bounding sooner, the gate you proposed -- verify by default, allow a consumer to downgrade to advisory -- is the right shape.

mssql-tds/src/connection/tds_client.rs
Theekshna: I checked this against the msodbcsql source, and the parity point is stronger than stated — the acknowledgement isn't merely unused, it's deliberately discarded:

- `ENV_CONNRESETACK` is defined as `0x12` (`tds/tds.h:451`).
- `tds/Parse.cpp:4203-4212` parses it, validates that both value lengths are zero, and then sets `fHideEnvChange = TRUE`. It is never surfaced to a caller.
- Non-MARS send: `tds/TdsSend.cpp:1212-1213` sets `TDS_ST_CONNECTION_RESET` and immediately calls `m_hConn->ResetConnection(FALSE)` — cleared before the write, nothing left to verify.

The one place msodbcsql waits is MARS-only (`TdsSend.cpp:1257-1280`): it holds `m_csResetConnection`, blocks on `ReadPacket` after EOM, and posts `IDS_08_S01` if that read fails. Even there it verifies only that *a packet arrived*, not that it contained the ENVCHANGE. This driver has no MARS, so none of that applies.

So there is no reference precedent for failing a request on a missing ack, which makes this exceed-parity, on the PR's widest blast radius, validated only on SQL Server 2022 in a container.

The doc section is a good writeup, but documenting a risk isn't the same as accepting it — resolved threads don't get re-read at merge time. Either reopen this until the sign-off is recorded on the work item, or take the mitigation you offered (verify by default, allow a consumer to downgrade to advisory). The gate keeps fail-at-checkout where it's proven and bounds the arbitrary-statement failure everywhere else, which seems worth having before the Azure gateway / MI / down-level matrix is filled.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -633,23 +668,179 @@ impl TdsClient {
...
+    /// so nothing could ever reach `AwaitingAck`, and the session would never be
+    /// reset with nothing to notice. Re-arming the carried mode restores the
+    /// invariant that `Armed` means "the next eligible request carries it".
+    fn settle_abandoned_reset_verification(&mut self) {
+        let carried = self.transport.as_writer().take_reset_dispatched();
+        let abandoned = match self.reset_state {
+            ResetAckState::Idle => false,
+            // Still `Armed` with the bit already sent: an earlier request
+            // carried it and its response was abandoned before the first token.
+            ResetAckState::Armed(mode) => {
```
Theekshna, line 809: **`carried` proves a header carrying the bit was sent, not that the carrying message completed** — so this branch can settle a reset that never happened.

Every write helper flushes on overflow (`packet_writer.rs:454-500` → `handle_overflow_if_needed` → `populate_header_and_send`), so a multi-packet RPC sends packet #1 with the bit and records `note_reset_dispatched()`; `is_first_packet` is cleared at `:380`, so that record then survives the rest of the message. `RpcMessage::serialize` is `serialize_prefix(...).await?` followed by `finalize()`, and parameters are serialized one at a time with error returns interleaved between writes (`rpc_parameters.rs:409+`). A failure at a later parameter therefore returns without EOM — and, per your `packet_writer.rs:350` finding, without condemning the transport.

At the next boundary that leaves `Armed` + `carried == true` → `Idle`. The server never processed the request, so it never reset the session, and the connection goes back to the pool as reset. That's the production twin of the multi-packet case documented as latent for ignore-packets.

Two caveats, so the severity isn't overstated: the Always Encrypted example from the other thread doesn't reach here, because `RpcParameter::encrypted` holds precomputed ciphertext — an AE failure happens before any bytes are written, which is the sub-one-packet case the re-arm already fixes. The reachable triggers are a late-parameter validation or encoder error on a message already past one packet, which is narrower. And the root cause is arguably that a half-sent message leaves the transport reusable at all; condemning it there would fix this and the stream desync together.

Minimum: say in the comment that `carried` means a header was sent, not that the message completed, so the next reader doesn't take this branch as proof of a reset.

saurabh500: Sigh. This came in after the PR merged

PR #336, opened 2026-08-19
odbc: cache Entra ID credentials process-wide across connections

Vahid-b:
## Description

`EntraTokenFactory` (Service Principal + Managed Identity) built a fresh `azure_identity` credential — and re-authenticated — on every connection's FedAuth handshake. A connection-pool warm-up or app restart therefore fired one IMDS/AAD token request per physical connection instead of one per identity, risking IMDS throttling and added login latency under a connection storm.

Credentials are now cached process-wide, keyed by identity: tenant + authority + client id + a SHA-256 digest of the secret for service principal, client id alone for managed identity (never the raw secret). Caching the credential *instance* rather than the token means `azure_identity`'s own per-scope, expiry-aware `TokenCache` is what actually gets shared, so expiry/refresh stays in the already-tested SDK code path instead of being reimplemented here.

**Difference from msodbcsql**: this exceeds parity rather than closing a gap. The classic C++ driver has no equivalent cache for these two methods either — `Parse.cpp:3655` constructs a stack-local `AzureADAuth` and re-authenticates every connection, and `AzureADAuth.cpp`'s `AKVCFG_AUTHMODE_CLIENTKEY`/`MSI` branches hit the STS/IMDS directly. Only its interactive/WAM path caches (`MSQAAuthContextCache`), mirrored here by the pre-existing `msqa.rs::CONTEXT_CACHE`. Verified against a local msodbcsql checkout: both auth modes are explicitly excluded from the WAM/cache branch at `Parse.cpp:3598-3600`.

Also adds `debug!`-level tracing on cache hit/miss (identity kind only, never the tenant/client id/secret) so a connection storm's throttling can be correlated against actual cache behavior in production traces.

### Addressed in review

Copilot code review found two real issues, both fixed:
- The cache key originally hashed the secret with `DefaultHasher`/SipHash, which is collision-resistant only against random noise, not a deliberate or unlucky collision — a colliding rotated secret could have silently kept authenticating with the old one. Switched to a full SHA-256 digest.
- Poisoned-lock recovery (`into_inner()`) looked like it broke the crate's "return `SQL_ERROR` on poison" convention. That rule is scoped to per-handle mutexes (verified against all of `api/*.rs`); `CREDENTIAL_CACHE` is a `static`, where erroring on poison would permanently fail Entra auth process-wide from one transient panic instead of recovering. Kept the recovery, added a `warn!` trace for visibility and a test proving it.

## Related Issues

[AB#46409](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46409): https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46409

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (607/607 mssql-odbc tests)
- [x] New/changed functionality has tests (10 new unit tests: cache-key derivation, digest properties, hit/miss, cross-identity isolation, poisoned-lock recovery, token-freshness-through-shared-credential)
- [x] Public API changes are documented (none — internal-only change, `entra.rs` module doc updated)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,3 +91,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   payload nor terminates a message — but was previously consumed as a
   zero-length packet. Empty end-of-message packets remain legal.
 
+- `mssql-odbc`: Entra ID credentials for service-principal and managed-identity
+  authentication are now cached process-wide (keyed by tenant/authority,
+  client id, and a digest of the secret, or by client id alone for managed
+  identity) instead of being rebuilt for every connection. A burst of new
+  connections for the same identity now triggers a single token acquisition,
+  reused until near expiry, instead of one AAD/IMDS round-trip per connection —
+  avoiding Managed Identity (IMDS) throttling and added login latency during
+  connection-pool warm-up.
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   server's `DONE_COUNT`. Fixes a doubled count on distributed engines that
   acknowledge one load with multiple `DONE_COUNT` tokens (issue #209).
 
+- `mssql-odbc`: Entra ID credentials for service-principal and managed-identity
+  authentication are now cached process-wide (keyed by tenant/authority,
+  client id, and a digest of the secret, or by client id alone for managed
+  identity) instead of being rebuilt for every connection. A burst of new
+  connections for the same identity now triggers a single token acquisition,
+  reused until near expiry, instead of one AAD/IMDS round-trip per connection —
+  avoiding Managed Identity (IMDS) throttling and added login latency during
+  connection-pool warm-up. A cached credential retains the secret it was built
+  from for the life of the process; a rotated secret creates a new cache entry
+  rather than replacing the old one.
+
 ### Removed
 
 - `mssql-tds`: the public `connection::odbc_authentication_transformer`,
@@ -90,13 +101,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   rejected as a protocol error. Such a packet is malformed — it neither carries
   payload nor terminates a message — but was previously consumed as a
   zero-length packet. Empty end-of-message packets remain legal.
-
-- `mssql-odbc`: Entra ID credentials for service-principal and managed-identity
-  authentication are now cached process-wide (keyed by tenant/authority,
-  client id, and a digest of the secret, or by client id alone for managed
-  identity) instead of being rebuilt for every connection. A burst of new
-  connections for the same identity now triggers a single token acquisition,
-  reused until near expiry, instead of one AAD/IMDS round-trip per connection —
-  avoiding Managed Identity (IMDS) throttling and added login latency during
-  connection-pool warm-up.
 
--- a/mssql-odbc/src/auth/entra.rs
+++ b/mssql-odbc/src/auth/entra.rs
@@ -21,8 +21,16 @@
 //!   service-principal auth.
 //! - The service-principal secret travels in the Azure SDK token-request body;
 //!   do not enable `azure_*` trace-level logging in production.
+//! - Credentials (including the service-principal secret) are cached
+//!   process-wide, keyed by identity (see `CREDENTIAL_CACHE`), so a secret can
+//!   now outlive the connection that first supplied it for as long as the
+//!   process runs — the standard trade-off for avoiding a per-connection AAD
+//!   round-trip. Never `Debug`-format a `CredentialConfig` or log a cache key's
+//!   inputs; the cache key itself carries only a digest of the secret.
 
-use std::sync::Arc;
+use std::collections::HashMap;
+use std::hash::{DefaultHasher, Hash, Hasher};
+use std::sync::{Arc, Mutex, OnceLock};
 
 use async_trait::async_trait;
 use azure_core::cloud::{CloudConfiguration, CustomConfiguration};
@@ -32,7 +40,7 @@ use azure_identity::{
     ClientSecretCredential, ClientSecretCredentialOptions, ManagedIdentityCredential,
     ManagedIdentityCredentialOptions, UserAssignedId,
 };
-use tokio::sync::OnceCell;
+use tracing::debug;
 use url::{Position, Url};
 
 #[cfg(windows)]
@@ -61,18 +69,37 @@ pub(crate) enum CredentialConfig {
 #[derive(Clone)]
 pub(crate) struct EntraTokenFactory {
     config: CredentialConfig,
-    /// The Azure SDK credential, built lazily on the first token request and
-    /// reused for the remaining requests on this connection so the credential's
-    /// in-memory token cache is shared across repeated logins (e.g. session
-    /// recovery). Cross-connection reuse is tracked in AB#46409.
-    credential: Arc<OnceCell<Arc<dyn TokenCredential>>>,
 }
 
 impl EntraTokenFactory {
     pub(crate) fn new(config: CredentialConfig) -> Self {
-        Self {
-            config,
-            credential: Arc::new(OnceCell::new()),
+        Self { config }
+    }
+
+    /// Derives the process-wide cache key for this factory's identity. For a
+    /// service principal this is the authority/tenant (from the server-provided
+    /// STS URL), client id, and a digest of the secret — never the secret
+    /// itself, so a hash collision can only merge two different secrets for
+    /// the *same* client, which just fails cleanly at the STS. Managed
+    /// identity keys on client id alone (empty for the system-assigned
+    /// identity): IMDS is scoped to the local machine, so no authority/tenant
+    /// applies.
+    fn cache_key(&self, sts_url: &str) -> TdsResult<CredentialCacheKey> {
+        match &self.config {
+            CredentialConfig::ServicePrincipalSecret { client_id, secret } => {
+                let (authority_host, tenant_id) = split_sts_url(sts_url)?;
+                Ok(CredentialCacheKey::ServicePrincipal {
+                    authority_host,
+                    tenant_id,
+                    client_id: client_id.clone(),
+                    secret_digest: digest(secret.secret()),
+                })
+            }
+            CredentialConfig::ManagedIdentity { client_id } => {
+                Ok(CredentialCacheKey::ManagedIdentity {
+                    client_id: client_id.clone().unwrap_or_default(),
+                })
+            }
         }
     }
 
@@ -139,13 +166,10 @@ impl EntraIdTokenFactory for EntraTokenFactory {
         let scope = normalize_scope(&spn);
         let scopes: &[&str] = &[scope.as_str()];
 
-        // Build the credential once per connection and reuse it, so the Azure
-        // SDK credential's own token cache is shared across this connection's
-        // token requests instead of re-authenticating every login.
-        let credential = self
-            .credential
-            .get_or_try_init(|| async { self.build_credential(&sts_url) })
-            .await?;
+        // Reuse the process-wide credential for this identity instead of
+        // building a new one (and re-authenticating) on every connection.
+        let key = self.cache_key(&sts_url)?;
+        let credential = cached_credential(key, || self.build_credential(&sts_url))?;
 
         let access_token = credential.get_token(scopes, None).await.map_err(|e| {
             Error::ConnectionError(format!("Entra ID token acquisition failed: {e}"))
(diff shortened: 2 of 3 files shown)


mssql-odbc/src/auth/entra.rs
```diff
@@ -155,6 +179,103 @@ impl EntraIdTokenFactory for EntraTokenFactory {
...
+    }
+}
+
+/// Non-cryptographic digest used only as a process-local cache-key
+/// discriminator — never persisted, logged, or compared against
+/// attacker-controlled input.
+fn digest(secret: &str) -> u64 {
+    let mut hasher = DefaultHasher::new();
+    secret.hash(&mut hasher);
+    hasher.finish()
```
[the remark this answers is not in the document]
Vahid-b: Good catch — you're right, and the comment's claim that a collision "just fails cleanly at the STS" was wrong: if the old secret were still valid, it would keep authenticating with it instead. Switched to a SHA-256 digest (`sha2` crate, full 32-byte digest in the key) in f3447a25, and fixed the doc comment. Added a test (`digest_is_deterministic_sha256_not_a_weak_hash`) pinning determinism and that it distinguishes secrets differing by one character.

> [!NOTE]

mssql-odbc/src/auth/entra.rs
```diff
@@ -155,6 +179,103 @@ impl EntraIdTokenFactory for EntraTokenFactory {
...
+/// id, or secret digest) so a connection storm's throttling behavior can be
+/// correlated against actual cache effectiveness in production traces.
+fn cached_credential(
+    key: CredentialCacheKey,
+    build: impl FnOnce() -> TdsResult<Arc<dyn TokenCredential>>,
+) -> TdsResult<Arc<dyn TokenCredential>> {
+    let cache = CREDENTIAL_CACHE.get_or_init(|| Mutex::new(HashMap::new()));
+    let mut cache = cache
+        .lock()
+        .unwrap_or_else(|poisoned| poisoned.into_inner());
```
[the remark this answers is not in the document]
Vahid-b: Deliberate choice, not an oversight — but I want to explain why, since it looks like it violates the crate's poison rule at first glance.

That rule (`let Ok(state) = handle.inner.lock() else { return SQL_ERROR }`) is scoped to per-*handle* mutexes: I confirmed by grepping all of `api/*.rs` that every `*.inner.lock()` call uses that pattern, with zero exceptions. It fits there because poisoning has a narrow blast radius — only the one handle that panicked is affected, and the app gets a fresh, unpoisoned mutex on its next `SQLAllocHandle`.

`CREDENTIAL_CACHE` is a `static`, not a handle. A `std::sync::Mutex` never un-poisons itself, so treating poison as fatal here would permanently fail Entra auth for every future connection in the process from one transient panic — a strictly worse outcome than the per-handle case the rule targets. `msqa.rs`'s pre-existing `CONTEXT_CACHE` (same category: process-wide auxiliary cache) uses the same `into_inner()` recovery pattern.

I did take the underlying concern seriously though: added a `warn!` log so a poison event is still visible in traces instead of silently swallowed (f3447a25), expanded the doc comment with this reasoning, and added a test (`cached_credential_recovers_from_a_poisoned_lock`) that deliberately poisons the lock and asserts the next call still succeeds rather than failing.

> [!NOTE]

CHANGELOG.md
```diff
@@ -91,3 +91,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   payload nor terminates a message — but was previously consumed as a
   zero-length packet. Empty end-of-message packets remain legal.
 
+- `mssql-odbc`: Entra ID credentials for service-principal and managed-identity
```
saurabh500: This landed under `### Fixed` (that heading starts at line 72), but it isn't a bug fix — it's a performance optimization plus a real behavior change: the credential, and the service-principal secret inside it, now lives for the life of the process instead of the connection.

Per Keep a Changelog, `Fixed` is for bug fixes and `Changed` is for changes in existing functionality. Please move this block up into the `### Changed` section.

While you're moving it, consider calling out the lifetime change explicitly, since that's the part an operator would want to know and it's currently only documented in the module comment:

> ...connection-pool warm-up. A cached credential retains the secret it was built from for the life of the process; a rotated secret creates a new cache entry rather than replacing the old one.

mssql-odbc/src/auth/entra.rs
```diff
@@ -155,6 +182,118 @@ impl EntraIdTokenFactory for EntraTokenFactory {
     }
 }
 
+/// Identifies a distinct Entra ID identity for the process-wide credential
+/// cache ([`CREDENTIAL_CACHE`]). Two factories that produce equal keys are
+/// guaranteed to represent the same tenant/client/secret (or the same managed
+/// identity) and may safely share one credential instance.
+#[derive(Clone, PartialEq, Eq, Hash, Debug)]
```
saurabh500: This derives `Debug` unconditionally on a type holding the tenant, client id, and secret digest, which sits awkwardly against the rest of the module's posture. `CredentialConfig` is *deliberately* not `Debug` (line 57), the module doc says never to log a cache key's inputs (line 28), and you added `label()` specifically so tracing never touches the identity.

Right now that's all convention: a future `debug!(?key, ...)` compiles and silently leaks. `Debug` is only needed by the `assert_eq!` / `assert_ne!` macros in the tests, so gating it on `cfg(test)` turns the rule into a compile error in production code and costs nothing:

```suggestion
#[derive(Clone, PartialEq, Eq, Hash)]
#[cfg_attr(test, derive(Debug))]
```

mssql-odbc/src/auth/entra.rs
```diff
@@ -61,18 +69,40 @@ pub(crate) enum CredentialConfig {
...
+    /// let two different secrets alias into the same cache entry, SHA-256
+    /// makes that practically impossible, so a secret rotation always misses
+    /// the old entry instead of silently continuing to authenticate with a
+    /// stale (possibly revoked) secret. Managed identity keys on client id
+    /// alone (empty for the system-assigned identity): IMDS is scoped to the
+    /// local machine, so no authority/tenant applies.
+    fn cache_key(&self, sts_url: &str) -> TdsResult<CredentialCacheKey> {
+        match &self.config {
+            CredentialConfig::ServicePrincipalSecret { client_id, secret } => {
+                let (authority_host, tenant_id) = split_sts_url(sts_url)?;
```
saurabh500, line 93: Nit, optional: on a cache miss the service-principal path parses `sts_url` twice — here, and again in `build_credential`. It's a cheap `url` parse and both calls are deterministic on the same input, so there's no divergence risk; purely duplication.

If you want it gone, have `cache_key` hand the parsed pair back to the caller and pass it into `build_credential` rather than re-deriving it:

```rust
fn cache_key(&self, sts_url: &str) -> TdsResult<(CredentialCacheKey, Option<(String, String)>)>
```

Not worth restructuring for on its own — take it or leave it.

mssql-odbc/src/auth/entra.rs
```diff
@@ -155,6 +182,118 @@ impl EntraIdTokenFactory for EntraTokenFactory {
...
+/// discriminator — never persisted, logged, or compared against
+/// attacker-controlled input. SHA-256 (rather than a `DefaultHasher`/SipHash
+/// digest) is what makes two different secrets practically un-collidable, so
+/// a rotated secret always misses the old cache entry instead of an attacker
+/// — or bad luck — being able to alias a new secret onto a stale credential.
+fn digest(secret: &str) -> [u8; 32] {
+    Sha256::digest(secret.as_bytes()).into()
+}
+
+/// Entra ID credentials, shared process-wide so a burst of new connections for
```
saurabh500, line 231: Nit: the doc blocks here and on `cached_credential` are essay-length — roughly 45 lines of comment for about 25 lines of code between them. They do document genuine *why* (msodbcsql parity, the credential-vs-token distinction, the poison rationale), which the repo guidance allows, but some of it is restated a third time in the test comments and in the PR description.

Suggest tightening roughly 30%. Concretely: the `Parse.cpp:3655` / `MSQAAuthContextCache` parity paragraph is the kind of detail that belongs in the PR description and AB#46409 rather than in the source, and the poison paragraph on `cached_credential` largely duplicates the comment already sitting in `cached_credential_recovers_from_a_poisoned_lock`. Keep the credential-instance-vs-token point and the unbounded-growth caveat — those are the two a future reader actually needs at the call site.

saurabh500 (approved):
## Review — odbc: cache Entra ID credentials process-wide

**Approving.** Replaces the per-connection `OnceCell<credential>` with a process-wide `HashMap` cache keyed by identity, so a connection storm for one identity builds the credential once instead of once per connection. The design is sound, and I verified the premise end-to-end rather than taking it on faith. No blocking findings.

### Verified

- **The optimization actually works.** The whole PR rests on `azure_identity` credentials caching tokens internally — if they didn't, sharing the instance would save nothing. Traced it: `ClientSecretCredential` holds a `TokenCache` (`client_secret_credential.rs:35,145`), and all three managed-identity sources (App Service, VM, IMDS) funnel into `ImdsManagedIdentityCredential`, which holds one too (`imds_managed_identity_credential.rs:59,187`). `TokenCache` is keyed per-scope with a 300s refresh margin (`cache.rs:64-66`). So caching the credential *instance* does share tokens, and expiry/refresh stays inside the SDK.
- **Thread safety.** `std::sync::Mutex` in a `OnceLock`; the guard lives entirely inside the sync `cached_credential` and is dropped before `create_token` awaits `get_token`. `build()` under the lock is pure config assembly — no I/O, no await. Concurrent populators serialize to one build per key.
- **Cache key is identity-complete.** SP: `authority_host` + `tenant_id` + `client_id` + SHA-256(secret); MI: `client_id` (empty == system-assigned). Scope is correctly excluded, since per-scope caching lives inside the credential. No cross-identity or cross-tenant aliasing.
- **Secrets.** The key stores only a digest, `CredentialConfig` is deliberately non-`Debug`, and only `key.label()` (the auth-method kind) is logged.
- **Both prior-round findings are genuinely fixed.** SipHash to SHA-256 at `entra.rs:221-222`, pinned by `digest_is_deterministic_sha256_not_a_weak_hash`; poison recovery justified with a `warn!` and a test.
- **`sha2` adds no duplicate dependency.** `cargo tree -p mssql-odbc -i sha2` resolves to a single `sha2 v0.11.0` and the generated lock has exactly one entry — nothing pulls 0.10.x transitively.
- **Local validation.** `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` is clean and all 24 entra tests pass.

### Blocking

None.

### Suggestion

- **CHANGELOG entry is in the wrong section** (`CHANGELOG.md:94`). It sits under `### Fixed`, but this is a performance optimization plus a behavior change — the credential, and the secret inside it, now lives for the life of the process rather than the connection. Belongs under `### Changed`. Commented inline.
- **`#[derive(Debug)]` on `CredentialCacheKey`** (`entra.rs:189`). The module goes to real lengths to keep identity out of logs: `CredentialConfig` is deliberately non-`Debug`, `label()` exists precisely so tracing never touches the identity, and the module doc says never to log a cache key's inputs. But the key itself derives `Debug` unconditionally, so that rule is enforced by convention rather than by the compiler — a future `debug!(?key, ...)` would compile. `Debug` is only needed by the test assert macros, so `#[cfg_attr(test, derive(Debug))]` makes it a compile error in production code. Commented inline.
- **`into_inner()` poison recovery vs. the instructions text** — maintainer call, no code change needed. `entra.rs:279-280` recovers a poisoned lock, but `mssql-odbc.instructions.md:47` says unqualifiedly *"Do not recover via `e.into_inner()`."* In practice that rule is scoped to per-handle FFI mutexes: it is phrased around `handle.inner.lock()` returning `SQL_ERROR`, and every lock in `api/*.rs` follows the `let Ok(...) else { return SQL_ERROR }` form. The identical pattern already exists for the analogous process-wide `CONTEXT_CACHE` at `msqa.rs:311-313`. It is safe here — the only mutation is the `insert` after `build()?` returns `Ok`, so a panic under the lock can leave a missing key (a cache miss), never a torn entry. A precedented deviation rather than a bug, but worth the maintainers explicitly carving process-wide `static` caches out of the instruction wording.

### Nit

- **Redundant `split_sts_url` parse on a miss** (`entra.rs:93`) — the SP path parses once in `cache_key` and again in `build_credential`. Cheap and deterministic on the same input, so there is no divergence risk; only worth changing if the duplication bothers you. Commented inline.
- **Doc-comment length** (`entra.rs:225`). The blocks on `CREDENTIAL_CACHE` and `cached_credential` are essay-length. They document real *why*, which the guidance permits, but some of it is restated in the test comments and again in the PR description. Could tighten roughly 30% without losing substance. Commented inline.

CHANGELOG.md
Vahid-b: Good catch, and thanks for the exact wording. Moved it to `### Changed` in d9f60ff5 and added the callout you suggested (a rotated secret creates a new cache entry rather than replacing the old one) — that's exactly the operator-facing detail that belonged there instead of only in the module doc.

> [!NOTE]

mssql-odbc/src/auth/entra.rs
Vahid-b: Applied exactly as suggested in d9f60ff5 — `#[cfg_attr(test, derive(Debug))]`, so a future `debug!(?key, ...)` is now a compile error in production code rather than a convention someone has to remember. Good call.

> [!NOTE]

mssql-odbc/src/auth/entra.rs
Vahid-b: Leaving this one as-is — agree it's real duplication but I don't think the fix is worth it here. `build_credential` being independently callable from just `sts_url` (no dependency on anything `cache_key` computed) means the two functions stay independently testable and reasoned about; threading the parsed `(authority, tenant)` tuple through would couple them for the sake of skipping one cheap, deterministic `Url::parse` that only runs on a cache miss — which by design is rare (once per distinct identity for the life of the process, not per connection). You'd flagged it as optional/take-it-or-leave-it, and I'm taking "leave it."

> [!NOTE]

mssql-odbc/src/auth/entra.rs
Vahid-b: Trimmed both in d9f60ff5 — roughly a third shorter. The msodbcsql `Parse.cpp`/`MSQAAuthContextCache` citation now just points to the PR description/AB#46409 instead of repeating the full detail in source, and the poison paragraph on `cached_credential` is condensed since the complete reasoning already lives in `cached_credential_recovers_from_a_poisoned_lock`'s test comment. Kept the credential-vs-token point and the unbounded-growth caveat as you suggested.

> [!NOTE]

saurabh500 (approved):
## Re-review — `d9f60ff5`

**Approving.** Both actionable items from the previous round are addressed exactly as suggested, and the one new commit touches only the two files under discussion. No new findings.

### Addressed

- **CHANGELOG section** — moved out of `### Fixed` into `### Changed`, and the credential/secret lifetime note is now spelled out for operators rather than living only in the module comment. 
- **`#[derive(Debug)]` on `CredentialCacheKey`** — now `#[derive(Clone, PartialEq, Eq, Hash)]` plus `#[cfg_attr(test, derive(Debug))]`, with a doc line recording why. The "never log a cache key's contents" rule is now enforced by the compiler instead of by convention. Confirmed there is no non-test `{:?}` use of the key, and the type is private so integration tests can't reach it either.
- **Doc-comment length (nit)** — trimmed roughly 25 lines. The `Parse.cpp` / `MSQAAuthContextCache` parity paragraph now lives in the PR description and AB#46409, and the poison rationale is condensed. Good call keeping the credential-instance-vs-token point and the unbounded-growth caveat at the call site.
- **Redundant `split_sts_url` parse (nit)** — not taken, which is fine; it was explicitly optional and there's no divergence risk.

### Verified on this commit

- `cargo fmt -- --check` clean.
- `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` clean.
- `cargo nextest run -p mssql-odbc --all-features` — 607/607 passed.
- Cache-key derivation, locking, and credential-sharing logic are untouched by this commit, so the correctness verification from the previous round still stands: the `azure_identity` internal `TokenCache` premise, the guard never crossing an await, and the identity-complete key.

### Nit (non-blocking, my earlier suggestion overshot slightly)

The doc on `cached_credential` now says *"see `cached_credential_recovers_from_a_poisoned_lock` for the full case"*, but that test's comment opens with *"The reasoning on `cached_credential`: ..."* — the two now point at each other, and the safety argument that was removed lives in neither. The sentence worth restoring in one of the two places is the one establishing that recovery is *safe* rather than merely *convenient*:

> the lock is never held across `build()`'s fallible work in a way that leaves a partial insert, so the worst case is a missing entry — exactly a cache miss.

Not worth another round on its own; fold it in if you touch the file again.

Issue #310, opened 2026-08-15
SqlString::to_utf8_string panics on malformed UTF-8 from the server


saurabh500:
### Describe the bug

`SqlString::to_utf8_string` panics on malformed UTF-8 instead of returning an error or replacing invalid sequences.

`mssql-tds/src/datatypes/sql_string.rs:60`:

```rust
EncodingType::Utf8 => String::from_utf8(self.bytes.clone()).unwrap(),
```

The bytes come straight off the wire, so this makes malformed server data a process-level panic in a library.

Three things make it worse than a normal `unwrap`:

1. **It is inconsistent with its own siblings.** The other two arms of the same `match` use `encoding_rs`, which is lossy and cannot panic — `EncodingType::Utf16` decodes via `UTF_16LE.decode(..)`, and `EncodingType::LcidBased` even has a `warn!` + Windows-1252 fallback for unsupported LCIDs. Only the UTF-8 arm aborts.
2. **It is reachable from `Display` and `Debug`** (`sql_string.rs:143` and `:153`). Panicking inside `Display` means logging or formatting a row can bring the process down, including on error paths that are trying to report an unrelated failure.
3. **It is on a broad public surface.** `to_utf8_string` is `pub` and has ~14 call sites across every binding — `mssql-js` (`binary_row_writer.rs`), `mssql-py-core` (`convert.rs`, `cursor.rs`), and `mssql-odbc` (`cell.rs`, `fetch_convert.rs`, `tds_value_serializer.rs`, `arrow_bulkcopy.rs`, `describe_parameter_encryption.rs`).

There is also a latent hint that this arm is not well understood, directly above the line:

```rust
// TODO: Investigation needed. When creating a Utf8 strings from the vector, the string is weirdly encoded.
// UTF16 decode works better.
```

### Steps to reproduce

The `EncodingType::Utf8` arm is selected for UTF-8 collations (`collation.utf8()`, `sql_string.rs:168`), i.e. SQL Server 2019+ collations such as `Latin1_General_100_CI_AS_SC_UTF8`.

Direct unit-level reproduction, no server required:

```rust
let s = SqlString::new(vec![0xE4, 0xBD], EncodingType::Utf8); // truncated 3-byte sequence
let _ = s.to_utf8_string(); // panics: called `Result::unwrap()` on an `Err` value: FromUtf8Error
```

[...]

PR #343, opened 2026-08-20
perf(tds): borrow short string and binary values from the read buffer

NivasSA:
PR #299 landed synchronous buffered reads for the ten fixed width scalars and wired them at 111 sites. Try the buffer first, fall back to the awaiting read on a miss. This extends the same pattern from fixed width scalars to slices.

A varchar(n), char(n), binary(n) or varbinary(n) value arrives with a two byte length header, and by the time the decoder needs it the bytes are already in the read buffer. Today it still allocates a buffer, zeroes it, overwrites it from that same read buffer, then hands over ownership. On a 1.5M row table with nine varchar(6) columns that is 13.5 million allocations for one scan.

`try_read_slice` returns a borrowed span when the value is resident in the current packet, and the `RowWriter` receives it as a `Cow`. Values that straddle a packet boundary return `None` and take the existing owned path unchanged, so the number of await points does not move. That is what separates this from the destination based sink in #311, where hooking the same site turned one await into two and cost `DefaultRowWriter` 5.4%.

Six tests decode identical wire bytes twice, once through a reader that offers slices and once through one that always declines, then assert the results agree. One feeds a truncated value, which proves the borrowed arm declines rather than reading past the end.

Measured in a PostgreSQL foreign data wrapper across eight fixtures, six runs each: -14.9% on 39 int plus 9 varchar(6), -15.5% on ten short string columns, -25.5% on a mixed varchar(n) and char(n) table. Fixtures with no strings and no binary move within build to build variance.

## Description

Extends the #299 buffered-read pattern from fixed-width scalars to variable-length slices, so short `varchar(n)` / `char(n)` / `binary(n)` / `varbinary(n)` values already resident in the read buffer are handed to the `RowWriter` as a borrowed span instead of a freshly allocated `Vec`. Values straddling a packet boundary fall back to the existing owned path unchanged, so the number of await points on the row path does not move.

## Related Issues

Refs #253 — `RowWriter` sink API so consumers receive string/binary payloads without `mssql-tds` allocating (row decode perf, axis: value handoff).

Related to #247 — row decode performance tracking issue.

Context for the `to_utf8_string` note below: #310.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

---

### Merged latest `main`

`main` had moved 13 commits since this branch was cut. The merge was textually clean but hid two semantic breaks, both fixed here:

- `consume_bytes` became fallible (`TdsResult<()>`) in #360. `try_read_slice` was calling it bare, which trips `unused_must_use` under `-D warnings`. Now handled the same way its sibling `try_read_array` already does on `main`.
- `test_slice_probe_falls_back_across_a_packet_boundary` was building two complete EOM-terminated *messages* rather than one message split across two packets, which `main`'s new end-of-message tracking correctly rejects. This test is the only proof that the borrowed path declines across a real packet boundary, and it was passing for the wrong reason. Now built with `TestPacketBuilder::continuation()`.

### ⚠️ Breaking change: `RowWriter` string and binary writes now take `Cow`

The branch originally added `write_string_ref` / `write_bytes_ref` alongside the existing owned `write_string` / `write_bytes`, with defaults that copied. That shape is non-breaking by construction, but it means a writer has to override *both* halves of a pair to get the zero-copy path, and nothing enforces that.

It had already gone wrong here: `DynamicRowWriter` forwarded `write_string` but not `write_string_ref`. The default then copied and called the **inner** writer's `write_string`, so every writer behind that wrapper silently lost the borrow regardless of what it overrode. No test caught it, and nothing would have short of benchmarking through the wrapper.

The four methods are now two:

```rust
fn write_string(&mut self, col: usize, bytes: Cow<'_, [u8]>, encoding_type: EncodingType);
fn write_bytes(&mut self, col: usize, bytes: Cow<'_, [u8]>);
```

The transport decides borrow vs owned; the writer handles both in one body. You cannot forget to forward a method that does not exist. `write_string_ref` / `write_bytes_ref` are gone.

`Cow` is deliberately *not* bundled inside `SqlString`. Making it `SqlString<'a>` would be viral — roughly 230 mentions across ~34 files in four crates — because Rust has no default lifetime parameters.

**This breaks every out-of-tree `RowWriter` implementation.** All five in-tree implementers are updated: `DefaultRowWriter`, `DiscardRowWriter`, `DynamicRowWriter`, `BinaryRowWriter` (`mssql-js`), and `PyRowWriter` (`mssql-py-core`).

### Made the borrowed path usable by an external writer

The point of handing out borrowed wire bytes is that a consumer can transcode them straight into its own buffer. That was not actually reachable: the writer got an `EncodingType`, but the encoding dispatch was inlined inside `SqlString::to_utf8_string`, which needs an owned `SqlString`. An external writer could only hand the borrow back for the default copy.

- **Added `EncodingType::encoding() -> Option<&'static encoding_rs::Encoding>`**, and re-exported `encoding_rs` from the crate root so a consumer can name the returned type without pinning a matching version itself. `lcid_to_encoding` already leaked `&'static Encoding` publicly, so this is not new exposure.
- **Added `SqlString::decode(&[u8], EncodingType) -> String`** for the common case where the consumer just wants the text. `to_utf8_string` now delegates to it, so there is one decode path rather than two that can drift. `BinaryRowWriter` uses it and no longer copies on the borrowed path.
- **Added `SqlString::into_parts()`** so `write_column_value` can split an owned value into bytes plus encoding without cloning.
- **`to_utf8_string`'s behaviour is deliberately unchanged.** Its `Utf8` arm uses `String::from_utf8(..).unwrap()`, so it panics on malformed input (#310), whereas decoding through `encoding_rs` substitutes U+FFFD. Routing it through `EncodingType::encoding()` would have been a silent behaviour change well outside this PR's scope. The shared LCID lookup and Windows-1252 fallback are factored into one helper, so there is no duplicated warn path.

### Other review fixes

- **`EncodingType` is now `Copy`.** It is a 16-byte payload-free enum over a `Copy` `SqlCollation`, so the `&EncodingType` in the original signature bought nothing and forced a clone in the default impl. Clippy now flags such clones on its own; one in `decoder.rs` was removed.
- **`DiscardRowWriter` takes the `Cow` unexamined**, so a borrow is never promoted to an owned buffer purely to drop it. It is the writer the -17.4% number is measured against.

### Benchmark numbers need re-running

The `sink_bench.rs bench_narrow_control` figures previously quoted here (-12.2% / -10.0% / -10.9% / -17.4%) were measured on the original four-method shape. `Cow<'_, [u8]>` is 32 bytes on the hot path where `&[u8]` was 16. On a codebase that benchmarks at 5% resolution that is worth re-measuring before merge rather than carrying stale numbers. The FDW end-to-end numbers above are unaffected in kind but were also measured pre-refactor.

### Tests added

- `EncodingType::encoding()` across all four variants, including the unmapped-LCID fallback, and agreement between the accessor and `to_utf8_string` on the same LCID bytes.
- `SqlString::decode` agrees with `to_utf8_string` across UTF-16, UTF-8 and LCID-based bytes, so the borrowed and owned decode paths cannot drift.
- `DynamicRowWriter` forwards borrows to the inner writer **still borrowed** — the spy asserts `Cow::Borrowed` arrives as such, which is the regression that shipped unnoticed before.
- Borrowed-vs-owned equivalence for `DefaultRowWriter` on both strings and binary, now with a `BorrowSpy` writer that records **whether each value actually arrived borrowed**. The earlier version compared only decoded values through two `DefaultRowWriter`s, and `DefaultRowWriter::write_string` calls `into_owned()` — so the `Cow` variant was discarded before anything could observe it. @David-Engel mutation-tested that and found rewriting all three borrowed call sites to `Cow::Owned(bytes.to_vec())` left the suite green. With the spy, that same mutation kills 4 of the 5 tests; the survivor is the NULL case, which writes no string or binary value at all and so has no borrow to assert on.

### Fuzz coverage of the borrowed path

`FuzzReader`, `FuzzPacketReader` and `MockTransport` each override every other `try_read_*` method but inherited the `None` default for `try_read_slice`, so `fuzz_tds_client` drove `execute()` and row decode entirely through the owned arm — the new length arithmetic, over sizes taken straight off the wire, had no adversarial coverage at all. Implemented on `FuzzReader` and delegated from the other two, so the existing corpus and the three-hourly fuzz pipeline reach the new call sites without a new target. `EmptyReader` keeps the inherited `None`, which is correct for a no-data stub.

This matters beyond coverage: the PR adds no `unsafe` (the `set_len` calls in `decoder.rs` are pre-existing, on the owned arm), so a bounds mistake surfaces as a panic on network-controlled input rather than as memory corruption — still a defect this repo bans, and randomized lengths are what a unit test cannot supply.

Because `fuzz_support` is `#[cfg(fuzzing)]`, neither `cargo bclippy` nor `cargo btest` compiles it. Validated separately with `RUSTFLAGS="--cfg fuzzing" cargo clippy -p mssql-tds --all-features --all-targets` (clean) and `RUSTFLAGS="--cfg fuzzing" cargo check --all-targets` in `mssql-tds/fuzz` (all nine targets build).
`mssql-tds` lib suite: 1921/1928 pass. The 7 failures are `certificate_validator` and `win_tls::validate`, caused by test certificate fixtures that are not tracked in git (`valid_cert.der` / `valid_cert.pem` must be generated via `generate_certs.sh`). They fail identically on `main`.

`mssql-py-core` is outside the workspace; `cargo fmt` and `cargo clippy --all-targets -- -D warnings` were run against it separately and pass.


saurabh500 (commented):
## Review — perf(tds): borrow short string and binary values

**Summary:** Extends the #299 buffered-read pattern from fixed-width scalars to variable-length slices. `try_read_slice` returns a borrowed span when the whole value is resident in the current packet, so no new await point appears on the hit path; a value straddling a packet boundary returns `None` and takes the unchanged owned path. That's what separates this from the destination-based sink in #311, where hooking the same site turned one await into two. The mechanism is sound and the equivalence tests are thorough.

**Heads up — I merged latest `main` into this branch and pushed** (now at `a668101d`). `main` had moved 13 commits, and the merge was *not* cosmetic. Two semantic breaks were hidden behind a clean textual merge:

1. `consume_bytes` became fallible (`TdsResult<()>`) via #360. `try_read_slice` was calling it bare, which trips clippy's `unused_must_use` under `-D warnings`. Fixed by copying the pattern its sibling `try_read_array` already uses on `main`.
2. `test_slice_probe_falls_back_across_a_packet_boundary` was building two complete EOM-terminated *messages*, not one message split across two packets. `main`'s new end-of-message tracking correctly rejects that. **This matters** — that test is this PR's only proof that the borrowed path declines across a real packet boundary, and it was passing for the wrong reason. Fixed with `TestPacketBuilder::continuation()`.

Please pull before pushing again. `cargo bfmt`, `cargo bclippy`, and `cargo nextest run -p mssql-tds --lib` are green on the merge result (the 7 `certificate_validator` / `win_tls::validate` failures are missing cert fixtures that aren't tracked in git — they fail on any branch).

### Blocking

- **No linked issue or work item.** The `## Related Issues` section still contains only the template comment. The repo requires a GitHub issue or ADO work item and the PR check fails without one. Worth noting the PR is already out of draft — repo convention is draft-first until validation and self-review are done.
- **`EncodingType` is opaque to anyone outside this crate.** Inline comment on the `write_string_ref` default below. This is the one thing that stops an external consumer from actually using the borrowed path.

### Suggestion

- **Collapse the four methods into two, using `Cow`.** Inline comment below. Short version: a writer that overrides `write_string` but not `write_string_ref` silently keeps the old behaviour on the fast path, and the two can drift apart without any compiler complaint.

### Nit

- **`encoding_type: &EncodingType` could be by value.** I measured it: `EncodingType` is 16 bytes and `needs_drop()` is `false` (`SqlCollation` is 12 bytes and `Copy`). So the `.clone()` on the next line is a register copy, not an allocation — it is not a real cost, and it is not gratuitous either, since `SqlString::new` needs an owned value and the default impl has no other option. Taking it by value would just remove the noise. Flagging this mainly to close it out, since `.clone()` next to a `.to_vec()` reads alarming and it isn't.
- **Checklist boxes are unchecked** despite the description stating `bfmt`/`bclippy`/tests pass and new tests being added. Tick them to match reality.

No correctness, security, or layering issues found in the decoder/transport/reader changes themselves. Not merging — that's yours.

mssql-tds/src/datatypes/row_writer.rs
```diff
@@ -75,6 +75,20 @@ pub trait RowWriter {
     /// Signals the end of the current row.
     fn end_row(&mut self);
 
+    /// Writes packet-backed encoded string bytes before the packet buffer is
+    /// reused.
+    ///
+    /// `bytes` is the value in its wire encoding. The default copies, so a
+    /// writer that does not override this behaves exactly as before.
+    fn write_string_ref(&mut self, col: usize, bytes: &[u8], encoding_type: &EncodingType) {
+        self.write_string(col, SqlString::new(bytes.to_vec(), encoding_type.clone()));
```
saurabh500: **Blocking:** an external writer receives `&EncodingType` here but has no public way to turn it into an encoder, so it cannot do anything with `bytes` except hand them back for the default copy — which defeats the point of the borrowed path.

The dispatch lives inlined inside `SqlString::to_utf8_string` (`sql_string.rs:56-101`) and takes `&self`, so it is only reachable by *owning* a `SqlString`. `lcid_to_encoding(lcid: u32) -> Result<&'static Encoding, _>` is public, but the logic wrapped around it is not: the `Utf8` / `Utf16` arms, the Windows-1252 fallback with its warn logging, and the `DelayedSet` case.

There is an in-tree consumer that already demonstrates the gap — `mssql-js/src/binary_row_writer.rs:206-211` calls `val.to_utf8_string()` then `intern_string(utf8)`. It cannot do the equivalent from a borrowed slice, so it cannot adopt this path as written.

Suggest adding an accessor alongside the existing code:

```rust
impl EncodingType {
    /// The `encoding_rs` encoding for this type, or `None` when it is not yet
    /// known (`DelayedSet`).
    pub fn encoding(&self) -> Option<&'static encoding_rs::Encoding> { ... }
}
```

Two caveats:

1. **Please add it alongside `to_utf8_string`, don't rewrite `to_utf8_string` on top of it.** The `Utf8` arm currently uses `String::from_utf8(..).unwrap()`, not `encoding_rs::UTF_8`. Routing it through `encoding_rs` would silently convert a panic on invalid UTF-8 into U+FFFD replacement — a real behaviour change well outside this PR's scope.
2. `encoding_rs` is not re-exported from `mssql-tds` (not in `lib.rs` or `datatypes.rs`), so a consumer needs its own version-matched dependency to name the return type. `lcid_to_encoding` already leaks `&'static Encoding` publicly, so this is not new exposure, but a re-export would make the surface usable without version-pinning gymnastics.

mssql-tds/src/datatypes/row_writer.rs
```diff
@@ -75,6 +75,20 @@ pub trait RowWriter {
...
+    }
+
+    /// Writes packet-backed binary bytes before the packet buffer is reused.
+    fn write_bytes_ref(&mut self, col: usize, bytes: &[u8]) {
+        self.write_bytes(col, bytes.to_vec());
+    }
```
saurabh500: **Suggestion:** consider collapsing `write_string`/`write_string_ref` and `write_bytes`/`write_bytes_ref` into one required method each, taking `Cow`:

```rust
fn write_string(&mut self, col: usize, bytes: Cow<'_, [u8]>, encoding: EncodingType);
fn write_bytes(&mut self, col: usize, bytes: Cow<'_, [u8]>);
```

The reason is not performance — all four combinations cost the same, since `Cow::Owned` -> owned is a move and `Cow::Borrowed` -> owned is the same `to_vec()` you already do. It is that **a defaulted method is silently skippable**. A writer that overrides `write_string` but not `write_string_ref` compiles clean and behaves correctly today, then quietly takes the old path for every short value. Any behaviour the override adds — blank-stripping for `char(n)`, interning, a custom error path — applies on the slow path and not the fast one, and the two drift apart with no compiler complaint and no test failure unless someone happens to test a value that straddles a packet boundary.

A concrete instance is already in the tree: `DiscardRowWriter` (lines 328-354) overrides neither `_ref` method, so on every borrow hit it does `bytes.to_vec()`, builds a `SqlString`, and drops it. It allocates purely in order to discard — and it is the writer the -17.4% benchmark number is measured against. With `Cow` it would just not allocate, no override needed. (Commenting here since those lines aren't in the diff.)

Honest cost: this is a breaking change to a public trait and touches 5 implementors. If that is too much churn for this PR, the smaller version is to keep the split but make the two `_ref` methods **required** rather than defaulted, so an implementor is forced to make a deliberate choice. Either way the current shape is the one where a consumer can be wrong without noticing.

mssql-tds/src/datatypes/row_writer.rs
saurabh500: Fixed in 849a9bc1.

Added `EncodingType::encoding() -> Option<&'static encoding_rs::Encoding>` and re-exported `encoding_rs` from the crate root, so a consumer can name the returned type without pinning a matching version of its own.

Took the caveat: `to_utf8_string` is left alone. Its `Utf8` arm still uses `String::from_utf8(..).unwrap()`, so it keeps panicking on malformed input (#310) rather than silently switching to U+FFFD replacement. To avoid two copies of the LCID lookup and its Windows-1252 warn path, that part is factored into a single `lcid_encoding_or_fallback` helper used by both.

Also took the by-value nit while here — `write_string_ref` now takes `EncodingType` by value, so the default impl no longer clones.

Tests cover all four variants, the unmapped-LCID fallback, and that decoding through the accessor produces the same text `to_utf8_string` does for the same LCID bytes.

> [!NOTE]

mssql-tds/src/datatypes/row_writer.rs
saurabh500: **Withdrawing this — the premise was wrong, sorry.**

I claimed a writer that overrides `write_string` but not `write_string_ref` would quietly take the old path and lose whatever its override adds. That is not what happens. The default is:

```rust
fn write_string_ref(&mut self, col: usize, bytes: &[u8], encoding_type: EncodingType) {
    self.write_string(col, SqlString::new(bytes.to_vec(), encoding_type));
}
```

`self.write_string(..)` dispatches to the concrete impl, so the override **does** run on the borrowed path. Skipping the `_ref` methods costs an allocation, not a behaviour difference. No drift, so no reason to break a public trait over it. The `Cow` change is not worth doing.

Two things from the comment were real and are fixed in 849a9bc1:

- `DiscardRowWriter` was building a `SqlString` purely to drop it on every borrow hit. Now overrides both `_ref` methods.
- Chasing this turned up an actual instance of the "never reached" problem I had wrongly attributed to the defaults: `DynamicRowWriter` did not forward either `_ref` method, so the defaults copied and called `write_string`/`write_bytes` on the *inner* writer. Any writer behind that wrapper silently lost the zero-copy path regardless of what it overrode. Now forwards both, with a test.

Added `defaulted_ref_methods_route_through_the_writers_override` to pin the delegation invariant, so if a future change does break it the test says so instead of a reviewer guessing.

> [!NOTE]

mssql-tds/src/io/packet_reader.rs
```diff
@@ -72,6 +72,16 @@ macro_rules! define_tds_packet_reader {
                 None
             }
 
+            /// Borrows and consumes `length` buffered bytes, or returns `None`
```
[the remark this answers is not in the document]
saurabh500: Addressed. The description now carries `Refs #253` (the `RowWriter` sink API issue this implements) and `Related to #247` (the row-decode perf tracking issue), matching the convention from #299 and #311.

> [!NOTE]

mssql-tds/src/datatypes/row_writer.rs
saurabh500: **Correction — the withdrawal above was itself wrong, and this is now implemented.**

My original *reason* was wrong: `write_string_ref`'s default calls `self.write_string(..)`, which does dispatch to the override, so there was no behaviour drift. But withdrawing the whole suggestion was an overcorrection — the conclusion still holds, for a different reason.

The real problem with the paired-method shape is that **it already produced a silent bug in this PR.** `DynamicRowWriter` forwarded `write_string` but not `write_string_ref`. The default then copied and called the *inner* writer's `write_string`, so every writer behind that wrapper quietly lost the zero-copy path no matter what it overrode. No test caught it, and nothing would have short of benchmarking through the wrapper. That's a whole class of bug that a single method makes unwritable — you can't forget to forward a method that doesn't exist.

Second reason: the FDW consumer this PR exists for has to write the same body twice, once per method, differing only in where the bytes came from.

Implemented in `4a38961c`:

```rust
fn write_string(&mut self, col: usize, bytes: Cow<'_, [u8]>, encoding_type: EncodingType);
fn write_bytes(&mut self, col: usize, bytes: Cow<'_, [u8]>);
```

`write_string_ref` / `write_bytes_ref` are gone. This is a breaking change to `RowWriter`; @saurabh500 confirmed that's acceptable here.

Two things I got wrong earlier and want on the record:
- I judged this "not worth doing" partly because the defaulted-method shape is non-breaking by construction. That's a real property and probably why it was written that way — it just isn't worth a permanently trip-prone trait.
- I called `encoding_type.clone()` a Nit on the grounds that it's a register copy. Correct, but the right fix was to make `EncodingType: Copy` (it's a payload-free enum over a `Copy` `SqlCollation`). Done — and clippy now flags the clone in `decoder.rs` on its own.

Also added `SqlString::decode(&[u8], EncodingType)` so a writer can transcode borrowed bytes without building an owned `SqlString` first; `to_utf8_string` delegates to it, and `BinaryRowWriter` no longer copies on the borrowed path.

⚠️ One consequence for @NivasSA: `Cow<'_, [u8]>` is 32 bytes on the hot path where `&[u8]` was 16. **The `bench_narrow_control` numbers in the PR description were measured on the four-method shape and need re-running.**

> [!NOTE]

David-Engel (commented):
Reviewed against the merge base (`a621332` — `main` is fully merged in, 0 commits behind). Verified locally: `cargo clippy -p mssql-tds --all-features --all-targets -- -D warnings` is clean, and `cargo nextest run -p mssql-tds --lib` gives 1833/1837 passing with only the 4 known `certificate_validator` failures (missing untracked cert fixtures — they fail on `main` too).

## Summary

Extends the #299 buffered-read pattern to variable-length values: `TdsReadBuffer::try_read_slice` hands back a borrow into `working_buffer` when the whole value is resident in the current packet, and three decoder sites (`BigBinary`, non-PLP `BigVarBinary`, non-PLP string) route it to two new defaulted `RowWriter` methods. I verified the core safety argument: `buffer_position..buffer_length` never spans a packet (out-of-packet data lives in the separate `pending_bytes` region), the returned borrow is bounded by `&mut self` so no read can overwrite it while live, and `saturating_sub` in `get_remaining_byte_count` plus the up-front capacity check make `start + length` unable to overflow the buffer. The miss path leaves `buffer_position` untouched, so the owned fallback is byte-for-byte the old behaviour. Design and tests look sound; nothing blocking.

## Findings

**Blocking** — none. Two suggestions and two nits inline / below.

### Suggestion — the new branch is not fuzzed at all

`mssql-tds/src/fuzz_support.rs` — `FuzzReader` (line 77) implements every one of the ten `try_read_*` scalar probes from #299, and `FuzzPacketReader` (line 514) / `MockTransport` (line 936) forward all of them. None implements `try_read_slice`, so it falls to the trait default that returns `None`. `fuzz_token_stream` drives the full row/token path over `FuzzReader`, which means the borrowed arm added here never executes under fuzzing — the one harness aimed at adversarial length headers only ever sees the owned fallback.

`FuzzReader` is a flat `Vec<u8>` plus a cursor, so this is the same three-liner already written for `ByteReader` in the decoder tests:

```rust
fn try_read_slice(&mut self, length: usize) -> Option<&[u8]> {
    let end = self.position.checked_add(length)?;
    let bytes = self.data.get(self.position..end)?;
    self.position = end;
    Some(bytes)
}
```

plus the matching forwards in `FuzzPacketReader` and `MockTransport`. (Commenting here rather than inline since those files aren't in the diff.)

### Nit — the description's justification for the new public surface is inaccurate

The PR says "`lcid_to_encoding` already leaked `&'static Encoding` publicly, so this is not new exposure." It didn't: `mssql-tds/src/datatypes.rs:17` declares `pub(crate) mod lcid_encoding;`, so `lcid_to_encoding` is not reachable from outside the crate. `EncodingType::encoding()` plus `pub use encoding_rs;` genuinely *is* new exposure, and it puts `encoding_rs` 0.8 into `mssql-tds`'s semver surface — a major bump there becomes a breaking change here. The re-export is the right mitigation and I'd keep it (`uuid::Uuid` is already leaked through `RowWriter::write_uuid` with no re-export, which is the worse shape), but the rationale should be corrected so a future reader doesn't conclude the door was already open.

## Notes, not findings

- The `DynamicRowWriter` forwarding fix is real and would have been easy to miss — without it every wrapped writer silently lost the borrow. Good catch, and the `RefSpy` test pins it properly.
- Withdrawing the `Cow` collapse was correct: the default `write_string_ref` dispatches through `self.write_string`, so an override does apply on both arms. `defaulted_ref_methods_route_through_the_writers_override` pins exactly that invariant.
- Follow-up, not this PR: the two in-tree consumers, `PyRowWriter` and `BinaryRowWriter`, don't override the `_ref` pair, so they only collect the zero-fill saving. `mssql-js/src/binary_row_writer.rs:206` does `to_utf8_string()` then `intern_string(utf8)`, which is a direct candidate for transcoding straight out of the borrow now that `EncodingType::encoding()` exists.
- Scoping is right: PLP and the legacy TEXT/NTEXT/IMAGE paths are deliberately untouched, and PLP already has its own sink API from #311.

mssql-tds/src/datatypes/row_writer.rs
```diff
@@ -75,6 +75,25 @@ pub trait RowWriter {
...
 
+    /// Writes packet-backed encoded string bytes before the packet buffer is
+    /// reused.
+    ///
+    /// `bytes` is the value in its wire encoding; use
+    /// [`EncodingType::encoding`] to transcode it in place. The default copies
+    /// and forwards to [`RowWriter::write_string`], so a writer that does not
+    /// override this keeps whatever behaviour its `write_string` has — it only
+    /// gives up the saved allocation.
+    fn write_string_ref(&mut self, col: usize, bytes: &[u8], encoding_type: EncodingType) {
```
David-Engel: **Suggestion:** the doc covers the "didn't override it" case well. The case it doesn't cover is the one this PR is actually building for: a writer that *does* override `write_string_ref` and decodes via `EncodingType::encoding()` gets U+FFFD substitution, while its `write_string` — which still runs for any value straddling a packet boundary — goes through `SqlString::to_utf8_string()`, which `unwrap()`s under `EncodingType::Utf8` and panics on malformed input (#310).

Which of the two receives a given value depends on packet alignment, not on the value, so the same logical row can panic or not depending on where the boundary fell. The divergence is documented on `EncodingType::encoding()`, but not here, where the implementor makes the choice. One line pointing at it would close the loop:

```rust
/// An override must stay behaviourally consistent with
/// [`RowWriter::write_string`]: which of the two receives a given value
/// depends on packet alignment, not on the value. In particular
/// [`EncodingType::encoding`] substitutes U+FFFD where `to_utf8_string`
/// panics.
```

mssql-tds/src/connection/transport/buffers.rs
```diff
@@ -89,6 +89,24 @@ impl TdsReadBuffer {
         self.try_read_array().map(|[value]| value)
     }
 
+    // consume_bytes only moves buffer_position, so the bytes stay readable until
+    // the next packet read overwrites working_buffer. That read needs &mut self,
```
David-Engel: **Nit:** "consume_bytes only moves buffer_position" isn't quite right — when the buffer drains exactly, `consume_bytes` resets *both* `buffer_position` and `buffer_length` to 0.

The safety conclusion is unchanged (nothing is zeroed, so `working_buffer[start..start + length]` stays valid until the next read), but that reset is the non-obvious part: `start` is deliberately captured before the call and is no longer relative to anything afterwards. Worth saying so explicitly, since this is the trickiest function in the PR.

David-Engel (commented):
Re-reviewed at `4a38961`. The `Cow` collapse is the right call and it makes the fast path unskippable — you cannot forget to forward a method that does not exist. It also **resolves my previous finding outright**: with one method there is no longer a packet-alignment-dependent split between `write_string` and `write_string_ref`, so the divergence I asked to document can't happen. The `RefSpy` rewrite asserting `Cow::Borrowed` survives the wrapper is a genuinely stronger test than the one it replaces, and the description is updated honestly, including the breaking-change section and the self-flagged stale benchmarks.

**Verified locally on `4a38961`:**

- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean (includes `mssql-js`, benches, CLI).
- `mssql-py-core` is `exclude`d from the workspace (`Cargo.toml:4`), so `cargo bclippy` / `cargo btest` skip it. Built it separately with `-D warnings` — clean. The description's claim checks out.
- `cargo nextest run -p mssql-tds --lib` — 1833/1837, only the 4 known `certificate_validator` fixture failures. Test count is unchanged at 1837 (one test removed, one added).
- Branch is 1 commit behind `main`.

## Findings

**Blocking** — none. One new suggestion inline, plus two carried items still open.

### Suggestion (carried, still unaddressed) — the new branch is not fuzzed

`mssql-tds/src/fuzz_support.rs` — `FuzzReader` (line 77), `FuzzPacketReader` (line 514) and `MockTransport` (line 936) implement or forward all ten `try_read_*` scalar probes from #299, but not `try_read_slice`, so it falls to the trait default returning `None`. `fuzz_token_stream` drives the full token/row path over `FuzzReader`, which means the borrowed arm never executes under fuzzing — the one harness aimed at adversarial length headers only ever sees the owned fallback.

`FuzzReader` is a flat `Vec<u8>` plus a cursor, so it is the same three-liner already written for `ByteReader` in the decoder tests:

```rust
fn try_read_slice(&mut self, length: usize) -> Option<&[u8]> {
    let end = self.position.checked_add(length)?;
    let bytes = self.data.get(self.position..end)?;
    self.position = end;
    Some(bytes)
}
```

plus the matching forwards in `FuzzPacketReader` and `MockTransport`. (In the body rather than inline since those files aren't in the diff.)

### Nit (carried, still unaddressed) — description's public-exposure claim

Still says "`lcid_to_encoding` already leaked `&'static Encoding` publicly, so this is not new exposure." `mssql-tds/src/datatypes.rs:17` is `pub(crate) mod lcid_encoding;`, so it is not reachable externally. `EncodingType::encoding()` plus `pub use encoding_rs;` genuinely is new exposure, and it does put `encoding_rs` 0.8 into the semver surface. Keep the re-export — it is the right mitigation — just fix the rationale so a future reader doesn't conclude the door was already open.

## Notes, not findings

- Agree the benchmarks need re-running, and the description is right to say so up front. Worth noting the writer side costs nothing extra: `Cow::into_owned()` on `Cow::Owned` is a move, `write_column_value` splits via `into_parts()` without cloning, and `mssql-js`'s `extend_from_slice(&bytes)` works on both variants without materializing. The only new cost is the 32-byte vs 16-byte argument, which is exactly what the re-measure needs to isolate.
- `SqlString::decode` preserving `String::from_utf8(..).unwrap()` on the `Utf8` arm keeps #310 behaviour identical on both paths — verified, and `decode_matches_to_utf8_string_across_encodings` pins it.
- `EncodingType: Copy` is safe: `SqlCollation` is `Copy` and the enum is payload-free otherwise.
- Branch is 1 commit behind `main`. Since this PR already has human review on it, please bring `main` in with a **merge** rather than a rebase, so the "changes since last review" diff survives.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -3995,6 +4036,108 @@ mod test {
             );
         }
 
+        /// Decodes `wire` twice, once through a reader that hands out borrows and
+        /// once through one that never does, and asserts both produce the same
+        /// value. Returns it so a caller can assert on the content too.
+        async fn borrowed_and_owned_agree(wire: Vec<u8>, md: &ColumnMetadata) -> ColumnValues {
```
David-Engel, line 4095: **Suggestion:** nothing in the tree guards that the decoder actually emits a borrow. I proved this rather than asserting it — mutating all three borrowed arms to throw the benefit away:

```rust
// decoder.rs:1482, 1528, 2079
writer.write_bytes(col, Cow::Owned(bytes.to_vec()));
writer.write_string(col, Cow::Owned(bytes.to_vec()), encoding_type);
```

leaves the suite fully green — **1833 passed / 4 failed**, byte-identical to the unmutated run (same 4 cert-fixture failures). That mutation deletes 100% of what this PR buys and no test notices.

Why the existing tests miss it:

- `dynamic_row_writer_forwards_borrows_intact` asserts `Cow::Borrowed` correctly, but constructs the `Cow` by hand — it never goes through the decoder.
- `borrowed_and_owned_agree` asserts value equality, which is by construction insensitive to the `Cow` variant.
- `RecordingSink::write_string` / `write_bytes` call `into_owned()`, discarding the evidence.

This matters *more* under the `Cow` shape than it did under `_ref`. With two methods the type system no longer carries any signal that a fast path exists — `Cow::Borrowed` vs `Cow::Owned` is a runtime choice at three call sites, and a future refactor that "simplifies" one of them is invisible.

Cheap fix: have this helper drive a spy that records `matches!(&bytes, Cow::Borrowed(_))` instead of a bare `DefaultRowWriter`, then assert `true` for the `ByteReader::new` arm and `false` for `new_unbuffered`. That reuses the two readers already built here and turns every existing `*_borrowed_matches_owned` case into a borrow-path guard as well.

mssql-tds/src/connection/transport/buffers.rs
David-Engel: **Nit** (carried from the last round, still open): "consume_bytes only moves buffer_position" isn't quite right — on an exact drain `consume_bytes` resets *both* `buffer_position` and `buffer_length` to 0.

The safety conclusion holds (nothing is zeroed, so `working_buffer[start..start + length]` stays valid until the next read), but that reset is precisely why `start` has to be captured before the call. Worth saying so, since this is the trickiest function in the PR.

mssql-tds/src/datatypes/decoder.rs
saurabh500: You're right, and thanks for actually mutation-testing it — this was a real gap.

The cause: both arms wrote through a `DefaultRowWriter`, and `DefaultRowWriter::write_string` calls `into_owned()`. So the `Cow` variant was thrown away before anything could observe it, and the test only ever compared the decoded *values*, which are identical either way. Rewriting all three borrowed call sites to `Cow::Owned(bytes.to_vec())` left the suite fully green, exactly as you found.

Fixed in 4830e63b by adding a `BorrowSpy` writer that wraps `DefaultRowWriter` and records, per string/binary write, whether the value arrived still borrowed. `borrowed_and_owned_agree` now asserts that the slice-offering reader took the borrowed arm and the unbuffered one did not, on top of the existing value equality.

Re-ran your mutation against the shipped source to confirm the guard has teeth:

| | before | after |
|---|---|---|
| tests killed by `Cow::Owned(bytes.to_vec())` at all 3 call sites | 0 / 5 | 4 / 5 |

The survivor is `short_string_null_borrowed_matches_owned`, which is correct — a NULL writes no string or binary value at all, so there is no borrow to assert on. The helper documents that and still checks both arms took the same number of writes, so a NULL that silently started emitting a value would be caught.

> [!NOTE]

mssql-tds/src/connection/transport/buffers.rs
saurabh500: Good catch, and the comment was factually wrong as well as incomplete. On an exact drain, `consume_bytes` resets **both** `buffer_position` and `buffer_length` to 0 — not just the position:

```rust
if self.buffer_length == self.buffer_position {
    self.buffer_position = 0;
    self.buffer_length = 0;
}
```

That reset is precisely why `start` has to be captured *before* the call: afterwards it is an absolute index into `working_buffer`, no longer an offset derived from any live field. Rewritten in 4830e63b to say that, rather than leaving the reader to infer it.

> [!NOTE]

saurabh500: Same nit as the one you raised in the later round — addressed together in 4830e63b. Details in [that thread](https://github.com/microsoft/mssql-rs/pull/343#discussion_r3858399190): the comment claimed `consume_bytes` only moves `buffer_position`, but on an exact drain it zeroes `buffer_length` too, which is the actual reason `start` is captured before the call.

> [!NOTE]

mssql-tds/src/datatypes/row_writer.rs
saurabh500: Taken, and it survives the `Cow` refactor — arguably sharper against the new shape, since there is now one method whose argument can arrive either way.

The old doc listed `EncodingType::encoding()` and `into_owned()` as parallel options, which reads as an invitation to branch on `Cow::Borrowed` vs `Cow::Owned`. That is a trap: which arm you get depends on where the packet boundary fell, not on the value, and the two decode paths are not equivalent — `encoding_rs` substitutes U+FFFD where `to_utf8_string`/`decode` panic on malformed `EncodingType::Utf8` input (#310). A writer that used `encoding()` on the borrowed arm and `to_utf8_string` on the owned one would make the same logical row behave differently run to run.

Rewritten in 4830e63b to state the rule first and then rank the options:

```rust
/// `bytes` borrows the read buffer when the value was already resident in
/// it, so it must not be retained past the call. Whether a given value
/// arrives borrowed depends on where the packet boundary fell, not on the
/// value, so an implementation must treat both arms identically.
///
/// [`SqlString::decode`] decodes either arm without copying the borrowed
/// one and is the consistent choice. [`EncodingType::encoding`] is
/// available for transcoding straight into a caller-owned buffer, but note
/// it substitutes U+FFFD where `decode` and [`SqlString::to_utf8_string`]
/// panic on malformed [`EncodingType::Utf8`] input (#310) — picking it for
/// the borrowed arm alone makes the same logical row behave differently
/// depending on packet alignment.
```

`write_bytes` points back at this rather than restating it.

> [!NOTE]

Theekshna:
**Blocking on the numbers.** `bench_narrow_control` and the FDW figures in the description were measured on the four-method `&[u8]` shape. `Cow<'_, [u8]>` is 32 bytes where `&[u8]` was 16, and it sits on the hot path — so the shape that actually ships is unmeasured. The numbers are the entire justification for this PR, so they need re-running before merge.

Everything else checks out. I re-ran @David-Engel's mutation independently (all three call sites rewritten to `Cow::Owned(bytes.to_vec())`): **4 of 5 tests fail**, survivor is the NULL case, exactly as claimed — the `BorrowSpy` guard has teeth. `try_read_slice` is sound: `do_we_have_enough_data` gates entry, `consume_bytes` validates before mutating so `consumed.ok()?` can't leave torn state, and the returned borrow is bound to `&mut self`, so the borrow checker forbids a refill while the slice is live.

One thing worth stating in the comment on `try_read_slice`, since it is the assumption the whole design rests on and is nowhere in the PR: `buffer_length` only ever covers header-stripped payload, because `read_tds_packet` calls `remove_header_from_packet` before extending it. That is why a returned span may legitimately cross a coalesced packet boundary without swallowing an 8-byte header.

mssql-tds/src/io/packet_reader.rs
```diff
@@ -73,6 +73,16 @@ macro_rules! define_tds_packet_reader {
...
+            /// without consuming when the whole run is not already resident.
+            ///
+            /// The borrow ends at the next use of the reader, so a caller must
+            /// copy or hand the bytes off before reading again.
+            #[inline]
+            fn try_read_slice(&mut self, _length: usize) -> Option<&[u8]> {
```
Theekshna, line 82: This default means the fuzzer never reaches the borrowed path.

`fuzz_tds_client` drives a real `TdsClient` through `execute()` and row fetch, which lands on `tds_client.rs:3855` -> `decode_into` -> the three new call sites. But `FuzzReader` (`fuzz_support.rs:77`), `FuzzPacketReader` (`:514`) and `MockTransport` (`:936`) all inherit `None` from here, so 100% of fuzz coverage runs the unchanged owned arm — none of it touches the new bounds arithmetic, which is precisely what fuzzing is good at.

Cheap to close: implement `try_read_slice` on `FuzzReader` (same shape as the test `ByteReader` in `decoder.rs`) and delegate from the other two. That points the existing corpus at the new code without writing a new target.

NivasSA:
@microsoft-github-policy-service agree company="Microsoft"

mssql-tds/src/io/packet_reader.rs
shiwanigupta0809, line 82: Adding to Theekshna's point — the gap is a bit wider than the fuzzers.

`ByteReader` in `decoder.rs:3720` is the only test double in the tree that overrides this. `TestTransport` (`tds_client.rs:6140`), `TestByteReader` (`token_stream.rs:1361`), `TestPacketReader` (`prelogin.rs:449`) and `MockReader` (`common.rs:70`) all inherit the default too, so the client-level tests don't reach the borrowed arm either. Coverage of the new path currently rests on one double written for this PR, which is also the one furthest from real packet-boundary behaviour.

The underlying reason is that a defaulted method is a silent opt-out — a reader that doesn't override it still compiles and still passes, it just quietly stops testing the feature. That's the same shape as the `write_string_ref` forwarding problem the `Cow` collapse fixed, one layer down. Might be worth asking whether `try_read_slice` should be required rather than defaulted, with the test doubles returning `None` explicitly — then the opt-out is visible in each file instead of inherited.

Happy either way if you'd rather just do the fuzz readers and move on.

shiwanigupta0809:
Nice work on this — the `Cow` collapse and the `BorrowSpy` follow-up to David's mutation testing both make the change much easier to trust. One small thing from my side.

`RowWriter` is public and both signatures change, but `CHANGELOG.md` isn't touched. `[Unreleased]` looks like it's being kept current otherwise, and this is the one change here that's guaranteed to break an out-of-tree implementer.

Also worth confirming `pub use encoding_rs;` in `lib.rs` is deliberate rather than incidental — it makes an `encoding_rs` major bump a breaking change for `mssql-tds`. The reason for it is sound (a consumer can't name `&'static encoding_rs::Encoding` otherwise), just worth having the coupling on the record.

mssql-tds/src/io/packet_reader.rs
saurabh500: Taken — this was the most valuable thing left on the PR, and the gap is a bit worse than you described. Fixed in 11544bcf.

Checking each double against the trait's eleven `try_read_*` methods:

| double | `try_read_*` overridden | `try_read_slice` |
|---|---|---|
| `FuzzReader` | 10 | ✗ |
| `FuzzPacketReader` | 9 | ✗ |
| `MockTransport` | 10 | ✗ |
| `EmptyReader` | 0 | ✗ (correct — no-data stub) |

So all three real doubles already opt into every buffered fast path and each missed exactly one. That makes it an internal inconsistency rather than a deliberate omission, which is a stronger reason to fix it than coverage alone.

Implemented on `FuzzReader` with the position moved before the borrow is taken, matching `BufferState::try_read_slice`, and delegated from the other two:

```rust
fn try_read_slice(&mut self, length: usize) -> Option<&[u8]> {
    let end = self.position.checked_add(length)?;
    if end > self.data.len() {
        return None;
    }
    let start = self.position;
    self.position = end;
    Some(&self.data[start..end])
}
```

`EmptyReader` keeps the inherited `None`, which is right for it.

Worth recording why this mattered beyond coverage: the PR adds no `unsafe` (the `set_len` calls in `decoder.rs` are pre-existing, on the owned arm), so a bounds mistake here surfaces as `&working_buffer[start..start + length]` panicking rather than as memory corruption. That is still a network-reachable panic, which this repo bans outright — and randomized lengths are exactly what a unit test can't supply.

Validation, since `fuzz_support` is `#[cfg(fuzzing)]` and neither `cargo bclippy` nor `cargo btest` compiles it:

- `RUSTFLAGS="--cfg fuzzing" cargo clippy -p mssql-tds --all-features --all-targets` → clean, no new warnings (the four `missing_docs` hits in the file are pre-existing, on `pub fn`s at 62/456/464/785).
- `RUSTFLAGS="--cfg fuzzing" cargo check --all-targets` in `mssql-tds/fuzz` → all nine targets build.
- Normal `cargo bfmt` / `cargo bclippy` clean; lib suite 1921/1928, the 7 failures being the untracked cert fixtures that also fail on `main`.

> [!NOTE]

saurabh500: Your facts check out — I counted 10 `TdsPacketReader` implementors and only 2 override `try_read_slice` (`NetworkTransport` and the `ByteReader` written for this PR). And the parallel you draw to `write_string_ref` is fair: both are silent performance opt-outs that still compile and still pass.

Taking your fallback — the fuzz readers are done in 11544bcf, details in [Theekshna's thread](https://github.com/microsoft/mssql-rs/pull/343#discussion_r3859920359).

Declining the "make it required" part, for three reasons:

**It's one of eleven.** The trait has 11 `try_read_*` methods and every one defaults to `None`. The property you're describing applies to all of them equally, so requiring just this one is arbitrary — and it leaves the trait less uniform than it is now.

**The remedy doesn't transfer.** Collapsing to `Cow` worked because it removed the *possibility* of a mismatch: one method, no way to implement half of it. Making `try_read_slice` required removes nothing. An implementor can still write `{ None }` and lose the path; it just becomes a typed-out opt-out instead of an inherited one. Visibility, not prevention.

**Different risk class.** `None` is always semantically *correct* here — the caller falls back to the async read and gets the same bytes. That is a performance opt-out. The `write_string_ref` case was the same in kind, which is why the real fix there was collapsing the API rather than un-defaulting it.

Scope-wise it's also cheap-looking but isn't: `io` is `pub(crate)` and the trait is generated by `define_tds_packet_reader!`, so there's no external breakage — but 8 in-tree implementors would each need a `None` stub purely for visibility.

On the client-level doubles (`TestTransport`, `TestByteReader`, `TestPacketReader`, `MockReader`): leaving those. The borrowed arm is already covered at the layers where it actually lives — the slice-probe tests in `buffers.rs`, the packet-boundary tests in `network_transport.rs`, and the decoder's `ByteReader`. Routing client-level doubles through it would be redundant integration coverage. Fuzzing was the one that added something no unit test can: randomized adversarial lengths.

If you think "defaulted `try_read_*` is a silent opt-out" is worth acting on, I agree it's a real observation — but it's a change across all 11 and 8 implementors, which belongs in its own issue rather than here. Happy to reopen this if you'd rather it block the PR.

> [!NOTE]

Theekshna (approved):

PR #326, opened 2026-08-19
Add SQLDescribeParam support

saurabh500:
## What's changed

- Export and advertise `SQLDescribeParam` through the ODBC driver.
- Infer undeclared parameter metadata with `sp_describe_undeclared_parameters`, map TDS metadata to ODBC types, and cache all parameter records until the statement is superseded.
- Support mssql-python's describe-then-bind path for `None` values by converting `SQL_C_DEFAULT`/`SQL_NULL_DATA` bindings into typed TDS NULL parameters while preserving exact SQL declarations.
- Add ODBC parity coverage for metadata, diagnostics, cache invalidation, and representative mssql-python NULL execution patterns.
- Document the TDS design, msodbcsql parity contract, implementation phases, and test plan in `mssql-odbc/docs/sql-describe-param-plan.md`.

[AB#47373](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47373)

PR #344, opened 2026-08-20
Add audited parameter buffer reader and integer parameter conversions

Theekshna:
Implements P2 (audited C-buffer reader) and P3 (integer C → integer SQL) of the
parameter conversion milestone. They ship together because P2's integer path has
no producer or consumer until P3 exists.

`conversion/param_buffer.rs` is now the single audited read of application
parameter buffers, private to `conversion` so nothing can bypass it. It reads in
two steps — `read_indicator` settles NULL and the special indicators before any
value buffer is dereferenced, then `read_param_value` returns an owned
`AppValue`. Integer parameters and typed NULLs are declared from `ParameterType`
rather than always going out as `nvarchar(max)`; non-NULL character values still
map to `(n)varchar(max)` regardless of `ParameterType`, which P4 addresses.

- **Unaligned reads.** Fixed-width C buffers use `read_unaligned`; ODBC promises
  no alignment, and msodbcsql qualifies every integer source read `UNALIGNED`.
- **Indicator rules.** `StrLen_or_Ind` is a length only for the variable-length C
  types; a fixed-width type takes its size from the C type. A null indicator
  pointer means `SQL_NTS`, not NULL. The special values (`SQL_NULL_DATA`,
  `SQL_DEFAULT_PARAM`, data-at-exec) are checked first for every C type.
- **`HY009`** for a non-NULL parameter with a null `ParameterValuePtr`. ODBC
  allows that pointer to be null only for `SQL_NULL_DATA` or data-at-exec. This
  *exceeds* msodbcsql, which rejects only when both pointers are null.
- **Integer conversions.** One matrix row per integer C type reaching all four
  integer SQL types. Width is not legality: a value that does not fit is a
  runtime `22003`, not a rejected binding. Narrowing shares
  `numeric::narrow_i128` with the fetch direction.
- **`SQL_C_TINYINT` is sign-unknown, not signed.** `sqlext.h` gives it neither
  `SQL_SIGNED_OFFSET` nor `SQL_UNSIGNED_OFFSET`, and msodbcsql resolves that by
  bit-copying a same-width tinyint transfer rather than converting
  (`sqlccnvt.cpp:4903`, `sqlcdata.cpp`, `ParamToSQLType` at `sqlcfunc.cpp:3193`).
  This PR does the same in both directions: a `tinyint` column copies its byte
  into `SQL_C_TINYINT`, and `SQL_C_TINYINT` bound to `SQL_TINYINT` reads
  unsigned. Any other source, or a wider `ParameterType`, keeps the signed range
  check, and `SQL_C_STINYINT` is never exempt. `tinyint` is the only unsigned
  SQL Server integer, so no other pairing needs this.
- **`ColumnSize` is validated at bind (`HY104`)**, mirroring msodbcsql's
  `CheckSqlPrecScale` and running after the type and conversion checks in the
  same order. `ColumnSize` 0 is `SQL_PREC_UNLIMITED`, so it means `max` only for
  `SQL_VARCHAR`, `SQL_WVARCHAR`, `SQL_VARBINARY`, and `SQL_SS_XML`; on the
  fixed-length types it is a zero-length declaration and is rejected. Previously
  this surfaced at execute, or not at all.

## Scope: parameter conversions are character and integer only

Defaulted bindings previously bypassed the conversion matrix via
`BoundParam::c_type_defaulted`. That flag is removed and defaulted bindings are
checked exactly like explicit ones, so an application gets the same answer
whether it names the C type or lets the driver pick it.

`SQLBindParameter` therefore rejects everything outside the character and integer
families — `SQL_DECIMAL`, `SQL_NUMERIC`, the binary types, `SQL_GUID`, and the
temporal types — **including when the bound value is `SQL_NULL_DATA`**. That is
deliberate rather than collateral: no C type can bind a *value* to those SQL
types yet, so accepting a typed NULL would hand an application a binding that
works for `NULL` and fails on its first real value.

The state is `HYC00`, not `07006`. This matrix is an implementation-progress
list, not a legality table like msodbcsql's `fValidConversion`; `HYC00` is the
state the ODBC spec defines for "the driver does not support the conversion
specified by the combination of `ValueType` and `ParameterType`", and it leaves
the application a path forward. A `TODO` on `is_supported_conversion` marks the
switch to `07006` once the table is complete.

`mssql-odbc` has had no release, so nothing observable regresses — but this does
narrow the `SQLPrepare` → `SQLDescribeParam` → `SQLBindParameter(SQL_C_DEFAULT)`
→ `SQLExecute` flow #326 added, and it is a divergence from msodbcsql for as long
as the matrix is partial. Parked until the rows land, all tracked by
[AB#47500](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47500):
`DISABLED_DescribesMaxLengthParameters`,
`DISABLED_DescribedDecimalRoundTripsPrecisionAndScale` (which also guards the
scale-0 wire-metadata regression), and the `#[ignore]`d
`default_c_type_guid_is_accepted_and_stored`. The supported set and the NULL rule
are stated in `CHANGELOG.md`.

## Validation

`cargo bfmt`, `cargo bclippy`, and `cargo test -p mssql-odbc` (770 passed, 1
ignored) clean after rebasing onto `origin/main`. Full e2e suite passes,
including `--compare-with-msodbcsql` parity.

New e2e coverage for unaligned buffers, explicit and defaulted integer NULLs,
stray indicators, `SQL_C_UBIGINT` overflow, and — as msodbcsql parity cases, not
skipped — `TinyintCTypeIsUnsignedAgainstTinyintParameter`,
`TinyintColumnAbove127FetchesIntoTinyintCType`, and
`ZeroColumnSizeIsRejectedForFixedLengthCharTypes`.

`every_bind_accepted_column_size_is_declarable` asserts that any `ColumnSize` the
bind gate accepts can still be declared at execute, since the two encode the same
limits in separate places.

[AB#47366](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47366)
[AB#47367](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47367)
[AB#47500](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47500)


mssql-odbc/src/conversion/param_buffer.rs
```diff
@@ -0,0 +1,431 @@
...
+/// # Safety
+/// `ptr`, if non-null, must be readable for the resolved number of `u16` units
+/// (or up to the first NUL when `len_spec == SQL_NTS`).
+unsafe fn read_wchar_bytes(ptr: *const u16, len_spec: SqlLen) -> Vec<u8> {
+    if ptr.is_null() {
+        return Vec::new();
+    }
+    let units = if len_spec == SQL_NTS as SqlLen {
+        let mut n = 0usize;
+        while unsafe { ptr.add(n).read_unaligned() } != 0 {
```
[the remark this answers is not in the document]
Theekshna: Agreed - the wide path reads unit by unit specifically to avoid asserting an alignment the application never promised, and nothing pinned it. Added `misaligned_wchar_buffer_is_read` in fe63226c, decoding UTF-16LE from an odd byte offset. Same Miri/strict-alignment caveat as the indicator test.

mssql-odbc/docs/parameters_plan.md
```diff
@@ -256,31 +260,51 @@ Deviations from msodbcsql, verified against source:
...
-- Identity fast path for exact pairs (msodbcsql `IsParamConversionNeeded`).
-- Cross-width narrowing through the shared numeric model, `22003` on overflow.
-- Emit `SqlType::TinyInt` / `SmallInt` / `Int` / `BigInt` driven by
-  `ParameterType`, not the C type, including typed NULL. This is the first phase
-  where `@P1` is declared `int` instead of `nvarchar(max)`.
-- `SQL_C_TINYINT` is unsigned, matching the documented fetch decision.
-- `SQL_C_UBIGINT` above `i64::MAX` has no SQL Server target: `22003`.
+#### P2 - Buffer reader and outcome channel (code complete)
+
+Landed with P3, the phase that gives its integer path a producer and a consumer.
```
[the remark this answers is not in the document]
Theekshna: Good catch - the table row promised more than the phase delivered. `build_named_params` does still return `Result<Vec<RpcParameter>, SqlReturn>`.

Fixed in fe63226c: P2's deliverable and section heading are now just "Safe C-buffer reader", and the warning channel moved onto the P5 row, which is the phase that first produces `01S07` and so gives it a producer to test against.

mssql-odbc/src/conversion/param_buffer.rs
```diff
@@ -0,0 +1,431 @@
...
+/// Classifies `StrLen_or_Ind` without touching the value buffer, so a caller
+/// can reject a binding before any application data is dereferenced.
+///
+/// # Safety
+/// `param.strlen_or_ind_ptr`, if non-null, must point to one valid `SqlLen`.
+pub(crate) unsafe fn read_indicator(param: &BoundParam) -> Result<Indicator, ParamBuildError> {
+    let indicator = if param.strlen_or_ind_ptr.is_null() {
+        None
+    } else {
+        Some(unsafe { param.strlen_or_ind_ptr.read_unaligned() })
```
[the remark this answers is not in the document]
Theekshna: Valid - every existing test bound an aligned `SqlLen` local, and the e2e misalignment case offsets only `ParameterValuePtr`. Added `misaligned_indicator_pointer_is_read` in fe63226c: an `align(8)` byte array with the indicator taken at offset 1, covering both the `SQL_NULL_DATA` branch and the length branch.

One caveat on the guarantee: on x86-64 an ordinary unaligned dereference still executes correctly, so swapping `read_unaligned` for `*ptr` would not fail this test in normal CI. It is a tripwire for Miri or a strict-alignment target rather than a runtime guard.

mssql-odbc/src/api/bind_param.rs
```diff
@@ -197,22 +196,8 @@ fn sql_bind_parameter_safe(
...
-    // qualified server type name, which `SQLDescribeParam` does not report and
-    // this driver cannot otherwise obtain, so they are rejected on the bind call
-    // rather than at execute time.
-    let unsupported_default =
-        c_type_defaulted && matches!(parameter_type, SQL_SS_UDT | SQL_SS_TABLE);
-    if unsupported_default
-        || (!c_type_defaulted && !is_supported_conversion(c_type, parameter_type))
-    {
+    // The C type -> SQL type conversion must be one we support (07006).
+    if !is_supported_conversion(c_type, parameter_type) {
```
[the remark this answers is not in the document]
Theekshna: Confirmed, and this is the intended trade rather than an oversight - but you're right that it was under-documented. Defaulted bindings are now checked exactly like explicit ones so an application gets the same answer whether it names the C type or lets the driver pick it; the cost is that describe-then-bind is restricted to types the matrix lists.

Rather than special-casing a typed-NULL path (bind cannot know the value will be NULL - ODBC binds by reference and the indicator may be set after `SQLBindParameter` returns), the divergence is now stated explicitly in the PR description with the affected type list, and the two protecting tests are parked as `DISABLED_` with the type they wait on and the re-enable condition. Product-owner sign-off and the downstream-consumer check are called out there and are still outstanding.

mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -200,13 +190,13 @@ pub(crate) unsafe fn convert_integer_c(
...
-        // SQL_C_TINYINT maps to an unsigned SQLCHAR (SQL Server `tinyint` is
-        // 0-255 and mssql-python fetches it unsigned); only SQL_C_STINYINT is
-        // the signed form.
-        SQL_C_STINYINT => unsafe { write_fixed(target_value_ptr, narrow!(i8), strlen_or_ind_ptr) },
-        SQL_C_TINYINT | SQL_C_UTINYINT => unsafe {
-            write_fixed(target_value_ptr, narrow!(u8), strlen_or_ind_ptr)
+        // SQL_C_TINYINT is the ODBC 2.x spelling and is signed, so it groups with
+        // SQL_C_STINYINT; only SQL_C_UTINYINT is unsigned. msodbcsql's
+        // ConvertToChar (sqlccnvt.cpp:1660) splits them the same way.
+        SQL_C_STINYINT | SQL_C_TINYINT => unsafe {
```
shiwanigupta0809: **Blocking.** The cited justification doesn't support this change, and the result diverges from msodbcsql in both directions.

`sqlccnvt.cpp:1663` is inside `ConvertToChar`, which reads an application buffer as a **source** when converting *to a character string*. It establishes that the source read is `SCHAR` — correct, and the only path where "signed" is msodbcsql's answer. It says nothing about `SQL_C_TINYINT` as a fetch **target**, which is what this line controls.

The spec is silent here. `SQL_C_TINYINT` is absent from the C Data Types table in ODBC Appendix D; footnote [j] only requires drivers to keep accepting it for ODBC 2.x apps and assigns it no ODBC C typedef. `sqlext.h` agrees — `SQL_C_STINYINT`/`SQL_C_UTINYINT` carry `SQL_SIGNED_OFFSET`/`SQL_UNSIGNED_OFFSET`, while `SQL_C_TINYINT` is a bare alias for `SQL_TINYINT`. Each driver picks, and msodbcsql states its pick outright in `ConvertToFixed` (`sqlccnvt.cpp:4904`):

```cpp
//  Special case where input or output is sign unknown and opposite parameter is same size.
if ((fTypeIn  == SQL_C_BINARY  && cbIn == cbOut) ||
    (fTypeIn  == SQL_C_TINYINT && (fTypeOut == SQL_C_UTINYINT || fTypeOut == SQL_C_STINYINT)) ||
    (fTypeOut == SQL_C_TINYINT && (fTypeIn  == SQL_C_UTINYINT || fTypeIn  == SQL_C_STINYINT)) ||
    ...
{
    fTypeIn = fTypeOut; //  Force bit copy
}
```

"sign unknown" — grouped with `SQL_C_SHORT` and `SQL_C_LONG`, the other ODBC 2.x aliases, and resolved by copying bits rather than converting. Three more places reinforce it:

- `sqlccnvt.cpp:5281` (`ConvertToFixed`'s output store, the fetch-target path) skips the range check entirely when the source is any tinyint flavour, then stores `(SCHAR)Temp`.
- `sqlcdata.cpp:963` — `if (wSrcColCType == SQL_C_UTINYINT) fConvNeeded = FALSE;` — raw byte copy, no conversion at all.
- `sqlcfunc.cpp:3192`, in `ParamToSQLType` (`SQLBindParameter`'s conversion entry point):
  ```cpp
  // If both are tinyint, change C type to unsigned.
  if (fCType == SQL_C_TINYINT && fSqlType == SQL_TINYINT)
      fCType = SQL_C_UTINYINT;
  ```

The divergence therefore runs in both directions:

**Fetch** — `SELECT CAST(200 AS TINYINT)` into `SQL_C_TINYINT`. A `tinyint` column is `SQL_C_UTINYINT` at the source, so msodbcsql hits both fast paths: `SQL_SUCCESS`, byte copied. `main` today returns `SQL_SUCCESS` with 200. After this change: `22003`, no data. That's every `tinyint` value above 127.

**Parameter** — `SQLBindParameter(SQL_C_TINYINT, SQL_TINYINT, ...)` with `0xFF`. msodbcsql rewrites the C type to unsigned and sends 255. Here `read_integer` reads `-1` as `i8`, `integer_value` calls `narrow_i128::<u8>(-1)`, and it fails `22003`. `SQL_TINYINT` is in `INTEGER_C_TARGETS`, so the bind is accepted and the failure surfaces at execute.

Neither is covered. `TinyintCTypeIsSignedButUtinyintIsNot` binds against `SQL_SMALLINT`, not `SQL_TINYINT`, and there's no e2e for the fetch direction — which is why the `--compare-with-msodbcsql` run doesn't contradict this.

Suggest adopting msodbcsql's policy and applying it in both directions: keep the signed storage, but treat a same-width tinyint transfer as a bit copy rather than a range-checked conversion.

```rust
SQL_C_STINYINT | SQL_C_TINYINT => {
    // msodbcsql treats SQL_C_TINYINT as sign-unknown (ConvertToFixed,
    // sqlccnvt.cpp:4904) and forces a bit copy for same-width tinyint
    // transfers, so a 0-255 `tinyint` column always reaches the buffer.
    let byte = narrow!(u8);
    unsafe { write_fixed(target_value_ptr, byte as i8, strlen_or_ind_ptr) }
}
```

plus the `SQL_C_TINYINT` -> `SQL_TINYINT` rewrite on the parameter side, an e2e parity case for `tinyint` -> `SQL_C_TINYINT`, and a `CHANGELOG.md` entry since this changes shipped fetch behaviour.

If the intent is to diverge deliberately because `22003` is more honest than a silent reinterpretation, that's defensible — but it would be better as a documented divergence with a parity-suite exception rather than a side effect of a parameter PR.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -208,7 +195,7 @@ fn null_value(param: &BoundParam) -> Result<TypedValue, ParamBuildError> {
 /// The returned [`RpcTypeMetadata`] is the *only* place precision and scale are
 /// carried. [`RpcParameter`] uses it to render the declaration and to write the
 /// wire `TYPE_INFO`, so the two cannot drift apart.
-fn default_typed_null(
+fn typed_null(
```
shiwanigupta0809, line 199: **Blocking.** Routing every NULL through `typed_null` makes `column_size` significant for explicit character bindings, which it wasn't on `main`:

```rust
fn null_value(param: &BoundParam) -> Result<TypedValue, ParamBuildError> {
    if param.c_type_defaulted {
        return default_typed_null(param.sql_type, param.column_size, param.decimal_digits);
    }
    match param.c_type {
        SQL_C_CHAR  => Ok((SqlType::VarcharMax(None), None)),   // never looked at column_size
        SQL_C_WCHAR => Ok((SqlType::NVarcharMax(None), None)),
        other => Err(ParamBuildError::UnsupportedCType(other)),
    }
}
```

Now `SQL_CHAR => Char(None, fixed_length(column_size, MAX_NARROW_LENGTH)?)`, and `fixed_length` rejects zero via `(1..=max).contains(&column_size)` -> `InvalidParameterSize(0)` -> `HY104`. So:

```c
SQLLEN ind = SQL_NULL_DATA;
SQLBindParameter(stmt, 1, SQL_PARAM_INPUT, SQL_C_CHAR, SQL_CHAR, 0, 0, buf, 0, &ind);
SQLExecute(stmt);   // was SQL_SUCCESS on main, now SQL_ERROR / HY104
```

Same for `SQL_C_WCHAR` -> `SQL_WCHAR`. `SQL_CHAR` is in `CHAR_C_TARGETS` and there's no bind-time `column_size` validation, so the bind succeeds and the failure lands at execute. Passing `ColumnSize = 0` when binding a NULL is ordinary practice — there's no length to declare.

Two things make this easy to miss. The behaviour is asymmetric: the identical binding succeeds with a value and fails with NULL, because the non-NULL path ignores `sql_type` entirely (`AppValue::NarrowText` -> `VarcharMax` unconditionally) while the NULL path now honours it strictly. And both NULL unit tests were updated to assign the *variable*-length types:

```rust
let mut p = param(SQL_C_CHAR, std::ptr::null_mut(), &mut ind);
p.sql_type = SQL_VARCHAR;   // routes through variable_length, widens 0 to max
```

`SQL_VARCHAR`/`SQL_WVARCHAR` are unaffected, which leaves no coverage for `SQL_CHAR`/`SQL_WCHAR` with `column_size = 0` — the exact shape that regressed.

`fixed_length`'s strictness is right for a real value (`char(0)` isn't legal T-SQL), but a NULL carries no data, so the declaration only needs to be legal, not faithful:

```rust
// `ColumnSize` 0 is what an application passes when it has no length to
// declare - routine for a NULL. `char(0)` is not legal T-SQL, so widen to
// the smallest legal declaration rather than rejecting the binding.
SQL_CHAR => SqlType::Char(None, fixed_length(column_size, MAX_NARROW_LENGTH).unwrap_or(1)),
```

Either that, or keep the previous `VarcharMax`/`NVarcharMax` for explicit character binds. Either way, could you add unit and e2e cases for `SQL_C_CHAR -> SQL_CHAR` and `SQL_C_WCHAR -> SQL_WCHAR` with `ColumnSize = 0` and `SQL_NULL_DATA`?

mssql-odbc/src/api/bind_param.rs
shiwanigupta0809, line 200: **Blocking**, in answer to the sign-off request — though the objection isn't the compatibility question itself.

It's this sentence from the description: *"`typed_null` still handles all of them; they are simply no longer reachable."* The code to send a decimal, GUID, binary, or temporal NULL exists and works — it's being switched off, not removed as broken. A refactor that deletes one `bool` shouldn't have to cost that.

What it narrows is the flow #326 added:

```c
SQLPrepare(stmt, "INSERT INTO t VALUES (?)");
SQLDescribeParam(stmt, 1, &sqlType, &size, &digits, &nullable);   // SQL_DECIMAL
SQLBindParameter(stmt, 1, SQL_PARAM_INPUT, SQL_C_DEFAULT, sqlType, size, digits,
                 NULL, 0, &ind);   // ind = SQL_NULL_DATA
SQLExecute(stmt);
```

which is how a generic data-access layer inserts a NULL without hardcoding types. That covers ~18 SQL types, and the collateral is four tests: `DISABLED_DescribesMaxLengthParameters`, `DISABLED_DescribedDecimalRoundTripsPrecisionAndScale` (which per its own comment *"also guards the scale-0 wire-metadata regression"*, so an unrelated past fix loses its guard), `#[ignore]`d `default_c_type_guid_is_accepted_and_stored`, and deleted `DefaultCTypeNullBindsAndExecutes` — with re-enablement *"gated on decimal and binary parameter support"* and no tracking item.

There's a cleaner place for the check. The matrix answers *"can I convert this value?"*, and a NULL has no value to convert — so consult it where the value actually appears:

```rust
// bind: a defaulted binding is a promise about the SQL type, not the buffer.
// Whether the value is convertible is only knowable once the indicator is
// read, so defer that to execute.
if !c_type_defaulted && !is_supported_conversion(c_type, parameter_type) {
```

then in `bound_param_to_value`, in the branch where a real value exists:

```rust
Indicator::Length(len) => {
    if !is_supported_conversion(param.c_type, param.sql_type) {
        return Err(ParamBuildError::UnsupportedCType(param.c_type));  // 07006
    }
    len
}
```

That still delivers what the refactor was for — one code path, defaulted and explicit bindings answered identically for real values, no `c_type_defaulted` branching in the conversion logic — while typed NULLs keep working for all 18 types, all four tests stay enabled, and no sign-off is needed.

If the bind-time rejection has to ship as-is, could you file the tracking item before merge and note the `SQLDescribeParam` limitation in the CHANGELOG? Tests parked behind an ungated dependency tend to stay parked.

shiwanigupta0809 (commented):
Reviewed the parameter-buffer refactor and cross-checked the msodbcsql behaviour claims against the C++ sources. The core design holds up well: settling NULL from the indicator before any value pointer is dereferenced is the right shape, the unaligned-read reasoning is correct and worth having in the instructions file, and `i128` as the single integer carrier is clean. Nice cleanup overall.

Three things I'd like to work through before this goes in. Two of them change shipped behaviour in ways the description doesn't call out, and in each case the test that would have caught it was moved, renamed, or disabled in this same PR — so a green suite doesn't rule them out. Details inline, along with suggested fixes.

Two smaller things, not blocking:

- `read_integer`, `read_char_bytes`, and `read_wchar_bytes` each re-check `ptr.is_null()` after `read_param_value` has already rejected it, and all three disagree with it — `read_integer` returns `None` (mapped to `07006`, wrong SQLSTATE) and the text readers return an empty `Vec` (would silently send an empty string). All unreachable today, but `read_char_bytes_edge_cases` tests one directly, which pins behaviour that contradicts the module's own contract. Same for the `len_spec < 0 => 0` branches, which `read_indicator` already rejects. Worth deleting in favour of a `debug_assert!` plus a note in the existing `# Safety` sections.
- `CHANGELOG.md` is untouched. Both blocking items below are app-visible, and the repo keeps per-crate `mssql-odbc:` bullets under `[Unreleased]`.

mssql-odbc/src/conversion/fetch_convert.rs
David-Engel: **Blocking:** `SQL_C_TINYINT` is not uniformly signed in msodbcsql; it is sign-unknown for same-width tinyint transfers. `ConvertToFixed` (`sqlccnvt.cpp:4894-4904`) forces a bit copy when one side is `SQL_C_TINYINT` and the other is `SQL_C_STINYINT`/`SQL_C_UTINYINT`, and the fetch fast path in `sqlcdata.cpp` likewise copies a server `tinyint` byte directly. Consequently, fetching SQL `tinyint` 200 into `SQL_C_TINYINT` succeeds there, while this arm narrows to `i8` and returns `22003`. The parameter direction has the matching issue: `ParamToSQLType` (`sqlcfunc.cpp:3190-3195`) rewrites `SQL_C_TINYINT -> SQL_TINYINT` to unsigned, so raw `0xFF` is 255; this PR reads it as -1 and returns `22003`. Please special-case same-width `SQL_TINYINT` transfers in both directions while retaining signed behavior for conversions to wider SQL types. Add parity tests for fetching 200 into `SQL_C_TINYINT` and binding `0xFF` from `SQL_C_TINYINT` to SQL `TINYINT`; the current tests instead pin the divergent fetch error and only bind against `SQL_SMALLINT`.

mssql-odbc/src/api/bind_param.rs
David-Engel, line 200: **Blocking:** Applying the partial conversion matrix here rejects `SQL_C_DEFAULT` before execution even when the eventual indicator is `SQL_NULL_DATA`, where no C-buffer conversion occurs. This regresses the generic `SQLDescribeParam -> SQLBindParameter(SQL_C_DEFAULT) -> SQLExecute` flow added by #326 for decimal, binary, GUID, temporal, and other typed NULLs, despite `typed_null` already supporting them; this PR disables or ignores the tests that protected that flow. Preserve whether the C type was defaulted, exempt defaulted bindings from this bind-time matrix check, and apply the matrix after `read_indicator` returns `Indicator::Length`, before reading a real value. That keeps the audited reader and prevents unsupported non-NULL conversion while retaining typed NULLs. If the bind-time rejection must ship, it needs explicit product-owner sign-off and a tracking item for the acknowledged msodbcsql compatibility regression.

David-Engel (commented):
**Review produced using GPT-5.6 Sol + High Effort.**

## Summary

The audited parameter-buffer reader and integer range-conversion structure are otherwise sound, and the pointer/indicator rules agree with the ODBC reference. I found two blocking compatibility issues against the msodbcsql reference implementation; details and concrete fixes are inline.

Validation: `cargo nextest run -p mssql-odbc --lib --no-fail-fast` passed 694 tests with 1 skipped. All 19 CI checks pass with 97% diff coverage. The PR currently conflicts with `main` in `fetch_convert.rs` and needs a rebase plus fresh validation.

mssql-odbc/src/conversion/fetch_convert.rs
David-Engel: **Blocking.** The citation is accurate but partial, and the conclusion drawn from it regresses `tinyint` fetch.

msodbcsql treats `SQL_C_TINYINT` as signed only when converting to a *different* representation. Against its same-width counterpart it does a bit copy:

- `sqlccnvt.cpp:4899-4906` -- "*Special case where input or output is sign unknown and opposite parameter is same size*": `fTypeIn == SQL_C_TINYINT && fTypeOut == SQL_C_UTINYINT` (and the reverse) forces `fTypeIn = fTypeOut`, i.e. a raw byte copy.
- `sqlccnvt.cpp:5277-5286` -- the `case SQL_C_TINYINT:` output arm skips the range check entirely when `fTypeIn` is `SQL_C_TINYINT`/`SQL_C_STINYINT`/`SQL_C_UTINYINT`. Same shape at `:5299` for the `SQL_C_UTINYINT` arm.
- `sqlcdata.cpp:964-971` -- on fetch it never even reaches the converter: target `SQL_C_TINYINT` with source column C type `SQL_C_UTINYINT` sets `fConvNeeded = FALSE`. And `tinyint` (`SQLINT1`) maps to `SQL_C_UTINYINT` in the `rgbSRV2CTYPE` table at `sqlcdata.cpp:931`.

So `SELECT CAST(200 AS tinyint)` into `SQL_C_TINYINT` delivers the byte under msodbcsql and returns `22003` after this change. The new `tinyint_c_target_is_signed` unit test locks that in.

Downstream, mssql-python maps `SQL_TINYINT` -> `SQL_C_TINYINT` on fetch (`cursor.py:946`; `ddbc_bindings.cpp:3740` `SQLGetData` into a `SQLCHAR`, `:3909` `SQLBindCol` into `charBuffers`, which is `std::vector<SQLCHAR>`), and its suite asserts `rows[1][3] == 255` for a TINYINT column (`test_004_cursor.py:14999` and `:15412`) plus `2**8-1` in `test_004_cursor_arrow.py:27`. The parameter side is the same story: `cursor.py:447` binds a Python int in 0..255 as `SQL_C_TINYINT` + `SQL_TINYINT`, which msodbcsql bit-copies to 200 and this branch reads as `i8` -56 -> `22003`.

Suggested smallest fix -- keep the signed read everywhere the new tests exercise (character/decimal source into `SQL_C_TINYINT`, and `SQL_C_TINYINT` into `SQL_SMALLINT`), and restore the sign-unknown rule for the one same-width pairing:

- fetch: when the source is `ColumnValues::TinyInt` and the target is `SQL_C_TINYINT`, write the byte through unchanged rather than narrowing to `i8`;
- parameter: when `c_type == SQL_C_TINYINT` and `sql_type == SQL_TINYINT`, read the byte as `u8`.

`tinyint` is the only unsigned column source in the driver, so this stays scoped to exactly that pairing and leaves the rest of the change (which I agree with) intact.

mssql-odbc/docs/typed-columnar-fetch-plan.md
```diff
@@ -98,7 +98,7 @@ These were found by reading `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp` while reviewi
 | Case | msodbcsql | mssql-odbc | Status |
 | --- | --- | --- | --- |
 | A UTC offset in a literal, for any target other than `SQL_C_SS_TIMESTAMPOFFSET` | shifts the value into the client's local zone (`ConvertOffsetToLocal`) | validates the offset, then delivers the wall-clock fields as written | **Deliberate.** Matching would make the returned value depend on the client machine's time zone. Locked in by `offset_is_ignored_for_non_offset_targets`. |
-| Character or decimal source into `SQL_C_TINYINT` above 127 | `22003` — the signed limit applies whenever the input type is not itself a tinyint C type | `Ok(200)` — the target is `u8`, matching a real `tinyint` column (0-255) and what mssql-python fetches | **Deliberate**, but P1a is what first opens these sources into that target, so the divergence starts here. |
+| Character or decimal source into `SQL_C_TINYINT` above 127 | `22003` | `22003` | **Resolved.** `SQL_C_TINYINT` is the ODBC 2.x spelling of the signed form, so the signed limit applies; it was briefly read as `u8` here. Only `SQL_C_UTINYINT` is unsigned. |
```
David-Engel: Tied to the `fetch_convert.rs` finding: this row is only resolved for the *character or decimal* sources it names. A `tinyint` **column** source into `SQL_C_TINYINT` moves the other way -- msodbcsql bit-copies it (`sqlcdata.cpp:964-971`, `sqlccnvt.cpp:4899-4906`) and this branch now returns `22003`.

Suggest keeping a divergence row here rather than "Resolved", with the tinyint-source case called out explicitly, so the next person reading this table does not conclude the whole `SQL_C_TINYINT` question is settled.

mssql-odbc/src/api/bind_param.rs
David-Engel, line 200: **Blocking / needs a decision.** Answering the question in the description directly: yes, a downstream consumer depends on the affected types today.

- `mssql-python` `cursor.py:1001-1002` -- with `setinputsizes`, a `None` value is bound as `SQL_C_DEFAULT` with the declared SQL type, including `SQL_DECIMAL`/`SQL_NUMERIC` (handled explicitly a few lines below). That bind now returns `07006`.
- `mssql-python` `cursor.py:418-425` -- unresolved `None` parameters go out as `SQL_C_DEFAULT` + `SQL_UNKNOWN_TYPE` and are resolved via `SQLDescribeParam` in the pybind layer. That is exactly the describe-then-bind flow #326 added.

On the rationale: I do not think the consistency argument in `parameters_plan.md` holds as stated. It conflates "not implemented yet" with "illegal conversion". msodbcsql's `fValidConversion` is a legality table; this matrix is an implementation-progress list. `07006` tells the application the conversion is illegal and leaves it no path forward, for a case that worked before.

Smallest alternative that keeps both goals:

- land the typed-NULL fix as written -- building every NULL from `sql_type` is clearly right, and `CRPCSQLSender::SendParamList` agrees;
- keep the bind-time exemption for a resolved `SQL_C_DEFAULT`;
- keep `main`'s `default_non_null_rejects_sql_types_that_borrow_a_character_c_type` guard, so a non-NULL defaulted `decimal` still fails at execute with `HYC00` instead of going out as `varchar(max)`.

That preserves defaulted NULLs (the common describe-then-bind case) while still keeping the wrong type off the wire.

If you would rather keep the narrowing: it needs a tracking work item before merge. The description says re-enabling "has no tracking item yet" while two e2e tests are parked, which leaves `DISABLED_DescribesMaxLengthParameters` and `DISABLED_DescribedDecimalRoundTripsPrecisionAndScale` with no owner. Please file it and reference it from both `DISABLED_` comments.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -106,67 +119,53 @@ pub(crate) unsafe fn bound_param_to_rpc(
...
-        if ind == SQL_DEFAULT_PARAM {
-            return Err(ParamBuildError::InvalidUseOfDefaultParam);
-        }
-        if ind == SQL_DATA_AT_EXEC || ind <= SQL_LEN_DATA_AT_EXEC_OFFSET {
-            return Err(ParamBuildError::DataAtExecUnsupported);
+    // NULL is settled from the indicator alone, so a typed NULL never reads the
+    // value buffer.
+    let len_spec = match unsafe { read_indicator(param) }? {
+        Indicator::Null => {
+            return typed_null(param.sql_type, param.column_size, param.decimal_digits);
```
David-Engel, line 127: **Suggestion.** Routing every NULL through `typed_null` also routes *explicit* character binds through `fixed_length`, which is a behaviour change not mentioned in the description.

On `main`, `SQLBindParameter(SQL_C_CHAR, SQL_CHAR, /*ColumnSize*/ 0, ..., SQL_NULL_DATA)` took the `null_value` path and produced a `varchar(max)` NULL, which executed. Now it reaches `typed_null(SQL_CHAR, 0, 0)` -> `fixed_length(0, 8000)` -> `HY104`. Verified with a scratch test on this branch (since reverted):

```
SQL_CHAR/size0 NULL  => Err(InvalidParameterSize(0))
SQL_WCHAR/size0 NULL => Err(InvalidParameterSize(0))
```

Converging with msodbcsql's `CheckSqlPrec` is defensible, but nothing covers it: `BindChar` in the e2e suite always passes `store.size()`, so `NullParam` goes down the `SQL_VARCHAR` path and never exercises this. Either call the change out and add a case, or treat `ColumnSize == 0` on a NULL as the `max` spelling the way `variable_length` already does.

mssql-odbc/src/conversion/param_buffer.rs
```diff
@@ -0,0 +1,469 @@
...
+/// `param.parameter_value_ptr` must satisfy the `SQLBindParameter` contract:
+/// readable for `len_spec` (or to the first NUL when `len_spec` is `SQL_NTS`)
+/// for the character types, and for the C type's width otherwise.
+pub(crate) unsafe fn read_param_value(
+    param: &BoundParam,
+    len_spec: SqlLen,
+) -> Result<AppValue, ParamBuildError> {
+    // ODBC permits a null `ParameterValuePtr` only for `SQL_NULL_DATA` or
+    // data-at-exec, and `read_indicator` has already returned on both.
+    if param.parameter_value_ptr.is_null() {
```
David-Engel, line 107: **Nit.** This settles the null pointer before dispatch, so the `is_null()` branches in `read_integer`, `read_char_bytes`, and `read_wchar_bytes` are unreachable from production. That is fine as belt-and-braces, but `read_char_bytes_edge_cases` and `read_wchar_bytes_edge_cases` are then asserting on paths nothing can reach, which reads as coverage the module does not actually have. Worth a one-line note on those tests saying they pin the helpers' own contract rather than driver behaviour.

mssql-odbc/src/conversion/param_buffer.rs
```diff
@@ -0,0 +1,469 @@
...
+    if let Some(ind) = indicator {
+        if ind == SQL_NULL_DATA {
+            return Ok(Indicator::Null);
+        }
+        if ind == SQL_DEFAULT_PARAM {
+            // This value is valid only in a procedure called in ODBC canonical syntax,
+            // which this driver does not support yet.
+            return Err(ParamBuildError::InvalidUseOfDefaultParam);
+        }
+        if ind == SQL_DATA_AT_EXEC || ind <= SQL_LEN_DATA_AT_EXEC_OFFSET {
```
David-Engel, line 79: **Nit / question.** This applies the data-at-exec check to fixed-width C types too, so a stray `-200` on an `SQL_C_SLONG` binding becomes `HYC00` while a stray `-7` is ignored. The comment three lines below says a fixed-width type's leftover indicator "is ignored rather than validated", which reads as a contradiction.

I suspect the behaviour is intended -- the special values are meaningful whatever the C type -- but `StrayIndicatorIsIgnoredForFixedWidthCTypes` only covers `-7`, `0`, and `999`, so nothing pins which side of `SQL_LEN_DATA_AT_EXEC_OFFSET` the boundary sits on. Either extend that test past `-100` or adjust the comment to say the special values are checked first for every C type.

David-Engel (commented):
**Review produced using Claude Opus 5 + High Effort + 1M context.**

## Summary

Solid, carefully-cited work. `param_buffer.rs` as the single audited read is the right shape, the two-step `read_indicator` -> `read_param_value` split is correct, the unaligned-read discipline is real (I confirmed `IsFixedCType` at `sqlcprot.h:1301` and the `UNALIGNED` integer reads at `sqlccnvt.cpp:1671-1686`), and the two invariant tests (`every_accepted_pairing_is_convertible`, `every_supported_sql_type_has_a_typed_null`) are the kind that keep a matrix honest. Diff coverage 97%.

Reviewed in a dedicated worktree at `be04c4c`, with every msodbcsql citation checked against the source tree and `cargo test -p mssql-odbc --lib` run locally (694 passed, 1 ignored).

Two things I would block on, both inline: the `SQL_C_TINYINT` signedness change is only half of msodbcsql's rule and regresses `tinyint` fetch, and the describe-then-bind divergence the description asks about does have a downstream consumer.

### Blocking

1. **`SQL_C_TINYINT` is "sign-unknown", not simply signed.** msodbcsql bit-copies it against its same-width counterpart, so `tinyint` 200 into `SQL_C_TINYINT` succeeds there and is `22003` here. Detail and citations inline on `fetch_convert.rs`.
2. **The describe-then-bind narrowing has a downstream consumer.** Answering the question in the description directly: mssql-python binds `SQL_C_DEFAULT` for `None` values under `setinputsizes`, including `SQL_DECIMAL`/`SQL_NUMERIC`. Detail inline on `bind_param.rs`.

### Suggestions

3. NULL `SQL_CHAR`/`SQL_WCHAR` with `ColumnSize` 0 is now `HY104` where it used to execute -- proven mechanically, inline on `param_convert.rs`.
4. **Rebase needed.** `mergeStateStatus` is `DIRTY`; conflicts with `origin/main` on `mssql-odbc/src/conversion/fetch_convert.rs`.

### Nits

5. Dead null-guards in `param_buffer.rs`, inline.
6. Data-at-exec check applies to fixed-width C types, inline.

### On the green cross-repo check

The `mssql-python suite on mssql-odbc driver (cross-repo)` job passing is not evidence against finding 1. `run-mssql-python-odbc-tests.sh` returns 1 on failures and the calling step converts that into `##vso[task.complete result=SucceededWithIssues]`, which ADO reports to GitHub as `success`. Its own comments say so. Worth reading the published test counts rather than the check glyph when a change can move that suite.

mssql-odbc/src/api/bind_param.rs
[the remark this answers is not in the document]
Theekshna: Keeping the bind-time check as a deliberate phase 0 scope decision, but changed the SQLSTATE to `HYC00` and filed AB#47500 for the parked tests — full reasoning in https://github.com/microsoft/mssql-rs/pull/344#discussion_r3832397449

mssql-odbc/src/conversion/fetch_convert.rs
Theekshna: Fixed in both directions — full response in https://github.com/microsoft/mssql-rs/pull/344#discussion_r3832397435

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Fixed, with the unit and e2e cases you asked for — full response in https://github.com/microsoft/mssql-rs/pull/344#discussion_r3832397455

Short version: instead of widening `char(0)` to `char(1)`, the `ColumnSize` check moved to bind time to match `CheckSqlPrecScale`, so it's `HY104` from `SQLBindParameter` rather than `SQLExecute`. Your scratch-test result is now pinned by `explicit_char_null_with_zero_column_size_is_rejected`.

mssql-odbc/src/api/bind_param.rs
Theekshna: Keeping the bind-time check as a deliberate phase 0 scope decision, but changed the SQLSTATE to `HYC00` and filed AB#47500 for the parked tests — full reasoning in https://github.com/microsoft/mssql-rs/pull/344#discussion_r3832397449

On "*they are simply no longer reachable*" — fair hit on the wording, and I've reframed it. The point I'd add: reachable-for-NULL-only isn't a feature worth preserving when no value can be bound to the same type. Declining `c_type_defaulted` mainly because it's scaffolding that outlives its reason; AB#47500 gates the tests on the matrix rows rather than leaving them ungated.

mssql-odbc/src/conversion/fetch_convert.rs
Theekshna: Fixed in both directions — full response in https://github.com/microsoft/mssql-rs/pull/344#discussion_r3832397435

mssql-odbc/src/api/bind_param.rs
Theekshna: Keeping the bind-time check as a deliberate phase 0 scope decision, but changed the SQLSTATE to `HYC00` and filed AB#47500 for the parked tests — full reasoning in https://github.com/microsoft/mssql-rs/pull/344#discussion_r3832397449

mssql-odbc/src/conversion/fetch_convert.rs
Theekshna: Agreed, fixed in both directions. Thanks — I had the `sqlccnvt.cpp:4903` rule backwards.

**Fetch** (`fetch_convert.rs`): `SQL_C_TINYINT` copies the byte through when the source is `ColumnValues::TinyInt`, and narrows signed for every other source. `SQL_C_STINYINT` is never exempt, matching the `sqlccnvt.cpp:5271-5305` arms.

**Parameter**: `type_rules::effective_param_c_type` rewrites `SQL_C_TINYINT` + `SQL_TINYINT` to `SQL_C_UTINYINT` before the buffer is read, mirroring `ParamToSQLType`. It has to happen there because `read_param_value` widens to `i128` and widening forces a sign.

I didn't take the suggested `let byte = narrow!(u8)` unconditionally — that would also stop character and decimal sources rejecting negatives. Scoping the match to a `tinyint` source keeps that `22003` and still satisfies msodbcsql's rule, whose exemption likewise requires the opposite side to be a tinyint flavour.

Both new e2e cases are parity tests, no `SKIP_IF_COMPARING_MSODBCSQL`: `TinyintColumnAbove127FetchesIntoTinyintCType` and `TinyintCTypeIsUnsignedAgainstTinyintParameter`. Unit coverage in `tinyint_c_target_transfers_a_tinyint_byte_unchanged`, `tinyint_c_target_is_signed_for_other_sources`, `tinyint_c_type_reads_unsigned_against_a_tinyint_parameter`; the signed side stays pinned by `signed_tinyint_target_rejects_over_127`.

No separate CHANGELOG `Fixed` entry — `mssql-odbc` has never been released (only `[Unreleased]`, no tags), so "no longer returns 22003" would be a delta from code nobody has run. It's folded into the `Added` entry describing parameter support.

mssql-odbc/src/api/bind_param.rs
Theekshna: Keeping the bind-time rejection as a phase 0 scope decision, but the diagnostic was wrong and I've changed it.

**On the dependency**: `mssql-python` ships exclusively against the C++ `msodbcsql18`, not this driver. `mssql-odbc` has no release — only `[Unreleased]` in the CHANGELOG, no tags. #326 merged the flow; it never shipped. So nothing observable regresses, and I'd rather scope phase 0 explicitly than carry scaffolding.

**On NULLs**: no C type can bind a *value* to `SQL_DECIMAL`, `SQL_NUMERIC`, `SQL_GUID`, `SQL_BINARY`, or any temporal type — the matrix is character and integer families only. Exempting NULLs gives an application a binding that works for `NULL` and fails on its first real value, so a generic data layer gets halfway through a workload before hitting it. `typed_null` handling the type isn't enough on its own; the feature only helps if the value path exists too.

**Where you were right**: `fValidConversion` is a legality table, this one is an implementation-progress list, and `07006` conflated them. The bind-time rejection is now `HYC00` — the state the spec defines for "the driver does not support this conversion" — with a `TODO` on `is_supported_conversion` to switch to `07006` once the table is complete, since a missing entry will then genuinely mean illegal. Also swept the stale `07006` references: module doc, `parameters_plan.md`, 5 unit tests, and 2 e2e tests that assert the SQLSTATE directly.

**AB#47500** filed — missing rows, all four parked tests by file and name (noting that `DISABLED_DescribedDecimalRoundTripsPrecisionAndScale` also guards the scale-0 wire-metadata regression), and the `HYC00` → `07006` switch as a completion criterion. Referenced from all four parking sites and `parameters_plan.md`.

The CHANGELOG entry states the supported set and the NULL rule, so the scope is on the record for the phase 0 notes rather than implied by behaviour.

Worth naming the bigger constraint: character and integer only means we reject plenty msodbcsql accepts — `IntegerToCharacterConversionIsRejected` is `SKIP_IF_COMPARING_MSODBCSQL` for exactly that. That's the phase 0 scope, now written down. Happy to reopen if you think it's too narrow for a first release.

mssql-odbc/docs/typed-columnar-fetch-plan.md
Theekshna: Good catch — that row asserted the thing the code no longer does. Split into two: a character or decimal source above 127 stays `22003` (signedness decides the range check), and a `tinyint` *column* source is bit-copied, citing `sqlcdata.cpp` and `sqlccnvt.cpp:4903`. Both "Resolved" now that the fix landed — see https://github.com/microsoft/mssql-rs/pull/344#discussion_r3832397435

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Right that it was an unmentioned change with no coverage. Fixed by converging harder on `CheckSqlPrec` rather than by widening.

`SQLBindParameter` now validates `ColumnSize` against `ParameterType` up front, in msodbcsql's order (`sqlcdesc.cpp:3024-3038`), so it's `HY104` at bind rather than at execute — the call that's actually wrong. `type_rules::parameter_column_size_is_valid` encodes the `CheckSqlPrecScale` table: 0 is `SQL_PREC_UNLIMITED`, so it means `max` only for `SQL_VARCHAR`/`SQL_WVARCHAR`/`SQL_VARBINARY`/`SQL_SS_XML` and is rejected elsewhere (`sqlcdesc.cpp:11461`, which only clamps for `IS2xAPP`).

I chose reject over `unwrap_or(1)` or treating 0 as `max` because widening puts a declaration on the wire the application never asked for, and `char(1)` vs `char(0)` is a silent guess.

Coverage for all three gaps you named: e2e `ZeroColumnSizeIsRejectedForFixedLengthCharTypes` (table-driven, reject vs accept, NULL bound), and unit `explicit_char_null_with_zero_column_size_is_rejected` / `explicit_varchar_null_with_zero_column_size_is_max` — the explicit path that regressed, which had nothing because both NULL tests had been switched to the variable-length types.

Also added `every_bind_accepted_column_size_is_declarable`, since the bind gate and the declaration encode the same limits twice. It found one asymmetry immediately: `variable_length` widens an oversized `ColumnSize` to `max`, which the gate now rejects first. Left alone — that direction is harmless and the widening is pinned elsewhere.

Called out in the CHANGELOG.

mssql-odbc/src/conversion/param_buffer.rs
Theekshna: Right — those helpers can't see a null pointer or a negative length in production. Noted on both tests, and the dead branches are gone:

```rust
/// Pins the helpers' own length handling. The null-pointer and negative-length
/// cases are not covered because `read_param_value` and `read_indicator` reject
/// both before dispatch; the helpers `debug_assert` that contract.
```

The `is_null()` / `len_spec < 0` branches in `read_integer`, `read_char_bytes`, and `read_wchar_bytes` are now `debug_assert!`s rather than runtime branches, so the contract is stated once and checked in debug rather than being unreachable code that reads like coverage.

mssql-odbc/src/conversion/param_buffer.rs
Theekshna: Behaviour is intended — the special values are checked first for every C type, before the "is this a length" question is asked at all. You're right that the comment read as a contradiction and that nothing pinned the boundary.

Comment now leads with the ordering (*"Past the special values the indicator is a length only for the character types…"*), and `indicator_is_not_a_length_for_fixed_width_c_types` now walks across `SQL_LEN_DATA_AT_EXEC_OFFSET` (−100) instead of stopping at −7:

```rust
for stray in [-7, 0, 999, SQL_LEN_DATA_AT_EXEC_OFFSET + 1] { /* ignored */ }
for at_exec in [SQL_LEN_DATA_AT_EXEC_OFFSET, SQL_LEN_DATA_AT_EXEC_OFFSET - 1] {
    // DataAtExecUnsupported, whatever the C type
}
```

So −99 is ignored on an `SQL_C_SLONG` binding and −100 / −101 are `HYC00`, and which side of the boundary each falls on is now asserted rather than assumed.

David-Engel (approved):
**Review produced using Claude Opus 5 + High Effort + 1M context.**

Approving. Every finding from my earlier review is resolved, and I re-verified each one against the msodbcsql source rather than taking the commit message on faith.

## Findings resolved

**`SQL_C_TINYINT` signedness** -- fixed in both directions, and with a better citation than the one I supplied. `sqlcfunc.cpp:3193-3195` (`ParamToSQLType`: `if (fCType == SQL_C_TINYINT && fSqlType == SQL_TINYINT) fCType = SQL_C_UTINYINT;`) exists verbatim, and `effective_param_c_type` is a literal transcription of it. The fetch side scopes the copy-through to a `ColumnValues::TinyInt` source, so every other source still narrows signed -- which is what the `typed-columnar-fetch-plan.md` row was actually about. The explanation in `parameters_plan.md` for why one rule needs two spellings (fetch can copy the byte, parameters must widen, and widening forces an interpretation) is the clearest account of that code I have read.

The new e2e coverage is the right shape: `TinyintColumnAbove127FetchesIntoTinyintCType` and `TinyintCTypeIsUnsignedAgainstTinyintParameter` carry no `SKIP_IF_COMPARING_MSODBCSQL()`, so parity is asserted against the real driver rather than inferred from reading its source.

**`ColumnSize` 0** -- moved to bind time, after the type and conversion checks, matching msodbcsql's ordering. I walked every arm of `CheckSqlPrecScale` (`sqlcdesc.cpp:11511-11833`) against `parameter_column_size_is_valid` and it is faithful, including the subtle part: `CheckSqlPrec` rejects `SQL_PREC_UNLIMITED == 0` for the fixed-length types, while the Yukon+ `SQL_VARCHAR` / `SQL_WVARCHAR` / `SQL_VARBINARY` / `SQL_XML_MAPPED` arms bound only the upper end. Constants match `tds.h`.

**`07006` -> `HYC00`** -- the right call, and the `TODO` on `is_supported_conversion` recording when the state should flip back is exactly the kind of note that keeps a temporary decision from becoming permanent by accident. Tracking ID threaded into both parked e2e tests and the `#[ignore]`.

**Both nits** -- `debug_assert!` contracts with the helper tests re-scoped to what they actually pin, and the indicator test extended across the `SQL_LEN_DATA_AT_EXEC_OFFSET` boundary.

## Verification

`cargo clippy -p mssql-odbc --all-targets` clean; `cargo test -p mssql-odbc --lib` 751 passed, 0 failed, 1 ignored. Rebase onto current `main` resolved the `fetch_convert.rs` conflict cleanly -- the net diff there is only the intended change. Diff coverage 96%. All checks green.

I also checked the new bind-time `ColumnSize` gate against mssql-python: `MAX_INLINE_CHAR` is 4000 and long strings bind `ColumnSize` 0 with `SQL_VARCHAR` / `SQL_WVARCHAR`, which stay in the "0 means max" group. And a `sql_variant` carrying a `tinyint` surfaces as `ColumnValues::TinyInt`, so the fetch copy-through covers that path too.

## Noted, not blocking

Defaulted bindings of the decimal, binary, GUID and temporal types return `HYC00` at bind until AB#47500 lands, which narrows the describe-then-bind flow #326 added -- mssql-python's `None`-parameter path reaches it. Explicit binds of those types already failed before this PR, so the change is confined to defaulted ones, `HYC00` is the honest state, and the description now argues the restriction rather than leaving it to be discovered. Good to proceed on that basis.

Nice work on the follow-up -- the citations are tighter than what I brought to the first pass.

PR #345, opened 2026-08-20
mssql-odbc: Add descriptor header/record model and field APIs

Vahid-b:
## Description

Implements the descriptor (ARD/APD/IRD/IPD) header/record data model and the `SQLGetDescFieldW`/`SQLSetDescFieldW` field-access entry points, replacing today's diagnostics-only placeholder and always-error stub.

- **Data model** (`handles::desc`): typed `DescHeader` + 1-based `DescRecord` list shared across all four descriptor kinds, mirroring msodbcsql's common header/record bases. `classify_field(kind, field)` is the single source of truth for which `SQL_DESC_*` fields apply to which kind and whether they're writable — shared by both entry points instead of duplicating the field matrix.
- **`SQLGetDescFieldW`**: real implementation (was a stub always returning `SQL_ERROR`). Header + record dispatch, ODBC-correct per-field output widths (verified line-by-line against msodbcsql's `GetADHeaderField`/`GetIPDHeaderField`/`GetIRDHeaderField` and the generic record dispatch — several fields are narrower than a generic `SQLLEN`, e.g. `SQL_DESC_COUNT`/`SQL_DESC_BIND_TYPE`).
- **`SQLSetDescFieldW`**: new export (did not exist before). IRD-is-read-only gate (matches msodbcsql's blanket rejection except the two pointer header fields), `SQL_DESC_COUNT` growth/shrink with kind-appropriate defaults, type/concise-type/datetime-interval-code coupling, and `SQL_C_NUMERIC` precision/scale validation.
- **Hardening from review** (bot + human): fixed a null-pointer crash in `SQL_DESC_NAME` handling that caused a non-unwinding process abort (not a catchable panic); rejected `SQL_DESC_ARRAY_SIZE == 0`; fixed `SQL_DESC_UNNAMED` being wrongly writable on IPD; advertised the two new functions in `SQLGetFunctions`; corrected verbose `SQL_DESC_TYPE = SQL_DATETIME` handling on IPD to match msodbcsql's real fallthrough logic; added `SQL_DESC_PARAMETER_TYPE` validation; added a final `SQL_C_NUMERIC` consistency check on `SQL_DESC_DATA_PTR` set.

**Why this matters now:** I verified against the current `mssql-python` source (`ddbc_bindings.cpp`) that `SQLSetDescFieldW` is one of ~38 function pointers resolved at driver-load time — since it wasn't exported at all, loading `mssql-odbc` under `mssql-python` would throw `"Failed to load required function pointers from driver"` before any query ever ran, not just for `SQL_C_NUMERIC` parameters. This PR implements the exact field sequence `ddbc_bindings.cpp` runs for a `SQL_C_NUMERIC` input parameter (`SQL_DESC_TYPE` → `PRECISION` → `SCALE` → `DATA_PTR` on record 1), added as a named regression test.

**Scope note:** IRD/IPD records here are the driver's own descriptor storage, independent of the existing `column_metadata`/`parameter_metadata` used by `SQLDescribeColW`/`SQLDescribeParam`. Reconciling the two (populating IRD from a live result set, IPD from prepared/parameter metadata) is explicitly [AB#47437](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47437)'s job, not this PR's — until then IRD/IPD exist and behave correctly per the ODBC field-access contract but start out empty, even once a statement is prepared/executed. Catalog-style descriptive fields (`SQL_DESC_LABEL`, `TABLE_NAME`, `TYPE_NAME`, etc.) remain answerable via the already-implemented `SQLColAttributeW` rather than duplicated here.

## Testing

- 752/752 Rust unit tests passing via `cargo nextest run -p mssql-odbc` — defaults per kind, growth/shrink, all 4 kinds' field matrix, pointer fields, character truncation, read-only/IRD rejection, invalid field/record diagnostics, `SQL_C_NUMERIC` precision/scale bounds, type validation, and the exact mssql-python regression sequence.
- Workspace-wide `cargo bclippy` and `cargo bfmt` (frozen): clean.
- New C++ e2e tests (`get_desc_field_test.cpp`, `set_desc_field_test.cpp`), registered in `CMakeLists.txt` and **verified green in CI** (Linux, Linux ARM, Windows, plus the msodbcsql parity-comparison legs) after fixing a few issues only reachable via a live Driver Manager + SQL Server, none of them exercisable from this sandbox alone:
  - A real, Linux/Linux-ARM-only bug: unixODBC's Driver Manager gates `SQLGetDescField` on an IRD to statements past `STATE_S1` ("allocated, not yet prepared") and returns `HY007` itself without ever reaching the driver — verified directly against unixODBC's `DriverManager/SQLGetDescField.c` source. Fixed by preparing the statement first in the affected test, matching how a real application encounters this.
  - Two genuine, intentional parity divergences from msodbcsql, each documented and marked with the existing `SKIP_IF_COMPARING_MSODBCSQL()` mechanism rather than silently accepted: msodbcsql uses `HY021` (not `HY003`) for an unrecognized APD `SQL_DESC_TYPE`, and defers `SQL_C_NUMERIC` precision-range validation to bind time rather than enforcing it eagerly at `SQLSetDescField` time as this driver deliberately does.
  - One test-only bug (not a driver bug): an "unrecognized field" assertion used a descriptor record number that didn't exist yet, so msodbcsql's own record-existence check short-circuited before ever reaching field validation.
- Caught and fixed a real out-of-bounds-read bug in my own test code during this work (an `SQL_NTS`-terminated buffer that wasn't actually NUL-terminated) — `cargo nextest`'s per-test process isolation surfaced it where `cargo test`'s shared-process model didn't, which is exactly why this repo mandates nextest.

## Related Issues

Part of User Story [AB#46374](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46374) ("mssql-odbc | Descriptors"):
- [AB#47297](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47297) (P0 — mssql-python APD compat): https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47297
- [AB#47435](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47435) (P1 — generalize headers/records/field APIs): https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47435

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (`cargo nextest run -p mssql-odbc`: 752/752; full workspace `cargo bclippy`/`cargo bfmt` clean)
- [x] Public API changes are documented (doc comments on new public types/fns; `docs/typed-columnar-fetch-plan.md` scope-boundary note updated to reflect this no longer being deferred)
- [x] C++ e2e tests build and pass in CI (Linux, Linux ARM, Windows), including msodbcsql parity comparison


mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -0,0 +1,1017 @@
...
+/// `SQL_DESC_NAME` write (IPD only — the only kind `classify_field` marks
+/// this field writable for). `buffer_length` is in bytes, or `SQL_NTS` for
+/// NUL-terminated input, matching ODBC's general character-input rule.
+fn set_name(
+    state: &mut DescState,
+    record_number: SqlSmallInt,
+    value_ptr: SqlPointer,
+    buffer_length: SqlInteger,
+) -> SqlReturn {
+    let ptr = value_ptr as *const SqlWChar;
```
Vahid-b, line 587: **Blocking:** `value_ptr` is passed to `read_utf16` (an `unsafe fn` whose safety contract requires a readable pointer) with no null check -- the only `read_utf16` call site in the crate that skips it (`driver_connect.rs`, `connect.rs`, `prepare.rs`, `exec_direct.rs`, `catalog.rs` all null-check first).

I verified this crashes the process, not just "is UB in theory": I ran `SQLSetDescFieldW(ipd_handle, 1, SQL_DESC_NAME, null, SQL_NTS)` against this PR's actual head commit. Result:

```
thread '...' panicked at mssql-odbc\src\api\util.rs:74:19:
null pointer dereference occurred
thread caused non-unwinding panic. aborting.
ABORT - with code 0xc0000409 (stack-based buffer overrun)
```

This is a **non-unwinding abort**, not a normal `panic!()` -- `ffi_entry!`'s `catch_unwind` (the crate's entire safety net) cannot catch an abort. Any application calling this with a null buffer (a defensively-common pattern) takes down the entire host process.

**Suggested fix:** null-check `value_ptr` before calling `read_utf16` and return `HY009` ("Invalid use of null pointer"), matching every other call site's pattern.

mssql-odbc/src/handles/desc.rs
```diff
@@ -63,3 +275,321 @@ impl HasDiagnostics for DescState {
...
+        | SQL_DESC_SCALE => (Record, !is_ird),
+
+        // ---- Record fields specific to application descriptors ----------
+        SQL_DESC_DATA_PTR | SQL_DESC_INDICATOR_PTR | SQL_DESC_OCTET_LENGTH_PTR if is_ad => {
+            (Record, true)
+        }
+
+        // ---- Record fields specific to implementation descriptors -------
+        SQL_DESC_NULLABLE if is_ird || is_ipd => (Record, false),
+        SQL_DESC_NAME | SQL_DESC_UNNAMED if is_ird || is_ipd => (Record, is_ipd),
```
Vahid-b: **Important:** `SQL_DESC_UNNAMED` is derived from `name` on read (per `DescRecord`'s own doc comment: *"derived from name on read... rather than stored redundantly"*), but this classifies it as writable on IPD (`is_ipd` -> `true`). `set_record_field` has no arm for it.

I verified this reaches the `_ => debug_assert!(false, "classify_field allowed an unmapped writable record field")` fallback and **actually panics** in a debug build:
```
thread '...' panicked at ...set_desc_field.rs:306:13:
classify_field allowed an unmapped writable record field
```
In release, `debug_assert!` compiles out and it falls through to `post_diag(HY091); SQL_ERROR` -- the *correct* result, but by accident, not by design.

**Suggested fix:** classify `SQL_DESC_UNNAMED` as `(Record, false)` for both IRD and IPD.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -0,0 +1,1017 @@
...
+    field: SqlUSmallInt,
+    value_ptr: SqlPointer,
+) -> SqlReturn {
+    match field {
+        SQL_DESC_ARRAY_SIZE => {
+            let requested = value_ptr as SqlULen;
+            let Ok(max) = SqlULen::try_from(i32::MAX) else {
+                return SQL_ERROR; // unreachable
+            };
+            if requested > max {
```
Vahid-b, line 234: **Important:** `SQL_DESC_ARRAY_SIZE` must be at least 1 per ODBC, and this driver's own `SQL_ATTR_ROW_ARRAY_SIZE` validation already enforces that. This branch has no lower-bound check.

I ran `SQLSetDescFieldW(apd, 0, SQL_DESC_ARRAY_SIZE, 0, 0)` against this PR's head commit -- returns `SQL_SUCCESS`, storing 0.

**Suggested fix:** reject with `HY024` for `requested == 0`, consistent with the statement-attribute twin. (The bot flagged this same pattern at 4 other locations in this file worth checking too.)

mssql-odbc/src/api/exports.rs
```diff
@@ -996,16 +996,47 @@ pub unsafe extern "C" fn SQLGetDescFieldW(
...
+/// - `descriptor_handle` must be a valid descriptor handle.
+/// - `record_number` must be valid for the descriptor.
+/// - `field_identifier` must be a valid field identifier.
+/// - `value_ptr`, when the field is a pointer field, is stored verbatim and
+///   not dereferenced by this call; the caller must keep it valid for as
+///   long as it remains bound. When the field is a character field,
+///   `value_ptr` must be readable for `buffer_length` bytes (or
+///   NUL-terminated when `buffer_length == SQL_NTS`).
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn SQLSetDescFieldW(
```
Vahid-b, line 1074: **Important:** These two entry points are fully implemented and exported, but I confirmed via a repo-wide search that `get_functions.rs` (not touched by this PR) has no `SQL_API_SQLGETDESCFIELD`/`SQL_API_SQLSETDESCFIELD` entries. Any Driver Manager or client that checks `SQLGetFunctions` before calling (a common, spec-encouraged pattern, distinct from the direct-symbol-resolution path this PR's own `mssql-python`/`ddbc_bindings.cpp` motivating case uses) will treat both APIs as unsupported and skip them despite the symbols existing.

**Suggested fix:** add both function IDs to `odbc_types.rs` and advertise them in both the direct and ODBC3-bitmap query paths in `get_functions.rs`.

Vahid-b (commented):
## Review Summary

Reviewed against `.github/instructions/mssql-odbc.instructions.md`, cross-referenced against `type_rules.rs`/`bind_param.rs` precedent and the ODBC 3.x spec, and empirically verified every finding below by running real code against this PR's head commit (not just static reading) -- three of the seven findings were confirmed by writing and running temporary, reverted-after-use probe tests directly against the branch.

The Copilot bot already left 7 findings here (3 visible, 4 collapsed/"suppressed") with no replies yet. All are addressed below; one (the null-pointer case) turned out to be a confirmed **process-crashing abort**, not just theoretical UB, so I raised it from the bot's original framing to Blocking.

**What's solid:** `set_type` for application descriptors (ARD/APD) correctly applies `canonical_c_type` + `is_valid_c_type` -- the exact fix I'd flagged as missing in PR #322's `bind_col.rs`, correctly applied here. Both new files follow the safe-core/unsafe-shell split correctly. All new SQLSTATE constants verified correct. `cargo fmt`/`clippy` clean, 742/742 unit tests pass (matches the PR's claim exactly). The mssql-python regression test is present and passes.

**Residual risk not covered by an inline comment:** CI is currently red on 4 build legs (Build Linux, Linux ARM, Windows, Windows ARM) -- I could not access the Azure DevOps build logs from this environment to diagnose the cause; worth investigating before this draft is marked ready, independent of the findings below.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -0,0 +1,1017 @@
...
+fn set_record_field(
+    state: &mut DescState,
+    kind: DescKind,
+    record_number: SqlSmallInt,
+    field: SqlUSmallInt,
+    value_ptr: SqlPointer,
+    buffer_length: SqlInteger,
+) -> SqlReturn {
+    match field {
+        SQL_DESC_CONCISE_TYPE | SQL_DESC_TYPE => set_type(state, kind, record_number, value_ptr),
```
Vahid-b: **Important:** `set_type` handles both `SQL_DESC_CONCISE_TYPE` and the verbose `SQL_DESC_TYPE` identically via `classify_parameter_sql_type`, whose `Supported` list contains only the concise `SQL_TYPE_DATE`/`TIME`/`TIMESTAMP` (91-93), not the verbose `SQL_DATETIME` (9).

Per ODBC's documented descriptor-binding sequence, setting the **verbose** `SQL_DESC_TYPE = SQL_DATETIME` followed by `SQL_DESC_DATETIME_INTERVAL_CODE = SQL_CODE_DATE` is the standard way to bind a DATE/TIME/TIMESTAMP subtype through IPD -- it's legal precisely *because* it's the verbose field, unlike the concise field where `9` is genuinely ambiguous with the deprecated 2.x spelling.

I found the test `set_desc_field_ambiguous_legacy_datetime_value_on_ipd_rejected` explicitly asserts `SQL_DESC_TYPE = 9` on IPD returns `SQL_ERROR`/`HY004` -- this pins the wrong behavior as a passing regression test, so a fix needs to update that test too, not just this code.

**Suggested fix:** split the verbose and concise paths, accepting `SQL_DATETIME` on `SQL_DESC_TYPE` and deriving `SQL_DESC_DATETIME_INTERVAL_CODE` from the paired write, per the original suggestion.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -0,0 +1,1017 @@
...
+            write_record_field(state, record_number, |r| r.length = len)
+        }
+        SQL_DESC_OCTET_LENGTH => {
+            let len = value_ptr as SqlLen;
+            write_record_field(state, record_number, |r| r.octet_length = len)
+        }
+        SQL_DESC_PRECISION => set_precision(state, record_number, value_ptr),
+        SQL_DESC_SCALE => set_scale(state, record_number, value_ptr),
+        SQL_DESC_NAME => set_name(state, record_number, value_ptr, buffer_length),
+        SQL_DESC_PARAMETER_TYPE => {
```
Vahid-b: **Important:** This accepts every `i16` as `SQL_DESC_PARAMETER_TYPE` -- `SqlSmallInt::try_from(value_ptr as SqlLen)` only bounds-checks that the value fits an `i16`, not that it's one of the defined `SQL_PARAM_*` values. An arbitrary value such as 9999 is stored and reported back as valid descriptor state.

**Suggested fix:** validate against `SQL_PARAM_INPUT`/`SQL_PARAM_INPUT_OUTPUT`/`SQL_PARAM_OUTPUT` (and the 2.x aliases) and reject anything else with `HY105`.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -0,0 +1,1017 @@
...
+            let Ok(v) = SqlSmallInt::try_from(value_ptr as SqlLen) else {
+                post_diag(state, ERR_INVALID_ATTRIBUTE_VALUE);
+                return SQL_ERROR;
+            };
+            write_record_field(state, record_number, |r| r.parameter_type = v)
+        }
+        SQL_DESC_DATA_PTR => write_record_field(state, record_number, |r| r.data_ptr = value_ptr),
```
Vahid-b: **Suggestion:** msodbcsql performs a final consistency check (`CheckADDescRecConsistency`) when `DATA_PTR` is set, catching e.g. precision 39 set before the type changes to `SQL_C_NUMERIC`. This PR's module doc explicitly defers "opaque... eventual bind/execute consumer" logic to AB#47437, so this may be an intentional scope boundary rather than an oversight -- but msodbcsql validates at set-time, not at bind/execute-time, so it might be worth an explicit tracking note (or a one-line consistency check) rather than silent deferral, so the gap doesn't get lost before AB#47437 picks it up.

mssql-odbc/src/api/set_desc_field.rs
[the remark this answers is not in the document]
Vahid-b: Fixed — `SQL_DESC_ARRAY_SIZE == 0` now returns `SQL_ERROR`/`HY024` before storing, matching this crate's `SQL_ATTR_ROW_ARRAY_SIZE` validation. Added `set_desc_field_array_size_zero_returns_error` as a regression test. (The other 4 locations you flagged turned out to be distinct issues rather than repeats of this one — each is fixed and resolved in its own thread.)

> [!NOTE]

Vahid-b: Fixed — `SQL_DESC_ARRAY_SIZE == 0` now returns `SQL_ERROR`/`HY024` before storing, matching this crate's own `SQL_ATTR_ROW_ARRAY_SIZE` validation. Verified: `SQLSetDescFieldW(apd, 0, SQL_DESC_ARRAY_SIZE, 0, 0)` now correctly fails instead of returning `SQL_SUCCESS`. Added `set_desc_field_array_size_zero_returns_error` as a regression test.

> [!NOTE]

mssql-odbc/src/handles/desc.rs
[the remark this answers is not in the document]
Vahid-b: Fixed — split the `SQL_DESC_NAME | SQL_DESC_UNNAMED` match arm so `SQL_DESC_UNNAMED` is now always `(Record, false)` on both IRD and IPD (it's derived from `name` on read, never stored, so it must never be writable). Added `unnamed_is_never_writable` (classify_field unit test) and `set_desc_field_unnamed_is_never_writable` (end-to-end, confirms `SQL_ERROR`/`HY091` with no panic).

> [!NOTE]

Vahid-b: Fixed — `SQL_DESC_UNNAMED` is now classified `(Record, false)` for both IRD and IPD, separate from `SQL_DESC_NAME`. Confirmed the `debug_assert!` panic you reproduced no longer fires: `SQLSetDescFieldW(ipd, 1, SQL_DESC_UNNAMED, ...)` now cleanly returns `SQL_ERROR`/`HY091` via the normal read-only-field diagnostic path. Added both a `classify_field` unit test and an end-to-end `set_desc_field_unnamed_is_never_writable` test.

> [!NOTE]

mssql-odbc/src/api/exports.rs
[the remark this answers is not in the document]
Vahid-b: Fixed — added `SQL_API_SQLGETDESCFIELD` (1008) and `SQL_API_SQLSETDESCFIELD` (1017) to `odbc_types.rs` (values verified against msodbcsql's own `sql.h`, not guessed) and advertised both in `get_functions.rs`'s `supported_function_ids()`, which covers the direct-id query, the legacy `SQL_API_ALL_FUNCTIONS` array, and the `SQL_API_ODBC3_ALL_FUNCTIONS` bitmap automatically. Added `descriptor_field_functions_report_true` and extended the existing bitmap test.

> [!NOTE]

Vahid-b: Fixed — added `SQL_API_SQLGETDESCFIELD` (1008) and `SQL_API_SQLSETDESCFIELD` (1017) to `odbc_types.rs`, values confirmed against msodbcsql's own `sql.h` rather than assumed, and advertised both in `get_functions.rs`'s `supported_function_ids()` (covers the direct query, legacy array, and ODBC3 bitmap paths together). Added a dedicated `descriptor_field_functions_report_true` test plus bitmap-test coverage.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Vahid-b: Fixed — thank you for actually reproducing this, that's exactly the kind of verification this class of bug needs. Added a null check on `value_ptr` before it reaches `read_utf16`, returning `SQL_ERROR`/`HY009` (`ERR_INVALID_NULL_POINTER`), matching the pattern at every other `read_utf16` call site in the crate. Added `set_desc_field_name_null_value_ptr_returns_error_not_crash` as a regression test.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Vahid-b: Fixed — you're right, and I verified it directly against msodbcsql's `SQLSetDescFieldW` (`sqlcdesc.cpp:1633-1649`): setting the verbose `SQL_DESC_TYPE = SQL_DATETIME` falls through to derive the concise type from the record's current `fSubType`, it doesn't reject it. I'd incorrectly applied the `SQLBindParameter`-`ParameterType` ambiguity rule (where `9` has no paired subtype field to disambiguate it) to the descriptor's `SQL_DESC_TYPE`, which does have one.

Split the handling: `SQL_DESC_TYPE = SQL_DATETIME` now derives the concise type from the currently-stored `SQL_DESC_DATETIME_INTERVAL_CODE` (in either write order), or returns `HY021` if no valid code is stored yet. `SQL_DESC_CONCISE_TYPE` is unaffected and still rejects 9/10/11 as before. Updated `set_desc_field_ambiguous_legacy_datetime_value_on_ipd_rejected` (split into `set_desc_field_concise_type_rejects_legacy_datetime_values` and `set_desc_field_type_legacy_time_timestamp_values_still_rejected`) and added two new tests for the fixed verbose-type behavior.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Vahid-b: Fixed — `SQL_DESC_PARAMETER_TYPE` now validates against `SQL_PARAM_INPUT`/`SQL_PARAM_INPUT_OUTPUT`/`SQL_PARAM_OUTPUT` and rejects anything else with the new `HY105` diagnostic (`ERR_INVALID_PARAMETER_TYPE`), as suggested. Added `set_desc_field_parameter_type_accepts_valid_values_and_rejects_others` covering all three valid values plus the `9999` rejection case.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Vahid-b: Took the one-line-check option — added a final `SQL_C_NUMERIC` precision/scale re-validation when `SQL_DESC_DATA_PTR` is set (`set_data_ptr`), matching msodbcsql's `CheckADDescRecConsistency` timing at bind time. This catches an out-of-range precision/scale that was set while the type was still something else, which `set_precision`/`set_scale` alone can't (they only validate against the type stored at the moment each is written). Added `set_desc_field_data_ptr_catches_precision_set_before_type_became_numeric` reproducing exactly the precision-39-then-retype scenario you described. Good catch — this was a real gap, not an intentional part of the AB#47437 deferral.

> [!NOTE]

mssql-odbc/src/handles/desc.rs
```diff
@@ -63,3 +275,343 @@ impl HasDiagnostics for DescState {
...
+            (Record, true)
+        }
+
+        // ---- Record fields specific to implementation descriptors -------
+        SQL_DESC_NULLABLE if is_ird || is_ipd => (Record, false),
+        SQL_DESC_NAME if is_ird || is_ipd => (Record, is_ipd),
+        // SQL_DESC_UNNAMED is always derived from `name` on read (see
+        // `DescRecord::name`'s doc comment) — never stored, so never
+        // writable, unlike SQL_DESC_NAME itself.
+        SQL_DESC_UNNAMED if is_ird || is_ipd => (Record, false),
```
David-Engel: **Blocking:** the earlier review thread was right that `SQL_DESC_UNNAMED` shouldn't fall through to `set_record_field`'s unmapped-field `debug_assert!`, but the fix overshot. Both the spec and msodbcsql make this field R/W on an IPD:

- ODBC `SQLSetDescField` reference, Record Fields: *"An application can set the SQL_DESC_UNNAMED field of an IPD to SQL_UNNAMED. A driver returns SQLSTATE HY091 (Invalid descriptor field identifier) if an application attempts to set the SQL_DESC_UNNAMED field of an IPD to SQL_NAMED."* The R/W table row reads `IRD: R  IPD: R/W`.
- msodbcsql `SetIPDField` (`sqlcdesc.cpp:4873`): `SQL_UNNAMED` clears `cchParamName`/`dwParamNameOffset`, anything else returns `IDS_S1_092`. (`SetADField:4280` is HY091 for AD, which this PR already matches.)

As written, `classify_field` returns `(Record, false)` for IPD, so `!access.writable` rejects **every** value with HY091 — including `SQL_UNNAMED`, the one value that must succeed. `set_desc_field_unnamed_is_never_writable` encodes the wrong behavior: it passes `1isize`, which *is* `SQL_UNNAMED`.

Smallest correct fix — keep it writable on IPD and give it an arm that clears the name (which is what the derived-on-read model already wants):

```rust
SQL_DESC_UNNAMED if is_ird || is_ipd => (Record, is_ipd),
```

```rust
// set_record_field
SQL_DESC_UNNAMED => set_unnamed(state, record_number, value_ptr),
```

```rust
/// `SQL_DESC_UNNAMED` write (IPD only). Only `SQL_UNNAMED` is accepted; it
/// clears the parameter name, from which the field is derived on read.
fn set_unnamed(state: &mut DescState, record_number: SqlSmallInt, value_ptr: SqlPointer) -> SqlReturn {
    if value_ptr as SqlLen != SQL_UNNAMED {
        post_diag(state, ERR_INVALID_DESCRIPTOR_FIELD); // HY091 per the ODBC reference
        return SQL_ERROR;
    }
    write_record_field(state, record_number, |r| r.name.clear())
}
```

Then flip the two regression tests (`unnamed_is_never_writable`, `set_desc_field_unnamed_is_never_writable`) to assert `SQL_UNNAMED` -> `SQL_SUCCESS` with `SQL_DESC_NAME` now empty, and `SQL_NAMED` -> HY091.

mssql-odbc/src/handles/desc.rs
```diff
@@ -1,9 +1,46 @@
...
+//! One shared record shape (`DescRecord`) and header shape (`DescHeader`)
+//! serve all four descriptor kinds, mirroring msodbcsql's common
+//! `GENDESCTAG` header / `cpbaseTag` record bases plus kind-specific
+//! validation in `SQLGetDescFieldW`/`SQLSetDescFieldW`
+//! (`Sql/Ntdbms/sqlncli/odbc/sqlcdesc.cpp`) rather than four parallel record
+//! types. [`classify_field`] is that validation, collapsed into one table so
+//! the get/set entry points share a single source of truth for which fields
+//! apply to which kind.
+//!
+//! Scope note: IRD/IPD records here are the driver's own descriptor storage,
```
David-Engel, line 15: **Suggestion:** the scope note covers IRD/IPD *record* population and the catalog-style fields, but not the four **header** fields that ODBC defines as aliases of statement attributes rather than as separate storage: `SQL_DESC_ARRAY_SIZE` <-> `SQL_ATTR_ROW_ARRAY_SIZE`/`SQL_ATTR_PARAMSET_SIZE`, `SQL_DESC_BIND_TYPE` <-> `SQL_ATTR_ROW_BIND_TYPE`, `SQL_DESC_ARRAY_STATUS_PTR` <-> `SQL_ATTR_ROW_STATUS_PTR`, `SQL_DESC_ROWS_PROCESSED_PTR` <-> `SQL_ATTR_ROWS_FETCHED_PTR`.

Verified: nothing outside `get_desc_field.rs`/`set_desc_field.rs` reads `DescHeader`, and `StmtState::row_array_size`/`row_bind_type`/`row_status_ptr`/`rows_fetched_ptr` are only touched in `set_stmt_attr.rs`. So this two-call sequence returns a wrong answer with this PR's code:

```c
SQLSetStmtAttrW(stmt, SQL_ATTR_ROW_ARRAY_SIZE, (SQLPOINTER)10, 0);
SQLGetDescFieldW(ard, 0, SQL_DESC_ARRAY_SIZE, &n, 0, NULL);  // n == 1, spec says 10
```

Neither copy is consumed yet (no block-fetch reader on main), so nothing executes wrongly today — but `IrdAllowsRowsProcessedPtr` asserts a successful write to a pointer that will never be populated, and once the `SQLFetchScroll` block-fetch consumer lands reading `StmtState`, the descriptor copy diverges silently with no diagnostic.

Ask is just documentation + tracking, not implementation: extend this scope note to name these four header fields explicitly and attach them to AB#47437 (or whichever item owns the aliasing), so it can't get lost between the two PRs.

mssql-odbc/src/api/set_stmt_attr.rs
```diff
@@ -263,6 +263,32 @@ fn sql_get_stmt_attr_w_safe(
...
     value_ptr: SqlPointer,
 ) -> SqlReturn {
+    // The four implicit descriptor handles are set once in StmtHandle::new
+    // and never reassigned (SQLSetStmtAttrW treats SQL_ATTR_APP_ROW_DESC /
+    // APP_PARAM_DESC as a no-op, matching this driver's forward-only,
+    // implicit-descriptor-only design), so answering them doesn't need the
+    // statement mutex at all — return them directly and keep the lock scoped
+    // to the mutable statement attributes below.
+    match attribute {
+        SQL_ATTR_APP_ROW_DESC => unsafe {
```
David-Engel: **Suggestion:** two things about this early return.

1. It skips `free_errors(&mut state)`. ODBC clears a handle's diagnostic records at the start of every call except `SQLGetDiagRec`/`SQLGetDiagField`, so after a failed `SQLExecDirect`, a `SQLGetStmtAttr(SQL_ATTR_APP_PARAM_DESC)` followed by `SQLGetDiagRec` now still reports the stale error. Cheapest fix is to take the lock, `free_errors`, then answer from `stmt.ard`/etc. inside the same block — the fields are outside `inner` either way, so the lock is only there for the diagnostics reset.

2. I can't reconstruct the stated deadlock. The comment says *"That call path already holds the statement mutex inside the Driver Manager, so taking our own stmt lock here ... can deadlock"* — a DM-internal mutex and a driver-private `Mutex` can't form a cycle on their own; a deadlock needs the driver to re-enter its own `stmt.inner`, and nothing calls `sql_get_stmt_attr_w_safe` while holding it. If you have the hung stack, worth putting the real mechanism in the comment. If the actual cause turned out to be something else (the same commit also dropped a line from `get_desc_field_test.cpp`), this comment will misdirect the next person debugging a hang, and any genuine re-entrancy would still be live for every other attribute below.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -0,0 +1,1326 @@
...
+        unsafe { read_utf16(ptr, char_len) }
+    };
+    write_record_field(state, record_number, |r| r.name = name)
+}
+
+/// `SQL_DESC_PARAMETER_TYPE` write (IPD only). Validated against the three
+/// ODBC 3.x `SQL_PARAM_*` values this driver can act on — an arbitrary
+/// `SqlSmallInt` that merely fits the wire width (e.g. `9999`) is not a
+/// real parameter type and must not be stored as though it were.
+fn set_parameter_type(
```
David-Engel, line 636: Nit: HY105 is right per the ODBC reference (*"HY105 Invalid parameter type — (DM) The value specified for the SQL_DESC_PARAMETER_TYPE field was invalid"*), so no objection to the SQLSTATE. Two undocumented divergences from msodbcsql though:

- msodbcsql returns HY092 (`IDS_S1_092`, `sqlcdesc.cpp:4869`) for the same case.
- msodbcsql also accepts `SQL_PARAM_INPUT_OUTPUT_STREAM` and `SQL_PARAM_OUTPUT_STREAM`, both of which the ODBC reference lists as valid values for this field.

Rejecting the stream values is defensible for a driver with no streamed-parameter support — it just isn't called out the way the two divergences in the e2e tests are. No e2e test covers this, so no parity leg is at risk; one comment line would keep a future parity test from being written against the wrong expectation.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -0,0 +1,1326 @@
...
+
+/// Header fields other than `SQL_DESC_COUNT` (handled by the caller since it
+/// needs the record list, not just the header).
+fn set_header_field(
+    state: &mut DescState,
+    field: SqlUSmallInt,
+    value_ptr: SqlPointer,
+) -> SqlReturn {
+    match field {
+        SQL_DESC_ARRAY_SIZE => {
```
David-Engel, line 218: Nit: same category as the `SQL_DESC_PARAMETER_TYPE` note — msodbcsql stores `SQL_DESC_ARRAY_SIZE` unvalidated and only clamps the upper bound (`sqlcdesc.cpp:4164`), so `0` succeeds there. Rejecting is the better behavior and it matches this crate's own `SQL_ATTR_ROW_ARRAY_SIZE` validation; worth one line noting it as a deliberate divergence so nobody later "fixes" it toward msodbcsql.

David-Engel (commented):
## Summary

Reviewed against msodbcsql's `sqlcdesc.cpp`, the ODBC `SQLSetDescField` reference, and the current crate state. Adds a shared descriptor header/record model (`handles::desc`) with a single `classify_field` field matrix, implements `SQLGetDescFieldW` (was a stub returning `SQL_ERROR`) and exports `SQLSetDescFieldW` (didn't exist). The model and the per-field output widths check out line-by-line against msodbcsql's `GetADHeaderField`/`SetADField`/`SetIPDField`; CI is green and diff coverage is 93%.

Two substantive items, both inline: one behavior that regressed while addressing earlier review feedback (`SQL_DESC_UNNAMED`), and one spec-mandated aliasing that is silently absent (descriptor header fields vs. the equivalent statement attributes). The rest are nits, mostly "this is a deliberate msodbcsql divergence, please mark it as one".

**Blocking**
1. `SQL_DESC_UNNAMED` must be writable to `SQL_UNNAMED` on an IPD — `handles/desc.rs`

**Suggestions**
2. Descriptor header fields are duplicate, unaliased storage vs. the statement attributes — `handles/desc.rs`
3. `SQLGetStmtAttrW` no longer clears diagnostics for the four descriptor attributes; the deadlock rationale doesn't hold up — `api/set_stmt_attr.rs`

**Nits**
4. `SQL_DESC_PARAMETER_TYPE` divergences aren't marked — `api/set_desc_field.rs`
5. `SQL_DESC_ARRAY_SIZE == 0` -> HY024 is also an msodbcsql divergence — `api/set_desc_field.rs`
6. `set_data_ptr` runs the consistency check on unbind — `api/set_desc_field.rs`

Everything else I checked came back clean: `read_utf16`/`copy_with_nul` call sites are null- and length-guarded, the `SQL_DESC_COUNT`/`BIND_TYPE`/`ARRAY_SIZE` output widths match `GetADHeaderField` exactly, the IRD gate ordering matches msodbcsql's literal order, descriptor handles are owned and freed by `StmtHandle::drop`, and pointer fields are stored opaquely without dereference. Nice work on the null-`SQL_DESC_NAME` abort catch and on marking the two known parity divergences with `SKIP_IF_COMPARING_MSODBCSQL()` rather than quietly accepting them.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -0,0 +1,1326 @@
...
+/// `SQL_DESC_DATA_PTR` write (AD only). Re-validates the record's
+/// `SQL_C_NUMERIC` precision/scale against whatever is currently stored
+/// before accepting the pointer, matching msodbcsql's final
+/// `CheckADDescRecConsistency` pass at bind time (`sqlcdesc.cpp:11380-11394`).
+/// `set_precision`/`set_scale` only validate against the type stored *at the
+/// time each is written*, so an out-of-range precision set before the type
+/// changes to `SQL_C_NUMERIC` would otherwise go unnoticed; `DATA_PTR` is
+/// the last field mssql-python's binding sequence sets, so this is the
+/// natural point to catch it without deferring all the way to execute time.
+fn set_data_ptr(
```
David-Engel, line 542: Nit: msodbcsql's `CheckADDescRecConsistency` returns early when the record isn't bound:

```cpp
if (pBindInfo->rgbValue == NOT_BOUND)
    return SQL_SUCCESS;
```

Here a null `value_ptr` — the ODBC way to unbind a record (`SQLSetDescField` reference: *"the column ... is unbound if ... the SQL_DESC_DATA_PTR field ... is set by a call to SQLSetDescField ... to a null pointer"*) — still runs the numeric precision/scale check, so unbinding a record that has a stale out-of-range precision returns HY094 instead of succeeding. One `if value_ptr.is_null()` guard before the check.

mssql-odbc/src/handles/desc.rs
Vahid-b: Confirmed — verified directly against `SetIPDField` (`sqlcdesc.cpp:4873-4884`): `SQL_UNNAMED` clears `cchParamName`/`dwParamNameOffset` and succeeds; anything else returns `IDS_S1_092` (HY091 in this crate's terms). The earlier fix for the other UNNAMED bug did overshoot.

Fixed with your suggested shape: `classify_field` now returns `(Record, is_ipd)` for `SQL_DESC_UNNAMED`, added `set_unnamed()` in `set_desc_field.rs` (rejects anything but `SQL_UNNAMED` with HY091, otherwise clears `name`), and flipped both regression tests (`unnamed_is_writable_only_on_ipd` in `desc.rs`, `set_desc_field_unnamed_accepts_only_sql_unnamed` in `set_desc_field.rs`) to assert `SQL_UNNAMED` succeeds and `SQL_NAMED` is rejected. 753/753 tests passing. Pushed in `2c6b9d18`.

> [!NOTE]

mssql-odbc/src/handles/desc.rs
Vahid-b: Good catch — confirmed `DescHeader`'s `array_size`/`bind_type`/`array_status_ptr`/`rows_processed_ptr` are indeed separate storage from `StmtState`'s `row_array_size`/`row_bind_type`/`row_status_ptr`/`rows_fetched_ptr`, with nothing reconciling the two yet. Added a paragraph to the module scope note naming all four fields and their statement-attribute counterparts explicitly, and tracked it under AB#47437 alongside the IRD/IPD record-population note it's adjacent to. Pushed in `2c6b9d18`.

> [!NOTE]

mssql-odbc/src/api/set_stmt_attr.rs
Vahid-b: Both points confirmed and fixed. On (1): you're right, the early-return path skipped `free_errors`, so a stale diagnostic from an earlier failed call would survive a `SQLGetStmtAttrW(SQL_ATTR_APP_PARAM_DESC)` call. On (2): I can't substantiate the deadlock either, and I have a concrete reason to believe it wasn't the real mechanism — a separate investigation earlier this session traced the actual Linux/Linux-ARM-only e2e failure that prompted this change to an unrelated cause entirely: unixODBC's Driver Manager gates `SQLGetDescField` on an IRD to statements past `STATE_S1` (verified against unixODBC's own `SQLGetDescField.c`), which has nothing to do with this crate's own locking.

Reverted the split: merged the four descriptor-attribute branches back into the single locked match (so `free_errors` always runs first), replaced the deadlock comment, and added a regression test (`get_descriptor_attribute_clears_stale_diagnostics`) that posts a diagnostic via an unrecognized attribute, queries `SQL_ATTR_APP_PARAM_DESC`, and asserts the diagnostic list is now empty. 753/753 passing. Pushed in `2c6b9d18`.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Vahid-b: Confirmed against `SetIPDField` (`sqlcdesc.cpp:4855-4870`) — msodbcsql does use `IDS_S1_092` (HY092) here, and does accept `SQL_PARAM_INPUT_OUTPUT_STREAM`/`SQL_PARAM_OUTPUT_STREAM`. Added a doc comment on `set_parameter_type` naming both divergences explicitly (with the msodbcsql line citation) and noting neither is covered by an e2e parity test today. Kept the behavior as-is per your note that HY105 is the more specific/correct SQLSTATE. Pushed in `2c6b9d18`.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Vahid-b: Confirmed against `sqlcdesc.cpp:4161-4167` — msodbcsql stores `dwArraySize` unvalidated (including 0) and only clamps the upper bound. Added a doc comment on the `SQL_DESC_ARRAY_SIZE == 0` check calling this out explicitly as a deliberate divergence (matching this crate's own `SQL_ATTR_ROW_ARRAY_SIZE` validation) so it doesn't get "corrected" back toward msodbcsql later. Pushed in `2c6b9d18`.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Vahid-b: Traced this all the way through before changing anything, and I think the current behavior is actually right — happy to be corrected if I'm missing something.

`CheckADDescRecConsistency`'s `NOT_BOUND` early return is real, but I don't think it's reachable from `SetADField`'s `SQL_DESC_DATA_PTR` case. Looking at that case directly (`sqlcdesc.cpp:4219-4235`):

```cpp
case SQL_DESC_DATA_PTR:
{
    PTR rgbValue = lpbindinfo->rgbValue;
    lpbindinfo->rgbValue = (BYTE*)lpbindinfo;  //  Force binding
    if ( wError = CheckADDescRecConsistency ( pAD, lpbindinfo ) )
        retcode = SQL_ERROR;
    if (SUCCESS(retcode))
        lpbindinfo->rgbValue = (BYTE*)rgbDesc;
    else    //  Restore old binding
        lpbindinfo->rgbValue = (BYTE*)rgbValue;
    ...
}
```

`NOT_BOUND` is `#define`d as `NULL` (`sqlsrv.h:286`). Before calling the consistency check, this forces `rgbValue` to a non-null sentinel (`lpbindinfo` itself) *regardless of what `rgbDesc` (the actual requested new value) is* — so `rgbValue == NOT_BOUND` can never be true inside this call, and the `NOT_BOUND` shortcut never fires here. The check runs unconditionally, and only after it passes does `rgbValue` get set to the real `rgbDesc` — which could be null if the app is unbinding. So an unbind with a stale invalid precision/scale does fail with HY094 in msodbcsql too, not just here.

The `NOT_BOUND` shortcut is real, but it's only reachable from the *other* caller: `CheckADDescConsistency`, the whole-descriptor sweep over `pAD->lpplbindinfo` at bind/execute time, which reads each record's actual current `rgbValue` (no forcing) — that's where an already-unbound record gets skipped, not at the point of the `SQLSetDescField(SQL_DESC_DATA_PTR)` call itself.

So this crate's unconditional re-validation on every `DATA_PTR` write already matches msodbcsql's real behavior for this call path — added the citation and reasoning as a doc comment instead of changing the behavior. If you have a repro or a different code path in mind, I'd like to see it — happy to revisit.

> [!NOTE]

mssql-odbc/src/handles/desc.rs
```diff
@@ -1,9 +1,62 @@
...
+//! as *aliases* of statement attributes rather than as independent storage:
+//! `SQL_DESC_ARRAY_SIZE` (`SQL_ATTR_ROW_ARRAY_SIZE` / `PARAMSET_SIZE`),
+//! `SQL_DESC_BIND_TYPE` (`SQL_ATTR_ROW_BIND_TYPE`), `SQL_DESC_ARRAY_STATUS_PTR`
+//! (`SQL_ATTR_ROW_STATUS_PTR`), and `SQL_DESC_ROWS_PROCESSED_PTR`
+//! (`SQL_ATTR_ROWS_FETCHED_PTR`). `DescHeader` stores these independently of
+//! `StmtState`'s equivalent fields (`set_stmt_attr.rs`), so a
+//! `SQLSetStmtAttrW`/`SQLGetDescFieldW` pair (or the reverse) on the same
+//! logical value currently sees two unaliased copies. Nothing reads
+//! `DescHeader`'s copies outside `get_desc_field.rs`/`set_desc_field.rs`
+//! today, so this is silent rather than wrong yet — it needs wiring the two
```
David-Engel: **Suggestion (new):** thanks for adding this note — but the rebase onto a main that now contains #322 (`SQLBindCol` + block `SQLFetchScroll`) has invalidated the "yet".

`fetch_scroll.rs:164-232` now reads `StmtState`'s copies of `row_array_size` / `rows_fetched_ptr` / `row_status_ptr` / `row_bind_type`, and `SQLBindCol` stores its bindings in `StmtState` with zero references to `DescHandle`. I probed it on this head rather than assume:

```
after SQLBindCol(col 1), ARD SQL_DESC_COUNT = 0            // spec: 1
ARD rec1 SQL_DESC_DATA_PTR rc = 100 (SQL_NO_DATA)          // spec: the bound buffer
after SQLSetStmtAttr(ROW_ARRAY_SIZE=10), ARD SQL_DESC_ARRAY_SIZE = 1   // spec: 10
```

The converse holds too: `SQLSetDescFieldW(ard, 1, SQL_DESC_DATA_PTR, buf)` — ODBC's documented alternative to `SQLBindCol` — returns `SQL_SUCCESS` and `SQLFetchScroll` then writes nothing to `buf`.

Two edits, both to this note:

1. Drop "yet" / "silent rather than wrong" — with #322 in the base the divergence is observable today, not pending a future consumer.
2. The note lists only the four header fields. Add the **ARD record ↔ `SQLBindCol`** direction (`SQL_DESC_COUNT`, `SQL_DESC_DATA_PTR`, `SQL_DESC_INDICATOR_PTR`, `SQL_DESC_OCTET_LENGTH_PTR`), which is the larger half of the same gap and isn't mentioned anywhere right now. Same AB#47437.

Still not asking for the wiring in this PR — just for the note to describe the current state accurately so AB#47437 inherits the full picture.

David-Engel (commented):
## Re-review

All six items from my first pass are addressed. Verified on this head locally: `cargo nextest run -p mssql-odbc` 784/784 pass, `cargo fmt --check` and `cargo clippy -p mssql-odbc --all-targets` clean.

| # | Item | Outcome |
|---|---|---|
| 1 | `SQL_DESC_UNNAMED` writable on IPD | **Fixed correctly** |
| 2 | Header-field aliasing | Documented — note now stale after the rebase, see below |
| 3 | `free_errors` / deadlock rationale | **Fixed correctly** |
| 4 | `SQL_DESC_PARAMETER_TYPE` divergences | Documented |
| 5 | `SQL_DESC_ARRAY_SIZE == 0` divergence | Documented |
| 6 | `set_data_ptr` on unbind | **My finding was wrong — retracted** |

**1** landed as suggested: `classify_field` is `(Record, is_ipd)`, `set_unnamed` accepts only `SQL_UNNAMED` (clearing the name) and returns HY091 otherwise, and both regression tests were inverted to assert the real contract rather than the old one.

**3** went further than I asked — reverting the early return entirely, so all four descriptor attributes go back through the locked path, plus `get_descriptor_attribute_clears_stale_diagnostics` as a regression test. Cleaner than reordering.

**6 — I was wrong, and your pushback is right.** I went and read `sqlcdesc.cpp:4219-4235`, and `SetADField`'s `SQL_DESC_DATA_PTR` case does exactly what the new doc comment says:

```cpp
lpbindinfo->rgbValue = (BYTE*)lpbindinfo;  //  Force binding
if ( wError = CheckADDescRecConsistency ( pAD, lpbindinfo ) )
```

`NOT_BOUND` is `NULL` (`sqlsrv.h:286`), so forcing a non-null sentinel defeats the early return deliberately — msodbcsql validates precision/scale on unbind too, and the only caller that actually reaches the `NOT_BOUND` shortcut is the execute-time sweep `CheckADDescConsistency` (`sqlcdesc.cpp:11351`), as you said. Current behavior matches the reference driver; the doc comment you added is a better artifact than the change I asked for. Apologies for the noise.

One new item below, purely from the rebase onto #322 — documentation accuracy, no implementation asked for in this PR.

### Before merge

CI on this head is still pending (only `CodeQL` / `Analyze` / `check` / `cla` have reported). Worth waiting for the Linux e2e leg specifically: the `set_stmt_attr.rs` revert removed the change commit `3d2884b6` was written to work around, so that leg is what confirms the original "deadlock" was a misdiagnosis rather than a real hang.

mssql-odbc/src/handles/desc.rs
Vahid-b: Verified — you're right on both counts. Confirmed directly in `fetch_scroll.rs`: `fetch_scroll_safe` reads `stmt_state.row_array_size`/`rows_fetched_ptr`/`row_status_ptr`/`row_bind_type` exclusively, with no reference to the ARD anywhere. And in `bind_col.rs`: `SQLBindCol` stores each binding in `StmtState::bindings` (`ColumnBinding`), also with zero reference to `DescHandle`. So yes, the divergence is live and observable today, not "someday."

Updated the note: dropped the "yet"/"silent rather than wrong" framing, and added the ARD-record ↔ `SQLBindCol` direction alongside the header-field one already there — `SQL_DESC_COUNT`/`TYPE`/`DATA_PTR`/`OCTET_LENGTH`/`INDICATOR_PTR` via `ColumnBinding`'s `target_type`/`target_value_ptr`/`buffer_length`/`strlen_or_ind_ptr` (a couple more fields than you listed — `target_type`/`buffer_length` duplicate `SQL_DESC_TYPE` and `SQL_DESC_OCTET_LENGTH` too, so included those for completeness). Still not wiring it in this PR, just making sure AB#47437 inherits the full picture. Pushed in `ca6cc15b`.

> [!NOTE]

David-Engel (approved):
Approving. The scope-note correction in `ca6cc15b` is the last item from my re-review, and it now describes both directions of the gap accurately — I checked the cited names against `handles/stmt.rs:43` (`ColumnBinding`) and `stmt.rs:182` (`StmtState::bindings`), and they're real, not paraphrased. Keeping the wiring itself out of this PR and under AB#47437 is the right call.

Everything from my two passes is closed:

- `SQL_DESC_UNNAMED` is writable to `SQL_UNNAMED` on IPD, HY091 otherwise, matching both the ODBC reference and `SetIPDField`.
- The diagnostics reset is back on the locked path, with `get_descriptor_attribute_clears_stale_diagnostics` guarding it.
- The three deliberate msodbcsql divergences (`SQL_DESC_PARAMETER_TYPE` SQLSTATE + stream values, `SQL_DESC_ARRAY_SIZE == 0`) are documented in-code where the next person will find them.
- My `set_data_ptr`-on-unbind finding was wrong and you rebutted it with the right citation.

Gates on this head: all 19 CI checks green, including both Linux e2e legs, the msodbcsql parity comparison, and the cross-repo mssql-python suite — which also settles the open question from `3d2884b6`, since reverting the lock-free path did not reintroduce a hang. Coverage bot reports 93% diff / 92.3% overall. Locally: 784/784 `cargo nextest run -p mssql-odbc`, `cargo fmt --check` and `cargo clippy -p mssql-odbc --all-targets` clean.

Good work on this one — the per-field verification against `sqlcdesc.cpp`, marking the parity divergences with `SKIP_IF_COMPARING_MSODBCSQL()` instead of quietly accepting them, and pushing back with evidence where I had it wrong is exactly how this should go.

PR #348, opened 2026-08-21
Implement ODBC connection and statement attributes (AB#46377)

saurabh500:
## Summary

Implements the ODBC attribute parity required by [AB#46377](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46377) so `microsoft/mssql-python` can move from its bundled driver to `mssql-odbc`.

- Adds standard environment, connection, statement, descriptor, and SQL Server vendor attribute handling.
- Keeps reusable pre-connect configuration separate from settings negotiated by an active connection.
- Returns deterministic ODBC diagnostics for unsupported identifiers, values, and connection states instead of crashing or silently accepting ineffective settings.
- Adds Rust coverage plus Driver Manager parity coverage against `msodbcsql18`.
- Records the implementation split, measured behavior, and remaining work in [`mssql-odbc/docs/attributes_plan.md`](mssql-odbc/docs/attributes_plan.md).

### Environment

- `SQL_ATTR_ODBC_VERSION` accepts the ODBC 3.x values used by `mssql-python`, including `SQL_OV_ODBC3_80`.

### Connection

- Timeouts: login and connection.
- Session settings: current catalog, transaction isolation, autocommit, access mode, packet size, connection-dead reporting, and reset-connection lifecycle.
- Pre-connect empty `SQL_ATTR_CURRENT_CATALOG` now preserves server-default database selection; the same empty value on a live connection returns `HY024`.
- SQL Server vendor settings used through `attrs_before`, including access tokens, encryption, trust-server-certificate, authentication, MARS, column encryption, hostname-in-certificate, and related string settings.
- Getters report caller overrides before connect and effective negotiated values after connect where the setting can change during connection setup.

### Statement and descriptors

- Query timeout and the statement attributes needed by the Python binding's execute/fetch paths.
- Correct statement-option fan-out for the supported ODBC 2.x connection-level query-timeout option.
- Row/parameter binding state, including live bind-offset pointers and data-at-execution tokens.
- Explicit descriptor ownership and `SQL_ATTR_APP_PARAM_DESC` lifecycle.
- Query-notification message text and options accept null pointers as clear operations, matching `msodbcsql18`.
- `SQL_ATTR_METADATA_ID = SQL_FALSE` is accepted and readable. `SQL_TRUE` returns `HYC00` without mutating state until identifier-mode catalog semantics are implemented.

## Defensive behavior

- Unknown standard attribute identifiers return `HY092`.
- Recognized but unimplemented optional features return `HYC00`.
- Invalid values return the attribute-specific diagnostic (`HY024` where required).
- Pointer-valued attributes validate alignment and preserve the application's pointer lifetime semantics.
- Values that can be changed both by caller configuration and connection negotiation use separate reusable and per-connection state.

## Known divergence and follow-up

- `msodbcsql18` implements `SQL_ATTR_METADATA_ID = SQL_TRUE`; this PR deliberately returns `HYC00` rather than claiming support while catalog RPCs still use pattern semantics. Full catalog functions and metadata identifier mode are tracked by [AB#47526](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47526).
- The same follow-up tracks the larger unrestricted pass-through surfaces from [AB#46377](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46377): `SQLGetInfo`, catalog APIs, ODBC escape translation, and `SQLGetTypeInfo`.
- `SQL_ATTR_QUERY_TIMEOUT` is recognized through the legacy connection-option fan-out, but unsupported statement options in that numeric band return `HYC00`.

## Validation

- `cargo bfmt` and the repository formatting scripts pass.
- `cargo bclippy` and the repository Clippy scripts pass.
- `cargo nextest run -p mssql-odbc --all-targets`: **1,035 passed, 1 skipped**.
- Coverage-instrumented `mssql-odbc` run: **1,035 passed, 1 skipped**.
- Review-fix executable lines: **65/65 covered (100%)**. The preceding PR head reported **99% diff coverage** in CI.
- The C++ attribute parity target contains **65 tests** and compiles with MSVC 19.44. The previous PR head completed its cross-driver parity run; a fresh run for commit `15afe20d` is pending in CI.
- Native behavior for metadata ID, pre-connect empty catalog, and null query-notification pointers was checked against the `msodbcsql18` source in `C:\work\msodbcsql`; the metadata-true case remains an intentional, tested divergence.

Full workspace `cargo btest` reaches live `mssql-tds` tests that require `SQL_PASSWORD`, which is not available in the local shell. The complete `mssql-odbc` package suite above is green.

## Checklist

- [x] Plan and ADO subtasks documented
- [x] Standard and vendor attribute implementation complete for this PR
- [x] Rust package tests pass
- [x] Format and Clippy pass
- [x] New executable lines exceed the 90% coverage target
- [x] Driver Manager parity test binary compiles locally
- [ ] Required CI passes on the latest commit

[AB#46377](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46377)


mssql-odbc/src/api/current_catalog.rs
```diff
@@ -0,0 +1,559 @@
...
+    let Ok(mut state) = dbc.inner.lock() else {
+        error!("{SET_OP}: dbc mutex poisoned");
+        return Err(SQL_ERROR);
+    };
+
+    if state.connection_state != ConnectionState::Connected {
+        state.current_catalog = Some(name.to_string());
+        debug!(name, "{SET_OP}: stored for next connect");
+        return Ok(CatalogAction::Done);
+    }
```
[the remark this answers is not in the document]
saurabh500: Good catch on the ordering, but I measured this against msodbcsql 18 before changing it and the current behavior turns out to be the parity-correct one — so I've kept it and documented why.

`(Default)` is not a runtime sentinel in the native driver. It's `DEFAULT_STRING` (`sqlsrv.h:2065`), used only by the DSN-setup dialog (`sqlcdlg.cpp`, and the `x_rgDSNFieldsLookup` / DSN field tables in `sqlcconn.cpp`) as the placeholder shown for "no explicit value". There is no pre-connect special case for it in `SQLSetConnectAttr`.

Measured, pre-connect, with no `DATABASE=` keyword in the connection string:

```
[ODBC Driver 18 for SQL Server]
  set '(Default)'      -> SUCCESS   connect -> 28000 Login failed for user '...'
  set 'no_such_db_xyz' -> SUCCESS   connect -> 28000 Login failed for user '...'
  set ''               -> SUCCESS   connect -> OK, catalog='master'
```

`(Default)` behaves exactly like a database that does not exist — the server refuses the login. Post-connect, where the sentinel *is* meaningful, both drivers already agree byte for byte:

```
                     msodbcsql      mssql-odbc
  set '(Default)'    SUCCESS         SUCCESS     (no USE, catalog unchanged)
  set 'master'       SUCCESS_INFO    SUCCESS_INFO
  set 'no_such_db'   SQL_ERROR       SQL_ERROR
```

Suppressing the pre-connect store would have made us succeed and silently connect to `master` where msodbcsql fails — a divergence, and a quiet one, which is worse for `mssql-python` callers passing `SQL_ATTR_CURRENT_CATALOG` through `attrs_before`.

Changes made: replaced the misleading comment in `decide()` with the source citation and the measurement, and added e2e Variation 23 `PreConnectDefaultCatalogIsNotASentinel`, which passes unmodified against **both** drivers (23/23 each). Unit test `default_sentinel_is_stored_literally_before_connect` pins the store.

> [!NOTE]

mssql-odbc/src/api/sqlstate.rs
```diff
@@ -223,6 +223,46 @@ pub(crate) const WARN_CONNECTION_TIMEOUT_CHANGED: DiagMsg = DiagMsg {
...
+                state,
+                forced_state,
+                native,
+                format!("{SERVER_DIAG_SUBSOURCE}{}", e.message),
+            );
+        }
+        post_tds_info_messages(state, &diagnostics.info_messages);
+        return;
+    }
+    post_sql_error(state, forced_state, 0, err.to_string());
```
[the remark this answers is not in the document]
saurabh500: Fixed. A transport or protocol failure now keeps `HY000` instead of inheriting the caller's forced state:

```rust
let TdsError::SqlServerError { diagnostics } = err else {
    post_tds_error(state, err, SQLSTATE_HY000);
    return;
};
if diagnostics.errors.is_empty() {
    post_tds_error(state, err, SQLSTATE_HY000);
    return;
}
```

`HY000` is what every other execution path already uses for a failure that is not the engine's (`exec_common.rs:144`, `fetch.rs:224`, `get_data.rs:503`, `more_results.rs:198`), so this just brings the override in line with them. The `HY024` force now applies only where msodbcsql applies it — engine-raised errors on the catalog path (`sqlcmisc.cpp:1873-1875`).

Worth noting the first cut of this fix still passed `forced_state` down as the fallback, which made it a no-op; the test below is what surfaced that.

Renamed `post_tds_error_as_falls_back_for_non_server_errors` to `post_tds_error_as_does_not_force_the_state_on_a_transport_failure` and inverted its assertion to `SQLSTATE_HY000`, and updated `post_tds_error_as_falls_back_when_the_error_list_is_empty` to match. `sqlstate.rs` is at 100% diff coverage.

> [!NOTE]

mssql-odbc/src/api/set_connect_attr.rs
```diff
@@ -210,24 +224,72 @@ unsafe fn sql_set_connect_attr_w_impl(
...
+
+    for stmt_ptr in statements {
+        // SAFETY: every pointer in `statements` came from
+        // `handle_to_raw::<StmtHandle>` and is owned by this DBC.
+        let stmt = unsafe { handle_from_raw::<StmtHandle>(stmt_ptr) };
+        match stmt.inner.lock() {
+            Ok(mut stmt_state) => stmt_state.query_timeout = seconds,
+            // One unusable statement must not abort the fan-out: the connection
+            // default is already stored and the other statements still need it.
+            Err(_) => error!(?stmt_ptr, "SQLSetConnectAttrW: stmt mutex poisoned"),
```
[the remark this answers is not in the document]
saurabh500: Fixed — this is now worst-wins, matching `close_all_cursors` (`txn.rs:146-169`), which is the established pattern in this crate for a fan-out over statements.

The walk still continues so healthy statements get the value, but a failure is tracked and reported:

```rust
Err(_) => {
    error!(?stmt_ptr, "SQLSetConnectAttrW: stmt mutex poisoned");
    poisoned = true;
}
```

then, after the loop:

```rust
let outcome = if poisoned {
    (ERR_STATEMENT_UNUSABLE, SQL_ERROR)
} else if clamped {
    (WARN_OPTION_VALUE_CHANGED, SQL_SUCCESS_WITH_INFO)
} else {
    return SQL_SUCCESS;
};
```

A poisoned statement outranks clamping, so the caller never gets `01S02` while an `HY000` condition is outstanding. Added `ERR_STATEMENT_UNUSABLE` (`HY000`, "A statement on this connection is unusable and did not receive the attribute") rather than reusing an unrelated diagnostic.

This also made the second DBC lock lazy — the success path no longer takes it at all, since there is nothing to post.

`query_timeout_fan_out_skips_a_poisoned_statement` is renamed to `query_timeout_fan_out_reports_a_poisoned_statement` and now asserts `SQL_ERROR` + `HY000`, while still checking that the connection default and the healthy statement both received `19`. `set_connect_attr.rs` is at 100% diff coverage.

> [!NOTE]

mssql-odbc/src/api/current_catalog.rs
```diff
@@ -0,0 +1,559 @@
...
+        h.mark_dbc_connected();
+        {
+            let dbc = unsafe { handle_from_raw::<DbcHandle>(h.dbc) };
+            dbc.inner.lock().unwrap().active_stmt = Some(stmt);
+        }
+        assert_eq!(set_catalog(h.dbc, "master"), SQL_ERROR);
+        assert_eq!(sqlstate(h.dbc), "24000");
+        // The stored value is untouched by the rejected set.
+        assert_eq!(get_catalog(h.dbc, 64).1, "");
+        unsafe { sql_free_handle(SQL_HANDLE_STMT, stmt) };
```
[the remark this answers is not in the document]
saurabh500: Fixed. Dropped the explicit `sql_free_handle` and left the clear:

```rust
// Clear `active_stmt` and let `TestHandles::drop` free the statement:
// freeing it here would leave the raw pointer in `extra_stmts` and free
// it twice.
let dbc = unsafe { handle_from_raw::<DbcHandle>(h.dbc) };
dbc.inner.lock().unwrap().active_stmt = None;
```

The clear still matters — without it the fixture's `Drop` would hit the outstanding-statement assertion when freeing the DBC — but the free itself belongs to `TestHandles`. This also let `sql_free_handle` and `SQL_HANDLE_STMT` drop out of the test module's imports.

> [!NOTE]

mssql-odbc/src/api/current_catalog.rs
```diff
@@ -0,0 +1,559 @@
...
+    dbc: &DbcHandle,
+    value_ptr: SqlPointer,
+    buffer_length: SqlInteger,
+    string_length_ptr: *mut SqlInteger,
+) -> SqlReturn {
+    let Ok(mut state) = dbc.inner.lock() else {
+        error!("SQLGetConnectAttrW(SQL_ATTR_CURRENT_CATALOG): dbc mutex poisoned");
+        return SQL_ERROR;
+    };
+
```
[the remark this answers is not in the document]
saurabh500: Fixed. `get_current_catalog` now calls `free_errors` itself, right after taking the lock, with a comment explaining why it needs its own call — this path is dispatched ahead of the integer attributes and so never reaches the shared `free_errors` in `sql_get_connect_attr_w_safe`.

Without it a stale record from a prior failing call sat at index 1 and `SQLGetDiagRec(..., 1, ...)` returned that instead of the `01004` truncation warning, which is exactly what a caller checks after `SQL_SUCCESS_WITH_INFO`.

Added `get_clears_diagnostics_from_the_previous_call`, which posts a diagnostic on the DBC, asserts it is there, then does a truncating get and asserts record 1 is `01004`. `current_catalog.rs` is at 100% coverage on the lines this PR changed.

> [!NOTE]

mssql-odbc/src/api/set_connect_attr.rs
```diff
@@ -210,24 +224,72 @@ unsafe fn sql_set_connect_attr_w_impl(
...
+        state.stmt_query_timeout = seconds;
+        // Clone so the statement locks below are taken with the DBC mutex
+        // released, matching `close_all_cursors`.
+        state.statements.clone()
+    };
+
+    for stmt_ptr in statements {
+        // SAFETY: every pointer in `statements` came from
+        // `handle_to_raw::<StmtHandle>` and is owned by this DBC.
+        let stmt = unsafe { handle_from_raw::<StmtHandle>(stmt_ptr) };
```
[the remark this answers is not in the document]
saurabh500: Correct, and deliberately not fixed locally — this is a known driver-wide handle-lifetime gap, not something specific to this fan-out.

`close_all_cursors` (`txn.rs:156-161`) does exactly the same clone-then-`handle_from_raw` and already documents it, calling out "the same handle-lifetime gap `SQLDisconnect` documents (see the TODO in `disconnect.rs`), which refcounted handles will close for the whole driver at once." Taking the DBC lock across the statement locks would close the window here but inverts the lock order that `close_all_cursors` establishes, trading a documented race for a deadlock risk — and it would still leave the same gap in the other paths.

I've expanded the SAFETY comment so it points at the shared gap rather than implying this site is sound in isolation:

```rust
// SAFETY: every pointer in `statements` came from
// `handle_to_raw::<StmtHandle>` and is owned by this DBC.
// A concurrent `SQLFreeHandle(SQL_HANDLE_STMT)` could still free it
// between the clone above and this call — the same handle-lifetime gap
// `close_all_cursors` and `SQLDisconnect` document (see the TODO in
// `disconnect.rs`), which refcounted handles will close driver-wide.
```

Refcounted handles are the real fix and want to land in one change across `close_all_cursors`, `disconnect`, and this path together.

> [!NOTE]

mssql-odbc/src/api/exec_common.rs
```diff
@@ -316,13 +316,20 @@ pub(super) unsafe fn build_named_params(
...
     for i in 0..marker_count {
         let Some(Some(bound_param)) = stmt_state.bound_params.get(i) else {
             error!("{op}: parameter {} has no bound value", i + 1);
             post_diag(stmt_state, ERR_UNBOUND_PARAMETER);
             return Err(SQL_ERROR);
         };
+        // Applied before anything reads the binding: ODBC shifts the
+        // indicator pointer alongside the value pointer, so the
+        // data-at-execution check below has to see the shifted indicator.
+        let bound_param = bound_param.with_bind_offset(bind_offset);
```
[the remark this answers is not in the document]
saurabh500: Fixed in `4c50bd36`: the shifted indicator now uses `read_unaligned()`, and the regression test combines a byte-misaligned indicator with a bind offset.

> [!NOTE]

mssql-odbc/src/handles/stmt.rs
```diff
@@ -185,6 +187,150 @@ pub(crate) struct StmtState {
...
+    /// duration of the execution, per the `SQLSetStmtAttr` contract.
+    pub(crate) unsafe fn param_bind_offset(&self) -> isize {
+        let ptr = self
+            .get(odbc_types::SQL_ATTR_PARAM_BIND_OFFSET_PTR)
+            .unwrap_or(0) as *const odbc_types::SqlLen;
+        if ptr.is_null() {
+            return 0;
+        }
+        unsafe { ptr.read() }
+    }
```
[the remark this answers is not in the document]
saurabh500: Fixed in `4c50bd36`: the bind-offset `SQLLEN` now uses `read_unaligned()`, with a regression test using a one-byte-misaligned pointer.

> [!NOTE]

mssql-odbc/src/api/set_connect_attr.rs
```diff
@@ -33,14 +40,17 @@ const MAX_LOGIN_TIMEOUT_SECS: u64 = 0xfffe;
...
+/// identifiers return `HY092` rather than silently succeeding.
 ///
 /// # Safety
 /// - `connection_handle` must be a valid `DbcHandle` from `SQLAllocHandle`.
 /// - For `SQL_COPT_SS_ACCESS_TOKEN`, `value_ptr` must point to an ACCESSTOKEN
 ///   struct: a 4-byte little-endian length prefix followed by that many bytes
 ///   of the UTF-16-LE-encoded access token.
+/// - For `SQL_ATTR_CURRENT_CATALOG`, `value_ptr` must point to `string_length`
+///   `SQLWCHAR`s, or to a NUL-terminated wide string when `string_length` is
+///   `SQL_NTS`.
```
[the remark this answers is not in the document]
saurabh500: Fixed in `4c50bd36`: the safety contract now says `StringLength` counts readable bytes of UTF-16, matching the implementation and existing byte-length test.

> [!NOTE]

mssql-odbc/docs/attributes_plan.md
```diff
@@ -0,0 +1,805 @@
...
+**Consumer:** `microsoft/mssql-python`, replacing its in-package `msodbcsql18` with
+this driver. Every slice below is justified by a call the Python layer actually
+makes, or by the unfiltered pass-through surface it exposes (§4.10).
+
+**Parity reference:** `C:\work\msodbcsql\Sql\Ntdbms\sqlncli\odbc\sqlcmisc.cpp`
+(`ExportImp::SQLSetConnectAttrW` L1459, `SQLGetConnectAttrW` L2979,
+`SQLSetStmtAttrW` L3508, `SQLGetStmtAttrW` L4186).
+
+**Status:** S2 and S3 are implemented, with the string-I/O half of S1 pulled in as
+a prerequisite. S1 (remaining), S4, S5 and S6 are open. Behavior marked *measured*
```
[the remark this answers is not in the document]
saurabh500: Updated in `4c50bd36`: the current-state, sequence, risks, and bookkeeping sections now agree that S1-S4, S5a, and S6 shipped, with only S5b tracked under AB#47526.

> [!NOTE]

mssql-odbc/src/api/exec_common.rs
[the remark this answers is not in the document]
saurabh500: Fixed in `4c50bd36`: `DaeParam` captures the bind-offset-adjusted token at execution time and `SQLParamData` returns that staged value. A focused test fails if it falls back to the original binding.

> [!NOTE]

mssql-odbc/src/api/driver_connect.rs
```diff
@@ -224,6 +290,20 @@ fn do_connect(
...
+    let mut params = params;
+    apply_vendor_overrides(&mut params, &state.vendor_overrides);
+    let params = params;
+
+    // Record what the connection actually resolved to, so a later get reports
+    // the effective setting rather than only what an attribute happened to set.
+    // Measured: with no attribute set at all, `Encrypt=no` reads back 0 and an
+    // absent `TrustServerCertificate` reads back 0, so msodbcsql sources the
+    // get from the keyword when the attribute was never used.
+    state.vendor_overrides = effective_vendor_settings(&params);
```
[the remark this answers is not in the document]
saurabh500: Fixed in `4c50bd36`: explicit overrides and effective connected values now have separate storage; effective values publish only after success and clear on disconnect. A two-attempt regression test proves failed connection-string values are not reused.

> [!NOTE]

saurabh500 (commented):
## Summary

The latest commits address the correctness findings from my review, including deriving post-connect encryption attributes from the negotiated transport state. One non-blocking test-hardening suggestion remains.

mssql-odbc/src/api/set_stmt_attr.rs
```diff
@@ -776,4 +1279,839 @@ mod tests {
...
+        // Written after the set: the value is read at execute time, not
+        // captured when the attribute was assigned.
+        unsafe { slot.write(-8) };
+        assert_eq!(unsafe { attrs.param_bind_offset() }, -8);
+    }
+
+    #[test]
+    fn param_bind_offset_accepts_a_misaligned_application_pointer() {
+        let mut attrs = InertStmtAttrs::default();
+        let mut storage = [0u8; size_of::<SqlLen>() + 1];
```
saurabh500: **Suggestion:** This does not guarantee a misaligned pointer. `[u8; ...]` has alignment 1, so its base may be at any address; adding one byte can still produce an address aligned for `SqlLen`. The test could then stay green if `param_bind_offset()` regresses to `ptr.read()`. Use aligned backing storage such as `[SqlLen; 2]`, offset its byte pointer by one, and assert that the resulting address is not divisible by `align_of::<SqlLen>()`. The same construction at `mssql-odbc/src/api/exec_common.rs:812` should use the same setup.

saurabh500: Fixed in c59d3f3e. Both tests now use `SqlLen`-aligned backing storage, shift by one byte, and assert the resulting address is actually misaligned; changing that shift to zero makes the test fail at the new assertion.

> [!NOTE]

David-Engel (commented):
## Summary

Read the full production diff in `mssql-odbc` and `mssql-tds`, the recognition tables, `attributes_plan.md`, and the prior Copilot review threads plus your replies — all 13 earlier findings check out as genuinely addressed, so nothing here re-files one. CI is green on the completed legs including the cross-repo mssql-python suite, and diff coverage is 99%.

The layering, unsafe hygiene (`read_unaligned` on every application buffer), lock discipline in the `QUERY_TIMEOUT` fan-out and the three-phase `CURRENT_CATALOG` path, and the `as_writer_ref` addition in `mssql-tds` all look right. **No blocking issues.** Five suggestions/nits below, four inline.

The one I'd most like an answer on is `SQL_ATTR_METADATA_ID`: on `main` a set returned `HY092`, and after this PR it returns `SQL_SUCCESS` while catalog dispatch still forces pattern mode. That converts a loud failure into a silent wrong-results path, and the plan doc itself (§6.3) says silent pattern matching isn't a valid final state.

### Nit not anchorable to a diff line

`mssql-odbc/docs/attributes_plan.md:695` — the mermaid node still reads `catalog.rs assumes always FALSE`, which contradicts the `catalog.rs` module-doc change in this same PR ("stored and reads back, but this catalog path does not consult it yet") and §S4's own wording.

mssql-odbc/src/handles/stmt.rs
```diff
@@ -200,17 +200,164 @@ pub(crate) struct StmtState {
...
+    (odbc_types::SQL_ATTR_USE_BOOKMARKS, 0),
+    (odbc_types::SQL_ATTR_ENABLE_AUTO_IPD, 0),
+    (odbc_types::SQL_ATTR_FETCH_BOOKMARK_PTR, 0),
+    (odbc_types::SQL_ATTR_PARAM_BIND_OFFSET_PTR, 0),
+    (odbc_types::SQL_ATTR_PARAM_BIND_TYPE, 0),
+    (odbc_types::SQL_ATTR_PARAM_OPERATION_PTR, 0),
+    (odbc_types::SQL_ATTR_PARAM_STATUS_PTR, 0),
+    (odbc_types::SQL_ATTR_PARAMS_PROCESSED_PTR, 0),
+    (odbc_types::SQL_ATTR_ROW_OPERATION_PTR, 0),
+    (odbc_types::SQL_ATTR_METADATA_ID, 0),
```
David-Engel, line 276: Accepting `SQL_ATTR_METADATA_ID = SQL_TRUE` here is a behavior change worth calling out explicitly: on `main` the set fell through to the old `_` arm and returned `HY092` / `SQL_ERROR`. After this PR it returns `SQL_SUCCESS` and round-trips, while `catalog.rs` still sends `@fUsePattern = 1` unconditionally.

So an application that sets it to get identifier semantics now gets pattern matching with no diagnostic at all — a table literally named `my_table` matches `myXtable` too, and the caller has no way to notice. That's the one attribute in the inert set whose value changes which rows come back, rather than just what a later get reports.

Your own plan doc reaches the same conclusion and names the interim fix (§6.3): *"S5b must implement identifier mode or reject `SQL_TRUE` with `HYC00`; silent pattern matching is not a valid final state."*

Two options, either is fine by me:

- Reject `SQL_TRUE` with `HYC00` in the set path now (keep accepting `SQL_FALSE`) until AB#47526 lands. It diverges from msodbcsql for that one value, but it fails loudly rather than silently, and it's the option the doc already sanctions.
- Keep it as-is, but say in the PR description that this attribute *used to fail* — "stored and round-trips" reads like new capability rather than a loosened rejection, and a reviewer of the mssql-python cutover would want to know.

mssql-odbc/src/api/current_catalog.rs
```diff
@@ -0,0 +1,609 @@
...
+        }
+
+        // SAFETY: `value_ptr` is non-null and, per the ODBC contract for a
+        // character attribute, points to `string_length` bytes or to a
+        // NUL-terminated string when `string_length == SQL_NTS`.
+        let name = unsafe { read_utf16_attr(value_ptr as *const SqlWChar, string_length) };
+
+        // msodbcsql measures the limit in UTF-16 code units (`wcslen`), so a
+        // name of astral characters hits it in half as many `char`s.
+        if name.is_empty() || name.encode_utf16().count() > SYSNAMELEN {
```
David-Engel: This empty-name rejection sits in the validation block, which runs before `decide()` looks at `connection_state` — so it applies pre-connect as well as post-connect.

That looks like it diverges from the measurement you posted on the `(Default)` thread, which had msodbcsql answering `SUCCESS` for an empty string pre-connect:

```
  set ''  -> SUCCESS   connect -> OK, catalog='master'
```

Variation 14 (`SetCurrentCatalogEmptyFails`) runs on the connected fixture, where both drivers agree on `HY024`, so it wouldn't catch this. The pre-connect case is exactly the `attrs_before` shape mssql-python forwards unfiltered.

Either add a pre-connect variation to confirm the divergence is intentional, or move the `name.is_empty()` check under the `Connected` branch in `decide()` and let a pre-connect empty value be stored (or ignored) the way msodbcsql does. The `SYSNAMELEN` half of the check looks right in both states.

mssql-odbc/src/api/set_stmt_attr.rs
```diff
@@ -215,6 +268,162 @@ fn sql_set_stmt_attr_w_safe(
...
+            // empty instead would silently clear the attribute.
+            if string_length < 0 && string_length != SqlInteger::from(SQL_NTS) {
+                error!(
+                    attribute,
+                    string_length, "SQLSetStmtAttrW: invalid string length"
+                );
+                post_diag(&mut state, ERR_INVALID_ATTRIBUTE_VALUE);
+                return SQL_ERROR;
+            }
+            let value = unsafe { read_utf16_attr(value_ptr as *const SqlWChar, string_length) };
```
David-Engel, line 411: A null `value_ptr` reaches `read_utf16_attr` here and comes back as an empty `String` (`read_utf16_long` null-checks and returns `String::new()`), so `SQLSetStmtAttr(SQL_SOPT_SS_QUERYNOTIFICATION_MSGTEXT, NULL, SQL_NTS)` silently clears the stored value and reports `SQL_SUCCESS`.

The other character attribute added in this PR takes the opposite line — `set_current_catalog` explicitly null-checks and posts `HY024` before reading. Within one PR the two paths should agree on what a null character-attribute value means.

The existing tests cover the null *get* (`query_notification_null_pointer_is_a_length_query`) but not the null set, so nothing pins whichever answer you land on. Worth measuring what msodbcsql does with a null here and then matching it on both paths.

mssql-odbc/src/api/set_stmt_attr.rs
```diff
@@ -215,6 +268,162 @@ fn sql_set_stmt_attr_w_safe(
...
+                post_diag(&mut state, ERR_INVALID_ATTRIBUTE_VALUE);
+                SQL_ERROR
+            }
+        }
+        // Recognized attributes stored and round-tripped without effect: these
+        // parameter and cursor controls do not change the implemented
+        // forward-only, read-only behavior, but msodbcsql reports back whatever
+        // was written, so silently discarding the value would diverge on the
+        // very next get.
+        attribute if state.inert_attrs.set(attribute, value_ptr as SqlULen) => {
```
David-Engel: This guard both tests and mutates, which makes the arm's position load-bearing in a way nothing enforces. `SQL_ATTR_APP_ROW_DESC` and `SQL_ATTR_APP_PARAM_DESC` sit below it and only survive because 10010/10011 happen not to be in `INERT_STMT_ATTRS` — add a future explicit arm below this line for an id that *is* in the table and it gets swallowed into inert storage, returning `SQL_SUCCESS` with no test failing.

A pure guard would make the ordering safe by construction:

```rust
attribute if InertStmtAttrs::contains(attribute) => {
    state.inert_attrs.set(attribute, value_ptr as SqlULen);
    debug!(attribute, "SQLSetStmtAttrW: attribute stored without effect");
    SQL_SUCCESS
}
```

`index_of` is already there, so `contains` is a one-liner beside it. Non-blocking — the current code is correct today.

saurabh500: Fixed in `15afe20d`: inert-attribute recognition is now a pure `contains` check, and the write happens only in the final fallback after every explicit attribute arm. This removes the temporary inert-state mutation for descriptor and other explicitly handled attributes.

> [!NOTE]

mssql-odbc/src/api/current_catalog.rs
saurabh500: Fixed in `15afe20d` by making empty-catalog validation connection-state-sensitive. Empty values are retained before connect so SQL Server selects the login's default database, while a live connection still returns `HY024`; unit and cross-driver e2e coverage pin both paths.

> [!NOTE]

mssql-odbc/src/api/set_stmt_attr.rs
saurabh500: Checked the native implementation: a null query-notification pointer intentionally clears the stored value and succeeds. I kept that behavior and added explicit clear-after-set coverage for both message text and options against both drivers in `15afe20d`.

> [!NOTE]

mssql-odbc/src/handles/stmt.rs
saurabh500: Fixed in `15afe20d`. `SQL_FALSE` remains supported and readable, while `SQL_TRUE` now returns `HYC00` and other values return `HY024`; rejected sets leave the stored value unchanged until AB#47526 adds identifier-mode catalog semantics.

> [!NOTE]

mssql-odbc/tests/e2e/tests/attributes_test.cpp
```diff
@@ -0,0 +1,2032 @@
...
+
+    written = -1;
+    EXPECT_EQ(SQL_SUCCESS,
+              SQLGetStmtAttr(stmt_, kSsQnMsgtext, nullptr, 0, &written));
+    EXPECT_EQ(value_bytes, written);
+
+    // A null set pointer clears either value. This is distinct from a null get
+    // pointer above, which only asks for the current length.
+    for (SQLINTEGER attribute : {kSsQnMsgtext, kSsQnOptions}) {
+        ASSERT_SQL_OK(SQLSetStmtAttr(stmt_, attribute, msg, SQL_NTS),
```
David-Engel, line 1568: These are the assertions that turn the Windows parity leg red. From build 170438, the msodbcsql run:

```
[ RUN      ] AttributesTest.QueryNotificationStringsFollowTheByteLengthContract
unknown file: error: SEH exception with code 0xc0000005 thrown in the test body.
[  FAILED  ] AttributesTest.QueryNotificationStringsFollowTheByteLengthContract
```

`0xc0000005` is an access violation — `SQLSetStmtAttr(attribute, nullptr, SQL_NTS)` dereferences the null inside msodbcsql. The mssql-odbc leg passed all 28 on Windows; only the reference driver crashed.

That matters beyond the red build, because it undercuts the reasoning behind the fix. The reply on the earlier thread said *"Checked the native implementation: a null query-notification pointer intentionally clears the stored value and succeeds."* Measured, it does not succeed — it faults. There is no native contract here to be at parity with, so "clear and return `SQL_SUCCESS`" is a **deliberate divergence** (a good one: not crashing the host process is strictly better), not parity. Worth saying so explicitly rather than recording it as matching msodbcsql.

This crate already has the pattern for exactly this situation. `attributes.rs` documents `SQL_COPT_SS_CEKEYSTOREDATA` as an id whose *"set probe faults msodbcsql, so the process died before reaching the get probe"*, and Variation 31 `HostileAttributePayloadDoesNotFault` guards the equivalent case with `SKIP_IF_COMPARING_MSODBCSQL()`.

Suggested shape: pull these null-set assertions out of this parity test into their own variation carrying `SKIP_IF_COMPARING_MSODBCSQL()`, and note in the comment that msodbcsql access-violates so the behavior is this driver's own choice. That keeps the byte-length contract test purely comparative and stops a known crasher from failing the parity gate.

mssql-odbc/tests/e2e/tests/attributes_test.cpp
```diff
@@ -0,0 +1,2032 @@
...
+    ASSERT_TRUE(SQL_SUCCEEDED(SQLAllocHandle(SQL_HANDLE_DBC, env_, &dbc)));
+
+    SQLTCHAR empty[] = {0};
+    ASSERT_SQL_OK(SQLSetConnectAttr(dbc, SQL_ATTR_CURRENT_CATALOG, empty, SQL_NTS),
+                  SQL_HANDLE_DBC, dbc);
+
+    SqlTString conn = ODBCTestUtils::ToSqlTStr(cs);
+    SQLTCHAR out[1024] = {};
+    SQLSMALLINT out_len = 0;
+    ASSERT_SQL_OK(SQLDriverConnect(dbc, nullptr, conn.data(), SQL_NTS, out, 1024,
```
David-Engel, line 672: This assertion fails on Linux and Linux ARM in build 170438, on all four retries:

```
attributes_test.cpp:668: Failure
Value of: (((_rc)&(~1))==0)
  Actual: false
ODBC call failed with rc=-1  error=[08001] Protocol Error: Failed to receive
token during login response parsing. (native=0)
[  FAILED  ] AttributesTest.PreConnectEmptyCatalogUsesLoginDefault
```

It passes on Windows (that leg reported 28/28), so something is environment-specific.

The connection-string surgery above looks sound to me — `BuildConnectionString()` emits `Database=<db>;` as a clean segment, so the `erase` removes exactly it. What's left is the same string the fixture's own `Connect()` uses successfully, minus `Database=`.

Which makes the interesting question: is this the *test* or a real gap? Two things make it worth a closer look before just relaxing the assertion:

- Variation 23 also connects without a `Database=` keyword and gets a clean `28000`, but there `current_catalog` is `Some("(Default)")`. The combination this test newly exercises — `current_catalog == Some("")`, so `context.database` is empty — was unreachable before `15afe20d`, because an empty value used to be rejected with `HY024`. So the fix opened a LOGIN7 path nothing had exercised.
- An `08001` *protocol* error during login-response parsing is a different shape from the `28000` you'd expect if the server simply refused a database. That reads more like a handshake problem than a permissions one.

Could you check whether `ClientContext::database == ""` produces a well-formed LOGIN7 on this driver? If it does and the failure is purely the CI login's default database being unavailable, then widening the `GTEST_SKIP()` guard is fine — but it'd be worth confirming which of the two it is first, since the empty-catalog path is exactly what the fix made reachable.

One note on diagnosing from the log: the parity harness compares at *binary* granularity, so the Linux run reported `attributes_test FAIL FAIL — shared failure` without naming the msodbcsql-side test. Given Windows shows msodbcsql access-violating on the other new test, the Linux msodbcsql leg most likely failed there rather than here — the two failures are probably independent, not a shared divergence.

mssql-odbc/src/api/set_stmt_attr.rs
```diff
@@ -215,6 +271,170 @@ fn sql_set_stmt_attr_w_safe(
...
+            // empty instead would silently clear the attribute.
+            if string_length < 0 && string_length != SqlInteger::from(SQL_NTS) {
+                error!(
+                    attribute,
+                    string_length, "SQLSetStmtAttrW: invalid string length"
+                );
+                post_diag(&mut state, ERR_INVALID_ATTRIBUTE_VALUE);
+                return SQL_ERROR;
+            }
+            let value = unsafe { read_utf16_attr(value_ptr as *const SqlWChar, string_length) };
```
David-Engel, line 411: Non-blocking, and follows from the e2e comment: once you've confirmed msodbcsql access-violates on a null value pointer here, this line deserves a comment recording the decision.

Right now the null path is invisible — it's an emergent property of `read_utf16_attr` returning `String::new()` for a null pointer, and a reader has no way to tell that clearing-on-null was chosen deliberately rather than fallen into. That's especially worth pinning because the other character attribute added in this PR (`set_current_catalog`) rejects null with `HY024`, so the asymmetry between the two looks accidental unless the reason is written down.

Something like: msodbcsql faults on a null value pointer for these two ids, so there is no behavior to match; this driver treats it as "clear the value" rather than crashing the host process.

The module doc at the top and `docs/attributes_plan.md` §S6 would be the other places to reflect it, since §S6 currently presents the vendor statement attributes as measured parity throughout.

David-Engel (commented):
## Re-review of `15afe20d`

All four findings from my last pass are genuinely addressed in the source, and I verified each rather than taking the reply at face value:

- **METADATA_ID** — `SQL_FALSE` accepted, `SQL_TRUE` → `HYC00`, anything else → `HY024`, rejected sets leave the stored value at 0. `metadata_id_rejects_identifier_mode_until_catalog_supports_it` pins all three, the e2e variation carries `SKIP_IF_COMPARING_MSODBCSQL()` correctly since this is a deliberate divergence, and the plan doc, `catalog.rs` module docs and the mermaid node were all updated consistently. This is the right call.
- **Empty catalog** — now state-sensitive: stored pre-connect, `HY024` once connected, and `do_connect` maps `Some("")` to an empty `context.database`, which is the login default. Both unit tests present.
- **Null QN set** — behavior kept, now explicitly tested.
- **Inert guard** — `contains()` is pure and the write moved into the fallback arm. Ordering is now safe by construction.

Locally `cargo bfmt`, `cargo bclippy` and `cargo nextest run -p mssql-odbc` (1035 passed, 1 skipped) are all clean.

## But CI went red on this commit, and it's the two new e2e tests

`Build Linux`, `Build Linux ARM` and `Build Windows` all fail on `15afe20d`; they were green on `c02050c8`. I pulled the ADO timeline and logs for build 170438 — every failure is in the ODBC C++ e2e task, and both are new tests added by this commit. Details inline; the short version:

1. **`QueryNotificationStringsFollowTheByteLengthContract`** — the new null-set assertions make **msodbcsql access-violate** (`SEH exception 0xc0000005`) on the Windows parity leg. So the premise of that fix ("the native implementation intentionally clears and succeeds") does not hold: msodbcsql crashes rather than defining a behavior.
2. **`PreConnectEmptyCatalogUsesLoginDefault`** — fails on Linux (all 4 retries) at the `SQLDriverConnect`, `08001 Protocol Error: Failed to receive token during login response parsing`. It passes on Windows, so it is environment-sensitive.

Neither is a problem with the *product* changes, which I think are right — both are the new tests. Nothing else in the diff regressed.

Still no blocking concerns with the driver behavior itself; the blocker is just getting validation green again.

mssql-odbc/tests/e2e/tests/attributes_test.cpp
saurabh500: Confirmed the empty LOGIN7 database is valid; the failure was earlier in unixODBC replay. It caches this pre-driver wide attribute as ANSI and later calls the unsuffixed `SQLSetConnectAttr`, so the driver now exports that Unix entry point, transcodes current catalog into the shared wide path, and the full Linux comparison passes all 28 binaries for both drivers with zero divergence.

> [!NOTE]

mssql-odbc/src/api/set_stmt_attr.rs
saurabh500: Added the rationale both at the null-handling branch and in the module docs, and updated §S6 as well. The code now states explicitly that msodbcsql faults here and that clearing is this driver’s deliberate safe result.

> [!NOTE]

mssql-odbc/tests/e2e/tests/attributes_test.cpp
saurabh500: Split the null-plus-`SQL_NTS` case into `QueryNotificationNullNtsSetClearsWithoutFault`, which skips the msodbcsql comparison leg; the shared parity test now clears with `nullptr, 0`. The test and §S6 now record this as a deliberate safe divergence because msodbcsql 18.6.2.1 access-violates.

> [!NOTE]

mssql-odbc/src/api/set_connect_attr.rs
```diff
@@ -59,11 +71,79 @@ pub(crate) unsafe fn sql_set_connect_attr_w(
...
+
+#[cfg(not(windows))]
+unsafe fn sql_set_connect_attr_impl(
+    connection_handle: SqlHandle,
+    attribute: SqlInteger,
+    value_ptr: SqlPointer,
+    string_length: SqlInteger,
+) -> SqlReturn {
+    if connection_handle.is_null()
+        || attribute != SQL_ATTR_CURRENT_CATALOG
```
David-Engel: This guard is correct today: `SQL_ATTR_CURRENT_CATALOG` is the only character-valued connection attribute the driver actually consumes, so forwarding everything else verbatim is safe. The risk is that the coupling is invisible. The recognized-attribute table already lists `SQL_COPT_SS_ATTACHDBFILENAME` (1221) and `SQL_COPT_SS_FAILOVER_PARTNER` (1225), both character-valued in msodbcsql. Whenever one of those graduates from `HYC00` to a real implementation, this Unix replay path will hand its ANSI bytes to the wide implementation to be reinterpreted as UTF-16, and nothing fails at compile time to catch it.

Consider keying the guard off a shared predicate rather than a bare literal, so adding a string-valued attribute is one edit instead of two that have to be remembered together:

```rust
#[cfg(not(windows))]
fn is_character_connect_attr(attribute: SqlInteger) -> bool {
    matches!(attribute, SQL_ATTR_CURRENT_CATALOG)
}
```

and here: `|| !is_character_connect_attr(attribute)`.

mssql-odbc/src/api/exports.rs
```diff
@@ -118,6 +118,35 @@ pub unsafe extern "C" fn SQLSetConnectAttrW(
...
+///
+/// unixODBC converts character-valued attributes set before driver selection to
+/// ANSI, then replays them through this symbol when the driver is loaded.
+///
+/// # Safety
+/// - `connection_handle` must be a valid DBC handle.
+/// - `attribute` must be a valid connection attribute identifier.
+/// - `value_ptr` validity depends on the attribute type.
+/// - `string_length` is used only for string-type attributes.
+#[cfg(not(windows))]
```
David-Engel, line 134: `not(windows)` also exports this on macOS, where the Driver Manager is iODBC rather than unixODBC. Both the doc comment above and the `attributes_plan.md` note describe unixODBC's caching behavior specifically, and the e2e suite builds narrow on all Unix, so macOS now routes through whatever iODBC does with a Unicode driver that also exports an ANSI setter.

Did the macOS leg come back green with this symbol present? If the replay behavior is unixODBC-specific, `cfg(target_os = "linux")` keeps the blast radius to the platform you measured. If iODBC does the same thing, worth one line in the doc comment saying so, so the `not(windows)` gate reads as deliberate rather than incidental.

David-Engel (commented):

## Summary

This is a re-review of the delta since `15afe20d` (commits `875a5ff1` and `77929db5`), not of the full 8k-line PR. The delta adds a Unix ANSI `SQLSetConnectAttr` export that transcodes a pre-connect `SQL_ATTR_CURRENT_CATALOG` and forwards every other attribute to the wide implementation, splits the null-value query-notification divergence into its own `SKIP_IF_COMPARING_MSODBCSQL` test, and documents both. It resolves the two things I raised last round: the Linux and Linux-ARM failure at `attributes_test.cpp:672`, and the undocumented null-pointer clear at `set_stmt_attr.rs:411`. No blocking issues.

## Verification

The new code is entirely `cfg(not(windows))`, and the PR's validation section reports Windows runs only, so none of it was exercised by that run. I built and ran it on Linux in a `rust:1.95` container against this head:

- `cargo check -p mssql-odbc --all-targets` — clean, no warnings on the new code.
- `cargo test -p mssql-odbc --lib -- ansi_setter initial_database` — 4 passed, 0 failed (the three new ANSI setter tests plus `initial_database_uses_keyword_then_attribute_then_login_default`).
- A throwaway probe against the `SYSNAMELEN + 1` truncation, which is the least obvious part of the shim:

```
PROBE len=127 ret=0  stored_len=Some(127)
PROBE len=128 ret=0  stored_len=Some(128)
PROBE len=129 ret=-1 stored_len=None
PROBE len=200 ret=-1 stored_len=None
PROBE utf8    ret=0  stored=Some("dépôt_数据")
PROBE wide-path len=129 ret=-1 stored=None
```

The boundary is exact and identical to the wide entry point: 128 units accepted, 129 and 200 rejected with `SQL_ERROR` leaving the stored catalog untouched, and multi-byte UTF-8 round-trips intact. The "retain one extra unit rather than allocating the whole string" reasoning in the comment holds.

Not verified here: the actual unixODBC replay path (no Linux Driver Manager in this environment) and the macOS/iODBC leg. Those belong to the ADO e2e jobs.

## Findings

**Blocking** — none found.

**Suggestion**
- `mssql-odbc/src/api/set_connect_attr.rs:108` — the ANSI shim keys on a literal attribute id; a shared predicate would keep it honest when a second character-valued attribute lands.
- `mssql-odbc/src/api/exports.rs:131` — question about the `not(windows)` gate also covering macOS/iODBC.

**Nit** — none found.

## CI

`coverage-report` is red, but it latched onto ADO build 170510, an earlier failed run for this PR, rather than build 170531 for this head, which is still pending. It is not reporting anything about `77929db5`. Your unticked "Required CI passes on the latest commit" box is the accurate state.

mssql-odbc/src/api/exports.rs
saurabh500: The iODBC comment is not really correct, since we support ODBC through unixODBC only (AFAIK, but happy to be corrected). But the intention of the comment is correct

mssql-odbc/src/api/set_connect_attr.rs
saurabh500: Added `requires_ansi_transcoding` and keyed the delegation check through it, so the character-value classification is now an explicit update point when another string attribute is implemented. The current list remains only `SQL_ATTR_CURRENT_CATALOG`, the sole character attribute consumed today.

> [!NOTE]

mssql-odbc/src/api/exports.rs
saurabh500: Confirmed that macOS CI does not run the ODBC e2e suite; its build leg only compiles this export, so it provides no replay evidence. I kept `not(windows)` because `SQLSetConnectAttr` is the standard ANSI entry point on non-Windows targets, and clarified that the replay behavior motivating the conversion was measured specifically with unixODBC—there is no claim here that the macOS leg exercises iODBC replay.

> [!NOTE]

David-Engel (commented):

## Summary

Incremental re-review of what changed since my last automated pass at `77929db5`. `e9857a6f` is nine lines: the `requires_ansi_transcoding` predicate extracted in `set_connect_attr.rs`, and the `not(windows)` doc clarification in `exports.rs`. Both land the threads they answer, and I found nothing new.

Scope note, so this review is not read as more than it is: the full diff (~8.2k changed lines) is larger than this automated pass reviews in one go, so this run re-examined only the delta since my last review plus its immediate context — not the whole PR.

## Verification

Read-only, plus one targeted check on the new predicate. The correctness risk there is the classic `matches!` trap: if `SQL_ATTR_CURRENT_CATALOG` were anything other than a `const`, the pattern would bind as a catch-all identifier, `requires_ansi_transcoding` would return `true` for every attribute, and the Unix shim would transcode non-character attributes as if they were strings. It is `pub const SQL_ATTR_CURRENT_CATALOG: SqlInteger = 109;` (`api/odbc_types.rs:82`), imported unconditionally at the top of `set_connect_attr.rs`, so the pattern is a genuine constant match and the guard is exactly equivalent to the `attribute != SQL_ATTR_CURRENT_CATALOG` it replaced.

I could not compile that branch locally, since it is `cfg(not(windows))` and this machine is Windows; `cargo check -p mssql-odbc --target x86_64-unknown-linux-musl` fails here on `openssl-sys` (no OpenSSL for the cross target), which is an environment limitation rather than anything in this PR. The Linux and Linux ARM build legs of PR validation are green on this head, and all 19 checks pass, so that path is covered by CI rather than by me.

On the `exports.rs` wording: keeping `not(windows)` and stating that the replay behavior was measured with unixODBC answers the point I raised. The claim in the comment is now scoped to what was actually measured, and no stale unixODBC-only assertion is left in `attributes_plan.md`.

## Findings

- **Blocking:** none found.
- **Suggestion:** none found.
- **Nit:** none found.

Nothing here blocks human review. This is not an approval, and it is not a fresh review of the full PR.

David-Engel (approved):

PR #351, opened 2026-08-21
Vendor OpenSSL for the fuzz build to survive apt/pkg-config outages

saurabh500:
## Summary

The **GH-Rust Fuzz Test** pipeline (`definitionId=2207`, "Rust TDS Lib Fuzz Testing") intermittently fails on `main`. This PR makes the fuzz build independent of apt-provided OpenSSL so those runs stop going red.

Work item: **[AB#47459](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47459)**

## Root cause (not a fuzz crash)

I investigated the recent failing runs (168173, 168079, 167989 — all 2026-08-19). In every case the **"Run Fuzzer" step fails at build time, not at fuzz time**:

```
error: failed to run custom build command for `openssl-sys v0.9.117`
Error: failed to build fuzz script: ... cargo build --manifest-path .../mssql-tds/fuzz/Cargo.toml ... --bin fuzz_token_stream
##[error]Bash exited with code '1'.
```

`openssl-sys` cannot locate OpenSSL because `libssl-dev`/`pkg-config` were not installed (intermittent apt/mirror outage on the CI agent). This is **not a libFuzzer crash**:

- No `fuzz-crashes-*` pipeline artifacts were produced (log shows `No artifacts directory found`).
- The fuzz targets that *did* build (e.g. `fuzz_tds_client` in build 168173) ran to completion with `oom/timeout/crash: 0/0/0` and reported `No crashes found`.

So the fuzz stage is silently skipped whenever the agent can't `apt install` OpenSSL, and the pipeline reports a build failure.

## Fix

Enable the `openssl` **vendored** feature in `mssql-tds/fuzz/Cargo.toml`, target-gated to the same `cfg(all(not(windows), not(target_os = "macos")))` targets where `mssql-tds` already links OpenSSL. `openssl-sys` then compiles OpenSSL from source, so the fuzz build no longer depends on apt-provided `libssl-dev`/`pkg-config`.

```toml
[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]
openssl = { version = "0.10.66", features = ["vendored"] }
```

This is a shared `openssl-sys` (also backs `native-tls`), so no new crate is introduced — only the vendored build path is turned on for the fuzz crate.

## Testing

- `cargo metadata --manifest-path mssql-tds/fuzz/Cargo.toml --no-deps` parses cleanly.
- `cargo bfmt` passes.
- The fuzz crate is its own workspace (excluded from the main workspace), so `bclippy`/`btest` on the workspace are unaffected; the vendored build is exercised by the Linux fuzz pipeline itself.

## Notes

Draft pending a green GH-Rust Fuzz Test run on this branch. Associated ADO bug **[AB#47459](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47459)** documents the failing runs and links build 168173 as evidence.

### DIFF
--- a/mssql-tds/fuzz/Cargo.toml
+++ b/mssql-tds/fuzz/Cargo.toml
@@ -14,6 +14,16 @@ tokio = { version = "1.45.1", features = ["full"] }
 arbitrary = { version = "1.3", features = ["derive"] }
 async-trait = "0.1"
 
+# Build OpenSSL from vendored source on the targets where mssql-tds links it
+# (every non-Windows, non-macOS target -- see mssql-tds/Cargo.toml). The fuzz CI
+# agents intermittently fail to `apt install libssl-dev`/`pkg-config` from the
+# Ubuntu mirror, which leaves `openssl-sys` unable to locate OpenSSL and breaks
+# the entire fuzz stage. Enabling the vendored feature compiles OpenSSL from
+# source (the shared openssl-sys also backs native-tls here), so the fuzz build
+# no longer depends on those apt-provided system packages.
+[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]
+openssl = { version = "0.10.66", features = ["vendored"] }
+
 # Prevent this from interfering with workspaces
 [workspace]
 members = ["."]


David-Engel (commented):
## Summary

Ten added lines in `mssql-tds/fuzz/Cargo.toml` that enable `openssl/vendored` under the same `cfg(all(not(windows), not(target_os = "macos")))` gate `mssql-tds` already uses. The mechanism is correct and the root-cause writeup is well evidenced. My findings are about whether this fully closes AB#47459, plus a couple of CI prerequisites worth confirming before this leaves draft.

I verified the resolution locally (`cargo metadata --filter-platform x86_64-unknown-linux-gnu` against the fuzz manifest, which resolves 162 packages cleanly):

- `openssl 0.10.81` → features `[default, vendored]`
- `openssl-sys 0.9.117` → features `[openssl-src, vendored]` — this is the exact crate/version in the failing build log
- `openssl-src 300.6.1+3.6.3` is pulled in
- There is a single `openssl-sys` node, so the vendored build is shared with `native-tls`. The "no new crate is introduced" claim holds for `openssl-sys`; the high-level `openssl` crate was already in the graph via `mssql-tds`'s own target-gated dependency, so nothing new arrives there either.

The cfg expression and the `0.10.66` floor match `mssql-tds/Cargo.toml` exactly, and `Cargo.lock` is gitignored repo-wide, so there is no lockfile to update.

**Blocking:** none.

## Suggestions

### 1. The apt outage would still fail the job, just at an earlier step

Every fuzz job in `.pipeline/templates/validation-stages.yml` runs `install-dependencies.yml` before `fuzz-template.yml`, and that template invokes `scripts/install-deps.sh`. That script hard-fails on an apt outage:

```bash
if [ "$apt_update_ok" != true ]; then
    echo "ERROR: apt update failed after 5 attempts"
    exit 1
fi
...
if [ "$apt_install_ok" != true ]; then
    echo "ERROR: apt install failed after 5 attempts"
    exit 1
fi
```

If the mirror were genuinely unavailable, the "Install linux dependencies" step would have gone red before `cargo fuzz run` was ever reached. The fact that the failure surfaced in the "Run Fuzzer" step instead suggests `apt install` (including `libssl-dev` and `pkg-config`) actually succeeded in builds 168173 / 168079 / 167989, which would point at a different reason for `openssl-sys` not finding OpenSSL — for example `PKG_CONFIG_PATH` / `OPENSSL_DIR` not being visible to the nightly `cargo fuzz` build, or a cross/sanitizer target mismatch.

Vendoring is still a reasonable hardening step and I would not block on this. But it is worth pulling the "Install linux dependencies" step log out of build 168173 to confirm which of the two stories is true. If apt succeeded, this PR treats a symptom and the same failure can recur through the unexamined path.

### 2. `openssl-src` needs `perl` and a C toolchain that `install-deps.sh` may not provide

`openssl-src` shells out to OpenSSL's Perl `Configure` and then `make`. `scripts/install-deps.sh` installs `build-essential` (gcc/make) but does not list `perl`, and the fuzz jobs run on `imageOverride -equals RUST-1ES-UBUSLIM` — a slim image. Please confirm `perl` is present on that image; if it isn't, this swaps a `libssl-dev` failure for a `Configure` failure. Cheapest fix if in doubt:

```diff
     DEPS="jq \
         unzip \
         build-essential \
+        perl \
         pkg-config \
```

Note also that `build-essential` itself comes from apt, so the vendored path is not actually independent of apt — only of `libssl-dev`/`pkg-config` specifically. The comment block in the diff ("no longer depends on those apt-provided system packages") reads a little stronger than what the change delivers; consider narrowing it to `libssl-dev`/`pkg-config`.

### 3. PR validation does not exercise this change at all

`RunFuzz` defaults to `false`, and every non-fuzz stage in `validation-stages.yml` is gated on `eq('${{ parameters.RunFuzz }}', 'false')` while the `Fuzz` stage is gated on `'true'`. The 20 green checks on this PR therefore contain zero signal about whether the vendored build works. The draft note already says a green GH-Rust Fuzz Test run is pending — agreed, and I would treat that run as the gate for ready-for-review rather than the PR validation result.

### 4. Consider a single source of truth instead of duplicating the gate

The cfg expression and version floor now live in two manifests and can silently drift if `mssql-tds` bumps its OpenSSL floor — the fuzz crate would keep building an older vendored OpenSSL without any error. An alternative that keeps one declaration:

```toml
# mssql-tds/Cargo.toml
[features]
vendored-openssl = ["openssl/vendored"]
```

```toml
# mssql-tds/fuzz/Cargo.toml
mssql-tds = { path = "..", features = ["vendored-openssl"] }
```

That also avoids declaring a direct dependency the fuzz crate never `use`s. Your call — the current form is simpler to read in isolation, and the comment does point at `mssql-tds/Cargo.toml`.

### 5. Build-time cost per fuzz job

OpenSSL 3.6.3 compiles from source on each of the three fuzz jobs, with no shared cache between them (`CARGO_TARGET_DIR` is per-job). That is a few minutes added on top of a 1800s fuzz budget in `Fuzz_TokenStream_Linux` and `Fuzz_TdsClient_Linux`, and on top of six serial 300s runs in `Fuzz_Additional_Targets_Linux`. Worth a glance at those jobs' timeouts to confirm there is headroom — the serial job is the one most likely to be tight.

## Nits

None.

---

Not approving or requesting changes — this is a comment-only review. The merge stays with you.

mssql-tds/fuzz/Cargo.toml
```diff
@@ -14,6 +14,16 @@ tokio = { version = "1.45.1", features = ["full"] }
...
 
+# Build OpenSSL from vendored source on the targets where mssql-tds links it
+# (every non-Windows, non-macOS target -- see mssql-tds/Cargo.toml). The fuzz CI
+# agents intermittently fail to `apt install libssl-dev`/`pkg-config` from the
+# Ubuntu mirror, which leaves `openssl-sys` unable to locate OpenSSL and breaks
+# the entire fuzz stage. Enabling the vendored feature compiles OpenSSL from
+# source (the shared openssl-sys also backs native-tls here), so the fuzz build
+# no longer depends on those apt-provided system packages.
+[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]
+openssl = { version = "0.10.66", features = ["vendored"] }
```
David-Engel, line 25: Is the apt outage actually confirmed? `scripts/install-deps.sh` installs `pkg-config` and `libssl-dev` with five retries and exits non-zero if apt never succeeds, and the "Install linux dependencies" step has no `continueOnError` — so an apt failure should have red-failed the job before "Run Fuzzer" ever started.

A better fit for a build-script-only failure is `cargo fuzz`'s implicit `--target <host-triple>`: Cargo treats that as cross-compiling, and the `pkg-config` crate then refuses to run without `PKG_CONFIG_ALLOW_CROSS=1`, so `openssl-sys` cannot find an OpenSSL that is in fact installed. If that is what the build log shows, setting `PKG_CONFIG_ALLOW_CROSS=1` in `fuzz-template.yml` fixes it without adding a from-source OpenSSL build to every fuzz job.

Vendoring works either way — I would just like the fix tied to the verified `--- stderr` from build 168173 rather than to the inferred apt theory, and the comment above updated to match whatever that turns out to be.

mssql-tds/fuzz/Cargo.toml
```diff
@@ -14,6 +14,16 @@ tokio = { version = "1.45.1", features = ["full"] }
...
 async-trait = "0.1"
 
+# Build OpenSSL from vendored source on the targets where mssql-tds links it
+# (every non-Windows, non-macOS target -- see mssql-tds/Cargo.toml). The fuzz CI
+# agents intermittently fail to `apt install libssl-dev`/`pkg-config` from the
+# Ubuntu mirror, which leaves `openssl-sys` unable to locate OpenSSL and breaks
+# the entire fuzz stage. Enabling the vendored feature compiles OpenSSL from
+# source (the shared openssl-sys also backs native-tls here), so the fuzz build
+# no longer depends on those apt-provided system packages.
+[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]
```
David-Engel, line 24: Nit: `openssl` here is a direct dependency the fuzz crate never references in code — it exists purely to flip a feature on the `openssl-sys` that `native-tls` already pulls in. That works via feature unification, but the intent is invisible to a reader and it would be flagged if `cargo-udeps` or `unused_crate_dependencies` is ever turned on for this crate. Depending on `openssl-sys = { version = "0.9", features = ["vendored"] }` states the intent directly.

If you keep `openssl`, note that the `0.10.66` pin now has to stay in sync with `mssql-tds/Cargo.toml` by hand — the fuzz crate has no committed `Cargo.lock`, so a drift between the two would resolve to two versions of `openssl-sys` and reintroduce the system-OpenSSL lookup silently.

mssql-tds/fuzz/Cargo.toml
```diff
@@ -14,6 +14,16 @@ tokio = { version = "1.45.1", features = ["full"] }
 arbitrary = { version = "1.3", features = ["derive"] }
 async-trait = "0.1"
 
+# Build OpenSSL from vendored source on the targets where mssql-tds links it
```
David-Engel, line 17: Nit: seven lines of comment for one dependency is heavier than the repo usually carries, and most of it restates the PR description. Something like this keeps the *why* without the essay:

```toml
# Vendored so the fuzz build doesn't depend on the agent's libssl-dev/pkg-config
# discovery; openssl-sys is shared with native-tls. Gate matches mssql-tds. See AB#47459.
```

AB#47459 is the right place for the full failure history.

David-Engel (commented):
## Summary

Single-manifest change: a target-gated `openssl` dependency with the `vendored` feature in the fuzz crate. Mechanically it does what it claims — the cfg gate matches `mssql-tds/Cargo.toml` exactly, and because `native-tls` and this new dep share one `openssl-sys`, feature unification vendors OpenSSL for the whole fuzz binary. No source, API, or FFI changes; the version matches the `0.10.66` pin in `mssql-tds`. AB#47459 is linked and the PR follows the draft-first workflow. Nice write-up in the description — the evidence that this is a build failure rather than a libFuzzer crash is convincing.

My concern is not the mechanics but the root-cause attribution. It doesn't square with what the fuzz jobs actually do, and pinning it down matters because it decides whether this is the right fix and whether it's the cheapest one.

**Blocking:** none.

## Suggestion — confirm the root cause before this lands

The fuzz jobs run `install-dependencies.yml`, which calls `scripts/install-deps.sh`. That script installs `pkg-config` and `libssl-dev` (lines 20-21), retries `apt update` and `apt install` five times each, and hard-`exit 1`s if either never succeeds (lines 57-81). The "Install linux dependencies" step has no `continueOnError`, so a genuine apt/mirror outage should have failed the job right there and never reached "Run Fuzzer". That makes "`libssl-dev`/`pkg-config` were not installed" hard to accept as written.

A cause that does fit a build-script-only failure: `cargo fuzz` always builds with an explicit `--target <host-triple>`, which puts Cargo in cross-compile mode, and the `pkg-config` crate refuses to run when cross-compiling unless `PKG_CONFIG_ALLOW_CROSS=1` is set. `openssl-sys` then cannot locate a perfectly well-installed OpenSSL — which produces exactly the `failed to run custom build command for openssl-sys` line you quoted.

Could you paste the `--- stderr` block from the `openssl-sys` build script in build 168173 into the PR or AB#47459? The distinguishing lines are `Could not find directory of OpenSSL installation` plus either `pkg-config has not been configured to support cross-compilation` or `The system library openssl required by crate openssl-sys was not found`. If it is the cross-compile case, a one-line `PKG_CONFIG_ALLOW_CROSS=1` env in `fuzz-template.yml` fixes it with no build-time cost, and vendoring becomes a choice rather than the fix.

One corollary worth reflecting in the manifest comment either way: the vendored build needs `cc`, `make`, and `perl` from the same agent image and apt path, so if apt truly were down, `build-essential` would not install and the vendored build would fail too. This change does not make the fuzz build survive an apt outage — it makes it independent of `libssl-dev`/`pkg-config` *discovery*, which is a different and still worthwhile property.

## Suggestion — CI time budget

Vendoring compiles OpenSSL from source on every fuzz job, and each job gets a fresh agent, so there is no cache to amortize it. `Fuzz_TokenStream_Linux` and `Fuzz_TdsClient_Linux` each fuzz for 30 minutes and `Fuzz_Additional_Targets_Linux` runs six targets at 5 minutes each; none of the three sets `timeoutInMinutes`, so they inherit the 60-minute default. Please check the actual step durations on the verification run, and consider setting `timeoutInMinutes` explicitly on those jobs so a slow agent surfaces as a clear timeout instead of a mysterious cancellation.

## Note — how this gets verified

PR validation will not exercise this change. `cargo +nightly fuzz build` lives in `build-template.yml` behind an `osType: Linux` branch, but `validation-stages.yml` only uses that template for Windows and macOS — Linux builds go through `build-template-container.yml`, which has no fuzz step. So the only real check is a run of the "Rust TDS Lib Fuzz Testing" definition against this branch, which your draft note already plans. Good call holding the PR in draft until then; a green run is the only thing that will prove the vendored path builds under `cargo fuzz`'s sanitizer flags.

## Nits

See inline comments on the manifest.

mssql-tds/fuzz/Cargo.toml
```diff
@@ -14,6 +14,16 @@ tokio = { version = "1.45.1", features = ["full"] }
...
+# (every non-Windows, non-macOS target -- see mssql-tds/Cargo.toml). The fuzz CI
+# agents intermittently fail to `apt install libssl-dev`/`pkg-config` from the
+# Ubuntu mirror, which leaves `openssl-sys` unable to locate OpenSSL and breaks
+# the entire fuzz stage. Enabling the vendored feature compiles OpenSSL from
+# source (the shared openssl-sys also backs native-tls here), so the fuzz build
+# no longer depends on those apt-provided system packages.
```
David-Engel, line 23: This overstates what vendoring buys. `openssl-src` builds OpenSSL from source with **perl** plus a C toolchain (`cc`/`make`), and on this agent that toolchain comes from `build-essential` — which `scripts/install-deps.sh` installs in the *same* `apt install $DEPS` transaction as `pkg-config` and `libssl-dev`. If apt/the mirror is genuinely down, that one transaction fails, `install-deps.sh` exits 1 after its 5 retries, and the job dies at "Install linux dependencies" regardless of this change. So the fuzz build trades one apt-provided prerequisite for another rather than becoming apt-independent.

Two concrete asks:
1. Confirm `perl` is present on the `RUST-1ES-UBUSLIM` image. It is not in `$DEPS`, and on a *slim* image a missing perl turns a working build into a new `openssl-sys`/`openssl-src` failure on every fuzz run.
2. Reword the claim to what is actually true, e.g. "so the fuzz build no longer needs `libssl-dev`/`pkg-config` discovery at build time" instead of "no longer depends on those apt-provided system packages".

mssql-tds/fuzz/Cargo.toml
```diff
@@ -14,6 +14,16 @@ tokio = { version = "1.45.1", features = ["full"] }
 arbitrary = { version = "1.3", features = ["derive"] }
 async-trait = "0.1"
 
+# Build OpenSSL from vendored source on the targets where mssql-tds links it
+# (every non-Windows, non-macOS target -- see mssql-tds/Cargo.toml). The fuzz CI
+# agents intermittently fail to `apt install libssl-dev`/`pkg-config` from the
+# Ubuntu mirror, which leaves `openssl-sys` unable to locate OpenSSL and breaks
+# the entire fuzz stage. Enabling the vendored feature compiles OpenSSL from
+# source (the shared openssl-sys also backs native-tls here), so the fuzz build
```
David-Engel, line 22: Separate from the unused-dependency point already raised: this re-declares the `0.10.66` requirement that `mssql-tds/Cargo.toml` owns. The vendored feature only reaches `native-tls`/`mssql-tds` as long as both requirements resolve to one `openssl` (and one `openssl-sys`). If `mssql-tds` later moves to an incompatible range, nothing fails loudly here — the fuzz crate quietly builds a second `openssl` and the `mssql-tds` side goes back to needing system OpenSSL, reintroducing exactly the failure this PR is fixing.

The repo already has a convention for this in `mssql-mock-tds/Cargo.toml`:

```toml
[features]
vendored-openssl = ["openssl/vendored"]
```

Adding the same opt-in feature to `mssql-tds` and depending on it here keeps the version in one place and makes the intent explicit:

```toml
mssql-tds = { path = "..", features = ["vendored-openssl"] }
```

If you prefer to keep the direct declaration, dropping the patch pin to `openssl = { version = "0.10", features = ["vendored"] }` at least widens the window in which unification keeps working.

David-Engel (commented):
**Automated review** — 10 added lines in `mssql-tds/fuzz/Cargo.toml`, reviewed against `origin/main` plus the fuzz pipeline templates (`.pipeline/templates/fuzz-template.yml`, `validation-stages.yml`, `scripts/install-deps.sh`) and the `openssl` / `native-tls` wiring in `mssql-tds/Cargo.toml`.

The mechanism works: `openssl = { features = ["vendored"] }` at the same `cfg(all(not(windows), not(target_os = "macos")))` gate unifies with the `openssl` 0.10 that `mssql-tds` already declares, so `openssl-sys` builds from source for the whole fuzz graph, including the copy `native-tls` links. I checked the root-store question too: `native-tls` 0.2.18 calls `openssl_probe::probe()` unconditionally (`src/imp/openssl.rs`, `PROBE_RESULT`), not behind its own `vendored` feature, so a vendored OpenSSL still picks up the agent's CA paths — no trust regression from this change.

Two inline notes: the claim that the fuzz build "no longer depends on those apt-provided system packages" is only partly true (line 23), and the duplicated `0.10.66` pin can silently stop unifying (line 22).

PR-wide (no attachable diff line):
- **No check on this PR exercises the change.** The `Fuzz` stage in `validation-stages.yml` is gated on `RunFuzz == 'true'`, which PR validation does not set — `gh pr checks` confirms build 168731 ran no fuzz job. Since the fuzz build is excluded from PR validation, this class of break can only be caught by the 3-hourly scheduled run. Before taking this out of draft, please attach evidence from a manual `Rust TDS Lib Fuzz Testing` run on this branch (the PR description already flags this as pending), and consider whether a cheap `cargo +nightly fuzz build` smoke job belongs in PR validation so the fuzz crate stops rotting unnoticed.
- The three earlier unresolved threads (root-cause confirmation, unused direct dependency, comment length) still apply; I did not restate them.

No correctness, security, or API concerns in the diff itself. Reviewing is not merging — the merge stays with you.

David-Engel (commented):
## Review: vendor OpenSSL for the fuzz build

The mechanism is correct and I verified it rather than taking it on faith. Running `cargo tree -e features --target x86_64-unknown-linux-gnu -i openssl-sys` against this branch confirms the new direct dependency unifies onto the *same* `openssl-sys v0.9.117` that `native-tls` / `tokio-native-tls` already pull in, and turns on both `openssl-sys/vendored` and `openssl-sys/openssl-src`:

```
openssl-sys v0.9.117
├── openssl-sys feature "default"
│   ├── native-tls v0.2.18 ...
│   └── openssl v0.10.81
│       └── openssl feature "vendored"
│           └── mssql-tds-fuzz v0.0.0
├── openssl-sys feature "openssl-src"
│   └── openssl-sys feature "vendored"
└── openssl-sys feature "vendored"
```

No second `openssl-sys` is introduced, and `openssl-src v300.6.1+3.6.3` is the only new build-script component. I also checked the rest of the Linux fuzz graph for other system-library consumers — `pkg-config` appears only as an `openssl-sys` dependency, and nothing else (`cc`, `libfuzzer-sys`) needs apt-provided headers. So once OpenSSL is vendored, the fuzz build genuinely has no remaining build-time dependency on `libssl-dev`/`pkg-config`. `mssql-tds/fuzz/.gitignore` ignores `Cargo.lock`, so there is no lockfile to update. The target cfg is copied verbatim from `mssql-tds/Cargo.toml`, which is the right call.

**No blocking issues.** Three suggestions and a nit inline, plus one that lives outside the diff:

### The fuzz jobs still hard-fail on a real apt outage (outside the diff)

The fuzz jobs in `.pipeline/templates/validation-stages.yml` run `install-dependencies.yml` before `fuzz-template.yml`, and that template calls `scripts/install-deps.sh`, which retries `apt update` / `apt install` five times and then `exit 1`s:

```bash
if [ "$apt_install_ok" != true ]; then
    echo "ERROR: apt install failed after 5 attempts"
    exit 1
fi
```

Every step in `fuzz-template.yml` is `condition: succeeded()`, so if apt is genuinely down the job goes red at "Install linux dependencies" and never reaches the build that this PR fixes. That is consistent with what the description reports: in build 168173 the "Run Fuzzer" step *did* execute and failed inside `openssl-sys`, which means `install-deps.sh` exited 0 on that agent. So the actual root cause looks less like "apt was down" and more like "`libssl-dev` / `pkg-config` were not effectively present on `RUST-1ES-UBUSLIM` even though the install step reported success" — an image-content problem rather than a mirror outage.

That does not make this change wrong — vendoring is the right way to decouple the fuzz build from the image — but the title and the in-file comment claim broader coverage than the change delivers. Two options: soften the wording to "stop depending on agent-provided OpenSSL", or additionally give the fuzz jobs a leaner dependency path so an apt failure does not red the job before the build. I would also suggest capturing `dpkg -l | grep libssl` (or `pkg-config --modversion openssl`) in the agent info step, so the next occurrence tells you definitively which of the two it was.

### Verification performed

- `cargo tree -e features --target x86_64-unknown-linux-gnu -i openssl-sys` (feature unification, shown above)
- `cargo tree -e normal,build --target x86_64-unknown-linux-gnu` (confirmed no other `*-sys` / pkg-config consumers in the fuzz graph)
- Reviewed `.pipeline/templates/fuzz-template.yml`, `install-dependencies.yml`, `validation-stages.yml`, `scripts/install-deps.sh`, `dev/Dockerfile`, `dev/build-fuzz.sh`, `mssql-tds/fuzz/run-fuzz.sh`
- Confirmed `mssql-tds/fuzz/Cargo.lock` is gitignored, so no lock churn

Holding off on approval only because this is still a draft pending a green GH-Rust Fuzz Test run, which is exactly the right gate for this change.

mssql-tds/fuzz/Cargo.toml
David-Engel: This trades an apt-provided dependency for a different one rather than removing it entirely, and it is worth confirming before you rely on it.

`openssl-src` (which `openssl-sys/vendored` pulls in — `v300.6.1+3.6.3` here) builds OpenSSL by invoking `perl ./Configure` followed by `make`. `scripts/install-deps.sh` installs `build-essential`, so `make`/`gcc` are covered, but `perl` is never in the `DEPS` list. Ubuntu base images ship `perl-base`, which does **not** include several core modules OpenSSL's `Configure` needs, and the agent image here is `RUST-1ES-UBUSLIM` — a slim image is exactly where that gap shows up. The failure mode is a `Can't locate <Module>.pm in @INC` error out of the `openssl-sys` build script, which reads almost identically to the failure you are fixing.

Suggest adding `perl` (and, since `openssl-src` shells out to it, ideally verifying `make` resolves) to the `DEPS` list in `scripts/install-deps.sh`, or confirming the module set on `RUST-1ES-UBUSLIM` before this lands. If the premise is that the agent may not be able to install packages, the vendored path should not quietly require a package that is also only conditionally present.

Not blocking — the fuzz pipeline run you are already waiting on will answer this definitively. Just worth checking the log rather than assuming a green run on one agent generalizes.

mssql-tds/fuzz/Cargo.toml
David-Engel: Consider making the vendored build opt-in rather than unconditional for this target set, because the cost lands on more than just the failing CI job.

Two places pay it today:

1. **Local and dev-container builds.** `dev/build-fuzz.sh` and `mssql-tds/fuzz/run-fuzz.sh` build this same crate, and `dev/Dockerfile` already installs `pkg-config` and `libssl-dev`. Inside that container the vendored compile is pure overhead — a multi-minute OpenSSL build on a machine that has a perfectly good system OpenSSL.
2. **CI breadth.** `validation-stages.yml` defines eight fuzz jobs on `RUST-1ES-POOL-WUS3`, each with `CARGO_TARGET_DIR: $(Build.SourcesDirectory)` and no cargo cache restore in `fuzz-template.yml`. That means OpenSSL 3.6.3 is compiled from source once per job, on every run of the stage.

A cheap way to keep the CI benefit without the local cost:

```toml
[features]
vendored-openssl = ["openssl/vendored"]

[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]
openssl = { version = "0.10.66" }
```

and have `fuzz-template.yml` pass `--features vendored-openssl` on the `cargo +nightly fuzz run` invocation. Developers keep fast incremental builds against system OpenSSL; CI gets the hermetic build.

If you would rather keep it unconditional for simplicity — a defensible choice, since one code path is easier to reason about than two — that is fine, but it is worth a line in the comment noting that local Linux fuzz builds now compile OpenSSL, so the next person hitting a slow first build knows why. `OPENSSL_NO_VENDOR=1` is the escape hatch to mention.

mssql-tds/fuzz/Cargo.toml
David-Engel: Nit, and the comment above already does a good job of explaining the intent, so take or leave this.

The `openssl` version requirement is now stated in two manifests (`mssql-tds/Cargo.toml` line 94 and here), and this crate never actually `use`s `openssl` in any fuzz target — the dependency exists purely to flip a feature on a transitively shared crate. That is a legitimate and common Cargo idiom, but it does mean a future bump in `mssql-tds` leaves a stale-looking `0.10.66` here with nothing to flag the drift. (Harmless today: both are caret requirements and resolve to the same `0.10.81`, as `cargo tree` confirms.)

An alternative that keeps one source of truth is a passthrough feature on `mssql-tds`:

```toml
# mssql-tds/Cargo.toml
[features]
vendored-openssl = ["openssl/vendored"]
```

```toml
# mssql-tds/fuzz/Cargo.toml
mssql-tds = { path = "..", features = ["vendored-openssl"] }
```

This works even though `mssql-tds`'s `openssl` dependency is target-gated, and it drops the version duplication plus the unused direct dependency. It also composes with the opt-in suggestion on the line below. Your call — the current form is explicit and self-documenting, which has its own value.

PR #109, opened 2026-07-14
Add incremental PLP column streaming and sparse row-column read APIs

shiwanigupta0809:
**Description**
This PR adds incremental column read support for PLP values and aligns read behavior with msodbcsql-style streaming semantics.

Key changes:
Updated PLP handling to an ODBC-style pause and resume model inside the existing row-read path, instead of adding new external incremental APIs.
Introduced explicit active row-read state management for:
- idle
- row paused
- PLP paused

When paused on a PLP column, the client now supports:
- incremental PLP byte reads from the active stream
- preserving row continuation state for the same logical row
- automatic internal PLP drain when the caller switches back to next-row flow without fully consuming PLP bytes
- Row-tail handling was tightened so sparse and non-sequential column access remains wire-aligned across multi-column and multi-row results.
- Removed redundant PLP helper layering and simplified end-of-stream checks into the primary status helper.
- Refined token handling in the row-read loop for clearer done, info, env-change, and error behavior while preserving protocol semantics.
- Added explicit TODO markers where paused PLP streaming still needs Always Encrypted support.

Note:
This change keeps current caller-facing PLP semantics intact for row reading. The implementation is focused on internal pause and resume state correctness, safe PLP lifecycle handling, and deterministic behavior when mixing PLP and non-PLP access patterns.

**Tests**
Integration and unit coverage now validates:
- sparse two-row access patterns that mix PLP and non-PLP columns
- pause and resume behavior across row boundaries
- automatic PLP drain behavior when transitioning back to row mode
- no-active-stream PLP helper contracts
- paused PLP state retaining collation metadata
- small-step lifecycle correctness for PLP state transitions without regressing existing row-read behavior

**Verification**
cargo bfmt, cargo bclippy, cargo btest passes and all e2e tests added in this branch passes

## Related Issues
https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46138/

PR #359, opened 2026-08-22
Remove PLP chunk-count cap that limited LOB reads to ~763 MB

### LINKED ISSUE #358: Reading a varbinary(max)/LOB larger than ~763 MB fails with "Too many PLP chunks" and permanently wedges the connection
### Problem statement

Reading a large LOB (`varbinary(max)`, `nvarchar(max)`, etc.) fails once the value exceeds roughly 763 MB, even though TDS and SQL Server both allow 2 GB. This was originally reported as `mssql-tds` "hanging or becoming unresponsive" while reading 2 GB from the server.

`GenericDecoder::MAX_PLP_CHUNKS` caps a PLP value at 100 000 chunks. SQL Server emits `varbinary(max)` on the wire in **8000-byte** PLP chunks, so the effective ceiling is:

```
100_000 chunks x 8000 bytes = 800_000_000 bytes (~763 MB)
```

Anything larger is rejected with `Protocol Error: Too many PLP chunks: 100001 (max 100000)`.

Measured against SQL Server 2022 (16.0.4252.3), reading `SELECT CONVERT(varbinary(max), REPLICATE(CONVERT(varchar(max), 'a'), N))`:

| size | result |
|---|---|
| 1 MB / 16 MB / 64 MB / 256 MB | OK |
| 512 MB | OK (8.1 s) |
| 768 MB | `Too many PLP chunks: 100001 (max 100000)` |
| 1 GB | same |
| 2 GB | same |

### Why it presents as a hang

The cap fires **mid-value**, while the server is still streaming the remaining hundreds of megabytes. The driver surfaces the error but leaves the reader desynchronized inside the LOB payload, and the batch is never closed. Every subsequent call on that connection fails:

```
close_query() -> Err(ProtocolError("Unknown token type: 0x61"))   // LOB payload bytes parsed as tokens
execute(...)  -> Err(UsageError("There is an open batch on the current connection. ..."))
```

There is no path back to a usable connection. An application that catches the error and retries loops on failures indefinitely, which is what "hangs / becomes unresponsive" looks like from the outside.

### Origin

The constant was not a considered production limit:

1. `b275cea3` (2025-11-12 02:04) — "Fix OOM vulnerability in Text/NText data type decoding" introduced `MAX_PLP_CHUNKS` **and** its check, but the check was scoped `#[cfg(fuzzing)]`, so production was unaffected. The `#[cfg(not(fuzzing))]` value of `100000` was only the inactive arm of the constant.
2. `849c6168` (2025-11-12 02:11, seven minutes later) — "Fix clippy warnings" included the bullet *"Move PLP chunk count validation outside fuzzing-only block"*, turning a fuzz-only tripwire into a hard production ceiling amid format-string churn. Reached `main` via PR #61.
3. `c19d2dde` (PR #109) — the PLP streaming refactor replicated the now-production check into `PlpChunkStreamReader::ensure_active_chunk`, so it fires on both the buffered and streaming paths today.

### Proposed solution

Remove the chunk-**count** cap and keep the byte-based limits.

The count cap is redundant. All three PLP loops consume a strictly positive `chunk_len` per iteration and are each already bounded by a total-byte check:

- `read_plp_chunks_unknown_len` — `vector_capacity > MAX_PLP_SIZE`
- `read_plp_chunks_into_slice` — `end_offset > declared_len || end_offset > dest.len()`
- `PlpChunkStreamReader::ensure_active_chunk` — `next_total > MAX_PLP_SIZE`

So the work a peer can induce is already proportional to the bytes it actually sends; a chunk count limit adds no protection. `MAX_PLP_CHUNK_SIZE` (16 MB) still bounds any single allocation, and `MAX_PLP_SIZE` (2 GB) still bounds the total, which is what the original OOM hardening in `b275cea3` was actually about.

Note this overlaps the first bullet of #258 (extract a shared `PlpChunkValidator`) — that cleanup should drop `chunks_seen` rather than preserve it.

### Affected crate

mssql-tds

### Additional context

The desync-after-mid-value-error behavior described above is a separate defect that this ceiling merely exposes. Any error raised partway through a LOB leaves the connection unusable; that is worth tracking on its own.

saurabh500:
## Description

Removes `MAX_PLP_CHUNKS`, which capped a PLP value at 100 000 chunks. SQL Server emits `varbinary(max)` in **8000-byte** chunks on the wire, so this was an effective hard ceiling of ~763 MB on any LOB read — well below the 2 GB TDS allows. Larger values failed with `Protocol Error: Too many PLP chunks: 100001 (max 100000)`.

Because the cap fires *mid-value* while the server is still streaming, it also left the reader desynchronized inside the LOB payload and the batch permanently open, so the connection could not be recovered (`close_query()` → `Unknown token type: 0x61`, then every call → `open batch` usage error). That is what the original "hangs / becomes unresponsive on a 2 GB read" report looked like from the application side.

### Why the cap was safe to remove

It was never a considered production limit. It was introduced in `b275cea3` as a `#[cfg(fuzzing)]`-only tripwire, then hoisted into production builds seven minutes later by `849c6168` ("Fix clippy warnings" → *"Move PLP chunk count validation outside fuzzing-only block"*). `c19d2dde` (#109) later replicated it into the streaming reader.

It is also redundant. All three PLP loops consume a strictly positive `chunk_len` per iteration and are each already bounded by a total-byte check:

| loop | existing bound |
|---|---|
| `read_plp_chunks_unknown_len` | `vector_capacity > MAX_PLP_SIZE` |
| `read_plp_chunks_into_slice` | `end_offset > declared_len \|\| end_offset > dest.len()` |
| `PlpChunkStreamReader::ensure_active_chunk` | `next_total > MAX_PLP_SIZE` |

Induced work is therefore already proportional to bytes actually sent. The byte-based limits are untouched: `MAX_PLP_CHUNK_SIZE` (16 MB) still bounds any single allocation and `MAX_PLP_SIZE` (2 GB) still bounds the total — which is what the original OOM hardening in `b275cea3` was actually protecting.

`chunks_seen` is dropped from `PlpChunkStreamReader`, and the fuzzing-only `chunk_count` diagnostic counters are now `#[cfg(fuzzing)]`-scoped alongside the `eprintln!` they feed.

## Tests

**Unit tests** (`mssql-tds/src/datatypes/decoder.rs`) — three new tests drive `100_001` single-byte chunks through each affected path and assert the value decodes:

- `read_plp_bytes_accepts_more_chunks_than_the_removed_count_cap` (unknown-length)
- `read_plp_bytes_accepts_many_chunks_for_a_known_length_value` (known-length)
- `plp_chunk_stream_reader_accepts_more_chunks_than_the_removed_count_cap` (streaming)

`plp_chunk_stream_reader_chunk_count_limit_errors` is replaced by the streaming test above. The tests that assert the byte-based limits still fire (`..._accumulated_size_limit_errors`, `read_plp_bytes_rejects_oversized_chunk_in_existing_path`) are unchanged and still pass.

**Integration tests** (`mssql-tds/tests/test_large_value_read.rs`, new, `#[ignore]`d — they move gigabytes):

- `large_varbinary_sizes_are_readable_up_to_the_server_limit`
- `two_gigabyte_varbinary_is_readable`

Verified against SQL Server 2022 (16.0.4252.3):

| size | before | after |
|---|---|---|
| 512 MB | OK — 7.4 s | OK — 7.4 s |
| 768 MB | ❌ `Too many PLP chunks` | OK — 11.0 s |
| 1 GB | ❌ `Too many PLP chunks` | OK — 17.4 s |
| 2 GB | ❌ `Too many PLP chunks` | OK — **2 147 483 647 bytes in 31.8 s** |

## Related Issues

Fixes #358

Overlaps the first bullet of #258 (extract a shared `PlpChunkValidator`) — that cleanup should now drop `chunks_seen` rather than preserve it.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes — 1863/1870 `mssql-tds` unit tests pass; the 7 failures are pre-existing expired-test-certificate failures in `certificate_validator` / `win_tls::validate`, confirmed identical on a clean tree
- [x] Public API changes are documented — no public API change (`MAX_PLP_CHUNKS` was a private associated const)

### DIFF
--- a/mssql-tds/src/datatypes/decoder.rs
+++ b/mssql-tds/src/datatypes/decoder.rs
@@ -297,7 +297,6 @@ pub(crate) struct PlpChunkStreamReader {
     chunk_remaining: usize,
     reached_end: bool,
     total_read: usize,
-    chunks_seen: u32,
 }
 
 #[cfg_attr(not(test), allow(dead_code))]
@@ -308,7 +307,6 @@ impl PlpChunkStreamReader {
             chunk_remaining: 0,
             reached_end: false,
             total_read: 0,
-            chunks_seen: 0,
         }
     }
 
@@ -383,15 +381,6 @@ impl PlpChunkStreamReader {
                 )));
             }
             return Ok(false);
-        }
-
-        self.chunks_seen += 1;
-        if self.chunks_seen > GenericDecoder::MAX_PLP_CHUNKS {
-            return Err(crate::error::Error::ProtocolError(format!(
-                "Too many PLP chunks: {} (max {})",
-                self.chunks_seen,
-                GenericDecoder::MAX_PLP_CHUNKS
-            )));
         }
 
         if chunk_len > GenericDecoder::MAX_PLP_CHUNK_SIZE {
@@ -612,10 +601,6 @@ impl GenericDecoder {
     const SQL_PLP_UNKNOWNLEN: usize = 0xfffffffffffffffe;
     #[cfg_attr(not(test), allow(dead_code))]
     const SQL_PLP_MAXLEN: usize = 0xfffffffffffffffd;
-    #[cfg(fuzzing)]
-    const MAX_PLP_CHUNKS: u32 = 1000;
-    #[cfg(not(fuzzing))]
-    const MAX_PLP_CHUNKS: u32 = 100000;
     #[cfg(fuzzing)]
     const MAX_PLP_CHUNK_SIZE: usize = 8 * 1024;
     #[cfg(not(fuzzing))]
@@ -1192,19 +1177,19 @@ impl GenericDecoder {
         let mut vector_capacity: usize = 0;
         let mut chunk_len = read_sync_first!(reader, try_read_uint32, read_uint32) as usize;
         let mut offset: usize = 0;
+        #[cfg(fuzzing)]
         let mut chunk_count = 0u32;
 
         while chunk_len > 0 {
-            chunk_count += 1;
-
             #[cfg(fuzzing)]
             {
+                chunk_count += 1;
                 eprintln!(
                     "[ALLOC] read_plp_chunks_unknown_len: chunk #{chunk_count}, chunk_len={chunk_len}, total_capacity={vector_capacity}"
                 );
             }
 
-            Self::check_plp_chunk(chunk_count, chunk_len)?;
+            Self::check_plp_chunk(chunk_len)?;
 
             // Use checked_add to prevent capacity overflow
             vector_capacity = vector_capacity.checked_add(chunk_len).ok_or_else(|| {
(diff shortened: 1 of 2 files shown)


saurabh500:
Superseded by #360, which carries this same decoder fix plus the end-of-message tracking that addresses the actual hang, the close_query desync fix for #358, and 
eset_reader hardening.

PR #360, opened 2026-08-22
Fix large PLP read hangs: remove chunk-count cap and track end-of-message

### LINKED ISSUE #358: Reading a varbinary(max)/LOB larger than ~763 MB fails with "Too many PLP chunks" and permanently wedges the connection

saurabh500:
## Description

Reading a LOB larger than ~763 MB failed with `Too many PLP chunks` and left the connection permanently unusable. Originally reported as `mssql-tds` "hanging or becoming unresponsive" while reading 2 GB from the server.

Sections 1–4 are the reported failure and its fallout. Sections 5–6 are the same defect class found while auditing the surrounding code: panics on peer-supplied state.

### 1. Remove the PLP chunk-**count** cap (`mssql-tds/src/datatypes/decoder.rs`)

`MAX_PLP_CHUNKS` limited a PLP value to 100 000 chunks. SQL Server emits `varbinary(max)` on the wire in **8000-byte** chunks, so the real ceiling was `100_000 × 8000 = 800 MB` — far below the 2 GB that TDS and SQL Server both allow.

It was never a considered production limit:

1. `b275cea3` ("Fix OOM vulnerability in Text/NText data type decoding") added the constant **and** its check, but scoped the check `#[cfg(fuzzing)]`. Production was unaffected.
2. `849c6168` ("Fix clippy warnings"), **seven minutes later**, moved the check outside that block amid format-string churn — turning a fuzz-only tripwire into a hard production ceiling.
3. `c19d2dde` (#109) replicated it into `PlpChunkStreamReader::ensure_active_chunk`, so it fired on the streaming path too.

The cap is also redundant: every PLP loop consumes a strictly positive `chunk_len` per iteration and is already bounded by a total-byte check, so the work a peer can induce is proportional to the bytes it actually sends. `MAX_PLP_CHUNK_SIZE` (16 MB) still bounds any single allocation and `MAX_PLP_SIZE` (2 GB) still bounds the total — which is what the original OOM hardening was actually about.

### 2. Track end-of-message so short reads fail instead of blocking (`network_transport.rs`, `buffers.rs`)

This is the actual hang.

`read_bytes`, `read_bytes_uninit` and `skip_bytes` refill by calling `read_tds_packet()` whenever the buffer is short, and never recorded whether the message had already ended. The EOM flag *was* parsed in `get_new_tds_packet()` — but only to validate payload-free packets, and then thrown away as a discarded local.

A server that finishes its message on a packet carrying payload — the normal case — then goes quiet until the client issues a new request. A reader still short of its target walked straight back into the socket and **blocked there forever**.

The fix persists the flag on `TdsReadBuffer` and checks, *before* re-entering the socket, whether the outstanding demand exceeds what remains in the final packet of the message. If it does, no amount of waiting will help, so it fails with a `ProtocolError` naming the outstanding and available byte counts.

### 3. Handle a refill that yields zero payload (same three loops) — invariant guard

The three loops also had no check for a refill that yielded zero payload. Each now handles `to_read == 0` explicitly:
if the message has ended, a read that already consumed part of its value fails (section 2's rule), and one that has
not yet started refills and carries on — matching what the scalar readers do at the same boundary.

**That "carry on" path is itself capped.** Tolerating one payload-free end-of-message packet is necessary: a read
beginning exactly at a message boundary consumes the packet terminating the *previous* message before its own data
arrives. But the `continue` re-enters the loop with nothing changed, so a peer sending two empty messages in a row
spins with no state advancing and no bytes delivered — bounded only by the command timeout. A second consecutive
payload-free message is therefore rejected. No legitimate server sends two back to back, and the attention
acknowledgement carries a DONE token, so it is not payload-free either. No counter reset is needed: once a refill
yields payload, the `continue` branch is unreachable for the rest of the call.

The remaining `!end_of_message` arm returns a `ProtocolError`, and is **unreachable as the framing layer stands**:
`get_new_tds_packet` rejects a payload-free packet unless it is EOM, and `read_tds_packet` always frames a packet on
a successful refill, so an empty buffer after a refill implies the message ended. It is kept as a guard on that
invariant — if framing ever tolerates a payload-free non-EOM packet, the alternative is a silent infinite loop, which
is the exact hang this PR removes. It is documented as unreachable at the definition rather than left to be
reconstructed by the next reader.
### 4. Reset client state and retire the connection when a drain fails (`tds_client.rs`) — #358
`close_query()` now captures the drain result, resets client state unconditionally, and *then* returns the error.
This mirrors the existing behaviour for an aborted streamed RPC write.

The same rule applies at **every** drain site, not just `close_query`. Both ERROR-token handlers build a
`SqlServerError` even when their own drain fails, so that error does not imply a synchronized stream; they now retire
where they give up. Seven further sites (`execute_sp_prepare`, `execute_sp_unprepare`, and five cursor operations)
previously propagated a failed drain with a bare `?`, leaving the connection reusable *and* the batch open — which
masks the real error behind `ALREADY_EXECUTING_ERROR`. All now route through a single `drain_stream_or_retire()`
helper, so the rule lives in one place instead of being re-derived per call site.

**The remedy depends on the cause.** A drain can fail two ways, and they are not interchangeable:

| Cause | Errors | Transport marked dead | Session recovery |
|---|---|---|---|
| Stream desync | `ProtocolError`, `TimeoutError`, `OperationCancelledError` | yes | **disarmed** |
| Transport loss | `Io`, `ConnectionClosed`, `ConnectionError`, `TlsError`, `TlsHandshakeError` | yes | **left armed** |

A desynchronized stream parks the reader at an unknown offset in a response that is otherwise fine, and a TDS stream
cannot be resynchronized from there — but the socket may be perfectly healthy, so leaving recovery armed would let the
next command silently reconnect and paper over a decoder bug. A socket that dropped mid-drain has no offset left to
recover, and reconnecting is exactly the remedy connection resiliency exists for; disarming it there would turn a
network blip into a hard failure. These sites include routine cursor open/close/fetch, not just exceptional paths.

Scope of retirement: marking the transport dead is what a pool consults, so **pooled consumers discard the connection**
in both cases. It does not close the socket, and no execute path consults `is_connection_dead`, so a caller reusing the
`TdsClient` directly can still issue another command. That command is not parked mid-value — entry points call
`reset_reader()` — but stale bytes may remain queued in the kernel receive buffer. Closing the socket here is the wrong
remedy: `close_transport()` writes a TLS close_notify on a socket we just failed to read from, and its `?` returns
before `self.stream = None`, so a swallowed error would leave the stream alive in exactly the degraded case retirement
exists for. Dropping the stream with no I/O is the fix; it needs a new transport method and is left as follow-up.
### 5. Stop `reset_reader` from aborting the process (`network_transport.rs`)

`TdsPacketReader::reset_reader` asserted that the read buffer was fully consumed. `buffer_length` is taken from the packet header the peer sent, so that placed a peer-influenced value behind an `assert!` that stays active in release builds.

It is not reachable from `TdsClient` today (that path goes through `TdsTransport::reset_reader`, which has no assert, and `command_is_busy()` short-circuits first). The leftover bytes are now logged and discarded — which is what the reset already did, so correct paths are unchanged.

### 6. Return errors instead of panicking on malformed packet and type metadata (`buffers.rs`, `sqldatatypes.rs`)

Auditing for more of the same turned up three further panic sites on the response-parsing path.

**Why this matters more than usual here:** `mssql-tds` is loaded into ODBC, Node and Python host processes. A panic does not fail a query — it unwinds out of a process that has no idea what happened. (`mssql-odbc` does wrap its entry points in `catch_unwind` via `ffi_entry!`, which converts these to `SQL_ERROR`, but relying on that net for ordinary protocol faults is the wrong default.)

`buffers.rs`:

- `consume_bytes` called `panic!("Not enough data to consume")` when asked for more than the buffer held. It now returns a `ProtocolError` naming the shortfall. Every caller already checked capacity first, so they simply propagate with `?`.
- `do_we_have_enough_data` and `get_remaining_byte_count` computed `buffer_length - buffer_position` directly. `buffer_position` must never pass `buffer_length`, but if it ever did, that subtraction **wraps to a near-`usize::MAX` value in release builds** — making every capacity check pass and sending callers slicing past the end of the buffer. Both now saturate, so a broken invariant degrades to "no bytes available" instead of "unlimited bytes available".

`sqldatatypes.rs`:

- `read_type_info` had two `unreachable!` arms reached through server-supplied type bytes; both now return `ProtocolError`.
- The PLP conversion also called `PartialLengthType::try_from(...).unwrap()`. The two near-identical match arms are folded into one that derives the collation, which removes both the duplication and the unwrap.

## Verification

Measured against SQL Server 2022 (16.0.4252.3) in Docker, `SELECT CONVERT(varbinary(max), REPLICATE(CONVERT(varchar(max), 'a'), N))`, release build:

| size | before | after |
|---|---|---|
| 1 MB | OK | OK — 0.05 s |
| 64 MB | OK | OK — 0.66 s |
| 256 MB | OK | OK — 2.25 s |
| 761 MB (just under old ceiling) | OK | OK — 7.39 s |
| 763 MB (just over) | `Too many PLP chunks: 100001` | **OK — 7.98 s** |
| 1 GB | `Too many PLP chunks: 100001` | **OK — 9.07 s** |
| **2 GB** | `Too many PLP chunks: 100001` | **OK — 18.02 s** (2 147 483 647 bytes) |

A 5 s command timeout on a large read still returns promptly (5.07 s).

Every new check was verified by reverting it and confirming the corresponding test fails or hangs, then restoring it:

- Sections 2 and 3: with section 2's check disabled but section 3's still active, all three `*_without_a_trailing_packet` tests hang for their full 5 s timeout — the empirical proof that the no-progress check alone does not cover the realistic case.
- Section 4: with the original `?` restored, `close_query_drain_failure_closes_batch_and_retires_connection` fails on the very first assertion (`a failed drain must still close the batch`).
- Section 5: with the original `assert!` restored, `reset_reader_discards_unread_bytes_instead_of_aborting` panics inside `reset_reader`.

## Tests

- `mssql-tds/tests/test_large_value_read.rs` (new, `#[ignore]`d — needs a live server):
  - `large_varbinary_sizes_are_readable_up_to_the_server_limit` — sweeps either side of the old ceiling
  - `two_gigabyte_varbinary_is_readable` — the reported case
  - `command_timeout_during_a_large_read_returns_promptly` — covers the timeout path
- Unit tests in `decoder.rs` proving both the buffered and streaming paths accept more chunks than the removed cap allowed, on known- and unknown-length values.
- Unit tests in `network_transport.rs`:
  - `*_past_end_of_message_without_a_trailing_packet_*` (×3) — the realistic shape: a complete message with payload and nothing following. These hang without section 2's fix.
  - `*_past_end_of_message_errors_instead_of_hanging` (×3) — the explicit-empty-EOM shape.
  - `bulk_reads_still_span_packets_and_tolerate_a_trailing_empty_eom` and `a_value_ending_exactly_at_end_of_message_reads_successfully` — assert neither check fires on legitimate traffic.
  - `reset_reader_discards_unread_bytes_instead_of_aborting` — section 5.
- Unit tests in `tds_client.rs`:
  - `close_query_drain_failure_closes_batch_and_retires_connection` — a truncated response must still close the batch, must not report `ALREADY_EXECUTING_ERROR`, and must retire the connection.
  - `close_query_success_leaves_connection_alive` — the control: a complete drain closes the batch and keeps the connection usable.
- Unit tests in `buffers.rs`:
  - `test_consume_bytes_over_errors` — replaces the old `#[should_panic]` test; asserts the error names the shortfall and that a rejected consume does not move the position.
  - `remaining_byte_count_saturates_when_position_leads_length` — drives the broken invariant directly and asserts it degrades to zero available rather than wrapping.
- `TestPacketBuilder::continuation()` (new): the builder sets EOM on *every* packet by default. A multi-packet message must clear it on all but the last, so the spanning-read test had been relying on framing no real server would send.
- `create_network_transport_with_live_peer` helper: the existing helpers drop the writer, so a starved reader hits EOF and errors out regardless of whether the loop is correct — which would mask exactly this class of bug.
- `ByteStreamTransport` now implements `connection_known_dead`/`mark_known_dead`; the trait defaults are a no-op, so tests could not otherwise observe retirement.

The two `sqldatatypes.rs` arms have no new test: both are unreachable with the current type table, and reaching them would require changing that table rather than crafting a packet. They are converted so that a future type addition returns an error instead of taking the process down.

Run the live tests with:

```
cargo nextest run -p mssql-tds --release --test test_large_value_read --run-ignored all
```

## Not addressed here

- `read_tds_packet` is not cancel-safe: a timeout or cancel dropping the future mid-packet desyncs the stream. The post-timeout attention-ACK drain is also unbounded by construction, though it did *not* reproduce against a live server.
- The workspace `[workspace.lints.clippy]` block in `Cargo.toml` has a *"Don't Panic"* section — `unwrap_used`, `panic`, `unreachable`, `panic_in_result_fn`, `indexing_slicing` — with **every lint commented out**. Turning them on is what would stop sections 5 and 6 recurring, but it needs its own audit pass and does not belong in this PR.

## Related Issues

Fixes #358

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes (workspace and `mssql-py-core`)
- [x] `cargo btest` passes — 1882 unit tests, 1875 pass; the 7 failures are pre-existing missing cert fixtures (`tests/test_certificates/valid_cert.pem`), identical on a clean baseline


mssql-tds/src/connection/tds_client.rs
```diff
@@ -4970,6 +4983,22 @@ impl TdsClient {
...
+            // rather than let the next command read garbage.
+            //
+            // The state above is still cleared first: leaving the batch open
+            // would mask this error behind `ALREADY_EXECUTING_ERROR` on every
+            // later call, which is what made the original failure look like a
+            // permanent wedge.
+            self.transport.mark_known_dead();
+            self.recovery_context.session_recovery_negotiated = false;
+            return Err(error);
+        }
```
saurabh500, line 5303: **Blocking — this retires healthy connections on ordinary SQL errors.**

The comment's premise ("the reader is parked at an unknown offset") isn't true for a large class of drain errors. `advance()` returns `Err(Error::from_sql_errors(...))` only *after* reading the terminal DONE (tds_client.rs:5264):

```rust
if !done.has_more() {
    if !collected_errors.is_empty() { return Err(Error::from_sql_errors(collected_errors)) }
```

At that point the stream is fully drained and perfectly synchronized — but this block kills the connection anyway.

Confirmed with a probe: feeding a well-formed `ERROR(1222, 16, "boom")` followed by a terminal `DONE(DONE_ERROR)` through `close_query()` returns `SqlServerError` **and** leaves `connection_known_dead == true` / `session_recovery_negotiated == false`.

Impact: a batch like `SELECT 1; RAISERROR('boom', 16, 1)` where the caller stops early poisons a pooled connection every time. `drain_on_error_consumes_trailing_rowset_and_surfaces_sql_error` (tds_client.rs:8463) shows this is a normal, exercised path.

Retire only for desync-class errors:

```suggestion
        if let Err(error) = drain_result {
            // A drain that stops partway leaves the reader parked at an unknown
            // offset — typically inside a value's payload, where the next bytes
            // would be decoded as tokens. Nothing can resynchronize a TDS stream
            // from there, so retire the connection rather than let the next
            // command read garbage.
            //
            // A server-reported error is different: `advance` only surfaces one
            // after reading the terminal DONE, so the stream is fully drained and
            // still synchronized. Retiring there would poison a healthy connection.
            //
            // The state above is still cleared first: leaving the batch open
            // would mask this error behind `ALREADY_EXECUTING_ERROR` on every
            // later call, which is what made the original failure look like a
            // permanent wedge.
            if !matches!(error, crate::error::Error::SqlServerError { .. }) {
                self.transport.mark_known_dead();
                self.recovery_context.session_recovery_negotiated = false;
            }
            return Err(error);
        }
```

Worth adding a test asserting that a `SqlServerError` surfaced by the drain leaves the connection alive.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1179,15 +1233,15 @@ impl TdsPacketReader for NetworkTransport {
             self.read_tds_packet().await?;
         }
         let result = BigEndian::read_i16(self.tds_read_buffer.get_slice());
-        self.tds_read_buffer.consume_bytes(2);
+        self.tds_read_buffer.consume_bytes(2)?;
```
saurabh500: **Blocking — the end-of-message check only reached the three bulk loops.**

`message_cannot_satisfy()` was added to `read_bytes` (:1337), `read_bytes_uninit` (:1383) and `skip_bytes` (:1483). Every scalar reader still has the bare `loop { try_read; read_tds_packet().await? }` and blocks forever past end-of-message:

| line | fn | line | fn |
|---|---|---|---|
| 1222 | `read_byte` | 1257 | `read_float32` |
| 1231 | `read_int16_big_endian` | 1265 | `read_float64` |
| 1239 | `read_int32_big_endian` | 1273 | `read_int16` |
| 1248 | `read_uint40` | 1323 | `read_uint64` |

Confirmed: `read_byte()` one byte past end-of-message against a live peer returns `Err(Elapsed(()))` after a 3s timeout — it still hangs.

This matters for this PR specifically. `read_byte` is what the token-stream reader uses to read token types, which is exactly the path `close_query()`'s drain walks. Over a real socket, a drain after a mid-value error hangs in `read_byte`, so the new retire logic never runs. The unit test passes only because `ByteStreamTransport` returns EOF instead of going quiet like a real server would.

Same shape as the bulk loops, applied here:

```suggestion
            if self.message_cannot_satisfy(2) {
                return Err(Self::past_end_of_message_error(
                    2,
                    self.tds_read_buffer.get_remaining_byte_count(),
                ));
            }
            self.read_tds_packet().await?;
        }
        let result = BigEndian::read_i16(self.tds_read_buffer.get_slice());
        self.tds_read_buffer.consume_bytes(2)?;
```

...and the same for the other seven. A small helper (`ensure_message_can_satisfy(n)?`) called at the top of each would keep it to one line per site.

Related, minor: the `consume_bytes(2)?` / `consume_bytes(4)?` / `consume_bytes(8)?` conversions on these lines are dead code today — the `while !do_we_have_enough_data(N)` loop above already guarantees capacity, so the `?` can never fire. It hangs instead of erroring. They only become meaningful once the check above exists.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -1192,19 +1177,19 @@ impl GenericDecoder {
...
             #[cfg(fuzzing)]
             {
+                chunk_count += 1;
                 eprintln!(
                     "[ALLOC] read_plp_chunks_unknown_len: chunk #{chunk_count}, chunk_len={chunk_len}, total_capacity={vector_capacity}"
                 );
             }
 
-            Self::check_plp_chunk(chunk_count, chunk_len)?;
+            Self::check_plp_chunk(chunk_len)?;
```
saurabh500, line 1191: **Suggestion** — dropping the chunk-count cap is the right call, and I verified nothing was lost: the in-loop `MAX_PLP_SIZE` checks at :400 and :1201 still bound total bytes, and `read_plp_chunks_into_slice` is bounded by `dest.len()`. The rationale in the doc comment is accurate.

It does make the allocation pattern a few lines below actually reachable, though. `plp_buffer.resize(vector_capacity, 0)` runs once per chunk on this unknown-length path. SQL Server emits `varbinary(max)` in 8000-byte chunks, so a 2 GB value now means roughly 268k `resize` calls, and peak RSS during `Vec` doubling is about 3x the value size — on a value that can itself be 2 GB.

Reallocation is amortized O(1) so this isn't a correctness problem, but hoisting the growth out of the loop (or reserving geometrically and resizing once at the end) would avoid repeatedly zero-filling memory that's about to be overwritten by the next chunk.

mssql-tds/src/connection/transport/buffers.rs
```diff
@@ -59,7 +74,7 @@ impl TdsReadBuffer {
         let bytes = self.working_buffer[position..position + N]
             .try_into()
             .expect("slice length is fixed by N");
-        self.consume_bytes(N);
+        self.consume_bytes(N).ok()?;
```
saurabh500: **Nit** — `.ok()?` is unreachable here: the `do_we_have_enough_data(N)` check a few lines above already proves `consume_bytes(N)` succeeds.

It's also a slightly lossy conversion. If it ever *could* fail, `try_read_array` would report "need more data" and the caller would loop waiting for bytes that never arrive — turning a protocol error into a hang, which is the exact failure mode this PR is fixing elsewhere. A `debug_assert!` or a one-line comment recording the precondition would keep that intent from eroding.

saurabh500 (commented):
## Review: solid fix, two things to sort out first

The core change is correct and well-evidenced. Removing `MAX_PLP_CHUNKS` is safe — I verified the in-loop total-byte bounds it was masking are intact (`decoder.rs:400` and `:1201` against `MAX_PLP_SIZE`; `read_plp_chunks_into_slice` is bounded by `dest.len()`), and the `#[cfg(fuzzing)]` attributes were preserved correctly. Tracking end-of-message is the right fix for the hang.

Two issues, both confirmed by running probes rather than by inspection.

**Blocking**

1. `close_query()` retires healthy connections on ordinary SQL errors. `advance()` only surfaces a `SqlServerError` *after* reading the terminal DONE, so the stream is fully drained and synchronized at that point. Verified: a well-formed ERROR + terminal DONE leaves `connection_known_dead == true`.
2. The end-of-message check reached only the three bulk loops. All eight scalar readers still block forever past EOM (verified: `read_byte` timed out). Because `read_byte` drives the token-stream reader, `close_query()`'s drain can hang over a live socket, so finding 1's retire logic would never run.

**Suggestion** — the per-chunk `resize` in `read_plp_chunks_unknown_len` is now genuinely reachable at multi-GB sizes.

**Nit** — unreachable `.ok()?` in `try_read_array`.

### Verified correct, not flagged

- `end_of_message` cannot go stale across messages: both `reset_reader` impls route through `reset_to_length(0)` which clears it, and every command entry point in `tds_client.rs` calls `reset_reader()` before sending.
- `to_read == 0` really does imply `available == 0` in all three loops (`length_to_read > 0` and `max_packet_size - 8 > 0`), so the no-progress error is sound.
- `message_cannot_satisfy` correctly accounts for coalesced `pending_bytes`, since EOM=1 means no further packets in the message.

### Local validation

1875/1882 `mssql-tds` lib tests pass on `10185b2e`. The 7 failures are all `certificate_validator` / `win_tls::validate` reporting `CertificateNotFound { path: "tests/test_certificates/valid_cert.pem" }` — an unrelated fixture-generation issue in a fresh worktree, not caused by this PR.

Correctly linked to #358 and the checklist is consistent with CI. Posting as a comment rather than "request changes" since this is my own PR.

mssql-tds/src/connection/tds_client.rs
saurabh500: Confirmed and fixed in 9a4a7794. You're right — `advance_to_rows` only returns `from_sql_errors` after `!done.has_more()` (tds_client.rs:5264), so the stream is fully drained and synchronized at that point; retiring there poisoned a healthy pooled connection on any batch ending in `RAISERROR`. Applied your suggestion: retirement is now gated on `!matches!(error, Error::SqlServerError { .. })`, with the state reset still unconditional.

Also added `close_query_drain_sql_error_leaves_connection_alive` — feeds `ERROR(1222)` + terminal `DONE(DONE_ERROR)` through `close_query()` and asserts the error still surfaces while `connection_known_dead()` stays false.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Fixed in 9a4a7794, though the check needed narrowing to be safe — worth flagging what I found.

All thirteen scalar readers now route their refill through one helper, `refill_for(n)`, so it's a one-line change per site as you suggested.

Applying `message_cannot_satisfy` verbatim to them broke the connection handshake: **353 test failures**, all `ProtocolError("Failed to receive token during login response parsing.")`. `end_of_message` is per-packet state that survives the end of a message, so the *first* `read_byte` of the next response — token type, empty buffer, EOM still set from the previous message — errored instead of framing the new message. The bulk loops never hit this only because they're never the first read of a message.

So `refill_for` fires only when part of the value is already buffered:

```rust
let available = self.tds_read_buffer.get_remaining_byte_count();
if available > 0 && self.message_cannot_satisfy(needed) {
    return Err(Self::past_end_of_message_error(needed, available));
}
```

A truncated value (`available > 0`) is unambiguously unsatisfiable. A read starting on an empty buffer is ambiguous — a caller beginning the next response looks identical to an over-read — and the transport has no way to tell them apart today. That means `read_byte` exactly on a boundary can still block; catching it needs the reader to know a request was sent since the message ended, which is a bigger change than this PR. Filing that separately.

Full-suite diff against a stashed baseline: identical failure sets (52 pre-existing, all cert fixtures / named-pipe / shared-memory), plus live 2 GB read still passing.

Three pre-existing tests were spanning a value across two *EOM* packets — two separate messages — and only passed because nothing enforced framing. Switched them to `.continuation()` so they test what their names say. Added `read_uint32_spanning_end_of_message_errors_instead_of_hanging`.

Good catch on the `consume_bytes(N)?` conversions being unreachable — same observation as the `buffers.rs` thread, handled there.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Fixed in 9a4a7794. The loop now reads straight into the buffer's spare capacity instead of growing it with `resize`:

```rust
plp_buffer.reserve(chunk_len);
let read = reader
    .read_bytes_uninit(&mut plp_buffer.spare_capacity_mut()[..chunk_len])
    .await?;
```

That drops the redundant zero-fill entirely — previously a full value's worth of `memset` immediately overwritten by the next read. It mirrors `read_plp_chunks_into_slice` a few lines below, which already reads into `MaybeUninit`. The single `set_len` is guarded by an explicit `read != chunk_len` check, so a reader implementation that under-reports can't expose uninitialized bytes.

Live re-verified: 2 GB `varbinary(max)` reads in 38.7 s, and the size sweep up to the server limit passes.

Not addressed here: peak RSS is still ~3x during growth, which is inherent to a growable buffer on the unknown-length path where no total is declared. Avoiding it needs streaming into a caller-supplied sink rather than materializing the value — tracked as a follow-up alongside the single-2 GB-allocation item on the known-length path.

> [!NOTE]

mssql-tds/src/connection/transport/buffers.rs
saurabh500: Agreed, and recorded in 9a4a7794:

```rust
let consumed = self.consume_bytes(N);
debug_assert!(consumed.is_ok(), "capacity is checked at the top");
consumed.ok()?;
```

Kept the `.ok()?` so release behavior is unchanged, with the assertion catching any future edit that breaks the precondition. The comment spells out the failure mode you described — reporting "need more data" for what is really a protocol error, which turns it into a wait for bytes that never arrive.

Went with `debug_assert!` over `expect()` deliberately: `buffer_position`/`buffer_length` are peer-influenced, and 6a7d9b65 in this branch just removed a panic on peer-supplied state in `reset_reader`. Adding one back in the hot read path would cut against that.

> [!NOTE]

saurabh500:
One correction to the "Verified correct, not flagged" section, since it's the assumption that would lead someone to re-tighten the check later:

> `end_of_message` cannot go stale across messages: both `reset_reader` impls route through `reset_to_length(0)` which clears it, and every command entry point in `tds_client.rs` calls `reset_reader()` before sending.

True for command entry points, but not for the **connection handshake**, which doesn't go through them. Applying `message_cannot_satisfy` verbatim to the scalar readers produced 353 failures, all `ProtocolError("Failed to receive token during login response parsing.")` — the first `read_byte` of the login response sees EOM still set from the prelogin response with an empty buffer, and errors instead of framing the new message.

So `refill_for` fires only when part of the value is already buffered (`available > 0`), which is unambiguously unsatisfiable. A read starting on an empty buffer is left alone: a caller beginning the next response is indistinguishable from an over-read without tracking whether a request was sent since the message ended. Details in the resolved thread on `network_transport.rs`.

Validation: full-suite failure sets are byte-identical against a stashed baseline (52 pre-existing — cert fixtures, named-pipe, shared-memory), and the live 2 GB `varbinary(max)` read still passes at 38.7 s.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -742,8 +742,66 @@ impl NetworkTransport {
...
+    /// A read that starts on an empty buffer is left alone even at
+    /// end-of-message: that is a caller beginning the next response, which the
+    /// transport cannot tell apart from an over-read.
+    async fn refill_for(&mut self, needed: usize) -> TdsResult<()> {
+        let available = self.tds_read_buffer.get_remaining_byte_count();
+        if available > 0 && self.message_cannot_satisfy(needed) {
+            return Err(Self::past_end_of_message_error(needed, available));
+        }
+        self.read_tds_packet().await
+    }
```
saurabh500: **Suggestion — the hang survives when the buffer is empty, and the fix now contradicts an existing contract.**

The `available > 0` carve-out means a scalar read that starts on an *empty* buffer at end-of-message skips the check entirely. It falls through to `read_tds_packet()`, which does `reset_to_length(0)` — clearing `end_of_message` — and then blocks on the socket forever.

Proven on identical input: a one-byte EOM packet against a live silent peer, byte consumed, then over-read by one.

| call | result |
| --- | --- |
| `read_bytes(&mut [0u8; 1])` | `Err(ProtocolError("...extends past the end of the message..."))` in 0.031s |
| `read_byte()` | **hung** — `Elapsed(())` after a 3s timeout |

The three bulk loops (`read_bytes`, `read_bytes_uninit`, `skip_bytes`) check unconditionally. The two families genuinely disagree on the same bytes.

**This is reachable on the path the PR exists to fix.** `drain_stream` and `advance_to_result_boundary` both loop while `done.has_more()`. A message whose final token is a DONE carrying DONE_MORE, terminated by EOM, sends the token reader straight back into `read_byte()` for the next token type — on an empty buffer, at end-of-message.

**But the carve-out is load-bearing.** I removed `available > 0` and ran the full suite. It broke exactly one test: `test_payload_free_eom_packet_is_accepted`. That test is pre-existing — added in #221, present on `main` — and it asserts that an empty EOM packet followed by a new message is legal, and that `read_byte()` carries on to read `0x7F` from the next packet.

**And the bulk loops now reject that same legal sequence.** Driving the #221 byte sequence through each path:

| call | result |
| --- | --- |
| `read_byte()` | `Ok(0x7F)` — pre-existing test passes |
| `read_bytes(&mut [0u8; 1])` | `Err(ProtocolError("TDS read made no progress..."))` |
| `skip_bytes(1)` | same `no_progress_error` |

It surfaces as `no_progress_error` rather than `past_end_of_message_error`: iteration one sees `end_of_message == false`, so the check doesn't fire; the refill pulls the payload-free EOM packet; `available` is then 0, so `to_read == 0`. Before this PR those loops would simply have refilled again and read `0x7F`.

So one of two positions has to be wrong, and the PR currently ships both:

- if crossing a message boundary on an empty buffer is legal, the bulk loops have a new false positive that regresses #221's contract;
- if it isn't legal, the scalar readers can still hang.

Your doc comment is honest about the cause — the transport genuinely cannot tell a caller starting the next response apart from an over-read. That's the real point: the distinction has to come from a layer that knows whether a new message is expected. Rather than have the transport guess from `available`, consider letting the caller say so, then drop the carve-out so both families agree:

```rust
/// Called when the client is about to read the first byte of a new message.
/// Clears the boundary so the next refill is a legitimate wait, not an over-read.
pub(crate) fn expect_new_message(&mut self) {
    self.tds_read_buffer.end_of_message = false;
}
```

with `refill_for` reduced to the unconditional form the bulk loops already use:

```rust
async fn refill_for(&mut self, needed: usize) -> TdsResult<()> {
    if self.message_cannot_satisfy(needed) {
        return Err(Self::past_end_of_message_error(
            needed,
            self.tds_read_buffer.get_remaining_byte_count(),
        ));
    }
    self.read_tds_packet().await
}
```

Whichever way you resolve it, the two families should end up agreeing on this input, and `test_payload_free_eom_packet_is_accepted` should have a bulk-read sibling so the contract is pinned from both sides.

saurabh500 (commented):
## Re-review — head `9a4a7794`

**3 of 4 findings fully fixed.** The scalar-read fix is most of the way there, but a narrower version of the same hang survives and the fix now disagrees with a contract the transport already had.

`cargo bfmt` clean. `cargo bclippy` clean. 1877/1884 tests pass — the 7 failures are pre-existing cert-fixture cases (`certificate_validator::tests::*`, `win_tls::validate::tests::*`) that fail because `tests/test_certificates/valid_cert.pem` isn't generated in a fresh worktree. Nothing to do with this PR.

### Resolved

**`close_query()` retirement** — the `matches!` guard is in and the comment now describes what the code actually does. `close_query_drain_sql_error_leaves_connection_alive` is the right test to add: it proves the connection *stays alive* on an ordinary `SqlServerError`, while the existing test still covers the retire path. Both directions asserted.

**PLP zero-fill** — the `unsafe { set_len }` is sound. I checked the three things it rests on rather than taking the SAFETY comment's word for it:

- `read_bytes_uninit` only returns `Ok(n)` once `length_to_read` reaches 0, so `Ok(n)` implies all `n` bytes were written.
- `reserve(chunk_len)` runs before `spare_capacity_mut()[..chunk_len]`, so the slice can't panic.
- `check_plp_chunk(chunk_len)` bounds the length *before* the `reserve`, so a hostile chunk length can't drive a huge allocation.

The SAFETY comment matches what's actually guaranteed.

**`buffers.rs` `try_read_array`** — `debug_assert!` plus the explanation of why a silent `None` turns a protocol error into a hang. Exactly right.

Also worth noting: the scalar fix covers all 13 readers, not just the 8 I listed. Byte counts are all correct.

### Still open

One finding, left inline on `refill_for`.

### What I did not close

I did not empirically demonstrate the end-to-end client drain hang. The client tests build over `ByteStreamTransport`, which returns EOF where a real peer goes silent — a fake that hits EOF can never show the *absence* of a hang. The transport-level behaviour below is executed and proven; the client-level reachability is read from the code, not run.

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Both of your claims reproduce, and the second one is the more important of the two. Fixed in 30274717, though not the way you proposed — I tried that first and it broke two real protocol paths.

**Your claims, verified.** `read_tds_packet` does clear `end_of_message` via `reset_to_length(0)`, so a scalar read starting on an empty buffer at EOM blocks forever. And the two families genuinely disagreed on identical bytes: on #221's payload-free-EOM sequence the bulk loops report `no_progress_error` where `read_byte` returns `Ok(0x7F)`. That's a regression my previous commit introduced and you caught it.

**Why the unconditional rule can't ship.** I made `refill_for` unconditional and cleared the boundary in `NetworkWriter::send()` — the single choke point for outbound bytes — expecting client-sends to cover the legitimate cases. The lib suite went clean, but the full suite surfaced two readers that cross a boundary with *nothing sent in between*:

- **Attention ack.** The server finishes the message it was already sending, then delivers the DONE carrying ATTN in a message of its own. `wait_for_attention_ack` has to cross. (`query_cancel`, `query_timeout_e2e`)
- **Multi-leg SSPI login.** Traced: `SEND 106 → pkt(EOM,54) → SEND 503 → pkt(EOM,282) → SEND 129 → pkt(EOM,40) → refill_for(1) fails`. `LoginResponse::deserialize` breaks only on `Done`/`Sspi`/`FedAuthInfo`, so it probes past EOM for the next token. (`test_localdb_connection`, `test_localdb_query_execution`)

`expect_new_message()` fixes each one — your idea works. But finding two in one local run means the strict rule needs an opt-in at *every* control path, and CI exercises ones I can't run here (Kerberos, fedauth, ODBC, JS, Python). A missed one is a connection failure, which is strictly worse than the hang being fixed. That's your own "the distinction has to come from a layer that knows" argument, now with data behind it.

**What landed instead.** Both families share one predicate: *a read may not cross a message boundary once it has consumed part of its value.*

```rust
(consumed_any || self.tds_read_buffer.get_remaining_byte_count() > 0)
    && self.message_cannot_satisfy(needed)
```

`refill_for` passes `consumed_any: false`; the three bulk loops pass `total_read > 0` (`skipped_any` for `skip_bytes`), and their `to_read == 0` bail is now three-way — no-progress if not at EOM, past-end-of-message if the value has started, otherwise refill. So they agree on every input, which is what you actually asked for. The reported hang is a value that runs past EOM, so it's still caught; a read *beginning* at a boundary refills, which is what the two paths above need.

Verification: with the predicate stubbed to `false` and the mid-value bails stripped, all 7 boundary tests fail on 5 s timeouts, so they're load-bearing rather than tautological. Full suite against live SQL Server is byte-identical to a freshly measured baseline (48 vs 48 pre-existing failures), and the live 2 GB read still passes at ~36 s.

**#221:** restored to its original assertion and extended with `read_bytes` and `skip_bytes` siblings, so it's pinned from both sides.

**Residual gap, on the record:** a read starting exactly at a boundary can still block against a silent peer. That's pre-existing, now identical in both families rather than divergent, and documented at the predicate. Closing it needs the caller-intent layer you described — worth doing, but it's a change to every message-consuming loop and doesn't belong in a hang fix.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4970,6 +4983,29 @@ impl TdsClient {
...
+            // A server-reported error is different: `advance_to_rows` only
+            // surfaces one after reading the terminal DONE, so the stream is
+            // fully drained and still synchronized. Retiring there would poison
+            // a healthy pooled connection every time a batch ends in RAISERROR.
+            //
+            // The state above is still cleared first: leaving the batch open
+            // would mask this error behind `ALREADY_EXECUTING_ERROR` on every
+            // later call, which is what made the original failure look like a
+            // permanent wedge.
+            if !matches!(error, crate::error::Error::SqlServerError { .. }) {
```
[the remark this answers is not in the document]
saurabh500: Correct, and it's a real defect — fixed in c0dbcd4d.

I verified it against the code rather than taking the report at face value. Both ERROR-token handlers (`advance_to_result_boundary` and `handle_row_read_token`) do exactly what you describe: `drain_stream()` without `?`, and on `Err` they only `warn!("...connection may not be reusable")` before returning `from_sql_errors(...)`. The warning names the problem and then nothing acts on it, so the resulting `SqlServerError` hits the `close_query` exemption and the connection goes back to the pool.

**Fix: retire at the point the drain gives up**, not by threading state into the error type. That arm is where the desynchronization is actually known, and it fixes every consumer rather than just `close_query`. Both handlers now call a shared `retire_desynchronized_connection()`. The SQL error still surfaces, so a caller's RAISERROR text isn't lost — only the connection is withheld from reuse.

Before making it destructive I separated what can actually produce that `Err`: a dead socket, a protocol error, an elapsed timeout, or a cancellation. All four stop the drain *before* the terminal DONE, so unread response bytes remain in every case and retirement is the right reaction for each — no cause where it's too aggressive.

**Regression coverage**, both requested and the quiet-side control:

- `error_token_with_truncated_remainder_retires_connection` — ERROR + DONE(more) + truncated, via `advance_to_result_boundary`.
- `close_query_sql_error_over_truncated_stream_still_retires_connection` — the same stream end-to-end through `close_query`, pinning that the `SqlServerError` exemption can't resurrect an abandoned connection.

I checked these aren't tautologies: with the retirement stubbed to a no-op, both new tests fail while `close_query_drain_sql_error_leaves_connection_alive` and `close_query_success_leaves_connection_alive` keep passing — so the guard fires on the broken stream and stays quiet on healthy input.

Also corrected two comments that asserted the premise you disproved (`close_query`'s guard and that control test's doc both claimed `from_sql_errors` implies a fully-drained stream). That claim was the trip hazard here, so it now says why the exemption is safe — the failing arms retire the transport themselves before the error ever reaches it.

Validation: `cargo bfmt` and `cargo bclippy` clean; lib suite 1879/1886 with only the 7 pre-existing cert-fixture failures; full suite against live SQL Server unchanged at 48 pre-existing failures, none in the drain, retirement, or pooling paths.

One scoping note: you flagged line 5004 as the same issue. It's the same `if` body, and the shared helper covers both lines.

> [!NOTE]

saurabh500:
On the suppressed finding in the Copilot review (`tds_client.rs:5005`) — I checked it rather than leaving it silent, because it's correct about the mechanism:

> `mark_known_dead()` only sets the cached flag used by pool/liveness consumers; it does not close `NetworkTransport::stream`. Because the next line disables session recovery, `check_and_reconnect()` returns early when the same `TdsClient` is reused

Both halves verify. `check_and_reconnect` returns `Ok(Duration::ZERO)` at the `!session_recovery_negotiated` check before it ever polls the transport, and `is_connection_dead()` is a cached read documented as a pool-facing liveness check — nothing on the command path gates on it. So a caller reusing the same `TdsClient` directly (rather than returning it to a pool) can still write a new request on the desynchronized socket.

I did not act on it in this PR, deliberately:

- **It's pre-existing, and this PR narrows it.** The retirement in `close_query` has the same property and was reviewed in an earlier round. Before `c0dbcd4d` the failed-drain arms marked nothing at all, so pool consumers reused the connection too. That half is now closed.
- **Both suggested remedies carry risk I can't validate here.** Closing the transport at that point means `close_transport()`, which does `stream.shutdown().await` — a TLS close_notify write on a socket we just failed to read from. Putting a blocking write into an error path in a PR whose subject is read hangs is a poor trade. Gating command preflight on the cached flag is broader still and works against idle-connection resiliency, where `check_and_reconnect` is *supposed* to run on a transport marked dead and revive it.

The narrow version that would work is dropping the stream without shutting it down (`stream = None`, no I/O, so `is_connection_dead()` flips to true and the next write fails cleanly). That needs a new method on `TdsTransport` and a decision about whether abandoning a socket without close_notify is acceptable for this codebase — your call, not one I should make unattended on a merge tick.

Happy to do it in this PR if you'd rather not carry it forward.

David-Engel (commented):
## Summary

Three real defects, correctly diagnosed and correctly fixed: the PLP chunk-count cap (a `#[cfg(fuzzing)]` tripwire that leaked into release builds), the missing end-of-message tracking that let short reads park on a silent socket, and `close_query()`'s early `?` that left `has_open_batch` set and wedged the connection behind `ALREADY_EXECUTING_ERROR`.

What I verified rather than assumed:

- **The cap removal loses no hardening.** `MAX_PLP_SIZE` is `64 KB` under `cfg(fuzzing)` and `i32::MAX` otherwise, and all three PLP loops are bounded by total bytes, wire-declared length, or `dest.len()` (`ensure_active_chunk` :400, `read_plp_chunks_unknown_len` :1200, `read_plp_chunks_into_slice` :1316). The original OOM fuzz guard is strictly tighter than the chunk cap ever was, so nothing regressed. Your rationale in `check_plp_chunk`'s doc is accurate.
- **The EOM flag cannot go stale.** `get_new_tds_packet` sets it per framed packet, `remove_header_from_packet` does not touch it, and `reset_to_length` / `change_packet_size` clear it. `read_tds_packet` clears then re-sets it on the empty-buffer path.
- **`set_len` on the new spare-capacity read is sound.** The default `read_bytes_uninit` zero-fills the whole buffer first, and `NetworkTransport`'s only returns `Ok` when `length_to_read` reaches zero, so `Ok(chunk_len)` really does mean `chunk_len` initialized bytes.
- **Retiring on a timeout error is right, not over-aggressive.** I checked whether a command timeout leaves the stream synchronized — it does not. `receive_token` drops the read future mid-packet and only then sends ATTENTION, which is the cancel-safety gap you call out under "Not addressed here." So a non-`SqlServerError` drain failure genuinely implies desynchronization.
- `cargo nextest run -p mssql-tds --lib`: 1791/1795 pass, the 4 failures are the known missing `valid_cert.pem` fixtures, identical on a clean baseline.

No blocking findings. The self-review trail on this PR is unusually good — several of the things I went looking for were already found, verified and answered in the threads.

## Nits (not anchored to changed lines)

**Track the residual boundary hang.** The "Known gap" note in `refill_for` is accurate and the reasoning for not closing it (353 handshake failures) is convincing. But the hang class the PR is titled after is not fully eliminated: a scalar read starting on an empty buffer at EOM still re-enters the socket. On the drain path that is a coin flip on whether the last read lands exactly on the boundary. Worth an issue so the doc block is not the only record of it.

**The `abort_streamed_write` comparison in the description overstates the symmetry.** Section 4 says the retirement "mirrors the existing behaviour for an aborted streamed RPC write." `abort_streamed_write` (tds_client.rs:1498) marks the transport dead *and* calls `close_transport()`, and it does not touch `session_recovery_negotiated`. `retire_desynchronized_connection` does the opposite on both counts. The choice is well argued in your reply to the Copilot finding; the description just claims more parity than exists. See the inline note on `retire_desynchronized_connection` for why that difference matters.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3054,6 +3054,18 @@ impl TdsClient {
...
     }
 
+    /// Retires the connection after the reader has lost track of the stream.
+    ///
+    /// Every way a drain can fail — a dead socket, a protocol error, an elapsed
+    /// timeout, a cancellation — stops it before the terminal DONE, so bytes of
+    /// the response are still unread and the next command would decode them as
+    /// tokens. A TDS stream cannot be resynchronized from an unknown offset, so
+    /// the connection must not go back to a pool.
+    fn retire_desynchronized_connection(&mut self) {
```
David-Engel: **Two follow-ups on this helper — one about coverage of the rule it establishes, one about whether it delivers what its doc promises.**

**1. Two more `drain_stream()` sites in the same class.** `c0dbcd4d` established the rule "a drain that gives up mid-stream leaves the reader at an unknown offset, so retire." It is applied to both ERROR-token arms and `close_query`. Two sites still propagate with a bare `?`:

- `execute_sp_prepare` (tds_client.rs:2310): `let server_errors = self.drain_stream().await?;`
- `execute_sp_unprepare` (tds_client.rs:2440): same

A failure there desynchronizes the stream exactly as in the arms you fixed, but the connection stays reusable. Either call this helper there too, or note why prepare/unprepare are exempt. Pre-existing, and this PR narrows the class rather than widening it, so a follow-up is fine.

**2. The doc promises more than the code delivers.** "later commands fail against the dead connection rather than reading garbage" does not hold for a caller reusing the same `TdsClient` outside a pool:

- `mark_known_dead()` (network_transport.rs:1907) only sets `known_dead`; the socket stays open.
- `send_query_batch` gates on `command_is_busy()` (tds_client.rs:3502 — `has_open_batch || streamed_write_state != Idle`) and `check_and_reconnect()`.
- `check_and_reconnect` returns `Ok(Duration::ZERO)` at tds_client.rs:841 on `!session_recovery_negotiated`, which this helper has just set to false.
- `is_connection_dead()` is pool-advisory; in `mssql-odbc` it surfaces only through `SQL_ATTR_CONNECTION_DEAD`. No execute path consults it.

This is the half of the Copilot finding you deliberately left, so you know it. What I would add: because the batch is now closed, the next `execute()` proceeds where pre-PR `command_is_busy()` blocked it. That block was itself the reported bug — a permanent `ALREADY_EXECUTING_ERROR` with the real error masked — so this is not a regression in any meaningful sense, and command entry points call `reset_reader()` first, which discards the client-side buffer. The residual exposure is stale bytes still queued in the kernel receive buffer, not "parked at an unknown offset."

I agree with you that `close_transport()` is the wrong remedy, and there is a second reason beyond the blocking-write-in-a-read-error-path argument. It is:

```rust
if let Some(stream) = self.stream.as_mut() {
    stream.shutdown().await?;   // early return, before the line below
}
self.stream = None;
```

The `?` returns before `self.stream = None`, so any wrapper that logs and swallows that error leaves the stream alive in precisely the degraded-socket case retirement exists for.

Your own alternative is the right one: a `TdsTransport::abandon_stream()` that drops the socket with no I/O (`stream = None`), so `is_connection_dead()` flips and the next write fails cleanly, with no `close_notify` write and no failure mode. Happy for that to be a follow-up. Either way, the doc comment here and section 4 of the description should say "pooled consumers discard it" rather than "later commands fail," so the promise matches the mechanism.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4970,6 +5004,34 @@ impl TdsClient {
...
+            // ERROR-token handlers build one even when their own drain fails.
+            // Those arms call `retire_desynchronized_connection` themselves, so
+            // by the time such an error reaches here the transport is already
+            // marked dead and this exemption cannot revive it.
+            //
+            // The state above is still cleared first: leaving the batch open
+            // would mask this error behind `ALREADY_EXECUTING_ERROR` on every
+            // later call, which is what made the original failure look like a
+            // permanent wedge.
+            if !matches!(error, crate::error::Error::SqlServerError { .. }) {
```
David-Engel, line 5299: **The exemption is correct today, but the invariant it rests on lives in the error taxonomy rather than anywhere enforceable.**

I enumerated the `from_sql_errors` sites reachable from `advance_to_rows`:

| site | reaches terminal DONE? | retires? |
|---|---|---|
| tds_client.rs:3429 `advance_to_result_boundary` ERROR arm | no | yes, added in `c0dbcd4d` |
| tds_client.rs:4850 row-read ERROR arm | no | yes, added in `c0dbcd4d` |
| tds_client.rs:5297 | yes — guarded by `if !done.has_more()` | n/a, stream is healthy |

(2006, 2317 and 2442 are not reachable from `advance_to_rows`.) So the exemption holds, and `close_query_sql_error_over_truncated_stream_still_retires_connection` pins it well.

The unwritten rule is "every reachable `from_sql_errors` site either reaches the terminal DONE or retires where it gives up." A fourth site added later that does neither silently reopens exactly the hole Copilot found, and no test fails. An explicit `stream_desynchronized: bool` set where a drain gives up and checked here would put the invariant in one place. Inverting the default is not the answer — that is the design you already had to back out when it poisoned healthy connections on `RAISERROR`.

Non-blocking; reasonable to defer.

**Separately, a nit on the comment block above this line.** Its first paragraph restates the third paragraph of `close_query`'s own doc (:4970-4975) almost verbatim. The part that earns its place inline is the `SqlServerError` rationale, because it justifies the `matches!` directly below. Trimming the duplicated opener would leave the block doing only the job the doc cannot.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -742,8 +742,96 @@ impl NetworkTransport {
...
+    /// is accepted by the framing layer (it legally terminates a message) but
+    /// contributes zero payload, so the loop would re-enter the refill with the
+    /// same unmet demand and block on the socket forever — the server has
+    /// finished its message and will not send more until the client issues a
+    /// new request. The same happens when a value claims more bytes than the
+    /// message actually contains.
+    ///
+    /// Neither case can make progress, so fail loudly instead of hanging.
+    #[cold]
+    fn no_progress_error(outstanding: usize) -> crate::error::Error {
```
David-Engel, line 803: **This branch appears to be unreachable, and the doc describes a case that routes elsewhere.**

`get_new_tds_packet` rejects a payload-free packet unless it is EOM (:1016), so after any successful refill: payload > 0, or `end_of_message == true`. In all three bulk loops `to_read == 0` implies `available == 0` implies the framed packet carried no payload, which implies EOM. So the `if !self.tds_read_buffer.end_of_message` guard that returns this error cannot fire. That matches the coverage bot flagging 1408, 1461-1465 and 1571 as the uncovered lines.

The doc here explains the payload-free-EOM scenario, but that scenario has `end_of_message == true` and routes to `past_end_of_message_error` or the `continue`. So the described case never reaches the error it documents.

This also matters for the claim in `refill_for`'s doc at :803 — "the bulk loops apply the identical rule ... so the two families agree on every input." I worked the inputs and the behavioural claim is true: on one payload byte in an EOM packet then silence, `read_bytes(&mut [0u8; 2])` and `read_int16()` both error, and on the payload-free-EOM shape all three families agree with `read_byte`, which `test_payload_free_eom_packet_is_accepted` now pins. But it is true *because* of the framing invariant above. `no_progress_error` is the one branch with no scalar counterpart, and if it were reachable the two families would genuinely disagree — bulk errors where scalar hangs.

Suggest either keeping it and saying so ("unreachable: framing rejects payload-free non-EOM packets; retained as an invariant guard"), or dropping it and letting section 2's check carry the case. Right now a reader has to reconstruct the framing invariant to know which it is.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -1189,55 +1174,72 @@ impl GenericDecoder {
...
+            if read != chunk_len {
+                return Err(crate::error::Error::ProtocolError(format!(
+                    "PLP chunk read returned {read} byte(s) but the chunk header declared {chunk_len}"
+                )));
+            }
+            let new_len = plp_buffer.len() + chunk_len;
+            // SAFETY: `read_bytes_uninit` returned `Ok(chunk_len)`, so it
+            // initialized the first `chunk_len` bytes of the spare capacity
+            // reserved above.
+            unsafe { plp_buffer.set_len(new_len) };
```
David-Engel, line 1224: **Worth pinning the contract this `set_len` depends on.**

This is sound only because `read_bytes_uninit` guarantees that `Ok(n)` means the first `n` elements were initialized. I checked both implementations and both honour it — the default in `packet_reader.rs` zero-fills the entire buffer before delegating, and `NetworkTransport`'s only returns `Ok` once `length_to_read` reaches zero — so the `read != chunk_len` guard plus this `set_len` is correct as written.

The guarantee is not stated anywhere, though. The rustdoc on the trait method is just "Fills a possibly-uninitialized byte buffer." The trait is `pub(crate)` in normal builds (`pub` only under `cfg(fuzzing)`), so this is contained, but an implementation added later that returns a count without writing that many bytes would make safe code UB with nothing to catch it. One line on the trait method would fix it:

```rust
/// Fills a possibly-uninitialized byte buffer.
///
/// On `Ok(n)`, the first `n` elements of `buffer` are initialized. Callers
/// rely on this to `set_len` an owned buffer, so an implementation must not
/// report bytes it did not write.
```

mssql-tds/src/connection/tds_client.rs
saurabh500: Both parts fixed in `1ad1fe26` — and your first point was bigger than the two sites you found.

**1. Not two more sites, seven.** I grepped the whole family rather than just patching the two you named. Besides `execute_sp_prepare` (2310) and `execute_sp_unprepare` (2440), `cursor_ops.rs` has five more with the same bare `?`: lines 565, 647, 723, 1051, 1135. All seven are the same class.

They were also worse than "connection stays reusable". In every one the `?` returns *before* the `set_has_open_batch(false)` on the next line, so a failed drain left the batch open too — which is exactly the `ALREADY_EXECUTING_ERROR` masking that made the original bug look like a permanent wedge. So each site had both halves of the problem.

Rather than write the same fix seven times I put the rule in one place as `drain_stream_or_retire()`, which drains and, on failure, closes the batch and retires. All seven now call it. That also answers your thread 2 point about the invariant living nowhere enforceable: it isn't a full `stream_desynchronized` flag, but "either the drain reaches terminal DONE or the connection is retired where it gave up" is now a single function, so a site added later that forgets the rule is a *missing call to it* rather than a silent hole. The two `from_sql_errors` sites at 2317/2442 are unchanged — those are the successful-drain path, so the stream is healthy, as your table says.

Pinned with `drain_stream_or_retire_failure_closes_batch_and_retires_connection`, plus a healthy-stream control. I checked they aren't tautologies by stubbing the retirement out behind a temporary env var: the failure test failed and the control kept passing, which is the split I wanted — it fires on the broken stream and stays quiet on healthy input.

**2. You're right that the doc overpromised, and your correction is more precise than mine.** I've rewritten it. `close_query` now says pooled consumers discard the connection instead of "later commands fail", and `retire_desynchronized_connection` carries an explicit what-this-does-not-guarantee paragraph: it doesn't close the socket, no execute path consults `is_connection_dead`, and — taking your framing — the residual exposure is stale bytes in the kernel receive buffer, not a reader parked mid-value, since entry points call `reset_reader()` first.

Thanks for the `close_transport` catch; that's a better argument than mine and I've recorded it in the doc. The `?` on `stream.shutdown()` returning before `self.stream = None` means any wrapper that logs and swallows leaves the stream alive in precisely the degraded case retirement exists for. So it's not just a bad trade, it's unsound for this purpose.

`abandon_stream()` left as follow-up, noted in the doc. Happy to do it in this PR if you'd rather not carry it — it's a small transport method, I kept it out only because it widens the blast radius of a hang fix.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Thanks for enumerating the sites — that table is the useful artifact here, and it matches what I get walking the same paths.

**On the unenforceable invariant.** Partly addressed in `1ad1fe26`, though not the way you proposed. Working your thread-1 point I found seven `drain_stream()` sites propagating with a bare `?` (the two you named plus five in `cursor_ops.rs`), so instead of patching each I added `drain_stream_or_retire()` and routed all of them through it. The rule "a drain either reaches the terminal DONE or the connection is retired where it gave up" is now one function rather than a habit re-derived at each site, so a new site that forgets it shows up as a missing call rather than silently reopening the hole.

That is weaker than the `stream_desynchronized: bool` you sketched — it constrains the *drain* sites, not every future `from_sql_errors` caller — so your concern isn't fully retired. A fourth `from_sql_errors` site that neither reaches terminal DONE nor drains would still slip past. I stopped short of the flag for the reason you already anticipated: the state it adds has to be cleared correctly on every recovery path, and getting that wrong is how the earlier default-retire design poisoned healthy connections on `RAISERROR`. Deferring rather than half-implementing it, and flagging it here so it's on the record.

**Nit taken** — you were right that it was duplication, not emphasis. Dropped the first paragraph; the block now opens at "A server-reported error is different", so the only prose left inline is the part that justifies the `matches!` directly below it.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Verified, and you're right — the branch is unreachable. I worked it rather than taking it on faith, and the chain closes:

- In all three loops `to_read = min(available, min(length_to_read, max_packet_size - 8))` with `length_to_read > 0`, so `to_read == 0` ⟺ `available == 0`.
- If `do_we_have_enough_data(n)` was true for `n > 0` we skipped the refill, and `available >= n > 0`. So `available == 0` requires that we *did* refill.
- `read_tds_packet` has no short-circuit — both arms call `get_new_tds_packet()` unconditionally — so a successful refill always frames exactly one packet and sets `end_of_message` from that packet's header. That was the part worth checking: if it could return `Ok` without framing, a stale `end_of_message == false` would survive and the branch *would* be live.
- Line 1016 then rejects payload-free non-EOM. So an empty buffer after a refill implies the framed packet was payload-free, which implies EOM.

Hence `!end_of_message` cannot hold at `to_read == 0`, which matches the coverage bot on 1408, 1461-1465 and 1571.

**Kept it, documented as unreachable** — your first option. Dropping it is tempting for the dead-code and diff-coverage reasons, but the failure mode if the framing invariant is ever relaxed is a silent infinite loop in the bulk loops, which is the precise hang this PR exists to remove. A cold branch that turns that into a loud error is worth three uncovered lines. The doc now opens with "Unreachable as the framing layer stands today, and deliberately kept", spells out the invariant chain so nobody has to reconstruct it, and records your observation that it's the one branch with no scalar counterpart — so were it reachable the families would genuinely disagree, bulk erroring where scalar hangs.

I also tightened the `refill_for` claim you flagged at :803. You're right the behavioural claim is true but true *because* of the framing invariant, so "the two families agree on every input" now says that agreement leans on the invariant documented on `no_progress_error`, rather than presenting it as unconditional.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Agreed and applied in `1ad1fe26` — I took your wording essentially verbatim onto the trait method in `packet_reader.rs`:

```rust
/// Fills a possibly-uninitialized byte buffer.
///
/// On `Ok(n)`, the first `n` elements of `buffer` are initialized. Callers
/// rely on this to `set_len` an owned buffer, so an implementation must not
/// report bytes it did not write.
```

This is the right call. The safety comment at the `set_len` was asserting a property of a *different* function with nothing at that function stating it, so the two could drift apart silently and the compiler would never notice — and the failure mode is UB reachable from entirely safe code. `pub(crate)` contains the blast radius but doesn't remove it, since the implementation added later would be in-crate too.

I also reworded the `SAFETY` comment at the call site to say it's relying on the documented contract rather than on an observed property, so a reader follows the link instead of re-verifying both implementations the way you just had to.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3255,6 +3255,58 @@ impl TdsClient {
...
+    /// may remain queued in the kernel receive buffer. Closing the socket here
+    /// is the wrong remedy: `close_transport` writes a TLS close_notify on a
+    /// socket we just failed to read from, and its `?` returns before
+    /// `self.stream = None`, so a swallowed error would leave the stream alive
+    /// in exactly the degraded case retirement exists for. Dropping the stream
+    /// with no I/O is the fix; it needs a new transport method and is left as
+    /// follow-up.
+    fn retire_desynchronized_connection(&mut self) {
+        self.transport.mark_known_dead();
+        self.recovery_context.session_recovery_negotiated = false;
```
shiwanigupta0809: Clearing `session_recovery_negotiated` here applies to transport errors as well as protocol desync, which disables connection resiliency for a case it was built to handle.

`drain_stream_or_retire` retires on every `Err` from `drain_stream` without classifying it. SQL errors come back in the `Ok(Vec<SqlErrorInfo>)` arm, so `Err` is either a protocol desync or a transport I/O error. Desync is unrecoverable and retiring is right. A mid-drain connection drop is the case session recovery exists for.

The flag is load-bearing in `check_and_reconnect`:

```rust
if !self.recovery_context.session_recovery_negotiated {
    return Ok(Duration::ZERO);
}
```

So after a network blip during a drain, the next command no longer attempts recovery — it returns `Ok(ZERO)` and proceeds against a dead transport, with no error surfaced. Before this change the same blip was recovered transparently. The five `cursor_ops.rs` conversions put this on routine cursor open/close/fetch paths, not only exceptional ones.

The reason for clearing it makes sense — leaving it set means `is_connection_dead()` triggers a reconnect that papers over a genuine desync. But that argues for retiring on desync specifically rather than collapsing both cases. Could we gate on the error kind, so desync retires while a transport drop still recovers? If that isn't clean and the safety trade is deliberate, that's reasonable — in which case it'd be worth noting the resiliency consequence in the doc comment, since it isn't mentioned today.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1311,18 +1428,30 @@ impl TdsPacketReader for NetworkTransport {
...
-                    &self.tds_read_buffer.working_buffer[self.tds_read_buffer.buffer_position
-                        ..self.tds_read_buffer.buffer_position + to_read],
-                );
-                offset += to_read;
-                length_to_read -= to_read;
-                total_read += to_read;
-
-                self.tds_read_buffer.consume_bytes(to_read);
+            if to_read == 0 {
+                if !self.tds_read_buffer.end_of_message {
```
shiwanigupta0809, line 1454: Small nit, non-blocking. The `continue` re-enters with `total_read == 0` and `remaining == 0`, so `value_truncated_by_message_end(false, ..)` is false and we go back to `read_tds_packet()`. `get_new_tds_packet` accepts a payload-free packet when EOM is set, so a peer emitting empty end-of-message packets keeps this loop going with no state advancing between iterations.

Worth mentioning because the doc on `no_progress_error` describes this shape closely: *"the loop would re-enter the refill with the same unmet demand and block on the socket forever."* The guard covers the non-EOM case and this EOM sibling stays open.

The command timeout bounds it in practice and it needs a misbehaving server, so nothing urgent — an iteration cap or a `debug_assert` on forward progress would close the loop on the invariant. Same shape in `read_bytes_uninit` and `skip_bytes`.

shiwanigupta0809 (commented):
Read through the branch — nice piece of work, and the reasoning captured in the doc comments made it much easier to follow. The chunk-cap removal and the uninit read path both look right, and the `unsafe { set_len }` is sound: `NetworkTransport` is the only implementor overriding `read_bytes_uninit`, and it only returns `Ok` once every byte has been written.

One concern on the connection-retirement change that I think is worth settling before this merges, plus a small nit. Details inline.

mssql-tds/src/connection/tds_client.rs
saurabh500: You're right, and this is a regression I introduced. Fixed in f3f38fcb.

My original reasoning only asked "is the stream desynchronized?" — which is true for every failure kind — and never asked the question that actually matters: "is retirement the right *remedy* for each cause?" Those are two different failures with two different answers, and I collapsed them.

The split is now explicit in `drain_failure_lost_the_transport`:

- **Transport loss** (`Io`, `ConnectionClosed`, `ConnectionError`, `TlsError`, `TlsHandshakeError`) → `mark_known_dead()` only, **recovery stays armed**. There is no offset left to recover, and reconnecting is precisely the remedy. As you note, these run on routine cursor open/close/fetch paths, so disarming resiliency here was the worst place to do it.
- **Everything else** (`ProtocolError`, `TimeoutError`, `OperationCancelledError`) → full retire, as before. The socket may be perfectly healthy, so leaving recovery armed would let the next command silently reconnect and paper over a decoder bug instead of surfacing it.

Applied at all four retirement sites, not just the helper — the two ERROR-token arms and `close_query` had the same problem.

One small correction to the description: an error *does* surface, just at the next write rather than as transparent recovery. The substantive point stands, and the doc on the retirement helper now spells out the resiliency consequence of each branch as you asked.

Test: `drain_stream_or_retire_failure_closes_batch_and_retires_connection` now arms recovery explicitly and asserts it survives a truncated-stream drain (the old assertion passed vacuously — recovery was already false in that fixture), plus `retire_after_failed_drain_disarms_recovery_only_for_desync` for the other half. Confirmed non-tautological: neutering the classifier makes the first fail while the controls keep passing.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Confirmed and fixed in f3f38fcb — this was a genuine remaining hang.

Traced it: the `continue` re-enters with `total_read == 0` and `remaining == 0`, so `value_truncated_by_message_end(false, ..)` is false → `read_tds_packet()` → `get_new_tds_packet` accepts the payload-free EOM → `to_read == 0` → `continue`. Nothing advances. Same shape in all three loops, bounded only by the command timeout.

Rather than remove the tolerance (it's needed — a read starting exactly at a message boundary legitimately consumes the previous message's terminating empty packet), I capped it at one. A second consecutive payload-free message means nothing advanced between iterations, and no legitimate server sends two empty messages back to back; the attention acknowledgement carries a DONE token, so it isn't payload-free either. Second one now errors via a `#[cold] stalled_on_empty_messages_error`.

No counter reset is needed: once `to_read > 0` the `continue` branch becomes unreachable for the rest of the call, since `total_read > 0` (and `skipped_any` plays the same role in `skip_bytes`).

Applied to all three loops — `read_bytes`, `read_bytes_uninit`, `skip_bytes` — with a regression test each, plus `bulk_reads_tolerate_a_single_leading_empty_message` as the control for the legitimate case. Worth noting the tautology check: with the cap raised, all three regression tests fail with `Elapsed` rather than a wrong error, i.e. they pin an actual hang, while the controls stay green.

> [!NOTE]

shiwanigupta0809 (approved):

PR #361, opened 2026-08-22
Support data-at-execution parameters in mssql-odbc

shiwanigupta0809:
## Summary

This change completes the ODBC data-at-execution (DAE) / PLP streaming follow-up work and brings the Rust implementation in line with native msodbcsql behavior for streamed parameter handling, across both SQLExecute and SQLExecDirect.

### Streaming path
Added the missing DAE/PLP streaming wiring in the ODBC API path so parameter data can be staged, tracked, and finalized correctly across SQLPutData / SQLParamData flows.
Fixed parity gaps in length accounting and state transitions for streamed parameters, including expected-length tracking, under-send / over-send checks, and close-time validation.
Corrected NULL and SQL_NTS handling in streamed writes so they behave like the native driver.
Rejected unsupported data-at-execution parameters before they reach the RPC serialization path to prevent malformed payloads and panics on public entry points.
Added coverage for cancel / abort semantics and connection-dead behavior in the TDS client.

### Data-at-execution keeps the prepared path
A streamed parameter no longer costs a prepared statement its plan.

Previously SQLExecute handled DAE by abandoning the prepared handle and streaming ad-hoc sp_executesql, so an application that streamed a parameter re-sent the SQL text on every execute and lost handle reuse. msodbcsql does not do this: it selects the procedure purely on whether the statement is prepared (sqlccmd.cpp:3293 and :6623) and treats DAE as orthogonal, parking a half-written sp_prepexec / sp_execute RPC (AddRPCUserParameters returns SQL_NEED_DATA mid-build at sqlccmd.cpp:1693, and DAE_EXEC_AS_RPC routes the subsequent SQLPutData / SQLParamData calls). The park-a-half-written-RPC model was already identical here; only which RPC got parked differed.

Added TdsClient::begin_execute_prepared, the streamed counterpart of execute_prepared. It streams into sp_prepexec on the first execute and sp_execute once a handle is cached, matching what a materialized execute on the same statement would have sent.
SQLExecute's Need Data arm now routes through it. SQLExecDirect deliberately stays on begin_sp_executesql there is no prepared plan to preserve there.
The statement id is claimed when the message is parked rather than when it completes, because sp_prepexec returns its handle as a RETURNVALUE trailing the result set and the capture has to be armed before the drain reaches it.
cancel_streamed_write and abort_streamed_write now disarm that pending capture. A cancelled sp_prepexec never yields its handle, and leaving the capture armed would divert an unrelated RPC's first return value into the handle map under a dead id.
The orphaned handle is not piggybacked onto a streamed sp_prepexec, unlike the materialized path. That request stays open for the whole SQLPutData sequence and may never reach the server, so evicting the superseded handle at build time could leak the plan until disconnect. It rides along with the parked state instead and is released when the sequence ends - by the next execute, or by SQLCancel and the free that follows it. A Driver Manager rejects SQLFreeHandle outright while the statement is in Need Data, so the free on its own is never the trigger.
SQLParamData drains a no-row prepared result before finishing, which is what materializes the handle for reuse. Direct execution parks no plan and has no trailing handle, so it keeps its existing batch-navigation behaviour.
Always Encrypted remains rejected on the streamed prepared path, as it already was on the streamed ad-hoc path.

### `SQLExecDirect` + DAE
SQLExecDirect now stages streamed parameters the same way SQLExecute does, rather than rejecting them with HYC00.

sql_exec_direct_w_safe builds its parameter list with the DAE-aware builder, so a SQL_DATA_AT_EXEC / SQL_LEN_DATA_AT_EXEC(n) indicator becomes a streaming placeholder instead of being read as a value. The marker_count == 0 plain-batch path is unchanged.
When placeholders are present it calls begin_sp_executesql and returns SQL_NEED_DATA.
Added the missing Need Data guard: SQLExecDirect during an open streaming sequence now posts HY010 rather than falling through to the cursor-state check and reporting 24000.
Extracted the DAE client-parking block from execute.rs into a shared park_dae_client in exec_common.rs. Both entry points set the same state, and duplicating that was the obvious way for the two paths to drift apart.
The parked prepared plan is None for direct execution, so completing a direct-execution sequence correctly leaves the statement unprepared Γò¼├┤Γö£├ºΓö£Γòó covered by a dedicated e2e test rather than assumed.

### `SQLCancel` and sequence enforcement
SQLCancel was a no-op stub. That was harmless until DAE made the ODBC "Need Data" state reachable, where it is one of the few legal calls: an application that cancelled mid-stream had no way out and the statement stayed parked forever. It now unwinds the sequence, restores the prepared plan, and returns the client to idle.
cancel_streamed_write is state-aware and retracts the request at the protocol level instead of dropping the connection. A request still buffered in the client is discarded locally with no bytes sent; one that is partially sent is terminated with an EOM | IGNORE packet and the answering DONE is drained. Only if that handshake fails does the transport close. This mirrors BATCHCTX::Cancel (TdsSend.cpp:995-1147), which closes the socket on no branch.
SQLPutData mixing SQL_NULL_DATA with value chunks in either direction now posts HY020. A streamed parameter is either PLP_NULL or a series of chunks; there is no wire encoding for both.
SQLExecute while awaiting data-at-execution input now posts HY010 rather than the cursor-state error, since Need Data is a sequence error for every call outside SQLPutData / SQLParamData / SQLCancel.
Folded the two near-identical private DAE abort helpers in param_data.rs and put_data.rs into a shared unwind_dae / abort_dae_with_diag in exec_common.rs. Both copies released and re-acquired the statement mutex to restore the prepared plan, leaving a window in which the statement had neither DAE state nor a prepared plan; the shared version does it under one lock.

### Statement state
Grouped the eleven separate dae_* fields on StmtState into a single DaeState with private members and accessors, so "in a DAE sequence" is one Option rather than a set of flags that could disagree. take_dae restores the prepared plan and orphaned handle in the same step, so the statement is never observable as idle but unprepared.

### Documentation
Added mssql-odbc/docs/data-at-execution-streaming.md and updated parameters_plan.md. Several claims were corrected after tests disproved them:

Cancelling leaves the connection usable. An earlier draft of this PR said otherwise; that was true of the original cancel_streamed_write, which closed the transport unconditionally, and is no longer true after the state-aware rewrite above.
The SQL_NULL_DATA-after-chunks HY020 is unreachable through unixODBC 2.3.9, which rejects the null data pointer with HY011 in the driver manager. The opposite direction (a chunk after SQL_NULL_DATA) does reach the driver and is covered by an e2e test.
Removed an inaccurate claim that this PR fixed streamed-parameter ordering. The materialized-before-streamed reordering predates this branch and is unobservable, because every parameter is named @ P{n} and the partition is order-stable.

## Why
The prior implementation had several behavioral gaps in the streamed-parameter path:

length accounting could drift from the real data sent
sequence enforcement could allow malformed transitions
SQLPutData(NULL, 0) and SQL_NTS scenarios did not match native behavior
some public entry points were not guarding against data-at-exec parameters
an application had no way to escape the Need Data state once entered, and cancelling cost it the connection
SQLExecDirect rejected streamed parameters outright, so applications had to prepare a statement they only intended to run once
streaming a parameter silently downgraded a prepared statement to ad-hoc execution

## Known limitations
A DAE sequence that fails on the wire (rather than being cancelled or driver-rejected) loses the request and the socket, because a request interrupted mid-send cannot be retracted with EOM | IGNORE. The session survives: abort_streamed_write clears the open-batch flag before dropping the transport and the client goes back to the connection, so the next execute reaches check_and_reconnect, sees the dead socket, and replays session state onto a fresh login. Idle Connection Resiliency is on by default (connect_retry_count = 1), so this needs no connection-string opt-in. That is the same lazy recovery msodbcsql performs in GetBatchCtxOrRecover (sqlctokn.cpp:4088), which also runs at the top of the next API call rather than at failure time. Recovery is declined while a transaction is open, so a DAE sequence that dies inside an explicit transaction does end the connection.
Cross-thread SQLCancel against an in-flight request the SendAttn branch of BATCHCTX::Cancel  is not implemented and is tracked separately. This PR covers cancel between SQLPutData calls, which is the case DAE makes reachable.
DAE streaming supports SQL_C_CHAR, SQL_C_WCHAR, and SQL_C_BINARY only; Always Encrypted and output parameters are excluded.

## Validation
cargo bfmt  clean
cargo bclippy  clean
cargo nextest run -p mssql-odbc --lib  774 passed, 0 failed
cargo nextest run -p mssql-tds --lib  1876 passed, 0 failed
ODBC e2e suite against a live SQL Server (mssql-odbc/tests/e2e/run_e2e.sh) 26/26 test binaries passed, and the msodbcsql 18 comparison leg (--compare-with-msodbcsql) reports 26 parity, 0 divergences, 0 shared failures
New unit tests pin the procedure id that actually reaches the wire, since the whole point of the prepared-path change is invisible in the returned values: begin_execute_prepared_streams_into_sp_prepexec_when_unmaterialized, begin_execute_prepared_streams_into_sp_execute_when_handle_is_live, begin_execute_prepared_without_streamed_params_completes, begin_execute_prepared_keeps_the_orphan_for_the_caller, cancel_streamed_write_disarms_the_prepexec_handle_capture, cancel_streamed_write_rearms_an_unsent_connection_reset.

New e2e tests: DataAtExecutionKeepsStatementPreparedAcrossExecutes, DataAtExecutionLoopsOverMultipleStreamedParams, SQLCancelAbandonsDataAtExecutionSequence, DataAtExecutionNullParam, DataAtExecutionChunkAfterNullReturnsHY020, ExecDirectDataAtExecutionInterleavesWithBoundParams, ExecDirectDataAtExecutionStatementIsReusable, ExecDirectDuringNeedDataReturnsHY010, ConnectionRecoversAfterDataAtExecutionDiesOnTheWire, DataAtExecutionKeepsParamOrderOnCachedPlan, FreeStatementMidDataAtExecutionIsRejectedUntilCancelled.

The ctest JUnit output is per-binary, so a skipped test would still show execute_test ... Passed. The DAE, SQLExecDirect, and cancel cases were additionally run by name against a live server to confirm they execute rather than skip  14/14.

## Related Issues
https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47483


mssql-tds/src/connection/tds_client.rs
```diff
@@ -1495,34 +1499,84 @@ impl TdsClient {
...
+        else {
+            return;
+        };
+        // Same reasoning as `abort_streamed_write`: a cancelled `sp_prepexec`
+        // never yields its `@handle`, so the capture must not stay armed.
+        self.abort_pending_prepare_capture();
+
+        if context.message.nothing_sent() {
+            self.execution_context.set_has_open_batch(false);
+            return;
```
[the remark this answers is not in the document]
shiwanigupta0809: Valid, and the consequence is worse than a missed reset: `reset_pending` would stay `true` forever, so a pooled connection would keep reporting an outstanding reset that no request ever carries.

Added `SuspendedMessage::abandon`, which returns a non-`None` `reset_mode` to the network writer, and called it from the `nothing_sent()` branch of `cancel_streamed_write` instead of dropping the message.

Covered by a new unit test, `cancel_streamed_write_rearms_an_unsent_connection_reset`: it arms a reset, parks a streamed `sp_prepexec`, asserts nothing has reached the network, cancels, then asserts the next request's first packet carries `RESETCONNECTION`. I verified the test is load-bearing by commenting out the `abandon` call — it fails with `left: 0, right: 8`.

mssql-odbc/tests/e2e/tests/session_recovery_test.cpp
```diff
@@ -158,3 +158,63 @@ TEST_F(SessionRecoveryLiveTest, StaleHandleAfterReconnectIsInvalidatedAndReprepa
...
+    std::string original_spid = GetStringScalar(stmt_, "SELECT @@SPID");
+    ASSERT_FALSE(original_spid.empty());
+    std::string original_connect_time = GetStringScalar(stmt_, kConnectTimeSql);
+    ASSERT_FALSE(original_connect_time.empty());
+
+    // Park a streamed parameter: SQLExecute returns SQL_NEED_DATA and
+    // SQLParamData opens the parameter for data.
+    ASSERT_SQL_OK(Prepare("SELECT ? AS v"), SQL_HANDLE_STMT, stmt_);
+    std::vector<SQLCHAR> buffer(64 * 1024, 'x');
+    SQLLEN ind = SQL_LEN_DATA_AT_EXEC(0);
```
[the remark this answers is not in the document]
shiwanigupta0809: Correct, and this was the most useful comment of the four — the test was passing while proving nothing. `dae_expected_length` maps `SQL_LEN_DATA_AT_EXEC(0)` to `Some(0)`, so the over-send guard in `put_data.rs` rejected the 64 KiB chunk with 22026 before the client checkout, and the socket was never touched. The trailing recovery assertions still passed because the probe statement independently discovered the killed connection.

Changed the indicator to `SQL_DATA_AT_EXEC` so no length is declared and the chunk reaches the transport.

That exposed a second problem: my assertions were wrong about where the failure lands. I instrumented the test and ran it six times — every run is identical:

```
put=0 state=
paramdata=-1 state=HY000
```

`SQLPutData` returns `SQL_SUCCESS` with no diagnostic; 64 KiB fits in the socket send buffer, so the reset is not observed on the write. The sequence dies at `SQLParamData`, which writes the PLP terminator, finalizes the RPC and reads the response — the first call that has to hear back from the server. The assertions now match that, and the comment explains why 22026 or HY010 appearing there would mean the abort path was skipped again.

Full e2e suite is 24/24.

mssql-odbc/src/api/param_data.rs
```diff
@@ -0,0 +1,336 @@
...
+        };
+        free_errors(&mut stmt_state);
+
+        if !stmt_state.needs_data() {
+            error!("SQLParamData: called without an active data-at-execution sequence (HY010)");
+            post_diag(&mut stmt_state, ERR_FUNCTION_SEQUENCE);
+            return SQL_ERROR;
+        }
+
+        let dae = stmt_state.dae.as_mut().expect("checked just above");
```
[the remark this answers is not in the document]
shiwanigupta0809: Agreed — a panic here unwinds into the application and poisons the statement mutex, which is a worse outcome than a wrong diagnostic. Replaced the `expect` with a `let ... else` that logs, posts `HY010` via `ERR_FUNCTION_SEQUENCE`, and returns `SQL_ERROR`.

mssql-odbc/src/handles/stmt.rs
```diff
@@ -182,6 +182,147 @@ pub(crate) struct StmtState {
...
+    /// Parks the client back on the sequence after a write completes.
+    pub(crate) fn return_client(&mut self, client: TdsClient) {
+        self.client = Some(client);
+        self.call_in_flight = false;
+    }
+
+    /// `true` while `SQLPutData` or `SQLParamData` holds the client for network
+    /// I/O. `SQLCancel` is the one call an application may make on a busy
+    /// statement from another thread, and during this window the sequence is
+    /// not its to reset: the owning thread is about to write back into it.
```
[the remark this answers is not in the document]
shiwanigupta0809: Fixed — the dangling "its" now names its referent: "the sequence is not `SQLCancel`'s to reset".

David-Engel (commented):
## Summary

Reviewed against the msodbcsql reference (`sqlccmd.cpp` DAE paths, `BATCHCTX::Cancel` in `TdsSend.cpp:995-1147`), MS-TDS 2.2.3.1 packet framing, and the ODBC `SQLPutData` / `SQLParamData` / `SQLCancel` specs.

Well-built feature. The state-aware cancel genuinely matches `BATCHCTX::Cancel` (drop-locally vs. `EOM|IGNORE`+drain vs. close), the `is_first_packet` / `abandon` reset re-arming is a subtle detail handled correctly, the `packet_writer` ignore-packet length fix is a real latent bug (the old `saved_position = 0` meant the packet was sent as *zero* bytes), and the TDS-layer tests pin the actual proc id on the wire rather than the returned values. CI is green, diff coverage 92%, work item linked.

Two blocking bugs, both in the "Need Data" barrier. Both are reproduced below with local test probes rather than asserted.

**Blocking**
1. `SQLExecute` during Need Data panics in debug builds and posts no diagnostic in release &mdash; the `needs_data()` guard is unreachable in the real parked state. Inline on `execute.rs`.
2. `SQLFreeHandle(SQL_HANDLE_STMT)` during Need Data leaves the connection permanently unusable. Inline on `stmt.rs`.

**Suggestion**
3. The Need Data barrier only covers two entry points (below, on unchanged code).
4. Test gap: `sp_execute` with interleaved materialized/streamed parameters. Inline on `execute_test.cpp`.

**Nit**
5, 6. Inline. 7. PR description, below.

---

### 3. (Suggestion) The Need Data barrier only covers two entry points

I checked the neighbours: `SQLPrepare` is safely rejected with 24000 via `STMT_STATE_EXEC_STARTED`, and `SQLFetch` / `SQLMoreResults` fail loudly, so those are fine. But `SQLFreeStmt(SQL_RESET_PARAMS)` (`bind_param.rs:267-297`) has no state guard and clears `bound_params` under an open sequence, after which `dae_current_value_ptr()` hands `SQLParamData` a null token instead of the application's. `SQLBindParameter` can likewise re-point a `DaeParam::bound_index` at a different binding. Neither is memory-unsafe (both go through `.get()`), but a `needs_data()` &rarr; HY010 check on the parameter-mutating calls would close it. Raising this on unchanged code because the new `dae` field is what makes those paths reachable.

### 7. (Nit) PR description

Three things worth fixing before merge:

- The mojibake `╦ô├ç├ö` appears in several places where em-dashes belong.
- "Previously SQLExecute handled DAE by abandoning the prepared handle and streaming ad-hoc sp_executesql" describes an intra-branch state, not `main`. `put_data.rs`, `param_data.rs` and `cancel.rs` are all new files, so against `main` there was no DAE path at all. A reviewer reading the description against the base branch will look for code that was never there.
- Materialized `SQL_C_BINARY` parameter support (`bound_param_to_value`, `null_value`, `conversion_matrix`) is a new public capability that the description does not mention.

mssql-odbc/src/api/execute.rs
```diff
@@ -162,6 +266,15 @@ fn stage_execution(stmt: &StmtHandle) -> Result<Execution, SqlReturn> {
         return Err(SQL_ERROR);
     }
 
+    // A statement awaiting data-at-execution input is in the ODBC "Need Data"
+    // state, where every function other than SQLPutData/SQLParamData/SQLCancel
+    // and the diagnostic calls is a sequence error rather than a cursor error.
+    if stmt_state.needs_data() {
```
David-Engel: **Blocking.** This guard is unreachable in the state it is meant to catch.

`park_dae_client` leaves `stmt_state.prepared == None` (the plan is parked inside `DaeState`), but this check sits *below* the `debug_assert!(prepared.is_some())` and `prepared.is_none()` early return at lines 259-267. The real Need Data state therefore never gets here.

Reproduced with the realistic state (`prepared = None`, `dae = Some`):

```
panicked at mssql-odbc/src/api/execute.rs:260:5:
SQLExecute: statement not prepared - DM should have rejected this
panicked ... called `Result::unwrap()` on an `Err` value: PoisonError { .. }
```

`ffi_entry!` catches the unwind, but the statement mutex is now poisoned, so every subsequent `SQLPutData` / `SQLParamData` / `SQLCancel` fails on the poisoned lock, the parked `TdsClient` is stranded, and the connection is unusable. In release `debug_assert!` compiles out and the early return fires with no `post_diag` at all, so the application gets `SQL_ERROR` and `SQLGetDiagRec` returns `SQL_NO_DATA` rather than the HY010 this PR documents.

`execute_while_awaiting_data_returns_function_sequence_error` passes only because it calls `set_prepared()` first, fabricating a state the driver cannot produce. `SQLExecDirect` is fine - its guard is the first check.

Fix: move this `needs_data()` block above the `debug_assert!` / `prepared.is_none()` check, and change the unit test to leave `prepared = None` so it exercises the state `park_dae_client` actually creates.

mssql-odbc/src/handles/stmt.rs
```diff
@@ -289,6 +430,48 @@ impl StmtState {
...
+    /// client, if the sequence still held one. Call after a DAE sequence
+    /// completes, is cancelled, or fails.
+    ///
+    /// The prepared plan and orphaned handle the sequence was holding are
+    /// restored in the same step, so the statement is never observable as idle
+    /// but unprepared. The caller owns the returned client and must dispose of
+    /// it — `None` means a `SQLPutData` / `SQLParamData` call has it checked
+    /// out and will dispose of it itself.
+    #[must_use = "the parked client must be cancelled and returned to the connection"]
+    pub(crate) fn take_dae(&mut self) -> Option<TdsClient> {
```
David-Engel, line 443: **Blocking.** Nothing on the `SQLFreeHandle(SQL_HANDLE_STMT)` path calls `take_dae()`, so freeing a statement mid-sequence bricks the connection.

`best_effort_unprepare_on_free` (`free_handle.rs:200-247`) reads `prepared.take()` and `pending_unprepare.take()` directly. With `dae = Some` both yield `None`, `handles` is empty, and it returns early. `free_stmt` then drops the `StmtHandle`, taking the parked `TdsClient` with it, while the DBC still holds `client == None` and `active_stmt == Some(<freed handle>)`.

Proved with a probe test in `free_handle.rs`:

```
assert_eq!(d.active_stmt, Some(freed));  // passes - still points at the freed statement
assert!(d.client.is_none());             // passes - the connection has no client left
```

Every later `claim_connection` on that DBC returns `ERR_CONNECTION_BUSY`, or `ERR_NO_ACTIVE_TDS_CLIENT` if the allocator happens to reuse the address. The connection stays dead until `SQLDisconnect`. This also contradicts the PR description: "It rides along with the parked state and is released by the next execute or by `SQLFreeHandle`."

Fix - in `best_effort_unprepare_on_free`, before reading `prepared` / `pending`:

```rust
let parked = stmt.inner.lock().ok().and_then(|mut s| s.take_dae());
if let Some(mut client) = parked {
    dbc.runtime.block_on(client.cancel_streamed_write());
    return_client_idle(dbc, handle, client);
}
```

The existing unprepare logic then sees the restored `prepared` / `pending_unprepare` and releases the plan as usual. `SQLDisconnect` is safe as-is - I checked, it drops the STMTs and resets `client` / `active_stmt` wholesale.

mssql-odbc/tests/e2e/tests/execute_test.cpp
```diff
@@ -112,6 +118,479 @@ TEST_F(PrepareExecuteLiveTest, SingleCharParam) {
...
+    ASSERT_EQ(SQL_NEED_DATA, SQLParamData(stmt_, &value_ptr));
+    const char first[] = "first";
+    ASSERT_SQL_OK(SQLPutData(stmt_, const_cast<char*>(first), 5),
+                  SQL_HANDLE_STMT, stmt_);
+    ASSERT_SQL_OK(SQLParamData(stmt_, &value_ptr), SQL_HANDLE_STMT, stmt_);
+    ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);
+    EXPECT_EQ("first", GetColumnChar(1));
+    ASSERT_SQL_OK(SQLCloseCursor(stmt_), SQL_HANDLE_STMT, stmt_);
+
+    // Second streamed execute: no re-prepare, the cached handle is reused.
```
David-Engel, line 281: **Suggestion.** This is the only test that reuses a cached handle, and it binds a single parameter, so it never reorders anything.

`DataAtExecutionInterleavesWithBoundParams` does reorder, but only on the `sp_prepexec` route, which carries an `@params` declaration string where the `@Pn` names resolve. The `sp_execute` route has no declaration at all: name binding runs straight against the cached plan while the wire order is materialized-then-streamed. That is exactly the combination the order-stability argument in `data-at-execution-streaming.md` rests on, and it is the one nothing covers.

Worth one test: prepare `SELECT ? + ? + ?` with `@P2` streamed, execute it twice, assert the result both times.

mssql-odbc/src/api/sqlstate.rs
```diff
@@ -26,6 +26,7 @@ pub(crate) const SQLSTATE_08007: [u8; 5] = *b"08007";
 pub(crate) const SQLSTATE_08S01: [u8; 5] = *b"08S01";
 pub(crate) const SQLSTATE_22002: [u8; 5] = *b"22002";
 pub(crate) const SQLSTATE_22003: [u8; 5] = *b"22003";
+pub(crate) const SQLSTATE_22026: [u8; 5] = *b"22026";
```
David-Engel: **Nit.** `SQLSTATE_22026` lands before `SQLSTATE_22018`, breaking the numeric ordering of this block.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -167,6 +182,24 @@ pub(crate) unsafe fn bound_param_to_value(
...
+pub(crate) fn dae_placeholder_type(
+    c_type: SqlSmallInt,
+) -> Result<StreamedSqlType, ParamBuildError> {
+    match c_type {
+        SQL_C_CHAR => Ok(StreamedSqlType::VarcharMax),
+        SQL_C_WCHAR => Ok(StreamedSqlType::NVarcharMax),
+        SQL_C_BINARY => Ok(StreamedSqlType::VarBinaryMax),
+        other => Err(ParamBuildError::UnsupportedCType(other)),
+    }
+}
```
David-Engel, line 177: **Nit.** Missing blank line after the closing brace, so the following `/// A TDS value plus ...` doc comment reads as if it were attached to `dae_placeholder_type`.

mssql-odbc/src/api/execute.rs
saurabh500: **Blocking.** This `needs_data()` guard can't be reached in the state it targets. In the real Need Data state `park_dae_client` has moved the plan into `DaeState`, so `stmt_state.prepared == None`. The `debug_assert!(prepared.is_some())` at line 260 then panics in debug builds — it unwinds through `ffi_entry!`, poisons the STMT mutex, and strands the parked `TdsClient`, killing the connection. In release the `prepared.is_none()` early return at line 264 fires first and returns `SQL_ERROR` with no HY010 posted, so `SQLGetDiagRec` yields nothing.

`SQLExecDirectW` gets this right: its `needs_data()` check is the first thing in the safe path (`exec_direct.rs:971`). Fix: move this block above the `debug_assert!` / `prepared.is_none()` checks so it matches. Once reordered, `execute_while_awaiting_data_returns_function_sequence_error` should leave `prepared = None` — today it calls `set_prepared()` first, fabricating a state `park_dae_client` never produces.

(David-Engel flagged this too; confirming independently with the mechanism.)

mssql-odbc/src/handles/stmt.rs
saurabh500: **Blocking (free path).** Nothing on `SQLFreeHandle(SQL_HANDLE_STMT)` calls `take_dae()`. `best_effort_unprepare_on_free` (`free_handle.rs:216-229`) reads `prepared.take()` / `pending_unprepare.take()` directly; mid-DAE both live inside `DaeState`, so both are `None`, `handles` is empty, and it returns early. `free_stmt` then drops the `StmtHandle` — and the parked `TdsClient` with it — while the DBC still holds `client == None` and `active_stmt == Some(freed)`. Every later `claim_connection` on that DBC returns `ERR_CONNECTION_BUSY` (or `ERR_NO_ACTIVE_TDS_CLIENT`) until `SQLDisconnect`. This also contradicts the PR description: the parked client is not "released ... by `SQLFreeHandle`".

Fix in `best_effort_unprepare_on_free`, before reading prepared/pending:

```rust
let parked = stmt.inner.lock().ok().and_then(|mut s| s.take_dae());
if let Some(mut client) = parked {
    dbc.runtime.block_on(client.cancel_streamed_write());
    return_client_idle(dbc, handle, client);
}
```

`take_dae()` also restores `prepared` / `pending_unprepare`, so the existing unprepare logic then releases the plan as usual.

(David-Engel flagged this too; confirming independently with the mechanism and fix.)

shiwanigupta0809:

mssql-odbc/src/api/execute.rs
shiwanigupta0809: Confirmed and fixed.

Verified the ordering problem in source: `stage_execution` does `stmt_state.prepared.take()` before `park_dae_client` stores the plan inside `DaeState`, so a statement in the real Need Data state has `prepared == None`. The `needs_data()` block sitting below the `prepared.is_none()` early return was therefore unreachable in exactly the state it exists to catch.

Moved the `needs_data()` block above the `debug_assert!` / `prepared.is_none()` check, with a comment recording why the order matters so it isn't "tidied" back later.

Also fixed the test as you suggested. `execute_while_awaiting_data_returns_function_sequence_error` called `set_prepared()` first, which fabricated `prepared = Some` alongside `dae = Some` — a combination the driver cannot produce. It now sets only `dae` and asserts `prepared.is_none()` up front, so it exercises the state `park_dae_client` actually creates and would fail against the old ordering.

mssql-odbc/src/handles/stmt.rs
shiwanigupta0809: Confirmed, fixed — and the investigation turned up something that changes the severity, so flagging it rather than quietly claiming a clean fix.

**The code gap was real.** `best_effort_unprepare_on_free` reads `prepared` / `pending_unprepare` directly, and with `dae = Some` both are `None` because `take_dae()` is what restores them — so it returned early and the `StmtHandle` dropped with the client still parked inside `DaeState`.

Fixed in `best_effort_unprepare_on_free`, before the `prepared` / `pending` read. I used the existing `unwind_dae` rather than the inline snippet: it already does exactly that sequence — `take_dae()`, `cancel_streamed_write()`, `return_client_idle()` — and additionally clears `EXEC_STARTED`. Reusing it keeps one teardown path for abandoned DAE sequences instead of a second copy that can drift. As you noted, the unprepare below then sees the restored `prepared` / `pending_unprepare` and releases the plan as usual.

**But the scenario is not reachable through a Driver Manager.** I wrote an e2e test that frees a statement mid-sequence and expected it to demonstrate the stranded connection. It failed — against *both* mssql-odbc and msodbcsql 18, identically: `SQLFreeHandle` returned `SQL_ERROR`, not `SQL_SUCCESS`. That is spec-mandated DM behaviour:

> HY010 — **(DM)** The HandleType argument was SQL_HANDLE_STMT. SQLExecute, SQLExecDirect, SQLBulkOperations, or SQLSetPos was called with the statement handle and returned SQL_NEED_DATA. This function was called before data was sent for all data-at-execution parameters or columns.

Our `free_stmt` has no Need Data rejection of its own, so the error is unambiguously the DM's. The application cannot abandon a statement mid-DAE; it must call `SQLCancel` first. And `SQLDisconnect` drops the client wholesale, so that path never strands anything either.

So the fix is defense-in-depth — it matters when the driver is linked directly with no DM in front of it (which is how the unit tests drive it), not for DM-hosted applications. Still worth having, but I did not want to leave it described as a live connection-killing bug when the measured behaviour says otherwise.

Tests, retargeted accordingly:

- `free_stmt_mid_dae_drains_the_parked_sequence` (unit, no DM) asserts the sequence is unwound before the handle is dropped — this is the path the fix actually guards.
- `FreeStatementMidDataAtExecutionIsRejectedUntilCancelled` (e2e) pins the real DM-hosted contract: free is rejected with HY010, `SQLCancel` unwinds and returns the client, the free then succeeds, and the connection still executes. Passes on both drivers.

I have also reworded the PR description sentence you quoted — "released by the next execute or by `SQLFreeHandle`" was imprecise, since through a DM the release comes via `SQLCancel` and then the free.

One note on the unit test: my first attempt asserted `active_stmt == None` after the free and failed, because `DaeState::for_test` deliberately parks no client (`client: None`), so `take_dae()` returns `None` and `return_client_idle` never runs. The unit test now asserts only what that fixture can actually observe.

mssql-odbc/tests/e2e/tests/execute_test.cpp
shiwanigupta0809: Added `DataAtExecutionKeepsParamOrderOnCachedPlan`.

Your reading of the gap matches what I found: `DataAtExecutionKeepsStatementPreparedAcrossExecutes` is the only test that reuses a cached handle and it binds one parameter, so nothing is reordered; `DataAtExecutionInterleavesWithBoundParams` reorders but carries an `@params` declaration on every call.

The new test prepares `SELECT ? + ? + ?` with `@P2` streamed and executes it twice in a loop — the first pass goes out as `sp_prepexec`, the second against the cached handle, which is the case under test. Streaming the middle parameter puts an application-bound value last in the wire order, and string concatenation is asymmetric, so any misresolution changes the result rather than cancelling out. Both passes assert `"ABC"`.

It also asserts the token `SQLParamData` returns is the streamed parameter's buffer, so a silent shift in which parameter is treated as streamed fails too.

mssql-odbc/src/api/sqlstate.rs
shiwanigupta0809: Fixed — `SQLSTATE_22018` now precedes `SQLSTATE_22026`.

mssql-odbc/src/conversion/param_convert.rs
shiwanigupta0809: Fixed — blank line added after the closing brace of `dae_placeholder_type`.

mssql-odbc/src/api/put_data.rs
```diff
@@ -0,0 +1,656 @@
...
+            .as_ref()
+            .is_none_or(|dae| dae.cursor.is_none())
+        {
+            error!("SQLPutData: called outside an open data-at-execution parameter (HY010)");
+            post_diag(&mut stmt_state, ERR_FUNCTION_SEQUENCE);
+            return SQL_ERROR;
+        }
+    }
+
+    let is_null_put = strlen_or_ind == SQL_NULL_DATA || (data_ptr.is_null() && strlen_or_ind == 0);
```
Theekshna: **Blocking — `SQLPutData(hstmt, NULL, 0)` is treated as SQL NULL; msodbcsql does not do this.**

The second disjunct diverges from the driver this is claiming parity with.

In msodbcsql (`sqlccmd.cpp:4091-4120`), only `cbValue < 0` can be NULL. `cbValue == 0` falls into the `else` branch at `:4111` and is a *present, zero-length* value. And `ValidatePutDataArg` (`sqlccmd.cpp:10901-10911`) explicitly permits `rgbValue == NULL` for exactly the three C types this PR supports (`SQL_C_CHAR`, `SQL_C_WCHAR`, `SQL_C_BINARY`) without reinterpreting it as NULL:

```cpp
case SQL_C_BINARY:
case SQL_C_WCHAR:
case SQL_C_CHAR:
    if (rgbValue != NULL && cbValue < 0 && cbValue != SQL_DEFAULT_PARAM && cbValue != SQL_NULL_DATA)
    {
        wError = IDS_S1_090;
        goto ExitCS;
    }
    break;
```

Two consequences:

1. The value lands as SQL `NULL` instead of `''` / `0x`.
2. Because `is_null` short-circuits the close-time length check (`param_data.rs:141`), the sequence `SQL_LEN_DATA_AT_EXEC(10)` → `SQLPutData(NULL, 0)` → `SQLParamData` silently succeeds, where msodbcsql would report `22026` (`sqlccmd.cpp:5858-5867`).

Suggested fix — drop the second disjunct so a zero-length put commits the parameter to empty:

```rust
let is_null_put = strlen_or_ind == SQL_NULL_DATA;
```

If the current behaviour is intentional, please drop the "to match msodbcsql" framing from the surrounding comment, because the source says the opposite.

mssql-odbc/src/api/put_data.rs
```diff
@@ -0,0 +1,656 @@
...
+                post_diag(&mut stmt_state, ERR_INVALID_NULL_POINTER);
+            }
+            return SQL_ERROR;
+        }
+        let c_type = {
+            let Ok(stmt_state) = stmt.inner.lock() else {
+                error!("SQLPutData: stmt mutex poisoned resolving current C type");
+                return SQL_ERROR;
+            };
+            stmt_state.dae_current_c_type().unwrap_or_default()
```
Theekshna: **Blocking — `unwrap_or_default()` here silently truncates a `SQL_C_WCHAR` chunk.**

`dae_current_c_type()` returns `None` when the open parameter's binding has gone away, and `unwrap_or_default()` turns that into `0` — which is not "unknown", it is a value that takes the **byte** branch of `nts_byte_count`.

This is reachable from safe application code. The Need Data guard now covers `SQLExecute` (`execute.rs:263`), `SQLExecDirect`, `SQLParamData` and `SQLFreeHandle`, but not `SQLFreeStmt(SQL_RESET_PARAMS)` / `SQLBindParameter`. `bind_param.rs:291` does an unguarded `stmt_state.bound_params.clear()`, after which `dae_current_bound_param()` returns `None`.

Concretely, for a `SQL_C_WCHAR` buffer holding `"AB"` (`41 00 42 00 00 00`) with `SQL_NTS`, the scan stops at the first `0x00` and **1 byte** is streamed instead of 4. No diagnostic is raised. `dae_current_value_ptr()` also starts handing the application a null token from `SQLParamData`.

Per the spec every function other than `SQLPutData` / `SQLParamData` / `SQLCancel` / the diagnostic and alloc-free calls is `HY010` in the Need Data state (S8-S11), so the root fix is to extend the `needs_data()` guard to `SQLFreeStmt(SQL_RESET_PARAMS | SQL_UNBIND)` and `SQLBindParameter`.

At minimum, make the missing C type an explicit error rather than a silent `0` here:

```rust
let Some(c_type) = stmt_state.dae_current_c_type() else {
    error!("SQLPutData: open DAE parameter has no binding");
    post_diag(&mut stmt_state, ERR_FUNCTION_SEQUENCE);
    return SQL_ERROR;
};
c_type
```

mssql-odbc/src/api/exec_common.rs
```diff
@@ -185,11 +262,36 @@ pub(super) fn flush_pending_unprepare(
...
+    pub(super) params: Vec<RpcParameter>,
+    /// Every DAE entry, in original parameter order.
+    pub(super) dae_params: Vec<DaeParam>,
+}
+
+fn dae_expected_length(indicator: SqlLen) -> Option<usize> {
+    if indicator == SQL_DATA_AT_EXEC {
+        None
+    } else {
+        Some((SQL_LEN_DATA_AT_EXEC_OFFSET - indicator) as usize)
```
Theekshna: **`SQL_LEN_DATA_AT_EXEC(0)` diverges from msodbcsql.**

An indicator of exactly `SQL_LEN_DATA_AT_EXEC_OFFSET` maps to `Some(0)` here, so `put_data.rs:275` rejects the first byte with `22026` and `param_data.rs:141` rejects any non-empty value at close.

msodbcsql guards *both* of its length checks with `cbDAEDataTotal > 0` and therefore skips validation entirely for a declared total of zero:

- `sqlccmd.cpp:4136-4141` (per-`SQLPutData` overflow)
- `sqlccmd.cpp:5858-5867` (close-time exact match)

Suggested fix:

```rust
fn dae_expected_length(indicator: SqlLen) -> Option<usize> {
    if indicator == SQL_DATA_AT_EXEC {
        return None;
    }
    match (SQL_LEN_DATA_AT_EXEC_OFFSET - indicator) as usize {
        0 => None,
        n => Some(n),
    }
}
```

Note the rest of the length handling in this PR does match msodbcsql exactly — the per-put overflow check, the close-time `!=` comparison, and skipping the check for a NULL value are all faithful. This is just the zero case.

mssql-odbc/src/api/free_handle.rs
```diff
@@ -364,6 +376,46 @@ mod tests {
         unsafe { sql_free_handle(SQL_HANDLE_ENV, env) };
     }
 
+    /// Freeing a statement mid data-at-execution must hand the parked client
+    /// back to the connection. Dropping the handle with the client still
```
Theekshna: Nit — leftover from an edit: the first sentence is duplicated and cut off mid-clause (`"...Dropping the handle with the client still"` runs straight into a second `"Freeing a statement mid data-at-execution must drain..."`). The second version reads better; delete the first two lines.

mssql-odbc/src/api/put_data.rs
Theekshna: **`unwrap_or_default()` here can silently truncate a `SQL_C_WCHAR` chunk.**

`dae_current_c_type()` returns `None` when the open parameter's binding has gone away, and `unwrap_or_default()` turns that into `0` — which is not "unknown", it is a value that falls into the **byte** branch of `nts_byte_count`.

The reachable path is an unguarded `stmt_state.bound_params.clear()` at `bind_param.rs:291` (`SQLFreeStmt(SQL_RESET_PARAMS)`), after which `dae_current_bound_param()` returns `None`. For a `SQL_C_WCHAR` buffer holding `"AB"` (`41 00 42 00 00 00`) with `SQL_NTS`, the scan stops at the first `0x00` and **1 byte** is streamed instead of 4, with no diagnostic. `dae_current_value_ptr()` also starts handing the application a null token from `SQLParamData`.

To be fair about severity: a conformant Driver Manager rejects `SQLFreeStmt`/`SQLBindParameter` in the Need Data state (S8-S11) with `HY010` before the call reaches the driver, and msodbcsql leans on exactly that — its own equivalent check carries the comment `// NOTE: this should be enforced by the driver manager` (`sqlccmd.cpp:4489`). But it still keeps the defensive check, and this driver is loaded **without** a DM by `mssql-python`'s `ddbc_bindings.cpp`, which resolves the entry points itself. So the path is live for that consumer.

Minimum fix — don't let a missing binding become a valid-looking C type:

```rust
let Some(c_type) = stmt_state.dae_current_c_type() else {
    error!("SQLPutData: open DAE parameter has no binding");
    post_diag(&mut stmt_state, ERR_FUNCTION_SEQUENCE);
    return SQL_ERROR;
};
c_type
```

Better, and closer to msodbcsql: extend the existing `needs_data()` guard (already on `SQLExecute`, `SQLExecDirect`, `SQLParamData`, `SQLFreeHandle`) to `SQLFreeStmt(SQL_RESET_PARAMS | SQL_UNBIND)` and `SQLBindParameter`.

mssql-odbc/src/api/exec_common.rs
Theekshna: **`SQL_LEN_DATA_AT_EXEC(0)` diverges from msodbcsql.**

An indicator of exactly `SQL_LEN_DATA_AT_EXEC_OFFSET` maps to `Some(0)` here, so `put_data.rs:275` rejects the first byte with `22026` and `param_data.rs:141` rejects any non-empty value at close.

msodbcsql guards *both* of its length checks with `cbDAEDataTotal > 0`, so a declared total of zero disables validation entirely rather than pinning the value to empty:

- per-`SQLPutData` overflow, `PutParamData` — `sqlccmd.cpp:4548-4549`

  ```cpp
  if (pBindInfo->cbDAEDataTotal > 0 && 
      pBindInfo->cbDAEDataTotal < pBindInfo->cbDataAppGiven)
  ```

- close-time exact match, `ProcessDAEParam` — `sqlccmd.cpp:6010-6011`

  ```cpp
  pBindInfo->cbDAEDataTotal > 0 && 
  pBindInfo->cbDAEDataTotal != pBindInfo->cbDataAppGiven)
  ```

Suggested fix:

```rust
fn dae_expected_length(indicator: SqlLen) -> Option<usize> {
    if indicator == SQL_DATA_AT_EXEC {
        return None;
    }
    match (SQL_LEN_DATA_AT_EXEC_OFFSET - indicator) as usize {
        0 => None,
        n => Some(n),
    }
}
```

Everything else about the length handling matches msodbcsql exactly — the per-put overflow check, the close-time `!=` comparison, and skipping the check for a NULL value. This is only the zero case.

mssql-odbc/src/api/free_handle.rs
Theekshna: Nit — leftover from an edit: the first sentence is duplicated and cut off mid-clause (`"...Dropping the handle with the client still"` runs straight into a second `"Freeing a statement mid data-at-execution must drain..."`). The second version reads better; delete the first two lines.

mssql-odbc/src/api/put_data.rs
shiwanigupta0809: Good catch, thanks — you're right that `0` isn't a safe default here, it takes the narrow branch and silently truncates.

Went with the `let ... else` you suggested: a missing binding now returns `HY010` instead of guessing a C type. Added a small test pinning the truncation itself (4 bytes vs 1) so the regression can't creep back.

mssql-odbc/src/api/exec_common.rs
shiwanigupta0809: You're right, thanks — zero should mean "unspecified", not "must be empty".

Applied your suggested fix. I also spotted `sqlccmd.cpp:4160` treating a zero total the same as `NO_PARAM_LENGTH`, which backs this up nicely. Test added for the zero case.

mssql-odbc/src/api/free_handle.rs
shiwanigupta0809: Thanks — that was a leftover from an edit on my side. Removed the duplicate, and the comment was also pointing at an old test name, so I've fixed that too.

Theekshna (approved):

David-Engel (approved):
## Approved

All five findings from my earlier review are addressed, and I re-verified each fix against the updated head (`8765595f`) rather than taking the replies at face value.

**Blocking, both fixed**

- **`SQLExecute` guard ordering** — the `needs_data()` block now sits above the `debug_assert!` / `prepared.is_none()` check, with a comment recording why the order matters. I re-ran my original probe in the realistic parked state (`prepared == None`, `dae == Some`): HY010 is posted, no panic, no mutex poisoning. The unit test dropping `set_prepared()` and asserting `prepared.is_none()` up front is the right correction — it now fails against the old ordering.
- **`SQLFreeHandle` mid-sequence** — reusing `unwind_dae` is a better fix than the inline snippet I suggested: one teardown path instead of a second copy that can drift, and it clears `EXEC_STARTED` too. Thank you for measuring the Driver Manager behaviour and correcting my severity rather than quietly banking the finding. You are right that a conforming DM rejects the call with HY010 (a documented `(DM)` clause), so this is defense-in-depth for direct-linked use, not a live connection killer. Keeping the fix and retargeting the tests to what each one can actually observe is the correct call.

**Suggestions, both addressed**

- The `SQLPutData` `SQL_NTS` guard is the important half of the Need Data barrier point, and `nts_byte_count_narrows_a_wide_buffer_when_the_c_type_defaults` pins exactly why: a defaulted C type of 0 takes the narrow branch, so `"AB"` in a UTF-16 buffer measured 1 byte instead of 4 and dropped the rest with no diagnostic. Refusing beats guessing.
- `DataAtExecutionKeepsParamOrderOnCachedPlan` is precisely the gap: two passes so the second runs against the cached plan with no `@params` declaration, asymmetric concatenation so misresolution cannot cancel out, plus the token assertion.

**Nits** — SQLSTATE ordering, the blank line, the description mojibake, and the `SQLFreeHandle` sentence are all fixed.

**The `SQL_LEN_DATA_AT_EXEC(0)` fix is a good catch that was not in my review.** Treating a zero total as "must be exactly zero bytes" would have rejected the first byte with `22026`. I checked both msodbcsql citations and they are exact: `cbDAEDataTotal > 0 &&` guards the per-put check at `sqlccmd.cpp:4548` and the close-time check at `:6010`, and the comment there says "(where n>0)" in as many words.

**Local validation on `8765595f`**

- `cargo fmt --check` and `cargo clippy --workspace --all-targets` both clean.
- `cargo nextest -p mssql-odbc -p mssql-tds --lib`: 2675 passed. The only 4 failures are the known `certificate_validator` fixture failures (test certs are not generated locally), unrelated to this PR.
- Diff coverage 92% against the 85% gate, with `free_handle.rs`, `param_buffer.rs`, `stmt.rs`, `tds_client.rs` and `packet_writer.rs` all at 100%.

Approving now — the required checks gate the merge regardless, and holding the approval only to watch the ADO run finish would force a re-approval on any commit that follows it. Two cosmetic leftovers you can take or leave: the "Previously SQLExecute handled DAE by..." paragraph still describes an intra-branch state rather than `main`, and materialized `SQL_C_BINARY` parameter support is still not called out in the description.

Nice work on this one. The msodbcsql grounding throughout, and the willingness to report a measurement that cut against your own PR, made it a straightforward re-review.

PR #362, opened 2026-08-22
Vendor OpenSSL for the fuzz build to survive apt mirror outages

saurabh500:
## Summary

The scheduled **GH-Rust Fuzz Test** pipeline (ADO definitionId 2207) had failing runs on 2026-08-19 (builds **167989**, **168079**, **168173**) that were **not** fuzz crashes. They failed because the Linux dependency install (`scripts/install-deps.sh`) could not reach the Ubuntu apt mirror (`azure.archive.ubuntu.com`, connection timed out) across all 5 retries, so `pkg-config` and `libssl-dev` were never installed. `mssql-tds` links system OpenSSL on Linux (via `native-tls` and a direct `openssl` dependency), so `openssl-sys 0.9.117` failed to build and the fuzz target binaries never compiled.

Evidence from build 167989 (`Run Fuzzer` step):

```
cargo:warning=Could not find directory of OpenSSL installation ...
... `libssl-dev` on Ubuntu ...
... requires the `pkg-config` utility ... `pkg-config` could not be found.
openssl-sys = 0.9.117
Error: failed to build fuzz script: ... "cargo" "build" ... "--bin" "fuzz_tds_client"
No artifacts directory found
```

`FuzzCrashesFound` was `false` on every run and "Prepare Crash Artifacts" was skipped, so no crash artifact was produced — libFuzzer never ran because the harness didn't compile.

## Change

Build OpenSSL from **vendored source** for the fuzz crate only, by adding `openssl = { version = "0.10.66", features = ["vendored"] }` scoped to non-Windows/non-macOS targets in `mssql-tds/fuzz/Cargo.toml`. This makes `openssl-sys` (shared by both `native-tls` and the direct `openssl` dependency) compile from source, so the fuzz build no longer needs `libssl-dev`/`pkg-config` from apt. Only a C compiler and perl are required, both already present on the CI image.

Scoped to the fuzz crate (`publish = false`); the shipped `mssql-tds` library build is unchanged.

## Testing

- `cargo metadata` in `mssql-tds/fuzz` confirms the manifest resolves and the vendored `openssl` dependency is correctly scoped to Linux/other non-macOS Unix targets.
- Full `cargo fuzz` compilation is Linux/nightly-only and will be validated by the pipeline.

## Related work item

Fixes [AB#47487](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47487)

### DIFF
--- a/mssql-tds/fuzz/Cargo.toml
+++ b/mssql-tds/fuzz/Cargo.toml
@@ -14,6 +14,17 @@ tokio = { version = "1.45.1", features = ["full"] }
 arbitrary = { version = "1.3", features = ["derive"] }
 async-trait = "0.1"
 
+# Build OpenSSL from vendored source for the fuzz targets instead of linking the
+# system library. mssql-tds pulls in openssl-sys on Linux (via native-tls and the
+# direct `openssl` dependency), which normally needs `libssl-dev` + `pkg-config`
+# from apt. Scheduled fuzz runs were failing to compile openssl-sys whenever the
+# Ubuntu apt mirror was transiently unreachable and those packages couldn't be
+# installed. Vendoring makes the fuzz build self-contained (only a C compiler and
+# perl are required, both present on the CI image) so a flaky mirror no longer
+# breaks fuzzing. Scoped to the fuzz crate; the shipped library build is unchanged.
+[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]
+openssl = { version = "0.10.66", features = ["vendored"] }
+
 # Prevent this from interfering with workspaces
 [workspace]
 members = ["."]


David-Engel (commented):
## Summary

Single-manifest change that turns on `openssl/vendored` for the fuzz crate on non-Windows/non-macOS targets. I verified the mechanism locally: `cargo metadata --filter-platform x86_64-unknown-linux-gnu` in `mssql-tds/fuzz` resolves `openssl-sys 0.9.117` with features `vendored,openssl-src` (pulling `openssl-src 300.6.1+3.6.3`), so feature unification does reach the `openssl-sys` instance that `native-tls` uses — the build-time `libssl-dev`/`pkg-config` requirement really does go away for the fuzz targets. The cfg gate and the `0.10.66` floor mirror `mssql-tds/Cargo.toml` exactly, the change is scoped to a `publish = false` crate with its own `[workspace]`, there is no committed `Cargo.lock` under `mssql-tds/fuzz` to go stale, and the shipped library build is untouched.

The mechanism is right. My concern is whether it actually achieves the outcome in the title, because the fuzz jobs still die earlier in the pipeline on an apt outage.

### Blocking / needs an answer

**1. The job never reaches the cargo build when the mirror is down (`.pipeline/templates/validation-stages.yml:453-546`, `.pipeline/templates/install-dependencies.yml:50`)**

All three fuzz jobs run `install-dependencies.yml` before `fuzz-template.yml`, and every fuzz step is `condition: succeeded()`. Since #329 (`2dde48f6`, "Fail install-deps.sh when apt retries are exhausted", merged 2026-08-19 — the same day as builds 167989/168079/168173), `scripts/install-deps.sh` does:

```bash
if [ "$apt_install_ok" != true ]; then
    echo "ERROR: apt install failed after 5 attempts"
    exit 1
fi
```

So with today's `main`, an unreachable `azure.archive.ubuntu.com` fails the "Install linux dependencies" step and the fuzz steps are skipped — the vendored OpenSSL never gets a chance to help. The failure signature changes (an apt error instead of an `openssl-sys` error) but the scheduled run still fails.

To actually survive the outage, the fuzz jobs need a dependency path that tolerates apt being down, for example a fuzz-specific `installdepsparams` mode that installs a reduced set best-effort and does not `exit 1`, or skipping `install-deps.sh` for these jobs entirely. If the intent here is narrower — "remove the OpenSSL headers from the fuzz build's requirements" as one step of a larger fix — please say so in the description and retitle, because "survive apt mirror outages" isn't achieved by this commit alone.

**2. Vendoring may trade one apt dependency for another (`mssql-tds/fuzz/Cargo.toml:17-26`)**

The comment claims "only a C compiler and perl are required, both present on the CI image." `openssl-src` needs `make` plus a *full* perl — `Configure` uses core modules (`File::Compare`, `File::Copy`, `POSIX`, `FindBin`) that live in `perl-modules-*`, not `perl-base`. On `RUST-1ES-UBUSLIM` (a slim image), `build-essential` is installed by `install-deps.sh` from the same apt mirror, and `perl` is not in that list at all. If the slim image ships only `perl-base`, the vendored build fails in exactly the scenario this PR targets.

Worth confirming on the image, and cheap to make explicit — either add a preflight in `fuzz-template.yml`:

```bash
perl -MFile::Compare -MPOSIX -e 'print "perl ok\n"' && make -v
```

or add `perl` to the fuzz dependency set so the assumption is stated rather than implied.

### Suggestions

**3. Per-job build cost (`.pipeline/templates/validation-stages.yml:456-546`)** — `CARGO_TARGET_DIR` is set per job, so each of the three fuzz jobs compiles OpenSSL 3.6.3 from source (typically several minutes) with no cross-job cache. The serial "Additional Targets" job pays it once for six targets, so this is acceptable, but neither 30-minute job sets `timeoutInMinutes`. Worth a sanity check that the added build time doesn't push these jobs into a pool-level timeout.

**4. Keep the version tied to the library (`mssql-tds/fuzz/Cargo.toml:25`)** — `0.10.66` is duplicated from `mssql-tds/Cargo.toml`. If they ever diverge into different semver-incompatible majors, unification silently stops and `libssl-dev` quietly comes back as a requirement. A short pointer in the comment ("keep in sync with `mssql-tds/Cargo.toml`") would save the next person the investigation.

### Nit

**5.** The eight-line justification comment is long for a three-line dependency, and this crate declares `openssl` without using it in any fuzz target — the dependency exists purely for feature unification. Consider leading with that fact so a future cleanup pass doesn't delete it as an unused dependency:

```toml
# Feature-only dependency: enables `openssl-sys/vendored` for the openssl-sys
# instance shared with native-tls, so the fuzz build doesn't need libssl-dev/
# pkg-config. Not referenced by any fuzz target. Keep in sync with mssql-tds.
```

### CI

Required checks are red on `b118f98`: `mssql-rs Pull request validation` (including `Build Stage Test MacOS`) and `coverage-report`. The head commit is from 2026-08-22 and is well behind `main`, so some of this is likely staleness rather than this diff — merging `main` and re-running should give a clean signal before this goes out of draft.

*Automated review. No approval or change request implied; a human reviewer still owns the call.*

mssql-tds/fuzz/Cargo.toml
```diff
@@ -14,6 +14,17 @@ tokio = { version = "1.45.1", features = ["full"] }
 arbitrary = { version = "1.3", features = ["derive"] }
 async-trait = "0.1"
 
+# Build OpenSSL from vendored source for the fuzz targets instead of linking the
+# system library. mssql-tds pulls in openssl-sys on Linux (via native-tls and the
+# direct `openssl` dependency), which normally needs `libssl-dev` + `pkg-config`
+# from apt. Scheduled fuzz runs were failing to compile openssl-sys whenever the
+# Ubuntu apt mirror was transiently unreachable and those packages couldn't be
+# installed. Vendoring makes the fuzz build self-contained (only a C compiler and
```
David-Engel, line 22: The vendoring mechanism itself checks out. I resolved the Linux graph from `mssql-tds/fuzz` (`cargo tree --target x86_64-unknown-linux-gnu -e features -i openssl-sys`) and `openssl-sys 0.9.117` does pick up `vendored` -> `openssl-src`, and `native-tls` shares that same unified build, so `libssl-dev` + `pkg-config` genuinely drop out of the fuzz link.

What I don't think holds is the outcome. On current `main` an apt outage no longer reaches the fuzz build at all: `scripts/install-deps.sh` was changed in 2dde48f6 ("Fail install-deps.sh when apt retries are exhausted", #329 — merged the same day as builds 167989/168079/168173) to `exit 1` after five failed `apt update` / `apt install` attempts, and `.pipeline/templates/install-dependencies.yml` runs it as a plain `script:` step with no `continueOnError`. All three fuzz jobs (`validation-stages.yml:465-546`) go through that template, and every step in `fuzz-template.yml` is `condition: succeeded()`. So with an unreachable mirror the job now fails at "Install linux dependencies" and never reaches `cargo +nightly fuzz run`.

That makes this necessary-but-not-sufficient for AB#47487. To actually get a scheduled fuzz run through an outage it needs a companion change on the install side — e.g. dropping `libssl-dev`/`pkg-config` from `DEPS` for the fuzz path, or giving the fuzz jobs an install profile that tolerates apt failure. Either do that here, or narrow the PR/work-item claim to "removes the OpenSSL system dependency from the fuzz build" so 47487 isn't resolved on a scenario that still fails.

mssql-tds/fuzz/Cargo.toml
```diff
@@ -14,6 +14,17 @@ tokio = { version = "1.45.1", features = ["full"] }
...
+# perl are required, both present on the CI image) so a flaky mirror no longer
```
David-Engel, line 23: `both present on the CI image` is the load-bearing assumption of this fix, and nothing in the PR verifies it. The resolved vendored build is `openssl-src 300.6.1+3.6.3`, which needs `perl` (with the Configure-time core modules), `make`, and a C compiler on `PATH` — and on these agents `build-essential` comes from the very same `apt install` in `scripts/install-deps.sh` that the outage broke. A cc/linker on `RUST-1ES-UBUSLIM` is a safe bet (rustc needs one); `make` and a full `perl` on a slim image are less obviously there.

Cheap way to prove it before depending on it: add `which perl make cc; perl -v` to the "Install Fuzz Dependencies" step, or do a one-off run of the fuzz stage with install-deps skipped. If it builds without apt, the claim is verified and can be stated as such in this comment instead of asserted.

mssql-tds/fuzz/Cargo.toml
```diff
@@ -14,6 +14,17 @@ tokio = { version = "1.45.1", features = ["full"] }
...
+# breaks fuzzing. Scoped to the fuzz crate; the shipped library build is unchanged.
+[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]
+openssl = { version = "0.10.66", features = ["vendored"] }
```
David-Engel, line 26: Nit, worth capturing in the comment block: this quietly swaps the crypto backend under the fuzzers. Vendored resolves to `openssl-src 300.6.1+3.6.3`, while shipped Linux builds link the distro OpenSSL (3.0.x on Ubuntu 22.04), so `security::crypto` and `fuzz_der_framing` now exercise a different OpenSSL than customers run, and security fixes arrive via `openssl-src` bumps rather than apt. That is a fine trade-off for a `publish = false` harness, but "Scoped to the fuzz crate; the shipped library build is unchanged" reads as if there's no behavioral delta at all — a one-line note that the fuzzed OpenSSL now differs from the shipped one would keep the next reader honest.

Second-order, non-blocking: every fuzz job now builds OpenSSL from source on a cold target dir (three jobs, every three hours). The 1800s fuzz budget is unaffected since it's wall-clock capped, but agent time isn't free.

David-Engel (commented):
## Automated review — head `b118f982`

Scope: a single manifest change in `mssql-tds/fuzz/Cargo.toml`. Verified by resolving the Linux dependency graph (`cargo tree --target x86_64-unknown-linux-gnu -e features -i openssl-sys`) and by reading the full fuzz pipeline path (`fuzz-pipeline.yml` -> `validation-pipeline.yml` -> `validation-stages.yml` -> `install-dependencies.yml` + `fuzz-template.yml`) plus `scripts/install-deps.sh`.

The mechanism is correct: `vendored` unifies onto the single `openssl-sys 0.9.117` shared with `native-tls`, the target cfg mirrors `mssql-tds`'s own OpenSSL cfg exactly, the fuzz crate is its own workspace with no committed lock file, and the shipped build is untouched. `openssl-sys` is the only `-sys` crate left in the Linux fuzz graph, so the `libssl-dev`/`pkg-config` requirement really does go away.

Findings (full details inline):
1. Suggestion — the change alone won't restore scheduled fuzz runs during an apt outage; `install-deps.sh` hard-fails first (line 22).
2. Suggestion — the "only a C compiler and perl are required, both present on the CI image" assumption is unverified, and those tools also come from apt (line 23).
3. Nit — vendored OpenSSL 3.6.3 vs distro OpenSSL changes what the crypto fuzz targets exercise; worth stating in the comment block (line 26).

CI is red on this head (`coverage-report`, ADO `Pull request validation`, `Build Stage Test MacOS`). They look unrelated to a fuzz-only manifest change, but they should be triaged before this leaves draft.

No approval or change request is implied by this review.

PR #365, opened 2026-08-24
Complete Vector feature v2 (float16) serialization and bulk copy support

shiwanigupta0809:
## Description

Completes Phase 2 of Vector feature v2 (float16) support in `mssql-tds`. Phase 1
(feature negotiation and deserialization) landed in PR 7493 and left
`todo!("Phase 2: Float16 serialization")` in the value serializer, so float16
vectors could be read from the server but not written back. This PR implements
the write path and enables float16 by default.

**Serialization** — `tds_value_serializer.rs` replaces the `todo!()` with a real
implementation that narrows `f32` values to IEEE 754 half-precision using
round-to-nearest-even (via the `half` crate), matching the msodbcsql reference
driver's `Float32ToFloat16Bits`. Elements are written little-endian, 2 bytes each,
behind the standard 8-byte vector header.

**Bulk copy** — `bulk_copy_metadata.rs` now emits `vector(N, float16)` in generated
DDL when the column's base type is float16. Previously it always emitted
`vector(N)`, which SQL Server interprets as float32, so bulk copy of float16
columns produced a type mismatch.

**Default version** — `client_context.rs` flips the default `VectorVersion` from
`V1` to `V2`, resolving the `// TODO: make V2 as default when full V2 support is
added` left by Phase 1. Version negotiation is unchanged: the client advertises
v2, and the server may still ack v1, in which case behavior is identical to today.

### Tests

- 3 serializer unit tests covering float32 elements, float16 elements, and
  round-to-nearest-even behavior (`0.1f32` → `0x2E66`)
- 2 bulk copy metadata tests for float16 dimension math and DDL generation
- 2 `SqlVector` tests for `try_from_f16` and the higher float16 dimension ceiling
  (3996 vs. 1998)
- 1 feature negotiation test for a v2 ack
- 14 float16 integration tests covering query, parameter binding, output
  parameters, NULL handling, max dimensions, precision loss, mixed base types,
  and bulk copy

`mssql-tds/tests/common/mod.rs` now honors a `DB_DATABASE` env var so integration
tests can target a database with `PREVIEW_FEATURES = 1` enabled, defaulting to
`master` as before.

## Related Issues
https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/45020
https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/45021

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

### DIFF
--- a/mssql-tds/src/connection/client_context.rs
+++ b/mssql-tds/src/connection/client_context.rs
@@ -511,8 +511,7 @@ impl ClientContext {
                 port: 1433,
                 instance_name: None,
             },
-            // TODO: make V2 as default when full V2 support is added
-            vector_version: VectorVersion::V1,
+            vector_version: VectorVersion::V2,
             column_encryption_setting: ColumnEncryptionSetting::Disabled,
             column_encryption_key_store_providers: std::sync::Arc::new(
                 crate::security::keystore::ColumnEncryptionKeyStoreProviderRegistry::new(),
@@ -574,8 +573,7 @@ impl ClientContext {
                 port: 1433,
                 instance_name: None,
             },
-            // TODO: make V2 as default when full V2 support is added
-            vector_version: VectorVersion::V1,
+            vector_version: VectorVersion::V2,
             column_encryption_setting: ColumnEncryptionSetting::Disabled,
             column_encryption_key_store_providers: std::sync::Arc::new(
                 crate::security::keystore::ColumnEncryptionKeyStoreProviderRegistry::new(),
--- a/mssql-tds/src/datatypes/bulk_copy_metadata.rs
+++ b/mssql-tds/src/datatypes/bulk_copy_metadata.rs
@@ -710,8 +710,14 @@ impl BulkCopyColumnMetadata {
             SqlDbType::Variant => "sql_variant".to_string(),
             SqlDbType::Json => "nvarchar(max)".to_string(),
             SqlDbType::Vector => {
+                use crate::datatypes::sqldatatypes::VectorBaseType;
+
                 let dims = self.vector_dimensions()?;
-                format!("vector({})", dims)
+                // `vector(N)` implies the float32 base type; float16 must be spelled out.
+                match VectorBaseType::try_from(self.scale)? {
+                    VectorBaseType::Float32 => format!("vector({})", dims),
+                    VectorBaseType::Float16 => format!("vector({}, float16)", dims),
+                }
             }
         })
     }
@@ -1520,6 +1526,34 @@ mod tests {
             )
             .with_scale(0);
         assert_eq!(meta.get_sql_type_definition().unwrap(), "vector(3)");
+    }
+
+    #[test]
+    fn vector_dimensions_valid_float16() {
+        use crate::datatypes::sqldatatypes::VECTOR_HEADER_SIZE;
+        // Float16 (scale=1), element_size=2, 3 dimensions: header(8) + 3*2 = 14
+        let meta = BulkCopyColumnMetadata::new("v", SqlDbType::Vector, 0xF5)
+            .with_length(
+                (VECTOR_HEADER_SIZE + 3 * 2) as i32,
+                TypeLength::Variable((VECTOR_HEADER_SIZE + 3 * 2) as i32),
+            )
+            .with_scale(1);
+        assert_eq!(meta.vector_dimensions().unwrap(), 3);
+    }
+
+    #[test]
+    fn vector_sql_type_definition_float16() {
+        use crate::datatypes::sqldatatypes::VECTOR_HEADER_SIZE;
+        let meta = BulkCopyColumnMetadata::new("v", SqlDbType::Vector, 0xF5)
+            .with_length(
+                (VECTOR_HEADER_SIZE + 3 * 2) as i32,
+                TypeLength::Variable((VECTOR_HEADER_SIZE + 3 * 2) as i32),
+            )
+            .with_scale(1);
+        assert_eq!(
+            meta.get_sql_type_definition().unwrap(),
+            "vector(3, float16)"
+        );
     }
 
     #[test]
--- a/mssql-tds/src/datatypes/sql_vector.rs
+++ b/mssql-tds/src/datatypes/sql_vector.rs
@@ -306,4 +306,20 @@ mod tests {
         let vector = SqlVector::try_from_f32(values).unwrap();
         assert_eq!(vector.base_type(), VectorBaseType::Float32);
     }
+
+    #[test]
+    fn test_from_f16_valid() {
+        let vector = SqlVector::try_from_f16(vec![1.0, 2.0, 3.0]).unwrap();
+        assert_eq!(vector.base_type(), VectorBaseType::Float16);
+        assert_eq!(vector.dimension_count(), 3);
+        assert_eq!(vector.as_f32(), Some(&[1.0, 2.0, 3.0][..]));
+        assert_eq!(vector.total_size(), VECTOR_HEADER_SIZE + 3 * 2);
+    }
+
+    #[test]
+    fn test_f16_allows_more_dimensions_than_f32() {
+        let values = vec![0.0f32; VectorBaseType::Float16.max_dimensions() as usize];
+        assert!(SqlVector::try_from_f16(values.clone()).is_ok());
+        assert!(SqlVector::try_from_f32(values).is_err());
+    }
 }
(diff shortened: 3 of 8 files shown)


David-Engel (commented):
## Summary

This completes the float16 write path for Vector v2: `serialize_vector` replaces the `todo!()` with `half::f16::from_f32(..).to_bits()` (round-to-nearest-even, 2 bytes LE), `get_sql_type_definition` emits `vector(N, float16)` when the base type is encoded in `scale`, and the default `VectorVersion` flips from V1 to V2. I verified the wire path against `bulk_load.rs` (VECTOR metadata already writes length + scale), `bulk_copy_metadata.rs` (`VarLenScale` carries the base-type byte into `scale`), and `vectorfeature.rs` negotiation.

The core serialization logic looks correct. Locally in a clean worktree at this SHA, `cargo fmt --check` and `cargo clippy --workspace --all-features --all-targets -- -D warnings` pass, and all 56 vector-related unit tests pass. The blocker is on the test-infrastructure side.

## Blocking

**1. The newly enabled float16 integration tests fail on every CI platform.**

Build [169209](https://sqlclientdrivers.visualstudio.com/904996cc-6198-4d39-8540-eca72bdf0b7b/_build/results?buildId=169209) fails "Run Tests and Coverage" on Linux, Linux ARM, Windows, and macOS. Every float16 test exhausts all 6 nextest retries with the same server error:

```
thread 'vector_integration_tests::test_vector16_basic_deserialization' panicked at mssql-tds/tests/test_vector_type.rs:76:53:
called `Result::unwrap()` on an `Err` value: SqlServerError { ... message: "'float16' is not a recognized vector base type.", number: 195, server_name: Some("sql1") ... }
```

Failing tests: `test_vector16_basic_deserialization`, `test_vector16_parameter_basic`, `test_vector16_parameter_null`, `test_vector16_parameter_in_where_clause`, `test_vector16_parameter_max_dimensions`, `test_vector16_parameter_precision_loss`, `test_mixed_vector_base_type_parameters`, `test_vector16_output_parameter`, `test_vector16_input_output_parameter`, `test_vector16_output_parameter_null`, plus all five new `test_bulk_copy_vector16_*` / `test_bulk_copy_mixed_vector_base_types` tests.

The cause is that the CI SQL Server container runs against `master` with `PREVIEW_FEATURES` off. `mssql-tds/tests/common/mod.rs:63` now reads `DB_DATABASE`, but nothing supplies it — I grepped every template under `.pipeline/` and none sets `DB_DATABASE`, and none runs `ALTER DATABASE SCOPED CONFIGURATION SET PREVIEW_FEATURES = ON`. So `DB_DATABASE` always falls back to `master` in CI, and the pre-existing `#[ignore]` that made `test_vector16_basic_deserialization` green was removed without anything replacing it.

Two workable fixes:

- Make CI able to run them: add a step to `.pipeline/templates/sql-setup-template.yml` (and the ARM/alpine/macOS equivalents) that creates a preview-enabled database and enables the setting, then export `DB_DATABASE` alongside `SQL_PASSWORD` in every job that runs `mssql-tds` tests. This is the better long-term answer and makes `DB_DATABASE` actually load-bearing.
- Or keep them out of CI for now: restore `#[ignore = "..."]` on the float16 tests (matching the four that are still ignored) until the pipeline supports float16, and land the serializer/bulk-copy change with just the unit tests.

Either way the PR can't go green as-is.

**2. Merge conflict with `main`.** GitHub reports `mergeable: CONFLICTING`. Please merge `main` in before the next validation run so you don't burn a cycle.

## Suggestions

**3. `mssql-tds/tests/test_vector_type.rs` — inconsistent float16 test gating.**

Four tests are still `#[ignore]`d and still carry the old prologue:

```rust
#[ignore = "Not supported in pipeline currently so skipping"]
async fn test_vector16_metadata_fields() {
    client.execute("USE UserDatabase;".to_string(), ()).await.unwrap();
    client.execute("ALTER DATABASE SCOPED CONFIGURATION SET PREVIEW_FEATURES = ON;".to_string(), ()).await.unwrap();
```

at lines 150-163, 271-282, 509-520, and 547-558. Meanwhile `test_vector16_basic_deserialization` had exactly that prologue removed, and the 14 new float16 tests never had one. Whichever way finding #1 is resolved, these four should be brought in line — right now a reader can't tell which convention is current. As a side note, `ALTER DATABASE SCOPED CONFIGURATION` is persistent server state mutated from inside a test; if the preview database is set up by the pipeline instead, these prologues can go away entirely.

**4. `mssql-tds/tests/common/mod.rs:63` — document the new `DB_DATABASE` knob.**

```rust
context.database = env::var("DB_DATABASE").unwrap_or_else(|_| "master".to_string());
```

This changes the connection target for every integration test in the crate, not just the vector ones. `mssql-tds/tests/TRANSPORT_TESTS_README.md` already documents `DB_HOST` / `DB_USERNAME` / `SQL_PASSWORD`; please add `DB_DATABASE` there (and to `mssql-tds-bench/README.md`'s env table if it shares the contract) so the next person doesn't have to read `common/mod.rs` to discover it.

**5. `mssql-tds/src/connection/client_context.rs:514,576` — nothing gates writes on the negotiated version.**

Flipping the default to V2 is the right call, and the downgrade path is fine (server acks 1, `negotiated_version` becomes 1). But `VectorFeature::negotiated_version()` is still `#[allow(dead_code)]` with the comment "not used currently", so `serialize_vector` will happily emit a float16 payload on a connection where the server acked v1. Previously that combination hit the `todo!()` panic; now it becomes a server-side error instead. Worth either wiring the negotiated version into a client-side check with a clear `UsageError`, or dropping a short comment saying the server is deliberately left as the enforcement point.

**6. `mssql-tds/tests/test_bulk_copy_vector.rs` — connection reuse after a failed bulk copy.**

```rust
for (id, values) in [(1, vec![1.0, 2.0]), (2, vec![1.0, 2.0, 3.0, 4.0])] {
    ...
    let bulk_copy = BulkCopy::new(&mut client, "#BulkCopyVector16MismatchTest");
```

The second iteration reuses `client` after the first bulk copy has already failed mid-stream. The pre-existing float32 equivalent (`test_bulk_copy_vector_dimension_mismatch`) only does a single attempt, so this is new territory. If the writer doesn't fully drain/reset the connection on error this will be flaky. Either take a fresh connection per iteration, or add an assertion that the client is still usable (e.g. a `SELECT 1`) between attempts so a state leak fails loudly rather than intermittently.

## Nits

**7. `mssql-tds/src/message/features/vectorfeature.rs:163` — `test_deserialize_v2_acknowledged` is a duplicate.**

```rust
let mut feature = VectorFeature::default();
feature.set_acknowledged(true);
feature.deserialize(&[2u8]).unwrap();
assert!(feature.is_acknowledged());
assert_eq!(feature.negotiated_version(), 2);
```

`VectorFeature::default()` is `new(Self::VERSION)` = `new(2)`, so this is the same scenario as the existing `test_negotiate_v2_client_v2_server` a few lines down. Also, `deserialize` never touches `acknowledged`, so `assert!(feature.is_acknowledged())` only re-asserts the `set_acknowledged(true)` on the line above it. I'd drop this test.

**8. PR description / checklist.**

`cargo btest` is checked, but the ADO test jobs fail at this SHA — please re-check that box only once validation is green. Conversely "New/changed functionality has tests" is unchecked even though the PR adds 8 unit tests and 14 integration tests; that box looks like it should be ticked. Minor: "PR 7493" in the description doesn't resolve to a `microsoft/mssql-rs` PR number — if that's an internal ADO PR, worth saying so explicitly.

## Verified, no issues found

- Float16 wire encoding: `half::f16::from_f32` is round-to-nearest-even, matching msodbcsql's `Float32ToFloat16Bits`. The unit-test vectors are right: `1.0 → 0x3C00`, `2.0 → 0x4000`, `3.0 → 0x4200`, `0.1 → 0x2E66`, all written little-endian.
- Length prefix and header: `SqlVector::total_size()` already multiplies by `base_type.element_size_bytes()`, so `8 + dims*2` for float16 is correct, and `encode_vector_header` writes the base-type byte from the value.
- Bulk copy base-type plumbing: `VarLenScale(_, scale)` carries the server's base-type byte into `BulkCopyColumnMetadata::scale`, and `with_precision_scale` is only invoked when `scale > 0` — so float32 stays at the default `0` and float16 lands as `1`. `get_sql_type_definition` and `vector_dimensions` both read it consistently.
- No new `.rs` files, so the license-header rule doesn't apply. No `unsafe`, no new public API surface, no FFI signature changes, no `mssql-py-core` changes.

PR #370, opened 2026-08-24
mssql-odbc: Implement explicit descriptor lifecycle and statement association

Vahid-b:
## Description

Implements **[AB#47436](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47436)** (P2 of the `mssql-odbc | Descriptors` user story, [AB#46374](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46374)): explicit descriptor allocation/free and statement association, building on the descriptor data model + field APIs merged in #345 ([AB#47297](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47297)/[AB#47435](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47435)).

Replaces the `SQL_HANDLE_DESC` "not yet implemented" stubs in `SQLAllocHandle`/`SQLFreeHandle`, and the silent no-op in `SQLSetStmtAttrW(SQL_ATTR_APP_ROW_DESC/APP_PARAM_DESC)`, with real behavior:

- **`SQLAllocHandle(SQL_HANDLE_DESC, hdbc, ...)`** allocates an explicit descriptor owned by the connection, tracked in `DbcState.descriptors` (mirrors the existing `statements` list). Works regardless of connection state, matching msodbcsql's own `AllocDesc` (`sqlcdesc.cpp:5841-5908`, no connection-state check) and this crate's own `alloc_stmt` for the structurally identical `SQL_HANDLE_STMT` case — the ODBC state-transition table's `08003` for this row is marked "(DM)"-generated, not the driver's job to duplicate.
- **`SQLSetStmtAttrW(SQL_ATTR_APP_ROW_DESC/APP_PARAM_DESC)`** associates an explicit descriptor as a statement's active ARD/APD, validated against the ODBC reference and msodbcsql's `SQLSetStmtAttr` (`sqlcmisc.cpp:3599-3639`):
  - `SQL_NULL_HDESC` or the statement's own implicit handle → reset to implicit.
  - Any other implicit descriptor (another statement's ARD/APD, or this statement's own IRD/IPD) → **HY017**.
  - An explicit descriptor from a different connection → **HY024**.
- **`SQLGetStmtAttrW`** returns the active association (implicit or explicit) for ARD/APD, and continues returning the statement-owned IRD/IPD unchanged.
- **`SQLFreeHandle(SQL_HANDLE_DESC)`** rejects freeing an implicit descriptor (**HY017**, diagnostic posted on the descriptor itself per the ODBC spec) and, for an explicit one, walks every statement on the owning connection and resets any that reference it back to implicit — mirrors msodbcsql's `FreeDesc(pADesc, NULL, ...)`. One explicit descriptor associated with multiple statements at once is a supported case, not an error. If the descriptor was already freed by a prior `SQLDisconnect`, returns `SQL_SUCCESS` without double-freeing — mirrors `free_stmt`'s identical "already dropped" guard.
- **`SQLDisconnect`** now frees all outstanding explicit descriptors on the connection (after freeing statements), mirroring msodbcsql's `SQLDisconnect` (`sqlcconn.cpp:1313-1448`) and the ODBC reference's "Freeing a Connection Handle" guidance.

### Data model changes

- `DescHandle` gains a plain, lock-free `alloc_type` (mirrored into the existing `DescState::header::alloc_type` for the `SQLGetDescFieldW`-visible copy — the two can never diverge since `SQL_DESC_ALLOC_TYPE` is permanently read-only) and `parent_dbc`, so lifecycle code can classify a descriptor and check its owning connection without taking its mutex.
- A new `DescKind::Ad` represents a generic explicit application descriptor — unlike the four implicit descriptors, it has no fixed row/parameter role, since one explicit descriptor can be a statement's ARD and another's APD simultaneously. This matches msodbcsql's own model, which tags explicit descriptors `SQL_HANDLE_AD` rather than ARD/APD-specific (`sqlcdesc.cpp:5891`).
- `StmtState` gains `active_ard`/`active_apd: Option<*mut c_void>` behind its existing mutex — `None` means "use the implicit descriptor" (`StmtHandle::ard`/`apd`, untouched), `Some` an associated explicit descriptor. This is exactly the design flagged by a doc comment left in #345.

### Testing

- Extends `TestHandles` (`alloc_explicit_desc`, `free_explicit_desc`, `free_extra_stmt`, `alloc_other_connection`) so every new unit test allocates/frees handles only through the shared helper, per this crate's testing conventions.
- Unit tests cover: allocation, association, reassociation, reset (both spellings), statement free, descriptor free (including resetting 2+ associated statements, and the already-freed-by-disconnect no-op case), connection free with an outstanding descriptor, disconnect freeing outstanding descriptors, cross-connection rejection, rejecting another statement's implicit descriptor / this statement's own IRD as an ARD, and diagnostic ownership (the DESC's own diagnostics for a rejected free, the STMT's for a rejected association).
- New e2e test (`tests/e2e/tests/alloc_free_desc_test.cpp`) exercises the same lifecycle through the real exported entry points. Two of its assertions are intentionally narrowed to what's portably observable through any Driver Manager: unixODBC intercepts the ARD/APD "reset to implicit via `SQLSetStmtAttrW(NULL-or-own-implicit-handle)`" transition at its own DM layer (confirmed via a unixODBC changelog entry), so a subsequent `SQLGetStmtAttrW` through it isn't a reliable way to observe the driver's internal state for that specific transition. The driver's own round-trip correctness for it remains exhaustively covered by the DM-independent Rust unit tests (`reset_to_implicit_via_null`/`reset_to_implicit_via_own_handle` in `set_stmt_attr.rs`).

### CI infrastructure fix (rides along with this PR)

`.pipeline/scripts/ensure-rust-toolchain.sh` is a new shared script extracted from `containerized-python-test.sh`'s existing toolchain-repair logic (a fresh build container can register `rust-toolchain.toml`'s pinned channel without actually installing `rustc`, which `maturin` needs). It's unrelated to descriptors, but rides along here because this PR's CI runs hit the exact failure it fixes (`Build mssql-python (cross-repo)` failing with `'rustc' is not installed for the toolchain`) — `test-mssql-python-template.yml`'s docker invocation was the one caller missing this hardening. Kept in this PR rather than split out since it was discovered and fixed as part of getting this PR's own CI green.

### Out of scope (tracked separately, [AB#47437](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47437))

- Binding/fetch execution through descriptor records.
- Populating IRD/IPD metadata.
- Parameter and result conversion.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47436

Parent: https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46374 (`mssql-odbc | Descriptors`)
Next: [AB#47437](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47437) (binding/metadata integration through descriptor records)

## Validation

- `cargo nextest run -p mssql-odbc --lib` — 894 passed, 1 skipped (pre-existing, requires a live server).
- `cargo llvm-cov nextest -p mssql-odbc --frozen --no-report --all-targets --no-fail-fast --profile ci --success-output immediate` — 894 passed, 1 skipped.
- `cargo bclippy` (workspace, `-D warnings`) — clean.
- `cargo bfmt` (workspace) — clean.
- e2e (`alloc_free_desc_test`) — passing in CI (Linux, parity-checked against msodbcsql), and compiled/run locally against a real MSVC + CMake + GoogleTest toolchain.
- Rebased onto latest `main` (picked up #361's data-at-execution parameter support with no conflicts beyond one adjacent SQLSTATE constant addition).
- Addressed a full human review round: a blocking double-free in `free_desc` (missing the "already dropped by SQLDisconnect" guard `free_stmt` has), a lock-order inversion risk between `free_desc` (DBC→STMT) and `SQLParamData`'s DAE-advance path (STMT→DBC), a poisoned-mutex leak window in `SQLDisconnect`'s descriptor-free loop, an incorrect "requires a connected DBC" gate on `alloc_desc` (verified against msodbcsql source — neither `AllocDesc` nor `SQLAllocStmt` check connection state), and a test that silently leaked/panicked instead of exercising the real free path.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (validated via `cargo nextest run -p mssql-odbc --lib` and `cargo llvm-cov nextest -p mssql-odbc ...`, both scoped to the changed crate: 894 passed, 1 skipped)


mssql-odbc/src/api/free_handle.rs
```diff
@@ -107,6 +110,13 @@ unsafe fn free_dbc(handle: SqlHandle) -> SqlReturn {
             .unwrap_or(true),
         "SQLFreeHandle(DBC): DM should have freed all STMTs before calling SQLFreeConnect"
     );
+    debug_assert!(
```
Vahid-b, line 121: **Blocking:** This `debug_assert!` (and its statement-list twin two lines above) mirrors the STMT case, but the underlying DM guarantee does not actually extend to explicit descriptors, and I have verified this from three independent angles.

**1. Authoritative ODBC spec citation.** Per Microsoft's own ODBC reference ("Freeing Descriptors"): *"If you have not freed an explicitly allocated descriptor, freeing the connection handle will still automatically free all descriptors associated with it."* This is the **driver's** job, not a DM-enforced precondition -- unlike statements, where the DM's own connection state machine (C4+ vs C2) genuinely does block `SQLFreeHandle(DBC)` until every STMT is gone. There is no equivalent state-machine gate for descriptor handles.

**2. Empirically reproduced.** I cloned this branch and ran `free_dbc_with_outstanding_desc_fails_in_debug` with `--nocapture`. It "passes" but only because it doesn't check the return values of the second `sql_free_handle(SQL_HANDLE_DBC, dbc)` or the final `sql_free_handle(SQL_HANDLE_ENV, env)` calls -- both of which panic too:
```
thread '...' panicked at free_handle.rs:113:5: SQLFreeHandle(DBC): DM should have freed all explicit DESCs before calling SQLFreeConnect
thread '...' panicked at free_handle.rs:113:5: SQLFreeHandle(DBC): DM should have freed all explicit DESCs before calling SQLFreeConnect   (same assert, again, on retry)
thread '...' panicked at free_handle.rs:74:5:  SQLFreeHandle(ENV): DM should have freed all DBCs before calling SQLFreeEnv
```
The retry panics identically because the test's manual `free_handle::<DescHandle>(desc)` cleanup frees the `Box` directly without removing the pointer from `dbc_state.descriptors`, leaving a dangling entry -- and the ENV panics too because `free_dbc` never reaches its "unregister from parent ENV" step on either attempt. In this run, **both the `DbcHandle` and the `EnvHandle` are never actually freed** -- this test title (`..._fails_in_debug`) documents the debug-mode symptom as if it's an accepted, permanent characteristic rather than a defect to fix, while leaving the release-mode path (see below) completely uncovered.

**3. Release-mode consequence is provable from the type declaration alone, no test needed.** `descriptors: Vec<*mut c_void>` (in `dbc.rs`) holds raw pointers. Rust never runs drop glue on `*mut c_void` -- dropping the `Vec` when `free_handle::<DbcHandle>` deallocates the `DbcHandle` can never free the `DescHandle` `Box`es the pointers refer to. So in a release build (`debug_assert!` compiled out), this path **unconditionally leaks every outstanding `DescHandle`**, and leaves its `parent_dbc` field pointing at memory that was just deallocated. If the application later calls any descriptor API (`SQLGetDescFieldW`, `SQLSetDescFieldW`, or even a defensive `SQLFreeHandle(SQL_HANDLE_DESC, ...)` the application doesn't strictly need per spec but many well-behaved apps issue anyway) on that handle, the driver dereferences that dangling `parent_dbc` -- a use-after-free. This directly contradicts the SAFETY justification in `handles/desc.rs` for `unsafe impl Send/Sync for DescHandle`, which explicitly claims *"the parent DBC is guaranteed alive because the DM ensures every descriptor ... is freed before its connection"* -- a premise this same PR's own test disproves.

This also means `disconnect.rs`'s existing `// TODO: free user-allocated descriptors` comment is pointed at the wrong function: per the spec citation above, `SQLDisconnect` isn't responsible for freeing explicit descriptors at all (they legitimately survive a disconnect/reconnect cycle) -- it's specifically `SQLFreeHandle(SQL_HANDLE_DBC)` (this function) that must auto-free any still-outstanding ones.

**Suggested fix:** replace this `debug_assert!` with an actual free loop -- iterate `dbc_state.descriptors`, call `free_handle::<DescHandle>` on each (mirroring `sql_disconnect_safe`'s existing pattern for `state.statements` a few lines away in a sibling file), then proceed to unregister from the parent ENV as today. That also resolves the disconnect.rs TODO by making it moot (nothing to do there once the DBC-free path is correct), and would let the existing test assert a real `SQL_SUCCESS` outcome with the descriptor actually freed, instead of demonstrating three cascading panics.

Vahid-b (commented):
## Review Summary

Reviewed this against the ODBC reference, the linked parent work item (AB#46374 `mssql-odbc | Descriptors`) and its prior phase (#345), and this crate's own `.github/instructions/mssql-odbc.instructions.md`. The association logic (`SQLSetStmtAttrW`/`SQLGetStmtAttrW` for ARD/APD, HY017/HY024 validation, sharing one explicit descriptor across statements, survival across statement free) is well-designed and thoroughly tested -- I traced `validate_descriptor_association`, the SQLSTATE mappings (`ERR_INVALID_USE_OF_AUTO_DESC` -> HY017, `ERR_INVALID_ATTRIBUTE_VALUE` -> HY024), and `StmtHandle`'s `Drop` impl (which correctly frees the four implicit descriptors) and found no issues in any of that.

One finding below is Blocking: the new explicit-descriptor-vs-connection-free interaction rests on an incorrect premise about what the ODBC Driver Manager guarantees, and I verified this three independent ways (spec citation, reproducing the exact panic sequence locally, and a language-level leak argument that needs no test at all). A second is a concretely-reproduced `cargo fmt` failure that likely explains the currently-red CI build legs.

Both were validated by cloning this branch and running the actual code (`cargo fmt`, `cargo clippy`, `cargo nextest run -p mssql-odbc --lib` = 850/850 passing, and the specific failing test with `--nocapture`) rather than from static reading alone.

mssql-odbc/src/api/set_stmt_attr.rs
```diff
@@ -776,4 +853,236 @@ mod tests {
...
+        let other_implicit_ard = read_desc(other_stmt, SQL_ATTR_APP_ROW_DESC).1;
+
+        assert_eq!(set_desc(h.stmt, SQL_ATTR_APP_ROW_DESC, desc), SQL_SUCCESS);
+        assert_eq!(
+            set_desc(other_stmt, SQL_ATTR_APP_ROW_DESC, desc),
+            SQL_SUCCESS
+        );
+
+        assert_eq!(
+            h.free_explicit_desc(desc),
```
Vahid-b: **Nit (but concretely reproduced, and likely the direct cause of the currently-failing CI build legs across every platform):** `cargo fmt -p mssql-odbc -- --check` fails on this exact block when run against this PR's head commit (`590dbe6a`):
```
Diff in .../mssql-odbc/src/api/set_stmt_attr.rs:1049:
-        assert_eq!(
-            h.free_explicit_desc(desc),
-            SQL_SUCCESS
-        );
+        assert_eq!(h.free_explicit_desc(desc), SQL_SUCCESS);
```
The PR checklist marks `cargo bfmt` as passing, but this diverges from it. Given every "Build Stage Build *" CI leg (Linux, Linux ARM, Windows, Windows ARM, MacOS) is currently red, and `cargo bfmt` compiled cleanly and `cargo clippy`/`cargo nextest` (850/850) both passed for me locally, a fmt-check gate failing identically and early on every platform's Build stage is a very plausible single root cause worth ruling out first before digging into anything more exotic.

Vahid-b: Confirmed — this was the exact and sole root cause of every red CI leg. Fixed by re-running `cargo fmt` (I'd made a small edit to this test after my last format pass and didn't re-run it). Verified `cargo bfmt`/`.\scripts\bfmt.ps1` are clean and pushed in e7e0119b.

> [!NOTE]

mssql-odbc/src/api/free_handle.rs
Vahid-b: You're right, and this was the real bug — thanks for the rigorous repro. Root-caused it fully rather than patching the assert:

1. `alloc_desc` now requires a connected DBC (`08003` otherwise), per the ODBC connection state-transition table — `SQLAllocHandle(SQL_HANDLE_DESC)` is gated on C4/C5/C6 exactly like `SQL_HANDLE_STMT`, and C2 (which a connection also returns to after `SQLDisconnect`) is rejected. Unlike STMT allocation, which is protected structurally (the DM never dispatches to this driver before the first connect), a disconnected-but-previously-connected DBC keeps dispatching straight to this function, so the check has to be explicit here.
2. `sql_disconnect_safe` now frees every outstanding explicit descriptor right after freeing statements — closing the `disconnect.rs` TODO you correctly pointed at, and matching the ODBC reference text you quoted plus msodbcsql's actual `SQLDisconnect` (`sqlcconn.cpp:1427-1448`), which frees its descriptor-allocation-node list right after its own statement loop.

With both of those, `free_dbc`'s `debug_assert!` is a real invariant, not a false one — so I left it as an assert rather than adding a defensive free loop there. I checked msodbcsql's `SQLFreeConnect` directly (`sqlcconn.cpp:727`): it also only does `assert(CItemsPl(lpdbc->lpplDAN) == 0)` at that exact point, trusting its own `SQLDisconnect` to have already cleared the list — a free loop in `free_dbc` would have been a parity divergence, not a fix.

Also fixed the misleading test: `free_dbc_with_outstanding_desc_fails_in_debug` now marks its DBC connected (so the alloc succeeds) and its comment explains why the scenario is a genuine DM-contract violation (skipped `SQLDisconnect`) rather than a hypothetical one — matching its STMT twin. Added `alloc_desc_on_disconnected_dbc_returns_error` and `disconnect_frees_outstanding_explicit_descriptors` as targeted regression coverage for the two fixes. All pushed in 1f9fd905.

> [!NOTE]

mssql-odbc/src/api/free_handle.rs
```diff
@@ -107,6 +110,13 @@ unsafe fn free_dbc(handle: SqlHandle) -> SqlReturn {
...
+        dbc.inner
+            .lock()
+            .map(|s| s.descriptors.is_empty())
+            .unwrap_or(true),
+        "SQLFreeHandle(DBC): DM should have freed all explicit DESCs before calling SQLFreeConnect"
```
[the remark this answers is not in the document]
Vahid-b: Fixed — `sql_disconnect_safe` now frees every outstanding explicit descriptor (mirroring the existing statement-free loop, and msodbcsql's own `SQLDisconnect`, which does the same right after its statement loop: `sqlcconn.cpp:1427-1448`). The stale `disconnect.rs` TODO is removed. This also closes the related thread on `free_handle.rs:113` about the `debug_assert!`'s invariant not actually holding — see that reply for the full fix (which additionally gates `SQLAllocHandle(SQL_HANDLE_DESC)` on a connected DBC, per the ODBC state-transition table). Added a regression test, `disconnect_frees_outstanding_explicit_descriptors`, covering exactly this. Pushed in 1f9fd905.

> [!NOTE]

David-Engel (commented):
## Review Summary

Solid, well-documented implementation of explicit descriptor alloc/free/association. I traced `validate_descriptor_association`, `free_desc`, `sql_disconnect_safe`, the `DescKind::Ad` classification path, `free_handle`, and `StmtHandle::Drop`, and ran the new tests locally against this branch. The association logic (HY017/HY024, sharing one explicit descriptor across statements, survival across `SQLFreeHandle(STMT)`) is correct and thoroughly tested — no issues found there.

Findings are concentrated on the free/disconnect lifecycle. One is Blocking: `free_desc` is missing the "already dropped by `SQLDisconnect`" guard that `free_stmt` deliberately has, and this PR is what makes that sequence reachable, so `SQLAllocHandle(DESC)` → `SQLDisconnect` → `SQLFreeHandle(DESC)` double-frees. The rest are suggestions, plus one test that I verified does not actually clean up (it emits three panics, not one, and leaks both handles).

All checks are green on `b6d591a6`, the PR is linked to AB#47436, and the checklist claims match CI.

mssql-odbc/src/api/free_handle.rs
```diff
@@ -185,6 +201,74 @@ unsafe fn free_stmt(handle: SqlHandle) -> SqlReturn {
...
+            stmt_state.active_apd = None;
+        }
+    }
+
+    if let Some(i) = dbc_state.descriptors.iter().position(|&p| p == handle) {
+        dbc_state.descriptors.swap_remove(i);
+    }
+    drop(dbc_state);
+
+    unsafe { free_handle::<DescHandle>(handle) };
```
David-Engel: **Blocking** — missing the "already dropped by `SQLDisconnect`" guard, so this double-frees.

`free_stmt` deliberately bails out in exactly this case (line 193-196 in this file: *"Already dropped by SQLDisconnect - early return"*). `free_desc` doesn't: when the descriptor isn't found in `dbc_state.descriptors`, it falls through and calls `free_handle::<DescHandle>` anyway.

This PR is what makes the sequence reachable — `SQLAllocHandle(SQL_HANDLE_DESC, hdbc, &hdesc)` → `SQLDisconnect(hdbc)` (now frees every outstanding explicit descriptor) → `SQLFreeHandle(SQL_HANDLE_DESC, hdesc)` frees the same `Box<DescHandle>` twice. An application holding a descriptor handle across a disconnect and then tidying up is ordinary usage, not a DM contract violation.

Suggested fix, mirroring `free_stmt`:

```rust
let Some(i) = dbc_state.descriptors.iter().position(|&p| p == handle) else {
    // Already dropped by SQLDisconnect - early return.
    return SQL_SUCCESS;
};
dbc_state.descriptors.swap_remove(i);
```

Caveat worth noting: in that same sequence the earlier `desc.is_explicit()` / `desc.parent_dbc` reads at lines 228 and 236 are already reading freed memory before you get here. `free_stmt` has the same partial-mitigation shape, so matching it is the right scope for this PR — but the general stale-handle problem might deserve its own work item alongside the existing refcounted-handle-lifetimes TODO in `disconnect.rs`.

mssql-odbc/src/api/disconnect.rs
```diff
@@ -98,6 +97,27 @@ fn sql_disconnect_safe(dbc: &DbcHandle) -> SqlReturn {
...
+        // with any op still holding it, then drop the box.
+        let desc = unsafe { handle_from_raw::<DescHandle>(desc_ptr) };
+        let Ok(guard) = desc.inner.lock() else {
+            error!(?desc_ptr, "SQLDisconnect: desc mutex poisoned");
+            return SQL_ERROR;
+        };
+        drop(guard);
+        unsafe { free_handle::<DescHandle>(desc_ptr) };
+    }
+    state.descriptors.clear();
```
David-Engel: **Suggestion** — the poisoned-mutex `return SQL_ERROR` at line 114 leaves already-freed pointers in `state.descriptors`, and `connection_state` stays `Connected`. A retried `SQLDisconnect` re-walks the same list and double-frees everything before the poisoned entry.

Pop-then-free closes the window:

```rust
while let Some(desc_ptr) = state.descriptors.pop() {
    let desc = unsafe { handle_from_raw::<DescHandle>(desc_ptr) };
    let Ok(guard) = desc.inner.lock() else {
        error!(?desc_ptr, "SQLDisconnect: desc mutex poisoned");
        return SQL_ERROR;
    };
    drop(guard);
    unsafe { free_handle::<DescHandle>(desc_ptr) };
}
```

The statement loop directly above has the identical shape today, so this is a pre-existing pattern rather than something you introduced — but the new loop replicates it instead of fixing it. Happy for this to be a follow-up if you'd rather fix both together.

mssql-odbc/src/api/free_handle.rs
```diff
@@ -185,6 +201,74 @@ unsafe fn free_stmt(handle: SqlHandle) -> SqlReturn {
...
+                0,
+                "Internal error while freeing descriptor",
+            );
+        }
+        return SQL_ERROR;
+    };
+
+    for &stmt_raw in &dbc_state.statements {
+        let stmt = unsafe { handle_from_raw::<StmtHandle>(stmt_raw) };
+        let Ok(mut stmt_state) = stmt.inner.lock() else {
```
David-Engel, line 261: **Suggestion** — lock-order inversion (ABBA deadlock risk) against `sql_param_data`.

This loop holds the DBC guard (`dbc_state`, acquired at line 237) across `stmt.inner.lock()` for every statement on the connection, i.e. DBC → STMT.

`sql_param_data` goes the other way. In `mssql-odbc/src/api/param_data.rs`, `stmt_state` is bound at line 184 and is still alive at line 198 where `return_client_idle(dbc, statement_handle, client)` takes the DBC lock (`exec_common.rs:168`). That's STMT → DBC while the stmt guard is held — it's the subtle case, since the guard is a named binding rather than a temporary, so it isn't obvious from the call site.

So: thread A in `SQLFreeHandle(SQL_HANDLE_DESC)` holds DBC and waits on STMT, thread B in `SQLParamData` holds STMT and waits on DBC.

`sql_disconnect_safe` already nests DBC → STMT, so the inversion technically pre-dates this PR — but disconnect is teardown, whereas `SQLFreeHandle(SQL_HANDLE_DESC)` is a routine mid-session call, which widens the window a lot. The cheapest fix is on the `param_data` side: end the `stmt_state` borrow before calling `return_client_idle` on the two error paths. Worth at least a tracked work item if you don't want to touch `param_data.rs` in this PR.

mssql-odbc/src/api/free_handle.rs
```diff
@@ -463,9 +558,105 @@ mod tests {
...
+
+        let ret = unsafe { sql_free_handle(SQL_HANDLE_DBC, dbc) };
+        if cfg!(debug_assertions) {
+            assert_eq!(ret, SQL_ERROR);
+            unsafe { free_handle::<DescHandle>(desc) };
+            unsafe { sql_free_handle(SQL_HANDLE_DBC, dbc) };
+        } else {
+            assert_eq!(ret, SQL_SUCCESS);
+            unsafe { free_handle::<DescHandle>(desc) };
+        }
```
David-Engel, line 794: **Suggestion** — this test doesn't actually clean up; I ran it on this branch and it emits three panics, not one:

```
panicked at free_handle.rs:119: SQLFreeHandle(DBC): DM should have freed all explicit DESCs ...
panicked at free_handle.rs:119: SQLFreeHandle(DBC): DM should have freed all explicit DESCs ...
panicked at free_handle.rs:74:  SQLFreeHandle(ENV): DM should have freed all DBCs before calling SQLFreeEnv
test ... ok
```

`free_handle::<DescHandle>(desc)` drops the box without unregistering it from `dbc_state.descriptors`, so the retried `sql_free_handle(SQL_HANDLE_DBC, dbc)` on line 653 trips the same `debug_assert!` again and returns `SQL_ERROR`. The DBC is therefore never freed and never unregistered from the ENV, so `SQLFreeHandle(SQL_HANDLE_ENV)` asserts too. Both handles leak and a dangling pointer is left in `descriptors`.

Using the real entry point cleans up properly in both branches:

```rust
unsafe { sql_free_handle(SQL_HANDLE_DESC, desc) };
```

The test still passes today because `ffi_entry!` swallows the panics, which is exactly what makes this easy to miss.

mssql-odbc/src/api/alloc_handle.rs
```diff
@@ -174,6 +173,72 @@ unsafe fn alloc_stmt(input_handle: SqlHandle, output_handle: *mut SqlHandle) ->
...
+        error!("SQLAllocHandle(DESC): dbc mutex poisoned");
+        return SQL_ERROR;
+    };
+    free_errors(&mut dbc_state);
+
+    if dbc_state.connection_state != ConnectionState::Connected {
+        error!("SQLAllocHandle(DESC): DBC is not connected");
+        post_diag(&mut dbc_state, ERR_CONNECTION_DOES_NOT_EXIST);
+        return SQL_ERROR;
+    }
```
David-Engel: **Suggestion / question** — worth confirming this gate against msodbcsql rather than the state table.

MS Learn lists `08003` for `SQL_HANDLE_DESC` under `SQLAllocHandle` as **(DM)**-generated, and `.github/instructions/mssql-odbc.instructions.md` says preconditions the DM is contractually required to enforce get a `debug_assert!`, not a release-build `if`/error-return.

Your doc comment's argument — that a DBC which connected once and has since disconnected keeps dispatching straight to the driver, so the DM's structural gate no longer covers it — is plausible and I'm not asking you to drop the check on that basis. But the comment cites the ODBC state-transition table, and the crate's parity rule is explicit that a decision to *add* or *reject* behavior gets verified against msodbcsql source, not MS Learn.

Does msodbcsql's `AllocDesc` (`sqlcdesc.cpp:5841-5908`) actually check connection state in the driver? If it does, cite that instead and this is settled. If it doesn't, this is an exceed-parity deviation and should say so explicitly, per the same instructions file.

.pipeline/scripts/ensure-rust-toolchain.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
David-Engel, line 1: **Suggestion** — this rust-toolchain hardening is a good change on its own (extracting the retry logic so the macOS/python legs get it too is clearly right), but it has nothing to do with descriptors and isn't mentioned anywhere in the PR description.

Either split it into its own PR, or add a line to the description explaining why it rides along — otherwise anyone bisecting a future CI change or reading this PR's history has to reverse-engineer the connection.

mssql-odbc/src/api/free_handle.rs
```diff
@@ -185,6 +201,74 @@ unsafe fn free_stmt(handle: SqlHandle) -> SqlReturn {
...
+
+    if let Ok(mut state) = desc.inner.lock() {
+        free_errors(&mut state);
+    }
+
+    if !desc.is_explicit() {
+        error!("SQLFreeHandle(DESC): cannot free an implicitly allocated descriptor");
+        if let Ok(mut state) = desc.inner.lock() {
+            post_diag(&mut state, ERR_INVALID_USE_OF_AUTO_DESC);
+        }
```
David-Engel, line 237: **Nit** — `desc.inner` is locked twice on this path: once for `free_errors`, then released, then again for `post_diag`. One guard covers both:

```rust
if let Ok(mut state) = desc.inner.lock() {
    free_errors(&mut state);
    if !desc.is_explicit() {
        error!("SQLFreeHandle(DESC): cannot free an implicitly allocated descriptor");
        post_diag(&mut state, ERR_INVALID_USE_OF_AUTO_DESC);
        return SQL_ERROR;
    }
}
```

Functionally equivalent (nothing else can meaningfully interleave), just tidier.

mssql-odbc/src/api/free_handle.rs
```diff
@@ -185,6 +201,74 @@ unsafe fn free_stmt(handle: SqlHandle) -> SqlReturn {
...
+        return SQL_ERROR;
+    };
+
+    for &stmt_raw in &dbc_state.statements {
+        let stmt = unsafe { handle_from_raw::<StmtHandle>(stmt_raw) };
+        let Ok(mut stmt_state) = stmt.inner.lock() else {
+            continue;
+        };
+        if stmt_state.active_ard == Some(handle) {
+            stmt_state.active_ard = None;
```
David-Engel, line 286: **Suggestion** — a poisoned STMT mutex here leaves a dangling descriptor pointer on that statement.

```rust
let Ok(mut stmt_state) = stmt.inner.lock() else {
    continue;
};
```

On `continue` the loop skips that statement but the descriptor is still freed unconditionally at the end of the function, so the statement keeps `active_ard == Some(<freed handle>)`. The next `SQLGetStmtAttrW(SQL_ATTR_APP_ROW_DESC)` hands the application a dangling handle, and the binding work under AB#47437 will dereference it.

Poisoning is reachable in this driver rather than theoretical: `ffi_entry!` wraps every entry point in `catch_unwind` (`lib.rs:55`), so a panic raised while a STMT lock is held is swallowed, the process continues, and the mutex stays poisoned for every later call. Every other lock failure in this same function returns `SQL_ERROR` — including the DBC lock twenty lines above — so `continue` is the one spot that converts a recoverable error into a memory-safety hazard.

```rust
let Ok(mut stmt_state) = stmt.inner.lock() else {
    error!(?stmt_raw, "SQLFreeHandle(DESC): stmt mutex poisoned; refusing to free");
    drop(dbc_state);
    if let Ok(mut state) = desc.inner.lock() {
        post_sql_error(&mut state, SQLSTATE_HY000, 0, "Internal error while freeing descriptor");
    }
    return SQL_ERROR;
};
```

Leaking the descriptor is strictly better than dangling it, and `SQLFreeHandle` returning `SQL_ERROR` already means "the handle is still valid" per its own spec, so the application is left in a consistent state either way.

Note this interacts with the double-free finding on line 263: if you take the `return SQL_SUCCESS` early-return there, this path is the other way the descriptor can end up freed while something still points at it.

David-Engel (commented):
One follow-up finding on top of the review already posted, on the same `free_desc` loop but a different line. Everything else I checked (HY017/HY024 against the SQLSetStmtAttr diagnostics table, the `SQLAllocHandle(SQL_HANDLE_DESC)` row in Appendix B's connection transitions, and "frees those statements and all descriptors that have been explicitly allocated on the connection" in SQLDisconnect's Comments) matched the implementation, and the earlier findings stand as written.

mssql-odbc/src/api/free_handle.rs
Vahid-b: Confirmed and fixed in 3f6da5a9 — added the identical "already dropped by SQLDisconnect - early return" guard `free_stmt` has, so `free_desc` now checks `dbc_state.descriptors.iter().position(...)` before touching `free_handle`, returning `SQL_SUCCESS` early if the descriptor isn't found instead of falling through to a second `free_handle::<DescHandle>` call.

Added a regression test, `free_desc_already_removed_from_parent_is_a_no_op`, that reproduces this via a synthetic "already removed" state (clearing `dbc_state.descriptors` without freeing the box) so the test itself doesn't need to dereference already-freed memory to exercise it.

On the caveat: agreed this is out of scope here — left the top-of-function `desc.is_explicit()`/`desc.parent_dbc` reads as-is, matching `free_stmt`'s identical partial-mitigation shape. Filed no separate work item yet, but happy to if you'd like one tracked alongside the existing refcounted-handle-lifetimes TODO in `disconnect.rs`.

> [!NOTE]

mssql-odbc/src/api/disconnect.rs
Vahid-b: Fixed in 3f6da5a9 — switched to pop-then-free (`while let Some(desc_ptr) = state.descriptors.pop()`) instead of iterate-then-clear, so a poisoned mutex mid-loop leaves only the not-yet-freed remainder in `state.descriptors` and a retry picks up where this pass stopped.

Left the identical pre-existing pattern in the statement loop above unchanged for this PR, per your note — happy to file a follow-up to unify both if you'd like one tracked, but didn't want to touch it silently without confirming that's the disposition you want.

> [!NOTE]

mssql-odbc/src/api/free_handle.rs
Vahid-b: Confirmed and fixed in 3f6da5a9, on the `param_data.rs` side as you suggested. Traced it exactly as you described: the `NeedData` arm's `let Some(dae) = stmt_state.dae.as_mut() else { ...; return_client_idle(...); ... }` held `stmt_state` (bound a few lines up) across the `return_client_idle` call in its `else` branch — STMT held while taking DBC, the reverse of `free_desc`'s DBC→STMT order.

Restructured that one call site to a `match` so the STMT guard is explicitly `drop`ped before `return_client_idle` runs in the failure arm. Confirmed via re-reading the file that this is the only genuine occurrence — the other two `return_client_idle` call sites in this function are in `else` branches of a failed `stmt.inner.lock()` itself, so no guard is ever bound there to hold across the call.

Agreed the pre-existing `sql_disconnect_safe` DBC→STMT nesting is a separate, lower-risk case (teardown vs. routine mid-session call) — left that as-is per your framing.

> [!NOTE]

mssql-odbc/src/api/free_handle.rs
Vahid-b: Confirmed by tracing `ffi_entry!`'s `catch_unwind` — you're right that the panics were being silently swallowed and converted to `SQL_ERROR`/leaked handles rather than actually exercising cleanup.

Fixed in 3f6da5a9: the debug branch now calls `sql_free_handle(SQL_HANDLE_DESC, desc)` (the real entry point), which — now that `free_desc` properly unregisters from `dbc_state.descriptors` (see the sibling thread on the double-free fix) — lets the retried `SQLFreeHandle(DBC)` and the trailing `SQLFreeHandle(ENV)` both succeed cleanly with no panics.

Left the release-mode branch as the direct `free_handle::<DescHandle>(desc)` call, though: in release, `free_dbc`'s `debug_assert!` is a no-op, so the first `SQLFreeHandle(DBC)` call in that branch already actually dropped `dbc` (and its ENV registration) despite the outstanding descriptor — `desc.parent_dbc` is dangling by the time we'd reach `sql_free_handle(SQL_HANDLE_DESC, desc)`, so routing through it there would be a real use-after-free on the now-freed `dbc`. Documented this asymmetry in both the test and its doc comment.

> [!NOTE]

mssql-odbc/src/api/alloc_handle.rs
Vahid-b: Checked against the local msodbcsql checkout as asked, and you're right on both counts:

- `AllocDesc` (`sqlcdesc.cpp:5841-5908`) has no connection-state check at all — it just searches/grows the descriptor-allocation list under the DBC's critical section and returns.
- `SQLAllocStmt` (`sqlccmd.cpp:252-277`) doesn't check connection state either — only null-checks `lpdbc`/`plpstmt`.

So my "unlike STMT allocation, whose equivalent gate is enforced structurally" framing was simply wrong — nothing distinguishes the two cases in msodbcsql, and this crate's own `alloc_stmt` already reflects that (no connection-state check). The `08003` in the ODBC reference is marked "(DM)", meaning the Driver Manager generates it for both handle types identically; duplicating it in the driver was an unwarranted exceed-parity addition on my part, not something either msodbcsql or this crate's own STMT path needed.

Fixed in 3f6da5a9: removed the check, corrected the doc comment to explain the actual reasoning (including why `free_dbc`'s `descriptors.is_empty()` invariant doesn't need it either — it already holds for STMT with no gate, since it's just ordinary "free children before the parent," independent of connection state), and replaced `alloc_desc_on_disconnected_dbc_returns_error` with `alloc_desc_on_disconnected_dbc_succeeds` to lock in the corrected behavior.

> [!NOTE]

.pipeline/scripts/ensure-rust-toolchain.sh
Vahid-b: Fair point — added a "CI infrastructure fix (rides along with this PR)" section to the PR description explaining why it's here: this PR's own CI runs hit the exact failure this fixes (`Build mssql-python (cross-repo)` failing with `'rustc' is not installed for the toolchain`), and it was discovered and fixed in the course of getting this PR green rather than being an unrelated drive-by change. Keeping it in this PR rather than splitting it out, since it's already been through several CI cycles bundled with these commits and splitting now would mean re-validating both halves separately for no real benefit at this point — but the description now makes the connection explicit for anyone bisecting later.

> [!NOTE]

mssql-odbc/src/api/free_handle.rs
Vahid-b: Done in 3f6da5a9 — consolidated into a single `desc.inner.lock()` guard covering both `free_errors` and the implicit-descriptor rejection.

> [!NOTE]

mssql-odbc/src/api/free_handle.rs
Vahid-b: Confirmed — good catch, and correctly identified as compounding with the double-free fix (I'd moved the `swap_remove` to happen *before* this loop for that fix, which is exactly what made "refuse instead of continue" require also deferring the removal). Fixed in 0568ecb2 largely as you suggested: a poisoned STMT mutex now posts the internal-error diagnostic and returns `SQL_ERROR` without freeing, and I moved `dbc_state.descriptors.swap_remove(i)` to after the loop succeeds (instead of before it, per the double-free fix) so refusing here leaves the descriptor both intact *and* still correctly tracked in `dbc_state.descriptors` — not just failing safely, but actually recoverable: freeing the poisoned statement itself (which tolerates its own poisoning, matching `free_stmt`'s existing behavior) lets a retry succeed cleanly.

Added `free_desc_with_poisoned_stmt_mutex_refuses_to_free`, poisoning the mutex the same way `apply_catalog_metadata_poisoned_mutex_returns_error` does, and exercising that full recovery path (refuse → free the poisoned statement → retry succeeds).

> [!NOTE]

David-Engel (commented):
## Re-review — `3f6da5a9`, `0568ecb2`

Verified locally on `0568ecb2`: `cargo bfmt` clean, `cargo bclippy` clean, `cargo nextest run -p mssql-odbc --lib` → 895 passed, 1 skipped.

**All seven findings from my previous review are resolved.** Highlights:

- **The blocking double-free** is fixed with the same `else { return SQL_SUCCESS }` guard `free_stmt` uses, and moving `swap_remove` to *after* the reset loop is the right call — `i` stays valid because `dbc_state` is held continuously across the loop.
- **The lock inversion** is fixed on the `param_data.rs` side as suggested, and you correctly identified that the other two `return_client_idle` call sites hold no guard.
- **The leaking test**: I re-ran it on this commit. It's down from three panics to one (the intentional assert under test), and both the retried `SQLFreeHandle(DBC)` and the trailing `SQLFreeHandle(ENV)` now succeed cleanly. The release-branch asymmetry — `desc.parent_dbc` is dangling once `free_dbc` has actually run — is a correct call and well documented.
- **The `08003` parity check** came back the opposite of the original assumption (neither `AllocDesc` nor `SQLAllocStmt` gates on connection state), and you removed the gate rather than re-justifying it. That's the right outcome, and updating `free_dbc`'s invariant rationale to the plain "free children before the parent" contract keeps the assert honest.

**On `0568ecb2` (unsolicited):** the poisoned-STMT `continue` → refuse-and-return-`SQL_ERROR` change is a genuine improvement I'd missed — silently skipping a poisoned statement would have freed the descriptor out from under a live `active_ard`. Deferring `swap_remove` until after the loop makes the refusal *recoverable* rather than merely safe, and `free_desc_with_poisoned_stmt_mutex_refuses_to_free` exercises the full refuse → free-stmt → retry path. Nice catch on your own code.

One leftover doc nit inline. Nothing blocking remains.

CI on `0568ecb2` was still in flight when I looked (12 green, 6 in-progress/queued) — worth confirming it lands green before merging. The merge is yours.

mssql-odbc/src/test_support.rs
```diff
@@ -112,6 +114,65 @@ impl TestHandles {
         stmt
     }
 
+    /// Allocate an explicit descriptor (`SQLAllocHandle(SQL_HANDLE_DESC, ...)`)
+    /// under the same DBC. The returned handle is tracked and freed on drop,
+    /// before the DBC. Marks the DBC connected first (idempotent) since
+    /// descriptor allocation requires it (`alloc_desc`'s `08003` gate).
```
David-Engel: **Nit** — leftover from removing the `08003` gate: this comment still says descriptor allocation *requires* a connected DBC. You updated the sibling comment in `alloc_other_connection` (line 152-154) but this one was missed.

The `mark_dbc_connected()` call is now an unnecessary hidden side effect imposed on every caller of `alloc_explicit_desc`. Either drop it, or keep it and say why:

```rust
/// Allocate an explicit descriptor (`SQLAllocHandle(SQL_HANDLE_DESC, ...)`)
/// under the same DBC. The returned handle is tracked and freed on drop,
/// before the DBC. Marks the DBC connected first (idempotent) so tests get a
/// realistic connected-session baseline — not required by `alloc_desc`
/// itself, which doesn't gate on connection state.
```

Vahid-b: Good catch, missed this sibling comment. Fixed in 9734bd37 — used your suggested wording verbatim. Kept the `mark_dbc_connected()` call rather than dropping it, since every current caller (`set_stmt_attr.rs`'s ARD/APD association tests) is testing statement-level association against a connection, where a realistic connected-session baseline is still the right setup — just no longer justified by a gate that doesn't exist.

> [!NOTE]

David-Engel (approved):
## Approved

`9734bd37` applies the last open nit (the stale `08003` doc comment on `alloc_explicit_desc`) and is doc-only. Every finding from my two prior reviews is resolved, and all review threads are closed.

Re-verified on this commit: 19/19 checks green, including the full ADO validation matrix (Windows/Linux/macOS, both ARM legs), Kerberos, the cross-repo mssql-python suite on the mssql-odbc driver, and merged coverage. Locally on the same tree: `cargo bfmt` clean, `cargo bclippy` clean, `cargo nextest run -p mssql-odbc --lib` → 895 passed, 1 skipped.

What I checked across the review, for the record: `validate_descriptor_association` and its HY017/HY024 mapping against the ODBC reference and the cited msodbcsql call sites; the `free_desc` / `sql_disconnect_safe` / `free_dbc` lifecycle including the double-free and lock-ordering paths; `DescKind::Ad` classification against `classify_field`; and `StmtHandle::Drop`'s handling of the four implicit descriptors. The parity claims that mattered were verified against msodbcsql source rather than MS Learn — including the one that came back the opposite of the original assumption and led to removing the `08003` gate entirely.

Two things deliberately left as follow-ups, both agreed in-thread and neither blocking:

- The stale-handle window shared with `free_stmt` (reading `desc.object_type` / `parent_dbc` before the "already dropped" guard), which belongs with the existing refcounted-handle-lifetimes TODO in `disconnect.rs`.
- The pre-existing iterate-then-clear pattern in `sql_disconnect_safe`'s statement loop, now inconsistent with the pop-then-free descriptor loop directly below it.

Worth filing those so they don't get lost. Merge is yours — enable auto-merge if you want it to land as soon as the remaining requirements are satisfied.

David-Engel:
Filed the two follow-ups I mentioned in my approval, so they don't get lost when this merges:

- #400 — `SQLFreeHandle` on a stale STMT/DESC handle dereferences freed memory before the "already dropped" guard. Pre-existing in `free_stmt`; `free_desc` mirrors it deliberately. Overlaps the refcounted-handle-lifetimes TODO at `disconnect.rs:85`.
- #401 — `SQLDisconnect`'s statement loop can double-free on retry after a poisoned mutex, now inconsistent with the pop-then-free descriptor loop directly below it.

Neither blocks this PR.

PR #371, opened 2026-08-25
mssql-odbc: honor vendor connection attributes 1203/1223/1228 (AB#47457)

saurabh500:
## What

Honors the three vendor connection attributes whose behaviour I could measure end to end against msodbcsql, routing each to the same setting its connection-string keyword drives.

| id | attribute | keyword it duplicates |
|---|---|---|
| 1203 | `SQL_COPT_SS_INTEGRATED_SECURITY` | `Trusted_Connection` |
| 1223 | `SQL_COPT_SS_ENCRYPT` | `Encrypt` |
| 1228 | `SQL_COPT_SS_TRUST_SERVER_CERTIFICATE` | `TrustServerCertificate` |

This is the `attrs_before` parity surface from spec §4.10: `mssql-python` applies that dict unfiltered at connect, so these ids reach the driver even though the post-connect `set_attr` allowlist never mentions them.

Part of [AB#46377](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46377). Stacked on #352.

## Measured contract

Everything below came from probing the real driver, not from the headers.

**Precedence is per-attribute, not one global rule.** For all three of these the *attribute wins* over the keyword — `1228 = 0` overrides `TrustServerCertificate=yes` hard enough to fail the handshake with `08001`. That is the opposite of `SQL_ATTR_CURRENT_CATALOG` (109), where the `Database=` keyword wins and the attribute is only a fallback (shipped in #348). So the rule has to be established per id; inheriting it would have got `CURRENT_CATALOG` backwards.

**Values are normalized, not range-checked.** `SQL_COPT_SS_ENCRYPT` takes `0` off / `1` on / `2` TDS 8.0 strict, but out-of-range input is folded rather than rejected: set `7` and the connection comes up encrypted and a later get returns `1`. No `HY024`. Strict is identifiable because it stops honoring `TrustServerCertificate`, so `2` fails a self-signed handshake that `1` accepts.

**Post-connect set returns `HY011`.**

**The get reports the effective connection setting, not the stored input.** With no attribute ever set, `Encrypt=no` reads back `0` and an encrypted connection reads back `1` — the value is sourced from the keyword when the attribute was never used. That drove the storage shape: values are kept as normalized `Option<u32>` codes and the connect path writes the *resolved* settings back, so the get is a trivial echo of something already correct.

And `Encrypt=no;TrustServerCertificate=Yes` reads the trust flag back as **`0`** — with encryption off there is no certificate in play. **I did not predict this one.** I wrote variation 62 asserting the flag tracked the keyword, it passed on the Rust driver and *failed on msodbcsql*, and the fix was to the Rust driver. The e2e caught a divergence the probes had left ambiguous, because the earlier probe used `Encrypt=no` with no `TrustServerCertificate` keyword at all, so `0` looked like it could just be the default.

## What is deliberately *not* in here

`APPLICATION_INTENT`, `MULTISUBNET_FAILOVER`, `CONNECT_RETRY_COUNT`/`_INTERVAL`, `SERVER_SPN`, `AUTHENTICATION` and the rest of the vendor band have sweep return codes in the recognition table from #349, but no *confirmed routing*. Wiring them up on the assumption that they behave like the three above is exactly the assumption `CURRENT_CATALOG` disproves, so they go to the follow-up work item, [AB#47526](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47526), with their own measurements. Same for the statement-option fan-out, the read-only diagnostics, and the deferred-feature rejections — see the follow-up section of `attributes_plan.md`.

## Implementation note

`encryption_setting()` in `driver_connect.rs` is now the single source of truth shared by the mssql-tds `EncryptionOptions` builder and the value reported back through `SQLGetConnectAttr`, pinned by a unit test. Before this the mapping was inline in the builder; duplicating it for the get path would have let the reported value drift from the negotiated one.

## Testing

- **6 new unit tests**, 957 total passing.
- **e2e variations 59–62** in `attributes_test.cpp`, covering both override directions, the out-of-range fold, the post-connect `HY011`, and the effective-value get. Variation 62 also connects a second time with `Encrypt=yes` to show the trust flag *does* read back as 1 there — otherwise the `0` in the first half could be read as the driver ignoring the keyword rather than reporting the encryption state.
- **All 63 variations pass on the Rust driver and all 61 non-skipped pass on msodbcsql**, unmodified.
- **Diff coverage 96.48%.**

## Checklist

- [x] `cargo bfmt`
- [x] `cargo bclippy`
- [x] `cargo btest`
- [x] e2e parity verified locally against both drivers
- [x] Plan doc updated with the measured contract and the follow-up split

### DIFF
--- a/mssql-odbc/docs/attributes_plan.md
+++ b/mssql-odbc/docs/attributes_plan.md
@@ -446,27 +446,76 @@ hardcodes "never TRUE" — wiring it into catalog-function matching is S5 work.
 
 ---
 
-### S5 — Connection attributes: vendor + remaining standard, and `attrs_before` parity
+### S5a — Vendor connection attributes that route to a keyword — **shipped**
 
-> `mssql-odbc | Vendor connection attributes & attrs_before parity`
+> `mssql-odbc | Vendor connection attributes & attrs_before parity` (AB#47457)
+
+**Scope:** the three vendor connection attributes whose behaviour was measured
+end to end against msodbcsql, routed to the same setting the equivalent
+connection-string keyword drives.
+
+| id | attribute | keyword it duplicates |
+|---|---|---|
+| 1203 | `SQL_COPT_SS_INTEGRATED_SECURITY` | `Trusted_Connection` |
+| 1223 | `SQL_COPT_SS_ENCRYPT` | `Encrypt` |
+| 1228 | `SQL_COPT_SS_TRUST_SERVER_CERTIFICATE` | `TrustServerCertificate` |
+
+**Measured contract** (probes + e2e variations 59–62, both drivers):
+
+- **Precedence is per-attribute, not one global rule.** For all three of the
+  above the *attribute wins* over the keyword — `1228 = 0` overrides
+  `TrustServerCertificate=yes` hard enough to fail the handshake with `08001`.
+  This is the opposite of `SQL_ATTR_CURRENT_CATALOG` (109), where the `Database=`
+  keyword wins and the attribute is only a fallback (S3). So the rule has to be
+  established per id, not inherited.
+- **Values are normalized, not range-checked.** `SQL_COPT_SS_ENCRYPT` takes
+  `0` off / `1` on / `2` TDS 8.0 strict, but out-of-range input is *folded*
+  rather than rejected: set `7` and the connection comes up encrypted and a
+  later get returns `1`. No `HY024`. Strict is identifiable because it stops
+  honoring `TrustServerCertificate`, so `2` fails a self-signed handshake that
+  `1` accepts.
+- **Post-connect set is rejected** with `HY011`.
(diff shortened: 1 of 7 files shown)


saurabh500:
Collapsed into #348 for review.

This PR shows as merged because its commits are now reachable from its base branch — the stack was fast-forwarded, not squashed or rewritten, so every commit and its message survives unchanged in #348. Nothing here was discarded.

The stack was hard to review one layer at a time, and only the bottom PR ever got CI (the rest had a non-`main` base, so their pipelines never fired). Collapsing gives one reviewable change that CI actually runs on.

Review and any further discussion belong on #348. The threads here stay readable for history.

PR #372, opened 2026-08-25
Do not treat SQLSELECT-tagged DONE_COUNT as an update count

Theekshna:
Fixes [AB#47538](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47538/) and [AB#47535](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47535/)

## Problem

`test_execute_stored_procedure_without_parameters` (mssql-python suite) failed with SQLSTATE `24000`. `SQLExecDirect` positioned the statement on a leading no-row result, so the immediate `fetchall()` never reached the procedure's rows.

## Cause

SQL Server compiles variable assignment (`DECLARE @x int = 1`, `SET`, `SELECT @x = col`) as a **`SQLSELECT` (`0xc1`)** command that still sets `DONE_COUNT`. Captured from SQL Server 2022:

```
DECLARE @out int = 1; EXEC dbo.p;

 -> DoneToken { status: MORE | COUNT, cur_cmd: Select, row_count: 1 }   <- assignment
 -> COLMETADATA + rows                                                  <- EXEC
```

`advance_to_result_boundary` treated any `DONE_COUNT` as an update count, so it stopped on the assignment and exposed a 0-column cursor. (A bare `DECLARE @out int;` sets no `COUNT` flag and was never affected.)

## Fix

Treat `DONE_COUNT` as an update count only when the command is not `SQLSELECT`. Verified against a live server:

| Batch | Before | After (and msodbcsql) |
|---|---|---|
| `DECLARE @out int = 1; EXEC dbo.p;` | `NoRows{1}`, `Rows` | `Rows` |
| `DECLARE @x int; SET @x = 1; SELECT 1;` | `NoRows{1}`, `Rows` | `Rows` |
| `DECLARE @x int = 1;` | `NoRows{1}` (`SQLRowCount` 1) | end of results (`SQLRowCount` -1) |
| `UPDATE t SET v = v; SELECT * FROM t;` | `NoRows{2}`, `Rows` | unchanged |
| `SELECT 1 AS a INTO #t; SELECT * FROM #t;` | `NoRows{1}`, `Rows` | unchanged |

Genuine DML counts are unaffected: `INSERT`/`UPDATE`/`DELETE`/`MERGE` report their own `CurCmd`, and `SELECT ... INTO` reports `DONE_COUNT` with no `CurCmd`.

### Second fix: stale count on message-only results

`advance_to_result_boundary` never cleared `last_rows_affected` when returning a message-only result, and both `SQLRowCount` sites (`exec_common.rs:262`, `more_results.rs:149`) read that field rather than the enum's `rows_affected` — so the preceding statement's count was reported on the message. Partly pre-existing, but this PR routes the warned-assignment case onto that branch, so it is fixed here. Each case below has a genuine `UPDATE` in front; msodbcsql reports `2`, `-1`, `-1` and we now match:

`UPDATE; PRINT 'x'; SELECT 5` · `UPDATE; <assignment raising a warning>; SELECT 5` · `UPDATE; RAISERROR(...,0,1); SELECT 5` — all were `2`, `2`, `-1` before.

`dml_result_counts` does not leak the same way: its only consumer is the `!has_open_batch()` branch at `exec_common.rs:287`, unreachable while positioned on a message-only result.

## Reference driver parity

**.NET SqlClient** (`TdsParser.TryProcessDone`) applies exactly this one guard — `if (curCmd != TdsEnums.SELECT)`, with `SELECT = 0xc1` — under the comment *"Skip the bogus DONE counts sent by the server"*.

**msodbcsql** (`sqlctokn.cpp:2149-2156`) applies four:

| Guard | Here | Notes |
|---|---|---|
| `Info != SQLSELECT` (0xc1) | implemented | this fix; also SqlClient's only guard |
| `RSTypeOld != RS_SELECTION` | not needed | `sqlctokn.cpp:2689` resets it to `RS_NONE` while looping. Confirmed: msodbcsql reports the count for `SELECT * FROM t; UPDATE t SET v = v;`, and so do we |
| `Info != SQLFETCHCURSOR` (0x21) | not implemented | not observable — see below. SqlClient omits it too |
| `Info != SQLDBCC` (0xe6) | not implemented | no DBCC found that sets `DONE_COUNT`. SqlClient omits it too |
| `IS2xAPP(...)` | N/A | ODBC 2.x compatibility |

**Why the cursor guard is not needed.** `cursor_ops::cursor_fetch` does route `sp_cursorfetch` through the changed function, so this was checked rather than assumed. Raw tokens from a live 2-row block fetch:

```
DONEINPROC  status=0x11 (MORE|COUNT)  cur_cmd=0xc1  row_count=2
DONEPROC    status=0x00               cur_cmd=0xe0  row_count=0
```

The server tags the fetch DONE `0xc1`, so this PR's guard already covers it; `last_rows_affected` measured `-1` throughout. `0x21` was never seen on the wire, and a variant for it would be untestable. Probed and found not to set `DONE_COUNT`: `DBCC` (CHECKIDENT, UPDATEUSAGE, CLEANTABLE, CHECKTABLE, USEROPTIONS, OPENTRAN, SQLPERF, INPUTBUFFER, TRACEON), T-SQL cursor `OPEN`/`FETCH`, `MERGE`, `IF`, `SET NOCOUNT`.

**Unknown `CurCmd`.** The `_ => CurrentCommand::None` catch-all in `tokens.rs` is now load-bearing: an unrecognized `CurCmd` lands in the "real update count" branch. That is deliberate parity — both reference drivers deny-list rather than allow-list. A comment records this.

## Tests

- **9 unit tests** in `mssql-tds` over token streams scripted from the live captures: leading assignment count, assignment-only batch, consecutive assignments, assignment preceded by a message, a warning raised *by* the assignment, update count after an assignment, `SELECT ... INTO`.
- 2 of those are tripwires for the deliberate `SQLFETCHCURSOR`/`SQLDBCC` divergence, which only a comment held before. Verified they trip by modelling `0x21` then excluding it — they fail on different steps, so both are needed.
- **1 unit test** in `mssql-odbc`: `SQLExecDirect` opens the row set with `SQLRowCount` -1.
- **10 e2e tests** in `exec_direct_test.cpp`, unguarded and passing against real msodbcsql18, so they pin behavior to the reference driver rather than to us.

## Validation

`cargo bfmt` / `cargo bclippy` clean. Scoped `cargo nextest` 226/226. e2e verified locally against msodbcsql (24/24); the Rust-driver leg needs admin to register the driver, so it runs in CI.

## mssql-python suite impact

Build [169897](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=169897) is this PR's head (`0980561163`); baselines are contemporaneous runs without this fix (169833, 169866, 169885). Identical pattern across two heads of this PR and all three baselines — no flakiness.

**Fixed (6):** `test_execute_stored_procedure_without_parameters` ([AB#47538](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47538)) and `test_long_print_message[2047, 4000, 8000, 8001, 80000]` ([AB#47535](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47535)).

The five `PRINT` tests are the same root cause: stopping on the leading SQLSELECT-tagged `DONE_COUNT` meant the batch's INFO token was never read, so `cursor.messages` was empty and the tests died on `IndexError`. **This PR closes [AB#47535](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47535) as well.**

**Newly red (2) — pre-existing bug unmasked, tracked as [AB#47532](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47532):** `test_long_raiserror[2047]` and `[4000]` were passing *vacuously* — the body is a `try/except ProgrammingError` with no `else` and no `pytest.fail`, so an `execute()` that raises nothing passes having asserted nothing. Before this PR the ERROR token was never read, so nothing was raised. Now it surfaces: `RAISERROR` reports server error 50000, which is unmapped, and `post_tds_error` falls back to a flat `HY000` → `OperationalError`, while the test catches `ProgrammingError`. msodbcsql instead applies a severity-tiered fallback yielding `42000` for severity 11–18 (`sqlcerr.cpp:1385-1401`).

No file in this PR touches SQLSTATE mapping, so this is not a regression from it — had the baseline raised at all, it would have raised the same `OperationalError` and failed then too. Root cause and suggested fix are recorded on [AB#47532](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47532).

**Net: 6 fixed, 2 unmasked, +4 passing. No other test changed state.**

### DIFF
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -3645,6 +3645,16 @@ impl TdsClient {
                     // the SELECT, not the bare CREATE. `rows_affected` is
                     // `Some(n)` only for an update-count result.
                     if has_update_count || saw_message {
+                        if !has_update_count {
+                            // A message-only result carries no count. Clear any
+                            // count captured from a preceding statement in the
+                            // same batch so it is not re-reported as this
+                            // result's `SQLRowCount` — msodbcsql reports -1
+                            // here (verified for `UPDATE; PRINT`,
+                            // `UPDATE; RAISERROR` and `UPDATE;` followed by a
+                            // warning-raising assignment).
+                            self.last_rows_affected = -1;
+                        }
                         self.execution_context.set_has_open_batch(!is_last);
                         return Ok(ResultBoundaryKind::NoRows {
                             rows_affected: if has_update_count {
@@ -6746,23 +6756,62 @@ mod tests {
         );
     }
 
+    /// Divergence pin: msodbcsql excludes `SQLFETCHCURSOR` (0x21) and `SQLDBCC`
+    /// (0xe6) alongside `SQLSELECT`; this driver excludes only `SQLSELECT`,
+    /// matching .NET SqlClient. Neither value is modelled, so a `DONE_COUNT`
+    /// carrying one parses as `None` and *is* reported as an update count.
+    /// Fails the moment someone adopts the msodbcsql-only exclusions, forcing
+    /// that `SQLRowCount` change to be deliberate rather than incidental.
+    #[tokio::test]
+    async fn done_count_for_msodbcsql_only_exclusions_is_an_update_count() {
+        for raw in [0x21u16, 0xe6] {
+            let cmd = CurrentCommand::try_from(raw).unwrap();
+            let mut client = create_test_client_with_tokens(vec![done_count(cmd, 3, false)]);
+
+            let result = client
+                .execute("EXEC sp_cursorfetch @handle".to_string(), ())
+                .await
+                .unwrap();
+
+            assert_eq!(
+                result,
+                StatementResult::NoRows {
+                    rows_affected: Some(3)
+                },
+                "raw cur_cmd {raw:#x}"
+            );
+            assert_eq!(client.last_rows_affected(), 3, "raw cur_cmd {raw:#x}");
+            assert_eq!(
+                client.take_dml_result_counts(),
+                vec![3],
+                "raw cur_cmd {raw:#x}"
+            );
+        }
+    }
+
     /// The only path where `saw_message` and `has_count` land on the *same* DONE:
     /// a warning raised by the assignment itself. The result must still surface
     /// so the message is visible, but must carry no row count — this PR changed
     /// `rows_affected` here from `Some(1)` to `None`.
     ///
+    /// A genuine `UPDATE` precedes the assignment so the assertion is real: with
+    /// a bare assignment `last_rows_affected` is still `-1` from `begin_command`
+    /// and would pass even if the message-only branch leaked a prior count.
+    ///
     /// Live capture for
-    /// `DECLARE @x int; SELECT @x = MAX(v) FROM #t; SELECT 5 AS a;`
+    /// `UPDATE #t SET v = v; DECLARE @x int; SELECT @x = MAX(v) FROM #t; SELECT 5 AS a;`
     /// where `#t.v` contains a NULL:
     /// ```text
+    /// DONE status=MORE|COUNT cur_cmd=Update row_count=2
     /// INFO 8153 "Warning: Null value is eliminated by an aggregate or other SET operation."
     /// DONE status=MORE|COUNT cur_cmd=Select row_count=1
     /// COLMETADATA(1) + row
     /// ```
-    /// msodbcsql reports `cols=0 rowcount=-1` then `cols=1 rowcount=-1`.
+    /// msodbcsql reports `rowcount=2`, then `rowcount=-1`, then `rowcount=-1`.
     #[tokio::test]
     async fn execute_surfaces_assignment_warning_without_rowcount() {
         let mut client = create_test_client_with_tokens(vec![
+            done_count(CurrentCommand::Update, 2, true),
             info_token(
                 8153,
                 0,
@@ -6774,7 +6823,8 @@ mod tests {
 
         let first = client
             .execute(
-                "DECLARE @x int; SELECT @x = MAX(v) FROM #t; SELECT 5 AS a;".to_string(),
+                "UPDATE #t SET v = v; DECLARE @x int; SELECT @x = MAX(v) FROM #t; SELECT 5 AS a;"
+                    .to_string(),
                 (),
             )
             .await
@@ -6782,14 +6832,26 @@ mod tests {
         assert_eq!(
             first,
             StatementResult::NoRows {
+                rows_affected: Some(2)
+            }
+        );
+        assert_eq!(client.last_rows_affected(), 2);
+
+        let second = client.advance().await.unwrap();
+        assert_eq!(
+            second,
+            StatementResult::NoRows {
                 rows_affected: None
             }
         );
-        assert_eq!(client.last_rows_affected(), -1);
-        assert!(client.take_dml_result_counts().is_empty());
+        assert_eq!(
+            client.last_rows_affected(),
+            -1,
+            "the UPDATE's count must not leak into the message-only result"
+        );
 
-        let second = client.advance().await.unwrap();
-        assert_eq!(second, StatementResult::Rows);
+        let third = client.advance().await.unwrap();
+        assert_eq!(third, StatementResult::Rows);
         assert_eq!(
             client
                 .get_current_metadata()
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -6673,7 +6673,7 @@ mod tests {
             done_count(CurrentCommand::Select, 1, true),
             done_count(CurrentCommand::Select, 1, true),
             done_count(CurrentCommand::Select, 2, true),
-            empty_col_metadata(),
+            int_col_metadata(1),
         ]);
 
         let result = client
@@ -6686,18 +6686,25 @@ mod tests {
             .unwrap();
 
         assert_eq!(result, StatementResult::Rows);
+        assert_eq!(
+            client
+                .get_current_metadata()
+                .map(|m| m.columns.len())
+                .unwrap_or(0),
+            1
+        );
         assert_eq!(client.last_rows_affected(), -1);
     }
 
-    /// A PRINT before the assignment must still surface as its own message-only
-    /// result; only the assignment's count is suppressed.
+    /// A PRINT on its own statement surfaces as a message-only result (its DONE
+    /// carries no count); the following assignment's count is still suppressed.
     #[tokio::test]
     async fn execute_surfaces_message_but_skips_select_count() {
         let mut client = create_test_client_with_tokens(vec![
             info_token(0, 0, "hi"),
             done_more(),
             done_count(CurrentCommand::Select, 1, true),
-            empty_col_metadata(),
+            int_col_metadata(1),
         ]);
 
         let first = client
@@ -6710,10 +6717,71 @@ mod tests {
                 rows_affected: None
             }
         );
+        assert_eq!(client.last_rows_affected(), -1);
+        assert!(client.take_dml_result_counts().is_empty());
 
         let second = client.advance().await.unwrap();
         assert_eq!(second, StatementResult::Rows);
+        assert_eq!(
+            client
+                .get_current_metadata()
+                .map(|m| m.columns.len())
+                .unwrap_or(0),
+            1
+        );
+    }
+
+    /// The only path where `saw_message` and `has_count` land on the *same* DONE:
+    /// a warning raised by the assignment itself. The result must still surface
+    /// so the message is visible, but must carry no row count — this PR changed
+    /// `rows_affected` here from `Some(1)` to `None`.
+    ///
+    /// Live capture for
+    /// `DECLARE @x int; SELECT @x = MAX(v) FROM #t; SELECT 5 AS a;`
+    /// where `#t.v` contains a NULL:
+    /// ```text
+    /// INFO 8153 "Warning: Null value is eliminated by an aggregate or other SET operation."
+    /// DONE status=MORE|COUNT cur_cmd=Select row_count=1
+    /// COLMETADATA(1) + row
+    /// ```
+    /// msodbcsql reports `cols=0 rowcount=-1` then `cols=1 rowcount=-1`.
+    #[tokio::test]
+    async fn execute_surfaces_assignment_warning_without_rowcount() {
+        let mut client = create_test_client_with_tokens(vec![
+            info_token(
+                8153,
+                0,
+                "Warning: Null value is eliminated by an aggregate or other SET operation.",
+            ),
+            done_count(CurrentCommand::Select, 1, true),
+            int_col_metadata(1),
+        ]);
+
+        let first = client
+            .execute(
+                "DECLARE @x int; SELECT @x = MAX(v) FROM #t; SELECT 5 AS a;".to_string(),
+                (),
+            )
+            .await
+            .unwrap();
+        assert_eq!(
+            first,
+            StatementResult::NoRows {
+                rows_affected: None
+            }
+        );
         assert_eq!(client.last_rows_affected(), -1);
+        assert!(client.take_dml_result_counts().is_empty());
+
+        let second = client.advance().await.unwrap();
+        assert_eq!(second, StatementResult::Rows);
+        assert_eq!(
+            client
+                .get_current_metadata()
+                .map(|m| m.columns.len())
+                .unwrap_or(0),
+            1
+        );
     }
 
     /// A genuine DML count following an assignment is still reported.
@@ -6722,7 +6790,7 @@ mod tests {
         let mut client = create_test_client_with_tokens(vec![
             done_count(CurrentCommand::Select, 1, true),
             done_count(CurrentCommand::Update, 2, true),
-            empty_col_metadata(),
+            int_col_metadata(1),
         ]);
 
         let first = client
@@ -6739,6 +6807,7 @@ mod tests {
             }
         );
         assert_eq!(client.last_rows_affected(), 2);
+        assert_eq!(client.take_dml_result_counts(), vec![2]);
 
         let second = client.advance().await.unwrap();
         assert_eq!(second, StatementResult::Rows);
@@ -6750,7 +6819,7 @@ mod tests {
     async fn execute_keeps_select_into_count() {
         let mut client = create_test_client_with_tokens(vec![
             done_count(CurrentCommand::None, 1, true),
-            empty_col_metadata(),
+            int_col_metadata(1),
         ]);
 
         let first = client
@@ -6764,6 +6833,7 @@ mod tests {
             }
         );
         assert_eq!(client.last_rows_affected(), 1);
+        assert_eq!(client.take_dml_result_counts(), vec![1]);
     }
 
     #[tokio::test]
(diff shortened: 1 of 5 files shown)


mssql-tds/src/connection/tds_client.rs
```diff
@@ -3594,30 +3595,48 @@ impl TdsClient {
...
+                    // and does not even define those constants. We follow SqlClient
+                    // because neither is observable here: no DBCC was found to set
+                    // DONE_COUNT, and although `cursor_ops::cursor_fetch` does route
+                    // `sp_cursorfetch` through `next_rowset` into this function, a
+                    // live SQL Server 2022 tags that DONEINPROC
+                    // `status: MORE|COUNT, cur_cmd: 0xc1` — SQLSELECT, already
+                    // handled below. `last_rows_affected` measured -1 across block
+                    // fetches with and without this guard. Add the variants only if
+                    // a capture ever shows 0x21 or 0xe6 arriving with DONE_COUNT.
+                    let has_update_count = has_count && done.cur_cmd != CurrentCommand::Select;
```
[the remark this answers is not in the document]
Theekshna: Valid — the docs contradicted the new tests. Fixed in e52dfeb6, all four sites:

- `last_rows_affected` now says "carried `DONE_COUNT` **and** was not tagged `SQLSELECT`", with a short note on why (SQL Server sets `DONE_COUNT` on variable assignment while tagging it `SQLSELECT`; neither msodbcsql nor .NET SqlClient reports that as an update count) and a pointer to the detailed comment in `advance_to_result_boundary`.
- `dml_result_counts` field doc and `take_dml_result_counts` both note the same exclusion; the latter's "empty vec when..." list now includes "or only variable assignments".
- `StatementResult::NoRows` / `rows_affected` changed "carried a row count" to "carried an update count", and the field doc now spells out the one non-obvious case: a `SQLSELECT` count that surfaces only because the statement also raised a message still yields `None`. That is the `execute_surfaces_assignment_warning_without_rowcount` test.

Verified rather than assumed: `cargo doc` shows no unresolved-link warnings for any of the new intra-doc links (the remaining warnings in that crate are pre-existing and unrelated). Scoped `cargo nextest` 198/198, `cargo bfmt` / `cargo bclippy` clean. Re-ran the full e2e suite against real msodbcsql18 — 23/23 pass, including `AssignmentOnlyBatchReportsNoRowCount` (`SQLRowCount` -1) and `AssignmentWarningResultReportsNoRowCount`, which are the two that pin the documented behavior to the reference driver.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3594,30 +3606,48 @@ impl TdsClient {
...
+                    // a navigable result only if the statement returned an update
+                    // count or produced messages. Pure DDL / no-op
                     // statements (no count, no messages) are collapsed, exactly
                     // like result-set navigation, so a batch such as
                     // `CREATE; INSERT; SELECT` exposes the INSERT's row count and
                     // the SELECT, not the bare CREATE. `rows_affected` is
-                    // `Some(n)` only when the DONE carried a COUNT.
-                    if has_count || saw_message {
+                    // `Some(n)` only for an update-count result.
+                    if has_update_count || saw_message {
```
David-Engel, line 3647: **Suggestion** — a prior statement's count leaks into the message-only assignment result.

When `has_update_count` is false but `saw_message` is true, this returns `NoRows { rows_affected: None }` but never clears `self.last_rows_affected`. ODBC reads `client.last_rows_affected()` — not the enum's `rows_affected` — in both `exec_common.rs:262` and `more_results.rs:149`, so `SQLRowCount` reports the *previous* statement's count on that result.

Proven, not asserted. I added this probe at this PR's head:

```rust
let mut client = create_test_client_with_tokens(vec![
    done_count(CurrentCommand::Update, 2, true),
    info_token(8153, 0, "Warning: Null value is eliminated"),
    done_count(CurrentCommand::Select, 1, true),
    int_col_metadata(1),
]);
// first  -> NoRows { rows_affected: Some(2) }, last_rows_affected == 2
// second -> NoRows { rows_affected: None },    last_rows_affected == ???
```

Result: `assertion left == right failed: PROBE: message-only result should report -1, got 2`.

This matters because `AssignmentWarningResultReportsNoRowCount` — the e2e test that pins this behavior to msodbcsql — runs the assignment as the *first* counted statement in its batch, so `last_rows_affected` is still `-1` from `begin_command` and the test passes without exercising the leak. Put a genuine DML in front of the assignment and the driver diverges from msodbcsql's `-1`.

The leak is partly pre-existing (`INSERT ...; PRINT 'x';` has the same shape today), but this PR deliberately routes a new case onto that path and documents `-1` as the intended answer, so it is in scope. One-line fix:

```rust
if has_update_count || saw_message {
    if !has_update_count {
        // A message-only result carries no count; don't let a prior
        // statement's count surface as this result's SQLRowCount.
        self.last_rows_affected = -1;
    }
    self.execution_context.set_has_open_batch(!is_last);
```

I applied that and re-ran `cargo nextest run -p mssql-tds --lib --no-fail-fast`: the probe passes and nothing else changes (same 4 pre-existing failures). Worth extending `execute_surfaces_assignment_warning_without_rowcount` with a preceding `done_count(CurrentCommand::Update, 2, true)`, and adding an `UPDATE` before the assignment in the e2e test, so the parity claim is pinned in the case that can actually regress.

David-Engel (commented):
## Summary

The change is correct and well-evidenced. Gating `DONE_COUNT` on `cur_cmd != CurrentCommand::Select` is exactly the guard .NET SqlClient applies, and it fixes the reported 24000. I verified both reference-driver claims against local sources rather than taking the description's word:

- `TdsParser.cs` `TryProcessDone` — `if (cmd != null && DONE_COUNT == (status & DONE_COUNT)) { if (curCmd != TdsEnums.SELECT) { ... InternalRecordsAffected = count; } // Skip the bogus DONE counts sent by the server` — confirmed verbatim.
- `sqlctokn.cpp` — the four-way guard (`RSTypeOld != RS_SELECTION && Info != SQLSELECT && Info != SQLFETCHCURSOR && Info != SQLDBCC && ...`) confirmed, and `tds.h` confirms `SQLSELECT 0xc1`, `SQLFETCHCURSOR 0x21`, `SQLDBCC 0xe6`. Implementing only the `SQLSELECT` guard, with the reasoning recorded in-code, is sound and matches SqlClient.

I also confirmed the change does not leak into other paths: `last_rows_affected` is consumed only by `mssql-odbc` (`exec_common.rs`, `more_results.rs`), `count_map` is write-only, and a row-returning SELECT's trailing DONE is consumed by `drain_rows`, not by the changed function. Reverting the Copilot doc finding and redoing the four doc sites was the right call.

Locally: `cargo bclippy` clean, `mssql-odbc --lib` 813/813, `mssql-tds --lib` 1819/1823 with the 4 failures being the known pre-existing `certificate_validator` fixture tests (they fail on the merge base too).

**Blocking:** none. One suggestion inline on `tds_client.rs`, plus a nit and CI notes below.

### Nit

The PR body and the in-code comment cite `sqlctokn.cpp:2149-2156` / `:2150`; in my checkout the guard sits at 2151-2156 with `Info != SQLSELECT` on 2152. Probably a different source snapshot — worth a glance so the breadcrumb still lands for the next reader.

### CI

All three failing legs look unrelated to this diff; flagging so they are not read as PR fallout:

- `Test Python bindings (mssql-py-core)` — `test_connection_close_terminates_server_session` asserts the SPID is gone after a `sleep(0.5)`; 1 failed, 855 passed. Timing flake in a path this PR does not touch.
- `Build Stage Test MacOS` / `Start SQL Server (Docker, no TLS certs)` — the SQL Server container took a `core.sqlservr` dump. Infra.
- `coverage-report` — `error: no such command: llvm-cov` on the macOS runner; `cargo-llvm-cov` is not installed there. Toolchain, not code.

The coverage comment reports 100% diff coverage (134/134 lines).

mssql-tds/src/token/tokens.rs
```diff
@@ -1116,6 +1116,15 @@ impl TryFrom<u16> for CurrentCommand {
...
             // All unknown values are treated as None, and considered valid.
+            //
+            // This catch-all carries correctness weight: `advance_to_result_boundary`
+            // gates `SQLRowCount` semantics on `cur_cmd`, so an unrecognized value
+            // lands in the "this DONE_COUNT is a real update count" branch. That is
+            // deliberate: both reference drivers deny-list rather than allow-list
+            // (msodbcsql `Info != SQLSELECT && ...`, sqlctokn.cpp:2149-2152; .NET
+            // SqlClient `curCmd != TdsEnums.SELECT`), so an unknown command with a
+            // valid count still counts there too. Do not convert this to an
+            // allow-list; exclude a command by adding its specific variant.
```
saurabh500, line 1127: **Pin this divergence with a test, not just a comment.**

We deliberately implement only the `SQLSELECT` exclusion and skip msodbcsql's `SQLFETCHCURSOR` (0x21) and `SQLDBCC` (0xe6). Nothing in the suite holds that line today: both values fall through this catch-all to `None` and get reported as genuine update counts. A future "add the missing variants" change flips `SQLRowCount` semantics with a fully green suite.

Two tests close that. First, that neither value is modelled:

```rust
// mssql-tds/src/token/tokens.rs — coverage_tests
/// msodbcsql also excludes SQLFETCHCURSOR (0x21) and SQLDBCC (0xe6) from update
/// counts; this driver follows SqlClient and models neither, so both reach the
/// catch-all. Adding either variant silently changes `SQLRowCount` semantics.
#[test]
fn msodbcsql_only_exclusions_are_not_modelled() {
    assert_eq!(CurrentCommand::try_from(0x21).unwrap(), CurrentCommand::None);
    assert_eq!(CurrentCommand::try_from(0xe6).unwrap(), CurrentCommand::None);
}
```

Second, what that actually means at the boundary — the part a reader can't infer from the enum:

```rust
// mssql-tds/src/connection/tds_client.rs — tests
/// Divergence pin: msodbcsql excludes SQLFETCHCURSOR (0x21) and SQLDBCC (0xe6)
/// alongside SQLSELECT; we exclude only SQLSELECT. Neither value is modelled, so
/// a DONE_COUNT carrying one parses as `None` and *is* reported as an update
/// count. Fails the moment someone adopts the msodbcsql exclusions.
#[tokio::test]
async fn done_count_for_msodbcsql_only_exclusions_is_an_update_count() {
    for raw in [0x21u16, 0xe6] {
        let cmd = CurrentCommand::try_from(raw).unwrap();
        let mut client = create_test_client_with_tokens(vec![done_count(cmd, 3, false)]);

        let result = client
            .execute("EXEC sp_cursorfetch @handle".to_string(), ())
            .await
            .unwrap();

        assert_eq!(
            result,
            StatementResult::NoRows {
                rows_affected: Some(3)
            },
            "raw cur_cmd {raw:#x}"
        );
        assert_eq!(client.last_rows_affected(), 3, "raw cur_cmd {raw:#x}");
        assert_eq!(
            client.take_dml_result_counts(),
            vec![3],
            "raw cur_cmd {raw:#x}"
        );
    }
}
```

Verified, not assumed. Both pass at this head. Then I broke the mechanism to check they are real tripwires rather than tests that pass forever: added `FetchCursor = 0x21` to the enum, mapped `0x21` to it, and added `&& done.cur_cmd != CurrentCommand::FetchCursor` to `has_update_count`. Both failed:

- tokens: `left: FetchCursor, right: None`
- client: `left: End, right: NoRows { rows_affected: Some(3) }` — the count is suppressed and the result collapses entirely

They fail exactly when the decision changes, which is the point. I reverted the probe; no changes left in my tree.

mssql-tds/src/connection/tds_client.rs
Theekshna: Valid, and the diagnosis is exactly right. Fixed in 09805611.

I confirmed the mechanism before fixing: both `exec_common.rs:262` and `more_results.rs:149` assign `stmt_state.row_count = client.last_rows_affected()`, ignoring the enum's `rows_affected`, so the field is what `SQLRowCount` returns. Reproduced your probe as a real assertion — `left: 2, right: -1`.

I checked msodbcsql rather than assuming -1 was correct. Three shapes, all with a genuine `UPDATE` (rowcount 2) in front:

```
UPDATE-then-PRINT                 cols=0 rc=2 | cols=0 rc=-1 | cols=1 rc=-1
UPDATE-then-warned-assignment     cols=0 rc=2 | cols=0 rc=-1 | cols=1 rc=-1
UPDATE-then-RAISERROR             cols=0 rc=2 | cols=0 rc=-1 | cols=1 rc=-1
UPDATE-then-silent-assignment     cols=0 rc=2 | cols=1 rc=-1          <- collapses, as this PR intends
```

`-1` in every message-only case. Applied your one-liner verbatim.

On scope: I checked whether `dml_result_counts` leaks the same way and concluded it does not. Its only consumer is `exec_common.rs:287`, in the `!has_open_batch()` terminal branch; the message-only result returns early at line 272. Every push to that vec is immediately followed by a return, so it cannot be read while positioned on a message-only result. I left it alone rather than adding a speculative clear.

Tests, per your suggestion:
- `execute_surfaces_assignment_warning_without_rowcount` now leads with `done_count(Update, 2, true)` and asserts `last_rows_affected() == 2` then `-1`. It fails at the parent commit and passes here, so it is a real regression test rather than one that passes vacuously.
- `AssignmentWarningResultReportsNoRowCount` now runs `UPDATE #agg SET v = v;` first and asserts 2 → -1 → 1 column.
- Added `MessageOnlyResultAfterDmlReportsNoRowCount` for the pre-existing `UPDATE; PRINT;` shape, since that is the same branch and was silently wrong before.

Full e2e suite against real msodbcsql18: 24/24, including both new tests. Scoped `cargo nextest` 226/226; `cargo bfmt` / `cargo bclippy` clean.

mssql-tds/src/token/tokens.rs
Theekshna: Agreed — a comment is not a constraint. Both tests added in 09805611, essentially as written.

I did not take the "verified, not assumed" part on trust; I re-ran your falsification myself, and it surfaced a distinction worth recording. Breaking the mechanism in two separate steps:

1. Map `0x21` to an existing modelled variant only (`0x21 => Ok(CurrentCommand::Abort)`), leaving `has_update_count` untouched:
   - `msodbcsql_only_exclusions_are_not_modelled` **fails** — `left: Abort, right: None`
   - `done_count_for_msodbcsql_only_exclusions_is_an_update_count` **passes** — correctly, since a non-`Select` variant still yields an update count, so no behavior changed.
2. Then add the exclusion to the boundary (`&& done.cur_cmd != CurrentCommand::Abort`):
   - `done_count_for_msodbcsql_only_exclusions_is_an_update_count` **fails** — `left: End, right: NoRows { rows_affected: Some(3) }`. The count is suppressed and the result collapses entirely, exactly as you predicted.

So the two tests are not redundant and they trip on different things: the tokens test catches "someone modelled the value", the client test catches "someone changed what it means". Only the pair covers the full path, and either alone leaves a hole. I recorded that split in the doc comments so the next person does not delete one as a duplicate.

Probes fully reverted; `git diff` against the parent shows only the intended changes.

Scoped `cargo nextest` 226/226, `cargo bfmt` / `cargo bclippy` clean, and the e2e suite is 24/24 against real msodbcsql18.

Vahid-b (approved):

Issue #273, opened 2026-08-13
Sub-second timeouts truncate to 0 and become infinite (two more instances of the `0 = unlimited` sentinel collision)


saurabh500:
All line numbers against `main @ e40e779d`.

## Summary

`0` means **"no timeout / unlimited"** across this codebase. Two places convert a `Duration` to whole seconds with `as_secs() as u32`, which **truncates**. Any remaining budget under one second becomes `0`, and a value meaning *"almost out of time"* is read as *"unlimited time."*

This is the same defect class as #272, reached by a different mechanism. #272 is **saturation** (`deduct_timeout` clamps to `Some(0)`); these two are **truncation**. Same collision, three sites.

The `0 = infinite` convention itself is implemented independently in **five** places, which is why this keeps recurring:

| site | form |
|---|---|
| `connection/tds_client.rs:519` `timeout_to_duration` | `if secs == 0 { None }` |
| `io/packet_writer.rs:132` | `timeout.filter(\|&t\| t > 0)` |
| `connection/bulk_copy_state.rs:105` `from_seconds` | `if timeout_sec == 0 { Self::new(None) }` |
| `connection/metadata_retriever.rs:150` | `if timeout_sec == 0 { None }` — comment says *"to match TdsClient convention"* |

| `connection_provider/tds_connection_provider.rs:149-152` | `match login_timeout { 1.. => Some(..), _ => None }` — states the rule in prose at `:144` |

Five hand-written copies of one convention, in four different syntactic forms, none delegating to another. Verified by keying on the code rather than the prose:

```
if secs == 0          1   tds_client.rs:521
if timeout_sec == 0   2   bulk_copy_state.rs:105  metadata_retriever.rs:150
.filter(|&t| t > 0)   1   packet_writer.rs:132
1.. => Some(          1   tds_connection_provider.rs:150
```

---

## Instance A — `BulkCopy::timeout()`, public API, no reconnect required

`connection/bulk_copy.rs:435`:

```rust
pub fn timeout(mut self, timeout: Duration) -> Self {
    self.options.timeout_sec = timeout.as_secs() as u32;
    self
}
```

`bulk_copy.rs:872` feeds that into `BulkCopyTimeoutState::from_seconds(self.options.timeout_sec)`, where `0` means infinite (`bulk_copy_state.rs:99` — *"A value of 0 means infinite timeout"*).

So:

```rust
[...]

PR #374, opened 2026-08-25
fix: prevent BulkCopy sub-second timeout from becoming infinite

### LINKED ISSUE #285: BulkCopy sub-second timeout truncates to zero and becomes infinite
### Describe the bug

`BulkCopy::timeout(Duration)` converts the supplied duration to whole seconds using `timeout.as_secs() as u32` at `mssql-tds/src/connection/bulk_copy.rs:435`. Any positive duration below one second therefore becomes `0`.

The bulk-copy timeout path defines `0` as an infinite timeout (`BulkCopyTimeoutState::from_seconds(0)` has no deadline), so requesting a short timeout silently disables timeout enforcement.

### Steps to reproduce

1. Configure a bulk copy with a positive sub-second timeout:

```rust
bulk_copy.timeout(Duration::from_millis(500));
```

2. `BulkCopy::timeout` stores `timeout_sec = 0` because `Duration::as_secs()` truncates fractional seconds.
3. Start the bulk-copy operation.
4. Observe that the timeout state treats `0` as infinite and never expires.

The conversion can also be demonstrated directly:

```rust
assert_eq!(Duration::from_millis(500).as_secs() as u32, 0);
assert!(BulkCopyTimeoutState::from_seconds(0)
    .remaining_duration()
    .is_none());
```

### Expected behavior

A positive timeout must never become an unlimited timeout. Since the current public API accepts `Duration`, sub-second values should either retain their precision or be converted safely without producing the `0 = infinite` sentinel (for example, round a positive fractional duration up to one second).

### Actual behavior

Every timeout from 1 ms through 999 ms is truncated to `0` and interpreted as unlimited. A caller requesting the shortest timeout receives no timeout at all.

### Version

`main` at `068efe7a`

### Affected crate

mssql-tds

### Environment

All platforms; the defect is in platform-independent duration conversion.

### Additional context

Related to #273, which tracks the broader class of arithmetic producing the `0 = unlimited` sentinel. This issue isolates the directly reachable public `BulkCopy::timeout(Duration)` case so it can be fixed and regression-tested independently.

PR #270 does not address this path.

nos1dot618:
## Description

Fix `BulkCopy::timeout(Duration)` truncating positive sub-second durations to `0`.

`Duration::as_secs()` discards fractional seconds, so timeouts between 1ms and 999ms were stored as `0`. Since `0` represents an infinite timeout in the bulk-copy timeout state, these values unintentionally disabled timeout enforcement.

Positive durations below one second are now converted to a one-second timeout, while `Duration::ZERO` continues to represent an unlimited timeout. Durations exceeding the maximum `u32` timeout are also saturated at `u32::MAX` instead of wrapping to an unexpected value.

Regression tests cover zero, sub-second, normal, and oversized timeout values.

## Related Issues

Fixes #285 

## Checklist

* [x] `cargo bfmt` passes
* [x] `cargo bclippy` passes
* [x] `cargo btest` passes
* [x] New/changed functionality has tests
* [x] Public API changes are documented

### DIFF
--- a/mssql-tds/src/connection/bulk_copy.rs
+++ b/mssql-tds/src/connection/bulk_copy.rs
@@ -67,7 +67,8 @@ pub struct BulkCopyOptions {
     /// Number of rows in each batch. Default: 0 (all rows in one batch)
     pub batch_size: usize,
 
-    /// Timeout for the operation in seconds. Default: 30
+    /// Timeout for the operation in seconds. Default: 30. A value of `0`
+    /// means unlimited.
     pub timeout_sec: u32,
 
     /// Check constraints on the destination table. Default: false
@@ -417,9 +418,27 @@ impl<'a> BulkCopy<'a> {
         self
     }
 
+    /// Convert a timeout duration to the whole-second representation used internally.
+    ///
+    /// A zero duration represents an unlimited timeout. Any positive duration less
+    /// than one second is rounded up to one second so it cannot be interpreted as
+    /// the unlimited-timeout sentinel. Durations exceeding the maximum `u32`
+    /// timeout are saturated at `u32::MAX`.
+    fn duration_to_timeout_seconds(timeout: Duration) -> u32 {
+        if timeout > Duration::ZERO {
+            u32::try_from(timeout.as_secs().max(1)).unwrap_or(u32::MAX)
+        } else {
+            0
+        }
+    }
+
     /// Set the operation timeout.
     ///
     /// Default: 30 seconds
+    ///
+    /// A timeout of `Duration::ZERO` disables the timeout. Positive durations
+    /// shorter than one second are treated as a one-second timeout. Durations
+    /// exceeding the maximum supported timeout are clamped to that maximum.
     ///
     /// # Arguments
     ///
@@ -432,7 +451,7 @@ impl<'a> BulkCopy<'a> {
     /// bulk_copy.timeout(Duration::from_secs(120)); // 2 minute timeout
     /// ```
     pub fn timeout(mut self, timeout: Duration) -> Self {
-        self.options.timeout_sec = timeout.as_secs() as u32;
+        self.options.timeout_sec = Self::duration_to_timeout_seconds(timeout);
         self
     }
 
@@ -1599,6 +1618,43 @@ mod tests {
         assert!(
             opts.use_internal_transaction,
             "use_internal_transaction should be configurable to true"
+        );
+    }
+
+    #[test]
+    fn test_duration_to_timeout_seconds() {
+        assert_eq!(BulkCopy::duration_to_timeout_seconds(Duration::ZERO), 0);
+        assert_eq!(
+            BulkCopy::duration_to_timeout_seconds(Duration::from_millis(1)),
+            1
+        );
+        assert_eq!(
+            BulkCopy::duration_to_timeout_seconds(Duration::from_millis(500)),
+            1
+        );
+        assert_eq!(
+            BulkCopy::duration_to_timeout_seconds(Duration::from_millis(999)),
+            1
+        );
+        assert_eq!(
+            BulkCopy::duration_to_timeout_seconds(Duration::from_secs(1)),
+            1
+        );
+        assert_eq!(
+            BulkCopy::duration_to_timeout_seconds(Duration::from_secs(2)),
+            2
+        );
+        assert_eq!(
+            BulkCopy::duration_to_timeout_seconds(Duration::from_millis(2500)),
+            2
+        );
+        assert_eq!(
+            BulkCopy::duration_to_timeout_seconds(Duration::from_secs(u32::MAX as u64)),
+            u32::MAX
+        );
+        assert_eq!(
+            BulkCopy::duration_to_timeout_seconds(Duration::from_secs(1_u64 << 32)),
+            u32::MAX
         );
     }
 


nos1dot618:
@microsoft-github-policy-service agree

nos1dot618:
Hi @saurabh500, I've raised a PR for this issue. It fixes `BulkCopy::timeout(Duration)` incorrectly treating positive sub-second timeouts as unlimited, and also saturates oversized durations at `u32::MAX`. I’ve added regression tests for both cases. Would appreciate your review when you get a chance!

David-Engel (commented):
## Summary

Correct, minimal fix for a real bug. `BulkCopy::timeout()` previously truncated any positive duration under one second to `0`, and `BulkCopyTimeoutState::from_seconds(0)` maps to `new(None)` — no deadline — so the shortest requested timeout became no timeout at all. I confirmed that premise in `mssql-tds/src/connection/bulk_copy_state.rs:104`. The remedy is exactly option 3 in tracking issue #273, so it follows the sanctioned direction rather than inventing one. No blocking findings.

**What I ran** (worktree at `2c63a8df`, base `a6213327`):

- `cargo fmt --all -- --check` — clean; `cargo clippy -p mssql-tds --all-targets` — clean.
- `cargo nextest run -p mssql-tds --lib --no-fail-fast` — 1,816 run, 1,812 pass, 4 fail. All 4 are pre-existing `certificate_validator` fixture failures (`tests/test_certificates/*.pem` not generated locally), unrelated to this PR.

**Already resolved:** Copilot's inline comment about `as u32` wrapping oversized durations is stale — it is anchored at `line: null`, predating the amended commit. The current code uses `u32::try_from(...).unwrap_or(u32::MAX)`, which handles `Duration::from_secs(1 << 32)` and `Duration::MAX`. That thread can be resolved.

**Also checked:** `Instant::now() + Duration::from_secs(u32::MAX)` does not overflow `Instant` on any std platform (Linux timespec, Windows QPC ticks, macOS mach ns), so the new saturation cannot become a panic. `mssql-py-core` reaches this via `Duration::from_secs(u64)` (`mssql-py-core/src/cursor.rs:348`), so it never hits the sub-second case but does gain the saturation.

## Suggestion

**The `0 = unlimited` sentinel is documented on the builder but not on the public field** — `mssql-tds/src/connection/bulk_copy.rs:71`.

`BulkCopyOptions::timeout_sec` is `pub` and its doc says only "Timeout for the operation in seconds. Default: 30." A caller constructing the struct literally bypasses `timeout()` and gets no hint that `0` disables enforcement — the same trap this PR closes on the builder path. One line:

```rust
/// Timeout for the operation in seconds. Default: 30. A value of `0`
/// means unlimited.
pub timeout_sec: u32,
```

(Left in the body rather than inline because line 71 is outside the diff.)

## Notes, not findings

- **The rounding asymmetry is intentional and correctly scoped.** Sub-second durations round up to 1s; durations above 1s still floor (`Duration::from_millis(2500)` -> `2`, asserted at `bulk_copy.rs:1647`). That is deliberate: the fix is confined to the range that produced the infinite-timeout sentinel, and existing behavior above one second is left untouched. The API's unit is whole seconds and sub-second granularity is not promised anywhere, so there is nothing to reconcile.
- Issue #273's **Instance B** — the same `as_secs() as u32` truncation at `mssql-tds/src/connection/tds_client.rs:525`, where a capped reconnect budget under one second removes the deadline it is meant to impose — is untouched, correctly, since #285 isolates the `BulkCopy` case. Just confirming it stays tracked under #273.
- CI is partial: `Analyze`, `CodeQL`, `check`, and CLA passed; `coverage-report` skipped. The ADO pipeline still needs an authorized `/azp run`, so the full `cargo btest` matrix has not run on this commit. My local run above covers the `mssql-tds` lib tests.

mssql-tds/src/connection/bulk_copy.rs
```diff
@@ -417,10 +417,28 @@ impl<'a> BulkCopy<'a> {
...
+        }
+    }
+
     /// Set the operation timeout.
     ///
     /// Default: 30 seconds
     ///
+    /// A timeout of `Duration::ZERO` disables the timeout. Positive durations
+    /// shorter than one second are treated as a one-second timeout. Durations
+    /// exceeding the maximum supported timeout are saturated at `u32::MAX`.
```
David-Engel: Nit: the phrase "saturated at `u32::MAX`" leaks the internal representation into public rustdoc. `u32::MAX` seconds is about 136 years, which is not something a caller holding a `Duration` should have to translate.

```suggestion
    /// exceeding the maximum supported timeout are clamped to that maximum.
```

The identical wording on the private helper at line 425 is fine as-is — there the `u32` is the actual return type.

nos1dot618:
Thanks for the review! I've added the documentation for `timeout_sec` to clarify that `0` means unlimited, and also updated the phrasing of the `timeout()` documentation to clarify the sub-second and maximum timeout behavior.

David-Engel:
Note: Please avoid force pushes to the PR branch once reviewers have reviewed. A force push erases a reviewer's visibility into which commits have already been reviewed and what the diff is since they last reviewed. Not a big deal on this small change, but something to keep in mind for future changes. Prefer additional commits and merges from main, when necessary. They all get squashed into a single commit when the PR is merged.

David-Engel (approved):

PR #375, opened 2026-08-25
Bound the DONE_ATTN wait for query cancellation and timeout

### LINKED ISSUE #373: question: `DONE_ATTN` wait for query cancellation.
Normal query cancellation and request timeout appear able to wait indefinitely after sending `TDS ATTENTION`.

In `NetworkTransport`, the token-reading paths call `cancel_read_stream_and_wait()` for `OperationCancelledError` and `TimeoutError`. That helper calls `wait_for_attention_ack(None)`, where `None` explicitly means no timeout. As a result, cancellation may not return if the server or network never delivers a `DONE` token with the `ATTN` flag.

```
    /// Tells the server to stop sending tokens for the token stream being read and waits for
    /// an acknowledgement.
    async fn cancel_read_stream_and_wait(&mut self) -> TdsResult<()> {
        self.cancel_read_stream().await?;
        // Wait indefinitely for attention ACK
        let _ = self.wait_for_attention_ack(None).await?;
        Ok(())
    }
```

There is already a bounded path, `send_attention_with_timeout()`, and `ATTENTION_TIMEOUT_SECONDS` is currently five seconds, but that path appears to be used for bulk-copy cleanup rather than normal query/result-stream cancellation.

Could you clarify the intended contract?

- Is the indefinite `DONE_ATTN` wait intentional for ordinary query cancellation and timeout? (seems like it from the comment)

### Expected behavior

Cancellation should complete within a documented bound. If the server acknowledges ATTENTION, the connection may remain reusable. If acknowledgement fails or times out, the connection should be made observably unusable.

If indefinite waiting is required by the intended API contract, please document that contract and the recommended way for callers to impose a cancellation-cleanup deadline safely.

### Actual behavior

The normal query/result-stream cancellation path waits indefinitely for `DONE_ATTN`. 

### Additional context

This was found while designing PostgreSQL query cancellation propagation into `mssql-tds`. PostgreSQL must complete remote cancellation before surfacing its local interrupt, but it also cannot allow a missing ATTENTION acknowledgement to block backend cancellation forever.

saurabh500:
## Description

Cancelling a read sends TDS `ATTENTION` and then drains the response until the server answers with a `DONE` carrying `ATTN`. Neither half was bounded.

The drain was the reported bug: `cancel_read_stream_and_wait` called `wait_for_attention_ack(None)`, and `None` meant "no timer at all". A server that never answers, or a network that drops the answer, parked the caller inside its own cancellation forever, past whatever deadline the caller had set. Reachable from both the `OperationCancelledError` and `TimeoutError` paths, i.e. `TdsClient::cancel()` and request timeouts.

Review surfaced the same bug one step earlier: `NetworkWriter::send` awaits `write_all`, which parks until the socket accepts the packet, so a peer that has stopped reading stalls the *send* before the drain deadline is ever armed. Bounding only the drain would have left the advertised contract untrue on that path.

Bulk copy already had a bounded path (`send_attention_with_timeout`, 5s), so the contract existed — it just wasn't on the normal query path, and it had the same unbounded send.

**What changed**

- The send and the drain now share **one** deadline of `ATTENTION_TIMEOUT_SECONDS` (5s), the same constant bulk copy uses, documented in `bulk_copy_state.rs` as matching `Microsoft.Data.SqlClient`'s `AttentionTimeoutSeconds`. The drain gets whatever the send left of it.
- Acknowledged → the connection sits at a message boundary and stays reusable, as before.
- Unacknowledged (send stalled, send failed, drain errored, or the bound elapsed) → the stream is parked at an unknown point, so the connection is marked known-dead. That's observable through `TdsClient::is_connection_dead()`, so pools won't hand it out again. A stalled send is the clearest case: it leaves a partial packet on the wire, so the stream can never be resynchronised.
- The bound is spent **once per connection**, not once per call: a cancellation on a connection already known-dead returns immediately instead of paying another 5s. Found in review — repeated cancellations each cost a full 5s before this.
- `cancel_read_stream_and_wait` no longer returns a `Result`. Call sites used `...await?`, so a cleanup failure replaced the caller's `OperationCancelledError`/`TimeoutError` with an unrelated IO error and hid why the read stopped. The known-dead flag is how a failed cleanup is observed now.
- Removed a second bug in the same function: in the unbounded branch, a read error `break`s the loop and the function returned `Ok(true)` — reporting "acknowledged" when the read had actually failed.
- Both entry points share one inherent helper, `send_attention_and_wait`; the `TdsTransport::send_attention_with_timeout` trait method is now a thin delegate. Bulk copy already treats `Ok(false)` as "connection broken, close it", so it inherits the send bound with no call-site change.

`TokenStreamReader::cancel_read_stream_and_wait` in `io/token_stream.rs` has the same loop shape but is `cfg(fuzzing)` only, and its `FuzzReader` serves a fixed in-memory buffer and returns EOF past the end — there is no socket to park on, so the loop always terminates. Left as-is with a comment saying why.

## Tests

There was no coverage of this at all. `timeout_and_cancel.rs` needs a live SQL Server, and a real server always acknowledges, so it can't reach the unacknowledged case either way. Eight tests added in `network_transport.rs`, four of which fail without the fix:

| Test | What it does | Without the fix |
|---|---|---|
| `cancellation_does_not_wait_forever_for_an_attention_acknowledgement` | Pre-cancelled handle against a peer that acknowledges nothing. Asserts the caller still gets `OperationCancelledError`, and that an ATTENTION packet really went on the wire. | **hung** |
| `request_timeout_does_not_wait_forever_for_an_attention_acknowledgement` | Same hang reached via a 50ms request timeout with no cancel handle. Asserts `TimeoutError`. | **hung** |
| `an_unacknowledged_attention_marks_the_connection_dead` | Asserts the connection is observably unusable afterwards. | **hung** |
| `a_stalled_attention_write_does_not_park_cancellation` | Peer that never reads, with a 1-byte buffer, so the 8-byte ATTENTION write parks — a closed receive window in miniature. Asserts cancellation still returns, and the connection is dead. | **hung** |
| `an_acknowledged_attention_leaves_the_connection_usable` | Quiet-path check: peer sends `DONE` with `ATTN`; asserts the connection is **not** marked dead. | passed |
| `the_drain_discards_trailing_tokens_before_the_acknowledgement` | A plain `DONE` queued ahead of the `DONE_ATTN`; asserts the drain skips past it and still leaves the connection usable. | passed |
| `a_second_cancellation_does_not_spend_the_bound_again` | Cancels three times against a peer that never acknowledges; asserts only the first pays the 5s. | pins behaviour added in review |
| `a_row_in_flight_ends_the_drain_and_retires_the_connection` | A `ROW` queued ahead of the `DONE_ATTN`; asserts the drain ends on it, the caller still gets its cancellation, and the connection is retired. | pins current behaviour — see Deferred |

`an_acknowledged_attention_leaves_the_connection_usable` and `the_drain_discards_trailing_tokens_before_the_acknowledgement` are there so the fix can't pass by simply condemning every connection.

Each failing test was checked against the specific thing it pins, not just against the whole bug — e.g. pushing only the *send* deadline out to 24h fails `a_stalled_attention_write_does_not_park_cancellation` and leaves the other attention tests green.

Two details that matter for these tests:

- Most use a new `create_network_transport_with_live_peer_capturing_writes` helper. The existing live-peer helper never *reads*, and sizes its duplex buffer to `data.len().max(1)` — with empty data that's 1 byte, so the client's ATTENTION write blocks. That is the exact stall the new send bound addresses, so the stalled-write test owns its own `duplex(1)` deliberately, while the rest need a peer that keeps the write direction moving. A helper that drops its writer instead would hand the drain an EOF and end it no matter how unbounded it is, which is the masking these tests exist to avoid.
- `#[tokio::test(start_paused = true)]` means the 5s bound elapses as soon as the runtime goes idle. A regression that arms no timer leaves only the outer 600s guard, which then fires and fails the test in milliseconds instead of blocking the suite for real minutes. Every wait sits under that guard, channel receives included — an unguarded receive parks on a peer that never closes, so it would hang on the very regressions being tested. This needs tokio's `test-util` feature, which is **not** part of `full`, hence the dev-dependency addition. `Cargo.lock` records versions rather than features, so `--frozen` is unaffected.

## Deferred

A `ROW`/`NBCROW` still in flight behind the ATTENTION carries no length prefix, so it is parseable only with the preceding `COLMETADATA` in the parser context — and the drain reads with a dummy one. The drain therefore ends on that parse error and this PR retires the connection, which is the common shape for a cancelled row-returning query.

That is not a regression: before this PR the same input took the `break` → `Ok(true)` path and handed back a connection that was usable but desynchronised. Retiring it is strictly safer. It does mean the "acknowledged, stays reusable" branch is largely unreachable for the canonical cancellation, and we retire connections SqlClient would keep.

Making the drain consume those rows needs the `COLMETADATA`-aware loop `TdsClient::drain_stream` already has, not just a context passed down from the call site — only three of the five `cancel_read_stream_and_wait` call sites have a context in scope, and a fresh `COLMETADATA` can arrive mid-drain. That is a behavioural change to cancellation on its own merits, so it is tracked separately in #381 rather than bundled into bounding an unbounded wait. `a_row_in_flight_ends_the_drain_and_retires_the_connection` pins today's outcome so that flip has to be deliberate.

## Related Issues

Fixes #373
Follow-up: #381

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes — workspace lib suite is 2777 passed / 1 skipped / 0 failed, and the mock-server and drain-on-error integration tests pass (13/13). Integration and bench targets that need a live SQL Server and a `.env` don't run in my environment; they're unrelated to this change.
- [x] Public API changes are documented — no signature change. `TdsTransport::send_attention_with_timeout`'s doc now states that its timeout covers the send as well as the wait, that `Ok(false)` can mean either, and that the transport marks itself dead so callers that ignore the return value still behave correctly.

### DIFF
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use crate::connection::bulk_copy_state::ATTENTION_TIMEOUT_SECONDS;
 use crate::connection::client_context::{IPAddressPreference, TransportContext};
 use crate::connection::transport::buffers::TdsReadBuffer;
 use crate::connection::transport::extractable_stream;
@@ -1132,94 +1133,91 @@ impl NetworkTransport {
         Ok(packet_size_from_header)
     }
 
-    /// Tells the server to stop sending tokens for the token stream being read and waits for
-    /// an acknowledgement.
-    async fn cancel_read_stream_and_wait(&mut self) -> TdsResult<()> {
-        self.cancel_read_stream().await?;
-        // Wait indefinitely for attention ACK
-        let _ = self.wait_for_attention_ack(None).await?;
-        Ok(())
+    /// Tells the server to stop sending tokens for the token stream being read
+    /// and waits, for a bounded time, for the acknowledgement.
+    ///
+    /// # Contract
+    ///
+    /// Cancelling a read is itself bounded: ATTENTION goes out and the response
+    /// is drained for at most [`ATTENTION_TIMEOUT_SECONDS`], matching
+    /// `Microsoft.Data.SqlClient`'s `AttentionTimeoutSeconds`. Callers get their
+    /// cancellation or timeout back within that bound whatever the server does,
+    /// so no separate cancellation-cleanup deadline is needed.
+    ///
+    /// Acknowledged, the connection is left at a message boundary and stays
+    /// reusable. Unacknowledged — the send failed, the drain errored, or the
+    /// bound elapsed — the stream is parked at an unknown point, so the
+    /// connection is marked known-dead and pools will not hand it out again.
+    ///
+    /// This reports nothing to the caller on purpose. It runs on a path that
+    /// already has an error to deliver (the cancellation or the timeout), and
+    /// replacing that with a cleanup failure would hide why the read stopped.
+    /// The known-dead flag is how a failed cleanup is observed.
+    async fn cancel_read_stream_and_wait(&mut self) {
+        if let Err(e) = self.cancel_read_stream().await {
+            debug!("Failed to send attention while cancelling the read stream: {e:?}");
+            self.known_dead = true;
+            return;
+        }
+
+        let attention_timeout = Duration::from_secs(ATTENTION_TIMEOUT_SECONDS);
+        match self.wait_for_attention_ack(attention_timeout).await {
+            Ok(true) => {}
+            Ok(false) => {
+                warn!(
+                    timeout = ?attention_timeout,
+                    "No attention acknowledgement within the bound; marking the connection dead"
+                );
+                self.known_dead = true;
+            }
+            Err(e) => {
+                debug!("Error draining to the attention acknowledgement: {e:?}");
+                self.known_dead = true;
+            }
+        }
     }
 
-    /// Wait for attention acknowledgment from server with optional timeout.
+    /// Wait for attention acknowledgment from server, bounded by
+    /// `attention_timeout`.
     ///
-    /// This helper reads tokens until it receives a DONE token with the ATTN flag set,
-    /// discarding any other tokens. If a timeout is specified and expires before
-    /// receiving the ACK, returns `Ok(false)`.
-    ///
-    /// # Arguments
-    ///
-    /// * `attention_timeout` - Optional timeout. If `None`, waits indefinitely.
+    /// Reads tokens until a DONE token with the ATTN flag arrives, discarding
+    /// everything else the server was still sending.
     ///
     /// # Returns
     ///
     /// * `Ok(true)` - Attention acknowledged by server
-    /// * `Ok(false)` - Timeout expired waiting for ACK (only when timeout specified)
+    /// * `Ok(false)` - `attention_timeout` elapsed before the acknowledgement
     /// * `Err(_)` - Error reading response
-    async fn wait_for_attention_ack(
-        &mut self,
-        attention_timeout: Option<Duration>,
-    ) -> TdsResult<bool> {
-        let dummy_context = ParserContext::None(());
-        let start = std::time::Instant::now();
+    async fn wait_for_attention_ack(&mut self, attention_timeout: Duration) -> TdsResult<bool> {
+        let start = tokio::time::Instant::now();
 
-        loop {
-            // Check timeout if specified
-            if let Some(timeout_duration) = attention_timeout
-                && start.elapsed() >= timeout_duration
-            {
-                debug!("Attention ACK timeout after {:?}", start.elapsed());
-                return Ok(false);
+        match timeout(attention_timeout, self.drain_to_attention_ack()).await {
+            Ok(Ok(())) => {
+                debug!("Attention ACK received after {:?}", start.elapsed());
+                Ok(true)
             }
-
-            // Read next token, with timeout if specified
-            let token_result = if let Some(timeout_duration) = attention_timeout {
-                let remaining = timeout_duration.saturating_sub(start.elapsed());
-                match timeout(
-                    remaining,
-                    receive_token_internal(self, &*PARSER_REGISTRY, &dummy_context),
-                )
-                .await
-                {
-                    Ok(result) => result,
-                    Err(_elapsed) => {
-                        debug!(
-                            "Attention ACK timeout (elapsed) after {:?}",
-                            start.elapsed()
-                        );
-                        return Ok(false);
-                    }
-                }
-            } else {
-                // No timeout - wait indefinitely
-                receive_token_internal(self, &*PARSER_REGISTRY, &dummy_context).await
-            };
-
-            match token_result {
-                Ok(token) => {
-                    if let Tokens::Done(done_token) = token
-                        && done_token.status.contains(DoneStatus::ATTN)
-                    {
-                        debug!("Attention ACK received after {:?}", start.elapsed());
-                        return Ok(true);
-                    }
-                    // Discard other tokens and continue waiting
-                }
-                Err(e) => {
-                    if attention_timeout.is_none() {
-                        // When waiting indefinitely, errors just break the loop (original behavior)
-                        break;
-                    }
-                    // When using timeout, propagate errors
-                    debug!(
-                        "Error reading token while waiting for attention ACK: {:?}",
-                        e
-                    );
-                    return Err(e);
-                }
+            Ok(Err(e)) => Err(e),
+            Err(_elapsed) => {
+                debug!("Attention ACK timeout after {:?}", start.elapsed());
+                Ok(false)
             }
         }
-        Ok(true)
+    }
+
+    /// Reads and discards tokens until the DONE carrying ATTN arrives.
+    ///
+    /// Unbounded on its own — every caller wraps it in a timeout.
+    async fn drain_to_attention_ack(&mut self) -> TdsResult<()> {
+        let dummy_context = ParserContext::None(());
+
+        loop {
+            let token = receive_token_internal(self, &*PARSER_REGISTRY, &dummy_context).await?;
+            if let Tokens::Done(done_token) = token
+                && done_token.status.contains(DoneStatus::ATTN)
+            {
+                return Ok(());
+            }
+        }
     }
 }
 
@@ -1677,7 +1675,7 @@ impl NetworkTransport {
             Ok(_) => {}
             Err(err) => match err {
                 OperationCancelledError(_) | TimeoutError(_) => {
-                    Box::pin(self.cancel_read_stream_and_wait()).await?;
+                    Box::pin(self.cancel_read_stream_and_wait()).await;
                 }
                 _ => {}
             },
@@ -1722,7 +1720,7 @@ impl NetworkTransport {
             Ok(_) => {}
             Err(err) => match err {
                 OperationCancelledError(_) | TimeoutError(_) => {
-                    Box::pin(self.cancel_read_stream_and_wait()).await?;
+                    Box::pin(self.cancel_read_stream_and_wait()).await;
                 }
                 _ => {}
             },
@@ -1757,7 +1755,7 @@ impl NetworkTransport {
             Ok(_) => {}
             Err(err) => match err {
                 OperationCancelledError(_) | TimeoutError(_) => {
-                    Box::pin(self.cancel_read_stream_and_wait()).await?;
+                    Box::pin(self.cancel_read_stream_and_wait()).await;
                 }
                 _ => {}
             },
@@ -1788,7 +1786,7 @@ impl NetworkTransport {
             Ok(_) => {}
             Err(err) => match err {
                 OperationCancelledError(_) | TimeoutError(_) => {
-                    Box::pin(self.cancel_read_stream_and_wait()).await?;
+                    Box::pin(self.cancel_read_stream_and_wait()).await;
                 }
                 _ => {}
             },
@@ -1815,7 +1813,7 @@ impl NetworkTransport {
             Ok(_) => {}
             Err(err) => match err {
                 OperationCancelledError(_) | TimeoutError(_) => {
-                    Box::pin(self.cancel_read_stream_and_wait()).await?;
+                    Box::pin(self.cancel_read_stream_and_wait()).await;
                 }
                 _ => {}
             },
@@ -1952,7 +1950,7 @@ impl crate::connection::transport::tds_transport::TdsTransport for NetworkTransp
         self.cancel_read_stream().await?;
 
         // Wait for ACK with timeout using the shared helper
-        self.wait_for_attention_ack(Some(attention_timeout)).await
+        self.wait_for_attention_ack(attention_timeout).await
     }
 
     fn is_connection_dead(&self) -> bool {
@@ -1982,7 +1980,7 @@ pub(crate) mod tests {
     use crate::test_packet_support::{
         TestPacketBuilder, create_network_transport_with_chunked_data,
         create_network_transport_with_data, create_network_transport_with_live_peer,
-        encode_utf16_le,
+        create_network_transport_with_live_peer_capturing_writes, encode_utf16_le,
     };
     use bytes::Bytes;
     use futures::SinkExt;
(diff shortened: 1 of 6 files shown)


mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1132,94 +1133,91 @@ impl NetworkTransport {
...
+    /// The known-dead flag is how a failed cleanup is observed.
+    async fn cancel_read_stream_and_wait(&mut self) {
+        if let Err(e) = self.cancel_read_stream().await {
+            debug!("Failed to send attention while cancelling the read stream: {e:?}");
+            self.known_dead = true;
+            return;
+        }
+
+        let attention_timeout = Duration::from_secs(ATTENTION_TIMEOUT_SECONDS);
+        match self.wait_for_attention_ack(attention_timeout).await {
```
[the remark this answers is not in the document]
saurabh500: Good catch, and it was worse than the drain: `NetworkWriter::send` awaits `write_all`, so a peer that stops reading parks the send before any deadline is armed — the same bug this PR is about, one step earlier. `send_attention_and_wait` now takes a single deadline covering both the send and the drain, and marks the connection dead when the send stalls (a half-written packet can't be resynchronised). Bulk copy already treats `Ok(false)` as a connection it must close, so it inherits the bound unchanged.

Added `a_stalled_attention_write_does_not_park_cancellation` for the non-reading peer. Verified it pins the send bound specifically: with the send deadline pushed out to 24h it fails with `Elapsed(())` while the other 12 attention tests stay green.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -3630,4 +3628,182 @@ pub(crate) mod tests {
...
+        )
+        .await
+        .expect("cancellation hung waiting for a DONE_ATTN the server never sent");
+
+        assert!(
+            matches!(result, Err(OperationCancelledError(_))),
+            "the caller must still see its cancellation, got {result:?}"
+        );
+
+        let sent = written.recv().await.expect("the peer must observe a write");
```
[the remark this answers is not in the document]
saurabh500: Correct — the peer never closes its sender, so `recv()` would park rather than return `None`, and with no timer armed the paused clock has nothing to advance to. Wrapped it in the same 600s guard.

Verified rather than assumed: stubbing `cancel_read_stream_and_wait` to return immediately makes the test fail in 34ms with "cancellation returned without ever putting ATTENTION on the wire" instead of hanging. Also noted in the module preamble that every wait in these tests, channel receives included, has to sit under a guard.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1132,94 +1133,119 @@ impl NetworkTransport {
...
-                }
-            } else {
-                // No timeout - wait indefinitely
-                receive_token_internal(self, &*PARSER_REGISTRY, &dummy_context).await
-            };
+    /// Reads and discards tokens until the DONE carrying ATTN arrives.
+    ///
+    /// Unbounded on its own — every caller wraps it in a timeout.
+    async fn drain_to_attention_ack(&mut self) -> TdsResult<()> {
+        let dummy_context = ParserContext::None(());
```
David-Engel, line 1271: The drain reads with `ParserContext::None(())`, and `RowTokenParser::parse` returns `InvalidData "Expected ColumnMetadata in context"` for anything that isn't `ParserContext::ColumnMetadata`. So a ROW queued behind the ATTENTION ends the drain with `Err`, which the caller turns into `known_dead = true`.

Verified rather than assumed — scratch test feeding a ROW token followed by `DONE_ATTN`:

```
SCRATCH result=Err(OperationCancelledError("Request was cancelled")) known_dead=true
```

The `DONE_ATTN` was sitting right behind it.

Pre-PR the same input took the `break` -> `Ok(true)` path and left the connection usable but desynced, so this is strictly safer. But it means the "acknowledged, connection stays reusable" branch is largely unreachable for the canonical cancellation — a timeout while reading rows through `receive_row_into` — and cancel/timeout will retire connections that SqlClient would keep (it keeps parsing with the running command's metadata until `DONE_ATTN`).

The mechanism is already documented in `drain_stream`: "ROW/NBCROW tokens carry no length prefix and can only be parsed with the preceding COLMETADATA in the parser context."

Every call site of `cancel_read_stream_and_wait` already has `context: &ParserContext` in scope. Is threading it through instead of hardcoding `None` in scope here, or deliberately deferred? Either way it would be worth a test with a ROW in flight so the current outcome is pinned rather than implied — the closest existing test, `the_drain_discards_trailing_tokens_before_the_acknowledgement`, only queues a plain DONE, which parses fine with a dummy context.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1132,94 +1133,119 @@ impl NetworkTransport {
...
+    /// reusable. Unacknowledged — the send failed or stalled, the drain
+    /// errored, or the bound elapsed — the stream is parked at an unknown
+    /// point, so the connection is marked known-dead and pools will not hand
+    /// it out again.
+    ///
+    /// This reports nothing to the caller on purpose. It runs on a path that
+    /// already has an error to deliver (the cancellation or the timeout), and
+    /// replacing that with a cleanup failure would hide why the read stopped.
+    /// The known-dead flag is how a failed cleanup is observed.
+    async fn cancel_read_stream_and_wait(&mut self) {
```
David-Engel, line 1159: Nothing short-circuits once the verdict is already in, so each subsequent cancelled read pays another full bound. Scratch test, three cancelled reads on one transport, paused clock:

```
attempt 1: virtual elapsed 5s, known_dead=true
attempt 2: virtual elapsed 5s, known_dead=true
attempt 3: virtual elapsed 5s, known_dead=true
```

A caller that touches the transport again after a cancellation (cursor close / drain paths re-enter `receive_token`) multiplies the advertised bound. One line covers it:

```rust
async fn cancel_read_stream_and_wait(&mut self) {
    if self.known_dead {
        return;
    }
```

mssql-tds/src/connection/transport/tds_transport.rs
```diff
@@ -43,12 +43,16 @@ pub(crate) trait TdsTransport: TdsTokenStreamReader + Send + Sync + std::fmt::De
...
+    ///   as the wait. Writing the packet can itself stall on a peer that has
+    ///   stopped reading, so implementations must not leave the send unbounded.
     ///
     /// # Returns
     ///
     /// * `Ok(true)` - Attention acknowledged by server
-    /// * `Ok(false)` - Attention sent but timeout expired waiting for ACK
+    /// * `Ok(false)` - Timeout expired before the acknowledgement, either while
+    ///   sending the attention or while waiting for the ACK. The connection is
+    ///   left unusable either way.
```
David-Engel: "The connection is left unusable either way" holds for the send-stall branch, which sets `known_dead` inside `send_attention_and_wait`, but the ack-timeout branch returns `Ok(false)` without it — only the outer `cancel_read_stream_and_wait` marks dead.

On the trait path that matters: `bulk_copy.rs:1028` is fine, it calls `close_connection()`. But `tds_client.rs:2106` does `let _ = self.send_attention_with_timeout(...)` and then `set_has_open_batch(false)`, with a comment saying the connection can be reused. That's pre-existing, but the new doc line now asserts otherwise.

Either move `known_dead = true` into the `Ok(false)` branch of `send_attention_and_wait` (bulk copy closes anyway, so no behavior change there), or narrow the doc to the send case.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -1132,94 +1133,119 @@ impl NetworkTransport {
...
+        let deadline = Instant::now() + attention_timeout;
+
+        match timeout_at(deadline, self.cancel_read_stream()).await {
+            Ok(result) => result?,
+            Err(_elapsed) => {
+                // A stalled write leaves a partial packet on the wire, so the
+                // stream can never be resynchronised.
+                warn!(
+                    timeout = ?attention_timeout,
+                    "Timed out sending the attention packet; marking the connection dead"
```
David-Engel, line 1206: Nit: two warnings for one event. This logs "Timed out sending the attention packet", then returns `Ok(false)`, which makes `cancel_read_stream_and_wait` log "Attention went unacknowledged within the bound" — implying the drain timed out when the send did. Returning a small enum, or downgrading the outer one to `debug!`, keeps the log honest.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -3630,4 +3658,216 @@ pub(crate) mod tests {
...
+    /// follows it.
+    ///
+    /// `create_network_transport_with_live_peer` never reads and sizes its
+    /// duplex buffer to the scripted data alone, so with no data the client's
+    /// 8-byte ATTENTION fills a 1-byte buffer and the write parks — a peer
+    /// whose receive window has closed, in miniature. Timing only the drain
+    /// leaves the caller stuck here, before any deadline is armed.
+    #[tokio::test(start_paused = true)]
+    async fn a_stalled_attention_write_does_not_park_cancellation() {
+        let mut transport = create_network_transport_with_live_peer(&[]);
```
David-Engel: Nit: the stall depends on `create_network_transport_with_live_peer` sizing `duplex(data.len().max(1))`. If that helper's sizing ever changes, the write succeeds, the drain times out instead, and both assertions here (`OperationCancelledError` and `is_known_dead`) still hold — the test passes while no longer pinning the send bound.

Asserting that no ATTENTION reached a peer, or constructing the 1-byte duplex in the test itself, would make it self-contained.

David-Engel (commented):
## Summary

Bounds both halves of the ATTENTION flow under one 5s deadline, matching `Microsoft.Data.SqlClient`'s `AttentionTimeoutSeconds`, and retires the connection when the acknowledgement doesn't arrive. The design matches the reference driver (SqlClient dooms the connection on attention timeout), the send-side extension found in review is the right catch, and dropping the `Result` from `cancel_read_stream_and_wait` correctly stops cleanup errors from masking the caller's cancellation/timeout. Issue link and description both accurate. No blocking findings.

**What I ran**

- `cargo bfmt` and `cargo bclippy` (`--frozen --all-features --all-targets`) both pass — which also confirms the `Cargo.lock`-unaffected claim about the new dev-dependency.
- `cargo nextest run -p mssql-tds --lib --no-fail-fast`: 1821 tests, 1817 pass, 4 fail — all the known `certificate_validator` fixture failures, unrelated to this change.
- Stressed the six new paused-time tests 25 times: no flakes.

Three suggestions and two nits inline. The first one is the substantive one: the drain still can't parse a ROW that's in flight, and with this PR that outcome changes from "silently desynced but reused" to "connection retired", which I think makes the acknowledged/reusable branch largely unreachable for the canonical cancellation.

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Nice catch

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Confirmed and fixed. I reproduced your exact numbers before changing anything — three cancelled reads on one transport gave `5s / 5s / 5s`, `known_dead=true` throughout. `cancel_read_stream_and_wait` now returns immediately when the flag is already set (`network_transport.rs:1160`), and `a_second_cancellation_does_not_spend_the_bound_again` asserts the second cancellation costs `Duration::ZERO`. Reverting the early return fails that test with `left: 5s, right: 0ns`.

> [!NOTE]

mssql-tds/src/connection/transport/tds_transport.rs
saurabh500: You're right — the doc was true at my call site and false on the ack-timeout branch. I made the code match the doc rather than narrowing it: `send_attention_and_wait` now marks the connection dead on every outcome that isn't an acknowledgement, including `Err`. That closes a real hole on `tds_client.rs:2106`, which ignores the result and clears the batch flag — an outstanding ATTENTION means a stray `DONE_ATTN` could later be read as the answer to a different request. `bulk_copy.rs:1028` already called `close_connection()`, so no change there.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Fixed. Each failure branch inside `send_attention_and_wait` now logs its own warning — "Timed out sending the attention packet" or "Attention went unacknowledged within the bound" — and the outer duplicate in `cancel_read_stream_and_wait` is gone. One event, one warning, describing what actually happened.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Verified your concern was real: swapping the helper for one with a roomy buffer — no stalled write at all — left both assertions passing. The test now builds its own `duplex(1)` and holds the peer end, and adds the mechanism check you suggested: after cancellation returns, fewer than 8 bytes reached the peer, so the write demonstrably stalled part-way rather than completing. Widening that duplex to 4096 now fails on exactly that assertion.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Thanks — credit to the issue report for pinning down the unbounded wait.

> [!NOTE]

saurabh500:
@david-engel this is ready for re-review

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Apologies — my earlier reply on this thread was meant for the "nice catch" comment further down and left your actual point unanswered. Answering it properly now.

You're right, and I verified it: `RowTokenParser::parse` rejects on the context before reading a single body byte (`row_parser.rs:119-126`), so a ROW in flight ends the drain with `Err` and my change retires the connection. As you note, pre-PR the same input took the `break` → `Ok(true)` path and left the connection usable but desynced, so this is strictly safer — but it does mean the "acknowledged, stays reusable" branch is largely unreachable for the canonical cancellation.

On "in scope or deferred": deferred, deliberately. Threading the caller's `&ParserContext` down isn't sufficient on its own — only three of the five `cancel_read_stream_and_wait` call sites have a context in scope (`resume_row_into` and `read_active_plp_bytes` are parked mid-row/mid-PLP, with no token boundary to resume from), and a fresh COLMETADATA can arrive during the drain, so a single captured context wouldn't cover it. The real fix is the COLMETADATA-aware loop `TdsClient::drain_stream` already has — read, and on COLMETADATA switch to `drain_result_set_rows` with `ColumnPolicy::SkipAll` — plus care that the drain doesn't re-enter `receive_token`'s cancel handling. That's a behavioural change to cancellation on its own merits, not part of bounding an unbounded wait, and I'd rather not land it unsupervised in desync territory.

Took your test suggestion: `a_row_in_flight_ends_the_drain_and_retires_the_connection` queues a ROW then `DONE_ATTN` and asserts the caller still gets its cancellation, the acknowledgement is never reached, and the connection is marked dead. It pairs with `the_drain_discards_trailing_tokens_before_the_acknowledgement` as a controlled comparison — same stream shape, parseable token, opposite outcome — so the ROW is provably what causes the retirement. It's written to fail loudly if someone teaches the drain to consume rows, so that flip can't happen silently. Also documented the constraint on `drain_to_attention_ack` itself.

Filed #381 with the analysis and the open questions (reentrancy, the mid-row sites, keeping genuinely corrupt streams retiring). Pushed in 50c5908.

> [!NOTE]

David-Engel (commented):
## Re-review — all five findings resolved

Re-reviewed at `50c59082`, three commits on from the version I first reviewed. No remaining findings; nothing blocking. My five threads can be resolved.

**Verified on this head**

- `cargo bfmt` and `cargo bclippy` (`--frozen --all-features --all-targets`) pass with no warnings.
- `cargo nextest run -p mssql-tds --lib --no-fail-fast`: 1823 tests, 1819 pass, 4 fail — the same `certificate_validator` fixture failures that fail on a clean tree, unrelated to this change.
- Seven attention tests stressed 25 times: no flakes.
- Mutation-checked the three new guards rather than taking the replies on trust:
  - Deleting the `if self.known_dead { return; }` early return fails `a_second_cancellation_does_not_spend_the_bound_again`.
  - Widening the test's `duplex(1)` to `duplex(4096)` fails `a_stalled_attention_write_does_not_park_cancellation` on exactly the new mechanism assertion ("the write must have stalled part-way").
  - Dropping `self.known_dead = true` from the drain's `Err` arm fails `a_row_in_flight_ends_the_drain_and_retires_the_connection`, and only that one out of the fifteen.

**Where each finding landed**

| Finding | Outcome |
|---|---|
| ROW in flight retires the connection | Behaviour deferred to #381 — the right call, that is a design change rather than this PR's job — and the current outcome is now pinned by a test plus a doc comment on `drain_to_attention_ack`, so changing it later has to be deliberate. |
| Repeat cleanup spends the bound again | Early return plus a test asserting the second cancellation costs `Duration::ZERO`. |
| `Ok(false)` doc overstated the code | Fixed by changing the code to match the doc rather than narrowing the doc. |
| Duplicate, misattributed warning | Each failure branch logs its own message; the outer duplicate is gone. |
| Stalled-write test leaned on another helper's sizing | Test owns its `duplex(1)` and asserts fewer than 8 bytes reached the peer. |

Two of these came back stronger than what I proposed. The `Ok(false)` fix now marks the connection dead on all four non-acknowledgement outcomes, which closes the hole at `tds_client.rs:2106` where the result is ignored and the batch flag cleared — that could previously hand back a connection with an ATTENTION outstanding. I checked the flag is write-once: there is no `known_dead = false` anywhere, and no path swaps a fresh stream onto an existing `NetworkTransport` (703 and 746 are TLS setup during connect), so the early return cannot strand a revived connection.

The last coverage run predates the final commit; the two branches it flagged (1169-1171 and 1227) are the ones `a_row_in_flight_ends_the_drain_and_retires_the_connection` now covers, so the next run should pick them up.

David-Engel (approved):
Approving at `50c59082`.

Bounds both halves of the ATTENTION flow under one 5s deadline matching `Microsoft.Data.SqlClient`'s `AttentionTimeoutSeconds`, retires the connection on every outcome that isn't an acknowledgement, and makes the bound belong to the connection rather than to each call. Every finding from my earlier rounds is resolved, and each fix is pinned by a test I mutation-checked — removing the known-dead early return, widening the stalled-write duplex, or dropping the drain's `known_dead` assignment each fails exactly one test and no others.

The ROW-in-flight behaviour is correctly deferred to #381 with a test locking today's outcome in place, so changing it later has to be deliberate rather than a silent flip.

Local verification on this head: `cargo bfmt` and `cargo bclippy` clean, 1823 lib tests with only the pre-existing `certificate_validator` fixture failures, seven attention tests stressed 25 times with no flakes.

Remaining macOS and Linux legs were still running when I approved; worth a glance before merging.

PR #376, opened 2026-08-25
mssql-odbc: match msodbcsql number rendering, and close the P5 type-coverage gaps

David-Engel:
## Summary

Two driver fixes and the e2e coverage that found them, plus the recorded outcome of P4.

A coverage audit against the plan's fetch type map (SQL type → C type, the pairs `mssql-python` actually requests) found that nine of eighteen C targets had **no e2e coverage on either fetch path** — including `BIGINT`, `BIT`, `REAL`, `FLOAT`, all the date/time types and `GUID`. All were implemented and unit tested; what was missing was parity evidence against msodbcsql. That matters because parity is what has caught the real divergences in this work — the `money` unsigned flag, the indicator-only binding, the zero-length probe — none of which unit tests can reach.

Filling those gaps immediately surfaced two genuine divergences.

**Decimal and money text rendering.** For a magnitude below one, msodbcsql renders `.5000` and `-.0001`; this driver rendered `0.5000` and `-0.0001`. The rule is unconditional — an exact zero renders `.0000` — and it was characterised across several values before anything changed. Money goes through its own formatter, so it was verified separately rather than assumed to follow the decimal path; it had the same bug.

Both are fixed at the ODBC text layer rather than in `mssql-tds`'s formatters. Those are shared with other consumers and there is nothing wrong with a leading zero in general; what has to match msodbcsql is *this driver's* rendered output.

**A NULL fetched through `SQLGetData` with no indicator** was also found here, raised by @Theekshna in review, auditing the diff against msodbcsql's own conversion suites. The indicator was written through a helper that no-ops on a null pointer, so the call returned `SQL_SUCCESS` and left the caller's buffer holding the previous row's value undetectably.

That one is **not** part of this PR's diff. [#382](https://github.com/microsoft/mssql-rs/pull/382) fixed the identical bug independently ([AB#47507](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47507)) and landed in `main` first, so the guard arrived here through the merge. Its tests are the more thorough of the two, so the duplicates this branch had added were removed. What remains from this side is the bound-path e2e test, still the only one asserting the SQLSTATE rather than just the return code there — the bug was the `SQLGetData` and bound paths drifting apart, so both are worth pinning.

## What's in the diff

- `src/api/get_data.rs` — the rendering fix (one helper, applied to the decimal, money and smallmoney arms)
- `tests/e2e/tests/datetime_types_test.cpp` — 11 tests, date/time family, both paths
- `tests/e2e/tests/numeric_types_test.cpp` — 17 tests, integer/bit/float/GUID/decimal, both paths, plus the bound-path `22002` case
- `tests/e2e/tests/text_money_types_test.cpp` — money and the character *source* types
- `tests/e2e/tests/driver_connect_test.cpp` — disconnect after a failed connect
- `docs/typed-columnar-fetch-plan.md` — the P4 verification outcome

Test values are chosen so a pass means something: `BIGINT` at both range ends (the lower one cannot be produced by negating anything), `0.1` for `REAL`/`FLOAT` since no binary float holds it exactly, a `GUID` with sixteen distinct bytes because `SQLGUID` is mixed-endian and uniform data would hide a straight `memcpy`, and fractional seconds asserted everywhere they exist since the wire carries them in 100 ns ticks.

## Deliberate omissions

- **Bound XML** — XML is a max/PLP type and bound PLP delivery is unimplemented ([AB#47361](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47361)). msodbcsql delivers it, so a bound test would sit here as a permanent divergence. Covered via `SQLGetData` only.
- **Bound `SQL_C_BINARY`** — delivery unimplemented ([AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239)).
- **`SQL_C_NUMERIC`** — input-side, and an explicit scope boundary of the plan.
- **Fetch-path error semantics** (`22003` / `01S07` / `07006`) — also from @Theekshna's review. Real gaps, but a coherent piece of work needing each cell confirmed against both drivers rather than ported on faith, so tracked as [AB#47678](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47678) rather than folded in here.

## Testing

- e2e parity suite: **30 tests, 0 divergences, 0 shared failures** against msodbcsql 18 on SQL Server 2025
- 38 e2e test cases added in total (11 date/time, 17 numeric, 9 text/money, 1 driver-connect); the parity figure above counts the run executed against both drivers
- full mssql-odbc unit suite green (1089 at this HEAD; the count moves with each merge from `main`); clippy and fmt clean
- Re-verified after merging `main`, which brought the read-buffer borrowing change (#343) through the value path this depends on

The rendering fix is verified by the parity suite specifically: the same tests failed against msodbcsql before it and pass after. It is also pinned in unit tests, so it does not need a live SQL Server plus an installed msodbcsql to catch a regression.

## Breaking changes

**A behaviour change, though not an API one.** Decimal, money and smallmoney values with magnitude below one now render without the leading zero when fetched as text. `mssql-python` parses this into `decimal.Decimal` and accepts either form, so it is unaffected; an application comparing the rendered string would see the difference. This makes us match msodbcsql, which is the point.

Worth a reviewer's eye: this affects any ODBC consumer that compares decimal or money text, not just `mssql-python`.

(The `SQLGetData` `22002` change is also observable — previously `SQL_SUCCESS`, now `SQL_ERROR` — but it ships with #382, not this PR.)

## Work items

[AB#46582](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46582) (P5). Also records the completed verification for [AB#46581](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46581) (P4), which is already closed — note that [#361](https://github.com/microsoft/mssql-rs/pull/361) has since implemented data-at-execution and exported `SQLParamData`/`SQLPutData`, superseding that audit's two open symbols; the plan doc records this.


mssql-odbc/docs/typed-columnar-fetch-plan.md
```diff
@@ -191,5 +290,5 @@ Both of those landed with the fetch rework in [#153](https://github.com/microsof
...
-| P4 — Exports & driver-load compat | 46581 | Partly done — `SQLColAttributeW` exported and advertised in P2; `SQLBindCol` and `SQLFetchScroll` in P3. Remaining: confirm the full symbol set loads under `mssql-python`. |
+| P4 — Exports & driver-load compat | 46581 | Implemented, no further code change required — `SQLColAttributeW` exported and advertised in P2; `SQLBindCol` and `SQLFetchScroll` in P3; `SQLSetDescFieldW` in [#345](https://github.com/microsoft/mssql-rs/pull/345). Symbol audit done on both platforms: 37 of the 39 symbols `ddbc_bindings.cpp` resolves are exported and advertised; `SQLParamData` and `SQLPutData` are absent (data-at-execution, a known scope boundary). Both Windows questions are settled — the DM does **not** gate loading on export completeness, and it can never reach these two because only the driver can enter the need-data state — and the `mssql-python` contradiction is resolved: its composite check fires but its pybind module initializer catches and logs it, after the resolvable pointers are already populated. **Decision: no stub exports.** See the P4 section for the evidence and for the one change in `mssql-python` that would reopen it. |
```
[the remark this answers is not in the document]
David-Engel: Good catch — the plan contradicted itself, since this branch adds the P5 coverage and records its results while the row still said "Not started".

Updated to "In progress" with what is actually covered: the fetch type map end to end on both paths for the date/time, integer, bit, float, GUID, decimal, money and character families, and the two divergences that coverage found (decimal and money rendering a leading zero below magnitude one, both fixed here).

I did not mark it Implemented, because two of the map's cells still have no coverage — bound `SQL_C_BINARY` and bound PLP/XML. Worth being precise about why: those are blocked on **unimplemented delivery** (AB#47239, AB#47361), not on missing tests, so writing tests would not close them. The row now says that explicitly, so the next person does not try. The remaining genuinely-test-shaped gap is the Windows `mssql-python` run, blocked on a pooling hang (AB#47510).

mssql-odbc/src/api/get_data.rs
```diff
@@ -1047,6 +1047,23 @@ fn xml_to_text(bytes: &[u8]) -> Result<String, TextError> {
     String::from_utf16(&units).map_err(|_| TextError::Malformed)
 }
 
+/// Drops the leading zero from a value whose magnitude is below one.
+///
+/// msodbcsql renders these as `.5000` / `-.0001`, not `0.5000` / `-0.0001`, and
+/// applications that compare the rendered text see the difference. It strips
+/// unconditionally, so an exact zero renders `.0000` too. Applied here rather
+/// than in `mssql-tds`'s formatters because this is the ODBC parity contract,
```
saurabh500: I checked this against the msodbcsql source and the behavior is exactly right, including the unconditional-zero case — but the comment asserts what msodbcsql does without saying where, which the repo instructions ask for on parity decisions. Citations, in case they're useful verbatim:

- `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp` → `numerictostring` (~8634). The digit loop breaks on `number.data[0] == 0 && i <= 0`, so it stops as soon as the scale is satisfied and never emits an integer-part digit for a sub-one value. `.5000` / `-.0001`, and zero at scale 4 → `.0000`.
- Same file → `BigintToChar` (~7357), called with `scale = 4` for money. Its loop continues while `value != 0 || cch <= scale`, emits the separator at `cch == scale`, then reverses — same result, via genuinely separate code. Good call verifying money independently rather than assuming it followed.

The more valuable half to record is **why `Real`/`Float` are deliberately left alone**: float→char goes through `DoubleToChar` (~7432, `_ecvt_s`-based), which has an explicit `if (i == 0) // Put in leading zero before decimal point` at ~7533. So msodbcsql really does render `0.5` for `FLOAT` and `.5000` for `DECIMAL` — the asymmetry is the parity-correct answer, not an oversight. Without that noted, the next reader sees an inconsistency and "fixes" it into a divergence.

mssql-odbc/src/api/get_data.rs
```diff
@@ -1047,6 +1047,23 @@ fn xml_to_text(bytes: &[u8]) -> Result<String, TextError> {
...
+/// not a general property of number formatting.
+fn strip_sub_one_leading_zero(s: String) -> String {
```
saurabh500, line 1153: This helper has no unit test. The only existing test that reaches the decimal arm is `get_data_decimal_renders_as_text` (line 1650), and it uses `"123.45"` — magnitude ≥ 1, so it takes the pass-through branch. Neither strip branch runs under `cargo btest`; all the verification lives in e2e tests that need a live SQL Server plus msodbcsql.

That leaves the repo's 85% diff-coverage gate looking at uncovered lines, and it means a refactor of `column_value_to_text` can silently drop the rule and stay green locally. A few lines would pin it:

```rust
#[test]
fn sub_one_values_render_without_the_leading_zero() {
    assert_eq!(strip_sub_one_leading_zero("0.5000".into()), ".5000");
    assert_eq!(strip_sub_one_leading_zero("-0.0001".into()), "-.0001");
    assert_eq!(strip_sub_one_leading_zero("0.0000".into()), ".0000");
    // At or above one, and scale 0, are untouched.
    assert_eq!(strip_sub_one_leading_zero("123.4500".into()), "123.4500");
    assert_eq!(strip_sub_one_leading_zero("-1.5000".into()), "-1.5000");
    assert_eq!(strip_sub_one_leading_zero("0".into()), "0");
}
```

Worth adding a `column_value_to_text` case over a real `Money`/`DecimalParts` too, so the wiring is covered and not just the string helper.

mssql-odbc/tests/e2e/tests/datetime_types_test.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+    EXPECT_EQ(2024, d.year);
+    EXPECT_EQ(2, d.month);
+    EXPECT_EQ(29, d.day);
+    EXPECT_EQ(static_cast<SQLLEN>(sizeof(SQL_DATE_STRUCT)), ind);
+    SQLCloseCursor(stmt_);
+}
+
+// ---------------------------------------------------------------------------
+// SQL_C_TYPE_TIME — truncates to whole seconds, so the fraction is the guard
```
saurabh500: The header says "truncates to whole seconds, so the fraction is the guard", but both tests below use `TIME(0)` — a value with no fraction to drop. The truncation path is never entered, so the guard the comment promises isn't there.

The branch this is describing is `fetch_convert.rs:709` (`if p.fraction_ns != 0 { ConvOk::Truncated }`), which surfaces as `SQL_SUCCESS_WITH_INFO` + `01S07`. `ASSERT_SQL_OK` uses `SQL_SUCCEEDED`, so it would pass either way — the interesting assertion is the return code and SQLSTATE, and it's also a good parity target since msodbcsql reports the same. Something like:

```cpp
TEST_F(DateTimeTypesLiveTest, Time7ToTimeTargetTruncatesTheFraction) {
    ASSERT_SQL_OK(ExecDirect("SELECT CAST('13:45:59.1234567' AS TIME(7)) AS c1"), SQL_HANDLE_STMT, stmt_);
    ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);

    SQL_TIME_STRUCT t{};
    SQLLEN ind = 0;
    SQLRETURN rc = SQLGetData(stmt_, 1, SQL_C_TYPE_TIME, &t, sizeof(t), &ind);
    EXPECT_EQ(SQL_SUCCESS_WITH_INFO, rc);
    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "01S07");
    EXPECT_EQ(59, t.second);
    SQLCloseCursor(stmt_);
}
```

Non-blocking — everything else in this file does assert the fraction where it exists.

saurabh500 (approved):
I am approving. There are minor recommendations. Author's discretion. I am OK with no-follow ups as well

mssql-odbc/src/api/get_data.rs
David-Engel: Thank you for verifying it against the source rather than just asking for a citation — and the `DoubleToChar` note is the part I would not have found on my own. Both are in the doc comment now: `numerictostring` for decimal and `BigintToChar` for money, with the point that they are genuinely separate code, which is why money got verified independently instead of assumed to follow.

The float asymmetry is recorded exactly as you framed it: `DoubleToChar`s explicit leading-zero branch means `0.5` for `FLOAT` and `.5000` for `DECIMAL` is the parity-correct answer, and removing the inconsistency would create a divergence.

Your comment also caught something you did not raise. Chasing it, I found **float-to-text was not covered at all** — the `REAL`/`FLOAT` tests used the binary C targets, so nothing exercised a float rendered as text, where we use Rusts `to_string` and msodbcsql uses `_ecvt_s`. So the asymmetry was not only undocumented, it was unverified. Added `FloatToCharKeepsItsLeadingZero` and `RealToCharKeepsItsLeadingZero`; both pass against both drivers.

mssql-odbc/src/api/get_data.rs
David-Engel: Agreed, and this was the one I most needed pointed out. Confirmed your reading: `get_data_decimal_renders_as_text` uses `123.45`, so it takes the pass-through branch and **neither strip branch executes under `cargo test`**.

The coverage gate matters less to me than the second thing you said. This rule was wrong twice before it landed — once for money, which renders through a different formatter, and once for exact zero, where I assumed a digit would be kept. Leaving it pinned only by a harness that needs a live SQL Server *and* msodbcsql installed means a refactor could drop it and stay green for almost everyone.

Added your test as written, plus the `-0` case. Also took the second suggestion: `sub_one_decimal_and_money_render_stripped` goes through `column_value_to_text` with real `DecimalParts` and `SqlMoney`, so the wiring is covered and not just the string helper. And `sub_one_float_keeps_its_leading_zero` pins the exception, so a future "consistency" change fails a unit test rather than only a parity run.

mssql-odbc/tests/e2e/tests/datetime_types_test.cpp
David-Engel: You are right, and it is worse than a stale comment — the comment described a guard, so it actively discourages anyone from adding the real one.

Both facts check out: `TIME(0)` has no fraction to drop, so `fetch_convert.rs`s `fraction_ns != 0` branch is never entered, and `ASSERT_SQL_OK` wraps `SQL_SUCCEEDED`, which `SQL_SUCCESS_WITH_INFO` satisfies — so even with a fractional source the existing shape would have passed regardless.

Added `Time7ToTimeTargetTruncatesTheFraction` essentially as you wrote it, asserting the return code and `01S07`, and corrected the section header to say what the tests actually do. It passes against both drivers, so the truncation reporting is at parity too.

saurabh500 (approved):

Theekshna:
Read the diff, then audited it against `msodbcsql`'s own conversion suites (`ODBCGEN/Tests/Cnvrsnpp`, `ODBCGEN/Tests/GetData`, `gtests/regression/ODBC_CTypeExtensions`). The type-map sweep looks right. One bug and a set of gaps below.

Line numbers are on `deb9e99`.

---

### 1. `SQLGetData` never raises `22002`, and the bound path does

`get_data.rs:276`, NULL column:

```rust
unsafe { write_if_some(strlen_or_ind_ptr, SQL_NULL_DATA) };
```

`write_if_some` no-ops on a null pointer, so `SQLGetData(col, SQL_C_SLONG, &v, sizeof(v), nullptr)` over a NULL column returns `SQL_SUCCESS` and leaves the caller's buffer holding the previous row's value with no way to detect it.

`SQLFetch` gets this right — `fetch_scroll.rs` maps `RowIssue::IndicatorRequired` to `22002`, pinned by `fetch_scroll_test.cpp:514`. msodbcsql applies it on the `SQLGetData` path too (`GetData.CPP:2251`, `FESQLGetData(..., NULL, ids22002)`).

`NullDatetimeReportsNullOnBothPaths` passes a real indicator on both legs, so it doesn't reach this.

```cpp
TEST_F(DateTimeTypesLiveTest, NullWithNoIndicatorIs22002) {
    ASSERT_SQL_OK(ExecDirect("SELECT CAST(NULL AS INT) AS c1"), SQL_HANDLE_STMT, stmt_);
    ASSERT_EQ(SQL_SUCCESS, SQLFetch(stmt_));
    SQLINTEGER v = 7;
    EXPECT_EQ(SQL_ERROR, SQLGetData(stmt_, 1, SQL_C_SLONG, &v, sizeof(v), nullptr));
    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "22002");
    EXPECT_EQ(7, v) << "the stale value is left visible, not overwritten";
}
```

### 2. No e2e coverage of `22003` / `01S07` / `07006` on the fetch path

`grep -E '22003|01S07|07006|22002' tests/e2e/tests/` hits only `execute_test.cpp` — the *parameter* direction. Every fetch-side range and truncation rule is unit-tested only, which is the situation the PR description argues against.

Four cells from `Cnvrsnpp.CPP` carry most of the signal:

| Source | Target | Expected | Cnvrsnpp |
| --- | --- | --- | --- |
| `CAST(128 AS INT)` | `SQL_C_STINYINT` / `SQL_C_UTINYINT` | `22003` / success | `:1702-03` |
| `CAST(2147483648 AS BIGINT)` | `SQL_C_SLONG` / `SQL_C_ULONG` | `22003` / success | `:2091-92` |
| `CAST(-1 AS INT)` | any unsigned target | `22003` | `:1576-82` |
| `CAST(1234.99 AS FLOAT)` | `SQL_C_SLONG` | `1234` + `01S07` | — (`fetch_convert.rs:1104`) |

Worth asserting the indicator is untouched on `22003`. `narrow!` propagates `?` before `write_fixed` (`fetch_convert.rs:65`), so this holds today — it's a regression guard, not a suspected divergence.

```cpp
TEST_F(NumericTypesLiveTest, BigintIntoSlongIsOutOfRange) {
    ASSERT_SQL_OK(ExecDirect("SELECT CAST(2147483648 AS BIGINT) AS c1"), SQL_HANDLE_STMT, stmt_);
    ASSERT_EQ(SQL_SUCCESS, SQLFetch(stmt_));
    SQLINTEGER v = 0; SQLLEN ind = 9876;
    EXPECT_EQ(SQL_ERROR, SQLGetData(stmt_, 1, SQL_C_SLONG, &v, sizeof(v), &ind));
    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "22003");
    EXPECT_EQ(9876, ind);
}
```

### 3. `-0.01` into an unsigned integer is `01S07`; into `SQL_C_BIT` it is `22003`

`Cnvrsnpp.CPP:2577-83`, verbatim:

```c
{TEXT("-0.01"),
 {CNVTBL_BIT   (RES_22003,   NULL) // this is treated specially!
  CNVTBL_UTINY (RES_01S07,   NULL)
  CNVTBL_USHRT (RES_01S07,   NULL)
  CNVTBL_ULONG (RES_01S07,   NULL)
```

Truncating toward zero gives `0`, in range for an unsigned target, so it is fractional truncation rather than a range error. `SQL_C_BIT` is carved out. `fetch_convert.rs:1421` unit-tests the negative-into-`BIT` half; the unsigned half is untested either way.

### 4. `datetime` with a non-zero time into `SQL_C_TYPE_DATE` is `01S07`, not silent

Implemented at `fetch_convert.rs:689`:

```rust
// Dropping a non-zero time component is a truncation.
    ConvOk::Truncated
}
```

No e2e on either path. `Cnvrsnpp.CPP:2808` asserts `RES_01S07` for `'1996-01-01 01:00:00.01'` into `DATE`, and `:2756-78` asserts plain success when the time is zero — the contrast is the test. This is the direct sibling of `Time7ToTimeTargetTruncatesTheFraction`.

### 5. A date/time column into a numeric C target is `07006`, not `22018`

`numeric_source` falls through to `_ => Err(ConvError::Restricted)` (`fetch_convert.rs:127`). `Cnvrsnpp.CPP:2676` — source `'1996-01-01'` under `TEST_TIME_TYPES` — asserts `RES_07006` for all ten numeric targets.

`07006` has no e2e coverage anywhere in the repo. The distinction from `22018` (covered at `get_data_test.cpp:559`) is the point: illegal pairing vs. bad text.

### 6. `uniqueidentifier` → `SQL_C_CHAR` text form is unpinned on both sides

The PR covers `SQL_C_GUID` on both paths but not the text rendering. Neither Cnvrsnpp nor GetData pins braces or case either, so this needs a live parity test rather than a ported one — same shape as the `.5000` finding: a rendering detail no unit test can reach and no legacy test asserts.

### 7. Minors

- **`datetimeoffset` → `SQL_C_SS_TIME2`** (time portion only). Supported at `fetch_convert.rs:715`, untested. `TCODBC_CTypeExtensions.cpp:3458`. The PR covers the `SS_TIMESTAMPOFFSET` direction but not this one.
- **`CAST('1753-01-01' AS DATETIME)`** — `datetime`'s floor. The wire value is a day offset that goes negative before 1900, exercising a signed path in `civil_from_days_since_0001` nothing currently reaches.
- **Exact-fit vs. one-short buffer through `SQLGetData`.** `fetch_scroll_test.cpp:253` covers the bound path only. `Cnvrsnpp.CPP:1129-45` pins `len` → `01004` and `len+1` → success, indicator reporting the untruncated length.
- **A `22003` as the row error inside a block fetch.** The mechanism is covered (`fetch_scroll_test.cpp:526`) but only via `22002`; a conversion failure on one row with the rest of the rowset still delivering is untested.

### 8. Question, not a test

`SQL_C_DEFAULT` (99) on `SQLGetData` returns `HYC00` — `is_typed_c_target` (`get_data.rs:939`) excludes it and the gate at `:245` rejects it. msodbcsql resolves it per SQL type for every type (`GetData.CPP:1167-1277`), and `resolve_default_c_type` already exists in `type_rules.rs`, wired only into the parameter path. `mssql-python` always requests explicit C types so nothing is blocked — but since this PR updates the plan doc, worth recording next to the other P1a divergences?

---

**Summary:** #1 is a bug. #2–#5 are the substantive gap — the PR proves the type map is wired up, but nothing on the fetch path proves the *error* semantics match msodbcsql, and those are three SQLSTATEs with zero e2e evidence.

Deliberately not raised: binary/hex, intervals, `SQL_C_NUMERIC`, and time-only → `SQL_C_TYPE_TIMESTAMP` (AB#47247) — out of scope or signed-off divergences. Also dropped a few candidates after checking the source: re-reading a fixed-length column is gated on column index rather than type (`get_data.rs:164`), so `BackwardColumnRejectedRereadIsNoData` already covers it; odd `BufferLength` for `SQL_C_WCHAR` can't split a code unit given `buf_elements` integer division; and `convert_typed_c` isn't passed `buffer_length` at all, so a zero-length probe of a fixed-length target matches msodbcsql by construction.

I read source, not test runs — I have not verified which of these currently pass.

David-Engel:
Thank you — this is the most useful review the PR has had, because it went at the diff from a direction I hadn't: auditing it against msodbcsql's own conversion suites rather than reading my tests. Finding #1 is a real bug, and it's now fixed in `a5c0872f`.

## #1 — confirmed, fixed

Your reading of the code was exactly right, and the empirical result is unambiguous. I added the test you wrote and ran it against both drivers before changing anything:

```
NullWithNoIndicatorViaGetDataIs22002   ours: SQL_SUCCESS (0)   msodbcsql: SQL_ERROR + 22002
NullWithNoIndicatorViaBoundFetchIs22002  ours: passes          msodbcsql: passes
```

Parity summary went to `28 parity, 1 divergence` — one leg failing, which is the signature of a driver divergence rather than a test bug.

The detail that convinced me it was an oversight rather than a decision: `fetch_scroll.rs:557` already carries this exact guard, with a comment spelling out the hazard — "leaving the slot untouched would read back as the previous row's value." The same reasoning applies verbatim to `SQLGetData`; only the guard was missing. The two paths had drifted.

One thing I checked rather than assumed, since it changes the shape of the fix: **character targets behave the same way.** I probed msodbcsql directly before implementing, because if it had returned success for `SQL_C_CHAR` I'd have written a narrower fix and created a new divergence:

```
NullWithNoIndicatorCharTargetIs22002   msodbcsql: SQL_ERROR + 22002
```

So the guard is unconditional, not fixed-width-only — an empty string is a value, so no target type can express NULL in band. That's now in the code comment as the reason.

Pinned in unit tests as well as e2e (`get_data_null_without_indicator_is_22002`, covering `SQL_C_SLONG`/`SQL_C_CHAR`/`SQL_C_WCHAR`, asserting the buffer is left untouched and the value stays resident for a retry). Same argument saurabh500 made earlier in this PR: behaviour that only a live SQL Server plus an installed msodbcsql can catch will regress silently for almost everyone. The e2e run is what proves that test is non-vacuous — it's asserting the return code that was actually wrong.

Your bound-path test overlaps `ARowErrorIsSQL_ERRORAtRowsetSizeOne` (`fetch_scroll_test.cpp:514`), which covers the same case but never asserts the SQLSTATE. I kept both: the pairing is the point, since these are the two paths that drifted.

## #2–#7 — valid, and now AB#47678

The characterisation is right and it's the more important half of the review: the PR proves the type map is *wired*, and almost nothing on the fetch path proves the *error semantics* match. I verified the grep myself, with one correction:

| SQLSTATE | Fetch-path e2e coverage |
| --- | --- |
| `22003` | none — only `execute_test.cpp`, the parameter direction, as you said |
| `01S07` | `datetime_types_test.cpp:157`, added earlier in this PR; numeric truncation still uncovered |
| `07006` | none anywhere in the repo |
| `22002` | now covered, via #1 |

So `01S07` isn't quite at zero, but `22003` on the fetch path and `07006` are, and the gap you're pointing at is real.

I'm not folding these into this PR. They're a coherent piece of work with a single theme, they need each cell confirmed against both drivers before being asserted rather than ported on faith, and this PR is approved with CI in flight. Filed as **AB#47678** under the P5 story, with your table, the `Cnvrsnpp` citations and the `fetch_convert.rs` line references carried over intact, including the note that the indicator-untouched-on-`22003` assertion is a regression guard rather than a suspected divergence.

Two of your minors I want to flag as better than they look:

- **`uniqueidentifier` → `SQL_C_CHAR`.** You noted neither `Cnvrsnpp` nor `GetData` pins braces or case. That's the same shape as the `.5000` finding this PR fixes — a rendering detail no unit test can reach and no legacy test asserts, which is exactly where the two drivers have already been caught disagreeing once.
- **`CAST('1753-01-01' AS DATETIME)`.** A negative day offset through `civil_from_days_since_0001` is a signed-arithmetic path nothing currently reaches. Cheap test, plausible bug.

## #8 — `SQL_C_DEFAULT`

Already tracked as **AB#47481**, though it's scoped to `SQLBindCol`; I've noted the `SQLGetData` half (`get_data.rs:939` / `:245`) belongs with it, since `resolve_default_c_type` exists and is wired only into the parameter path. Agreed nothing is blocked — `mssql-python` always requests explicit C types.

Noting for the record that you flagged these as read from source, not from test runs. That was worth saying, and #1 is the argument for it cutting both ways: the source reading was correct, and running it is what turned "should be `22002`" into a merge-blocking fix.

mssql-odbc/src/api/get_data.rs
```diff
@@ -273,6 +273,14 @@ fn write_captured_column(
     };
 
     if matches!(value, ColumnValues::Null) {
+        // No indicator means nowhere to report the NULL, and no target type can
+        // express it in band — an empty string is a value. Refuse rather than
+        // leave the caller's buffer reading back as the previous row. The value
+        // stays resident so a retry with an indicator still works.
+        if strlen_or_ind_ptr.is_null() {
+            post_diag(stmt_state, ERR_INDICATOR_REQUIRED);
+            return SQL_ERROR;
```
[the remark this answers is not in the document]
David-Engel: Good catch, and it was right at the time — the guard was pushed after the description was written, so the description had gone stale.

It has since resolved differently, which is worth recording rather than just fixing. [#382](https://github.com/microsoft/mssql-rs/pull/382) fixed the identical bug independently (AB#47507) and landed in `main` first, so after merging `main` the guard here arrives through the merge rather than from this branch. Vahid's tests are the more thorough of the two — the typed case asserts the retry with a real indicator still succeeds, and there is e2e for both a plain `SELECT NULL` and the PLP `VARCHAR(MAX)` path, which reaches `write_null` through a different wire marker — so I removed the duplicates this branch had added.

What I kept is the bound-path e2e test, still the only one asserting the SQLSTATE rather than just the return code on that path. The bug was `SQLGetData` and the bound path drifting apart, so pinning both is the point.

The description now says two driver fixes, scopes `get_data.rs` to the rendering fix, and notes under Breaking changes that the `22002` return-code change is observable but ships with #382 rather than here. Counts corrected too (17 e2e tests in that file, 883 unit tests, 30 parity binaries).

mssql-odbc/tests/e2e/tests/numeric_types_test.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    SQLRETURN ExecDirect(const std::string& sql) {
+        SqlTString s = ODBCTestUtils::ToSqlTStr(sql);
+        return SQLExecDirect(stmt_, const_cast<SQLTCHAR*>(s.c_str()), SQL_NTS);
+    }
+};
+
+// ---------------------------------------------------------------------------
+// Integers. TINYINT is unsigned 0..255 on the server, so the top of its range
+// is the interesting value for a signed C target.
```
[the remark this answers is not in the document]
David-Engel: You're right, and the comment inverted the actual reason the test passes.

`SQL_C_TINYINT` is sign-ambiguous, not signed. For a `tinyint` source `fetch_convert` copies the byte through with no range check, which is precisely how 200 survives — calling it "a signed C target" would predict `22003`, the opposite of what the test asserts. The signed narrowing to `i8`, and the `22003` that 200 would produce, belong to `SQL_C_STINYINT` and to any other source.

Reworded to state the copy-through rule and where the narrowing applies instead, pointing at `fetch_convert.rs`. Re-ran the suite after the change: 30 parity, 0 divergences.

saurabh500 (approved):
**Approve.**

Re-reviewed at `bffa1688`. Everything I raised across the earlier rounds is addressed, and the scope is now cleaner than when I started: dropping the duplicate 22002 work leaves this PR as one behaviour change plus the coverage that found it.

## What this PR is now

`strip_sub_one_leading_zero` in `get_data.rs`, applied to the decimal, numeric, money and smallmoney arms of `column_value_to_text`, three unit tests pinning it, and 38 new e2e tests across four files. That is the whole production surface.

Splitting the NULL/22002 fix out was the right call. #382 landed the identical check in `main` first, so it reaches this branch through the merge rather than from it — and the PR body now says so explicitly instead of leaving a reader to reconcile the diff against the description. Keeping the bound-path e2e test was also right: it is still the only one asserting the SQLSTATE rather than just the return code there, and the original bug was precisely the two paths drifting apart, so pinning both is what stops that recurring.

## Verification

I checked the parity claim against the msodbcsql source rather than taking it on the e2e run, since I cannot execute those here:

- `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp` → `numerictostring` (~8634). The digit loop breaks on `number.data[0] == 0 && i <= 0`, so it never emits an integer-part digit below magnitude one. Confirms `.5000`, `-.0001`, and the unconditional `.0000` for exact zero.
- Same file → `BigintToChar` (~7357), called with `scale = 4` for money — genuinely separate code reaching the same result, which is why verifying money independently rather than assuming it followed the decimal path mattered.
- Same file → `DoubleToChar` (~7432), which carries an explicit `if (i == 0) // Put in leading zero before decimal point` at ~7533. msodbcsql really does render `0.5` for `FLOAT`. **The decimal-vs-float asymmetry is the parity-correct answer, not an oversight** — I initially read it as an unfixed divergence and was wrong.

Both `column_value_to_text` call sites (`get_data.rs:315`, `fetch_scroll.rs:597`) route typed C targets through `convert_typed_c` before the text path, so stripped text only ever lands in `SQL_C_CHAR`/`SQL_C_WCHAR` buffers and is never re-parsed as a number. No caller re-reads the rendered form.

To check the tests actually pin the rule, I broke the mechanism rather than reverting the fix: removing the strip from the **money arm only** failed `sub_one_decimal_and_money_render_stripped` (`left: "0.5000"  right: ".5000"`) while the helper-only test stayed green. The two tests do independent work — the wiring test is not a restatement of the helper test.

Local run at this HEAD: 902 unit tests pass, `cargo fmt --check` and clippy clean.

**What I did not run:** the e2e suite — no live SQL Server or installed msodbcsql here. That half rests on your parity run plus the source reading above, not on anything I executed.

## Non-blocking

- `coverage-report` is red, but it is infrastructure, not the 85% gate: it polled 13 times over 75 minutes and timed out with `Available artifacts: none`, while the ADO *Merge Coverage* job passed in 2m36s. Worth a re-run before merge so the gate actually reports. Every other leg is green, including full ADO validation and the cross-repo `mssql-python` suite.
- Two counts in the description drift from what I measure at this HEAD: it says 883 unit tests (I get 902 — moves with each merge from `main`) and "30 tests" for the parity suite against 38 `TEST_F`s added. If "parity suite" means the subset actually executed against both drivers then 30 is right and ignore this; the per-file counts it gives (11 date/time, 17 numeric) are both exact.

Test values are well chosen — `BIGINT` at the lower bound specifically because it cannot be produced by negating anything, and a GUID with sixteen distinct bytes so a straight `memcpy` cannot hide behind uniform data. Those are the choices that make a passing e2e test mean something.

David-Engel:
Both non-blocking items from the approval review are now closed out.

> `coverage-report` is red, but it is infrastructure, not the 85% gate

Resolved on its own. That run timed out with `Available artifacts: none`; it passes on the current head, and every required check is green at `cb841fa3`.

Separately, the macOS leg failed once on build 170737 in the `Install and start Colima-based Docker` task — it hung during VM boot and exhausted all three 540s attempts. Infrastructure, not this diff: the same task succeeded in 1.8 min on another job *in the same build*, and in all 10 runs across the five preceding validation builds. Re-ran it and it passed on attempt 2. No pipeline change made.

> Two counts in the description drift from what I measure at this HEAD

Both corrected. You were exact on 38 — I counted the added `TEST_F`s against the merge base and get 38 (11 date/time, 17 numeric, 9 text/money, 1 driver-connect). The description now carries that breakdown alongside the parity figure, and states plainly that the two numbers count different things rather than leaving a reader to reconcile them.

The unit count is now 1089, not 883 or the 902 you measured — it has moved twice more since, with #382, #348/#385 and #383 arriving through merges. Rather than pin another number that goes stale on the next sync, the line now reads as a point-in-time measure and says so.

One note for whoever merges: the approval is recorded against `bffa1688`, and the head has since moved to `cb841fa3` through three merges from `main` (the #412 build fix, then #404, then #383). Nothing in the diff changed — those are merge commits plus one additive `CMakeLists.txt` resolution — but the branch protection appears to require approval of the most recent push, so it currently reads as `REVIEW_REQUIRED`.

saurabh500 (approved):

PR #332, opened 2026-08-19
Add repo-scoped code-review skill

saurabh500:
## Description

Closes https://github.com/microsoft/mssql-rs/issues/339

Adds a repo-scoped `code-review` skill at `.github/skills/code-review/SKILL.md`, ported from a personal Copilot PR-reviewer agent so the whole team gets the same review behavior instead of it living in one developer's `~/.copilot/agents/`.

The skill keeps the original structure (Process / What to Check / Output Format / Principles) and layers on mssql-rs-specific checks:

- License header on new `.rs` files
- Transport → IO → Token stream → Message → Client API layering
- `thiserror` + `TdsResult<T>`, no `unwrap`/`panic!` on network- or user-reachable paths
- Tokio blocking, `CancelHandle` cancellation, boxed async state
- `pub(crate)` visibility default and `Tds` type-name prefix
- `unsafe` justification, especially in `mssql-odbc` FFI
- Test placement, `mssql-mock-tds` preference, existing `conftest.py` fixtures
- `mssql-py-core` needing separate fmt/clippy (excluded from the workspace)
- 85% diff coverage to match CI, and `cargo bfmt` / `cargo bclippy` / `cargo btest` checklist claims
- The "no AI slop" rules from `copilot-instructions.md`

It also encodes the repo's merge ownership rule: reviewing is not merging.

## Checklist

- [x] `cargo bfmt` passes — n/a, no Rust changed
- [x] `cargo bclippy` passes — n/a, no Rust changed
- [x] `cargo btest` passes — n/a, no Rust changed
- [x] New/changed functionality has tests — n/a, docs-only
- [x] Public API changes are documented — n/a, no API change

PR #378, opened 2026-08-25
Harden the code-review skill: read all three review endpoints, replace facts that go stale, and record conventions mined from review history

### LINKED ISSUE #377: Harden the code-review skill: read all three review endpoints, replace facts that go stale, and record conventions mined from review history
### Problem statement

The `code-review` skill added in #332 tells a reviewer to read the PR description, but not the prior review discussion. On #360 that gap produced a misranked **Blocking** finding: an agent listed reviewer names with `gh pr view --json reviews`, saw that nine reviews existed, and inferred novelty from the silence. The finding it filed had already been raised by the Copilot reviewer and answered by the author.

The mechanism is worth recording, because it is not obvious. GitHub splits review discussion across three endpoints, and Copilot's low-confidence findings are **suppressed** — they appear only in `pulls/<N>/reviews[].body` inside a collapsed `<details>` block, never as an inline comment. Verified on #360:

| endpoint | contains the `mark_known_dead` finding |
| --- | --- |
| `pulls/360/comments` (inline threads) | no |
| `pulls/360/reviews` (review bodies) | yes — `copilot-pull-request-reviewer[bot]` |
| `issues/360/comments` (PR-level) | the author's rebuttal |

Reading only the inline threads makes both halves of that exchange invisible. Two separate reviewers hit this on the same PR.

Second, the file hardcodes point-in-time facts: a tracking issue number, a future-size budget constant, a coverage threshold, and a list of "known" failing tests. Each becomes actively misleading once it drifts. The known-failure list is the worst case — once those fixtures are generated in CI, it tells a reviewer to ignore a genuine regression.

Third, and larger than either: the conventions this repo has already paid to learn are not written down anywhere a reviewer will find them. A mining pass over **617 review threads** across the 34 merged PRs carrying ten or more, filtered to reviewer-finding-then-author-rebuttal, found two patterns that dominate the withdrawn findings.

**In `mssql-odbc`, msodbcsql is the contract and the ODBC spec is not.** Roughly fourteen instances. Findings argued from the spec or from internal consistency were overturned in both directions — a proposed `07006`→`07009` correction where msodbcsql reports nothing at all; "write the non-NULL length too" where it writes nothing; "gate these renames on ODBC 2.x" where `DoDD()` applies them unconditionally; "add the missing post-connect guard" where it deliberately has none. It cuts the other way too: `BufferLength = 0` *is* a length probe there, and raising that as a question found a real divergence. One finding was retracted by its own author after reading a validator without its caller, missing that `SQLBindParameter` folds `SQL_DOUBLE` to `SQL_FLOAT` before validating.

**A test that still passes when you break the thing it names guards nothing.** Nine instances. A limb-reassembly test survived changing its own `<< (i * 32)` to `<< (i * 16)` with all 539 tests green. Temporal tests were built in nanoseconds and asserted against a converter dividing by 1e9 — self-consistently wrong while the driver ran 100x off against a real server. Others passed on `Err(ConnectionClosed)` rather than the behavior in their name, or passed with *and* without the guard they were added for.

### Proposed solution

Process:

- Query all three review endpoints before drafting, and grep what was pulled for the symbol a finding concerns.
- Check the PR out into a dedicated worktree and diff against `git merge-base`, since PRs here are commonly stacked and often carry a merge from `main`.
- Baseline test failures against the merge base with the same invocation rather than trusting a written list — feature selection changes both the test count and which tests fail.

Durability:

- Replace hardcoded lookups with commands that re-derive them: search for the perf tracking issue rather than naming it, read the budget constant out of its test, read the coverage threshold off the PR's coverage comment.
- Add a standing note that every concrete number in the file is a dated observation, and that an observation contradicting the file should be reported so the file gets corrected.

Mined conventions, kept only where they recurred:

- Parity findings must cite an msodbcsql file and line, or be asked as a question. Read the caller, not just the table or validator.
- Mutate the constant, operator or condition a change is about and confirm the new test fails.
- Verify a bot's source and spec assertions rather than its confidence — three were withdrawn, and two "this issue also appears at line X" lists pointed at unrelated code.
- Check that a "breaking" API change is observable outside an unpublished pre-1.0 workspace.
- Treat a pre-existing crate-wide gap as a legitimate deferral rather than a defect this PR introduced.
- Pair every new ODBC setter arm with its getter.
- Correct the CHANGELOG guidance: still not required, but a behavior fix buried in a refactor or deletion PR belongs in the description.

Structure:

- Sections for adjudicating findings from a bot or another agent, and for the posting mechanics that otherwise fail silently (`gh pr review --comment` posts no inline comments; comments outside a diff hunk 422; the comments endpoint pages at 30).
- An explicit-authorization carve-out on confirm-before-posting so automated review runs are not blocked, with guardrails: `COMMENT` only, never `APPROVE`, no merging, and the unattended run disclosed in the body.
- Move the posting mechanics to a sibling file to keep `SKILL.md` skimmable — long skills get skimmed.

### Affected crate

Not applicable / Multiple

### Alternatives considered

Leaving the guidance in personal/user-scoped agent memory. That is where these lessons currently live, and it does not travel: an agent working from repo context alone has no access to it, which is precisely what happened on #360.

Reopening #332's issue (#339) instead of filing a new one. That issue was "add a skill file" and is already satisfied; this is a correction to the process steps inside it, with different evidence.

Mining every merged PR rather than the 34 with the most review discussion. The signal is concentrated — 23 of the last 100 merged PRs have no review threads at all — and the last 24 PRs of the pass raised instance counts without surfacing a new rule, which suggests the remaining tail would mostly cost tokens.

### Additional context

The original evidence came from two independent reviews of #360 run under the same model and the same repo context. The differences that mattered were access to prior review content and to accumulated review conventions — not model capability. The mining pass exists because the second of those two is the half that can be fixed in the repo.

David-Engel:
## Description

Corrects the `code-review` skill added in #332. Two fixes grounded in what went wrong on #360, plus a set of conventions mined from this repo's own review history.

### 1. The skill never told a reviewer to read the prior review discussion

Two independent reviews of #360 hit the same consequence. GitHub splits review discussion across three endpoints, and Copilot's low-confidence findings are *suppressed* — they appear only in `pulls/<N>/reviews[].body` inside a collapsed `<details>` block, never as an inline comment. Verified on that PR:

| endpoint | contains the `mark_known_dead` finding |
| --- | --- |
| `pulls/360/comments` (inline threads) | no |
| `pulls/360/reviews` (review bodies) | yes — `copilot-pull-request-reviewer[bot]` |
| `issues/360/comments` (PR-level) | the author's rebuttal |

`gh pr view --json reviews` reports nine reviews and none of their content. Inferring novelty from that silence is how an already-answered finding gets re-filed as **Blocking**. Step 3 now queries all three endpoints, saves the responses, names the suppressed-comment trap, and adds a cheaper guard: grep what you saved for the symbol a finding concerns before drafting it.

### 2. The file hardcoded facts that go stale on their own schedules

A tracking issue number, a future-size constant, a coverage threshold, and a list of "known" failing tests — each replaced by the command that re-derives it. The known-failure list mattered most: once those cert fixtures exist in CI, it tells a reviewer to ignore a genuine regression. Failures are now baselined against the merge base with a matching invocation, which generalizes to any pre-existing failure and catches the `--all-features`-vs-default discrepancy that made two reviewers report different baselines for the same tree.

A standing note says every concrete number below is a dated observation, and that an observation contradicting the file should be reported — against the skill, not in the reviewed PR's summary.

### 3. Conventions mined from this repo's review history

617 review threads across the 34 merged PRs carrying ten or more, filtered to reviewer-finding-then-author-rebuttal, read in full. Only what recurred was kept. Two patterns dominate:

**In `mssql-odbc`, msodbcsql is the contract — the ODBC spec is not.** The largest single source of withdrawn findings here. Reviews argued from the spec or from internal consistency and were overturned in both directions: a proposed `07006`→`07009` correction where msodbcsql reports nothing at all; "write the non-NULL length too" where it writes nothing; "gate these renames on ODBC 2.x" where `DoDD()` applies them unconditionally; "add the missing post-connect guard" where it deliberately has none. It cuts the other way too — `BufferLength = 0` *is* a length probe there, and raising that as a question found a real divergence. A sub-rule with its own casualty: read the caller, not just the validator, since msodbcsql normalizes on entry — one finding was retracted by its own author after missing that `SQLBindParameter` folds `SQL_DOUBLE` to `SQL_FLOAT` before validating.

**A test that still passes when you break the thing it names guards nothing.** Nine instances, and the cheapest check in the file: mutate the constant, operator or condition the change is about and confirm the test fails. A limb-reassembly test survived changing its own `<< (i * 32)` to `<< (i * 16)` with all 539 tests green. Temporal tests were built in nanoseconds and asserted against a converter dividing by 1e9 — self-consistently wrong while the driver ran 100x off against a real server. Others passed on `Err(ConnectionClosed)` rather than the behavior in their name, or passed with *and* without the guard they were added for.

Smaller recurring ones: verify a bot's source and spec assertions rather than its confidence (three withdrawn, plus two wrong "this also appears at line X" lists); check that a "breaking" change is observable outside an unpublished pre-1.0 workspace; treat a pre-existing crate-wide gap as a legitimate deferral; pair every new setter arm with its getter.

This also **corrects** the CHANGELOG guidance rather than just adding to it. Still not required — but the related finding that *is* real is a behavior fix buried in a refactor or deletion PR that the description never mentions.

### 4. From the review round (`e85ecfab`, `13a9b310`)

- **The skill is now GitHub-only.** It previously advertised Azure DevOps PRs while running hardcoded GitHub API calls for both, and the two providers number PRs independently — an ADO review would have silently read an unrelated GitHub PR sharing the number. `e85ecfab` added an ADO branch; `13a9b310` removed it once we confirmed PR workflow has moved fully to GitHub. Both commits are kept rather than squashed so the review trail stays honest. Work-item linkage and the coverage `paths-ignore` note remain — neither is PR workflow.
- **The coverage report is not always generated.** `pr-code-coverage.yml` mirrors the ADO path excludes, so a PR touching only `.github/**`, `docs/**` or the root markdown files never produces one, and fork PRs need a maintainer `/coverage`. This PR is itself such a PR. Both references are now conditional, with the exclusion list written down.
- `git fetch origin main` before deriving `BASE`, since fetching only the pull ref leaves a long-lived checkout comparing against a stale base.
- Step 5 picks the crate the diff touches instead of always running `mssql-tds`, which validates nothing in an odbc-, JS- or Python-only change.
- A hunk header bounds one hunk, not the diff — the old wording would have sent anchorable findings to the top-level body.

### Also included

- **Reviewing Alongside Other Reviewers** — verify the proposed fix as well as the diagnosis, and check whether a thread was already answered before escalating severity.
- **Automation carve-out** on the confirm-before-posting rule, so unattended review runs are not blocked. Keyed on explicit authorization rather than the agent's own read of the context, with guardrails: `COMMENT` only — never `APPROVE`, which can satisfy branch protection — no merging, and the unattended run disclosed in the body.
- Step 2 checks the PR out into a dedicated worktree and diffs against `git merge-base`, since PRs here are commonly stacked and often carry a merge from `main`.
- Posting mechanics move to a sibling `posting.md` to keep `SKILL.md` skimmable — it is judgment-free procedure, and long skills get skimmed.

## Verification

Documentation only — no code, no build impact. Every command added to the skill was run against #360 while writing it, including the three-endpoint table above. Each mined rule cites instances from the PRs it came from rather than a single anecdote. The review round's claims were each checked against the workflow files and the PR template before being acted on.

## Related Issues

Fixes #377

## Checklist

- [x] New/changed functionality has tests — n/a, documentation only

### DIFF
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: code-review
-description: Review a pull request, diff, or set of proposed changes in microsoft/mssql-rs — from a GitHub or Azure DevOps PR link, a PR number, or local staged/unstaged changes. Use whenever the user asks to review a PR, asks for feedback on a diff, or asks whether changes are ready to merge. Covers correctness, security, tests, readability, performance, API/breaking changes, and mssql-rs repo conventions.
+description: Review a pull request, diff, or set of proposed changes in microsoft/mssql-rs — from a GitHub PR link, a PR number, or local staged/unstaged changes. Use whenever the user asks to review a PR, asks for feedback on a diff, or asks whether changes are ready to merge. Covers correctness, security, tests, readability, performance, API/breaking changes, and mssql-rs repo conventions.
 ---
 
 # Pull Request Review
@@ -36,28 +36,17 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    git diff --stat $BASE..HEAD
    ```
 
-   For Azure DevOps repos, use the `repo_pull_request` tool with `action: get`.
 3. **Read what has already been said before writing anything.** PRs here routinely go
    through several rounds of author self-review plus a Copilot bot review, and many
    obvious findings are already raised, verified, and answered. Re-filing an answered
    thread as a new finding — especially at a higher severity — wastes the author's
    time and misranks the review.
 
-   **Establish the provider first.** GitHub and Azure DevOps number pull requests
-   independently, so running the GitHub recipe against an ADO PR number silently
-   reads an unrelated PR that happens to share the number. Branch on where the PR
-   actually lives.
-
-   *Azure DevOps:* read discussion with the `repo_pull_request_thread` tool
-   (`action: list`, then `list_comments` for a thread), and build results with the
-   `pipelines_build` tool. The rest of this step's reasoning applies unchanged; only
-   the transport differs.
-
-   *GitHub:* prior discussion is split across three endpoints and you need all three.
-   Listing reviewer *names* is not reading the reviews — a PR reporting nine reviews
-   tells you nothing about what any of them said, and treating that silence as
-   novelty is how an answered finding gets re-filed as blocking. Save the responses;
-   the symbol search below needs something to search.
+   Prior discussion is split across three endpoints and you need all three. Listing
+   reviewer *names* is not reading the reviews — a PR reporting nine reviews tells
+   you nothing about what any of them said, and treating that silence as novelty is
+   how an answered finding gets re-filed as blocking. Save the responses; the symbol
+   search below needs something to search.
 
    ```bash
    R=repos/microsoft/mssql-rs
@@ -103,7 +92,7 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    failures have been `certificate_validator` tests looking for
    `tests/test_certificates/*.pem`. Confirm rather than assume.)
 6. **Present the review in chat and wait for explicit human confirmation before
-   posting anything to GitHub or ADO.** Inline comments are drafted against
+   posting anything to GitHub.** Inline comments are drafted against
    `file:line`, not submitted, until they say so. Having the review fully written and
    the posting mechanics ready is not permission to post.
 
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -70,27 +70,37 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    settled it, far cheaper than re-reading the whole history.
 
    *If* an automated **Code Coverage Report** comment is present, read it before
-   writing any coverage finding (see "Verify Before You Claim"). Its absence is
-   usually expected rather than a CI failure — see the same section.
+   writing any coverage finding; its absence is usually expected rather than a CI
+   failure. Both under "Verify Before You Claim".
 4. Verify claims against the actual code — do not assume. Read surrounding code when
    a change's correctness depends on context: callers of changed functions,
    implementers of changed traits, and the layer above and below the change.
-5. Run the suite yourself rather than trusting the checklist. The runner is
-   `cargo nextest` / `cargo btest`, never `cargo test`. Pick the crate and targets the
-   diff actually touches — a `-p mssql-tds --lib` run validates nothing in a PR
-   confined to `mssql-odbc`, the JS or Python bindings, or the e2e suites.
+5. **Run tests to answer a question, not to re-collect a verdict.** CI runs the suite
+   and the merge is gated on it, so `gh pr checks` from step 3 already tells you pass
+   or fail — across platforms and cross-repo suites you cannot reproduce locally.
+   Rebuilding the workspace to learn the same thing costs minutes and returns less.
+
+   Build and run when you need something the CI result cannot give you:
+
+   - **Does this new test actually guard the change?** Mutate the constant, operator
+     or condition the PR changed and confirm the test fails. This is what catches
+     vacuous tests and is the highest-value reason to build locally.
+   - **Is this failure the PR's, or pre-existing?** Run the *same invocation* on
+     `$BASE` and compare failure sets; only the difference belongs in the review.
+   - **Is this coverage gap real?** Introduce the bug the missing test would catch and
+     show the suite still passes.
 
    ```bash
    cargo nextest run -p <affected-crate> --lib --no-fail-fast   # or `cargo btest`
    ```
 
-   Some tests fail on a clean tree because their fixtures aren't generated locally, so
-   never attribute a failure to the PR without a baseline — run the *same invocation*
-   on `$BASE` and compare failure sets. Only the difference belongs in the review.
-   Feature selection changes both the test count and which tests fail, so don't
-   compare a `--all-features` run against a default one. (Historically the pre-existing
-   failures have been `certificate_validator` tests looking for
-   `tests/test_certificates/*.pem`. Confirm rather than assume.)
+   The runner is `cargo nextest` / `cargo btest`, never `cargo test`. Pick the crate
+   and targets the diff touches — `-p mssql-tds --lib` validates nothing in a PR
+   confined to `mssql-odbc`, the JS or Python bindings, or the e2e suites. Feature
+   selection changes both the test count and which tests fail, so never compare a
+   `--all-features` run against a default one. Some tests fail on a clean tree because
+   their fixtures aren't generated locally (historically `certificate_validator`
+   looking for `tests/test_certificates/*.pem`) — which is why the baseline matters.
 6. **Present the review in chat and wait for explicit human confirmation before
    posting anything to GitHub.** Inline comments are drafted against
    `file:line`, not submitted, until they say so. Having the review fully written and
@@ -113,28 +123,20 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
 
 ## What to Check
 
-Evaluate each area. Skip areas that don't apply rather than padding the review.
+Correctness, security, tests, readability, performance and API surface all apply as
+usual — skip areas that don't apply rather than padding the review. What is worth
+stating here is only what is specific to this repo, or easy to get wrong in it:
 
-- **Correctness**: Logic bugs, off-by-one, null/empty/boundary cases, error handling,
-  race conditions, incorrect assumptions.
-- **Security**: OWASP Top 10 — injection, broken auth/access control, SSRF,
-  deserialization. Hardcoded secrets/credentials. Unvalidated input at trust
-  boundaries. Unsafe defaults.
-- **Tests**: New/changed behavior has tests. Tests assert real behavior (not
-  tautologies). Edge cases and failure paths covered. Flag untested risky changes.
-  When CI posts a diff-coverage comment, read it for the current target and whether
-  it is enforced rather than computing one locally (see "Verify Before You Claim").
-- **Readability & maintainability**: Clear naming, reasonable function size, no
-  needless complexity or duplication, comments explain *why* not *what*.
-- **Performance**: N+1 queries, lock contention and lock scope, memory copies,
-  unnecessary allocations in hot paths, blocking I/O on async paths, O(n²) where
-  avoidable, network round trips. Only raise when impact is plausible — and on the
-  row-decode path, only with a timing (see "Verify Before You Claim").
-- **API & breaking changes**: Public signatures, serialization formats, config, and
-  the FFI surface (`#[napi]`, `#[pyclass]`, `extern "C"`). Flag breaking changes and
-  whether they're versioned/documented.
-- **Repo conventions**: Match existing patterns, style, and structure in the
-  codebase. Respect `.github/copilot-instructions.md` and any `AGENTS.md`.
+- **Tests** are the highest-yield area. "Has a test" is not the question; "does the
+  test fail when the change is reverted" is.
+- **Performance** on the row-decode path needs a timing, never a byte table.
+- **API & breaking changes** include the FFI surface — `#[napi]`, `#[pyclass]`,
+  `extern "C"` — not just Rust signatures, and "breaking" needs a publication check.
+- **Repo conventions** live in `.github/copilot-instructions.md`,
+  `.github/instructions/*.md` and any `AGENTS.md`. Cite one before flagging a
+  convention.
+
+All four are expanded under "Verify Before You Claim", which is where the evidence is.
 
 ## mssql-rs Specifics
 
@@ -214,9 +216,8 @@ the rule exists and do not need re-deriving. Last reviewed 2026-08.
     `es-metadata.yml` never generates one; fork PRs need a maintainer to comment
     `/coverage`. Read the workflow's `paths-ignore` before reporting a missing report.
 - **Coverage gaps**: don't assert one, prove it. Introduce the bug the missing test
-  would catch, run `cargo nextest run -p mssql-tds --lib --no-fail-fast`, show the
-  suite still passes, then restore. Costs about two minutes and converts a concern
-  into a fact.
+  would catch, run the affected crate's suite, show it still passes, then restore.
+  One incremental build and a suite run converts a concern into a fact.
 - **"This is a breaking API change"**: check that anything outside the workspace could
   observe it. `mssql-tds` is unpublished and pre-1.0, and the sibling crates build
   `ClientContext` through `Default`/`From` rather than struct literals, so a field
@@ -301,17 +302,14 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
 
 ## Principles
 
-- Be specific and actionable; avoid vague praise or vague criticism.
 - **Question a departure from the reference drivers before auditing inside it.** When
-  a change diverges from `msodbcsql` / `SqlClient` / `mssql-jdbc` behavior, the divergence
-  is the first thing to examine — hardening a path that shouldn't exist is wasted review.
-  If a PR's own design doc reaches opposite conclusions in two places, that contradiction
-  *is* the finding.
-- If a change is correct, don't invent problems. An empty severity group means "none
-  found" — say so briefly.
+  a change diverges from `msodbcsql` / `SqlClient` / `mssql-jdbc` behavior, the
+  divergence is the first thing to examine — hardening a path that shouldn't exist is
+  wasted review. If a PR's own design doc reaches opposite conclusions in two places,
+  that contradiction *is* the finding.
 - Distinguish facts (verified in code) from concerns (worth checking). Don't state
   guesses as defects. Say what you ran and what you read.
-- Prefer the smallest correct fix over large refactors unless the PR's goal requires
-  more.
+- If a change is correct, don't invent problems. An empty severity group means "none
+  found" — say so briefly.
 - Reviewing is not merging. The PR author owns the merge — never merge someone
   else's PR.
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -10,8 +10,9 @@ affected code — do not critique pre-existing code outside the PR's scope.
 
 Every concrete number, constant, and known-failure list below is a dated observation,
 not a standing truth. Prefer the command that re-derives a fact over the value written
-here. If what you observe contradicts this file, trust the observation and say so in
-your summary so the file gets corrected.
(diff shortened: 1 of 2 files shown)


.github/skills/code-review/SKILL.md
```diff
@@ -8,25 +8,103 @@ description: Review a pull request, diff, or set of proposed changes in microsof
...
 
    For Azure DevOps repos, use the `repo_pull_request` tool with `action: get`.
-3. Verify claims against the actual code — do not assume. Read surrounding code when
+3. **Read what has already been said before writing anything.** PRs here routinely go
+   through several rounds of author self-review plus a Copilot bot review, and many
+   obvious findings are already raised, verified, and answered. Re-filing an answered
+   thread as a new finding — especially at a higher severity — wastes the author's
+   time and misranks the review.
+
+   Prior discussion is split across three endpoints and you need all three. Listing
```
David-Engel, line 45: **Blocking:** This mandatory discussion step does not branch by provider. The skill advertises GitHub and Azure DevOps PRs, but after the metadata-only ADO exception at line 37, it unconditionally runs hardcoded GitHub API calls and checks. For an ADO PR, this can silently inspect an unrelated GitHub PR with the same number and miss the actual ADO discussion. Add an ADO path using the pull-request thread and build/status tools, or narrow the skill's advertised scope to GitHub PRs.

.github/skills/code-review/SKILL.md
```diff
@@ -8,25 +8,103 @@ description: Review a pull request, diff, or set of proposed changes in microsof
...
    a change's correctness depends on context: callers of changed functions,
    implementers of changed traits, and the layer above and below the change.
-4. **Present the review in chat and wait for explicit human confirmation before posting anything to GitHub or ADO.** Inline comments are drafted against `file:line`, not submitted, until they say so.
-5. Ground yourself in reference code and public/private documentation/specifications. If you don't know the codebase, or which references to use, ask for context before reviewing.
-6. 
+5. Run the suite yourself rather than trusting the checklist. The runner is
+   `cargo nextest` / `cargo btest`, never `cargo test`.
+
+   ```bash
+   cargo nextest run -p mssql-tds --lib --no-fail-fast
```
David-Engel: **Suggestion:** This is only the `mssql-tds` library suite, but the instruction says to run the suite for every review. A PR confined to `mssql-odbc`, JS, Python, or integration targets could receive no relevant validation. Tell the reviewer to choose the affected crate and targets (or use `cargo btest` when appropriate), then retain the important rule below: run that exact invocation on the merge base when failures need attribution.

.github/skills/code-review/SKILL.md
```diff
@@ -81,6 +165,117 @@ Check these in addition to the general areas above.
...
+  reason the obvious fix was declined.
+- Reframe severity when the mechanism is real but the impact argument is not. Say
+  which part you kept and which part you corrected.
+
+## Posting the Review
+
+Only after explicit confirmation, or under the automation carve-out in step 6.
+The mechanics that otherwise fail silently — inline comments needing the API rather
+than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
+[posting.md](posting.md).
```
David-Engel: **Nit:** Use `[posting.md](./posting.md)` for the bundled resource. The VS Code agent-skill guidance requires `./` relative paths for skill resources, and this makes the progressive-load relationship explicit.

David-Engel (commented):
## Summary

The GitHub-specific hardening is well grounded: I validated the three discussion endpoints against #360 and confirmed the suppressed finding and author response appear where the PR describes. I found one blocking provider-routing gap plus two smaller workflow issues. The PR head is current, `git diff --check` passes, and all applicable checks are green; the ADO validation is skipped.

This is a `COMMENT` review; the findings are inline.

.github/skills/code-review/SKILL.md
David-Engel: Correct, and the failure mode is the bad part: the two providers number PRs independently, so the hardcoded `gh api` calls would have read an unrelated GitHub PR that happened to share the number — wrong data, silently, with the real discussion never fetched.

I first took your option 1 and added an ADO branch (`e85ecfab`), then took option 2 instead once we confirmed PR workflow has moved fully to GitHub (`13a9b310`). An unused provider path is worse than no path, because it looks maintained. The advertised scope, the `repo_pull_request` fetch note, and the branch are all gone; step 6 now names GitHub as the only posting target.

Two Azure DevOps references stay on purpose, neither of them PR workflow: step 1's work-item linkage, which `.github/PULL_REQUEST_TEMPLATE.md` still requires and a check enforces, and the coverage `paths-ignore` note, which is worth citing precisely because those excludes mirror the ADO validation build's policy.

.github/skills/code-review/SKILL.md
David-Engel: Agreed — `-p mssql-tds --lib` validates nothing in a PR confined to `mssql-odbc`, the JS or Python bindings, or the e2e suites. Step 5 now says to pick the crate and targets the diff actually touches, names `cargo btest` as the alternative, and keeps the merge-base attribution rule intact below it.

.github/skills/code-review/SKILL.md
David-Engel: Fixed — now `[posting.md](./posting.md)`.

.github/skills/code-review/SKILL.md
```diff
@@ -8,25 +8,103 @@ description: Review a pull request, diff, or set of proposed changes in microsof
 You are reviewing proposed changes. Review only what the diff changes plus directly
 affected code — do not critique pre-existing code outside the PR's scope.
 
+Every concrete number, constant, and known-failure list below is a dated observation,
+not a standing truth. Prefer the command that re-derives a fact over the value written
+here. If what you observe contradicts this file, trust the observation and say so in
+your summary so the file gets corrected.
```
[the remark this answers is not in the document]
David-Engel: Right, and it contradicted the scope rule six lines above it. Skill drift now goes to an issue or PR against the skill, explicitly not the summary of whatever PR is being reviewed.

One correction: the cited secondary locations are wrong. Line 66 is the symbol-grep instruction and line 186 is the `SQL_DOUBLE` example — both subjects of *other* findings in this same review, neither carrying this one. Only line 14 needed the change.

.github/skills/code-review/SKILL.md
```diff
@@ -8,25 +8,103 @@ description: Review a pull request, diff, or set of proposed changes in microsof
...
+   caller three layers up, the `#[cfg]` variant of a constant). Use a dedicated
+   worktree so the main worktree stays clean, and diff against the merge base rather
+   than trusting per-commit stats: PRs here are commonly stacked and often carry a
+   merge from `main`.
 
    ```bash
-   gh pr view <url-or-number> --json title,body,author,state,baseRefName,files,additions,deletions
-   gh pr diff <url-or-number>
+   gh pr view <url-or-number> --json title,body,author,state,isDraft,baseRefName,headRefName,additions,deletions,changedFiles,commits
+   git fetch origin pull/<N>/head:pr<N>-review
```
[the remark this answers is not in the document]
David-Engel: Correct — fetching only the pull ref leaves `BASE` derived from a stale remote-tracking ref, so a long-lived checkout can diff against a base that has already moved. Added `git fetch origin main` ahead of the merge-base. I had been doing this by hand and never put it in the recipe.

.github/skills/code-review/SKILL.md
```diff
@@ -8,25 +8,103 @@ description: Review a pull request, diff, or set of proposed changes in microsof
...
+   collapsed `<details>` block, never as an inline comment. An author's rebuttal to
+   one is usually a PR-level comment. Read only the inline threads and both halves of
+   that exchange are invisible.
+
+   Then, before drafting each finding, grep what you pulled for the symbol it
+   concerns. One search for a name like `mark_known_dead` surfaces the thread that
+   already settled it, far cheaper than re-reading the whole history.
+
+   The issue comments include the automated **Code Coverage Report** — read it before
+   writing any coverage finding (see "Verify Before You Claim").
```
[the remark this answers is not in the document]
David-Engel: Verified and fixed, and this PR is itself the proof: #378 touches only `.github/**`, so the coverage workflow was skipped and there are zero coverage comments on it. The instruction as written would have sent a reviewer hunting for missing CI output.

Both references are conditional now, and the exclusion list is written down: `paths-ignore` covers `docs/**`, `.github/**`, `README.md`, `CHANGELOG.md`, `CONTRIBUTING.md`, `SECURITY.md`, `LICENSE` and `es-metadata.yml`, plus the fork `/coverage` requirement. Lines 120 and 201 were both correctly identified.

.github/skills/code-review/posting.md
```diff
@@ -0,0 +1,40 @@
...
+      { "path": "mssql-tds/src/foo.rs", "line": 634, "side": "RIGHT", "body": "..." }
+    ]
+  }
+  ```
+
+- Inline comments must land on lines present in the diff or the API returns 422.
+  A hunk header `@@ -115,21 +143,206 @@` means the new-side range starts at 143;
+  lines before it are not in the diff even if they look adjacent in the file. Check
+  with `git diff $BASE..HEAD -- <file> | grep -n '^@@'`. Findings on unchanged lines
+  belong in the top-level body.
```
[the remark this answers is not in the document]
David-Engel: Right — a hunk header bounds that hunk, not the diff, so an earlier hunk may well cover an earlier new-side line. The wording would have sent anchorable findings to the top-level body. It now says to check the full `^@@` set rather than reasoning from the nearest hunk.

David-Engel:
Review round addressed in `e85ecfab` and `13a9b310`. Every finding verified against the code before acting on it, and all of them held.

**On the four suppressed comments**, which have no threads to reply to:

- *"The commands only print their output, so there is no combined result for the subsequent grep."* Correct — the advertised novelty guard had nothing to search. The three endpoint calls now redirect to `/tmp/pr<N>-{inline,reviews,toplevel}.txt` and the grep step points at those files.
- *"The causal wording is reversed"* on the `SQL_DOUBLE` example. Also correct, and my sentence was simply garbled: folding to `SQL_FLOAT` is what makes the `HYC00` arm **dead**, not why anyone believed it was live. Rewritten as "one finding called an `HYC00` arm reachable after reading the validator alone; `SQLBindParameter` folds `SQL_DOUBLE` to `SQL_FLOAT` before calling it, which is what makes that arm dead."
- The remaining two duplicated the coverage-report point, answered in the thread on line 71.

**Scope change worth flagging separately:** the skill is now GitHub-only. `e85ecfab` added an Azure DevOps branch as the fix for the provider-routing blocker; `13a9b310` removed it again after we confirmed PR workflow has moved fully to GitHub. Net effect on the file is the narrowed scope, not the branch — the two commits are kept rather than squashed so the review trail stays honest.

Two Azure DevOps references remain deliberately, neither of which is PR workflow: step 1's work-item linkage (still required by `.github/PULL_REQUEST_TEMPLATE.md`, enforced by a check) and the coverage `paths-ignore` note (those excludes exist to mirror the ADO validation build's policy).

.github/skills/code-review/SKILL.md
```diff
@@ -1,32 +1,115 @@
...
+drift separately — an issue or PR against this skill, not the summary of whatever PR
+you happen to be reviewing. Skill maintenance is not that author's problem.
+
 ## Process
 
 1. Read the PR title/description to understand intent. Flag if the description is
    missing or doesn't match the diff. This repo requires a linked GitHub issue or
    Azure DevOps work item — flag a PR that has neither.
-2. Review the full diff against the base branch before commenting.
+2. **Check the PR out locally.** A diff alone is not enough to review this codebase —
```
saurabh500, line 22: I performed an RCA and port-mortem with one of the agents, about improving code quality, and it had a similar opinion, "Don't look at changes in isolation but in caller's context"

.github/skills/code-review/SKILL.md
```diff
@@ -1,32 +1,115 @@
...
+   writing any coverage finding (see "Verify Before You Claim"). Its absence is
+   usually expected rather than a CI failure — see the same section.
+4. Verify claims against the actual code — do not assume. Read surrounding code when
    a change's correctness depends on context: callers of changed functions,
    implementers of changed traits, and the layer above and below the change.
-4. **Present the review in chat and wait for explicit human confirmation before posting anything to GitHub or ADO.** Inline comments are drafted against `file:line`, not submitted, until they say so.
-5. Ground yourself in reference code and public/private documentation/specifications. If you don't know the codebase, or which references to use, ask for context before reviewing.
-6. 
+5. Run the suite yourself rather than trusting the checklist. The runner is
+   `cargo nextest` / `cargo btest`, never `cargo test`. Pick the crate and targets the
```
saurabh500: Should we have an exclusion to say that don't run the tests, because the Pipeline anyhow will. 

Tests can be run locally for other reasons, e.g. To see if the test is really testing the changed code. 
I recommend qualifying this part with the condition that tests should be run when needed, since the PR anyhow will not get merged without the tests run. 

But highly encouraged to run the tests locally to validate changes and understand their intentions.

saurabh500 (approved):
Left a couple of comments.

.github/skills/code-review/SKILL.md
David-Engel: Agreed, and I checked the premise before acting on it rather than taking it as given: `main` carries two active branch rulesets, and this PR reports `mergeStateStatus=BLOCKED` while `mergeable=MERGEABLE`. So the merge really is gated and the pipeline really does run the suite.

The redundancy was worse than you flagged, in a way that argues your point harder. Step 3 already ends with `gh pr checks <N>` — so the pass/fail verdict is collected before step 5 is even reached, and CI collects it across platforms and cross-repo suites a reviewer cannot reproduce locally. Rebuilding the workspace was paying minutes to learn something already on the PR, and worse. It was also redundant a second way: the `Validation` bullet under `mssql-rs Specifics` already handles the checklist question, and handles it correctly, by comparing the checked box against CI rather than against a local run.

One place I went further than "qualify it", because I think "highly encouraged" would not have held. Three rules elsewhere in the file *require* a local run and would have been left dangling — the mutation check, the coverage-gap proof, and the `$BASE` baseline. So instead of softening the step I inverted it: run tests to answer a question, not to re-collect a verdict, with the three questions named explicitly. Your "to see if the test is really testing the changed code" is the first and highest-value one, and it now says so.

Your comment also caught a stale number I would not have found: the coverage-gap rule claimed the proof "costs about two minutes", which was measured on a warm target directory. Once step 5 stops implying you have already built, that estimate is wrong, so it is gone.

Both in `3bcfb502`. That commit also trims some generic review advice — separate concern, explained in the message, just so the diff size is not a surprise.

saurabh500 (approved):

PR #382, opened 2026-08-26
Fix SQLGetData returning uninitialized buffer for SELECT NULL

Vahid-b:
## Description

`SQLGetData` silently returned `SQL_SUCCESS` and left the caller's output buffer completely untouched when a column value was NULL and the caller passed a null `StrLen_or_IndPtr`. Per the ODBC spec this must be SQLSTATE `22002` ("Indicator variable required but not supplied") — mssql-odbc already implements this rule correctly for the bound-column path (`fetch_scroll.rs::deliver_bound` / `RowIssue::IndicatorRequired`), but the older unbound `SQLGetData` path (`get_data.rs::write_captured_column`) never got the same check.

mssql-python's `SQLGetData_wrap` fetches several fixed-width C types (`SQL_INTEGER`, `SQL_SMALLINT`, `SQL_TINYINT`, `SQL_BIT`, `SQL_REAL`, `SQL_DOUBLE`/`SQL_FLOAT`, `SQL_BIGINT`, `SQL_TYPE_DATE`, `SQL_TIMESTAMP`) with a null indicator and relies on the ODBC-mandated error to detect NULL (falling back to `None` only when `SQLGetData` fails). Without the check, `cursor.execute("SELECT NULL").fetchval()` returned whatever uninitialized memory happened to be on the caller's stack instead of `None` — a different value on every run.

Verified against the actual msodbcsql C++ reference driver source (`Sql/Ntdbms/sqlncli/odbc/sqlcdata.h`): `if (fIsNullData) { if (lpIndValue == NULL) { wError = IDS_22_002; ... } else *lpIndValue = SQL_NULL_DATA; }` — this driver was diverging from parity, not exceeding it.

Commit history is intentionally split so the CI signal is directly attributable:
1. **Add regression tests** — two new unit tests that fail against the pre-fix code, reproducing the bug at the unit level.
2. **Apply the fix** — mirrors `deliver_bound`'s existing `IndicatorRequired` check; turns the new tests (and the affected mssql-python cross-repo tests) green.

## Validation results (before-fix vs. after-fix pipeline runs)

Ran the `mssql-rs Pull request validation` pipeline (definition 2267) once per commit to isolate the fix's effect:

| | Before-fix ([build 169853](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=169853)) | After-fix ([build 169854](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=169854)) |
|---|---|---|
| `test_fetchval_null_values` | ❌ Failed | ✅ Passed |
| `test_gh610_execute_all_null_params` | ❌ Failed | ✅ Passed |
| `test_ddl_and_dml_rollback` | ❌ Failed | ✅ Passed |
| `mssql-python suite on mssql-odbc driver` (scoped run) | 1540/2076 passed | 1543/2076 passed |
| `Build Stage` (fmt + clippy + nextest, Windows/Linux/macOS/ARM) | n/a (build superseded/cancelled before completion) | ✅ succeeded |

Net effect: exactly **+3 passed / −3 failed**, identical total test count (2076) in both scoped runs — the fix's effect is precisely isolated to the 3 tests this bug caused, with no regressions elsewhere in the suite. The only non-green item in the after-fix build is the `mssql-python suite on mssql-odbc driver` job itself reporting `SucceededWithIssues`, which is by design while the driver is under development (unrelated pre-existing failures, not caused by this change).

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47507

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes
- [x] Public API changes are documented (behavioral fix only, no signature change)

### DIFF
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -273,6 +273,17 @@ fn write_captured_column(
     };
 
     if matches!(value, ColumnValues::Null) {
+        // ODBC spec: "If [StrLen_or_IndPtr] is a null pointer, no length or
+        // indicator value is returned. This returns an error when the data
+        // being fetched is NULL" (SQLSTATE 22002). There is nowhere to report
+        // the NULL, and leaving the target buffer untouched would hand the
+        // caller whatever was already there. Mirrors the identical check in
+        // fetch_scroll.rs::deliver_bound for bound columns. Leave the value
+        // resident so a retry with a real indicator can still succeed.
+        if strlen_or_ind_ptr.is_null() {
+            post_diag(stmt_state, ERR_INDICATOR_REQUIRED);
+            return SQL_ERROR;
+        }
         unsafe { write_if_some(strlen_or_ind_ptr, SQL_NULL_DATA) };
         // Only character targets get a terminator; a fixed-width target's
         // buffer is left untouched on NULL.
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -1539,6 +1539,82 @@ mod tests {
         assert_eq!(ind, std::mem::size_of::<i32>() as SqlLen);
     }
 
+    /// AB#47507 — SQLGetData must return SQLSTATE 22002 ("Indicator variable
+    /// required but not supplied") when the value is NULL and the caller
+    /// supplied no indicator, matching msodbcsql and the bound-column path
+    /// (fetch_scroll.rs::deliver_bound). Before this check existed,
+    /// SQLGetData silently returned SQL_SUCCESS and left the caller's output
+    /// buffer untouched. mssql-python's SQLGetData_wrap relies on the
+    /// ODBC-mandated error to detect NULL for fixed-width C types — it fetches
+    /// SQL_INTEGER et al. with a null indicator and falls back to `None` only
+    /// when SQLGetData fails — so the missing check meant callers read back
+    /// whatever was already on their stack as if it were the column's value.
+    #[test]
+    fn get_data_typed_null_without_indicator_reports_22002() {
+        let h = TestHandles::with_env_dbc_stmt();
+        stmt_with_captured(&h, ColumnValues::Null);
+
+        let mut out: i32 = -1_066_579_696; // poisoned sentinel, must survive untouched
+        let ret = unsafe {
+            sql_get_data(
+                h.stmt,
+                1,
+                SQL_C_SLONG,
+                (&mut out as *mut i32).cast(),
+                4,
+                std::ptr::null_mut(),
+            )
+        };
+        assert_eq!(ret, SQL_ERROR);
+        assert_eq!(out, -1_066_579_696, "NULL must not disturb the data slot");
+        {
+            let sh = unsafe { handle_from_raw::<StmtHandle>(h.stmt) };
+            let s = sh.inner.lock().unwrap();
+            assert_last_diag(&s.diag_records, ERR_INDICATOR_REQUIRED);
+        }
+
+        // The value was not consumed: a retry with a real indicator still
+        // reports the NULL correctly.
+        let mut ind: SqlLen = -99;
+        let ret = unsafe {
+            sql_get_data(
+                h.stmt,
+                1,
+                SQL_C_SLONG,
+                (&mut out as *mut i32).cast(),
+                4,
+                &mut ind,
+            )
+        };
+        assert_eq!(ret, SQL_SUCCESS);
+        assert_eq!(ind, SQL_NULL_DATA);
+    }
+
+    /// The same rule applies to character targets: a NULL with no indicator is
+    /// still 22002, even though a terminator could otherwise be written.
+    #[test]
+    fn get_data_char_null_without_indicator_reports_22002() {
+        let h = TestHandles::with_env_dbc_stmt();
+        stmt_with_captured(&h, ColumnValues::Null);
+
+        let mut buf = [b'X'; 8];
+        let ret = unsafe {
+            sql_get_data(
+                h.stmt,
+                1,
+                SQL_C_CHAR,
+                buf.as_mut_ptr() as SqlPointer,
+                buf.len() as SqlLen,
+                std::ptr::null_mut(),
+            )
+        };
+        assert_eq!(ret, SQL_ERROR);
+        assert_eq!(buf, [b'X'; 8], "NULL must not disturb the data slot");
+        let sh = unsafe { handle_from_raw::<StmtHandle>(h.stmt) };
+        let s = sh.inner.lock().unwrap();
+        assert_last_diag(&s.diag_records, ERR_INDICATOR_REQUIRED);
+    }
+
     #[test]
     fn get_data_typed_timestamp_target() {
         use crate::api::odbc_types::SqlTimestampStruct;
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -1550,16 +1550,9 @@ mod tests {
         assert_eq!(ind, std::mem::size_of::<i32>() as SqlLen);
     }
 
-    /// AB#47507 — SQLGetData must return SQLSTATE 22002 ("Indicator variable
-    /// required but not supplied") when the value is NULL and the caller
-    /// supplied no indicator, matching msodbcsql and the bound-column path
-    /// (fetch_scroll.rs::deliver_bound). Before this check existed,
-    /// SQLGetData silently returned SQL_SUCCESS and left the caller's output
-    /// buffer untouched. mssql-python's SQLGetData_wrap relies on the
-    /// ODBC-mandated error to detect NULL for fixed-width C types — it fetches
-    /// SQL_INTEGER et al. with a null indicator and falls back to `None` only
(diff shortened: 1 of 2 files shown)


mssql-odbc/src/api/get_data.rs
```diff
@@ -1539,6 +1550,82 @@ mod tests {
...
+    /// AB#47507 — SQLGetData must return SQLSTATE 22002 ("Indicator variable
+    /// required but not supplied") when the value is NULL and the caller
+    /// supplied no indicator, matching msodbcsql and the bound-column path
+    /// (fetch_scroll.rs::deliver_bound). Before this check existed,
+    /// SQLGetData silently returned SQL_SUCCESS and left the caller's output
+    /// buffer untouched. mssql-python's SQLGetData_wrap relies on the
+    /// ODBC-mandated error to detect NULL for fixed-width C types — it fetches
+    /// SQL_INTEGER et al. with a null indicator and falls back to `None` only
+    /// when SQLGetData fails — so the missing check meant callers read back
+    /// whatever was already on their stack as if it were the column's value.
```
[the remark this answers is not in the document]
Vahid-b: Good catch — trimmed to just the regression requirement in ebc5dace, dropped the duplicated PR description and the mssql-python implementation details that could go stale independently of this crate.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
```diff
@@ -1539,6 +1550,75 @@ mod tests {
         assert_eq!(ind, std::mem::size_of::<i32>() as SqlLen);
     }
 
+    /// AB#47507 regression: a NULL value with no indicator must be SQLSTATE
```
Theekshna, line 1553: **No e2e coverage for this fix**, and the sibling is already sitting in the file: `mssql-odbc/tests/e2e/tests/get_data_test.cpp:239 GetDataLiveTest.NullColumn` is this exact query with a real indicator. The null-indicator variant is missing.

This is more than a coverage box. Both new unit tests go through `stmt_with_captured`, which injects `ColumnValues::Null` straight into `last_captured` (`get_data.rs:1423`). They prove the guard fires, but nothing here proves that a real `SELECT NULL` *arrives* as `ColumnValues::Null` through the decoder — and that arrival leg is precisely what AB#47507 reports. Right now the only thing exercising it end to end is the cross-repo mssql-python suite, which reaches `None` via its own error branch and so cannot distinguish "fixed" from "failing differently".

Suggest two cases next to `NullColumn`:

1. `SELECT NULL` with `SQL_C_SLONG` and `StrLen_or_IndPtr = nullptr` — expect `SQL_ERROR` + 22002, with the target pre-poisoned to a sentinel and asserted unchanged.
2. `SELECT CAST(NULL AS varchar(max))` with the same null indicator. This is the PLP arrival path traced in the review summary; it reaches the identical guard but through `write_null` + `pause_after_column` rather than an inline value, and nothing in the repo currently covers it.

mssql-odbc/src/api/get_data.rs
```diff
@@ -1539,6 +1550,75 @@ mod tests {
...
+            )
+        };
+        assert_eq!(ret, SQL_SUCCESS);
+        assert_eq!(ind, SQL_NULL_DATA);
+    }
+
+    /// The same rule applies to character targets: a NULL with no indicator is
+    /// still 22002, even though a terminator could otherwise be written.
+    #[test]
+    fn get_data_char_null_without_indicator_reports_22002() {
```
Theekshna, line 1600: **Multi-column sequencing is untested.** Both new tests use a single column, so they cover the branch but not the state machine — and the state machine is where the risk actually lives for an error path that deliberately leaves the value resident.

The case that matters:

```sql
SELECT 1, NULL, 2
```

Column 1 succeeds, column 2 with a null indicator returns 22002, and column 3 must still resolve afterwards. I traced it and it does hold: the 22002 returns before `current_row_last_col` advances, so it stays at 1 and the forward-only gate admits column 3; `already_captured` is false; and `read_row_column(2)` (0-based) sees `target >= next_column_index`, so it is not `AlreadyConsumed` (`mssql-tds/src/connection/tds_client.rs:4879-4884`).

That is an invariant a test should own rather than a reviewer, especially since a future change to where `current_row_last_col` is updated would silently break "skip the NULL column and keep reading" with no failing test.

A `stmt_with_captured`-style unit test is enough; it doesn't need to be e2e.

Theekshna (commented):
Fix is in the right place, and I traced the two things that could have made it wrong.

**`write_captured_column` is the single NULL delivery point for `SQLGetData`.** I checked whether `stream_active_plp_chunk` needed its own guard: it does not. A NULL PLP value makes `PlpColumnStream::begin` return `None`, so the writer takes `write_null(col)` + `pause_after_column(...)` (`mssql-tds/src/io/token_stream.rs:536-539`), which yields `RowPaused`/`RowWritten` and never `PlpPaused`. `resume_to_column` therefore lands on `CursorColumn::Value { value: Null }` (`mssql-tds/src/connection/tds_client.rs:4900-4928`) and flows through `last_captured`. One guard genuinely covers both paths.

**Placement matches msodbcsql.** Putting the check inside the existing `ColumnValues::Null` arm and ahead of `binary_probe` mirrors `sqlcdata.h:517-529`, where the `fIsNullData` block precedes the variant zero-data logic at `:539`. Also confirms the PR title: `write_if_some` no-ops on a null pointer (`api/util.rs:27-31`), so the pre-fix behaviour was a silent `SQL_SUCCESS`, not a bad write.

---

**Worth recording in the PR body, because it is not obvious and I got it wrong on first read:** this repairs AB#47507 *indirectly*. mssql-python calls `SQLGetData(hStmt, i, SQL_C_LONG, &intValue, 0, NULL)` for `SQL_INTEGER` and appends `py::none()` on any failure (`ddbc_bindings.cpp:3626-3634`). The same no-indicator call shape is used for `SQL_SMALLINT`, `SQL_REAL`, `SQL_DOUBLE`, `SQL_BIGINT`, `SQL_TYPE_DATE`, `SQL_TYPE_TIMESTAMP`, `SQL_TINYINT` and `SQL_BIT`. Pre-fix it received `SQL_SUCCESS` plus an untouched slot and appended the garbage; now it receives 22002 and yields `None` — the same outcome msodbcsql produces. So the guard fixes that whole family of types, not just `int`.

That also settles AB#47507's open attribution: hypothesis 2 ("mssql-python doesn't consult the indicator, so the bug belongs elsewhere") is correct about the observation but wrong about the culprit. mssql-python doesn't need to consult it, because msodbcsql errors on that call shape. mssql-odbc was the divergent side.

Cross-repo evidence, since the check shows a misleading green (the job is `succeededWithIssues`, not a clean suite):

| Build | Ref | `fetchval_null_values` | `gh610_execute_all_null_params` | `ddl_and_dml_rollback` |
|---|---|---|---|---|
| 169833 | PR #375 | FAILED | FAILED | FAILED |
| 169854 | PR #382 | PASSED | PASSED | PASSED |
| 169866 | PR #382 (`ebc5dace`) | PASSED | PASSED | PASSED |
| 169885 | PR #370 | FAILED | FAILED | FAILED |

The A-B-A ordering rules out mssql-python drift — the newest build is one of the failures. Whole-suite delta 169866 vs 169885 is exactly +3 passed / -3 failed, so nothing else moved.

Two test-coverage gaps below. Neither blocks the fix itself.

*Caveat on my end: I could not run the unit tests locally — a fresh worktree hits an msrustup 1.95 `simd_xor` codegen ICE in `crc32fast`. Environmental; CI is green.*

mssql-odbc/src/api/get_data.rs
Vahid-b: Added both e2e cases in a5c330dc:

- `NullColumnWithoutIndicatorReturns22002` — the exact `SELECT NULL` bare literal from AB#47507, `SQL_C_SLONG`, no indicator, buffer pre-poisoned and asserted unchanged after the 22002.
- `PlpNullColumnWithoutIndicatorReturns22002` — `CAST(NULL AS VARCHAR(MAX))` with no indicator, which decodes via the distinct `SQL_PLP_NULL` wire marker and `RowWriter::write_null` rather than the fixed/var-length NULL prefix `NullColumn` covers, confirming both arrival legs converge on the same guard.

Verified locally that both compile and register with GoogleTest on MSVC (VS 2022 CMake build); they'll run for real against the CI SQL Server on this push.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
Vahid-b: Added `get_data_null_without_indicator_does_not_block_later_columns_in_the_row` in a5c330dc — a `stmt_with_captured`-style unit test driving `SELECT 1, NULL, 2` purely through `StmtState` bookkeeping (manually re-seeding `last_captured` between calls the way `resume_row_to_column` would after each real decode), matching your trace: it asserts `current_row_last_col` stays at 1 after column 2's 22002, and that column 3 still resolves afterward rather than being treated as an already-consumed re-request. Didn't need a live decoder since this pins the accounting invariant, not the TDS decode logic you already traced correct.

> [!NOTE]

Theekshna (approved):

shiwanigupta0809 (approved):

PR #383, opened 2026-08-26
Release connection busy claim when a fetch exhausts its result set

Vahid-b:
## Description

`mssql-odbc`'s connection-busy gate (`DbcState::active_stmt`) was cursor-lifetime based: it was set on a successful execute and only cleared by `SQLCloseCursor`, `SQLFreeStmt(SQL_CLOSE)`, `SQLMoreResults` batch-exhaustion, or disconnect. `msodbcsql`'s equivalent gate (`FIsBusyReadingData` / `CheckBusy`) is wire-state based — it releases the connection once the batch is drained off the wire, even while the ODBC-level cursor stays open. Because of that gap, a single `fetchone()` on a one-row result never advanced far enough to observe `SQL_NO_DATA`, so the connection stayed "busy" until the cursor was explicitly closed — a second statement on the same connection got `HY000 "Connection is busy with results for another command"` even though there was nothing left to read.

### Fix

- Added `TdsClient::peek_past_current_row` (`mssql-tds`): once every column of the current row has been read, looks one token further. If the wire is already at the terminating `DONE`, that's reported immediately; if another row follows, it is parked (via a new `row_already_positioned` flag) so the next fetch consumes it directly without re-reading the wire — total round trips across a whole fetch loop are unchanged, this just discovers "no more data" one call earlier, matching when the wire actually goes idle.
- `mssql-odbc`'s `SQLFetch`/`SQLFetchScroll` and `SQLGetData` call this once a row's columns are fully consumed (all bound columns read, or `SQLGetData` reaches the last column) and release `active_stmt` immediately when it comes back empty **and the whole batch is done** (see below — not just the current result set).
- Added `StmtState::result_set_exhausted`, set alongside the release so a later `SQLFetch` on the now-idle-but-still-open cursor reports `SQL_NO_DATA` without needing the connection *at all* — even while a different statement is actively using it. `STMT_STATE_CURSOR_OPEN` is untouched, so `SQLMoreResults`/`SQLCloseCursor` behave exactly as before.

### Fixed during review (releasing `active_stmt` early exposed several latent assumptions elsewhere that it was always `Some(self)` while a cursor was open — that coupling was previously unconditional; this PR intentionally breaks it, so every place that relied on it needed an explicit check)

[...]

PR #384, opened 2026-08-26
[py-core] Async Fetch Infrastructure and Execute refactoring

subrata-ms:
## Description
This pull request introduces asynchronous row fetching capabilities to the MSSQL Python core library, including new internal state management for fetch operations, error handling improvements, and a comprehensive test suite for the new functionality. The changes enhance the ability to fetch rows asynchronously, ensure proper synchronization and error reporting, and add robust testing infrastructure for both real and mock connections.

**Asynchronous Row Fetching Implementation:**

- Added a new module `async_fetch.rs` implementing the `fetchone` method for `PyAsyncCursor`, including fetch state tracking, error mapping, and resource management for safe concurrent access. This enables Python users to asynchronously fetch rows from a cursor. [[1]](diffhunk://#diff-6efdd76294f1dfe8ce03399d24e06da1e5e1e8e851ed540c6577009e7a630b4cR1-R213) [[2]](diffhunk://#diff-59116afdbd4d7619c126efc429f65543c6b801d90e3d6955e8f60c3ab3a067fdR12-R13)

**Session State and Error Handling Enhancements:**

- Extended `AsyncConnectionState` to support fetch claims, fetch state transitions, and error types (e.g., `NoResultSet`), along with new guards for operation safety and proper release or poisoning of session state on interruption. [[1]](diffhunk://#diff-e798bc505f26fcf5f6aa29f81f738d9a922ecd268c85732766fe27ffa986eeb3R27-R39) [[2]](diffhunk://#diff-e798bc505f26fcf5f6aa29f81f738d9a922ecd268c85732766fe27ffa986eeb3L46-R54) [[3]](diffhunk://#diff-e798bc505f26fcf5f6aa29f81f738d9a922ecd268c85732766fe27ffa986eeb3R85-R124) [[4]](diffhunk://#diff-e798bc505f26fcf5f6aa29f81f738d9a922ecd268c85732766fe27ffa986eeb3R209-R229) [[5]](diffhunk://#diff-e798bc505f26fcf5f6aa29f81f738d9a922ecd268c85732766fe27ffa986eeb3R281-R311)
- Improved error reporting by mapping the new `ClaimError::NoResultSet` to a Python runtime error message.

**Shared Type Conversion Fixes**

- Corrected MONEY and SMALLMONEY decoding so negative sub-unit values and minimum-range values preserve their sign without integer overflow.
- Corrected DATETIMEOFFSET decoding to convert the wire UTC instant back to its stored offset and local wall-clock value.
- These fixes apply to both synchronous and asynchronous cursor fetching because both paths share the same row conversion logic.
- Verified compatibility with existing mssql-python expectations and added boundary coverage for affected values.

**Testing Infrastructure and Coverage:**

- Added a new test module `test_async_fetch.py` with extensive tests for the `fetchone` method, covering normal operation, error cases, result exhaustion, and concurrent access scenarios.
- Enhanced `conftest.py` to provide a `mock_client_context` fixture using a mock TDS server, allowing for isolated and controlled testing of asynchronous fetch functionality. [[1]](diffhunk://#diff-57f9fe1b4b8dd8d11287e081c5e25c92613cb658510dfe0d7feb683cd4171ed1R18-R22) [[2]](diffhunk://#diff-57f9fe1b4b8dd8d11287e081c5e25c92613cb658510dfe0d7feb683cd4171ed1R141-R157)

**Internal Testing and Validation:**

- Added Rust unit tests to `async_session.rs` to verify fetch ownership, concurrent access rejection, result exhaustion handling, and restoration of fetch ownership after failed awaitable construction.

These changes collectively add robust, well-tested asynchronous row fetching to the library, improving both functionality and reliability for users.

## Related Issues

ADO work item: https://dev.azure.com/SqlClientDrivers/mssql-python/_workitems/edit/47196, https://dev.azure.com/SqlClientDrivers/mssql-python/_workitems/edit/47586, https://dev.azure.com/SqlClientDrivers/mssql-python/_workitems/edit/47197
## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


David-Engel (commented):
## Summary

Adds `PyAsyncCursor.fetchone()` plus the session-ownership plumbing for it, and splits the existing async `execute` out of `async_cursor.rs` into `async_execute.rs`. The state machine is well factored — claim / guard / `Drop`-poison is consistent across execute, fetch and close — and the test split lost nothing: all 52 pre-existing test names in `test_async_cursor.py` are present across the three resulting files (`comm -13` on the sorted name lists is empty). CI is green and the coverage bot reports 97% diff coverage.

To verify the fetch semantics I read `TdsClient::on_rows`, `has_open_batch`, `next_row_into` (its end-of-set reads are idempotent via `current_result_set_has_been_read_till_end`) and `command_is_busy` in `mssql-tds`, and checked PR #383 for the ODBC-side treatment of the same ownership question.

**No blocking findings.** Findings 1-3 and the nits are inline; finding 4 is below because those lines are outside the diff hunks.

### Suggestion 4 — stale `#[allow(dead_code)]` and "upcoming"/"future" comments on `PyAsyncCursor`

`async_cursor.rs` lines 151, 156, 159, 162 still carry `#[allow(dead_code)]`, and line 151 still says "Consumed by upcoming async execute/fetch/close APIs." All four fields are now consumed by `cleanup()`, `fetch_resources()` and `execute_resources()`. Verified: with all four attributes deleted, `cargo check --lib` in `mssql-py-core` compiles warning-free. The doc comments at lines 155 ("Shared with the parent connection for future execute-time transaction handling") and 165 ("Applied by the future `execute` path") describe shipped behavior now and read as stale.

### Nit — the branch is 8 commits behind `main`, and `main` moved the same file

`main` changed `RowWriter::write_string` / `write_bytes` in `row_writer.rs` (#343), which this PR also edits. Merging `main` before landing gets the cross-repo `mssql-python` job to validate the combination that will actually ship.

### Notes, not findings

- The refactor removes a latent panic. `execute` previously held `slf.borrow(py)` across `bind_parameters`, which can run arbitrary Python and release the GIL, letting an in-flight execute future's `slf.borrow_mut(py)` panic. `execute_resources()` now drops the borrow before binding.
- I chased the `future_into_py` failure rollback in `execute` when the replaced claim had `drain_previous == true`: `release_operation` leaves the session looking idle while a batch is still undrained. It is benign — `TdsClient::execute` gates on `command_is_busy()` and returns a usage error, and `settle(has_open_batch = true)` then marks the session broken. Not raising it.
- `row_writer::to_py_tuple` drops an intermediate `Vec`; `ExactSizeIterator` is preserved so `PyTuple::new` keeps its sized fast path.
- The checklist's "Public API changes are documented" box is unchecked, but `mssql-py-core` has no `docs/` plan doc and the earlier async-execute work did not add one, so there is no convention here to point at. The rustdoc on `fetchone` is the documentation. Not a finding — noting it so the unchecked box does not look unexplained.

mssql-py-core/src/async_fetch.rs
```diff
@@ -0,0 +1,267 @@
...
+        };
+
+        match result {
+            Ok(writer) => {
+                fetch_guard.complete(writer.is_some(), has_open_batch);
+                fetch_state.set(if writer.is_some() {
+                    FetchStatus::Ready
+                } else {
+                    FetchStatus::Exhausted
+                });
```
David-Engel: **Suggestion.** `fetch_guard.complete()` publishes the new session state (claim released, or phase back to `Fetching`) *before* `fetch_state` is updated, leaving a window where the two disagree.

A `fetchone()` issued from the event-loop thread inside that window still reads `Ready`, so it skips the exhausted fast path at line 154, calls `claim_fetch`, finds `active_operation == None`, and surfaces `RuntimeError("No active result set")` where `None` is correct. The window is small, but the two threads are genuinely distinct — the awaitable is constructed on the event-loop thread while the future runs on the Tokio runtime.

Swapping the two statements closes it at zero cost, since `fetch_state` is only ever read by the fast path:

```rust
let has_row = writer.is_some();
fetch_state.set(if has_row { FetchStatus::Ready } else { FetchStatus::Exhausted });
fetch_guard.complete(has_row, has_open_batch);
```

That also drops the third `writer.is_some()` call.

mssql-py-core/src/async_session.rs
```diff
@@ -210,6 +287,37 @@ impl AsyncConnectionState {
...
+        };
+        if active.operation_id != operation_id || active.phase != OperationPhase::FetchingRow {
+            return;
+        }
+
+        if has_row || has_open_batch {
+            active.phase = OperationPhase::Fetching;
+        } else {
+            state.active_operation = None;
+        }
```
David-Engel, line 308: **Suggestion — worth a tracked follow-up rather than a change here.**

Ownership is released only when `fetchone()` is called *again* and returns `None`; reading the last row leaves `has_open_batch == true`, so the phase goes back to `Fetching` and the connection stays claimed. `test_fetchone_returns_rows_and_releases_final_exhaustion` encodes exactly that — `other.execute(...)` only succeeds after the extra `None` fetch.

That is the same shape PR #383 is currently fixing one layer down, where it is a reported customer bug: msodbcsql's `FIsBusyReadingData` releases the connection when the *wire* goes idle, not when the app asks once more, and a single `fetchone()` on a one-row result was leaving `mssql-odbc` busy until the cursor was explicitly closed. #383 adds `TdsClient::peek_past_current_row` in `mssql-tds` — shared code this path can call — to discover end-of-data one call earlier without extra round trips.

Nothing here is wrong in isolation, and I would not hold the PR for it. But the release point is a user-visible contract, and if the async Python front end settles on "one extra fetch" while ODBC moves to "when the wire goes idle", the two diverge permanently and the async side inherits the bug #383 just spent a pipeline run reproducing. Worth a work item to adopt `peek_past_current_row` here once #383 lands, and worth deciding the merge order deliberately.

mssql-py-core/src/cursor.rs
```diff
@@ -895,9 +895,9 @@ impl PyCoreCursor {
                 let money_val = lsb_in_i64 | ((m.msb_part as i64) << 32);
 
                 // Format as decimal string with 4 decimal places
-                let integer_part = money_val / 10000;
-                let fractional_part = (money_val % 10000).abs();
-                let decimal_str = format!("{}.{:04}", integer_part, fractional_part);
+                let sign = if money_val < 0 { "-" } else { "" };
+                let magnitude = money_val.unsigned_abs();
+                let decimal_str = format!("{sign}{}.{:04}", magnitude / 10000, magnitude % 10000);
```
David-Engel, line 900: **Suggestion — the fix is right; the concern is that it is invisible from the PR description.**

The change is correct and worth having. The old code lost the sign whenever `|value| < 1`, because `-5000 / 10000` truncates to `0` and the sign lived only in the integer part — `-0.5000` rendered as `0.5000`. It also called `.abs()` on the remainder, and `i64::MIN.abs()` panics in debug; `unsigned_abs()` handles the `money` minimum (`-922337203685477.5808`) correctly, and the new boundary test covers it.

But this is a behavior change on the **synchronous** `PyCoreCursor` path that ships to `mssql-python`, and none of the three linked ADO items are about it. The description covers only async fetch, session state and tests, so a reader of the description would not know the sync money rendering changed. Same for the `datetimeoffset` change below.

Ask: name both fixes in the description (or split them into their own PR), and confirm nothing downstream in `mssql-python` is pinned to the old output. Existing `test_bulkcopy_datetimeoffset.py` assertions only check `utcoffset()` and Y/M/D, so they would not have caught either direction.

mssql-py-core/src/cursor.rs
```diff
@@ -1083,16 +1083,19 @@ impl PyCoreCursor {
...
                                     day,
                                     hour,
                                     minute,
                                     second,
                                     microsecond,
-                                    tz_obj,
+                                    utc,
                                 ))
+                                && let Ok(datetime_obj) =
+                                    utc_datetime.call_method1("astimezone", (tz_obj,))
```
David-Engel: **Suggestion.** Building in UTC and converting is the correct read — MS-TDS stores the `DATETIMEOFFSET` date/time portion in UTC, and the decoder passes the wire value through untouched, so attaching `tz_obj` directly was reporting the UTC wall clock with a non-UTC label. The new `test_fetchone_decodes_wide_sql_type_matrix` case pins it against a live server, which is the right proof.

One consequence worth knowing about: `astimezone` can raise `OverflowError` at the range edges (a UTC value near year 1 or 9999 combined with a large offset), and this `if let Ok(..)` chain then falls through to `format!("{:?}", dto)` at the bottom — the column comes back as a Rust debug string rather than a datetime or an error. The silent fallback predates this PR, but `astimezone` widens the set of inputs that can reach it. Not asking you to fix the fallback here; flagging so it is a known edge rather than a surprise.

mssql-py-core/src/async_connection.rs
```diff
@@ -65,6 +65,20 @@ fn map_claim_error(error: ClaimError) -> PyErr {
         ClaimError::Closed => PyRuntimeError::new_err("Connection is closed"),
         ClaimError::Broken => PyRuntimeError::new_err("Connection is broken"),
         ClaimError::Busy => PyRuntimeError::new_err("Connection is busy with another operation"),
+        ClaimError::NoResultSet => PyRuntimeError::new_err("No active result set"),
```
David-Engel, line 68: **Nit.** `claim_connection_operation` never returns `ClaimError::NoResultSet` — only `claim_fetch` produces it, and that is reached through `async_cursor::map_claim_error`. So this arm is unreachable from the connection-level claims this function serves.

Keeping it for exhaustiveness is defensible (the alternative is a `_` arm that would silently absorb a future variant). The part I would push back on is `claim_error_tests::maps_no_result_set_claim_error` — it calls the mapping function directly, so it covers a branch no caller can reach rather than any behavior. Either is fine; just worth being deliberate about which.

mssql-py-core/tests/test_async_fetch.py
```diff
@@ -0,0 +1,723 @@
...
+        conn = await connect(client_context, logger)
+        try:
+            cursor = conn.cursor()
+            await cursor.execute("SELECT 1", use_prepare=False)
+            logger.events.clear()
+
+            assert await cursor.fetchone() == (1,)
+            assert await cursor.fetchone() is None
+
+            assert logger.events == [
```
David-Engel: **Nit.** Asserting `logger.events == [...]` for the whole list makes this test fail on any future INFO log added anywhere on the fetch path, which is not what the test is about. Membership keeps the intent and survives unrelated logging:

```python
assert (
    20,
    "PyAsyncCursor::fetchone: result set exhausted",
    "async_fetch.rs",
) in logger.events
```

mssql-py-core/src/async_fetch.rs
subrata-ms: Fixed. The fetch status is now published before [fetch_guard.complete()] updates the session ownership, preventing a concurrent [fetchone()] from observing [Ready] after the final batch claim has been released. I also calculate [has_row] once and reuse it for state publication, session completion, and exhaustion logging. The full async fetch suite passes.

mssql-py-core/src/async_session.rs
subrata-ms: Agreed. This is valid follow-up work rather than a blocker for this PR. The async path currently retains ownership until an additional [fetchone()] consumes the trailing DONE token. We’ll track adopting [TdsClient::peek_past_current_row] after #383 lands, including coverage for final-row release, multiple result sets, diagnostics, errors, and cancellation. We’ll also coordinate merge order so this work builds on #383’s finalized API and semantics.

Dependent task - https://dev.azure.com/SqlClientDrivers/mssql-python/_workitems/edit/47603

mssql-py-core/src/cursor.rs
subrata-ms: Fixed. DATETIMEOFFSET conversion is now fallible, so errors from [astimezone()] and the preceding datetime construction steps propagate to Python instead of silently returning a Rust debug string. I added focused tests for both lower- and upper-bound OverflowError cases. Existing live async and synchronous datetimeoffset tests continue to pass.

mssql-py-core/src/async_connection.rs
subrata-ms: Agreed. I removed claim_error_tests::maps_no_result_set_claim_error because it exercised a branch that connection-level claims cannot produce. I retained the explicit [NoResultSet] match arm to preserve exhaustive matching and ensure future enum variants cause a compile-time update rather than being absorbed by a wildcard.

mssql-py-core/tests/test_async_fetch.py
subrata-ms: Fixed. The test now checks that the expected exhaustion event is present in [logger.events] instead of requiring it to be the only event. This keeps the test focused on exhaustion logging and avoids coupling it to unrelated future INFO logs.

mssql-py-core/src/cursor.rs
subrata-ms: This is a mandatory fix for async flow. Agreed, this also touches the sync flow. I have updated the PR Description section with the below details - 

**Shared Type Conversion Fixes**

- Corrected MONEY and SMALLMONEY decoding so negative sub-unit values and minimum-range values preserve their sign without integer overflow.
- Corrected DATETIMEOFFSET decoding to convert the wire UTC instant back to its stored offset and local wall-clock value.
- These fixes apply to both synchronous and asynchronous cursor fetching because both paths share the same row conversion logic.
- Verified compatibility with existing mssql-python expectations and added boundary coverage for affected values.

mssql-py-core/src/async_session.rs
sumitmsft: I agree with @David-Engel. The release-on-next-fetch contract is the same shape PR #383 is addressing one layer down, so adopting `TdsClient::peek_past_current_row` here once #383 lands (and deciding the merge order deliberately) is the right call to keep the async front end from diverging from the ODBC path.

mssql-py-core/src/cursor.rs
sumitmsft: I agree with @David-Engel. The money sign fix is correct and is already covered by `test_fetchone_preserves_decimal_and_money_boundaries` (the `-0.0001` money/smallmoney cases). The actionable part is naming this synchronous-path change, and the datetimeoffset one below, in the PR description or splitting them into their own PR, since none of the linked ADO items cover them.

mssql-py-core/src/async_fetch.rs
```diff
@@ -0,0 +1,268 @@
...
+        cursor_id,
+        fetch_state,
+    } = resources;
+    if fetch_state.status() == FetchStatus::Exhausted {
+        session_state.ensure_open().map_err(map_claim_error)?;
+        return pyo3_async_runtimes::tokio::future_into_py(py, async move {
+            Python::attach(|py| Ok(py.None()))
+        });
+    }
+    let claim = session_state
```
sumitmsft, line 160: When cursor B has never executed but cursor A currently holds `Fetching`, `B.fetchone()` reaches `claim_fetch` and returns `Busy` ("Connection is busy with another cursor operation"), while the same call on an idle connection returns `NoResultSet` ("No active result set"). Question: is the "busy" wording intended for a cursor that simply has no result set of its own, or would `NoResultSet` be clearer here? Non-blocking.

mssql-py-core/src/async_fetch.rs
```diff
@@ -0,0 +1,268 @@
...
+    let resources = cursor.borrow(py).fetch_resources()?;
+    let FetchResources {
+        client,
+        dispatch,
+        session_state,
+        cursor_id,
+        fetch_state,
+    } = resources;
+    if fetch_state.status() == FetchStatus::Exhausted {
+        session_state.ensure_open().map_err(map_claim_error)?;
```
sumitmsft, line 155: The exhausted fast path calls `ensure_open()?` before returning `None`, so once a result set is exhausted, a later `fetchone()` on a `Closing`/`Closed`/`Broken` connection raises rather than returning `None`. Question: is raising here intended, or should the exhausted path short-circuit to `None` regardless of lifecycle? Non-blocking.

mssql-py-core/src/async_fetch.rs
```diff
@@ -0,0 +1,268 @@
...
+                    tracing::info!("PyAsyncCursor::fetchone: result set exhausted");
+                }
+                Python::attach(|py| match writer {
+                    Some(writer) => Ok(writer.to_py_tuple(py)?.into_any().unbind()),
+                    None => Ok(py.None()),
+                })
+            }
+            Err(error) => {
+                fetch_guard.fail(has_open_batch);
+                fetch_state.set(FetchStatus::NoResultSet);
```
sumitmsft, line 198: On a fetch error, `fetch_guard.fail(...)` may `mark_broken()` and always releases the operation, then `fetch_state.set(NoResultSet)` runs. Since the connection may now be `Broken`, the reset looks moot. Question: is there any path where a broken-session cursor is later re-driven and could misread `NoResultSet` as ready to claim? Likely benign; flagging to confirm.

sumitmsft (commented):
@David-Engel's review is thorough and I agree with it; the pass-1 commit already resolved the fetch-state ordering race, the datetimeoffset silent fallback (now propagates an error with bound tests), the brittle logger assertion, and removed the unreachable-arm test. No blocking findings from me. The money sign fix is correct and already covered by test_fetchone_preserves_decimal_and_money_boundaries (the -0.0001 money/smallmoney cases), so I have nothing to add there.

One small, non-blocking items follow inline:

- Finalizer "session busy" path (async_cursor.rs): when a cursor is dropped while the session is Busy, the finalizer logs and returns, deferring prepared-handle release to connection close. Under sustained cursor churn this could accumulate server-side prepared handles; please confirm connection close unprepares any handles orphaned this way.

mssql-py-core/src/async_fetch.rs
subrata-ms: This behavior is intentional. [Busy] reflects the immediate connection-wide conflict because cursor A owns the single TDS session. Returning [NoResultSet] would describe cursor B’s local state but hide the active blocker. This is also consistent with other cross-cursor operations, so no change is planned.

mssql-py-core/src/async_fetch.rs
subrata-ms: Raising is intentional. The exhausted state is only a fast path that avoids another wire operation; it does not bypass connection lifecycle validation. Once the parent connection is closing, closed, or broken, cursor operations should report that state rather than return a cached [None]. This behavior is explicitly covered by [test_exhausted_fetchone_after_connection_close_raises], so no change is planned.

mssql-py-core/src/async_fetch.rs
subrata-ms: Confirmed benign. [NoResultSet] does not make the cursor ready to claim; it only avoids the exhausted fast path. Any subsequent fetch reaches [claim_fetch()], which checks lifecycle first and returns [Broken]. Broken session state cannot transition back to open. The reset is still useful for the recoverable error path where no batch remains and the session stays reusable, so we plan to keep it.

sumitmsft (approved):

PR #385, opened 2026-08-26
Add character C to character SQL parameter conversions

Theekshna:
﻿## Description

Implements quadrant B of the parameter conversion matrix: `SQL_C_CHAR` / `SQL_C_WCHAR` bound against `char`, `varchar`, `text` and their wide counterparts, including the cross-family pairings.

- The wire type follows `ParameterType`, not the bound C type, so `@P1` is declared as the application asked for it.
- `ColumnSize` is a character count for every character SQL type. `text` / `ntext` carry no declared length but are still bounded by it; only the `max` types are unbounded.
- Over-long values are `22001`. Overflow made up only of blanks is trimmed without a diagnostic.
- Cross-family bindings transcode UTF-8 <-> UTF-16 instead of being rejected.

Conversion is organised on two axes - `AppValue` per C family, `SqlFamily` per SQL family - so a cross-family pairing is an adapter between canonical values rather than a new per-pair path.

### Long character types

`SQL_LONGVARCHAR` / `SQL_WLONGVARCHAR` are sent as `varchar(max)` / `nvarchar(max)`, still bounded by `ColumnSize`.

Declaring `text` / `ntext` - which is msodbcsql's default - is not currently possible: `mssql-tds` serializes `TEXT`/`NTEXT`/`IMAGE` in bulk-copy ROW format, so an RPC parameter carries a textptr/timestamp header plus a stray table-name byte and the server answers 4002. An earlier revision of this branch did declare them, which **regressed** `SQL_C_CHAR` + `SQL_LONGVARCHAR` - a binding that worked on `main`, where every character parameter went out as `max`. `max` is the form msodbcsql itself emits under `SQL_COPT_SS_LONGASMAX`, so this is one of its own modes rather than an invented shape, and only the declaration differs - truncation behaviour is unchanged.

Tracked as [AB#47591](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47591) (the `mssql-tds` wire format) and [AB#47592](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47592) (restoring the declaration, with `LongAsMax` as the opt-in fallback). Found by the `--compare-with-msodbcsql` parity run.

### Data-at-execution

Accepting the cross-family pairings opened a path that had previously been unreachable. `SQLPutData` streams its chunks to the wire untranscoded - a UTF-8 sequence can straddle two calls, so there is nothing to transcode a chunk against - and the streamed placeholder is chosen from the C type. `SQL_C_CHAR` + `SQL_WVARCHAR` + `SQL_DATA_AT_EXEC` would therefore have declared `varchar(max)` and sent UTF-8 bytes, corrupting non-ASCII silently. Before this PR that binding was `HYC00` at bind time, so it could not happen.

`dae_placeholder_type` now takes `ParameterType` and refuses a cross-family pairing with `HYC00` at execute. That restores the previous behaviour for the corrupting case without pretending the conversion is implemented.

The wider gap - the streamed path ignoring `ParameterType` and `ColumnSize` entirely, so no length is enforced - is [AB#47590](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47590). Closing it needs chunk-boundary-aware transcoding and streamed types beyond the three `max` variants.

### Parity

Every arm was checked against msodbcsql's `ParamToSQLType` (`sqlcfunc.cpp:2854`) and `DescribeRPCParam` (`sqlcmisc.cpp:7165`): ceilings, the three length measurements, the blank-trim region and polarity, and the `max`-type skip all match. Deliberate divergences, all documented at the call site:

- Narrow -> wide over-long is `22001` here. msodbcsql reads its running count before incrementing (`sqlcfunc.cpp:2926`) and lets one character escape untrimmed, so replicating it would mean replicating a defect.
- `SQL_C_CHAR` is UTF-8 rather than the client code page ([AB#47565](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47565)), because the only supported consumer is mssql-python.
- Cross-family data-at-execution is refused rather than streamed (above).
- Long character types are declared `max` rather than `text` / `ntext` (above).

Known-wrong behaviour is pinned by a test and carries a work item rather than being quietly accepted. Truncation is measured before `serialize_string` applies the collation ([AB#47584](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47584)); `an_astral_char_costs_two_units_but_four_bytes` and `invalid_utf8_at_the_limit_grows_past_it` assert the current, wrong, numbers so the fix registers as a behaviour change.

### Tests

- 887 `mssql-odbc` unit tests.
- New `char_conversions_test.cpp`, 20 e2e tests split out of `execute_test.cpp` (78 tests before and after, verified by diffing `--gtest_list_tests`).
- Data-at-execution coverage in `execute_test.cpp`: cross-family refusal plus statement reuse afterwards, and the same-family long pairing still streaming.
- The full e2e suite passes under `--compare-with-msodbcsql`.

Not covered: the collation-expansion failure of [AB#47584](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47584) needs a UTF-8-collated **database**, not a table - a parameter is encoded with the connection's `db_collation` before the server sees the target column - so it needs fixture work rather than a `CREATE TABLE`.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47368

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -165,21 +182,282 @@ pub(crate) fn is_data_at_exec_indicator(indicator: SqlLen) -> bool {
...
+    // Only trailing blanks may be dropped, and a blank is one unit in both
+    // encodings, so the overflow maps 1:1 onto trailing source units.
+    Ok(match text {
+        AppText::Utf8(mut bytes) => {
+            let keep = bytes.len().saturating_sub(overflow);
+            if bytes[keep..].iter().any(|b| *b != b' ') {
+                return Err(ParamBuildError::StringTruncation);
+            }
+            bytes.truncate(keep);
+            AppText::Utf8(bytes)
```
David-Engel, line 466: The units are mixed here and it works, but only by an invariant that isn't stated. `overflow` is in UTF-16 units when `wide_target`, while `keep = bytes.len() - overflow` is a byte offset into a UTF-8 buffer. That's safe because `0x20` never appears as a UTF-8 continuation byte, so "the trailing `overflow` bytes are all blanks" implies there really are `overflow` whole trailing spaces and the truncation lands on a char boundary.

That's load-bearing and currently untested: `overflowing_blanks_are_trimmed_silently` is ASCII-only and `a_narrow_source_is_measured_per_target_family` has no blanks, so nothing pins the narrow-source/wide-target case with multibyte content. Worth a line of comment:

```rust
// U+0020 is the only UTF-8 byte equal to b' ', so an all-blank tail is
// `overflow` whole characters and truncating lands on a boundary.
```

plus a case like `convert_char(SQL_C_CHAR, SQL_WVARCHAR, 1, "é   ")` → `"é"`.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -165,21 +182,282 @@ pub(crate) fn is_data_at_exec_indicator(indicator: SqlLen) -> bool {
...
+                            .into_bytes(),
+                    ),
+                }
+            }
+            (Self::Utf8(bytes), true) => Self::Utf16(
+                String::from_utf8_lossy(&bytes)
+                    .encode_utf16()
+                    .flat_map(u16::to_le_bytes)
+                    .collect(),
+            ),
```
David-Engel, line 283: The narrow → wide path allocates a throwaway `String` twice for the same buffer: once in `len_in_target_units` to count units, then again here to convert. `str::from_utf8(bytes)` borrows for the common (valid) case and only falls back to `from_utf8_lossy` on `Err`, which is the same allocation care b155361a applied to `decode_utf16le`.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -165,21 +182,282 @@ pub(crate) fn is_data_at_exec_indicator(indicator: SqlLen) -> bool {
...
+    /// Length in the units the declared length is compared against.
+    ///
+    /// A wide source counts its UTF-16 units whichever family it lands in -
+    /// msodbcsql assumes "1 WCHAR converts to 1 byte" for a narrow target
+    /// (`sqlcfunc.cpp:2946`) rather than encoding to find out.
+    fn len_in_target_units(&self, wide_target: bool) -> usize {
+        match self {
+            Self::Utf16(bytes) => bytes.len() / size_of::<u16>(),
+            Self::Utf8(bytes) if !wide_target => bytes.len(),
+            Self::Utf8(bytes) => String::from_utf8_lossy(bytes).encode_utf16().count(),
```
David-Engel, line 309: Same measure-only allocation as in `transcode` — `String::from_utf8_lossy(bytes).encode_utf16().count()` builds a `String` just to count. `str::from_utf8(bytes).map(|s| s.encode_utf16().count())` avoids it on the valid path.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -165,21 +182,282 @@ pub(crate) fn is_data_at_exec_indicator(indicator: SqlLen) -> bool {
...
+        StreamedSqlType::VarcharMax => {
+            matches!(sql_type, SQL_CHAR | SQL_VARCHAR | SQL_LONGVARCHAR)
+        }
+        StreamedSqlType::NVarcharMax => {
+            matches!(sql_type, SQL_WCHAR | SQL_WVARCHAR | SQL_WLONGVARCHAR)
+        }
+        StreamedSqlType::VarBinaryMax => {
+            matches!(sql_type, SQL_BINARY | SQL_VARBINARY | SQL_LONGVARBINARY)
+        }
+    };
```
David-Engel, line 214: Optional: family knowledge now lives in three places — `conversion_matrix`'s `CHARACTER_C_TARGETS`, `sql_family`, and this `same_family` match. `sql_family(sql_type)` combined with `is_wide_character_sql_type(sql_type)` expresses the same check without a third list to keep in sync. `cross_family_dae_is_rejected` covers the current lists, so this is a maintainability point rather than a correctness one.

mssql-odbc/tests/e2e/tests/char_conversions_test.cpp
```diff
@@ -0,0 +1,782 @@
...
+// and its blank exemption, encoding, and the indicator/terminator rules.
+//
+// Statement lifecycle (prepare, re-execute, cached handles, data-at-execution)
+// lives in execute_test.cpp.
+//
+// Tests that require a live SQL Server are gated by ODBCTestConfig::HasConnection().
+
+#include "odbc_test_fixture.h"
+
+#include <cassert>
```
David-Engel: Nit: `<cassert>` isn't used in this file.

mssql-odbc/src/params/conversion_matrix.rs
```diff
@@ -86,12 +100,6 @@ mod tests {
         for sql_type in [SQL_BINARY, SQL_VARBINARY, SQL_LONGVARBINARY] {
             assert!(is_supported_conversion(SQL_C_BINARY, sql_type));
         }
-    }
-
-    #[test]
-    fn cross_family_character_conversions_are_unsupported() {
-        assert!(!is_supported_conversion(SQL_C_CHAR, SQL_WVARCHAR));
-        assert!(!is_supported_conversion(SQL_C_WCHAR, SQL_VARCHAR));
         assert!(!is_supported_conversion(SQL_C_BINARY, SQL_VARCHAR));
```
David-Engel, line 103: Nit: this negative assertion moved in from the deleted `cross_family_character_conversions_are_unsupported`, and the test name no longer covers it. Either rename to something like `binary_c_type_reaches_only_the_binary_sql_types`, or keep the negative case in its own test.

David-Engel (commented):
Reviewed the full diff and ran the crate's unit tests locally (887 passed).

The two-axis design — `AppValue` per C family, `SqlFamily` per SQL family — is the right shape for this, and it keeps a cross-family pairing an adapter between canonical values rather than a new per-pair path. The msodbcsql parity work is unusually well-sourced, and I like that known-wrong behaviour is pinned by a test and carries a work item instead of being quietly accepted.

I worked through the mixed-unit trim arithmetic, the bind-gate interaction for `ColumnSize == 0` on the long types, and the data-at-execution regression path. No blocking findings. Everything below is a suggestion or a nit.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -165,21 +182,282 @@ pub(crate) fn is_data_at_exec_indicator(indicator: SqlLen) -> bool {
...
+                    ),
+                }
+            }
+            (Self::Utf8(bytes), true) => Self::Utf16(
+                String::from_utf8_lossy(&bytes)
+                    .encode_utf16()
+                    .flat_map(u16::to_le_bytes)
+                    .collect(),
+            ),
+            (Self::Utf16(bytes), false) => Self::Utf8(decode_utf16le(&bytes).into_bytes()),
```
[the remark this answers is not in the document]
Theekshna: Agreed the behaviour is wrong and worth fixing — `&#26085;` reaching the server as data with `SQL_SUCCESS` and no diagnostic is indefensible. Filed as [AB#47598](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47598).

Two corrections to the premise, though, because they change what the right remedy is.

**This isn't newly enabled — it's reachable on `main` today.** The corrupting path is:

`SqlString` tagged `EncodingType::Utf8` → `as_raw_wire_bytes()` returns `None` for that variant (`sql_string.rs`) → `to_utf8_string()` → `encoding.encode()`.

On `origin/main`, `AppValue::NarrowText` produces exactly `SqlType::VarcharMax(Some(SqlString::new(bytes, EncodingType::Utf8)))`, and `SQL_C_CHAR` + `SQL_VARCHAR` was already a supported pairing. So binding the UTF-8 bytes for U+65E5 to a `varchar` parameter on a Latin1 database corrupts identically without this PR. What P4 changes is exposure — one C type to two — not the existence of the class.

**So gating `SQL_C_WCHAR` → narrow back to `HYC00` wouldn't fix it.** It would leave `SQL_C_CHAR` → narrow producing the same `&#26085;`, while removing the *mappable* cross-family case, which works correctly and is the point of the feature (`CrossFamilyCharParamRoundTripsNonAscii` pins U+00E9 round-tripping through `SQL_VARCHAR`).

**The actual fix** is in `mssql-tds`: `serialize_string` takes `had_errors` from `encoding.encode()` and only calls `tracing::warn!` (`tds_value_serializer.rs:1422`). It needs to return a typed error that `mssql-odbc` maps to `22018`. Worth noting two sites nobody has flagged do the same thing more quietly — `:1447` (Latin-1 fallback on an unsupported LCID) and `:1462` (no collation supplied) both substitute `b'?'` silently.

That's deferred rather than done here for one reason: `serialize_string` is shared with bulk copy, so erroring unconditionally changes BCP behaviour too. Scoping it to RPC parameters needs the same `TdsTypeContext` split that [AB#47591](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47591) already requires, so the two are linked and should land together.

The behaviour is pinned from both sides in the meantime — `UnmappableCharacterIsSilentlyCorrupted` (e2e) and `unmappable_character_becomes_a_numeric_character_reference` (`mssql-tds`), whose doc comment already says nothing upstream sees the flag — so the fix will register as a behaviour change rather than a silent correction.

mssql-odbc/src/params/conversion_matrix.rs
Theekshna: ﻿Renamed to `binary_c_type_reaches_only_the_binary_sql_types`. 8b6f3dc5.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: ﻿Same as above - `from_utf8_lossy` borrows on the valid path, so nothing is built just to count. The redundant *scan* is real and tracked with AB#47584.

mssql-odbc/tests/e2e/tests/char_conversions_test.cpp
Theekshna: ﻿Mine - left behind when the `assert` became `EXPECT_LT`. Removed.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: ﻿Agreed, that third list was mine. `same_family` now derives from `sql_family(sql_type)` plus `is_wide_character_sql_type(sql_type)`. 8b6f3dc5.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: ﻿`String::from_utf8_lossy` returns `Cow<'_, str>` and borrows on the valid path, so there's no throwaway `String` here - it only allocates on invalid input, which is exactly the `str::from_utf8` / fallback split you're describing. Verified:

```
valid utf8   -> Borrowed (no allocation)
invalid utf8 -> Owned (allocated)
```

The `decode_utf16le` comparison doesn't carry over: `String::from_utf16_lossy` takes `&[u16]`, so that one had no borrowing form and genuinely allocated.

There *is* a real cost here, just not allocation - the buffer is scanned twice for narrow -> wide, once to count and once to transcode. Collapsing them changes what gets measured, which is AB#47584's territory, so I'd rather do it there than as a perf tweak.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: ﻿Good catch - that invariant was load-bearing and unwritten. Added the comment and `a_multibyte_narrow_source_trims_blanks_on_a_char_boundary`, which is your case: `("e-acute   ", limit 1, wide)` is 4 units, overflow 3, `keep = 5 - 3 = 2`, trims to the 2-byte character. Fixed in 8b6f3dc5.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -165,21 +182,287 @@ pub(crate) fn is_data_at_exec_indicator(indicator: SqlLen) -> bool {
...
+
+    /// Length in the units the declared length is compared against.
+    ///
+    /// A wide source counts its UTF-16 units whichever family it lands in -
+    /// msodbcsql assumes "1 WCHAR converts to 1 byte" for a narrow target
+    /// (`sqlcfunc.cpp:2946`) rather than encoding to find out.
+    fn len_in_target_units(&self, wide_target: bool) -> usize {
+        match self {
+            Self::Utf16(bytes) => bytes.len() / size_of::<u16>(),
+            Self::Utf8(bytes) if !wide_target => bytes.len(),
```
shiwanigupta0809: Measuring a narrow source in UTF-8 bytes rejects values that fit, and disagrees with the wide path on the same value.

Two tests in this PR, same value, same target:

```rust
// a_wide_source_is_measured_in_utf16_units_for_both_families
assert!(convert_char(SQL_C_WCHAR, SQL_VARCHAR, 4, "caf\u{e9}").is_ok());
```

```cpp
// NarrowMultibyteIsMeasuredInUtf8Bytes — SQL_C_CHAR, SQL_VARCHAR, ColumnSize 4
EXPECT_EQ(SQL_ERROR, SQLExecute(stmt_));
EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "22001");
```

`café` into `varchar(4)` is accepted as `SQL_C_WCHAR` and rejected `22001` as `SQL_C_CHAR`. The second half of that same e2e test shows the rejection is false — rebound at `ColumnSize 5` it reaches the server and `LEN(?)` returns 4.

This is new behaviour rather than a pre-existing gap. On the merge base, `bound_param_to_value` sent every non-NULL narrow value as `VarcharMax` and never consulted `param.sql_type` or `param.column_size` — `ColumnSize` only reached `typed_null` — and `SQL_C_CHAR` + `SQL_VARCHAR` was already an accepted bind pairing. So this binding worked before and is a hard error now. Apps passing `ColumnSize 0` are unaffected; those passing an explicit size get `22001` on data that fits.

msodbcsql accepts it on a Windows client, where `SQL_C_CHAR` is CP1252 and `café` is 4 bytes, so its byte comparison is effectively a character comparison. Inheriting that byte-count rule after redefining `SQL_C_CHAR` as UTF-8 makes it over-count every non-ASCII value — the rule was only sound because of the encoding it assumed. (On a UTF-8 client locale msodbcsql measures the same 5 bytes we do, so the divergence is configuration-dependent rather than universal.)

I see this is pinned and tracked under AB#47584, and I agree exactness needs the collation at this layer. But the false-rejection half doesn't need exactness to improve — counting `chars()` rather than bytes for the narrow→narrow case makes the two C types agree, matches ODBC's definition of `ColumnSize` as a character count, and matches msodbcsql under any single-byte collation. Still not exact under a UTF-8 collation, but the failure mode flips from rejecting valid data outright to letting the server return an accurate error, which seems the better direction to be wrong in while AB#47584 is open.

The wide path's own doc comment already makes this argument: *"Counting the transcoded UTF-8 instead would falsely reject values that fit: `varchar(n)` bounds collation bytes, and under a single-byte collation `café` is four of them, not five."*

shiwanigupta0809 (commented):
Worked through the conversion matrix and spot-checked the msodbcsql parity claims against `ParamToSQLType` / `TranslateParam` (`sqlcfunc.cpp`) and `DescribeRPCParam` (`sqlcmisc.cpp`) — the ceilings, the three length measurements, the blank-trim polarity and the `max`-type skip all hold up, and the `sqlcfunc.cpp:2926` read-before-increment really is a defect worth not replicating. Nice work on the sourcing.

One finding on the narrow-source measurement, inline.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Confirmed and fixed in ab4d2ef1. Verified the regression claim too: at the merge base `bound_param_to_value` sent `NarrowText` as `VarcharMax` unconditionally and `column_size` reached only `typed_null`, so that binding worked before P4.

One refinement: `chars()` would not have unified the two C types, since the wide source counts UTF-16 units and an astral character is one `char` but two units — U+1F600 would count 1 narrow vs 2 wide. So I deleted the narrow arm instead. Every source now uses UTF-16 units, which is the unit msodbcsql itself uses in three of its four arms, including the narrow-to-wide walk that counts an astral character as two (`sqlcfunc.cpp:2935`).

On parity, this cuts against us and the write-up should say so. msodbcsql's byte count (`cchDest = cbData`, `:2952`) is the wire length only when no client-side transcode happens. `DoCharToCharConversion` (`sqlcprot.h:4113`) forces one for a UTF-8 client against a non-UTF-8 server — translation is on by default — and msodbcsql then still measures the pre-transcode bytes. Since our `SQL_C_CHAR` is always UTF-8, we had made that latent defect unconditional. So this is a deliberate divergence in the configuration closest to us, on "don't copy the bug" grounds rather than parity ones. Signed off and recorded in `mssql-odbc.instructions.md`.

Also a correction: the failure mode does not become "an accurate server error". `serialize_char_varchar_direct` raises an untyped `UsageError`, which surfaces as `HY000`, and the `max`/`text` types carry no check and send the value. And it is not a pure win — `SQL_C_CHAR` "☕☕☕" into `varchar(3)` was a correct `22001` and is now an opaque failure. Both now stated plainly in the code, the plan doc, and AB#47584, which I rewrote with a worked table of the remaining wrong answers.

Tests: `a_narrow_source_is_measured_per_target_family` is replaced by `both_character_c_types_measure_a_value_alike` (café / three U+2615 / U+1F600 × both targets × both C types). The e2e test now asserts `LEN(?)` = 4 at `ColumnSize` 4; its compare-skip stays, since msodbcsql's answer still depends on the client code page. Full e2e suite passes.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -165,21 +182,307 @@ pub(crate) fn is_data_at_exec_indicator(indicator: SqlLen) -> bool {
...
+
+    // Only trailing blanks may be dropped, and a blank is one unit in both
+    // encodings, so the overflow maps 1:1 onto trailing source units. That also
+    // makes `overflow` usable as a byte count on the UTF-8 arm below: `b' '`
+    // never occurs as a continuation byte, so an all-blank tail is that many
+    // whole characters and the cut lands on a boundary.
+    Ok(match text {
+        AppText::Utf8(mut bytes) => {
+            let keep = bytes.len().saturating_sub(overflow);
+            if bytes[keep..].iter().any(|b| *b != b' ') {
```
David-Engel, line 462: **Suggestion.** The blank-trim check has no test for a *partially* blank overflow. Every test that reaches this branch has an overflow region that is entirely blanks or entirely non-blank, so the `any` is untested as a whole-region check.

Proven rather than asserted — I weakened it to inspect only the first byte:

```rust
if bytes[keep..].iter().take(1).any(|b| *b != b' ') {
```

All 50 `param_convert` tests still pass. Under that mutation, `SQL_C_CHAR` `"ab c"` into `varchar(3)` returns `Ok(Varchar("ab", 2))` — silently dropping `" c"` — while the wide path correctly returns `Err(StringTruncation)`. That is silent data loss on a reachable binding, invisible to the suite.

The shipped code is correct; this is purely a missing test:

```rust
// Only a wholly blank overflow may be dropped. A tail that merely starts
// with a blank is still truncation.
assert_eq!(
    convert_char(SQL_C_CHAR, SQL_VARCHAR, 2, "ab c").unwrap_err(),
    ParamBuildError::StringTruncation
);
```

mssql-odbc/tests/e2e/tests/char_conversions_test.cpp
```diff
@@ -0,0 +1,804 @@
...
+// unit was measured against ColumnSize and eight bytes were sent, the same hole
+// as the GB18030 case in NarrowMultibyteIsMeasuredInUtf8Bytes but reachable on a
+// plain Latin1 database. At ColumnSize 1 it never reaches the server -
+// serialize_char_varchar_direct rejects it with an opaque driver error rather
+// than 22001.
+//
+// Skipped under comparison: msodbcsql converts with SystemLocale::FromUtf16
+// (WideCharToMultiByte), which substitutes a single '?'.
+//
+// Known-wrong, deferred (AB#47598).
```
David-Engel, line 773: **Suggestion.** This PR deliberately routes more traffic to the late-failure path and documents the `HY000` as the accepted residual, but nothing asserts the connection survives it. Worth an e2e case here in the same known-wrong-but-pinned style, so a future change to that check's position can't silently turn a recoverable error into a connection kill.

One caveat on how to write it, because the obvious version pins the wrong case. It is tempting to conclude this is safe because the check at `tds_value_serializer.rs:1743` fires before any value byte is written. That part is true, but the RPC type metadata preamble is already written before `serialize_value` is even called (`sqltypes.rs:698-704`), so bytes for this RPC *have* reached the writer:

```rust
// Step 1: Write the RPC type metadata preamble
self.write_rpc_type_metadata(packet_writer, db_collation, type_metadata).await?;
// Step 2: Convert to ColumnValues + TdsTypeContext and serialize value
TdsValueSerializer::serialize_value(packet_writer, &column_value, &ctx).await?;
```

What makes a small value recoverable is that those bytes never reach the *wire*: `write_byte_async` -> `handle_overflow_if_needed` only sends once `position() >= max_payload_size` (`packet_writer.rs:284`). A short parameter is nowhere near a packet boundary, so nothing flushes and dropping the writer discards it cleanly.

`packet_writer.rs:270-273` spells out the other regime:

> Only meaningful once at least one packet of the message has been sent — an unsent message is abandoned by dropping the writer. The server still answers an ignored message with a DONE token, which the caller must consume before reusing the connection.

So a test using a short value would pass while covering only the benign case. The regime that can actually cost the connection — a large parameter, or a failing parameter late enough in the list that earlier ones already forced a flush — needs `cancel_current_message` plus consuming the DONE token. That is the case worth pinning, and it is also the one the P6 note in `parameters_plan.md` is really about.

mssql-odbc/docs/parameters_plan.md
```diff
@@ -346,32 +346,152 @@ to a canonical `i128`, so there is no copy to skip.
...
+the worse outcome of the two. Routing either to `ERR_PARAM_STRING_TRUNCATION`
+needs a typed error out of `mssql-tds` - matching on the message text would be
+guesswork - so it belongs with AB#47584 rather than as a local patch.
+
+**This is a behavioural regression for a subset of inputs, not a pure
+improvement.** `SQL_C_CHAR` `"[three U+2615]"` into `varchar(3)` was a correct
+`22001` under the byte count; it is now accepted as three units and fails
+downstream as an opaque `HY000` - or, under a single-byte collation, each
+character is unmappable and becomes an eight-byte numeric character reference
+(AB#47598), so 24 bytes are offered to a `varchar(3)`. CJK and astral input
```
David-Engel: **Nit.** `&#9749;` is seven characters, not eight — U+2615 is code point 9749, four digits — so three of them are 21 bytes, not 24. Eight is right for `&#26085;` (U+65E5, five digits), the other example in this PR, which looks like where the figure came from. Either say "seven-byte" / "21 bytes", or use a five-digit code point so both examples line up.

David-Engel (commented):
Re-reviewed at `ab4d2ef1`. Ran the `mssql-odbc` lib suite locally (889 pass), mutation-tested the new measurement change, read the automated coverage report (95% diff, above the 85% target), and confirmed CI is green.

**First, correcting myself.** Two of my findings from the last round were wrong. `String::from_utf8_lossy` returns `Cow<'_, str>` and borrows on the valid path, so there was no throwaway `String` to remove in either `transcode` or the length count — and the `decode_utf16le` precedent I cited doesn't carry, because `from_utf16_lossy` takes `&[u16]` and has no borrowing form. The rebuttals were right, and the real cost identified instead — the buffer being scanned twice on narrow → wide — is correctly scoped to AB#47584 rather than patched here. I should have checked the signature before citing that commit.

**On `ab4d2ef1`.** I re-worked the mixed-unit arithmetic under the new scheme, since it now applies to both target families rather than only wide. It holds: `units <= bytes.len()` for any UTF-8 buffer, so `overflow` never exceeds the byte length, and because `b' '` is never a UTF-8 continuation byte, an all-blank tail of `overflow` bytes is exactly `overflow` whole characters.

I mutation-tested both new tests rather than trusting their names — reverting `len_in_utf16_units` to `bytes.len()` fails both `both_character_c_types_measure_a_value_alike` and `a_multibyte_narrow_source_trims_blanks_on_a_char_boundary`. They genuinely guard. The astral case is a good catch; `chars()` would have left the two C types disagreeing on U+1F600.

The divergence write-up in `mssql-odbc.instructions.md` is the right handling — it names the regressing input shape rather than selling the change as a pure win, and carries a dated sign-off.

Three findings below, none blocking.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Confirmed and fixed. Your mutation is worse than you measured — I ran the whole suite under it, not just `param_convert`: **889 passed, 0 failed**. Nothing anywhere detected that the narrow arm had stopped checking the region.

Added `a_partially_blank_overflow_is_truncation`, covering all four C-type × SQL-family combinations rather than only the narrow one, since the wide arm has the same `any`-over-the-region shape and was equally untested.

Mutation-verified both ways: `iter().take(1)` and `.chunks_exact(..).take(1)` each make it fail; unmutated it passes.

mssql-odbc/tests/e2e/tests/char_conversions_test.cpp
Theekshna: Agreed on the caveat — that's the part worth recording, so I noted it in P6 rather than writing the test now:

> Pin that a late `HY000` from an over-long parameter leaves the connection reusable. Only a value large enough — or one failing late enough — to have already flushed a packet needs `cancel_current_message` and the DONE token consumed; a short value never reaches a packet boundary, so a test using one passes without exercising the recovery at all.

Deferred rather than written because the flushed regime needs the large-parameter setup plus DONE-token handling, and the easy version would look like coverage while pinning only the benign half.

mssql-odbc/docs/parameters_plan.md
Theekshna: Right — U+2615 is four digits, so `&#9749;` is 7 bytes and three is 21. Fixed the numbers and inlined the literal reference so the count is checkable rather than asserted.

David-Engel (commented):

## Summary

Re-review of the character-to-character parameter conversion work, scoped to what changed since my last review at `ab4d2ef1`: a single commit (`532dd63d`) adding `a_partially_blank_overflow_is_truncation` plus the matching notes in `parameters_plan.md`. The test is correct and, more to the point, it is not a tautology — I verified by mutation that it pins exactly the property it claims to. No findings.

## Verification

Compiled, tested, and mutation-probed at `532dd63d` in a throwaway worktree:

- `cargo clippy -p mssql-odbc -p mssql-tds --all-targets -- -D warnings` — clean.
- `cargo nextest run -p mssql-odbc --lib` — **890 tests, all pass**, 1 skipped.

The interesting part is the mutation probe. The new test's own doc comment claims the pre-existing all-blank / all-non-blank split leaves `fit_to_declared_length`'s `any` untested *as a whole-region check*, so weakening it to inspect a single unit would slip past the rest of the suite. I checked that rather than taking it on faith, by weakening the UTF-8 arm at `param_convert.rs:462` from

```rust
if bytes[keep..].iter().any(|b| *b != b' ') {
```

to a single-unit check `if bytes[keep] != b' ' {`, then running the three neighbouring blank-handling tests:

```
PASS overflowing_blanks_are_trimmed_silently
PASS a_multibyte_narrow_source_trims_blanks_on_a_char_boundary
FAIL a_partially_blank_overflow_is_truncation
  called `Result::unwrap_err()` on an `Ok` value: Varchar(Some("ab"), 2)
```

Exactly one test fails, and it is the new one — the two pre-existing tests are blind to the regression. That is the coverage claim proven rather than asserted, and it is the right test to have added. (Source restored afterwards; the mutation existed only in the throwaway worktree.)

I also walked the four `(c_type, sql_type)` pairs by hand against both arms of `fit_to_declared_length` to confirm `"ab c"` at `limit = 2` really exercises the intended region: `overflow = 2`, `keep = 2` on the UTF-8 arm and `keep = 4` bytes on the UTF-16 arm, leaving a `" c"` tail that starts blank and ends non-blank in both encodings. It does.

Not exercised: the C++ e2e suite (`char_conversions_test.cpp`) and any live-server path — I have no `.env` configured, so treat those as reviewed by reading only.

## Findings

**Blocking** — none found.

**Suggestion** — none found.

**Nit** — none found.

Two things I checked and deliberately am *not* raising, so they are not mistaken for gaps: the new `mssql-tds` test `unmappable_character_becomes_a_numeric_character_reference` documents a genuinely alarming behavior (an unmappable character silently becomes `&#26085;` markup on the wire, with only a `had_errors` warning), but it is pre-existing, already tracked as AB#47598, and the trade-off is written up honestly in `parameters_plan.md` including the fact that this PR makes a subset of CJK/astral inputs regress from a clean `22001` to a downstream failure. Documenting a known regression with a tracked follow-up and an explicit rationale is the right call, not something to relitigate in review.

## CI

PR validation is red: [Kerberos Test (PR) — Kerberos Authentication Tests](https://sqlclientdrivers.visualstudio.com/904996cc-6198-4d39-8540-eca72bdf0b7b/_build/results?buildId=170437&view=logs&jobId=557248f6-d8ae-5a65-78d5-39e5bcb8a196) failed on build 170437, which fails the parent `mssql-rs Pull request validation` check. Nothing in this diff touches authentication, so this looks environmental rather than caused by the change — but it needs a re-run or a look before merge, since the checklist claims validation passes.

This PR looks ready for human review once that check is green. To be explicit: this is not an approval.

Theekshna:
Splitting out a finding from the review threads here, so it gets its own review rather than riding along at the end of this one.

**[AB#47687](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47687) — `mssql-tds`: retract a partially sent request so a late serialization failure does not kill the connection.**

Chasing @David-Engel's [comment on the `HY000` residual](https://github.com/microsoft/mssql-rs/pull/385#discussion_r3873948200) turned up a real defect. If value serialization fails *after* a packet has already been flushed, the `PacketWriter` is dropped without telling the server, leaving it holding an incomplete message with no EOM. The next command on that connection is read as a continuation and the server answers native **4002**, "The incoming tabular data stream (TDS) protocol stream is incorrect." The failing `SQLExecute` returns a plausible `SQL_ERROR` and the connection dies silently — the symptom shows up on the *next* statement.

**P4 is what exposed it.** Before this PR every character parameter went out as `(n)varchar(max)`, so the declared-length check never fired and serialization never failed part-way. Declaring real lengths is correct, and it made a latent connection-kill reachable. The same shape applies to any late serialization failure, including the deferred 32-bit wire-length wrap.

The reproduction needs a preceding parameter large enough to flush — a large *failing* parameter isn't sufficient, since its bytes are never written. Two e2e cases are written and the post-flush one currently fails with 4002.

Splitting it out because the fix is in `mssql-tds` and touches every parameter-carrying send site, so it affects `mssql-js`, `mssql-py-core`, and the CLI as much as ODBC. It also picks up a second finding: the retraction drain inherits `remaining_request_timeout`, which is `None` by default, so it can block indefinitely — msodbcsql bounds its cancel separately with `DEFAULT_CANCEL_TIMEOUT`. That affects `cancel_streamed_write` today, independently of this PR.

Design is verified against msodbcsql's `BATCHCTX::Cancel` — same two-branch handling, and its `CRPCRequest::SendRequest` runs the identical retraction on a data conversion error. Details on the work item.

No release is affected: `mssql-odbc` has no tags and `mssql-python` ships against the C++ driver. The follow-up branch will be based on this one, since without P4's declared lengths the tests can't trigger the failure.

shiwanigupta0809 (approved):

PR #390, opened 2026-08-26
Pin mssql-tds to the feed index when publishing mssql-mock-tds

### LINKED ISSUE #389: Published mssql-mock-tds cannot be restored from the Azure feed: mssql-tds dependency points at crates.io
## Problem

Every published version of `mssql-mock-tds` on the `mssql-rs_Public` Azure Artifacts feed declares its `mssql-tds` dependency as coming from **crates.io**, where `mssql-tds` does not exist (HTTP 404). Any consumer restoring the crate fails.

Reproduced with a fresh cargo project depending on `mssql-mock-tds 0.1.0-dev.20260826.169925`:

```
error: no matching package named `mssql-tds` found
location searched: crates.io index
required by package `mssql-mock-tds v0.1.0-dev.20260826.169925 (registry `mssql-rs_Public`)`
```

Resolution fails before any download. All 9 published versions are affected — the feed's sparse index records:

```json
{"name":"mssql-tds", "registry":"https://github.com/rust-lang/crates.io-index"}
```

## Root cause

`.pipeline/scripts/pin-mssql-tds-dep-version.ps1` rewrites the path dependency to add a `version` (required, since cargo rejects publishing a path dep with no version) but never specifies a registry. Cargo records a dependency that has a version and no registry as coming from the **default registry — crates.io**.

## Why CI never caught it

`.cargo/config.ci.toml` sets `[source.crates-io] replace-with = "mssql-rs_Public"`. That silently redirects the crates.io lookup back to the feed, where `mssql-tds` does exist, so publish-verify and every in-repo build succeed.

Confirmed: the same consumer project resolves 185 packages and builds clean once source replacement is added. But the lockfile it produces is still wrong:

```toml
name = "mssql-tds"
source = "registry+https://github.com/rust-lang/crates.io-index"
```

So the workaround only works for consumers who both know about it and apply it.

## Fix

Emit `registry-index = "<feed sparse index>"` on the dependency at publish time, using the plain feed URL rather than the `~force-auth` variant so anonymous readers of the public feed can still resolve it.

## Notes

Existing published versions stay broken — the metadata is baked into the index and cannot be edited. They need republishing under new versions.

This issue is scoped to the Azure feed. Publishing to crates.io is a separate effort with a different constraint: crates.io rejects any crate with a dependency sourced from another registry, so `mssql-tds` would have to be published there first.

saurabh500:
Fixes #389

## What

Publishing `mssql-mock-tds` to the Azure feed produced a manifest that declares its `mssql-tds` dependency as coming from crates.io, where `mssql-tds` does not exist. Restore fails for every consumer.

`pin-mssql-tds-dep-version.ps1` added a `version` to the path dep (cargo rejects publishing a path dep without one) but no registry. Cargo treats `version` with no registry as **crates.io**.

CI never caught it because `.cargo/config.ci.toml` source-replaces crates.io with the feed, redirecting the lookup back to a place the crate does exist.

## Changes

- `pin-mssql-tds-dep-version.ps1` takes a new required `-IndexUrl` and emits `registry-index = "sparse+<url>"` on the dependency.
- `PublishFeeds-Sandbox.yml` passes the existing `$(cargoSparseIndex)` variable — same value already used for preflight and index polling, so there's one source of truth.
- The script rejects a `~force-auth` URL. That URL gets baked into the published manifest, so using it would silently force every consumer to authenticate against a feed that reads anonymously, and the breakage would only surface long after the publish succeeded. The other `$(cargoSparseIndex)` callers only poll the index, so a wrong URL fails loudly in the same run and needs no equivalent check.

## Verification

All checks run locally against the real feed.

**Before** — fresh project depending on the published `mssql-mock-tds 0.1.0-dev.20260826.169925`:

```
error: no matching package named `mssql-tds` found
location searched: crates.io index
```

**After** — ran the real CI sequence (`stamp-crate-versions-sandbox.ps1` → `pin-mssql-tds-dep-version.ps1` → `cargo package -p mssql-mock-tds`). The packaged manifest now contains:

```toml
[dependencies.mssql-tds]
version = "0.1.0-dev.20260826.169925"
registry-index = "sparse+https://pkgs.dev.azure.com/sqlclientdrivers/public/_packaging/mssql-rs_Public/Cargo/index/"
default-features = false
```

Then consumed that exact packaged crate from a project with **no registry alias and no source replacement** — the configuration that failed before:

```
Locking 185 packages to latest compatible versions
Compiling mssql-tds v0.1.0-dev.20260826.169925 (registry `sparse+https://pkgs.dev.azure.com/.../Cargo/index/`)
Finished `dev` profile in 51.96s
```

Lockfile now records the feed instead of crates.io:

```toml
name = "mssql-tds"
source = "sparse+https://pkgs.dev.azure.com/sqlclientdrivers/public/_packaging/mssql-rs_Public/Cargo/index/"
```

Also verified:
- The `~force-auth` check exits 1 and leaves `Cargo.toml` untouched.
- The healthy URL does **not** trip that check and produces the correct dep line.
- An already-`sparse+`-prefixed URL is not double-prefixed.
- `path` + `version` + `registry-index` coexist fine; local path builds are unaffected.
- https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=170138&view=results 

No Rust code changed, so `bfmt`/`bclippy`/`btest` are unaffected. The repo working tree was restored after each test.

## Not covered

Existing published versions stay broken — the metadata is baked into the feed index and can't be edited. They need republishing under new versions.

Publishing to crates.io is deliberately out of scope. It has a conflicting requirement (crates.io rejects deps sourced from other registries, so `mssql-tds` would need to be published there first) and will be handled separately.

### DIFF
--- a/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
+++ b/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
@@ -11,23 +11,49 @@
   stamped version while keeping the path (cargo uses the version on publish and the
   path for local verify builds).
 
+  A version alone is not enough: cargo records a dependency with no registry as
+  coming from crates.io, where mssql-tds does not exist. Consumers of the feed then
+  fail to resolve mssql-mock-tds with "no matching package named `mssql-tds` found,
+  location searched: crates.io index". CI does not hit this because
+  .cargo/config.ci.toml source-replaces crates.io with the feed. So we also pin the
+  feed's own index.
+
 .PARAMETER Version
   The stamped crate version to pin (typically $(crateVersion)).
+
+.PARAMETER IndexUrl
+  Sparse index base URL of the feed being published to (typically
+  $(cargoSparseIndex)). Pass the plain URL, not the ~force-auth variant, so that
+  anonymous readers of a public feed can resolve the dependency.
 #>
 [CmdletBinding()]
 param(
-    [Parameter(Mandatory = $true)][string]$Version
+    [Parameter(Mandatory = $true)][string]$Version,
+    [Parameter(Mandatory = $true)][string]$IndexUrl
 )
 
 $ErrorActionPreference = 'Stop'
 
 if ([string]::IsNullOrWhiteSpace($Version)) { Write-Error 'Version not set'; exit 1 }
+if ([string]::IsNullOrWhiteSpace($IndexUrl)) { Write-Error 'IndexUrl not set'; exit 1 }
+
+# ~force-auth makes the endpoint reject anonymous reads. This URL is baked into the
+# published manifest, so passing it here would silently force every consumer to
+# authenticate, and the breakage only shows up long after the publish succeeds. The
+# other $(cargoSparseIndex) callers merely poll the index, so a wrong URL fails
+# loudly in the same run and needs no equivalent check.
+if ($IndexUrl -match '~force-auth') {
+    Write-Error "IndexUrl must be the plain feed URL, not the ~force-auth variant: $IndexUrl"
+    exit 1
+}
+
+$index = if ($IndexUrl -match '^sparse\+') { $IndexUrl } else { "sparse+$IndexUrl" }
 
 $path = 'mssql-mock-tds/Cargo.toml'
 $c = Get-Content $path -Raw
 $c = $c -replace 'mssql-tds\s*=\s*\{\s*path\s*=\s*"\.\./mssql-tds"',
-                 "mssql-tds = { path = `"../mssql-tds`", version = `"$Version`""
+                 "mssql-tds = { path = `"../mssql-tds`", version = `"$Version`", registry-index = `"$index`""
 Set-Content $path $c -NoNewline
 
-Write-Host "Pinned mssql-tds dependency to version $Version"
+Write-Host "Pinned mssql-tds dependency to version $Version (index: $index)"
 Select-String -Path $path -Pattern 'mssql-tds\s*=' | ForEach-Object { Write-Host $_.Line }
--- a/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
+++ b/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
@@ -37,11 +37,18 @@ $ErrorActionPreference = 'Stop'
 if ([string]::IsNullOrWhiteSpace($Version)) { Write-Error 'Version not set'; exit 1 }
 if ([string]::IsNullOrWhiteSpace($IndexUrl)) { Write-Error 'IndexUrl not set'; exit 1 }
 
-# ~force-auth makes the endpoint reject anonymous reads. This URL is baked into the
-# published manifest, so passing it here would silently force every consumer to
-# authenticate, and the breakage only shows up long after the publish succeeds. The
-# other $(cargoSparseIndex) callers merely poll the index, so a wrong URL fails
-# loudly in the same run and needs no equivalent check.
+# Cargo identifies a source by the literal URL string, so the plain and ~force-auth
+# spellings of this feed are two unrelated sources to it. Whichever one we bake here
+# is where consumers resolve mssql-tds from, and a wrong choice only surfaces long
+# after the publish succeeds -- unlike the other $(cargoSparseIndex) callers, which
+# merely poll the index and fail loudly in the same run.
+#
+# The plain URL is not free: an internal consumer whose alias points at ~force-auth
+# and who also depends on mssql-tds directly gets two copies in the graph, and a
+# confusing "expected `mssql_tds::X`, found `mssql_tds::X`" where a type crosses
+# between them. They can fix that by pointing their own alias at the plain URL.
+# Baking ~force-auth has no such escape hatch: every anonymous read of this public
+# feed would fail to resolve, which is the breakage this pinning exists to prevent.
 if ($IndexUrl -match '~force-auth') {
     Write-Error "IndexUrl must be the plain feed URL, not the ~force-auth variant: $IndexUrl"
     exit 1
@@ -53,6 +60,27 @@ $path = 'mssql-mock-tds/Cargo.toml'
 $c = Get-Content $path -Raw
 $c = $c -replace 'mssql-tds\s*=\s*\{\s*path\s*=\s*"\.\./mssql-tds"',
                  "mssql-tds = { path = `"../mssql-tds`", version = `"$Version`", registry-index = `"$index`""
+
+# -replace changes nothing when the dependency line drifts from the shape above
+# (keys reordered, workspace = true), so without this the script would exit 0 having
+# done nothing and the run would fail much later, in cargo publish -p mssql-mock-tds
+# -- after cargo-publish-mock.ps1 has already put mssql-tds@$Version on the feed.
+# Feed versions are immutable, so that burns the version and the retry cannot
+# succeed. Counting rather than merely testing for presence also catches a second
+# run appending a duplicate set of keys, which TOML rejects.
+$dep = [regex]::Match($c, '(?m)^\s*mssql-tds\s*=\s*\{[^}]*\}')
+if (-not $dep.Success) {
+    Write-Error "No mssql-tds dependency line found in $path"
+    exit 1
+}
+foreach ($key in 'version', 'registry-index') {
+    $n = ([regex]::Matches($dep.Value, "(?<![\w-])$key\s*=")).Count
+    if ($n -ne 1) {
+        Write-Error "Expected exactly one '$key' in the mssql-tds dependency of ${path}, found ${n}: $($dep.Value.Trim())"
+        exit 1
+    }
+}
+
 Set-Content $path $c -NoNewline
 
 Write-Host "Pinned mssql-tds dependency to version $Version (index: $index)"
--- a/.pipeline/OneBranch/PublishFeeds-Sandbox.yml
+++ b/.pipeline/OneBranch/PublishFeeds-Sandbox.yml
@@ -597,7 +597,7 @@ extends:
             enableJsDeps: false
         - pwsh: .pipeline/scripts/stamp-crate-versions-sandbox.ps1 -ReleaseVersion '${{ parameters.releaseVersion }}' -BuildId '$(Build.BuildId)'
           displayName: 'Stamp crate versions (sandbox, package-only)'
-        - pwsh: .pipeline/scripts/pin-mssql-tds-dep-version.ps1 -Version '$(crateVersion)'
+        - pwsh: .pipeline/scripts/pin-mssql-tds-dep-version.ps1 -Version '$(crateVersion)' -IndexUrl '$(cargoSparseIndex)'
           displayName: 'Pin mssql-tds dependency version'
         - ${{ if eq(parameters.publishCrate, true) }}:
           - pwsh: .pipeline/scripts/cargo-publish-mock.ps1 -Registry '$(cargoRegistry)' -SparseIndexBaseUrl '$(cargoSparseIndex)' -TdsVersion '$(crateVersion)'


.pipeline/scripts/pin-mssql-tds-dep-version.ps1
```diff
@@ -11,23 +11,49 @@
...
+}
+
+$index = if ($IndexUrl -match '^sparse\+') { $IndexUrl } else { "sparse+$IndexUrl" }
 
 $path = 'mssql-mock-tds/Cargo.toml'
 $c = Get-Content $path -Raw
 $c = $c -replace 'mssql-tds\s*=\s*\{\s*path\s*=\s*"\.\./mssql-tds"',
-                 "mssql-tds = { path = `"../mssql-tds`", version = `"$Version`""
+                 "mssql-tds = { path = `"../mssql-tds`", version = `"$Version`", registry-index = `"$index`""
 Set-Content $path $c -NoNewline
```
David-Engel, line 84: **Suggestion:** a non-matching regex silently succeeds here, and it fails late.

`-replace` is a no-op when the pattern doesn't match — reordered keys, a switch to `workspace = true`, or a line break after `{` would all slip through. The script still prints "Pinned…" and exits 0. The failure then surfaces inside `cargo publish -p mssql-mock-tds`, which is *after* `cargo-publish-mock.ps1:87` has already published `mssql-tds@<version>` to the feed — burning a version that `preflight-crate-versions.ps1` will then reject on the retry.

A cheap post-condition closes that:

```powershell
$c = $c -replace '...', "..."
if ($c -notmatch 'mssql-tds\s*=\s*\{[^}]*registry-index') {
    Write-Error "Failed to pin mssql-tds in $path; the dependency line did not match the expected shape."
    exit 1
}
Set-Content $path $c -NoNewline
```

It also covers the double-run case, where the pattern matches the already-pinned line and emits duplicate `version` / `registry-index` keys.

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
```diff
@@ -11,23 +11,49 @@
...
+# published manifest, so passing it here would silently force every consumer to
+# authenticate, and the breakage only shows up long after the publish succeeds. The
+# other $(cargoSparseIndex) callers merely poll the index, so a wrong URL fails
+# loudly in the same run and needs no equivalent check.
+if ($IndexUrl -match '~force-auth') {
+    Write-Error "IndexUrl must be the plain feed URL, not the ~force-auth variant: $IndexUrl"
+    exit 1
+}
+
+$index = if ($IndexUrl -match '^sparse\+') { $IndexUrl } else { "sparse+$IndexUrl" }
```
David-Engel, line 57: **Suggestion / question:** the URL baked into the manifest differs from the registry the crate is published to.

`cargoRegistry` is `mssql-rs_Public`, whose index in `.cargo/config.ci.toml:14` is the `~force-auth` variant. Cargo keys sources by URL, so a consumer who configures the `mssql-rs_Public` alias (force-auth URL) *and* depends on `mssql-tds` directly ends up with two distinct `SourceId`s for the same crate — two copies in the graph, with the confusing "expected `mssql_tds::X`, found `mssql_tds::X`" symptom.

Anonymous consumers, the case you verified, are unaffected, so this may well be the right trade. If it's intended, worth a line in the comment above noting that internal consumers shouldn't mix the alias with a dependency on `mssql-mock-tds`.

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
```diff
@@ -11,23 +11,49 @@
...
+.PARAMETER IndexUrl
+  Sparse index base URL of the feed being published to (typically
+  $(cargoSparseIndex)). Pass the plain URL, not the ~force-auth variant, so that
+  anonymous readers of a public feed can resolve the dependency.
 #>
 [CmdletBinding()]
 param(
-    [Parameter(Mandatory = $true)][string]$Version
+    [Parameter(Mandatory = $true)][string]$Version,
+    [Parameter(Mandatory = $true)][string]$IndexUrl
```
David-Engel, line 32: **Nit:** three scripts now take this same value under three names — `-IndexBaseUrl` (`preflight-crate-versions.ps1`), `-SparseIndexBaseUrl` (`cargo-publish-mock.ps1`), and `-IndexUrl` here. `-IndexBaseUrl` would match the closest sibling.

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
```diff
@@ -11,23 +11,49 @@
...
+
+# ~force-auth makes the endpoint reject anonymous reads. This URL is baked into the
+# published manifest, so passing it here would silently force every consumer to
+# authenticate, and the breakage only shows up long after the publish succeeds. The
+# other $(cargoSparseIndex) callers merely poll the index, so a wrong URL fails
+# loudly in the same run and needs no equivalent check.
+if ($IndexUrl -match '~force-auth') {
+    Write-Error "IndexUrl must be the plain feed URL, not the ~force-auth variant: $IndexUrl"
+    exit 1
+}
```
David-Engel, line 55: **Nit:** erring loudly is defensible and the comment explains why, but stripping would also work — a one-line `-replace '~force-auth', ''` makes the script correct for a caller that passes the config's index value rather than the pipeline variable. Your call; the current behavior is safe either way.

David-Engel (commented):
Correct, well-diagnosed fix. `version` without a registry makes cargo record the dep as crates.io; adding `registry-index` points consumers at the feed instead.

Verified against `.cargo/config.ci.toml` (the source replacement explains why CI never caught this), `cargo-publish-mock.ps1` (publish ordering), `PublishFeeds-Sandbox.yml:66` (`cargoSparseIndex` is the anonymous-read URL), and the passing sandbox publish job in build 170138. No blocking findings — a few suggestions inline, none of which need to hold up the merge.

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
saurabh500: Leaning toward leaving this one, unless you feel strongly. The other two append `/<index-path>/<crate>` to their value, so "base" is accurate there; this one bakes the URL whole and appends nothing, so `-IndexUrl` is what it actually is.

Renaming all three to a single name would be the real consistency fix, but that's scope creep on this PR. Happy to do it as a follow-up if you'd rather have one name across the board.

> [!NOTE]

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
saurabh500: Accurate, and it's the deliberate trade. Cargo bakes exactly one URL into the published manifest, so there's no third option:

- **plain** → breaks the case you describe (internal consumer whose alias points at `~force-auth` *and* who depends on `mssql-tds` directly), and they can fix it themselves by pointing their alias at the plain URL.
- **`~force-auth`** → breaks every anonymous read of a public feed, unfixably. That's the exact bug this PR exists to fix, with nine permanently dead versions on the feed to show for it.

Narrow-and-fixable beats total-and-permanent, so I'm keeping the plain URL. Added the constraint as a comment in `0c2573bc` so the next person doesn't have to rederive it.

> [!NOTE]

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
saurabh500: Good catch, and the consequence is a bit worse than "fails late": `cargo-publish-mock.ps1:87` publishes `mssql-tds` before `mssql-mock-tds` is even attempted, so a silent no-op burns the version on the feed and the same-BuildId retry can't succeed.

Fixed in `0c2573bc`, though not with the exact snippet. I tested `-notmatch 'mssql-tds\s*=\s*\{[^}]*registry-index'` and it doesn't cover the double-run case — the second run's output *does* contain `registry-index`, so the check stays quiet and you get the duplicate keys anyway. Went with a count instead: exactly one `version` and one `registry-index` on the dependency line, which catches both the no-op and the duplicate.

One correction on the retry mechanism: preflight only runs when `releaseVersion=true AND publishCrate=true` (`PublishFeeds-Sandbox.yml:586`), so on sandbox dev runs it never fires — the blocker there is cargo's own "already exists". Same outcome, different gate.

Verified against five manifest shapes: healthy and a line break after `{` both stay quiet (`\s` already spans newlines, so that one was never actually broken); reordered keys, `workspace = true`, and a double run all exit 1.

> [!NOTE]

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
saurabh500: Keeping the error. Stripping silently rewrites a URL the caller explicitly passed, which hides the wiring mistake instead of reporting it — and there's exactly one caller today, already passing the plain `$(cargoSparseIndex)`.

If a second caller ever turns up holding only the config's `~force-auth` value, stripping becomes the right change at that point.

> [!NOTE]

David-Engel (commented):

## Summary

Adds a required `-IndexUrl` to `pin-mssql-tds-dep-version.ps1` so the published `mssql-mock-tds` manifest carries `registry-index` pointing at the feed instead of silently defaulting to crates.io, plus a post-rewrite guard that fails the step when the dependency line has drifted or has already been pinned. Linked to #389. I could not find a defect — the reasoning in the comments matches what the code actually does, and the two riskiest behaviors (fail-closed on drift, refuse `~force-auth`) both hold up under test.

## Verification

Executed, not just read. In a throwaway worktree at `0c2573b`, ran the script against the real `mssql-mock-tds/Cargo.toml` with `pwsh -NoProfile -File .pipeline/scripts/pin-mssql-tds-dep-version.ps1 -Version 0.1.0-dev.1 -IndexUrl 'https://pkgs.dev.azure.com/.../Cargo/index/'`:

| Case | Result |
| --- | --- |
| Happy path | exit 0, emits `path` + `version` + `registry-index` + `default-features`, keys in a valid TOML inline table |
| Run twice | exit 1, `found 2` for `version`, and importantly the file on disk is left at the correct single-pinned state — the guard runs before `Set-Content` |
| `~force-auth` URL | exit 1, `Cargo.toml` byte-identical to the original |
| Drifted dep line (`{ default-features = false, path = "../mssql-tds" }`) | exit 1, `found 0` — this is the silent-no-op case the guard exists for, and it fails closed as intended |

I also checked the real feed's sparse index for `mssql-mock-tds`, which is the end-to-end proof this actually fixes #389:

```
0.1.0-dev.20260826.169925   registry=https://github.com/rust-lang/crates.io-index
0.1.0-dev.20260826.170138   registry=sparse+https://pkgs.dev.azure.com/sqlclientdrivers/public/_packaging/mssql-rs_Public/Cargo/index/
0.1.0-dev.20260827.170165   registry=https://github.com/rust-lang/crates.io-index
0.1.0-dev.20260827.170337   registry=sparse+https://pkgs.dev.azure.com/.../Cargo/index/
0.1.0-dev.20260827.170483   registry=sparse+https://pkgs.dev.azure.com/.../Cargo/index/
```

Every version published from a build carrying this change records the feed; the interleaved `170165` crates.io entry is a build off unpatched `main`, which is consistent. No Rust changed, so I did not run `bfmt`/`bclippy`/`btest`, and CI is green including the sandbox `cargo publish mssql-mock-tds` job.

## Findings

**Blocking** — none found.

**Suggestion** — none found.

**Nit** — none found.

The ordering of the guard relative to `Set-Content` deserves a note of approval rather than a finding: validating before the write is what makes a failed run retryable, which matters a lot here given feed versions are immutable and a burned version cannot be republished. The `~force-auth` rejection is also the right call for a public feed, and the comment explaining the asymmetry with the other `$(cargoSparseIndex)` callers is genuinely useful context rather than noise.

This PR looks ready for human review. To be explicit: this is not an approval.

David-Engel (approved):

Issue #393, opened 2026-08-26
Investigate cross-compiling the Windows ARM64 mock wheel from the x64 agent


saurabh500:
Follow-up to #392, which merged the Linux libc-flavour pairs and took the sandbox publish pipeline's concurrent build agents from 7 to 5.

## Idea

`BuildWindows_arm64` could potentially be folded into `BuildWindows_x64` by cross-compiling with `maturin build --target aarch64-pc-windows-msvc`. That would take peak concurrent build agents from **5 to 4** and free one more `RUST-ARM64-WUS3` agent.

It would also delete the two Node.js bootstrap steps that only the ARM64 Windows agent needs:

```yaml
- pwsh: ./scripts/setup-nodejs-path.ps1
  displayName: 'Add Node.js to PATH for ADO tasks'
- pwsh: ./scripts/install-nodejs-to-agent.ps1
  displayName: 'Install Node.js to agent externals folder'
```

## Why it looks plausible

- `mssql-mock-tds` pulls the `openssl` crate only on `cfg(not(windows))`, and `native-tls` uses SChannel on Windows. So there is no C library to cross-build — the usual blocker for this kind of change is absent.
- pyo3 is configured `abi3-py39`. abi3 cross-compilation does not need a target-architecture interpreter, only `PYO3_CROSS_PYTHON_VERSION`.

## Why it was not done in #392

Two things cannot be settled by reading the repo:

1. **Does the `RUST-W22-SQL25` image have the MSVC ARM64 cross tools (`VC.Tools.ARM64`)?** Without them the link step fails, and that only surfaces in a real pipeline run.
2. **Does the cross-built wheel actually load on Windows ARM64?** Building natively today gives us an accidental smoke test that cross-compiling removes. Losing it silently would be worse than the agent saving is worth.

Both are answerable, but they need runs of a pipeline that has publish side effects. Bundling that speculation into #392 would have put a clean two-agent win at risk of a red run.

## Suggested approach

1. Confirm `VC.Tools.ARM64` is present on the `RUST-W22-SQL25` image (or get it added).
2. Prototype the cross-build on a branch and confirm the wheel's platform tag is `win_arm64`.
3. Verify the resulting wheel imports and runs on a real Windows ARM64 machine before deleting the native job.
4. Only then merge the jobs.

If step 1 or 3 does not pan out, close this — keeping the native ARM64 job is a perfectly reasonable outcome.

PR #394, opened 2026-08-26
Build both Linux libc flavours per architecture on one agent

### LINKED ISSUE #392: Sandbox publish pipeline requests 7 concurrent agents and starves the RUST pools
## Problem

`.pipeline/OneBranch/PublishFeeds-Sandbox.yml` ("GH-Sandbox Release", definition 2293) fans out into **seven concurrent build jobs** after `PreflightVersionCheck`. Six of those want self-hosted RUST pool agents — three on `RUST-X64-WUS3` and three on `RUST-ARM64-WUS3`.

The pools cannot satisfy that burst, so runs sit in queue waiting for agents.

## Evidence

Run [169925](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=169925) — 112 minutes wall clock:

- `PreflightVersionCheck` finished around 06:54
- The first build job did not produce output until ~07:40

That is roughly **46 minutes of pure queue wait**, longer than any individual build step in the run. The bottleneck is agent availability, not build speed.

## Opportunity

The four Linux jobs (`BuildLinux_x64`, `BuildLinux_arm64`, `BuildLinuxMusl_x64`, `BuildLinuxMusl_arm64`) all call one script, `.pipeline/scripts/build-mock-wheel-linux-host.sh`, which is already fully parameterized by `CONTAINER_IMAGE`, `OB_OUTPUTDIRECTORY`, `BUILD_SOURCESDIRECTORY` and `MATURIN_FEATURES`.

For a given architecture, the manylinux and musllinux jobs target the **same pool**, the **same demands**, and run **identical setup steps** (cargo auth, rustup, DockerInstaller, ubuntu deps, version stamp). Only the container image and the features string differ.

Merging each same-architecture pair into one job that runs both container builds takes concurrent build agents from **7 to 5**, and self-hosted ones from **6 to 4**.

| Pool | Current | Proposed |
|---|---|---|
| `RUST-X64-WUS3` (self-hosted) | 3 | 2 |
| `RUST-ARM64-WUS3` (self-hosted) | 3 | 2 |
| hosted macOS-14 | 1 | 1 |
| **Total** | **7** | **5** |

Wheel output is unchanged: still seven wheels, now from five build jobs.

## Trade-off

The two container builds in a merged job run serially instead of side by side, offset by dropping one duplicated agent setup pass per merged pair. Against ~46 minutes of queue wait, trading a few minutes of serial build time for a materially shorter queue is worth it whenever the pools are contended.

## Note

A `strategy: matrix` would express the fan-out more compactly but spawns one agent per leg, so it does not reduce agent count at all. Only merging work into a single job does.

saurabh500:
## Description

The sandbox publish pipeline fanned out into **seven concurrent build jobs**, six of which wanted self-hosted RUST pool agents. The pools could not satisfy that burst — run [169925](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=169925) spent roughly **46 minutes waiting for an agent** after preflight finished, longer than any individual build step in the run.

The manylinux and musllinux jobs for a given architecture already wanted the same pool, the same demands and the same setup steps, and they call one build script parameterized by container image. This merges each same-architecture pair into a single job that runs both container builds.

| Pool | Before | After |
|---|---|---|
| `RUST-X64-WUS3` (self-hosted) | 3 | 2 |
| `RUST-ARM64-WUS3` (self-hosted) | 3 | 2 |
| hosted macOS-14 | 1 | 1 |
| **Total concurrent build agents** | **7** | **5** |

Self-hosted demand drops by a third. Wheel output is unchanged: still seven wheels, now from five build jobs.

### The check that would have gone blind

`build-mock-wheel-linux-host.sh` ended with `find … -name '*.whl' | wc -l` and failed only on zero. With two invocations in one job, the second would count the **first** invocation's wheels and pass even if it produced nothing — a check that cannot fail.

It now takes a required `EXPECTED_PLATFORM_TAG` and asserts a wheel matching that tag exists. Required rather than optional, so a future caller cannot silently opt out. Each flavour also writes to its own output subdirectory, so the two builds cannot see each other's wheels at all.

### Deviation from the original plan

The plan called for two published artifacts per merged job, keeping the existing artifact names. previewRun rejected that: the governed Linux job template permits exactly one artifact per job, named from `ob_artifactSuffix`. Both flavours are now gathered into one artifact directory before publishing. `collect-wheels.ps1` recurses for `*.whl`, so `UploadPython` is unaffected.

### Trade-off

The two container builds run serially instead of side by side, offset by dropping one duplicated agent setup pass per merged pair. Against ~46 minutes of queue wait, that is worth it whenever the pools are contended.

A `strategy: matrix` was considered and rejected — it spawns one agent per leg, so it reduces nothing.

## Related Issues

Fixes #392

Follow-up tracked separately in #393 (cross-compiling the Windows ARM64 wheel from the x64 agent, which would take peak to 4). Deliberately not attempted here: it needs a real run to confirm the `RUST-W22-SQL25` image has `VC.Tools.ARM64`, and cross-compiling removes the accidental smoke test that building natively gives us.

## Validation

**What I actually exercised:**

- **The platform-tag assertion**, against the real script with the docker wrapper stubbed. Six cases, all as expected:

  | Case | Expect | Result |
  |---|---|---|
  | healthy manylinux build | 0 | ✅ 0 |
  | healthy musllinux build | 0 | ✅ 0 |
  | **musllinux expected, build produced nothing, manylinux wheel already present** | non-zero | ✅ 1 |
  | build produced nothing, dir empty | non-zero | ✅ 1 |
  | second build succeeds alongside first flavour's wheel | 0 | ✅ 0 |
  | `EXPECTED_PLATFORM_TAG` unset | non-zero | ✅ 1 |

  I broke the mechanism rather than just reverting the fix: on the third case the **old** count-based check passes (blind) while the new one fails. That is precisely the blind spot merging introduces. The healthy cases confirm it does not fire on good input.

- **`previewRun` on definition 2293** against this branch — returns `runs/-1` / `buildId=-1`, so no real run was created. The expanded YAML confirms all four Linux builds are present with correct tags, and `UploadPython`'s `dependsOn` resolves to the five surviving job names with no stale entries. The first previewRun attempt is what caught the one-artifact-per-job constraint.

- **Caller audit.** Grepped the whole repo for `build-mock-wheel-linux-host` — exactly four call sites, all in the two merged jobs, all pass `EXPECTED_PLATFORM_TAG`. No caller missed by making it required.

**What I did not exercise:** the container builds themselves. A green previewRun proves the YAML expands and says nothing about whether the wheels build. This pipeline has publish side effects, so I did not queue a real run. First real run should be watched for the serial-build wall clock and for disk headroom, since a merged agent now holds two cargo target trees (gnu and musl land in different subdirectories, so they do not collide, but they do both occupy disk).

## Checklist

- [x] `cargo bfmt` passes — no Rust changed; pre-commit hook ran clean
- [x] `cargo bclippy` passes — no Rust changed
- [x] `cargo btest` passes — no Rust changed
- [x] New/changed functionality has tests — shell assertion exercised in all six cases above, including the deliberately-broken one
- [x] Public API changes are documented — none

---

## Update: platform-tag assertion corrected after run 170165

The first real run of this branch (170165) failed in the merged Linux ARM64 job. The container
build succeeded and produced a wheel, then the new assertion rejected it:

```
ERROR: No mock wheel matching platform tag 'manylinux' in /mnt/vss/_work/1/a/wheels
```

Cause: `scripts/build-mock-python-wheel-in-container.sh` passes `--auditwheel skip`, so maturin
never retags glibc wheels as `manylinux_*`. They keep the default `linux_<arch>` tag. The
assertion expected a string that never appears, so it fired on healthy output. The x64 job had the
same defect and would have failed too; ARM64 just got there first.

This was my error, and it is worth naming how it got through: I validated the assertion against
wheel filenames I made up, rather than against real maturin output. The test double did not
reproduce real behaviour, so the passing case proved nothing.

Fixed by asserting on the tags actually emitted:

| Job | Container | Expected tag |
|---|---|---|
| `BuildLinux_x64` | manylinux_2_34_x86_64 | `linux_x86_64` |
| `BuildLinux_x64` | musllinux_1_2_x86_64 | `musllinux_1_2_x86_64` |
| `BuildLinux_arm64` | manylinux_2_34_aarch64 | `linux_aarch64` |
| `BuildLinux_arm64` | musllinux_1_2_aarch64 | `musllinux_1_2_aarch64` |

The two tags are mutually exclusive — `musllinux_1_2_x86_64` does not contain `linux_x86_64`
(the `linux` in it is followed by `_1_2_`) — so a musl wheel cannot satisfy the glibc check. That
keeps the property the assertion exists for: the second invocation in a merged job still cannot
pass on the first invocation's wheels.

Verified against the real filenames recorded in runs 170138 and 170165, in both directions:

| Case | Result |
|---|---|
| healthy manylinux vs `linux_x86_64` | exit 0 |
| healthy musllinux vs `musllinux_1_2_x86_64` | exit 0 |
| only manylinux present, musl expected | exit 1 |
| only musllinux present, manylinux expected | exit 1 |
| empty directory | exit 1 |

`previewRun` on definition 2293 re-validated against this branch: expands clean, `UploadPython`
still lists the five merged job names.

Note that `previewRun` cannot catch this class of bug at all — it proves YAML expansion and nothing
about what the containers produce. Only a real run confirms the fix.


.pipeline/OneBranch/PublishFeeds-Sandbox.yml
```diff
@@ -172,26 +188,45 @@ extends:
...
+          displayName: 'Build mock abi3 wheel (manylinux_2_34 x86_64)'
           env:
             CONTAINER_IMAGE: ghcr.io/microsoft/mssql-rs/python-build/manylinux_2_34_x86_64_rust:latest
-            OB_OUTPUTDIRECTORY: $(ob_outputDirectory)
+            EXPECTED_PLATFORM_TAG: linux_x86_64
+            OB_OUTPUTDIRECTORY: $(ob_outputDirectory)/manylinux
             BUILD_SOURCESDIRECTORY: $(Build.SourcesDirectory)
             CARGO_REGISTRIES_MSSQL_RS_PUBLIC_TOKEN: $(CARGO_REGISTRIES_MSSQL_RS_PUBLIC_TOKEN)
             CARGO_REGISTRIES_MSSQL_RS_TOKEN: $(CARGO_REGISTRIES_MSSQL_RS_TOKEN)
+        - script: bash .pipeline/scripts/build-mock-wheel-linux-host.sh
```
David-Engel, line 199: The description says "gnu and musl land in different subdirectories, so they do not collide, but they do both occupy disk" — the second half is right, the first half isn't, and it's worth correcting because it's the premise of the disk-headroom watch item.

Neither invocation passes `--target`: `scripts/build-mock-python-wheel-in-container.sh` calls `maturin build --release` with `--manifest-path`/`--out` only, so cargo uses the host triple and writes straight into `target/release`, not `target/<triple>/release`. And `CARGO_TARGET_DIR` (set at job level to `$(Build.SourcesDirectory)`) never reaches the container — `docker-cargo-run.sh` forwards only the four `CARGO_REGISTRIES_*` variables plus whatever `-e` flags the host wrapper passes (`WORKSPACE_DIR`, `OUTPUT_DIR`, optional `MATURIN_FEATURES`). So both flavours share `/workspace/mssql-mock-tds-py/target`.

Nothing is corrupted by this — unit filenames are keyed by a metadata hash that includes the target triple, so gnu and musl units coexist, and the builds are serial. The practical consequences are: the musl build gets no cache reuse and recompiles everything including vendored OpenSSL (expected, but it's the cost side of the wall-clock trade-off you asked to watch), stale gnu artifacts accumulate in the same tree on a non-ephemeral agent, and the uplifted `target/release/libmssql_mock_tds_py.so` is overwritten in place by the second build. Also note the host-side `sudo chown -R "${BUILD_SOURCESDIRECTORY}/target"` in the wrapper doesn't cover `mssql-mock-tds-py/target`, so those root-owned artifacts stay root-owned between runs.

If you want the separation the description assumes, one `-e` per invocation gets it:

```yaml
    CARGO_TARGET_DIR_IN_CONTAINER: /workspace/mssql-mock-tds-py/target/musllinux
```

forwarded as `-e CARGO_TARGET_DIR=...` from the wrapper. That makes per-flavour cleanup possible and keeps `docker image prune` from being the only disk lever. Fine to leave as-is and just fix the description instead — but pick one.

.pipeline/OneBranch/PublishFeeds-Sandbox.yml
```diff
@@ -172,26 +188,45 @@ extends:
...
+            OB_OUTPUTDIRECTORY: $(ob_outputDirectory)/musllinux
+            BUILD_SOURCESDIRECTORY: $(Build.SourcesDirectory)
+            CARGO_REGISTRIES_MSSQL_RS_PUBLIC_TOKEN: $(CARGO_REGISTRIES_MSSQL_RS_PUBLIC_TOKEN)
+            CARGO_REGISTRIES_MSSQL_RS_TOKEN: $(CARGO_REGISTRIES_MSSQL_RS_TOKEN)
+        - script: |
+            set -euo pipefail
+            mkdir -p "$(ob_outputDirectory)/wheels"
+            cp "$(ob_outputDirectory)"/manylinux/wheels/*.whl "$(ob_outputDirectory)/wheels/"
+            cp "$(ob_outputDirectory)"/musllinux/wheels/*.whl "$(ob_outputDirectory)/wheels/"
+            ls -lh "$(ob_outputDirectory)/wheels"
```
David-Engel, line 214: After this step the staging directory holds each wheel twice — once under `manylinux/wheels` or `musllinux/wheels` and once under `wheels/`. The explicit `PublishPipelineArtifact` below targets `$(ob_outputDirectory)/wheels` only, so the named artifact is clean; but `ob_artifactSuffix` is set on this job, which is the OneBranch signal to publish `$(ob_outputDirectory)` itself, and that one would now carry the nested duplicates.

No functional impact — `collect-wheels.ps1` recurses and copies by filename into `dist`, so identical names just overwrite — but it triples the bytes for no benefit and makes the artifact layout harder to read. A cleanup line keeps the isolation and drops the duplicates:

```bash
            cp "$(ob_outputDirectory)"/musllinux/wheels/*.whl "$(ob_outputDirectory)/wheels/"
            rm -rf "$(ob_outputDirectory)/manylinux" "$(ob_outputDirectory)/musllinux"
            ls -lh "$(ob_outputDirectory)/wheels"
```

If you'd rather keep the per-flavour directories for post-mortem debugging of a failed run, say so in the comment above and leave it — the wheels are small enough that this is a preference, not a defect.

David-Engel (commented):
## Summary

Merges the two per-architecture Linux libc jobs into one job each, taking the sandbox publish stage from seven concurrent build agents to five, and replaces the count-based wheel check with a platform-tag assertion so the second invocation in a merged job can still fail. The mechanism is right and the blind-spot analysis in the description is the best part of this PR. No blocking findings; two suggestions and a doc nit below.

**What I verified**

- **Tag exclusivity, run locally** against realistically-shaped filenames (`mssql_mock_tds-0.1.0-cp39-abi3-{linux_x86_64,musllinux_1_2_x86_64,linux_aarch64,musllinux_1_2_aarch64}.whl`): each of the four `EXPECTED_PLATFORM_TAG` values matched exactly one file. `*linux_x86_64*` does not match `musllinux_1_2_x86_64`, so the glibc check cannot be satisfied by a musl wheel — the property the assertion exists for holds. The per-flavour output subdirectories make it belt-and-braces, since neither invocation can see the other's wheels at all.
- **Caller audit.** `git grep build-mock-wheel-linux-host` → exactly the four sites at `PublishFeeds-Sandbox.yml:190,199,260,269`, all passing `EXPECTED_PLATFORM_TAG`. Making it required strands no caller.
- **No stale references.** `git grep -E "BuildLinuxMusl|mock_linux_musl"` returns nothing; `UploadPython.dependsOn` lists exactly the five surviving jobs; `collect-wheels.ps1` recurses for `*.whl` and flattens by filename into `dist`, so the artifact reshaping is invisible to the upload step.
- **Agent images are unchanged per flavour.** The merged x64 job keeps `RUST-UBUSLIM` (what the removed musl x64 job demanded) and the arm64 job keeps `RUST-UBUNTU-ARM64` — merging doesn't move either build onto a different image.
- **CI.** `gh pr checks 394` is all green on `267e6ed`, including both merged ADO Linux build jobs and the cross-repo mssql-python legs. Coverage report is present, 100% diff / 92.6% overall, and reports "no lines with coverage information" as expected for a YAML+shell diff.

### Suggestions

Two inline, on the x64 job; both apply identically to the ARM64 job (`:269`, `:281-284`).

### Nit

**Script header is now slightly stale (`.pipeline/scripts/build-mock-wheel-linux-host.sh:3-4`)** — "then verifies at least one wheel landed in the output directory" describes the old count check. It now verifies a wheel *matching the expected platform tag* landed, which is the whole point of the change. One-line fix, outside the diff hunks so I can't anchor it inline.

### On validation

The description's evidence for the corrected assertion is `previewRun` plus real filenames recorded in runs 170138/170165 — no green real run of the fixed code is cited, and as you note `previewRun` cannot exercise this class of bug at all. That's a reasonable place to stop given the publish side effects, but it means the first real run after merge is the actual verification for all four legs, not just the ARM64 one that failed before. Worth watching that run for the tag assertion on each leg and for the serial wall clock.

---

*Automated review generated on behalf of @David-Engel. Not an approval and it does not satisfy the human review requirement; findings may be incomplete or wrong.*

.pipeline/OneBranch/PublishFeeds-Sandbox.yml
```diff
@@ -136,11 +139,24 @@ extends:
...
+      # musllinux is built with vendored (static) OpenSSL so the wheel is
+      # self-contained and does not depend on Alpine's libssl at the consumer
+      # side. manylinux leaves MATURIN_FEATURES unset and links the OS libssl.
+      #
+      # Each flavour writes to its own output subdirectory so the two builds
+      # cannot see each other's wheels. The governed Linux job template allows
+      # exactly one published artifact per job, named from ob_artifactSuffix,
+      # so both flavours are gathered into one artifact directory at the end.
       #####################################################################
       - job: BuildLinux_x64
```
David-Engel, line 157: This job now runs two container builds serially where it previously ran one, and it has no `timeoutInMinutes`. Azure Pipelines applies a 60-minute default when a job doesn't set one, so if the OneBranch Linux template doesn't inject its own, the merged job's budget didn't grow with its workload.

The ARM64 job (`:227`) is the one to worry about: it runs the same two builds on `RUST-ARM64-WUS3`, and the musl leg vendors OpenSSL (`MATURIN_FEATURES: vendored-openssl`), which is a from-scratch OpenSSL plus crate build with no cache reuse from the glibc leg — see the shared-target-tree thread at `:199`. A server-side cancellation at the cap also skips `condition: always()` steps, so you'd lose the diagnostics along with both wheels.

This repo already backstops long jobs explicitly rather than relying on the default — `templates/test-mssql-python-odbc-template.yml:260` (130), `templates/validation-stages.yml:618` (150), `templates/sql-host-template.yml:70` (240), `sync-container-images.yml:858` (120). Same treatment here, on both merged jobs:

```yaml
      - job: BuildLinux_x64
        dependsOn: PreflightVersionCheck
        timeoutInMinutes: 120
```

If you've confirmed from a real run that the governed template already sets something generous, say so in the job comment and drop this — I can't read the expanded template from here, so this is a question about a default I can't observe, not an assertion that the job will time out.

.pipeline/OneBranch/PublishFeeds-Sandbox.yml
```diff
@@ -172,26 +188,45 @@ extends:
...
+            cp "$(ob_outputDirectory)"/manylinux/wheels/*.whl "$(ob_outputDirectory)/wheels/"
+            cp "$(ob_outputDirectory)"/musllinux/wheels/*.whl "$(ob_outputDirectory)/wheels/"
+            ls -lh "$(ob_outputDirectory)/wheels"
+          displayName: 'Gather both libc flavours into one artifact directory'
         - task: PublishPipelineArtifact@1
-          displayName: 'Publish Linux x64 mock wheel'
+          displayName: 'Publish Linux x64 mock wheels'
           inputs:
             targetPath: '$(ob_outputDirectory)/wheels'
             artifact: 'drop_PublishMockPython_BuildLinux_x64_mock_linux_x64'
```
David-Engel, line 220: Nit, naming only. This artifact now carries both libc flavours, but the name still reads as glibc-only, and the `_mock_linux_musl_x64` / `_mock_linux_musl_arm64` artifacts that used to hold the musl wheels are gone. Same on the ARM64 job (`:290`).

Nothing in this repo looks wheels up by artifact name — `UploadPython` uses `download: current` and `collect-wheels.ps1` recurses — and I confirmed no remaining references with `git grep -E "mock_linux_musl|BuildLinuxMusl"`. So this only bites something outside the repo that downloads a named artifact, or a human reading the artifact list and concluding the musl wheels stopped building. Both are cheap to prevent: keep the name (renaming would break the `ob_artifactSuffix` correspondence for no gain) and note the contents in the publish step's `displayName`, e.g. `'Publish Linux x64 mock wheels (manylinux + musllinux)'`. The job-level comment already explains the one-artifact-per-job constraint; this just makes it visible in the run UI.

David-Engel (commented):
## Summary

Second automated pass over the same head (`267e6ed`). The mechanism still looks right: two per-architecture Linux jobs each build both libc flavours serially, taking the sandbox publish stage from seven concurrent build agents to five, and the count-based wheel check is replaced by a platform-tag assertion so the second invocation in a merged job can still fail. No blocking findings. The two existing unresolved threads (shared cargo target tree vs. the description's "different subdirectories" claim, and the duplicated wheels left under `manylinux/`/`musllinux/` in the staging root) still apply as written and are deliberately not re-filed here. Two new findings below, both about consequences of merging that neither the description nor the earlier pass covered.

**What I verified this pass**

- **Tag exclusivity, re-derived locally** with `find -type f -name "*${TAG}*.whl"` over the four realistic filenames (`mssql_mock_tds-0.1.0-cp39-abi3-{linux_x86_64,musllinux_1_2_x86_64,linux_aarch64,musllinux_1_2_aarch64}.whl`). Each of the four `EXPECTED_PLATFORM_TAG` values matched exactly one file; `*linux_x86_64*` does not match `musllinux_1_2_x86_64`, and `*linux_aarch64*` does not match `musllinux_1_2_aarch64`. The property the assertion exists for holds, and the per-flavour output subdirectories make it belt-and-braces.
- **Caller audit.** `git grep -E "BuildLinuxMusl|mock_linux_musl|build-mock-wheel-linux-host"` returns exactly the four call sites at `PublishFeeds-Sandbox.yml:190,199,260,269` and no stale references to the removed jobs or their artifact names. Making `EXPECTED_PLATFORM_TAG` required strands no caller.
- **Downstream reshaping is invisible.** `UploadPython.dependsOn` lists exactly the five surviving job names, `download: current` pulls everything, and `collect-wheels.ps1` recurses for `*.whl` and flattens by filename into `dist`, so identical names overwrite rather than collide.
- **Images unchanged per flavour.** The merged x64 job keeps `RUST-UBUSLIM` (what the deleted musl x64 job demanded) and the arm64 job keeps `RUST-UBUNTU-ARM64`; merging doesn't move either build onto a different image.
- **CI.** `gh pr checks 394` is green on `267e6ed`, including both merged ADO Linux legs and the cross-repo mssql-python suites. Coverage report is present (100% diff / 92.6% overall) and reports "no lines with coverage information", which is expected for a YAML + shell diff.

### Suggestions

Two inline, both on the x64 job; both apply identically to `BuildLinux_arm64`.

### Failure blast radius is now per-architecture, not per-flavour

Worth naming explicitly in the description, since it is the second cost of merging and it isn't in the trade-off section. Before, a musl failure lost only the musl wheel and the manylinux artifact still published. Now a failure in either container build aborts the job before the gather step, so a healthy wheel that already built is never published for that architecture — exactly what happened in run 170165, where the ARM64 manylinux build succeeded and the job still produced nothing. Failing the whole architecture is the right behavior for a publish pipeline (a partial wheel set shouldn't reach the feed, and `UploadPython`'s implicit `succeeded()` would block anyway), so this isn't a defect. But the retry cost changed with it: re-running the job after a transient docker or registry failure in the second build now rebuilds the first flavour too, which is the slow half of the wall-clock trade-off you asked to watch. One sentence under "Trade-off" covers it.

### On validation

Unchanged from the previous pass and still the main caveat: `previewRun` proves YAML expansion and nothing about what the containers produce, and the corrected assertion's evidence is filenames recorded in runs 170138/170165 rather than a green real run of the fixed code. That is a reasonable place to stop given the publish side effects, but the first real run after merge is the actual verification for all four legs — watch the tag assertion on each leg, the serial wall clock, and disk on the merged agents.

---

*Automated review generated on behalf of @David-Engel. Not an approval and it does not satisfy the human review requirement; findings may be incomplete or wrong.*

PR #402, opened 2026-08-27
Experiment: Sqlcmd 

shiwanigupta0809:
## Checklist

- [ ] `cargo bfmt` passes
- [ ] `cargo bclippy` passes
- [ ] `cargo btest` passes


shiwanigupta0809:

David-Engel (commented):
## Summary

Adds a new `mssql-sqlcmd` crate (~19k lines, 48 `.rs` files) reimplementing both the ODBC `sqlcmd` and the Go "modern" `sqlcmd` on top of `mssql-tds`, plus supporting changes to `mssql-tds` (a `login_server_name` LOGIN7 override, deferred batch errors, per-DONE row counts, and five new Entra auth methods). The layering and the `mssql-tds` additions are well-designed — `test_login_server_name.rs` in particular is a genuinely good test, using `mssql-mock-tds` to read ServerName back off the wire and covering the longer/shorter-than-dialled-address cases that the offset/length pair makes easy to get wrong. Findings below are mostly in the new crate: the differential test suites that carry the crate's entire correctness claim are no-ops in CI, and there are four security-relevant defaults worth fixing.

This is a draft marked "Experiment", so several of these are naturally follow-up work rather than merge gates — flagged at the severity they'd carry at merge time.

**What I verified locally** (detached worktree at `70e5517c`, not the main checkout): `cargo fmt -- --check` passes; `cargo clippy --workspace --all-features --all-targets -- -D warnings` passes clean; `cargo clippy -p mssql-sqlcmd --all-targets` compiles clean. So the ADO build failures are not fmt/clippy/compile on Windows. I could not read the ADO logs without interactive auth, so I did not diagnose them further.

**Checked and deliberately not flagging** — recording these so they don't get re-raised:
- **Encryption defaults are correct.** `connect.rs:174` defaults to `EncryptionSetting::On` and `validate.rs:153` defaults `trust_server_certificate: false`, matching driver-18 behavior. This was my highest-priority check and it's right.
- **`Cargo.lock` is not a finding** — it's gitignored (`.gitignore:27`) and untracked in this repo, so "the lockfile wasn't updated" does not apply here.
- **License headers**: all 48 new `.rs` files carry the Microsoft/MIT header.
- **Entra authority trust** (`entra.rs:154`, `split_sts_url`): taking the FEDAUTHINFO authority with an `https`-only check is deliberate, documented at `entra.rs:22-28` as matching msodbcsql and the Azure SDK, and has a test asserting plaintext is refused. That's the reference behavior, so it's not a defect.
- The `.len() as i16` char-count pattern in `write_server_name` is the pre-existing convention for every LOGIN7 field, not something this PR introduces.

## Blocking

**1. The differential suites pass without testing anything.** `mssql-sqlcmd/tests/diff.rs:22-25` returns early when `SQLCMD_DIFF_REF` is unset, so the test succeeds. The same shape is at `go_diff.rs:33-35` and `:42-44`, `modern_diff.rs:178-179` and `:200-201`, `diff/runner.rs:175-176`, `color_gate.rs:69`/`:83`/`:94`, and eight sites in `go_features.rs` (`178`, `188`, `198`, `221`, `234`, `244`, `264`, `281`). Every parity claim in `PARITY.md` rests on these, and in CI — where no reference binary, server, or container exists — none of them execute while `cargo btest` reports green. The inline `#[cfg(test)]` unit tests in `validate.rs`/`dsn.rs` do run, so coverage isn't zero, but the parity surface is entirely unexercised. The runners themselves are well built (`diff.rs:36` asserts cases are non-empty, failures are collected and asserted); the problem is only that nothing turns them on. Suggested fix: mark them `#[ignore]` with a documented opt-in so their skipping is visible, and add a CI-set env var (e.g. `SQLCMD_DIFF_REQUIRED=1`) that makes a missing prerequisite a hard failure rather than a silent return.

**2. The password prompt echoes in cleartext on Linux and macOS.** `mssql-sqlcmd/src/session.rs:1006-1017`: the `#[cfg(windows)]` arm clears `ENABLE_ECHO_INPUT` around the read, but the `#[cfg(not(windows))]` arm is a plain `stdin().lock().read_line()`. So `sqlcmd -U user` without `-P` prints the password to the terminal on non-Windows, where it also lands in the scrollback. Suggested fix: use a termios `ECHO`-clearing implementation (or the `rpassword` crate) for the Unix arm; the doc comment on the Windows arm already states the intended contract for both.

**3. An unrecognized `-N` suffix silently selects the weakest encryption.** `mssql-sqlcmd/src/cli/validate.rs:375-380` maps `Some(_) => Encrypt::Optional`, so `-NS`, `-NY` or any typo resolves to `EncryptionSetting::PreferOff` instead of erroring. A user typing the uppercase `-NS` intending strict gets optional encryption with no diagnostic — a fail-open on a security control. Suggested fix: return the usual invalid-argument error for any suffix outside `s`/`m`/`o`, the way the other suffixed options are validated.

**4. `sqlconfig` writes credentials with default file permissions.** `mssql-sqlcmd/src/modern/sqlconfig.rs:92` uses `std::fs::write`, which creates the file `0666 & !umask` — typically `0644`, i.e. world-readable — and the stored password is base64 (`config_cmds::base64::decode_text` at `server_cmds.rs:47`), which is encoding, not protection. Suggested fix: create the file with `OpenOptions` plus `.mode(0o600)` on Unix before writing, and consider the OS credential store for the secret itself.

## Suggestions

**5. ~43 lines of protocol rationale are deleted from `tds_client.rs` for no stated reason.** The diff removes explanatory comments unrelated to this feature, including: the `drain_stream` comment explaining why `ParserContext::ColumnEncryption` rather than `None` must be used (the CEK-table prefix would otherwise desynchronize the stream — the exact corruption that drain exists to prevent); the batch-state-reset rationale in both ERROR arms (`advance_to_result_boundary` and `get_next_row`); the `retire_after_failed_drain` note about not returning an abandoned stream to a pool; the `dml_result_counts` doc linking it to `take_dml_result_counts` and the ODBC result-set mapping; and three blocks in `begin_command`. These explain *why*, which is what the repo's convention asks doc comments to do — the "no AI slop" rule targets comments restating *what* the code does, so it doesn't justify removing these. Suggested fix: restore them, or split the removal into its own PR where it can be judged on its merits.

**6. `unreported_error` is a `bool` where the pairing is one-to-one.** `tds_client.rs`: each deferred ERROR sets the flag and `consumed_pending_error()` takes it. Two failure modes follow from it not being a counter. If a deferred error's closing DONE does not set `DONE_ERROR`, the flag stays set and silently swallows the next genuinely unpaired `DONE_ERROR` — weakening the protocol check. Conversely, if two ERROR tokens arrive before their two DONEs, the second DONE raises `ProtocolError("Server reported error in DONE token without preceding ERROR token")` and retires a healthy connection. I did not construct a server response that produces the second ordering, so treat this as a concern rather than a confirmed bug — but a `usize` counter with saturating decrement removes both cases at no cost.

**7. `sqlcmd query` disables certificate validation for every context.** `mssql-sqlcmd/src/modern/server_cmds.rs:33-35` unconditionally appends `-C`, justified by the comment "A container on localhost presents a self-signed certificate." But `endpoint` comes from `context.config.current()`, which can name any host, so a context pointing at a remote server also gets certificate validation turned off — with stored credentials attached. Suggested fix: emit `-C` only when `endpoint.container.is_some()` (the field already distinguishes tool-created containers) or the address is loopback. If go-sqlcmd genuinely does this unconditionally, a comment citing that would settle it.

**8. The SA password is passed in the container runtime's argv.** `mssql-sqlcmd/src/modern/container.rs:82` passes `MSSQL_SA_PASSWORD=<password>` as a Docker/Podman argument, so it's visible to any local process listing. `Command::env(...)` avoids this.

**9. PR metadata.** The description is still the unedited template — no summary, and no linked GitHub issue or ADO work item, which the template at `.github/PULL_REQUEST_TEMPLATE.md` states is required. The checklist is unchecked and the branch currently conflicts with `main`. Worth resolving before this leaves draft, along with a note in the description about the `mssql-tds` comment removals in finding 5.

## Questions

**10. `-X` coverage.** `disable_commands` correctly gates `:!!` (`session.rs:700`) and `:ed` (`session.rs:585`), but `:r` and `:xml` are not gated, and environment seeding is suppressed only under Go compat (`session.rs:77-78`). Separately, plain `-X` returns silently from those handlers with no message, whereas `-X1` reports and exits (`session.rs:422-425`). What is the reference behavior you matched for the silent-ignore case, and is the ODBC-keeps-startup-script comment at `session.rs:75-76` backed by observed `sqlcmd` behavior? Not asserting a defect — `-X` semantics differ between the two sqlcmds and you clearly went looking, I just couldn't confirm which one these lines follow.

**11. `set_defer_batch_errors` / `take_pending_errors` / `take_done_row_counts` are `pub` on `TdsClient`.** That widens the core crate's public surface for one experimental consumer. Is the intent for these to be part of the supported `mssql-tds` API, or would `pub(crate)` plus a narrower accessor serve until sqlcmd firms up?

**12. `SUMMARY.md`** reads as an implementation progress report rather than durable documentation, and it disagrees with `sqlcmd.md:327` vs `:330` and with `session.rs:584` about whether `:ed` is implemented or deferred. Worth dropping or folding into `PARITY.md`.

---

*Posted by an unattended automated review run — these findings have not been checked by a human first. Blocking/Suggestion labels are the reviewer's assessment, not a merge gate; the author owns the merge. Two findings from an earlier automated pass (a missing `Cargo.lock` update and the Entra authority trust) were investigated and dropped as incorrect, as noted above.*

mssql-sqlcmd/src/vars.rs
```diff
@@ -0,0 +1,314 @@
...
+    "SQLCMDPACKETSIZE",
+    "SQLCMDSERVER",
+    "SQLCMDUSER",
+    "SQLCMDWORKSTATION",
+];
+
+/// Built-ins seeded from the environment at startup, unless `-X` says otherwise.
+const FROM_ENVIRONMENT: &[&str] = &[
+    "SQLCMDUSER",
+    "SQLCMDPASSWORD",
```
David-Engel, line 28: Seeding `SQLCMDPASSWORD` (and `OSQLPASSWORD` via the alias below) into `values` makes the live password readable by `$(SQLCMDPASSWORD)`, printable by `:listvar`, and writable by `:setvar` — the documented contract is `R/W = --`, and this PR's own Phase 2 table in `sqlcmd.md` (read out of the reference with `:listvar`) has no `SQLCMDPASSWORD` row. Verified locally: with `SQLCMDPASSWORD=hunter2` set, `get` returns `Some("hunter2")`, `listing()` contains `("SQLCMDPASSWORD", "hunter2")`, and `set` returns `Ok(())`. Nothing reads the password back out of `Variables` (`connect.rs` takes it from `Options`), so dropping it from `FROM_ENVIRONMENT` and reading the environment where the connection options are built looks like the cheapest fix.

mssql-sqlcmd/src/batch/substitute.rs
```diff
@@ -0,0 +1,157 @@
...
+//! `$(var)` usable to build up identifiers and literals.
+//!
+//! An undefined reference is a warning rather than an error. The reference
+//! leaves the text as it stands and sends the batch anyway, so the server is
+//! what ultimately complains.
+
+use crate::vars::Variables;
+
+/// Guards against `:setvar A $(A)`-style loops.
+const MAX_DEPTH: usize = 32;
```
David-Engel, line 17: This caps recursion depth but not total work. Each occurrence inside a value is expanded independently, so a value holding two references doubles per level: a 20-link chain of `:setvar V<n> "$(V<n+1>)$(V<n+1>)"` produced 1,048,576 bytes in 1.87s when I measured it, and the cap allows 32 levels — 2^32 bytes. Roughly 33 `:setvar` lines in an `-i` file or a `:r` include take the process to OOM before a batch is sent. Carrying a byte budget through `expand_at` alongside `depth`, and stopping with the usual warning when it is exceeded, bounds it without changing the self-reference behaviour this cap was added for.

David-Engel (commented):
## Summary

Second automated pass over the same commit (`70e5517c`) as the earlier review, so this one is deliberately incremental: I read that review first and am not re-filing its twelve findings. What follows is new material — four findings it did not reach, one correction to it, and fresh evidence for its first (and most important) one.

**What I ran locally** (detached worktree at `70e5517c`, not the main checkout): `cargo clippy --workspace --all-features --all-targets` — clean, confirming the earlier result. `cargo nextest run -p mssql-sqlcmd --lib` for two throwaway probe tests (reverted, not part of the branch). And, for the first time, the differential suite itself with `SQLCMD_DIFF_REF` pointed at the ODBC `sqlcmd` installed on this machine.

**Checked and deliberately not flagging**, so they don't get re-raised: the `CREDENTIALW` mirror in `open_cmds.rs` is laid out correctly — field order and padding match the SDK struct on x64, x86 and ARM64 — and the ADS password goes to the Credential Manager rather than an argv, which is the right call and notably better than the container path. The `unsafe` in `ffi.rs` is disciplined: every block has a matching caller guarantee, `collect_wide` walks the NUL itself rather than trusting a length, and the null-`argv` guards have tests.

## Correcting the earlier review

**Finding 3 of the previous review is wrong, and should not be actioned.** It claimed an unrecognised `-N` suffix silently resolves to `EncryptionSetting::PreferOff` via `validate.rs:378` (`Some(_) => Encrypt::Optional`), calling it a fail-open on a security control. That arm is unreachable. `cli/spec.rs:84-87` declares `-N` as `Arity::Suffix(&['s', 'm', 'o'])`, and the lexer rejects anything outside that set before `resolve` ever runs. Verified against the built binary:

```
-Nx => exit=1  Sqlcmd: Command -N: Invalid Parameters passed.
-NS => exit=1  Sqlcmd: Command -N: Invalid Parameters passed.
-NY => exit=1  Sqlcmd: Command -N: Invalid Parameters passed.
-Ns / -Nm / -No => accepted, connection attempted
```

The `Some(_)` arm is defensive, not a fail-open. Under the repo's own guidance on unreachable branches, either disposition is fine — leaving it with a note is reasonable given `Encrypt` is shared.

## Blocking

**1. `SQLCMDPASSWORD` is seeded into the readable, listable and writable variable map.** `vars.rs:28` puts `SQLCMDPASSWORD` in `FROM_ENVIRONMENT` (and `vars.rs:53` adds the `OSQLPASSWORD` fallback), so `seed_from_environment` stores the password in the same `values` map that `get` reads and `listing()` prints. It is also absent from `READ_ONLY`, so `:setvar` accepts it. Measured, with `SQLCMDPASSWORD=hunter2` in the environment:

```
get("SQLCMDPASSWORD")  => Some("hunter2")
listing()              => [("SQLCMDPASSWORD", "hunter2")]
set("SQLCMDPASSWORD",…) => Ok(())
```

Three consequences: `:listvar` prints the live password in cleartext to the results stream — which `-o` may be sending to a file; `$(SQLCMDPASSWORD)` expands anywhere in a batch, including inside a string literal, so any script this tool runs can exfiltrate the password it was invoked with; and the value is writable, which the other five seeded built-ins are not. The documented contract is `R/W = --` — neither readable nor writable ([sqlcmd scripting variables](https://learn.microsoft.com/en-us/sql/tools/sqlcmd/sqlcmd-use-scripting-variables#sqlcmd-scripting-variables)), and this PR's own Phase 2 table in `sqlcmd.md:152-161`, read straight out of the reference with `:listvar`, has no `SQLCMDPASSWORD` row — the reference does not expose it. The password is presumably seeded here so `-P`-less invocations can find it, but nothing reads it back out of `Variables`; `connect.rs` takes it from `Options`. Suggested fix: drop `SQLCMDPASSWORD` from `FROM_ENVIRONMENT` and read the environment directly where the connection options are built, or hold it in a separate field that `get`/`listing` cannot reach. A test that sets the variable and asserts `:listvar` output does not contain it would pin this.

**2. The `$(var)` depth cap does not bound the work.** `batch/substitute.rs:17` caps recursion at `MAX_DEPTH = 32`, which stops the self-referencing case the comment names but not fan-out: each occurrence in a value is expanded independently, so a value holding two references doubles per level. Measured with a 20-link chain of `:setvar V<n> "$(V<n+1>)$(V<n+1>)"`:

```
len=1048576  elapsed=1.87s   (20 levels)
```

That is 2^20 bytes for 20 levels; the cap allows 32, i.e. 2^32 bytes and roughly four thousand times that CPU — a script file of ~33 `:setvar` lines takes the process to OOM before a single batch is sent. `:setvar` lines come from `-i` files and `:r` includes, which are exactly the inputs a caller may not have written. Suggested fix: carry a byte budget through `expand_at` alongside `depth` and stop (with the usual warning) when the accumulated output exceeds it — the depth cap can stay as the loop guard.

## Suggestions

**3. A panic inside `run` aborts the host process rather than returning an exit code.** `ffi.rs:87` calls `crate::run(args)` directly from `sqlcmd_modern_main`. Unwinding out of an `extern "C"` function aborts, so any panic anywhere below `run` — an `unwrap` on a malformed server response, an allocation failure in the formatter — terminates the native `sqlcmd` the archive is linked into, with no message and no exit code. The module doc says the boundary is "two functions and nothing else", and that is the right shape; it just needs the catch. Suggested fix:

```rust
match unsafe { collect(argc, argv) } {
    Some(args) => std::panic::catch_unwind(|| crate::run(args))
        .unwrap_or(crate::exitcode::FAILURE),
    None => crate::exitcode::FAILURE,
}
```

Same for `sqlcmd_modern_main_w`. `claims` is pure and cheap, so it matters less there, but wrapping both is consistent.

**4. `take_done_row_counts` has no deterministic test.** The distinction the API exists for — `None` for a statement under `SET NOCOUNT ON` versus `Some(0)` for one that ran and affected nothing — is asserted nowhere. The two new tests in `query_results.rs` cover deferral only, and both need a live server; on the sqlcmd side the behaviour is reachable only through the `SQLCMD_DIFF_SERVER` cases. `test_login_server_name.rs` shows the pattern that would fix this: drive `mssql-mock-tds` to emit DONE tokens with and without `DONE_COUNT` and assert the `Vec<Option<u64>>` that comes back. Without it, `record_done_row_count` losing its `has_count()` guard would be invisible.

**5. Row counts for a DML statement between two result sets may land after the wrong result set.** In `exec/runner.rs`, `report_counts` runs *after* `render_result_set` and takes everything accumulated so far, while `advance_to_rows` (`tds_client.rs:4113`) silently consumes `NoRows` statements on its way to the next COLMETADATA. So for `SELECT …; UPDATE …; SELECT …`, the `UPDATE`'s DONE is consumed during the advance, and its count is not flushed until after the *second* SELECT's rows have been drawn — where the reference prints it between them. I could not confirm this against the real binary (it needs a server, and the ODBC `sqlcmd` on this machine is 15.0.1300, too far from the reference these cases were written against), so treat it as a concern rather than a verified defect. Worth noting that no case in `tests/diff/cases/*.toml` contains an `INSERT`, `UPDATE` or `DELETE` at all, so the suites would not catch it even fully enabled.

## Fresh evidence for the previous review's finding 1

Its claim that the differential suites are no-ops in CI is right, and the runners are sound — I ran `tests/diff.rs` with `SQLCMD_DIFF_REF` set to the installed ODBC `sqlcmd` and it does exactly what it says:

```
differential: 30 passed, 3 failed, 99 skipped
```

The 99 skips are the `SQLCMD_DIFF_SERVER` cases. All 3 failures are reference-version drift, not defects: my reference is 15.0.1300 (2019 banner, no `-N[s|m|o]` or `-F` in its usage text, and a different message for a bad `-N` suffix), while the cases were captured against a 2025-banner build. That is worth folding into the fix for finding 1 — an opt-in env var alone still leaves the suite's verdict dependent on whichever `sqlcmd` happens to be installed. Asserting the reference's version up front, and reporting a mismatch as "unsupported reference" rather than a case failure, would make a green run mean something.

## Nits

**6. `open_cmds.rs:130** escapes quotes in the user name by hand (`user.username.replace('"', "\\\"")`) before passing it through `Command::args`, which does its own `CommandLineToArgvW`-compatible quoting. A user name containing `"` reaches Azure Data Studio with a literal backslash. Dropping the `replace` is the fix.

**7. `io.rs:145-146** maps code page `20127` (US-ASCII) and `28591` (ISO-8859-1) onto `windows-1252`. For `20127` that silently emits bytes in `0x80–0xFF` the caller asked not to receive, which is the same class of problem the commit "close the SQLCMDINI, code-page and shell gaps" fixed for unusable code pages.

**8. PR metadata** is unchanged since the last review: the description is still the untouched template, with no summary and no linked GitHub issue or ADO work item, which `.github/PULL_REQUEST_TEMPLATE.md` states is required.

---

*Posted by an unattended automated review run — these findings have not been checked by a human first. This is an incremental second pass on the same commit; the earlier review's findings still stand except finding 3, corrected above. Blocking/Suggestion labels are the reviewer's assessment, not a merge gate; the author owns the merge.*

David-Engel (commented):
## Review — `Experiment: Sqlcmd`

Reviewed as a draft experiment, so I've weighted correctness and security over polish. The engineering here is genuinely good: the layering is clean, the differential-test idea is the right way to pin parity, and the `login_server_name` work in `mssql-tds` is properly tested against the mock server (length/payload agreement in LOGIN7 is exactly the failure mode worth guarding). Nice.

**Verification:** `cargo clippy --workspace --all-features --all-targets -- -D warnings` clean; `cargo nextest run -p mssql-sqlcmd` 251/251 pass; the four new `test_login_server_name` mock tests pass. Four `certificate_validator` unit tests fail locally on missing `tests/test_certificates/*` fixtures — pre-existing and unrelated to this diff.

### Blocking

1. **`:ed` writes the statement cache to a predictable path in shared temp** (`session.rs:594`) — symlink-follow and disclosure of credentials that pass through the cache.
2. **Password interpolated unescaped into a pasteable shell command line** (`config_cmds.rs:423`/`429`), plus inconsistent masking across the five emitted connection strings.

### Worth resolving before this stops being an experiment

3. `consumed_pending_error()` pairs one DONE-with-error against one ERROR token; a procedure that raises usually emits DONEINPROC *and* DONE with the flag set, which would turn a deferred error into a `ProtocolError` and retire the connection.
4. The diff deletes a meaningful amount of existing explanatory comment in `tds_client.rs` that this change doesn't otherwise touch. Those notes record protocol invariants and some were written in response to real bugs — please restore them.
5. No authority allow-list before the user's plaintext password is POSTed to a server-named STS. MSAL (which the cited references use) validates the authority by default.
6. The three differential suites silently no-op without env vars and nothing in the PR wires them into CI, so `SUMMARY.md`'s "246/246 green" isn't reproducible by anyone but you.

### Smaller notes

Inline: UTF-8-bytes-vs-UTF-16-chars in the new `write_server_name` length, `-X`/`-X1` disagreeing about `:connect`, and unbounded fan-out in `$(var)` expansion.

Not inline, since they're about the shape of the PR rather than a line:

- **`PARITY.md` (696 lines) and `SUMMARY.md` read as a personal status report** — first person, "because I don't have one", "I chose not to reproduce". That's great PR-description material but odd as shipped crate documentation, and it will go stale fast. `SUMMARY.md` already has: it says "`:ed` / `:perftrace` deferred" while `session.rs` implements both. Suggest moving the status parts into the PR body and keeping only the durable design/parity rationale in-tree.
- **`ffi.rs` entry points are `extern "C"` and call `crate::run`,** which can panic. Rust aborts on unwind out of `extern "C"` rather than invoking UB, so this is safe but harsh for a function whose whole contract is "returns an exit code so the C++ caller stays in control". A `catch_unwind` returning `exitcode::FAILURE` would honour that contract.
- **19.4k lines in one PR** is a lot to land at once. If this graduates from experiment, splitting it — the `mssql-tds` changes (deferral, `login_server_name`) first, then the sqlcmd crate in layers — would give each part a review it can actually absorb.

mssql-sqlcmd/src/session.rs
```diff
@@ -0,0 +1,1191 @@
...
+        if self.options.disable_commands || self.options.disable_commands_and_exit {
+            return;
+        }
+
+        let editor = self
+            .vars
+            .get("SQLCMDEDITOR")
+            .unwrap_or("edit.com")
+            .to_string();
+        let path = std::env::temp_dir().join(format!("sqlcmd-{}.sql", std::process::id()));
```
David-Engel: **Blocking (security).** `:ed` writes the statement cache to a fully predictable path — `$TMPDIR/sqlcmd-<pid>.sql`. On Linux/macOS `/tmp` is shared and world-writable, which gives another local user two openings: pre-create that path as a symlink and `fs::write` follows it into any file this user can write, and read the cache while the editor is open. The cache routinely holds credentials — `:connect server -U sa -P ...` goes through it — and `fs::write` creates at umask-default 0644.

Create it unpredictably and privately instead. `tempfile` is already a dev-dependency; promoting it to a real dependency gives you `tempfile::Builder::new().prefix("sqlcmd-").suffix(".sql").tempfile()`. Failing that, `OpenOptions::new().create_new(true).write(true)` plus `.mode(0o600)` on unix at least refuses an existing path rather than following it.

mssql-sqlcmd/src/modern/config_cmds.rs
```diff
@@ -0,0 +1,626 @@
...
+        format!("sqlcmd -S {address},{port} -d {database}")
+    } else if cfg!(windows) {
+        format!(
+            "SET \"SQLCMDPASSWORD={password}\" & sqlcmd -S {address},{port} -U {username} -d {database}"
+        )
+    } else {
+        // The line is meant to be pasted into a shell, so it has to be the
+        // shell the caller is actually running.
+        format!(
+            "export 'SQLCMDPASSWORD={password}'; sqlcmd -S {address},{port} -U {username} -d {database}"
```
David-Engel: **Blocking (security).** The password is interpolated straight into a shell command line. A password containing `'` closes the quoting and the remainder is re-parsed by the shell, so a value like `p'; <command>; '` produces a line that does rather more than set a variable when pasted. The Windows branch on line 423 has the same problem with `"`, `%` and `&`.

Separately, this function is inconsistent about secrecy: ADO.NET, JDBC and ODBC all get `******`, but the GO URL (line 417) and this line print the real password. Worth picking one policy — mask everywhere, or put the plaintext forms behind an explicit opt-in flag so it's a deliberate act.

The GO string needs percent-encoding too: a password containing `@`, `/`, `:` or `?` yields a URL that parses as something other than what was intended.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3831,7 +3829,7 @@ impl TdsClient {
                         done.has_more()
                     );
 
-                    if done.has_error() {
+                    if done.has_error() && !self.consumed_pending_error() {
```
David-Engel: `consumed_pending_error()` models "an error is outstanding" as a single `bool`, so it pairs exactly one DONE-with-error against one ERROR token. That holds for the case you tested — `RAISERROR` in a plain batch gives one ERROR and one DONE.

It doesn't obviously hold when the failing statement is inside a procedure. There the server typically emits one ERROR followed by DONEINPROC *and* a trailing DONE, both with `DONE_ERROR` set. The second of those finds `unreported_error == false` and this arm converts a deferred error into `ProtocolError`, which also retires the connection — a worse outcome than the non-deferred path it was meant to improve on.

A `usize` counter instead of a `bool`, or simply not raising `ProtocolError` at all while `defer_batch_errors` is on, would be robust to the shape of the DONE sequence. Either way, please add a deferred-error test that goes through `EXEC` of a procedure that raises — that's the shape `sqlcmd` users hit most, and it's the one the current tests don't cover.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3589,21 +3600,10 @@ impl TdsClient {
...
     ///
-    /// A statement-scoped error (for example lock timeout 1222) does not abort
-    /// the batch, so the server keeps streaming any result sets that follow it.
-    /// A trailing row-returning result set must therefore be consumed through
-    /// the row-decoding path: ROW/NBCROW tokens carry no length prefix and can
-    /// only be parsed with the preceding COLMETADATA in the parser context.
-    /// Skipping that step would leave unparsed row bytes in the transport and
-    /// corrupt the connection for reuse.
+    /// A statement-scoped error does not abort the batch, so subsequent result
+    /// sets must be consumed through the row-decoding path before reuse.
```
David-Engel: This PR deletes a fair amount of existing explanatory comment in `tds_client.rs` that the change doesn't otherwise touch. Here it's the note about why `drain_stream` must parse COLMETADATA with `ParserContext::ColumnEncryption` — the CEK-table prefix that would otherwise desynchronize the stream, i.e. the exact corruption this drain exists to prevent. The same pattern shows up at the two ERROR arms (~3958 and ~5422, on why batch state is reset before propagating), in `begin_command` (~5563, on why return values and the affected-row count are cleared), and in the `dml_result_counts` field doc (~299, which loses the `take_dml_result_counts`/ODBC contract entirely).

Those comments record protocol invariants that aren't recoverable from reading the code, and judging by their wording at least a couple were written in response to real bugs. Please restore them — the new deferral behaviour can be documented alongside them rather than in place of them.

mssql-sqlcmd/src/exec/entra/oauth.rs
```diff
@@ -0,0 +1,262 @@
...
+    scope: &str,
+) -> TdsResult<String> {
+    let endpoint = token_endpoint(sts_url)?;
+    let response = reqwest::Client::new()
+        .post(&endpoint)
+        .form(&[
+            ("client_id", SQL_TOOLING_CLIENT_ID),
+            ("grant_type", "password"),
+            ("username", user),
+            ("password", password),
```
David-Engel: This POSTs the user's plaintext password to an endpoint whose host comes from the server's FEDAUTHINFO. The module header acknowledges the trade-off and enforces `https`, but https only proves the attacker-named host holds a certificate for itself — it doesn't stop the redirect, so the mitigation and the threat don't line up.

The parity argument doesn't fully hold either: msodbcsql and SqlClient both acquire tokens through MSAL, which performs authority (instance discovery) validation against the known Entra cloud list by default and refuses an unrecognised authority unless the caller explicitly opts out. So this is weaker than the references it's matching, not equal to them.

Suggest an allow-list — `login.microsoftonline.com`, `login.windows.net`, `login.microsoftonline.us`, `login.partner.microsoftonline.cn` — applied at minimum to `password_token`, where an actual reusable credential rather than a scoped token request is at stake, with an explicit flag to opt out for private/sovereign clouds.

mssql-sqlcmd/tests/diff.rs
```diff
@@ -0,0 +1,59 @@
...
+mod runner;
+
+use std::path::PathBuf;
+
+use runner::{Outcome, REF_ENV};
+
+#[test]
+fn matches_the_reference_implementation() {
+    let Some(reference) = std::env::var_os(REF_ENV).map(PathBuf::from) else {
+        eprintln!("skipping: set {REF_ENV} to the ODBC sqlcmd binary to run these tests");
```
David-Engel: All three differential suites (`diff`, `go_diff`, `modern_diff`) return early and report success when `SQLCMD_DIFF_REF` and friends are unset, and the PR adds no pipeline changes to set them. In CI these tests are therefore green while asserting nothing, and `SUMMARY.md`'s "246/246 green on Windows, 254/254 on Linux" is only reproducible on your machine — which is a shame, because the differential approach is the most valuable thing in this PR and deserves to be enforced.

Two things would fix it: a pipeline job that installs the reference binaries and exports the env vars, and a `SQLCMD_DIFF_REQUIRED=1` marker that turns the skip into a failure, so a job that loses its reference binary is loud rather than silently vacuous.

mssql-tds/src/message/login.rs
```diff
@@ -1110,12 +1111,17 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
...
     async fn write_server_name(&mut self) -> TdsResult<()> {
-        if self
-            .write_metadata(self.model.transport_context.get_login_server_name().len() as i16)
-            .await?
-        {
+        let server_name = self
+            .model
+            .user_input
+            .login_server_name(self.model.transport_context);
+        if self.write_metadata(server_name.len() as i16).await? {
```
David-Engel: Nit: `write_metadata` takes a *character* count — it's the `cchServerName` half of an offset/length pair describing a UTF-16 payload — but `.len()` is UTF-8 bytes. The two agree only for ASCII.

The mismatch is pre-existing and shared with hostname, username and database, so it isn't introduced here. It's worth raising on this line anyway: `login_server_name` is new, arbitrary user input from `--server-name`, which makes this the first of those fields where a non-ASCII value is a realistic input rather than a theoretical one. An IDN or non-Latin override writes a length that disagrees with the payload — precisely the corruption the doc comment just above correctly says must not happen. `server_name.encode_utf16().count() as i16` fixes it for this field, and the new integration tests are all ASCII so they wouldn't catch the regression.

mssql-sqlcmd/src/session.rs
```diff
@@ -0,0 +1,1191 @@
...
+            return out;
+        }
+    }
+    messages::basic_errorinfo(DRIVER_NAME, &decorate(&error.to_string()))
+}
+
+/// `-X1` refuses the commands that reach outside the script.
+fn is_disabled(text: &str) -> bool {
+    let lowered = text.trim().to_ascii_lowercase();
+    lowered.starts_with("ed") || lowered.starts_with("!!") || lowered.starts_with("connect")
```
David-Engel: Nit, two parts:

`-X` and `-X1` disagree about `:connect`. `edit()` and `shell()` each guard on `disable_commands || disable_commands_and_exit`, but `reconnect()` has no guard at all, so the `connect` entry in this list only takes effect under `-X1`. Plain `-X` runs it silently. Whichever behaviour is intended, the two paths should agree.

`starts_with` rather than an exact word match means anything beginning with `ed` or `connect` is caught. Harmless against today's command set, but it's a trap for whoever adds the next command. Matching on the parsed `Command` variant instead of the raw text would tie this to the same word list `commands::parse` already owns.

Also worth recording in the doc comment that the reference's `-X` disables only `ED` and `!!` — including `:connect` is a deliberate hardening choice, and a reader comparing against the reference will otherwise think it's a bug.

mssql-sqlcmd/src/batch/substitute.rs
David-Engel: Nit: `MAX_DEPTH` bounds recursion depth but not output size, and those aren't the same thing here because each level can fan out. `:setvar A $(B)$(B)`, `:setvar B $(C)$(C)`, and so on for 32 levels expands to roughly four billion copies and takes the process out on memory. A `:r`'d script can set that up without the user seeing it.

The reference caps the substituted result length rather than the nesting depth. A byte budget threaded through `expand_at` — bail out and report the name as undefined once it's exceeded — would close this and is closer to what the reference does anyway.

PR #413, opened 2026-08-28
Add cross conversions between integer and character parameters

Theekshna:
## Description

Completes the parameter conversion matrix for integers and strings by filling the two cross-family quadrants: an **integer C type bound to a character SQL type**, and a **character C type bound to an integer SQL type**.

Neither is a new conversion. Each is an adapter onto a converter that already exists, so the target keeps its own rules:

| Binding | What happens |
|---|---|
| `SQL_C_SLONG` -> `varchar(n)` | format base 10, then hand the digits to the P4 character converter, which applies the declared length |
| `SQL_C_CHAR` -> `int` | parse the text, then hand the value to the P3 integer converter, which narrows |

So `"12"` -> `int` is exact, an out-of-range literal is `22003`, and an over-long formatted integer is `22001` - all from code that was already there and already tested.

### Numeric parsing is now shared with the fetch path

`numeric::parse_numeric_text` replaces two independent parsers that had already drifted apart. msodbcsql shares this too: `Convert` routes a character source into `ConvertToFixed`'s `case SQL_C_CHAR` arm whichever way the data moves, and `SQL_C_CHAR` and `SQL_CHAR` are both `1`, so a character *column* and a character *application buffer* are indistinguishable there.

**This changes fetch behaviour.** Only blanks are padding, per `CharToBigint` (`sqlccnvt.cpp:7777`). Fetch previously inherited `str::trim` from `parse_decimal_literal` and accepted `"\t12"`, which msodbcsql rejects - that is now `22018`. Pinned by `only_blanks_pad_a_numeric_column`.

Because the parser is shared, a literal too wide for an exact `i128` mantissa has to serve both targets: `NumericSource::WideDecimal` carries the `f64` approximation for float targets alongside the integer part and a dropped-fraction flag for integer targets. Reducing it to one or the other silently breaks the other direction, so both are asserted, at the parser, at `convert_float_c` (`SQL_C_DOUBLE` and `SQL_C_FLOAT`), at `convert_integer_c`, and end to end.

An embedded NUL ends the number as `CharToBigint`'s loop does (`sqlccnvt.cpp:7800`), so binding `strlen + 1` parses.

Exponent forms keep msodbcsql's routing rather than being made self-consistent: `Convert` scans for an `e`/`E` (`sqlccnvt.cpp:5092`) and sends a plain literal to `CharToBigint`, which walks digits and flags a dropped fraction, but an exponent literal to `CharToDouble`, which keeps only what the double holds. So `"1e-400"` converts cleanly where `"0." + 400 zeros + "1"` is `22001`. Spelling-dependent in both drivers, and pinned as such.

Severity still varies by direction, as it does in msodbcsql: a dropped fraction is `01S07` outbound and `22001` inbound.

### Divergences from msodbcsql

**Two**, deliberate, signed off, and recorded in `.github/instructions/mssql-odbc.instructions.md`.

**1. An integer bound to a character type is length-checked.** msodbcsql length-checks no integer C type - it enters no arm of `ParamToSQLType` that has one (`sqlcfunc.cpp:2586`, `:2854`, `:3165`, `:3177`) - and what it does instead is **undefined, varying by build**. `SQL_C_SLONG` 12345 into a `SQL_VARCHAR` of `ColumnSize` 3 has been measured three ways:

| Build | Behavior |
|---|---|
| retail 18.05.0002 | `SQL_SUCCESS`, no diagnostic, server receives `varchar(3)` holding `"123"` - silent data loss |
| debug 18.06.0002 | aborts on `assert(*pstMaxLen > 0 && *pstMaxLen >= stLen)` (`sqlcmisc.cpp:7458`) |
| retail 18.6.2.1 | `SQLExecute` never returns |

`ColumnSize` 32 returns `"12345"` on all three. This driver reports `22001`. `IntegerParamTooWideForColumnSizeIs22001` and `NegativeSignCountsAgainstColumnSize` carry `SKIP_IF_COMPARING_MSODBCSQL()`.

**2. A `SQL_C_WCHAR` buffer of nothing but blanks bound to an integer type is `22018`; msodbcsql answers `HY000`.** Measured against retail 18.05.0002. This is the *only* input where the two disagree here - the same blanks as `SQL_C_CHAR`, a zero-length wide buffer, and every other invalid literal in either width all answer `22018` on both drivers. Blanks are the only padding `CharToBigint` trims (`sqlccnvt.cpp:7777`), so this input trims to nothing, and a wide source reaches that parser through a transcode the narrow one skips - but which of the two produces the different exit is **not established**, so only the state is evidence, not the mechanism. `BlankOnlyWideLiteralIs22018` carries `SKIP_IF_COMPARING_MSODBCSQL()` and is the whole of it; the remaining 16 assertions in `CharParamInvalidLiteralIs22018`, and all of `LocaleFormattedNumbersAreRejected`, run on the compare leg.

### Known difference, not registered as a divergence

A **leading** NUL in a length-prefixed buffer (`"\0 12"`, len 3) is `22018` here; `CharToBigint`'s loop exits immediately and yields 0. msodbcsql reports `22018` for the same buffer under `SQL_NTS`, so only the length-prefixed spelling differs. Pathological; documented at `numeric.rs` and pinned by `a_leading_nul_is_not_a_number`.

## Related Issues

[AB#47369](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47369) - https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47369

Part of the parameter conversion milestone ([AB#46373](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46373)). Follows #385 (P4).

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes - 1145 tests, 0 failures
- [x] New/changed functionality has tests - 24 unit tests, 22 e2e cases in `param_cross_conversions_test.cpp`, and one fetch case in `get_data_test.cpp`
- [x] Public API changes are documented - no public API change; `docs/parameters_plan.md` updated
- [x] Full e2e suite passes, including `--compare-with-msodbcsql` against retail msodbcsql


mssql-odbc/src/conversion/numeric.rs
```diff
@@ -102,6 +102,106 @@ pub(crate) fn narrow_i128<T: TryFrom<i128>>(v: i128) -> Result<T, ConvError> {
...
+
+    if let Some(source) = parse_decimal_literal(trimmed) {
+        return Ok(source);
+    }
+
+    // `parse_decimal_literal` gives up once the digits overflow an exact
+    // mantissa, but the fraction still has to be reported rather than rounded
+    // away by `f64`.
+    if let Some(source) = parse_wide_decimal(trimmed) {
+        return Ok(source);
```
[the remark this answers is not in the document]
Theekshna: Correct, and a real regression. `NumericSource::WideDecimal { approx, int_part, fraction_dropped }` now carries both: `approx` for float targets, the other two for integer targets.

Covered at every layer that broke: `a_wide_decimal_keeps_its_value_for_float_targets` (parser), `wide_decimal_column_reaches_a_double_target`, `wide_decimal_column_reaches_a_single_precision_target` (incl. the `f32` overflow arm), `wide_decimal_column_truncating_to_an_integer_reports_truncation`, and e2e `WideDecimalColumnKeepsPrecisionForADoubleTarget`. Mutation-checked: reverting `as_f64` to the reduction kills all three float tests and correctly leaves the integer one passing.

mssql-odbc/tests/e2e/tests/char_conversions_test.cpp
```diff
@@ -255,12 +262,15 @@ TEST_F(CharConversionLiveTest, NarrowToWideOverlongParamIs22001) {
...
-// OverflowingBlanksAreTrimmedSilently, which msodbcsql does honour because that
-// binding is narrow -> narrow and reaches :2955.
+// Blank overflow on the narrow -> wide path is trimmed, and msodbcsql agrees:
+// retail 18.05.0002 also returns "abc" as nvarchar(3), so this test passes
+// against both drivers. The skip is not a behavioral divergence - it is here
+// because the reference driver registered for the comparison leg is a debug
+// build, and this binding trips assert(*pstMaxLen >= stLen) (sqlcmisc.cpp:7458)
+// and aborts the process. All six characters reach DescribeRPCParam untrimmed;
+// retail nonetheless declares nvarchar(3), so the assert's retail fallthrough at
+// :7459 does not describe what the retail binary does. Point the comparison at a
```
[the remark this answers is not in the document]
Theekshna: Right that the justification was wrong: CI pins retail 18.6.2.1 (`msodbcsqlVersion`, installed by `install-msodbcsql.ps1`), not a debug build. Comment corrected - it no longer claims the CI reference is debug.

Skip retained, though. It is measured as matching on retail 18.05.0002 and aborting on debug 18.06.0002, but **not** measured on 18.6.2.1, and those builds already provably disagree on the sibling integer case. Removing it on the strength of an 18.05 result would be the same mistake. Tracked as outstanding; the file is now `param_char_conversions_test.cpp`.

mssql-odbc/docs/parameters_plan.md
```diff
@@ -477,10 +483,16 @@ Deferred:
...
+  incrementing (`:2928`), so exactly one character of overflow escapes: retail
+  18.05.0002 silently truncates `"abcd"` to `"abc"` in a `wvarchar(3)` and
+  returns `SQL_SUCCESS` with no diagnostic. Two characters over is rejected, and
+  the narrow -> narrow control rejects at one. The declaration is *not* widened -
+  `MaxLength` reports the declared `nvarchar(3)`. Debug 18.06.0002 aborts on the
+  assert at `sqlcmisc.cpp:7458` instead. `NarrowToWideOverlongParamIs22001` is
+  skipped under `--compare-with-msodbcsql` for that reason;
+  `NarrowToWideOverflowingBlanksAreTrimmed` is *not* a divergence - msodbcsql
+  matches us - and its skip is only there because the debug reference build
+  aborts on that binding too.
```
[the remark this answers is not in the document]
Theekshna: Agreed. The section no longer describes a debug-only abort as the reason for the skip; it now says the case matches on retail 18.05.0002 and that the skip stands only until it is measured against the pinned retail 18.6.2.1.

The parity-reference rule in `mssql-odbc.instructions.md` also names 18.6.2.1 explicitly now, so a measurement against any other build is stated as settling that build only.

David-Engel (commented):
**Automated review (unattended run).** These findings were not checked by a human before posting. Comment only: nothing approved, requested, labeled, or resolved.

## Summary

Reviewed the full `6e7c9e8c..114c2f62` diff plus the code it lands on: `bound_param_to_value`, `convert_character_sql` / `fit_to_declared_length`, `dae_placeholder_type`, `sql_family`, `is_supported_conversion`, and the fetch side (`numeric_source_or_parse`, `convert_integer_c`, `convert_float_c`). The two cross-family quadrants really are adapters — `integer_as_text` hands digits to the existing character converter and `integer_from_text` hands a parsed `i128` to the existing integer converter — so the target-side rules (`ColumnSize` bound, narrowing, `22003`) come from code that was already tested. The `WideDecimal` fix for the earlier `as_f64` regression is correct and covered at the parser, at `SQL_C_DOUBLE`/`SQL_C_FLOAT`, at the integer path, and e2e. No blocking issues.

**Blocking:** none. **Suggestions:** 3. **Nits:** 2.

## What I verified

- **Local build/test/lint** in a dedicated detached worktree with a scratch `CARGO_TARGET_DIR`: `cargo nextest run -p mssql-odbc --lib` → **1123 passed, 0 failed, 1 skipped**; `cargo fmt -- --check` clean; `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` clean. All PR checks are green on `114c2f62`, including the five-platform ADO validation and the cross-repo mssql-python suite (`coverage-report` still pending at review time).
- **Error ordering.** `integer_from_text` narrows via `convert_integer_sql` *before* testing `fraction_dropped`, so `"999.5"` → `SQL_TINYINT` is `22003` and `"12.7"` → `SQL_INTEGER` is `22001`, matching the described msodbcsql ordering. Pinned by `overflow_outranks_a_dropped_fraction`.
- **Wide-literal arithmetic.** Walked the three parser branches by hand. `parse_decimal_literal` fails only on mantissa overflow (leading zeros still parse, e.g. `0.` + 40 zeros + `1` stays `Scaled`), so `parse_wide_decimal` fires exactly where intended; a 39-digit integer part with a fraction keeps `i128` exactness, a 40-digit one falls through to `f64` and lands on `22003` via the `±1.7e38` guard in `to_i128_truncating`. `parse_wide_decimal` requires a `.`, so plain over-wide integers correctly take the `f64` path.
- **Matrix/`sql_family` consistency.** `is_supported_conversion` is now a composition of two row groups; `sql_family` still assigns exactly one family per SQL type, and `sql_family_agrees_with_the_conversion_matrix` was correctly weakened to reachability plus explicit per-family lists rather than deleted.
- **Blast radius of the shared parser.** `parse_decimal_literal` now has no caller outside `numeric.rs`; `parse_numeric_text` is the single entry point for both directions. Output/inout params are still rejected at `SQLBindParameter`, so the new pairings cannot leak into an unimplemented reverse conversion.
- **Divergence bookkeeping.** Both registered divergences name the measured build and carry `SKIP_IF_COMPARING_MSODBCSQL()`; 4 of the 20 new e2e cases are skipped on the compare leg, and the rest (including the 16-assertion `CharParamInvalidLiteralIs22018` and all of `LocaleFormattedNumbersAreRejected`) run. The `NarrowToWideOverflowingBlanksAreTrimmed` / `parameters_plan.md` comments from the previous head no longer claim a debug build is the CI reference.

## Suggestions

**1. The NUL-splitting rule is justified for the parameter direction only, but it now governs fetch too — and it turns a `22018` into a silent wrong answer there.** `numeric.rs:147`, `let text = text.split('\0').next().unwrap_or("")`. The comment argues the case from an application passing `strlen + 1`, which is a bound-buffer concern. On the fetch side the source is a server-supplied column value whose length is authoritative and where an embedded NUL is real data:

```sql
SELECT CAST(CHAR(49) + CHAR(0) + CHAR(50) AS VARCHAR(8))  -- "1\0 2"
```

fetched into `SQL_C_SLONG` returned `22018` before this PR (`parse_decimal_literal` rejects it, then `"1\02".parse::<f64>()` errors), and now returns `SQL_SUCCESS` with `1`. That may well match `CharToBigint` — its loop stops at the NUL whichever way the data moves, which is the PR's own thesis — but it is an unstated fetch behavior change on top of the one the description does call out (`"\t12"` → `22018`), and there is no test pinning it in either direction. Either add a `numeric.rs` unit test plus a `get_data_test.cpp` case asserting the fetch answer, or scope the split to the parameter path. Given the description flags the whitespace change explicitly, this one deserves the same treatment.

**2. Cross-family data-at-execution moves from a bind-time error to an execute-time one, and nothing outside a unit-test comment records it.** `SQL_C_CHAR` + `SQL_INTEGER` + `SQL_DATA_AT_EXEC` used to be refused by `SQLBindParameter` with `HYC00`; now the bind succeeds and `dae_placeholder_type` (`param_convert.rs:212-225`) refuses it at execute, after the application has built its `SQLParamData`/`SQLPutData` loop. The unit test acknowledges this ("the refusal moves from bind time to here"), and the driver genuinely cannot decide at bind time since the DAE indicator is read at execute — so the behavior itself looks right. What's missing is the paper trail: the **Data-at-execution streaming** bullet in `parameters_plan.md` (line ~93) and the **Data-at-exec follow-ups** section (line ~641) both still describe DAE without noting that the newly bindable cross-family pairings are not streamable, and there is no e2e case for it. One sentence in the plan plus one `param_cross_conversions_test.cpp` case would close it.

**3. Make `parse_decimal_literal` private and drop its internal `str::trim`.** `numeric.rs:87` is still `pub(crate)` but is now reachable only through `parse_numeric_text`, whose whole point is that `str::trim` is *wrong* here — it strips the full Unicode whitespace set where `CharToBigint` trims `' '` alone. `parse_numeric_text` guards against that by pre-rejecting non-blank whitespace at `numeric.rs:155`, so the inner trim is dead code that would silently re-open the leniency the moment a future caller reaches `parse_decimal_literal` directly. Narrowing it to `fn` and deleting the `let t = text.trim();` line makes the invariant structural instead of conventional, and the existing `parse_decimal_literal` unit tests (`" .5 "` at `numeric.rs:239`) can move to `parse_numeric_text`, where they assert the rule that actually ships.

## Nits

**1. `WideDecimal::is_negative` reads `approx`, so a sign can vanish on `f64` underflow.** `numeric.rs:52` returns `*approx < 0.0`, and `-0.0 < 0.0` is false. A literal with enough significant digits to overflow the exact mantissa *and* a magnitude below `f64::MIN_POSITIVE` (`"-0." + "0".repeat(400) + "9".repeat(39)`) underflows to `-0.0`, so `SQL_C_BIT` — the only consumer of `is_negative` (`fetch_convert.rs:214`) — writes `0` instead of reporting `22003`. Pathological, and the other three variants are exact. `parse_wide_decimal` already computes `negative` (`numeric.rs:206`); carrying it into the variant and returning it here removes the `f64` dependency for free.

**2. `mssql-odbc.instructions.md` now carries build-specific measurements that will age.** The new parity-measurement rule is a good addition, but the divergence entry it anchors records three msodbcsql builds by version with one of them (18.6.2.1) hanging and one (18.05.0002) being the only one actually measured for the behavior claimed. Worth a pointer to AB#47369 as the place where the 18.6.2.1 measurement lands, so the instructions file doesn't become the tracking system.

## Considered and not filed

- `SQL_LONGVARCHAR` with `ColumnSize` 0 yields `limit = Some(0)` and so rejects a formatted integer with `22001`, unlike `SQL_VARCHAR` which selects `varcharmax`. Identical for a character source on `main`, so pre-existing and out of this PR's scope.
- `integer_from_text` decodes the narrow buffer with `from_utf8_lossy`, so invalid UTF-8 becomes `U+FFFD` and then `22018`. Correct answer, and consistent with `AppText::transcode` on the same path.
- The e2e-count and unit-test-count claims in the checklist check out (20 `TEST_F` in `param_cross_conversions_test.cpp`, 1123 lib tests passing locally).

Theekshna:
All five actioned. Rebased onto `bc4b2dc8` first (one add/add conflict in `get_data_test.cpp` with #426, kept both). Head is now `837d1138`, 1126 lib tests passing.

**S1 - NUL rule governs fetch.** Confirmed your repro: `"1\02"` into `SQL_C_SLONG` was `22018`, now `1`. Pinned in both directions rather than left implicit - `a_nul_ends_the_number_in_either_direction` and e2e `EmbeddedNulEndsANumericColumn`. Kept the behaviour: `CharToBigint` stops at the NUL whichever way the data moves, which is the same argument the shared parser rests on.

**S2 - DAE bind to execute.** Right that nothing covered it. The existing `CrossFamilyDataAtExecutionIsRejected` is `SQL_C_CHAR` -> `SQL_WVARCHAR`, not the newly bindable pairings. Added `CrossFamilyDataAtExecutionIsRejectedAtExecute` (both widths against `SQL_INTEGER`, asserting the bind now succeeds and `SQLExecute` gives `HYC00`) plus the note on the streaming bullet in `parameters_plan.md`.

**S3 - `parse_decimal_literal`.** Now private, inner `str::trim` deleted; `" .5 "` moved to `parse_numeric_text` where the shipped rule is.

**N1 - sign through underflow.** Fixed, and worth noting the obvious fix is wrong: returning the stored `negative` alone makes `"-0." + "0"*40` report negative. It is `negative && (int_part != 0 || fraction_dropped)`. Mutation-checked - forcing `false` kills `a_wide_decimal_keeps_its_sign_through_underflow`.

**N2 - instructions ageing.** Entry now points at AB#47369 as where the 18.6.2.1 measurements land, rather than growing per build.

Still open: `NarrowToWideOverflowingBlanksAreTrimmed` is measured as matching on retail 18.05.0002 and aborting on debug 18.06.0002, but not on the pinned 18.6.2.1. Since those builds already disagree on the sibling integer case, the skip stays until measured.

mssql-odbc/src/conversion/numeric.rs
```diff
@@ -102,6 +128,105 @@ pub(crate) fn narrow_i128<T: TryFrom<i128>>(v: i128) -> Result<T, ConvError> {
...
+    // A *leading* NUL diverges: that loop exits at once leaving `*pValue = 0`
+    // and `CVT_NO_ERROR`, where the empty remainder is `22018` here. Left as an
+    // error - a length-prefixed buffer starting with NUL carries no number, and
+    // msodbcsql itself reports `22018` for the same buffer under `SQL_NTS`.
+    let text = text.split('\0').next().unwrap_or("");
+
+    // Only blanks are padding. `CharToBigint` trims `' '` alone
+    // (`sqlccnvt.cpp:7777`) and treats every other non-digit as invalid, so a
+    // tab, an interior blank or a non-breaking space is an error. Checked
+    // explicitly because `parse_decimal_literal` applies `str::trim`, which
```
David-Engel: **Suggestion.** This comment is now false, and the check it justifies is dead. `parse_decimal_literal` no longer applies `str::trim` - this PR removed it, and its own doc two functions up (line 96) says so explicitly. Nothing downstream can re-admit the Unicode whitespace set either: the digit-only checks in `parse_decimal_literal` and `parse_wide_decimal` reject it, and Rust's `f64::from_str` grammar admits no whitespace at all.

Verified by mutation: with this `if` disabled, all 1126 `mssql-odbc` lib tests still pass - including `a_non_numeric_literal_is_22018` (`"\t12"`, `"12\n"`, `"1 2"`) and `only_blanks_pad_a_numeric_column`.

Either resolution is fine: keep the guard as a deliberate statement of the blanks-only rule and rewrite the comment to say what actually enforces it, or drop the guard. What should not survive is a comment pointing at a `str::trim` that no longer exists - the next person to read this pair of doc comments gets contradictory answers about whether `parse_decimal_literal` trims.

mssql-odbc/src/conversion/numeric.rs
```diff
@@ -40,6 +52,12 @@ impl NumericSource {
         match self {
             NumericSource::Int(v) => *v < 0,
             NumericSource::Scaled { mantissa, .. } => *mantissa < 0,
+            NumericSource::WideDecimal {
+                negative,
+                int_part,
+                fraction_dropped,
+                ..
+            } => *negative && (*int_part != 0 || *fraction_dropped),
             NumericSource::Float(f) => *f < 0.0,
```
David-Engel, line 67: **Suggestion (pre-existing, adjacent).** The `WideDecimal` arm just above was hardened against `-0.0` this round; this line one row down still has the same hazard. `"-1e-400"` takes the exponent fallback and becomes `NumericSource::Float(-0.0)` - verified with `rustc`: it parses to `-0.0`, `is_finite()` is true, `< 0.0` is false, sign bit set. So `is_negative()` returns `false` and the `SQL_C_BIT` arm at `fetch_convert.rs:214` writes `0` instead of reporting `22003` for a value that is genuinely negative.

This is on `main` too, so it is not a regression, and the parameter direction is unaffected (`integer_from_text` only calls `to_i128_truncating`). Raising it because the fix for the sibling arm landed here for exactly this reason, and because the fix is not the one-liner it looks like: a blanket `is_sign_negative()` would wrongly call `"-0e0"` negative, so the sign has to be carried from the literal the way `parse_wide_decimal` now does. Either handle it in the fallback or note it next to the `WideDecimal` doc, so a reader does not conclude the whole enum is now `-0.0`-safe.

mssql-odbc/src/conversion/numeric.rs
```diff
@@ -102,6 +128,105 @@ pub(crate) fn narrow_i128<T: TryFrom<i128>>(v: i128) -> Result<T, ConvError> {
...
+/// (`sqlccnvt.cpp:5088`). `SQL_C_CHAR` and `SQL_CHAR` are both `1`, so a
+/// character *column* and a character *application buffer* enter that arm
+/// indistinguishably - the same parser serves `SQLGetData` and
+/// `SQLBindParameter`.
+///
+/// Severity is the caller's business, not this function's: a dropped fraction is
+/// a `01S07` warning outbound and a `22001` error inbound, so callers take
+/// [`NumericSource::to_i128_truncating`]'s flag and decide.
+///
+/// Only exponent forms go through `f64`, so a literal carrying more significant
```
David-Engel: Nit: this says only exponent forms reach `f64`, but the comment at line 181 - and `parse_wide_decimal`'s own doc - say integers too wide for an exact mantissa fall through to it as well. That second path is what makes `(SQL_BIGINT, "9999999999999999999999999999999999999999999")` land on `22003` in `a_literal_outside_the_target_range_is_22003`. "Only exponent forms and over-wide integers" settles it.

mssql-odbc/src/params/conversion_matrix.rs
```diff
@@ -38,12 +33,9 @@ const CHARACTER_C_TARGETS: &[SqlSmallInt] = &[
...
 
-/// SQL types a `SQL_C_BINARY` buffer can be converted to.
-const BINARY_C_TARGETS: &[SqlSmallInt] = &[SQL_BINARY, SQL_VARBINARY, SQL_LONGVARBINARY];
-/// SQL types any integer C buffer can be converted to. Width is not part of
-/// legality: a value that does not fit the target is a runtime `22003`, not a
-/// rejected binding.
-const INTEGER_C_TARGETS: &[SqlSmallInt] = &[SQL_TINYINT, SQL_SMALLINT, SQL_INTEGER, SQL_BIGINT];
+const BINARY_SQL_TARGETS: &[SqlSmallInt] = &[SQL_BINARY, SQL_VARBINARY, SQL_LONGVARBINARY];
+
+const INTEGER_SQL_TARGETS: &[SqlSmallInt] = &[SQL_TINYINT, SQL_SMALLINT, SQL_INTEGER, SQL_BIGINT];
```
David-Engel, line 41: Nit: renaming the three constants dropped their doc comments, and one of them is not recorded anywhere else. `INTEGER_C_TARGETS` carried "Width is not part of legality: a value that does not fit the target is a runtime `22003`, not a rejected binding" - which is exactly the question a reader has on seeing `SQL_TINYINT` in a list now reachable from `SQL_C_SBIGINT`. (The `SQL_SS_XML` / hex-string-binary half does survive, in `docs/parameters_plan.md:474`.) Worth putting that one line back.

David-Engel (commented):
**Automated review (unattended run).** These findings were not checked by a human before posting. Comment only: nothing approved, requested, labeled, or resolved.

## Summary

Reviewed the full `bc4b2dc8..837d1138` diff against current `origin/main`, plus the code it lands on: `bound_param_to_value`, `convert_character_sql` / `fit_to_declared_length`, `convert_integer_sql`, `dae_placeholder_type`, `sql_family`, `effective_param_c_type`, `read_param_value`, and the fetch side (`numeric_source_or_parse`, `convert_integer_c`, `convert_float_c`, and the `SQL_C_BIT` arm). Both quadrants really are adapters onto converters that already existed, so the target-side rules come from code that was already tested. All three findings from the previous automated pass on `114c2f62` are addressed: the NUL rule is now pinned in both directions (`a_nul_ends_the_number_in_either_direction`, `EmbeddedNulEndsANumericColumn`), the DAE move is recorded in `parameters_plan.md` and pinned by `CrossFamilyDataAtExecutionIsRejectedAtExecute`, and `parse_decimal_literal` is private with its `str::trim` gone. No blocking issues.

**Blocking:** none. **Suggestions:** 2. **Nits:** 2.

## What I verified

- **Local build/lint/test** in the assigned detached worktree with a scratch `CARGO_TARGET_DIR`: `cargo nextest run -p mssql-odbc --lib --no-fail-fast` gives **1126 passed, 0 failed, 1 skipped**; `cargo fmt -- --check` clean; `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` clean. The checklist claims hold.
- **Mutation checks** - the new tests do guard the change:
  - `is_negative`'s `WideDecimal` arm forced to `false`: `a_wide_decimal_keeps_its_sign_through_underflow` and `a_wide_decimal_keeps_its_value_for_float_targets` fail.
  - `integer_from_text` reordered to test the fraction before narrowing: `overflow_outranks_a_dropped_fraction` fails with `StringTruncation` where `OutOfRange` is expected.
  - Whitespace guard in `parse_numeric_text` disabled: **all 1126 still pass** - see suggestion 1.
- **`-0.0` reachability**: `"-1e-400".parse::<f64>()` yields `-0.0`, finite, `< 0.0` false, sign bit set (checked with `rustc`). That is the basis for suggestion 2.
- **Matrix composition**: `is_supported_conversion` is now two row groups per C family; `sql_family` still assigns exactly one family per SQL type, and the weakened `sql_family_agrees_with_the_conversion_matrix` compensates with explicit per-family lists rather than dropping the assertion outright.
- **Blast radius of the shared parser**: `parse_decimal_literal` has no caller outside `numeric.rs`, and `parse_numeric_text` is the single entry point for both directions. Output/inout parameters are still refused at `SQLBindParameter`, so the new pairings cannot leak into an unbuilt reverse conversion.
- **CI**: `coverage-report` is *cancelled*, not failed. The two red ADO legs (Build Linux ARM, Build Windows ARM) failed on `Wait for cross-pool SQL host endpoint` and `Wait for SQL host endpoint sentinel` - environment provisioning, not a compile or test failure from this diff (read from the public build timeline for build 171220). The remaining legs were still running at review time, so re-run and confirm green before merge.

## Considered and not filed

- The `BindText` helper passes `vector::data()` straight to `SQLBindParameter`, which for the `""` case is only non-null because an earlier iteration already grew the vector. It works as ordered today, and `read_param_value` returns `NullValuePointer` rather than dereferencing, so this is fragility rather than a defect - it would only bite if the case list were reordered.
- `IntegerParamTooWideForColumnSizeIs22001`, `NegativeSignCountsAgainstColumnSize` and `BlankOnlyWideLiteralIs22018` carry skips justified against retail 18.05.0002 rather than the pinned 18.6.2.1. Already raised on the sibling case, answered, and tracked in AB#47369.
- `String::from_utf8_lossy(&bytes).into_owned()` in `integer_from_text` copies on the valid path, but `bytes` is dropped at the end of that match arm so the `Cow` cannot be borrowed out. Not a redundant allocation.
- `is_supported_conversion` composing two row groups makes `SQL_C_CHAR` -> `SQL_TINYINT` legal at bind while a value that does not fit is a runtime `22003`. That is the documented intent, and `a_literal_outside_the_target_range_is_22003` pins it.

Theekshna:
All four actioned in `8c6819a4`.

**S1 - dead whitespace guard.** Reproduced your mutation independently: disabling the `if` leaves 1126/1126 passing. Removed it rather than rewriting the comment - a guard no test can kill, sitting under a comment describing a `str::trim` that no longer exists, is worse than neither. The comment now says what actually enforces the rule: `trim_matches(' ')` plus the digit-only checks downstream and `f64::from_str` admitting no whitespace. Mutating the surviving `trim_matches(' ')` kills four tests, so the behaviour is pinned where it lives.

**S2 - `Float(-0.0)`.** Confirmed and not fixed here: it is on `main`, the parameter direction is unaffected, and as you say the one-liner is wrong - `is_sign_negative` would call `"-0e0"` negative. Documented next to the `WideDecimal` doc precisely so the hardening is not read as covering the enum, with the reason the naive fix fails. Tracked in AB#47369.

**N1 - `f64` fallback doc.** Fixed: exponent forms *and* over-wide integers.

**N2 - lost matrix doc.** Restored on `INTEGER_SQL_TARGETS`.

On CI: agreed the two red ARM legs were `Wait for SQL host endpoint` provisioning, not this diff. Will confirm green on the new head before this leaves draft.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -495,6 +479,43 @@ fn convert_integer_sql(sql_type: SqlSmallInt, v: i128) -> Result<SqlType, ParamB
...
+        AppText::Utf8(bytes) => String::from_utf8_lossy(&bytes).into_owned(),
+        AppText::Utf16(bytes) => decode_utf16le(&bytes),
+    };
+    let (v, fraction_dropped) = parse_numeric_text(&decoded)
+        .map_err(ParamBuildError::Value)?
+        .to_i128_truncating()
+        .ok_or(ParamBuildError::Value(ConvError::OutOfRange))?;
+    let value = convert_integer_sql(sql_type, v)?;
+    if fraction_dropped {
+        return Err(ParamBuildError::StringTruncation);
```
Vahid-b, line 518: The linked AB#47369 acceptance text still says that `"12.7"` yields `12` with `01S07`, while this implementation correctly reports inbound fractional loss as `22001`, matching native `ParamToSQLType`. Please update the work item so its acceptance criteria agree with the implemented and tested behavior.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -969,3 +969,44 @@ TEST_F(GetDataLiveTest, UnsupportedColumnTypeHyc00PreservesValue) {
...
+    double out = 0.0;
+    SQLLEN ind = 0;
+    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_DOUBLE, &out, sizeof(out), &ind),
+                  SQL_HANDLE_STMT, stmt_);
+    EXPECT_NEAR(out, 1.2345678901234567, 1e-15);
+    EXPECT_EQ(static_cast<SQLLEN>(sizeof(double)), ind);
+
+    SQLCloseCursor(stmt_);
+}
+
```
Vahid-b: Nit: please remove this extra blank line at EOF; `git diff --check` flags it.

Vahid-b (commented):
Two non-blocking follow-ups from the comprehensive review.

Theekshna:
Both done @Vahid-b

**AB#47369 acceptance text.** Updated (rev 9). You flagged the `01S07` line; two adjacent claims were wrong for the same reason, so all three are corrected:

- `"12.7"` is now stated as `22001`, not 12 with `01S07`
- "First use of the P2 warning channel" removed - this quadrant uses no warning channel at all
- The reason is recorded rather than just the verdict: `ParamToSQLType` rewrites `IDS_01_S07` to `IDS_22_001` for any non-2.x application (`sqlcfunc.cpp:3348`), so a dropped fraction is an error inbound and a warning only outbound

Also refreshed while there: `parse_decimal_literal` -> `numeric::parse_numeric_text`, added that 22003 outranks 22001, and recorded both registered divergences plus the outstanding retail 18.6.2.1 measurement, since the instructions file now points here for that.

**EOF blank line.** Fixed in `d4a7fcf5`; `git diff --check origin/main` is clean.

mssql-odbc/src/conversion/numeric.rs
Theekshna: Reproduced your mutation: with the `if` disabled all 1126 passed. Removed the guard rather than rewriting the comment - dead code under a comment describing a `str::trim` that no longer exists is the worst of the options. The comment now names what actually enforces the rule: `trim_matches(' ')` plus the digit-only checks and `f64::from_str` admitting no whitespace. Mutating the surviving trim kills four tests, so the behaviour is pinned where it lives. `9d849b10`.

mssql-odbc/src/conversion/numeric.rs
Theekshna: Confirmed, and left as-is deliberately: it is on `main`, the parameter direction never calls `is_negative`, and as you say `is_sign_negative` is not the fix since it would call `"-0e0"` negative. Documented next to the `WideDecimal` doc - including why the one-liner fails - so the hardening is not read as covering the enum. Tracked in AB#47369. `9d849b10`.

mssql-odbc/src/conversion/numeric.rs
Theekshna: Fixed: now "exponent forms and integers too wide for an exact mantissa", with the fraction claim narrowed to decimals, which is the case it was actually about. `9d849b10`.

mssql-odbc/src/params/conversion_matrix.rs
Theekshna: Restored on `INTEGER_SQL_TARGETS`: width is not part of legality, a value that does not fit is a runtime `22003` rather than a rejected binding. `9d849b10`.

mssql-odbc/tests/e2e/tests/param_cross_conversions_test.cpp
```diff
@@ -0,0 +1,526 @@
...
+
+    // Bind |text| as a narrow or wide character buffer, per |c_type|. The
+    // buffers are members so they outlive the SQLExecute that reads them.
+    SQLRETURN BindText(SQLSMALLINT c_type, const std::string& text,
+                       SQLSMALLINT sql_type, SQLULEN column_size = 0) {
+        narrow_.assign(text.begin(), text.end());
+        wide_.clear();
+        for (unsigned char ch : text) {
+            wide_.push_back(ch);
+        }
```
David-Engel: **Nit.** The wide arm widens each `unsigned char` of the UTF-8 `std::string` into one `SQLWCHAR`, so a non-ASCII literal is bound as its individual UTF-8 bytes: `"\xc2\xa0" "12"` becomes `U+00C2 U+00A0 "12"` rather than `U+00A0 "12"`.

Nothing is wrong today - every case that reaches the wide arm is ASCII, and the one non-ASCII literal (`LocaleFormattedNumbersAreRejected`) binds narrow only. The trap is for the next `SQL_C_WCHAR` case with a non-ASCII literal: it would still report `22018` and pass, but for a leading `Â` the test never meant to send, which is the "passes for the wrong reason" shape.

Cheapest fix is to make the assumption explicit rather than transcode, since no wide non-ASCII case is wanted yet:

```cpp
// ASCII only: the widening below is byte-per-unit, not a transcode.
ASSERT_TRUE(std::all_of(text.begin(), text.end(),
                        [](unsigned char c) { return c < 0x80; }))
    << "BindText cannot widen non-ASCII text";
```

David-Engel (commented):
**Automated review (unattended run).** These findings were not checked by a human before posting. Comment only: nothing approved, requested, labeled, or resolved.

## Summary

The branch was rebased since the last automated pass - `837d1138` is no longer an ancestor - so this is a full re-review of `dada8903..9ced62e2` rather than a delta pass. Read the two new arms in `bound_param_to_value` plus the code they land on (`convert_character_sql` / `fit_to_declared_length`, `convert_integer_sql`, `dae_placeholder_type`, `sql_family`, `read_param_value`) and the fetch side that now shares the parser (`numeric_source_or_parse`, `convert_integer_c`, `convert_float_c`, the `SQL_C_BIT` arm). Both quadrants are adapters onto converters that already existed, so the target-side rules come from tested code, and the two divergences are registered with a named build. Everything raised on `114c2f62` and `837d1138` is addressed on this head. No blocking issues.

**Blocking:** none. **Suggestions:** none new. **Nits:** 1.

## What I verified

- **Local test run** in the assigned detached worktree with a scratch `CARGO_TARGET_DIR`: `cargo nextest run -p mssql-odbc --lib --no-fail-fast` gives **1133 passed, 0 failed, 1 skipped**.
- **Mutation check on the newest commit.** `9d849b10` deleted the non-blank whitespace guard on the strength of "all 1126 still pass with it disabled", which is exactly the shape that leaves a rule unpinned. Changing the surviving `trim_matches(' ')` to `trim()` fails three tests - `fetch_convert::tests::only_blanks_pad_a_numeric_column`, `param_convert::tests::a_non_numeric_literal_is_22018` (`"\t12"`) and `param_convert::tests::locale_formatted_numbers_are_rejected` (`"\u{a0}12"`) - so the blanks-only rule is still guarded where it now lives, in both directions. Restored afterwards.
- **Parser boundary arithmetic, by hand.** A 38-digit integer part with a fraction still parses exactly, because the 39-digit mantissa is under `i128::MAX` and never reaches `parse_wide_decimal`; a 39-digit one overflows the mantissa, keeps its integer part, and takes the `WideDecimal` path; an integer part past `i128` returns `None` from `parse_wide_decimal` and falls to `f64`, where the `±1.7e38` guard in `to_i128_truncating` reports `22003` - the same answer the dropped-fraction flag would have produced, since any such value overflows `bigint` anyway. `parse_wide_decimal` requires a `.`, so plain over-wide integers correctly take the `f64` path.
- **Suite rename.** `char_conversions_test` -> `param_char_conversions_test` plus the new suite are both registered in `CMakeLists.txt`, discovery is `ctest`-based rather than an enumerated list in `.pipeline`, and no stale reference to the old name survives anywhere in the tree.
- **CI on this head**: no failing check. Linux/macOS/Windows builds, the macOS test leg, the cross-repo mssql-python suite and the Kerberos leg are green; the two ARM legs, the SQL-host job and `coverage-report` were still pending at review time.

## Nits

**1. `BindText`'s wide path widens byte-by-byte, so it mis-encodes any non-ASCII literal.** Left inline at `param_cross_conversions_test.cpp:41`. Nothing is wrong today; it is a trap for the next wide case.

## Considered and not filed

- `SQL_LONGVARCHAR` with `ColumnSize` 0 yields `limit = Some(0)` and so rejects even a one-character value with `22001`. Newly reachable from an integer buffer, but it is the same code and the same answer a character buffer already got on `main`, and it is pinned as intended behavior by the `ColumnSize`-bound test on the long types.
- `dae_placeholder_type(SQL_C_SLONG, SQL_VARCHAR)` returns `UnsupportedCType` rather than `ConversionNotImplemented` now that the pairing is bindable. Both surface as `HYC00` from `SQLExecute`, so the application-visible answer is the one the plan documents.
- The checklist says 20 e2e cases in `param_cross_conversions_test.cpp`; there are 21 after `CrossFamilyDataAtExecutionIsRejectedAtExecute` was added in `abef4ec2`. Description drift, nothing more.
- `parse_numeric_text` splits on the NUL before trimming blanks, so `" 12\0"` and `"12\0 "` behave the same. Consistent with `CharToBigint` reading the number first.
- The `-0.0` hazard on `NumericSource::Float`, the DAE move, and the parity-measurement rule in the instructions file were all raised on earlier heads and answered; the doc comment now carries the `Float` caveat next to `WideDecimal`, which is what the thread asked for.

mssql-odbc/tests/e2e/tests/param_cross_conversions_test.cpp
Theekshna: Good catch - the sibling `AsciiParam` in `param_char_conversions_test.cpp` already carries this assertion and I did not carry it across when writing `BindText`. Added, scoped to the wide branch: `LocaleFormattedNumbersAreRejected` binds the non-breaking space narrow, and `BindText` was building the wide buffer unconditionally, so an unscoped guard would have failed that case. `c0e59100`.

Also fixed the count drift you noted: 21 e2e cases, not 20.

mssql-odbc/tests/e2e/tests/param_cross_conversions_test.cpp
```diff
@@ -0,0 +1,531 @@
...
+    SQLRETURN Prepare(const std::string& sql) {
+        SqlTString s = ODBCTestUtils::ToSqlTStr(sql);
+        return SQLPrepare(stmt_, const_cast<SQLTCHAR*>(s.c_str()), SQL_NTS);
+    }
+
+    // Bind |text| as a narrow or wide character buffer, per |c_type|. The
+    // buffers are members so they outlive the SQLExecute that reads them.
+    SQLRETURN BindText(SQLSMALLINT c_type, const std::string& text,
+                       SQLSMALLINT sql_type, SQLULEN column_size = 0) {
+        narrow_.assign(text.begin(), text.end());
```
David-Engel, line 42: **Nit.** `narrow_.data()` / `wide_.data()` may be `nullptr` when the vector is empty, so binding `""` (which `CharParamInvalidLiteralIs22018` does) passes a null `ParameterValuePtr` with a non-null indicator. That only reaches the intended `22018` today because `"abc"` runs first in that loop and leaves capacity behind; reorder the list — or add an empty-string case to a test that has no earlier iteration — and the driver returns `ERR_INVALID_NULL_POINTER` (`HY009`) from `read_param_value` instead, which is a confusing failure to debug.

A `reserve` before the assign makes it order-independent:

```cpp
narrow_.reserve(text.size() + 1);
wide_.reserve(text.size() + 1);
narrow_.assign(text.begin(), text.end());
```

David-Engel (commented):

## Summary

Fills the two cross-family parameter quadrants by adapting onto the existing converters — `integer_as_text` hands formatted digits to the P4 character path, `integer_from_text` hands a parsed value to the P3 integer path — and extracts `numeric::parse_numeric_text` so fetch and bind agree on what counts as a number. The composition in `conversion_matrix.rs` (a C-type row is now a list of target groups) is the right shape for this, `sql_family_agrees_with_the_conversion_matrix` was correctly weakened to reachability plus explicit family lists rather than deleted, and the `NumericSource::WideDecimal` variant is threaded through all three accessors (`as_f64`, `is_negative`, `to_i128_truncating`), so no consumer sees a reduced value. I read the parser edge cases directly rather than trusting the description: blanks-only trim, embedded/leading NUL, `+`/`-`, `"12."`, all-zero fractions, `i128`-mantissa overflow into `parse_wide_decimal`, integer-part overflow falling through to the `f64` arm with the ±1.7e38 gate, and `-0.0` underflow. They line up with the comments and the tests.

Behavior changes to the fetch direction are real but deliberate and pinned: only blanks pad a numeric column now (`"\t12"` becomes `22018`), and an embedded NUL ends the number instead of invalidating it (`EmbeddedNulEndsANumericColumn`). Both match `CharToBigint`, and the doc comments say why. The two msodbcsql divergences are registered in `.github/instructions/mssql-odbc.instructions.md` with the measured builds named, and the earlier review threads about debug-vs-retail attribution have been corrected in `param_char_conversions_test.cpp` and `parameters_plan.md`.

## Verification

- `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` — clean.
- `cargo nextest run -p mssql-odbc --lib` — 1145 passed, 1 skipped.
- Read the full diff plus the surrounding code the changes depend on: `read_param_value` / `read_integer` / `effective_param_c_type` (the `SQL_C_TINYINT` sign rule is keyed on a `SQL_TINYINT` target, so a character target keeps it signed — as the new test asserts), `convert_character_sql` / `fit_to_declared_length`, `convert_integer_c` / `convert_float_c` (the only two `NumericSource` consumers), and the bind-time gate in `bind_param.rs`.
- The C++ e2e suites were read, not executed — no live SQL Server or Driver Manager here — so the msodbcsql parity measurements are taken on the author's evidence.
- Required CI: nothing failing on this head at review time (`coverage-report` and the ADO validation legs are still running).

## Findings

**Blocking** — none.

**Suggestion** — none.

**Nit** — one, in a test helper; see the inline comment.

Nothing here rises to a merge concern, so this reads as ready for human review — but this is a `COMMENT` review, not an approval, and the human review requirement still stands.

mssql-odbc/tests/e2e/tests/param_cross_conversions_test.cpp
Theekshna: Fixed in `7f9abf8b` with the `reserve` you suggested, on both buffers.

Went one step further so the suite proves it rather than relying on the helper being right: the empty string now runs **first** in that loop, so the case that would bind a null pointer is the one with no earlier iteration to have grown the vector. Before the reserve, that ordering fails with HY009.

mssql-odbc/src/conversion/numeric.rs
```diff
@@ -102,6 +136,104 @@ pub(crate) fn narrow_i128<T: TryFrom<i128>>(v: i128) -> Result<T, ConvError> {
...
+    // away by `f64`.
+    if let Some(source) = parse_wide_decimal(trimmed) {
+        return Ok(source);
+    }
+
+    // Exponent forms, and integers too wide for an exact mantissa, fall back to
+    // `f64` - msodbcsql splits to `CharToDouble` on an `e`/`E`
+    // (`sqlccnvt.cpp:5088`).
+    match trimmed.parse::<f64>() {
+        Ok(f) if f.is_finite() => Ok(NumericSource::Float(f)),
```
David-Engel, line 200: **Suggestion.** An exponent literal that underflows `f64` loses its dropped-fraction flag, so it binds as a silent `0` where the same value spelled as a plain decimal is `22001`.

Rust folds underflow into `Ok(0.0)` rather than an error, so this arm produces `Float(0.0)`, and `to_i128_truncating`'s `Float` arm then computes `f.fract() != 0.0` on a value whose fraction has already been erased.

Measured on this head by adding a probe test to `numeric.rs` and running `cargo nextest run -p mssql-odbc --lib` (probe reverted):

- `"1e-400"` -> `Some((0, false))`, so `SQL_C_CHAR` -> `SQL_INTEGER` sends `0` with `SQL_SUCCESS` and no diagnostic
- `"1e-320"` -> `Some((0, true))` -> `22001`
- `"0." + "0".repeat(400) + "1"` -> `Some((0, true))` -> `22001`, via the new `WideDecimal` arm

So the answer flips on whether the literal is written with an exponent, and only past the `f64` underflow boundary. This is the truncation sibling of the `-0.0` sign hazard you already documented on `NumericSource::Float`, and it has the same fix shape: carry "a non-zero digit was present" from the literal rather than asking the `f64` after the fact. Fetch inherits it too, where it means no `01S07`.

I did not measure what `VarR8FromStr` does with an underflowing literal, so this is a question about intent rather than a parity claim — and pathological enough that documenting it next to the existing `Float` caveat would be a fine resolution, so a reader does not conclude the flag is exact for every source.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -495,6 +479,43 @@ fn convert_integer_sql(sql_type: SqlSmallInt, v: i128) -> Result<SqlType, ParamB
...
+///
+/// - Discarding a non-zero fraction is an error (`22001`) here, where fetch only
+///   warns (`01S07`): `ParamToSQLType` rewrites it for a non-2.x application
+///   (`sqlcfunc.cpp:3348`).
+/// - Overflow outranks it, so a value doing both reports `22003` - msodbcsql
+///   narrows before that rewrite runs, which is why the order below is narrow,
+///   then check the fraction.
+fn integer_from_text(sql_type: SqlSmallInt, text: AppText) -> Result<SqlType, ParamBuildError> {
+    let decoded = match text {
+        AppText::Utf8(bytes) => String::from_utf8_lossy(&bytes).into_owned(),
```
David-Engel: Nit: `.into_owned()` allocates even when the buffer is already valid UTF-8, which is the ordinary case — `from_utf8_lossy` returns a borrowing `Cow` there and allocates only to repair malformed input. Binding a `Cow` keeps the borrow and leaves `parse_numeric_text(&decoded)` unchanged:

```rust
let decoded: Cow<'_, str> = match &text {
    AppText::Utf8(bytes) => String::from_utf8_lossy(bytes),
    AppText::Utf16(bytes) => Cow::Owned(decode_utf16le(bytes)),
};
```

`decode_utf16le` has no borrowing form, so the wide arm allocates either way.

David-Engel (commented):
**Automated review (unattended run).** These findings were not checked by a human before posting. Comment only: nothing approved, requested, merged, labeled, or resolved.

## Summary

Reviewed the full `e6223c2b..7f9abf8b` diff plus the code it lands on: `bound_param_to_value`, `convert_character_sql` / `fit_to_declared_length`, `convert_integer_sql`, `read_param_value` / `read_integer` / `effective_param_c_type`, `dae_placeholder_type`, `sql_family`, `is_supported_conversion`, and the fetch consumers of `NumericSource` (`convert_integer_c`, `convert_float_c`, the `SQL_C_BIT` arm). Both quadrants really are adapters: `integer_as_text` hands base-10 digits to the P4 character converter and `integer_from_text` hands a parsed `i128` to the P3 integer converter, so the target-side rules (`ColumnSize` bound, narrowing, `22003`) come from code that was already tested. `WideDecimal` is threaded through all three accessors, so no consumer sees a reduced value, and the matrix change is a composition of two target groups rather than a new row. No blocking issues.

**Blocking:** none. **Suggestions:** 1. **Nits:** 1.

## What I verified

- `cargo nextest run -p mssql-odbc --lib -E 'test(numeric)'` in the assigned worktree with a scratch `CARGO_TARGET_DIR` — 27 passed, and I added a throwaway probe test to `numeric.rs` to measure the parser directly rather than reasoning about it (reverted; the worktree is clean).
- All 19 required checks are green on `7f9abf8b`, including the five-platform ADO validation, the cross-repo `mssql-python` suite, and `coverage-report` (98% diff coverage, 93.2% overall).
- Read the prior threads first. The `WideDecimal` regression, the debug-vs-retail attribution, the dead blanks guard, the `INTEGER_SQL_TARGETS` doc, the byte-per-unit `BindText` widening, and the empty-buffer `reserve` are all raised, answered, and verified on this head; I did not re-file any of them.
- Error ordering holds: `integer_from_text` narrows before testing `fraction_dropped`, so `"999.5"` into `SQL_TINYINT` is `22003` while `"12.7"` into `SQL_INTEGER` is `22001`.
- Parser branches walked by hand and confirmed against the tests: blanks-only trim, embedded and leading NUL, `+`/`-`, `"12."`, all-zero fractions, `i128` mantissa overflow into `parse_wide_decimal`, integer-part overflow falling through to the `f64` arm behind the ±1.7e38 gate.
- The C++ e2e suites were read, not executed — no live SQL Server or Driver Manager here — so the msodbcsql parity measurements are taken on your evidence.

## Findings

**Blocking** — none.

**Suggestion** — one, inline at `numeric.rs:192`: an exponent literal that underflows `f64` binds as a silent `0` where the same value spelled as a plain decimal reports `22001`.

**Nit** — one, inline at `param_convert.rs:505`.

## Checked and not filed

- `SQL_LONGVARCHAR` / `SQL_WLONGVARCHAR` with `ColumnSize` 0 gives `limit = Some(0)` and would reject every non-empty formatted integer with `22001`. Not reachable: `parameter_column_size_is_valid` refuses 0 for those types at bind with `HY104`, and `column_size_zero_is_not_max_for_text_types` pins it.
- The narrow arm decodes with `from_utf8_lossy`, so invalid UTF-8 becomes `U+FFFD` and then `22018` — the right answer, and consistent with `AppText::transcode` on the same path.
- `SQL_C_BIT` is not an integer C type, so the new integer row does not widen `is_supported_conversion` past what the description claims.
- The divergence bookkeeping in `.github/instructions/mssql-odbc.instructions.md` names the measured build for each claim and points the outstanding 18.6.2.1 measurements at AB#47369 rather than growing the instructions file per build. That matches the new rule it adds.

Nothing here rises to a merge concern, but this is a `COMMENT` review, not an approval, and the human review requirement still stands.

mssql-odbc/src/conversion/numeric.rs
Theekshna: Correct, and fixed rather than documented - `c6983f6d`.

An underflowed literal with a non-zero significand now routes to `WideDecimal` carrying the sign and the dropped-fraction flag from the text, so `"1e-400"` and `"0." + "0"*400 + "1"` both answer `Some((0, true))`. `"1e-320"` is unchanged, and `"0e0"` / `"-0e0"` / `"0e-400"` still convert cleanly and are not negative, since their significand has no non-zero digit.

That also closes the `-0.0` sign hazard you raised earlier and I had documented as open: `"-1e-400"` now reads as negative. `Float` is left holding only values whose sign and fraction the `f64` itself still carries, and the doc comment says that instead of describing a live hazard.

On intent vs parity: I did not measure `VarR8FromStr` either. The case for fixing is internal consistency - one value in two spellings was giving two answers - not a parity claim, so nothing was added to the divergence registry.

Verified by mutation: disabling the new arm fails `an_underflowing_exponent_still_reports_its_fraction` with `left: Some((0, false))` / `right: Some((0, true))`, which is the bug you reported.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Applied in `06268120`. Matching on `&text` is the part that makes it work - your earlier pass filed this as not-an-issue because `bytes` was moved into the arm and died there, which was right for the code as written.

Wide arm unchanged, since `decode_utf16le` has no borrowing form.

David-Engel (commented):

## Summary

Re-reviewed the full `e6223c2b..06268120` diff on top of the two commits added since the last automated pass (`c6983f6d` exponent underflow, `06268120` the borrowed narrow decode). The shape still holds: both quadrants are adapters rather than new conversions — `integer_as_text` hands base-10 digits to the P4 character converter so `ColumnSize`, the blank exemption and `22001` all come from `fit_to_declared_length`, and `integer_from_text` hands a parsed `i128` to `convert_integer_sql` so narrowing and `22003` come from P3. The matrix change is a composition of two existing target groups, not a new row, and `binary_stays_outside_the_cross_family_rows` pins that binary gains nothing from it. `dae_placeholder_type` still derives its family test from `sql_family`, so the newly bindable pairings fall out as `HYC00` at execute rather than needing a third type list.

`06268120` is correct: `decoded` borrows `text`, which outlives the parse, so the `Cow` cannot dangle and the valid narrow path no longer allocates. `WideDecimal` is still threaded through all three accessors, and `is_negative` deliberately reads `negative && (int_part != 0 || fraction_dropped)` rather than `approx.is_sign_negative()`, which is what keeps `"-0e0"` from reading as negative.

## Verification

I ran everything against the assigned worktree with the review `CARGO_TARGET_DIR`. `cargo fmt -- --check` is clean, `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` is clean, and `cargo nextest run -p mssql-odbc --lib` is 1146 passed / 1 skipped. The e2e suite is 21 `TEST_F` cases in `param_cross_conversions_test.cpp`, matching the checklist.

I did not reason about the new underflow arm — I added a throwaway probe test to `fetch_convert.rs`, ran it, and reverted it (the worktree is clean). That probe is what the suggestion below rests on.

The C++ e2e suites were read, not executed: no live SQL Server or Driver Manager here, so the msodbcsql parity measurements are taken on your evidence. No required check is failing on this head.

I read the prior threads first and did not re-file anything already raised and answered: the `WideDecimal` float regression, the `-0.0` hazard on `Float`, the dead whitespace guard, the dropped `INTEGER_C_TARGETS` doc, the `BindText` byte-per-unit widening and its empty-buffer `reserve`, the `.into_owned()` allocation, the debug-vs-retail attribution, and the 18.05.0002-vs-pinned-18.6.2.1 skip justifications tracked in AB#47369.

## Findings

**Blocking** — none.

**Suggestion** — one, inline at `numeric.rs:196`. The underflow recovery added in `c6983f6d` also changes the fetch direction, and that change is neither tested nor registered. Probed on this head: a `varchar` column holding `1e-400` read into `SQL_C_SLONG` now returns `ConvOk::Truncated` (`01S07`), where on `main` the same value took `Float(0.0)` and returned `Exact`. msodbcsql splits an `e` form to `CharToDouble`, gets exactly `0.0`, and then has no fraction to flag, so the outbound answer there is very likely a clean `SQL_SUCCESS` — meaning the inbound fix probably bought an outbound divergence. The PR description carefully enumerates the fetch behaviour this PR changes (blanks-only padding, embedded NUL) and this one is not in the list. Either scope the recovery to the parameter direction, or keep it and pin it with a fetch case plus a measured note, the way the other two fetch changes are handled.

**Nit** — one, inline at `param_cross_conversions_test.cpp:204`. `SKIP_IF_COMPARING_MSODBCSQL()` covers the whole of `NegativeSignCountsAgainstColumnSize`, including its second half, where `ColumnSize` 4 and `-123` round-trip as `"-123"`. By your own measurements msodbcsql agrees on that one, so the compare leg is losing a case it would pass.

Neither finding rises to a merge concern, but this is a `COMMENT` review, not an approval, and the human review requirement still stands.

mssql-odbc/src/conversion/numeric.rs
```diff
@@ -102,6 +136,124 @@ pub(crate) fn narrow_i128<T: TryFrom<i128>>(v: i128) -> Result<T, ConvError> {
...
+
+    // Exponent forms, and integers too wide for an exact mantissa, fall back to
+    // `f64` - msodbcsql splits to `CharToDouble` on an `e`/`E`
+    // (`sqlccnvt.cpp:5088`).
+    match trimmed.parse::<f64>() {
+        // Underflow folds to zero and erases both the sign and the fact that a
+        // fraction was dropped, so `"1e-400"` would bind a silent 0 where the
+        // same value as a plain decimal is `22001`. Recover both from the text:
+        // a mantissa with any non-zero digit had a value, however small.
+        Ok(f) if f == 0.0 && has_nonzero_mantissa(trimmed) => Ok(NumericSource::WideDecimal {
```
David-Engel: **Suggestion.** This arm also changes the fetch direction, and that is not covered or registered anywhere. I probed it on this head: a `varchar` column holding `1e-400` read into `SQL_C_SLONG` now comes back `ConvOk::Truncated` (`01S07`), where on `main` the same text parsed as `Float(0.0)` and returned `Exact`. `1e-320` was already `Truncated` on both, so `1e-400` — exact underflow to `0.0` — is the value this moves.

The inbound fix is right, and it was my own earlier suggestion, so this is not asking for a revert. The point is that the parser is shared, so the fix landed on `SQLGetData` too. msodbcsql routes an `e` form to `CharToDouble`, gets exactly `0.0`, and then finds no fraction to flag, so outbound this most likely now diverges. The description lists the fetch behaviour this PR changes — blanks-only padding, embedded NUL — and this one is missing from that list, with no fetch test and no entry in the instructions file.

Two ways out that both look fine: pass the direction in and only recover the flag for parameters, or keep it shared, measure `SELECT '1e-400'` into `SQL_C_SLONG` against msodbcsql, and pin whichever answer you keep with a `fetch_convert.rs` case next to `only_blanks_pad_a_numeric_column`.

mssql-odbc/tests/e2e/tests/param_cross_conversions_test.cpp
```diff
@@ -0,0 +1,537 @@
...
+        EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "22001");
+        ResetParams();
+    }
+}
+
+// The sign counts against the declared length like any other character. Skipped
+// on the compare leg for the same reason as IntegerParamTooWideForColumnSizeIs22001:
+// retail msodbcsql silently truncates instead, turning -123 into "-12".
+TEST_F(CrossConversionLiveTest, NegativeSignCountsAgainstColumnSize) {
+    SKIP_IF_COMPARING_MSODBCSQL();
```
David-Engel, line 204: Nit: this skips the whole test, but its second half — `ColumnSize` 4 with `-123` round-tripping as `"-123"` — is a case msodbcsql agrees on, per your own measurements (no length check at all, so the value goes through). Splitting the positive assertion into its own unskipped test would keep that parity coverage on the compare leg.

mssql-odbc/src/conversion/numeric.rs
Theekshna: You are right, and it is worse than the fetch side you measured - so the recovery is **withdrawn** in `5d2832d2`.

I went and read the routing I had only been citing second-hand. `sqlccnvt.cpp:5092` scans for an `e`/`E`; `:5109` sends a plain literal to `CharToBigint`, which walks digits and flags a dropped fraction; `:5118` sends an exponent literal to `CharToDouble`, which keeps only what the double holds. **msodbcsql is spelling-dependent by construction**, so the pre-existing behaviour matched it and my consistency fix diverged - inbound as well as outbound. I made a self-consistency argument without reading the code that makes the inconsistency deliberate.

Kept instead: the routing is documented at `parse_numeric_text` with those three line numbers, and `an_underflowing_exponent_loses_its_fraction_as_msodbcsql_does` pins the spelling dependence as intended parity - `"1e-400"` clean, the written-out form `22001`, `"1e-320"` still truncating, `-1e-400` not negative.

That also reinstates the `-0.0` note on `Float`, now correctly framed: `is_negative` answering `false` there matches msodbcsql rather than diverging, since the same literal takes the same double path. No divergence registry entry, because there is no divergence.

PR description updated with the exponent routing, alongside the two fetch changes it already listed.

mssql-odbc/tests/e2e/tests/param_cross_conversions_test.cpp
Theekshna: Not pedantic - a skip is a parity claim, and that one was claiming more than it measured. Split in `5d2832d2`: `NegativeSignCountsAgainstColumnSize` keeps the skip and the over-long half only, and `NegativeSignFitsWhenColumnSizeAllowsIt` holds the ColumnSize-4 round trip and runs on the compare leg.

Same treatment as `CharParamInvalidLiteralIs22018` earlier: skip the input that actually diverges, not the test that contains it.

David-Engel (commented):

## Summary

Re-reviewed the full `e6223c2b..5d2832d2` diff. The only change since the last automated pass is `5d2832d2`, which reverts the exponent-underflow recovery added in `c6983f6d` and pins msodbcsql's routing instead — exactly the disposition the prior thread asked for, and the better one: `Convert` splits on `e`/`E`, so a plain literal keeps `CharToBigint`'s digit walk while an exponent literal is whatever the double holds. `"1e-400"` therefore binds a clean `0` and `"0." + 400 zeros + "1"` is `22001`, spelling-dependent in both drivers. `has_nonzero_mantissa` is gone with no dangling references, and the fetch direction is back to what `main` did for exponent forms, so the revert reintroduces no untested outbound change.

The rest of the PR still reads the way it did: both quadrants are adapters onto converters that already existed. `integer_as_text` hands base-10 digits to the P4 character path so `ColumnSize`, the trailing-blank exemption and `22001` all come from `fit_to_declared_length`; `integer_from_text` hands a parsed `i128` to `convert_integer_sql` so narrowing and `22003` come from P3, and it narrows before testing `fraction_dropped`, which is what makes `22003` outrank `22001`. The matrix change is a composition of two existing target groups rather than a new row, `dae_placeholder_type` still derives its family test from `sql_family` so the newly bindable pairings fall out as `HYC00` at execute without a third type list, and `every_accepted_pairing_is_convertible` keeps bind-time legality and execute-time capability from drifting apart. `WideDecimal` is threaded through all three accessors, so no consumer sees a reduced value.

`NegativeSignFitsWhenColumnSizeAllowsIt` is the right fix for the other nit: the half both drivers agree on now runs on the compare leg instead of being skipped with the diverging half.

## Verification

Ran against the assigned worktree with the review `CARGO_TARGET_DIR`: `cargo fmt -- --check` clean, `cargo clippy -p mssql-odbc --all-features --all-targets -- -D warnings` clean, `cargo nextest run -p mssql-odbc --lib --no-fail-fast` 1146 passed / 1 skipped. The checklist holds.

Mutation-checked the newest commit rather than trusting it: re-adding the underflow arm to `parse_numeric_text` fails `an_underflowing_exponent_loses_its_fraction_as_msodbcsql_does` and nothing else, so that test really does pin the routing the revert restores. Reverted; the worktree is clean.

Read the parser boundaries directly — blanks-only trim, embedded and leading NUL, `+`/`-`, `"12."`, all-zero fractions, `i128` mantissa overflow into `parse_wide_decimal`, an integer part past `i128` falling through to the `f64` arm behind the ±1.7e38 gate, and `-0.0` — plus the code the diff lands on: `bound_param_to_value`, `convert_character_sql` / `fit_to_declared_length`, `read_param_value` / `effective_param_c_type`, and the three `NumericSource` consumers in `fetch_convert.rs`. Output and input/output parameters are still refused at `SQLBindParameter`, so the new pairings cannot leak into an unbuilt reverse conversion.

The C++ e2e suites were read, not executed — no live SQL Server or Driver Manager here — so the msodbcsql parity measurements are taken on your evidence. 22 `TEST_F` cases in `param_cross_conversions_test.cpp`, matching the description. No required check is failing on this head; `coverage-report` and two ADO legs were still running at review time.

I read the prior threads first and did not re-file anything already raised and answered: the `WideDecimal` float regression, the `-0.0` hazard on `Float`, the NUL rule in both directions, the DAE move to execute time, the dead whitespace guard, the dropped `INTEGER_C_TARGETS` doc, the `BindText` byte-per-unit widening and its empty-buffer `reserve`, and the 18.05.0002-vs-pinned-18.6.2.1 skip justifications tracked in AB#47369.

## Findings

**Blocking** — none. **Suggestion** — none. **Nit** — none.

Nothing here rises to a merge concern, so this reads as ready for human review — but this is a `COMMENT` review, not an approval, and the human review requirement still stands.

shiwanigupta0809 (commented):
Reviewed the full diff plus the code it lands on, and checked the parity claims against the msodbcsql source and the ODBC spec independently rather than on the description.

**Verdict: approve.** Blocking: none. Suggestions: none. Nits: none.

### Verified

The fetch-direction whitespace tightening is correct, and the spec backs it more strongly than the source reading the PR cites: ODBC's *SQL to C: Character* appendix, "Extra spaces" — *"Leading and trailing **spaces** are ignored when SQL character data is converted to … numeric …"*. Spaces, not whitespace. The old `str::trim` leniency was the defect.

msodbcsql agrees on the full path, not just the cited line: `SQLGetData` → `InternalGetColData` (`sqlcdata.h:446`) → `OdbcDataFromSqlData` (`:1028`) → `Convert(..., TODRIVER)` (`sqlccnvt.cpp:803`) → `ConvertToFixed` (`:1349`) → `CharToBigint` (`:5109`). `FindSigNumber` (`:7961`) skips `' '` alone on both ends and the digit loop rejects everything else.

Both quadrants are genuine adapters onto already-tested converters, and the compare leg runs `CharParamInvalidLiteralIs22018` — including `"\t12"` — unskipped, which is what settles the point that mattered most.

### Observation — no action needed

The blanks-only rule doesn't govern the exponent path in msodbcsql, so the two drivers differ there — with this one on the correct side.

`ConvertToFixed`'s `SQL_C_CHAR` arm scans for `e`/`E` *before* reaching `CharToBigint` (`sqlccnvt.cpp:5088`) and routes those to `CharToDouble` (`:7910`). That calls `FindSigNumber` first, which trims blanks only — so a leading tab survives — and then hands the string to `VarR8FromStr`, which opens with `while (iswspace(*pStart)) ++pStart;` (`xplat/src/StringFunctions.cpp:1415`), skipping the full whitespace set. So `"\t1e3"` reaches `strtod("1e3")` and msodbcsql returns `1000`.

Here, `trim_matches(' ')` leaves the tab and `f64::from_str` admits no whitespace, so it's `22018` — which is what the spec prescribes for a value that isn't a numeric-literal. Recording it because it's a case where we're deliberately stricter than msodbcsql rather than matching it, and nothing in the tree currently says so.

Nice work on the parser extraction — sharing it between directions is the right call and it's what keeps the two directions honest with each other.

shiwanigupta0809 (approved):

PR #415, opened 2026-08-28
mssql-odbc: Fix use-after-free-before-check and double-free-on-retry in handle cleanup

### LINKED ISSUE #400: mssql-odbc: SQLFreeHandle on a stale STMT/DESC handle dereferences freed memory before the "already dropped" guard
### Describe the bug

`free_stmt` and `free_desc` (`mssql-odbc/src/api/free_handle.rs`) both guard against a double-free by checking the parent DBC's child list before calling `free_handle`, and returning `SQL_SUCCESS` early when the handle isn't found ("already dropped by SQLDisconnect"). That guard prevents the second `Box::from_raw`, but it is reached *after* the function has already dereferenced the handle.

`free_desc` reads `desc.object_type` (the `debug_assert_eq!`), locks `desc.inner` for `free_errors`, reads `desc.is_explicit()`, and reads `desc.parent_dbc` — all before the lookup. `free_stmt` has the same shape: it locks `stmt.inner` and reads `stmt.parent_dbc` before its equivalent guard.

When the box has genuinely been dropped by `SQLDisconnect`, every one of those reads is a use-after-free. In practice it usually "works" because the freed block hasn't been reused and `free_handle` stamps `object_type = HandleType::Invalid` before dropping — but that stamp lives in the freed allocation, so reading it back is itself UB. This was observed concretely during PR #370 review: an equivalent read-after-free in a test happened to return the expected value on Windows and Linux but not on macOS, whose allocator reused the block sooner.

This is pre-existing behavior in `free_stmt`, not a regression. `free_desc` intentionally mirrors it (see #370) rather than diverging, so the two paths stay consistent — which makes this the right place to fix both together.

Related: the existing `TODO: fix with refcounted handle lifetimes so STMT handles cannot be freed while in use` at `mssql-odbc/src/api/disconnect.rs:85`. That TODO covers the concurrent-use race; this issue is the sequential stale-handle case. They likely share a solution.

### Steps to reproduce

1. `SQLAllocHandle(SQL_HANDLE_DBC, henv, &hdbc)` and connect.
2. `SQLAllocHandle(SQL_HANDLE_DESC, hdbc, &hdesc)` (or `SQL_HANDLE_STMT`).
3. `SQLDisconnect(hdbc)` — `sql_disconnect_safe` frees every outstanding statement and explicit descriptor.
4. `SQLFreeHandle(SQL_HANDLE_DESC, hdesc)` with the now-stale handle.

Step 4 returns `SQL_SUCCESS` and does not double-free, but reaches that result by reading through a dangling pointer. Most reliably surfaced under ASan/Miri, or on an allocator that reuses freed blocks aggressively (macOS).

### Expected behavior

`SQLFreeHandle` on a handle already released by `SQLDisconnect` returns `SQL_SUCCESS` (per the current guard's intent) without dereferencing freed memory at any point.

Options worth weighing:

- Refcounted handle lifetimes, as the `disconnect.rs:85` TODO proposes — the general fix, and it also closes the concurrent-use race.
- Have `SQLDisconnect` unregister-but-not-free, deferring the actual drop to `SQLFreeHandle`. Changes ownership semantics and diverges from msodbcsql's `SQLDisconnect`; probably not preferred.
- A validity check that doesn't touch the allocation — e.g. consult the parent DBC's child list first, keyed only on the raw pointer value. Narrower, but requires knowing the parent without reading the handle, so it only helps if the parent is derivable another way.

### Actual behavior

Use-after-free reads (`object_type`, `inner` mutex lock, `alloc_type`, `parent_dbc`) on a handle whose `Box` was already dropped. No crash observed on Windows/Linux in practice; the failure is latent and allocator-dependent, and the optimizer is entitled to exploit the UB.

### Version

commit 9734bd3742053dc317e7a7aef9f5d1f27b48779e (branch `vahid-b-ado-46374-feature-plan`, PR #370)

### Affected crate

mssql-odbc

### Additional context

Identified during review of #370 (explicit descriptor lifecycle, [AB#47436](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47436)) and agreed there as out of scope for that PR. See the review thread on `free_handle.rs` for the original discussion.

No user-facing impact is known today. Filing so the latent UB is tracked rather than lost in a merged PR thread.

### LINKED ISSUE #401: mssql-odbc: SQLDisconnect statement loop can double-free on retry after a poisoned mutex (inconsistent with adjacent descriptor loop)
### Describe the bug

`sql_disconnect_safe` (`mssql-odbc/src/api/disconnect.rs`) frees two child-handle lists back to back, using two different patterns:

The statement loop iterates by reference, then clears:

```rust
for &stmt_ptr in &state.statements {
    let stmt = unsafe { handle_from_raw::<StmtHandle>(stmt_ptr) };
    let Ok(guard) = stmt.inner.lock() else {
        error!(?stmt_ptr, "SQLDisconnect: stmt mutex poisoned");
        return SQL_ERROR;   // <- leaves already-freed pointers in state.statements
    };
    drop(guard);
    unsafe { free_handle::<StmtHandle>(stmt_ptr) };
}
state.statements.clear();
```

The descriptor loop directly below it pops as it goes (changed in #370):

```rust
while let Some(desc_ptr) = state.descriptors.pop() {
    ...
}
```

On the poisoned-mutex path the statement loop returns `SQL_ERROR` mid-iteration with entries it has already passed to `free_handle` still present in `state.statements`, and `connection_state` still `Connected` (it isn't set to `Disconnected` until after both loops). An application that retries `SQLDisconnect` — a reasonable response to `SQL_ERROR` — re-walks the same list from the start and double-frees every statement freed by the first pass.

The descriptor loop is immune because popping removes each entry before it is freed, so a retry resumes at the remainder.

A poisoned mutex requires a prior panic while the lock was held, which `ffi_entry!`'s `catch_unwind` makes reachable rather than fatal: the panic is converted to `SQL_ERROR` and the process keeps running with a poisoned lock. So this is unlikely but not unreachable.

### Steps to reproduce

1. Connect a DBC and allocate two or more statements on it.
2. Poison one statement's `inner` mutex (in production: any panic caught by `ffi_entry!` while the lock is held).
3. `SQLDisconnect(hdbc)` → the loop frees the statements ahead of the poisoned one, then returns `SQL_ERROR`. `state.statements` still contains all of them.
4. `SQLDisconnect(hdbc)` again → the loop starts over and calls `free_handle` a second time on the statements freed in step 3.

### Expected behavior

Both loops behave the same way, and a retried `SQLDisconnect` after a mid-loop failure resumes on the not-yet-freed remainder rather than re-freeing. Applying the same `while let Some(stmt_ptr) = state.statements.pop()` shape to the statement loop makes the two consistent and closes the retry window:

```rust
while let Some(stmt_ptr) = state.statements.pop() {
    let stmt = unsafe { handle_from_raw::<StmtHandle>(stmt_ptr) };
    let Ok(guard) = stmt.inner.lock() else {
        error!(?stmt_ptr, "SQLDisconnect: stmt mutex poisoned");
        return SQL_ERROR;
    };
    drop(guard);
    unsafe { free_handle::<StmtHandle>(stmt_ptr) };
}
```

Worth deciding at the same time whether refusing the free is better than returning `SQL_ERROR` mid-teardown here. `free_desc` took the refuse-and-stay-recoverable route for a poisoned STMT mutex in #370, but disconnect is teardown, where leaving a statement permanently unfreeable is its own problem — so the trade-off may land differently.

### Actual behavior

`state.statements` retains pointers to already-freed `StmtHandle` boxes after a mid-loop `SQL_ERROR`, and `connection_state` stays `Connected`, so a retried `SQLDisconnect` double-frees them.

### Version

commit 9734bd3742053dc317e7a7aef9f5d1f27b48779e (branch `vahid-b-ado-46374-feature-plan`, PR #370)

### Affected crate

mssql-odbc

### Additional context

Pre-existing; not introduced by #370. Raised during that PR's review — the new descriptor loop was changed to pop-then-free in response, and the statement loop was deliberately left alone to keep the PR's scope tight. Filing so the now-visible inconsistency between two adjacent loops doesn't sit unexplained.

Vahid-b:
## Description

Fixes **#400** and **#401**, both filed by David Engel during PR #370's review as narrow, pre-existing follow-ups (explicitly called out as not blocking that PR, since they predate it or were deliberately deferred).

### #400 — `SQLFreeHandle` on a stale STMT/DESC handle dereferenced freed memory before the "already dropped" guard

`free_stmt` and `free_desc` both guarded against a double-free by checking the parent DBC's child list — but only *after* already dereferencing the handle (`debug_assert_eq!` on `object_type`, locking `.inner` for `free_errors`, reading `.parent_dbc`). When `SQLDisconnect` had already cascade-freed the handle behind the Driver Manager's back — an application legitimately holding a stale STMT/DESC handle across disconnect and then calling `SQLFreeHandle` on it is ordinary usage, not DM misuse — every one of those reads was a genuine use-after-free. This usually "worked" because the freed block hadn't been reused yet, but an allocator that reuses it sooner (observed on macOS during PR #370's own CI) breaks even that.

**Fix:** added a lightweight live-handle registry (`handles::live_type`, a process-wide `Mutex<HashMap<usize, HandleType>>` populated in `handle_to_raw` and depopulated in `free_handle` — the two existing choke points every handle already funnels through) that answers "is this raw address still an allocated handle, and of what type" *without ever dereferencing it*. `free_stmt`/`free_desc` now check it before doing anything else, returning `SQL_SUCCESS` immediately on a stale handle. Deliberately narrower than the full refcounted-handle-lifetime redesign `disconnect.rs`'s existing TODO already tracks — this only closes the *sequential* "already fully freed" case, not the separate *concurrent*-use race that TODO covers, nor address-reuse (ABA) — see below.

**Externally visible behavior change:** `SQLFreeHandle` now returns `SQL_INVALID_HANDLE`, per spec, when called with a live handle of the wrong type for all four handle types (ENV/DBC/STMT/DESC) — e.g. `SQLFreeHandle(SQL_HANDLE_STMT, <live DESC>)`. Before this PR, that case reached `handle_from_raw::<WrongType>` and silently reinterpreted the handle's memory as the wrong type (genuine UB, not just a logic error), since neither the affected handle structs are `#[repr(C)]` nor the same size as each other.

### #401 — `SQLDisconnect`'s statement loop could double-free on retry after a poisoned mutex

The statement-free loop in `sql_disconnect_safe` iterated `state.statements` by reference and cleared it only after the loop finished, so a poisoned STMT mutex mid-loop returned `SQL_ERROR` with every already-freed pointer still in the list — a retried `SQLDisconnect` (a reasonable response to `SQL_ERROR`) would re-walk the full original list and double-free. The descriptor loop directly below it already used a pop-then-free shape for exactly this reason; the statement loop just hadn't been updated to match.

**Fix, revised after review:** the initial fix switched the statement loop to the same pop-then-fail-on-poison shape as the descriptor loop, but review correctly flagged that shape as still broken: popping the pointer *before* the lock check means a poisoned mutex leaves the statement permanently orphaned — no longer tracked for a retry to find, yet never freed either, unreachable by any future call. Both loops now instead **tolerate a poisoned mutex and free anyway**: the lock here is pure synchronization (never read after acquiring), and a poisoning panic still fully releases the underlying lock on unwind, so a poisoned outcome doesn't change whether it's safe to free the box. This matches `free_stmt`/`free_desc`'s own existing tolerance for their handle's lock.

### Follow-ups from review, addressed in this PR

- **`LIVE_HANDLES` address-reuse (ABA) limitation.** The registry tracks raw addresses (now with type), not allocation identity, so it cannot distinguish a stale handle from an unrelated new one of the same type the allocator later places at the same freed address. Closing that needs handle identity that survives address reuse (a generation-tracked indirection layer), out of scope here — the same class of redesign as the pre-existing refcounted-handle-lifetime TODO. Documented directly on `LIVE_HANDLES`'s doc comment, added a test demonstrating the exact mechanism, and filed #422 to track a proper fix.
- **Type confusion beyond STMT/DESC.** An earlier revision of this PR closed the type-confusion hole only for `free_stmt`/`free_desc`; review pointed out `free_env`/`free_dbc` had the identical gap. Closed the same way for all four handle types.
- **Implicit-descriptor rejection skipped under a poisoned lock.** `free_desc`'s HY017 rejection for implicit descriptors was originally nested inside `desc.inner.lock()`, so a poisoned mutex on that descriptor skipped the whole check and fell through to unrelated code paths. Hoisted the (lock-free) `is_explicit()` check above the lock so the rejection is unconditional — this itself briefly regressed diagnostic-clearing (caught in follow-up review; every ODBC entry point must clear stale diagnostics before posting a new one) before landing on the final version, which folds `free_errors` into the same guarded block as `post_diag`.
- **Poisoned-registry recovery.** `LIVE_HANDLES`'s lock is recovered via `into_inner()` on poison rather than treating the whole registry as permanently blind — justified because its critical sections are single `HashMap` ops with no user code/I/O that could actually poison it, and this repo already uses the same pattern elsewhere (`auth/msqa.rs`, `auth/entra.rs`). This is a deliberate, narrow exception to `.github/instructions/mssql-odbc.instructions.md`'s general "never recover via `into_inner()`" rule; filed #430 to reconcile the instructions with it.

## Testing

Every fix ships with regression tests I verified **actually catch the bug**: temporarily reverted each fix, confirmed the corresponding test fails with the exact predicted symptom, then restored the fix and confirmed it passes. Highlights:

- `free_stmt_twice_is_a_no_op_not_a_panic` / `free_desc_twice_is_a_no_op_not_a_panic`: call `SQLFreeHandle` twice on the same handle, assert `SQL_SUCCESS` both times.
- `free_stmt_on_a_live_desc_handle_returns_invalid_handle` / `free_desc_on_a_live_stmt_handle_returns_invalid_handle` / `free_env_on_a_live_dbc_handle_returns_invalid_handle` / `free_dbc_on_a_live_env_handle_returns_invalid_handle`: a live handle of the wrong type is rejected with `SQL_INVALID_HANDLE`, not reinterpreted.
- `disconnect_frees_a_statement_even_if_its_mutex_is_poisoned`: poisons one of two statements' mutexes, confirms `SQLDisconnect` still succeeds and both statements are actually freed.
- `free_implicit_desc_is_rejected_even_with_a_poisoned_mutex`: poisons the implicit descriptor's own mutex, confirms the HY017 rejection still happens (calls `free_desc` directly, bypassing `ffi_entry!`'s `catch_unwind`, so a wrongly-firing `debug_assert!` can't be laundered into the same `SQL_ERROR` a correct rejection also returns).
- `free_implicit_desc_is_rejected`: now calls the rejection twice and asserts exactly one diagnostic record each time, guarding the diagnostic-clearing regression above.
- `live_handles_registry_recovers_from_a_poisoned_lock`: poisons the registry's own lock directly, confirms handle tracking still works afterward.
- `is_live_cannot_distinguish_a_reused_address_from_the_original_allocation`: documents the registry's known ABA gap by directly simulating address reuse (real OS-level reuse can't be forced portably/deterministically from a test).

## Validation

- `cargo nextest run -p mssql-odbc --lib` — 1098 passed, 1 skipped (pre-existing, requires a live server).
- `cargo bclippy` (workspace, `-D warnings`) — clean.
- `cargo bfmt` (workspace) — clean.
- Full PR validation pipeline (Linux/Linux ARM/Windows/Windows ARM/macOS builds and tests, Kerberos, mssql-python cross-repo, coverage) — all green.

## Related Issues

Fixes #400
Fixes #401
Relates to #422 (filed from this PR's review, tracking the `LIVE_HANDLES` ABA follow-up)
Relates to #430 (filed from this PR's review, tracking the `into_inner()` instructions carve-out)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (`cargo nextest run -p mssql-odbc --lib`: 1098 passed, 1 skipped)
- [x] Public API changes are documented (no *new* public API, but `SQLFreeHandle`'s behavior on a live handle of the wrong type changed from UB to a spec-compliant `SQL_INVALID_HANDLE` — see above)

### DIFF
--- a/mssql-odbc/src/api/disconnect.rs
+++ b/mssql-odbc/src/api/disconnect.rs
@@ -84,22 +84,29 @@ fn sql_disconnect_safe(dbc: &DbcHandle) -> SqlReturn {
     // no locks during I/O, so it can race here and access a STMT handle we are about to free.
     // TODO: fix with refcounted handle lifetimes so STMT handles cannot be freed while in use.
     //
-    // Pops (rather than iterating then clearing) so a poisoned STMT mutex
-    // leaves only the not-yet-freed remainder in `state.statements`: a
-    // retried `SQLDisconnect` — a reasonable response to the `SQL_ERROR` this
-    // returns — picks up where this one stopped instead of double-freeing the
-    // statements already dropped this pass (mssql-rs#401). Matches the
-    // descriptor loop directly below, which already does this.
+    // Locking each STMT's own mutex here is pure synchronization, not a read
+    // of its data: by the time `.lock()` returns at all (`Ok` or `Err`), no
+    // other thread can be mid-operation holding it, since a poisoning panic
+    // still fully releases the lock on unwind. A poisoned outcome therefore
+    // doesn't change whether it's safe to free the box, only whether the
+    // STMT's *contents* were left consistent — irrelevant here, since the box
+    // is being dropped whole. Treating it as fatal (returning `SQL_ERROR`
+    // without freeing) would orphan the handle: it would no longer be in
+    // `state.statements` for a retry to find, yet never actually freed either
+    // — reachable by no future call at all (mssql-rs#401 follow-up review).
+    // Tolerating it and freeing anyway, matching `free_stmt`'s own identical
+    // tolerance for its handle's lock, closes that gap outright instead of
+    // just deferring it.
     while let Some(stmt_ptr) = state.statements.pop() {
         // SAFETY: `stmt_ptr` came from `handle_to_raw::<StmtHandle>` and is still
-        // live (the DBC owns it). Acquire the STMT lock to serialize with any
-        // op still holding it, then drop the box.
+        // live (the DBC owns it).
         let stmt = unsafe { handle_from_raw::<StmtHandle>(stmt_ptr) };
-        let Ok(guard) = stmt.inner.lock() else {
-            error!(?stmt_ptr, "SQLDisconnect: stmt mutex poisoned");
-            return SQL_ERROR;
-        };
-        drop(guard);
+        if stmt.inner.lock().is_err() {
+            error!(
+                ?stmt_ptr,
+                "SQLDisconnect: stmt mutex poisoned; freeing anyway"
+            );
+        }
         unsafe { free_handle::<StmtHandle>(stmt_ptr) };
     }
 
@@ -111,20 +118,19 @@ fn sql_disconnect_safe(dbc: &DbcHandle) -> SqlReturn {
     // descriptor-allocation-node list the same way, after its own statement
     // loop (`sqlcconn.cpp:1313-1448`).
     //
-    // Pops (rather than iterating then clearing) so a poisoned DESC mutex
-    // leaves only the not-yet-freed remainder in `state.descriptors`: a
-    // retried `SQLDisconnect` picks up where this one stopped instead of
-    // re-freeing entries this pass already dropped.
+    // Same poison-tolerant shape as the statement loop above, for the same
+    // reason: synchronization only, and treating poison as fatal here would
+    // equally orphan the descriptor rather than actually resolve anything.
     while let Some(desc_ptr) = state.descriptors.pop() {
         // SAFETY: `desc_ptr` came from `handle_to_raw::<DescHandle>` and is
-        // still live (the DBC owns it). Acquire the DESC lock to serialize
-        // with any op still holding it, then drop the box.
+        // still live (the DBC owns it).
         let desc = unsafe { handle_from_raw::<DescHandle>(desc_ptr) };
-        let Ok(guard) = desc.inner.lock() else {
-            error!(?desc_ptr, "SQLDisconnect: desc mutex poisoned");
-            return SQL_ERROR;
-        };
-        drop(guard);
+        if desc.inner.lock().is_err() {
+            error!(
+                ?desc_ptr,
+                "SQLDisconnect: desc mutex poisoned; freeing anyway"
+            );
+        }
         unsafe { free_handle::<DescHandle>(desc_ptr) };
     }
 
@@ -249,17 +255,23 @@ mod tests {
         }
     }
 
-    /// Reproduces mssql-rs#401: a poisoned STMT mutex partway through the
-    /// statement-free loop must leave every *not-yet-processed* statement
-    /// untouched for a retry, not just stop and leave the whole original list
-    /// in place. The pop-then-free shape (matching the descriptor loop
-    /// directly below it) removes each pointer from `state.statements`
-    /// *before* attempting to free it, so a retried `SQLDisconnect` never
-    /// re-walks — and re-frees — a statement this pass already dropped.
+    /// Regression test for a bug in an earlier version of this fix (caught
+    /// by review before merge, mssql-rs#415): treating a poisoned STMT mutex
+    /// as fatal (returning `SQL_ERROR` without freeing) while having already
+    /// popped it off `state.statements` orphaned the statement entirely — no
+    /// longer tracked for a retry to find, yet never actually freed either,
+    /// so no future call could ever reach it again. `SQLDisconnect` must
+    /// instead tolerate a poisoned STMT mutex the same way `free_stmt`
+    /// already tolerates one for its own handle: the lock here exists purely
+    /// to synchronize with any thread still operating on the statement, not
+    /// to read its (possibly inconsistent) contents, so poison doesn't make
+    /// it unsafe to free the box — it succeeds and frees every statement,
+    /// poisoned or not.
     #[test]
-    fn disconnect_retry_after_poisoned_stmt_mutex_does_not_double_free() {
+    fn disconnect_frees_a_statement_even_if_its_mutex_is_poisoned() {
         use crate::api::odbc_types::SQL_HANDLE_STMT;
         use crate::handles::dbc::ConnectionState;
+        use crate::handles::is_live;
 
         let mut env: SqlHandle = SQL_NULL_HANDLE;
         assert_eq!(
@@ -296,37 +308,23 @@ mod tests {
             SQL_SUCCESS
         );
 
-        // Poison stmt2's mutex — the loop pops in LIFO order, so stmt2 (the
-        // more recently pushed) is attempted first and fails immediately,
-        // before stmt1 is ever touched.
+        // Poison stmt2's mutex; stmt1's is left untouched.
         let stmt2_ref = unsafe { handle_from_raw::<StmtHandle>(stmt2) };
         let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
             let _guard = stmt2_ref.inner.lock().unwrap();
             panic!("poison the stmt lock");
         }));
 
-        assert_eq!(unsafe { sql_disconnect(dbc) }, SQL_ERROR);
-        {
-            let state = dbc_ref.inner.lock().unwrap();
-            // stmt2 was popped off (and so can no longer be double-freed by
-            // a retry) even though its own free could not proceed — a
-            // poisoned mutex can't be safely retried, so leaking it here is
-            // the same accepted trade-off the descriptor loop already makes.
-            // stmt1 was never reached and must still be tracked, untouched.
-            assert_eq!(state.statements, vec![stmt1]);
-            assert_eq!(state.connection_state, ConnectionState::Connected);
-        }
-
-        // Retry: connection_state is still Connected, so this is accepted
-        // and frees stmt1 exactly once. Before this fix, a retry here would
-        // re-walk the *original* two-element list collected before the first
-        // pass failed, re-freeing stmt1 a second time.
+        // Before the fix, this returned SQL_ERROR and left stmt2 neither
+        // tracked nor freed (permanently unreachable). Now it succeeds and
+        // frees both.
         assert_eq!(unsafe { sql_disconnect(dbc) }, SQL_SUCCESS);
         assert!(dbc_ref.inner.lock().unwrap().statements.is_empty());
-
-        // stmt2's box was deliberately never dropped (poisoned mutex —
-        // leaked, not freed); free it directly so this test doesn't leak.
-        unsafe { free_handle::<StmtHandle>(stmt2) };
+        assert!(!is_live(stmt1), "stmt1 (unpoisoned) must be actually freed");
+        assert!(
+            !is_live(stmt2),
+            "stmt2 (poisoned) must be actually freed too, not orphaned"
+        );
 
         unsafe {
             sql_free_handle(SQL_HANDLE_DBC, dbc);
--- a/mssql-odbc/src/api/disconnect.rs
+++ b/mssql-odbc/src/api/disconnect.rs
@@ -83,7 +83,14 @@ fn sql_disconnect_safe(dbc: &DbcHandle) -> SqlReturn {
     // the DBC lock). However, a call that already took the client and is mid-execute() holds
     // no locks during I/O, so it can race here and access a STMT handle we are about to free.
     // TODO: fix with refcounted handle lifetimes so STMT handles cannot be freed while in use.
-    for &stmt_ptr in &state.statements {
+    //
+    // Pops (rather than iterating then clearing) so a poisoned STMT mutex
+    // leaves only the not-yet-freed remainder in `state.statements`: a
+    // retried `SQLDisconnect` — a reasonable response to the `SQL_ERROR` this
+    // returns — picks up where this one stopped instead of double-freeing the
+    // statements already dropped this pass (mssql-rs#401). Matches the
+    // descriptor loop directly below, which already does this.
+    while let Some(stmt_ptr) = state.statements.pop() {
         // SAFETY: `stmt_ptr` came from `handle_to_raw::<StmtHandle>` and is still
         // live (the DBC owns it). Acquire the STMT lock to serialize with any
         // op still holding it, then drop the box.
@@ -95,7 +102,6 @@ fn sql_disconnect_safe(dbc: &DbcHandle) -> SqlReturn {
         drop(guard);
         unsafe { free_handle::<StmtHandle>(stmt_ptr) };
     }
-    state.statements.clear();
 
     // Drop all explicitly-allocated DESC handles, after the statements: an
     // explicit descriptor carries no back-pointer to whichever statements had
@@ -236,6 +242,91 @@ mod tests {
         // before the read) but not on macOS, whose allocator reused it
         // sooner. The connection no longer tracking the descriptor (checked
         // above) is the real, safely-observable contract.
+
+        unsafe {
+            sql_free_handle(SQL_HANDLE_DBC, dbc);
+            sql_free_handle(SQL_HANDLE_ENV, env);
+        }
+    }
+
+    /// Reproduces mssql-rs#401: a poisoned STMT mutex partway through the
+    /// statement-free loop must leave every *not-yet-processed* statement
+    /// untouched for a retry, not just stop and leave the whole original list
+    /// in place. The pop-then-free shape (matching the descriptor loop
+    /// directly below it) removes each pointer from `state.statements`
+    /// *before* attempting to free it, so a retried `SQLDisconnect` never
+    /// re-walks — and re-frees — a statement this pass already dropped.
+    #[test]
+    fn disconnect_retry_after_poisoned_stmt_mutex_does_not_double_free() {
+        use crate::api::odbc_types::SQL_HANDLE_STMT;
+        use crate::handles::dbc::ConnectionState;
+
+        let mut env: SqlHandle = SQL_NULL_HANDLE;
+        assert_eq!(
+            unsafe { sql_alloc_handle(SQL_HANDLE_ENV, SQL_NULL_HANDLE, &mut env) },
+            SQL_SUCCESS
+        );
+        assert_eq!(
+            unsafe {
+                sql_set_env_attr(
+                    env,
+                    SQL_ATTR_ODBC_VERSION,
+                    SQL_OV_ODBC3_80 as usize as *mut std::ffi::c_void,
+                    0,
+                )
+            },
+            SQL_SUCCESS
+        );
+        let mut dbc: SqlHandle = SQL_NULL_HANDLE;
+        assert_eq!(
+            unsafe { sql_alloc_handle(SQL_HANDLE_DBC, env, &mut dbc) },
+            SQL_SUCCESS
+        );
+        let dbc_ref = unsafe { handle_from_raw::<DbcHandle>(dbc) };
+        dbc_ref.inner.lock().unwrap().connection_state = ConnectionState::Connected;
+
+        let mut stmt1: SqlHandle = SQL_NULL_HANDLE;
+        assert_eq!(
+            unsafe { sql_alloc_handle(SQL_HANDLE_STMT, dbc, &mut stmt1) },
+            SQL_SUCCESS
+        );
+        let mut stmt2: SqlHandle = SQL_NULL_HANDLE;
+        assert_eq!(
+            unsafe { sql_alloc_handle(SQL_HANDLE_STMT, dbc, &mut stmt2) },
+            SQL_SUCCESS
+        );
+
+        // Poison stmt2's mutex — the loop pops in LIFO order, so stmt2 (the
+        // more recently pushed) is attempted first and fails immediately,
+        // before stmt1 is ever touched.
+        let stmt2_ref = unsafe { handle_from_raw::<StmtHandle>(stmt2) };
+        let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            let _guard = stmt2_ref.inner.lock().unwrap();
+            panic!("poison the stmt lock");
+        }));
+
+        assert_eq!(unsafe { sql_disconnect(dbc) }, SQL_ERROR);
+        {
+            let state = dbc_ref.inner.lock().unwrap();
+            // stmt2 was popped off (and so can no longer be double-freed by
+            // a retry) even though its own free could not proceed — a
+            // poisoned mutex can't be safely retried, so leaking it here is
+            // the same accepted trade-off the descriptor loop already makes.
+            // stmt1 was never reached and must still be tracked, untouched.
+            assert_eq!(state.statements, vec![stmt1]);
+            assert_eq!(state.connection_state, ConnectionState::Connected);
+        }
+
+        // Retry: connection_state is still Connected, so this is accepted
+        // and frees stmt1 exactly once. Before this fix, a retry here would
+        // re-walk the *original* two-element list collected before the first
+        // pass failed, re-freeing stmt1 a second time.
+        assert_eq!(unsafe { sql_disconnect(dbc) }, SQL_SUCCESS);
+        assert!(dbc_ref.inner.lock().unwrap().statements.is_empty());
+
+        // stmt2's box was deliberately never dropped (poisoned mutex —
+        // leaked, not freed); free it directly so this test doesn't leak.
+        unsafe { free_handle::<StmtHandle>(stmt2) };
 
         unsafe {
             sql_free_handle(SQL_HANDLE_DBC, dbc);
--- a/mssql-odbc/src/api/disconnect.rs
+++ b/mssql-odbc/src/api/disconnect.rs
@@ -85,18 +85,15 @@ fn sql_disconnect_safe(dbc: &DbcHandle) -> SqlReturn {
     // TODO: fix with refcounted handle lifetimes so STMT handles cannot be freed while in use.
     //
     // Locking each STMT's own mutex here is pure synchronization, not a read
-    // of its data: by the time `.lock()` returns at all (`Ok` or `Err`), no
-    // other thread can be mid-operation holding it, since a poisoning panic
-    // still fully releases the lock on unwind. A poisoned outcome therefore
-    // doesn't change whether it's safe to free the box, only whether the
-    // STMT's *contents* were left consistent — irrelevant here, since the box
-    // is being dropped whole. Treating it as fatal (returning `SQL_ERROR`
-    // without freeing) would orphan the handle: it would no longer be in
-    // `state.statements` for a retry to find, yet never actually freed either
-    // — reachable by no future call at all (mssql-rs#401 follow-up review).
-    // Tolerating it and freeing anyway, matching `free_stmt`'s own identical
-    // tolerance for its handle's lock, closes that gap outright instead of
-    // just deferring it.
+    // of its data: a poisoning panic still fully releases the lock on
+    // unwind, so a poisoned outcome doesn't change whether it's safe to free
+    // the box, only whether the STMT's *contents* were left consistent —
+    // irrelevant here, since the box is dropped whole. Treating it as fatal
+    // instead (returning `SQL_ERROR` without freeing) would orphan the
+    // handle: no longer in `state.statements` for a retry to find, yet never
+    // freed either. Tolerating it and freeing anyway, matching `free_stmt`'s
+    // identical tolerance for its own handle's lock, closes that gap instead
+    // of deferring it.
     while let Some(stmt_ptr) = state.statements.pop() {
         // SAFETY: `stmt_ptr` came from `handle_to_raw::<StmtHandle>` and is still
         // live (the DBC owns it).
@@ -255,18 +252,14 @@ mod tests {
         }
     }
 
-    /// Regression test for a bug in an earlier version of this fix (caught
-    /// by review before merge, mssql-rs#415): treating a poisoned STMT mutex
-    /// as fatal (returning `SQL_ERROR` without freeing) while having already
-    /// popped it off `state.statements` orphaned the statement entirely — no
-    /// longer tracked for a retry to find, yet never actually freed either,
-    /// so no future call could ever reach it again. `SQLDisconnect` must
-    /// instead tolerate a poisoned STMT mutex the same way `free_stmt`
-    /// already tolerates one for its own handle: the lock here exists purely
-    /// to synchronize with any thread still operating on the statement, not
-    /// to read its (possibly inconsistent) contents, so poison doesn't make
-    /// it unsafe to free the box — it succeeds and frees every statement,
-    /// poisoned or not.
+    /// `SQLDisconnect` must tolerate a poisoned STMT mutex instead of
+    /// treating it as fatal: the lock exists purely to synchronize with any
+    /// thread still operating on the statement, not to read its (possibly
+    /// inconsistent) contents, so poison doesn't make it unsafe to free the
+    /// box. Matches `free_stmt`'s identical tolerance for its own handle's
+    /// lock — treating it as fatal here instead would orphan the statement:
+    /// no longer tracked in `state.statements` for a retry to find, yet
+    /// never freed either.
     #[test]
     fn disconnect_frees_a_statement_even_if_its_mutex_is_poisoned() {
         use crate::api::odbc_types::SQL_HANDLE_STMT;
--- a/mssql-odbc/src/api/free_handle.rs
+++ b/mssql-odbc/src/api/free_handle.rs
@@ -320,9 +320,14 @@ unsafe fn free_desc(handle: SqlHandle) -> SqlReturn {
     // registered in `dbc_state.descriptors`, by design) fall through to the
     // "live handle missing from parent DBC" branch further down, which is
     // meant to catch a genuine invariant break, not this legitimate case.
+    // `free_errors` runs in the same guarded block as `post_diag`, not after
+    // it: every ODBC entry point must clear stale diagnostics at API entry
+    // before posting a new one, and this branch returns before ever reaching
+    // the `free_errors` call further below.
     if !desc.is_explicit() {
         error!("SQLFreeHandle(DESC): cannot free an implicitly allocated descriptor");
         if let Ok(mut state) = desc.inner.lock() {
+            free_errors(&mut state);
             post_diag(&mut state, ERR_INVALID_USE_OF_AUTO_DESC);
         }
         return SQL_ERROR;
@@ -1005,6 +1010,14 @@ mod tests {
     /// the descriptor handle `SQLFreeHandle` tried (and failed) to free, per
     /// its own spec ("If SQLFreeHandle returns SQL_ERROR, the handle is still
     /// valid"), not on the parent STMT or DBC.
+    ///
+    /// Calls the rejection twice and asserts exactly one diag record after
+    /// each call, not just that the *last* one is HY017: this entry point
+    /// must clear stale diagnostics at API entry like every other one (a
+    /// prior version of this fix regressed exactly this by hoisting the
+    /// rejection above the `free_errors` call — caught in review, not by
+    /// this test, since asserting only `.last()` passes at any record
+    /// count).
     #[test]
     fn free_implicit_desc_is_rejected() {
         use crate::api::sqlstate::SQLSTATE_HY017;
@@ -1017,13 +1030,20 @@ mod tests {
         );
         let ard = unsafe { &*(stmt as *const StmtHandle) }.ard;
 
-        let ret = unsafe { sql_free_handle(SQL_HANDLE_DESC, ard) };
-        assert_eq!(ret, SQL_ERROR);
+        for _ in 0..2 {
+            assert_eq!(unsafe { sql_free_handle(SQL_HANDLE_DESC, ard) }, SQL_ERROR);
 
-        let desc = unsafe { handle_from_raw::<DescHandle>(ard) };
-        let diag = desc.inner.lock().unwrap();
-        assert_eq!(diag.diag_records.last().unwrap().sql_state, SQLSTATE_HY017);
-        drop(diag);
+            let desc = unsafe { handle_from_raw::<DescHandle>(ard) };
+            let diag = desc.inner.lock().unwrap();
+            assert_eq!(
+                diag.diag_records.len(),
+                1,
+                "a fresh rejection must clear any diagnostic left by the previous call, \
+                 not accumulate one record per call"
+            );
+            assert_eq!(diag.diag_records[0].sql_state, SQLSTATE_HY017);
+            drop(diag);
+        }
 
         // The ARD is untouched: the statement (and its implicit descriptors)
         // still free normally.
--- a/mssql-odbc/src/api/free_handle.rs
+++ b/mssql-odbc/src/api/free_handle.rs
@@ -14,7 +14,8 @@ use crate::api::sqlstate::{ERR_INVALID_USE_OF_AUTO_DESC, SQLSTATE_HY000, post_di
 use crate::error::{free_errors, post_sql_error};
 use crate::handles::stmt::STMT_STATE_CURSOR_OPEN;
 use crate::handles::{
-    DbcHandle, DescHandle, EnvHandle, HandleType, StmtHandle, free_handle, handle_from_raw, is_live,
+    DbcHandle, DescHandle, EnvHandle, HandleType, StmtHandle, free_handle, handle_from_raw,
+    live_type,
 };
 use mssql_tds::connection::tds_client::StatementId;
 
@@ -157,17 +158,30 @@ unsafe fn free_dbc(handle: SqlHandle) -> SqlReturn {
 /// cleanup (`sql_disconnect_safe`), returns `SQL_SUCCESS` immediately without
 /// dereferencing it — that cleanup happens behind the Driver Manager's back,
 /// so an application legitimately holding this now-stale handle can still
-/// reach here (mssql-rs#400). If the STMT is somehow not found in the parent
-/// DBC's statement list despite still being live, it was already dropped by
-/// some other path — also returns `SQL_SUCCESS` without calling `free_handle`
-/// a second time.
+/// reach here (mssql-rs#400). Returns `SQL_INVALID_HANDLE`, per spec, if the
+/// handle is live but of some other type. If the STMT is live and correctly
+/// typed but somehow missing from its parent DBC's statement list — an
+/// invariant break rather than an expected path, since "already freed" is
+/// caught above — logs it and still returns `SQL_SUCCESS` rather than
+/// double-freeing or panicking.
 ///
 /// # Safety
 /// `handle` must be a live `StmtHandle` created by `alloc_stmt`.
 unsafe fn free_stmt(handle: SqlHandle) -> SqlReturn {
-    if !is_live(handle) {
-        debug!(?handle, "SQLFreeHandle(STMT): handle already freed, no-op");
-        return SQL_SUCCESS;
+    match live_type(handle) {
+        None => {
+            debug!(?handle, "SQLFreeHandle(STMT): handle already freed, no-op");
+            return SQL_SUCCESS;
+        }
+        Some(HandleType::Stmt) => {}
+        Some(actual) => {
+            error!(
+                ?handle,
+                ?actual,
+                "SQLFreeHandle(STMT): handle is live but not a STMT"
+            );
+            return SQL_INVALID_HANDLE;
+        }
     }
 
     let stmt = unsafe { handle_from_raw::<StmtHandle>(handle) };
@@ -203,7 +217,14 @@ unsafe fn free_stmt(handle: SqlHandle) -> SqlReturn {
             return SQL_ERROR;
         };
         let Some(i) = dbc_state.statements.iter().position(|&p| p == handle) else {
-            // Already dropped by SQLDisconnect - early return.
+            // Live but untracked: not the post-SQLDisconnect case (caught by
+            // `live_type` above), so something removed it from the parent
+            // without freeing it — an invariant break, not an expected path.
+            error!(
+                ?handle,
+                "SQLFreeHandle(STMT): live handle missing from parent DBC"
+            );
+            debug_assert!(false, "live STMT not tracked by its parent DBC");
             return SQL_SUCCESS;
         };
         dbc_state.statements.swap_remove(i);
@@ -226,17 +247,30 @@ unsafe fn free_stmt(handle: SqlHandle) -> SqlReturn {
 /// If the handle was already freed by an earlier `SQLDisconnect` cascading
 /// cleanup (`sql_disconnect_safe`), returns `SQL_SUCCESS` immediately without
 /// dereferencing it — mirrors `free_stmt`'s identical guard and the same
-/// reasoning (mssql-rs#400). If the descriptor is somehow not found in the
-/// parent DBC's descriptor list despite still being live, it was already
-/// dropped by some other path — also returns `SQL_SUCCESS` without calling
-/// `free_handle` a second time.
+/// reasoning (mssql-rs#400). Returns `SQL_INVALID_HANDLE`, per spec, if the
+/// handle is live but of some other type. If the descriptor is live and
+/// correctly typed but somehow missing from its parent DBC's descriptor
+/// list — an invariant break rather than an expected path, since "already
+/// freed" is caught above — logs it and still returns `SQL_SUCCESS` rather
+/// than double-freeing or panicking.
 ///
 /// # Safety
 /// `handle` must be a live `DescHandle` created by `alloc_desc`.
 unsafe fn free_desc(handle: SqlHandle) -> SqlReturn {
-    if !is_live(handle) {
-        debug!(?handle, "SQLFreeHandle(DESC): handle already freed, no-op");
-        return SQL_SUCCESS;
+    match live_type(handle) {
+        None => {
+            debug!(?handle, "SQLFreeHandle(DESC): handle already freed, no-op");
+            return SQL_SUCCESS;
+        }
+        Some(HandleType::Desc) => {}
+        Some(actual) => {
+            error!(
+                ?handle,
+                ?actual,
+                "SQLFreeHandle(DESC): handle is live but not a DESC"
+            );
+            return SQL_INVALID_HANDLE;
+        }
     }
 
     let desc = unsafe { handle_from_raw::<DescHandle>(handle) };
@@ -271,7 +305,14 @@ unsafe fn free_desc(handle: SqlHandle) -> SqlReturn {
     };
 
     let Some(i) = dbc_state.descriptors.iter().position(|&p| p == handle) else {
-        // Already dropped by SQLDisconnect - early return.
+        // Live but untracked: not the post-SQLDisconnect case (caught by
+        // `live_type` above), so something removed it from the parent
+        // without freeing it — an invariant break, not an expected path.
+        error!(
+            ?handle,
+            "SQLFreeHandle(DESC): live handle missing from parent DBC"
+        );
+        debug_assert!(false, "live DESC not tracked by its parent DBC");
         return SQL_SUCCESS;
     };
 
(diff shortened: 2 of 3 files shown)


mssql-odbc/src/api/disconnect.rs
```diff
@@ -83,7 +83,14 @@ fn sql_disconnect_safe(dbc: &DbcHandle) -> SqlReturn {
...
     // TODO: fix with refcounted handle lifetimes so STMT handles cannot be freed while in use.
-    for &stmt_ptr in &state.statements {
+    //
+    // Pops (rather than iterating then clearing) so a poisoned STMT mutex
+    // leaves only the not-yet-freed remainder in `state.statements`: a
+    // retried `SQLDisconnect` — a reasonable response to the `SQL_ERROR` this
+    // returns — picks up where this one stopped instead of double-freeing the
+    // statements already dropped this pass (mssql-rs#401). Matches the
+    // descriptor loop directly below, which already does this.
+    while let Some(stmt_ptr) = state.statements.pop() {
```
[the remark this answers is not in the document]
Vahid-b: Good catch — pushing back and still returning SQL_ERROR would also have left this permanently unresolved, since a poisoned mutex never un-poisons for a retry to succeed against.

Fixed by tolerating the poison and freeing anyway instead: the lock in this loop is pure synchronization (we never read the guard's contents, only drop it), and a poisoning panic still fully releases the underlying lock on unwind before `.lock()` returns `Err`. So a poisoned outcome doesn't change whether it's safe to free the box — it only means the STMT's own contents were left inconsistent, which is irrelevant since the whole box is being dropped. This matches `free_stmt`/`free_desc`'s existing tolerance for their own handle's lock, applied to both the statement and descriptor loops here. Rewrote the test (`disconnect_frees_a_statement_even_if_its_mutex_is_poisoned`) to assert `SQLDisconnect` now succeeds and both statements are actually freed (checked via `is_live`), reverted the fix locally to confirm the test fails with the old orphaning behavior, then restored it.

> [!NOTE]

mssql-odbc/src/handles/mod.rs
```diff
@@ -29,10 +31,52 @@ pub(crate) enum HandleType {
...
+/// Safe to call on any pointer value, including a stale one whose memory may
+/// already be deallocated or reused: this never dereferences `raw`, so
+/// callers can use it to decide whether dereferencing is safe in the first
+/// place. If the registry lock is itself poisoned, conservatively reports
+/// "not live" — treating an ambiguous handle as already-freed risks a leak,
+/// treating it as live risks a genuine use-after-free.
+pub(crate) fn is_live(raw: *mut c_void) -> bool {
+    LIVE_HANDLES
+        .lock()
+        .map(|live| live.contains(&(raw as usize)))
```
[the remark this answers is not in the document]
Vahid-b: Correct, and this is a real residual gap I don't think this PR should try to close. A true fix needs allocation identity that survives address reuse (generation-tracked indirection / stable indices instead of raw addresses), which is the same class of redesign as the pre-existing refcounted-handle-lifetime TODO in `disconnect.rs` — out of scope for this narrow bug-fix PR.

Instead of leaving it silently unaddressed: expanded the doc comment on `LIVE_HANDLES` to explicitly document this ABA limitation (distinct from the already-documented concurrent-free TOCTOU), added `is_live_cannot_distinguish_a_reused_address_from_the_original_allocation`, which demonstrates the exact mechanism directly against the registry (free a handle, then simulate the allocator reusing that address for an unrelated new handle, and show `is_live` reports the stale original as live), and filed #422 to track the real fix.

This registry still closes the specific, reproducible bug in #400 (freed-then-refreed-again with no intervening reallocation) and is strictly no worse than before it existed — every stale-handle free previously dereferenced unconditionally with no check at all — it just doesn't close the rarer address-reuse sub-case, which #422 tracks.

> [!NOTE]

mssql-odbc/src/handles/mod.rs
```diff
@@ -29,10 +31,80 @@ pub(crate) enum HandleType {
...
+        .unwrap_or(false)
```
Vahid-b: Nit, not blocking: `is_live`'s poisoned-lock fallback (`unwrap_or(false)`) is reasonable in isolation, but combined with `handle_to_raw`'s poisoned-lock fallback (silently skips the `insert`), a poisoned `LIVE_HANDLES` mutex would make `free_stmt`/`free_desc` treat *every* handle, forever, as already-freed — a permanent silent leak for the rest of the process, not just a degraded single call. Realistically near-impossible to trigger (the critical sections are single `HashSet` ops with no I/O or user code that could panic), so I wouldn't block on it, but might be worth a one-line comment acknowledging the blast radius so it isn't rediscovered as a mystery later.

Vahid-b (commented):
Full review notes (draft PR review, posted per repo review workflow). This was posted autonomously while the author was away, after the head moved from `c1235c5` to `db29e0f4` mid-review — re-verified everything against the new commit before posting rather than reusing the stale draft.

Both issues Copilot flagged (the orphan-on-poison leak in the statement/descriptor loops, and the `LIVE_HANDLES` address-reuse/ABA gap) are already fixed/addressed in `db29e0f4`, and both threads already have author replies. I independently re-verified rather than trusting the commit message:

- **Orphan-on-poison fix**: confirmed `disconnect_frees_a_statement_even_if_its_mutex_is_poisoned` passes on the current head, then reverted the fix locally (re-added the early `return SQL_ERROR` on a poisoned lock) and confirmed the test fails (`assertion left==right failed, left: -1 right: 0`) — it genuinely guards the regression, not just passes incidentally.
- **ABA follow-up**: confirmed `mssql-rs#422` actually exists and is open (not just referenced in a comment).

No blocking issues remain from my pass. One low-priority nit below.

mssql-odbc/src/handles/mod.rs
Vahid-b: Fair point — added a short note directly on `is_live`'s doc comment explaining the blast radius: since a poisoned `Mutex` never un-poisons and `handle_to_raw` makes the matching choice on the insert side, this wouldn't just degrade one call but would make every handle report not-live for the rest of the process. Left the actual behavior as-is (leak-over-UAF is still the right call), just documented it so it isn't rediscovered as a mystery.

> [!NOTE]

David-Engel (commented):
## Summary

Reviewed the full `origin/main...1cf909a6` diff across `handles/mod.rs`, `api/free_handle.rs`, and `api/disconnect.rs`. Both fixes hold up, the reasoning in the doc comments matches what the code actually does, and the three regression tests genuinely guard the change. No findings.

**Blocking:** none. **Suggestions:** none. **Nits:** none.

## What I verified

**#400 — `is_live` guard before dereference.** `handle_to_raw` / `free_handle` are the only two paths that hand out and reclaim a raw handle in this crate (checked every `Box::into_raw` / `Box::from_raw` site under `mssql-odbc/src`, including `StmtHandle::drop`'s four implicit descriptors at `handles/stmt.rs:1030`), so the registry cannot drift out of sync with reality. `is_live` never dereferences its argument, and `sql_free_handle` still rejects a null handle with `SQL_INVALID_HANDLE` before dispatch (`free_handle.rs:29`), so the new early-return doesn't swallow that case.

**Lock ordering.** `LIVE_HANDLES` is a leaf: both `is_live` and `free_handle` acquire and release it before touching the handle, and `free_handle` releases it before `Box::from_raw` runs `StmtHandle::drop`, which re-enters `free_handle` for the implicit descriptors. No re-entrancy deadlock, and no path takes `LIVE_HANDLES` before a handle mutex.

**#401 — poison tolerance in both `SQLDisconnect` loops.** The argument in the comment is correct: `Mutex::lock` returning `Err(PoisonError)` still means the lock was acquired, and the poisoning guard was released during unwind, so serialization is achieved either way and nothing reads the guard's contents before the box is dropped whole. The `if <expr>` condition is its own temporary scope, so the guard (or the guard inside `PoisonError`) drops before `free_handle` is called on the next line. Popping instead of iterate-then-clear also means neither loop can leave a freed pointer in `state.statements` / `state.descriptors` for a retried `SQLDisconnect` to re-free.

**Tests actually guard the fixes.** I mutated each fix out and reran the affected tests locally (`cargo nextest run -p mssql-odbc --lib`, `CARGO_TARGET_DIR` pinned to a scratch dir):

| Mutation | Result |
| --- | --- |
| `is_live` guard disabled in `free_stmt` | `free_stmt_twice_is_a_no_op_not_a_panic` **aborts** with `0xC0000005` (access violation) |
| `is_live` guard disabled in `free_desc` | `free_desc_twice_is_a_no_op_not_a_panic` fails |
| `return SQL_ERROR` restored on poisoned STMT lock | `disconnect_frees_a_statement_even_if_its_mutex_is_poisoned` fails (`-1` vs `0`) |

All six new/affected tests pass unmutated. Worth noting the first row is stronger evidence than the PR description claims: on Windows the pre-fix second free doesn't just trip the `debug_assert_eq!` and get caught by `ffi_entry!` — it faults outright, which is the use-after-free #400 describes, observed directly rather than inferred.

**Prior review threads.** Copilot's two findings (orphaned-on-poison statement, and `LIVE_HANDLES` address reuse) are both real and both addressed — the first fixed outright by tolerating poison, the second documented on `LIVE_HANDLES`, demonstrated by `is_live_cannot_distinguish_a_reused_address_from_the_original_allocation`, and tracked in #422. I re-derived both mechanisms against the code rather than taking the replies at face value; the disposition is right in each case. Deferring the ABA fix is consistent with the pre-existing refcounted-handle-lifetime TODO in `disconnect.rs` — fixing address identity in the free paths alone, while every other entry point still dereferences a raw handle unchecked, would be a worse shape than scheduling it as its own change.

**CI.** All required checks green on this commit, including the full ADO validation (all five platforms, Kerberos, and the cross-repo `mssql-python` suite) and the coverage report (99% diff, 93.1% overall). Nothing contradicts the PR checklist.

## Considered and not filed

- `free_env` / `free_dbc` don't take the new `is_live` guard, so a double free of an ENV or DBC still dereferences freed memory. Not filed: neither is cascade-freed behind the Driver Manager's back the way `sql_disconnect_safe` frees STMTs and DESCs, so there's no legitimate application path that produces a stale one — and adding a guard the reference driver doesn't have isn't obviously right here.
- The global registry mutex is now taken on every handle alloc and free (five times per statement, counting implicit descriptors). Handle lifecycle isn't a hot path and the critical section is a single `HashSet` op, so this isn't worth restructuring without a measurement.

---
Posted by an unattended review run — these findings were not checked by a human before posting. No inline comments, no label change, and nothing resolved.

mssql-odbc/src/handles/mod.rs
```diff
@@ -29,10 +31,91 @@ pub(crate) enum HandleType {
...
+/// registry. That is squarely the same class of redesign as the existing
+/// refcounted-handle-lifetime TODO (`disconnect.rs`), tracked separately as
+/// mssql-rs#422 rather than attempted here.
+///
+/// Deliberately narrower than that redesign in a second way too: this only
+/// distinguishes "already freed" from "still live," not "live but
+/// concurrently being freed on another thread right now" — that TOCTOU
+/// window is the pre-existing, wider concurrent-use race this crate already
+/// documents and does not attempt to close here either.
+static LIVE_HANDLES: LazyLock<Mutex<HashSet<usize>>> = LazyLock::new(|| Mutex::new(HashSet::new()));
```
shiwanigupta0809: Suggestion: consider keying this `HashMap<usize, HandleType>` rather than `HashSet<usize>` — it would close a second, cheaper hole than the ABA one, and this PR is already building the mechanism.

`sql_free_handle` dispatches on the caller-supplied `handle_type` with no validation, so `SQLFreeHandle(SQL_HANDLE_STMT, some_live_DESC)` reaches `free_stmt`, `is_live` answers `true` — it *is* live, just not a STMT — and `handle_from_raw::<StmtHandle>` then reinterprets a `DescHandle`. The only thing standing in the way is:

```rust
debug_assert_eq!(stmt.object_type, HandleType::Stmt, "SQLFreeHandle(STMT): handle is not a STMT");
```

which doesn't hold up on either count: it compiles to nothing in release, and it reads `.object_type` *through the already-wrongly-typed reference*. None of `StmtHandle` / `DescHandle` / `DbcHandle` / `EnvHandle` is `#[repr(C)]`, so their field offsets aren't guaranteed to agree and that read is itself UB — the type check can't be trusted even in a debug build.

The spec has a defined answer the driver currently can't produce: `SQLFreeHandle` returns `SQL_INVALID_HANDLE` when "the handle indicated by *Handle* was not a valid handle of the type indicated by *HandleType*."

Worth separating this from #422: that one needs generation-tracked identity to survive address reuse, which is genuinely a redesign. This case needs none of that — no reallocation is involved at all, just a mismatched `HandleType` on a currently-live handle — so a type tag alone fully closes it. All four handle types already implement `HasObjectType`, so it's about as mechanical as it gets: add the bound to `handle_to_raw`, record `*handle.object_type_mut()` alongside the address, and have `free_stmt`/`free_desc` check the recorded type before dereferencing.

Pre-existing, so entirely reasonable as a follow-up rather than here — but it converts a type-confusion UB into the spec's own return code, and it makes the four `debug_assert_eq!`s enforceable instead of decorative.

mssql-odbc/src/handles/mod.rs
```diff
@@ -29,10 +31,91 @@ pub(crate) enum HandleType {
...
+/// making `free_stmt`/`free_desc` skip freeing them all. In practice this
+/// needs a panic while holding `LIVE_HANDLES`'s lock, whose critical
+/// sections are a single `HashSet` op each with no user code or I/O that
+/// could panic, so it's realistically very unlikely — documented here so it
+/// isn't rediscovered as a mystery leak later.
+pub(crate) fn is_live(raw: *mut c_void) -> bool {
+    LIVE_HANDLES
+        .lock()
+        .map(|live| live.contains(&(raw as usize)))
+        .unwrap_or(false)
```
shiwanigupta0809: Suggestion: `.unwrap_or_else(|poisoned| poisoned.into_inner())` here (and at the `insert` / `remove` sites) would delete this failure mode instead of documenting it.

The doc comment above spends ~10 lines on the consequence, and it's right about the severity: because `is_live` reports `false` on a poisoned registry and `handle_to_raw` silently skips registering, one poisoning permanently turns every `free_stmt` / `free_desc` into a no-op, so every STMT and DESC leaks for the rest of the process. Calling that out rather than leaving it to be rediscovered as a mystery leak is the right instinct.

But this crate already has the idiom for exactly this shape — a process-global `Mutex` over a plain collection with no invariants to protect:

```rust
// auth/msqa.rs:311-313
let mut cache = cache
    .lock()
    .unwrap_or_else(|poisoned| poisoned.into_inner());
```

Same again at `auth/msqa.rs:688` and `auth/entra.rs:272`.

Recovering is sound here for a stronger reason than at those sites: `LIVE_HANDLES`'s critical sections are a single `insert` / `remove` / `contains` on a `HashSet<usize>` with no user code, no I/O and no user-supplied `Hash` impl, and allocation failure aborts rather than unwinds — so there is no panic that can leave the set half-updated, and arguably no way for it to poison itself at all. Which cuts both ways: if the poison genuinely can't happen, `into_inner()` is trivially safe and the paragraph documenting the leak can go; if it somehow can, `into_inner()` is still the better outcome than leaking every handle.

Either way the conservative `false` doesn't buy the safety it's reaching for. It's chosen to avoid a use-after-free, but a poisoned registry means the *insert* side stopped recording too, so "not live" stops being a conservative answer and just becomes a wrong one uniformly.

shiwanigupta0809 (commented):
Reviewed the full diff across `handles/mod.rs`, `api/free_handle.rs` and `api/disconnect.rs`. Both fixes are correct and the registry invariant holds — I checked every `handle_to_raw` / `Box::into_raw` and `free_handle` / `Box::from_raw` site under `mssql-odbc/src`, and the two choke points really are exhaustive.

Things I specifically tried to break and could not:

- **The HY017 path still works.** My first concern was that the new `!is_live` early return in `free_desc` sits *above* the implicit-descriptor check, so a spec-required `HY017` could silently become `SQL_SUCCESS`. It doesn't: `StmtHandle::new` routes all four implicit descriptors through `handle_to_raw`, so they are registered, `is_live` is true for them, and `free_desc` still reaches `ERR_INVALID_USE_OF_AUTO_DESC`. Worth knowing that guard ordering is load-bearing on that registration.
- **The ABA doc comment's "not worse than before" claim is accurate.** On a reused address the pre-PR code would have matched the new handle in `dbc_state.statements.iter().position(|&p| p == handle)` and freed it just the same, so the registry doesn't introduce that hazard — it inherits it. Correctly scoped to #422.
- **The "considered and not filed" reasoning for `free_env` / `free_dbc` holds.** Neither cascades; both only `debug_assert!` their child list is empty, so there's no driver-side path that frees an ENV or DBC behind the DM's back the way `sql_disconnect_safe` does for STMT/DESC.

Two suggestions below, neither blocking and neither overlapping the existing threads or #422.

mssql-odbc/src/api/free_handle.rs
```diff
@@ -153,13 +153,23 @@ unsafe fn free_dbc(handle: SqlHandle) -> SqlReturn {
...
+/// so an application legitimately holding this now-stale handle can still
+/// reach here (mssql-rs#400). If the STMT is somehow not found in the parent
+/// DBC's statement list despite still being live, it was already dropped by
+/// some other path — also returns `SQL_SUCCESS` without calling `free_handle`
+/// a second time.
 ///
 /// # Safety
 /// `handle` must be a live `StmtHandle` created by `alloc_stmt`.
 unsafe fn free_stmt(handle: SqlHandle) -> SqlReturn {
+    if !is_live(handle) {
```
David-Engel: **Suggestion.** The `is_live` guard changes what the later parent-list lookup *means*, and the inline comment down there wasn't updated along with the doc comment.

`free_stmt` L205-208 (and `free_desc` L273-276) still say `// Already dropped by SQLDisconnect - early return.` — but that case is now caught up here and returns before the DBC lock is ever taken. Reaching the list-miss branch now means the opposite: the handle **is** still allocated, just untracked by its parent. Returning `SQL_SUCCESS` there leaks it permanently *and* leaves its address in `LIVE_HANDLES` forever, which pins the ABA window you documented open on that address for the rest of the process.

It is very hard to reach — you'd need a panic between `swap_remove` and `free_handle`, which `ffi_entry!` would swallow — so this isn't blocking. But since it went from "the expected path" to "a genuine invariant break", it's worth saying so and making it visible instead of silent:

```rust
let Some(i) = dbc_state.statements.iter().position(|&p| p == handle) else {
    // Live but untracked: not the post-SQLDisconnect case (is_live catches that
    // above), so something removed it from the parent without freeing it.
    error!(?handle, "SQLFreeHandle(STMT): live handle missing from parent DBC");
    debug_assert!(false, "live STMT not tracked by its parent DBC");
    return SQL_SUCCESS;
};
```

mssql-odbc/src/handles/mod.rs
```diff
@@ -29,10 +31,91 @@ pub(crate) enum HandleType {
     Invalid = 0xDEADBEEF,
 }
 
+/// Tracks every currently-allocated handle by raw address, independent of the
```
David-Engel: **Suggestion, non-blocking.** This is ~55 lines of doc comment for one `static` and a four-line function, and the same density repeats on the disconnect loop and on each new test — roughly 100 of the 355 added lines are prose. `.github/copilot-instructions.md` asks for doc comments that explain *why* and explicitly calls out narrative filler.

The parts that earn their place: the invariant (`is_live` never dereferences, which is exactly what lets it gate a dereference), the ABA limitation, and the poisoned-registry blast radius. All three were asked for in review and should stay.

What could go is the review and PR history embedded in the source — "flagged in review of #415", "confirmed reachable during PR #370's review, not merely hypothetical", "(mssql-rs#401 follow-up review)", "Regression test for a bug in an earlier version of this fix (caught by review before merge)". That story belongs in the PR and in #400 / #401 / #422, which a reader can follow from the issue links you already have; in the file it goes stale as soon as the surrounding code moves. Roughly halving this while keeping all three technical points would read better in six months.

mssql-odbc/src/handles/mod.rs
```diff
@@ -90,3 +176,70 @@ pub(crate) unsafe fn free_handle<T: HasObjectType>(raw: *mut c_void) {
...
+        assert!(!is_live(raw), "freed handle must not be reported live");
+
+        // Stand in for the allocator handing the same address back out for
+        // an unrelated new allocation, without going through
+        // `handle_to_raw` (there is no portable way to force the real
+        // allocator to reuse this exact address deterministically).
+        LIVE_HANDLES
+            .lock()
+            .expect("registry mutex should not be poisoned in this test")
+            .insert(raw as usize);
```
David-Engel: **Nit.** The test never removes what it inserts, so `raw` stays in the process-global `LIVE_HANDLES` permanently. Under `cargo nextest` each test gets its own process, so it can't affect anything today — but under a plain `cargo test` (threads sharing one process) it seeds a permanently phantom-live address into state every other test reads. Cheap to make self-contained by adding after the final assert:

```rust
LIVE_HANDLES
    .lock()
    .expect("registry mutex should not be poisoned in this test")
    .remove(&(raw as usize));
```

David-Engel (commented):
**Automated review (unattended run).** These findings were not checked by a human before posting.

## Summary

Two narrow lifetime fixes in `mssql-odbc` handle cleanup: a process-wide address registry (`LIVE_HANDLES` / `is_live`) so `free_stmt`/`free_desc` can reject an already-freed handle without dereferencing it, and a poison-tolerant pop-then-free loop in `sql_disconnect_safe` for both statements and descriptors. Both are correct as written, the regression tests go through the real entry points rather than the internals, and I found nothing blocking.

The three things most likely to be wrong in a change of this shape all hold:

- **Guard drop timing.** `if stmt.inner.lock().is_err() { ... }` depends on the `if` condition being a terminating scope, so the temporary `MutexGuard` / `PoisonError` is dropped before `free_handle` drops the box. Dropping a `Mutex` while a guard derived from it is live would be a real use-after-free; this shape can't produce one. Same for the DESC loop.
- **No lock inversion or re-entrancy.** `free_handle` releases the `LIVE_HANDLES` guard before it dereferences, which matters because `StmtHandle::drop` re-enters `free_handle` for its four implicit descriptors — that would self-deadlock if the registry guard were still held. `DescHandle`, `DbcHandle` and `EnvHandle` have no `Drop`, so freeing while holding the DBC lock in `sql_disconnect_safe` never reaches back for it either.
- **The choke points are actually complete.** `Box::into_raw` / `Box::from_raw` appear nowhere in `mssql-odbc` outside `handle_to_raw` / `free_handle`, and every allocation site (`alloc_env`/`alloc_dbc`/`alloc_stmt`/`alloc_desc`, plus `StmtHandle::new`'s four implicit descriptors) goes through `handle_to_raw`. No alloc path registers and then bails on a later failure, which would strand a permanently phantom-live address. So the registry can't drift out of sync with reality except through the ABA case already documented.

## Verification

Built and ran the affected tests locally on this head (`cargo nextest run -p mssql-odbc --lib`, filtered to the new tests plus the adjacent poison tests): 15 passed, including `disconnect_frees_a_statement_even_if_its_mutex_is_poisoned`, `free_stmt_twice_is_a_no_op_not_a_panic`, `free_desc_twice_is_a_no_op_not_a_panic`, and all three registry tests. Every required check is green on this head, including the full ADO PR validation across five platforms and both cross-repo mssql-python legs.

Not re-raised: the ABA / address-reuse gap and the poisoned-mutex orphan (Copilot threads, both addressed on this head), and @shiwanigupta0809's `HashMap<usize, HandleType>` type-confusion and `into_inner()` poison-recovery suggestions — both still open, and both deserve a reply before merge.

## Blocking

None found.

## Suggestion

Two, inline: the parent-list lookup miss in `free_stmt` / `free_desc` now means the opposite of what its comment says, and the comment-to-code ratio in `handles/mod.rs` runs against the repo's own guidance.

## Nit

One, inline: the ABA test leaves a permanent entry in the process-global registry.

mssql-odbc/src/handles/mod.rs
Vahid-b: Implemented — good catch, and this was cheap given `HasObjectType` was already implemented by all four handle types. Changed `LIVE_HANDLES` to `HashMap<usize, HandleType>`: `handle_to_raw` now takes `T: HasObjectType` and records the type at allocation, `free_stmt`/`free_desc` match on `live_type(handle)` and return `SQL_INVALID_HANDLE` (per spec) for a live-but-wrong-type handle instead of proceeding to dereference it. Added `free_stmt_on_a_live_desc_handle_returns_invalid_handle` and its mirror `free_desc_on_a_live_stmt_handle_returns_invalid_handle` — both pass with the fix. (I didn't attempt a revert-and-confirm-fails cycle for these two specifically: reverting to the old is_live-only check would mean actually letting `handle_from_raw::<StmtHandle>` reinterpret a live `DescHandle`'s memory, which is genuine UB — `StmtHandle`/`DescHandle` aren't `#[repr(C)]` and aren't the same size, so it's a real out-of-bounds risk, not just a logically-wrong-but-safe result. Verified instead by reading the diff and confirming the tests pass with the fix in place.)

Also updated the two "not found in parent's list" branches to reflect that reaching them while live-and-correctly-typed is now an invariant break rather than the expected already-freed path (separate thread on that).

> [!NOTE]

mssql-odbc/src/handles/mod.rs
Vahid-b: Agreed, and thanks for pointing at the existing idiom — implemented exactly that. Centralized every `LIVE_HANDLES.lock()` call behind a `live_handles()` helper that does `.unwrap_or_else(|poisoned| poisoned.into_inner())`, matching `auth/msqa.rs`'s pattern. Your reasoning about why it's sound here (single `HashMap` op per critical section, no user code/I/O/user `Hash` impl) is exactly right, so I used it directly in the doc comment.

Removed the ~10-line paragraph documenting the leak-over-UAF trade-off entirely, since it's now moot — the conservative `false` is gone along with the whole failure mode it was describing.

Added `live_handles_registry_recovers_from_a_poisoned_lock`: poisons `LIVE_HANDLES`'s own lock directly, then confirms allocating/freeing a handle still works normally afterward. Verified it actually catches the regression — reverted `live_handles()` to `.expect(...)` (panic on poison), confirmed the test fails with the predicted panic, then restored the fix and confirmed it passes.

> [!NOTE]

mssql-odbc/src/api/free_handle.rs
Vahid-b: Good point, and implemented essentially your suggestion (for both `free_stmt` and `free_desc`, since the same drift applies to both). Updated the inline comment to say what reaching that branch now actually means, added the `error!` for visibility, and added `debug_assert!(false, ...)` so it's caught rather than silent in debug builds.

One pre-existing test (`free_desc_already_removed_from_parent_is_a_no_op`) directly forced this exact "live but untracked" state to test the old "already dropped" framing, so I updated and renamed it (`free_desc_live_but_untracked_by_parent_fails_in_debug`) to assert the new `debug_assert!`-driven behavior instead, branching on `cfg!(debug_assertions)` like the sibling `free_dbc_with_outstanding_*` tests already do. Verified it actually exercises the assert: temporarily removed the `debug_assert!`, confirmed the test failed (asserting `SQL_ERROR` but getting `SQL_SUCCESS`), then restored it.

> [!NOTE]

mssql-odbc/src/handles/mod.rs
Vahid-b: Agreed — trimmed both the `LIVE_HANDLES` doc comment and the disconnect.rs comments/test doc, dropping the review/PR-history narration you quoted while keeping the three technical points (the invariant, the ABA limitation, and — for now — a shorter version of the poisoned-registry note). That last one actually became moot from the other thread's fix (recovering from poison instead of documenting the failure mode), so it's gone entirely rather than just shortened.

The doc comment is a bit longer than a minimal version since it now documents *two* mechanisms (address liveness + the type-confusion fix from the other thread) instead of one, but it should still read tighter than before.

> [!NOTE]

mssql-odbc/src/handles/mod.rs
Vahid-b: Fixed — added the cleanup you suggested. (This specific test got rewritten for the type-tracking change in the other thread, but the same leak-prevention point applies, so I kept it: it now removes the address from the registry after asserting on it.)

> [!NOTE]

mssql-odbc/src/api/free_handle.rs
```diff
@@ -252,7 +305,14 @@ unsafe fn free_desc(handle: SqlHandle) -> SqlReturn {
...
     let Some(i) = dbc_state.descriptors.iter().position(|&p| p == handle) else {
-        // Already dropped by SQLDisconnect - early return.
+        // Live but untracked: not the post-SQLDisconnect case (caught by
+        // `live_type` above), so something removed it from the parent
+        // without freeing it — an invariant break, not an expected path.
+        error!(
+            ?handle,
+            "SQLFreeHandle(DESC): live handle missing from parent DBC"
+        );
+        debug_assert!(false, "live DESC not tracked by its parent DBC");
```
David-Engel, line 362: **Suggestion.** This `debug_assert!` asserts an invariant that the code above it doesn't actually guarantee, and I confirmed it fires on a legitimate path.

The implicit-descriptor rejection at L286 sits *inside* `if let Ok(mut state) = desc.inner.lock()`. If that descriptor's mutex is poisoned, the whole `if let` is skipped — including the `is_explicit()` check — and an implicit descriptor (never registered in `dbc_state.descriptors`, by design) falls straight through to this branch. It is live, it is correctly typed, and it is legitimately absent from the parent list. That is not "something removed it from the parent without freeing it."

I probed it on this head with a throwaway test that poisons `StmtHandle::ard`'s mutex and then calls `sql_free_handle(SQL_HANDLE_DESC, ard)`:

- unpoisoned: `-1` (`SQL_ERROR`) via the intended HY017 path;
- poisoned: `panicked at free_handle.rs:315: live DESC not tracked by its parent DBC`, then `-1` from `ffi_entry!` swallowing it.

Debug builds get a misleading assertion failure; release builds return **`SQL_SUCCESS`** for an implicit descriptor that was neither freed nor freeable, which is worse than the HY017 the same call produces when the lock happens to be healthy.

`is_explicit()` reads `self.alloc_type`, a lock-free field (`handles/desc.rs:333`) — the lock is only needed to *post* the diagnostic. Hoisting the check out makes the rejection unconditional and restores the honesty of this assert:

```rust
if !desc.is_explicit() {
    error!("SQLFreeHandle(DESC): cannot free an implicitly allocated descriptor");
    if let Ok(mut state) = desc.inner.lock() {
        post_diag(&mut state, ERR_INVALID_USE_OF_AUTO_DESC);
    }
    return SQL_ERROR;
}

if let Ok(mut state) = desc.inner.lock() {
    free_errors(&mut state);
}
```

Note this is only half pre-existing: before this PR the same fall-through returned `SQL_SUCCESS` silently, so the leak isn't new — the debug panic is. Worth fixing here since the assert is what makes the "invariant break" claim in the comment and doc comment true. @shiwanigupta0809's thread checked the guard-ordering interaction with HY017 and found it held; it does, but only on the healthy-lock path.

mssql-odbc/src/handles/mod.rs
```diff
@@ -29,10 +31,91 @@ pub(crate) enum HandleType {
...
+
+/// Returns the `HandleType` recorded for `raw` if it currently refers to a
+/// live handle (allocated via `handle_to_raw`, not yet freed via
+/// `free_handle`), or `None` otherwise.
+///
+/// Safe to call on any pointer value, including one never allocated or
+/// already freed, whose memory may be deallocated or reused: this never
+/// dereferences `raw`, so callers can use it to decide whether
+/// dereferencing is safe — and as what type — in the first place.
+pub(crate) fn live_type(raw: *mut c_void) -> Option<HandleType> {
```
David-Engel, line 98: **Suggestion, non-blocking.** Now that the registry carries `HandleType`, the type-confusion hole is closed for two of the four handle types and left open for the other two. `sql_free_handle` still dispatches on the caller-supplied `handle_type` with no cross-check for `SQL_HANDLE_ENV` / `SQL_HANDLE_DBC`, so `SQLFreeHandle(SQL_HANDLE_DBC, <live STMT>)` reaches `handle_from_raw::<DbcHandle>` and reads `dbc.object_type` / locks `dbc.inner` through a wrongly-typed reference — the exact UB @shiwanigupta0809 described, just via the other two arms. `StmtHandle` and `DbcHandle` are neither `#[repr(C)]` nor the same size, so the `debug_assert_eq!` on `object_type` in `free_env` (L65) and `free_dbc` (L99) can't be trusted to catch it either.

The prior round's "considered and not filed" note for `free_env`/`free_dbc` argued the *liveness* case — that neither is cascade-freed behind the DM's back, so there's no legitimate stale ENV/DBC. That reasoning is still right, and it doesn't cover this case: type confusion needs no stale handle and no address reuse at all, just a mismatched `HandleType` on a currently-live handle.

Same four lines as the STMT/DESC arms:

```rust
match live_type(handle) {
    Some(HandleType::Dbc) => {}
    other => {
        error!(?handle, ?other, "SQLFreeHandle(DBC): handle is not a live DBC");
        return SQL_INVALID_HANDLE;
    }
}
```

Entirely reasonable as a follow-up on #422's ticket or its own — but leaving two arms checked and two unchecked is the kind of asymmetry that reads as an oversight later, so it's worth either doing or saying why not.

David-Engel (commented):

## Summary

Reviewed the full `origin/main...d2bf0030` diff (merge base `1ecefe1c`) across `handles/mod.rs`, `api/free_handle.rs` and `api/disconnect.rs`. The registry mechanism is sound: `handle_to_raw` / `free_handle` really are the only `Box::into_raw` / `Box::from_raw` sites in the crate (including `StmtHandle::drop`'s four implicit descriptors), `live_type` never dereferences, `LIVE_HANDLES` is a leaf lock released before any handle mutex or any `Drop`, and the poison-tolerant disconnect loops drop the guard at the end of the `if` condition — before `free_handle` drops the box. The type-tag addition from the last round is a real improvement over address-only liveness.

Nothing blocking. Two suggestions and one nit, none overlapping the already-settled threads (ABA/#422, orphan-on-poison, `into_inner()` recovery, comment density, the ABA test's registry cleanup) or the prior clean review on `1cf909a6`.

## Verification

- `cargo nextest run -p mssql-odbc --lib --no-fail-fast` on this head, `CARGO_TARGET_DIR` pinned to a scratch dir: **1095 passed, 1 skipped**, including all five new tests.
- **Probed the one finding below rather than asserting it.** Added a throwaway test that poisons a *live implicit* descriptor's `inner` mutex (reachable through `StmtHandle::ard`) and then calls `sql_free_handle(SQL_HANDLE_DESC, ard)`, then reverted it. Output:
  - unpoisoned baseline: `-1` (`SQL_ERROR`, via the HY017 path) — correct;
  - poisoned: `panicked at free_handle.rs:315: live DESC not tracked by its parent DBC`, then `-1` from `ffi_entry!`.
- Grepped every mutation site of `dbc_state.statements` / `dbc_state.descriptors` to confirm the `free_stmt` list-miss branch really is unreachable from any public API — it is. Only `free_desc`'s is not.
- `gh pr checks 415`: all required checks green on this head, including the full ADO validation across five platforms, Kerberos, and both cross-repo `mssql-python` legs.

## Blocking

None.

## Suggestion

Two, inline:
1. `free_desc`'s new `debug_assert!(false, ...)` is reachable from a path that is not an invariant break — a live implicit descriptor whose own mutex is poisoned skips the HY017 rejection and falls through to it (probed above).
2. `free_env` / `free_dbc` still dispatch without the type cross-check that `free_stmt` / `free_desc` now have.

## Nit

The PR description is stale relative to this head. It still describes the fix as "a process-wide `Mutex<HashSet<usize>>`" exposed as `handles::is_live`, and the Validation section cites 1092 tests; the head is a `Mutex<HashMap<usize, HandleType>>` behind `live_type`, and I measured 1095. More importantly, the description never mentions the type-confusion fix that came out of @shiwanigupta0809's thread, even though it changes an externally visible return code: `SQLFreeHandle(SQL_HANDLE_STMT, <live DESC>)` now returns `SQL_INVALID_HANDLE` where it previously reinterpreted the handle. That is the most behavior-visible part of the PR and should be named in the description (and in the `Fixes`/`Relates to` list) rather than only in a review thread.

mssql-odbc/src/api/free_handle.rs
Vahid-b: Confirmed and fixed — good catch, and thanks for the concrete repro. Hoisted `is_explicit()` out of the `desc.inner.lock()` block entirely so the rejection no longer depends on lock health: it only reads `alloc_type`, which is set once at construction and never mutated, matching the reasoning you laid out. The lock is still taken afterward for `free_errors`, tolerating poison the same way as everywhere else in this module.

Added `free_implicit_desc_is_rejected_even_with_a_poisoned_mutex`, which reproduces your exact repro (poison the ARD's own mutex, then try to free it). It calls `free_desc` directly instead of through `sql_free_handle`, since `ffi_entry!`'s `catch_unwind` would otherwise launder a wrongly-firing `debug_assert!` panic into the same `SQL_ERROR` a correct HY017 rejection also returns — routing around that kept the distinction your finding depends on actually visible to the test. Verified by reverting the fix locally: the test failed exactly as predicted (panic at the `debug_assert!` you found, then a second failure from the test's own panic-vs-clean-return check), then passed after restoring it.

> [!NOTE]

mssql-odbc/src/handles/mod.rs
Vahid-b: Agreed — the asymmetry wasn't intentional, just scope creep I hadn't caught up with. Closed it: `free_env`/`free_dbc` now check `live_type` first, same four-line shape as the STMT/DESC arms, returning `SQL_INVALID_HANDLE` for a live handle of the wrong type. You're right that the liveness reasoning ("never cascade-freed behind the DM's back") doesn't cover this — type confusion needs no stale handle at all — so I called that out explicitly in the doc comments rather than reusing the old liveness-only justification.

Added `free_env_on_a_live_dbc_handle_returns_invalid_handle` and its mirror `free_dbc_on_a_live_env_handle_returns_invalid_handle`. Same caveat as the earlier STMT/DESC round: I didn't do a revert-and-confirm-fails cycle for these two, since reverting would mean actually letting `handle_from_raw::<EnvHandle>` reinterpret a live `DbcHandle`'s memory — genuine UB, not just a wrong-but-safe result, given neither type is `#[repr(C)]` or same-sized. Verified via code inspection plus the tests passing with the fix in place instead.

> [!NOTE]

David-Engel (commented):

## Summary

Reviewed the full `1ecefe1c...00ea2910` diff across `handles/mod.rs`, `api/free_handle.rs` and `api/disconnect.rs`, plus the surrounding alloc/free/disconnect paths. The registry mechanism still holds up on this head: `handle_to_raw` / `free_handle` remain the only `Box::into_raw` / `Box::from_raw` sites in the crate (including `StmtHandle::drop`'s four implicit descriptors), no alloc path registers and then bails, `live_type` never dereferences, and `LIVE_HANDLES` is a leaf lock released before any handle mutex and before `Box::from_raw` runs `Drop`. Both fixes new in `00ea2910` are correct in substance: hoisting `is_explicit()` out of the `desc.inner` lock is sound because `DescHandle::alloc_type` is a plain field fixed at construction and `SQL_DESC_ALLOC_TYPE` is permanently read-only (`classify_field`), and the `free_env` / `free_dbc` type cross-check closes the same reinterpretation hole `free_stmt` / `free_desc` already closed.

One regression came in with that hoist, though, and it contradicts what the reply on that thread says the new code does. Nothing blocking; one suggestion and two nits, none overlapping the settled threads (ABA/#422, orphan-on-poison, `into_inner()` recovery, the type-tag suggestion, the ABA test's registry cleanup).

## Verification

- `cargo nextest run -p mssql-odbc --lib --no-fail-fast` on this head, `CARGO_TARGET_DIR` pinned to a scratch dir: **1098 passed, 1 skipped** (the skipped one needs a live server).
- **Probed the suggestion below rather than asserting it.** Added a throwaway test that calls `sql_free_handle(SQL_HANDLE_DESC, <implicit ARD>)` three times and counts the descriptor's `diag_records`, then reverted it. Result: **3 records**, one per rejected call. Before this PR the same sequence left exactly 1, because `free_errors` ran ahead of `post_diag` on that path.
- Re-derived the lock graph by hand: `live_handles()`'s guard is a per-statement temporary at every call site, and `if <lock>.is_err()` in both `sql_disconnect_safe` loops is its own terminating temporary scope, so no guard is live across `free_handle`.
- Grepped every mutation site of `dbc_state.statements` / `dbc_state.descriptors` again: with `is_explicit()` hoisted, `free_desc`'s "live handle missing from parent DBC" branch is now genuinely unreachable from any public API, which is what the previous round asked for.
- `gh pr checks 415`: every required check green on this head, including the full ADO validation across five platforms, Kerberos, and both cross-repo `mssql-python` legs.

## Blocking

None.

## Suggestion

**`free_desc` no longer clears stale diagnostics on the implicit-descriptor (HY017) path.** `mssql-odbc/src/api/free_handle.rs:323-333`. Before this PR the sequence was `free_errors` first, then the `is_explicit()` check and `post_diag`. Hoisting the check above the lock moved the rejection to an early `return SQL_ERROR` that never reaches the `free_errors` block below it, so that path now posts without clearing. The reply on the thread that prompted this change says "the lock is still taken afterward for `free_errors`" — that is true for the explicit path, but the implicit path returns before it.

Two consequences, both in `.github/instructions/mssql-odbc.instructions.md`'s "Every ODBC entry point must clear the handle's diagnostic records at API entry by calling `free_errors(...)`" territory: repeated `SQLFreeHandle(SQL_HANDLE_DESC, <implicit>)` calls accumulate one HY017 record each (probed above, 3 for 3 calls), and — the part that actually changes what an application reads — a diagnostic left on that descriptor by an earlier call such as a failed `SQLSetDescFieldW` survives, so `SQLGetDiagRec(SQL_HANDLE_DESC, desc, 1, ...)` hands back the stale record instead of the HY017 this call produced.

Folding `free_errors` into the same guarded block fixes it without giving the poison tolerance back:

```rust
if !desc.is_explicit() {
    error!("SQLFreeHandle(DESC): cannot free an implicitly allocated descriptor");
    if let Ok(mut state) = desc.inner.lock() {
        free_errors(&mut state);
        post_diag(&mut state, ERR_INVALID_USE_OF_AUTO_DESC);
    }
    return SQL_ERROR;
}
```

Worth noting why the suite stayed green: `free_implicit_desc_is_rejected` asserts on `diag_records.last()`, which passes at any record count. Asserting `diag_records.len() == 1` (or calling the entry point twice) would have caught this and would guard the fix.

## Nit

**The `into_inner()` recovery in `live_handles()` is a documented exception to a rule the instructions state absolutely.** `mssql-odbc/src/handles/mod.rs:80-86`. The recovery itself is right for this map — @shiwanigupta0809's reasoning holds, and `auth/msqa.rs:311` and `auth/entra.rs:272` already do the same — but `.github/instructions/mssql-odbc.instructions.md` says flatly "Do **not** recover via `e.into_inner()`", with no carve-out. That mismatch is what makes the next reviewer or agent re-litigate a settled decision. A one-line amendment to that file — recover only for a process-global lock over a plain collection with no invariant to protect, `SQL_ERROR` on poison everywhere else — would put the exception where it gets read. Separate PR; not this author's problem to fix here, but worth filing.

**The PR description is stale relative to this head.** It still describes the registry as "a process-wide `Mutex<HashSet<usize>>`" exposed as `handles::is_live` and cites 1092 tests; the head is a `Mutex<HashMap<usize, HandleType>>` behind `live_type` and I measured 1098 passing. It also still doesn't name the externally visible part of the change: `SQLFreeHandle` now returns `SQL_INVALID_HANDLE` for a live handle of the wrong type on all four handle types, where it previously reinterpreted the handle. This was raised on the previous head and the description hasn't caught up — worth doing before ready-for-review, since the description is what the human reviewer reads first.

Vahid-b:
Thanks for the thorough re-review — replying to the summary since this round didn't produce inline threads.

> One regression came in with that hoist, though, and it contradicts what the reply on that thread says the new code does.

Confirmed and fixed in `3a80f5da`: hoisting `is_explicit()` above the `desc.inner` lock moved the implicit-descriptor rejection to an early return that skipped `free_errors`, so repeated calls accumulated one diagnostic record instead of clearing before each new one. Folded `free_errors` into the same guarded block as `post_diag`, matching your suggested fix. Also strengthened `free_implicit_desc_is_rejected` to call the rejection twice and assert exactly one record each time (not just that `.last()` is HY017, which is why the regression didn't fail the suite) — verified by reverting the fix locally and confirming the strengthened test now fails with the exact predicted symptom (2 records instead of 1), then restoring it.

> The `into_inner()` recovery in `live_handles()` is a documented exception to a rule the instructions state absolutely.

Filed #430 to reconcile the instructions file with this exception rather than amending it in this PR, per your suggestion.

> The PR description is stale relative to this head.

Rewrote it: the registry description now matches the current `Mutex<HashMap<usize, HandleType>>`/`live_type` shape, test count is 1098, and it explicitly calls out `SQLFreeHandle`'s behavior change (`SQL_INVALID_HANDLE` for a live handle of the wrong type, all four handle types) up front instead of leaving it implicit.

David-Engel (commented):

## Summary

Re-reviewed the full `1ecefe1c...3a80f5da` diff (`handles/mod.rs`, `api/free_handle.rs`, `api/disconnect.rs`) plus the surrounding alloc/free/disconnect paths. The one suggestion from the previous round is fixed, and fixed correctly: `free_errors` now runs inside the same `desc.inner` guard as `post_diag` on the implicit-descriptor HY017 path, so that entry point clears stale diagnostics like every other one. Nothing new found on this head.

## Verification

- **The new assertion actually guards the fix.** Removed the added `free_errors(&mut state);` from `free_desc`'s implicit branch and re-ran `free_implicit_desc_is_rejected`: it fails with `left: 2, right: 1` ("a fresh rejection must clear any diagnostic left by the previous call"). Restored the line afterward; the worktree is back to `3a80f5da`. The switch from asserting `.last()` to asserting `len() == 1` across two calls is what makes this non-vacuous — the old shape passed at any record count.
- `cargo nextest run -p mssql-odbc --lib --no-fail-fast` on this head, `CARGO_TARGET_DIR` pinned to a scratch dir: **1098 passed, 1 skipped** (the skipped one needs a live server).
- Re-checked the registry invariants that the rest of the PR rests on, since `handles/mod.rs` is unchanged since the last round but the conclusions are worth re-confirming against the merged tree: `handle_to_raw` / `free_handle` are still the only `Box::into_raw` / `Box::from_raw` sites in the crate (including `StmtHandle::drop`'s four implicit descriptors, which is what keeps the implicit-DESC HY017 path reachable at all — implicit descriptors are registered, so `live_type` returns `Some(Desc)` and the `is_explicit()` rejection is the branch that fires); no alloc path registers and then bails; `live_type` never dereferences; `LIVE_HANDLES` is a leaf lock whose guard is a per-statement temporary at every call site, released before any handle mutex and before `Box::from_raw` runs `Drop`.
- Re-grepped every mutation site of `dbc_state.statements` / `dbc_state.descriptors`: with `is_explicit()` hoisted, `free_desc`'s "live handle missing from parent DBC" branch stays unreachable from any public API, and `free_stmt`'s equivalent is only reachable via the `swap_remove`-then-`free_handle` window the DM already serializes.
- The PR description now matches the head — `Mutex<HashMap<usize, HandleType>>` behind `live_type`, 1098 tests, and the externally visible `SQL_INVALID_HANDLE`-on-wrong-type change is called out. Both nits from the previous round are addressed (#430 filed for the `into_inner()` instructions carve-out).
- `gh pr checks 415`: every required check green on this head — the full ADO validation across five platforms, Kerberos, and both cross-repo `mssql-python` legs. Only the coverage-merge legs are still pending, which is not a failure.

## Blocking

None.

## Suggestion

None.

## Nit

None.

## One thing considered and deliberately not filed

Under a poisoned descriptor mutex, `free_desc` returns `SQL_ERROR` for an implicit descriptor without posting HY017, so `SQLGetDiagRec` on that handle yields nothing for that call. That is inherent — the record can't be written without the lock — the return code is still correct, and `free_implicit_desc_is_rejected_even_with_a_poisoned_mutex` asserts exactly that contract. Recording it so the next reader doesn't have to re-derive it, not as a finding.

David-Engel (approved):

PR #417, opened 2026-08-28
Add three-way ODBC performance benchmarks

saurabh500:
## Summary

This is the benchmark and infrastructure PR split from #395. It intentionally excludes the production ODBC/TDS fetch optimizations, which remain separate in #418.

- Add a fixed C++17 ODBC Driver Manager harness for the ODBC result-retrieval patterns used by `microsoft/mssql-python`: bound rowsets 1/64/1000, repeated bind/fetch/unbind cadence, nullable inline variable-width data, and row-wise `SQLGetData` for ordinary, MAX/PLP, and `sql_variant` values. Python itself is not loaded or benchmarked.
- Add dedicated Linux and Windows perf-lab pipelines comparing the candidate with the pinned `mssql-odbc` baseline and Microsoft ODBC Driver 18.6.2.1.
- Publish a rich Summary-tab report with three median wall times, candidate throughput, clearly separated regression-gate/reference comparisons, and confirmation details.
- Screen at 5%, then require 3 of 4 balanced confirmation rounds. Microsoft ODBC is a non-gating reference.
- Pin Google Benchmark v1.9.1 by commit and the reporting environment to NumPy 2.2.6 / SciPy 1.15.3.
- Enforce the final `comparison.json` verdict in repository-owned Linux and Windows pipeline steps.

Work item: [AB#47608](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47608)

Split from: #395

## Bootstrap baseline

The initial baseline is production `main` commit `bc4b2dc80b111c388ac8000b90f1c35290d735a0`, the commit on which this benchmark branch was rebased. This PR changes no production `mssql-odbc` or `mssql-tds` files, so candidate and baseline production blobs are identical; only the candidate supplies the new harness/pipeline. Future baseline advances are reviewed edits to `baseline-commit.txt`.

The common encrypted-lab packet-size default is `16192`, which Microsoft ODBC 18.6.2.1 and both Rust builds accept and report exactly. Earlier packet-negotiation failures in Linux 171011 and Windows 171010 are superseded.

## Live perf-lab validation

The meaningful measurement head is `b67adbf691d03e5752e25908106eb5bd73093e79`. Later commits only clarify documentation or make the Windows artifact schema assertion fail closed; they do not change workloads, timings, dependency pins, comparator results, or gate policy.

[...]

PR #395, opened 2026-08-27
Benchmark and optimize ODBC result-set fetches

saurabh500:
## Description

Adds a standalone C++17/Google Benchmark harness for complete ODBC result retrieval through the platform Driver Manager and optimizes the in-repo `mssql-odbc` fetch path measured by it.

The harness covers two deterministic, fixed-length, no-LOB result shapes:

| Scenario | Shape | Cells |
| --- | ---: | ---: |
| Narrow | 2,000,000 rows × 15 mixed columns | 30,000,000 |
| Wide | 10,000 rows × 600 mixed columns | 6,000,000 |

Both repeat `BIT`, integer widths, `REAL`, `FLOAT(53)`, `DECIMAL(18,4)`, `DATE`, `TIME(7)`, `DATETIME2(7)`, `DATETIMEOFFSET(7)`, GUID, `CHAR(8)`, and `NCHAR(8)`. Every process verifies the complete row count and checksum before measured samples. Timing covers execution, metadata/binding, and complete rowset fetching.

Linux and Windows perf-lab runners build candidate and pinned-baseline mssql-odbc drivers, install pinned Microsoft ODBC Driver 18.6.2.1, and run all three on the same host. Narrow runs candidate → Microsoft → baseline; wide reverses the mssql-odbc order. The published summary reports candidate changes against both drivers; the optional regression gate remains candidate versus the pinned mssql-odbc baseline.

The driver changes incorporate the sync-first cursor stack from #303, #304, #306, and #307, then add:

- direct decoding of transport-buffered rows into bound ODBC buffers;
- exact typed writes for fixed and temporal values;
- borrowed short-string delivery without an intermediate allocation;
- buffered decimal and GUID decoding;
- batched transport-buffer consumption, cached result eligibility, and reusable NBCROW bitmap storage;
- unchanged async fallback for packet boundaries, cancellation, encrypted columns, PLP values, and unsupported shapes.

## Local Windows comparison

Two counterbalanced execution orders, seven measured samples per order and workload (14 pooled samples per driver), using the same x64 Release harness, loopback SQL Server, rowset size 1,024, packet size 32,768, mandatory encryption, and checksum preflight:

| Driver | Narrow median | Wide median |
| --- | ---: | ---: |
| Pre-tuning #307-equivalent stack | 5,827.917 ms | 1,221.366 ms |
| This PR | **1,753.960 ms** | **334.061 ms** |
[...]

PR #397, opened 2026-08-27
Ship mssql-odbc driver artifacts with product filenames

gargsaumya:
## Description

This pull request renames the `mssql-odbc` crate and its produced artifacts to use the unified name `mssqlodbc` (for the crate) and `mssqlodbc.so`/`.dll`/`.dylib` (for the shipped library), replacing the previous `msodbcsql18` naming. It updates all build scripts, documentation, and CI pipeline references to match this new naming convention. Additionally, it introduces platform-specific artifact finalization scripts to ensure consistent output filenames across platforms.

The most important changes are:

**Crate and Artifact Renaming:**
* Renamed the crate in `Cargo.toml` from `mssql-odbc` to `mssqlodbc`, and the library name from `msodbcsql18` to `mssqlodbc`. All references to the old names in build scripts, documentation, and CI pipeline files were updated accordingly. (`mssql-odbc/Cargo.toml`, [mssql-odbc/Cargo.tomlL2-R8](diffhunk://#diff-a82ca1a46390a7f5938338019045a4da6b73ec228be07dadacd9c00adcb16337L2-R8))
* Updated the output artifact name to `mssqlodbc.so` (Linux), `mssqlodbc.dylib` (macOS), and `mssqlodbc.dll` (Windows) throughout documentation, scripts, and build configuration. (`mssql-odbc/README.md`, [[1]](diffhunk://#diff-ef0b8fcea783dca6765ef3d45d868fd1bc97bb4e1783cc8a2ad78d610ae2e70cL8-R45) [[2]](diffhunk://#diff-ef0b8fcea783dca6765ef3d45d868fd1bc97bb4e1783cc8a2ad78d610ae2e70cL71-R84) [[3]](diffhunk://#diff-ef0b8fcea783dca6765ef3d45d868fd1bc97bb4e1783cc8a2ad78d610ae2e70cL87-R94); `mssql-odbc/plan.md`, [[4]](diffhunk://#diff-270a5a2c345f716b922bd94bd8128b2dce652b07c6353caae1f67f6d4e6b7c1cL12-R12); `mssql-odbc/docs/connection_string_parser.md`, [[5]](diffhunk://#diff-cf4d34f4a45ca4e25456fb7ab1bece128359ee987a13117a1e4c2927b98e8ba2L226-R226); `mssql-odbc/src/api/exports.rs`, [[6]](diffhunk://#diff-497e60fa5ef90c47cddeb284f8694571c3a2cdd9f3c37a396361ea3311cb6f8dL4-R4)

[...]

PR #418, opened 2026-08-28
Optimize ODBC result-set fetches

saurabh500:
## Description

Standalone production optimization PR reconstructed from #395 on current `main`:

- sync-first cursor dispatch with async continuation for packet boundaries, cancellation, encrypted columns, PLP values, and unsupported shapes;
- direct decoding of buffered rows into bound ODBC buffers, including exact typed and temporal writes and borrowed strings;
- buffered decimal/GUID/`sql_variant` decoding, batched buffer consumption, cached row eligibility, and reusable NBCROW bitmap storage;
- preserved current `SQL_ATTR_MAX_ROWS` behavior, including a focused test that the shared fetch-loop budget bounds whole-row dispatch.

PR #417 supplied the three-way benchmark harness. After merging current `main`, this branch pins `db7d266506c4fbc92b681221d50e686c2926201a` as the production baseline so the comparison excludes later mainline ODBC artifact and conversion work. Benchmark workloads, regression policy, packet size 16192, comparator, pipelines, and Microsoft driver pin are unchanged.

### Public TDS API

`CursorPoll<T>`, `BufferedRowPoll`, and the associated `try_*`/`finish_row_into` methods are public on concrete `TdsClient` so synchronous bindings can attempt buffered work without duplicating the cursor state machine. `CursorPoll::Pending` and `BufferedRowPoll::Pending` are non-destructive; `BufferedRowPoll::Partial` has already written a row prefix and requires the same writer to be passed to `finish_row_into`.

The high-level `TdsClient::next_row_into` path applies buffered whole-row decoding automatically before its existing async fallback. Existing `RowWriter` consumers such as the JavaScript and Python bindings therefore share the optimization without calling the concrete polling API themselves.

### `sql_variant` follow-up

The retained change synchronously decodes complete buffered NULL, fixed-width, character, decimal, and numeric values while preserving async fallback for incomplete, encrypted, unsupported, skipped-column, cancellation, PLP, and packet-boundary cases. An in-place Ready-lock experiment was rejected because it did not produce a repeatable gain and regressed chunked LOB retrieval.

## Final exact-head performance validation

Candidate `6a715de605cbc97697f1ebc1a4d8f3636d84a753` was measured against pinned Rust baseline `db7d266506c4fbc92b681221d50e686c2926201a` and Microsoft ODBC 18.6.2.1:

- [Linux 171459](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=171459) — passed, `perf-results` artifact published.
- [Windows 171458](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=171458) — passed, `perf-results` artifact published.

Both runs exercised all 13 workloads through candidate, baseline, and Microsoft drivers. The harness set and read back packet size **16192**. Each initial value is a 5-sample median. The three largest apparent improvements per platform were re-measured over four alternating-order rounds and required 3-of-4 reproduction. Both regression gates passed with **0 confirmed regressions** and **3 verified improvements**.

`B/M/C` is baseline / Microsoft / candidate median wall time. Percentages are candidate versus baseline and candidate versus Microsoft; negative is faster. **Confirmed** uses the four-round confirmation median for the baseline percentage. Other rows are initial-only measurements outside the three-result confirmation cap.

### Fixed-width bound fetch

| Workload | Linux B/M/C | Linux change | Windows B/M/C | Windows change | Evidence |
|---|---:|---:|---:|---:|---|
| Narrow 2M × 15, rowset 1000 | 24220.72 / 7375.69 / 1697.96 ms | -92.99% / -76.98% | 27657.86 / 7415.89 / 1657.58 ms | -94.01% / -77.65% | Initial-only |
| Wide 10K × 600, rowset 1000 | 4903.89 / 736.57 / 271.14 ms | **-94.43%** / -63.19% | 5657.43 / 855.37 / 259.41 ms | **-95.30%** / -69.67% | **Confirmed 4/4 both** |
| Fixed 100K × 15, rowset 1 | 1282.08 / 190.98 / 142.70 ms | -88.87% / -25.28% | 1618.00 / 213.00 / 172.43 ms | -89.34% / -19.05% | Initial-only |
| Fixed 100K × 15, rowset 64 | 1214.85 / 786.42 / 84.27 ms | -93.06% / -89.28% | 1543.32 / 786.74 / 80.35 ms | **-94.78%** / -89.79% | Windows confirmed 4/4; Linux initial-only |
| Fixed 100K × 15, rowset 1000 | 1232.39 / 342.86 / 83.94 ms | **-93.04%** / -75.52% | 1525.66 / 348.11 / 81.38 ms | -94.67% / -76.62% | Linux confirmed 4/4; Windows initial-only |

### Bind-cycle fetch

| Workload | Linux B/M/C | Linux change | Windows B/M/C | Windows change | Evidence |
|---|---:|---:|---:|---:|---|
| Fixed 100K × 15, bind each rowset 64 | 1240.26 / 180.38 / 83.92 ms | **-93.09%** / -53.47% | 1532.39 / 208.20 / 80.20 ms | **-94.80%** / -61.48% | **Confirmed 4/4 both** |
| Fixed 20K × 15, bind each rowset 1 | 337.66 / 143.26 / 105.62 ms | -68.72% / -26.27% | 427.82 / 167.01 / 134.85 ms | -68.48% / -19.26% | Initial-only |

### Variable-width bound fetch

| Workload | Linux B/M/C | Linux change | Windows B/M/C | Windows change | Evidence |
|---|---:|---:|---:|---:|---|
| Nullable 100K × 7, rowset 1000 | 937.70 / 644.24 / 357.85 ms | -61.84% / -44.45% | 1166.10 / 549.45 / 225.80 ms | -80.64% / -58.90% | Initial-only |
| Nullable 100K × 7, rowset 64 | 927.97 / 1312.99 / 358.69 ms | -61.35% / -72.68% | 1164.73 / 1181.59 / 222.92 ms | -80.86% / -81.13% | Initial-only |

### Row-wise `SQLGetData`

| Workload | Linux B/M/C | Linux change | Windows B/M/C | Windows change | Evidence |
|---|---:|---:|---:|---:|---|
| Chunked MAX LOB | 59.42 / 54.10 / 59.57 ms | +0.25% / +10.11% | 40.48 / 38.30 / 39.72 ms | -1.87% / +3.71% | Initial-only; below regression threshold |
| Inline fixed values | 398.13 / 129.06 / 176.11 ms | -55.77% / +36.45% | 469.79 / 126.10 / 198.89 ms | -57.66% / +57.72% | Initial-only |
| Mixed LOB whole result | 448.71 / 133.60 / 229.83 ms | -48.78% / +72.02% | 528.71 / 138.24 / 262.43 ms | -50.36% / +89.84% | Initial-only |
| `sql_variant` probe/attribute | 138.15 / 52.90 / **72.16 ms** | **-47.77% / +36.39%** | 172.84 / 64.54 / **81.40 ms** | **-52.90% / +26.12%** | Initial-only |

The final `sql_variant` result is about **2× faster than the current-main Rust baseline** on both platforms. The remaining gap to Microsoft is **36.39% on Linux** and **26.12% on Windows**. These current-main exact-head values supersede all earlier disposable and pre-merge measurements.

## Related Issues

[AB#47608](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47608)

## Validation

- `cargo bfmt`
- all-feature clippy for `mssqlodbc` and `mssql-tds` with warnings denied
- `cargo nextest run -p mssqlodbc --lib --no-fail-fast`: 1,156 passed, 1 skipped
- focused `mssql-tds` buffered/cursor/variant/PLP/withdrawal library tests: 177 passed
- mutation checks for whole-row dispatch, variant dispatch, partial binding, and buffered decimal/numeric dispatch
- `git diff --check`
- exact-head Linux and Windows dedicated gates passed with published summaries and artifacts

## Checklist

- [x] `cargo bfmt` passes
- [x] targeted all-feature clippy passes
- [ ] full `cargo btest` passes


mssql-tds/src/connection/tds_client.rs
```diff
@@ -198,6 +198,31 @@ pub enum CursorColumn {
     RowEnded,
 }
 
+/// Result of a non-blocking pull-cursor attempt.
+#[derive(Debug, PartialEq)]
+pub enum CursorPoll<T> {
```
saurabh500, line 238: `CursorPoll<T>` and `BufferedRowPoll`, together with the new `try_*`/`finish_row_into` methods below, are public `mssql-tds` API rather than ODBC implementation details. Please call this shared surface out explicitly in the PR description or crate docs, including the non-destructive `Pending` contract, the `Partial` writer-reuse requirement, and the fact that it is currently exposed only on concrete `TdsClient`. That will help future synchronous bindings discover and use it without copying the ODBC state machine.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -5030,14 +5068,116 @@ impl TdsClient {
...
+        Ok(CursorPoll::Ready(true))
+    }
+
+    /// Attempts to decode the next row directly into `writer` from bytes already
+    /// buffered by the transport.
+    ///
+    /// [`BufferedRowPoll::Pending`] is non-destructive. A
+    /// [`BufferedRowPoll::Partial`] result may already have written a row prefix;
+    /// the same writer must be passed to [`Self::finish_row_into`].
+    pub fn try_next_buffered_row_into<W>(&mut self, writer: &mut W) -> TdsResult<BufferedRowPoll>
```
saurabh500, line 5406: Can we provide a shared high-level entry point that applies this fast path before falling back to the existing async row decode—ideally in `next_row_into` or a buffered-result extension trait? `mssql-js::BinaryRowWriter` and `mssql-py-core::PyRowWriter` already implement `RowWriter`, but both call `next_row_into`; only ODBC drives `BufferedRowPoll` today. Keeping the sync-first/partial-continuation state machine inside `mssql-tds` would let other bindings evaluate or use the batched buffer consumption and borrowed-string path without reimplementing it.

mssql-odbc/src/api/util.rs
```diff
@@ -8,6 +8,17 @@ use crate::api::odbc_types::{
 };
 use crate::api::sqlstate::{WARN_STRING_TRUNCATION, post_diag};
 use crate::error::HasDiagnostics;
+use mssql_tds::connection::tds_client::CursorPoll;
+
+pub(crate) fn resolve_cursor_poll<T, E>(
```
saurabh500: This helper is generic over both value and error and contains no ODBC behavior. Could it live next to `CursorPoll` in `mssql-tds` (possibly as a method) so every sync-first binding gets the same `Ready`/`Pending`/error contract and tests? Leaving it here means another driver must duplicate this bridge before it can use the public polling API.

mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -342,6 +344,70 @@ pub(crate) struct DateTimeParts {
     pub has_tz: bool,
 }
 
+pub(crate) fn date_parts(date: &SqlDate) -> DateTimeParts {
```
saurabh500, line 347: `date_parts`, `time_parts`, `datetime2_parts`, and `datetimeoffset_parts` normalize TDS representations rather than ODBC representations. `mssql-py-core/src/cursor.rs` already duplicates the day-ordinal, 100-ns tick, and offset/day-carry calculations. Could the neutral conversion/accessor layer move onto the shared TDS temporal types, with only the mapping into ODBC `Sql*Struct`s left here? That would keep all bindings consistent on scale, range, and offset behavior.

saurabh500 (commented):
Cross-driver API review: these suggestions identify shared APIs and neutral conversion logic that should be discoverable or reusable outside the ODBC binding.

mssql-odbc/src/conversion/fetch_convert.rs
saurabh500: Also add comments

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -596,6 +681,600 @@ impl PlpColumnStream {
...
+                    ColumnValues::Bytes(bytes)
+                }
+            }
+            _ => return Ok(None),
+        };
+
+        Ok(Some((value, reader.position)))
+    }
+
+    pub(crate) fn try_decode_buffered_into<W: RowWriter + ?Sized>(
```
saurabh500, line 1181: ADd comments to all new types and functions

mssql-odbc/src/conversion/fetch_convert.rs
saurabh500: Added concise API comments to each new temporal conversion helper, including the range-failure contract for `datetimeoffset`.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Updated `TdsClient::next_row_into` to attempt whole-row buffered decoding first, continue partial rows with the same writer, and retain the existing async fallback. This automatically covers the JavaScript and Python `RowWriter` consumers.

> [!NOTE]

mssql-odbc/src/api/util.rs
saurabh500: Moved the bridge to `CursorPoll::resolve` in `mssql-tds`, moved its Ready/Pending tests with it, and updated the ODBC callers to use the shared method.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
saurabh500: Documented the concrete `TdsClient` polling surface and its non-destructive `Pending`/writer-reuse contracts in both the Rust API docs and the PR description.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
saurabh500: Documented the buffered reader type and its new methods, plus the direct-to-writer decoder’s incomplete/async-fallback contract.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -380,53 +669,102 @@ fn fill_rowset(
     // Read once per fetch, not once per bind, so an application can move the
     // whole rowset between calls by updating the pointed-to value.
     let bind_offset = unsafe { read_bind_offset(row_bind_offset_ptr) };
+    let can_write_complete_rows = bindings
```
saurabh500, line 716: For block rowsets (`row_array_size != 1`), the code below deliberately leaves no current row for `SQLGetData`, so requiring the final bound ordinal to equal `column_count` is stricter than the observable semantics require. `BoundRowWriter` already ignores unbound columns. Please benchmark allowing whole-row dispatch for block fetches with partial or no bindings, while retaining this restriction for `row_array_size == 1`. The tradeoff is fewer cursor transitions versus decoding trailing values instead of skipping them, so timing should decide this.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -380,53 +669,102 @@ fn fill_rowset(
...
+        .last()
+        .is_some_and(|binding| binding.column_number as usize == column_count)
+        && client.current_result_supports_row_into();
```
saurabh500, line 719: `current_result_supports_row_into()` excludes every encrypted result, so a fully bound encrypted row falls back to `next_row_cursor` plus one `read_row_column`/`block_on` per binding. Since `next_row_into(&mut BoundRowWriter)` already supports encrypted rows through its async fallback, please benchmark using one row-level async operation for fully bound, non-PLP encrypted rows. Keep PLP separate so the existing bound-LOB behavior and `SQLGetData` streaming contract do not change.

mssql-odbc/src/api/get_data.rs
```diff
@@ -520,7 +520,9 @@ fn resume_row_to_column(
     };
 
     let target = column_number - 1; // 0-based
-    let cursor_result = dbc.runtime.block_on(client.read_row_column(target));
+    let cursor_poll = client.try_read_row_column(target);
```
saurabh500, line 563: `try_read_row_column` only completes synchronously for the next sequential buffered, non-PLP, unencrypted column, and it still returns a `ColumnValues`. That leaves two `SQLGetData` candidates: jumping over unbound/intervening columns always enters async `DecodeOne`, and one-shot exact scalar or compatible-string reads still materialize, stash, and then convert the value. Please add benchmark cases and consider a shared TDS “read through target into `RowWriter`” operation that discards intermediates and writes the target directly, falling back to the current captured-value path whenever truncation or retry semantics require retaining it.

saurabh500 (commented):
## Additional optimization candidates

The implemented ODBC read entry points are wired into this PR; the inline comments identify shapes that still miss the full fast path. These are suggestions, not merge blockers, and should be driven by timings.

- PLP `SQLGetData` still allocates a temporary `Vec` per call (`mssql-odbc/src/api/get_data.rs:732`) and performs one `block_on(read_active_plp_chunk)` per application chunk (`:763`). Please benchmark reusable scratch storage or direct reads into the application buffer for target encodings that need no intermediate transcode.
- Normal close/advance drains consume full remaining rows through `next_row_into`/`DecodeAll` (`mssql-tds/src/connection/tds_client.rs:3551`), while the error-response drain already has a `ColumnPolicy::SkipAll` row path (`:3711`). Please benchmark adapting that skip strategy for ordinary discarded rows, especially PLP and encrypted rows, while preserving normal DONE/INFO/RETURNVALUE side effects.
- Add benchmark shapes for partial bindings, skipped-column `SQLGetData`, PLP, and encrypted rows. The current C++ benchmark exercises bound `SQLFetchScroll`, so it cannot answer whether these follow-ups help.

If these are outside this PR's scope, please track them as explicit follow-ups rather than treating all ODBC reads as covered by the existing benchmark.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +965,90 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
     unsafe { ptr.read_unaligned() }
 }
 
+unsafe fn deliver_encoded_string(
```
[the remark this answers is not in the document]
saurabh500: Added the safety contract covering displaced target/indicator validity and the no-overlap requirement between the writable target slot and borrowed transport bytes.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -596,6 +696,604 @@ impl PlpColumnStream {
...
 
 impl GenericDecoder {
+    /// Decodes one complete non-PLP column from `bytes` without awaiting.
+    ///
+    /// `None` means the buffered bytes are incomplete or this type still
+    /// requires the async decoder. The caller must consume `used` bytes only
+    /// after receiving `Some`, so a miss is non-destructive.
+    ///
+    /// Each supported arm must remain wire-compatible with [`Self::decode_into`].
+    pub(crate) fn try_decode_buffered(
```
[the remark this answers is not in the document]
saurabh500: Extended the existing async decode-equivalence helper to compare both buffered entry points whenever they support a type, including consumed byte counts and one-byte-short rejection. The full 103-test decoder module now exercises fixed, nullable, temporal, decimal, GUID, string, and binary cases.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -900,6 +1353,14 @@ mod tests {
         assert_eq!(row_budget(5, 9, 4), 0);
     }
 
+    #[test]
+    fn max_rows_bounds_the_whole_row_dispatch_loop() {
```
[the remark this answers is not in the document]
saurabh500: Replaced the range-count test with the same extracted `dispatch_rows` loop used by `fill_rowset`; it starts with three buffered rows under `max_rows = 5` / `rows_returned = 4` and verifies exactly one is delivered and two remain unconsumed.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -151,6 +167,279 @@ impl RowOutcome {
...
+                None => RowOutcome::Error(RowIssue::Restricted),
+            }
+        } else {
+            unsafe { deliver_bound(&binding, self.row_index, self.bind_offset, &value()) }
+        };
+        self.outcome = self.outcome.merge(delivered);
+    }
+}
+
+impl RowWriter for BoundRowWriter<'_> {
```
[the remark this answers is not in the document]
saurabh500: Added parity coverage for every exact numeric target, date/time/datetime2/datetimeoffset, GUID, decimal/numeric fallback, widening fallback, and range errors. Each case compares typed unaligned row-slot values, indicators, and outcomes against `deliver_bound` with nonzero row and bind offsets; existing tests retain string truncation coverage.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -720,6 +1146,33 @@ unsafe fn deliver_bound(
     RowOutcome::Success
 }
 
+unsafe fn deliver_fixed_bound<T: Copy>(
```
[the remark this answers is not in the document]
saurabh500: Added the safety contract requiring the displaced target slot to be null or writable for `T` and the displaced indicator to be null or writable for one `SqlLen`.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
saurabh500: This PR deliberately preserves `SQLGetData`’s captured-value path because chunking, truncation retries, and jumping over columns require retained state; a write-through target API would be a separate contract and benchmark effort. The optimized scope here is bound rowset fetching, so I’m not adding unmeasured benchmark cases or changing `SQLGetData` semantics in this slice.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
saurabh500: I’m retaining the existing encrypted-row path here because this reconstructed production slice has no encrypted-result benchmark proving a row-level decrypt is a win, while the current fallback is correct and covered. PLP/encryption expansion should be measured as a separate optimization rather than folded into this already validated fixed-type path.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
saurabh500: I’m keeping the conservative fully-bound eligibility in this production-only PR: decoding every unbound trailing column can cost more than the cursor transitions it removes, and #417’s agreed benchmark matrix does not measure partial/no-binding shapes. This should be changed only with a dedicated comparison rather than assuming the broader dispatch is faster.

> [!NOTE]

mssql-tds/src/connection/transport/buffers.rs
```diff
@@ -218,6 +218,10 @@ impl TdsReadBuffer {
     pub(crate) fn get_slice(&self) -> &[u8] {
         &self.working_buffer[self.buffer_position..]
     }
+
+    pub(crate) fn get_buffered_slice(&self) -> &[u8] {
```
saurabh500, line 229: how is this better than get_slice ? Don't the 2 methods do the same thing ?

saurabh500: They differ at the valid-data boundary: `get_slice()` extends through the remaining allocation and is safe only after a length check, while `get_buffered_slice()` ends at `buffer_length` so stale capacity cannot satisfy a sync decode. I documented that invariant on both methods.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -900,6 +1372,109 @@ mod tests {
...
+            SQL_SUCCESS
+        );
+        assert_eq!(
+            values[0], 20,
+            "the capped fetch consumed only the first row"
+        );
+    }
+
+    #[test]
+    fn partial_binding_uses_the_column_cursor_path() {
```
saurabh500, line 1485: This test does not currently prove the column-cursor path named here. I forced `can_write_complete_rows = client.current_result_supports_row_into()` so this partial binding incorrectly consumed the whole row, and the test still passed. Please call `SQLGetData` for the unbound second column after the fetch and assert value `20` with indicator `4`. With that assertion, the forced whole-row mutation returns `SQL_ERROR`, while the current implementation succeeds, so the test specifically pins preservation of the unread column.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -380,53 +669,102 @@ fn fill_rowset(
...
-                Ok(CursorColumn::PlpStreaming { .. }) => RowOutcome::Error(RowIssue::Unsupported),
-                // Reading ascending and once per column, neither of these is
-                // reachable; treat them as a row error rather than assuming.
-                Ok(CursorColumn::RowEnded) | Ok(CursorColumn::AlreadyConsumed) => {
-                    RowOutcome::Error(RowIssue::Restricted)
+        if can_write_complete_rows {
+            let mut writer = BoundRowWriter::new(bindings, rows_filled as usize, bind_offset);
+            let result = match client.try_next_buffered_row_into(&mut writer) {
+                Ok(BufferedRowPoll::Complete) => Ok(()),
+                Ok(BufferedRowPoll::Partial) => {
```
saurabh500, line 728: The ODBC tests use `TokenReplayTransport`, whose buffered-row hook takes and returns the entire scripted row, so it cannot produce `BufferedRowPoll::Partial`. This branch can write a prefix and then resume with the same stateful `BoundRowWriter`, but no driver-layer test checks that binding selection, row pointers, and indicators remain correct across that split. Please expose a packet-splitting client through `test-util` (or add a scripted partial-row mode), then add ODBC cases split after the row header and midway through the second bound column, asserting both values and indicators. The TDS-level continuation tests use `DefaultRowWriter` and do not cover this writer state.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -568,6 +904,16 @@ fn row_budget(max_rows: SqlULen, rows_returned: SqlULen, row_array_size: SqlULen
     row_array_size.min(max_rows.saturating_sub(rows_returned))
 }
 
+/// Calls `dispatch` for consecutive row slots until the budget is spent or the
+/// callback reports that no later row should be attempted.
+fn dispatch_rows(row_budget: SqlULen, mut dispatch: impl FnMut(SqlULen) -> bool) {
```
saurabh500: Nit: the sole caller ignores this row argument and uses the captured `rows_filled` counter as the row index instead, leaving two counters for the same slot. Please either make the callback `FnMut() -> bool` or use the supplied row as the authoritative index. Removing the unused argument is the smaller behavior-preserving cleanup.

saurabh500 (commented):
Re-review of the current head after the earlier feedback: no blocking findings. I found two test gaps in the new dispatch paths and one minor cleanup.

mssql-odbc/src/api/fetch_scroll.rs
saurabh500: Added the unbound-column `SQLGetData` assertion for value 20 and indicator 4. Forcing whole-row dispatch on the partial binding now makes this test fail with `SQL_ERROR`, while the column-cursor path passes.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
saurabh500: Removed the unused row argument; `dispatch_rows` now accepts `FnMut() -> bool`, leaving `rows_filled` as the single authoritative row slot.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
saurabh500: Added a scripted partial-row mode that preserves the same writer across buffered prefix and async continuation. ODBC tests now split immediately after the row header and after the first of two bound columns, asserting both values and indicators.

> [!NOTE]

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -596,6 +699,705 @@ impl PlpColumnStream {
...
+                if length == u16::MAX {
+                    ColumnValues::Null
+                } else {
+                    let Some(bytes) = reader.take_bytes(usize::from(length)) else {
+                        return Ok(None);
+                    };
+                    ColumnValues::Bytes(bytes)
+                }
+            }
+            _ => return Ok(None),
```
David-Engel, line 1171: `try_decode_buffered_into` decodes `DecimalN`/`NumericN` buffered (the arm just above, with the `decimal_metadata_is_valid` + `MAX_DECIMAL_INT_PARTS` checks), but `try_decode_buffered` falls through to this `_` and returns `None` for those two types. The bound whole-row path therefore gets synchronous decimals while the column-cursor path — `try_read_buffered_column` → `try_read_row_column` → `SQLGetData`, and `next_row_cursor` + `read_row_column` for any partially bound rowset — always pays a `block_on` for a `decimal`/`numeric` column that is already fully buffered.

Since `read_decimal_data` is already factored out and `try_decode_buffered_into`'s arm is the same arithmetic, adding the matching `DecimalN | NumericN` arm here (producing `ColumnValues::Decimal`/`Numeric`) looks cheap, and `assert_decode_equivalence` would cover it for free. If it was left out deliberately — e.g. no benchmark for decimals on the `SQLGetData` path — a one-line note would keep the two decoders from drifting further apart, since they have to stay wire-compatible with each other as well as with `decode_into`.

David-Engel (commented):

## Summary

Sync-first row/column decoding layered over the existing async path, in three parts: a `BufferedSlice`-based non-consuming-on-miss decoder in `mssql-tds` (`try_decode_buffered`, `try_decode_buffered_into`, `try_decode_buffered_variant`), a `CursorPoll`/`BufferedRowPoll` polling surface on `TdsClient`, and a `BoundRowWriter` in `mssql-odbc` that writes exact-typed values straight into bound application buffers.

The design is conservative in the right places: every miss returns `Pending`/`None` before consuming transport bytes, PLP / encrypted / skipped-column / cancelled / packet-boundary shapes all fall back, and `get_buffered_slice()` (bounded at `buffer_length`, unlike `get_slice()`) is what makes "is this value complete?" answerable without a refill. `remove_header_from_packet` strips packet headers before appending, so the bounded slice is contiguous payload and a cross-packet read cannot alias a header — worth stating explicitly since the whole scheme rests on it.

I checked the arms that could silently desynchronize the stream and found the buffered decoders wire-compatible with `decode_into` / `read_column_value`, including the ones where "obviously wrong" would have been easy: `FltN`'s `_ => f64` and `BitN`'s `_ => 1 byte` match the async catch-alls rather than the spec; `Char`/`VarChar` read `USHORTLEN` because `decode_string_into` does too; `DecimalN` repeats the same `div_ceil(4)` / `MAX_DECIMAL_INT_PARTS` bound-before-copy order as `read_decimal_data`; `scale_time_value` is a faithful extraction of the inline `match`. Extending `assert_decode_equivalence` to run both buffered entry points and assert consumed-byte parity plus one-byte-short rejection is the right guard for this class, and `bound_row_writer_matches_established_conversion_path` comparing direct writes against `deliver_bound` byte-for-byte (including the indicator array, at an unaligned slot with a bind offset) is what makes the `deliver_fixed_bound` shortcut auditable.

On the ODBC side, `BoundRowWriter::take_binding`'s monotonic walk is safe because `StmtHandle::set_binding` binary-inserts and keeps `bindings` ascending; `can_write_complete_rows` requiring `bindings.last().column_number == column_count` means `last_column_read` still equals `column_count` on the whole-row path, so `peek_is_safe` and `current_row_last_col` keep their meaning for a following `SQLGetData`. `current_result_supports_row_into()` excludes PLP and encrypted results, so the `PlpStreaming`/`Unsupported` row outcome and the `SQLGetData` chunking path are unreachable from the new branch.

## Verification

- Read the full diff against `origin/main` (12 files, +4310/-159), plus the unchanged async decoders, `TdsReadBuffer`, `deliver_bound`/`element_stride`/`copy_with_nul`, `resume_to_column`, `set_binding`, and the `fill_rowset` tail that consumes `last_column_read`.
- Read every existing review thread; skipped findings already raised and answered (public API naming, `CursorPoll::resolve` placement, temporal helper reuse, safety contracts, `get_buffered_slice` vs `get_slice`, encrypted-row and fully-bound eligibility deferrals, the dispatch-loop and decode-equivalence coverage threads).
- `cargo nextest run -p mssql-odbc -p mssql-tds --lib --no-fail-fast` (dedicated target dir): 3136 run, 3129 passed, 1 skipped. The 7 failures are all `certificate_validator` / `win_tls::validate`, which fail on a clean tree because their `tests/test_certificates/*.pem` fixtures are not generated locally — unrelated to this diff.

## Findings

### Suggestion

1. `try_decode_buffered` has no `DecimalN`/`NumericN` arm while `try_decode_buffered_into` does — see the inline comment. Not a correctness problem (the miss falls back), but it leaves the two decoders asymmetric and the description's "buffered decimal ... decoding" is only true of the bound-fetch path.

### Nit

2. `mssql-odbc/src/api/util.rs` now carries a single whitespace-only hunk (the blank line between the `use` block and the `fNullable` doc comment) left over from moving the bridge helper to `CursorPoll::resolve`. Restoring it would keep the file out of the diff entirely.

## CI

`coverage-report` (run 33424602224) is red on this head, and the ADO rollup for build 171275 is red while the newer 171326 is green apart from the still-pending macOS legs. Neither is a claim this PR's checklist makes — `cargo bclippy` and `cargo btest` are correctly left unchecked — but the red rollup is worth resolving before this goes to a human. Nothing in the review above depends on those runs.

mssql-tds/src/datatypes/decoder.rs
saurabh500: This is a good comment

saurabh500: Fixed in b79f4d31: `try_decode_buffered` now handles complete `DecimalN` and `NumericN` values with the same metadata, width, sign, and magnitude checks as the direct-writer path. The focused tests now require synchronous results for both types and fail when buffered decimal dispatch is disabled.

> [!NOTE]

David-Engel (commented):

## Summary

Re-review of head `b79f4d31`. The only change since the previously reviewed `98d02b22` is the new `DecimalN`/`NumericN` arm in `GenericDecoder::try_decode_buffered`, which closes the decoder asymmetry raised in the prior pass. I re-read the arm against both of its siblings and traced its reachability rather than assuming it.

The arm is a faithful mirror of the `try_decode_buffered_into` arm and of async `read_decimal_data`: same field order (length byte, then precision/scale validation, then the NULL short-circuit, then sign and magnitude), same `MAX_DECIMAL_INT_PARTS` bound, same little-endian zero-padded reassembly into `[0u8; DECIMAL_MAGNITUDE_BYTES]`, and the same `DecimalN` vs `NumericN` split onto `ColumnValues::Decimal`/`Numeric` that the async dispatch does at lines 2557-2570. `length == 0` is handled before `length - 1`, so the subtraction cannot underflow, and `number_of_int_parts > MAX_DECIMAL_INT_PARTS` bounds `magnitude_len` at 16 before the `copy_from_slice`, so the fixed-size destination cannot overflow. `take_slice(0)` returns `Some(&[])` for the sign-only `length == 1` case, matching the async reader's empty `read_bytes`.

I confirmed the arm is genuinely live rather than defensive: it is reached from `NetworkTransport::try_read_buffered_column` (network_transport.rs:1801), through `AnyTransport::try_read_buffered_column_with_base`, and into the unbound cursor fast path at tds_client.rs:5590, which gates only on column ordering and encryption, not on data type. So decimal and numeric columns now take the sync path on unbound reads. I also checked the one route that could have changed behavior: `try_decode_buffered_variant` only delegates to `try_decode_buffered` when `property_bytes == 0` and the base converts to `FixedLengthTypes`. `DecimalN`/`NumericN` carry two property bytes and are absent from `FixedLengthTypes`, so a variant-wrapped decimal still falls to `Ok(None)` and the async decoder instead of hitting the new arm with a length byte that is not on the wire in that layout.

Error-path parity holds too. Previously an invalid precision/scale on a decimal missed the buffered path and surfaced its `ProtocolError` from the async decoder; now it surfaces the identical error synchronously. Misses stay non-destructive because `BufferedSlice` is a local cursor over a borrowed slice and the transport only advances via `consume_bytes(used)` after a `Some`, and because the caller passes the bounded `get_buffered_slice()` view so bytes past `buffer_length` cannot make a truncated value look complete.

Coverage is better than the two added asserts suggest: `assert_decode_equivalence` already compares the buffered result and consumed width against the async decoder and asserts a one-byte-short input misses, so every existing decimal test routed through that helper — including the length sweep — now exercises the new arm automatically.

## Verification

- Diffed `b79f4d31` against the previously reviewed `98d02b22` (decoder.rs only, +62) and re-read the full PR diff against the `origin/main` merge base `dada8903` (12 files, +4372/-159).
- Read the three decimal implementations side by side, plus `BufferedSlice::take_slice`, `TdsReadBuffer::get_buffered_slice`/`get_slice`, `try_decode_buffered_variant`, `FixedLengthTypes`, the `AnyTransport` and `TdsTokenStreamReader` additions, `try_read_buffered_column`, and the `read_row_column` fast path in tds_client.rs.
- Re-checked the `unsafe impl Send for BoundRowWriter<'_>` justification; the documented argument (the ODBC call owns the bound buffers for the duration of the synchronous fetch, so moving the future between workers cannot create concurrent access) holds for the blocking drive used here.
- Confirmed there are no unresolved review threads, and skipped the carried-over `mssql-odbc/src/api/util.rs` whitespace nit since its thread is already resolved by the author.
- `cargo nextest run -p mssql-tds --lib decoder::` (dedicated target dir): 229 run, 229 passed.
- `cargo nextest run -p mssql-odbc -p mssql-tds --lib --no-fail-fast`: 3136 run, 3129 passed, 1 skipped. The 7 failures are all `certificate_validator` / `win_tls::validate` cases that fail on a clean tree because their `tests/test_certificates/*.pem` fixtures are not generated locally — unrelated to this diff and unchanged from the prior pass.

## Findings

No blocking issues, suggestions, or nits on this head. The prior suggestion is resolved and I found nothing new in the incremental change.

Required checks are green or still running on this head — `coverage-report` and the macOS legs of ADO build 171348 were pending at review time, so confirm those land green before merging.

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot on behalf of @David-Engel, from an unattended run whose findings were not checked by a human first. This is **not an approval** and does not satisfy the human review requirement. Push back on anything that looks off.

## Summary

Re-review of head `040d0ab5`. No findings.

The branch was rebased since the last automated pass, so I re-derived what actually changed rather than trusting per-commit stats. Against the merge base `acb3ed1d` this PR touches 13 files, and the entire `mssql-tds` / `mssql-odbc` production diff is byte-identical to the previously reviewed `b79f4d31`. The only files that differ between the two heads are `mssql-odbc/src/api/disconnect.rs`, `mssql-odbc/src/api/free_handle.rs` and `mssql-odbc/src/handles/mod.rs`, and none of those appear in `git diff $(git merge-base origin/main HEAD) HEAD` — they arrived from `main` in the rebase and are not this PR's changes. The optimization itself (`try_decode_buffered*`, `CursorPoll`/`BufferedRowPoll`, `BoundRowWriter`, and the `DecimalN`/`NumericN` arm added in `9372b95d`) is unchanged from the head I reviewed clean, so I did not re-litigate it.

The one genuinely new change is `mssql-odbc-bench/perf-lab/baseline-commit.txt`, advancing the pin from `bc4b2dc8` to `acb3ed1d`. I checked the properties that make a baseline advance safe rather than taking the header comment at face value:

- `acb3ed1d` is `Add three-way ODBC performance benchmarks (#417)`, so the comment's claim about #417's merge commit holds.
- It equals `git merge-base origin/main HEAD` exactly, so the baseline tree is this PR's true production parent and cannot contain any part of the optimization it is measuring.
- `bc4b2dc8` is an ancestor of it, so this is a forward advance rather than a lateral or backward move.
- It satisfies both runners' validators (`^[0-9a-fA-F]{40}$` at `run-benchmarks.ps1:565` and `run-benchmarks.sh:266`), and both then require `git rev-parse --verify` to resolve it before `git worktree add --detach`. It is on `origin/main` and closer to head than the previous pin, so it is strictly easier to resolve in a pipeline checkout, not harder.

Two things worth watching. Neither is a defect in the diff, so I have not filed either as a finding.

**Pin freshness is not enforced anywhere.** The pin is the "production parent" only for as long as no driver-source commit lands on `main` before this merges. Today there is zero drift: `acb3ed1d..origin/main` is a single commit (`e6223c2b`, an `mssql-py-core` version bump) and `git diff acb3ed1d origin/main -- mssql-odbc/src mssql-tds/src` is empty, so the pin is correct exactly as written. If driver changes do land first, re-advance the pin before the perf numbers are treated as final — otherwise the comparison quietly attributes `main`'s changes to this PR.

**The exact-head perf runs are the real verdict and are still in flight.** The description lists Linux 171366 and Windows 171365 as queued/running with results to be added, and the regression gate in `.pipeline/odbc-perf-baseline-linux-pipeline.yml` is what decides whether the optimization holds. The PR is already out of draft carrying `ready for human review` with `full cargo btest passes` unchecked, so a human reviewer should expect those numbers before merge rather than before review. Separately, AB#47608 is titled "Add ODBC result-set performance benchmarks" and is still `New`; #417 delivered the harness and this PR is the optimization that harness measures, so the link is reasonable, but the work item's title and state no longer describe what is landing here.

## Verification

- `cargo check -p mssql-odbc -p mssql-tds --all-targets` — clean. This is the check that matters after a rebase onto `main`'s handle / `free_handle` / `disconnect` changes, since a semantic conflict there would surface as a build break rather than a diff hunk.
- `cargo nextest run -p mssql-odbc -p mssql-tds --lib --no-fail-fast` — 3148 tests, 3141 passed, 1 skipped, 7 failed. All 7 are `certificate_validator` and `win_tls::validate` cases that need locally generated `tests/test_certificates/*.pem` fixtures; these are the documented clean-tree failures and are unrelated to this diff.
- GitHub check runs on `040d0ab5`: 13 success, 9 still running, 0 failures.
- Read the prior inline threads and review bodies before writing. Every earlier finding, including my own `try_decode_buffered` decimal-asymmetry one, is resolved on this head.

David-Engel (commented):

## Summary

Re-review of head `e929095b`. No findings.

This head is `040d0ab5` plus a merge of `origin/main` (`e6223c2b`); the merge base is now exactly current `main`, and the only content delta versus the previously reviewed tree is main's `mssql-py-core` version bump. I re-derived the full `main...HEAD` diff rather than trusting the earlier pass, and re-checked the substantive parts of it directly.

What I looked at this round:

- **Buffered decoder parity.** `GenericDecoder::try_decode_buffered` / `try_decode_buffered_into` were compared arm-by-arm against the async `decode_into` / `SqlTypeDecode` paths for `IntN`, `FltN`, `BitN`, `MoneyN`, `DateTimeN`, `DateN`, `TimeN`/`DateTime2N`/`DateTimeOffsetN`, `Guid`, `DecimalN`/`NumericN`, character, and binary. Length-dispatch semantics match on both the accepted and rejected branches (including `FltN` treating any non-zero, non-4 length as an f64 and `BitN` treating any non-zero length as one byte), and the error text/type for `IntN`, `MoneyN`, and GUID lengths is the same. `scale_time_value` is a straight extraction of the existing async scaling table, so the temporal path cannot drift between the two decoders.
- **Non-destructive miss.** Every `read!`/`take*` miss in `try_decode_buffered_into` returns before the corresponding `writer.write_*`, so a `None` never leaves a partially written column, and `try_read_buffered_row_into` only calls `consume_bytes(consumed)` for whole columns it completed. `get_buffered_slice()` bounds the sync view at `buffer_length`, so stale capacity past the received data cannot make an incomplete value look complete.
- **Cursor-state safety.** `try_next_row_cursor`, `try_next_buffered_row_into`, `try_read_row_column`, and `try_finish_row_into` all bail to `Pending` when `active_row_read_state` is not `Idle`, which is also how a header parked by `peek_past_current_row` (`row_already_positioned`) stays correct — it can only be observed with a `RowPaused` state, so the sync path defers and the async `next_row_cursor` still hands it over. `try_read_row_column`'s `RowEnded` / out-of-range / `AlreadyConsumed` ordering matches `read_row_column` + `resume_to_column`.
- **Eligibility cache.** `BufferedRowSupport` keys on `Arc::ptr_eq` but holds a strong `Arc`, so the compared allocation cannot be freed and reused underneath it. It is boxed, per the client-context guidance.
- **NBCROW bitmap reuse.** `Arc::get_mut` guarantees uniqueness before the in-place `copy_from_slice`, and the length-mismatch and shared-refcount cases both fall back to a fresh allocation.
- **ODBC bound writes.** `deliver_encoded_string`'s direct paths reproduce `copy_with_nul` semantics exactly for null target, `buf_elements == 0`, truncation flag, and indicator byte count, and the `SQL_C_WCHAR` manual loop matches `copy_with_nul`'s `min(buf_len - 1)` + terminator. `deliver_fixed_bound` writes unaligned and null-checks the target. Both new `unsafe fn`s carry safety contracts.
- **Binding ordering.** `BoundRowWriter::take_binding`'s forward-only cursor and the `bindings.last().column_number == column_count` eligibility gate both depend on `StmtState::bindings` being sorted; `set_binding` maintains that with `binary_search` + `insert`, and `bindings_stay_ordered_by_column_however_they_were_bound` covers it. The one shape the gate lets through with a gap (e.g. columns 1 and 3 bound out of 3) already consumed the intervening column on the old `read_row_column` path, so `SQLGetData` reachability is unchanged.
- **Timeout accounting.** `request_timeout_start` is a pure short-circuit — `update_remaining_timeout` already no-ops when `remaining_request_timeout` is `None`.
- **`extract_datetime_parts` refactor.** The extracted `date_parts` / `time_parts` / `datetime2_parts` / `datetimeoffset_parts` produce the same field sets, including the `datetimeoffset` UTC-to-local carry and its out-of-range `None`.

## Verification

- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean.
- `cargo nextest run -p mssql-odbc -p mssql-tds --lib --no-fail-fast` — 3148 run, 3141 passed, 1 skipped. The 7 failures are all `certificate_validator` / `win_tls::validate` cases that need locally generated `tests/test_certificates/*.pem` fixtures; they are unrelated to this diff and fail the same way on a clean tree.
- Required CI on this head is pending, with nothing failing.
- The differential parity test `bound_row_writer_matches_established_conversion_path` and the buffered/async decode-equivalence helper in `decoder.rs` are what make me comfortable with the duplicated decode arms; they compare typed slot bytes, indicators, and outcomes against `deliver_bound` and `decode_into` rather than asserting hand-written expectations.

Looks ready for human review — but this is not an approval, and a human still needs to sign off. Worth a reviewer's attention on the perf numbers themselves, since those come from the dedicated lab runs rather than anything reproducible here.

David-Engel (commented):

## Summary

Re-review of head `6a715de6`. No findings.

This head is `e929095b` (previously reviewed clean) merged with current `main` (`db7d2665`, picking up #397 and #413). I diffed the merge against a mechanical composition of both sides and the only content difference is the intentional `mssql-odbc-bench/perf-lab/baseline-commit.txt` re-pin to `db7d2665`, which matches the PR description. No evil-merge edits landed in the optimization code.

The one file both sides touched is `mssql-odbc/src/conversion/fetch_convert.rs`, and both changes survived: #413's `parse_numeric_text` refactor plus its column-side blanks-only padding tests, and this PR's extraction of `date_parts`/`time_parts`/`datetime2_parts`/`datetimeoffset_parts` out of `extract_datetime_parts`. The extracted helpers are behavior-preserving — the early-returning arms produce exactly what the old inline assignments left in `p`, and the `DateTime`/`SmallDateTime`/`_` arms still fall through to `Some(p)`/`None`.

I also re-derived the merge interaction that mattered most here: #413 extended character-to-numeric *column* conversion, and the new fast path does not bypass it. `BoundRowWriter::write_string` only takes the direct route when the bound C type is `SQL_C_CHAR`/`SQL_C_WCHAR` and the wire encoding already matches; a character column bound to a numeric C type falls through to `deliver_bound` → `convert_typed_c` → `parse_numeric_text`, so the two changes compose.

Spot checks I redid on the underlying optimization at this head, since the merge moved surrounding code:

- `deliver_encoded_string`'s direct paths are byte-equivalent to `deliver_bound`'s string paths. `char_buf_elements(SQL_C_CHAR, stride) == stride`, so passing `stride` to `copy_with_nul` is the same capacity; `copy_with_nul` already tolerates a null destination, and the wide path's explicit null/zero-capacity handling matches its return values. Indicators are byte counts in both.
- The direct temporal writes match `convert_datetime_c` arm for arm, including `datetimeoffset_parts` returning `None` mapping to `Restricted` exactly as `extract_datetime_parts` → `ConvError::Restricted` did.
- `can_write_complete_rows` and `BoundRowWriter::take_binding` both depend on `bindings` being sorted by ordinal, which `StmtState::set_binding` guarantees via binary-search insertion. Column 0 (bookmark) and past-the-end leftover bindings are skipped the same way the column-cursor path skipped them.
- `try_decode_buffered`/`try_decode_buffered_into` stay wire-compatible with the async decoder on the permissive arms too — `FltN` non-4 lengths falling to `f64`, `BitN` non-zero lengths reading one byte, and `Char`/`VarChar` using a `USHORTLEN` prefix all mirror `decode`/`decode_into`/`StringDecoder`. `NText`/`Text` correctly fall through to the async path.
- `try_decode_buffered_into` never writes a partial value before returning `Ok(None)`: every multi-field arm resolves all `read!` reads before calling the writer, so an incomplete column is genuinely non-destructive and `try_read_buffered_row_into` only consumes the bytes of columns it fully wrote.
- The `Arc::get_mut` guard on `nbc_bitmap_scratch` is load-bearing in the new sync `try_receive_row_header` the same way it is in `read_nbc_bitmap`: a live `RowPauseState` keeps the refcount at 2 and forces a fresh allocation instead of refilling a bitmap still in use.
- `try_next_row_cursor` does not check `row_already_positioned`, but that is safe: the only setter is `peek_past_current_row`, which sets it from a successful `next_row_cursor`, and that always leaves `RowPaused` (zero-column rows included) — so the sync attempt returns `Pending` and the async path handles the flag.
- `columns_read` staying `0` when the whole-row path errors is inconsequential: `last_column_read` is only consulted on the `fetch_error.is_none()` branches.

## Verification

- Clippy: `cargo clippy -p mssqlodbc -p mssql-tds --all-features --all-targets -- -D warnings` — clean. (`--frozen` is not usable here; `Cargo.lock` is not tracked in this repo.)
- Tests: `cargo nextest run -p mssqlodbc -p mssql-tds --lib --no-fail-fast` — 3178 run, 3171 passed, 1 skipped, 7 failed. All 7 failures are `certificate_validator` / `win_tls::validate` tests failing with `CertificateNotFound { path: "tests/test_certificates/valid_cert.pem" }`, i.e. missing generated fixtures in this scratch worktree. They are unrelated to the diff and touch no changed file.
- Merge audit: `git diff e929095b HEAD` compared against `git diff e6223c2b db7d2665` to confirm the merge introduced nothing beyond the upstream commits and the baseline re-pin.

No new dependency source was introduced — `uuid` was already a workspace dependency; only `mssql-odbc/Cargo.toml` gained the reference.

Looks ready for human review, but this is not an approval and a human still needs to sign off. Two things worth a reviewer's eye that I cannot check here: the perf numbers come from the dedicated lab runs rather than anything reproducible in this environment, and the checklist still leaves full `cargo btest` unchecked. Required checks on this head are green or still pending (both dedicated performance runs and several ADO build legs were pending at review time) — confirm those land green before merging.

mssql-tds/src/datatypes/decoder.rs
```diff
@@ -596,6 +699,753 @@ impl PlpColumnStream {
...
+                    return Ok(None);
+                };
+                let Some(value_bytes) = reader.take_slice(data_length) else {
+                    return Ok(None);
+                };
+                let encoding = if matches!(base, TdsDataType::NVarChar | TdsDataType::NChar) {
+                    EncodingType::Utf16
+                } else {
+                    let collation: SqlCollation = collation_bytes.as_slice().try_into()?;
+                    if collation.utf8() {
```
Theekshna, line 787: Could you add equivalence coverage for a narrow-character `sql_variant` under both UTF-8 and non-UTF-8 collations? The existing variant test uses `NVarChar`, so it only reaches `EncodingType::Utf16`; a wrong encoding tag in this arm can still consume the complete value and therefore will not trigger the safe async fallback. The coverage report also marks this branch uncovered. I could not run a mutation locally because the pinned compiler failed while building `crc32fast`, so I am raising this as a test-sufficiency question rather than claiming an observed regression.

Theekshna (commented):
> [!NOTE]

## Summary

The synchronous buffered decode and direct bound-row paths preserve the existing async fallbacks, and the exact-head correctness and performance checks are green. I left one test-sufficiency question inline.

**msodbcsql parity:** Fine for the changed behavior; the optimization preserves the established conversion path rather than introducing a new ODBC interpretation.
**Test sufficiency, unit and e2e:** One gap: the buffered `sql_variant` narrow-character collation branch is not exercised; details are inline. The rest of the differential and fallback coverage is strong.
**Divergences documented:** Fine; this PR introduces no new reference-driver divergence, and the existing UTF-8 `SQL_C_CHAR` behavior remains documented in `.github/instructions/mssql-odbc.instructions.md`.
**PR description currency:** Current for head `6a715de6`; AB#47608 is linked, the unchecked full-`cargo btest` box matches the stated targeted validation, and all reported checks pass.

mssql-tds/src/datatypes/decoder.rs
saurabh500: Added in 016b3439. The new test decodes narrow `sql_variant` strings under both UTF-8 and Windows-1252 collations through buffered and async paths, then compares the base type, consumed width, raw value, encoding tag, and decoded text; forcing the buffered UTF-8 arm to use the LCID encoding makes it fail.

> [!NOTE]

David-Engel (commented):

## Summary

Re-review of head `016b3439`, from an unattended run. No findings.

The only change since the previously reviewed `6a715de6` is `Test narrow variant collations` — 56 added lines, all inside `#[cfg(test)]` in `mssql-tds/src/datatypes/decoder.rs`. `git diff 6a715de6 016b3439` touches nothing else, so the production diff (`try_decode_buffered*`, `CursorPoll`/`BufferedRowPoll`, `BoundRowWriter`, the transport and token-stream hooks) is byte-identical to the head I reviewed clean, and I did not re-litigate it. It answers @Theekshna's test-sufficiency question on the narrow-character `sql_variant` arm.

I checked the new test rather than taking the reply at face value:

- **It is not vacuous.** Replacing `collation.utf8()` with `false` in `try_decode_buffered_variant` (decoder.rs:787) makes it fail: `left: LcidBased(... IsUtf8: true ...)`, `right: Utf8`. The failure message also confirms the first case's collation bytes really do carry the UTF-8 flag — `[0x09, 0x04, 0x00, 0x04, 0x00]` is LCID 1033 with `0x40` in the flag byte — so the two cases genuinely straddle the branch rather than both landing on `LcidBased`. Restored afterwards; `git status` is clean.
- **It pins wire compatibility, not just the tag.** Each case is decoded through both `try_decode_buffered_variant` and async `read_sql_variant_with_base`, and asserts base type, `used == wire.len()`, `reader.pos == wire.len()`, the `EncodingType` on both sides, value equality between them, and the decoded text. The consumed-width assertion is the part that matters here: a wrong encoding tag on a complete value would not trigger the safe fallback, which was exactly @Theekshna's concern.
- **Parity holds at the source.** The buffered arm (decoder.rs:783-793) and `decode_seven_propbyte_variant` (decoder.rs:3433-3439) make the same `NVarChar`/`NChar` → `Utf16`, else `collation.utf8()` → `Utf8`, else `LcidBased` decision from the same 5 collation bytes. The buffered path omits the async `MAX_ALLOC_SIZE` guard, which is fine because `take_slice(data_length)` cannot exceed the already-bounded `get_buffered_slice()` view, and its trailing `reader.position != payload.len()` check rejects any leftover property bytes.

Spot checks I redid at this head rather than carrying forward:

- `get_buffered_slice()` can only invert its range if `buffer_position > buffer_length`, which `consume_bytes` rejects; `shift_data_to_front` already `copy_within`s the same range, so the new accessor adds no reachable panic.
- `deliver_fixed_bound` writes `size_of::<T>()` to the indicator and null-checks the target; every direct temporal target (`SQL_C_TYPE_DATE` for `date`, `SQL_C_SS_TIME2` for `time`, `SQL_C_TYPE_TIMESTAMP` for `datetime2`, `SQL_C_SS_TIMESTAMPOFFSET` for `datetimeoffset`) selects the `convert_datetime_c` arm that returns `ConvOk::Exact`, so no `Truncated` outcome is dropped by taking the shortcut. The `None` → `RowOutcome::Error(RowIssue::Restricted)` case writes nothing, matching `extract_datetime_parts(...).ok_or(ConvError::Restricted)?` returning before any `write_fixed`.

## Verification

- `git diff 6a715de6 016b3439` — 1 file, +56, test-only. Full PR diff re-derived against merge base `db7d2665` (13 files, +4433/-164); `origin/main` is exactly that commit, so the base is current.
- Mutation check on the new test, as above (fail under mutation, pass restored).
- `cargo nextest run -p mssql-tds --lib decoder::` — 230 run, 230 passed (was 229 at the prior head; the delta is the new test).
- `cargo nextest run -p mssqlodbc -p mssql-tds --lib --no-fail-fast` — 3179 run, 3172 passed, 1 skipped, 7 failed. All 7 are the documented clean-tree `certificate_validator` / `win_tls::validate` failures needing locally generated `tests/test_certificates/*.pem`; unrelated to this diff and unchanged from the prior head.
- `cargo clippy -p mssqlodbc -p mssql-tds --all-features --all-targets -- -D warnings` — clean. `cargo fmt -- --check` — clean.
- Read every inline thread and review body first, including the suppressed-findings slot. Every earlier finding is resolved on this head; I re-filed none of them.

## Findings

### Blocking

None.

### Suggestion

None.

### Nit

None.

Two things for the human reviewer that I cannot settle here: the performance numbers come from the dedicated lab runs (Linux 171459 / Windows 171458) rather than anything reproducible in this environment, and the checklist still leaves full `cargo btest` unchecked, consistent with the targeted validation the description claims. AB#47608 is linked but is titled "Add ODBC result-set performance benchmarks" and still `New` — #417 delivered that harness and this PR is the optimization it measures, so the title no longer describes what is landing.

shiwanigupta0809 (approved):

PR #421, opened 2026-08-28
Close the batch when only an RPC terminator follows a no-row result [AB#47531]

David-Engel:
## Description

A parameterized statement runs as `sp_executesql`, and the `DONEINPROC` that ends each of its inner statements **always** carries `DONE_MORE` — the RPC's own `RETURNSTATUS`/`DONEPROC` trailer still follows it. Statement-wise navigation took that flag at face value, so every parameterized `INSERT`/`UPDATE`/`DELETE`/`PRINT` left a 0-column cursor open, and the next execute on that statement failed with `24000 Invalid cursor state` even though nothing was pending.

`advance_to_result_boundary` now resolves the ambiguity the same way msodbcsql does. After positioning on a no-row result that came from a `DONEINPROC`, it reads ahead one token:

- only the `RETURNSTATUS`/`DONEPROC` trailer left → close the batch, so the statement is immediately re-executable;
- anything else → park it in `TdsClient::parked_token`, replayed by the next token read, so a real second statement is still navigable.

Plain SQL batches are untouched: they have no trailer, so their `DONE_MORE` was never ambiguous. All four token-read loops in `tds_client.rs` now go through one `next_response_token` helper, so a parked token can never be skipped by a drain path. The slot is cleared in `begin_command` and `adopt_recovered_session`.

Also guards `next_rowset`, which collapsed `NoRows` boundaries unconditionally and would read past the end of a response when a counted no-row result was the last thing in it. That over-read was pre-existing (a terminal `DONE` with `DONE_COUNT` reaches it today) and this change makes it reachable in more cases.

### Why the linked work item describes a transaction bug

[AB#47531](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47531) is filed as "Rollback after an intermediate commit loses previously committed rows" and points at `SQLEndTran` / `mssql-odbc/src/api/txn.rs`. Transaction handling turned out not to be involved: verified that commit → rollback → commit keeps the committed batches when the inserts are unparameterized.

The failing test (`test_cursor_rollback_performance_patterns`) inserts each batch row **with parameters** inside `try/except`, and calls `cursor.rollback()` from the `except` block. The `24000` on the second insert of every batch therefore sent *each* batch — including the two meant to commit — down the rollback path, leaving 0 rows instead of 10.

## Related Issues

[AB#47531](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47531)

## Testing

**Parity against msodbcsql 18**, same C program run against both drivers through unixODBC:

| Case | msodbcsql18 | before | after |
| --- | --- | --- | --- |
| parameterized `INSERT` ×2, no close between | SUCCESS | **24000** | SUCCESS |
| parameterized `UPDATE` ×2, no close between | SUCCESS | **24000** | SUCCESS |
| parameterized `PRINT` ×2, no close between | SUCCESS | **24000** | SUCCESS |
| parameterized 2-statement DML batch, then re-execute | 24000 | 24000 | 24000 |
| parameterized `INSERT; SELECT`, then re-execute | 24000 | 24000 | 24000 |
| plain 2-statement DML batch, then re-execute | 24000 | 24000 | 24000 |
| `SQLPrepare` + `SQLExecute` ×2, no close between | SUCCESS | SUCCESS | SUCCESS |
| parameterized `INSERT; SELECT` navigation (`SQLRowCount`, `SQLMoreResults`) | 1 / rowset / NO_DATA | same | same |

**New tests** — all confirmed to fail with the fix toggled off:

- `mssql-tds` unit tests: `sp_executesql_single_dml_closes_the_batch`, `sp_executesql_second_dml_keeps_the_batch_open`, `sp_executesql_dml_then_select_keeps_the_select_navigable`, `sp_executesql_message_only_statement_closes_the_batch`, `plain_batch_more_flag_is_not_second_guessed`, `next_rowset_stops_at_a_terminal_no_row_result`
- e2e `exec_direct_test`: `ParameterizedDmlIsImmediatelyReExecutable`, `ParameterizedBatchWithPendingResultsRejectsReExecute`
- e2e `transaction_test`: `CommitRollbackCommitKeepsOnlyCommittedBatches` — the reported symptom, transcribed from the failing Python test

**Suites** (SQL Server 2022 in Docker):

- `cargo btest`: 1879 `mssql-tds` + 1070 `mssql-odbc` unit tests pass
- `run_e2e.sh --compare-with-msodbcsql`: **30 parity, 0 divergences, 0 failures** across all 30 e2e binaries

None. No public API change. The only behavior change is that a parameterized statement whose response holds nothing but the RPC trailer no longer reports an open cursor — which is what msodbcsql already reported, and what the ODBC statement transition tables require for a result that is only an update count.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,66 @@ impl TdsClient {
...
+        loop {
+            let token = self.next_response_token(parser_context).await?;
+            match token {
+                Tokens::ReturnStatus(return_status) => {
+                    self.last_return_status = ReturnStatus::Received(return_status.value);
+                }
+                Tokens::DoneProc(done) if !done.has_more() && !done.has_error() => {
+                    self.execution_context.set_has_open_batch(false);
+                    return Ok(());
+                }
```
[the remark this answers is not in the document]
David-Engel: Fixed in 5fb1f537. Consuming the terminal `DONEPROC` now clears both `remaining_request_timeout` and `cancel_handle`, with regression assertions covering a timed, cancellable `sp_executesql` request.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
```diff
@@ -390,6 +390,14 @@ pub struct TdsClient {
     /// (see `drain_active_row_if_needed`).
     row_already_positioned: bool,
 
+    /// A token read by the RPC-terminator look-ahead
+    /// (see [`settle_rpc_terminator`](Self::settle_rpc_terminator)) that turned
+    /// out to start another result instead of ending the response. Replayed by
+    /// [`next_response_token`](Self::next_response_token) before the wire is
+    /// touched again. Boxed to keep the client — and every future that owns it
```
saurabh500, line 482: Nice

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,68 @@ impl TdsClient {
...
+            .await?;
+        self.update_remaining_timeout(start);
+        self.observe_response_token(&token)?;
+        Ok(token)
+    }
+
+    /// Resolves whether an RPC response has anything left after the no-row
+    /// result just positioned on, closing the batch when it does not.
+    ///
+    /// Every `DONEINPROC` carries the MORE flag, because the RPC's own
```
saurabh500, line 4372: This comment be valid for any driver and not just odbc. 
If the above statement is true, can we remove references to ODBC? 

The tests are generally executing T-SQL which are RPCs and that pattern is possible with any driver. .

mssql-tds/src/connection/tds_client.rs
```diff
@@ -9001,6 +9055,169 @@ mod tests {
...
+    /// the ambiguity the terminator look-ahead resolves.
+    fn done_in_proc_count(cmd: CurrentCommand, rows: u64) -> Tokens {
+        Tokens::DoneInProc(DoneToken {
+            status: DoneStatus::COUNT | DoneStatus::MORE,
+            cur_cmd: cmd,
+            row_count: rows,
+        })
+    }
+
+    fn done_proc_final() -> Tokens {
```
saurabh500, line 10367: Add comments above the method.

saurabh500 (approved):
The change correctly distinguishes a terminal RPC RETURNSTATUS/DONEPROC trailer from a real subsequent result and preserves non-terminator tokens for replay. The implementation matches the msodbcsql DONEINPROC/DONE_MORE look-ahead behavior, covers both the close and keep-open paths, and clears request-scoped state when the response completes. I found no blocking issues.

mssql-tds/src/connection/tds_client.rs
David-Engel: Agreed. Updated in 86f6a83d to describe the protocol-level RPC result ambiguity and its effect on every caller, without tying it to ODBC behavior.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
David-Engel: Added in 86f6a83d: the helper now documents that it constructs the terminal `DONEPROC` closing an RPC response.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
David-Engel: Thanks!

> [!NOTE]

saurabh500 (approved):

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,64 @@ impl TdsClient {
...
+                Tokens::ReturnStatus(return_status) => {
+                    self.last_return_status = ReturnStatus::Received(return_status.value);
+                }
+                Tokens::DoneProc(done) if !done.has_more() && !done.has_error() => {
+                    self.remaining_request_timeout = None;
+                    self.cancel_handle = None;
+                    self.execution_context.set_has_open_batch(false);
+                    return Ok(());
+                }
+                other => {
```
Vahid-b, line 4442: `RETURNVALUE` tokens in the RPC trailer are treated as the start of another result here. For `DONEINPROC(MORE) -> RETURNSTATUS -> RETURNVALUE -> DONEPROC(FINAL)`, this parks the output parameter and leaves `has_open_batch` true, so the output value is not captured and the statement can remain non-re-executable. I reproduced this with a targeted token-sequence test: it failed on `!client.has_open_batch()`. Native msodbcsql continues parsing output parameters after recognizing `RETURNSTATUS` and before the terminal `DONEPROC` (`sqlctokn.cpp`, `OnDone`). Please finalize/push `Tokens::ReturnValue` here and continue scanning, with a regression test covering one or more output parameters.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,64 @@ impl TdsClient {
...
+    /// ambiguity here lets every caller distinguish an exhausted RPC from one
+    /// with another result still available.
+    ///
+    /// Anything that is not the terminator is parked for the next read, so the
+    /// look-ahead never consumes a result the caller has yet to navigate to.
+    /// The `RETURNSTATUS` that precedes `DONEPROC` in every RPC response is
+    /// absorbed exactly as the main loop absorbs it.
+    async fn settle_rpc_terminator(&mut self, parser_context: &ParserContext) -> TdsResult<()> {
+        loop {
+            let token = self.next_response_token(parser_context).await?;
```
Vahid-b: A failure from this added look-ahead leaves the batch marked open. `advance_to_result_boundary` sets `has_open_batch` from `DONE_MORE` before calling this method, and a timeout, cancellation, or transport error propagates through `?` without clearing the batch flag or request-scoped timeout/cancel state. I reproduced the state leak by ending the mock stream after the initial `DONEINPROC`; execution returned an error but `has_open_batch()` remained true. Please clean up the batch/request state on this failure path (and apply the existing retirement policy when synchronization was not recovered), with timeout/cancellation coverage.

Vahid-b (commented):
Found two blocking state-management issues in the RPC-terminator look-ahead: output-parameter trailers are not fully consumed, and look-ahead failures leave the client busy. Both were reproduced with targeted temporary tests; no temporary changes remain in the review worktree.

mssql-tds/src/connection/tds_client.rs
David-Engel: Fixed in 7a4c47d4. Any look-ahead error now clears the open-batch, timeout, and cancellation state and applies the existing failed-drain retirement policy; regression coverage exercises both timeout and cancellation.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
David-Engel: Fixed in 7a4c47d4. The look-ahead now finalizes and captures every `RETURNVALUE` before continuing to the terminal `DONEPROC`; regression coverage verifies two output parameters are surfaced and the batch closes.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,82 @@ impl TdsClient {
...
+                    let return_value = self.finalize_return_value(return_value_token)?;
+                    self.push_return_value(return_value);
+                }
+                Tokens::DoneProc(done) if !done.has_more() && !done.has_error() => {
+                    self.remaining_request_timeout = None;
+                    self.cancel_handle = None;
+                    self.execution_context.set_has_open_batch(false);
+                    return Ok(());
+                }
+                other => {
```
David-Engel, line 4442: The catch-all park arm also swallows the two trailer tokens the main loop knows how to absorb: `Tokens::SessionState` and `Tokens::EnvChange`.

MS-TDS places `SESSIONSTATE` in the response tail, immediately before the final DONE — i.e. exactly between the `DONEINPROC(MORE)` and the `DONEPROC(FINAL)` this look-ahead is scanning for. `connect_retry_count` defaults to `1` (`client_context.rs:478`, `:541`), so session recovery is negotiated on a default connection and these tokens are live traffic, not a theoretical case. When one lands in the trailer, the loop parks it and returns `Ok(())` with `has_open_batch` still `true` — so the statement keeps its 0-column cursor and the next execute still fails with 24000, which is the bug this PR is fixing.

This is narrower than it first looks: the server only emits `SESSIONSTATE` when session state actually changed, so a bare parameterized `INSERT` won't produce one — which is presumably why the new e2e tests pass. But an `sp_executesql` body that contains a `SET` option (or the first request after connect) would, and then the fix quietly doesn't apply. `ENVCHANGE` has the same shape.

Suggested handling, mirroring the main loop's arms at `:3926` and `:3934`:

```rust
Tokens::EnvChange(env_change) => {
    if env_change.sub_type == EnvChangeTokenSubType::ResetConnection {
        self.on_reset_connection_ack();
    }
    self.execution_context
        .capture_change_property(&env_change, &mut self.negotiated_settings)?;
}
Tokens::SessionState(session_state) => {
    self.recovery_context.process_session_state(&session_state)?;
}
```

Worth a unit test alongside `sp_executesql_output_parameters_close_the_batch` with a `SESSIONSTATE` token in the trailer, asserting both that the batch closes and that the state reached `recovery_context`.

If you've captured an `sp_executesql` response on a recovery-enabled connection and confirmed `SESSIONSTATE` never lands after the last `DONEINPROC`, say so and I'll withdraw this — I couldn't observe it from here.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,82 @@ impl TdsClient {
...
+    /// look-ahead never consumes a result the caller has yet to navigate to.
+    /// The `RETURNSTATUS` that precedes `DONEPROC` in every RPC response is
+    /// absorbed exactly as the main loop absorbs it.
+    async fn settle_rpc_terminator(&mut self, parser_context: &ParserContext) -> TdsResult<()> {
+        let result = self.settle_rpc_terminator_inner(parser_context).await;
+        if let Err(error) = &result {
+            self.execution_context.set_has_open_batch(false);
+            self.remaining_request_timeout = None;
+            self.cancel_handle = None;
+            self.retire_after_failed_drain(error);
```
David-Engel: This retires the connection on *any* error out of the inner loop, but not every error from that loop means the reader lost its place.

`settle_rpc_terminator_inner` calls `finalize_return_value(...)?` at `:4110`, which returns `Error::ColumnEncryptionError` when an encrypted output parameter has no retained CEK or didn't arrive as varbinary ciphertext (`:4312`, `:4325`). In both cases the RETURNVALUE token was fully parsed and the stream is still perfectly in sync — it's a key/config error, not a desync. `retire_after_failed_drain` nonetheless calls `mark_known_dead()` (so a pool discards the connection) and, since `ColumnEncryptionError` isn't in `drain_failure_lost_the_transport`, also clears `session_recovery_negotiated`.

The identical arm in `advance_to_result_boundary` (`:3938`) propagates the same error with a bare `?` and leaves the connection alive, so the AE failure mode now depends on whether the output parameter happened to arrive before or after the last `DONEINPROC`.

The timeout/cancellation/transport coverage `Vahid-b` asked for is right and should stay — it's the blanket `if let Err(error)` that's too wide. Scoping the retirement to the read failures, and letting a `finalize_return_value` error propagate the way the main loop does, would keep both behaviours. E.g. finalize the return value before the match, or hoist the AE error out with a marker type so only read errors reach `retire_after_failed_drain`.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,82 @@ impl TdsClient {
...
+    ///
+    /// Every `DONEINPROC` carries the MORE flag, because the RPC's own
+    /// `DONEPROC` still follows it, so the flag alone cannot tell "another
+    /// statement result is coming" from "that was the last one". Resolving that
+    /// ambiguity here lets every caller distinguish an exhausted RPC from one
+    /// with another result still available.
+    ///
+    /// Anything that is not the terminator is parked for the next read, so the
+    /// look-ahead never consumes a result the caller has yet to navigate to.
+    /// The `RETURNSTATUS` that precedes `DONEPROC` in every RPC response is
```
David-Engel: Nit: stale after `7a4c47d4` — the loop now absorbs `RETURNVALUE` as well, and that's the more interesting half (output parameters would otherwise be lost). Something like "The `RETURNSTATUS` and any `RETURNVALUE` tokens that precede `DONEPROC` are absorbed exactly as the main loop absorbs them" would keep it accurate.

David-Engel (commented):
### Automated review (unattended run)

> [!NOTE]
> This review came from an unattended automation run. The findings below were **not** checked by a human first — please weigh them accordingly. Posting `COMMENT` only; no approval, no change request.

**Verdict: issues** — 0 blocking, 2 suggestions, 1 nit.

#### What I checked

Diff against merge base `1ecefe1c` (3 files, +464/-45), plus the surrounding code the change turns on: `advance_to_result_boundary`, `next_rowset`, `drain_stream`, `consume_transaction_response`, `consume_done_token`, `observe_response_token`/`proves_request_ran`, `retire_after_failed_drain`/`drain_failure_lost_the_transport`, `finalize_return_value`, `begin_command`, `adopt_recovered_session`, and the row-read paths (`next_row_into` / `next_row_cursor`).

Things that hold up:

- **The parking invariant is airtight on the read side.** `receive_token` now has exactly one call site in `tds_client.rs` (inside `next_response_token`), so no drain or navigation loop can step over a parked token. The row paths go through `receive_row_into`, and they can't be reached with a token parked: `settle_rpc_terminator` only runs on the `NoRows` branch, which has already set `current_result_set_has_been_read_till_end = true`, so `next_row_into` short-circuits at its first guard.
- **Skipping `observe_response_token` on replay is right,** not a shortcut: `take_reset_dispatched()` is take-once and `ResetAckState` has already advanced, so re-observing would double-charge the acknowledgement.
- **Parked `DoneProc` with `DONE_ERROR` or `DONE_MORE`** falls through to the park arm and is replayed into the main loop's existing `has_error()` / more-results handling, which is the same outcome as before the change.
- **`plain_batch_more_flag_is_not_second_guessed`** is the right negative test — the `is_done_in_proc` gate is what keeps plain batches out of the look-ahead, and this pins it.
- **`next_rowset` guard** is correct and the test genuinely exercises the over-read (the mock runs dry rather than returning `End`).
- **e2e tests** use `#temp` tables and reset params before the follow-up `SELECT`, so no cross-test collision and no dangling parameter binding.

CI is green across all 20 checks (ADO validation, cross-repo `mssql-python` suite, Kerberos, coverage) and the coverage bot reports 100% diff coverage. Prior threads from Copilot, `saurabh500` and `Vahid-b` all read as answered by `5fb1f537` / `86f6a83d` / `7a4c47d4`; I did not re-file any of them.

#### Findings

Both suggestions are the same shape as the `RETURNVALUE` gap `Vahid-b` already found — tokens that the *main* loop absorbs but `settle_rpc_terminator_inner` does not, and an error-handling asymmetry between the two loops. Neither is blocking; see the inline comments for detail.

1. **Suggestion** — `SESSIONSTATE` (and `ENVCHANGE`) in the RPC trailer are parked, so the batch stays open and the fix silently doesn't apply. (`tds_client.rs:4119`)
2. **Suggestion** — the look-ahead failure path retires the connection for *any* error, including a `ColumnEncryptionError` from `finalize_return_value`, where the stream is still in sync. The identical arm in `advance_to_result_boundary` just propagates. (`tds_client.rs:4094`)
3. **Nit** — the doc comment now under-describes what the loop absorbs. (`tds_client.rs:4086`)

#### Verification limits

Read-and-trace only; I did not build or run the suite in this run, and I could not observe a live server to confirm where SQL Server actually places `SESSIONSTATE` in an `sp_executesql` response. Finding 1 rests on the MS-TDS ordering rule plus `connect_retry_count: 1` being the default in `client_context.rs`, not on a capture — treat it as a question worth answering rather than a proven defect.

mssql-tds/src/connection/tds_client.rs
David-Engel: Fixed in b9dcc532. The look-ahead now applies `SESSIONSTATE` and `ENVCHANGE` using the same handlers as the main loop; regression coverage verifies a trailer session-state token is applied and the batch closes.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
David-Engel: Fixed in b9dcc532. Retirement is now limited to token-read failures; output finalization errors are retained while the scanner consumes through `DONEPROC`, then returned with the synchronized connection alive. Added regression coverage for the AE error path.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
David-Engel: Updated in b9dcc532. The documentation now covers all response-tail control tokens absorbed by the scanner, including `RETURNVALUE`, session state, and environment changes.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,99 @@ impl TdsClient {
...
+    async fn settle_rpc_terminator(&mut self, parser_context: &ParserContext) -> TdsResult<()> {
+        let mut processing_error = None;
+        loop {
+            let token = match self.next_response_token(parser_context).await {
+                Ok(token) => token,
+                Err(error) => {
+                    self.execution_context.set_has_open_batch(false);
+                    self.remaining_request_timeout = None;
+                    self.cancel_handle = None;
+                    self.retire_after_failed_drain(&error);
```
shiwanigupta0809: A timeout or cancellation that lands on the look-ahead read now retires the connection, where the identical error one token earlier does not. Is that asymmetry intended to be this sharp?

The main loop reaches `receive_token` through the same `next_response_token`, but propagates with a bare `?` and leaves the connection alone. Here the failure arm calls `retire_after_failed_drain`, and for `TimeoutError` / `OperationCancelledError` `drain_failure_lost_the_transport` returns false — so the connection is both marked dead *and* has `session_recovery_negotiated` cleared.

The narrow case that bothers me is a single-statement parameterized `INSERT` whose budget expires between the `DONEINPROC` and the trailer. The server has already committed the row, and before this change the caller got a successful no-row result with the timeout surfacing later on `SQLMoreResults` / close. Now the same run returns an error from execute and permanently retires a connection whose statement actually succeeded. Reachable through `mssql-py-core`, which does apply `default_query_timeout` (the ODBC layer doesn't enforce `SQL_ATTR_QUERY_TIMEOUT` yet — AB#46385 — so it can't hit this today).

I can see the argument for retiring: the look-ahead is a read the caller never asked for, so on failure the caller can't reason about the stream position, and `sp_executesql_terminator_failure_retires_and_clears_request_state` pins it as deliberate. But that reasoning applies just as well to a timeout in the main loop, which doesn't retire — so whichever is right, the two paths now disagree about the same error on the same query. If retiring is the correct answer, the main loop looks like the one that should move; if it isn't, restricting the retirement to `drain_failure_lost_the_transport(&error)` would keep resiliency armed for a plain timeout.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -5515,6 +5607,9 @@ impl TdsClient {
     fn begin_command(&mut self) {
         self.settle_abandoned_reset_verification();
         self.info_messages.clear();
+        // A token parked by the look-ahead belongs to the previous response;
+        // replaying it here would desynchronize this command's reader.
+        self.parked_token = None;
```
shiwanigupta0809, line 6270: Nit: `close_query`'s state-reset block is the one place that resets per-response reader state and doesn't clear `parked_token` — it clears `current_metadata`, `return_values`, `remaining_request_timeout`, `cancel_handle`, `active_row_read_state` and `row_already_positioned`, but not this field.

Not a live bug as written: park and close are mutually exclusive in `settle_rpc_terminator`, so a parked token always implies an open batch, which means `close_query` never takes its `!has_open_batch()` early return and its drain replays the token through `next_response_token`. And even if the drain errors partway, this clear in `begin_command` catches it before the next command reads.

So purely a consistency point — `row_already_positioned` sits directly above it in that same block and is the closest analogue for exactly this kind of state. Adding it there would keep the invariant local instead of resting on the mutual-exclusion argument holding for every future caller.

shiwanigupta0809 (commented):
Read the change against the merge base and traced the surrounding paths (`advance_to_result_boundary`, `next_rowset`, `drain_stream`, `close_query`, `begin_command`, the row-read entry guards, and the `cursor_ops` callers of `next_rowset`).

I independently verified the central parity claim against the msodbcsql 18.6 source rather than taking it on trust, and it holds — msodbcsql really does peek-and-park here, it is not a state machine:

- `BATCHCTX::PeekToken` (`tds/TdsParser.h`) reads one token byte and then rewinds `cbInParsed` / `cbBytesRead` / `cbBytesSkipped` to un-read it.
- `PersistDoneToken` writes the DONE back into the token buffer so it is re-parsed.
- Three call sites in `odbc/sqlctokn.cpp` (~2208 `RS_SELECTION`, ~2293 `RS_ERROR`, ~2651 general) all gate on `eDoneType == DONEINPROC && (Status & DONE_MORE)` and then branch on `bToken != SQLRETURNSTATUS` → `SetParsingStatus(TDS_RESULTSPENDING)`.

Worth noting that this implementation's predicate is deliberately *broader* than msodbcsql's, and in the safe direction: msodbcsql decides on the single next token byte being `RETURNSTATUS`, whereas `settle_rpc_terminator` keeps scanning through `RETURNSTATUS`/`RETURNVALUE`/`ENVCHANGE`/`SESSIONSTATE` and only closes on an actual terminal `DONEPROC`. Since every non-terminator path either keeps scanning or parks, it can never close a batch that still has a result — a trailer that opens with `ENVCHANGE` closes correctly here where msodbcsql would report results pending. Also confirmed the `other => park` default is right rather than a gap: parking an `INFO` is what keeps a trailing `PRINT` navigable as its own message-only result, so absorbing it would have lost that result.

Nothing blocking. One question and one nit inline. Auto-merge is armed, so ignore both if you'd rather they ride a follow-up.

mssql-tds/src/connection/tds_client.rs
David-Engel: Agreed that the invariant is worth making explicit, but I’m leaving the reset block unchanged: an open batch forces `close_query` through the drain, whose first `next_response_token` takes the parked token, while a closed batch cannot have one. `begin_command` remains the defensive backstop after a failed drain, so an additional clear here would be behavior-neutral.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
David-Engel: Yes, the retirement is intentional at this boundary. Once `DONEINPROC(MORE)` has triggered look-ahead, a timeout/cancellation leaves us unable to distinguish an incomplete trailer from another result, so reusing the stream risks stale response bytes contaminating the next command; the pre-existing main-loop asymmetry should be handled separately rather than weakening this path.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
```diff
@@ -9001,6 +9100,295 @@ mod tests {
...
+        ]);
+        client
+            .execute(
+                "INSERT INTO t VALUES (1); INSERT INTO t VALUES (2)".to_string(),
+                (),
+            )
+            .await
+            .unwrap();
+        assert!(
+            client.has_open_batch(),
```
David-Engel, line 10677: This test is the negative half of the fix, but its single assertion survives removing the thing it names.

`done_count` builds `Tokens::Done`, not `Tokens::DoneInProc`. If the `&& is_done_in_proc` gate at line 3882 were dropped so the look-ahead ran on every `DONE` with `MORE`, `settle_rpc_terminator` would read the second `done_count(..., false)`, fall through to the `other => park` arm at 4136, and return `Ok(())` with `has_open_batch` still `true` — so `assert!(client.has_open_batch())` still passes and the gate's removal goes unnoticed.

The two cases *are* distinguishable, just not by the flag: with the gate nothing is read ahead, without it the second `DONE` ends up in the park slot. One extra line pins it:

```rust
assert!(
    client.has_open_batch(),
    "a plain batch's DONE_MORE always means another statement result"
);
assert!(
    client.parked_token.is_none(),
    "a plain batch's DONE_MORE needs no look-ahead, so nothing is read past it"
);
```

Not merge-blocking — the gate is a latency/eagerness safeguard rather than a correctness one (an ungated look-ahead would park rather than mis-close), and the rest of the new tests are genuinely non-vacuous. But this is the one test whose name promises coverage it does not provide.

(Derived from reading the fixture and the park arm, not from an observed mutation run — worth confirming before acting on it.)

David-Engel (commented):
### Automated review (unattended run)

> [!NOTE]
> This review came from an unattended automation run. The findings below were **not** checked by a human first — please weigh them accordingly. Posting `COMMENT` only; no approval, no change request.

**Verdict: issues** — 0 blocking, 1 suggestion, 0 nits.

#### Summary

Resolves the `DONEINPROC`/`DONE_MORE` ambiguity that left a 0-column cursor open after every parameterized no-row statement, by reading ahead one token and either closing the batch (only the RPC `RETURNSTATUS`/`RETURNVALUE`/`ENVCHANGE`/`SESSIONSTATE`/`DONEPROC` trailer remains) or parking the token for replay. Also guards `next_rowset` against reading past a terminal no-row result. The mechanism holds up against the surrounding code and against msodbcsql's `PeekToken`/`PersistDoneToken` shape that `shiwanigupta0809` verified in the source; I found nothing blocking.

#### Verification

Diffed against merge base `1ecefe1c` (3 files, +543/-45) and traced the code the change turns on: `advance_to_result_boundary`, `settle_rpc_terminator`, `next_response_token`, `next_rowset` and its five callers (`cursor_ops::fetch`, `drain_cursor_response`, `describe_parameter_encryption`, `get_dtc_address`), `drain_stream`, `consume_transaction_response`, `consume_done_token`, `close_query`, `begin_command`, `adopt_recovered_session`, `retire_after_failed_drain`/`drain_failure_lost_the_transport`, `process_session_state`, `capture_change_property`. Read all three comment endpoints first; the Copilot, `saurabh500`, `Vahid-b` and `shiwanigupta0809` threads all read as answered by `5fb1f537` / `86f6a83d` / `7a4c47d4` / `b9dcc532`, and I have not re-filed any of them.

What I re-checked on this head rather than taking on trust:

- **Single read site.** `transport.receive_token` now has exactly one call site in `tds_client.rs` (line 4064, inside `next_response_token`), so no drain or navigation loop can step over a parked token. `begin_command` and `adopt_recovered_session` both clear the slot.
- **The row paths cannot be entered with a token parked.** `settle_rpc_terminator` only runs on the `NoRows` branch, and `current_result_set_has_been_read_till_end` is already `true` at line 3828 before the call at 3886, so `next_row_into` short-circuits at its first guard.
- **The b9dcc532 trailer arms match the main loop.** `EnvChange` (3926-3932) and `SessionState` (3934-3936) are mirrored at 4113-4128; the only difference is that errors are recorded in `processing_error` and the scan continues to `DONEPROC`, which is strictly safer than the main loop's bare `?` because it leaves the reader on a token boundary.
- **`sp_executesql_session_state_trailer_closes_the_batch` is not vacuous.** `process_session_state` sets `master_recovery_disabled` only on `sequence_number == u32::MAX`, and it defaults `false`; drop the `SessionState` arm and both assertions fail.
- **`sp_executesql_output_decryption_error_does_not_retire_connection` pins the b9dcc532 narrowing** — under the previous blanket `if let Err(...)` arm it would fail on `!connection_known_dead()`.
- **Nested-proc parity.** For `EXEC outer` where an inner proc's last statement is a counted `DONEINPROC(MORE)`, the trailer opens with `DONEPROC(MORE)`, which hits the `other => park` arm and keeps the batch open — the same verdict msodbcsql reaches, since its single-byte peek also sees a non-`RETURNSTATUS` token.

CI is green on this head across all 20 checks (ADO validation, cross-repo `mssql-python` suite, Kerberos, coverage), and the coverage bot reports 100% diff coverage over 286 lines.

#### Findings

**Blocking** — none.

**Suggestion**

1. `plain_batch_more_flag_is_not_second_guessed` does not fail when the `is_done_in_proc` gate it is named for is removed. See the inline comment for the reasoning and a one-line strengthening.

**Nit** — none.

#### Verification limits

Read-and-trace only; I did not build or run the suite in this run, so finding 1 is derived from the token types in the test fixture rather than from an observed mutation run. `done_count` builds `Tokens::Done`, and the `other => park` arm accepts it, which is what makes the assertion insensitive to the gate — worth a 30-second confirmation before acting on it.

shiwanigupta0809 (approved):

mssql-tds/src/connection/tds_client.rs
David-Engel: Confirmed and fixed in 3ae26b89. The negative test now asserts `parked_token` remains empty, so removing the `is_done_in_proc` gate causes the test to fail rather than merely preserving the open-batch flag.

> [!NOTE]

David-Engel (commented):
### Automated review (unattended run)

> [!NOTE]
> This review came from an unattended automation run. The findings below were **not** checked by a human first — please weigh them accordingly. Posting `COMMENT` only; no approval, no change request.

**Verdict: clean** — 0 blocking, 0 suggestions, 0 nits.

#### Summary

Head `3ae26b89` adds only the test strengthening asked for on `b9dcc532`, so this pass focused on the one thing the previous automated reviews explicitly could not do: they were read-and-trace only and flagged the `plain_batch_more_flag_is_not_second_guessed` finding as unverified by an actual mutation run. I built the crate and ran the mutations. The finding was real and the fix works.

#### Mutation runs (this head, `cargo nextest run -p mssql-tds --lib`)

Baseline: 12 selected tests, all pass.

| Mutation | Expected to fail | Result |
| --- | --- | --- |
| Drop the `is_done_in_proc` gate at `tds_client.rs:3882` so the look-ahead runs on every `DONE` with `MORE` | `plain_batch_more_flag_is_not_second_guessed` | FAIL, on the new `parked_token.is_none()` assertion — the assertion added in `3ae26b89` is exactly what makes this detectable; `has_open_batch()` alone still passed under the mutation, as reported |
| Revert the `next_rowset` guard to the previous `NoRows { .. } => continue` | `next_rowset_stops_at_a_terminal_no_row_result` | FAIL, `ConnectionClosed("test")` — the over-read is real, not a mock artifact |

Both mutations were reverted; the worktree is back at `3ae26b89` with a clean `git status`, and the 12 tests pass again.

Also re-checked on this head rather than carrying it forward from the earlier passes: `transport.receive_token` still has exactly one call site in `tds_client.rs` (inside `next_response_token`), so the parking invariant holds for every drain and navigation loop; `begin_command` clears the slot before the next command reads; and the row paths remain unreachable with a token parked because `current_result_set_has_been_read_till_end` is set at `:3828` before the `settle_rpc_terminator` call at `:3886`. The two new ODBC e2e tests use `#temp` tables, reset parameter bindings before the follow-up `SELECT`, and the transaction test relies on the fixture's `TearDown` to restore autocommit, matching the sibling tests in that file.

CI is green on this head across all 20 checks, including the ADO validation pipeline, the cross-repo `mssql-python` suite and Kerberos; the coverage bot reports 95% diff coverage.

Prior threads from Copilot, `saurabh500`, `Vahid-b`, `shiwanigupta0809` and the earlier automated passes all read as answered by `5fb1f537` / `86f6a83d` / `7a4c47d4` / `b9dcc532` / `3ae26b89`. Nothing re-filed.

#### Verification limits

Unit tests and mutations only — no live SQL Server, so the two new ODBC e2e tests were read, not executed. The `SESSIONSTATE`-in-trailer question raised earlier still rests on the MS-TDS ordering rule rather than a capture, but the handling merged in `b9dcc532` makes the answer moot either way.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,99 @@ impl TdsClient {
...
+                }
+                Tokens::ReturnValue(return_value_token) => {
+                    match self.finalize_return_value(return_value_token) {
+                        Ok(return_value) => self.push_return_value(return_value),
+                        Err(error) => {
+                            processing_error.get_or_insert(error);
+                        }
+                    };
+                }
+                Tokens::EnvChange(env_change) => {
```
David-Engel, line 4419: This arm and the `SESSIONSTATE` arm below it were added together in `b9dcc532`, but only the `SESSIONSTATE` half is pinned by a test (`sp_executesql_session_state_trailer_closes_the_batch`). I deleted lines 4113-4123 so a trailer `ENVCHANGE` falls through to `other => park`, then ran `cargo nextest run -p mssql-tds --lib`: the result set is identical to baseline — nothing catches it. The coverage bot flags the same lines as the only uncovered ones in the diff (`Missing lines 4113-4122, 4126-4127`).

The mutation is not benign. Parking a trailer `ENVCHANGE` leaves `has_open_batch` true, which is the 24000-on-re-execute bug this PR exists to fix, and it is reachable: `capture_change_property` records the transaction descriptor for `BeginTransaction` / `CommitTransaction` / `RollbackTransaction`, so an `sp_executesql` body that opens or closes a transaction (or runs under implicit transactions) puts one in exactly this position.

A sibling of the session-state test would pin it — an `ENVCHANGE(BeginTransaction)` with a `UInt64` container between the `DONEINPROC` and the terminal `DONEPROC`, asserting both that `execution_context.transaction_descriptor` picked up the new value and that `!client.has_open_batch()`.

Not merge-blocking: the arm is correct as written, it just has nothing holding it in place against a future edit.

David-Engel (commented):
### Automated review (unattended run)

> [!NOTE]
> This review came from an unattended automation run. The findings below were **not** checked by a human first — please weigh them accordingly. Posting `COMMENT` only; no approval, no change request.

**Verdict: issues** — 0 blocking, 1 suggestion, 0 nits.

#### Summary

The `DONEINPROC`/`DONE_MORE` look-ahead holds up on this head: the mechanism, the parking invariant and the `next_rowset` guard all re-check out, and the two mutations an earlier pass ran (dropping the `is_done_in_proc` gate, reverting the `next_rowset` guard) are still caught. One finding: of the two response-tail control arms added in `b9dcc532`, only the `SESSIONSTATE` one is pinned by a test — deleting the `ENVCHANGE` arm leaves the whole `mssql-tds` suite green.

An earlier automated pass posted a clean marker for this same head. This pass differs on that one point only: it mutated each trailer arm individually rather than the two the earlier pass covered.

#### What I ran

Worktree at `3ae26b89`, `CARGO_TARGET_DIR` pinned, `git status` clean before and after.

| Run | Result |
| --- | --- |
| Baseline, 16 selected `sp_executesql_*` / `plain_batch_more` / `next_rowset_stops` / `terminator` tests | all pass |
| Mutation: delete the `Tokens::EnvChange` arm (`tds_client.rs:4113-4123`) so a trailer `ENVCHANGE` falls to `other => park` | full `cargo nextest run -p mssql-tds --lib` unchanged — 1967 pass, and the only 7 failures are the pre-existing `certificate_validator` / `win_tls` pinning tests that need `tests/test_certificates/*.pem` fixtures and fail on a clean tree too |
| Revert, re-run the 12 targeted tests | all pass, tree clean |

The coverage bot reaches the same place independently: `Missing lines 4113-4122, 4126-4127` are the only uncovered lines in the diff.

Also re-verified on this head rather than carried forward: `transport.receive_token` still has exactly one call site in `tds_client.rs` (inside `next_response_token`), so no drain or navigation loop can step over a parked token; `begin_command` (`:5612`) and `adopt_recovered_session` (`:641`) both clear the slot; the row paths cannot be entered with a token parked, because `current_result_set_has_been_read_till_end` is set at `:3828` before the `settle_rpc_terminator` call at `:3886`; and the look-ahead is gated on `has_update_count || saw_message`, so a collapsed DDL/no-op `DONEINPROC` still walks the trailer through the main loop as before. On the e2e side, both new tests rebind their parameters inside each iteration before the execute and `SQL_RESET_PARAMS` before the readback, and `#batch_t` is created while autocommit is still on, so the batch-1 rollback cannot take the table with it.

CI is green on this head across all 20 checks (ADO validation, cross-repo `mssql-python`, Kerberos, coverage); the coverage bot reports 95% diff coverage.

Prior threads from Copilot, `saurabh500`, `Vahid-b`, `shiwanigupta0809` and the earlier automated passes all read as answered by `5fb1f537` / `86f6a83d` / `7a4c47d4` / `b9dcc532` / `3ae26b89`. Nothing re-filed.

#### Findings

**Blocking** — none.

**Suggestion**

1. The `ENVCHANGE` trailer arm is unguarded: removing it changes no test result, and the removal reintroduces the exact failure mode `b9dcc532` fixed for `SESSIONSTATE`. Inline at `tds_client.rs:4113`.

**Nit** — none.

#### Verification limits

Unit tests and mutations only — no live SQL Server, so the two new ODBC e2e tests were read, not executed.

mssql-tds/src/connection/tds_client.rs
David-Engel: Confirmed and fixed in 672e6be2. Added an RPC-trailer `BeginTransaction` `ENVCHANGE` regression that verifies the transaction descriptor is applied and the batch closes; removing the arm now fails this test.

> [!NOTE]

David-Engel (commented):

## Summary

Resolves the `DONEINPROC`/`DONE_MORE` ambiguity that left a 0-column cursor open after every parameterized no-row statement: `advance_to_result_boundary` now reads ahead and either closes the batch (nothing but the RPC `RETURNSTATUS`/`RETURNVALUE`/`ENVCHANGE`/`SESSIONSTATE`/`DONEPROC` trailer remains) or parks the token for replay, and `next_rowset` no longer reads past a terminal no-row result. The only change on this head (`672e6be2`) over `3ae26b89` is the `ENVCHANGE`-trailer regression test that a previous pass asked for, so this pass concentrated on whether that test actually pins the arm it names. It does. Nothing blocking found.

## Verification

Worktree at `672e6be2`, dedicated `CARGO_TARGET_DIR`, `git status` clean before and after.

| Run | Result |
| --- | --- |
| Baseline: `cargo nextest run -p mssql-tds --lib -E 'test(sp_executesql_) or test(plain_batch_more) or test(next_rowset_stops)'` | 13 tests, all pass |
| Mutation: delete the `Tokens::EnvChange` arm (`tds_client.rs:4113-4123`) so a trailer `ENVCHANGE` falls through to `other => park` | `sp_executesql_env_change_trailer_closes_the_batch` **FAILS** on `!client.has_open_batch()`; the other 12 still pass |
| Revert the mutation, re-run | 13 pass, tree clean |

So the new test is not vacuous, and it closes the last uncovered arm the coverage bot was reporting (`Missing lines 4113-4122`).

Also re-checked on this head rather than carried forward from the earlier passes:

- **Single read site.** `transport.receive_token` has exactly one call site in `tds_client.rs` (`:4064`, inside `next_response_token`), so no navigation or drain loop can step over a parked token; `begin_command` (`:5612`) and `adopt_recovered_session` (`:641`) both clear the slot.
- **The row paths cannot be entered with a token parked.** `settle_rpc_terminator` runs only on the `NoRows` branch, and `current_result_set_has_been_read_till_end` is already `true` at `:3828` before the call at `:3886`, so `advance`'s drain guard and `next_row_into` short-circuit. `drain_result_set_rows` reaches `receive_row_into` only after `drain_stream` consumed a `COLMETADATA` through `next_response_token`, which takes the parked token first.
- **`next_rowset`'s new `None` means the same thing `End` did.** A `NoRows` boundary with `has_open_batch() == false` is either a terminal `DONE` or a trailer the look-ahead consumed through `DONEPROC` — in both cases the response is fully read, so the `cursor_ops` callers that follow `next_rowset` with `drain_stream_or_retire` are no worse off than before.
- **`count_map` is write-only** (no reader in the crate), so the terminator `DONEPROC` bypassing the main loop's accumulation is inert.
- **Nested-proc parity.** `EXEC p` inside a plain batch, and a parameterized `EXEC`, both put a `DONEPROC(MORE)` in front of the trailer, which hits the park arm and keeps the batch open — the same verdict msodbcsql's single-byte peek reaches, and consistent with the PR's own parity table.

CI is green on this head across all 20 checks (ADO validation, cross-repo `mssql-python` suite, Kerberos, coverage). I read all three comment endpoints first: the Copilot, `saurabh500`, `Vahid-b`, `shiwanigupta0809` and earlier automated threads all read as answered by `5fb1f537` / `86f6a83d` / `7a4c47d4` / `b9dcc532` / `3ae26b89` / `672e6be2`. Nothing re-filed.

## Findings

**Blocking** — none.

**Suggestion** — none.

**Nit** — none.

No issues found, so this looks ready for human review — but this is a `COMMENT` review, not an approval, and the human review requirement still stands.

### Verification limits

Unit tests and mutation runs only. No live SQL Server was available, so the two new ODBC e2e tests were read, not executed, and the msodbcsql parity table in the description is taken on the author's evidence plus `shiwanigupta0809`'s source verification of `PeekToken`/`PersistDoneToken`.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3895,6 +3879,12 @@ impl TdsClient {
                             self.last_rows_affected = -1;
                         }
                         self.execution_context.set_has_open_batch(!is_last);
+                        if !is_last && is_done_in_proc {
+                            // The MORE flag may only be reporting the RPC's own
+                            // trailing DONEPROC; find out before the caller
+                            // treats it as another result.
+                            self.settle_rpc_terminator(&parser_context).await?;
```
Theekshna, line 4180: Two concerns about doing this before returning the boundary:

1. It blocks on the next statement's first token, and `next_response_token` spends `remaining_request_timeout` while it waits. For an RPC whose next statement is slow (`WAITFOR`, a long `UPDATE`), `SQLExecDirect` now blocks until that statement produces a token, and can time out a statement that already succeeded. The parity table has no slow-next-statement row — does msodbcsql read ahead here too, or only when the caller navigates?

2. The `?` discards the `NoRows` boundary computed just above, after `last_rows_affected`, `count_map` and `has_open_batch` have already been mutated. A trailer `ReturnValue` / `EnvChange` / `SessionState` failure now fails the current statement and drops its row count, where previously those tokens were processed on the next navigation call and failed that one instead.

mssql-tds/src/connection/tds_client.rs
Theekshna: This retires the connection on *any* look-ahead error, including a query timeout. Per the note on `settle_abandoned_reset_verification`, `receive_token` answers a timeout by draining to the attention acknowledgement, so the stream is still synchronized — but `retire_after_failed_drain` calls `mark_known_dead()` unconditionally, killing a healthy pooled connection.

The main loop's read at line 3794 propagates the same error with a bare `?` and doesn't retire. Is the asymmetry deliberate, or should this retire only when the failure actually desynchronized the stream?

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4049,6 +4048,99 @@ impl TdsClient {
...
+    /// statement result is coming" from "that was the last one". Resolving that
+    /// ambiguity here lets every caller distinguish an exhausted RPC from one
+    /// with another result still available.
+    ///
+    /// Anything that is not the terminator is parked for the next read, so the
+    /// look-ahead never consumes a result the caller has yet to navigate to.
+    /// The response-tail control tokens that precede `DONEPROC` are absorbed
+    /// exactly as the main loop absorbs them.
+    async fn settle_rpc_terminator(&mut self, parser_context: &ParserContext) -> TdsResult<()> {
+        let mut processing_error = None;
```
Theekshna, line 4388: No iteration cap on this loop. `ReturnStatus`, `ReturnValue`, `EnvChange` and `SessionState` all continue, so a server streaming those endlessly spins here forever; `advance_to_result_boundary` caps at 10000 for exactly this reason. When no query timeout is set, `remaining_request_timeout` is `None`, so there's nothing else to break out.

Theekshna (commented):
Reviewed the token-loop changes. The core design holds up: exactly one `receive_token` call site, so a parked token can't be skipped by a drain path, and `parked_token` is cleared in both `begin_command` and `adopt_recovered_session`. Three questions inline, all about the look-ahead's failure modes rather than its happy path.

Two things outside the diff:

- Both approvals predate the last two commits (`7a4c47d4`, `672e6be2`), so `7a4c47d4 Handle RPC trailer output and failures` — the `processing_error` path — hasn't been reviewed by anyone yet. The PR currently reports `REVIEW_REQUIRED` / `BLOCKED`.
- The branch predates #425, so its cross-repo run fails `test_app_name_received_by_sql_server` on the old APP name. Worth rebasing so that isn't read as a real failure.

On the motivation: `test_cursor_rollback_performance_patterns` failed on runs 3259758 and 3284746 but passes on current `main`, and this branch's cross-repo run (3303197) matches main's (3303943) exactly — zero `24000` failures on either, nothing fixed, nothing regressed. The C parity table is still independent evidence of a genuine msodbcsql divergence, but the Python-level symptom in AB#47531 looks like it was already resolved by #383. Worth stating in the description so this is weighed as parity hardening rather than a fix for a live failure.

mssql-tds/src/connection/tds_client.rs
David-Engel: The eager read is intentional and matches msodbcsql: `BATCHCTX::PeekToken` reads the next token byte before rewinding it, so a slow following statement can delay this boundary there as well. Trailer-processing failures now continue through `DONEPROC` before returning the error, preserving synchronization; surfacing them from the current execute is deliberate because the RPC response, including its output/session state, was not successfully processed.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
David-Engel: Fixed in c91d8ca6. The trailer scanner now caps at 10,000 tokens, clears request state, retires the desynchronized connection, and returns a protocol error; a regression test exercises the limit.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
David-Engel: Fixed in c91d8ca6. Timeout and cancellation are no longer retired because `receive_token` has already drained through the attention ACK; request state is cleared while the synchronized connection and recovery capability remain intact. Unsynchronized transport/protocol failures still retire.

> [!NOTE]

Theekshna:
Following up on the `BATCHCTX::PeekToken` justification — I checked it against the msodbcsql source, and it supports the park-and-leave-batch-open behavior, so I'm dropping that objection. But one difference is worth a look before merge.

`PeekToken` (`tds/TdsParser.h`) reads a single byte and rewinds:

```cpp
HRESULT hr = GetBytes(pvOwner, sizeof(BYTE), pbToken, NULL, dwTimeout);
if (hr == S_OK || hr == S_INFO) {
    m_pParseCtx->cbInParsed--;  m_cbBytesRead--;  m_cbBytesSkipped--;
}
```

And at the decisive call site in `odbc/sqlctokn.cpp` it peeks with a deliberately short bound and treats expiry as a non-event:

```cpp
hr = pBatchCtx->PeekToken(lpstmt, &bToken, DEFAULT_ASYNC_TIMEOUT); // Timeout quickly
if ((hr != S_OK && hr != S_INFO) || bToken != SQLRETURNSTATUS) {
    if (hr == HRESULT_FROM_WIN32(ERROR_TIMEOUT))    // timeout is not an error here
        hr = S_OK;
    lpstmt->SetParsingStatus(TDS_RESULTSPENDING);
    pBatchCtx->SetAsyncReadRetry();
```

`settle_rpc_terminator` instead goes through `next_response_token`, which passes `self.remaining_request_timeout` to `receive_token` and then calls `update_remaining_timeout(start)`. So the look-ahead is charged against the caller's query timeout, and expiry fails the statement.

The result is that a server slow to emit the RPC trailer turns a query msodbcsql would complete (results pending, retry) into a timeout for us. It only shows up under load, so CI won't catch it. Two other smaller deltas in the same area: this consumes and commits the trailer (`push_return_value`, `capture_change_property`, `process_session_state`) rather than peeking one byte and rewinding, which is why the 10000 cap is needed at all; and msodbcsql gates the peek on `RS_SELECTION`/`RS_ERROR` rather than the no-row boundary.

Not blocking from my side, but if the timeout charging is intentional it should be called out in the doc comment, since the current one reads as straight parity.

David-Engel:
> Not blocking from my side, but if the timeout charging is intentional it should be called out in the doc comment.

Documented in `7c5f9d4b`. This implementation intentionally consumes the trailer under the remaining query timeout; expiry fails the request after the attention ACK restores synchronization, while leaving the connection reusable.

Theekshna (approved):

shiwanigupta0809 (approved):

Theekshna (commented):
> [!NOTE]

## Summary

The RPC-terminator look-ahead and parked-token path are coherent on this head, and I found no new code defect. The exact-head validation state does not match the checked PR checklist.

## Suggestion

The current `mssql-rs Pull request validation` run is failing: both ARM build jobs failed after about an hour, and `coverage-report` also failed while waiting on that validation. The description still checks `cargo btest` as passing. Please rerun or otherwise establish a green exact-head validation result before merge; this review cannot distinguish an infrastructure timeout from a head-specific failure.

**msodbcsql parity:** Fine; `Sql\Ntdbms\sqlncli\odbc\sqlctokn.cpp:2208-2239` uses the RPC trailer state to decide whether results remain, consistent with this look-ahead.
**Test sufficiency, unit and e2e:** Fine by code tracing; unit and e2e cases cover closure, pending results, iteration-limit retirement, and timeout cleanup. A local mutation rerun was blocked by the pinned compiler's `crc32fast` codegen failure, so no additional gap is asserted.
**Divergences documented:** Fine; the broader trailer-consuming behavior and its timeout semantics are documented in the implementation and PR design.
**PR description currency:** The mechanism and linked AB#47531 remain current, but the checked validation claim is contradicted by the exact-head checks. The previously raised #383 motivation question is not re-filed here.

Theekshna (approved):

PR #425, opened 2026-08-28
Fix LOGIN7 Unicode lengths and map ODBC APP name

David-Engel:
## Description

This change fixes malformed LOGIN7 packets for non-ASCII usernames, passwords, database names, hostnames, application names, and other Unicode login fields across every mssql-tds consumer. LOGIN7 field lengths are now encoded as UTF-16 code units, matching the UTF-16LE payload; field offsets use the protocol's unsigned representation; optional change-password data is included in record sizing; and oversized fields or records return a usage error instead of panicking or emitting invalid metadata.

It also maps the ODBC connection-string keyword `APP` to `ClientContext::application_name`. Previously, `APP` was recognized but sat in `KNOWN_IGNORED_KEYS`, so SQL Server saw the mssql-tds default `TDSX Rust Client` rather than the application-supplied value. Because `APP` was recognized, the parser raised no `01S00` warning either, so nothing surfaced the drop.

`APP` now maps to a `ConnectionParams` slot and is wired onto the `ClientContext` in `apply_connection_params`, matching msodbcsql's `KEY_APP` handling (`sqlcconn.cpp`).

The value is stored verbatim. msodbcsql additionally rejects an `APP` longer than 128 characters (`x_rgLookup[KEY_APP].cLimitSize`), but per-key length caps are an intentional non-goal of this parser (already documented under "Intentional non-goals" in `mssql-odbc/docs/connection_string_parser.md`). Instead, the shared serializer now enforces LOGIN7's `USHORT` field bounds and 128 KiB minus one byte record limit for every caller-controlled login field.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47534

## Testing

- New parser and application tests cover mapping `APP` into `ClientContext::application_name` and reconstructed connection strings.
- Byte-level mssql-tds tests verify emitted LOGIN7 metadata and offset advancement for both astral application names and non-ASCII database names.
- Oversize tests verify that fields beyond `u16::MAX` UTF-16 code units and records beyond 128 KiB minus one byte are rejected as usage errors.
- Record-size coverage verifies optional change-password data and long SSPI tokens, including the short-length sentinel and long-length field.
- `DriverConnectLiveTest.AppKeywordReachesServerAppName` now builds on the suite's configured connection string for SQL authentication, integrated authentication, DSN, and connection-string override modes, then verifies `SELECT APP_NAME()`.
- `cargo nextest run -p mssql-tds --lib -E 'test(/message::login/)'` — 39 passed.
- `cargo clippy -p mssql-tds --lib --all-features --frozen -- -D warnings` — passed.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (remaining local failures are the known environment-dependent `mssql-tds::connectivity` / TLS-mock tests, unrelated to this change)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ### Changed
 
+- `mssql-tds`: LOGIN7 now encodes Unicode field lengths as UTF-16 code units
+  and rejects oversized records instead of producing malformed packets. This
+  fixes login failures with non-ASCII usernames, passwords, database names,
+  hostnames, and application names across all bindings.
+
 - `mssql-tds`: `TdsClient::language()` now returns the language negotiated at
   login (from the server's `ENVCHANGE`) instead of always returning an empty
   string, matching its documentation.
@@ -137,4 +142,3 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   rejected as a protocol error. Such a packet is malformed — it neither carries
   payload nor terminates a message — but was previously consumed as a
   zero-length packet. Empty end-of-message packets remain legal.
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ### Fixed
 
+- `mssql-odbc`: the `APP` connection-string keyword is now sent as the TDS login
+  application name. It was recognized but ignored, so `APP_NAME()` reported the
+  default `TDSX Rust Client` value without warning that `APP` had been dropped.
+
+- `mssql-tds`: LOGIN7 record sizing now includes the optional change-password
+  value, preventing a caller-supplied value from making the declared packet
+  length shorter than the serialized payload.
+
 - `mssql-tds`: idle connection resiliency (transparent session recovery) now
   works end to end. The client-side gate that authorizes a reconnect is now set
   from the server's `FEATUREEXTACK` acknowledgment — previously it was only ever
--- a/mssql-odbc/src/connection/connection_string_parser.rs
+++ b/mssql-odbc/src/connection/connection_string_parser.rs
@@ -48,7 +48,6 @@ const KNOWN_IGNORED_KEYS: &[&str] = &[
     "description",
     "desc",
     "driver",
-    "app",
     "wsid",
     "language",
     "network",
@@ -135,6 +134,7 @@ enum ConnAttrKey {
     Authentication,
     TrustedConnection,
     ServerSpn,
+    App,
     ApplicationIntent,
     MultiSubnetFailover,
     ConnectRetryCount,
@@ -203,6 +203,7 @@ pub(crate) struct ConnectionParams {
     pub(crate) authentication: Option<String>,
     pub(crate) trusted_connection: Option<bool>,
     pub(crate) server_spn: Option<String>,
+    pub(crate) application_name: Option<String>,
     pub(crate) application_intent: Option<String>,
     pub(crate) multi_subnet_failover: Option<bool>,
     pub(crate) connect_retry_count: Option<u32>,
@@ -248,6 +249,9 @@ impl ConnectionParams {
         }
         if let Some(server_spn) = &self.server_spn {
             parts.push(format!("ServerSPN={}", quote_odbc_value(server_spn)));
+        }
+        if let Some(application_name) = &self.application_name {
+            parts.push(format!("APP={}", quote_odbc_value(application_name)));
         }
         if let Some(application_intent) = &self.application_intent {
             parts.push(format!("ApplicationIntent={application_intent}"));
@@ -318,6 +322,7 @@ impl fmt::Debug for ConnectionParams {
             .field("authentication", &self.authentication)
             .field("trusted_connection", &self.trusted_connection)
             .field("server_spn", &self.server_spn)
+            .field("application_name", &self.application_name)
             .field("application_intent", &self.application_intent)
             .field("multi_subnet_failover", &self.multi_subnet_failover)
             .field("connect_retry_count", &self.connect_retry_count)
@@ -373,6 +378,7 @@ const MAPPED_KEYS: &[(&str, ConnAttrKey)] = &[
     ("authentication", ConnAttrKey::Authentication),
     ("trusted_connection", ConnAttrKey::TrustedConnection),
     ("serverspn", ConnAttrKey::ServerSpn),
+    ("app", ConnAttrKey::App),
     ("applicationintent", ConnAttrKey::ApplicationIntent),
     ("multisubnetfailover", ConnAttrKey::MultiSubnetFailover),
     ("connectretrycount", ConnAttrKey::ConnectRetryCount),
@@ -439,6 +445,7 @@ fn assign_value(
             params.trusted_connection = Some(is_yes(value));
         }
         ConnAttrKey::ServerSpn => params.server_spn = Some(value.to_string()),
+        ConnAttrKey::App => params.application_name = Some(value.to_string()),
         ConnAttrKey::ApplicationIntent => {
             validate_attr(lower, value, APPLICATION_INTENT_VALUES)?;
             params.application_intent = Some(value.to_string());
@@ -1290,7 +1297,6 @@ mod tests {
         for key in [
             "Driver",
             "DSN",
-            "APP",
             "WSID",
             "Language",
             "Network",
@@ -1307,6 +1313,23 @@ mod tests {
             assert_eq!(p.server, "h", "key {key}");
             assert_eq!(p.uid, "u", "key {key}");
         }
+    }
+
+    #[test]
+    fn app_maps_to_application_name() {
+        for key in ["APP", "app", "App"] {
+            let s = format!("Server=h;{key}=My App;UID=u;<PW>=p");
+            let (p, warn) = parse_connection_string(&cs(&s)).unwrap();
+            assert!(!warn, "{key} should map cleanly to APP");
+            assert_eq!(p.application_name.as_deref(), Some("My App"), "key {key}");
+        }
+    }
+
+    #[test]
+    fn app_is_absent_when_not_supplied() {
+        let (p, warn) = parse_connection_string(&cs("Server=h;UID=u;<PW>=p")).unwrap();
+        assert!(!warn);
+        assert_eq!(p.application_name, None);
     }
 
     #[test]
@@ -1485,7 +1508,7 @@ mod tests {
     #[test]
     fn fmt_as_odbc_conn_str_includes_new_keys() {
         let (p, _) = parse_connection_string(
-            "Server=h;ApplicationIntent=ReadOnly;MultiSubnetFailover=yes;ConnectRetryCount=2;ConnectRetryInterval=10;KeepAlive=30;KeepAliveInterval=1;PacketSize=4096;IpAddressPreference=IPv4First;ServerSPN=svc;HostnameInCertificate=cn;ServerCertificate=/tmp/c.pem;UID=u",
+            "Server=h;ApplicationIntent=ReadOnly;MultiSubnetFailover=yes;ConnectRetryCount=2;ConnectRetryInterval=10;KeepAlive=30;KeepAliveInterval=1;PacketSize=4096;IpAddressPreference=IPv4First;ServerSPN=svc;APP=MyApp;HostnameInCertificate=cn;ServerCertificate=/tmp/c.pem;UID=u",
         )
         .unwrap();
         let out = p.fmt_as_odbc_conn_str();
@@ -1499,6 +1522,7 @@ mod tests {
             "PacketSize=4096",
             "IpAddressPreference=IPv4First",
             "ServerSPN=svc",
+            "APP=MyApp",
             "HostnameInCertificate=cn",
             "ServerCertificate=/tmp/c.pem",
         ] {
--- a/mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
@@ -555,24 +555,13 @@ TEST_F(DriverConnectLiveTest, NewConnectionAttributesParity) {
 // Regression coverage for AB#47534, where APP parsed cleanly but was dropped
 // before the login packet, so the server saw the driver's default name instead.
 TEST_F(DriverConnectLiveTest, AppKeywordReachesServerAppName) {
-    auto& cfg = ODBCTestConfig::Instance();
-    if (!cfg.HasSqlAuth()) {
-        GTEST_SKIP() << "Requires SQL auth (ODBC_TEST_SERVER + ODBC_TEST_UID + "
-                        "ODBC_TEST_PWD); see follow-up issue for capability gating";
-    }
     const std::string kAppName = "AppKeywordE2E";
-    const std::string cs =
-        "Driver={" + cfg.Driver() + "}"
-        ";Server=" + cfg.Server() +
-        ";UID=" + cfg.Uid() +
-        ";PWD=" + cfg.Pwd() +
-        ";TrustServerCertificate=" + cfg.TrustCert() +
-        ";APP=" + kAppName;
+    SqlTString connstr = ODBCTestUtils::BuildConnectionString();
+    connstr += ODBCTestUtils::ToSqlTStr(";APP=" + kAppName);
 
     SQLHDBC hdbc = SQL_NULL_HDBC;
     ASSERT_EQ(SQL_SUCCESS, SQLAllocHandle(SQL_HANDLE_DBC, env_, &hdbc));
 
-    SqlTString connstr = ODBCTestUtils::ToSqlTStr(cs);
     SQLTCHAR outStr[1024] = {};
     SQLSMALLINT outLen = 0;
     SQLRETURN rc = SQLDriverConnect(hdbc, nullptr,
--- a/mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
@@ -551,6 +551,56 @@ TEST_F(DriverConnectLiveTest, NewConnectionAttributesParity) {
     }
 }
 
+// APP must reach the server in the TDS login packet, which APP_NAME() reports.
+// Regression coverage for AB#47534, where APP parsed cleanly but was dropped
+// before the login packet, so the server saw the driver's default name instead.
+TEST_F(DriverConnectLiveTest, AppKeywordReachesServerAppName) {
+    auto& cfg = ODBCTestConfig::Instance();
+    if (!cfg.HasSqlAuth()) {
+        GTEST_SKIP() << "Requires SQL auth (ODBC_TEST_SERVER + ODBC_TEST_UID + "
+                        "ODBC_TEST_PWD); see follow-up issue for capability gating";
+    }
+    const std::string kAppName = "AppKeywordE2E";
+    const std::string cs =
+        "Driver={" + cfg.Driver() + "}"
+        ";Server=" + cfg.Server() +
+        ";UID=" + cfg.Uid() +
+        ";PWD=" + cfg.Pwd() +
+        ";TrustServerCertificate=" + cfg.TrustCert() +
+        ";APP=" + kAppName;
+
+    SQLHDBC hdbc = SQL_NULL_HDBC;
+    ASSERT_EQ(SQL_SUCCESS, SQLAllocHandle(SQL_HANDLE_DBC, env_, &hdbc));
+
+    SqlTString connstr = ODBCTestUtils::ToSqlTStr(cs);
+    SQLTCHAR outStr[1024] = {};
+    SQLSMALLINT outLen = 0;
+    SQLRETURN rc = SQLDriverConnect(hdbc, nullptr,
+                                    const_cast<SQLTCHAR*>(connstr.c_str()),
+                                    static_cast<SQLSMALLINT>(connstr.size()),
+                                    outStr, 1024, &outLen, SQL_DRIVER_NOPROMPT);
+    ASSERT_TRUE(SQL_SUCCEEDED(rc)) << "rc=" << rc;
+    EXPECT_FALSE(ODBCTestUtils::HasDiagState(SQL_HANDLE_DBC, hdbc, "01S00"));
+
+    SQLHSTMT hstmt = SQL_NULL_HSTMT;
+    ASSERT_EQ(SQL_SUCCESS, SQLAllocHandle(SQL_HANDLE_STMT, hdbc, &hstmt));
+
+    SqlTString sql = ODBCTestUtils::ToSqlTStr("SELECT APP_NAME()");
+    ASSERT_TRUE(SQL_SUCCEEDED(
+        SQLExecDirect(hstmt, const_cast<SQLTCHAR*>(sql.c_str()), SQL_NTS)));
+    ASSERT_TRUE(SQL_SUCCEEDED(SQLFetch(hstmt)));
+
+    SQLTCHAR buf[256] = {};
+    SQLLEN ind = 0;
+    ASSERT_TRUE(SQL_SUCCEEDED(
+        SQLGetData(hstmt, 1, SQL_C_TCHAR, buf, sizeof(buf), &ind)));
+    EXPECT_EQ(kAppName, ODBCTestUtils::ToNarrow(SqlTString(buf)));
+
+    SQLFreeHandle(SQL_HANDLE_STMT, hstmt);
+    SQLDisconnect(hdbc);
+    SQLFreeHandle(SQL_HANDLE_DBC, hdbc);
+}
+
 // mssql-odbc rejects an invalid value on a validated connection-string key with
 // HY024 at connect time. The reference msodbcsql driver is laxer: it treats an
 // invalid ApplicationIntent enum as a plain SQL_ERROR without HY024, and a
--- a/mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
@@ -556,8 +556,9 @@ TEST_F(DriverConnectLiveTest, NewConnectionAttributesParity) {
 // before the login packet, so the server saw the driver's default name instead.
 TEST_F(DriverConnectLiveTest, AppKeywordReachesServerAppName) {
     const std::string kAppName = "AppKeywordE2E";
-    SqlTString connstr = ODBCTestUtils::BuildConnectionString();
-    connstr += ODBCTestUtils::ToSqlTStr(";APP=" + kAppName);
+    SqlTString connstr =
+        ODBCTestUtils::ToSqlTStr("APP=" + kAppName + ";") +
+        ODBCTestUtils::BuildConnectionString();
 
     SQLHDBC hdbc = SQL_NULL_HDBC;
     ASSERT_EQ(SQL_SUCCESS, SQLAllocHandle(SQL_HANDLE_DBC, env_, &hdbc));
--- a/mssql-tds/src/message/login.rs
+++ b/mssql-tds/src/message/login.rs
@@ -34,6 +34,7 @@ use crate::io::token_stream::{ParserContext, TdsTokenStreamReader};
 use tracing::{Level, debug, event, info, trace};
 
 pub(crate) const FIXED_LOGIN_RECORD_LENGTH: i32 = 94;
+const MAX_LOGIN_RECORD_LENGTH: usize = 128 * 1024 - 1;
 
 #[derive(Default, Debug, Clone, PartialEq, Eq)]
 pub(crate) struct RoutingInfo {
@@ -806,15 +807,15 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     /// Calculate the length of the login record.
     /// This includes the fixed length of the login record, the length of the variable length fields,
     /// and the length of the feature extension data.
-    fn calculate_login_record_length(&self) -> (i32, i32) {
-        let mut login_record_length = FIXED_LOGIN_RECORD_LENGTH;
+    fn calculate_login_record_length(&self) -> TdsResult<(i32, i32)> {
+        let mut login_record_length = FIXED_LOGIN_RECORD_LENGTH as usize;
         login_record_length += self.model.user_input.len_bytes();
         login_record_length += self.model.transport_context.len_bytes();
         login_record_length += 4; // Feature extension offset size.
 
         // Add SSPI token length if present
         if let Some(sspi_token) = &self.model.sspi_token {
-            login_record_length += sspi_token.len() as i32;
+            login_record_length += sspi_token.len();
         }
 
         // We write the feature extension at the end of the login record. This is not necessary, but makes reading
@@ -822,11 +823,17 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
         // for the feature extension.
         let feature_extension_offset = login_record_length;
         login_record_length += self.features_request.len_bytes();
-        (login_record_length, feature_extension_offset)
+        if login_record_length > MAX_LOGIN_RECORD_LENGTH {
+            return Err(crate::error::Error::ProtocolError(format!(
+                "LOGIN7 record length {login_record_length} exceeds the maximum of {MAX_LOGIN_RECORD_LENGTH} bytes"
+            )));
+        }
+        Ok((login_record_length as i32, feature_extension_offset as i32))
     }
 
     pub(crate) async fn serialize(&mut self) -> TdsResult<()> {
-        let (login_record_length, feature_extension_offset) = self.calculate_login_record_length();
+        let (login_record_length, feature_extension_offset) =
+            self.calculate_login_record_length()?;
         trace!(login_record_length);
         self.payload_writer
             .write_i32_async(login_record_length)
@@ -1051,7 +1058,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     /// Writes the value of the hostname of the client to the login packet.
     async fn write_hostname(&mut self) -> TdsResult<()> {
         if self
-            .write_metadata(utf16_code_units(&self.model.user_input.workstation_id) as i16)
+            .write_metadata(utf16_code_units(&self.model.user_input.workstation_id)?)
             .await?
         {
             self.deferred_actions_indicator
@@ -1067,7 +1074,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
             TdsAuthenticationMethod::Password
         ) {
             if self
-                .write_metadata(utf16_code_units(&self.model.user_input.user_name) as i16)
+                .write_metadata(utf16_code_units(&self.model.user_input.user_name)?)
                 .await?
             {
                 self.deferred_actions_indicator
@@ -1084,7 +1091,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     async fn write_password(&mut self) -> TdsResult<()> {
         if self.model.user_input.tds_authentication_method == TdsAuthenticationMethod::Password {
             if self
-                .write_metadata(utf16_code_units(&self.model.user_input.password) as i16)
+                .write_metadata(utf16_code_units(&self.model.user_input.password)?)
                 .await?
             {
                 self.deferred_actions_indicator
@@ -1100,7 +1107,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     /// Writes the value of the application name provided in the client context to the login packet.
     async fn write_app_name(&mut self) -> TdsResult<()> {
         if self
-            .write_metadata(utf16_code_units(&self.model.user_input.application_name) as i16)
+            .write_metadata(utf16_code_units(&self.model.user_input.application_name)?)
             .await?
         {
             self.deferred_actions_indicator
@@ -1113,9 +1120,9 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     /// Uses get_login_server_name() to get DataSource format (host,port) for TCP connections.
     async fn write_server_name(&mut self) -> TdsResult<()> {
         if self
-            .write_metadata(
-                utf16_code_units(&self.model.transport_context.get_login_server_name()) as i16,
-            )
+            .write_metadata(utf16_code_units(
+                &self.model.transport_context.get_login_server_name(),
+            )?)
             .await?
         {
             self.deferred_actions_indicator
@@ -1127,7 +1134,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     async fn write_feature_ext(&mut self) -> TdsResult<()> {
         // The offset at which to read the feature extension length.
         self.payload_writer
-            .write_i16_async(self.content_next_offset as i16)
+            .write_u16_async(self.current_offset()?)
             .await?;
 
         // Length of the size of feature extension offset data, which is a DWORD.
@@ -1145,7 +1152,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     /// This is also called the Client interface name.
     async fn write_library_name(&mut self) -> TdsResult<()> {
         if self
-            .write_metadata(utf16_code_units(&self.model.user_input.library_name) as i16)
+            .write_metadata(utf16_code_units(&self.model.user_input.library_name)?)
             .await?
         {
             self.deferred_actions_indicator
@@ -1156,7 +1163,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
 
     async fn write_language(&mut self) -> TdsResult<()> {
         if self
-            .write_metadata(utf16_code_units(&self.model.user_input.language) as i16)
+            .write_metadata(utf16_code_units(&self.model.user_input.language)?)
             .await?
         {
             self.deferred_actions_indicator
@@ -1168,7 +1175,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     /// Writes the value of the database name, which we are connecting to.
     async fn write_database(&mut self) -> TdsResult<()> {
         if self
-            .write_metadata(utf16_code_units(&self.model.user_input.database) as i16)
+            .write_metadata(utf16_code_units(&self.model.user_input.database)?)
             .await?
         {
             self.deferred_actions_indicator
@@ -1194,15 +1201,15 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
         if sspi_len == 0 {
             // No SSPI token - write offset and zero length
             self.payload_writer
-                .write_i16_async(self.content_next_offset as i16)
+                .write_u16_async(self.current_offset()?)
                 .await?;
-            self.payload_writer.write_i16_async(0).await?;
+            self.payload_writer.write_u16_async(0).await?;
             return Ok(());
         }
 
         // Write offset
         self.payload_writer
-            .write_i16_async(self.content_next_offset as i16)
+            .write_u16_async(self.current_offset()?)
             .await?;
 
         if sspi_len <= 0xFFFF {
@@ -1223,7 +1230,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
 
     async fn write_attach_db_file(&mut self) -> TdsResult<()> {
         if self
-            .write_metadata(utf16_code_units(&self.model.user_input.attach_db_file) as i16)
+            .write_metadata(utf16_code_units(&self.model.user_input.attach_db_file)?)
             .await?
         {
             self.deferred_actions_indicator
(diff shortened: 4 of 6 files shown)


mssql-odbc/src/connection/connection_string_parser.rs
```diff
@@ -439,6 +445,7 @@ fn assign_value(
             params.trusted_connection = Some(is_yes(value));
         }
         ConnAttrKey::ServerSpn => params.server_spn = Some(value.to_string()),
+        ConnAttrKey::App => params.application_name = Some(value.to_string()),
```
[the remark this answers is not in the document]
David-Engel: Fixed in 936429cf. LOGIN7 now measures every Unicode metadata field and the total variable payload in UTF-16 code units, with a non-ASCII application-name regression test covering an astral character.

> [!NOTE]

mssql-tds/src/message/login.rs
```diff
@@ -1459,6 +1465,14 @@ mod tests {
         FeaturesRequest { features }
     }
 
+    #[test]
+    fn application_name_length_uses_utf16_code_units() {
+        let application_name = "MSSQL \u{1f980}".to_string();
+
+        assert_eq!(utf16_code_units(&application_name), 8);
+        assert_eq!(application_name.len_bytes(), 16);
```
David-Engel: This test guards `String::len_bytes` (and therefore `calculate_login_record_length`), but not the `write_metadata(...)` arguments that actually emit the `cchXxx` fields and advance `content_next_offset` — which is the half the original finding was about.

I confirmed this by mutation: reverting just `write_app_name` back to

```rust
.write_metadata(self.model.user_input.application_name.len() as i16)
```

leaves `cargo nextest run -p mssql-tds --lib -E 'test(utf16) or test(login)'` at 59 passed, 0 failed. The malformed-packet bug is fully reintroducible with this test green, and the new e2e case uses the ASCII `"AppKeywordE2E"`, so it doesn't cover it either.

Cheapest fix using patterns already in the repo: build a `Serializer` over a `PacketWriter` and assert on the emitted bytes the way `message/prelogin.rs:480-489` does, or give `tests/test_redirection.rs:358-370` a non-ASCII server/instance name — `mssql-mock-tds` already parses LOGIN7 `ServerName` by offset (`protocol.rs:186-229`), so a bad `cch` value there shows up as a wrong parse.

mssql-odbc/src/connection/connection_string_parser.rs
David-Engel, line 448: Nit / question: `Server=h;APP=;...` stores `Some("")` here, and `apply_connection_params` then replaces the `TDSX Rust Client` default with the empty string, so the server reports a blank `APP_NAME()` instead of the driver default.

That matches how `ServerSPN` and `Database` behave in this parser, so it may be intended — but msodbcsql is the contract rather than internal consistency. Worth confirming what `KEY_APP` does with an empty value in `sqlcconn.cpp` and noting the answer in `connection_string_parser.md`.

David-Engel (commented):
## Summary

Reviewed the full `cf6139bd..936429cf` diff plus the surrounding LOGIN7 serializer. The `APP` wiring is correct and consistently documented, and the UTF-16 length fix in `936429cf` is the right correction to the Copilot finding it answers. No blocking issues. Two suggestions and one nit, the first backed by a local mutation run.

**Blocking:** none. **Suggestions:** 2. **Nits:** 1.

## What I verified

**`APP` end to end.** `"app"` moves out of `KNOWN_IGNORED_KEYS` into `MAPPED_KEYS` → `ConnectionParams::application_name` → `apply_connection_params` → `ClientContext::application_name` (default `"TDSX Rust Client"`, `client_context.rs:475`). Warning behavior is unchanged in both directions: recognized-ignored and mapped keys are equally silent, so no `01S00` appears or disappears. `fmt_as_odbc_conn_str` quotes the value through `quote_odbc_value` like the other pass-throughs. The docs mapping table, the pass-through list, and the "Intentional non-goals" section are all consistent with the code, and the 128-character `cLimitSize` divergence is called out rather than silently dropped.

**The UTF-16 fix is complete and internally consistent.** Every login field written with `write_string_unicode_async` now declares its length in UTF-16 code units, which is exactly the number of 2-byte units the payload writer emits, and `String::len_bytes` feeds `calculate_login_record_length` from the same count — so the declared record length, the `cchXxx` fields, and `content_next_offset` all move together. `.len()` no longer appears on any string login field; the remaining occurrences are `sspi_token` and `access_token_bytes`, which are byte counts and correct as-is. This also matches SqlClient, which writes `string.Length` (UTF-16 code units, so an astral character counts 2) — consistent with the new test's expectation of 8 for `"MSSQL 🦀"`.

**Baseline and CI.** `cargo nextest run -p mssql-tds --lib -E 'test(login)'` → 49 passed (`CARGO_TARGET_DIR` pinned to a scratch dir, dedicated worktree). All required checks are green on this SHA, including the five-platform ADO validation, Kerberos, and the cross-repo mssql-python suite that motivated the change; the coverage bot reports 100% diff coverage.

## Suggestions

**1. The regression test guards `len_bytes`, not the nine `write_metadata` call sites — and the call sites are what the original finding was about.** `application_name_length_uses_utf16_code_units` only exercises `utf16_code_units` and `String::len_bytes`, both of which feed `calculate_login_record_length`. The `cchXxx` fields and `content_next_offset` come from the separate `write_metadata(...)` arguments in `write_hostname` / `write_username` / `write_password` / `write_app_name` / `write_server_name` / `write_library_name` / `write_language` / `write_database` / `write_attach_db_file` / `write_change_password`, and none of them is covered. Verified by mutation: reverting `write_app_name` to `self.model.user_input.application_name.len() as i16` and running `cargo nextest run -p mssql-tds --lib -E 'test(utf16) or test(login)'` gives **59 passed, 0 failed** — the bug the PR fixes is fully reintroducible with the suite green. (File restored; worktree clean.) The new e2e case uses the ASCII `"AppKeywordE2E"`, so it doesn't close the gap either.

Two existing patterns make this cheap to cover without a live server:
- `message/prelogin.rs:480-489` builds a `Serializer` over a `PacketWriter` in a unit test and asserts on the produced bytes — the same shape works for `Serializer::serialize` here, asserting the `cchAppName` field and the following offsets for a non-ASCII name.
- `mssql-mock-tds` already parses LOGIN7 `ServerName` by offset (`protocol.rs:186-229`), and `tests/test_redirection.rs:358-370` asserts on the parsed value. A non-ASCII server or instance name there would have caught the metadata/payload mismatch end to end.

**2. Name the LOGIN7 wire-format fix in the PR description.** The description and Testing section cover only the `APP` mapping, but `936429cf` changes the length encoding of *every* Unicode LOGIN7 field — hostname, username, password, app name, server name, library, language, database, attachdbfilename, changepassword — plus the declared login record length. That is a protocol-level behavior change affecting all callers (JS, Python, CLI), not just the ODBC path, and today it is discoverable only from the commit subject and a reply on a resolved bot thread. The repo review skill calls this out specifically: a behavior fix carried alongside another change should be named in the description. Adding the new `mssql-tds` test to the Testing list would help too.

## Nit

**Empty `APP=` now sends a zero-length app name.** `ConnAttrKey::App => params.application_name = Some(value.to_string())` stores `""` for `Server=h;APP=;...`, and `apply_connection_params` then overwrites the `TDSX Rust Client` default with the empty string, so the server sees a blank `APP_NAME()` rather than the driver default. That is consistent with how the other pass-throughs (`ServerSPN`, `Database`) behave here, so it may well be deliberate — but msodbcsql's `KEY_APP` handling is the contract, not internal consistency. Worth a quick confirmation against `sqlcconn.cpp`, and a one-line note in the docs either way.

## Considered and not filed

- `utf16_code_units(...) as i16` truncates, and `write_metadata` computes `(char_length * 2) as i32` in `i16` arithmetic, so a login string longer than ~16K characters overflows (debug panic) or wraps into a bogus offset (release). This PR does newly expose `application_name` to unvalidated connection-string input while declining msodbcsql's 128-character cap, but the identical shape already exists on `main` for `UID` and `Database`, which are equally connection-string reachable — so it is a pre-existing, crate-wide gap and fixing it for one field only would be the worse shape. Better as its own change.
- `ClientContext::len_bytes` omits `change_password` from the record-length sum while `write_change_password` writes it. Currently harmless because nothing ever assigns a non-empty `change_password`, and it is pre-existing.

---
Posted by an unattended review run — these findings were not checked by a human before posting. Comment only: nothing approved, requested, labeled, or resolved.

mssql-tds/src/message/login.rs
David-Engel: Fixed in dc310935. The regression test now invokes `write_app_name` directly and asserts the emitted LOGIN7 offset, UTF-16 code-unit length, and resulting `content_next_offset`, so reverting that metadata call site breaks the test.

> [!NOTE]

mssql-odbc/src/connection/connection_string_parser.rs
David-Engel: Keeping the current behavior: an explicitly supplied empty `APP=` is distinct from omitting `APP` and intentionally overrides the default with an empty application name, consistent with other explicit pass-through values. No code or documentation change is needed for this edge case.

> [!NOTE]

mssql-tds/src/message/login.rs
```diff
@@ -1306,6 +1308,10 @@ trait SizedLoginItem {
     fn len_bytes(&self) -> i32;
 }
 
+fn utf16_code_units(value: &str) -> usize {
```
shiwanigupta0809: **The length arithmetic downstream of this helper is `i16`, and this PR is what makes it reachable from an uncapped connection-string value.**

`utf16_code_units` returns `usize`, but every call site casts to `i16` and `write_metadata` then does:

```rust
async fn write_metadata(&mut self, char_length: i16) -> TdsResult<bool> {
    ...
    self.payload_writer.write_i16_async(char_length).await?;
    self.content_next_offset += (char_length * 2) as i32;
```

`char_length * 2` is **`i16` multiplication**, widened to `i32` only afterwards. Three separate breakages, in increasing order of reachability:

1. `char_length > 16383` → the multiply overflows `i16`. Debug and `cargo nextest` builds **panic** ("attempt to multiply with overflow") — inside an `extern "C"` ODBC entry point. `[profile.release]` sets `lto`/`strip` but not `overflow-checks`, so release instead wraps to a negative and `content_next_offset` walks *backwards*, producing a LOGIN7 with overlapping field offsets that the driver still sends.
2. `char_length > 32767` → the `as i16` cast itself is lossy before the multiply ever runs. MS-TDS 2.2.6.4 declares `cchAppName` as **`USHORT`**, so the correct type here is `u16`, not `i16`.
3. MS-TDS 2.2.6.4 also states: *"The length of a LOGIN7 stream MUST NOT be longer than 128K-1(byte) bytes."* Nothing in `calculate_login_record_length` checks that, and it is now a caller-controlled MUST.

On reachability — all three are pre-existing for `Database`, `UID` and `PWD`, so this is not a regression. But the PR description says:

> msodbcsql additionally rejects an `APP` longer than 128 characters (`x_rgLookup[KEY_APP].cLimitSize`), but per-key length caps are an intentional non-goal of this parser

That reasoning treats the cap as an input-validation nicety. It is also the only thing that kept a connection string from reaching this arithmetic, and the parser explicitly has no length limits of its own — so declining it here leaves nothing between `APP=<20000 chars>` and a panic or a malformed login packet. Worth deciding deliberately rather than by inheritance.

Either resolution is fine by me:

- **Fix it at the protocol layer** (my preference, since it covers `Database`/`UID`/`PWD` too): widen `write_metadata` to `u16`, compute the offset advance in `i32` (`i32::from(char_length) * 2`), and return a `TdsResult::Err` when a field exceeds `u16::MAX` code units or the assembled record exceeds `128 * 1024 - 1`. A clear error beats a wrapped offset.
- **Or keep msodbcsql's 128 for `APP`** and note in the docs that this one is a protocol-adjacent bound rather than a validation preference, so it does not contradict the non-goal.

What I would avoid is landing the mapping with neither.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
```diff
@@ -551,6 +551,56 @@ TEST_F(DriverConnectLiveTest, NewConnectionAttributesParity) {
     }
 }
 
+// APP must reach the server in the TDS login packet, which APP_NAME() reports.
+// Regression coverage for AB#47534, where APP parsed cleanly but was dropped
+// before the login packet, so the server saw the driver's default name instead.
+TEST_F(DriverConnectLiveTest, AppKeywordReachesServerAppName) {
+    auto& cfg = ODBCTestConfig::Instance();
+    if (!cfg.HasSqlAuth()) {
```
shiwanigupta0809: **Suggestion:** this guard can make the regression test a no-op in the very configurations CI is most likely to use.

`HasSqlAuth()` is `!server_.empty() && !uid_.empty() && !pwd_.empty()`. A suite configured through `ODBC_TEST_CONNSTR`, a DSN, or integrated auth leaves `uid_`/`pwd_` empty and this silently `GTEST_SKIP`s. Since the whole point is a regression guard for AB#47534 — a bug that shipped precisely because nothing surfaced it — a guard that can skip without anyone noticing is worth avoiding.

The existing `HasSqlAuth()` sites at 309/416/516/613 need it for a documented reason: the fixture header says they "assemble their own strings from `Server()`/`Uid()`/`Pwd()` and corrupt individual auth tokens, so they require all three." This test does not corrupt anything — it only needs to append `;APP=<name>` to whatever connection the suite already has.

`ODBCTest::BuildConnectionString()` already exists and honours every configured mode. Building on it and appending `;APP=` drops the gate entirely and fixes a second problem in passing: the hand-built string here also ignores `cfg.Encrypt()` and `cfg.Database()`, so on a server that requires `Encrypt=Strict` this test fails to connect rather than skipping cleanly.

Minor: "see follow-up issue for capability gating" names no issue. The fixture header calls it "tracked separately" — quoting the actual number, or dropping the clause with the gate, keeps the message actionable.

mssql-tds/src/message/login.rs
```diff
@@ -1459,6 +1468,34 @@ mod tests {
         FeaturesRequest { features }
     }
 
+    #[test]
+    fn application_name_metadata_uses_utf16_code_units() {
```
shiwanigupta0809, line 1492: **Suggestion:** the mutation argument from the earlier thread applies to nine more call sites than the one it was raised on.

The fix for that thread was right and I re-derived it — this test now drives `write_app_name` directly and asserts the emitted offset, the `cchAppName` value and the resulting `content_next_offset`, so reverting *that* call site fails. But `write_hostname`, `write_user_name`, `write_password`, `write_server_name`, `write_library_name`, `write_language`, `write_database`, `write_attach_db_file` and `write_change_password` all changed identically in this PR and none is covered. Reverting any of them to `.len() as i16` leaves the suite green, which is the same gap the thread identified, just at the other nine sites.

This is worth closing rather than noting, for two reasons. First, `APP` is the *least* likely of the ten to actually be non-ASCII in the field — `Database`, `UID` and `PWD` are far more commonly non-ASCII, so the covered site is the one that matters least. Second, these are ten copies of one expression, which is exactly the shape where a future refactor reverts a subset.

A table-driven version of this test is cheap: parameterise over `(field setter, writer method, expected code units)` and cover `database` with a BMP non-ASCII value (`"café"` — 4 UTF-16 units, 5 UTF-8 bytes) alongside the existing astral `application_name` case. Two rows would already guard the class rather than one instance.

mssql-tds/src/message/login.rs
```diff
@@ -1084,7 +1084,7 @@ impl<'a, 'n, 'context> Serializer<'a, 'n, 'context> {
     async fn write_password(&mut self) -> TdsResult<()> {
         if self.model.user_input.tds_authentication_method == TdsAuthenticationMethod::Password {
             if self
-                .write_metadata(self.model.user_input.password.len() as i16)
+                .write_metadata(utf16_code_units(&self.model.user_input.password) as i16)
```
shiwanigupta0809: **Suggestion:** this line is a user-visible connectivity fix, and neither the title nor the summary says so.

Before this change `write_password` declared `cchPassword` as the UTF-8 byte length while writing a UTF-16LE payload, and `content_next_offset` advanced by twice the UTF-8 length. For a password like `café` that is 5 declared characters against 4 actually written, so `cchPassword` is wrong **and** every subsequent field offset — `AppName`, `ServerName`, `CltIntName`, `Language`, `Database`, `FeatureExt` — is shifted two bytes past its real data. The same held for `user_name`, `database`, `workstation_id` and `server_name`, and `String::len_bytes` over-declared the total record length on top of it.

In other words: **logging in with a non-ASCII password, username, database name or hostname was producing a malformed LOGIN7 before this PR**, on every mssql-tds consumer — `mssql-odbc`, `mssql-js`, `mssql-py-core`, the CLI. That is a bug a customer would hit on day one in a non-English deployment and would search for by exactly those words.

The PR describes it as "While adding non-ASCII `APP` support, this change also corrects the shared LOGIN7 serializer to measure every Unicode field in UTF-16 code units ... keeps each field's declared character length, subsequent offsets, and the total login-record length consistent". Accurate, but it reads as internal tidying subordinate to the `APP` work, and it is filed under a title about a connection-string keyword.

Ask: give it its own line in the summary stating the user-facing effect, and add a `CHANGELOG.md` entry. It is also worth its own work item — the `APP` mapping is AB#47534, but this fixes something with a different (and larger) blast radius, and linking them together means the login fix is not findable on its own.

shiwanigupta0809 (commented):
Reviewed the mapping against the parser's existing pass-through keys, and the LOGIN7 change against MS-TDS 2.2.6.4 and every remaining length site.

Two things I verified and want to note as correct, because they are the ones most likely to be wrong in a change of this shape:

- The UTF-16 conversion is **complete**. `SizedLoginItem for TransportContext` delegates to `String::len_bytes`, and `calculate_byte_length_for_authentication` uses `len_bytes` for both `user_name` and `password`, so `calculate_login_record_length` and the per-field `write_metadata` calls now agree for every field. I went looking for a straggler `.len()` and there is none.
- The astral test value is a good choice — `"MSSQL \u{1f980}"` is 7 `chars` but 8 UTF-16 units and 12 UTF-8 bytes, so all three counts differ and none of the three plausible wrong implementations passes.

One finding I'd like resolved before merge: this PR routes uncapped user input into a LOGIN7 serializer whose length arithmetic is `i16`, and the description explicitly declines the only cap that was standing in front of it. Details inline, with the MS-TDS citation.

**Blocking:** one (LOGIN7 length arithmetic vs. an uncapped `APP`).
**Suggestion:** three (the e2e guard skips in CI; mutation coverage on the other nine call sites; the LOGIN7 fix is under-billed).
**Nit:** none.

Not re-filing the three resolved threads (UTF-16 metadata, `write_app_name` mutation coverage, empty `APP=`); two of the suggestions below build on the latter two rather than repeating them.

mssql-odbc/tests/e2e/tests/driver_connect_test.cpp
David-Engel: Fixed in 3f6f02ac. The test now starts with `BuildConnectionString()` and appends `APP`, so it covers the suite's SQL auth, integrated auth, DSN, encryption, database, and full connection-string configurations without the `HasSqlAuth()` skip.

> [!NOTE]

mssql-tds/src/message/login.rs
David-Engel: Updated in 3f6f02ac. The PR title and opening summary now state the user-visible non-ASCII login fix, and `CHANGELOG.md` calls out affected usernames, passwords, database names, hostnames, and application names across all bindings. I kept the existing APP work item rather than creating an unrelated tracker from this review thread.

> [!NOTE]

mssql-tds/src/message/login.rs
David-Engel: Fixed in 3f6f02ac at the protocol layer. LOGIN7 metadata now uses `u16`, offset arithmetic widens before multiplication, field offsets are checked, strings beyond `u16::MAX` code units are rejected, and the complete record is capped at 128 KiB minus one byte with regression coverage for both limits.

> [!NOTE]

mssql-tds/src/message/login.rs
David-Engel: Expanded coverage in 3f6f02ac with a BMP non-ASCII database case alongside the astral application-name case. Both exercise their field writer and assert the emitted UTF-16 length plus offset advancement, covering the shared conversion across distinct LOGIN7 metadata call sites.

> [!NOTE]

David-Engel (commented):
Posted by an unattended review run — these findings were not checked by a human before posting. Comment only: nothing approved, requested, labeled, or resolved.

## Summary

Reviewed the full `cf6139bd..3f6f02ac` diff plus the surrounding LOGIN7 serializer and the ODBC connection-string path. The UTF-16 length fix is correct and complete, the `u16` metadata widening and the two new bounds checks are the right resolution of the previous round's blocking thread, and the `APP` mapping is wired and documented consistently. One finding: the new offset validation is applied to zero-length fields as well, which turns the long-SSPI (`cbSSPILong`) login path into a hard failure.

**Blocking:** none. **Medium:** 1. **Nit:** 1.

## Verification

**UTF-16 accounting is internally consistent.** Every `write_metadata` argument, `String::len_bytes`, and the deferred payload writer (`write_string_unicode_async` / `encode_utf16().flat_map(to_le_bytes)`) now count the same UTF-16 code units, and the `deferred_actions_indicator` push order matches the order `content_next_offset` advances, so the declared record length, each `cchXxx`, and every `ibXxx` move together. `TransportContext::len_bytes` delegates to `String::len_bytes`, and `calculate_byte_length_for_authentication` covers `user_name`/`password`, so `calculate_login_record_length` agrees with the per-field writes.

**The new tests are not vacuous.** `"MSSQL \u{1f980}"` is 8 UTF-16 units against 12 UTF-8 bytes and `"caf\u{e9}"` is 4 against 5, and both tests assert the emitted offset, the emitted length, and the resulting `content_next_offset` — so reverting either call site to `.len()` fails them. Eight of the ten `write_metadata` call sites are still uncovered; noting that rather than re-filing, since the earlier thread on this was answered by `dc310935`/`3f6f02ac`.

**Bounds.** `MAX_LOGIN_RECORD_LENGTH = 128 * 1024 - 1` matches MS-TDS 2.2.6.4 ("The length of a LOGIN7 stream MUST NOT be longer than 128K-1(byte) bytes"), and `calculate_login_record_length` runs before any byte is written, so an oversized record fails before touching the socket. `write_sspi_short`'s `sspi_len as u16` is guarded by the `<= 0xFFFF` branch, and `feature_extension_offset` stays a DWORD, so neither is affected by the `u16` narrowing.

**ODBC `APP`.** `"app"` moves out of `KNOWN_IGNORED_KEYS` into `MAPPED_KEYS` → `ConnectionParams::application_name` → `apply_connection_params` → `ClientContext::application_name` (default `"TDSX Rust Client"`). Warning behaviour is unchanged in both directions, `fmt_as_odbc_conn_str` quotes through `quote_odbc_value` like the other pass-throughs, and no stale `APP` entry is left in the ignored-key list or in the docs. The e2e test now builds on `ODBCTestUtils::BuildConnectionString()`, so it no longer skips under DSN, integrated-auth or `ODBC_TEST_CONNSTR` configurations, and all three helpers it uses exist in `odbc_test_fixture.h`.

**Tests and CI.** `cargo nextest run -p mssql-tds --lib` over the new login tests passes locally (dedicated worktree, scratch `CARGO_TARGET_DIR`). All checks are green on `3f6f02ac`, including the five-platform ADO validation, the Kerberos job, and the cross-repo mssql-python suite; the coverage bot reports 100% diff coverage.

**Not re-filed.** The resolved UTF-16, mutation-coverage and empty-`APP=` threads, and the `ClientContext::len_bytes` omission of `change_password` that a previous round already recorded as considered-and-not-filed.

## Findings

### Medium

**A LOGIN7 with an SSPI token larger than ~64 KB now fails instead of connecting, and `cbSSPILong` becomes unreachable.** `write_metadata` evaluates `let offset = self.current_offset()?;` *before* the `char_length == 0` early return, so an empty field still requires `content_next_offset <= u16::MAX`. `write_attach_db_file` and `write_change_password` run after `write_sspi_short` has advanced the cursor by the full token length, so any token that pushes it past 65535 aborts the login — even though the record is comfortably inside the new 128 KiB − 1 cap and both of those fields are empty.

That makes the `sspi_len > 0xFFFF` branch of `write_sspi_short` and all of `write_cb_sspi_long` dead code: the serializer deliberately supports the long form, but no login that needs it can be produced any more. Before this PR the same case wrapped `content_next_offset as i16` into a bogus `ibAttachDbFile`, which the server ignores because `cchAttachDbFile` is 0, so the login went through. Tokens that large are rare — a Kerberos AP-REQ carrying a large PAC — but that is precisely the case `cbSSPILong` exists for.

Verified on `3f6f02ac` with a temporary probe test (worktree restored afterwards): a 70,000-byte `sspi_token` for which `calculate_login_record_length()` returns `Ok` yields

```
Protocol Error: LOGIN7 field offset 70094 exceeds the maximum of 65535 bytes
```

from the *empty* `attach_db_file` field.

Suggested shape: keep the strict check for fields that carry data and move `current_offset()?` below the `char_length == 0` early return, writing the truncated offset (or `0`) for empty fields — MS-TDS lets the server ignore `ibXxx` when `cchXxx` is 0, which is what the reference drivers rely on. The `sspi_len == 0` arm of `write_sspi_short` has the same shape and deserves the same treatment.

### Nit

**`write_feature_ext` is now mixed-signedness.** The offset goes out through `write_u16_async(self.current_offset()?)` while the constant length immediately below is still `write_i16_async(4)`. Identical bytes on the wire, but `write_u16_async(4)` would match the field's unsigned type and the rest of the function after this change.

David-Engel (commented):

## Summary

Reviewed the full `cf6139bd..f62e8203` diff plus the surrounding LOGIN7 serializer, the ODBC connection-string path, and the mock server's LOGIN7 parser. The UTF-16 length fix is correct and complete, and `f62e8203` is the right resolution of the previous round's long-SSPI finding — I confirmed by mutation that the new test actually guards it. The `APP` mapping is wired and documented consistently. No blocking issues: two suggestions and two nits.

**Blocking:** none. **Suggestions:** 2. **Nits:** 2.

## Verification

**The long-SSPI fix works and its test is not vacuous.** Moving `current_offset()?` below the `char_length == 0` early return (and writing `0` for the empty `ibXxx`) is what makes `write_attach_db_file` / `write_change_password` survive a `content_next_offset` past `u16::MAX`. Mutation-checked on this SHA: restoring the old ordering (`let offset = self.current_offset()?;` above the early return, writing `offset` for empty fields) makes `empty_fields_do_not_reject_long_sspi_offset` fail and everything else pass. Worktree restored, `git status` clean. MS-TDS 2.2.6.4 lets the server ignore `ibXxx` when `cchXxx` is 0, so the zero offset is legal, and the ADO validation plus the Kerberos job are green on this SHA.

**UTF-16 accounting is still internally consistent.** Every `write_metadata` argument, `String::len_bytes`, and the deferred payload writers (`write_string_unicode_async`, `encode_utf16().flat_map(to_le_bytes)`) count the same UTF-16 code units; `TransportContext::len_bytes` delegates to `String::len_bytes` and `calculate_byte_length_for_authentication` covers `user_name`/`password`, so the declared `Length`, each `cchXxx`, and every `ibXxx` move together. `deferred_actions_indicator` push order still matches the order `content_next_offset` advances. `MAX_LOGIN_RECORD_LENGTH = 128 * 1024 - 1` matches the spec's "MUST NOT be longer than 128K-1(byte) bytes", and `calculate_login_record_length` runs before any byte is written, so an oversized record fails before the socket. Error paths inside `write_variable_length_section` abort within the first ~150 payload bytes, well under any legal `max_payload_size`, so no partial LOGIN7 can reach the wire — dropping the writer is sufficient.

**Ordering constraint holds.** `write_feature_ext` and `write_sspi_short` are the only remaining unconditional `current_offset()?` callers, and both run before the SSPI token advances the cursor, so neither can trip the `u16` bound on a legal record. `mssql-mock-tds`'s LOGIN7 parser (`protocol.rs:186-229`) gates on `server_name_length > 0`, so it is unaffected by the new zero offsets.

**ODBC `APP`.** `"app"` moves out of `KNOWN_IGNORED_KEYS` into `MAPPED_KEYS` → `ConnectionParams::application_name` → `apply_connection_params` → `ClientContext::application_name`. Warning behavior is unchanged in both directions, `fmt_as_odbc_conn_str` quotes through `quote_odbc_value` like the other pass-throughs, and the docs table, pass-through list and non-goals section all match the code.

**Tests and CI.** `cargo nextest run -p mssql-tds --lib -E 'test(login) or test(utf16) or test(sspi)'` → 79 passed (dedicated worktree, scratch `CARGO_TARGET_DIR`). All required checks pass on `f62e8203`, including the five-platform ADO validation, Kerberos, and the cross-repo mssql-python suite.

**Not re-filed.** The resolved UTF-16 metadata, `write_app_name` mutation-coverage, empty-`APP=`, `i16` arithmetic, and e2e `HasSqlAuth()` threads, and the mixed-signedness nit on `write_feature_ext` (fixed in `f62e8203`).

## Findings

### Suggestions

**1. `empty_fields_do_not_reject_long_sspi_offset` skips the SSPI slot it exists to protect, and `write_cb_sspi_long` has no test at all.** `mssql-tds/src/message/login.rs:1594` seeks to `PACKET_HEADER_SIZE + 4`, stepping straight over `ibSSPI`/`cbSSPI`, so the four asserts that follow only cover the two empty fields. The long form's actual wire encoding — `ibSSPI == 94` and the `0xFFFF` marker in `cbSSPI` (`login.rs:1218`) — is never asserted, and `write_cb_sspi_long` is referenced by no test in the crate. That is the branch the commit subject ("Preserve long SSPI LOGIN7 serialization") is about: today, mutating `write_u16_async(0xFFFF)` to `sspi_len as u16` or dropping the `sspi_len > 0xFFFF` arm of `write_cb_sspi_long` leaves the suite green.

Cheap fix inside the test you already have — read the two `u16`s you currently skip, and call the DWORD writer:

```rust
let mut payload = packet_writer.get_payload();
payload.set_position(PacketWriter::PACKET_HEADER_SIZE as u64);
assert_eq!(payload.read_u16::<LittleEndian>().unwrap(), FIXED_LOGIN_RECORD_LENGTH as u16);
assert_eq!(payload.read_u16::<LittleEndian>().unwrap(), 0xFFFF);
// ... existing four zero asserts for attach_db_file / change_password ...
```

plus a `block_on(serializer.write_cb_sspi_long()).unwrap()` before reading the payload and an assert that the trailing `u32` is `70_000`. Three extra asserts turn the test from "the empty fields don't error" into "the long form serializes correctly", which is the property the commit claims.

**2. `SizedLoginItem for ClientContext` still omits `change_password` while `write_change_password` writes it.** `login.rs:1359-1373` sums `workstation_id`, `application_name`, `library_name`, `language`, `database`, `attach_db_file` and the auth fields, but not `change_password` — so a non-empty `ClientContext::change_password` (a `pub` field, `client_context.rs:224`) makes the declared LOGIN7 `Length` short by `2 * code_units` while the bytes are still emitted, and `login_options.rs:432` already sets the change-password option bit from it. Nothing assigns it today, and it is pre-existing, so this is not a regression — but this PR is exactly the one that makes the declared record length authoritative (`MAX_LOGIN_RECORD_LENGTH`) and is already editing that `impl`, so folding in `client_context_item_length += self.change_password.len_bytes();` is a one-line close rather than leaving a known inconsistency behind a bound that now has teeth. A previous round recorded this as considered-and-not-filed; raising it once more only because the surrounding contract changed.

### Nits

**3. Stale comment in `write_sspi_short` (`login.rs:1202`).** `// No SSPI token - write offset and zero length` — as of `f62e8203` no offset is written; both fields are literal `0`. Worth replacing with the *reason*, since that is the non-obvious part and the next reader's instinct will be to "restore" the running offset: something like `// Empty field: ibSSPI is ignored when cbSSPI is 0 (MS-TDS 2.2.6.4), so emit 0 rather than an offset that may exceed u16 after a long token.` The same one-liner would earn its place above the `char_length == 0` arm of `write_metadata`.

**4. The e2e `APP` append is order-sensitive under `ODBC_TEST_CONNSTR`.** `driver_connect_test.cpp:556-557` appends `";APP=" + kAppName` to `BuildConnectionString()`, which already ends in `;` in every non-override branch (`odbc_test_utils.cpp:90-110`) — harmless, since Phase 1 of the parser skips runs of `;`, so no `01S00`. The order does matter though: recognized keys are **first-wins** (`connection_string_parser.rs:613`), so if a suite's `ODBC_TEST_CONNSTR` already carries an `APP=`, the appended one is discarded and this test fails with a confusing `APP_NAME()` mismatch rather than a clear message. Prepending instead of appending (`ToSqlTStr("APP=" + kAppName + ";") + connstr`) would make the test's value win in every configuration.

mssql-tds/src/message/login.rs
```diff
@@ -1344,6 +1369,7 @@ impl SizedLoginItem for ClientContext {
         client_context_item_length += self.language.len_bytes();
         client_context_item_length += self.database.len_bytes();
         client_context_item_length += self.attach_db_file.len_bytes();
+        client_context_item_length += self.change_password.len_bytes();
```
David-Engel, line 1372: Good catch folding this in â€” it closes the case where a caller-set `change_password` (a `pub` field on `ClientContext`, `client_context.rs:224`, and `login_options.rs:432` already sets the change-password option bit from it) made the declared LOGIN7 `Length` short by `2 * code_units` while the bytes were still emitted. That matters more now that `MAX_LOGIN_RECORD_LENGTH` gives the declared length teeth.

The fix has no test, though. I deleted this line on this SHA and all 38 `message::login*` tests still passed; the full `cargo nextest run -p mssql-tds --lib` also stayed green apart from the seven pre-existing `certificate_validator` / `win_tls::validate` fixture failures that fail on a clean tree too. `login_rejects_oversized_string_and_record` exercises `application_name` + `database`, so it cannot see this field.

A direct assert is a few lines:

```rust
#[test]
fn record_length_accounts_for_change_password() {
    let base = ClientContext {
        connect_retry_count: 0,
        ..ClientContext::default()
    };
    let with_change_password = ClientContext {
        connect_retry_count: 0,
        change_password: "abcd".to_string(),
        ..ClientContext::default()
    };
    assert_eq!(with_change_password.len_bytes(), base.len_bytes() + 8);
}
```

Worth naming in the PR description too â€” it is a behavior fix that currently rides along unmentioned under the Unicode-lengths story.

CHANGELOG.md
```diff
@@ -55,6 +55,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
 
 ### Changed
 
+- `mssql-tds`: LOGIN7 now encodes Unicode field lengths as UTF-16 code units
+  and rejects oversized records instead of producing malformed packets. This
+  fixes login failures with non-ASCII usernames, passwords, database names,
```
David-Engel, line 60: This entry covers the `mssql-tds` half of the PR, but the ODBC `APP` mapping â€” the AB#47534 fix, and the half a user actually notices â€” gets no entry at all. There is a `### Fixed` section at line 116 that fits it:

```markdown
- `mssql-odbc`: the `APP` connection-string keyword is now sent as the TDS login
  application name. It was recognized but ignored, so `APP_NAME()` reported the
  mssql-tds default `TDSX Rust Client`; because the key was recognized, no
  `01S00` warning surfaced the drop either.
```

Same section would be the right home for the `change_password` record-length fix flagged on `login.rs:1372`. Not merge-blocking â€” the changelog is in no checklist here â€” but this PR already writes one entry, so the omission looks like an oversight rather than a choice.

mssql-tds/src/message/login.rs
```diff
@@ -1303,28 +1319,37 @@ enum LoginDeferredPayload {
 
 /// Trait to calculate the size of the fields for the login records in bytes.
 trait SizedLoginItem {
-    fn len_bytes(&self) -> i32;
+    fn len_bytes(&self) -> usize;
+}
+
+fn utf16_code_units(value: &str) -> TdsResult<u16> {
+    u16::try_from(value.encode_utf16().count()).map_err(|_| {
+        crate::error::Error::ProtocolError(format!(
```
David-Engel: `Error::ProtocolError` renders as `Protocol Error: LOGIN7 string exceeds the maximum of 65535 UTF-16 code units`, which reads like the server sent something malformed. This check, `current_offset`, and the `MAX_LOGIN_RECORD_LENGTH` check all fire on caller-supplied input before a byte reaches the socket, so `Error::UsageError` â€” used for exactly this kind of input rejection in `mssql-odbc/src/connection/odbc_authentication_validator.rs:36` â€” describes it better.

Nothing downstream changes: `post_tds_error` takes the SQLSTATE from the call site rather than the variant, so this is purely about the message an mssql-python or mssql-js caller reads. Fine to leave as-is if you would rather not churn it.

David-Engel (commented):

Posted by an unattended run — these findings were not checked by a human before posting. Comment only: nothing approved, requested, labeled, or resolved.

## Summary

Re-reviewed the full `cf6139bd..dec41caf` diff plus the surrounding LOGIN7 serializer, the ODBC connection-string path, and the mock server's LOGIN7 parser. `dec41caf` closes all four items from the previous round: the long-SSPI test now asserts `ibSSPI`/`cbSSPI` and exercises `write_cb_sspi_long`, `change_password` is folded into `ClientContext::len_bytes`, the stale `write_sspi_short` comment is replaced with the reason, and the e2e test prepends `APP=` so its value wins under first-wins parsing. No blocking issues — one suggestion and two nits, all small.

**Blocking:** none. **Suggestions:** 1. **Nits:** 2.

## Verification

**The new long-SSPI assertions are not vacuous.** Mutation-checked on this SHA in a detached worktree with a scratch `CARGO_TARGET_DIR`: changing `write_u16_async(0xFFFF)` to `write_u16_async(sspi_len as u16)` (`login.rs:1219`) makes `empty_fields_do_not_reject_long_sspi_offset` fail and leaves the other 37 `message::login*` tests green. So the `0xFFFF` marker, the `ibSSPI` value and the `cbSSPILong` DWORD are all genuinely pinned now, which is the property `dec41caf`'s subject claims. Worktree restored afterwards, `git status` clean.

**Length accounting still balances.** `calculate_login_record_length` = 94 + `ClientContext::len_bytes` (workstation_id, application_name, library_name, language, database, attach_db_file, change_password, plus user_name/password under `Password` auth) + server name + 4 for the FeatureExt DWORD + `sspi_len` + `FeaturesRequest::len_bytes`. That matches exactly what `write_variable_length_section` emits and the order `deferred_actions_indicator` replays it in. `write_username` / `write_password` skip both the metadata and the payload off the `Password` path and `calculate_byte_length_for_authentication` gates on the same condition, so the declared `Length`, every `cchXxx`, and every `ibXxx` move together.

**Ordering constraint still holds.** `write_feature_ext` and `write_sspi_short` remain the only unconditional `current_offset()?` callers, and both run before the SSPI token advances the cursor, so a legal record cannot trip the `u16` bound there. `mssql-mock-tds`'s LOGIN7 parser (`protocol.rs:217-223`) reads only `ibServerName`/`cchServerName` and gates on `cchServerName > 0`, so the new zero offsets for empty fields don't affect it.

**Tests and CI.** `cargo nextest run -p mssql-tds --lib -E 'test(/message::login/)'` → 38 passed on this SHA. The full `-p mssql-tds --lib` run fails only the seven pre-existing `certificate_validator` / `win_tls::validate` cases that need locally generated fixtures and fail on a clean tree too. All required checks pass on `dec41caf`, including the five-platform ADO validation, the Kerberos job, and the cross-repo mssql-python suite.

**Not re-filed.** The four items from the `f62e8203` round, all fixed here; and the earlier resolved threads on UTF-16 metadata, `write_app_name` mutation coverage, empty `APP=`, `i16` arithmetic, and the e2e `HasSqlAuth()` question.

## Findings

Three inline, plus one that no diff hunk covers:

**Nit — leftover mixed signedness in the two auth fields.** `write_username` (`login.rs:1084-1085`) and `write_password` (`login.rs:1101-1102`) still emit their "field absent" pair with `write_i16_async(0)` while every other LOGIN7 metadata site in this file moved to `write_u16_async`. The bytes are identical so nothing is wrong on the wire, but it is the same mixed signedness you cleaned up in `write_feature_ext` last round, and it is two lines.

mssql-tds/src/message/login.rs
David-Engel: Added `record_length_accounts_for_change_password`, which pins the UTF-16 byte contribution independently of the oversized-record test. The PR description and changelog now also call out this behavior fix.

> [!NOTE]

mssql-tds/src/message/login.rs
David-Engel: Changed the field-length, offset, and total-record bound failures to `UsageError`, and extended the oversized-input test to assert that classification. I also changed the remaining absent username/password metadata writes to unsigned values for consistency.

> [!NOTE]

CHANGELOG.md
David-Engel: Added separate `Fixed` entries for the ODBC `APP` mapping and the change-password LOGIN7 sizing correction.

> [!NOTE]

David-Engel (commented):

Posted by an unattended run — these findings were not checked by a human before posting. Comment only: nothing approved, requested, or resolved.

## Summary

Re-reviewed the full `origin/main...94d2443b` diff plus the surrounding LOGIN7 serializer, the ODBC connection-string path, and the mock server's LOGIN7 parser. `94d2443b` closes the last round's item and the two open inline threads: the bound failures are now `UsageError` rather than `ProtocolError`, the absent username/password metadata pairs moved to `write_u16_async`, `record_length_accounts_for_change_password` pins the change-password contribution, and the `CHANGELOG` now carries separate `Fixed` entries for the ODBC `APP` mapping and the change-password sizing correction. No findings this round.

**Blocking:** none. **Suggestions:** none. **Nits:** none.

## Verification

**Length accounting balances field-for-field.** `calculate_login_record_length` = 94 + `ClientContext::len_bytes` (workstation_id, application_name, library_name, language, database, attach_db_file, change_password, plus user_name/password under `Password` auth) + server name + 4 for the FeatureExt DWORD + `sspi_len` + `FeaturesRequest::len_bytes`. That is exactly what `write_variable_length_section` emits, and the `deferred_actions_indicator` push order (HostName, UserName, Password, AppName, ServerName, FeatureExtOffset, Library, Language, Database, Sspi, AttachDbFile, ChangePassword) matches the order `content_next_offset` advances, so the declared `Length`, every `cchXxx` and every `ibXxx` move together. `write_username`/`write_password` gate on `TdsAuthenticationMethod::Password` and `calculate_byte_length_for_authentication` gates on the same condition. `TransportContext::len_bytes` delegates to `String::len_bytes`, and every `write_metadata` argument, `String::len_bytes` and the deferred payload writers (`write_string_unicode_async`, `encode_utf16().flat_map(to_le_bytes)`) count the same UTF-16 code units.

**The new test is not vacuous, and neither is the new classification assert.** Mutation-checked on this SHA in a detached worktree with a scratch `CARGO_TARGET_DIR`: deleting `client_context_item_length += self.change_password.len_bytes();` (`login.rs:1372`) makes `record_length_accounts_for_change_password` fail and leaves the other 38 `message::login*` tests green; reverting `utf16_code_units` to `Error::ProtocolError` makes `login_rejects_oversized_string_and_record` fail on the `matches!` assert it gained in this commit. Worktree restored after each, `git status` clean.

**`UsageError` is the right variant here.** All three new bounds — `utf16_code_units`, `current_offset`, `MAX_LOGIN_RECORD_LENGTH` — fire on caller-supplied input before a byte reaches the socket, which is what `UsageError` is used for elsewhere in the crate (`odbc_authentication_validator.rs:36`, `tds_value_serializer.rs`). `MAX_LOGIN_RECORD_LENGTH = 128 * 1024 - 1` matches MS-TDS 2.2.6.4 ("MUST NOT be longer than 128K-1(byte) bytes"), and the check runs before any write, so an oversized record fails before the socket rather than half-serialized.

**Ordering constraint still holds.** `write_feature_ext` and the non-empty arm of `write_sspi_short` remain the only unconditional `current_offset()?` callers, and both run before the SSPI token advances the cursor, so a legal record cannot trip the `u16` bound there. Empty fields write a literal `0` offset, which MS-TDS lets the server ignore when `cchXxx` is 0, and `mssql-mock-tds`'s LOGIN7 parser (`protocol.rs:217-223`) gates on `server_name_length > 0`, so the zero offsets don't affect it.

**ODBC `APP`.** `"app"` is out of `KNOWN_IGNORED_KEYS` and into `MAPPED_KEYS` → `ConnectionParams::application_name` → `apply_connection_params` → `ClientContext::application_name`. `fmt_as_odbc_conn_str` quotes through `quote_odbc_value` like the other pass-throughs, `Debug` carries the new field, and no stale `APP` entry is left in the ignored-key list or in `connection_string_parser.md` — the mapping table, the pass-through list and the "Intentional non-goals" section all match the code, including the `cLimitSize` divergence being named rather than dropped. The e2e case prepends `APP=` so its value wins under first-wins parsing in every configured mode.

**Tests and CI.** `cargo nextest run -p mssql-tds --lib -E 'test(/message::login/)'` → 39 passed on this SHA. All required checks pass on `94d2443b`: the five-platform ADO validation, the Kerberos job, and the cross-repo mssql-python suite. The coverage bot reports 96% diff coverage (reported, not enforced).

## Considered and not filed

- **The `current_offset` overflow branch (`login.rs:1284-1289`) has no test**, which is what the coverage bot's missing-lines list is pointing at. It is reachable only with an SSPI token above ~64 KB *and* a non-empty `attach_db_file` or `change_password`, and the behavior it produces — a clear `UsageError` instead of a wrapped offset — is correct, since `ibXxx` is a `USHORT` and a field past 65535 is unrepresentable. Six lines of error text on a path this narrow isn't worth a round trip.
- `write_server_name`'s conversion (`login.rs:1125`) is the other uncovered line. Same class as the "other nine call sites" thread that `3f6f02ac` answered with the `database` case; not re-filing.
- The resolved threads on UTF-16 metadata, `write_app_name` mutation coverage, empty `APP=`, `i16` arithmetic, the e2e `HasSqlAuth()` gate, `change_password` in `len_bytes`, the `write_sspi_short` comment, and the `CHANGELOG` omission — all fixed in `3f6f02ac`, `f62e8203`, `dec41caf` or `94d2443b`.

shiwanigupta0809 (approved):

PR #426, opened 2026-08-28
mssql-odbc: pin the fetch path's error semantics, and render GUIDs in upper case

David-Engel:
## Summary

Closes the fetch-path error-semantics gap from [AB#47678](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47678), and fixes a GUID rendering divergence the coverage found.

The P5 work proved the fetch **type map** is wired up. It did not prove the **error semantics** match msodbcsql: a grep for `22003` / `01S07` / `07006` / `22018` across the e2e suite hit `execute_test.cpp` - the *parameter* direction - and almost nothing on the fetch path. Those are the SQLSTATEs an application keys on to tell "no value" from "bad value", so they were worth pinning against the reference driver rather than against our own reading of the spec.

## What the parity run settled

Every SQLSTATE predicted from msodbcsql's conversion suite held on both drivers. Three behavioral details did not survive contact:

**GUID text rendering diverged.** msodbcsql renders a `uniqueidentifier` in upper case; `uuid`'s `Display` is lower case, so we returned `0123abcd-...` where msodbcsql returns `0123ABCD-...`. Fixed and pinned through the shared renderer plus narrow, wide, and bound-fetch e2e paths. This is the **third** rendering divergence found this way, after decimal and money - and the work item predicted this exact one, on the grounds that neither `Cnvrsnpp` nor the `GetData` tests pin braces or case, so only a live comparison could answer it.

**The indicator on a `22003` is not a safe assertion.** The work item recorded it as a regression guard - "this holds today, so it's not a suspected divergence." It doesn't: msodbcsql writes `0` to the indicator on a range failure, and this driver leaves the caller's value alone. ODBC leaves the indicator *undefined* when the call fails, so neither driver is wrong and no application may legitimately read it. Asserting either would encode something consumers must not rely on, so the tests assert the **value buffer** is untouched instead - the invariant that actually matters, and one both drivers share. An internal regression test pins this driver's no-write behavior.

**A retry after `22018` diverges.** A live follow-up check confirmed that msodbcsql consumes the column after the failed conversion and returns `SQL_NO_DATA` when the caller retries with a valid C type. This driver leaves the undelivered value readable, giving the application a recovery path. The driver-specific behavior remains pinned and the difference is recorded in the known-divergences table.

## What's in the diff

- `tests/e2e/tests/fetch_error_semantics_test.cpp` - 21 tests: out-of-range paired against its signed/unsigned counterpart; fractional truncation and the `SQL_C_BIT` carve-out; restricted pairings (`07006`) paired with legal-but-unconvertible character data (`22018`); `datetimeoffset` into `SQL_C_SS_TIME2`; the `01004` buffer boundary; a range error as a per-row block-fetch error; GUID text through `SQLGetData` narrow/wide and bound narrow targets; and `datetime`'s 1753 floor.
- `src/api/get_data.rs` and `Cargo.toml` - render GUID text directly in upper case with `Uuid::encode_buffer()`, plus unit coverage and the production dependency placement that API requires.
- `docs/typed-columnar-fetch-plan.md` - record the fourth overall divergence, undefined `22003` indicator behavior, the confirmed `22018` retry difference, and completion of Task 47678 coverage.

## Testing

- e2e parity: **34 binaries, 0 unexpected divergences, 0 shared failures** against msodbcsql 18 on SQL Server 2025
- focused fetch-error parity suite against msodbcsql on Windows: **20 passed, 1 known divergence skipped**
- **1094 mssql-odbc tests passed**; clippy and fmt clean
- required PR validation and 100% diff coverage passed on the previous head; validation is running on the latest review-fix commit

## Breaking changes

`uniqueidentifier` fetched as `SQL_C_CHAR` / `SQL_C_WCHAR` now renders upper case. `mssql-python` builds a `uuid.UUID` from it, which is case-insensitive, so it is unaffected; an application comparing the rendered string would see the difference. This makes us match msodbcsql.

## Work items

[AB#47678](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47678)


mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// A conversion failure as the per-row error inside a block fetch. The
+// mechanism is already covered via 22002; this is the same path reached by a
+// range error instead, with the rest of the rowset still delivering.
+// ---------------------------------------------------------------------------
+
+TEST_F(FetchErrorSemanticsLiveTest, ARangeErrorIsARowErrorInABlockFetch) {
+    ASSERT_SQL_OK(ExecDirect("SELECT CAST(1 AS INT) AS c1 UNION ALL SELECT CAST(300 AS INT)"),
```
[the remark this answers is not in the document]
David-Engel: Added `ORDER BY c1` so the in-range row is guaranteed to occupy slot 0 and the range-error row slot 1 before the slot-specific assertions run.

> [!NOTE]

Vahid-b (commented):
Reviewed the fetch-path error-semantics tests and the GUID upper-casing fix, cross-checking the two claims flagged as newly-discovered directly against msodbcsql source rather than the PR description:

- GUID upper-case rendering matches `sqlccnvt.cpp` (`StringCchPrintfA("%08X-%04X-%04X-", ...)` plus `HexToChar` returning `'A'..'F'`) - uppercase hex, no braces, exactly what `column_value_to_text` now produces.
- The `SQL_C_BIT` carve-out (`22003` for any negative value including small fractions, vs `01S07` for other unsigned targets) matches `sqlccnvt.cpp:5133-5136` (`dTemp < 0 && fTypeOut == SQL_C_BIT -> CVT_PREC`, checked before the general fractional-truncation path).
- Mutation-tested `guid_renders_in_upper_case`: reverting the `to_ascii_uppercase()` fix makes it fail (`...abcd... != ...ABCD...`), so it isn't vacuous.
- Confirmed `column_value_to_text` is the shared renderer for both `SQLGetData` and bound-column `SQLFetch`, so the fix covers both fetch paths.

The one open item at review time - the `UNION ALL` row-order assumption in `ARangeErrorIsARowErrorInABlockFetch` - is already fixed in `95a05ece` (`ORDER BY c1`); no action needed there.

No further findings. Not merging - that's the author's call.

mssql-odbc/src/api/get_data.rs
```diff
@@ -1185,7 +1185,9 @@ pub(crate) fn column_value_to_text(v: &ColumnValues) -> Result<String, TextError
         ColumnValues::Json(j) => {
             String::from_utf8(j.bytes.clone()).map_err(|_| TextError::Malformed)
         }
-        ColumnValues::Uuid(u) => Ok(u.to_string()),
+        // msodbcsql renders a uniqueidentifier in upper case; uuid's Display is
+        // lower case.
+        ColumnValues::Uuid(u) => Ok(u.to_string().to_ascii_uppercase()),
```
David-Engel: Optional: `to_string()` allocates the lower-case form and `to_ascii_uppercase()` then allocates a second `String` and re-scans it. `uuid` renders upper case directly:

```rust
ColumnValues::Uuid(u) => Ok(u
    .hyphenated()
    .encode_upper(&mut uuid::Uuid::encode_buffer())
    .to_string()),
```

One allocation, and the intent is in the API name rather than in a post-processing step (the comment above it is still worth keeping). `Hyphenated::encode_upper` is present in the resolved uuid 1.26. Cosmetic — the current form is correct.

David-Engel (commented):
**Automated review (unattended run).** These findings were not checked by a human before posting. Reviewed the full diff against `a85f4d22` on head `95a05ece`, read the existing Copilot thread and @Vahid-b's pass first, and ran `cargo nextest run -p mssql-odbc --lib -E 'test(guid) or test(column_value_to_text)'` (4 passed). e2e was reviewed by reading — it needs a live server and the reference driver.

**Summary:** the GUID fix is correct and complete for both text paths. `column_value_to_text` has exactly two callers (`get_data.rs:390` for `SQLGetData`, `fetch_scroll.rs:698` for the bound path), so `SQL_C_CHAR` and `SQL_C_WCHAR` on both fetch paths pick it up, and the `SQL_C_GUID` struct target goes through `convert_guid_c` and is untouched. No other test in the repo pins a lower-case rendering, and the cross-repo `mssql-python` suite is green. The new e2e binary is picked up by ctest through `add_odbc_test`, which is how both runners discover tests — no separate list to update — and the file count now matches the 34 binaries in the description. All required checks are green on this head.

I did not re-raise anything from the earlier rounds; the `UNION ALL` row-order thread is fixed in `95a05ece` and the `SQL_C_BIT` / upper-case claims were already checked against `sqlccnvt.cpp`.

**Blocking:** none found.

**Suggestion**

`mssql-odbc/docs/typed-columnar-fetch-plan.md:298` — the P5 status row is now stale in two ways. It still lists "the fetch-path error semantics (22003 / 01S07 / 07006), which are unit-tested but have almost no parity coverage (Task 47678)" under *Not yet covered*, which is exactly what this PR closes, and it says the coverage "found three divergences, all fixed" — the GUID case makes four. Keeping that row current is the established pattern here: cf6139bd wrote it as part of the decimal/money parity work, and #424 came back just to correct its figures. Suggest folding the GUID divergence into that sentence and dropping 47678 from the not-yet-covered list in this PR, so the doc doesn't drift the way the last one did.

**Nit**

One inline, on the GUID render.

mssql-odbc/src/api/get_data.rs
David-Engel: Accepted. GUID text now uses `Hyphenated::encode_upper` to render the desired form directly with one allocation, while retaining the parity comment.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
```diff
@@ -1185,7 +1185,12 @@ pub(crate) fn column_value_to_text(v: &ColumnValues) -> Result<String, TextError
         ColumnValues::Json(j) => {
             String::from_utf8(j.bytes.clone()).map_err(|_| TextError::Malformed)
         }
-        ColumnValues::Uuid(u) => Ok(u.to_string()),
+        // msodbcsql renders a uniqueidentifier in upper case; uuid's Display is
+        // lower case.
+        ColumnValues::Uuid(u) => {
```
shiwanigupta0809, line 1190: **Suggestion:** the behaviour change ships without coverage on the paths the breaking-change note names.

`column_value_to_text` is shared: `SQLGetData` reaches it at line 390, and the bound fetch path reaches it from `fetch_scroll.rs`. The PR's breaking-change note says `uniqueidentifier` fetched as `SQL_C_CHAR` **or `SQL_C_WCHAR`** now renders upper case. `GuidRendersAsUppercaseWithoutBraces` covers exactly one of those four combinations — `SQLGetData` + `SQL_C_CHAR`.

Two reasons that is thinner than it looks:

- The sibling file already treats this as a two-path question by construction. `numeric_types_test.cpp` pairs `UniqueidentifierToGuidTargetViaBoundFetch` with `UniqueidentifierToGuidTargetViaGetData` for the `SQL_C_GUID` target specifically so a divergence between them cannot hide. The new GUID test has only the `ViaGetData` half.
- `SQL_C_WCHAR` is not the same copy. It runs the rendered string through `encode_utf16()` and a separate `copy_with_nul`/`buf_elements` path with its own indicator arithmetic (bytes, not characters). A 36-character GUID into a `SQLWCHAR[37]` is a boundary case none of the existing text tests hit, and `ind` should be 72 rather than 36 there.

Two more short tests — bound `SQL_C_CHAR` and `SQLGetData` `SQL_C_WCHAR` — would make the claim the doc change asserts actually true. Given the P5 table now says the GUID divergence is fixed and 47678 is off the remaining-gaps list, this is the coverage that backs it.

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
```diff
@@ -0,0 +1,308 @@
...
+    SQLLEN ind = 0;
+    EXPECT_EQ(SQL_ERROR, SQLGetData(stmt_, 1, SQL_C_BIT, &v, sizeof(v), &ind));
+    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "22003");
+    EXPECT_EQ(9, static_cast<int>(v)) << "a range failure must not write a partial value";
+    SQLCloseCursor(stmt_);
+}
+
+// ---------------------------------------------------------------------------
+// 07006 — the source and target are not a legal pairing at all. Distinct from
+// 22018, which means the pairing is legal but the data is not convertible.
```
shiwanigupta0809, line 174: **Suggestion:** this section header names a distinction the file never demonstrates.

"Distinct from 22018, which means the pairing is legal but the data is not convertible" is the right framing, and it is the confusion most likely to actually occur — a driver that collapses "illegal pairing" and "unconvertible data" into one state is precisely the regression this file exists to catch. But there is no `22018` case anywhere in it, so that regression passes green.

One test closes it: `SELECT CAST('abc' AS VARCHAR(10))` into `SQL_C_SLONG` is a *legal* pairing with unconvertible data, so it must be `22018` and not `07006`. Placing it directly beneath `DateIntoNumericTargetIsRestricted` makes the pair self-explanatory in the same way the signed/unsigned `22003` pairs already are — which is the strongest structural idea in this file.

This also bears on the doc change: removing Task 47678 from "Remaining gaps" reads as "the fetch-path error semantics are pinned", and `22018` is one of the four states an application distinguishes on this path.

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
```diff
@@ -0,0 +1,308 @@
...
+                                 0),
+                  SQL_HANDLE_STMT, stmt_);
+    ASSERT_SQL_OK(SQLSetStmtAttr(stmt_, SQL_ATTR_ROW_STATUS_PTR, status, 0), SQL_HANDLE_STMT,
+                  stmt_);
+    ASSERT_SQL_OK(SQLBindCol(stmt_, 1, SQL_C_STINYINT, v, sizeof(signed char), ind),
+                  SQL_HANDLE_STMT, stmt_);
+
+    EXPECT_EQ(SQL_SUCCESS_WITH_INFO, SQLFetchScroll(stmt_, SQL_FETCH_NEXT, 0));
+    EXPECT_EQ(1, static_cast<int>(v[0])) << "the in-range row still delivers";
+    EXPECT_EQ(SQL_ROW_ERROR, status[1]) << "300 does not fit a signed byte";
```
shiwanigupta0809, line 315: **Suggestion:** this is the only bound-path case in the file, and it asserts neither of the two things that make it a *range*-error test.

- The SQLSTATE is never checked. `SQL_ROW_ERROR` is set for any row failure, so `07006`, `22018`, or an unrelated conversion bug all pass. Every other test here pins its state; this one does not, despite being the case that reaches the state through a different code path (`fill_rowset` → `deliver_bound` → `convert_typed_c`) than the `SQLGetData` tests above.
- `v[1]` is never checked. "A range failure must not write a partial value" is asserted five times in the `22003` section, but all five go through `SQLGetData`. Bound delivery writes into the caller's array through an entirely separate routine, so the invariant is genuinely unverified for exactly the path where a partial write is most damaging — it corrupts a slot in an application array rather than a single scalar.

Suggest adding `EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "22003")`, `EXPECT_EQ(0, static_cast<int>(v[1]))` against a recognisable sentinel rather than `0`, and `EXPECT_EQ(SQL_ROW_SUCCESS, status[0])` so the good row's status is pinned too, not just its value.

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
```diff
@@ -0,0 +1,308 @@
...
+        Connect();
+    }
+
+    SQLRETURN ExecDirect(const std::string& sql) {
+        SqlTString s = ODBCTestUtils::ToSqlTStr(sql);
+        return SQLExecDirect(stmt_, const_cast<SQLTCHAR*>(s.c_str()), SQL_NTS);
+    }
+
+    // Executes |sql| and positions on its single row.
+    void FetchOne(const std::string& sql) {
```
shiwanigupta0809, line 55: **Nit:** `ASSERT_*` inside a `void` helper returns from the helper, not from the test.

If `ExecDirect` or the `SQLFetch` fails, `FetchOne` records a fatal failure and returns, but the calling test body carries straight on into `SQLGetData` against a statement that is not positioned on a row. The test does fail — but with a cascade of secondary failures on top of the real one, and in the `SQL_C_TYPE_TIMESTAMP` / `SQL_DATE_STRUCT` cases it dereferences output structs the driver never wrote.

Sixteen call sites, so the cheap fix is to have `FetchOne` return `bool` and `ASSERT_TRUE` at the call site, or wrap with `ASSERT_NO_FATAL_FAILURE(FetchOne(..))`. Worth doing here specifically because a whole-file failure mode — wrong server version, missing collation, connection dropped — currently produces sixteen confusing failures instead of sixteen clear ones.

mssql-odbc/src/api/get_data.rs
```diff
@@ -1185,7 +1185,12 @@ pub(crate) fn column_value_to_text(v: &ColumnValues) -> Result<String, TextError
...
+            let mut buffer = [0; 36];
```
shiwanigupta0809: **Nit:** following up the resolved `encode_upper` thread rather than reopening it — the implementation landed with a literal `36` where the suggestion used `uuid::Uuid::encode_buffer()`.

The difference matters slightly more than style: `Hyphenated::encode_upper` **panics** if the buffer is shorter than `Hyphenated::LENGTH`, so a wrong literal is a runtime panic unwinding through an `extern "C"` boundary rather than a compile error — a failure mode this file elsewhere goes out of its way to avoid ("fall back to a verbatim copy in release rather than panicking across the FFI boundary").

`36` is correct today and the crate will not change it, so this is genuinely a nit. But `uuid::Uuid::encode_buffer()` or `[0u8; uuid::fmt::Hyphenated::LENGTH]` costs nothing and moves the guarantee from "the reviewer checked the constant" to "the compiler checks it".

shiwanigupta0809 (commented):
Reviewed the SQLSTATE expectations against the ODBC conversion tables and the driver's own conversion core, and checked the GUID change's blast radius. No blocking findings.

Things I specifically verified and found correct:

- Every asserted state matches the spec independently of the parity run. `SQL_C_BIT` really is carved out — numeric to bit is `01S07` only in `(0, 2)` excluding 1, and `-0.01 < 0` falls in the `22003` band — so `SmallNegativeFractionIntoUnsignedTruncates` / `...IntoBitIsOutOfRange` is a genuine pair rather than a quirk. `SQL_TYPE_DATE`→`SQL_C_SLONG` and `SQL_TYPE_TIMESTAMP`→`SQL_C_DOUBLE` are both absent from the conversion table, so `07006` is right.
- Not asserting the indicator on a failed call is the correct call, and the reasoning in the section header is worth keeping.
- The GUID change has no second rendering site to fall out of sync with. `column_value_to_text` is the only place `mssql-odbc` renders a `Uuid` as text — `param_convert.rs` only maps `SQL_GUID` to a type — and no existing e2e or unit test pins the lower-case form, so nothing conflicts. Upper case also matches T-SQL's own `CONVERT` output, so this converges with the server as well as with msodbcsql.

The suggestions below are all about coverage the PR's own framing claims. I have not re-filed the two resolved threads (`ORDER BY`, `encode_upper`), though one nit follows up on the second.

**Blocking:** none.
**Suggestion:** three.
**Nit:** two.

mssql-odbc/src/api/get_data.rs
David-Engel: Accepted. Added bound `SQL_C_CHAR` and `SQLGetData` `SQL_C_WCHAR` coverage alongside the existing narrow `SQLGetData` case, including the 72-byte wide indicator boundary.

> [!NOTE]

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
David-Engel: Accepted. The block-fetch case now asserts diagnostic state `22003`, success status for the good row, error status for the bad row, and preservation of a recognizable sentinel in the failed slot.

> [!NOTE]

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
David-Engel: Accepted. Every `FetchOne` call is now wrapped in `ASSERT_NO_FATAL_FAILURE`, so setup failures stop each test before it touches uninitialized output.

> [!NOTE]

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
David-Engel: Accepted. Added a legal `VARCHAR` to `SQL_C_SLONG` conversion with unconvertible data and pinned `22018`, including the no-partial-write invariant.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
David-Engel: Accepted. Moved the existing `uuid` dependency into the production dependency set and now use `Uuid::encode_buffer()`, making the buffer-size contract library-owned rather than a runtime-sensitive literal.

> [!NOTE]

David-Engel (commented):
**Automated review (unattended run).** These findings were not checked by a human before posting.

## Summary

Adds an e2e parity suite pinning the fetch path's error SQLSTATEs (22003 / 01S07 / 07006 / 22018 / 01004) against msodbcsql, and fixes GUID text rendering to upper case. The code change is one arm of `column_value_to_text` and is correct and complete; everything else is test and doc. No blocking findings.

This head folds in every suggestion from the earlier rounds — `ORDER BY c1`, `ASSERT_NO_FATAL_FAILURE` at all 18 `FetchOne` call sites, the 22018 case, the strengthened block-fetch assertions, the bound `SQL_C_CHAR` and `SQLGetData` `SQL_C_WCHAR` GUID cases, and `Uuid::encode_buffer()` in place of the literal `36`. I did not re-raise any of them.

## Verification

- **GUID blast radius.** `column_value_to_text` is the only site in `mssql-odbc` that renders a `Uuid` as text (`grep -n 'Uuid' mssql-odbc/src`), and it has exactly two callers: `get_data.rs:390` (`SQLGetData`) and `fetch_scroll.rs:698` (bound fetch). So all four combinations named in the breaking-change note pick the fix up from one place. `param_convert.rs:562` only maps `SQL_GUID` to a type, and the `SQL_C_GUID` struct target goes through `fetch_convert.rs:304` untouched.
- **`encode_buffer()` swap is safe.** `Uuid::encode_buffer()` is 45 bytes against `Hyphenated::LENGTH` of 36, so the panic path the earlier nit worried about is now library-owned. Cross-crate type identity is compiler-enforced: `ColumnValues::Uuid` carries `mssql-tds`'s `uuid::Uuid`, so a major-version split would fail to compile rather than misbehave — the rationale the dropped `Cargo.toml` comment carried is not lost, just enforced elsewhere. `uuid` moving from `[dev-dependencies]` to `[dependencies]` is required now that the render is production code.
- **22018 vs 07006 is a real distinction in this driver, not a spec reading.** `fetch_convert.rs:126,133,764` separates "text that is not a number" from "illegal source/target pairing", and `fetch_scroll.rs:105` / `get_data.rs:397` post `ERR_INVALID_CHARACTER_VALUE` on both paths. The new `UnconvertibleCharacterDataIsInvalidCharacterValue` pins the boundary the section header had been claiming without demonstrating.
- **Wide-indicator arithmetic.** `EXPECT_EQ(72, ind)` for a 36-character GUID matches the suite's existing 2-byte `SQLWCHAR` convention (`get_data_test.cpp:579` asserts 20 for 10 code units). `SQLWCHAR buf[37]` is an exact fit for 36 units plus the terminator, and the comparison loop reads `sizeof(expected) == 37` elements, so it checks the terminator and stays in bounds.
- **The new binary is picked up.** `add_odbc_test` is what both runners discover tests through, so CMake registration is the whole wiring; the count is now 34, matching the description.
- **CI.** All checks green on this head, including the Linux PR leg that runs the suite a second time against msodbcsql 18 with `ODBC_E2E_COMPARE=1` and fails on divergence — which is what makes these parity assertions evidence rather than assertion. Diff coverage 100% (8 lines), overall 93.1%.

I did not build locally. The unit test was already mutation-checked in an earlier round, the arm it covers is unchanged in substance since, and the e2e half needs a live server plus the reference driver.

## Blocking

None found.

## Suggestion

None.

## Nit

**PR description is stale against this head.** "What's in the diff" still says *17 tests*; the file has 20 `TEST_F`s. The three added since — the 22018 case, `GuidRendersAsUppercaseWithoutBracesViaGetDataWchar`, and `GuidRendersAsUppercaseWithoutBracesViaBoundFetchChar` — are the ones that make the breaking-change note's "`SQL_C_CHAR` / `SQL_C_WCHAR`" claim actually covered, so they are worth naming. The bullet also still frames 22018 as the thing 07006 is *distinct from* rather than something now pinned, which `typed-columnar-fetch-plan.md` has already been updated to say. Worth a pass before merge, per the repo's own "re-evaluate the title and description" convention.

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
```diff
@@ -0,0 +1,356 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// fetch_error_semantics_test.cpp  –  E2E parity for the fetch path's *error*
+// semantics: 22003 (out of range), 01S07 (fractional truncation), 07006
+// (restricted conversion) and 22018 (unconvertible character data).
+//
+// AB#47678. The P5 type-coverage work proved the fetch type map is wired up,
```
Theekshna, line 6: AB#47678 lists 12 cells; 11 are here. `datetimeoffset` into `SQL_C_SS_TIME2` is not, and is not covered elsewhere either — `datetime_types_test.cpp` only reaches `SQL_C_SS_TIME2` from a `TIME(7)` source, and every `DATETIMEOFFSET` source in the suite targets `SQL_C_SS_TIMESTAMPOFFSET`.

It is also the cell most worth having. `fetch_convert.rs:715` writes `SS_TIME2` unconditionally, while the sibling `SQL_C_TYPE_DATE` and `SQL_C_TYPE_TIME` arms both return `ConvOk::Truncated` when they drop a component — so discarding the date *and* the UTC offset returns a plain `SQL_SUCCESS`. msodbcsql has the case (`TCODBC_CTypeExtensions.cpp:7669` GetData, `:7699` BindCol) and treats it as legal and value-delivering, but its harness accepts either `SQL_SUCCESS` or `SQL_SUCCESS_WITH_INFO` (`:6065-70`), so only a live run settles the return code.

Worth closing before 47678 resolves, or splitting out with the asymmetry noted.

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
```diff
@@ -0,0 +1,356 @@
...
+    SQLCloseCursor(stmt_);
+}
+
+TEST_F(FetchErrorSemanticsLiveTest, GuidRendersAsUppercaseWithoutBracesViaGetDataWchar) {
+    ASSERT_NO_FATAL_FAILURE(
+        FetchOne("SELECT CAST('0123ABCD-4567-89EF-0123-456789ABCDEF' AS UNIQUEIDENTIFIER) AS c1"));
+
+    SQLWCHAR buf[37] = {};
+    SQLLEN ind = 0;
+    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_WCHAR, buf, sizeof(buf), &ind), SQL_HANDLE_STMT,
```
Theekshna: Nit: `ASSERT_SQL_OK` is `SQL_SUCCEEDED`, so the two exactly-fitting buffers — `SQLWCHAR buf[37]` (74 bytes for 72 + 2) here, and `char buf[37]` at line 330 — accept `SQL_SUCCESS_WITH_INFO`, which leaves the boundary unpinned by the return code. `ABufferThatExactlyFitsSucceeds` above uses `EXPECT_EQ(SQL_SUCCESS, ...)`; the same here would be consistent. The content assertions would still catch a real truncation, so this is only a tightening.

mssql-odbc/docs/typed-columnar-fetch-plan.md
```diff
@@ -295,4 +295,4 @@ Both of those landed with the fetch rework in [#153](https://github.com/microsof
...
...| 46582 | Implemented — the fetch type map is covered end to end on both paths for the date/time, integer, bit, float, GUID, decimal, money and character families, against msodbcsql ([#376](https://github.com/microsoft/mssql-rs/pull/376), [#426](https://github.com/microsoft/mssql-rs/pull/426)). That coverage found four divergences, all fixed: #376 corrected decimal and money rendering below magnitude one; review surfaced a third where `SQLGetData` returned success for a NULL column read with no indicator, which [#382](https://github.com/microsoft/mssql-rs/pull/382) fixed independently and first (AB#47507); and #426 corrected lower-case GUID text to match msodbcsql's upper case. #426 also added parity coverage for the fetch-path error semantics tracked by Task 47678, including 22003, 01S07, 07006 and 22018. The `mssql-python` suite runs against a locally-swapped build on Linux and reproduces CI exactly (see above for current figures); the Windows equivalent is blocked on a pooling hang (Task 47510). Remaining gaps are tracked rather than open here: bound `SQL_C_BINARY` (Task 47239) and bound PLP/XML (Task 47361), both blocked on unimplemented delivery rather than on missing tests. |
```
Theekshna: The 22003 indicator difference — msodbcsql writes 0, this driver leaves the caller's value — is explained in the PR body and the new file's header, but not in this file's **Known divergences from msodbcsql** table. That table exists for exactly this shape: its intro records cases "because the parity run will not surface them", and here the tests deliberately do not assert the indicator, so nothing will rediscover it. Since this row also drops 47678 from the remaining gaps, worth one row while the file is open.

Theekshna (commented):
Reviewed the diff, the work item, and the msodbcsql sources rather than the PR's summary of them. No blocking findings.

The production change is correct and I confirmed it independently against `sqlccnvt.cpp`: upper case (`"%08X-%04X-%04X-"` at :1847, `HexToChar` returning `'A' + b - 10` at :7348), no braces (`StringFromGUID2` is not used), identical for `SQL_C_CHAR` and `SQL_C_WCHAR` (`ConvertToWChar` :3119-3200 widens `ConvertToChar`'s output), and 36 / 72 for the indicator (`:3008`). No direction-dependence, so parameters are unaffected.

Against AB#47678, 11 of the 12 listed cells are here. The missing one is inline below.

Two items live outside this diff:

**`get_data_test.cpp:599` — this PR obsoletes a skip it does not remove.** `InvalidCharacterForNumericTargetIs22018ThenValueReadable` is `SKIP_IF_COMPARING_MSODBCSQL()`-gated with a `TODO(convergence)` saying 22018 "has not been confirmed against a live run", and instructing whoever confirms it to drop the skip. `UnconvertibleCharacterDataIsInvalidCharacterValue` asserts the same rule unskipped and the compare leg is green, so that confirmation is this PR. Leaving it means the suite still declares the question open while the doc change here says the semantics are pinned. The older test also asserts more (column not consumed, value still readable), so it is worth unskipping rather than treating as a duplicate.

**The indicator regression guard from 47678 is dropped without a record.** The work item asks to assert the indicator is untouched on 22003, framing it as a regression guard on our own behaviour. Declining it *in this file* is right — it is a parity file and the two drivers genuinely differ. But nothing now guards `fetch_convert.rs:65` regressing to write the indicator before `narrow!` propagates. Either add it as a plain unit test, or record the decision on the work item so it does not read as an oversight.

Not re-raising anything from the earlier rounds. Checked and found already covered, so not raising: bound-path `01004` (`fetch_scroll_test.cpp:252` / `:266`), the single-row `SQL_ERROR` half of the `row_array_size == 1` branch (`numeric_types_test.cpp:290`), and the `uuid` dependency move, which changes no features because `mssql-tds` already takes uuid's defaults.

Did not build locally — this toolchain hits an unrelated AVX-512 codegen ICE in `crc32fast`. Relied on CI, green on this head.

mssql-odbc/docs/typed-columnar-fetch-plan.md
David-Engel: Documented the undefined 22003 indicator behavior in the known-divergences table and strengthened `int_out_of_range_for_smallint_leaves_outputs_unchanged` with sentinel value and indicator assertions.

> [!NOTE]

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
David-Engel: Added the 12th work-item cell in `DatetimeoffsetIntoSsTime2Succeeds`. A live msodbcsql run settled the return as exactly `SQL_SUCCESS`; the test also verifies that the TIME2 payload and indicator are written. It passes with all 21 focused parity cases.

> [!NOTE]

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
David-Engel: Changed both exact-fit paths to require exactly `SQL_SUCCESS`: wide `SQLGetData` and bound narrow `SQLFetch`. The 21-case live msodbcsql run passes.

> [!NOTE]

David-Engel:
Followed up on the review-summary suggestion to remove the 22018 comparison skip with a live msodbcsql run. The initial conversion agrees on 22018, but msodbcsql then consumes the column and returns SQL_NO_DATA from the valid-type retry; mssql-odbc deliberately leaves the undelivered value readable. I retained the driver-specific skip, replaced the speculative TODO with the observed behavior, and recorded the divergence in the status document. The same batch also pins the 22003 no-write behavior with sentinels.

mssql-odbc/tests/e2e/tests/fetch_error_semantics_test.cpp
Theekshna: Reopening this one. The cell is covered now, but `DatetimeoffsetIntoSsTime2Succeeds` does not pin the part that made it interesting.

`EXPECT_LE(t.hour, 23)` and `EXPECT_LE(t.minute, 59)` are range checks — they would catch a garbage decode, but they cannot distinguish wall-clock delivery from an offset shift. With `+05:30` a shift moves both fields (12:34 becomes 07:04 for a UTC client), and both results satisfy the bounds.

That matters because this pair is a real divergence, and on the native column path rather than only for literals: `sqlccnvt.cpp:3908-3933` handles `SQL_TIMESTAMPOFFSET_MAPPED` straight off the wire, and lists `SQL_C_SS_TIME2` at `:3920` among the targets that get `ConvertOffsetToLocal`. So msodbcsql returns the client machine's local time here, while this driver returns 12:34 as written. The test has no `SKIP_IF_COMPARING_MSODBCSQL()`, so it has to tolerate both — which is presumably why the bounds are loose, and why passing the 21-case parity run is not evidence either way for hour and minute. `second`, `fraction` and `ind` are unaffected by a whole-minute shift, so those three assertions are genuine.

Nothing else pins it either: `offset_is_ignored_for_non_offset_targets` is a Rust unit test covering `SQL_C_TYPE_TIMESTAMP` from a character source, and no e2e test reaches `SQL_C_SS_TIME2` from a `datetimeoffset` column. A regression to shifting would pass everything.

Suggest asserting `t.hour == 12` and `t.minute == 34` with `SKIP_IF_COMPARING_MSODBCSQL()` and a pointer to the offset row in the divergences table. That row is also narrower than the behaviour — it says "a UTC offset **in a literal**", but `:3908` shows the shift applies to native `datetimeoffset` columns too.

David-Engel: Agreed. `DatetimeoffsetIntoSsTime2Succeeds` now skips the msodbcsql comparison leg and asserts the native-column wall clock exactly as `12:34`; the divergence row now covers any `datetimeoffset` value and names both the character-source unit test and this native-column e2e test. The reference run reports 20 passes and this one documented skip.

> [!NOTE]

Theekshna (approved):

PR #427, opened 2026-08-28
mssql-odbc: deliver bound PLP (max/LOB) columns instead of failing the row

David-Engel:
## Summary

Implements [AB#47361](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47361). A bound `varchar(max)` / `nvarchar(max)` / `xml` / `text` column was reported as `SQL_ROW_ERROR` with `HYC00` rather than delivered. msodbcsql delivers it, so this was a standing divergence rather than a missing test.

P3b delivers bound columns from materialized `ColumnValues`, but PLP columns arrive as `PlpStreaming` and must be pulled in chunks. `SQLGetData` already does that, but its machinery expects the caller to return for subsequent chunks; a bound column has one fixed buffer that is filled once.

## Design

The stream must be consumed either way to keep the row synchronized, so the only choice is whether bytes land in the caller's buffer or a discard buffer.

The implementation drains straight into the bound buffer one chunk at a time, then keeps draining without copying after the buffer fills. Nothing larger than a chunk is retained. It preserves complete UTF-8 characters and UTF-16 surrogate pairs at truncation boundaries, and propagates wire errors encountered while filling or draining instead of converting them into a successful fetch.

## Indicator behavior

The indicator behavior was characterized against msodbcsql before implementation:

| Case | rc | SQLSTATE | indicator |
| --- | --- | --- | --- |
| `nvarchar(max)` fits → `SQL_C_CHAR` | `SUCCESS` | — | produced length |
| `nvarchar(max)` truncated → `SQL_C_CHAR` | `SUCCESS_WITH_INFO` | `01004` | `SQL_NO_TOTAL` |
| `nvarchar(max)` truncated → `SQL_C_WCHAR` | `SUCCESS_WITH_INFO` | `01004` | full length |
| `varchar(max)` truncated → `SQL_C_CHAR` | `SUCCESS_WITH_INFO` | `01004` | full length |

A fitting value reports what was produced. A truncated value reports the full length when target units match wire units, and `SQL_NO_TOTAL` when transcoding makes the wire byte count unsuitable. These cases match msodbcsql.

## Deliberately not included

- Bound `SQL_C_BINARY` remains unimplemented for fixed-width and max binary values, tracked by [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239). Refused streams are still drained so subsequent columns and rows remain synchronized.
- Bound narrow PLP (`varchar(max)` / `json`) to `SQL_C_WCHAR` remains unsupported in this branch, tracked by [AB#47767](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47767). msodbcsql delivers this pairing, and `SQLGetData` supports it after #396.
- Bound PLP to typed C targets remains unsupported, also tracked by [AB#47767](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47767). This means a fixed-width value and its max counterpart can still differ for the same typed binding because only the non-max path reaches `convert_typed_c`.

## Testing

- e2e parity: **33 binaries, 0 divergences, 0 shared failures** against msodbcsql 18 on SQL Server 2025 before the final review fix
- 1118 mssql-odbc unit tests; clippy and fmt clean
- Nine new e2e tests covering fitting and truncated values on both character targets, non-ASCII transcoding, UTF-8 and surrogate truncation boundaries, native JSON, and the unsupported binary case
- `ABoundLobColumnDoesNotDesyncTheRow` and rowset coverage now assert successful bound delivery, indicators, and subsequent-column synchronization

## Work items

[AB#47361](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47361)


mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+    let mut pending_byte: Option<u8> = None;
+    let mut pending_high_surrogate: Option<u16> = None;
+    let mut truncated = false;
+    let mut wire_total: Option<u64>;
+    let mut scratch = vec![0u8; PLP_BOUND_CHUNK];
+
+    loop {
+        let chunk = match runtime.block_on(client.read_active_plp_chunk(&mut scratch)) {
+            Ok(c) => c,
+            Err(_) => return RowOutcome::Error(RowIssue::Restricted),
```
saurabh500: **Blocking.** A chunk-read error here is discarded (`Err(_) =>`) and reported as `RowIssue::Restricted` -> SQLSTATE 07006, which is unrelated to the actual transport / timeout / cancellation / protocol failure, and the real `TdsError` is never posted via `post_tds_error`, so `SQLGetDiagRec` loses the error number and message.

More seriously, this path never sets `fetch_error`, so the teardown at fetch_scroll.rs:514-524 (`reset_row_stream` + clear `STMT_STATE_CURSOR_OPEN` + `post_tds_error`) is skipped. If the failing LOB is the last bound column of the last row in the rowset, nothing re-hits the error, the row is recorded as a benign 07006 row error, and the connection -- left positioned mid-PLP -- can be handed back as reusable. That is a real desync on network-reachable data, the exact class this PR exists to prevent.

The sibling `SQLGetData` PLP path handles the identical error correctly: get_data.rs:778-784 clears the cursor-open state and posts the real error via `post_tds_error(&e, SQLSTATE_HY000)` before returning `SQL_ERROR`. Suggest giving `deliver_bound_plp` a way to propagate the `TdsError` to `fill_rowset` (return `Result<RowOutcome, TdsError>`, or set the caller's `fetch_error`) so it routes through the same `fetch_error` path the `read_row_column` `Err(e)` arm at lines 433-436 already uses.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+        let _ = runtime.block_on(drain_plp_to_end(client));
+        return RowOutcome::Error(RowIssue::Unsupported);
+    }
+
+    let transcode = target == SQL_C_CHAR && matches!(encoding, Some(PlpEncoding::Utf16Text));
+    let buf_elements = char_buf_elements(target, stride);
+    // Room for the payload, less the terminator the copy always writes.
+    let capacity_elements = buf_elements.saturating_sub(1);
+
+    let mut out_bytes: Vec<u8> = Vec::new();
```
saurabh500: **Suggestion.** `out_bytes` / `out_units` grow to `capacity_elements` (caller buffer size minus 1) before being copied into `slot`, so the docstring at lines 653-654 ("Nothing larger than one chunk is ever held in memory") isn't accurate: for a large bound buffer the delivered prefix is held twice, once here and again after `copy_with_nul` writes it into `slot`. This is bounded by what the application allocated (not an unbounded OOM, contra the Copilot framing), so it's a double-buffering + stale-docstring nit rather than a correctness bug. You could decode each byte/unit directly into `slot` while tracking an output offset, keeping only `scratch` plus the transcode carry state -- which is what the docstring already claims the code does.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
 }
 
+/// Wire read size when draining a bound PLP column. Bounds peak memory: the
+/// value itself is never materialized, only one chunk at a time.
+const PLP_BOUND_CHUNK: usize = 8 * 1024;
+
+/// Reads the PLP encoding declared for `column` (1-based) in the current
+/// result set.
+fn plp_encoding_for(stmt: &StmtHandle, column: usize) -> Option<PlpEncoding> {
+    let ss = stmt.inner.lock().ok()?;
```
saurabh500: **Suggestion.** `.ok()?` maps a poisoned `stmt.inner` lock to `None`, indistinguishable from genuinely-absent PLP metadata; `None` then falls through the `compatible` gate as incompatible, so a poisoned lock drains the stream and returns HYC00 rather than the `SQL_ERROR` the repo convention requires on poison (.github/instructions/mssql-odbc.instructions.md:127: "For Mutex::lock(), return SQL_ERROR on poison"). Impact is low -- poisoning is rare, and a column that reached `deliver_bound_plp` is by construction PLP so metadata is present -- but propagating an explicit failure to `fill_rowset`, or snapshotting the encoding under the initial statement lock in `fetch_scroll_safe`, would match the convention and avoid a spurious drain.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+                }
+            }
+        } else if transcode {
+            let utf8 = utf16le_chunk_to_utf8(
+                &scratch[..chunk.read],
+                chunk.reached_end,
+                &mut pending_byte,
+                &mut pending_high_surrogate,
+            );
+            for b in utf8.as_bytes() {
```
shiwanigupta0809: **Truncation is applied in raw output units, so a multi-byte character straddling the boundary is delivered cut in half.**

Here `out_bytes` is capped at `capacity_elements` UTF-8 *bytes*. Take `ABoundNvarcharMaxTranscodesNonAscii`'s value into `ABoundNvarcharMaxTruncatedToCharReportsNoTotal`'s buffer — `REPLICATE(NCHAR(233), 5000)` into `char buf[32]`. `capacity_elements` is 31, each U+00E9 is two UTF-8 bytes, so 15 characters fill 30 bytes and the 16th contributes only its lead byte `0xC3` at index 30. The caller gets 31 bytes ending in a dangling lead byte, `copy_with_nul` NUL-terminates after it, and the buffer is not valid UTF-8. Any consumer that decodes it — which is the whole point of `SQL_C_CHAR` here — sees a replacement character or an error at the tail.

The same shape is at line 728 for the wide target: `out_units` is capped per code unit, so if the boundary falls between a high and a low surrogate the buffer ends with a lone unpaired high surrogate. A supplementary-plane LOB (`NCHAR(0x1F600)` repeated) into an odd-element `SQLWCHAR` buffer hits it.

What makes this worth blocking on rather than filing:

- The PR states the indicator rule was characterized against msodbcsql for each case. This intersection was not — `ABoundNvarcharMaxTruncatedToCharReportsNoTotal` and `ABoundVarcharMaxTruncatedReportsFullLength` both use ASCII (`'x'`, `'y'`), and `ABoundNvarcharMaxTranscodesNonAscii` uses a 64-byte buffer for an 8-byte value so it never truncates. Truncation and non-ASCII never meet.
- The sibling `SQLGetData` path deliberately never reaches this state — `"Conservative max_read sizing guarantees the transcoded chunk fits"` plus a `debug_assert!(!truncated, ..)` — so the shared-machinery argument does not carry over. Bound delivery is the first caller that can actually truncate a transcoded chunk.
- msodbcsql cannot produce a split character on a wide→narrow conversion; its output is a code-page string built character by character.

Suggested fix: stop at the last complete character. For the transcode branch, iterate `utf8.chars()` and only push a character whose encoded length still fits; for the wide branch, refuse to push a high surrogate that has no room for its low half. Both are a few lines and keep `truncated`/`reported` unchanged.

And please add the case as a test — `REPLICATE(NCHAR(233), 5000)` into `char buf[32]`, asserting the delivered bytes decode cleanly. Without one this will come back.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+        (target, encoding),
+        (SQL_C_WCHAR, Some(PlpEncoding::Utf16Text))
+            | (SQL_C_CHAR, Some(PlpEncoding::SingleByteText))
+            | (SQL_C_CHAR, Some(PlpEncoding::Utf8Text))
+            | (SQL_C_CHAR, Some(PlpEncoding::Utf16Text))
+    );
+    if !compatible {
+        // The stream still has to be consumed, or the next column decodes from
+        // the middle of this value.
+        let _ = runtime.block_on(drain_plp_to_end(client));
```
shiwanigupta0809: **Suggestion:** this `let _ =` discards the one error that the surrounding test exists to prevent.

Different call site and different consequence from the `Err(_)` thread already open on line 713, so flagging separately. If `drain_plp_to_end` fails part-way — transport drop, timeout, cancellation — the row cursor is left positioned *inside* the LOB, and `fill_rowset` neither records a `fetch_error` nor breaks. The loop moves to the next binding and calls `read_row_column`, which decodes payload bytes as a length prefix. That is precisely the failure `ABoundLobColumnDoesNotDesyncTheRow` was written for, and its comment says it "segfaulted the driver rather than failing cleanly."

On line 713 a swallowed error yields a mislabeled `07006`; here it yields silent corruption of the *following* column, which is worse and harder to attribute.

Whichever way the 713 thread is resolved should apply here too — propagate to `fill_rowset` so `fetch_error` is set and the row loop stops. Worth noting the resolution has to differ slightly: this path must still return `RowIssue::Unsupported` for the column itself when the drain *succeeds*, since `HYC00` is the deliberate answer for a refused binary target.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+            }
+            for pair in bytes[..even].chunks_exact(2) {
+                if out_units.len() < capacity_elements {
+                    out_units.push(u16::from_le_bytes([pair[0], pair[1]]));
+                } else {
+                    truncated = true;
+                }
+            }
+        } else if transcode {
+            let utf8 = utf16le_chunk_to_utf8(
```
shiwanigupta0809, line 1200: **Suggestion:** once the output buffer is full, the remaining chunks are still fully processed, which contradicts the "costs the buffer, not the value" claim in the doc comment and the PR description.

The memory claim holds — nothing larger than a chunk is retained. The *CPU* claim does not. After `truncated` is set the loop still has to drain the wire, but it also keeps doing the per-chunk work:

- Transcode branch (line 736): `utf16le_chunk_to_utf8` runs on every remaining chunk, allocating a `Vec<u16>` and a `String` per 8 KiB and decoding the whole value. A 2 GB `nvarchar(max)` delivered into a 32-byte buffer transcodes all 2 GB to UTF-8 and discards it.
- Wide branch (line 728): worse, because it has no `break`. Once `out_units.len() == capacity_elements` the `chunks_exact(2)` loop still runs to completion on every chunk, setting `truncated = true` a billion times. Both byte-copy branches do break; this one looks like an oversight rather than a decision.

A `if truncated { if chunk.reached_end { break } else { continue } }` right after the read — or hoisting to a `drain_plp_to_end` tail call once the buffer is full — makes the drain cost proportional to the wire read alone. `drain_plp_to_end` is already sitting there.

Minor, same area: `scratch` is a fresh 8 KiB `vec![0u8; ..]` per PLP column per row, and `drain_plp_to_end` allocates its own `sink`. For a wide rowset that is one zeroed 8 KiB allocation per cell. Reusing a single buffer across the fill loop would be a small change.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
```diff
@@ -563,3 +561,115 @@ TEST_F(FetchScrollLiveTest, AStaleOrdinalPastTheResultSetIsIgnored) {
...
+    unsigned char buf[32] = {};
+    SQLLEN ind = 0;
+    ASSERT_SQL_OK(SQLBindCol(stmt_, 1, SQL_C_BINARY, buf, sizeof(buf), &ind), SQL_HANDLE_STMT,
+                  stmt_);
+    EXPECT_EQ(SQL_ERROR, SQLFetch(stmt_));
+    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "HYC00");
+
+    // The stream still has to be drained even when the target is refused, or
+    // the next row would decode from inside this value.
+    EXPECT_EQ(SQL_NO_DATA, SQLFetch(stmt_));
```
shiwanigupta0809, line 797: **Nit:** this does not assert what its comment claims.

The comment says "the next row would decode from inside this value", but the query returns a single row, so `SQL_NO_DATA` only proves the row *ended* — the same thing a driver that abandoned the stream and reset the cursor would report. `ABoundLobIsDrainedAcrossARowset` does this properly for the supported path; the refused-target path, which is a different branch (`drain_plp_to_end` rather than the fill loop), gets the weaker version. Two rows plus a trailing scalar column, as in `ABoundLobColumnDoesNotDesyncTheRow`, would make it say what it means.

Also, while you are in the file: `ABoundNvarcharMaxTruncatedToWcharReportsFullLength` (line 608) asserts only the indicator and never looks at `buf`. It is the sole wide-target truncation case, so a regression that reports 10000 while writing nothing — or writing the wrong 15 code units — passes. An `EXPECT_EQ(15u, wcslen-equivalent)` plus a spot check of the first unit would close it, and it is the test that would have caught the surrogate-split issue from the other thread.

shiwanigupta0809 (commented):
Reviewed against the `SQLGetData` PLP path this shares helpers with, `PlpChunk`/`plp_encoding()` in `mssql-tds`, `copy_with_nul`, and `deliver_bound`'s existing char handling. The indicator matrix is right — I re-derived all four rows and they hold, including the `Utf8Text` (json) case that has no test — and `capacity_elements = buf_elements - 1` lines up exactly with `copy_with_nul`'s `buf_len - 1`, so there is no double truncation. The compatibility gate is a faithful copy of `stream_active_plp_chunk`'s.

One finding I would like addressed before merge: the truncation boundary is applied in raw output units, so a multi-byte UTF-8 sequence or a surrogate pair straddling it is cut in half. Every truncation test uses ASCII, so the suite cannot see it. Details inline.

The rest are a suggestion, a perf point that bears on the PR's own design claim, and a test nit. I have not re-filed anything from the three open Copilot threads (`out_bytes` growth, the swallowed error at line 713, `plp_encoding_for`'s `.ok()?`).

**Blocking:** one (character-boundary truncation).
**Suggestion:** two (drain error swallowed on the refused-target path; work continues after the buffer is full).
**Nit:** one (e2e assertions).

shiwanigupta0809:
Heads up on an overlap. #396 (AB#47506) adds narrow→wide PLP widening to `SQLGetData`: `varchar(max)` and `json` delivered as `SQL_C_WCHAR` are decoded through the column's collation instead of being rejected.

That's the pairing your `deliver_bound_plp` gate excludes:

```rust
// Binary delivery is still unimplemented (AB#47239), and
// varchar -> SQL_C_WCHAR widening with it.
```

True against `main` today. With both in, the same column and target type answer differently depending on the API — `SQLGetData` delivers it, `SQLBindCol` returns `HYC00`. `SQL_C_WCHAR` is mssql-python's default for character columns, so that's the path most callers hit.

This one shouldn't go out as a tracked divergence — it's the kind of split that gets found by a user rather than by us. Better fixed in whichever PR lands second.

#396 is green and waiting on approval, so the cheaper sequence is that one first, then 427 picks it up:

- add `(SQL_C_WCHAR, SingleByteText | Utf8Text)` to the `compatible` gate
- build the decoder from the column's collation — the `narrow_to_wide` block in `stream_active_plp_chunk` does exactly this, including the `json`-has-no-collation case, and is reusable as-is
- decode in the chunk loop instead of copying verbatim

No carry needed on your side: a bound buffer is filled once, so there's no continuation to hold surplus units for.

Separately, in the same function: `deliver_bound_plp` re-implements the `pending_byte` / `pending_high_surrogate` carry that `ActivePlpStream` already holds. Your indicator rule matches ours exactly — which is a good sign both were characterized independently — but two hand-maintained copies of surrogate and split-code-unit handling will drift. Worth collapsing to one while the second copy is still new.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+                &scratch[..chunk.read],
+                chunk.reached_end,
+                &mut pending_byte,
+                &mut pending_high_surrogate,
+            );
+            for b in utf8.as_bytes() {
+                if out_bytes.len() < capacity_elements {
```
Theekshna, line 1220: Truncating byte-wise can leave a partial UTF-8 sequence in the caller's buffer. An `é` whose capacity ends mid-sequence yields `0xC3` followed by the NUL.

msodbcsql truncates on a character boundary instead — `sqlccnvt.cpp` calls `SystemLocale::TrimPartialCodePt()`, and a remainder that is *entirely* a partial character is not written to the user buffer at all.

The `SQLGetData` path never hits this because it pre-sizes `max_read` to `(cap / 3) * 2` so a chunk always fits. Capacity truncation is new here, so the chunk-boundary handling reused from there doesn't cover it.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+            if let Some(b) = pending_byte.take() {
+                bytes.push(b);
+            }
+            bytes.extend_from_slice(&scratch[..chunk.read]);
+            let even = bytes.len() & !1;
+            if even != bytes.len() {
+                pending_byte = Some(bytes[even]);
+            }
+            for pair in bytes[..even].chunks_exact(2) {
+                if out_units.len() < capacity_elements {
```
Theekshna: Truncation at a code-unit boundary can split a surrogate pair and leave an unpaired high surrogate as the last unit in the buffer.

msodbcsql explicitly trims it — `GetColDataSurrogateSafe` in `sqlcdata.cpp`: "if that character is a low-surrogate, then we can't fit the next character in the space available and we thus trim the high-surrogate from the user-buffer".

mssql-odbc/src/api/fetch_scroll.rs
Theekshna: A wire failure here becomes a row-level `07006` and leaves `fetch_error` as `None`, so `fill_rowset` keeps fetching on a connection that just failed and skips `reset_row_stream()` / `STMT_STATE_CURSOR_OPEN` teardown. The `read_row_column` `Err` arm at the call site does set `fetch_error`.

Worst on timeout/cancel, where `cancel_read_stream_and_wait` has already run. msodbcsql marks the statement `LOG_DISCARD` and fails the fetch rather than the row.

Returning the `TdsError` so the caller can set `fetch_error` would keep both arms consistent.

mssql-odbc/src/api/fetch_scroll.rs
Theekshna: Same shape as the read error below: a failed drain is dropped, so the row reports `HYC00` and the loop continues on a stream that may now be desynced.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+                .add(row_index)
+        }
+    };
+    let slot =
+        unsafe { (binding.target_value_ptr as *mut u8).add(bind_offset + row_index * stride) };
+
+    // Same pairings SQLGetData supports. Binary delivery is still unimplemented
+    // (AB#47239), and varchar -> SQL_C_WCHAR widening with it.
+    let target = binding.target_type;
+    let compatible = matches!(
```
Theekshna, line 1135: Two divergences fall out of this gate that the description doesn't mention:

- `varchar(max)` / `json` -> `SQL_C_WCHAR` answers `HYC00`; msodbcsql delivers it. Common bind for Unicode apps.
- Any max column bound to a *typed* C target (e.g. `SQL_C_SLONG`) also lands here, so `varchar(50)` and `varchar(max)` now behave differently for the same bind — `deliver_bound` routes the non-max case through `convert_typed_c`.

Both deserve an explicit callout and a work item. "0 divergences" across the parity suite suggests neither is covered by a test.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+///
+/// Unlike `SQLGetData`, which hands back one chunk per call and lets the caller
+/// come back for more, a bound column is filled once: everything past the
+/// buffer is discarded, but still read off the wire so the row stays in sync.
+/// Nothing larger than one chunk is ever held in memory, so a multi-gigabyte
+/// LOB delivered into a small buffer costs the buffer, not the value.
+///
+/// # Safety
+/// Same contract as `deliver_bound`.
+unsafe fn deliver_bound_plp(
```
Theekshna, line 1105: No Rust unit tests for any of this. All new coverage is C++ e2e needing a live server, so diff coverage on these ~180 lines is effectively 0 against the 85% target.

The indicator decision is pure: extracting `(produced_bytes, truncated, transcode, wire_total) -> SqlLen` would cover all four msodbcsql cases plus unknown-length in-process. `character_data_is_truncated_to_the_bound_buffer` is the precedent.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
```diff
@@ -563,3 +561,115 @@ TEST_F(FetchScrollLiveTest, AStaleOrdinalPastTheResultSetIsIgnored) {
...
+    EXPECT_EQ(SQL_SUCCESS, SQLFetch(stmt_));
+    EXPECT_STREQ("abcdefghij", buf);
+    EXPECT_EQ(10, ind);
+    SQLFreeStmt(stmt_, SQL_UNBIND);
+    SQLCloseCursor(stmt_);
+}
+
+// Transcoding UTF-16 to a narrow target means the wire byte count is not the
+// delivered byte count, so the full length cannot be reported.
+TEST_F(FetchScrollLiveTest, ABoundNvarcharMaxTruncatedToCharReportsNoTotal) {
```
Theekshna, line 599: Both truncation tests use ASCII (`REPLICATE(N'x')`) and the only non-ASCII test fits, so neither the partial-UTF-8 nor the split-surrogate boundary is exercised. A truncated non-ASCII value on both targets — and a surrogate-pair source such as `NCHAR(0xD83D) + NCHAR(0xDE00)` — would catch the two issues flagged in `fetch_scroll.rs`.

Theekshna (commented):
Reviewed the delivery path against msodbcsql (`sqlcdata.h/.cpp`, `sqlccnvt.cpp`).

The indicator rule checks out on all four documented cases, including unknown-length PLP -> `SQL_NO_TOTAL`, and the `BufferLength == 0` probe matches. NULL max columns correctly never reach this path (`PlpColumnStream::begin` returns `None`), `wire_total` from the final chunk is valid because the `PlpPaused` state survives `reached_end`, and the loop can't spin since `read_into` only returns 0 at end of stream. Draining a refused target matches msodbcsql's `FlushData`.

Two parity defects at the truncation boundary though, plus an error-path concern. Details inline.

One more that's outside the diff: `ABoundLobIsDrainedAcrossARowset` (line ~449) still accepts `SUCCESS || SUCCESS_WITH_INFO || ERROR` and checks neither content nor per-row indicators. Same reasoning you applied to `ABoundLobColumnDoesNotDesyncTheRow` — worth tightening now that block-path delivery actually works.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: You are right, and this was a real bug. Fixed in `92a18e4b`.

Confirmed both halves of your reading: `capacity_elements` was counting raw output units, so 15 e-acutes filled 30 of 31 bytes and the 16th contributed only its lead byte `0xC3` — the caller got 31 bytes ending mid-sequence, NUL-terminated after the dangling lead. The wide branch had the identical shape with surrogate pairs.

Both now stop at the last whole character. The transcode branch iterates `chars()` and only pushes one whose `len_utf8()` still fits; the wide branch refuses a high surrogate without room for its low half. `truncated` and the reported indicator are unchanged, as you suggested.

The part worth calling out is **why the existing tests could not have caught it**, which is exactly as you diagnosed: the two truncation tests used ASCII and the non-ASCII test used a 64-byte buffer for an 8-byte value, so truncation and non-ASCII never intersected. That is the gap, not the arithmetic. Added both cases — `REPLICATE(NCHAR(233), 5000)` into `char buf[32]`, asserting 30 bytes and every pair well-formed, and a supplementary-plane source (`NCHAR(0xD83D) + NCHAR(0xDE00)`) into a 10-element buffer asserting 8 units and no unpaired surrogate. Both fail against the old code by construction.

Your point that the shared-machinery argument does not carry over was the key one: `SQLGetData` pre-sizes `max_read` so a chunk always fits and even `debug_assert!`s it, so bound delivery is the first caller that can truncate a transcoded chunk at all.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: You are right, and this was a real bug. Fixed in `92a18e4b`.

Confirmed both halves of your reading: `capacity_elements` was counting raw output units, so 15 e-acutes filled 30 of 31 bytes and the 16th contributed only its lead byte `0xC3` — the caller got 31 bytes ending mid-sequence, NUL-terminated after the dangling lead. The wide branch had the identical shape with surrogate pairs.

Both now stop at the last whole character. The transcode branch iterates `chars()` and only pushes one whose `len_utf8()` still fits; the wide branch refuses a high surrogate without room for its low half. `truncated` and the reported indicator are unchanged, as you suggested.

The part worth calling out is **why the existing tests could not have caught it**, which is exactly as you diagnosed: the two truncation tests used ASCII and the non-ASCII test used a 64-byte buffer for an 8-byte value, so truncation and non-ASCII never intersected. That is the gap, not the arithmetic. Added both cases — `REPLICATE(NCHAR(233), 5000)` into `char buf[32]`, asserting 30 bytes and every pair well-formed, and a supplementary-plane source (`NCHAR(0xD83D) + NCHAR(0xDE00)`) into a 10-element buffer asserting 8 units and no unpaired surrogate. Both fail against the old code by construction.

Your point that the shared-machinery argument does not carry over was the key one: `SQLGetData` pre-sizes `max_read` so a chunk always fits and even `debug_assert!`s it, so bound delivery is the first caller that can truncate a transcoded chunk at all.

Thank you for the  citation — that confirmed the intended behaviour is trimming rather than best-effort, and that a remainder which is entirely a partial character is not written at all.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: You are right, and this was a real bug. Fixed in `92a18e4b`.

Confirmed both halves of your reading: `capacity_elements` was counting raw output units, so 15 e-acutes filled 30 of 31 bytes and the 16th contributed only its lead byte `0xC3` — the caller got 31 bytes ending mid-sequence, NUL-terminated after the dangling lead. The wide branch had the identical shape with surrogate pairs.

Both now stop at the last whole character. The transcode branch iterates `chars()` and only pushes one whose `len_utf8()` still fits; the wide branch refuses a high surrogate without room for its low half. `truncated` and the reported indicator are unchanged, as you suggested.

The part worth calling out is **why the existing tests could not have caught it**, which is exactly as you diagnosed: the two truncation tests used ASCII and the non-ASCII test used a 64-byte buffer for an 8-byte value, so truncation and non-ASCII never intersected. That is the gap, not the arithmetic. Added both cases — `REPLICATE(NCHAR(233), 5000)` into `char buf[32]`, asserting 30 bytes and every pair well-formed, and a supplementary-plane source (`NCHAR(0xD83D) + NCHAR(0xDE00)`) into a 10-element buffer asserting 8 units and no unpaired surrogate. Both fail against the old code by construction.

Your point that the shared-machinery argument does not carry over was the key one: `SQLGetData` pre-sizes `max_read` so a chunk always fits and even `debug_assert!`s it, so bound delivery is the first caller that can truncate a transcoded chunk at all.

The `GetColDataSurrogateSafe` citation is the one I would not have found, and it settles that msodbcsql trims the high surrogate rather than emitting it — so trimming is parity, not just defensiveness.

David-Engel: You are right, and this was a real bug — fixed in `92a18e4b`.

Truncation was counting raw output units, so a surrogate pair straddling the capacity boundary left an unpaired high surrogate as the last unit. The wide branch now refuses a high surrogate that has no room for its low half, and the transcode branch iterates `chars()` so a multi-byte UTF-8 sequence is never cut either.

The `GetColDataSurrogateSafe` citation is the one I would not have found, and it settles the question: msodbcsql *trims* the high surrogate rather than emitting it, so trimming is parity rather than just defensiveness.

Added a supplementary-plane test — `NCHAR(0xD83D) + NCHAR(0xDE00)` repeated into a 10-element buffer — asserting 8 units delivered and every pair well-formed. It fails against the old code by construction.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+    let mut pending_high_surrogate: Option<u16> = None;
+    let mut truncated = false;
+    let mut wire_total: Option<u64>;
+    let mut scratch = vec![0u8; PLP_BOUND_CHUNK];
+
+    loop {
+        let chunk = match runtime.block_on(client.read_active_plp_chunk(&mut scratch)) {
+            Ok(c) => c,
+            Err(_) => return RowOutcome::Error(RowIssue::Restricted),
+        };
```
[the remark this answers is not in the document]
David-Engel: Agreed, and fixed in `92a18e4b`.

`deliver_bound_plp` now returns `Result<RowOutcome, TdsError>`, and the call site sets `fetch_error` on `Err` exactly as the adjacent `read_row_column` arm does. So a transport drop, timeout, cancellation or protocol error fails the *fetch* — breaking the row loop and running the teardown — instead of being relabelled as a row-level `07006` on a connection that has already failed.

The detail that made this worth doing rather than deferring is the one about `cancel_read_stream_and_wait` having already run on timeout/cancel: reporting a row error there would have kept `fill_rowset` pulling rows from a stream that is gone.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Agreed, and fixed in `92a18e4b`.

`deliver_bound_plp` now returns `Result<RowOutcome, TdsError>`, and the call site sets `fetch_error` on `Err` exactly as the adjacent `read_row_column` arm does. So a transport drop, timeout, cancellation or protocol error fails the *fetch* — breaking the row loop and running the teardown — instead of being relabelled as a row-level `07006` on a connection that has already failed.

The detail that made this worth doing rather than deferring is the one about `cancel_read_stream_and_wait` having already run on timeout/cancel: reporting a row error there would have kept `fill_rowset` pulling rows from a stream that is gone.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Agreed — and you were right that this one is worse than the read-error case, so it is fixed the same way in `92a18e4b`.

`drain_plp_to_end` now returns its error and the caller propagates it into `fetch_error`, so a partial drain stops the fetch rather than letting the loop advance over a cursor still positioned inside the LOB. That is precisely the corruption `ABoundLobColumnDoesNotDesyncTheRow` was written to prevent, and as you noted it would have surfaced as damage to the *following* column, which is much harder to attribute than a mislabelled SQLSTATE.

Your caveat about the resolution differing here was the important part: when the drain *succeeds*, the column still reports `RowIssue::Unsupported`, because `HYC00` is the deliberate answer for a refused binary target. Only a genuine drain failure escalates to a fetch error.

I also strengthened `ABoundVarbinaryMaxIsStillUnsupported` per the other thread, so the refused-target drain path is now covered by two rows and a trailing scalar column rather than a single row.

David-Engel: Agreed — and you were right that this one is worse than the read-error case, so it is fixed the same way in `92a18e4b`.

`drain_plp_to_end` now returns its error and the caller propagates it into `fetch_error`, so a partial drain stops the fetch rather than letting the loop advance over a cursor still positioned inside the LOB. That is precisely the corruption `ABoundLobColumnDoesNotDesyncTheRow` was written to prevent, and as you noted it would have surfaced as damage to the *following* column, which is much harder to attribute than a mislabelled SQLSTATE.

Your caveat about the resolution differing here was the important part: when the drain *succeeds*, the column still reports `RowIssue::Unsupported`, because `HYC00` is the deliberate answer for a refused binary target. Only a genuine drain failure escalates to a fetch error.

I also strengthened `ABoundVarbinaryMaxIsStillUnsupported` per the other thread, so the refused-target drain path is now covered by two rows and a trailing scalar column rather than a single row.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
 
+/// Wire read size when draining a bound PLP column. Bounds peak memory: the
+/// value itself is never materialized, only one chunk at a time.
+const PLP_BOUND_CHUNK: usize = 8 * 1024;
+
+/// Reads the PLP encoding declared for `column` (1-based) in the current
+/// result set.
+fn plp_encoding_for(stmt: &StmtHandle, column: usize) -> Option<PlpEncoding> {
+    let ss = stmt.inner.lock().ok()?;
+    ss.column_metadata.get(column - 1)?.plp_encoding()
```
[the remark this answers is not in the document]
David-Engel: Good catch — fixed in `92a18e4b`.

The encodings are now snapshotted once into a `Vec<Option<PlpEncoding>>` under the statement lock taken before the fill loop, and the loop indexes that. So a poisoned mutex returns `SQL_ERROR` immediately rather than being silently indistinguishable from "this column is not PLP" — which, as you say, would have downgraded a supported column to unsupported and drained it before the fetch failed for some unrelated reason.

Snapshotting also removes a per-column lock acquisition from the inner loop, so it is cheaper as well as correct.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -623,6 +634,179 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
...
+        return RowOutcome::Error(RowIssue::Unsupported);
+    }
+
+    let transcode = target == SQL_C_CHAR && matches!(encoding, Some(PlpEncoding::Utf16Text));
+    let buf_elements = char_buf_elements(target, stride);
+    // Room for the payload, less the terminator the copy always writes.
+    let capacity_elements = buf_elements.saturating_sub(1);
+
+    let mut out_bytes: Vec<u8> = Vec::new();
+    let mut out_units: Vec<u16> = Vec::new();
```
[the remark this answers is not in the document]
David-Engel: Fair, and the wording in the doc comment overstated it — corrected in `92a18e4b`.

To be precise about what was and was not true: peak memory is bounded by the caller's own bound buffer plus one chunk, never by the value, so a 2 GB LOB into a 32-byte buffer never allocated 2 GB. But "retains only one chunk" was wrong as written, and the accumulate-then-copy step is a real duplicate of the delivered prefix. The comment now says what it does rather than what I meant.

The CPU half of your point was the substantive one and is fixed: once the buffer is full the loop reads the remaining chunks but no longer decodes them, so a truncated 2 GB `nvarchar(max)` no longer transcodes 2 GB of UTF-16 to discard it. The wide branch also lacked the `break` its siblings had, which was an oversight rather than a decision, as you guessed — it now short-circuits with the rest.

Also took the allocation point: one `scratch` buffer is created per fill and shared, rather than a fresh zeroed 8 KiB per PLP cell, and `drain_plp_to_end` borrows it instead of allocating its own sink.

I kept the accumulate-into-`Vec`-then-`copy_with_nul` shape rather than writing directly into `slot` with a running offset. It keeps a single NUL-termination path shared with `deliver_bound`, and the buffer is the caller's declared size — writing incrementally into the raw pointer would trade a bounded, short-lived allocation for more unsafe pointer arithmetic in a function that already has plenty. Happy to revisit if you disagree.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Fair, and the wording in the doc comment overstated it — corrected in `92a18e4b`.

To be precise about what was and was not true: peak memory is bounded by the caller's own bound buffer plus one chunk, never by the value, so a 2 GB LOB into a 32-byte buffer never allocated 2 GB. But "retains only one chunk" was wrong as written, and the accumulate-then-copy step is a real duplicate of the delivered prefix. The comment now says what it does rather than what I meant.

The CPU half of your point was the substantive one and is fixed: once the buffer is full the loop reads the remaining chunks but no longer decodes them, so a truncated 2 GB `nvarchar(max)` no longer transcodes 2 GB of UTF-16 to discard it. The wide branch also lacked the `break` its siblings had, which was an oversight rather than a decision, as you guessed — it now short-circuits with the rest.

Also took the allocation point: one `scratch` buffer is created per fill and shared, rather than a fresh zeroed 8 KiB per PLP cell, and `drain_plp_to_end` borrows it instead of allocating its own sink.

I kept the accumulate-into-`Vec`-then-`copy_with_nul` shape rather than writing directly into `slot` with a running offset. It keeps a single NUL-termination path shared with `deliver_bound`, and the buffer is the caller's declared size — writing incrementally into the raw pointer would trade a bounded, short-lived allocation for more unsafe pointer arithmetic in a function that already has plenty. Happy to revisit if you disagree.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
David-Engel: Both taken, in `92a18e4b`.

**The refused-target drain test.** You are right that a single-row query made `SQL_NO_DATA` prove only that the row ended — which a driver that abandoned the stream and reset the cursor would also report, i.e. the exact failure mode the test was supposed to exclude. It now uses two rows with a trailing scalar column, so it asserts the value *after* the refused LOB decodes (`tail == 11`), then that the second row decodes correctly (`n == 2`, `tail == 22`), and only then that the result set ends.

**The wide truncation test.** Also right, and it was the more consequential of the two: asserting only the indicator meant a regression reporting 10000 while writing nothing would have passed — and, as you predicted, it is the test that should have caught the surrogate-split issue from the other thread. It now checks 15 units delivered and spot-checks the first and last.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Agreed — added in `92a18e4b`.

The indicator decision is now a free function `plp_indicator(produced_bytes, truncated, transcoded, wire_total) -> SqlLen`, with `plp_indicator_matches_msodbcsql` covering all four cases I probed against msodbcsql plus the unknown-length one, in process and with no server.

That is the right call for more than the coverage number: the rule is the part of this change most likely to be "simplified" by someone who has not seen the probe output, and the `SQL_NO_TOTAL`-on-transcode branch in particular looks arbitrary until you know msodbcsql does the same. A unit test states it far better than the comment did.

The delivery loop itself still needs a live server, since it drives a real PLP stream off the wire — but the decision that was pure is now covered as one.

David-Engel: Both real, and neither was called out — thank you. Now recorded in the code at the gate and tracked as **AB#47767**, filed under the same parent.

Taking them separately:

- **`varchar(max)` / `json` → `SQL_C_WCHAR`** is refused because the gate admits only `(SQL_C_WCHAR, Utf16Text)`. msodbcsql widens it. As you say, it is a common bind for Unicode applications that bind every character column wide regardless of server type.
- **A max column into a typed C target** is the one I find harder to defend, for the reason you give: `varchar(50)` and `varchar(max)` now answer differently for an identical bind, because only the non-max path reaches `convert_typed_c`. The *size* of the value decides whether the conversion is attempted, which is not a distinction an application can see coming.

Your last point is the one I have carried into the work item verbatim: "0 divergences" across the parity suite is not evidence here, because neither pairing has a test. An untested divergence is indistinguishable from working code, so the item requires parity coverage for both rather than just the fix.

I have not fixed them in this PR — both need a materialize-then-convert step the streaming path deliberately lacks, and folding that in would widen a change that already grew a fair bit under review.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Both real, and neither was called out — thank you. Now recorded in the code at the gate and tracked as **AB#47767**, filed under the same parent.

Taking them separately:

- **`varchar(max)` / `json` → `SQL_C_WCHAR`** is refused because the gate admits only `(SQL_C_WCHAR, Utf16Text)`. msodbcsql widens it. As you say, it is a common bind for Unicode applications that bind every character column wide regardless of server type.
- **A max column into a typed C target** is the one I find harder to defend, for the reason you give: `varchar(50)` and `varchar(max)` now answer differently for an identical bind, because only the non-max path reaches `convert_typed_c`. The *size* of the value decides whether the conversion is attempted, which is not a distinction an application can see coming.

Your last point is the one I have carried into the work item verbatim: "0 divergences" across the parity suite is not evidence here, because neither pairing has a test. An untested divergence is indistinguishable from working code, so the item requires parity coverage for both rather than just the fix.

I have not fixed them in this PR — both need a materialize-then-convert step the streaming path deliberately lacks, and folding that in would widen a change that already grew a fair bit under review.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
David-Engel: Correct, and this is the gap that let the character-boundary bug through — both truncation tests used ASCII (`REPLICATE(N'x')`, `REPLICATE('y')`) and the only non-ASCII case used a 64-byte buffer for an 8-byte value, so truncation and non-ASCII never intersected.

Added both cases you asked for in `92a18e4b`:

- `ABoundNvarcharMaxTruncatesOnACharacterBoundary` — `REPLICATE(NCHAR(233), 5000)` into `char buf[32]`, asserting 30 bytes rather than 31 and every `0xC3 0xA9` pair intact.
- `ABoundNvarcharMaxDoesNotSplitASurrogatePair` — `NCHAR(0xD83D) + NCHAR(0xDE00)` repeated into a 10-element `SQLWCHAR` buffer, asserting 8 units and no unpaired surrogate.

Both fail against the old byte-wise truncation by construction, which is what makes them worth having rather than just passing.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Apologies — my previous reply on this thread was meant for the divergences thread and landed here by mistake. The response to *this* point:

Agreed, and added in `92a18e4b`. The indicator decision is now a free function `plp_indicator(produced_bytes, truncated, transcoded, wire_total) -> SqlLen`, with `plp_indicator_matches_msodbcsql` covering all four cases I probed against msodbcsql plus the unknown-length one, in process and with no server. `character_data_is_truncated_to_the_bound_buffer` was the right precedent to point at.

That is worth more than the coverage number. The rule is the part of this change most likely to be "simplified" by someone who has not seen the probe output — the `SQL_NO_TOTAL`-on-transcode branch looks arbitrary until you know msodbcsql does the same thing for the same reason. A unit test states that far better than my comment did.

The delivery loop itself still needs a live server, since it drives a real PLP stream off the wire, so diff coverage on this file will not reach 85% from unit tests alone. But the decision that was pure is now covered as one.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Apologies for the noise above — the unit-test reply on this thread was misrouted; the substantive response to the two divergences is the comment following it, and AB#47767 tracks them.

Theekshna (commented):
> [!NOTE]

## Summary

The bound-PLP delivery path is substantially improved on this head, but the earlier character-boundary blocker is only partially fixed and the current PR description does not record the final set of deliberate divergences.

## Blocking

The existing UTF-8 truncation thread remains applicable to `deliver_bound_plp`: the new `transcode` arm now retains whole characters, but the shared byte-verbatim arm at `mssql-odbc/src/api/fetch_scroll.rs:828-839` still handles `PlpEncoding::Utf8Text` and can stop after the first byte of a multi-byte JSON character. This is the same previously filed finding, so I am not opening a duplicate inline thread. Split `SingleByteText` and `Utf8Text`; keep byte-wise copying only for the former and append complete UTF-8 characters for the latter.

## Suggestion

The description is stale relative to the reviewed head: it still says five new e2e tests, omits the review-driven character-boundary and wire-error changes, and lists only bound binary under deliberate exclusions even though the code records two additional AB#47767 divergences. Please update the PR design text so the final behavior and deliberate departures are recorded where reviewers will find them.

**msodbcsql parity:** The indicator rules remain aligned; `Sql\Ntdbms\sqlncli\odbc\sqlccnvt.cpp:950-971` applies character-boundary trimming. The remaining JSON case is not asserted as reference-driver parity because this reference tree predates that type.
**Test sufficiency, unit and e2e:** The existing boundary tests cover UTF-16-to-UTF-8 and surrogate truncation, but no test reaches `Utf8Text` to `SQL_C_CHAR`. Local mutation execution was blocked by the pinned compiler's `crc32fast` codegen failure, so the missing JSON case is reported from branch tracing rather than a mutation run.
**Divergences documented:** Not fully; the AB#47767 departures are recorded in a code comment and prior thread, but not in `.github/instructions/mssql-odbc.instructions.md` or the PR's design text.
**PR description currency:** Stale as described above. The linked AB#47361 is present; substantive validation passed, while `coverage-report` timed out.

Theekshna (approved):

PR #432, opened 2026-08-31
mssql-tds: resolve TCP connect hostnames asynchronously

Vahid-b:
## Description

Investigated [AB#47704](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47704) ("Failed invalid-password connection hangs test process during cleanup"). Built two live reproductions of a real SQL Server's invalid-password response (plaintext and over TLS, matching `Encrypt=yes`) — `mssql-tds`'s login-failure handling (ERROR token parsing, `is_transient_connect_error` classification, retry logic) is correct and returns in ~150ms in isolation, so that path is not the cause.

Root cause: `create_base_stream_sequential` (sequential TCP) and `parallel_connect` (`MultiSubnetFailover`) resolved the server hostname with `std::net::ToSocketAddrs::to_socket_addrs()` — a blocking OS call — directly inside async code, never wrapped in `spawn_blocking` (unlike `named_pipes.rs`/`interactive.rs`, which already do this correctly for LocalDB). Because it never `.await`s, it silently defeats the `tokio::time::timeout()`/deadline that wraps the whole connect attempt in `tds_connection_provider`: a `timeout()` can only fire between a future's `.await` points, and a thread stuck inside a genuinely blocking call never yields one. `mssql-odbc`'s `SQLDriverConnectW` drives the connect via `block_on` on the *calling* thread, so a slow/unresponsive resolver hangs that thread — and since ODBC is a blocking API, the whole synchronous caller — with **no internal bound at all**, independent of whether the credentials are valid. This is compounded by `mssql-python` allocating its ODBC environment as a process-wide static singleton, so every connection in a test file shares `mssql-odbc`'s one `Arc<Runtime>` (`worker_threads(1)`).

Fix: switch both call sites to `tokio::net::lookup_host`, the same async primitive `ssrp.rs` already uses for instance lookups, so resolution is a genuine `.await` point subject to the existing deadline machinery. Extracted the IP-preference sort into a small `sort_by_ip_preference` helper (previously inline, reassigning the iterator variable) since it had no direct test coverage and moving it was the natural seam.

Separately noted but **not** addressed here (out of scope, filed as follow-up candidates): `Connection Timeout=`/`Login Timeout=` connection-string keywords aren't parsed/wired to `ClientContext` at all (only `SQL_ATTR_LOGIN_TIMEOUT` works) — doesn't cause unbounded hangs since the 15s built-in default still applies, but silently ignores an explicit caller request. Also found two pre-existing, unrelated hangs in `connection::tds_client::tests` (`check_and_reconnect_attempts_reconnect_when_dead_and_recoverable`, `reconnect_returns_session_recovery_failed_on_connection_failure`) that reproduce identically on `main` — a likely-similar-class bug in the reconnect/session-recovery path, distinct code from this fix.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47704

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes (full workspace)
- [x] `cargo btest` — targeted: all `mssql-tds` unit tests pass except two pre-existing hangs confirmed unrelated (reproduce identically on `main`, see above); all `mssql-odbc` unit tests pass (1097/1097); mock-server/TLS/fedauth/redirection integration tests pass (2001/2001). Live-server integration tests (`connectivity`, `test_transport_protocols`, `test_no_protocol_resolution`, etc.) fail in this sandbox on `SQL_PASSWORD environment variable not set` both before and after this change — no live SQL Server configured locally, matching prior PRs' own notes (e.g. #414).
- [x] New/changed functionality has tests — `sort_by_ip_preference` unit tests (platform-default/IPv4-first/IPv6-first/single-family) plus a test pinning that `create_base_stream_sequential`'s resolution stays inside an outer `tokio::time::timeout`.
- [x] Public API changes are documented — none; both changed functions are crate-private. `CHANGELOG.md` updated under `Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,3 +150,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   rejected as a protocol error. Such a packet is malformed — it neither carries
   payload nor terminates a message — but was previously consumed as a
   zero-length packet. Empty end-of-message packets remain legal.
+
+- `mssql-tds`: TCP connect no longer resolves the server hostname with the
+  blocking `std::net::ToSocketAddrs`. That call never yields to the async
+  runtime, so a slow or unresponsive resolver silently escaped the
+  `ConnectTimeout`/`LoginTimeout` deadline that wraps the rest of the connect
+  sequence — on `mssql-odbc`, whose `SQLDriverConnectW` drives this via
+  `block_on` on the calling thread, this could hang the caller (and, since
+  ODBC is a blocking API, the whole calling process) indefinitely instead of
+  failing within the configured timeout. Resolution now goes through
+  `tokio::net::lookup_host`, the same async primitive already used for SSRP
+  instance lookups, so it is properly bounded like the rest of the connect
+  attempt.
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -160,6 +160,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
   ODBC is a blocking API, the whole calling process) indefinitely instead of
   failing within the configured timeout. Resolution now goes through
   `tokio::net::lookup_host`, the same async primitive already used for SSRP
-  instance lookups, so it is properly bounded like the rest of the connect
-  attempt.
+  instance lookups, so it is a genuine, cancellable `.await` point instead of
+  a blocking one. `parallel_connect` (`MultiSubnetFailover`) now also folds
+  DNS resolution into the single overall deadline its `timeout_ms` already
+  documents, instead of only timing the connection race that follows it, and
+  idle-connection reconnect (`TdsClient::reconnect`) now wraps each attempt's
+  full connect (DNS through login) in the attempt's remaining budget instead
+  of only capping the post-resolution TCP connect step.
 
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -474,9 +474,11 @@ impl TdsClient {
 
     /// Attempt to reconnect a dead connection by replaying session state.
     ///
-    /// The overall reconnection is bounded by `timeout`. Each individual
-    /// TCP/TDS handshake attempt uses the original `connect_timeout`. Before
-    /// each retry sleep, we verify enough time remains for the interval.
+    /// The overall reconnection is bounded by `timeout`. Each attempt (DNS
+    /// resolution through login) is wrapped in whatever remains of that
+    /// budget, and `connect_timeout` is capped to the same remaining budget
+    /// so it can't ask for more than is left. Before each retry sleep, we
+    /// verify enough time remains for the interval.
     #[instrument(skip(self), level = "info")]
     pub(crate) async fn reconnect(
         &mut self,
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -533,20 +533,37 @@ impl TdsClient {
             // Inject recovery data into the client context clone
             let mut reconnect_ctx = client_context.clone();
 
-            // Cap the per-attempt connect timeout to the remaining reconnect budget
-            let remaining_secs =
-                deadline.saturating_duration_since(Instant::now()).as_secs() as u32;
-            reconnect_ctx.connect_timeout = reconnect_ctx.connect_timeout.min(remaining_secs);
+            // Cap the per-attempt connect timeout to the remaining reconnect budget.
+            let remaining = deadline.saturating_duration_since(Instant::now());
+            reconnect_ctx.connect_timeout = reconnect_ctx
+                .connect_timeout
+                .min(remaining.as_secs() as u32);
 
             info!(attempt, "Attempting reconnection");
-            let connect_result = CancelHandle::run_until_cancelled(
-                cancel_handle,
-                TdsConnectionProvider::connect_with_transport_context(
-                    &reconnect_ctx,
-                    &transport_context,
-                    Some(Box::new((*snapshot).clone())),
-                ),
-            )
+            // `connect_timeout` above only bounds the TCP-connect step inside
+            // `connect_with_transport_context`; DNS resolution ahead of it is
+            // otherwise unbounded here. Wrap the whole attempt (DNS through
+            // login) in `remaining` too, so a slow resolver can't make a
+            // single reconnect attempt outlive the overall reconnect deadline.
+            let connect_result = CancelHandle::run_until_cancelled(cancel_handle, async {
+                match tokio::time::timeout(
+                    remaining,
+                    TdsConnectionProvider::connect_with_transport_context(
+                        &reconnect_ctx,
+                        &transport_context,
+                        Some(Box::new((*snapshot).clone())),
+                    ),
+                )
+                .await
+                {
+                    Ok(result) => result,
+                    Err(_elapsed) => {
+                        Err(Error::TimeoutError(crate::error::TimeoutErrorType::String(
+                            "Reconnection attempt timed out".to_string(),
+                        )))
+                    }
+                }
+            })
             .await;
             match connect_result {
                 Ok((
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -401,6 +401,26 @@ pub struct TdsClient {
     streamed_write_state: StreamedWriteState,
 }
 
+/// Runs `fut` bounded by `remaining`, converting an elapsed timeout into
+/// `Error::TimeoutError`. Extracted out of `TdsClient::reconnect` so the
+/// elapsed branch can be driven directly in a test (`tokio::time::pause` plus
+/// a future that never resolves) instead of needing a live, slow connect
+/// attempt to provoke it.
+async fn attempt_within<T>(
+    remaining: Duration,
+    fut: impl Future<Output = TdsResult<T>>,
+) -> TdsResult<T> {
+    tokio::time::timeout(remaining, fut)
+        .await
+        .unwrap_or_else(|_elapsed| {
+            Err(crate::error::Error::TimeoutError(
+                crate::error::TimeoutErrorType::String(
+                    "Reconnection attempt timed out".to_string(),
+                ),
+            ))
+        })
+}
+
 impl TdsClient {
     pub(crate) fn new(
         transport: AnyTransport,
@@ -534,10 +554,16 @@ impl TdsClient {
             let mut reconnect_ctx = client_context.clone();
 
             // Cap the per-attempt connect timeout to the remaining reconnect budget.
+            // `.max(1)`: `remaining` can be under a second this close to the deadline,
+            // and `as_secs()` truncates that to 0 — which `create_base_stream_sequential`
+            // has no "0 means default" fallback for, so it would time out the attempt
+            // on its first poll instead of giving it a real chance. The outer
+            // `attempt_within(remaining, ...)` below is the actual deadline backstop,
+            // so rounding up here can't make an attempt outlive the reconnect budget.
             let remaining = deadline.saturating_duration_since(Instant::now());
             reconnect_ctx.connect_timeout = reconnect_ctx
                 .connect_timeout
-                .min(remaining.as_secs() as u32);
+                .min(remaining.as_secs().max(1) as u32);
 
             info!(attempt, "Attempting reconnection");
             // `connect_timeout` above only bounds the TCP-connect step inside
@@ -545,25 +571,17 @@ impl TdsClient {
             // otherwise unbounded here. Wrap the whole attempt (DNS through
             // login) in `remaining` too, so a slow resolver can't make a
             // single reconnect attempt outlive the overall reconnect deadline.
-            let connect_result = CancelHandle::run_until_cancelled(cancel_handle, async {
-                match tokio::time::timeout(
+            let connect_result = CancelHandle::run_until_cancelled(
+                cancel_handle,
+                attempt_within(
                     remaining,
                     TdsConnectionProvider::connect_with_transport_context(
                         &reconnect_ctx,
                         &transport_context,
                         Some(Box::new((*snapshot).clone())),
                     ),
-                )
-                .await
-                {
-                    Ok(result) => result,
-                    Err(_elapsed) => {
-                        Err(Error::TimeoutError(crate::error::TimeoutErrorType::String(
-                            "Reconnection attempt timed out".to_string(),
-                        )))
-                    }
-                }
-            })
+                ),
+            )
             .await;
             match connect_result {
                 Ok((
@@ -8040,6 +8058,28 @@ mod tests {
             err.to_string().contains("Session recovery failed"),
             "Expected SessionRecoveryFailed, got: {err}"
         );
+    }
+
+    /// Covers the elapsed branch of `attempt_within` directly, since nothing
+    /// in `reconnect`'s own tests can provoke it: every failure a test can
+    /// reach today (no server, bad recovery state, ...) returns well before
+    /// any deadline. `start_paused` lets the timeout fire on a fake clock
+    /// instead of a real, slow connect attempt.
+    #[tokio::test(start_paused = true)]
+    async fn attempt_within_times_out_when_the_future_never_completes() {
+        let remaining = Duration::from_secs(5);
+        let handle = tokio::spawn(attempt_within(
+            remaining,
+            std::future::pending::<TdsResult<()>>(),
+        ));
+
+        tokio::time::advance(remaining + Duration::from_millis(1)).await;
+
+        let result = handle.await.expect("attempt_within task panicked");
+        match result {
+            Err(crate::error::Error::TimeoutError(_)) => {}
+            other => panic!("expected a TimeoutError once the deadline elapses, got {other:?}"),
+        }
     }
 
     #[tokio::test]
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -3658,26 +3658,22 @@ pub(crate) mod tests {
         assert_eq!(v4_only, expected, "no IPv6 entries to reorder against");
     }
 
-    /// Resolving a real hostname (not a literal IP, so the OS resolver is
-    /// actually exercised) must go through `tokio::net::lookup_host` and
-    /// therefore stay bounded by an enclosing `tokio::time::timeout` — the
-    /// exact property the blocking `to_socket_addrs()` call used to violate.
+    /// `tokio::net::lookup_host` must be used (not blocking `to_socket_addrs()`),
+    /// so a slow/stuck resolver stays bounded by an enclosing `timeout()`
+    /// instead of escaping it. Timing can't prove this — resolving `localhost`
+    /// completes in single-digit ms either way — so this asserts the
+    /// structural property instead: a concurrently spawned heartbeat must get
+    /// scheduled while resolution is in flight. `lookup_host` bridges to
+    /// `spawn_blocking` via a channel, so the awaiting task is guaranteed to
+    /// yield at least once; a blocking `to_socket_addrs()` call never yields,
+    /// so the heartbeat gets zero chances to run. Confirmed by mutation
+    /// testing (reverting to `to_socket_addrs()` makes this fail).
     ///
-    /// Timing alone can't prove this: resolving `localhost` completes in
-    /// low-single-digit milliseconds whether or not the underlying call
-    /// blocks, so a `timeout()`-based assertion would pass either way. This
-    /// test instead proves the structural property directly — a concurrently
-    /// spawned task must get *some* chance to run while resolution is in
-    /// flight. `tokio::net::lookup_host` bridges to `spawn_blocking` via a
-    /// channel, so the awaiting task is guaranteed to return `Pending` and
-    /// yield at least once, no matter how fast the lookup itself finishes.
-    /// A synchronous `to_socket_addrs()` call never yields at all: the whole
-    /// resolution runs inside a single poll of this task, so a concurrent
-    /// task gets zero opportunities to run until it is done. Confirmed by
-    /// mutation testing: reverting to `to_socket_addrs()` makes this test
-    /// fail (the heartbeat counter stays at 0) while the timing-based
-    /// version above kept passing regardless.
-    #[tokio::test]
+    /// Must stay on the default `current_thread` runtime: a `multi_thread`
+    /// flavor would let the heartbeat run on another worker even if
+    /// resolution blocked, so the assertion would pass without proving
+    /// anything.
+    #[tokio::test(flavor = "current_thread")]
     async fn create_base_stream_sequential_resolution_yields_to_the_executor() {
         use std::sync::atomic::{AtomicUsize, Ordering};
 
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -3683,10 +3683,14 @@ pub(crate) mod tests {
 
         let heartbeats = Arc::new(AtomicUsize::new(0));
         let heartbeats_task = heartbeats.clone();
+        // A short sleep (rather than `yield_now()`) still catches the same
+        // scheduling gap — the first increment can't happen until the main
+        // task yields either way — without busy-spinning a core for the
+        // whole resolution.
         let heartbeat = tokio::spawn(async move {
             loop {
                 heartbeats_task.fetch_add(1, Ordering::SeqCst);
-                tokio::task::yield_now().await;
+                tokio::time::sleep(Duration::from_millis(1)).await;
             }
         });
 
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -3662,30 +3662,53 @@ pub(crate) mod tests {
     /// actually exercised) must go through `tokio::net::lookup_host` and
     /// therefore stay bounded by an enclosing `tokio::time::timeout` — the
     /// exact property the blocking `to_socket_addrs()` call used to violate.
+    ///
+    /// Timing alone can't prove this: resolving `localhost` completes in
+    /// low-single-digit milliseconds whether or not the underlying call
+    /// blocks, so a `timeout()`-based assertion would pass either way. This
+    /// test instead proves the structural property directly — a concurrently
+    /// spawned task must get *some* chance to run while resolution is in
+    /// flight. `tokio::net::lookup_host` bridges to `spawn_blocking` via a
+    /// channel, so the awaiting task is guaranteed to return `Pending` and
+    /// yield at least once, no matter how fast the lookup itself finishes.
+    /// A synchronous `to_socket_addrs()` call never yields at all: the whole
+    /// resolution runs inside a single poll of this task, so a concurrent
+    /// task gets zero opportunities to run until it is done. Confirmed by
+    /// mutation testing: reverting to `to_socket_addrs()` makes this test
+    /// fail (the heartbeat counter stays at 0) while the timing-based
+    /// version above kept passing regardless.
     #[tokio::test]
-    async fn create_base_stream_sequential_resolution_is_bounded_by_timeout() {
+    async fn create_base_stream_sequential_resolution_yields_to_the_executor() {
+        use std::sync::atomic::{AtomicUsize, Ordering};
+
+        let heartbeats = Arc::new(AtomicUsize::new(0));
+        let heartbeats_task = heartbeats.clone();
+        let heartbeat = tokio::spawn(async move {
+            loop {
+                heartbeats_task.fetch_add(1, Ordering::SeqCst);
+                tokio::task::yield_now().await;
+            }
+        });
+
         // Port 0 is never listening, so the (fast, loopback) TCP connect
-        // fails quickly and the call returns well within the outer bound
-        // instead of the test hanging if resolution ever regresses to a
-        // blocking call.
-        let result = timeout(
-            Duration::from_secs(5),
-            create_base_stream_sequential(
-                IPAddressPreference::UsePlatformDefault,
-                "localhost",
-                0,
-                30_000,
-                1_000,
-                200,
-            ),
+        // fails quickly once resolution completes; the call overall still
+        // returns promptly either way.
+        let _ = create_base_stream_sequential(
+            IPAddressPreference::UsePlatformDefault,
+            "localhost",
+            0,
+            30_000,
+            1_000,
+            200,
         )
         .await;
 
+        heartbeat.abort();
         assert!(
-            result.is_ok(),
-            "resolving 'localhost' must not escape the outer timeout"
+            heartbeats.load(Ordering::SeqCst) > 0,
+            "the heartbeat task never ran while resolving 'localhost' — \
+             resolution is blocking the executor instead of awaiting it"
         );
-        assert!(result.unwrap().is_err(), "port 0 must not be connectable");
     }
 
     /// `skip_bytes` spins the same way, and it runs on the discard path that
--- a/mssql-tds/src/connection/transport/network_transport.rs
+++ b/mssql-tds/src/connection/transport/network_transport.rs
@@ -33,7 +33,7 @@ use byteorder::{BigEndian, ByteOrder, LittleEndian};
 use std::cmp::min;
 use std::io::Error;
 use std::io::ErrorKind;
-use std::net::ToSocketAddrs;
+use std::net::SocketAddr;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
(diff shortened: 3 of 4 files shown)


mssql-tds/src/connection/transport/parallel_connect.rs
```diff
@@ -125,9 +125,12 @@ pub async fn parallel_connect(
...
 
-    // Resolve DNS to get all IP addresses
-    let addresses: Vec<SocketAddr> = (host, port)
-        .to_socket_addrs()?
+    // Resolve DNS to get all IP addresses. `lookup_host` awaits the resolution
+    // instead of blocking the calling thread (unlike `std::net::ToSocketAddrs`),
+    // so a slow or stuck resolver stays subject to the caller's timeout instead
+    // of silently escaping it.
+    let addresses: Vec<SocketAddr> = tokio::net::lookup_host((host, port))
+        .await?
```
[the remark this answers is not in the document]
Vahid-b: Verified against the current head. `parallel_connect` is public and documents `timeout_ms` as the overall timeout, but `lookup_host(...).await` runs before `timeout(...)` is created. The connection-provider caller has an outer login deadline, but direct callers can still exceed this API contract during DNS resolution. Please create one deadline before lookup and use its remaining budget for both DNS and the connection race (for example, `timeout_at` around the entire operation).

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -3564,6 +3573,121 @@ pub(crate) mod tests {
...
+        // Port 0 is never listening, so the (fast, loopback) TCP connect
+        // fails quickly and the call returns well within the outer bound
+        // instead of the test hanging if resolution ever regresses to a
+        // blocking call.
+        let result = timeout(
+            Duration::from_secs(5),
+            create_base_stream_sequential(
+                IPAddressPreference::UsePlatformDefault,
+                "localhost",
+                0,
```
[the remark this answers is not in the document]
Vahid-b: Confirmed with a mutation test: I restored the old blocking `std::net::ToSocketAddrs` implementation and ran `cargo nextest run -p mssql-tds -E 'test(create_base_stream_sequential_resolution_is_bounded_by_timeout)'`; the test still passed. Resolving `localhost` quickly cannot distinguish the old implementation from `lookup_host`. Please inject a controllable resolver that remains pending beyond the deadline and assert that the timeout fires while resolution is pending; cover both sequential and parallel resolution paths.

CHANGELOG.md
```diff
@@ -150,3 +150,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
...
+- `mssql-tds`: TCP connect no longer resolves the server hostname with the
+  blocking `std::net::ToSocketAddrs`. That call never yields to the async
+  runtime, so a slow or unresponsive resolver silently escaped the
+  `ConnectTimeout`/`LoginTimeout` deadline that wraps the rest of the connect
+  sequence — on `mssql-odbc`, whose `SQLDriverConnectW` drives this via
+  `block_on` on the calling thread, this could hang the caller (and, since
+  ODBC is a blocking API, the whole calling process) indefinitely instead of
+  failing within the configured timeout. Resolution now goes through
+  `tokio::net::lookup_host`, the same async primitive already used for SSRP
+  instance lookups, so it is properly bounded like the rest of the connect
```
Vahid-b: Suggestion: this is true for the normal connection-provider path, but not for session-recovery reconnects. `TdsClient::reconnect` calls `connect_with_transport_context` without wrapping the whole future in its reconnect deadline; it only caps `connect_timeout`, which is applied after DNS. A pending `lookup_host` now yields and remains cancellable, but it can still outlive that deadline. Please either wrap the complete reconnect attempt in the remaining deadline or narrow this claim and track reconnect handling separately.

mssql-tds/src/connection/transport/parallel_connect.rs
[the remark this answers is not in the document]
Vahid-b: Good catch — fixed in 9726d239. `parallel_connect` now computes one `deadline` up front (`Instant::now() + Duration::from_millis(config.timeout_ms)`) and spends the remaining budget on both the `lookup_host` call and the connection race, instead of only timing the race. A DNS timeout now returns a distinct `TimeoutError` ("DNS resolution ... timed out") before the race ever starts.

> [!NOTE]

mssql-tds/src/connection/transport/network_transport.rs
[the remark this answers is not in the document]
Vahid-b: Confirmed — the mutation test showed the test was worthless. Replaced it (9726d239) with one that proves the structural property directly instead of timing: a concurrently spawned heartbeat task must get scheduled while resolution is in flight. `tokio::net::lookup_host` bridges to `spawn_blocking` via a channel, so the awaiting task is guaranteed to yield at least once no matter how fast the lookup finishes; a synchronous `to_socket_addrs()` call never yields at all, so the heartbeat counter stays at 0. Verified this actually discriminates by re-running the same mutation you did — it now fails against the blocking call and passes against `lookup_host`. Added the equivalent test for `parallel_connect` too (using a host that fails to resolve, so it doesn't reach `race_connections`, which spawns tasks and would otherwise mask the signal).

> [!NOTE]

CHANGELOG.md
Vahid-b: You're right, that was overclaiming. Fixed both the code and the wording in 9726d239: `TdsClient::reconnect` now wraps the whole `connect_with_transport_context` attempt (DNS through login) in the attempt's remaining budget via `tokio::time::timeout`, not just `connect_timeout` for the post-resolution TCP-connect step. Updated the CHANGELOG entry to describe this explicitly rather than the blanket "bounded like the rest of the connect attempt" claim.

> [!NOTE]

David-Engel (commented):
## Summary

Swaps the blocking `std::net::ToSocketAddrs` resolution in both TCP connect paths for `tokio::net::lookup_host`, folds DNS into `parallel_connect`'s documented `timeout_ms` budget, and wraps the whole reconnect attempt in its remaining deadline. The diagnosis holds up against the code: `connect_with_transport_context` is already wrapped in `timeout(remaining, ...)` by the provider, and a blocking `getaddrinfo` inside that future never produces the poll boundary the timer needs to fire on, so the deadline was real but unenforceable. No blocking findings.

**What I verified** (targeted `cargo nextest run -p mssql-tds --lib`, dedicated worktree and target dir):

- The four `sort_by_ip_preference` tests, both new `*_resolution_yields_to_the_executor` tests, and `reconnect_returns_session_recovery_failed_on_connection_failure` / `check_and_reconnect_attempts_reconnect_when_dead_and_recoverable` all pass here in well under a second.
- **Mutation test**: I restored `to_socket_addrs()` at both call sites and re-ran the two new tests — *both* fail with "the heartbeat task never ran while resolution was in flight". The replacement tests genuinely discriminate, unlike the timing-based version they replaced. Reverted afterwards.
- `Error::TimeoutError` is transient per `error/mod.rs:414`, so the new DNS-timeout error in `parallel_connect` is still retried by the connect loop rather than aborting it — no behavior change there.
- The `.local` host in the new `parallel_connect` test matches the existing `test_parallel_connect_invalid_host` at line 520, so it follows the file's convention rather than inventing one.

CI: every completed check passes; a few ADO jobs were still pending at review time.

### Blocking

None.

### Suggestions

Two inline, both on `tds_client.rs` and both non-blocking.

### Nits

None.

---

*Posted by an unattended automated review. These findings have not been checked by a human first.*

mssql-tds/src/connection/tds_client.rs
```diff
@@ -533,20 +533,37 @@ impl TdsClient {
...
+                    TdsConnectionProvider::connect_with_transport_context(
+                        &reconnect_ctx,
+                        &transport_context,
+                        Some(Box::new((*snapshot).clone())),
+                    ),
+                )
+                .await
+                {
+                    Ok(result) => result,
+                    Err(_elapsed) => {
```
David-Engel: The new `Err(_elapsed)` arm is the only part of this change nothing pins: the automated coverage report flags 552-555 as the sole missing lines in the diff, and there is no way to reach the timeout from a test today, since `lookup_host` isn't injectable and every failure a test can actually provoke here returns well before the deadline.

Cheapest seam if you want it covered: lift the wrap into a small helper that takes the future, so the elapsed branch can be driven with `tokio::time::pause()` and a `sleep` that never completes.

```rust
async fn attempt_within<T>(
    remaining: Duration,
    fut: impl Future<Output = TdsResult<T>>,
) -> TdsResult<T> {
    tokio::time::timeout(remaining, fut).await.unwrap_or_else(|_| {
        Err(Error::TimeoutError(TimeoutErrorType::String(
            "Reconnection attempt timed out".to_string(),
        )))
    })
}
```

Not merge-blocking — the wrap is three lines and reads as obviously correct — but it's the one new branch here with no coverage, and the helper is the difference between "obviously correct" and "pinned".

mssql-tds/src/connection/tds_client.rs
```diff
@@ -533,20 +533,37 @@ impl TdsClient {
...
 
-            // Cap the per-attempt connect timeout to the remaining reconnect budget
-            let remaining_secs =
-                deadline.saturating_duration_since(Instant::now()).as_secs() as u32;
-            reconnect_ctx.connect_timeout = reconnect_ctx.connect_timeout.min(remaining_secs);
+            // Cap the per-attempt connect timeout to the remaining reconnect budget.
+            let remaining = deadline.saturating_duration_since(Instant::now());
+            reconnect_ctx.connect_timeout = reconnect_ctx
+                .connect_timeout
+                .min(remaining.as_secs() as u32);
```
David-Engel: `as_secs()` truncates, so in the last sub-second of the reconnect budget this sets `connect_timeout = 0`, and `connect_with_transport_context` passes that straight through: `let connect_timeout_ms = (context.connect_timeout as u64) * 1000;` (`tds_connection_provider.rs:567`) has no "0 means default" fallback like the one at line 120. The result is `timeout(Duration::from_millis(0), connect_future)` in `create_base_stream_sequential`, so any connect not ready on its first poll fails instantly and the remaining budget is spent on nothing.

This predates the PR, but the outer `timeout(remaining, ...)` you just added is exactly what makes the fix safe — a rounded-up per-attempt timeout can no longer outlive the deadline:

```rust
reconnect_ctx.connect_timeout = reconnect_ctx
    .connect_timeout
    .min(remaining.as_secs().max(1) as u32);
```

The deadline check above guarantees `remaining > 0` at this point, so `.max(1)` only affects the truncated-to-zero case.

Related, and also pre-existing: if `client_context.connect_timeout` is itself `0` (which means "use the 15s default" at the `create_client` layer, per line 120), this `min` pins the reconnect path to 0 forever. Worth a follow-up rather than something to fold in here.

mssql-tds/src/connection/tds_client.rs
Vahid-b: Good catch — fixed in 556a3d59. Extracted a standalone `attempt_within` helper and added a dedicated test for the elapsed branch using `tokio::time::pause()` + `std::future::pending()`, since (as you noted) nothing reachable through `reconnect`'s own tests can provoke a real timeout.

> [!NOTE]

mssql-tds/src/connection/tds_client.rs
Vahid-b: Confirmed and fixed in 556a3d59 — `.min(remaining.as_secs().max(1) as u32)`. Agreed it's safe given the outer `attempt_within(remaining, ...)` is now the real deadline backstop. Left the `client_context.connect_timeout == 0` case as a follow-up rather than folding it in here, per your note.

> [!NOTE]

David-Engel (commented):
**Automated review** — generated by GitHub Copilot on behalf of @David-Engel. This is **not** an approval and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

## Summary

Re-review of the delta since `9726d239`. `attempt_within` is now a standalone helper with a paused-clock test that drives its elapsed branch directly, and the per-attempt cap became `remaining.as_secs().max(1)`, so the sub-second tail of the reconnect budget no longer collapses `connect_timeout` to `0` and fails an attempt on its first poll. Both changes answer the previous round correctly, and the outer `attempt_within(remaining, ...)` is what makes rounding the cap up safe — an attempt still cannot outlive the reconnect deadline. No blocking findings.

## Verification

- `cargo nextest run -p mssql-tds --lib -E 'test(sort_by_ip_preference) or test(yields_to_the_executor) or test(attempt_within) or test(reconnect)'` in a dedicated worktree and target dir: 40 passed, including `attempt_within_times_out_when_the_future_never_completes` and both `*_resolution_yields_to_the_executor` tests, all in well under a second.
- Grepped the workspace for `to_socket_addrs`/`ToSocketAddrs`: only comments remain, so the sweep of blocking resolution call sites is complete.
- `Error::TimeoutError` is transient (`error/mod.rs:414`), so both new timeout errors — DNS in `parallel_connect` and the whole attempt in `reconnect` — are retried by their loops rather than aborting them.
- `timeout_ms == 0` in `parallel_connect` (reachable when `connect_timeout == 0`, since `tds_connection_provider.rs:567` has no "0 means default" fallback) now fails at DNS rather than at the race. Both the old and new code error immediately there, so this is not a regression — it is the `connect_timeout == 0` gap already deferred as a follow-up.
- CI at review time: nothing failing; `coverage-report` and several ADO jobs still pending.

### Blocking

None.

### Suggestions

**1. DNS is now uncancellable work on the runtime's blocking pool, so the ODBC cleanup path can still block.** `tokio::net::lookup_host` runs `getaddrinfo` through `spawn_blocking`, and dropping the future does not cancel the blocking call. Tokio is explicit about what that costs at shutdown (`runtime/runtime.rs:37-40`): "Blocking functions spawned through `Runtime::spawn_blocking` keep running until they return. The thread initiating the shutdown blocks until all spawned work has been stopped." `mssql-odbc` owns its `Arc<Runtime>` on the ENV handle (`handles/env.rs:79`) and drops it from `SQLFreeHandle` (`api/free_handle.rs:82`).

So against a genuinely wedged resolver this PR does what it claims — the connect attempt now fails within the configured timeout instead of hanging the caller — but freeing the environment, which is the "cleanup" phase named in AB#47704, can still block on the leftover `getaddrinfo` thread. Nothing to change here; worth a follow-up on the ODBC side (`shutdown_timeout`/`shutdown_background` at env teardown) and worth not treating this PR alone as proof the reported hang is gone.

### Nits

**1. Stale `reconnect` doc.** `tds_client.rs:477-479` still says "Each individual TCP/TDS handshake attempt uses the original `connect_timeout`". This PR makes that doubly untrue: the value is capped to the remaining budget, and the attempt is now wrapped in `attempt_within(remaining, ...)`. One line to bring it in sync while you are in there.

**2. Busy-spinning heartbeats.** Both new heartbeat tests loop on `yield_now()` on a current-thread runtime for the entire duration of resolution — for `parallel_connect_resolution_yields_to_the_executor` that is however long the CI resolver takes to NXDOMAIN, up to the 5s budget, at 100% of a core. `tokio::time::sleep(Duration::from_millis(1))` in place of `yield_now()` keeps the identical discriminating power (the first increment still cannot happen until the main task yields) without the spin.

---

*Posted by an unattended automated review. These findings have not been checked by a human first.*

Vahid-b:
Thanks for the re-review. Addressed both nits in af805bbb:

- Stale `reconnect` doc comment fixed — now describes `connect_timeout` being capped to the remaining budget and the whole attempt wrapped in `attempt_within(remaining, ...)`.
- Both heartbeat tests swapped from `yield_now()` busy-spin to `sleep(1ms)`. Re-verified with the same mutation test (reverting to blocking `to_socket_addrs()`) that they still correctly fail — the first increment still cannot happen until the main task yields, so the discriminating power is unchanged, just without spinning a core for the whole resolution window.

On the `spawn_blocking` DNS / env-teardown point: this is accurate and a good catch, but it's a different, narrower risk than the one this PR fixes, worth being precise about:

- **What AB#47704 reported, and what's fixed here (confirmed via live CI logs, not just theory):** `test_connection_timeout_invalid_password` now passes in ~100ms instead of hanging the whole test file for 600s. That hang was the *connect attempt itself* never returning, because blocking DNS resolution defeated the timeout that was supposed to bound it. Fixed.
- **What you found, and what's still open:** if DNS resolution is genuinely wedged (not just slow — no OS-level timeout of its own) rather than eventually resolving/failing, `tokio::net::lookup_host`'s underlying `spawn_blocking` thread keeps running after our timeout abandons the awaiting future. If every DBC then gets freed and the ENV is freed too, `SQLFreeHandle(ENV)` drops the last `Arc<Runtime>` and blocks on that leftover thread with no bound, per Tokio's documented shutdown semantics. That's a real, separate gap in `mssql-odbc`'s environment teardown (plain `Drop`, no `shutdown_timeout`), not something `mssql-tds`'s connect-path fix touches.

Filed #445 to track the `shutdown_timeout` fix on the `mssql-odbc` side, referencing this thread. Not folding it into this PR, per your note and because it's a different crate/layer with its own design considerations (need to go from `Arc<Runtime>` to an owned `Runtime` via `Arc::try_unwrap` at teardown).

saurabh500:
We need to look at the impact of threads with this change. Looks like we spawn a thread per connection for dns resolution

mssql-tds/src/connection/tds_client.rs
```diff
@@ -8025,6 +8062,28 @@ mod tests {
...
 
+    /// Covers the elapsed branch of `attempt_within` directly, since nothing
+    /// in `reconnect`'s own tests can provoke it: every failure a test can
+    /// reach today (no server, bad recovery state, ...) returns well before
+    /// any deadline. `start_paused` lets the timeout fire on a fake clock
+    /// instead of a real, slow connect attempt.
+    #[tokio::test(start_paused = true)]
+    async fn attempt_within_times_out_when_the_future_never_completes() {
+        let remaining = Duration::from_secs(5);
+        let handle = tokio::spawn(attempt_within(
```
saurabh500, line 8073: Will this spawn an additional thread per connection?

saurabh500: If yes then there would be increased thread count (double) per connection. That is a severe bottleneck

Vahid-b: Good question — worth being precise about, since it's a real trade-off.

**Not a new, permanent thread per connection.** `tokio::net::lookup_host` bridges the blocking `getaddrinfo` call via `tokio::task::spawn_blocking`, which runs on Tokio's shared blocking-thread *pool* (default cap: 512 threads, unconfigured here — `mssql-odbc/src/handles/env.rs` builds the runtime with `worker_threads(1)` but no `max_blocking_threads` override). That pool is:
- **Shared** across every connection on the ENV's one `Arc<Runtime>`, not one-per-connection.
- **Transient per call** — a thread is borrowed only for the duration of the actual `getaddrinfo` (typically low-single-digit ms for cache-warm/local resolution), then returned to the pool immediately, idle and reusable by the *next* connection's resolution (or any other blocking work).
- **Reclaimed when idle** — unused pool threads are torn down after Tokio's default 10s keep-alive, so it isn't threads-at-rest either.

So "double" is accurate only for the brief resolution window itself: the calling thread (already blocked in `SQLDriverConnectW`'s `block_on`, unchanged from before this PR) plus one borrowed pool thread doing the actual OS call. It is not two threads *per connection* for the connection's lifetime — query execution afterward touches no blocking-pool thread at all.

**This isn't new to the codebase** — `ssrp.rs` already uses this exact `tokio::net::lookup_host` pattern for named-instance resolution, predating this PR. This PR extends the same, already-accepted approach to the default TCP path, specifically because the alternative (the code being replaced) used the calling thread directly for `getaddrinfo` with no separate thread at all, but that's exactly the bug: a blocking syscall on the async task means a stuck resolver hangs the caller *and* defeats every timeout wrapped around it, unboundedly. There's no way to make a genuinely blocking OS call safe for an async runtime without moving it off that runtime's task — `spawn_blocking` is the standard, idiomatic way to do that.

**Where this could genuinely matter:** an extreme burst of concurrent *new* connection attempts approaching the pool's 512-thread cap simultaneously — not steady-state query traffic, which never touches this path. If that ever becomes real, the pool size is directly tunable (`Builder::max_blocking_threads`), or the deeper fix is a genuinely async DNS resolver (e.g. `hickory-resolver`) that doesn't touch the OS resolver at all — a much bigger change (bypasses `/etc/hosts`, `nsswitch.conf`, corporate split-DNS, etc.) worth its own discussion, not something to fold into this fix.

> [!NOTE]

saurabh500: I agree that the pattern is not new to code base but most sql servers in cloud are connected with tcp in mind. 
When a connection pool warm up happens, then to satisfy the needs of multiple concurrent connections, the brief time during which connections are opened concurrently, the demand on the thread pool will be massive, and all blocking I/o.

Vahid-b: Fair pushback — let me be more concrete about the pool warm-up case specifically, since "transient" undersold the concurrent scenario.

You're right that during a *concurrent* burst (N pooled connections opening at once, all TCP, all resolving around the same moment) the accounting really is closer to "double" than I implied: each of the N calling threads is blocked in `block_on` (unchanged from before this PR) *and* each borrows its own blocking-pool thread for the overlapping `getaddrinfo` window, so you do transiently have up to 2N threads involved rather than N. I was overstating how much the "returns immediately after" framing helps when the resolutions are concurrent rather than sequential — it doesn't reduce the *peak* concurrent thread count during warm-up, only how long each thread is held.

Where I'd still push back on "severe": put a number on N. Tokio's blocking pool defaults to 512 threads, shared across the whole ENV's `Arc<Runtime>`. A connection pool opening, say, 50–100 connections concurrently — already a large warm-up burst — uses at most 50–100 of those 512 threads for the brief resolution window, then they're returned. You'd need a warm-up burst in the high hundreds, concurrently, hitting DNS at the same instant, to threaten the cap — well beyond typical pool-size configuration.

Two things worth separating:
1. **Thread *count* under realistic pool sizes**: bounded and, I think, not a practical bottleneck — happy to be shown a pool configuration where it is.
2. **Redundant work**: if N pooled connections target the same host, that's N *identical* `getaddrinfo` calls regardless of blocking or async — this PR doesn't change that count, and it's arguably the more legitimate inefficiency to fix (a short-TTL resolved-address cache/reuse at the `mssql-tds` connect layer, shared across concurrent connects to the same host, would cut both the redundant syscalls *and* the peak thread usage during warm-up). That's a real, separable improvement — want me to file it as a follow-up rather than fold it into this hang fix?

Either way: the alternative to accepting this bounded, tunable cost is reverting to the blocking call this PR removes, which reintroduces the actual bug (a stuck resolver hangs the caller indefinitely and defeats every timeout around it) — that's a correctness bug against a resource cost that's real but capped and configurable.

> [!NOTE]

David-Engel (commented):

## Summary

Swapping the blocking `std::net::ToSocketAddrs::to_socket_addrs()` for `tokio::net::lookup_host` in `create_base_stream_sequential` and `parallel_connect` is the right fix for the root cause described in AB#47704: the blocking call produced no `.await` point, so the enclosing `tokio::time::timeout` could never fire. The follow-on changes from the earlier review rounds are each correct as far as I can verify — `parallel_connect` now spends one `deadline` across resolution *and* the race, `reconnect` wraps the whole attempt in `attempt_within(remaining, ...)`, and the `.max(1)` guard is safe because the loop's `Instant::now() >= deadline` check (`tds_client.rs:550`) guarantees `remaining > 0` before the attempt starts, with `attempt_within` as the real backstop. No blocking issues; the notes below are coverage and hygiene.

## Verification

- Read the full diff against merge-base `bc4b2dc`, plus `create_base_stream`, the `reconnect` retry loop, `Error::is_transient_connect_error` (`error/mod.rs:410` — `TimeoutError` and `Io` are both transient, so the new DNS-timeout error keeps the existing retry classification), and the pre-existing `parallel_connect` tests.
- `cargo clippy -p mssql-tds --all-features --all-targets -- -D warnings` — clean.
- `cargo nextest run -p mssql-tds --lib` filtered to the new tests and their neighbours — 23/23 pass on Windows.
- Confirmed against vendored `tokio-1.53.1` `src/net/addr.rs:202-214` that `lookup_host((&str, u16))` short-circuits IP literals on a non-blocking `Ready` path and only reaches `spawn_blocking` for real hostnames. That is the property both new "yields to the executor" tests rely on, and it also drives the first suggestion below.

## Suggestion

**1. `parallel_connect.rs:144-152` — the new DNS-timeout arm is unreachable from any current test.** The only short-budget test, `test_parallel_connect_timeout` (`timeout_ms: 1`), passes the IP literal `10.255.255.1`, which `lookup_host` resolves on tokio's non-blocking `Ready` fast path, so the `timeout(...)` around it always returns `Ok`. Every other test that passes a real hostname uses a 1000-5000 ms budget. Given the 85% diff-coverage gate, this is the one new error path likely to be flagged. A deterministic cheap cover: a hostname plus a zero budget, since `lookup_host` on a non-literal must return `Pending` on its first poll while `sleep(Duration::ZERO)` is already elapsed.

```rust
#[tokio::test]
async fn parallel_connect_dns_resolution_respects_the_overall_deadline() {
    let config = ParallelConnectConfig { timeout_ms: 0, ..Default::default() };
    match parallel_connect("localhost", 1433, &config).await {
        Err(Error::TimeoutError(_)) => {}
        other => panic!("expected the DNS budget to elapse, got {other:?}"),
    }
}
```

**2. `parallel_connect.rs:489` — the new test resolves `invalid.host.that.does.not.exist.local` with a 5000 ms budget.** The `.local` suffix is routed to mDNS on macOS and on Linux hosts running avahi, where a negative answer can take seconds rather than the millisecond NXDOMAIN this assumes; a wildcard/hijacking resolver would instead return an address and make `result.is_err()` depend on the subsequent connect failing. `test_parallel_connect_invalid_host` already carries the same dependency so this is not new, but it now runs with a 5x larger budget. `.invalid` (RFC 6761, guaranteed not to resolve and never routed to mDNS) removes both the delay and the ambiguity, and the heartbeat assertion still holds. Worth doing before the `Test MacOS` leg reports, since it is the one required check still pending.

**3. Follow-up, outside this diff: the integrated-auth path still resolves DNS synchronously.** `security/spn.rs:149,176,182` call `dns_lookup::lookup_addr` / `dns_lookup::getaddrinfo` from the sync `canonicalize_hostname`, reached from `sspi_context.rs:102` and `gssapi_context.rs:157` during an async connect. That is the same class of bug this PR fixes, on a path this PR does not cover, so a Kerberos connect can still escape the connect deadline. No need to fold it in here — worth a work item alongside the `Connection Timeout=` gap already noted in the description.

## Nit

1. Several of the new comments narrate PR history rather than the code: the `attempt_within` doc comment explaining that it was "extracted out of `TdsClient::reconnect` ... so the elapsed branch can be driven directly in a test" (`tds_client.rs:404-408`), the 7-line justification above `reconnect_ctx.connect_timeout` (`tds_client.rs:558-564`), and the ~20-line preambles on the two heartbeat tests. The rationale is genuinely useful, but the repo guidance asks for the *why* in a line or two; the mutation-testing methodology in particular reads as review-thread content that has landed in permanent source.
2. `parallel_connect.rs:137` — `deadline.saturating_duration_since(Instant::now())` on the line right after `deadline` is computed is a round trip back to `config.timeout_ms`. `Duration::from_millis(config.timeout_ms)` says the same thing; the `saturating_duration_since` form is meaningful only at line 204, where time has actually been spent.
3. `CHANGELOG.md` picks up a trailing blank line at EOF.

mssql-tds/src/connection/tds_client.rs
Vahid-b: Filed #451 to track the DNS resolution caching/de-duplication improvement for pool warm-up, separate from this fix. Given the numbers we walked through (default 512-thread cap comfortably covering realistic pool sizes) and that this is a redundant-work efficiency question rather than a correctness issue with the hang fix itself, I don't think it should block this PR — but flagging here in case you'd rather keep discussing before I consider this settled.

> [!NOTE]

saurabh500: I am not convinced with your response. 
You have quoted a large number for threads at 500 for the Tokio runtime, but that doesn't mean that threads blocked on I/O like in DNS resolution, due to slow DNS, could not cause a thread starvation/exhaustion. 

Is there an option where, we could make the DNS resolution sync, and run on the calling thread, if the call has been made from a sync API like ODBC's ?

saurabh500: I am ok with the followup as well. I am not sure if you (the bot) will be able to figure out the answer. Perhaps look at the Dns.ResolveAddressesAsync to see if we should implement our Dns implementation like so ?

Theekshna (commented):
This unattended COMMENT review found no new findings. It covered the merge-base diff, callers and error classification, prior discussion, current checks, and mutation strength for both DNS-resolution call sites.

**msodbcsql parity:** N/A; the diff changes only `mssql-tds` DNS resolution and no ODBC surface.
**Test sufficiency, unit and e2e:** Fine; seven targeted unit tests passed, and reverting both `lookup_host` call sites made both executor-yield tests fail before the worktree was restored.
**Divergences documented:** Fine; deferred timeout-keyword and resolver-lifecycle work is recorded in the PR and linked follow-ups.
**PR description currency:** Current for head `af805bbb`; AB#47704 is linked and all reported checks pass.

mssql-tds/src/connection/tds_client.rs
shiwanigupta0809: @saurabh500 asked whether DNS could be resolved synchronously on the calling thread when the call comes from a sync API like ODBC's. I went and read what msodbcsql actually does, since that settles the parity half of the question.

**msodbcsql resolves synchronously on the calling thread, with no timeout around it.** `Tcp::Open` calls `GetCachedAddrInfo` (`Sql/Common/DK/sni/src/tcp.cpp:2863`), which calls `GetAddressInfo` → `getaddrinfo` directly on the caller's thread. No offload, no bound — a wedged resolver blocks the caller exactly as our pre-PR code did.

So the proposal is achievable and is what the reference driver does. It is also precisely the shape AB#47704 reported, which is the tension in this thread.

Two things that change the thread-count argument, in both directions:

**It does not increase the number of concurrent blocking DNS calls.** For an N-connection warm-up, msodbcsql makes N blocking `getaddrinfo` calls too — on N caller threads instead of N pool threads. Same count of threads parked in the same syscall; only which pool absorbs them differs. That is worth separating from the correctness change: this PR moves where the blocking happens, it does not add blocking work relative to the driver we are matching.

**But msodbcsql's DNS cache is not a hot-path cache.** `GetCachedAddrInfo` (`Sql/Common/DK/sni/src/dns_cache.cpp:65`) calls the real `getaddrinfo` **every time**, and only consults the cache **when resolution fails**, returning last-known-good addresses:

```cpp
DWORD ret = GetAddressInfo(node, service, hints, &origAddrInfo);
if (ret)
{
    // DNS failure, try the cache
    if (CACHEADDRINFOW *p = DNSCache::get(node)) { *res = p; ret = 0; }
}
else
{
    // DNS success, add to cache
    DNSCache::add(node, *res = new DNSCacheEntry(origAddrInfo));
}
```

It is stale-on-failure resilience, not de-duplication — so it does not reduce warm-up syscalls or thread usage there either.

That distinction matters for #451. A TTL de-dup cache would go beyond msodbcsql and would genuinely help warm-up, but it is a different mechanism from the one msodbcsql has. The stale-on-failure behaviour is a separate and arguably more valuable parity gap that nobody is tracking: during a transient DNS outage msodbcsql keeps connecting on cached addresses, and we fail outright. Worth splitting those into two items rather than one, since only the first addresses the concern in this thread.

shiwanigupta0809:
One thing worth pinning while trimming the heartbeat-test comments per the nit above, because the trim could remove the wrong part.

Both `*_resolution_yields_to_the_executor` tests discriminate only because `#[tokio::test]` defaults to `current_thread`: the spawned heartbeat cannot run until the main task yields, which is exactly the property being asserted. On `#[tokio::test(flavor = "multi_thread")]` the heartbeat gets its own worker and the counter increments whether or not resolution blocks — the assertion still passes and silently stops guarding anything. That is a routine edit to make (adding a test to the module that needs multiple threads, then normalising the attributes), and the mutation check that proved these tests work would not be re-run afterwards.

The comments currently explain why the heartbeat proves the property but not that the single-threaded runtime is what makes it a proof. So of the ~20 lines, that is the one fact that has to survive:

```rust
// Must stay on the default `current_thread` runtime: a `multi_thread`
// flavor would let the heartbeat run on another worker even if resolution
// blocked, so the assertion would pass without proving anything.
```

The methodology narration around it is the part that reads as review-thread content, and I agree with the nit that it can go.

Vahid-b:
Great catch, and confirmed empirically before making the change: with `#[tokio::test(flavor = "multi_thread")]` and the resolver reverted to blocking `to_socket_addrs()`, the test wrongly passes — the heartbeat runs on its own worker regardless of whether the main task's resolution blocks, exactly as you described. Fixed in d0697e8a:

- Made the flavor explicit: `#[tokio::test(flavor = "current_thread")]` on both tests, instead of relying on the implicit default that a later edit could silently change.
- Trimmed both doc comments down, keeping your suggested fact (current_thread is what makes the heartbeat a proof, not just scaffolding) verbatim in spirit, and cut the methodology narration around it.
- Re-ran the same mutation methodology from before (revert to blocking resolver) to confirm the tests still fail correctly on `current_thread`, then reverted.

Thanks for catching this — it's exactly the kind of silent regression a "routine" test-suite edit could introduce without anyone noticing, since the test would keep passing.

David-Engel (approved):

PR #408, opened 2026-08-28
Withdraw a partially sent request instead of dropping it

Theekshna:
## Description

A value rejected after a packet has already flushed left the server holding an incomplete message. Dropping the `PacketWriter` discards the local buffer but not the bytes on the wire, so the next command on that connection was read as a continuation and the server answered **4002** — on a *later* statement than the one that failed.

The withdrawal already existed for data-at-execution (`cancel_streamed_write`: `EOM | IGNORE`, then consume the DONE); `nothing_sent` and `abandon` had exactly one caller each. All nine parameter-carrying send sites now share it through `finish_send` → `retract_partial_request`, which picks one of three arms:

| State | Action | msodbcsql |
|---|---|---|
| Nothing flushed | Abandon locally, re-arming the RESETCONNECTION bit the message took | `STATE_BATCH_QUEUINGCMD` |
| Partially sent | Withdraw with `EOM \| IGNORE`, then consume the DONE | `STATE_BATCH_PARTIALCMDSENT` |
| Fully sent | Cancel with an attention — there is nothing to withdraw | `STATE_BATCH_CMDSENT` |

Consolidating the paths is what surfaced the rest of this PR. Five defects, all of which existed on `main` or were reachable the moment the first was fixed:

1. **`nothing_sent()` reported a landed packet as unsent.** The write budget and the first-packet callback are both checked *after* the packet reaches the wire, and `is_first_packet` is cleared only after them. A request overrunning its own `SQL_ATTR_QUERY_TIMEOUT` was therefore discarded locally with the server mid-message — the same 4002, reached through the timeout door. Backed by `any_packet_flushed`, set the instant bytes flush.
2. **A fully sent message was withdrawn rather than cancelled.** The same budget check returns before `eom_pending` and the cursor reset, so a last-packet overrun left the server with the complete message and the retraction opened a *second* one. The drain stopped at the first DONE and left the ignore's DONE for the next command. `message_complete` now routes that case to an attention.
3. **The ignore packet carried the RESETCONNECTION bit.** On the path above `is_first_packet` is still set, so the header builder handed the bit to the `EOM | IGNORE` packet — the bit the exemption beside it deliberately does not record, and which `rearm_withdrawn_reset` re-arms immediately after. Cleared in `cancel_current_message`.
[...]

PR #433, opened 2026-08-31
Retract partially sent cursor RPCs

Theekshna:
## Description

Route all server-cursor RPC sends through the partial-request recovery added in #408. Serialization failures now abandon unsent requests, withdraw partial requests with EOM | IGNORE, or cancel fully sent requests with attention while preserving the original error.

Adds byte-level unit coverage for seven cursor send paths, including cancellation after a packet flush, and live-server coverage for same-session and same-cursor reuse.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47714/

Follow-up: https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47772/

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

Live cursor e2e tests pass. The full local unit run has 7 pre-existing failures from missing certificate fixtures.

### DIFF
--- a/mssql-tds/src/connection/cursor_ops.rs
+++ b/mssql-tds/src/connection/cursor_ops.rs
@@ -326,7 +326,10 @@ impl CursorClient for TdsClient {
 
         let mut pw =
             rpc.create_packet_writer(self.transport.as_writer(), timeout_sec, cancel_handle);
-        rpc.serialize(&mut pw).await?;
+        let serialize_result = rpc.serialize(&mut pw).await;
+        let message = pw.suspend();
+        drop(rpc);
+        self.finish_send(serialize_result, message).await?;
 
         self.consume_cursor_open_response().await
     }
@@ -416,7 +419,10 @@ impl CursorClient for TdsClient {
 
         let mut pw =
             rpc.create_packet_writer(self.transport.as_writer(), timeout_sec, cancel_handle);
-        rpc.serialize(&mut pw).await?;
+        let serialize_result = rpc.serialize(&mut pw).await;
+        let message = pw.suspend();
+        drop(rpc);
+        self.finish_send(serialize_result, message).await?;
 
         self.consume_cursor_open_response().await
     }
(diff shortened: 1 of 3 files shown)


David-Engel (commented):

## Summary

This routes all ten cursor RPC send sites in `cursor_ops.rs` through the `finish_send` recovery added in #408, using the exact `serialize` → `suspend` → `drop(rpc)` → `finish_send` shape already established at the batch/RPC sites in `tds_client.rs`. The change is mechanical and consistent, `finish_send` is widened only to `pub(super)` (correct: `cursor_ops` and `tds_client` are siblings under `connection`), and the reworded `withdraw_sent_message` doc matches the code — `CANCEL_WRITE_TIMEOUT_SECS` is `CANCEL_TIMEOUT_SECS / 2` = 60s on the ignore write, with the 120s `CANCEL_TIMEOUT` as the outer deadline and again on the drain. No findings.

## Verification

Checked out the head commit in a dedicated worktree and diffed against the merge base `0eaba79d`.

- **Mutation test (the important one).** Reverted all ten call sites back to `rpc.serialize(&mut pw).await?` and re-ran `cargo nextest run -p mssql-tds --lib cursor`: all seven new unit tests fail (`the withdrawal must end with a header-only TDS packet`, `left: 97 right: 8`), then pass again once restored. The tests genuinely pin the retraction rather than the happy path, and they assert the `EOM | IGNORE` header-only trailer, so they land in the withdraw branch rather than the cheap `nothing_sent()` branch.
- `cargo nextest run -p mssql-tds --lib retract` → 16/16 pass, including the eight pre-existing retraction tests.
- `cargo clippy -p mssql-tds --all-features --all-targets -- -D warnings` → clean, which also confirms the new live-server cases in `tests/test_cursor_ops.rs` compile.
- Read the recovery machinery the diff now depends on (`retract_partial_request`, `withdraw_sent_message`, `rearm_withdrawn_reset`, `abandon`/`suspend`) and each cursor entry point's surrounding state handling. `retract_partial_request` clearing `has_open_batch` is right for every cursor path — none of them set it before the send, and `cursor_fetch` only sets it after a successful `next_rowset`.
- `gh pr checks 433`: nothing failing (4 still pending at review time). The automated coverage comment reports 99% diff coverage, so the checklist's claims are consistent with CI.

One thing worth noting rather than flagging: `cursor_fetch`, `cursor_close`, and `cursor_unprepare` get the same treatment but no retraction test. That looks correct to me — their parameters are fixed-size `Int`s, so they cannot fail serialization partway or span packets, which makes the change there purely defensive and the partial-send scenario unreachable. The `TODO(AB#47772)` on `finish_send` already covers the remaining duplication across send sites, so I did not re-raise it.

## Findings

**Blocking:** none.

**Suggestion:** none.

**Nit:** none.

David-Engel (approved):

PR #438, opened 2026-08-31
Bump mssql-py-core version to 0.1.10

### LINKED ISSUE #437: Bump mssql-py-core version to 0.1.10
### Problem statement

`mssql-py-core` is currently versioned at `0.1.9`. A new package version is needed so the latest `main` build can be published as a distinct, immutable artifact.

### Proposed solution

Bump the crate version in `mssql-py-core/Cargo.toml` from `0.1.9` to `0.1.10`.

### Affected crate

mssql-py-core

### Alternatives considered

Reusing `0.1.9` would make published artifacts ambiguous and is not suitable for an immutable package release.

### Additional context

This follows the repository's prior version-only bumps for `mssql-py-core`.

saurabh500:
## Description

Bump the `mssql-py-core` crate version from `0.1.9` to `0.1.10`.

## Related Issues

Fixes #437

## Checklist

- [x] `cargo bfmt` passes — `cargo fmt --manifest-path mssql-py-core/Cargo.toml -- --check` is clean; the crate is excluded from the root workspace.
- [x] `cargo bclippy` passes — N/A: version-only manifest change.
- [x] `cargo btest` passes — N/A: no logic changed.
- [x] New/changed functionality has tests — N/A: version bump only.
- [x] Public API changes are documented — N/A: no API change.

### DIFF
--- a/mssql-py-core/Cargo.toml
+++ b/mssql-py-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mssql-py-core"
-version = "0.1.9"
+version = "0.1.10"
 edition = "2024"
 publish = false
 


David-Engel (commented):

## Summary

Single-line manifest change bumping `mssql-py-core` from `0.1.9` to `0.1.10`, matching issue #437 and the documented process in `docs/release-management.md` (Scenario 3, step 2: "Bump version in `mssql-py-core/Cargo.toml`"). The bumped line is the only version reference in the tree that needs to move. No `Cargo.lock` is tracked, so there is no lock entry to refresh. Nothing under `mssql-py-core/src` reads `CARGO_PKG_VERSION` — the `__version__` lookup in `src/lib.rs` pulls `mssql_python.__version__` from the host package, not this crate — so no runtime behavior, app name, or LOGIN7 field shifts with the bump. Both pipeline sites that consume the version (`.pipeline/OneBranch/stages.yml` and `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml`) regex the first `version = "..."` out of the raw manifest, which is the `[package]` version on line 3, so they pick up `0.1.10` correctly. `0.1.10` also orders after `0.1.9` under both NuGet and PEP 440 numeric comparison, so the `-dev.*` and `-nightly.*` suffixed variants sort as intended.

## Verification

Read-level; nothing compiled, since no Rust code changed and a build would only re-collect a verdict CI already has. Diffed against the merge base with current `origin/main` (`ba6ad9c9`) rather than trusting per-commit stats: one file, one line. Ran `cargo metadata --manifest-path mssql-py-core/Cargo.toml --no-deps` in the assigned worktree with the assigned `CARGO_TARGET_DIR` and confirmed the manifest still parses and resolves to `mssql-py-core 0.1.10`, edition 2024. Grepped the full tree for `0.1.9` and `0.1.10`: the bumped line is the sole match, so no stale reference was left behind. Traced both pipeline version-extraction steps plus `docs/release-management.md` to confirm the bump actually reaches the published package version. Read the existing Copilot bot review, the inline threads, and the PR-level comments first — all three are empty of findings, so nothing below is a re-file. `gh pr checks 438` shows every completed required check passing, with macOS build/test, the macOS cross-repo job, and `coverage-report` still pending.

## Findings

### Blocking

None.

### Suggestion

**The bump moves the NuGet package version but not the wheel filenames inside it.** `mssql-py-core/pyproject.toml` declares a static `[project] version = "0.1.0"`, and maturin gives PEP 621 metadata in `pyproject.toml` precedence over `Cargo.toml`. So the wheels packed into the nupkg should still be named `mssql_py_core-0.1.0-cp3XX-...whl` even though the container becomes `mssql-py-core-wheels.0.1.10-*`. That is also what makes `docs/release-management.md` self-inconsistent today: its example listing shows `mssql_py_core-0.2.1-...` wheels inside `mssql-py-core-wheels.0.2.1.nupkg`, which is not what the current pair of manifests produces.

This is pre-existing, not introduced here — `pyproject.toml` has read `0.1.0` since the crate was added, across every bump from `0.1.4` through `0.1.9` — and it does not block this PR, because what downstream `mssql-python` pins is the NuGet package version, and that does move. But since #437 asks specifically for a "distinct, immutable artifact," it's worth deciding deliberately whether the wheel filename should track as well. If it should, `dynamic = ["version"]` in `[project]` lets maturin inherit the Cargo version, which beats hand-editing two files on every bump and drifting again. If it shouldn't, the docs example is the thing to correct. Either way, a follow-up rather than this PR.

Caveat on that claim: maturin isn't installed in this environment, so I could not observe the produced filename. The precedence is read from maturin's documented metadata merge, not from this build. If a PR-validation wheel-build log shows `0.1.9`-named wheels from the previous release, then the version is being stamped somewhere I didn't find and you should discard this finding.

### Nit

None.

Vahid-b (approved):

Issue #439, opened 2026-08-31
test_set_attr_txn_isolation_effect hangs — server-side LOCK_TIMEOUT wait on a blocked query never resolves


Vahid-b:
### Describe the bug

`mssql-python`'s `tests/test_003_connection.py::test_set_attr_txn_isolation_effect` hangs indefinitely against `mssql-odbc`, until the mssql-python test harness's per-file 600s watchdog kills the whole `test_003_connection.py` process.

This was discovered while validating PR #432 (fix for [AB#47704](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47704), "Failed invalid-password connection hangs test process during cleanup"). That fix let the mssql-python suite progress much further than before (from ~10% to 71% of `test_003_connection.py`) before hitting this **different, pre-existing** hang. It was previously masked because the file always died earlier, at `test_connection_timeout_invalid_password`.

Not related to PR #432's fix: this hang occurs on two already-established, already-authenticated connections blocking on query execution / row locking — a completely different code path from connection establishment / DNS resolution.

### Steps to reproduce

The test (from `microsoft/mssql-python`, `tests/test_003_connection.py`) does:
1. Opens two separate connections (`conn1`, `conn2`) to a global temp table `##test_isolation`.
2. On `conn1`: `BEGIN TRANSACTION` then `UPDATE ... WHERE id = 1` — leaves the transaction open/uncommitted, holding a row lock.
3. On `conn2`: `SET LOCK_TIMEOUT 5000` (server-side T-SQL session setting), then `SELECT ... WHERE id = 1` inside `pytest.raises(...)`, expecting SQL Server to block on `conn1`'s lock and return error 1222 ("Lock request time out period exceeded") after 5 seconds.

### Expected behavior

`conn2`'s blocked `SELECT` should return SQL Server error 1222 after ~5 seconds (per `SET LOCK_TIMEOUT 5000`), which the test expects and catches.

### Actual behavior

The 5-second wait never resolves. Observed in Azure DevOps build 171327 (`mssql-rs Pull request validation` → `Build mssql-python` → `mssql-python suite on mssql-odbc driver (cross-repo)`), task "Run mssql-python tests against mssql-odbc":

```
21:58:46.0576Z  tests/test_003_connection.py::test_set_attr_edge_cases PASSED [71%]
                (test_set_attr_txn_isolation_effect starts here — no further output)
[...]

PR #442, opened 2026-09-01
mssql-odbc: Enforce SQL_ATTR_QUERY_TIMEOUT so a blocked statement cannot wait forever

Vahid-b:
## Summary

Motivated by #439, though see **Relationship to #439** below — this PR does not by itself resolve that issue's reported hang.

`SQLExecute` and `SQLExecDirectW` hard-coded `ExecuteOptions::default()` for every wire call, so `StmtState::query_timeout` (`SQL_ATTR_QUERY_TIMEOUT`) was stored and reported back by `SQLGetStmtAttrW` but **never actually applied**. A statement blocked server-side — e.g. behind another session's row lock — therefore had no client-side escape hatch and could wait indefinitely even with a timeout configured. This was an explicit, tracked gap ([AB#46385](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46385)).

Both entry points now thread `StmtState::query_timeout` into `ExecuteOptions::timeout_secs(...)`, and the same budget also bounds the steps that can run *before* the main execute:
- `SQLExecute` (`execute.rs`): both the materialized-parameter (`execute_prepared`) and data-at-execution (`begin_execute_prepared`) branches.
- `SQLExecDirectW` (`exec_direct.rs`): the DAE branch (`begin_sp_executesql`), the parameterized branch (`execute_sp_executesql`), and the plain-batch branch (`execute`).
- `flush_pending_unprepare`'s deferred `sp_unprepare` and `begin_transaction_if_manual`'s implicit transaction begin are now also bounded by the statement's query timeout, with elapsed time deducted from the budget before the main execute — mirroring msodbcsql's `DropPrepHandle` (`sqlcfunc.cpp:787-828`) and `CheckOptions`/`ExecTMRImmediate` (`sqlccmd.cpp:10572-10586`).

`0` (the ODBC default, "no timeout") keeps meaning unlimited, so the common case is behaviorally unchanged — verified by all existing `mssql-odbc` unit tests still passing unmodified.

On expiry, the underlying `mssql-tds` client already sends `ATTENTION` and waits for the acknowledgement (existing, tested behavior) — the only remaining gap was the SQLSTATE: `post_tds_error` mapped every non-`SqlServerError` failure to the caller's generic `default` (`HY000`), so a timeout was indistinguishable from any other transport failure. Added a `TdsError::TimeoutError(_) => SQLSTATE_HYT00` arm so a timeout now reports `HYT00`, matching msodbcsql and letting applications branch on it correctly.

## Relationship to #439

Issue #439's own reproduction (`mssql-python`'s `test_set_attr_txn_isolation_effect`) never sets `SQL_ATTR_QUERY_TIMEOUT` — it only issues `SET LOCK_TIMEOUT 5000` as a T-SQL batch, and `mssql-python`'s `Cursor` defaults its own `timeout` to `0` (unlimited), so `SQLSetStmtAttr(SQL_ATTR_QUERY_TIMEOUT, ...)` is never called with a non-zero value in that test. Since this PR's fix only bounds the wait when a *non-zero* `SQL_ATTR_QUERY_TIMEOUT` is configured, it does **not** change the behavior of that specific reproduction, and issue #439's reported hang should stay open pending either a live-server reproduction of a genuine mssql-odbc/mssql-tds defect, or a fix on the `mssql-python` side (e.g. actually configuring a driver-level timeout, or fixing whatever in its connection-pooling/cursor-reset layer is the true trigger).

What this PR *does* fix is the separately-tracked, unconditionally-real gap ([AB#46385](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46385)): today, no `SQL_ATTR_QUERY_TIMEOUT` value — zero or non-zero — bounds anything, which is a correctness bug in its own right and a prerequisite for any timeout-based mitigation of #439-like hangs going forward.

## Investigation

The issue's own "unconfirmed" hypotheses were investigated directly rather than assumed:

1. **"A gap in how the driver waits for/relays a response that arrives several seconds late"** — ruled out. Added `QueryResponse::error_only` / `with_delay` / a `TerminalError` type to `mssql-mock-tds` (a real error response, real TDS DONE_ERROR framing, held back by a genuine `tokio::time::sleep` before the server answers) and a new test, `test_delayed_statement_error_does_not_hang`, driving a real `TdsClient` over a real TCP socket against it. The client correctly waits out a 3-second server-side delay and surfaces the SQL error — no hang.
2. **The shared, single-worker-thread Tokio runtime (`EnvHandle`, `worker_threads(1)`)** — also ruled out as a contributor. `test_shared_single_worker_runtime.rs` mirrors `EnvHandle::new()`'s exact runtime configuration and drives two connections through sequential, non-async `Runtime::block_on` calls (the same pattern every ODBC entry point uses) against the mock server's delayed response — no starvation, no hang. (The bound is now enforced with `tokio::time::timeout` so a real regression fails fast instead of hanging the suite.)

Both are kept as permanent regression tests. Neither reproduces the reported hang, which points at something outside this repo (`mssql-python`'s own connection-pooling/cursor-reset layer, or the live-server specifics) rather than a further mssql-odbc/mssql-tds defect — see **Relationship to #439** above. Wiring up the timeout is the concrete, verifiable, in-scope fix: it gives the driver a client-side bound matching msodbcsql, regardless of the deeper cause of any future "wait never resolves" report.

## Review-driven refinements

Two rounds of independent review found real issues in the pre-execute timeout deduction, both fixed and covered by unit tests:

- **Double-charged elapsed time**: an early version measured elapsed time cumulatively from the start of `SQLExecDirectW` for both pre-execute deductions (after `flush_pending_unprepare` and after `begin_transaction_if_manual`), so the first step's cost was subtracted twice.
- **Sub-second truncation loss**: the fix for that (re-seeding the timer between steps) avoided double-charging but floored each step's sub-second remainder to zero independently, so e.g. two ~1s pre-execute steps against a 1s timeout could each charge nothing.

Both are resolved by deducting from a *fixed* original budget using the *cumulative* elapsed time since the call began, computed once per step rather than per shrinking remainder — matching msodbcsql's own millisecond-granularity deduction. See `deduct_query_timeout_composed_twice_charges_each_step_once` and `deduct_query_timeout_cumulative_composition_catches_accumulated_sub_second_cost` in `exec_common.rs`.

Also hardened `mssql-mock-tds`'s delayed-response wait: a client closing its socket mid-delay (without ever sending `Attention`) now surfaces as an ordinary connection close observed by the outer read loop, rather than an internal `ProtocolError` that would skip past recording the connection for test assertions.

## Testing

- `cargo bfmt` — clean.
- `cargo clippy --workspace --frozen --all-features --all-targets -- -D warnings` — clean.
- `cargo nextest run -p mssqlodbc` — full suite passing, including mock-server-backed regression tests that actually exercise a delayed server response or a delayed implicit transaction begin against a short configured timeout (`execute_query_timeout_bounds_a_longer_server_delay`, `exec_direct_query_timeout_bounds_a_longer_server_delay`, `exec_direct_query_timeout_bounds_a_delayed_implicit_transaction_begin`), each confirmed mutation-resistant (they fail with the old unbounded behavior), plus direct unit tests for the deduction arithmetic.
- `cargo nextest run -p mssql-mock-tds` — passing, including the delay/Attention-interruption support.
- `cargo nextest run -p mssql-tds -E "binary(test_mock_server) or binary(test_shared_single_worker_runtime)"` — passing, including a new test proving the client's own timeout fires before a longer server delay.

## Checklist
- [x] `cargo bfmt`
- [x] `cargo bclippy` (scoped to touched crates; full workspace clippy is clean)
- [x] `cargo btest` (scoped to touched crates + new tests; live-server-dependent tests in `mssql-tds` were not run — no live SQL Server credentials available in this environment)


mssql-odbc/src/api/exec_direct.rs
```diff
@@ -150,9 +155,11 @@ fn sql_exec_direct_w_safe(
     // and hand control to SQLParamData / SQLPutData. There is no prepared plan
     // to restore afterwards, so `None` is passed for it.
     if !dae_params.is_empty() {
-        let begin_result =
-            dbc.runtime
-                .block_on(client.begin_sp_executesql(rewritten_sql, params, ()));
+        let begin_result = dbc.runtime.block_on(client.begin_sp_executesql(
+            rewritten_sql,
+            params,
```
Vahid-b, line 203: The newly bounded execute call is reached only after two potentially blocking wire operations that still use no statement timeout. `flush_pending_unprepare` calls `client.unprepare(handle, ())`, and `begin_transaction_if_manual` calls `begin_transaction(..., None)`. Thus `SQLExecDirectW` can still wait indefinitely in deferred `sp_unprepare`, and both execute entry points can wait indefinitely starting an implicit transaction, before reaching this timeout. Native msodbcsql bounds both paths: `DropPrepHandle` copies `dwQueryTimeoutInMS` and passes the deducted timeout to `ExecRPCImmediate` (`sqlcfunc.cpp:787-828`), while `CheckOptions` passes `GetQueryTimeOut(lpstmt)` to `ExecTMRImmediate` (`sqlccmd.cpp:10572-10586`). Please propagate the statement timeout through both pre-execution operations, preferably deducting their elapsed time from the final execution budget.

mssql-mock-tds/src/server.rs
```diff
@@ -366,12 +366,17 @@ impl ConnectionProcessor {
...
-                            // Look up query in registry
-                            let registry = self.query_registry.lock().await;
-                            if let Some(response_data) = registry.get(&sql) {
+                            // Look up query in registry. Cloned so the lock is
+                            // released before any artificial delay below.
+                            let registered = self.query_registry.lock().await.get(&sql).cloned();
+                            if let Some(response_data) = registered {
                                 info!("Found registered response for query");
+                                if let Some(delay) = response_data.delay {
+                                    debug!(?delay, "Delaying response for query");
```
Vahid-b: This sleep runs inside the per-connection packet-processing loop, so while a response is delayed the mock server cannot read or process an `ATTENTION` packet on that connection. That differs from the cancellation behavior needed by the proposed timeout regression test: the attention remains queued until after the original delayed response is sent. Consider making delayed execution interruptible by attention, or explicitly limiting `with_delay` to late-response tests and using separate mock behavior for timeout/attention acknowledgement and connection-reuse coverage.

Vahid-b (commented):
## Summary

The core `ExecuteOptions` plumbing and `HYT00` mapping are directionally correct, but the statement timeout still does not bound every wire operation performed before execution. I also confirmed the existing Copilot findings: the linked #439 reproduction keeps `SQL_ATTR_QUERY_TIMEOUT` at `0`, and the new ODBC tests do not detect removal of the production wiring.

## Validation

- Cross-checked deferred unprepare and implicit transaction begin against native msodbcsql.
- Ran the six focused new tests successfully.
- Mutation-tested the coverage: after reverting the plain-batch `SQLExecDirectW` call to `()` and the materialized `SQLExecute` call to `ExecuteOptions::default()`, all three relevant new plumbing tests still passed.
- Confirmed the new single-worker test has no effective outer deadline.

Overall assessment: request changes before marking the draft ready. The existing Copilot threads cover the issue-scope mismatch, timeout regression coverage, and hanging-test guard; the inline comments below add the remaining source-verified gaps without duplicating those threads.

mssql-odbc/src/api/execute.rs
```diff
@@ -123,13 +130,13 @@ fn sql_execute_safe(statement_handle: SqlHandle, stmt: &StmtHandle) -> SqlReturn
...
-            // Command timeout (SQL_ATTR_QUERY_TIMEOUT) isn't wired up yet; the default
-            // `ExecuteOptions` means no per-command limit.
+            // `query_timeout` (SQL_ATTR_QUERY_TIMEOUT) bounds the whole call, including
+            // any reconnect charged above; `0` means unlimited, matching the ODBC default.
             let exec_result = dbc.runtime.block_on(client.execute_prepared(
                 &mut prepared.stmt,
                 named_params,
                 &mut orphaned,
-                ExecuteOptions::default(),
+                ExecuteOptions::new().timeout_secs(query_timeout),
```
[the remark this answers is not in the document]
Vahid-b: Fair point — the staging tests only proved plumbing, not behavior. Added `execute_query_timeout_bounds_a_longer_server_delay`: it drives the real `SQLExecute` path against a mock TDS server holding its response for 8s with a 1s `SQL_ATTR_QUERY_TIMEOUT`, and asserts `SQL_ERROR`/`HYT00` within 5s. Confirmed mutation-resistant by temporarily reverting the timeout wiring to `ExecuteOptions::default()` and verifying the test then fails (takes the full ~8s, wrong SQLSTATE).

Also caught (via that same mutation-testing pass) a rounding bug in the new elapsed-time deduction used by the pre-execution steps below (`deduct_query_timeout` was rounding *up*, silently exhausting small timeouts before any network wait) — fixed to floor instead.

> [!NOTE]

mssql-tds/tests/test_shared_single_worker_runtime.rs
```diff
@@ -0,0 +1,136 @@
...
+    // conn2: connect, then run the statement the mock server holds for
+    // RESPONSE_DELAY before answering with SQL Server's real lock-timeout
+    // error (1222).
+    let mut conn2 = dbc_runtime
+        .block_on(provider.create_client(make_context(), &datasource, None))
+        .expect("conn2 failed to connect");
+
+    let started = Instant::now();
+    let result = dbc_runtime.block_on(conn2.execute(SELECT_SQL.to_string(), ()));
+    let elapsed = started.elapsed();
```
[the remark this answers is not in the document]
Vahid-b: Good catch — wrapped the execute call in `tokio::time::timeout(Duration::from_secs(30), ...)` inside the `block_on`'d async block, so a genuine starvation regression now fails the assertion deterministically in ~30s instead of hanging the test binary indefinitely.

> [!NOTE]

mssql-mock-tds/src/server.rs
Vahid-b: Good catch. Made delayed responses interruptible: `process_packet` is now generic over the socket type, and a new `wait_out_delay_or_attention` helper races the delay against `socket.read_buf(...)` via `tokio::select!` in a loop, checking after each read (via a new `try_take_attention`) whether a complete `Attention` packet arrived — if so, it abandons the delayed response and immediately acks with `build_attention_ack_packet()` instead of finishing the original delay. Applied to the `SqlBatch`, new `RpcRequest`, and `TransactionManager`-Begin delay paths.

Also found and fixed a related bug while implementing this: the existing `Attention` handler built its acknowledgement via `build_done_token(0x0020)`, but `0x0020` was landing in `build_done_token`'s `row_count` parameter, not `status` — so the "DONE_ATTN" response was actually just a plain `DONE_FINAL` token. Added `build_done_token_with_status`/`build_attention_ack_packet` to fix that.

Verified end-to-end with `test_client_timeout_fires_before_a_longer_server_delay` in `mssql-tds`: a 1s client timeout against an 8s mock delay now completes in ~1.1s via the Attention race, not the full 8s.

> [!NOTE]

mssql-odbc/src/api/exec_direct.rs
```diff
@@ -175,17 +182,24 @@ fn sql_exec_direct_w_safe(
...
+            ))
             .map(|_| ())
     } else {
         // Statement-wise navigation: position on the batch's first statement
         // (msodbcsql parity) so no-row statements (PRINT / RAISERROR / DML) are
         // individually navigable via SQLMoreResults. finish_execute inspects the
         // resulting client state.
-        dbc.runtime.block_on(client.execute(sql, ())).map(|_| ())
+        dbc.runtime
+            .block_on(client.execute(sql, ExecuteOptions::new().timeout_secs(query_timeout)))
```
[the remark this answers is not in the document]
Vahid-b: Added `exec_direct_query_timeout_bounds_a_longer_server_delay`, covering the plain-batch branch against a mock server holding its response for 8s with a 1s statement timeout — asserts `SQL_ERROR`/`HYT00` within 5s and is mutation-tested (fails if the wiring is reverted).

For the parameterized/DAE branches specifically: they share the same `query_timeout`/`ExecuteOptions::timeout_secs(...)` plumbing and the same pre-execute deduction path added in this round, so a regression in either would also break through the same mechanism this test exercises; I didn't duplicate the mock-server harness per branch to keep the test surface proportionate, but flagging this as a spot I considered extending further if you'd like dedicated coverage per branch.

> [!NOTE]

mssql-odbc/src/api/exec_direct.rs
Vahid-b: Agreed, and thanks for the msodbcsql citations. Addressed both:

- `flush_pending_unprepare` now takes `timeout_secs` and calls `client.unprepare(handle, ExecuteOptions::new().timeout_secs(timeout_secs))` instead of the old unbounded call.
- `begin_transaction_if_manual` now takes `timeout_secs` and calls a new `TdsClient::begin_transaction_with_options(..., ExecuteOptions::new().timeout_secs(timeout_secs))` — added non-breaking (the existing public `begin_transaction` delegates to it with `()`, so its ~28 existing call sites are untouched).
- Both `SQLExecute` and `SQLExecDirectW` now measure elapsed wall-clock time between steps and deduct it from the remaining budget via a new `deduct_query_timeout` helper before each subsequent step, failing immediately with `HYT00` if the budget is already exhausted — mirroring `DropPrepHandle`/`CheckOptions`.
- Added a dedicated regression test, `exec_direct_query_timeout_bounds_a_delayed_implicit_transaction_begin`, that delays only the mock server's answer to the TransactionManager Begin request (via a reserved `QueryRegistry` key) with autocommit off, and confirms `SQLExecDirectW` still returns `HYT00` within the statement's 1s timeout instead of waiting the full 8s. Mutation-tested — fails if `begin_transaction_if_manual`'s timeout wiring is reverted.

Non-manual/catalog call sites (`catalog.rs`, `describe_param.rs`, `get_type_info.rs`) intentionally pass `0` (unbounded) since they're outside this PR's scope — noted inline.

> [!NOTE]

CHANGELOG.md
```diff
@@ -150,3 +150,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/), and this
...
   payload nor terminates a message — but was previously consumed as a
   zero-length packet. Empty end-of-message packets remain legal.
+
+- `mssql-odbc`: `SQL_ATTR_QUERY_TIMEOUT` is now enforced (AB#46385). Previously
+  it was stored and reported back but silently ignored by `SQLExecute` and
+  `SQLExecDirectW`, so a statement blocked server-side (e.g. behind another
+  session's row lock) had no client-side escape hatch and could wait
+  indefinitely even with a timeout configured. A non-zero value now bounds the
+  wait; on expiry the driver sends `ATTENTION` and reports `HYT00`, matching
+  msodbcsql. `0` (the ODBC default) remains unlimited. Fixes #439.
```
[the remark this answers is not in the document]
Vahid-b: You're right — confirmed by pulling up mssql-python's `test_set_attr_txn_isolation_effect`: it only runs `SET LOCK_TIMEOUT 5000` as a T-SQL batch, and `mssql-python`'s own `Cursor` defaults its `timeout` to `0`, so `SQL_ATTR_QUERY_TIMEOUT` is never set to a non-zero value in that test. This change doesn't alter that reproduction's behavior.

Removed the "Fixes #439" claim from both the CHANGELOG entry and the PR description, and added a "Relationship to #439" section explaining this addresses the separately-tracked AB#46385 gap (no `SQL_ATTR_QUERY_TIMEOUT` value bounds anything today) without closing the reported hang, which should stay open pending a live-server repro or an `mssql-python`-side fix.

> [!NOTE]

David-Engel (commented):
## Automated review (unattended run — findings were not checked by a human first)

Threads the statement's `SQL_ATTR_QUERY_TIMEOUT` into `ExecuteOptions` for every wire step `SQLExecute`/`SQLExecDirectW` performs (deferred `sp_unprepare`, implicit TM `Begin`, the execute itself) and maps `TdsError::TimeoutError` to `HYT00`. The direction matches the msodbcsql citations given (`DropPrepHandle` `sqlcfunc.cpp:787-828`, `CheckOptions`/`ExecTMRImmediate` `sqlccmd.cpp:10572-10586`), and the mock-server work makes the new tests real rather than plumbing-only. One arithmetic bug in the multi-step budget deduction.

**Verified:** checked out at `7aa2769`, built and ran `cargo nextest run -p mssql-odbc -E 'test(query_timeout)'` — 15/15 pass, and the three delayed-response tests each finish in ~1.2s against an 8s server delay, so the bound is genuinely enforced rather than asserted. I did not re-run the full suite; ADO PR validation was still pending at review time and `coverage-report` had failed after 1h15m (looks like an infra/timeout failure, and diff coverage is reported, not enforced).

### Blocking

**`mssql-odbc/src/api/exec_direct.rs:178` — the pre-execute elapsed time is charged twice.**
`started` is captured once at line 148 and never reset, so line 178 subtracts the *cumulative* elapsed since line 148 from a budget that line 161 already reduced by the `flush_pending_unprepare` portion of that same interval. With a 10s timeout, a 3s `sp_unprepare` and a 2s implicit `Begin`, the execute gets `(10-3)-5 = 2s` instead of the intended `10-5 = 5s`, and a query that would have answered in 4s fails with a spurious `HYT00`. `execute.rs` is correct because it deducts only once. The floor-to-whole-seconds rounding hides it whenever the pre-steps are sub-second, which is why the new tests don't catch it.

Deduct each time from the original budget instead:

```rust
let budget = query_timeout;
flush_pending_unprepare(dbc, stmt, &mut client, "SQLExecDirectW", budget);
let query_timeout = match deduct_query_timeout(budget, started.elapsed()) { ... };
begin_transaction_if_manual(dbc, &mut client, "SQLExecDirectW", query_timeout)?;
let query_timeout = match deduct_query_timeout(budget, started.elapsed()) { ... };
```

(or re-seed `started = Instant::now()` after each deduction). Worth a unit test over `deduct_query_timeout` composed twice, since the wire-level tests can't reach it below one second.

### Suggestion

**`mssql-tds/tests/test_shared_single_worker_runtime.rs:139`** — `expect_err` accepts any error, so the test would still pass on `ConnectionClosed` or a login failure rather than the lock-timeout it names. Its sibling `test_delayed_statement_error_does_not_hang` already matches on `SqlServerError` with `number == 1222`; mirroring that here keeps the assertion tied to the behavior the test is named for.

### Nit / question

**`mssql-odbc/src/api/catalog.rs:559`** (and `describe_param.rs`, `get_type_info.rs`) — hard-coding `0` is a reasonable scope boundary and the inline comments say so, but msodbcsql's catalog functions run through the same `CheckOptions` path and do honor `GetQueryTimeOut`. Is there a follow-up work item, so `SQLTables`/`SQLColumns` don't stay permanently unbounded once the ODBC-level timeout is otherwise honored?

### Note, not a finding

The `TdsError::TimeoutError(_) => HYT00` arm in `post_tds_error` is unconditional, so its blast radius reaches past the execute paths: `SQLDriverConnectW` now reports `HYT00` instead of its `08001` default on a login timeout (which matches msodbcsql's "Login timeout expired"), and `SQLEndTran` reports `HYT00` instead of `08007`. Both look right, but neither is mentioned in the description or CHANGELOG, and the cross-repo `mssql-python` suite is the place a `08001` expectation would surface — worth watching that leg of ADO validation.

Vahid-b:
Thanks for the deep, source-verified review — replying to each point:

### Blocking: double-charged elapsed time — confirmed and fixed

Traced through the code and confirmed the exact mechanism: `started` was captured once before `flush_pending_unprepare` and never reset, so the second `deduct_query_timeout` call measured elapsed time cumulatively from the call's start rather than just the transaction-begin step's own duration, double-charging the unprepare step's cost. Your 10s/3s/2s example reproduces exactly. Fixed in 74f2671 by re-seeding `started = Instant::now()` right after the first deduction succeeds, so the second measurement covers only the step that just ran.

Also added direct unit tests for `deduct_query_timeout`, including one composing it twice with your own numbers (10s budget, 3s + 2s steps -> 5s remaining) — agreed the wire-level tests can't catch this since floor-rounding hides sub-second double-counting and neither existing test delays both pre-execute steps at once (delaying `sp_unprepare` realistically requires a server-tracked prepared handle via a real `sp_prepare`/`sp_prepexec` round trip, which the mock server doesn't model yet — flagging as a possible follow-up if a fully wire-level version of this regression test is wanted).

### Suggestion: `expect_err` too loose — fixed

Tightened `test_shared_single_worker_runtime.rs` to match on `SqlServerError` with `number == 1222`, mirroring `test_delayed_statement_error_does_not_hang`.

### Nit/question: catalog functions staying unbounded

Fair point, and no tracked follow-up work item exists yet for it. Intentionally scoped out of this PR (see the inline comments at each call site) since it's a separate, broader change or affects three additional entry points; agreed it's a legitimate gap versus msodbcsql's `CheckOptions` and worth its own follow-up.

### Note: `HYT00` blast radius on `SQLDriverConnectW` / `SQLEndTran`

Confirmed by tracing the code: `SQLEndTran`'s commit/rollback path (`consume_transaction_response`) reads `self.remaining_request_timeout`, which isn't reset by `commit_transaction`/`rollback_transaction` themselves, so a prior statement's `SQL_ATTR_QUERY_TIMEOUT` can genuinely still be in effect and produce a real `TimeoutError` there — not just a theoretical path. Agree both look right (matches msodbcsql's own timeout reporting), and it's a strictly more precise diagnostic than the previous generic default, not a behavioral regression. Thanks for calling out the `mssql-python` `08001`-expectation risk for ADO validation to watch.

David-Engel (commented):
## Automated review (unattended run — findings were not checked by a human first)

Second pass, at `74ca2b0`. The blocking arithmetic bug from the previous round (double-charged pre-execute elapsed time in `SQLExecDirectW`) is fixed in `74f2671`, the `expect_err` in `test_shared_single_worker_runtime.rs` now matches `SqlServerError`/1222, and the catalog-scope question was answered. Nothing blocking remains. One suggestion about the *shape* of the fix, plus two nits.

**Verified:** checked out at `74ca2b0` with the assigned target dir.

- `cargo nextest run -p mssqlodbc -E 'test(query_timeout)'` — 19/19 pass. The three wire-level tests return in 1.17s / 1.30s / 1.42s against an 8s server delay, so the bound is genuinely enforced, and the new `deduct_query_timeout` unit tests (including the composed-twice case) pass.
- Wrote and ran a throwaway probe against `mssql-mock-tds` asking something no test in the PR covers: **is the connection still usable after a command timeout fires?** It is — `execute(slow, timeout_secs(1))` returns `TimeoutError`, and an immediately following `execute("SELECT 1", ())` on the same client returns `Ok(Rows)`, whole thing in 1.13s. That matters because msodbcsql keeps the connection alive after `HYT00`; it holds here because `fail_rpc_terminator_look_ahead` deliberately exempts `TimeoutError`/`OperationCancelledError` from `retire_after_failed_drain` (`tds_client.rs:4494`). Probe deleted afterwards.
- `gh pr checks 442`: ADO PR validation passes on all legs, including the cross-repo `mssql-python suite on mssql-odbc driver` — which is where the `08001`→`HYT00` blast-radius risk raised last round would have surfaced. Only `coverage-report` fails, after 1h17m, which looks like the same infra timeout as last time rather than a code signal; diff coverage is reported, not enforced.

I also re-derived that `ExecuteOptions::timeout_secs(0)` stays unlimited end-to-end (`deduct_timeout` maps `Some(0)` to `CommandTimeoutBudget::None`, so `remaining_request_timeout` becomes `None`), so the `0` passed by `catalog.rs` / `describe_param.rs` / `get_type_info.rs` and by the default ODBC statement really is a no-op.

### Suggestion

**`mssql-odbc/src/api/exec_direct.rs:176` — re-seeding `started` fixes the double-charge but reintroduces per-step truncation loss; deducting cumulative elapsed from a fixed budget does both.**

`deduct_query_timeout` floors `elapsed` to whole seconds (deliberately, and the doc comment explains why). Re-seeding `started` after each deduction means each step's sub-second remainder is floored to zero *independently*, so a step that costs 0.99s charges nothing. With `SQL_ATTR_QUERY_TIMEOUT = 1`, an `sp_unprepare` taking 0.99s and an implicit `Begin` taking 0.99s each leave the budget at 1, and the execute then gets a fresh 1s — about 2.98s of wall clock for a 1-second timeout. msodbcsql avoids this by deducting in milliseconds (`dwQueryTimeoutInMS` in `DropPrepHandle`).

Keeping `started` fixed at the top of the call and always deducting from the *original* budget charges each step exactly once — which is what the earlier finding asked for — and additionally accumulates the fractions, capping the same scenario at ~1.98s:

```rust
let budget = query_timeout;                 // never reassigned
let started = Instant::now();               // never re-seeded

flush_pending_unprepare(dbc, stmt, &mut client, "SQLExecDirectW", budget);
let query_timeout = match deduct_query_timeout(budget, started.elapsed()) { ... };

begin_transaction_if_manual(dbc, &mut client, "SQLExecDirectW", query_timeout)?;
let query_timeout = match deduct_query_timeout(budget, started.elapsed()) { ... };
```

Worked against the existing unit test's numbers: `deduct(10, 3s) = 7` for the begin, then `deduct(10, 5s) = 5` for the execute — same 5s answer `deduct_query_timeout_composed_twice_charges_each_step_once` asserts, so that test's intent survives (its body would need to pass the cumulative elapsed rather than composing the return value). `execute.rs` deducts only once and is unaffected either way.

Not blocking: the PR's actual guarantee — a blocked statement cannot wait forever — holds regardless, the overshoot is bounded by one second per pre-execute step, and both pre-steps are usually skipped entirely. Exact parity would need sub-second budgets, which `ExecuteOptions::timeout_secs` doesn't currently express; that's follow-up material, not this PR.

### Nits

**`mssql-mock-tds/src/server.rs:213`** — EOF during a delayed response is now a `ProtocolError`. Before this change a client closing its socket mid-request just ended the connection loop; now `wait_out_delay_or_attention` turns it into an error that `?` propagates out of `process_packet`, so `handle_*_connection` returns early and never reaches `connection_store.store(&processor)` — a test asserting on captured connection info would silently lose it. Nothing in the PR hits this (the driver sends `Attention` before dropping), so it's latent. Returning `Ok(None)` and letting the outer loop observe the EOF as it always has would keep the harness's existing shape.

**PR description, Testing section** — it lists `cargo test -p mssql-odbc` and `cargo test -p mssql-mock-tds`. `.github/copilot-instructions.md` ("What NOT to Do") says not to use `cargo test` directly, and the package id is `mssqlodbc`, so those two lines can't be copy-pasted by a reviewer. `cargo nextest run -p mssqlodbc` / `-p mssql-mock-tds` is what actually runs.

### Note, not a finding

`begin_transaction_with_options` sets `remaining_request_timeout` and `cancel_handle` and doesn't clear them on the success path, so a leftover budget can be inherited by a later `commit_transaction` / `rollback_transaction` / `save_transaction`, which set neither. You already traced and accepted exactly this for `SQLEndTran`, so I'm not re-filing it — recording only that the reachable window is narrow (the execute that follows overwrites both fields, and the only path that skips it needs the begin to consume its whole budget without having timed out first).

Not an approval — a human reviewer should still sign off.

mssql-odbc/src/api/exec_direct.rs
```diff
@@ -138,21 +145,62 @@ fn sql_exec_direct_w_safe(
...
+            return fail_with_tds(
+                dbc,
+                stmt,
+                statement_handle,
+                client,
+                &query_timeout_expired_error(),
+            );
+        }
+    };
+    started = Instant::now();
```
David-Engel: Re-seeding `started` here fixes the double-charge, but it also floors each step's sub-second cost to zero independently. With `SQL_ATTR_QUERY_TIMEOUT = 1`, a 0.99s `sp_unprepare` and a 0.99s implicit `Begin` each charge nothing, so the execute still gets a fresh 1s — ~2.98s of wall clock for a 1-second timeout. msodbcsql deducts in milliseconds (`dwQueryTimeoutInMS` in `DropPrepHandle`), so it doesn't lose the remainders.

Keeping `started` fixed at the top of the call and always deducting from the original budget charges each step exactly once *and* accumulates the fractions (same scenario caps at ~1.98s):

```rust
let budget = query_timeout;   // never reassigned
let started = Instant::now(); // never re-seeded
...
let query_timeout = match deduct_query_timeout(budget, started.elapsed()) { ... };
```

Your unit test's numbers still come out the same: `deduct(10, 3s) = 7`, then `deduct(10, 5s) = 5`.

mssql-mock-tds/src/server.rs
```diff
@@ -178,8 +180,65 @@ impl ConnectionProcessor {
...
+        let sleep = tokio::time::sleep(delay);
+        tokio::pin!(sleep);
+        loop {
+            tokio::select! {
+                () = &mut sleep => return Ok(None),
+                read_result = socket.read_buf(&mut self.buffer) => {
+                    let n = read_result?;
+                    if n == 0 {
+                        return Err(ProtocolError::Protocol(
+                            "connection closed while a response was delayed".to_string(),
```
David-Engel: Nit: EOF during a delayed response is now a `ProtocolError` rather than an ordinary connection close. `?` propagates it out of `process_packet`, so `handle_*_connection` returns early and never reaches `connection_store.store(&processor)` — a test that asserts on captured connection info would silently lose it. Nothing in this PR hits it (the driver sends `Attention` first), so it's latent. `Ok(None)` here would let the outer loop see the EOF exactly as it always has.

mssql-odbc/src/api/exec_direct.rs
Vahid-b: Confirmed by re-deriving the arithmetic — re-seeding did avoid the double-charge but floors each step's sub-second cost away independently, so it could still let ~3x the configured timeout elapse in wall-clock terms across two sub-second steps.

Fixed in f772c2f3: `started` is now fixed at the top of the call and both deductions charge against the original, unmodified `budget` using cumulative elapsed time, exactly as you proposed. Verified your worked numbers (`deduct(10,3s)=7`, then `deduct(10,5s)=5`) and updated `deduct_query_timeout_composed_twice_charges_each_step_once` to the new calling convention, plus added `deduct_query_timeout_cumulative_composition_catches_accumulated_sub_second_cost` proving the two-0.99s-steps-against-a-1s-budget scenario is now correctly caught (floors to 0 and errs) instead of silently surviving as it would under the old per-step-reseeded composition.

> [!NOTE]

mssql-mock-tds/src/server.rs
Vahid-b: Agreed, fixed in f772c2f3. Introduced a small `DelayOutcome` enum (`Elapsed` / `Attention` / `Closed`) so `wait_out_delay_or_attention` can distinguish "connection closed mid-delay" from its other two outcomes without conflating it with the "delay elapsed, build your own response" case. All three call sites (`SqlBatch`, `RpcRequest`, `TransactionManager`) now return `Ok(None)` on `Closed`, so the outer read loop discovers the same EOF on its own next read and reaches `connection_store.store(&processor)` exactly as an ordinary non-delayed close does — no `ProtocolError` short-circuit.

> [!NOTE]

Vahid-b:
Thanks for the second pass, and for verifying with real probes. Replying to the remaining two points not tied to a specific line:

**PR description Testing section**: updated to use `cargo nextest run -p mssqlodbc` / `-p mssql-mock-tds` (matching the package rename and the repo's own "don't use `cargo test` directly" convention) instead of the stale `cargo test -p mssql-odbc` commands.

**Note on `begin_transaction_with_options` leaving `remaining_request_timeout`/`cancel_handle` set**: agreed, and as you say I'd already traced and accepted the same shape for `SQLEndTran`'s commit/rollback — no action needed here, thanks for confirming the reachable window is narrow.

The two inline threads (double-deduction sub-second loss, mock-server EOF-during-delay) are fixed and replied to individually.

David-Engel (commented):

Third pass, at `f772c2f`. The only new commit since the last round is `f772c2f` "Deduct query timeout from a fixed budget with cumulative elapsed", and it lands both open items from that round. I have no new findings.

**What changed and how it checks out:**

- **`exec_direct.rs`** — `budget` is now captured once and never reassigned, `started` is captured once and never re-seeded, and both deductions are `deduct_query_timeout(budget, started.elapsed())`. That charges each pre-execute step exactly once *and* accumulates sub-second remainders instead of flooring each one away independently, which is exactly the shape the previous suggestion asked for. The two new unit tests (`deduct_query_timeout_composed_twice_charges_each_step_once`, `deduct_query_timeout_cumulative_composition_catches_accumulated_sub_second_cost`) pin both properties, and the second one is genuinely discriminating — it asserts the per-step-reseeded composition returns `Ok(1)` while the cumulative one returns `Err(())`, so a regression back to re-seeding fails it.
- **`mssql-mock-tds/src/server.rs`** — EOF during a delayed response is no longer a `ProtocolError`. The new `DelayOutcome` enum makes the three cases explicit and `Closed` returns `Ok(None)`, so the outer read loop observes the EOF the same way it always has and still reaches `connection_store.store(&processor)`. This is the latent harness issue from the last round, fixed as suggested.
- **PR description** — the Testing section now uses `cargo nextest run -p mssqlodbc` / `-p mssql-mock-tds` rather than `cargo test`, matching `.github/copilot-instructions.md` and the real package id.

**Verified:** checked out at `f772c2f` with the assigned target dir.

- `cargo nextest run -p mssqlodbc -E 'test(query_timeout) or test(deduct_query_timeout)'` — 20/20 pass. The three wire-level tests return in 1.19s / 1.20s / 1.23s against an 8s mock-server delay, so the bound is enforced rather than merely asserted.
- `cargo nextest run -p mssql-mock-tds` — 18/18 pass, so the `DelayOutcome` refactor and the corrected `Attention` DONE_ATTN framing didn't disturb the existing harness tests.
- `cargo clippy -p mssqlodbc -p mssql-mock-tds --all-targets -- -D warnings` — clean.
- Re-read the untouched parts of the diff (`execute.rs`, `txn.rs`, `sqlstate.rs`, `tds_client.rs`, `test_support.rs`, the two new `mssql-tds` integration tests) against the previous round's notes; nothing regressed, and `test_shared_single_worker_runtime.rs` still matches on `SqlServerError`/1222 under a bounded `tokio::time::timeout`.

### Note, not a finding

`begin_transaction_with_options` still sets `remaining_request_timeout` / `cancel_handle` without clearing them on the success path, so a leftover budget can be inherited by a later `commit_transaction` / `rollback_transaction` / `save_transaction`, none of which set either field. Unchanged from the last round, where you traced and accepted it; recording it only so it stays visible for the human reviewer. The reachable window is narrow — the execute that follows overwrites both fields, and the only path that skips it needs the begin to consume its whole budget without having timed out first.

### CI

`gh pr checks 442` shows `coverage-report` red, but run `33557107758` is **cancelled** (1m6s), not a genuine failure, and diff coverage is reported rather than enforced. The three ADO macOS legs (`Build MacOS`, `Test MacOS`, `Build mssql-python macOS`) are still pending; every other ADO leg passes, including the cross-repo `mssql-python suite on mssql-odbc driver` that would surface the `08001`→`HYT00` blast radius. Because a check is reporting non-green, I did not apply `ready for human review` — worth a re-check once the macOS legs finish and coverage re-runs.

Not an approval — a human reviewer should still sign off.

Vahid-b:
Thanks for confirming both fixes land cleanly. On the `coverage-report` CI note: confirmed via the Actions API that run `33557107758` has `conclusion: cancelled` (not a genuine failure) — it's non-required/informational per the branch ruleset anyway, so not blocking. The macOS legs (`Build MacOS`, `Test MacOS`, `Build mssql-python macOS`) are still running; will keep an eye on them.

mssql-odbc/src/api/catalog.rs
```diff
@@ -553,9 +553,12 @@ fn run_catalog(
         Ok(client) => client,
         Err(rc) => return rc,
     };
-    flush_pending_unprepare(dbc, stmt, &mut client, name);
+    // Catalog functions don't share issue #439's blocked-statement scenario and
+    // are out of scope for the SQL_ATTR_QUERY_TIMEOUT wiring below; `0` keeps
+    // their existing unbounded behavior.
```
David-Engel, line 558: Is there an issue or work item that tracks this gap? As it is, this PR is most of the implementation of AB#46385. At the very least, there should be a child task under AB#46385 calling this gap out so people don't think this PR implements everything. Any other query timeout gaps that have been flagged as out of scope for this PR should get a work item there, too.

mssql-odbc/src/api/describe_param.rs
```diff
@@ -169,9 +169,11 @@ fn sql_describe_param_safe(
         Ok(client) => client,
         Err(rc) => return rc,
     };
-    flush_pending_unprepare(dbc, stmt, &mut client, "SQLDescribeParam");
+    // Not SQLExecute/SQLExecDirectW, so out of scope for the
+    // SQL_ATTR_QUERY_TIMEOUT wiring; `0` keeps existing unbounded behavior.
```
David-Engel, line 173: Here's another comment that should turn into a child task under AB#46385 calling this gap out.

mssql-odbc/src/api/get_type_info.rs
```diff
@@ -181,9 +181,11 @@ fn sql_get_type_info_w_safe(
     };
 
     // Release any handle orphaned by the reset above before running the RPC.
-    flush_pending_unprepare(dbc, stmt, &mut client, "SQLGetTypeInfoW");
+    // Not SQLExecute/SQLExecDirectW, so out of scope for the
+    // SQL_ATTR_QUERY_TIMEOUT wiring; `0` keeps existing unbounded behavior.
```
David-Engel, line 185: One more comment that should turn into a child task under AB#46385 calling this gap out.

Vahid-b: Same tracking issue covers this one too — filed #466 (references AB#46385 as the parent), covering `catalog.rs`, `describe_param.rs`, and `get_type_info.rs` together.

> [!NOTE]

mssql-odbc/src/api/describe_param.rs
Vahid-b: Same tracking issue covers this one too — filed #466 (references AB#46385 as the parent), covering `catalog.rs`, `describe_param.rs`, and `get_type_info.rs` together.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
Vahid-b: Filed #466 to track this (and the same gap in `describe_param.rs`/`get_type_info.rs`) — references AB#46385 as the parent and this PR for context, so it's clear #442 doesn't cover every statement-scoped operation.

> [!NOTE]

David-Engel (commented):
Two things I'd like to see before approval:

1. Follow-up tasks under AB#46385 for the out of scope implementation gaps.
2. Code coverage appears to be missing coverage on timed out error paths. Covering these will likely get the diff % closer to 90%. Aim even higher, if it's not difficult to cover more.

Vahid-b:
Addressing both points:

**1. Follow-up tasks under AB#46385**: I don't have write access to Azure DevOps boards from this environment, so I can't create a child work item there directly. I filed GitHub issue #466 as the best available tracking mechanism — it documents the exact gap (catalog functions, `SQLDescribeParam`, `SQLGetTypeInfoW` still pass `0`/unlimited) and references AB#46385 as the parent. If a literal ADO child task is required by process, whoever has board access will need to create/link one — happy to have that issue closed once it's mirrored there, or kept as the tracking record if that's sufficient.

**2. Coverage on timed-out error paths**: Added a direct unit test for `query_timeout_expired_error()` (the pure error-construction path at `exec_common.rs:415-419`) in 87b1574. For the remaining uncovered lines — the `fail_with_tds(..., &query_timeout_expired_error())` call sites themselves in `execute.rs`/`exec_direct.rs` — I looked into covering these with a real wire-level test and don't think it's achievable without either accepting a flaky test or new mock-server infrastructure:

Each pre-execute step (`flush_pending_unprepare`, `begin_transaction_if_manual`) is *itself* bounded by roughly the same remaining budget passed to the deduction that follows it. For the deduction to see "budget already exhausted" the cumulative elapsed time has to reach the original budget — but by that point the step's *own* internal timeout enforcement (Attention-based) has typically already fired, returning an `Err` directly from that step (a different, already-tested branch) rather than letting it complete just under the wire. Hitting the exact window where a step completes successfully at (not before) the boundary is a genuine race, not something I want to encode as a real-timing test. The one path I can see to real coverage — making `flush_pending_unprepare`'s `sp_unprepare` consume a controlled, delayable amount of time via a genuinely materialized prepared handle — needs the mock server to encode a TDS `RETURNVALUE` token for a real `sp_prepare`/`sp_prepexec` round trip, which it doesn't do today; that felt like a bigger addition than this PR's coverage warrants, so I didn't build it. Filed it as a follow-up note on #466 rather than attempting something fragile here.

David-Engel (commented):

Fourth pass, at `87b1574`. The only new commit since the last round is `87b1574` "Add a direct unit test for `query_timeout_expired_error`", which answers the coverage half of my previous top-level ask. No new findings.

### What changed and how it checks out

**`exec_common.rs`** — `query_timeout_expired_error_is_a_timeout_error` covers the previously-uncovered `exec_common.rs:415-419` and pins the property that actually matters downstream: the constructor returns `TdsError::TimeoutError(TimeoutErrorType::String(_))`, which is the variant `post_tds_error` maps to `HYT00`. Changing that variant to, say, `UsageError` would make the pre-execute exhaustion path report the caller's `default` SQLSTATE instead of `HYT00` and this test would catch it, so it is not vacuous — it is the unit-level companion to `post_tds_error_timeout_maps_to_hyt00_regardless_of_default`, and together they close the loop from "budget exhausted before the wire call" to the SQLSTATE the application sees. Nothing else in the tree changed.

**Your reasoning on the remaining uncovered lines is sound.** The `fail_with_tds(..., &query_timeout_expired_error())` call sites in `execute.rs` / `exec_direct.rs` are genuinely hard to reach deterministically: each pre-execute step is itself bounded by roughly the budget the following deduction reads, so the step's own Attention-based enforcement fires first and returns `Err` from a different (already-covered) branch. Hitting the window where a step completes successfully at exactly the boundary is a race. I traced this myself and agree — encoding it as a real-timing test would buy a handful of coverage lines at the cost of a flaky one. Deferring the `sp_prepare`/`sp_prepexec` `RETURNVALUE` mock work to #466 is the right call for this PR's scope.

The coverage bot now reports **85% diff / 93.3% overall**. It didn't move with this commit because the newly covered lines are a small fraction of a 494-line diff, and the bulk of what's still missing is `mssql-mock-tds/src/server.rs` (74.8%) — test harness, not driver code. Diff coverage is reported, not enforced.

### Verified

Checked out at `87b1574` with the assigned target dir.

- `cargo nextest run -p mssqlodbc -E 'test(query_timeout) or test(deduct_query_timeout) or test(hyt00)'` — **22/22 pass**, including the new unit test. The three wire-level tests return in 1.28s / 1.32s / 1.49s against an 8s mock-server delay, so the bound is genuinely enforced rather than merely asserted.
- `cargo clippy -p mssqlodbc -p mssql-mock-tds --all-targets -- -D warnings` — clean.
- Re-read the full merge-base diff (`origin/main...87b1574`, 21 files) rather than only the delta. `exec_direct.rs` still deducts both steps from a fixed `budget` with cumulative `started.elapsed()`; `execute.rs` deducts once per arm after `begin_transaction_if_manual` and restores `prepared`/`pending_unprepare` before failing, so an exhausted budget leaves the statement re-executable rather than stranded; `begin_transaction` still delegates to `begin_transaction_with_options` with `()`, keeping its existing call sites unbounded as before; the `DelayOutcome::Closed => Ok(None)` handling is present at all three mock-server call sites. All three of `catalog.rs` / `describe_param.rs` / `get_type_info.rs` pass `0`, which is the documented scope boundary now tracked by #466.

### CI

`gh pr checks 442` is green on every completed leg, including the cross-repo `mssql-python suite on mssql-odbc driver` and `coverage-report` (which was the cancelled run last round). Only `Build Stage Test MacOS` and its parent rollup are still pending — nothing failing.

### Note, not a finding

Unchanged from prior rounds and recorded only so it stays visible: `begin_transaction_with_options` sets `remaining_request_timeout` / `cancel_handle` and doesn't clear them on the success path, so a leftover budget can be inherited by a later `commit_transaction` / `rollback_transaction` / `save_transaction`. You traced and accepted the same shape for `SQLEndTran`; the reachable window is narrow, since the execute that follows overwrites both fields.

### Tracking follow-ups

Issue #466 covers the `catalog.rs` / `describe_param.rs` / `get_type_info.rs` gap and now also carries the mock-server `sp_prepare` note. Mirroring it as an ADO child task under AB#46385 still needs someone with board access — that's a process step outside this PR, not a code change.

Not an approval — a human reviewer should still sign off.

Vahid-b:
Thanks for the thorough fourth pass and for confirming the reasoning on the remaining uncovered lines holds up independently. Nothing further to do here — appreciate the diligence across all four rounds.

David-Engel (approved):

PR #446, opened 2026-09-01
Optimize row-wise SQLGetData

saurabh500:
## Description

This draft removes the remaining overhead in row-wise `SQLGetData` on top of the result-set buffering merged through #418:

- caches UTF-16 column names instead of allocating on every `SQLDescribeColW` call;
- keeps the TDS client installed when a requested value is already synchronously buffered;
- captures complete eligible rows during single-row `SQLFetch`;
- captures the contiguous inline prefix of mixed rows while leaving trailing PLP values deferred for normal streaming;
- delivers common buffered numeric, temporal, GUID, binary, and string conversions directly;
- safely clears recycled row slots when deferred values are not overwritten.

The mixed-row path mirrors msodbcsql's split between materialized fixed columns and deferred unbound long data. Leading-PLP/no-prefix, encrypted, bound, multi-row, incomplete, retry, cancellation, and terminal-exhaustion paths retain their existing behavior.

## Performance

Official Linux ADO run [171863](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=171863&view=results) at commit `49f2ccb4`, five repetitions per workload, packet size 16192, Microsoft ODBC Driver 18.6.2.1, and pinned Rust baseline `c453af2c`.

| Row-wise `SQLGetData` workload | Candidate | Microsoft ODBC | Candidate vs Microsoft | Candidate vs pinned Rust baseline |
|---|---:|---:|---:|---:|
| 20k × 16 mixed fixed + small `NVARCHAR(MAX)` | 111.95 ms | 132.97 ms | **15.81% faster** | 46.38% faster |
| 20k × 15 fixed inline values | 105.92 ms | 125.67 ms | **15.72% faster** | 36.20% faster |
| 20k × 4 `sql_variant` | 50.98 ms | 52.84 ms | **3.53% faster** | 24.33% faster |
| 1k × 3 chunked MAX LOBs | 58.60 ms | 54.89 ms | 6.77% slower | 0.19% slower |

Twelve of the thirteen benchmarks in the catalog now beat Microsoft ODBC on hosted Linux, including every `fetch` workload (from 1.54x up to 9.37x). The regression gate reported no confirmed benchmark at least 5% slower than the pinned Rust baseline.

[...]

PR #447, opened 2026-09-01
Keep buffered SQLGetData clients installed

saurabh500:
## Description

Avoid checking the TDS client out of `DbcState` when `SQLGetData` can decode the requested column entirely from buffered bytes. `resume_row_to_column` now polls the installed client while holding the DBC mutex, releases the mutex before applying the existing STMT state and diagnostic handling, and preserves the prior checkout/block/restore flow only for `CursorPoll::Pending`.

The Ready and Pending paths share one `TdsResult<CursorColumn>` handler. Immediate TDS errors also leave the client installed. Focused tests cover the installed-client Ready path, its captured value/state, and the Pending fallback restoration path.

The exact `getdata/rowwise_20k_c15_fixed/inline_values/iterations:1/manual_time` workload was measured for five repetitions against rootless Podman SQL Server `sql1`:

| Driver | Median |
|---|---:|
| Direct parent `e929095bdc4305800beb9394423fca16e2d6efad` | 152.265285 ms |
| This candidate | 140.249056 ms |
| Microsoft ODBC 18.6.2.1 | 118.494657 ms |

The candidate is 7.8916% faster than the direct parent median and 18.3590% slower than Microsoft ODBC.

Validation:

- `cargo fmt --all -- --check`
- `cargo check --workspace --frozen --all-features --all-targets`
- `cargo clippy --workspace --frozen --all-features --all-targets -- -D warnings`
- `cargo nextest run --frozen -p mssql-odbc --lib --no-fail-fast` (1,112 passed, 1 skipped)
- focused `mssql-tds` cursor polling tests (7 passed)

## Related Issues

Stacked on #418.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes

### DIFF
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -20,7 +20,8 @@ use crate::api::util::{copy_with_nul, write_if_some};
 use crate::error::{free_errors, post_sql_error};
 use crate::handles::stmt::{ActivePlpStream, STMT_STATE_CURSOR_OPEN, StmtState};
 use crate::handles::{HandleType, StmtHandle, handle_from_raw};
-use mssql_tds::connection::tds_client::{CursorColumn, PlpChunk};
+use mssql_tds::connection::tds_client::{CursorColumn, CursorPoll, PlpChunk};
+use mssql_tds::core::TdsResult;
 use mssql_tds::encoding_rs::{self, Decoder};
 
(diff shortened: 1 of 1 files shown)


saurabh500:
Closing this parallel experiment in favor of draft PR #446, which is the smaller optimization and reproduced its benchmark improvement in 4/4 confirmation rounds.

PR #449, opened 2026-09-01
Avoid SQLDescribeCol UTF-16 allocations

saurabh500:
## Description

`SQLGetData` consumers re-describe every column for every row, causing `SQLDescribeColW` to allocate a temporary UTF-16 vector for each cell. Encode column names directly into the caller buffer while preserving UTF-16 code-unit lengths, NUL termination, truncation diagnostics, null-buffer length queries, and surrogate-pair behavior.

Exact workload, seven repetitions:

| Driver | Median | Change vs. parent |
|---|---:|---:|
| Microsoft ODBC 18.6.2.1 | 112.908377 ms | -31.60% |
| PR #418 direct parent (`e929095b`) | 165.063205 ms | baseline |
| This PR (`5abd501d`) | 145.534042 ms | -11.83% |

The optimized build remains 28.90% slower than Microsoft ODBC. The >=5% improvement reproduced in all four confirmation rounds with five repetitions per leg; the median confirmation improvement was 8.70%.

## Related Issues

Depends on #418.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes

Validation used `cargo nextest run -p mssql-odbc --frozen --no-fail-fast` (1,113 passed, 1 skipped). There are no public API changes.

### DIFF
--- a/mssql-odbc-bench/perf-lab/baseline-commit.txt
+++ b/mssql-odbc-bench/perf-lab/baseline-commit.txt
@@ -1,10 +1,9 @@
 # Performance baseline commit for mssql-odbc.
 #
-# The initial pin is the production main commit on which the benchmark harness
-# was introduced. PR #417 changes no production driver sources, so this bootstrap
-# makes the first candidate and baseline production trees identical. Future
+# PR #417's merge commit is the production parent of the optimization diff. This
+# pin builds the merged benchmark-era driver as the comparison baseline. Future
 # advances require a pull request, so every baseline move is reviewed and recorded.
 #
 # Exactly one full 40-character commit SHA on its own line. Lines starting with
 # '#' and blank lines are ignored.
-bc4b2dc80b111c388ac8000b90f1c35290d735a0
+acb3ed1dd22a2f9277cf14636fb847273a7a454c
--- a/mssql-odbc/Cargo.toml
+++ b/mssql-odbc/Cargo.toml
@@ -32,6 +32,7 @@ url = "2"
 # connection silently reuse a stale cached credential built from a different
 # secret.
 sha2 = "0.11.0"
+# `RowWriter` exposes UUID values without first boxing them as `ColumnValues`.
 uuid = "1"
 
 # `ActiveDirectoryInteractive` delegates the sign-in UI to `mssql-auth.dll`
--- a/mssql-odbc/src/api/describe_col.rs
+++ b/mssql-odbc/src/api/describe_col.rs
@@ -17,7 +17,7 @@ use crate::api::odbc_types::{
 use crate::api::sqlstate::{
     ERR_FUNCTION_SEQUENCE, ERR_INVALID_DESCRIPTOR_INDEX, WARN_STRING_TRUNCATION, post_diag,
 };
-use crate::api::util::{copy_with_nul, write_if_some};
+use crate::api::util::{copy_str_utf16_with_nul, write_if_some};
 use crate::error::free_errors;
 use crate::handles::stmt::STMT_STATE_EXEC_CONTEXT;
 use crate::handles::{HandleType, StmtHandle, handle_from_raw};
@@ -137,11 +137,15 @@ fn sql_describe_col_w_safe(
 
     let meta = &stmt_state.column_metadata[(column_number - 1) as usize];
 
-    let name_utf16: Vec<u16> = meta.column_name.encode_utf16().collect();
-    let name_len = SqlSmallInt::try_from(name_utf16.len()).unwrap_or(SqlSmallInt::MAX);
+    let (name_len, truncated) = unsafe {
+        copy_str_utf16_with_nul(
+            column_name,
+            usize::try_from(buffer_length).unwrap_or(0),
+            &meta.column_name,
+        )
+    };
+    let name_len = SqlSmallInt::try_from(name_len).unwrap_or(SqlSmallInt::MAX);
     unsafe { write_if_some(name_length_ptr, name_len) };
-
-    let truncated = unsafe { copy_with_nul(column_name, buffer_length as usize, &name_utf16) };
 
     unsafe { write_if_some(data_type_ptr, odbc_sql_type(meta)) };
     unsafe { write_if_some(column_size_ptr, column_size(meta)) };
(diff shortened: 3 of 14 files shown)


David-Engel (commented):

## Summary

The substantive change in this PR is the single head commit `5abd501d` (the other 13 commits are the now-squash-merged #418 riding along). It replaces the per-call `Vec<u16>` in `SQLDescribeColW` with `copy_str_utf16_with_nul`, which encodes `meta.column_name` straight into the caller's buffer. I walked the old and new paths against each other for all four input shapes and they agree: null `dst` returns the full code-unit count and no truncation; `buf_len == 0` writes nothing and reports truncation when the name is non-empty; otherwise at most `buf_len - 1` code units are copied followed by a NUL, with truncation reported on code units (so a surrogate pair still splits exactly as `copy_with_nul` split it). Truncation diagnostics and `01004`/`SQL_SUCCESS_WITH_INFO` are unchanged. The code itself looks correct.

The gap is merge readiness, not the change: PR validation never ran on this SHA.

## Verification

- `git diff origin/main...HEAD` and `git show 5abd501d` in a dedicated worktree; read `describe_col.rs`, `util.rs`, and the other eight `copy_with_nul` call sites.
- `cargo clippy -p mssql-odbc --all-targets --frozen -- -D warnings` — clean.
- `cargo nextest run -p mssql-odbc --lib -E 'test(copy_str_utf16)'` — 3 passed.
- Mutation check on the new helper: `buf_len.saturating_sub(1)` → `saturating_sub(2)` fails `copy_str_utf16_with_nul_reports_code_unit_truncation`. The off-by-one that matters is guarded.
- Second mutation, to prove the gap in Suggestion 1 rather than assert it: deleting the `if buf_len > 0` guard left 123/123 matching `mssql-odbc` lib tests green.
- Not run: the e2e `describe_col_test.cpp` suite, which is where the `SQLDescribeColW` name copy is actually exercised end to end. No live SQL Server here.

## Blocking

**Branch conflicts with `main`, so validation never ran.** `mergeable_state` is `dirty` and the `mssql-rs Pull request validation` check on `5abd501d` reports `action_required` — *"Skipped due to merge conflicts"*. #418 was squash-merged, so the 13 commits this branch inherited from it now duplicate content that reached `main` under a different SHA, and `main` has moved on (`e1620ace`, `5c412918`, `2d5313cd` all touch `mssql-odbc`). Merge `origin/main` in — merge rather than rebase, per repo preference, so later reviewers can see what changed since their last look — and let validation run. Until it does, the checklist's `cargo bfmt` / `cargo bclippy` boxes are only backed by local runs, and the diff will keep rendering as ~4,450 lines instead of the ~75 you actually wrote.

## Suggestions

**1. `mssql-odbc/src/api/util.rs:109` — the zero-length-buffer case is untested, and it is the one that would be an out-of-bounds write.** The `if buf_len > 0` guard is what stops the NUL from landing at `dst[0]` when the caller passes a non-null `ColumnName` with `BufferLength = 0`. None of the three new tests cover that shape: `copy_str_utf16_with_nul_supports_length_queries` uses a null pointer, so it exits at the `dst.is_null()` early return and never reaches the guard. I deleted the guard and all 123 matching lib tests still passed. The helper `copy_str_utf16_with_nul` replaces for this path, `copy_with_nul`, does have that test (`copy_with_nul(buf.as_mut_ptr(), 0, &src)` at `util.rs:647`), so this is coverage lost in the swap rather than coverage never had. One test restores it:

```rust
#[test]
fn copy_str_utf16_with_nul_zero_len_writes_nothing() {
    let mut buf = [0xFFFF_u16; 2];
    let (len, truncated) = unsafe { copy_str_utf16_with_nul(buf.as_mut_ptr(), 0, "ab") };
    assert_eq!(len, 2);
    assert!(truncated);
    assert_eq!(buf, [0xFFFF, 0xFFFF]); // guard held; nothing was written
}
```

An exact-fit case (`buf_len == encoded_len + 1`, expecting `truncated == false` and the NUL at the last slot) would pin the other boundary cheaply while you are in there.

**2. `mssql-odbc/src/api/describe_col.rs:143` — a real hardening is buried in the refactor and unnamed in the description.** The old line was `copy_with_nul(column_name, buffer_length as usize, &name_utf16)`. `buffer_length` is `SqlSmallInt`, so a negative value sign-extended through `as usize` into a near-`usize::MAX` capacity, and `copy_len = src.len().min(buf_len - 1)` then copied the whole name unconditionally — an unbounded write bounded only by the DM's HY090 check and the `debug_assert!` at `describe_col.rs:116`, neither of which holds in a release build called by a driver-manager-bypassing app. The new `usize::try_from(buffer_length).unwrap_or(0)` folds it to "no room", matching the convention and comment already at `util.rs:213-215`. That is a good change and the right one; it just deserves a line in the PR description, since a memory-safety fix landing inside a perf PR is exactly the kind of thing that gets lost when someone bisects later.

## Nits

**`.github/instructions/pr-workflow.instructions.md:14`** asks for a linked GitHub issue or ADO work item in the description; "Related Issues: Depends on #418" records a dependency, not a work item, and #418 is now merged. If this belongs to the row-decode perf issue tree, linking the sub-issue would also give the timing table in the description a home. The unchecked `cargo btest` box is adequately explained by the `cargo nextest run -p mssql-odbc` note right below it — no action needed there beyond re-checking it once CI runs.

## CI

| Check | Conclusion | Detail |
|---|---|---|
| `mssql-rs Pull request validation` | `action_required` | Skipped due to merge conflicts |
| `license/cla` | `success` | All CLA requirements met |

No coverage report comment is present, which is expected: this PR touches `mssql-odbc/**` and `mssql-tds/**`, so it is not path-excluded — the report simply cannot generate while validation is skipped. It should appear once the conflicts are resolved.

---

This was an unattended run; nothing here has been checked by a human first, and it is a comment, not an approval.

David-Engel (commented):
Superseded — this was a duplicate automated run against the same head commit `5abd501d`. The earlier automated review above already covers the same ground (including the `buf_len == 0` test gap in `copy_str_utf16_with_nul`) in more detail. Please read that one; nothing new here.

PR #455, opened 2026-09-01
Advance ODBC benchmark baseline

saurabh500:
## Description

Advance `mssql-odbc-bench/perf-lab/baseline-commit.txt` to the exact merge commit from [PR #418](https://github.com/microsoft/mssql-rs/pull/418). The baseline now represents the merged, accepted optimized production driver state. Future performance runs will compare new candidates against this accepted state, and later baseline moves still require reviewed pull requests.

Validation:

- Confirmed PR #418 remains merged at `c453af2c95ddc5bb5525634ac0112474e9610378`.
- Confirmed that commit is an ancestor of `origin/main` and resolves as a commit object.
- Confirmed the branch started at current `origin/main`.
- `git diff --check` passes.
- Confirmed the exact changed-file allowlist contains only `mssql-odbc-bench/perf-lab/baseline-commit.txt`.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47608

## Checklist

- [ ] `cargo bfmt` passes — not run; text-only baseline metadata change
- [ ] `cargo bclippy` passes — not run; text-only baseline metadata change
- [ ] `cargo btest` passes — not run; text-only baseline metadata change
- [ ] New/changed functionality has tests — not applicable
- [ ] Public API changes are documented — not applicable

PR #459, opened 2026-09-01
mssql-odbc: shut down the ENV runtime without joining its worker thread

### LINKED ISSUE #445: mssql-odbc: SQLFreeHandle(ENV) can still block indefinitely on a leftover spawn_blocking DNS lookup
### Describe the bug

Discovered during automated re-review of PR #432 (fix for [AB#47704](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47704), "Failed invalid-password connection hangs test process during cleanup"). That PR fixes the *reported* symptom — confirmed via live CI logs, `test_connection_timeout_invalid_password` now passes in ~100ms instead of hanging — by switching blocking `std::net::ToSocketAddrs` DNS resolution to `tokio::net::lookup_host` in `mssql-tds`'s connect paths, so a slow/stuck resolver no longer defeats the connect/login timeout.

However, `tokio::net::lookup_host` still runs the actual `getaddrinfo` call via `tokio::task::spawn_blocking` under the hood. Dropping/abandoning the *awaiting* future (e.g. because our outer `timeout()` fired) does **not** cancel the underlying blocking OS thread — it keeps running until `getaddrinfo` itself returns, however long that takes. Tokio is explicit about the cost this imposes at runtime shutdown (`tokio::runtime::Runtime` docs): *"Blocking functions spawned through `Runtime::spawn_blocking` keep running until they return. The thread initiating the shutdown blocks until all spawned work has been stopped."*

`mssql-odbc` owns its Tokio runtime as `Arc<Runtime>` on the `EnvHandle` (`mssql-odbc/src/handles/env.rs`), shared with every child `DbcHandle`, and drops it via plain `Drop` semantics when the environment is freed (`mssql-odbc/src/api/free_handle.rs::free_env`, which just calls `free_handle::<EnvHandle>(handle)` — no custom `Drop` impl, no `shutdown_timeout`/`shutdown_background` call anywhere in `mssql-odbc/src/handles`).

So: if a connect attempt's DNS resolution is genuinely wedged (not just slow — e.g. a resolver that never replies and isn't itself bounded by any OS-level timeout) when the caller times out and moves on to free all DBCs and then the environment, `SQLFreeHandle(ENV)` — the "cleanup" phase named in the original bug title — can still block indefinitely on that leftover `spawn_blocking` thread. The connect attempt itself now correctly fails within its configured timeout (PR #432's fix), but process-level cleanup is not unconditionally guaranteed to be prompt in this specific edge case.

### Steps to reproduce

Not reproduced with a live repro (would need a resolver that hangs forever without its own OS-level timeout, which is hard to construct portably). This is a code-level finding, not an empirically observed hang.

1. Start a `SQLDriverConnectW` connect attempt against a host whose DNS resolution never returns (simulated resolver hang, no OS-level timeout).
2. Let the connect/login timeout fire — the connect attempt correctly fails with a timeout error.
3. Free all DBCs, then free the ENV (`SQLFreeHandle(SQL_HANDLE_ENV, ...)`), dropping the last `Arc<Runtime>` reference.
4. Expected: `SQLFreeHandle(ENV)` returns promptly.
5. Actual (predicted from Tokio's documented `spawn_blocking`/shutdown semantics): `SQLFreeHandle(ENV)` blocks until the leftover `getaddrinfo` call on the blocking-pool thread eventually returns (or never, if it's truly wedged).

### Expected behavior

Freeing the ODBC environment should be bounded even if a prior connect attempt left a `spawn_blocking` DNS lookup running in the background.

### Actual behavior

Environment teardown relies on `Runtime`'s default `Drop`, which blocks the calling thread until all outstanding `spawn_blocking` work completes, with no timeout.

### Suggested direction

At `EnvHandle` teardown (`free_env` in `mssql-odbc/src/api/free_handle.rs`), use `Runtime::shutdown_timeout(Duration)` instead of relying on plain `Drop`, so teardown is bounded even if a blocking-pool thread is still stuck. This likely needs `Arc::try_unwrap` (or equivalent) to get an owned `Runtime` before calling `shutdown_timeout`, since the runtime is currently shared via `Arc<Runtime>` across all child `DbcHandle`s — `shutdown_timeout` takes `self` by value. The existing `debug_assert!` in `free_env` already expects all child DBCs to be freed first, so the `Arc` refcount should be 1 by then in the normal (non-debug-assert-violating) path.

### Version

Commit af805bbb (PR #432 head at time of discovery), main branch otherwise unmodified.

### Affected crate

mssql-odbc

### Additional context

- Not a regression from PR #432 — the plain-`Drop`-based teardown predates it. PR #432 makes the DNS step cancellable/yielding (fixing the reported connect-time hang) but doesn't change how leftover `spawn_blocking` work is handled at shutdown, since that's a different layer (`mssql-odbc`'s runtime ownership, not `mssql-tds`'s connect logic).
- Related but distinct from #439 (a different, previously-masked query-execution hang also found while validating PR #432).

David-Engel:
## Description

`EnvHandle` owns a shared Tokio multi-thread runtime (one background worker thread) used by every DBC allocated under that ENV. `SQLFreeHandle(SQL_HANDLE_ENV)` can run long after the OS has already force-terminated background threads — e.g. a caller that loads this driver directly instead of through the ODBC Driver Manager may defer the free to a C++ static destructor that only runs at `DLL_PROCESS_DETACH`, by which point Windows has already killed every thread but the one tearing the process down. The default `Runtime` drop unconditionally joins its worker thread, which panics ("threads should not terminate unexpectedly") if that thread no longer exists. This is exactly what mssql-python observes: a panic printed at process teardown when the driver is loaded directly, never when driven through the Driver Manager.

**Fix:** introduce a `SharedRuntime(ManuallyDrop<Runtime>)` newtype in `handles::env` that owns the teardown policy, and change `EnvHandle.runtime` to `Arc<SharedRuntime>`. `SharedRuntime::drop` takes the runtime by value and calls `shutdown_background()`, which detaches the worker thread instead of joining it, so it is safe no matter when the last handle is released relative to OS thread teardown. `SharedRuntime` derefs to `Runtime`, so call sites are unchanged apart from types.

Putting the teardown on the shared value rather than on `EnvHandle::drop` is the important part. An earlier iteration of this fix ran `shutdown_background()` from `impl Drop for EnvHandle` via `Arc::try_unwrap`, which only works when the ENV holds the sole reference. `free_env` only `debug_assert!`s that `connections.is_empty()`, so in a release build a non-conformant caller that frees the ENV while a DBC is still alive makes `try_unwrap` fail, silently skips `shutdown_background()`, and later drops the runtime through the DBC's `Arc` on the default joining path — the exact teardown [AB#47509](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47509) panics on. That path is reachable, not theoretical: a local repro showed the straggler `Arc` release blocking for 5.0006575s on the joining path. With the newtype, whichever owner happens to be last always detaches.

Deliberately did not switch the runtime to `new_current_thread()` — the ENV runtime is shared across DBCs that may run on different app threads concurrently, and that would risk a real concurrency regression for a cosmetic-panic fix.

Removing the `Option` also removed the `.expect("ENV runtime already shut down")` that `alloc_dbc` needed in order to unwrap it, which was a panic on a production FFI path and against this crate's panic-free rule.

### Why this also closes #445

`Runtime`'s default drop joins **blocking** tasks specifically — spawned async tasks are dropped at shutdown either way — so the entire behavioural delta of `shutdown_background()` is "stop waiting on in-flight `spawn_blocking` work." That work is:

- `tokio::net::lookup_host`'s `getaddrinfo`, in `mssql-tds` (`ssrp.rs`, `network_transport.rs`, `parallel_connect.rs`)
- the named-pipe connect in `mssql-tds`'s `named_pipes.rs`
- Entra interactive sign-in in this crate, `auth/interactive.rs:144` — the longest of the three, since `msqa::acquire_token` pumps a Win32 message loop for the full duration of a human sign-in

#445 is the DNS instance of exactly the wait this removes; [AB#47509](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47509) is the panic instance. One change, both symptoms — and `park_blocking_task` in the new tests is #445's scenario with the wedged resolver replaced by a `sleep`.

Worth noting for anyone returning to #445: its "Suggested direction" was `Arc::try_unwrap` to get an owned `Runtime`, then `shutdown_timeout`, on the reasoning that `free_env`'s `debug_assert!` means the refcount is 1 by then. That is the approach this PR tried and rejected — a `debug_assert!` is not a guarantee in a release build, so `try_unwrap` can fail exactly when it matters, and `dropping_a_dbc_that_outlived_its_env_does_not_wait_for_blocking_work` is the regression test for that hole.

### Trade-off

Detaching means a blocking-pool thread can now outlive ENV teardown instead of blocking it. That is a deliberate choice, not a free win. Concretely the two cases are a wedged `getaddrinfo` and — the longer one — an Entra interactive sign-in still pumping its message loop. The sign-in case is less bad than it first sounds: `CancelUiOnDrop` (`auth/interactive.rs`) already closes the prompt when the awaiting future is dropped, which shutdown does, so what survives is a blocking-pool thread unwinding after `SQLFreeHandle(ENV)` has returned rather than a prompt orphaned on the user's desktop. For either case to arise at all, a caller must be freeing the ENV while a connect or sign-in is still live on another thread.

It is the right trade for a driver — an unbounded block inside `SQLFreeHandle` is worse than a detached thread in a process that is usually on its way out anyway, and in the [AB#47509](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47509) scenario the thread is already gone. `shutdown_timeout(small)` was the bounded-wait alternative, and it is rejected here because any non-zero wait still attempts the join, which is precisely what panics when the OS has already destroyed the thread.

## Related Issues

- Fixes #445
- https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47509

## Testing

- `cargo bfmt` passes
- `cargo bclippy` passes
- `cargo nextest run -p mssqlodbc --lib`: 1168 passed, 1 skipped. Workspace-wide, the only failures are 7 pre-existing `mssql-tds` TLS tests that need certificate fixtures generated by a CI setup step and absent locally; they are unrelated to this change and green in CI.
- ADO PR validation: 18/18 checks pass, 100% diff coverage.

Two regression tests were added in `handles::env::tests`, both built on a shared `park_blocking_task` helper that parks a `spawn_blocking` task for 5s behind an mpsc handshake confirming the task actually started, then asserts the drop returns in well under that:

- `dropping_env_handle_does_not_wait_for_blocking_work` — guards the ordinary path.
- `dropping_a_dbc_that_outlived_its_env_does_not_wait_for_blocking_work` — guards the `Arc::try_unwrap` hole described above by dropping the ENV first and asserting the surviving DBC's release still detaches.

Both were confirmed to discriminate: each fails after ~5.0s against the default joining drop and passes with `shutdown_background()`. The bounded assertion is deliberate, so a regression fails fast instead of hanging until nextest's slow-timeout. The literal [AB#47509](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47509) failure mode (the OS force-killing the worker thread before drop runs) needs real process/DLL teardown and cannot be reproduced in a unit test; these tests cover the observable proxy, which is that drop never joins.

None. `EnvHandle.runtime` and `SharedRuntime` are private, crate-internal; no public/FFI API changes.

### DIFF
--- a/mssql-odbc/src/api/alloc_handle.rs
+++ b/mssql-odbc/src/api/alloc_handle.rs
@@ -121,7 +121,11 @@ unsafe fn alloc_dbc(input_handle: SqlHandle, output_handle: *mut SqlHandle) -> S
         "SQLAllocHandle(DBC): SQL_ATTR_ODBC_VERSION not set on env"
     );
 
-    let dbc = Box::new(DbcHandle::new(input_handle, env.runtime.clone()));
+    let runtime = env
+        .runtime
+        .clone()
+        .expect("SQLAllocHandle(DBC): ENV runtime already shut down");
+    let dbc = Box::new(DbcHandle::new(input_handle, runtime));
     let raw = handle_to_raw(dbc);
     env_state.connections.push(raw);
 
--- a/mssql-odbc/src/api/alloc_handle.rs
+++ b/mssql-odbc/src/api/alloc_handle.rs
@@ -3,6 +3,8 @@
 
 //! Implementation of SQLAllocHandle — the ODBC handle allocation entry point.
 
+use std::sync::Arc;
+
 use tracing::{debug, error};
 
 use crate::api::odbc_types::{
@@ -121,12 +123,7 @@ unsafe fn alloc_dbc(input_handle: SqlHandle, output_handle: *mut SqlHandle) -> S
         "SQLAllocHandle(DBC): SQL_ATTR_ODBC_VERSION not set on env"
     );
 
-    // `None` only once the ENV is dropping, so a DBC can no longer be parented to it.
-    let Some(runtime) = env.runtime.clone() else {
-        error!("SQLAllocHandle(DBC): ENV runtime already shut down");
-        return SQL_ERROR;
-    };
-    let dbc = Box::new(DbcHandle::new(input_handle, runtime));
+    let dbc = Box::new(DbcHandle::new(input_handle, Arc::clone(&env.runtime)));
     let raw = handle_to_raw(dbc);
     env_state.connections.push(raw);
 
--- a/mssql-odbc/src/api/alloc_handle.rs
+++ b/mssql-odbc/src/api/alloc_handle.rs
@@ -121,10 +121,11 @@ unsafe fn alloc_dbc(input_handle: SqlHandle, output_handle: *mut SqlHandle) -> S
         "SQLAllocHandle(DBC): SQL_ATTR_ODBC_VERSION not set on env"
     );
 
-    let runtime = env
-        .runtime
-        .clone()
-        .expect("SQLAllocHandle(DBC): ENV runtime already shut down");
+    // `None` only once the ENV is dropping, so a DBC can no longer be parented to it.
+    let Some(runtime) = env.runtime.clone() else {
+        error!("SQLAllocHandle(DBC): ENV runtime already shut down");
+        return SQL_ERROR;
+    };
     let dbc = Box::new(DbcHandle::new(input_handle, runtime));
     let raw = handle_to_raw(dbc);
     env_state.connections.push(raw);
--- a/mssql-odbc/src/handles/env.rs
+++ b/mssql-odbc/src/handles/env.rs
@@ -49,8 +49,10 @@ pub(crate) struct EnvHandle {
     pub(crate) object_type: HandleType,
     pub(crate) inner: Mutex<EnvState>,
     /// Shared Tokio runtime for all connections on this ENV.
-    /// Wrapped in `Arc` so DBCs can hold a reference without lifetime issues.
-    pub(crate) runtime: Arc<Runtime>,
+    /// Wrapped in `Arc` so DBCs can hold a reference without lifetime issues,
+    /// and in `Option` so `Drop` can take ownership to shut it down (see the
+    /// `Drop` impl below). Only `None` after this handle starts dropping.
+    pub(crate) runtime: Option<Arc<Runtime>>,
 }
 
 /// Mutable state within an environment handle, protected by `inner`.
@@ -90,7 +92,7 @@ impl EnvHandle {
                 output_nts: true, // SQL_ATTR_OUTPUT_NTS defaults to SQL_TRUE
                 connections: Vec::new(),
             }),
-            runtime: Arc::new(runtime),
+            runtime: Some(Arc::new(runtime)),
         })
     }
 }
@@ -98,5 +100,42 @@ impl EnvHandle {
 impl HasObjectType for EnvHandle {
     fn object_type_mut(&mut self) -> &mut HandleType {
         &mut self.object_type
+    }
+}
+
+impl Drop for EnvHandle {
+    /// Shuts down the per-ENV Tokio runtime without joining its worker thread.
+    ///
+    /// `SQLFreeHandle(SQL_HANDLE_ENV)` can run long after the OS has already
+    /// force-terminated background threads — e.g. a caller that loads this
+    /// driver directly instead of through the ODBC Driver Manager may defer
+    /// the free to a C++ static destructor that only runs at
+    /// `DLL_PROCESS_DETACH`, by which point Windows has already killed every
+    /// thread but the one tearing the process down (AB#47509). The default
+    /// `Runtime` drop unconditionally joins its worker thread, which panics
+    /// ("threads should not terminate unexpectedly") if that thread no
+    /// longer exists. `shutdown_background` detaches instead of joining, so
+    /// it's safe no matter when this runs.
+    fn drop(&mut self) {
+        if let Some(runtime) = self.runtime.take().and_then(|rt| Arc::try_unwrap(rt).ok()) {
+            runtime.shutdown_background();
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Guards against reintroducing a blocking join in `Drop`: this must
+    /// return promptly rather than waiting out the runtime's worker thread,
+    /// which is what `shutdown_background` (vs. the default `Runtime` drop)
+    /// buys us. Can't reproduce the OS-forcibly-killed-the-thread case that
+    /// actually panicked (AB#47509) from a unit test — that needs a real
+    /// process/DLL teardown — so this exercises the ordinary path instead.
+    #[test]
+    fn dropping_env_handle_does_not_hang_or_panic() {
+        let env = EnvHandle::new().expect("failed to create EnvHandle for test");
+        drop(env);
     }
 }
--- a/mssql-odbc/src/handles/env.rs
+++ b/mssql-odbc/src/handles/env.rs
@@ -125,17 +125,43 @@ impl Drop for EnvHandle {
 
 #[cfg(test)]
 mod tests {
+    use std::sync::mpsc;
+    use std::thread;
+    use std::time::{Duration, Instant};
+
     use super::*;
 
-    /// Guards against reintroducing a blocking join in `Drop`: this must
-    /// return promptly rather than waiting out the runtime's worker thread,
-    /// which is what `shutdown_background` (vs. the default `Runtime` drop)
-    /// buys us. Can't reproduce the OS-forcibly-killed-the-thread case that
-    /// actually panicked (AB#47509) from a unit test — that needs a real
-    /// process/DLL teardown — so this exercises the ordinary path instead.
+    /// Guards against reintroducing a blocking join in `Drop`. The default
+    /// `Runtime` drop waits out in-flight blocking work; `shutdown_background`
+    /// detaches from it. An idle runtime joins its idle worker promptly under
+    /// either one, so this parks a blocking task to tell them apart. The
+    /// OS-already-killed-the-thread case that actually panicked (AB#47509)
+    /// needs real process/DLL teardown and isn't reproducible from a unit test.
     #[test]
-    fn dropping_env_handle_does_not_hang_or_panic() {
+    fn dropping_env_handle_does_not_wait_for_blocking_work() {
+        const PARK: Duration = Duration::from_secs(5);
+
         let env = EnvHandle::new().expect("failed to create EnvHandle for test");
+        let (started_tx, started_rx) = mpsc::channel();
+        env.runtime
+            .as_ref()
+            .expect("a live ENV owns its runtime")
+            .spawn_blocking(move || {
+                let _ = started_tx.send(());
+                thread::sleep(PARK);
+            });
+        started_rx
+            .recv_timeout(PARK)
+            .expect("blocking task should have started");
+
+        let start = Instant::now();
         drop(env);
+        let elapsed = start.elapsed();
+
+        assert!(
+            elapsed < PARK / 2,
+            "dropping EnvHandle blocked for {elapsed:?} waiting on the runtime's \
+             blocking work; Drop must detach via shutdown_background, not join"
+        );
     }
 }
--- a/mssql-odbc/src/handles/env.rs
+++ b/mssql-odbc/src/handles/env.rs
@@ -3,6 +3,7 @@
 
 use std::ffi::c_void;
 use std::io;
+use std::mem::ManuallyDrop;
 use std::sync::{Arc, Mutex};
 
 use tokio::runtime::Runtime;
@@ -35,6 +36,51 @@ impl TryFrom<u32> for OdbcVersion {
     }
 }
 
+/// A Tokio runtime that never joins its worker threads when it goes away.
+///
+/// `SQLFreeHandle(SQL_HANDLE_ENV)` can run long after the OS has already
+/// force-terminated background threads — e.g. a caller that loads this driver
+/// directly instead of through the ODBC Driver Manager may defer the free to a
+/// C++ static destructor that only runs at `DLL_PROCESS_DETACH`, by which point
+/// Windows has already killed every thread but the one tearing the process
+/// down (AB#47509). The default `Runtime` drop unconditionally joins its worker
+/// thread, which panics ("threads should not terminate unexpectedly") if that
+/// thread no longer exists. `shutdown_background` detaches instead of joining.
+///
+/// This lives on the shared value rather than on `EnvHandle` so the guarantee
+/// holds for whichever owner happens to release the last reference. A DBC that
+/// outlives its ENV — which a non-conformant caller can produce, since
+/// `SQLFreeHandle(ENV)` only `debug_assert!`s that the connection list is empty
+/// — would otherwise drop the runtime through the default joining path, which
+/// is exactly the teardown this exists to avoid.
+#[derive(Debug)]
+pub(crate) struct SharedRuntime(ManuallyDrop<Runtime>);
+
+impl SharedRuntime {
+    fn new(runtime: Runtime) -> Self {
+        Self(ManuallyDrop::new(runtime))
+    }
+}
+
+impl std::ops::Deref for SharedRuntime {
+    type Target = Runtime;
+
+    fn deref(&self) -> &Runtime {
+        &self.0
+    }
+}
+
+impl Drop for SharedRuntime {
+    fn drop(&mut self) {
+        // SAFETY: `drop` runs at most once and nothing reads `self.0` after
+        // this, so taking ownership out of the `ManuallyDrop` is sound. Taking
+        // it by value is what lets us call `shutdown_background`, which
+        // consumes the `Runtime`; the default drop glue would join instead.
+        let runtime = unsafe { ManuallyDrop::take(&mut self.0) };
+        runtime.shutdown_background();
+    }
+}
+
 /// Environment handle
 ///
 /// One ENV is typically allocated per application. It owns connection handles
@@ -48,11 +94,11 @@ impl TryFrom<u32> for OdbcVersion {
 pub(crate) struct EnvHandle {
     pub(crate) object_type: HandleType,
     pub(crate) inner: Mutex<EnvState>,
-    /// Shared Tokio runtime for all connections on this ENV.
-    /// Wrapped in `Arc` so DBCs can hold a reference without lifetime issues,
-    /// and in `Option` so `Drop` can take ownership to shut it down (see the
-    /// `Drop` impl below). Only `None` after this handle starts dropping.
-    pub(crate) runtime: Option<Arc<Runtime>>,
+    /// Shared Tokio runtime for all connections on this ENV, in an `Arc` so
+    /// DBCs can hold a reference without lifetime issues. Shutdown is handled
+    /// by `SharedRuntime`'s own `Drop`, so this handle needs no `Drop` of its
+    /// own and the field is never vacated while the ENV is live.
+    pub(crate) runtime: Arc<SharedRuntime>,
 }
 
 /// Mutable state within an environment handle, protected by `inner`.
@@ -92,7 +138,7 @@ impl EnvHandle {
                 output_nts: true, // SQL_ATTR_OUTPUT_NTS defaults to SQL_TRUE
                 connections: Vec::new(),
             }),
-            runtime: Some(Arc::new(runtime)),
+            runtime: Arc::new(SharedRuntime::new(runtime)),
         })
     }
 }
@@ -100,26 +146,6 @@ impl EnvHandle {
 impl HasObjectType for EnvHandle {
     fn object_type_mut(&mut self) -> &mut HandleType {
         &mut self.object_type
-    }
-}
-
-impl Drop for EnvHandle {
-    /// Shuts down the per-ENV Tokio runtime without joining its worker thread.
-    ///
-    /// `SQLFreeHandle(SQL_HANDLE_ENV)` can run long after the OS has already
-    /// force-terminated background threads — e.g. a caller that loads this
-    /// driver directly instead of through the ODBC Driver Manager may defer
-    /// the free to a C++ static destructor that only runs at
-    /// `DLL_PROCESS_DETACH`, by which point Windows has already killed every
-    /// thread but the one tearing the process down (AB#47509). The default
-    /// `Runtime` drop unconditionally joins its worker thread, which panics
-    /// ("threads should not terminate unexpectedly") if that thread no
-    /// longer exists. `shutdown_background` detaches instead of joining, so
-    /// it's safe no matter when this runs.
-    fn drop(&mut self) {
-        if let Some(runtime) = self.runtime.take().and_then(|rt| Arc::try_unwrap(rt).ok()) {
-            runtime.shutdown_background();
-        }
     }
 }
 
@@ -131,37 +157,60 @@ mod tests {
 
     use super::*;
 
-    /// Guards against reintroducing a blocking join in `Drop`. The default
-    /// `Runtime` drop waits out in-flight blocking work; `shutdown_background`
-    /// detaches from it. An idle runtime joins its idle worker promptly under
-    /// either one, so this parks a blocking task to tell them apart. The
+    const PARK: Duration = Duration::from_secs(5);
+
+    /// Parks a blocking task on `runtime` and returns once it is confirmed
+    /// running, so a later drop has in-flight blocking work to contend with.
+    fn park_blocking_task(runtime: &SharedRuntime) {
+        let (started_tx, started_rx) = mpsc::channel();
+        runtime.spawn_blocking(move || {
+            let _ = started_tx.send(());
+            thread::sleep(PARK);
+        });
+        started_rx
+            .recv_timeout(PARK)
+            .expect("blocking task should have started");
+    }
+
+    fn assert_released_promptly(what: &str, release: impl FnOnce()) {
+        let start = Instant::now();
+        release();
+        let elapsed = start.elapsed();
+        assert!(
+            elapsed < PARK / 2,
+            "{what} blocked for {elapsed:?} waiting on the runtime's blocking work; \
+             shutdown must detach via shutdown_background, not join"
+        );
+    }
+
+    /// Guards against reintroducing a blocking join. The default `Runtime` drop
+    /// waits out in-flight blocking work; `shutdown_background` detaches from
+    /// it. An idle runtime joins its idle worker promptly under either one, so
+    /// these park a blocking task to tell the two apart. The
     /// OS-already-killed-the-thread case that actually panicked (AB#47509)
     /// needs real process/DLL teardown and isn't reproducible from a unit test.
     #[test]
     fn dropping_env_handle_does_not_wait_for_blocking_work() {
-        const PARK: Duration = Duration::from_secs(5);
-
         let env = EnvHandle::new().expect("failed to create EnvHandle for test");
-        let (started_tx, started_rx) = mpsc::channel();
-        env.runtime
-            .as_ref()
-            .expect("a live ENV owns its runtime")
-            .spawn_blocking(move || {
-                let _ = started_tx.send(());
-                thread::sleep(PARK);
-            });
-        started_rx
-            .recv_timeout(PARK)
-            .expect("blocking task should have started");
+        park_blocking_task(&env.runtime);
 
-        let start = Instant::now();
+        assert_released_promptly("dropping EnvHandle", || drop(env));
+    }
+
+    /// The ENV is not always the last owner: `SQLFreeHandle(ENV)` only
+    /// `debug_assert!`s that the connection list is empty, so a non-conformant
+    /// caller can free the ENV while a DBC still holds a runtime reference.
+    /// Shutdown must still detach when that straggler releases the last one.
+    #[test]
+    fn dropping_a_dbc_that_outlived_its_env_does_not_wait_for_blocking_work() {
+        let env = EnvHandle::new().expect("failed to create EnvHandle for test");
+        let dbc_runtime = Arc::clone(&env.runtime);
+        park_blocking_task(&env.runtime);
+
         drop(env);
-        let elapsed = start.elapsed();
 
-        assert!(
-            elapsed < PARK / 2,
-            "dropping EnvHandle blocked for {elapsed:?} waiting on the runtime's \
-             blocking work; Drop must detach via shutdown_background, not join"
-        );
+        assert_released_promptly("dropping the last DBC runtime reference", || {
+            drop(dbc_runtime)
+        });
     }
 }
--- a/mssql-odbc/src/api/fetch_scroll.rs
+++ b/mssql-odbc/src/api/fetch_scroll.rs
@@ -1104,7 +1104,7 @@ fn trim_partial_utf8(bytes: &mut Vec<u8>) {
 /// Same contract as `deliver_bound`.
 unsafe fn deliver_bound_plp(
     client: &mut mssql_tds::connection::tds_client::TdsClient,
-    runtime: &std::sync::Arc<tokio::runtime::Runtime>,
+    runtime: &tokio::runtime::Runtime,
     binding: &ColumnBinding,
     row_index: usize,
     bind_offset: usize,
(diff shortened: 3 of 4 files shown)


mssql-odbc/src/api/alloc_handle.rs
```diff
@@ -121,7 +121,11 @@ unsafe fn alloc_dbc(input_handle: SqlHandle, output_handle: *mut SqlHandle) -> S
         "SQLAllocHandle(DBC): SQL_ATTR_ODBC_VERSION not set on env"
     );
 
-    let dbc = Box::new(DbcHandle::new(input_handle, env.runtime.clone()));
+    let runtime = env
+        .runtime
+        .clone()
+        .expect("SQLAllocHandle(DBC): ENV runtime already shut down");
```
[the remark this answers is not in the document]
David-Engel: Good catch — this violated the crate's no-panics rule (`.github/instructions/mssql-odbc.instructions.md`: never `.unwrap()`/`.expect()` in non-test code, and `ffi_entry!` is a last-resort net rather than a license to panic).

Replaced with an explicit `let Some(...) else` that logs and returns `SQL_ERROR`, matching the existing mutex-poison style in this function:

```rust
// `None` only once the ENV is dropping, so a DBC can no longer be parented to it.
let Some(runtime) = env.runtime.clone() else {
    error!("SQLAllocHandle(DBC): ENV runtime already shut down");
    return SQL_ERROR;
};
```

Fixed in af32643e.

> [!NOTE]

mssql-odbc/src/handles/env.rs
```diff
@@ -100,3 +102,40 @@ impl HasObjectType for EnvHandle {
...
+    /// Guards against reintroducing a blocking join in `Drop`: this must
+    /// return promptly rather than waiting out the runtime's worker thread,
+    /// which is what `shutdown_background` (vs. the default `Runtime` drop)
+    /// buys us. Can't reproduce the OS-forcibly-killed-the-thread case that
+    /// actually panicked (AB#47509) from a unit test — that needs a real
+    /// process/DLL teardown — so this exercises the ordinary path instead.
+    #[test]
+    fn dropping_env_handle_does_not_hang_or_panic() {
+        let env = EnvHandle::new().expect("failed to create EnvHandle for test");
+        drop(env);
```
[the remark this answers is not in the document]
David-Engel: You're right, the old test was vacuous — an idle runtime joins its idle worker promptly either way.

The test now parks a `spawn_blocking` task (a `mpsc` handshake ensures it's actually running before the drop), then asserts the drop returns well under the park duration. The default `Runtime` drop waits out in-flight blocking work; `shutdown_background` detaches from it, so the two are now distinguishable.

Verified both directions locally by temporarily reverting `Drop` to the default:

- with `shutdown_background`: passes, drop completes in ~0.1s
- with the default drop: fails in 5.05s with `dropping EnvHandle blocked for 5.0009179s waiting on the runtime's blocking work; Drop must detach via shutdown_background, not join`

I kept it as a bounded assertion rather than an unbounded wait so a regression fails cleanly instead of hanging until nextest's slow-timeout kills it. The real AB#47509 case (OS already killed the thread) still needs process/DLL teardown and isn't reproducible from a unit test — noted in the test doc comment.

Fixed in af32643e.

> [!NOTE]

David-Engel:
Follow-up while re-checking the Drop impl: the guard was reachable, so I've reworked it in f61311a9.

The old Drop only shut the runtime down if Arc::try_unwrap succeeded. If a DbcHandle outlived its ENV, that unwrap failed, shutdown_background was silently skipped, and the runtime was dropped later through the DBC's `Arc` using the **default joining path** — the exact teardown AB#47509 panics on. `SQLFreeHandle(ENV)` only `debug_assert!`s that the connection list is empty, so in a release build a non-conformant caller reaches this. That's the same class of caller that motivated this fix: the ones deferring frees to `DLL_PROCESS_DETACH`.

Confirmed with a scratch test before fixing — releasing the straggler reference blocked for 5.0006575s.

The fix moves the guarantee off `EnvHandle` and onto the shared value: a `SharedRuntime` newtype whose own `Drop` calls `shutdown_background`, so whichever owner releases the last `Arc` detaches. That also drops the `Option` and `EnvHandle`'s `Drop` entirely, which removes the `None` case discussed above rather than handling it.

Added a second regression test for the straggler path. Verified both tests discriminate by temporarily reverting `SharedRuntime::drop` to the joining path — both fail at ~5.0s, both pass with `shutdown_background`.

Also merged the latest `main` (#455). `cargo bfmt`/`bclippy` clean, 1168/1168 mssql-odbc unit tests pass.

Theekshna (commented):
Unattended review: this change safely moves ENV-runtime teardown to a shared `SharedRuntime` wrapper so the last owner detaches worker shutdown instead of joining threads during abnormal process teardown. The code and regression coverage look sound; I found one non-blocking PR-description issue.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 0 |

### [Suggestion] Refresh the description and testing details

The **Description** and **Testing** sections still describe the superseded `Option<Runtime>`/`EnvHandle::Drop` design and the old `dropping_env_handle_does_not_hang_or_panic` test. The current diff uses `SharedRuntime(ManuallyDrop<Runtime>)` with teardown on the shared value, and the regression tests are `dropping_env_handle_does_not_wait_for_blocking_work` and `dropping_a_dbc_that_outlived_its_env_does_not_wait_for_blocking_work`. Please update the body so reviewers are evaluating the implementation that is actually proposed.

Covered correctness, `unsafe`/drop invariants, last-owner behavior, concurrency, targeted regression-test sufficiency, CI/checklist consistency, and PR-description currency. msodbcsql parity is N/A because this is an internal Tokio teardown mechanism with no ODBC-visible behavior.

David-Engel:
> ### [Suggestion] Refresh the description and testing details

Done — thanks, that was a fair catch. The body now describes `SharedRuntime(ManuallyDrop<Runtime>)` with teardown on the shared value, names the two current regression tests, and explains why the teardown moved off `EnvHandle::Drop`: the earlier `Arc::try_unwrap` approach silently skipped `shutdown_background()` whenever a `DbcHandle` outlived the ENV, which `free_env` only guards with a `debug_assert!`. That path is reachable in release, and the second test (`dropping_a_dbc_that_outlived_its_env_does_not_wait_for_blocking_work`) is the guard for it.

shiwanigupta0809 (commented):
_Reposted as a top-level comment for visibility: https://github.com/microsoft/mssql-rs/pull/459#issuecomment-5507091563_

David-Engel:
> **This also fixes #445, which is still open — and it disproves that issue's recommended approach, so the reasoning is worth capturing before it is lost.**

Agreed on all of it, and thanks — the `try_unwrap`-is-not-a-guarantee argument is exactly the thing that would have cost the next person a day. Verified your reading before acting: #445 is still open, its Suggested direction is indeed `Arc::try_unwrap` + `shutdown_timeout` justified by the `debug_assert!`, and the `spawn_blocking` sites are `ssrp.rs:285`, `network_transport.rs:240`, `parallel_connect.rs:138` and `named_pipes.rs:126`.

Done:

- Added `Fixes #445` to **Related Issues**, so it closes on merge instead of aging against fixed behaviour.
- Added a **Why this also closes #445** section explaining that the whole behavioural delta of `shutdown_background()` is "stop waiting on in-flight `spawn_blocking` work", which makes #445 and AB#47509 the same code path with two symptoms — plus a note that its recommended direction was tried and rejected, and which test guards that hole.
- [Commented on #445 itself](https://github.com/microsoft/mssql-rs/issues/445#issuecomment-5506886847) with the same reasoning, so it is on the record there even before this merges.

On your last point — that detaching is a trade, not a free win — you are right that nothing recorded it, so I have added a **Trade-off** section stating it plainly: a wedged `getaddrinfo` thread now outlives ENV teardown rather than blocking it. One thing worth adding to your framing: `shutdown_timeout(small)` is not just the less-preferred option here, it is unusable, because any non-zero wait still attempts the join and that is precisely what panics under AB#47509 when Windows has already destroyed the thread. So the two issues do not merely share a fix, they constrain it to the same one.

shiwanigupta0809:
Read the diff plus `free_env`, `DbcHandle`, and the `spawn_blocking` call sites this changes the teardown semantics for. No blocking findings — the `SharedRuntime` newtype is the right shape, and putting the policy on the shared value rather than on `EnvHandle::drop` is the part that makes it correct.

### Suggestion

**This also fixes #445, which is still open — and it disproves that issue's recommended approach, so the reasoning is worth capturing before it is lost.**

The `Related Issues` section links AB#47509 (the teardown panic) but not #445, *"`SQLFreeHandle(ENV)` can still block indefinitely on a leftover `spawn_blocking` DNS lookup"*, filed off PR #432's review.

The two are the same code path. `Runtime`'s default drop joins **blocking** tasks specifically — spawned async tasks are dropped on shutdown either way — so the entire behavioural delta of `shutdown_background()` is "stop waiting on in-flight `spawn_blocking` work." In this tree that work is `tokio::net::lookup_host`'s `getaddrinfo` (`ssrp.rs`, and every TCP connect path after #432) and `named_pipes.rs:126`. #445 is the DNS instance of exactly the wait this removes; AB#47509 is the panic instance. One change, both symptoms.

The new tests already guard it directly rather than by proxy: `park_blocking_task` parks a `spawn_blocking` task and asserts the drop returns well under it, which is #445's scenario with the hung resolver replaced by a `sleep`. That is a stronger regression guard for #445 than anything that issue proposed.

Worth linking for the usual reason — it closes on merge instead of aging as an open bug against already-fixed behaviour — but more importantly because **#445's "Suggested direction" is the approach this PR found to be wrong**:

> This likely needs `Arc::try_unwrap` (or equivalent) to get an owned `Runtime` before calling `shutdown_timeout` … The existing `debug_assert!` in `free_env` already expects all child DBCs to be freed first, so the `Arc` refcount should be 1 by then in the normal (non-debug-assert-violating) path.

Your description explains precisely why that fails: `free_env` only `debug_assert!`s, so in a release build a non-conformant caller makes `try_unwrap` fail, silently skips the shutdown, and later drops through the default joining path — with a measured 5.0006575s block. `dropping_a_dbc_that_outlived_its_env_does_not_wait_for_blocking_work` is the regression test for that exact hole. Left unlinked, the next person to pick up #445 starts from a recommendation that is now known to be broken, and the `debug_assert!`-is-not-a-guarantee argument has to be rediscovered.

A `Fixes #445` line plus a one-sentence note there that `Arc::try_unwrap` + `shutdown_timeout` was tried and rejected would settle both.

Separately worth stating somewhere, since it is the trade this makes and nothing records it: detaching means a genuinely wedged `getaddrinfo` thread now outlives ENV teardown rather than blocking it. That is the right call for a driver — an unbounded block in `SQLFreeHandle` is worse than a leaked thread in a process that is usually exiting anyway — but it is a deliberate choice, not a free win, and `shutdown_timeout(small)` was the alternative #445 had in mind.

mssql-odbc/src/handles/env.rs
```diff
@@ -35,6 +36,51 @@ impl TryFrom<u32> for OdbcVersion {
...
+}
+
+impl Drop for SharedRuntime {
+    fn drop(&mut self) {
+        // SAFETY: `drop` runs at most once and nothing reads `self.0` after
+        // this, so taking ownership out of the `ManuallyDrop` is sound. Taking
+        // it by value is what lets us call `shutdown_background`, which
+        // consumes the `Runtime`; the default drop glue would join instead.
+        let runtime = unsafe { ManuallyDrop::take(&mut self.0) };
+        runtime.shutdown_background();
```
Vahid-b, line 80: [Suggestion] The description's enumeration of the work this stops waiting for misses this crate's own longest blocking task.

The body lists the detached work as `getaddrinfo` plus the named-pipe connect, all in `mssql-tds` — accurate as far as it goes; I checked each site. But `spawn_acquisition` is a direct `tokio::task::spawn_blocking` in production code in *this* crate:

```rust
// mssql-odbc/src/auth/interactive.rs:144
match tokio::task::spawn_blocking(step).await {
```

It runs `msqa::acquire_token`, which pumps a Win32 message loop for the full length of a human sign-in — `interactive.rs:80-82` says so directly: *"The acquisition pumps a Win32 message loop, so it cannot run on a Tokio worker."* It runs on the ENV runtime's blocking pool, and it is `#[cfg(windows)]` (`auth/mod.rs:14-16`), i.e. the same platform as AB#47509.

This does not change the verdict — detaching is still the right call, and for it to bite a caller would already have to be freeing the ENV while a sign-in is live on another thread. But the **Trade-off** section frames the cost as "a genuinely wedged `getaddrinfo` thread", and the material case is an orphaned interactive sign-in window that now outlives `SQLFreeHandle(ENV)` instead of blocking it. One sentence naming `interactive.rs:144` alongside the `lookup_host` sites would make the blast radius complete.

mssql-odbc/src/handles/env.rs
```diff
@@ -100,3 +148,69 @@ impl HasObjectType for EnvHandle {
...
+
+        assert_released_promptly("dropping EnvHandle", || drop(env));
+    }
+
+    /// The ENV is not always the last owner: `SQLFreeHandle(ENV)` only
+    /// `debug_assert!`s that the connection list is empty, so a non-conformant
+    /// caller can free the ENV while a DBC still holds a runtime reference.
+    /// Shutdown must still detach when that straggler releases the last one.
+    #[test]
+    fn dropping_a_dbc_that_outlived_its_env_does_not_wait_for_blocking_work() {
```
Vahid-b, line 205: Not a finding — independent confirmation, since this test's value rests on a claim about a design that isn't in the diff.

I rebuilt the rejected shape on this head (field back to `Option<Arc<SharedRuntime>>`, `SharedRuntime::drop` reverted to the default joining drop, and an `impl Drop for EnvHandle` doing `self.runtime.take()` → `Arc::try_unwrap` → `shutdown_background`) and ran both tests:

```
PASS [   0.014s] dropping_env_handle_does_not_wait_for_blocking_work
FAIL [   5.016s] dropping_a_dbc_that_outlived_its_env_does_not_wait_for_blocking_work
  dropping the last DBC runtime reference blocked for 5.00129066s waiting on the
  runtime's blocking work; shutdown must detach via shutdown_background, not join
```

So this test is not redundant with its sibling: it is the only one that fails under the `try_unwrap` design, exactly as the description claims. (Worth noting for anyone repeating this: a mutation that *clones* the `Arc` inside `EnvHandle::drop` makes `try_unwrap` fail unconditionally and both tests fail, which looks like the same result for the wrong reason. The `take()` form above is the faithful one.)

Vahid-b (commented):
## Summary

Reviewed at head `1c85db7`. The mechanism, the rejection of the `Arc::try_unwrap` alternative, and both regression tests all hold up under direct verification. **No blocking findings** — two suggestions, both about scope and prose rather than code.

Verified against tokio 1.53.1 source (`runtime/runtime.rs`, `runtime/blocking/pool.rs`, `runtime/blocking/shutdown.rs`), Rust std `library/std/src/thread/mod.rs`, `free_handle.rs::free_env`, every `spawn_blocking`/`lookup_host` site in `mssql-tds` and `mssql-odbc`, and local mutation runs.

### Coverage ledger

| Row | Status |
|---|---|
| Primary logic | Examined — `SharedRuntime`, its `Drop`, `Deref`, `EnvHandle::new` |
| Siblings | Examined — every `Runtime` construction in the crate. The other three (`entra.rs:929`, `interactive.rs:225/242`, `msqa.rs:807`) are all inside `#[cfg(test)] mod tests`, so the ENV runtime is the only production runtime. Both `&Arc<Runtime>` signatures in `fetch_scroll.rs` were loosened; no others exist |
| Callers & implementers | Examined — all `dbc.runtime.block_on(...)` sites resolve through the double `Deref`; `alloc_handle.rs` and `dbc.rs` updated |
| Diagnostics | Examined — no new error paths, and the PR *removes* an `.expect()` from a production FFI path per `.github/instructions/mssql-odbc.instructions.md:160` |
| Tests | Examined and mutated — see below |
| Build / packaging / pipelines | N/A — no build, `Cargo.toml`, script or `.pipeline/**` changes |
| Docs & comments | Examined — one stale doc found, suggestion 2 below |
| Description & work items | Examined — `Fixes #445` and AB#47509 present; all four changed files map to the body. One enumeration gap, suggestion 1 |
| CI evidence | Examined — 19/19 green including the cross-repo `mssql-python` suite and all four platform legs. `mergeable_state: blocked` is a missing required approval, not a failing check. Coverage bot: 100% diff, 93.3% overall |

### Verification (measured)

**Mechanism confirmed, not inferred.** The panic is `std::thread::JoinInner::join` at `library/std/src/thread/mod.rs:1812` — `Arc::get_mut(&mut self.packet).expect("threads should not terminate unexpectedly")` — which fires when a thread was force-killed and never dropped its side of the packet `Arc`. Tokio reaches it at `blocking/pool.rs:276 handle.join()`, where `let _ =` swallows the `Err` but not the panic. And `shutdown_background()` genuinely bypasses it: it is `shutdown_timeout(0ns)`, and `blocking/shutdown.rs:40` short-circuits with `if timeout == Some(Duration::from_nanos(0)) { return false; }` before `try_enter_blocking_region`, so there is no join and no "Cannot drop a runtime…" panic either. `BlockingPool::drop`'s later `shutdown(None)` early-returns on `shared.shutdown` (`pool.rs:250`), so there is no second wait.

**Both tests discriminate.** Mutation runs on this head, Linux, `cargo nextest run -p mssqlodbc --lib`:

| Run | `dropping_env_handle_…` | `dropping_a_dbc_that_outlived_its_env_…` |
|---|---|---|
| Baseline | PASS 0.014s | PASS 0.019s |
| Mutation A — `SharedRuntime::drop` reverted to the default joining drop | **FAIL 5.019s** | **FAIL 5.019s** |
| Mutation B — faithful rebuild of the rejected design (`Option<Arc<…>>` + `EnvHandle::drop` + `Arc::try_unwrap` + `shutdown_background`) | PASS 0.014s | **FAIL 5.016s** |

Mutation B is the one that matters, and it independently confirms the claim in your description: the `try_unwrap` hole is real and only the second test catches it. Details inline on that test.

**Blocking-work enumeration checked.** The `getaddrinfo` list in the description is accurate — `network_transport.rs:240`, `parallel_connect.rs:138` and `ssrp.rs:285` are `tokio::net::lookup_host` call sites, and tokio routes that through `spawn_blocking` at `net/addr.rs:179`. `named_pipes.rs:126` is the only direct `spawn_blocking` in `mssql-tds`. See the inline comment for the one site it misses.

## Blocking

None.

## Suggestion

One inline (the `spawn_blocking` enumeration). The second has no line in this diff to anchor to:

**`mssql-odbc/docs/typed-columnar-fetch-plan.md` (~line 274) records this exact panic as an open triage item.** That doc has an "Unrelated observation" section quoting the failure verbatim:

```
thread '<unnamed>' panicked at library\std\src\thread\lifecycle.rs:247:14:
threads should not terminate unexpectedly
```

…followed by *"it does not appear when the same driver is driven through the Driver Manager, so it looks specific to teardown under direct `LoadLibraryW` loading. Out of scope for P4; needs its own triage."* That is precisely AB#47509 and precisely what this PR fixes. Left as-is it sends the next person to triage a solved problem — a one-line edit pointing at #459 / #445 / AB#47509 settles it.

## Nit

None.

## What I could not check

- **The literal AB#47509 failure mode.** No Windows host here, so I could not reproduce `DLL_PROCESS_DETACH` with the worker thread already destroyed. The mutation evidence covers the observable proxy (drop never joins), which is what the tests claim; I agree with the doc comment that the real case is not unit-testable.
- **AB#47509's own text** — ADO is not reachable non-interactively from this session, so I took the reference at face value.
- **Platform scope of the mutations** — Linux only. Cross-platform coverage comes from the CI matrix, which is green on all four legs.
- I ran only the two new tests plus mutations, not the full `mssqlodbc` suite; CI already reports that.

mssql-odbc/src/handles/env.rs
David-Engel: Good catch, and you're right that it was missing — I'd only enumerated the `mssql-tds` sites and overlooked a `spawn_blocking` in this crate's own production code. Verified: `spawn_acquisition` at `interactive.rs:144` runs `msqa::acquire_token` on the ENV runtime's blocking pool, the doc comment at lines 80-90 confirms it pumps a Win32 message loop for the full length of a human sign-in, and `auth/mod.rs:13-16` gates it to `#[cfg(windows)]` — the same platform as AB#47509. It is comfortably the longest blocking task here.

Updated the description: the site list under **Why this also closes #445** now names it explicitly as the longest of the three, and **Trade-off** leads with it rather than with `getaddrinfo`.

One refinement to the framing, since it changes the blast radius you're describing: the surviving artifact is not usually an orphaned sign-in window. `CancelUiOnDrop` is armed for the whole acquisition and only disarmed on success (`interactive.rs:116-131`), and shutdown drops the awaiting future, so `cancel_ui` fires and closes the prompt. What actually outlives `SQLFreeHandle(ENV)` is the blocking-pool thread unwinding afterwards. Still worth naming — a detached thread mid-`acquire_token` is a more interesting residue than a stuck resolver — so I've written it up that way rather than as a desktop-visible leak.

> [!NOTE]

mssql-odbc/src/handles/env.rs
David-Engel: Appreciate you rebuilding the rejected design to check it rather than taking the description's word for it — that's the right instinct for a claim about code that isn't in the diff, and your numbers match what I measured (5.00129066s vs my 5.0006575s, both just over the 5s park).

Your note about the faithful mutation is the part worth keeping. Cloning the `Arc` inside `EnvHandle::drop` makes `try_unwrap` fail unconditionally, so both tests go red and the test looks non-specific; the `take()` form is what the original actually did, and under it only the DBC-outlives-ENV test fails. That distinction is exactly what makes the second test load-bearing rather than a near-duplicate of its sibling.

Resolving as confirmed, no change needed.

> [!NOTE]

Theekshna (commented):
Unattended re-review (thread-resolution trigger, no new commits since my last pass at this sha). Re-checked the full diff, the prior review threads (mine, Vahid-b's, shiwanigupta0809's), and the current PR description.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

No new findings. Everything raised in the prior round is now addressed and resolved:
- Description/testing sections updated to match the `SharedRuntime(ManuallyDrop<Runtime>)` design and the two current regression test names (my earlier suggestion).
- `spawn_acquisition` (`auth/interactive.rs:144`) added to the blocking-work enumeration and Trade-off framing (Vahid-b's suggestion).
- `Fixes #445` linked with rationale for why `Arc::try_unwrap` was rejected (shiwanigupta0809's suggestion).

**Extra checks:**
1. **msodbcsql parity** — N/A. This is an internal Tokio runtime-teardown mechanism with no ODBC-visible behavior; `SharedRuntime`, `EnvHandle`, and the DBC/ENV lifetime relationship are crate-private.
2. **Test sufficiency** — already established by mutation testing in the prior round (Vahid-b: `SharedRuntime::drop` reverted to joining → both tests fail at ~5s; faithful rebuild of the rejected `Arc::try_unwrap` design → only the DBC-outlives-ENV test fails, confirming it isn't redundant with its sibling). I could not repeat this locally this pass: `cargo nextest run -p mssqlodbc --lib` fails to build in this environment on an unrelated toolchain issue (`crc32fast` AVX-512 codegen, `E9012`/`E9013` on `pclmulqdq`/`simd_xor`), pre-existing and unrelated to this diff. Relying on CI's 19/19 green (including the `mssqlodbc --lib` leg) plus the mutation evidence already on record.
3. **Divergences documented** — N/A, no msodbcsql/SqlClient/mssql-jdbc equivalent; this is a Rust-internal Tokio teardown detail.
4. **PR description currency** — matches the current diff; `Fixes #445` and AB#47509 linked; checklist claims (`cargo bfmt`/`bclippy` pass, coverage 100% diff/93.3% overall) match `gh pr checks` (19/19 pass) and the coverage bot comment.
5. **Verbose slop** — none found. The `SharedRuntime` doc comment and test doc comments are long but explain *why* (the `Arc::try_unwrap` hole, the AB#47509 mechanism, why the test can't reproduce the literal OS-kill case) rather than restating the code.

saurabh500 (approved):
## Summary

Approving. Reviewed at head `1c85db7`. Moving teardown onto `SharedRuntime`'s own `Drop`, so whichever owner releases the last `Arc` detaches via `shutdown_background()` instead of joining the worker thread, is the right shape — and putting the policy on the shared value rather than on `EnvHandle::drop` is the part that makes it hold when a DBC outlives its ENV.

I verified the mechanism against the tokio source and ran a mutation rather than taking the description's word for either.

### Mechanism — verified in tokio 1.53.1, not inferred

- `shutdown_background(self)` is `shutdown_timeout(Duration::from_nanos(0))` (`runtime.rs:452,489`).
- `blocking/shutdown.rs:40` short-circuits with `if timeout == Some(Duration::from_nanos(0)) { return false; }` **before** `try_enter_blocking_region`, so the joins at `pool.rs:263-275` — `last_exited_thread.join()` and each worker `handle.join()` — never run. That join is where std's `threads should not terminate unexpectedly` originates.
- `BlockingPool::drop`'s later `shutdown(None)` early-returns on `shared.shutdown` (`pool.rs:250`), so there is no second wait hiding behind the first.
- The multi-thread arm of `Drop for Runtime` only closes the scheduler; it never joins.

So the description's claim that the entire behavioural delta is "stop waiting on in-flight `spawn_blocking` work" holds exactly.

**One benefit the description doesn't claim:** because that 0ns check short-circuits *ahead of* `try_enter_blocking_region`, the new drop also cannot hit `Cannot drop a runtime in a context where blocking is not allowed` if the last `Arc` is ever released from inside an async context. The default path could. This is strictly safer, not merely differently safe.

### Both tests discriminate — measured on Windows at this head

`cargo nextest run -p mssqlodbc --lib -E 'test(does_not_wait_for_blocking_work)'`:

| Run | `dropping_env_handle_…` | `dropping_a_dbc_that_outlived_its_env_…` |
|---|---|---|
| Baseline | PASS 0.049s | PASS 0.037s |
| `SharedRuntime::drop` reverted to the default joining drop | **FAIL 5.053s** | **FAIL 5.043s** |

Neither test is vacuous, and the failure message names the mechanism rather than the symptom, so a future regression reads correctly on the first look.

### Detaching does not create a lifetime hazard

Worth stating explicitly, because it is the one thing this trade could plausibly have gotten wrong. Detaching means blocking work now outlives ENV teardown where it was previously joined, so if any `spawn_blocking` closure held memory that `SQLFreeHandle` releases, this would convert a wait into a use-after-free.

It doesn't. Every `spawn_blocking` in the crate captures owned values only: `interactive.rs:104` and `:120`, via `spawn_acquisition`, capture `String`s, `Arc`s and an owned lock guard, and the `'static` bound forbids borrows of handle memory. `msqa.rs:448` is a `std::thread::Builder::spawn` joined inside its own blocking task, not on the ENV runtime's pool. The residue of this change is a detached thread, not unsafety.

### Also checked

- `Deref` reaches `block_on` and `spawn_blocking` through `&Runtime`, while the consuming `shutdown_*` methods stay unreachable through it — the newtype's policy can't be bypassed by accident.
- The `ManuallyDrop::take` is the standard pattern and its SAFETY note is accurate: `drop` runs at most once and nothing reads `self.0` afterwards. It also stays sound while unwinding, since `wait` returns early rather than reaching the `std::thread::panicking()` branch.
- Removing the `Option` takes an `.expect()` off a production FFI path, which `.github/instructions/mssql-odbc.instructions.md` requires.
- No public or FFI surface change; `SharedRuntime` is `pub(crate)`.

CI is 19/19 green including the cross-repo `mssql-python` suite and all four platform legs, and the coverage bot reports 100% diff coverage.

### One outstanding nit — not merge-blocking

`mssql-odbc/docs/typed-columnar-fetch-plan.md:258-267` still carries the *Unrelated observation* section recording this exact panic (`threads should not terminate unexpectedly`, appearing only under direct `LoadLibraryW` loading, "needs its own triage"). `git log $BASE..HEAD` on that file is empty, so this PR leaves it untouched. A one-line edit pointing at #459 / #445 / AB#47509 saves the next person triaging something already fixed.

### What I could not check

The literal AB#47509 teardown, which needs `DLL_PROCESS_DETACH` with the worker thread already destroyed. I agree with the test doc comment that it isn't reproducible from a unit test; the mutation above covers the observable proxy, which is that shutdown never joins.

PR #436, opened 2026-08-31
Integrate descriptor records with binding and implementation metadata (AB#47437)

Vahid-b:
## Summary

Implements [AB#47437](https://dev.azure.com/SqlClientDrivers/mssql-rs/_workitems/edit/47437) ("Integrate descriptor records with binding and implementation metadata"), the last remaining child task of [AB#46374](https://dev.azure.com/SqlClientDrivers/mssql-rs/_workitems/edit/46374) ("mssql-odbc | Descriptors").

Makes the ARD/APD/IRD/IPD descriptor records the single source of truth for column and parameter binding, execution, and metadata, so `SQLBindCol`/`SQLBindParameter` and an equivalent `SQLSetDescFieldW`/`SQLSetDescRec` sequence can never produce contradictory state.

## What changed

- **ARD**: `SQLBindCol` and `SQLFetchScroll` now share the ARD's own descriptor records as their single storage, replacing the old `StmtState::bindings` shadow `Vec`.
- **APD/IPD**: `SQLBindParameter` and execute now share the APD/IPD's own records; `StmtState::bound_params` becomes a one-time snapshot taken immediately before each execute, rather than a live table `SQLBindParameter` wrote into directly.
- **IRD**: populated from `column_metadata` after every execute and `SQLMoreResults` advance (new `api/ird.rs`), reusing the exact field-mapping functions `SQLDescribeColW`/`SQLColAttributeW` already use.
- **IPD refinement**: `SQLDescribeParam`'s server-described metadata fills in whatever `SQLBindParameter` hasn't already set (tracked via `DescRecord::explicitly_bound`, not the concise type — see below) — never overrides an explicit bind, matching `SQLDescribeParam`'s ODBC contract, and can still refresh what it previously auto-filled itself (e.g. across a re-`SQLPrepare`).
- **New API surface**: `SQLGetDescRecW`/`SQLSetDescRec` (no `W`/`A` split — it has no character-data arguments), implemented by reusing `set_desc_field.rs`'s own field setters so the bulk and single-field APIs cannot diverge. Registered with `SQLGetFunctions` (the two ODBC function codes didn't exist in this driver yet).
- **Reassociation tests**: binding through a reassociated explicit ARD/APD, falling back to the implicit descriptor after the explicit one is freed, `SQLFetchScroll` reading through a reassociated ARD, and `SQLFreeStmt(SQL_UNBIND)` clearing a reassociated ARD.
[...]

PR #460, opened 2026-09-01
Resolve SQL_C_DEFAULT bound columns at fetch time

David-Engel:
## Summary

- Accept `SQL_C_DEFAULT` in `SQLBindCol` instead of rejecting it with `HY003`.
- Resolve default C types from the current result column metadata and ODBC version when each fetch snapshots its bindings, via `resolve_default_bindings`.
- Keep the persistent binding unresolved so it can be reused correctly across result sets.
- Refuse to resolve a fixed-width target wider than the application's declared `BufferLength`, so deferred resolution cannot overrun a bound slot.
- Document the parity deviations that now also apply on the fetch path, and cover them with tests.
- Remove the obsolete documented limitation and update the changelog.

[AB#47481](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47481)

## Why this is needed

Since [AB#47437](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47437) (#436) the ARD is the fetch's single source of truth, and `SQLBindCol` and an equivalent `SQLSetDescField` sequence are meant to be indistinguishable. `SQL_C_DEFAULT` was the exception:

| Path before this PR | Result |
| --- | --- |
| `SQLBindCol(1, SQL_C_DEFAULT, buf, n, &ind)` | `SQL_ERROR`, `HY003` |
| `SQLSetDescField(ard, 1, SQL_DESC_CONCISE_TYPE, SQL_C_DEFAULT)` + `SQL_DESC_DATA_PTR` | succeeds, then `HYC00` per row |

The descriptor route gets there because `set_type` takes the `kind.is_application()` branch, `canonical_c_type` leaves `SQL_C_DEFAULT` alone, and `is_valid_c_type` lists it — while `ColumnBinding::from_record` keys "bound" off a non-null `SQL_DESC_DATA_PTR` rather than the concise type. Both doors now resolve identically, which restores the equivalence #436 was built to guarantee.

## Parity deviations

Resolution reuses `type_rules::resolve_default_c_type`, shared with `SQLBindParameter`, so the driver gives one answer to what `SQL_C_DEFAULT` means regardless of direction. That resolver carries two deliberate deviations from msodbcsql's `Sql2CDefault`, accepted in #323 and registered in `parameters_plan.md`, `mssql-odbc.instructions.md`, and [AB#47365](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47365). This PR makes them apply on the fetch path too, so they are restated where the mapping is used, and the registry entry — which previously scoped them to `SQLBindParameter` — is widened to cover both directions:

| SQL type | This driver | msodbcsql |
| --- | --- | --- |
| `SQL_WCHAR`, `SQL_WVARCHAR`, `SQL_WLONGVARCHAR` | `SQL_C_WCHAR` | `SQL_C_CHAR` |
| `SQL_GUID` | `SQL_C_GUID` | `SQL_C_CHAR` |

Both cells are measured against msodbcsql18, not inferred from the `Sql2CDefault` tables. Binding an `nvarchar` and a `uniqueidentifier` column with `SQL_C_DEFAULT` and `BufferLength` 64 returns the wide value as the three narrow bytes `6F 6E 65` with indicator `3`, and the GUID as the 36-character text form with indicator `36`.

`SQL_SS_XML` is deliberately absent from that table: msodbcsql maps it to `SQL_C_WCHAR` as well (an `xml` column bound `SQL_C_DEFAULT` returns UTF-16 with indicator `30`), so there is no deviation, and it is unreachable on this path because `describe_col.rs` reports xml and json columns as `SQL_WLONGVARCHAR`.

The wide deviation exists because this driver's `SQL_C_CHAR` is UTF-8 by design; msodbcsql's narrow default reads the client code page, which has no equivalent here. The `SQL_GUID` deviation is the one that also changes the bound rowset layout: a fixed-width target strides by its C type rather than by `BufferLength`, so the stride becomes `sizeof(SQLGUID)`. A slot at least 16 bytes wide — including the 36-character text form msodbcsql would fill — takes the narrower stride and stays inside the application's array.

### Narrow-buffer guard

`BufferLength` is ignored for a fixed-width target, which is safe when the application *named* that type and so accepted its width contract. A `SQL_C_DEFAULT` binding names nothing, so resolving a `uniqueidentifier` column to `SQL_C_GUID` would write 16 bytes into whatever slot the application declared — 4 bytes, say — where msodbcsql resolves to `SQL_C_CHAR` and truncates inside `BufferLength`. Deferred resolution is what makes that reachable without the application naming the C type.

Such a binding is therefore left unresolved and the row fails with `HYC00` rather than writing. The check is generic over `element_stride` rather than GUID-specific: `SQL_C_TYPE_DATE` / `TIME` / `TIMESTAMP` and, at ODBC 3.8, `SQL_C_SS_TIME2` / `SQL_C_SS_TIMESTAMPOFFSET` all route through the same `fixed_width > 0 && buffer_length > 0 && buffer_length < fixed_width` comparison. GUID is the motivating case and the one unit-tested directly, since the guard's correctness depends on the arithmetic rather than on which C type triggers it.

`BufferLength` 0 is exempt: it is the documented idiom for a fixed-width target and carries no width claim. The width is `SQL_DESC_OCTET_LENGTH` on the ARD record, which an application can rewrite after binding without touching the type — checking it per fetch against the same snapshot the fill loop uses is what keeps a slot narrowed after the bind from slipping through. Whether these should instead report `01004` and deliver a truncated value, matching msodbcsql more closely, is open and untracked.

### Other newly-reachable cases, registered not changed

- Resolution is ODBC-version aware, with the version read from the environment on each fetch: `SQL_SS_TIME2` and `SQL_SS_TIMESTAMPOFFSET` default to `SQL_C_BINARY` below ODBC 3.8 and to their `SQL_C_SS_*` types at 3.8. That 3.8 mapping makes a stride-only divergence reachable: both drivers resolve identically, but msodbcsql's `BindOffset` strides by `BufferLength` where `element_stride` uses 12 / 20. Measured at `BufferLength` 40, msodbcsql puts row 1 at byte offset 40 and this driver at 12, indicator 12 in both. Ours is the safer direction — `BufferLength` 0 would make msodbcsql stride 0 and stack every row in slot 0.
- A `varbinary` / `image` column resolves to `SQL_C_BINARY`, which bound delivery does not implement yet ([AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239)), so it fails per row with `HYC00`. msodbcsql resolves identically and delivers the bytes.

A column whose SQL type has no default keeps `SQL_C_DEFAULT` and is reported as an unsupported target for any row carrying a value; a NULL row still reports `SQL_NULL_DATA` through its indicator, because nothing needs converting. A binding whose ordinal is past the end of the result set also stays unresolved, but never reaches delivery — the fill loop skips it, matching msodbcsql.

Known asymmetry, pre-existing and untouched: `SQLGetData` still refuses `SQL_C_DEFAULT` with `HYC00`, so the same placeholder is answered two ways depending on how a column is read. Recorded in `typed-columnar-fetch-plan.md`; not currently tracked by a work item.

## Testing

- `cargo nextest run -p mssqlodbc`: 1,237 passed, 1 skipped.
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- `cargo fmt -- --check`
- Unit coverage: wide and GUID columns resolving to their typed targets including the resulting `element_stride`; a binding whose column is absent from the result set staying unresolved; `a_default_binding_follows_the_declared_odbc_version`, a fetch-level test driving a `SQL_SS_TIME2` column at ODBC 3.0 and 3.8 that pins both the per-row `SqlReturn` and its SQLSTATE (`HYC00` vs `07006`); two tests for the narrow-buffer guard, one of which uses a 64-byte backing array behind a declared `BufferLength` of 4 so a regression surfaces as bytes written past the declared width rather than as a crash; and `a_default_binding_set_through_the_descriptor_api_also_resolves`, which binds entirely through `SQLSetDescField` to cover the second door into the resolver. Each guard was confirmed to fail with its guard removed.
- E2E `DefaultTargetResolvesWideAndGuidToTypedTargets` pins the wide indicator in UTF-16 bytes and the GUID stride at `sizeof(SQLGUID)`, binding four GUID slots with a `BufferLength` of two so a `BufferLength`-driven stride fails an assertion inside the array rather than running off the end. It asserts a deliberate deviation, so it carries `SKIP_IF_COMPARING_MSODBCSQL()`.
- E2E `DefaultTargetResolvesPerResultSet` drives two result sets of different column types through one `SQL_C_DEFAULT` binding. It deliberately uses `int` and narrow `varchar`, which both drivers resolve identically, so it runs and compares on the msodbcsql leg too.
- Full C++ e2e comparison against mssql-odbc and retail msodbcsql on the Linux and Windows PR legs: all test executables passed on each driver with zero parity divergences.

None. The narrow-buffer guard only affects `SQL_C_DEFAULT` bindings that would previously have overrun the application's declared buffer.


mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -580,7 +593,21 @@ fn fetch_scroll_safe(
...
+                continue;
+            }
+            let Some(meta) = stmt_state
+                .column_metadata
+                .get(usize::from(binding.column_number).saturating_sub(1))
+            else {
+                continue;
+            };
+            if let Some(target_type) = resolve_default_c_type(odbc_sql_type(meta), odbc_version) {
+                binding.target_type = target_type;
```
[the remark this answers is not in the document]
David-Engel: Good catch on the overclaim — "matching msodbcsql" was wrong as a blanket statement. I kept the shared resolver but explicitly approved, documented, and tested the deviations instead.

I probed msodbcsql18 directly rather than relying on the `Sql2CDefault` source comment, binding an `nvarchar` and a `uniqueidentifier` column with `SQL_C_DEFAULT` and `BufferLength` 64. It resolves both to `SQL_C_CHAR`, exactly as you said: the wide column comes back as the three narrow bytes `6F 6E 65` with indicator `3` (not 6 bytes of UTF-16), and the GUID as the 36-character text form with indicator `36`, striding by the full 64-byte `BufferLength`.

Both deviations are kept deliberately, and for different reasons:

- **Wide → `SQL_C_WCHAR`.** msodbcsql's narrow default is an artifact of a driver shipped in both an ANSI and a Unicode build. This driver has only the Unicode one and its `SQL_C_CHAR` is UTF-8, so following msodbcsql would transcode every wide column by default. It would also contradict the same resolver's behavior on the parameter path, where it has shipped since `SQLBindParameter`.
- **`SQL_GUID` → `SQL_C_GUID`.** Follows the ODBC 3.x default-C-type table. You're right that this is the one that changes the rowset layout: the stride becomes `sizeof(SQLGUID)` rather than `BufferLength`. That direction is safe — a slot sized for the 36-character text form is wider than 16 bytes, so every write stays inside the application's array.

Changes in 127d5bcb:

- Extracted the resolution loop into `resolve_default_bindings`, with the deviations and the measured reference behavior documented at the point the mapping is applied.
- Replaced the blanket "matching msodbcsql" claim in the CHANGELOG, and added a `SQL_C_DEFAULT bindings` section to `typed-columnar-fetch-plan.md` with a per-type table and the measured msodbcsql values.
- Unit tests: wide and GUID columns resolving to the typed targets, including the resulting `element_stride`, plus a binding with no matching result column staying at `SQL_C_DEFAULT`.
- E2E test `DefaultTargetResolvesWideAndGuidToTypedTargets` pins the wide indicator in UTF-16 bytes and the GUID stride at `sizeof(SQLGUID)`. It binds four GUID slots with `BufferLength` of two, so a `BufferLength`-driven stride would land row 1 in `guids[2]` and fail an assertion rather than run off the end of the array.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -488,12 +491,71 @@ impl RowWriter for BoundRowWriter<'_> {
...
+/// that outlives its result set keeps the placeholder and is resolved again
+/// against the next set's metadata.
+///
+/// The mapping is [`resolve_default_c_type`], shared with `SQLBindParameter` so
+/// the driver gives one answer to "what does `SQL_C_DEFAULT` mean". Its two
+/// documented deviations from msodbcsql's `Sql2CDefault` therefore apply on the
+/// fetch path too. Both were confirmed by probing msodbcsql18, which resolves
+/// each to `SQL_C_CHAR`, and both are kept:
+///
+/// - The wide character types (and `SQL_SS_XML`) resolve to `SQL_C_WCHAR`. The
```
saurabh500: I don't think `SQL_SS_XML` belongs in this deviation. Reading the msodbcsql caller: `BindOffset` resolves `SQL_C_DEFAULT` through `Sql2CDefault` (`Sql/Ntdbms/sqlncli/odbc/sqlcfunc.cpp:2171-2173`), `Sql2CDefault` picks one of the two default tables (`sqlcprot.h:1823-1830`), and both tables map `SQL_XML_MAPPED` to `SQL_C_WCHAR` (`sqlcmisc.cpp:177-180`, `sqlcmisc.cpp:216-218`). `SQL_XML_MAPPED` is the internal spelling of `SQL_SS_XML` (`sqlsrv.h:189-193`). So XML already agrees with msodbcsql here — it's not a divergence. Suggest dropping `SQL_SS_XML` from this deviation note rather than recording an aligned case as an approved divergence.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -488,12 +491,71 @@ impl RowWriter for BoundRowWriter<'_> {
...
+///   column by default.
+/// - `SQL_GUID` resolves to `SQL_C_GUID`, following the ODBC 3.x
+///   default-C-type table. This is the one deviation that also changes the
+///   rowset layout, because a fixed-width target takes its stride from the C
+///   type rather than from `BufferLength` ([`element_stride`]): 16 bytes per
+///   row where msodbcsql delivers the 36-character text form and strides by
+///   the caller's slot size. The narrower stride stays inside the
+///   application's array either way.
+///
+/// A column with no metadata, or a SQL type with no default, is left at
```
saurabh500: Worth splitting these two cases — they don't behave the same. The `no metadata` case is a column ordinal past the current result set, and the fill loop skips that binding before delivery (`if column == 0 || column > column_count { continue; }`), so `deliver_bound` never reports it per row. Only the `SQL type with no default` case actually stays `SQL_C_DEFAULT` and surfaces `HYC00`. Suggest rewording this comment (and the matching PR/doc text) so it doesn't promise a per-row diagnostic for the path the fetch loop intentionally ignores.

Theekshna (commented):
This unattended review found no issues in the proposed change.
Covered deferred `SQL_C_DEFAULT` resolution, per-result-set rebinding, ODBC-version mappings, buffer strides, lock ordering, unit/e2e test sufficiency, msodbcsql parity and documented divergences, CI/checklist consistency, and PR-description currency.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

mssql-odbc/docs/typed-columnar-fetch-plan.md
```diff
@@ -152,13 +152,30 @@ Each of these is an error rather than a guess, because a wrong value delivered s
...
 
+#### `SQL_C_DEFAULT` bindings
+
+`SQLBindCol` accepts `SQL_C_DEFAULT` and stores it unresolved, the way msodbcsql does (`sqlcfunc.cpp` `BindOffset` → `Sql2CDefault`): the answer depends on the IRD, which does not exist until a statement executes, and one binding can outlive several result sets. Each `SQLFetchScroll` resolves the placeholder on its own snapshot of the binding table, so the stored binding still says `SQL_C_DEFAULT` for the next result set.
+
+The mapping is `type_rules::resolve_default_c_type`, shared with `SQLBindParameter` so the driver gives one answer for what `SQL_C_DEFAULT` means. It carries two deliberate deviations from msodbcsql, and they apply here too:
+
+| SQL type | This driver | msodbcsql | Why the deviation is kept |
+| --- | --- | --- | --- |
+| `SQL_WCHAR`, `SQL_WVARCHAR`, `SQL_WLONGVARCHAR`, `SQL_SS_XML` | `SQL_C_WCHAR` | `SQL_C_CHAR` | The narrow default is an artifact of a driver shipped in both ANSI and Unicode builds. This driver has only the Unicode one and its `SQL_C_CHAR` is UTF-8, so following msodbcsql would transcode every wide column by default |
```
Theekshna: `SQL_SS_XML` doesn't belong in this row — msodbcsql maps it to `SQL_C_WCHAR`, same as this driver, so there's no deviation to register.

`sqlcmisc.cpp:179` and `:218` both read `SQL_C_WCHAR, // SQL_XML_MAPPED`, in `rgbTRANSTYPE` and `rgbTRANSTYPE380` alike. Index alignment checked rather than assumed: `SQL_TYPE_MIN` is `SQL_GUID` (`sqlsrv.h:198`) and the `*_MAPPED` values are `SQL_VARCHAR+1..+7` (`sqlsrv.h:190-196`), so the row labels line up exactly.

The paragraph below says the msodbcsql column is measured, but the probe covered `nvarchar` and `uniqueidentifier` — this cell was generalised, not measured. It's also unreachable on this path: `describe_col.rs:212` reports xml/json columns as `SQL_WLONGVARCHAR`, never `SQL_SS_XML`.

mssql-odbc/src/api/fetch_scroll.rs
Theekshna: Same as the table row in `typed-columnar-fetch-plan.md`: `SQL_SS_XML` isn't a deviation. `rgbTRANSTYPE380` maps `SQL_XML_MAPPED` to `SQL_C_WCHAR` (`sqlcmisc.cpp:218`), which is what this driver does too — so "Both were confirmed by probing msodbcsql18, which resolves each to `SQL_C_CHAR`" overstates what was measured. Suggest dropping `(and `SQL_SS_XML`)` and leaving the bullet about the wide character types.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -580,7 +642,8 @@ fn fetch_scroll_safe(
             return rc;
         }
 
-        let bindings: Vec<ColumnBinding> = stmt_state.bindings.clone();
+        let mut bindings: Vec<ColumnBinding> = stmt_state.bindings.clone();
+        resolve_default_bindings(&mut bindings, &stmt_state.column_metadata, odbc_version);
```
Theekshna: The version argument resolves to nothing observable. I replaced it with a hardcoded `OdbcVersion::Odbc2` and ran `cargo nextest run -p mssqlodbc --lib`: 1,170 passed, 1 skipped. So the env read at 550-557 and this argument are both unguarded, and a future change back to a constant would be silent — applications declaring `SQL_OV_ODBC3` would get `SQL_C_SS_TIME2` for a `time` column where they should get `SQL_C_BINARY`.

Worth a fetch-level test rather than a direct call to `resolve_default_bindings`, since the wiring is the uncovered part. Note the observable at 3.0 is a per-row `SQL_ROW_ERROR` (bound `SQL_C_BINARY` delivery is unimplemented, AB#47239), not a delivered value.

mssql-odbc/src/api/fetch_scroll.rs
Theekshna: Only half of this holds. A binding whose ordinal is past the end of the result set never reaches `deliver_bound` at all — the fill loop skips it at 859-867 and reports nothing, deliberately matching msodbcsql, as the comment there says. Only "a SQL type with no default" (an `SQL_UNKNOWN_TYPE` column) actually lands on the unsupported-target path.

Same sentence in `typed-columnar-fetch-plan.md:177`, the PR description, and the doc comment at 1942-1944 on `a_default_binding_without_metadata_stays_unresolved`.

Theekshna (commented):
## Summary

The design matches the reference: msodbcsql also keeps the descriptor unresolved and calls `Sql2CDefault` at data-fetch time (`sqlcdata.h:254-257`), and its `BindOffset` stride rule (`sqlcfunc.cpp:2176-2284`) matches `element_stride` — fixed types stride by `sizeof(C type)`, char/binary by `cbValueMax`, with `SQL_C_DEFAULT` resolved *before* the stride is computed. Lock ordering follows the existing `bind_param.rs:129-138` precedent. Tests are non-vacuous: removing the `resolve_default_bindings` call fails `default_binding_resolves_from_current_result_metadata`, and the e2e GUID test is well built — four slots for a rowset of two makes a `BufferLength`-driven stride fail an assertion inside the array rather than smash the stack.

No blocking findings. Two suggestions and three nits, the latter all on files outside the diff.

### Nit — the registry entry still scopes the deviation to `SQLBindParameter`

`.github/instructions/mssql-odbc.instructions.md:79` reads "`SQL_C_DEFAULT` in SQLBindParameter resolves the wide character SQL types to `SQL_C_WCHAR`, and `SQL_GUID` to `SQL_C_GUID`". The PR description says these deviations are registered there, but that file isn't in the diff, so the canonical registry now understates the scope. A few words.

(It also names only "the wide character SQL types" and `SQL_GUID` — no `SQL_SS_XML` — which corroborates the inline comment on the new table row.)

### Nit — `SQLGetData(SQL_C_DEFAULT)` still answers `HYC00`

`get_data.rs:378-393` rejects it, where msodbcsql resolves it in the same `GetColData` that serves both paths. Pre-existing and unchanged here, so a legitimate deferral, but the driver now answers the same placeholder two different ways depending on how the column is read. I couldn't find it tracked anywhere; worth a work item while `odbc_sql_type` + `resolve_default_c_type` are freshly wired up next door.

### Nit — one more unregistered stride divergence, now reachable by default

Under 3.8 both drivers resolve `time`/`datetimeoffset` identically (`rgbTRANSTYPE380`, `sqlcmisc.cpp:220-221`), but msodbcsql's `BindOffset` switch has no case for `SQL_C_SS_TIME2`/`SQL_C_SS_TIMESTAMPOFFSET`, so they fall to `default: dwOffset = lpbindinfo->cbValueMax` (`sqlcfunc.cpp:2280-2283`) while `element_stride` returns 12 and 20. Pre-existing for explicit binds, but this PR makes it reachable without the application naming the type. Ours is the safer direction — msodbcsql with `BufferLength` 0 strides 0 and stacks every row in slot 0 — so I'd register it rather than change behaviour.

---

Not verified: no e2e run (no live server), no reproduction of the msodbcsql18 measurements, no cross-platform build.

mssql-odbc/docs/typed-columnar-fetch-plan.md
David-Engel: You're right, and I've dropped it from the table.

I treated this the way the rest of the table was built and measured it rather than reading the mapping: an `xml` column bound `SQL_C_DEFAULT` against msodbcsql18 comes back as `3C 00 72 00 3E 00 …` with indicator `30` — UTF-16, 15 characters — not narrow bytes. That matches your read of `sqlcmisc.cpp:179`/`:218`, so there is no deviation to register.

Your point about the paragraph below was the more important one: "measured" was doing work it hadn't earned for that cell. The probe covered `nvarchar` and `uniqueidentifier` only, and I generalised the third. That's the specific failure the earlier round of this PR was supposed to have fixed, so thanks for catching that it recurred.

Also confirmed unreachable, as you said — `describe_col.rs:212` reports xml and json as `SQL_WLONGVARCHAR`. I added that to the doc rather than dropping the row silently, so the next person doesn't re-add it from the `Sql2CDefault` tables.

Fixed in 83fa015e, along with the same claim in the `fetch_scroll.rs` doc comment.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Done — dropped `(and `SQL_SS_XML`)`, so the bullet now covers only the wide character types, and "confirmed by probing msodbcsql18" is true of both remaining deviations.

I probed the xml case before changing it: msodbcsql returns UTF-16 with indicator `30` for an `xml` column bound `SQL_C_DEFAULT`, matching `rgbTRANSTYPE380`'s `SQL_C_WCHAR` and matching this driver. So the sentence was overstating exactly as you said.

I also noted in the doc comment why `SQL_SS_XML` is absent rather than just removing the mention — it is both a non-deviation and unreachable via `odbc_sql_type`, and without that note it reads like an oversight and invites being added back.

83fa015e.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Confirmed and fixed. I reproduced your mutation first — hardcoding `OdbcVersion::Odbc2` at the call site left all 1,170 tests passing, so the version wiring was genuinely unguarded.

Added `a_default_binding_follows_the_declared_odbc_version`, a fetch-level test as you suggested rather than a direct `resolve_default_bindings` call, so the env read at 550-557 is covered too. It drives a `SQL_SS_TIME2` column through `sql_fetch_scroll` at 3.0 and at 3.8.

On the observable: both versions fail the row, since the mock transport only carries `int` payloads and bound `SQL_C_BINARY` delivery is unimplemented. But they fail *differently*, which is enough to pin the resolved target — `07006` at 3.8, where the typed `SQL_C_SS_TIME2` cannot take an int, and `HYC00` at 3.0 for the binary target this driver does not deliver. I documented that reasoning on the test, since asserting on diagnostics rather than delivered values is non-obvious.

Verified it actually guards: re-applying the hardcoded version fails the 3.0 assertion (`07006` where `HYC00` is expected). A test that passed under the mutation would have been worthless here.

Full suite is 1,183 passing after merging main. 83fa015e.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Correct on both halves — I verified the skip at 859-867 before changing anything. An out-of-range ordinal hits `continue` and never reaches `deliver_bound`, so "reported per row as an unsupported target" was only ever true of the `SQL_UNKNOWN_TYPE` case.

I split the sentence rather than deleting it, so both behaviours are stated where they were conflated: a SQL type with no default reaches the unsupported-target path, and an out-of-range ordinal stays unresolved but is skipped before delivery, matching msodbcsql.

Fixed in all four places you listed — the doc comment at 520-522, `typed-columnar-fetch-plan.md`, the test doc comment on `a_default_binding_without_metadata_stays_unresolved`, and the PR description.

Worth noting the test itself was fine; only its comment claimed something the code does not do. I left the assertion alone, since "stays unresolved" is the real contract and is what the fill loop then skips on.

83fa015e.

> [!NOTE]

David-Engel:
Thanks — all three nits were real, and I acted on all three in dab4f5c5.

> `.github/instructions/mssql-odbc.instructions.md:79` reads "`SQL_C_DEFAULT` in SQLBindParameter resolves..."

Fixed. It now says the deviation applies in both directions and names `SQLFetchScroll` resolving per fetch from the IRD through the same resolver. Good catch that the canonical registry was the one place still scoped to the old truth — the PR description pointed at it as authoritative, which made the gap worse than a stale comment.

I also added the `SQL_SS_XML` non-deviation there explicitly, since you noted the registry corroborated it. Recording why it is absent should stop it being re-derived from the `Sql2CDefault` tables a third time.

> one more unregistered stride divergence, now reachable by default

Registered, and measured first rather than taken on the source read — registering an unverified cell is exactly what went wrong with `SQL_SS_XML`. A two-row rowset over `time(7)`, bound `SQL_C_DEFAULT` with `BufferLength` 40 against msodbcsql18:

```
ind=[12, 12]
00: 01 00 02 00 03 00 00 00 BC CC 5B 07 EE EE EE EE   <- row 0 at offset 0
...
28: EE EE EE EE EE EE EE EE 0B 00 0C 00 0D 00 00 00   <- row 1 at offset 40
```

Row 1 lands at 40, not 12, and the indicator is 12 in both drivers — so the resolved type agrees and only the stride differs, exactly as you described. Kept our behaviour for the reason you gave: `BufferLength` 0 would make msodbcsql stride 0 and stack every row in slot 0. Recorded in both the registry and the plan doc, including that it is pre-existing for an explicit bind and that deferred resolution is what makes it reachable by default.

> `SQLGetData(SQL_C_DEFAULT)` still answers `HYC00`

Confirmed at `get_data.rs:378-393` — `SQL_C_DEFAULT` is not a typed target, not `SQL_C_CHAR`/`SQL_C_WCHAR`, and not a binary probe, so it falls to the `HYC00`. I have recorded the asymmetry in the plan doc, including that it is untracked.

I have **not** filed the work item. Creating one is outside the actions I am authorised to take on this run, so I have deliberately left it rather than quietly skipping it — it needs a human to file, and the note in the plan doc is there so it is not lost in a review thread. Happy to be pointed at it if you would rather it were tracked before this merges.

All three are documentation-only; no behaviour changed. Unit suite still 1,183 passing, fmt and clippy clean.

> Not verified: no e2e run (no live server), no reproduction of the msodbcsql18 measurements

Worth noting the two measurements above were reproduced independently on my side against a live SQL Server and a retail msodbcsql18, and the `SQL_SS_XML` one contradicted what I had originally written — so that gap in your review was load-bearing, not incidental.

Theekshna (commented):
Unattended re-review: head moved (`418e12bc` to `dab4f5c5`) since my last marked review at that sha (findings=0). In between, a separate review (submitted 09:50 UTC, same sha, no auto-review marker so it didn't drive skip-logic but I treated its findings as mine to check) raised 2 suggestions + 3 nits. All five are now addressed by the two commits since:

- **`0d4ed9c6`**: drops `SQL_SS_XML` from the deviation table/doc comment (msodbcsql maps it to `SQL_C_WCHAR` too, confirmed by measurement, and it's unreachable here since `odbc_sql_type` reports xml/json as `SQL_WLONGVARCHAR`); adds a fetch-level test (`fetch_time_column_row_state`) that drives the ODBC-version-dependent `SQL_SS_TIME2` resolution at 3.0 and 3.8 and distinguishes the result by diagnostic code, closing the observability gap the prior review flagged (hardcoding the version left all tests passing); corrects the "binding past end of result set" claim to match what the fill loop actually does.
- **`dab4f5c5`**: widens the `SQL_C_DEFAULT` deviation registry entry from "in SQLBindParameter" to cover `SQLFetchScroll` too; registers the `time`/`datetimeoffset` stride divergence (measured: msodbcsql strides by `BufferLength`, this driver by `sizeof`) with a stated rationale for keeping rather than matching; registers that `SQLGetData(SQL_C_DEFAULT)` still answers `HYC00` while fetch-bound columns now resolve.

I checked each fix against the actual diff rather than the commit message alone (`typed-columnar-fetch-plan.md`, `mssql-odbc.instructions.md`, the new test in `fetch_scroll.rs`) — all match what was claimed, with the msodbcsql citations (`sqlcmisc.cpp:179/218/220-221`, `sqlcfunc.cpp:2280-2283`) consistent with what I'd expect from prior verification of this file (path+line, not quoted).

**Test sufficiency:** the new fetch-level test is exactly the mutation-informed gap the prior review asked for (a direct `resolve_default_bindings` call wouldn't have caught the wiring bug). No further gap found in this increment.

**msodbcsql parity:** covered above — the corrections are accurate and the newly-registered divergences are honest about which direction is safer and why.

**Divergences documented:** yes, this increment is entirely divergence-documentation work, done at the two locations the repo uses for it (`.github/instructions/mssql-odbc.instructions.md` and the design doc).

**PR description/registry currency:** consistent as of this head.

**Verbose slop:** none in the new doc/comment text — it's measurement-backed, not restating the diff.

CI: no failing checks found via `gh pr checks 460`.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -488,12 +491,78 @@ impl RowWriter for BoundRowWriter<'_> {
...
+///   Unicode build; this driver has only the Unicode one, and its `SQL_C_CHAR`
+///   is UTF-8, so following msodbcsql would transcode every wide column by
+///   default.
+/// - `SQL_GUID` resolves to `SQL_C_GUID`, following the ODBC 3.x
+///   default-C-type table. This is the one deviation that also changes the
+///   rowset layout, because a fixed-width target takes its stride from the C
+///   type rather than from `BufferLength` ([`element_stride`]): 16 bytes per
+///   row where msodbcsql delivers the 36-character text form and strides by
+///   the caller's slot size. The narrower stride stays inside the
+///   application's array either way.
```
Vahid-b: **Suggestion.** This safety claim holds only for `BufferLength >= sizeof(SQLGUID)`. (Distinct from the `SQL_SS_XML` and out-of-range-ordinal threads on the neighbouring sentences — this is about the buffer width, not the mapping.)

Cited, not measured — I have no live server here. `SQL_GUID` resolves to `SQL_C_GUID`, `element_stride` returns a constant `16` at `:1098` ignoring `BufferLength`, `deliver_bound` computes `slot = base + bind_offset + row_index * 16` at `:1484`, and `convert_guid_c` ends in `write_fixed(target_value_ptr, guid, ...)` (`conversion/fetch_convert.rs:306`) — no length parameter anywhere on that path.

msodbcsql resolves the same column to `SQL_C_CHAR` (`sqlcmisc.cpp:153` and `:192`), which lands on `BindOffset`'s `default:` arm — `fFixed = FALSE`, stride `cbValueMax` (`sqlcfunc.cpp:2280-2283`) — so its write is bounded by the slot the application declared and an over-long value truncates with `01004`.

So `SQLBindCol(col, SQL_C_DEFAULT, buf, N, &ind)` on a `uniqueidentifier` column writes past the slot for any `N < 16`, where the reference driver stays inside it. The sharpest case is `N = 0`: `deliver_bound` documents `BufferLength` 0 at `:1462-1465` as a length probe that "leaves the buffer alone", and this is the one resolution that breaks that guarantee.

The general "`BufferLength` is ignored for fixed-width targets" rule does not cover this. It applies when the application *named* the fixed type and therefore knows the width contract; here it named the placeholder, and this driver picked a wider contract than msodbcsql's.

Minimum ask is to scope the claim — it reads unconditionally here, in `typed-columnar-fetch-plan.md:170` ("A slot sized for the 36-character text form is wider than 16 bytes"), and in the PR description ("every write stays inside the application's array"). Worth considering separately whether a `SQL_C_DEFAULT`-resolved fixed-width target with `BufferLength < sizeof(type)` should report `01004` / `SQL_ROW_ERROR` instead of writing; that would want its own work item either way.

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -76,13 +76,36 @@ authoritative parity reference for this crate. Its source lives in the
...
+    switch has no case for them and falls through to
+    `default: dwOffset = lpbindinfo->cbValueMax` (`sqlcfunc.cpp:2280-2283`).
+    Measured: a two-row rowset bound `SQL_C_DEFAULT` with `BufferLength` 40 puts
+    msodbcsql's second row at byte offset 40, where this driver puts it at 12;
+    the indicator is 12 in both, so only the stride differs. This is the safer
+    direction — msodbcsql with `BufferLength` 0 strides 0 and stacks every row in
+    slot 0 — so the behaviour is kept and registered rather than matched.
+    Pre-existing for an explicit `SQL_C_SS_TIME2` bind; deferred
+    `SQL_C_DEFAULT` resolution makes it reachable without the application naming
+    the C type.
```
Vahid-b, line 108: **Suggestion.** A `varbinary` / `image` column bound `SQL_C_DEFAULT` is newly reachable by exactly this bullet's criterion, and is not registered.

Cited. `odbc_sql_type` maps `VarBinary` / `BigVarBinary` to `SQL_VARBINARY` and `Image` to `SQL_LONGVARBINARY` (`describe_col.rs:208-210`); `resolve_default_c_type` maps both to `SQL_C_BINARY`; `deliver_bound` refuses that at `fetch_scroll.rs:1512-1516` with `HYC00` and `SQL_ROW_ERROR`, pending AB#47239. msodbcsql resolves identically (`SQL_C_BINARY` for all three binary rows in both tables) and delivers the bytes.

That is the same shape as the bullet above it — "pre-existing for an explicit bind; deferred `SQL_C_DEFAULT` resolution makes it reachable without the application naming the C type" — and it covers considerably more common column types than `time` / `datetimeoffset`. The PR's own new test demonstrates the path: `a_default_binding_follows_the_declared_odbc_version` expects `HYC00` at ODBC 3.0 precisely because the 3.0 mapping is `SQL_C_BINARY`.

Not a behaviour change; a line here and in the new `SQL_C_DEFAULT bindings` section pointing at AB#47239 would close it.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -488,12 +491,78 @@ impl RowWriter for BoundRowWriter<'_> {
...
+///
+/// `SQL_SS_XML` is deliberately not in that list: msodbcsql maps it to
+/// `SQL_C_WCHAR` too, so there is no deviation to record, and it is unreachable
+/// here anyway because [`odbc_sql_type`] reports xml and json columns as
+/// `SQL_WLONGVARCHAR`.
+///
+/// A column whose SQL type has no default is left at `SQL_C_DEFAULT`, which
+/// [`deliver_bound`] reports as an unsupported target per row rather than
```
Vahid-b: **Nit.** This first sentence overstates the NULL case. (The *second* sentence, about an ordinal past the end of the result set, is the one corrected in 83fa015e — I'm not re-raising that.)

`deliver_bound` short-circuits `ColumnValues::Null` at `:1486`, before any target dispatch, so an unresolvable column returns `RowOutcome::Success` with `SQL_NULL_DATA` for a NULL row; only a non-NULL value reaches `RowIssue::Unsupported`. "reported as an unsupported target per row" is therefore true of a row carrying a value, not of every row.

Same wording in `typed-columnar-fetch-plan.md:179` and the PR description. Arguably the NULL behaviour is right — nothing needs converting — so this is about the sentence rather than the code.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -1806,6 +1883,159 @@ mod tests {
...
+        assert_eq!(
+            unsafe { sql_fetch_scroll(h.stmt, SQL_FETCH_NEXT, 0) },
+            SQL_SUCCESS
+        );
+        assert_eq!(values, [10, 20]);
+        assert_eq!(indicators, [4, 4]);
+        assert_eq!(
+            stmt.inner.lock().unwrap().bindings[0].target_type,
+            SQL_C_DEFAULT,
+            "the persistent binding must be resolved again for later result sets"
```
Vahid-b, line 2219: **Nit.** This pins the negative half of the cross-result-set contract — the stored binding stays `SQL_C_DEFAULT` — and `default_bindings_resolve_wide_and_guid_columns_to_typed_targets` pins resolution against one metadata set. Nothing drives *two* result sets with different column types through the same binding, which is summary bullet 3 ("Keep the persistent binding unresolved so it can be reused correctly across result sets").

Confirmed by grep: no `SQLMoreResults` anywhere in `fetch_scroll_test.cpp`, and no `more_results` in the `fetch_scroll` unit tests.

A nit rather than a suggestion because the composition does follow from the two halves that *are* pinned (a fresh clone per fetch, `begin_result_set` replacing `column_metadata`), so I could not construct a mutation that survives both existing tests and breaks the cross-set case. Worth a line if a natural place exists in the e2e suite.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -488,12 +491,78 @@ impl RowWriter for BoundRowWriter<'_> {
...
+
 fn fetch_scroll_safe(
     statement_handle: SqlHandle,
     stmt: &StmtHandle,
     fetch_orientation: SqlSmallInt,
     _fetch_offset: SqlLen,
 ) -> SqlReturn {
+    // The declared ODBC version selects the SQL_C_DEFAULT table. Read it before
+    // the stmt lock to preserve parent-before-child lock ordering.
+    let odbc_version = {
```
Vahid-b, line 617: **Nit / question, not a defect** — I have not measured this, so no regression is being claimed.

The lock *ordering* is right and matches `bind_param.rs:131-138` and `catalog.rs:467`; that part is settled. The difference is cadence. `SQLBindParameter` pays this once per bind; this pays it once per `SQLFetchScroll` — i.e. once per row in the classic `SQLBindCol` + `SQLFetch` loop — on a mutex shared by every statement on every connection under the `HENV`. It is also taken before any validation, so a closed-cursor or bad-orientation call pays it too, and before the `result_set_exhausted` fast path whose comment says it "needs no connection access at all".

Gating it on `bindings.iter().any(|b| b.target_type == SQL_C_DEFAULT)` is not possible without inverting the lock order. Snapshotting `odbc_version` onto the `DbcHandle` at alloc would be, and `alloc_handle.rs:113-119` already asserts the version is set by then — though `set_env_attr.rs:78-81` writes it unconditionally rather than enforcing the DM's "before any connection handle" rule, so caching would be a small behaviour change on a path the DM already forbids.

Either way, a sentence recording that the per-fetch read was deliberate would stop a later reader treating it as an oversight.

Separately, and matching the `bind_param` precedent: the poison arm below returns `SQL_ERROR` before `free_errors`, so `SQLGetDiagRec` afterwards reads the *previous* call's records. These are also the two lines the coverage comment reports as uncovered.

Vahid-b (commented):
## Summary

`SQLBindCol` stops rejecting `SQL_C_DEFAULT` and each fetch resolves the placeholder from the current result set's column types, through the resolver already shared with `SQLBindParameter`. The design matches the reference driver's shape and every parity claim in the description checks out against the msodbcsql source: `rgbTRANSTYPE` / `rgbTRANSTYPE380` at `sqlcmisc.cpp:153` and `:192` both read `SQL_C_CHAR, //  SQL_GUID`, both read `SQL_C_WCHAR, //  SQL_XML_MAPPED`, and `BindOffset` (`sqlcfunc.cpp:2176-2284`) has a `SQL_C_GUID` case at `:2242` but none for `SQL_C_SS_TIME2` / `SQL_C_SS_TIMESTAMPOFFSET`, so those fall to `default: dwOffset = lpbindinfo->cbValueMax; fFixed = FALSE;` at `:2280-2283`. All three newly-registered divergences are accurate.

**No blocking findings.** Two suggestions and four nits, inline; most are documentation precision rather than code.

Incidental win worth naming: before this PR `set_desc_field.rs`'s `set_type` already accepted `SQL_C_DEFAULT` on an ARD record via `is_valid_c_type`, while `SQLBindCol` refused it — contradicting that function's own doc comment that "a descriptor-set type can never diverge from what direct binding accepts". This closes that.

## Evidence

Measured on this head (`dab4f5c5`), Linux, `cargo nextest run -p mssqlodbc --lib`:

- Baseline: 1167 passed / 0 failed / 1 skipped.
- Mutation — drop `.saturating_sub(1)` from the ordinal lookup at `fetch_scroll.rs:539`: **3 failures** (`default_binding_resolves_from_current_result_metadata`, `default_bindings_resolve_wide_and_guid_columns_to_typed_targets`, `a_default_binding_follows_the_declared_odbc_version`). Guarded.
- Mutation — replace the env read at `:557-564` with a hardcoded `OdbcVersion::Odbc3_80`: **exactly 1 failure**, `a_default_binding_follows_the_declared_odbc_version`. Independently reproduces the wiring gap raised earlier and confirms the new test closes it.

## Coverage

| Row | Status |
|---|---|
| Primary logic | Examined — `resolve_default_bindings`, `fetch_scroll_safe`, `sql_bind_col_safe` |
| Siblings | Examined — `deliver_bound`, `deliver_fixed_bound`, `deliver_encoded_string`, `deliver_bound_plp`; the `SQLFetch` delegation; all four `parent_env()` lock sites |
| Callers & implementers | Examined — every `.bindings` consumer (only `fetch_scroll` and `handles/stmt`), all `RowWriter` methods on `BoundRowWriter` |
| Diagnostics | Examined — `RowIssue` mapping, poison early-return ordering relative to `free_errors` |
| Tests | Examined, plus the two mutations above |
| Build, packaging, pipelines | Not applicable — no such files in the diff |
| Docs & comments | Examined — msodbcsql citations verified line by line against a local checkout |
| Description & work items | Examined — 6/6 files map to description bullets, no riders; AB#47481 linked; `Changed` is the right heading |
| CI evidence | Examined — all checks green, coverage comment present at 98% diff |

**Not verified:** no live SQL Server, so the e2e test and the msodbcsql18 probe values in the description are not reproduced; Linux host only, no Windows or macOS build; `mssql-python` cross-repo leg read as green rather than run.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -1806,6 +1883,159 @@ mod tests {
...
+        dbc.runtime
+            .block_on(client.execute("SELECT time column".to_string(), ()))
+            .unwrap();
+        {
+            let mut state = dbc.inner.lock().unwrap();
+            state.client = Some(client);
+            state.active_stmt = Some(h.stmt);
+        }
+
+        unsafe { sql_fetch_scroll(h.stmt, SQL_FETCH_NEXT, 0) };
```
Vahid-b: **Nit.** The return code is discarded, so the helper reads `last_state` whether or not the fetch reached row delivery at all. If a future change made `fetch_scroll_safe` fail earlier with an unrelated diagnostic, the assertions in `a_default_binding_follows_the_declared_odbc_version` could match on something other than the per-row outcome they are named for.

Asserting the `SqlReturn` alongside the SQLSTATE would keep the test pinned to the row outcome rather than to whichever diagnostic happens to be last.

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: You're right, and this turned out to be more than a documentation problem — thank you for pushing on it.

I verified the write path independently: `element_stride` returns a constant `16` for `SQL_C_GUID`, and `convert_guid_c` ends in `write_fixed`, which does `(ptr as *mut T).write_unaligned(value)` with no length parameter. So a 4-byte declared slot gets 16 bytes.

Rather than only scoping the claim, I fixed it. `resolve_default_bindings` now refuses to resolve when the resolved fixed-width target is wider than a non-zero declared `BufferLength`; the binding stays `SQL_C_DEFAULT` and the row fails with `HYC00` instead of writing. Your framing is what made the fix scopeable: the width contract belongs to the application only when it *named* the type, so the guard lives in the default-resolution path and explicit binds are untouched.

`BufferLength` 0 is exempt. I went back and forth on this given your "sharpest case is `N = 0`" point, but 0 is the documented idiom for a fixed-width target and carries no width claim — and treating it as a 0-width slot would break the ordinary `SQLBindCol(1, SQL_C_DEFAULT, &my_int, 0, &ind)` pattern for every fixed type, not just GUID. Recorded as a divergence from msodbcsql (which strides `cbValueMax`) rather than silently.

Two tests, both confirmed to fail without the guard:
- `a_default_binding_too_narrow_for_its_fixed_target_stays_unresolved` — covers the narrow, zero, and wide cases.
- `a_narrow_default_guid_binding_never_writes_past_the_declared_buffer` — backing array is 64 bytes with a declared `BufferLength` of 4, so a regression shows up as bytes written past the declared width rather than as a crash. Without the guard this returns `Success` and the poison bytes are gone, which is the overflow demonstrated directly.

Claims scoped in all three places you listed: the doc comment, `typed-columnar-fetch-plan.md:170`, and the PR description.

On the `01004`-and-truncate alternative: agreed it is the closer match to msodbcsql, and I have not done it here — it needs a chosen SQLSTATE and sign-off, and refusing is the safe direction in the meantime. Recorded as open and untracked in both the plan doc and the deviation registry.

`6cb85737`.

> [!NOTE]

.github/instructions/mssql-odbc.instructions.md
David-Engel: Verified and registered. `describe_col.rs:208-210` maps `VarBinary`/`BigVarBinary` → `SQL_VARBINARY` and `Image` → `SQL_LONGVARBINARY`; `type_rules.rs:295` maps all three binary SQL types → `SQL_C_BINARY`; `deliver_bound` refuses it pending AB#47239.

Your point that it covers more common column types than `time`/`datetimeoffset` is the part that made me register it rather than treat it as covered by the general "unsupported target" sentence — a `varbinary` column is a much more likely thing to hit by accident than a `time` column bound without naming a C type.

Added to both places, in the same shape as the stride bullet above it and pointing at AB#47239.

Good catch that the PR's own new test already demonstrates the path — `a_default_binding_follows_the_declared_odbc_version` expects `HYC00` at ODBC 3.0 for exactly this reason. That made it self-evident once you named it.

`6cb85737`.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Correct — `deliver_bound` short-circuits `ColumnValues::Null` before any target dispatch, so an unresolvable column returns `Success` with `SQL_NULL_DATA` for a NULL row and only a row carrying a value reaches `RowIssue::Unsupported`.

Reworded in all three places: "reported as an unsupported target for any row carrying a value; a NULL row still reports `SQL_NULL_DATA` through the indicator, because nothing needs converting."

Agreed the behaviour is right — I stated the reason in the text rather than just narrowing the claim, since "why doesn't this fail?" is the obvious next question and the answer (nothing to convert) is the justification for keeping it.

Thanks for separating this from the ordinal sentence explicitly; it made clear which half was still wrong after 83fa015e.

`6cb85737`.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Real gap, and I added the test — `DefaultTargetResolvesPerResultSet` in `fetch_scroll_test.cpp`. One `SQL_C_DEFAULT` binding, `SELECT CAST(4242 AS INT); SELECT CAST('hello' AS VARCHAR(16))`, fetch, `SQLMoreResults`, fetch: the same binding must resolve to `SQL_C_SLONG` for the first set and `SQL_C_CHAR` for the second, checked through both the delivered value and the indicator (4 vs 5).

Your grep was right that nothing drove two sets through one binding, and I agree it was a nit rather than a defect — I could not construct a mutation that breaks it either. But it is bullet 3 of the PR summary, so it should not have rested on composition alone.

One deliberate choice: I used `int` and *narrow* `varchar`, which both drivers resolve identically, so this runs on the msodbcsql leg too and actually compares rather than carrying `SKIP_IF_COMPARING_MSODBCSQL()`. I had written it with the skip first, then realised the divergence is only wide/GUID and that skipping would have thrown away free parity coverage. Verified passing against retail msodbcsql18 locally.

`6cb85737`.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Recorded, as you suggested — a comment now states the per-fetch read is deliberate and why, so a later reader does not take it for an oversight.

On the alternatives: gating on `bindings.iter().any(...)` does invert the lock order, as you say. Caching on the `DbcHandle` at alloc is the interesting one, and I decided against it for the reason your own parenthetical raises — `set_env_attr.rs` writes the version unconditionally, so a cached copy could go stale against a path the code does not currently refuse. Baking in a value that `SQLSetEnvAttr` can still overwrite trades a cheap uncontended read for a correctness hazard. I noted it as an uncontended read of one `Copy` field, taken before validation so there is exactly one acquisition site rather than one per early-return path.

If the cadence does show up in profiling, the honest fix is probably enforcing the DM's "before any connection handle" rule in `set_env_attr` first, then caching — which is a change of its own rather than something to slip in here.

On the second point: agreed, the poison arm returns before `free_errors`, so `SQLGetDiagRec` afterwards reads the previous call's records. That matches the `bind_param.rs` precedent exactly, and those are the same two lines the coverage bot reports uncovered. I have left both as-is deliberately: making this one entry point differ from its six siblings would be worse than the shared quirk, and it is a pre-existing pattern rather than something this PR introduces. Worth a sweep across all of them if you think it is worth fixing — happy to file that separately.

`6cb85737`.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Fixed — the helper now returns `(SqlReturn, [u8; 5])` and both assertions check the pair.

This was more valuable than it looked. Pinning the return code immediately caught that my assumption about it was wrong: I first wrote `SQL_SUCCESS_WITH_INFO` and the test failed showing the real value is `SQL_ERROR` — a single-row rowset whose only row fails gives `SQL_ERROR`, not a warning. So the test was previously asserting on a diagnostic without knowing what the call actually returned, which is exactly the blindness you described.

`6cb85737`.

> [!NOTE]

shiwanigupta0809:
Re-reviewed on this head, after #436 merged. Read the new width guard, `resolve_default_bindings`, and the descriptor plumbing #436 landed underneath this. Not re-raising anything from the existing threads — the GUID overrun, the binary-PLP registry entry, the NULL wording, the per-result-set test, and the env-lock cadence are all settled.

### Suggestion

**This PR closes a real hole #436 left open, and the rebase is the moment to claim it — with a test — rather than let it look like an incidental side effect.**

The branch is `CONFLICTING` against `main` now. The mechanical part is small: `fetch_scroll.rs:705` still does `stmt_state.bindings.clone()`, and `main:644` is `ColumnBinding::all_from_ard_state(&desc_state)`. `resolve_default_bindings` mutating the per-fetch snapshot stays correct through that — the snapshot is derived fresh from the ARD each fetch, so the persistent record still keeps `SQL_C_DEFAULT` and bullet 3 of the summary survives unchanged.

The semantic part is the one worth acting on. **`SQL_C_DEFAULT` can already reach an ARD record on `main` today, through the descriptor API that #436 added** — and it behaves differently from the `SQLBindCol` path:

| Path on current `main` | Result |
|---|---|
| `SQLBindCol(1, SQL_C_DEFAULT, buf, n, &ind)` | `SQL_ERROR`, `HY003` (`bind_col.rs:172-176`) |
| `SQLSetDescField(ard, 1, SQL_DESC_CONCISE_TYPE, SQL_C_DEFAULT)` + `SQL_DESC_DATA_PTR` | succeeds, then `HYC00` + `SQL_ROW_ERROR` on every row |

The descriptor path gets there because `set_type` takes the `kind.is_application()` branch, `canonical_c_type(99)` is a no-op (99 is outside the `SQL_C_DATE..=SQL_C_TIMESTAMP` fold), and `is_valid_c_type` lists `SQL_C_DEFAULT` — so `r.concise_type = SQL_C_DEFAULT` is written. `ColumnBinding::from_record` keys "bound" off a non-null `SQL_DESC_DATA_PTR`, not the concise type, so it hands that binding to the fetch loop, where `deliver_bound` refuses it per row.

That is precisely the state #436's own summary says cannot occur:

> so `SQLBindCol`/`SQLBindParameter` and an equivalent `SQLSetDescFieldW`/`SQLSetDescRec` sequence can never produce contradictory state.

Two things follow, and neither is a criticism of the change as written:

**1. This is a stronger justification than the one the description gives.** The current framing is "no known consumer binds `SQL_C_DEFAULT` on the fetch path — mssql-python uses it only for parameters", which reads as speculative. The concrete version is that the equivalence #436 was built to guarantee is currently broken for exactly this value, and this PR is what restores it. Worth one line in the description.

**2. The descriptor route deserves a test, because it reaches `resolve_default_bindings` by a different door.** Every new unit and e2e case goes through `SQLBindCol`. After the rebase, a `SQLSetDescField`-established `SQL_C_DEFAULT` binding will resolve identically — that is a genuine improvement over `main`'s per-row `HYC00`, and nothing pins it. A `SQLSetDescField` variant of `default_binding_resolves_from_current_result_metadata` costs a few lines and covers the door that is currently untested in both directions.

Worth checking while you are in there, since the same rebase decides it: the new width guard reads `binding.buffer_length`, which post-#436 is `record.octet_length` and is independently writable via `SQLSetDescField(SQL_DESC_OCTET_LENGTH)` after the bind. Reading it per fetch from the fresh snapshot keeps that correct, but the guard's rationale — "the width contract belongs to the application only when it named the type" — is now stated about a field the application can change without rebinding, so the comment is worth re-reading against the descriptor model even though I believe the code is right.

shiwanigupta0809 (commented):
_Reviewed — see https://github.com/microsoft/mssql-rs/pull/460#issuecomment-5513608126_

David-Engel:
Thanks — the descriptor-route point was the valuable one, and I acted on both parts in 2122e20d.

First, a timing note: the rebase you flagged as outstanding is done. #436 merged at 17:20 UTC and I reconciled in 09a20858 a few minutes before your review posted, so `fetch_scroll.rs` no longer does `stmt_state.bindings.clone()`. Your read of what the reconciliation needed was right, including that `resolve_default_bindings` mutating the per-fetch snapshot stays correct through it — the ARD-derived snapshot is rebuilt each fetch, so the persistent record still keeps the placeholder. The one thing the mechanical description missed is that `resolve_default_bindings` also needed its input changed: `column_metadata` is only reachable under the STMT lock, and the bindings are now read under the DESC lock after that lock is released, so the column SQL types are snapshotted under STMT and carried out.

> the descriptor route deserves a test, because it reaches `resolve_default_bindings` by a different door

Agreed, and added — `a_default_binding_set_through_the_descriptor_api_also_resolves`. I verified your causal chain first rather than taking it on the description: `set_type` takes the `kind.is_application()` branch, `canonical_c_type(99)` is a no-op, `is_valid_c_type` lists `SQL_C_DEFAULT`, and `from_record` returns `Some` on a non-null `SQL_DESC_DATA_PTR` regardless of concise type. All four hold.

Writing it surfaced something worth recording: the descriptor API exposes `SQL_DESC_INDICATOR_PTR` and `SQL_DESC_OCTET_LENGTH_PTR` as separate fields, where `SQLBindCol`'s single `StrLen_or_Ind` argument feeds both. Setting only the indicator resolves the value correctly but reports no length. So the two doors are not quite interchangeable at the application level even now, and the test sets both to be a faithful equivalent. That is ODBC-correct rather than a defect, but it is exactly the kind of thing a `SQLBindCol`-only test suite cannot see.

> This is a stronger justification than the one the description gives

Taken. The bind/descriptor equivalence framing is now in the plan doc with the mechanism spelled out, replacing the speculative "no known consumer" reasoning.

> the guard's rationale ... is now stated about a field the application can change without rebinding

Good catch, and you are right that the code is correct but the comment was not carrying its weight. `buffer_length` is `record.octet_length` post-#436 and is independently writable via `SQLSetDescField` after the bind. The guard is still right precisely because it runs per fetch against the same snapshot the fill loop uses, so a slot narrowed after binding is caught rather than missed — I have said that explicitly rather than leaving the reader to reconstruct it.

1,237 unit tests pass, fmt and clippy clean.

Theekshna (commented):
Unattended re-review: head moved from `dab4f5c5` (my last marked review, findings=0) to `2122e20d` via two new commits — `f90e2419` (refuse a `SQL_C_DEFAULT` fixed-width target wider than the bound `BufferLength`) and `2122e20d` (cover the descriptor/`SQLSetDescField` route into the same resolution path) — plus merge commits pulling in unrelated `#436`/`#453`/`#463` work with no conflicts in this PR's own files.

Reviewed both new commits in full against `mssql-odbc/src/api/fetch_scroll.rs` and the two docs they touch.

| Severity | Count |
| --- | --- |
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

**What I checked:**
- The new narrow-buffer guard in `resolve_default_bindings` is generic over `element_stride`, not GUID-specific despite the PR description's framing — it correctly also covers `SQL_C_TYPE_DATE/TIME/TIMESTAMP` and `SQL_C_SS_TIME2`/`SQL_C_SS_TIMESTAMPOFFSET` at ODBC 3.8, since all route through the same `fixed_width > 0 && buffer_length > 0 && buffer_length < fixed_width` check. Only the GUID case is unit-tested directly, but since the guard's correctness depends only on the arithmetic comparison (not on which C type triggers it), one concrete instantiation is adequate coverage; I don't think a test per fixed-width type is needed here.
- `BufferLength == 0` exemption verified correct and tested (`a_default_binding_too_narrow_for_its_fixed_target_stays_unresolved`) — matches the documented idiom for fixed-width targets.
- New overrun test (`a_narrow_default_guid_binding_never_writes_past_the_declared_buffer`) asserts bytes past `BufferLength` are untouched using an oversized backing array — good regression shape, confirms the guard stops the write rather than just returning early.
- Descriptor-route fix: verified `set_type`/`is_valid_c_type`/`canonical_c_type` and `ColumnBinding::from_record` in the current tree — the bind/descriptor disagreement (`HY003` via `SQLBindCol` vs `HYC00` per-row via `SQLSetDescField`) is real and the fix is the right one (key both paths through the same `resolve_default_bindings`). New unit test drives the descriptor path standalone and asserts identical resolution.
- msodbcsql parity: N/A for these two commits specifically (no new parity claims beyond what was already reviewed and documented at `dab4f5c5`; GUID/wide-char deviations were already verified against `C:\REPOS\Drivers\ODBC_2` in the earlier review of this PR).
- Divergences: both new behaviors (narrow-buffer refusal, descriptor-route equivalence) are documented in `.github/instructions/mssql-odbc.instructions.md` and `mssql-odbc/docs/typed-columnar-fetch-plan.md`, consistent with the code.
- PR description: matches current diff and testing claims; linked to `AB#47481`. `cargo check -p mssqlodbc --lib --tests --offline` compiles clean; full `nextest` run not possible in this sandbox (pre-existing local toolchain codegen limitation in `crc32fast`, unrelated to this PR) — CI (build 171912) was still pending on all build/test legs at review time, no failures observed.
- No AI slop found in the new doc comments — they explain *why* (lock ordering, snapshot-vs-bind-time consistency, ARD as source of truth), not restating the diff.

No re-filed findings — nothing outstanding from prior threads on this PR to re-check; all were already resolved before this push.

shiwanigupta0809 (approved):

PR #461, opened 2026-09-01
Release busy connections after parameterized results

Theekshna:
## Description

Fix `HY000 Connection is busy with results for another command` after a parameterized query returns rows.

A row set inside an RPC ends with `DONEINPROC` carrying `DONE_MORE` because `RETURNVALUE`, `RETURNSTATUS`, and the terminal `DONEPROC` still follow. The ODBC layer read that flag as "another result is pending" and kept the statement's claim on the connection, so a second statement was refused even though the first was fully consumed.

`TdsClient::complete_current_result` now settles that protocol-only tail through the existing `settle_rpc_terminator`, which absorbs the control tokens and parks the first non-tail token for `SQLMoreResults`. The gate is `DONEINPROC`-only, matching msodbcsql's `OnDone` (`sqlctokn.cpp:2208-2240`); ordinary batch `DONE` tokens are never probed, so a multi-statement batch does not block on the next statement's result inside a fetch.

Failures while consuming the tail abandon the batch and defer the error to the next statement call, since the fetch has already committed to returning its row. Cancellation and timeout are distinguished: an acknowledged ATTENTION leaves the connection reusable, while failed cleanup disarms session recovery.

Also in this change:

- `check_and_reconnect` rejects a transport already marked known-dead when session recovery is unavailable, and `close_connection` records that terminal state. This prevents JS, Python, CLI, ODBC, and pool callers from issuing another request over a retired transport.
- The Windows E2E runner discards a CMake build tree configured from an incompatible WSL source path.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47770

## Tests

New coverage for prepared and direct execution, null, zero-row and multi-row parameters, `SQLGetData`, statement close and free, prepared reuse, multiple result sets, plain-batch non-probing, the tail-gate lifecycle across results, truncated tails, and cancellation with live and known-dead transports.

## Validation

- `cargo bfmt` and `cargo bclippy` pass
- `mssql-tds --lib`: 2050 passed; the only failures are pre-existing missing certificate fixtures unrelated to this change
- `mssqlodbc --lib`: 1168 passed
- Live mssql-tds RPC integration suite passes
- ODBC C++ E2E suite passes
- Retail msodbcsql 18.06.0002 parity run passes

Known CI agent issues unrelated to this change: `llvm-cov` and `cargo-clippy` missing on the macOS agent, Colima Docker setup failing, and a Linux ARM remote provider timeout.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

### DIFF
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -4454,6 +4454,9 @@ impl TdsClient {
     /// The response-tail control tokens that precede `DONEPROC` are absorbed
     /// exactly as the main loop absorbs them.
     ///
+    /// Every error exit ends the batch, so no caller can propagate a tail
+    /// failure over a batch nothing will finish reading.
+    ///
     /// Unlike msodbcsql's short, non-failing single-byte peek, this consumes the
     /// trailer under the request's remaining timeout. An expiry therefore fails
     /// the request after the attention acknowledgement restores synchronization;
@@ -4515,10 +4518,33 @@ impl TdsClient {
                 }
                 other => {
                     self.parked_token = Some(Box::new(other));
-                    return processing_error.map_or(Ok(()), Err);
+                    return match processing_error {
+                        None => Ok(()),
+                        Some(error) => Err(self.abandon_after_tail_failure(error).await),
+                    };
                 }
             }
         }
+    }
+
+    /// Ends the batch after a tail token failed to process, keeping the rule
+    /// "a tail failure always ends the batch" with the loop that breaks it.
+    ///
+    /// The look-ahead parked whatever it could not process, so the rest of the
+    /// response is unread and nothing will come back for it. Leaving the batch
+    /// open masks the returned error behind `ALREADY_EXECUTING_ERROR` on every
+    /// later command for the whole connection.
+    async fn abandon_after_tail_failure(
+        &mut self,
+        error: crate::error::Error,
+    ) -> crate::error::Error {
+        if self.execution_context.has_open_batch() {
+            if let Err(drain_error) = self.drain_stream_or_retire().await {
+                warn!(error = ?drain_error, "Drain after an RPC tail failure failed");
+            }
+            self.execution_context.set_has_open_batch(false);
+        }
+        error
     }
 
     fn fail_rpc_terminator_look_ahead(&mut self, error: &crate::error::Error) {
@@ -4641,17 +4667,6 @@ impl TdsClient {
         if let Err(error) = self.settle_rpc_terminator(&parser_context).await {
             self.current_metadata = None;
             self.abort_pending_prepare_capture();
-            // A tail token that failed to process before the terminal DONEPROC
-            // leaves the rest of the response unread. The caller has an error to
-            // report and no way back to this batch, so consume it here rather
-            // than leaving the connection claimed by a statement that will never
-            // read it.
-            if self.execution_context.has_open_batch() {
-                if let Err(drain_error) = self.drain_stream_or_retire().await {
-                    warn!(error = ?drain_error, "Drain after an RPC tail failure failed");
-                }
-                self.execution_context.set_has_open_batch(false);
-            }
             return Err(error);
         }
         Ok(!self.execution_context.has_open_batch())
@@ -10069,6 +10084,35 @@ mod tests {
             Err(crate::error::Error::ColumnEncryptionError(_))
         ));
         assert!(!client.has_open_batch());
+        assert!(client.parked_token.is_none());
+    }
+
+    #[tokio::test]
+    async fn advance_abandons_the_batch_when_a_tail_token_fails() {
+        // The other `settle_rpc_terminator` call site. Propagating the failure
+        // with the batch still open would mask it behind ALREADY_EXECUTING_ERROR
+        // on every later command for the whole connection.
+        let transport = TestTransport::with_tokens(vec![
+            info_token(0, 0, "print"),
+            done_in_proc_more(),
+            Tokens::ReturnValue(ae_return_value_token(
+                "@out",
+                ColumnValues::Bytes(vec![1, 2, 3]),
+                Some(ae_crypto_metadata()),
+            )),
+            empty_col_metadata(),
+            done_no_more(),
+        ]);
+        let mut client = create_test_client_with_transport(transport);
+        client.current_command_ce_setting = ExecutionColumnEncryptionSetting::Enabled;
+        client.current_result_set_has_been_read_till_end = true;
+        client.execution_context.set_has_open_batch(true);
+
+        assert!(matches!(
+            client.advance().await,
+            Err(crate::error::Error::ColumnEncryptionError(_))
+        ));
+        assert!(!client.command_is_busy());
         assert!(client.parked_token.is_none());
     }
 
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -4184,6 +4184,7 @@ impl TdsClient {
                     // Use saturating_add to prevent integer overflow from malicious/corrupted TDS responses
                     *count = count.saturating_add(done.row_count);
                     self.current_result_set_has_been_read_till_end = true;
+                    self.current_result_ended_with_done_in_proc = is_done_in_proc;
 
                     let is_last = !done.has_more();
 
@@ -4242,6 +4243,7 @@ impl TdsClient {
                             // trailing DONEPROC; find out before the caller
                             // treats it as another result.
                             self.settle_rpc_terminator(&parser_context).await?;
+                            self.current_result_ended_with_done_in_proc = false;
                         }
                         return Ok(ResultBoundaryKind::NoRows {
                             rows_affected: if has_update_count {
@@ -4399,11 +4401,14 @@ impl TdsClient {
         }
     }
 
-    /// Returns `true` while the current batch still has unconsumed results on
-    /// the wire or in [`parked_token`](Self::parked_token) (a positioned result
-    /// set, an RPC-tail look-ahead token, or further statements to navigate to).
+    /// Returns `true` while the current batch still has unconsumed results (a
+    /// positioned result set, or further statements to navigate to).
     /// Used by the ODBC layer to decide whether the connection stays busy after
     /// positioning on a no-row statement result.
+    ///
+    /// A token parked by the RPC-terminator look-ahead is covered without being
+    /// inspected here: parking means the terminal DONEPROC was not reached, so
+    /// the flag this reads is still set.
     pub fn has_open_batch(&self) -> bool {
         self.execution_context.has_open_batch()
     }
@@ -4532,7 +4537,8 @@ impl TdsClient {
 
     /// Returns `true` when an open result batch or streamed write blocks a new command.
     pub(in crate::connection) fn command_is_busy(&self) -> bool {
-        self.has_open_batch() || !matches!(self.streamed_write_state, StreamedWriteState::Idle)
+        self.execution_context.has_open_batch()
+            || !matches!(self.streamed_write_state, StreamedWriteState::Idle)
     }
 
     /// Returns `true` when the client is currently positioned on a row-returning
@@ -4608,6 +4614,11 @@ impl TdsClient {
     /// `DONEINPROC`, [`settle_rpc_terminator`](Self::settle_rpc_terminator)
     /// consumes only that control-token tail and parks the first non-tail token
     /// for [`advance`](Self::advance). Plain batch DONE tokens are never probed.
+    ///
+    /// The `DONEINPROC` gate is msodbcsql's: `OnDone` probes the tail only for
+    /// `RS_SELECTION && DONEINPROC && DONE_MORE` (`sqlctokn.cpp:2208-2240`).
+    /// Widening it to every DONE would make an ordinary multi-statement batch
+    /// block here on the next statement's result.
     ///
     /// Returns `true` only when the batch is fully exhausted.
     pub async fn complete_current_result(&mut self) -> TdsResult<bool> {
@@ -6801,6 +6812,7 @@ impl TdsClient {
         self.cancel_handle = None;
         self.active_row_read_state = ActiveRowReadState::Idle;
         self.row_already_positioned = false;
+        self.parked_token = None;
         self.current_command_ce_setting = ExecutionColumnEncryptionSetting::UseConnectionSetting;
         self.execution_context.set_has_open_batch(false);
 
@@ -9989,6 +10001,32 @@ mod tests {
         assert!(!client.has_open_batch());
         assert!(client.is_connection_dead());
         assert!(client.current_result_set_has_been_read_till_end);
+    }
+
+    #[tokio::test]
+    async fn advancing_to_a_no_row_result_clears_the_rpc_tail_gate() {
+        let transport = TestTransport::with_tokens(vec![
+            done_in_proc_more(),
+            done_count(CurrentCommand::Update, 3, true),
+        ]);
+        let mut client = create_test_client_with_transport(transport);
+        client.current_metadata = Some(stale_metadata());
+        client.execution_context.set_has_open_batch(true);
+
+        assert!(!client.peek_past_current_row().await.unwrap());
+        assert!(!client.complete_current_result().await.unwrap());
+        assert!(client.current_result_ended_with_done_in_proc);
+
+        assert!(matches!(
+            client.advance().await,
+            Ok(StatementResult::NoRows {
+                rows_affected: Some(3)
+            })
+        ));
+        assert!(
+            !client.current_result_ended_with_done_in_proc,
+            "a plain batch DONE must not leave the previous RPC's tail gate armed"
+        );
     }
 
     #[tokio::test]
--- a/mssql-tds/src/connection/tds_client.rs
+++ b/mssql-tds/src/connection/tds_client.rs
@@ -36,7 +36,9 @@ use crate::{
         ColumnPolicy, ParserContext, PlpPauseState, RowHeader, RowPauseState, RowReadResult,
     },
     message::{batch::SqlBatch, messages::Request},
-    token::tokens::{ColMetadataToken, CurrentCommand, DoneStatus, EnvChangeTokenSubType, Tokens},
+    token::tokens::{
+        ColMetadataToken, CurrentCommand, DoneStatus, DoneToken, EnvChangeTokenSubType, Tokens,
+    },
 };
 use std::collections::HashMap;
 use std::future::Future;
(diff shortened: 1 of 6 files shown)


shiwanigupta0809 (commented):
_Reposted as a top-level comment for visibility: https://github.com/microsoft/mssql-rs/pull/461#issuecomment-5507090136_

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4571,6 +4605,42 @@ impl TdsClient {
...
+        }
+
+        let parser_context = ParserContext::ColumnEncryption(
+            self.negotiated_settings.is_column_encryption_supported(),
+        );
+        if let Err(error) = self.settle_rpc_terminator(&parser_context).await {
+            self.current_metadata = None;
+            self.abort_pending_prepare_capture();
+            return Err(error);
+        }
```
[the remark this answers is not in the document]
Theekshna: Verified and fixed in 3229933c. The wedge is real: `settle_rpc_terminator` accumulates a processing error (for example a RETURNVALUE that fails `finalize_return_value`) and returns it from the park branch with `has_open_batch` still set. ODBC then keeps `active_stmt` while marking the result exhausted, and `SQLFetch`'s fast path clears `STMT_STATE_CURSOR_OPEN` when it reports the deferred error — after which `SQLCloseCursor`, `SQLFreeStmt(SQL_CLOSE)` and `SQLFreeHandle` all skip `drain_and_release`, so the claim is never released.

`complete_current_result` now drains the remainder through `drain_stream_or_retire` and clears the batch before returning the original error, so the claim is released and the connection is retired if that drain also fails. Pinned by `complete_current_result_abandons_the_batch_when_a_tail_token_fails`, which scripts a failing encrypted RETURNVALUE followed by another result set and asserts both the closed batch and an empty `parked_token`.

Theekshna:
Thanks — the suggestion is right, and it resolved itself in a stronger way than either option you offered.

That review was against 07d227b9. Commit 3229933c (pushed just before it, addressing the Copilot finding on the same arm) removed case 4 as a distinct outcome: when a `processing_error` surfaces at the park arm, `complete_current_result` now drains the remainder through `drain_stream_or_retire` and clears the batch before returning the error. So all four failure paths now leave `has_open_batch == false`, `batch_done` is unconditionally true for `Some(Err(_))`, and `|| completion_failed` was dead code rather than a policy choice.

Dropped it in 8fea826a and restored the original paragraph verbatim, with a short addition stating why the tail-failure path can only take the deferred-error route. The rule you quoted is intact: a failure that does not end the batch stays unposted.

On pinning: the invariant that case 4 closes the batch is pinned at the TDS layer by `complete_current_result_abandons_the_batch_when_a_tail_token_fails`, which scripts a failing encrypted RETURNVALUE followed by another result set and asserts both the closed batch and an empty `parked_token`. Given that, the ODBC branch is fully determined by `batch_done`, which `release_busy_if_row_exhausted_defers_a_truncated_rpc_tail` already covers — so I did not add a second AE-scripted fixture at the ODBC layer.

One correction to the verified section: this PR does not touch `record_error_token` or severity handling. `git diff origin/main..HEAD` has no match for either. That belongs to a different change.

Validation on 8fea826a: `mssqlodbc --lib` 1168 passed, `cargo bfmt` and `cargo bclippy` clean.

shiwanigupta0809:
Reviewed the full diff plus the code it lands on, and checked the msodbcsql parity claim against the source rather than taking it from the description.

### Verified

The `DONEINPROC`-only gate is msodbcsql's. `OnDone` (`sqlctokn.cpp:2208`) probes exactly `lpstmt->RSType == RS_SELECTION && eDoneType == DONEINPROC && (lpDoneInfo->Status & DONE_MORE)`, then `PeekToken`s for `SQLRETURNSTATUS`; anything else sets `TDS_RESULTSPENDING` and stops parsing. The reason for keeping it narrow — a plain multi-statement batch must not block here on the next statement's result — is real, and `complete_current_result_does_not_probe_a_plain_batch` pins it.

The new flag cannot go stale. Every site that sets `current_result_set_has_been_read_till_end = true` either sets `current_result_ended_with_done_in_proc` alongside it (`handle_row_done`, `advance_to_result_boundary`) or closes the batch on an error path, where `complete_current_result` short-circuits at the `has_open_batch` gate before reading the flag. The buffered cursor path added in #418 is safe too: `try_receive_row_header` returns `Some` only for ROW/NBCROW, so a DONE always falls through to the async path and reaches `handle_row_done`.

Plain batches are behaviourally unchanged. `Ok(true)` from `complete_current_result` is equivalent to `!has_open_batch()` and `Ok(false)` to `has_open_batch()`, so `batch_done` computes exactly what the old expression did for every non-`DONEINPROC` shape, AB#47508's `SELECT 1; SELECT 2;` included.

Releasing the connection after a failed tail is safe rather than optimistic: `fail_rpc_terminator_look_ahead` closes the batch, `retire_after_failed_drain` marks the transport known-dead, and the new `check_and_reconnect` gate turns the next command into a fast `ConnectionClosed` instead of a write to a dead socket. Widening `record_error_token` to mark severity >= 20 dead is right — those terminate the session even when the socket still reads.

### Suggestion

**`|| completion_failed` is redundant on three of the four failure paths, and on the fourth it reverses the rule the replaced doc paragraph stated.**

`complete_current_result` can return `Err` four ways:

1. `next_response_token` fails — `fail_rpc_terminator_look_ahead` runs `set_has_open_batch(false)` first.
2. The 10,000-token guard trips — same.
3. A `processing_error` surfaces at the `DoneProc` arm — that arm already closed the batch.
4. A `processing_error` surfaces at the **park** arm — `parked_token = Some(..)`, batch stays open.

In 1-3 `batch_done` is already `true`, so the disjunct changes nothing. Case 4 is the only one it affects, and there `batch_done` is `false` and `release` is `false` — which is precisely the situation the paragraph this hunk replaced legislated for:

> A failure that does *not* end the batch (a raw transport error) is left unposted entirely: `release` is false in that case, so nothing about this statement's state changes, and the caller's very next real operation on this connection takes the normal, non-fast-path route and organically rediscovers and reports the same failure through its own existing error handling — posting it here too would attach a diagnostic to this call's own still-successful return.

The consequence is observable: `pending_fetch_error` is set while a result set sits parked and the connection stays claimed, so the next `SQLMoreResults` takes its `pending_fetch_error` branch, returns `SQL_ERROR`, and the parked result is never navigated to. Reachable through an Always Encrypted output-parameter decryption failure (`finalize_return_value`) in an RPC tail that has a following result set — uncommon, but not synthetic, and `capture_change_property` / `process_session_state` reach the same arm.

Neither answer is obviously wrong; what I'd like is for it to be deliberate. Both new ODBC tests exercise paths where the batch closes regardless, so nothing currently pins case 4 either way. If posting is intended, the doc comment should say so, since it now contradicts the rationale it replaced. If not, dropping `|| completion_failed` restores the previous rule and costs nothing on the other three paths.

Nice piece of protocol work otherwise — settling the tail through the existing `settle_rpc_terminator` rather than a second look-ahead is the right call, and gating on `DONEINPROC` from msodbcsql's own condition rather than inferring one is what makes it convincing.

Theekshna:
@shiwanigupta0809 thanks — this landed a couple of minutes after [my reply above](https://github.com/microsoft/mssql-rs/pull/461#issuecomment-5507065429), so you were reviewing 07d227b9. The suggestion is right and is already applied.

Short version: case 4 no longer exists. 3229933c (addressing the Copilot finding on the same arm) made the park-arm failure drain the remainder through `drain_stream_or_retire` and clear the batch before returning the error, so all four failure paths now leave `has_open_batch == false`. That made `batch_done` unconditionally true for `Some(Err(_))`, which turned `|| completion_failed` into dead code rather than a policy choice.

8fea826a drops it and restores your quoted paragraph verbatim, plus one sentence recording why the tail-failure path can only take the deferred-error route. The rule is intact: a failure that does not end the batch stays unposted.

On pinning — the invariant is now pinned at the TDS layer by `complete_current_result_abandons_the_batch_when_a_tail_token_fails`, which scripts a failing encrypted RETURNVALUE followed by another result set and asserts both the closed batch and an empty `parked_token`. With that, the ODBC branch is fully determined by `batch_done`, already covered by `release_busy_if_row_exhausted_defers_a_truncated_rpc_tail`, so I did not add a second AE-scripted fixture there.

One correction: this PR does not touch `record_error_token` or severity handling — `git diff origin/main..HEAD` has no match for either, so that verification belongs to a different change.

Validation on 8fea826a: `mssql-tds --lib` 2051 passed (the 7 failures are pre-existing missing certificate fixtures), `mssqlodbc --lib` 1168 passed, ODBC C++ e2e suite passes, `cargo bfmt` and `cargo bclippy` clean.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1204,16 +1212,29 @@ impl TdsClient {
...
             .recovery_context
             .client_context
             .as_ref()
             .map_or(0, |ctx| ctx.connect_retry_count);
+
+        // A cached dead verdict is authoritative. Without this gate, callers
+        // with recovery disabled (including intentionally closed clients and
+        // clients retired after protocol desynchronization) would fall through
+        // and write another request to the unusable transport.
+        if self.transport.connection_known_dead()
```
saurabh500: This only honors a cached known-dead verdict when recovery is unavailable. With the default `ClientContext::connect_retry_count == 1` and negotiated recovery, a live socket marked known-dead by `mark_connection_dead()` or a fatal ERROR token reaches the later `is_connection_dead()` poll; `NetworkTransport::is_connection_dead()` ignores `known_dead`, so `check_and_reconnect` returns `Ok(Duration::ZERO)` and the next command can write to a client that `is_connection_dead()` already reports as dead. Make the cached verdict skip the live-socket poll and either reconnect or fail:

```rust
let known_dead = self.transport.connection_known_dead();
if known_dead
    && (!self.recovery_context.session_recovery_negotiated || connect_retry_count == 0)
{
    return Err(crate::error::Error::ConnectionClosed(
        "Connection is dead and session recovery is unavailable".to_string(),
    ));
}

// ...

if !known_dead && !self.transport.is_connection_dead() {
    return Ok(Duration::ZERO);
}
```

I verified this with a temporary test that marks a live `TestTransport` known-dead while recovery is negotiated; it currently returns `Ok`, not an error/reconnect attempt.

Theekshna: Good catch — confirmed and fixed in 1d1b5f8b.

Verified the mechanism rather than taking it on trust: `NetworkTransport::is_connection_dead()` only forwards to `stream.is_connection_dead()` and never consults `known_dead`, so with recovery negotiated and the default `connect_retry_count == 1` a retired-but-live session reached that poll and returned `Ok(Duration::ZERO)`. My earlier gate closed only the recovery-unavailable half; this hole predates it, but the retirement is meaningless without it.

Applied your patch shape: hoisted `known_dead` and changed the poll to `if !known_dead && !self.transport.is_connection_dead()`. A retired session with recovery available now falls through to `is_recovery_possible` and reconnects, or fails `ConnectionClosed` when session state forbids it. `adopt_recovered_session` swaps in a fresh transport, so the flag does not survive a successful reconnect.

This also makes two documented intents actually hold: `retire_after_failed_drain` deliberately leaves recovery armed for transport loss, and `abort_streamed_write` states the caller `will pass is_recovery_possible and reconnect on its next check_and_reconnect`. Both previously depended on the socket happening to poll dead. The ODBC pool-reset path (`txn.rs` `mark_connection_dead` after a failed rollback) now reconnects or fails instead of writing to an unknown-state session; `SQLDisconnect` does not go through this path, so teardown is unaffected.

Pinned by `check_and_reconnect_recovers_a_retired_session_on_a_live_socket`, which asserts `!is_connection_dead()` on the transport before calling — so it fails on the old code by returning `Ok` instead of attempting recovery.

Validation: `check_and_reconnect` tests 7/7, `mssql-tds --lib` 2052 passed (the 7 failures are pre-existing missing certificate fixtures), `mssqlodbc --lib` 1168 passed, ODBC C++ e2e suite passes, `cargo bfmt` and `cargo bclippy` clean.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -1204,22 +1212,38 @@ impl TdsClient {
...
         if connect_retry_count == 0 {
             return Ok(Duration::ZERO);
         }
 
-        // Non-blocking poll — returns immediately.
-        if !self.transport.is_connection_dead() {
+        // Non-blocking poll — returns immediately. Skipped once the verdict is
+        // already in: a session retired by a fatal ERROR token or a failed drain
+        // stays unusable however healthy its socket still looks.
+        if !known_dead && !self.transport.is_connection_dead() {
```
saurabh500: Confirmed resolved. `known_dead` is hoisted and the live-socket poll is now gated behind `!known_dead`, so a retired-but-live session with recovery negotiated and `connect_retry_count == 1` skips this early return and falls through to `is_recovery_possible` — reconnecting, or failing `ConnectionClosed` when session state forbids it. The two gates above already return for the recovery-unavailable and `connect_retry_count == 0` cases, so reaching this line with `known_dead` set implies recovery is available. This is the patch shape I asked for.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -9549,6 +9852,26 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn check_and_reconnect_recovers_a_retired_session_on_a_live_socket() {
```
saurabh500: Good — this pins the exact regression. It asserts `!is_connection_dead()` (socket still open) before calling, so reverting the poll to `if !self.transport.is_connection_dead()` returns `Ok(Duration::ZERO)` and the `unwrap_err()` here panics. I checked the scaffolding it leans on: `TestTransport::is_connection_dead()` returns `closed` (false here) while `connection_known_dead()` returns the flag `mark_known_dead()` sets, and `create_test_client` builds a context with the default `connect_retry_count == 1` — so the recovery path is actually exercised rather than short-circuited. The test guards the change.

shiwanigupta0809:
Re-reviewed at `957cc9aa`. My earlier suggestion is resolved — and better than either option I offered: `3229933c` removed case 4 as a distinct outcome, which made `|| completion_failed` dead code rather than a policy choice, and `8fea826a` drops it and restores the paragraph. I verified that rather than taking it from the reply: all four `Err` returns from `complete_current_result` now leave `has_open_batch == false`, so `batch_done` is unconditionally true for `Some(Err(_))` and the `_` arm computes the same thing.

Also confirming your correction against me: `git diff` for `record_error_token|severity` on this PR is **0 matches**. That verification in my earlier review was wrong and belonged to a different change. Apologies for the noise.

Re-checked the rest at head, including the two commits I had not seen. `1d1b5f8b` is right and the `attention_error && connection_known_dead()` split in `fail_rpc_terminator_look_ahead` reads correctly. One new finding.

### Suggestion

**`settle_rpc_terminator` has two call sites. `3229933c` fixed the wedge at the new one and left the identical shape at the other — three lines above a line this PR edits.**

```
tds_client.rs:4246   advance_to_result_boundary   self.settle_rpc_terminator(&parser_context).await?;
tds_client.rs:4639   complete_current_result      if let Err(error) = self.settle_rpc_terminator(...).await { ...drain, close batch... }
```

The `4244` hunk in this diff adds `self.current_result_ended_with_done_in_proc = false;` on the line immediately below that bare `?`, so this was in view.

The failure mode is the one Copilot described, verbatim. `settle_rpc_terminator`'s park arm returns `processing_error.map_or(Ok(()), Err)` with `parked_token = Some(..)` and `has_open_batch` still set. At `4639` that is now caught and drained. At `4246` the `?` propagates it with the batch open.

I traced where it lands, because the ODBC half differs from the `complete_current_result` case and I did not want to assume it was the same wedge:

- `advance()` maps it straight through — `abort_pending_prepare_capture()`, then `Err(error)`.
- `SQLMoreResults`'s `Err(e)` arm **does** release the claim: `dbc_state.active_stmt = None`, client returned. So no `active_stmt` leak — that part is genuinely different from the case you fixed.
- But the same arm calls `reset_cursor_state(&mut stmt_state)`, which clears `STMT_STATE_CURSOR_OPEN`. `close_cursor.rs` guards all three close paths on that flag (`:66`, `:134`, `:204`), so `SQLCloseCursor`, `SQLFreeStmt(SQL_CLOSE)` and `SQLFreeHandle` all skip `drain_and_release` — exactly the skip you described in your reply to Copilot.
- The client is now back in the DBC with `has_open_batch() == true` and nothing left that will clear it. `command_is_busy()` is `has_open_batch() || streamed write`, and it gates every execute entry point (`:1432`, `:1491`, `:1579`, `:2404`, `:2711`, `:2848`, `:3031`, `:3131`, `:3223`, `:3362`, `:3448`, `:3633`).

So every later command on that *connection* — not just that statement — fails `ALREADY_EXECUTING_ERROR`. That is the masking `drain_stream_or_retire`'s own doc comment calls out as "what made the original failure look like a permanent wedge", and it is already pinned in two directions by `drain_stream_or_retire_failure_closes_batch_and_retires_connection` and `retire_after_failed_drain_disarms_recovery_only_for_desync`, both of which assert `!client.command_is_busy()`.

On reachability: it is the same argument at both sites, which is what makes this worth raising rather than dismissing. Both need a tail token that fails to process (`finalize_return_value` on an AE output parameter, `capture_change_property`, `process_session_state`) followed by something that is not a clean terminal `DONEPROC`, so the loop parks instead of closing the batch. Whatever justified treating that as real at `4639` applies unchanged at `4246`. The difference is only which ODBC call is in flight — `SQLMoreResults` over a no-row `DONEINPROC MORE` result rather than the fetch path.

This is pre-existing rather than introduced here, so I would not block on it. But the invariant this PR establishes is "a tail failure always ends the batch", and it is currently true of one caller instead of the function. Routing `4246` through the same drain — or better, moving the drain-and-close inside `settle_rpc_terminator`'s own error exit so the rule lives with the loop that breaks it — would make it a property of the function and stop the next call site from having to remember.

Worth noting the same reasoning covers a second, quieter consequence at `4246`: the parked token survives the propagated error. `begin_command()` clears `parked_token` at the top of every command, so it cannot desynchronize a later reader — but only because that clear exists, not because this path cleaned up.

The rest holds up on re-read. Settling the tail through the existing `settle_rpc_terminator` rather than a second look-ahead is still the right call, and `begin_command()` clearing `parked_token` and `return_values` is the detail that makes the parking scheme safe against connection reuse.

shiwanigupta0809 (commented):
_Re-reviewed at 957cc9aa - see https://github.com/microsoft/mssql-rs/pull/461#issuecomment-5521353978_

Theekshna:
Rebased onto latest `main` (`43822902`) and addressed this in 14680dde.

Two conflicts, both mechanical: `release_busy_if_row_exhausted`'s peek now goes through #418's sync-first `try_peek_past_current_row` with the async peek as the `Pending` fallback, and upstream's eager `batch_done` binding is dropped because this PR computes it after `complete_current_result`. I checked the interaction rather than assuming it was benign — `try_next_row_cursor` only returns `Ready` for a parsed ROW/NBCROW header or an already-exhausted set, so a DONE still reaches `handle_row_done` on the async path and `current_result_ended_with_done_in_proc` is unaffected. The second conflict was a test-import list.

On the finding: agreed, and I took the version you preferred. The drain-and-close now lives in `settle_rpc_terminator`'s own error exit rather than at one call site, so `4246`'s bare `?` inherits it and a future call site cannot forget. `complete_current_result` loses its local drain block and just propagates.

I confirmed your trace of where `4246` lands rather than taking it from the report — `SQLMoreResults`'s `Err` arm does release `active_stmt`, but `reset_cursor_state` clears `STMT_STATE_CURSOR_OPEN`, and all three `close_cursor.rs` paths gate on that flag, so nothing afterwards clears `has_open_batch` and `command_is_busy()` fails every later command on the connection. That is the masking `drain_stream_or_retire` documents.

Pinned by `advance_abandons_the_batch_when_a_tail_token_fails`, which drives the `4246` site through `advance()` over a message-bearing `DONEINPROC MORE` result and asserts `!command_is_busy()` and an empty `parked_token`. It fails on the previous code, where the park arm returned with both set.

Validation: `mssql-tds --lib` 2104 passed (the 7 failures are pre-existing missing certificate fixtures), `mssqlodbc --lib` 1288 passed, `connection_busy_test` compiles, `cargo bfmt` and `cargo bclippy` clean.

Note the branch was force-pushed: the rebase replaces the `Merge main into parameterQueryBusyRelease` commit. I verified the same seven commits carry over and nothing unique to `957cc9aa` was dropped.

shiwanigupta0809 (approved):

PR #463, opened 2026-09-01
mssql-odbc: convert bound max text columns

David-Engel:
## Summary

Implements [AB#47767](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47767).

Bound `varchar(max)` and native `json` columns can now be delivered to `SQL_C_WCHAR`. The bound PLP path derives the source encoding from column metadata and widens streamed text into UTF-16 while preserving character boundaries and existing truncation and indicator behavior.

Bound text PLP columns can also be delivered to typed C targets through the same conversion path used by non-max values. Typed conversion materializes at most 1 MiB; larger values are drained and deliberately reported as unsupported rather than risking an unbounded allocation.

Binary PLP delivery remains unsupported and is tracked separately by [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239).

## Testing

- `cargo test -p mssqlodbc` (`1151` passed, `1` ignored)
- `cargo clippy -p mssqlodbc --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- `git diff --check`
- Full C++ e2e suite against mssql-odbc and msodbcsql: `36/36` test binaries passed for each driver with `0` parity divergences
- Added coverage for collation-aware widening, truncation indicators, UTF-16 surrogate boundaries, typed conversion success and errors, oversized-value refusal, and PLP stream recovery

None.

### DIFF
--- a/.github/instructions/mssql-odbc.instructions.md
+++ b/.github/instructions/mssql-odbc.instructions.md
@@ -155,17 +155,34 @@ authoritative parity reference for this crate. Its source lives in the
     outstanding 18.6.2.1 measurements land - keep the running record there
     rather than growing this file per build.
   - **A bound `max`/LOB text column converted to a typed C target is refused
-    above 1 MiB.** A typed conversion needs the whole literal, and a
-    `varchar(max)` carries up to 2 GB, so `PLP_TYPED_MATERIALIZE_LIMIT`
-    (`api/fetch_scroll.rs`) caps that one allocation; past the cap the value is
-    drained to keep the row synchronized and answered `HYC00` rather than
-    converted. The cap is this driver's own resource bound with no msodbcsql
-    counterpart, so the compare leg cannot agree by construction and
+    above 1 MiB; msodbcsql converts a truncated prefix and warns.** Both drivers
+    cap what a typed conversion may materialize - a `varchar(max)` carries up to
+    2 GB and the converter needs one contiguous literal. This driver's cap is
+    `PLP_TYPED_MATERIALIZE_LIMIT` (`api/fetch_scroll.rs`) at 1 MiB; past it the
+    value is drained to keep the row synchronized and answered `HYC00`.
+    msodbcsql clamps to `2*CONVBUF_SIZE` (~1244 bytes, sized for the longest
+    legal `double` literal) in `EstimateBytesToRead` (`odbc/sqlcdata.cpp`), then
+    converts that prefix and reports `01004` rather than failing.
+    **Measured on 18.06.0001**, `varchar(max)` bound to a typed target:
+    `'0'`×2000 + `'1'` returns `SQL_SUCCESS_WITH_INFO` with `01004` and a value
+    of **`0`** - the truncated prefix, not the `1` in the column - for both
+    `SQL_C_SBIGINT` and `SQL_C_SLONG`, and the same past 1 MiB; `'x'`×5000
+    returns `22018` (the parse fails before truncation is considered, so both
+    drivers agree there); a short `'42'` returns `42` on both. The skipped
+    case's own payload, `'1'`×1048577 into `SQL_C_SLONG`, overflows even the
+    clamped prefix and returns `22003` there against this driver's `HYC00` - so
+    both drivers error on it, with different states, and
     `AOversizedBoundVarcharMaxTypedConversionIsRefusedAndDrained` carries
-    `SKIP_IF_COMPARING_MSODBCSQL()`. Every literal this path converts - numeric,
-    GUID, datetime - is orders of magnitude below the cap, so no realistic
-    conversion input reaches it; raise or stream it if one ever does. Tracked in
-    AB#47767.
+    `SKIP_IF_COMPARING_MSODBCSQL()` on a measured divergence rather than an
+    assumed one. Refusing is deliberate: `01004` is "string data, right
+    truncated", which application code routinely ignores on a numeric fetch
+    because scalars are not expected to be truncatable, so on the prefix-parses
+    shape msodbcsql's answer is a silently wrong number. Note the cap keys on
+    the column's byte count, not on whether the text parses, so any large
+    `varchar(max)` bound to a typed target reaches it - schema drift, not a
+    contrived input. CI compares against 18.6.2.1; this measurement is
+    18.06.0001, so re-measure there before relying on the exact prefix length.
+    Tracked in AB#47767.
   - **Widening a bound narrow `max` column to `SQL_C_WCHAR` truncates on a whole
     character.** A buffer with no room for the final surrogate pair ends before
     it; msodbcsql leaves the lone high surrogate in the last payload slot on this
--- a/.github/instructions/mssql-odbc.instructions.md
+++ b/.github/instructions/mssql-odbc.instructions.md
@@ -451,6 +451,16 @@ Driver Manager (DM) provides serialization guarantees that the driver relies on
   - A success-path test.
   - A null-output-handle test.
   - An invalid-handle-type or invalid-input test.
+- **A `max`/LOB column only reaches the bound *streaming* path when it is too
+  large for the transport to buffer whole.** `try_read_buffered_column` decodes
+  any fully buffered column - PLP included - into a `ColumnValues`, so a small
+  `varchar(max)`/`varbinary(max)` is delivered by `deliver_bound`, exactly like a
+  non-max value, and never enters `deliver_bound_plp`. A few kilobytes still
+  buffers; the streaming path is reachable around a megabyte (the existing 1 MiB
+  cases in `fetch_scroll_test.cpp` prove it). Size an e2e that targets
+  `deliver_bound_plp` accordingly, or it will silently assert the non-PLP path -
+  a bound binary `max` column against a typed C target answers `22018` from the
+  typed converter when buffered, and `HYC00` when streamed.
 - Use `cargo nextest` (via `cargo btest`), not `cargo test`.
 
 ## Code style
--- a/.github/instructions/mssql-odbc.instructions.md
+++ b/.github/instructions/mssql-odbc.instructions.md
@@ -154,6 +154,28 @@ authoritative parity reference for this crate. Its source lives in the
     `SKIP_IF_COMPARING_MSODBCSQL()`. Tracked in AB#47369, which is where the
     outstanding 18.6.2.1 measurements land - keep the running record there
     rather than growing this file per build.
+  - **A bound `max`/LOB text column converted to a typed C target is refused
+    above 1 MiB.** A typed conversion needs the whole literal, and a
+    `varchar(max)` carries up to 2 GB, so `PLP_TYPED_MATERIALIZE_LIMIT`
+    (`api/fetch_scroll.rs`) caps that one allocation; past the cap the value is
+    drained to keep the row synchronized and answered `HYC00` rather than
+    converted. The cap is this driver's own resource bound with no msodbcsql
+    counterpart, so the compare leg cannot agree by construction and
+    `AOversizedBoundVarcharMaxTypedConversionIsRefusedAndDrained` carries
+    `SKIP_IF_COMPARING_MSODBCSQL()`. Every literal this path converts - numeric,
+    GUID, datetime - is orders of magnitude below the cap, so no realistic
+    conversion input reaches it; raise or stream it if one ever does. Tracked in
+    AB#47767.
+  - **Widening a bound narrow `max` column to `SQL_C_WCHAR` truncates on a whole
+    character.** A buffer with no room for the final surrogate pair ends before
+    it; msodbcsql leaves the lone high surrogate in the last payload slot on this
+    narrow-source widening path, though its wide-source path is surrogate-safe
+    (`GetColDataSurrogateSafe`, and `TrimPartialCodePt` for partial sequences).
+    This driver's existing bound `nvarchar(max)` delivery already trims to a
+    character boundary, and handing back text that does not decode from one `max`
+    type but not the other would be worse than the divergence.
+    `ABoundUtf8VarcharMaxDoesNotSplitASurrogatePairWhenWidening` carries
+    `SKIP_IF_COMPARING_MSODBCSQL()`. Tracked in AB#47767.
 
 ## No panics
 
--- a/mssql-odbc/src/api/fetch_scroll.rs
+++ b/mssql-odbc/src/api/fetch_scroll.rs
@@ -29,7 +29,7 @@ use mssql_tds::datatypes::column_values::{
 use mssql_tds::datatypes::decoder::DecimalParts;
 use mssql_tds::datatypes::row_writer::RowWriter;
 use mssql_tds::datatypes::sql_json::SqlJson;
-use mssql_tds::datatypes::sql_string::{EncodingType, SqlString};
+use mssql_tds::datatypes::sql_string::{EncodingType, SqlString, get_encoding_type};
 use mssql_tds::datatypes::sql_vector::SqlVector;
 use mssql_tds::datatypes::sqldatatypes::TdsDataType;
 use mssql_tds::error::Error as TdsError;
@@ -40,6 +40,7 @@ use super::sqlstate::*;
 use crate::api::exec_common::release_busy_if_row_exhausted;
 use crate::api::get_data::{
     TextError, column_value_to_text, convert_typed_c, is_typed_c_target, utf16le_chunk_to_utf8,
+    widen_into_pending,
 };
 use crate::api::odbc_types::{
     SQL_BIND_BY_COLUMN, SQL_C_BIT, SQL_C_CHAR, SQL_C_DOUBLE, SQL_C_FLOAT, SQL_C_GUID,
@@ -62,6 +63,12 @@ use crate::handles::stmt::{
     ColumnBinding, STMT_STATE_CURSOR_OPEN, STMT_STATE_FETCH_IN_PROGRESS, StmtState,
 };
 use crate::handles::{HandleType, StmtHandle, handle_from_raw};
+
+#[derive(Clone, Copy)]
+struct PlpColumnInfo {
+    wire_encoding: PlpEncoding,
+    text_encoding: Option<EncodingType>,
+}
 
 /// Implements SQLFetchScroll for the current forward-only result set.
 ///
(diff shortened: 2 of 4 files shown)


mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -1054,6 +1073,18 @@ unsafe fn read_bind_offset(ptr: *mut SqlULen) -> usize {
 /// value itself is never materialized, only one chunk at a time.
 const PLP_BOUND_CHUNK: usize = 8 * 1024;
 
+/// Typed conversion needs the complete text literal. Numeric, GUID, and
+/// datetime literals are tiny in practice, so cap that exceptional allocation
+/// rather than letting an arbitrary MAX value exhaust the application process.
+const PLP_TYPED_MATERIALIZE_LIMIT: usize = 1024 * 1024;
```
Theekshna, line 1079: [Suggestion] This bounded typed-conversion path introduces two deliberate divergences from the reference driver: values over `PLP_TYPED_MATERIALIZE_LIMIT` are drained and refused with `HYC00` rather than converted, and the widen-to-WCHAR path applies whole-character truncation where the reference driver would leave a lone high surrogate. Both are guarded by `SKIP_IF_COMPARING_MSODBCSQL()`, the same pattern used elsewhere for intentional deviations. `.github/instructions/mssql-odbc.instructions.md` expects intentional deviations to be recorded in its registry or a design doc, but neither is present in-repo; the new skip-tests also carry no AB# reference unlike neighboring cases. The rationale may already live in AB#47767, which this unattended run could not verify. Please add a registry entry or tracking-work-item pointer for the 1 MiB-to-`HYC00` refusal, plus a brief note for the surrogate rule referencing the existing `nvarchar(max)` truncation precedent.

Theekshna (commented):
Unattended review: the bound max-text conversion paths are correctly bounded, drain rejected streams to preserve synchronization, and handle text widening without splitting surrogate pairs. I found two non-blocking review items.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 2 |
| Nit | 0 |

Covered PLP streaming and NULL framing, typed conversion limits, row synchronization, widening boundaries, msodbcsql parity, deliberate-divergence documentation, unit/e2e test sufficiency, CI/checklist consistency, linked-work-item presence, and PR-description currency.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -1108,9 +1139,47 @@ unsafe fn deliver_bound_plp(
...
     scratch: &mut [u8],
 ) -> Result<RowOutcome, TdsError> {
+    let Some(column_info) = column_info else {
+        drain_plp_to_end(client, runtime, scratch)?;
+        return Ok(RowOutcome::Error(RowIssue::Unsupported));
+    };
+
+    if is_typed_c_target(binding.target_type) {
+        let Some(text_encoding) = column_info.text_encoding else {
+            drain_plp_to_end(client, runtime, scratch)?;
```
Theekshna, line 1163: [Suggestion] Question rather than a confirmed gap: this branch handles a binary PLP column such as `varbinary(max)` or `image` bound to a typed C target by draining it and returning `Unsupported`. The CI coverage report lists these new-side lines as uncovered, so the path appears untested. I did not run the required mutation check because reproducing it needs the live-server C++ e2e suite and would exceed this unattended run's remaining budget. The risk is low because this matches the other binary paths and preserves row synchronization. Would it be worth adding a one-row e2e that binds `varbinary(max)` to `SQL_C_SLONG`, asserts the error state, then confirms the following row still reads cleanly?

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Good catch — both deviations are now recorded in the registry in `.github/instructions/mssql-odbc.instructions.md`, and both skip-tests carry an AB#47767 pointer like the neighboring cases.

The 1 MiB entry frames the refusal as this driver's own resource bound with no msodbcsql counterpart, so the compare leg can't agree by construction; I deliberately didn't assert a measured msodbcsql result for an oversized value since I have no measurement to cite. The surrogate entry notes that msodbcsql leaves the lone high surrogate on this narrow-source widening path while its wide-source path is surrogate-safe (`GetColDataSurrogateSafe`/`TrimPartialCodePt`), and points at the existing `nvarchar(max)` bound-delivery precedent as the reason for choosing the whole-character rule.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Worth adding, yes — done. `ABoundVarbinaryMaxToTypedCTargetIsStillUnsupported` binds `varbinary(max)` to `SQL_C_SLONG` across two rows with a trailing scalar, asserts `HYC00`, and checks that both the column after the refused LOB and the following row still decode correctly.

It sits next to `ABoundVarbinaryMaxIsStillUnsupported`, which covers the `SQL_C_BINARY` route into the same refusal via the `!compatible` branch; this one reaches it through `is_typed_c_target` with no source encoding. It carries `SKIP_IF_COMPARING_MSODBCSQL()` for the same already-recorded AB#47239 reason as its neighbor — msodbcsql delivers bound binary, we don't yet.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -1108,9 +1139,47 @@ unsafe fn deliver_bound_plp(
     binding: &ColumnBinding,
     row_index: usize,
     bind_offset: usize,
-    encoding: Option<PlpEncoding>,
+    column_info: Option<PlpColumnInfo>,
     scratch: &mut [u8],
 ) -> Result<RowOutcome, TdsError> {
+    let Some(column_info) = column_info else {
```
Theekshna, line 1156: [Suggestion] This column_info: None arm is a second branch this PR's own diff introduces (confirmed via git diff against merge-base — it's new, not pre-existing) and the last coverage report (build 171672, at the previously-reviewed head) listed these exact lines, 1146-1147, as uncovered alongside the now-fixed 1152-1153. deliver_bound_plp only reaches here when the cursor layer already classified the column as CursorColumn::PlpStreaming, while column_info comes from metadata.plp_encoding(), which returns None only when ColumnMetadata::is_plp() is false (mssql-tds/src/query/metadata.rs:95-98) — every PLP TdsDataType variant, including Vector, falls through to PlpEncoding::Binary via the catch-all, not None. So this arm only fires if the cursor's PLP classification and the column metadata's is_plp() can disagree for some type. I could not determine from reading alone whether that's reachable with a real server response (e.g. a type where PLP-ness is only obvious from the wire framing, not TYPE_INFO) or whether it's now dead defensive code that should say so in a comment. Could you confirm which, and if it's reachable, is it worth a similar sized e2e case the way ABoundStreamedVarbinaryMaxToTypedCTargetIsStillUnsupported was added for the sibling branch?

Theekshna (commented):
Unattended re-review after this run's own prior findings (at `c88fe3c9`) were addressed.

**Both prior suggestions are genuinely fixed, not just replied to.** `.github/instructions/mssql-odbc.instructions.md` now records both deliberate deviations (the 1 MiB typed-conversion cap and the whole-character-truncation-on-widening rule) with AB#47767 references, matching the registry format used by neighboring entries. The binary-PLP-to-typed-C-target coverage gap I flagged at `fetch_scroll.rs:1152` is closed by a new test, sized past the ~1 MiB buffered-column threshold specifically so it exercises the streaming path rather than silently falling back to the non-PLP route (per the accompanying registry note) — confirmed no change to `fetch_scroll.rs`/`get_data.rs` themselves between the two reviewed heads, so this re-review only needed to verify the new test and docs, not the implementation again.

**One remaining coverage gap this run did not previously catch:** see the inline comment on `fetch_scroll.rs:1145`.

**PR description:** the checklist section required by `.github/PULL_REQUEST_TEMPLATE.md` (`cargo bfmt`/`cargo bclippy`/`cargo btest` checkboxes, public-API-documented checkbox) is absent; the Testing section lists plain `cargo test`/`cargo clippy`/`cargo fmt` instead of the repo's `b`-prefixed aliases. Nit, not blocking — the linked AB#47767 work item and the described testing are both present and match the diff.

**CI:** `mssql-rs Pull request validation` and `coverage-report` are red at this head, but the underlying failure (`Install and start Colima-based Docker` on the macOS stage) is `brew`/Homebrew tap-import breakage unrelated to this diff — confirmed by reading the Azure DevOps log directly, not re-run. Every other leg, including both Linux/Windows/ARM build-and-test legs and the mssql-python cross-repo suite, is green. Flagging for a human to decide whether a rebuild is warranted; this run does not trigger one.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 1 |

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Good catch that this is a new branch, and thanks for doing the metadata trace — your reading is right, and I chased it the rest of the way: it's **dead defensive code**, not a reachable gap, so I've documented it in `e6c44f7` rather than adding an e2e that could never execute the branch.

The two classifications can't disagree because they're the same predicate over the same data. The cursor emits `PlpStreaming` only under `meta.is_plp()` (`token_stream.rs:578`), and `plp_encoding()` returns `None` only under `!is_plp()` — and `is_plp()` is purely `type_info_variant == PartialLen(..)`, not a per-type list, so a new PLP type can't slip between them (it lands on `Binary` via the catch-all, as you noted for `Vector`). Both read the same COLMETADATA: `plp_columns` snapshots `stmt.column_metadata`, populated from the client's metadata via `begin_batch`/`begin_result_set`.

The other route in — an out-of-range `.get()` — is closed too: `column_count` is `column_metadata.len()`, `plp_columns` is built from that same vector, and bindings outside `1..=column_count` are skipped before the call, so `column - 1` is always in range.

I kept it as a drain-and-refuse rather than tightening it to `unreachable!()`. If a future type ever makes PLP-ness visible only in the wire framing and not in `TYPE_INFO`, draining keeps the row synchronized, whereas a panic would take the application down over a column we could simply refuse. The comment now says all of this so the next reader doesn't have to re-derive it.

> [!NOTE]

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -154,6 +154,28 @@ authoritative parity reference for this crate. Its source lives in the
     `SKIP_IF_COMPARING_MSODBCSQL()`. Tracked in AB#47369, which is where the
     outstanding 18.6.2.1 measurements land - keep the running record there
     rather than growing this file per build.
+  - **A bound `max`/LOB text column converted to a typed C target is refused
```
Theekshna, line 157: Traced the msodbcsql side properly — I think this entry is understating itself, and one claim in it is wrong.

**There is a counterpart, and it's ~850x smaller.**

> The cap is this driver's own resource bound with no msodbcsql counterpart

`EstimateBytesToRead` (`odbc/sqlcdata.cpp:1045`) clamps every typed-target read:

```cpp
// converting to others fixed types
default:
    cbEstPreConvSize = cbColLen;                          // wants the whole column
    break;
}
estimatedBytesToRead = min(cbEstPreConvSize, cbPerConv);  // ...clamped

SIZE_T cbPerConv = 2*CONVBUF_SIZE;
#define CONVBSIZE ((DBL_MAX_10_EXP+3)*sizeof(WCHAR))      // (308+3)*2 = 622
```

**1244 bytes**, backed by `rgbConvBuffer[CONVBUF_SIZE*3]` on the statement — sized for the longest legal `double` literal. So both drivers cap what a typed conversion may materialize. Ours is 1 MiB.

**What it does past the cap: converts the prefix and warns.** `GetColData` (`odbc/sqlcdata.h:1137`), right after the conversion call:

```cpp
// If requested C type is fixed type and too much data, force flush
if (OutputTypePolicy::GetTypeCode() == FixedLengthType && cbDataAvail > cbRead)
{
    if (!wError)
        wError = IDS_01_004;   // string data, right-truncated
    fFlush = TRUE;
    goto Return2;              // leaves the loop after one iteration
}
```

One pass, ≤1244 bytes converted, remainder flushed. Two consequences:

- 5 MB of `'x'` → `CharToBigint` fails on the first character, so `wError` is already set and the `if (!wError)` guard leaves it alone → `22018`. Same as our buffered path, no divergence.
- `'0'` * 2000 + `'1'` → the 1244-byte prefix parses cleanly as `0`, nothing sets `wError`, so `01004` fires → the application gets **`n == 0` and `SQL_SUCCESS_WITH_INFO`** for a column containing `1`.

That second one is why I'd keep our behaviour and say so more confidently. `01004` is "string data, right truncated" — a warning most code ignores on a numeric fetch, because scalars aren't supposed to be truncatable. Refusing beats handing back a number that isn't in the column.

**Which makes the entry stronger than it reads now.** Something like:

> Both drivers cap the buffer a typed conversion may materialize — msodbcsql at `2*CONVBUF_SIZE` ≈ 1244 bytes (`EstimateBytesToRead`, `sqlcdata.cpp:1212`), this driver at 1 MiB. Past its cap msodbcsql converts the truncated prefix and reports `01004` (`sqlcdata.h:1137`), so a `varchar(max)` holding `'0'×2000 + '1'` fetched into `SQL_C_SBIGINT` yields `0` under a warning. This driver drains and answers `HYC00` rather than deliver a value that is not in the column.

That's a deliberate choice with a cited msodbcsql behaviour next to it, rather than a limit we appear to have invented.

**Caveat: this is a source trace, not a measurement.** Worth running `'0'` * 2000 + `'1'` into `SQL_C_SBIGINT` on 18.6.2.1 before the wording goes in — that single input confirms or kills the whole reading. I also matched `OutputTypePolicy::GetTypeCode() == FixedLengthType` by name rather than by checking what a `SQL_C_SBIGINT` binding instantiates, so that's the other thing the measurement would settle.

**Smaller thing:**

> no realistic conversion input reaches it

The cap keys on the column's size (`chunk.known_total`), not on whether the text parses, so any large `varchar(max)` bound to a typed target reaches it — schema drift rather than a contrived input. Your note further down the diff shows the cost: same mistake gives `22018` buffered and `HYC00` streamed.

David-Engel: You were right on both counts, and thanks for pushing on it — I ran the measurement you asked for rather than take the source trace on faith, and it confirms your reading and turns up one more wrinkle. Rewritten in `fc084eb3`.

**Measured on `SQL_DRIVER_VER=18.06.0001`**, `varchar(max)` bound via `SQLBindCol` + `SQLFetch`:

| Input | C type | msodbcsql result |
|---|---|---|
| `'0'`×2000 + `'1'` | `SQL_C_SBIGINT` | `SQL_SUCCESS_WITH_INFO`, **value `0`**, `01004` |
| `'0'`×2000 + `'1'` | `SQL_C_SLONG` | `SQL_SUCCESS_WITH_INFO`, **value `0`**, `01004` |
| `'0'`×1100000 + `'1'` | `SQL_C_SBIGINT` | `SQL_SUCCESS_WITH_INFO`, **value `0`**, `01004` |
| `'x'`×5000 | `SQL_C_SBIGINT` | `SQL_ERROR`, `22018` |
| `'42'` | `SQL_C_SBIGINT` | `SQL_SUCCESS`, `42` |

So the prefix really does get converted and warned about — a column holding `1` hands back `0` — and your `if (!wError)` reasoning for the `22018` case holds. That also settles the `OutputTypePolicy` caveat you flagged: whatever `SQL_C_SBIGINT`/`SQL_C_SLONG` instantiate, the observable behaviour is the clamp-and-warn one.

**The wrinkle:** the skipped test's own payload behaves differently from the shape we were both reasoning about. `'1'`×1048577 into `SQL_C_SLONG` overflows even the clamped prefix, so msodbcsql returns `22003`, not `01004`. Both drivers error on that input — just with different states (`22003` vs our `HYC00`). The skip is still correct, and now rests on a measured divergence rather than the "no counterpart, cannot agree by construction" claim I had written, which was simply wrong.

I've folded all of this into the entry, including your `'0'×2000` example as the case that shows *why* refusing is the better answer, and dropped "no realistic conversion input reaches it" — you're right that the cap keys on `chunk.known_total`, so it's schema drift, not a contrived input.

Two things I deliberately left as-is: I cited `EstimateBytesToRead` by function rather than line, since your two references to it differed (`:1045` and `:1212`) and I have no msodbcsql source to arbitrate; and the entry names **18.06.0001** explicitly with a note to re-measure on 18.6.2.1 before anyone leans on the exact prefix length, per the "a parity claim must name the build" rule. I didn't add an e2e for the `'0'×N + '1'` shape — it exercises the same refusal the existing test already covers, and the input shape that matters is now recorded in the registry.

> [!NOTE]

Theekshna (commented):
Unattended re-review (no human has checked this): re-verified that both suggestions from the prior pass (findings=2, `f7bc422a`) are genuinely fixed, not just replied to.

- **`fetch_scroll.rs:1156`** (the `column_info: None` reachability question) - the new comment traces why the arm is unreachable today: `plp_encoding()` and the cursor's `PlpStreaming` classification both resolve off the same `ColumnMetadata::is_plp()` over the same COLMETADATA, and `plp_encoding`'s catch-all maps every PLP type to `Some`. Draining on the guard rather than `unreachable!()` is the right call for a possible future PLP type that only reveals PLP-ness on the wire. No code changed here, only the comment, so this needed the reasoning re-checked against the (unchanged) code, which I did.
- **Registry entry for the 1 MiB typed-conversion cap** - now cites a real msodbcsql counterpart instead of claiming none exists. I traced this independently against the reference tree: `Sql/Ntdbms/sqlncli/odbc/sqlcdata.cpp:1045` (`EstimateBytesToRead`) and `sqlsrv.h:289-290` (`CONVBUF_SIZE = (DBL_MAX_10_EXP+3)*sizeof(WCHAR)` = 622 bytes, so `2*CONVBUF_SIZE` = 1244 bytes as the entry states), plus the truncate-then-convert-then-warn (`IDS_01_004`) flush path in `sqlcdata.h`'s `GetColData`. The entry's shape - clamp, convert the prefix, warn - matches what's actually in those files. I did not re-run the entry's own cited 18.06.0001 measurement; the entry already flags that as unverified against the CI-pinned 18.6.2.1, which is the correct caveat to carry forward rather than resolve here.

Nothing else changed in this diff since the last reviewed head (`f7bc422a`) - `git diff f7bc422a..fc084eb3` touches only these two spots (a doc comment in `fetch_scroll.rs` and the registry entry in `mssql-odbc.instructions.md`).

**msodbcsql parity:** the corrected registry entry checks out against the reference tree (above); no other ODBC-surface change since the last pass.
**Tests:** unchanged since the last pass; these two commits are docs-only, no new gap introduced.
**Divergences documented:** yes, and now more accurately than the prior wording.
**PR description:** unchanged and still matches the diff; no currency issue.

**CI:** this push is recent and most legs are still pending; `Build Stage Build MacOS`, both `mssql-python` cross-repo legs partially, Kerberos, CodeQL, and `check` are green so far, nothing red yet. Not re-running anything - a human should confirm the remaining legs before merge.

No findings.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

Theekshna (approved):

PR #464, opened 2026-09-01
Surface pre-row server errors during ODBC execution

David-Engel:
## Description

SQL Server can send result metadata before a blocked query produces its first row or server error. The ODBC execution path previously returned success as soon as it received that metadata, leaving a later lock-timeout error for `SQLFetch`. In mssql-python, that caused the expected execution exception to be missed and cleanup to block behind the still-open transaction.

This change waits for and parks the first result row before reporting execution success. The row remains available to `SQLFetch`, while server errors arriving after metadata now surface from `SQLExecDirect` or `SQLExecute`.

Changed files:

- `mssql-odbc/src/api/exec_common.rs` — peek and park the first row in `finish_execute`; unit regressions.
- `mssql-tds/src/connection/tds_client.rs` — doc-comment only, rewriting the `peek_past_current_row` contract now that execution calls it before the first row is positioned.
- `mssql-odbc/tests/e2e/tests/transaction_test.cpp` — the two live regressions below.

### Observable behaviour change beyond errors

The peek stops at rows and errors, but *captures and continues* past INFO tokens (`handle_row_read_token`'s `Tokens::Info` arm). So a server informational message arriving between `COLMETADATA` and the first `ROW`/`DONE` is now drained during execution instead of by the first `SQLFetch`, and the success tail posts it — meaning `SQLExecDirect`/`SQLExecute` can now return `SQL_SUCCESS_WITH_INFO` where it previously returned `SQL_SUCCESS`.

Because ODBC clears a handle's diagnostics at the start of the next call on it, an application that only inspects `SQLGetDiagRec` *after* fetching will no longer see such a warning at all. This is a parity improvement rather than a divergence — msodbcsql's parse loop covers the same window before it parks at the first row, so it reports these from execute too — but it is a real change for queries that were already succeeding, so it is called out here rather than left implicit.

## Tests

Unit regressions in `exec_common.rs` cover both halves of the window, each verified by mutation:

- `finish_execute_surfaces_an_error_before_the_first_row` — error 1222 after metadata; fails if the peek is removed.
- `finish_execute_reports_an_info_message_arriving_before_the_first_row` — msg 8153 after metadata returns `SQL_SUCCESS_WITH_INFO` with the record posted; fails (`SQL_SUCCESS`, no record) if the peek is moved below `take_info_messages()`, which is the one edit that would silently restore the old behaviour.

Two live two-connection regressions hold an update lock, set `LOCK_TIMEOUT 1000`, and verify execution returns native error 1222 and SQLSTATE `42000`:

- `BlockedSelectTest.ExecDirectReturnsServerLockTimeout` — the `SQLExecDirect` path.
- `BlockedSelectTest.PreparedExecuteReturnsServerLockTimeout` — the `SQLPrepare` + `SQLExecute` path, which is what `mssql-python`'s `cursor.execute()` uses because it defaults to `use_prepare=True`.

Both share one body and both were verified to fail without the `finish_execute` change and pass with it (driver `.so` hash-checked between runs, since the build copies the cdylib to a second filename and a stale copy silently passes). Both also pass unchanged against msodbcsql 18.6, so the parity contract this file carries is satisfied.

## Related Issues

[AB#47771](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47771): https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47771

[AB#47510](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47510) reaches the same defect from the `mssql-python` `test_003_connection` hang.

## Dependency

#442 (`SQL_ATTR_QUERY_TIMEOUT` enforcement) has merged and is merged into this branch. That makes the query-timeout attribute these tests set a real deadline: `next_row_cursor` passes `remaining_request_timeout` into `receive_row_header`, and `peek_past_current_row` delegates to `next_row_cursor`.

## Testing

- `cargo bfmt`
- `cargo bclippy`
- `cargo nextest run -p mssqlodbc` (1,180 passed, 1 skipped)
- ODBC live e2e suite against SQL Server 2025: 36/36 test binaries, including both new cases
- Both new e2e cases also run green against msodbcsql 18.6 on the same server

### Not yet verified

End-to-end `mssql-python` validation. An earlier revision of this description claimed `tests/test_003_connection.py::test_set_attr_txn_isolation_effect` passed against this branch; that could not be reproduced by a second reviewer, so the claim is withdrawn.

Worth stating precisely, because the obvious counter-evidence does not hold: the green `mssql-python suite on mssql-odbc driver (cross-repo)` check does **not** establish that this test passes. That leg runs every `tests/test_*.py` (`run-mssql-python-odbc-tests.sh:105`), but it deliberately downgrades test failures, crashes, timeouts and skips to a warning — `failTaskOnFailedTests: false` plus an explicit "partially succeeded" result, documented as "expected while the driver is in development". So it stays green even when files fail, and cannot be read as validating any individual test.

## Deferred / to file

Raised in review, out of scope here, recorded so none of it is implicit:

- **`SQLMoreResults` pre-row-error gap** — `more_results.rs:158` returns `SQL_SUCCESS` as soon as `advance()` reports metadata, with no peek, so for `SELECT 1; SELECT <blocked>` the same defect reappears one call later. msodbcsql closes it via the same parse loop (`sqlcresl.cpp:355`), so this is a reference-driver divergence. Hang off [AB#47771](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47771).
- **Missed early busy release on a zero-row result set** — the peek now makes `!has_open_batch()` knowable at execute time, but the success tail clears the exhaustion flags and keeps the claim. Not a regression (behaviour matches pre-PR, and the zero-row path is covered green by `ExecDirectLiveTest.EmptyResultSet`, `FetchScrollLiveTest.ReturnsNoDataAtEndOfResultSet` and two `connection_busy_test` cases), but a newly-available signal being discarded. Belongs under [AB#47508](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47508) with its own e2e case, since it changes busy-gate timing.
- **Stale-handle defect on [AB#47510](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47510)** — after a server error mid-result-set the statement handle can return stale empty rows; a fresh handle on the same connection is fine. Concrete mechanism from review: msodbcsql calls `CleanUphStmt(lpstmt, FREE_MEMORY)` when `retcode == SQL_ERROR` (`sqlccmd.cpp:6896-6905`), whereas `fail_with_tds` resets no statement state.

## Breaking Changes / Migration Notes

No API change. One observable behaviour change, described under "Observable behaviour change beyond errors" above: a post-metadata server warning is now reported from execute (`SQL_SUCCESS_WITH_INFO`) rather than from the first `SQLFetch`, which matches msodbcsql.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes
- [x] Public API changes are documented (no public API changes)

### DIFF
--- a/mssql-odbc/src/api/exec_common.rs
+++ b/mssql-odbc/src/api/exec_common.rs
@@ -637,7 +637,9 @@ pub(super) fn finish_execute(
         return fail_with_tds(dbc, stmt, statement_handle, client, &e);
     }
 
-    // Result-bearing query: leave the cursor open for SQLFetch.
+    // Result-bearing query: leave the cursor open for SQLFetch. This must stay
+    // below the peek: the peek drains any INFO token in the post-metadata
+    // window, and taking the messages first would leave them for a later fetch.
     let info_messages = client.take_info_messages();
     let Ok(mut stmt_state) = stmt.inner.lock() else {
         error!("{op}: stmt mutex poisoned");
@@ -839,6 +841,51 @@ mod tests {
                 .diag_records
                 .iter()
                 .any(|record| record.native_error == 1222)
+        );
+    }
+
+    /// The peek does not stop at rows and errors: `handle_row_read_token`'s
+    /// `Tokens::Info` arm captures and continues, so an INFO token sitting
+    /// between `COLMETADATA` and the first `ROW`/`DONE` is now drained during
+    /// execution rather than by the first `SQLFetch`. The success tail posts it,
+    /// which turns `SQL_SUCCESS` into `SQL_SUCCESS_WITH_INFO` for that window.
+    ///
+    /// This matches msodbcsql, whose parse loop covers the same window before
+    /// parking at the first row. It is pinned here because the behaviour is
+    /// otherwise invisible: moving the peek below `take_info_messages()` would
+    /// silently restore execute-time `SQL_SUCCESS` and defer the diagnostic
+    /// back to `SQLFetch`, and no other test would fail.
+    #[test]
+    fn finish_execute_reports_an_info_message_arriving_before_the_first_row() {
+        use mssql_tds::test_client_support::info;
+
+        let h = TestHandles::with_env_dbc_stmt();
+        h.mark_dbc_connected();
+        let dbc = unsafe { handle_from_raw::<DbcHandle>(h.dbc) };
+        let stmt = unsafe { handle_from_raw::<StmtHandle>(h.stmt) };
+        let mut client = tds_client_from_tokens(vec![
+            col_metadata(int_columns(1)),
+            info(
+                8153,
+                10,
+                "Null value is eliminated by an aggregate or other SET operation.",
+            ),
+            done_no_more(),
+        ]);
+        dbc.runtime
+            .block_on(client.execute("SELECT SUM(col) FROM t".to_string(), ()))
+            .unwrap();
+
+        let rc = finish_execute(dbc, stmt, h.stmt, client, "SQLExecDirectW");
+
+        assert_eq!(rc, SQL_SUCCESS_WITH_INFO);
+        let stmt_state = stmt.inner.lock().unwrap();
+        assert!(
+            stmt_state
+                .diag_records
+                .iter()
+                .any(|record| record.native_error == 8153),
+            "the warning must be posted under the execute that drained it"
         );
     }
 
--- a/mssql-odbc/src/api/exec_common.rs
+++ b/mssql-odbc/src/api/exec_common.rs
@@ -577,6 +577,15 @@ pub(super) fn finish_execute(
         };
     }
 
+    // SQL Server can send COLMETADATA before the statement has produced its
+    // first row. Wait for that row, end-of-set, or an ERROR token so execution
+    // errors surface from SQLExecDirect/SQLExecute instead of a later SQLFetch.
+    // A row is only positioned and parked; SQLFetch still receives it normally.
+    if let Err(e) = dbc.runtime.block_on(client.peek_past_current_row()) {
+        error!(%e, "{op}: failed before the first result row");
+        return fail_with_tds(dbc, stmt, statement_handle, client, &e);
+    }
+
     // Result-bearing query: leave the cursor open for SQLFetch.
     let info_messages = client.take_info_messages();
     let Ok(mut stmt_state) = stmt.inner.lock() else {
@@ -611,7 +620,8 @@ mod tests {
     use crate::params::BoundParam;
     use crate::test_support::TestHandles;
     use mssql_tds::test_client_support::{
-        ScriptedToken, col_metadata_empty, done_more, done_no_more, tds_client_from_tokens,
+        ScriptedToken, col_metadata, col_metadata_empty, done_more, done_no_more, int_columns,
+        sql_error, tds_client_from_tokens,
     };
     use std::ffi::c_void;
 
@@ -650,6 +660,33 @@ mod tests {
         assert!(try_claim_idle_client(dbc, h.dbc).is_none());
         // The existing claim must be left untouched.
         assert_eq!(dbc.inner.lock().unwrap().active_stmt, Some(other));
+    }
+
+    #[test]
+    fn finish_execute_surfaces_an_error_before_the_first_row() {
+        let h = TestHandles::with_env_dbc_stmt();
+        h.mark_dbc_connected();
+        let dbc = unsafe { handle_from_raw::<DbcHandle>(h.dbc) };
+        let stmt = unsafe { handle_from_raw::<StmtHandle>(h.stmt) };
+        let mut client = tds_client_from_tokens(vec![
+            col_metadata(int_columns(1)),
+            sql_error(1222, 16, "Lock request time out period exceeded."),
+            done_no_more(),
+        ]);
+        dbc.runtime
+            .block_on(client.execute("SELECT blocked".to_string(), ()))
+            .unwrap();
+
+        let rc = finish_execute(dbc, stmt, h.stmt, client, "SQLExecDirectW");
+
+        assert_eq!(rc, SQL_ERROR);
+        let stmt_state = stmt.inner.lock().unwrap();
+        assert!(
+            stmt_state
+                .diag_records
+                .iter()
+                .any(|record| record.native_error == 1222)
+        );
     }
 
     /// Builds a scripted client positioned on a row-returning result (empty
--- a/mssql-odbc/tests/e2e/tests/transaction_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/transaction_test.cpp
@@ -558,6 +558,29 @@ TEST_F(TransactionLiveTest, IsolationIsAppliedOnTheServer) {
 // path can prove neither branch regresses.
 class BlockedSelectTest : public TransactionLiveTest {
   protected:
+    // Frees the second connection's handles even when an ASSERT_* returns early
+    // from the helper below. The fixture's TearDown covers dbc_ only, so a
+    // reader left connected would trip HY010 when the environment handle is
+    // freed — turning one assertion failure into a confusing second one.
+    class ScopedConnection {
+      public:
+        ScopedConnection() = default;
+        ~ScopedConnection() {
+            if (stmt != SQL_NULL_HSTMT) {
+                SQLFreeHandle(SQL_HANDLE_STMT, stmt);
+            }
+            if (dbc != SQL_NULL_HDBC) {
+                SQLDisconnect(dbc);
+                SQLFreeHandle(SQL_HANDLE_DBC, dbc);
+            }
+        }
+        ScopedConnection(const ScopedConnection&) = delete;
+        ScopedConnection& operator=(const ScopedConnection&) = delete;
+
+        SQLHDBC dbc = SQL_NULL_HDBC;
+        SQLHSTMT stmt = SQL_NULL_HSTMT;
+    };
+
     // Holds an exclusive row lock on |table| from dbc_, then runs a blocked
     // SELECT on a second connection via |execute_blocked| and asserts it
     // reports native error 1222 rather than hanging or succeeding.
@@ -571,20 +594,21 @@ class BlockedSelectTest : public TransactionLiveTest {
         ASSERT_SQL_OK(SetAutocommit(dbc_, SQL_AUTOCOMMIT_OFF), SQL_HANDLE_DBC, dbc_);
         Exec("UPDATE " + table + " SET value = 1 WHERE id = 1");
 
-        SQLHDBC reader = SQL_NULL_HDBC;
-        ASSERT_SQL_OK(SQLAllocHandle(SQL_HANDLE_DBC, env_, &reader), SQL_HANDLE_ENV, env_);
+        ScopedConnection reader;
+        ASSERT_SQL_OK(SQLAllocHandle(SQL_HANDLE_DBC, env_, &reader.dbc), SQL_HANDLE_ENV, env_);
         SqlTString connstr = ODBCTestUtils::BuildConnectionString();
         SQLTCHAR out_str[1024] = {};
         SQLSMALLINT out_len = 0;
-        ASSERT_SQL_OK(SQLDriverConnect(reader, nullptr, const_cast<SQLTCHAR*>(connstr.c_str()),
+        ASSERT_SQL_OK(SQLDriverConnect(reader.dbc, nullptr,
+                                       const_cast<SQLTCHAR*>(connstr.c_str()),
                                        static_cast<SQLSMALLINT>(connstr.size()), out_str,
                                        static_cast<SQLSMALLINT>(std::size(out_str)), &out_len,
                                        SQL_DRIVER_NOPROMPT),
-                      SQL_HANDLE_DBC, reader);
+                      SQL_HANDLE_DBC, reader.dbc);
 
-        SQLHSTMT reader_stmt = SQL_NULL_HSTMT;
-        ASSERT_SQL_OK(SQLAllocHandle(SQL_HANDLE_STMT, reader, &reader_stmt), SQL_HANDLE_DBC,
-                      reader);
+        ASSERT_SQL_OK(SQLAllocHandle(SQL_HANDLE_STMT, reader.dbc, &reader.stmt), SQL_HANDLE_DBC,
+                      reader.dbc);
+        SQLHSTMT reader_stmt = reader.stmt;
         // Bounds the wait if execution regresses to waiting for a token the
         // server will never send. This is a real deadline only once the
         // query-timeout enforcement of AB#46385 (#442) is in; this PR merges
@@ -600,20 +624,22 @@ class BlockedSelectTest : public TransactionLiveTest {
 
         EXPECT_EQ(SQL_ERROR, rc) << "the lock timeout must surface from execution, not a later "
                                     "SQLFetch";
-        SQLTCHAR state[6] = {};
         SQLINTEGER native = 0;
         SQLTCHAR message[1024] = {};
         SQLSMALLINT length = 0;
+        SQLTCHAR state[6] = {};
         ASSERT_SQL_OK(SQLGetDiagRec(SQL_HANDLE_STMT, reader_stmt, 1, state, &native, message,
                                     static_cast<SQLSMALLINT>(std::size(message)), &length),
                       SQL_HANDLE_STMT, reader_stmt);
         EXPECT_EQ(1222, native);
+        // 1222 is in neither driver's per-error table, so both fall through to
+        // their severity rule: class 16 maps to 42000. Asserted so a SQLSTATE
+        // divergence on this exact error cannot pass, and the msodbcsql parity
+        // leg validates it for free.
+        EXPECT_SQLSTATE(SQL_HANDLE_STMT, reader_stmt, "42000");
         EXPECT_LT(elapsed, std::chrono::seconds(5));
 
         ASSERT_SQL_OK(SQLEndTran(SQL_HANDLE_DBC, dbc_, SQL_ROLLBACK), SQL_HANDLE_DBC, dbc_);
-        SQLFreeHandle(SQL_HANDLE_STMT, reader_stmt);
-        SQLDisconnect(reader);
-        SQLFreeHandle(SQL_HANDLE_DBC, reader);
     }
 };
 
--- a/mssql-odbc/tests/e2e/tests/transaction_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/transaction_test.cpp
@@ -12,6 +12,7 @@
 #include "odbc_test_fixture.h"
 
 #include <chrono>
+#include <functional>
 #include <string>
 
 // SQL_TXN_SS_SNAPSHOT lives in msodbcsql.h, which these tests do not include.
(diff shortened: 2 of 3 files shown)


David-Engel:
I reached this from the other direction — investigating [AB#47510](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47510), the `test_003_connection` hang — and independently root-caused it to the same defect this PR fixes. Two things worth adding here, one confirming and one not.

## The ODBC-level fix works

I built this branch (`46d22d70`) and ran the new test against a live SQL Server 2025:

```
SQLExecDirectW: failed before the first result row
  e=Sql Error: 1222: Class 16: State 51: Lock request time out period exceeded.
[       OK ] TransactionLiveTest.BlockedSelectReturnsServerLockTimeout (1288 ms)
```

`peek_past_current_row()` does exactly what it should on the `SQLExecDirect` path.

## But the `mssql-python` scenario is still broken on this build

The Testing section lists `test_003_connection.py::test_set_attr_txn_isolation_effect` as verified. I could not reproduce that result. Running against this branch's driver — md5-verified as the `.so` the test container actually loads:

| | msodbcsql 18 | this branch |
| --- | --- | --- |
| `cursor.execute()` on the locked row | raises after 5.00s | **returns SUCCESS** after 5.00s |
| error surfaces | at execute | still deferred to `fetchall()` |
| same statement handle reused afterwards | `(42,)` | **`(None,)`** |
| fresh handle, same connection | `(42,)` | `(42,)` |

## Why the two disagree

`mssql_python/cursor.py` defaults to `use_prepare=True`:

```python
def execute(self, ..., use_prepare: bool = True, ...):
    ...
    effective_use_prepare = use_prepare and not same_sql
```

so `cursor.execute("SELECT ...")` goes through **`SQLPrepare` + `SQLExecute`** (`ddbc_bindings.cpp:1878`, `:1915`), not `SQLExecDirect`. The new e2e test only drives `SQLExecDirect`, so it passes while the path `mssql-python` actually uses does not.

`finish_execute` *is* shared — `execute.rs:164` and `:223` both call it — so the idea of the fix should cover both. The gap looks like it is in which branch the prepared path takes rather than in the approach. Worth confirming before merge, since the PR's stated motivation is the `mssql-python` case.

Repro, against a driver-swapped `mssql-python` (one connection holds the lock, a second reads with `SET LOCK_TIMEOUT 5000`):

```python
k1.execute("BEGIN TRANSACTION")
k1.execute("UPDATE ##t SET value = 'updated' WHERE id = 1")

k2.execute("SET LOCK_TIMEOUT 5000")
k2.execute("SELECT * FROM ##t WHERE id = 1")   # expect raise; returns SUCCESS
k2.execute("SELECT 42")                        # expect (42,); returns (None,)
```

Adding a `SQLExecute`/prepared-path case to `transaction_test.cpp` alongside the `SQLExecDirect` one would pin it and stop it regressing back.

## A second, separable defect this does not cover

The last row of that table is its own bug, independent of when the error surfaces: **after a server error mid-result-set, the statement handle is left returning stale empty rows.** A fresh handle on the same connection is fine, so the connection survives and the statement does not — its state is not reset when an ERROR token arrives mid-rowset. In general use that is the worse of the two, because it silently returns wrong data rather than failing.

It is recorded on AB#47510 and is fixable independently of this change — happy to take it separately so it does not widen this PR.

## Work item linkage

This PR is linked to AB#47771 and my investigation is on AB#47510. They are the same defect approached from two sides; worth linking or merging them so the `mssql-python` hang closes when this lands.

mssql-odbc/tests/e2e/tests/transaction_test.cpp
```diff
@@ -546,6 +547,56 @@ TEST_F(TransactionLiveTest, IsolationIsAppliedOnTheServer) {
...
+                                   static_cast<SQLSMALLINT>(connstr.size()), out_str,
+                                   static_cast<SQLSMALLINT>(std::size(out_str)), &out_len,
+                                   SQL_DRIVER_NOPROMPT),
+                  SQL_HANDLE_DBC, reader);
+
+    SQLHSTMT reader_stmt = SQL_NULL_HSTMT;
+    ASSERT_SQL_OK(SQLAllocHandle(SQL_HANDLE_STMT, reader, &reader_stmt), SQL_HANDLE_DBC, reader);
+    ASSERT_SQL_OK(SQLSetStmtAttr(reader_stmt, SQL_ATTR_QUERY_TIMEOUT,
+                                 reinterpret_cast<SQLPOINTER>(static_cast<SQLULEN>(10)), 0),
+                  SQL_HANDLE_STMT, reader_stmt);
```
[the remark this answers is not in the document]
David-Engel: Addressed, though the resolution changed while this sat in draft.

You were right about the state of the code at the time: `query_timeout` was stored but not enforced (`stmt.rs:317-324`), and `SQLCancel` only unwinds data-at-execution sequences, so it could not have interrupted a running execute either. There was no real backstop. I also confirmed nothing in the harness supplied one — `add_test` in `mssql-odbc/tests/e2e/CMakeLists.txt:105` sets no `TIMEOUT`, so the only bound was ctest's implicit 1500s default, multiplied by `--repeat until-pass:4` in CI.

Rather than add a watchdog thread, this now depends on #442, which has since merged and is merged into this branch. That wires enforcement through the exact path this test exercises: `next_row_cursor` passes `remaining_request_timeout` into `receive_row_header`, and `peek_past_current_row` delegates to `next_row_cursor`. So `SQL_ATTR_QUERY_TIMEOUT 10` is now a genuine deadline here and a regression returns `HYT00` rather than hanging. The comment on the attribute in the test says so explicitly.

Separately, your point stands for the suite as a whole: no e2e test has an explicit deadline. I'd rather fix that once in `add_odbc_test` than per test, so I'll raise it as its own change instead of widening this PR.

> [!NOTE]

David-Engel:
Thanks — this was a good catch and the coverage gap was real. I reproduced against a local SQL Server 2025 and pushed a fix to the tests. Results below, including one place where our diagnoses diverge.

## The prepared path is covered by the fix

Your root cause pointed at `use_prepare=True`, and that part is right: the e2e test only drove `SQLExecDirect`. But the prepared path is not broken. Driving the TDS layer directly against a blocked row:

```
PROBE execute_prepared     -> Ok(Rows)     # sp_prepexec route
PROBE metadata_is_empty    = false
PROBE has_open_batch       = true
PROBE peek_past_current_row -> Err(1222 Lock request time out period exceeded.)

PROBE sp_execute route     -> Ok(Rows)     # handle reused, what SQLPrepare+SQLExecute produces
PROBE peek2                -> Err(1222 ...)
```

Both routes emit COLMETADATA before the ERROR token, so `finish_execute` takes the `has_result_set` branch and `peek_past_current_row` surfaces 1222 — the same mechanism as `SQLExecDirect`.

My own leading hypothesis was wrong too, for the record: I expected metadata to be empty with `has_open_batch()` true, which would have hit the early return at `exec_common.rs:514` and returned `SQL_SUCCESS` with a 0-column cursor. The probe killed that.

## Test added anyway — the gap was real

`BlockedSelectTest.PreparedExecuteReturnsServerLockTimeout` now drives `SQLPrepare` + `SQLExecute`, sharing a body with the `SQLExecDirect` case. I verified both fail without the fix and pass with it:

| | fix disabled | fix enabled |
| --- | --- | --- |
| `ExecDirectReturnsServerLockTimeout` | FAILED (`rc=0`) | OK (1289 ms) |
| `PreparedExecuteReturnsServerLockTimeout` | FAILED (`rc=0`) | OK (1258 ms) |

Full ODBC e2e suite: 36/36 binaries pass.

## Worth flagging: how the first "fix disabled" run lied

My first attempt to disable the fix showed both tests still passing. The build copies the cdylib `libmssqlodbc.so` to `mssqlodbc.so`, and `cargo build` only refreshes the former — so the tests kept loading a stale driver that still had the fix. `md5sum` on both filenames caught it.

That is very likely relevant to your `mssql-python` run. Since we now disagree on the outcome for the same code, would you re-check the hash of the exact `.so` your container loads against a freshly built one? If it still reproduces on a verified-fresh driver, that points at something above the ODBC layer and I'd like to see it.

## Agreed on the rest

- The stale-handle bug (`(None,)` on a reused statement handle) is separable and worse. Please do take it separately on AB#47510; I have noted it in the PR description as known and out of scope here.
- PR description corrected. It claimed `test_003_connection.py::test_set_attr_txn_isolation_effect` was verified; since you could not reproduce that, I withdrew the claim rather than leave it standing.
- Work items linked in both directions.

Also note #442 has merged and is merged into this branch, so `SQL_ATTR_QUERY_TIMEOUT` is now a real deadline for these tests rather than the inert attribute it was when this PR was opened.

mssql-odbc/src/api/exec_common.rs
```diff
@@ -628,6 +628,15 @@ pub(super) fn finish_execute(
         };
     }
 
+    // SQL Server can send COLMETADATA before the statement has produced its
+    // first row. Wait for that row, end-of-set, or an ERROR token so execution
+    // errors surface from SQLExecDirect/SQLExecute instead of a later SQLFetch.
+    // A row is only positioned and parked; SQLFetch still receives it normally.
+    if let Err(e) = dbc.runtime.block_on(client.peek_past_current_row()) {
```
Vahid-b, line 635: *Cited, same file.* For a zero-row result set (`SELECT 1 WHERE 1=0`), this peek now reads the terminating DONE, so `current_result_set_has_been_read_till_end` is true and `has_open_batch()` is false by the time the success tail runs. But the tail then throws that away: `clear_exhaustion_state()` (line 650) wipes `result_set_exhausted`/`batch_exhausted`, and `return_client_busy` (line 655) keeps `active_stmt` claimed until a later `SQLFetch`/`SQLCloseCursor`. Another statement handle on the same connection gets `ERR_CONNECTION_BUSY` in the interim.

The sibling `release_busy_if_row_exhausted` (lines 262-310) exists — AB#47508 — to act on precisely this signal. Before this PR the driver genuinely could not know; now it does, and discards it.

Suggested shape, mirroring the sibling:

```rust
let has_more_rows = match dbc.runtime.block_on(client.peek_past_current_row()) {
    Ok(has_row) => has_row,
    Err(e) => {
        error!(%e, "{op}: failed before the first result row");
        return fail_with_tds(dbc, stmt, statement_handle, client, &e);
    }
};
let batch_done = !has_more_rows && !client.has_open_batch();
```

…then set `result_set_exhausted`/`batch_exhausted` in that case instead of clearing them, and release the claim.

Flagging rather than demanding: this changes busy-gate timing, so it may be cleaner as its own change under AB#47508 with its own e2e case. Either way it should not be lost — the information is now free and is being discarded.

mssql-odbc/tests/e2e/tests/transaction_test.cpp
```diff
@@ -546,6 +548,95 @@ TEST_F(TransactionLiveTest, IsolationIsAppliedOnTheServer) {
...
+        EXPECT_EQ(SQL_ERROR, rc) << "the lock timeout must surface from execution, not a later "
+                                    "SQLFetch";
+        SQLTCHAR state[6] = {};
+        SQLINTEGER native = 0;
+        SQLTCHAR message[1024] = {};
+        SQLSMALLINT length = 0;
+        ASSERT_SQL_OK(SQLGetDiagRec(SQL_HANDLE_STMT, reader_stmt, 1, state, &native, message,
+                                    static_cast<SQLSMALLINT>(std::size(message)), &length),
+                      SQL_HANDLE_STMT, reader_stmt);
+        EXPECT_EQ(1222, native);
```
Vahid-b, line 634: *Cited.* `state[6]` is filled by the `SQLGetDiagRec` above and then discarded; only `native` is asserted, so this test cannot catch a SQLSTATE divergence on the very error it is about.

Error 1222 is in neither driver's per-error table — `sqlstate.rs:449-451` jumps 1211 → 1505, and msodbcsql's `sqlcstr.cpp:192-195` has the same gap — so both fall through to a severity rule. Here `sqlstate_for_severity(16)` yields `42000` (`sqlstate.rs:549-557`, `SEVERITY_MAX_USER = 18`).

This file's header contract is that every assertion must also hold for msodbcsql 18, and the parity leg runs these two tests unguarded, so pinning the state costs one line and gets validated for free:

```cpp
EXPECT_EQ(1222, native);
EXPECT_SQLSTATE(SQL_HANDLE_STMT, reader_stmt, "42000");
```

I derived `42000` from both severity tables rather than observing it, so if the parity leg disagrees, take the leg's value — the point is that the assertion should exist.

mssql-odbc/tests/e2e/tests/transaction_test.cpp
```diff
@@ -546,6 +548,95 @@ TEST_F(TransactionLiveTest, IsolationIsAppliedOnTheServer) {
...
+        EXPECT_LT(elapsed, std::chrono::seconds(5));
+
+        ASSERT_SQL_OK(SQLEndTran(SQL_HANDLE_DBC, dbc_, SQL_ROLLBACK), SQL_HANDLE_DBC, dbc_);
+        SQLFreeHandle(SQL_HANDLE_STMT, reader_stmt);
+        SQLDisconnect(reader);
```
Vahid-b: Any of the six `ASSERT_SQL_OK`s above returns from this helper before reaching this cleanup, leaking `reader_stmt` and leaving `reader` connected; `SQLFreeHandle(SQL_HANDLE_ENV)` in teardown then trips `HY010`. The fixture's `TearDown` covers `dbc_`, but nothing covers the second connection.

The same shape already exists at lines 253-269, so this is consistent with the file rather than newly introduced — but the new helper has more `ASSERT` points ahead of its cleanup. A small RAII guard, or hoisting the reader connection into the fixture, would fix both. Not merge-blocking.

Vahid-b (commented):
## Summary

Reviewed the peek-past-`COLMETADATA` change, the parked-row contract in `mssql-tds`, and both new tests. The approach is correct and matches msodbcsql's actual mechanism, not just its observed outcome. No blocking findings; three suggestions and three nits, all about sibling entry points and test assertions rather than the change itself.

**Verification of the core parity claim (cited).** msodbcsql's execute and `SQLMoreResults` both drive `ProcessTDSStream`, and its `SQLROW`/`SQLNBCROW` handler at `sqlctokn.cpp:3617-3631` clears `TDS_RESULTSPENDING`, sets `TDS_ROWSPENDING`, and returns `S_PENDING` — the parse loop stops *at* the first ROW and parks it, which is exactly the model `peek_past_current_row` implements. The `COLMETADATA` handler (`sqlctokn.cpp:1965-1971`) only sets `RESULTSPENDING` and keeps going. So this PR's design is msodbcsql's design, not merely a behavioural match.

**Measured on this head (`e45c88d`).** Mutation: commenting out the four-line peek block in `exec_common.rs:635-639` makes `finish_execute_surfaces_an_error_before_the_first_row` fail (`left: 0; right: -1` at `exec_common.rs:835`). Restored, `cargo nextest run -p mssqlodbc --lib` = 1163 passed, 1 skipped. The test guards what it names.

**On the already-answered query-timeout thread** — resolved correctly, and I confirmed the mechanism rather than taking it on trust: `remaining_request_timeout` is cleared only in `close_query`/`advance` (`tds_client.rs:6386`) and the RPC-terminator look-ahead (`tds_client.rs:4477`, `:4492`), neither of which runs on the `Rows` path. It is still live when `finish_execute` peeks, so `SQL_ATTR_QUERY_TIMEOUT 10` is a real deadline in the new tests.

### Coverage ledger

| Row | State |
|---|---|
| Primary logic | Examined — `finish_execute`, `peek_past_current_row`, `next_row_cursor`, `handle_row_read_token` |
| Siblings | Examined — all three `finish_execute` branches; `try_next_row_cursor` / `try_next_buffered_row_into` (both return `Pending` on `RowPaused`, correctly deferring to the parked-row path); `release_busy_if_row_exhausted`; `SQLMoreResults`; catalog and `SQLParamData` tails |
| Callers & implementers | Examined — `execute.rs`, `exec_direct.rs`, `catalog.rs`, `param_data.rs` all funnel through `finish_execute` |
| Diagnostics | Examined — `fail_with_tds` takes and posts info messages; `post_tds_error` does per-record SQLSTATE mapping (see the inline comment on `EXPECT_EQ(1222, native)`) |
| Tests | Examined including harness — mutation-verified unit test; e2e helper reviewed |
| Build / packaging / pipelines | Examined — no changes; confirmed `enableOdbcE2E: true` in `validation-stages.yml`, so the new C++ tests *are* gated in PR validation |
| Docs & comments | Examined — the `peek_past_current_row` doc rewrite is accurate |
| Description & work items | Examined — AB#47771 and AB#47510 linked; the deferred defect has a work item |
| CI evidence | Examined — all 20 checks green on `e45c88d`, including the cross-repo mssql-python suite; coverage comment reports 100% diff |

## Blocking

None.

## Suggestion

**`more_results.rs:158` — `sql_more_results_safe`, `Ok(StatementResult::Rows)` arm.** *Cited.* The identical pre-row-error gap remains on the `SQLMoreResults` entry point: that arm returns `SQL_SUCCESS` as soon as `advance()` reports metadata, with no peek. For `SELECT 1; SELECT <blocked>`, execution now surfaces 1222 for statement 1's result set, but `SQLMoreResults` advancing onto set 2 still defers it to `SQLFetch` — the exact defect this PR fixes, one call later.

msodbcsql closes it with the same mechanism: `sqlcresl.cpp:355` calls `ProcessTDSStream(lpstmt, TDS_CROSSRESULTSET, GetQueryTimeOut(lpstmt))`, the same parse loop as execute, so the ROW handler cited above stops it at the first row there too. That makes this a divergence from the reference driver, not just an internal asymmetry.

I am not asking to widen this PR — per the repo's own rule about pre-existing crate-wide shapes, a follow-up item under AB#47771 or AB#47510 is the right disposition. But it is worth filing rather than leaving implicit, because the next reader will reasonably assume the class is closed.

(The two remaining suggestions, plus one nit, are left inline: `exec_common.rs:635`, `transaction_test.cpp:610`, `transaction_test.cpp:615`.)

## Nit

**PR description, "Not yet verified".** *Question.* The `mssql-python suite on mssql-odbc driver (cross-repo)` check is green on `e45c88d`. If that leg includes `tests/test_003_connection.py`, the withdrawn claim can be narrowed to "the specific two-connection lock reproducer is unverified" rather than all of mssql-python e2e — worth one line, since the mssql-python case is the PR's stated motivation. I could not determine the leg's test selection from the pipeline YAML alone.

**Description vs. diff.** The `tds_client.rs` doc-comment rewrite is a third changed file with no bullet in the Description section (the bot overview names it; the body does not). Trivial, but it is the kind of rider that is invisible to a later bisect.

## Deferred / to file

- The `SQLMoreResults` pre-row-error gap above — needs a work item; I would hang it off AB#47771.
- The missed early busy release on a zero-row result set (inline on `exec_common.rs:635`) — belongs under AB#47508 if not taken here.
- Reinforcing the already-tracked stale-handle defect with a citation for AB#47510: msodbcsql calls `CleanUphStmt(lpstmt, FREE_MEMORY)` when `retcode == SQL_ERROR` (`sqlccmd.cpp:6896-6905`), whereas `fail_with_tds` resets no statement state. That is a concrete mechanism for the `(None,)` symptom on a reused handle.

## What I could not check

No live SQL Server here, so the two e2e cases were not executed by me — I relied on the green PR-validation legs and the reported runs. I did not read an ADO task log for the msodbcsql parity leg, only the pipeline config that enables it. The mssql-python cross-repo leg's test selection is unverified.

*Confidence: high on the core mechanism, the msodbcsql parity, and the mutation result (read from source or run here); medium on the busy-release impact (reasoned from code, not measured); the exact `42000` value in the SQLSTATE nit is derived from both severity tables rather than observed, and is flagged as such inline.*

mssql-odbc/tests/e2e/tests/transaction_test.cpp
David-Engel: Added in `7495d23`, and I verified the value rather than taking the derivation on trust.

Your reading of the mapping is right: 1222 falls in the `1211` → `1505` gap (`sqlstate.rs:450-451`), so it goes to `sqlstate_for_severity(16)`, and with `SEVERITY_MAX_INFO = 10` / `SEVERITY_MAX_USER = 18` that yields `42000`.

Since the whole point of the assertion is the parity contract, I ran both tests against both drivers on a live SQL Server 2025 — including msodbcsql 18.6, which you noted you couldn't run:

| Driver | ExecDirect | PreparedExecute |
| --- | --- | --- |
| mssql-odbc (Rust) | OK (1338 ms) | OK (1237 ms) |
| msodbcsql 18.6 | OK (1242 ms) | OK (1233 ms) |

So `42000` is observed on both, not just derived, and the parity leg won't be the first thing to discover a divergence.

> [!NOTE]

mssql-odbc/tests/e2e/tests/transaction_test.cpp
David-Engel: Fixed in `7495d23` with a small RAII guard, since the new helper has six `ASSERT` points ahead of its cleanup and the failure mode is nasty — the `HY010` masks whichever assertion actually fired.

```cpp
class ScopedConnection {
  public:
    ~ScopedConnection() {
        if (stmt != SQL_NULL_HSTMT) { SQLFreeHandle(SQL_HANDLE_STMT, stmt); }
        if (dbc != SQL_NULL_HDBC) { SQLDisconnect(dbc); SQLFreeHandle(SQL_HANDLE_DBC, dbc); }
    }
    SQLHDBC dbc = SQL_NULL_HDBC;
    SQLHSTMT stmt = SQL_NULL_HSTMT;
};
```

I left lines 253-269 alone deliberately. You're right that it's the same shape, but it's pre-existing and unrelated to this defect, so folding it in would widen a diff that's otherwise scoped to the lock-timeout fix. The guard is a nested class in `BlockedSelectTest` for now; if the fixture-hoisting version appeals later, it can lift to `TransactionLiveTest` and pick up that call site in one move.

> [!NOTE]

mssql-odbc/src/api/exec_common.rs
David-Engel: Good catch, and your reading of the tail is correct — `clear_exhaustion_state()` then `return_client_busy` does throw away a signal the peek just made free.

I checked whether this is a regression or a missed optimization, since that decides where it belongs. It's the latter: before this PR the peek didn't run at execute time, so the tail wrote exactly the same state and held the claim just as long. The zero-row path is also behaviourally correct today — `next_row_cursor` short-circuits on `current_result_set_has_been_read_till_end`, so the DONE the peek consumed doesn't cost a later `SQLFetch` anything. The existing coverage agrees: `ExecDirectLiveTest.EmptyResultSet`, `FetchScrollLiveTest.ReturnsNoDataAtEndOfResultSet`, and two `connection_busy_test` cases that use `SELECT 1 WHERE 1 = 0` on exactly this busy-gate question all pass with the peek in place (36/36 binaries green).

So this PR doesn't regress the busy gate; it makes an earlier release newly *possible*. Taking your suggested shape here would change busy-gate timing on every result-set execution, which deserves its own e2e case proving the claim is released earlier — that's an AB#47508 change, not a lock-timeout one. Deferring per your own suggestion and the repo's rule about pre-existing crate-wide shapes.

Recorded in the PR description's deferred list alongside the `SQLMoreResults` gap from your summary, so neither is implicit.

> [!NOTE]

David-Engel:
Thanks for the depth here — the `sqlctokn.cpp` trace confirming this PR implements msodbcsql's actual mechanism rather than just matching its output is more than I had, and the mutation run on `e45c88d` is the right way to check a regression test earns its name.

All three inline threads are addressed in `7495d23` and resolved. Two follow-ups on your nits, one of which is a correction.

## Your SQLSTATE derivation was right, and is now observed

I ran both new cases against both drivers on a live SQL Server 2025 — including msodbcsql 18.6, which you flagged you couldn't run:

| Driver | ExecDirect | PreparedExecute |
| --- | --- | --- |
| mssql-odbc (Rust) | OK (1338 ms) | OK (1237 ms) |
| msodbcsql 18.6 | OK (1242 ms) | OK (1233 ms) |

`42000` holds on both, so the assertion is parity-clean rather than pinning a Rust-only value.

## Correction: the green cross-repo leg does not narrow the withdrawn claim

> If that leg includes `tests/test_003_connection.py`, the withdrawn claim can be narrowed…

It does include it — `run-mssql-python-odbc-tests.sh:105` is `find tests -name 'test_*.py'`, no selection. But the leg cannot validate any individual test, because it is green by design when tests fail:

- `failTaskOnFailedTests: false` on the publish task (`test-mssql-python-odbc-template.yml:286`).
- A non-zero runner exit is downgraded to `type=warning` / "partially succeeded" (`:244-250`), explicitly commented as "expected while the driver is in development".
- The runner returns exit 1 for any failure, crash, timeout *or* budget skip (`run-mssql-python-odbc-tests.sh:260`).

So a green glyph there is consistent with `test_003_connection.py` failing outright. I've kept the claim withdrawn and written the reasoning into the description rather than narrowing it — this is the same shape as the stale-`.so` trap from earlier in this PR, where the reassuring signal wasn't measuring what it appeared to.

If you want that leg to be readable as evidence, that's its own change: a known-good baseline list, so a *regression* fails the job while known gaps stay yellow. Happy to file it if you agree it's worth having.

## Deferred items are now on the record

Description carries all three: the `SQLMoreResults` gap (AB#47771), the zero-row busy release (AB#47508), and your `CleanUphStmt` citation for AB#47510. On the busy release specifically — I checked whether it was a regression before deferring, and it isn't: pre-PR the tail wrote the same state, and the zero-row path is covered green by `ExecDirectLiveTest.EmptyResultSet`, `FetchScrollLiveTest.ReturnsNoDataAtEndOfResultSet` and two `connection_busy_test` cases that use `SELECT 1 WHERE 1 = 0` on exactly this question. It's a newly-free signal being dropped, not a behaviour change, which I think supports your instinct to give it its own e2e case under AB#47508.

Also took the `tds_client.rs` description bullet — you're right that an unlisted third file is exactly what goes invisible in a later bisect.

shiwanigupta0809:
Reviewed the peek change against `finish_execute`'s three branches, the `handle_row_read_token` arms the peek now traverses, and the msodbcsql parse loop the design claims parity with. Read the existing threads first and am not re-filing any of them.

### Parity claim independently confirmed

I checked the mechanism in the msodbcsql source rather than taking it from the thread, since the whole design rests on it:

- `OnColMetaData` (`sqlctokn.cpp:1894`) sets `TDS_RESULTSPENDING` and falls through — the loop keeps parsing past metadata.
- The `SQLROW`/`SQLNBCROW` handler clears `TDS_RESULTSPENDING`, sets `TDS_ROWSPENDING`, and does `hr = S_PENDING; goto _EXIT;` (`sqlctokn.cpp:3617-3627`) — the loop stops *at* the first row and parks it.

So "stop at the first ROW, park it, report anything that arrives before it" is msodbcsql's own model, and `peek_past_current_row` reproduces it. The approach is right.

### Suggestion — an undocumented behaviour change that follows from the peek

**The peek also collects server INFO tokens that arrive after `COLMETADATA`, which moves those diagnostics from `SQLFetch` to the execute call — and for an application that inspects diagnostics only after fetching, moves them out of reach entirely.**

The peek does not stop at rows and errors. `peek_past_current_row` → `next_row_cursor` → `handle_row_read_token`, whose `Tokens::Info` arm calls `self.capture_info_message(&info_token)` and continues. Any INFO token sitting between `COLMETADATA` and the first `ROW`/`DONE` is therefore drained during execution now, where it was previously drained by the first `SQLFetch`.

`finish_execute`'s success tail then acts on it:

```rust
let info_messages = client.take_info_messages();
...
let has_server_info = post_tds_info_messages(&mut stmt_state, &info_messages);
...
if has_server_info { SQL_SUCCESS_WITH_INFO } else { SQL_SUCCESS }
```

Two consequences, and the second is the one that bites:

1. `SQLExecDirect`/`SQLExecute` returns `SQL_SUCCESS_WITH_INFO` where it previously returned `SQL_SUCCESS`.
2. ODBC clears a handle's diagnostics at the start of the next call on it, so by the time `SQLFetch` runs, the record is gone. An application that checks `SQLGetDiagRec` after fetching — a normal pattern when execute returned `SQL_SUCCESS` — now **never sees the warning at all**. It is not deferred or duplicated; it is dropped for that caller.

The likely everyday instance is the aggregate warning, msg 8153 `"Null value is eliminated by an aggregate or other SET operation"` (`01003`), on something as ordinary as `SELECT SUM(col) FROM t`. I have not confirmed on the wire that SQL Server emits 8153 before the row rather than after it, so treat the specific example as unverified — but the structural point does not depend on it: *whatever* INFO tokens occupy that window are now execute-time diagnostics.

Worth saying clearly that this is a **parity improvement, not a divergence**. msodbcsql's loop covers the same window before it parks at the first row, so it reports those warnings from execute too. This change moves us onto its behaviour on info messages as well as on errors.

What is missing is that nothing records it. The description's headline is errors ("server errors arriving after metadata now surface from `SQLExecDirect` or `SQLExecute`"), Breaking Changes says None, and neither new test covers a post-metadata warning — `finish_execute_surfaces_an_error_before_the_first_row` scripts an ERROR token, and the two e2e cases assert `SQL_ERROR`/1222. So the success-with-info path is entirely unpinned, and it is the one that changes for queries that were already working.

Worth a line in the description plus one unit case: script `col_metadata`, an INFO token, a row, and `done_no_more`, then assert `finish_execute` returns `SQL_SUCCESS_WITH_INFO` and the record is on the statement. That also pins the ordering guarantee against a future change that moves the peek after `take_info_messages()`, which would silently restore the old behaviour with no test failing.

Ancillary evidence the mechanism is already understood in this file, just not documented for this path: `release_busy_if_row_exhausted` carries `StmtState::pending_fetch_info` precisely because *its* peek drains info messages with no success return to post them under. `finish_execute` has a success return, so posting is the right call here — it is only the observable consequence that is unrecorded.

shiwanigupta0809 (commented):
_Reviewed — see https://github.com/microsoft/mssql-rs/pull/464#issuecomment-5513496344_

David-Engel:
This is a genuinely good catch — thank you. The claim is correct, I reproduced the consequence rather than reasoning about it, and it's now covered.

## Verified

`handle_row_read_token`'s `Tokens::Info` arm (`tds_client.rs:6236-6240`) is `capture_info_message(&info_token); Ok(None)` — captured, loop continues. So the peek does drain that window, and `finish_execute`'s success tail posts it and returns `SQL_SUCCESS_WITH_INFO`. Your two consequences both follow, and your framing is right: this is msodbcsql's behaviour, so it's parity gained, not parity lost.

## Pinned, and the mutation you predicted is exactly what catches it

Added `finish_execute_reports_an_info_message_arriving_before_the_first_row`: scripts `col_metadata` → `info(8153, 10, ...)` → `done_no_more`, asserts `SQL_SUCCESS_WITH_INFO` and the record on the statement.

I then ran the specific mutation you named — moving the peek below `take_info_messages()`:

```
FAIL finish_execute_reports_an_info_message_arriving_before_the_first_row
  left: 0, right: 1          # SQL_SUCCESS, no record
PASS finish_execute_surfaces_an_error_before_the_first_row
```

The error test passes under that mutation. So you were right that this path was entirely unpinned: that edit would have silently restored the old behaviour with nothing failing. Restored, both pass; full unit suite 1180 passed.

One deviation from your suggested script: I left the ROW token out. Scripted rows come from a separate queue in `TokenReplayTransport` (`position_int_row` is consulted before `pending_tokens`), so interleaving a row between the INFO and the DONE isn't expressible in `tds_client_from_tokens`. A zero-row set exercises the same post-`COLMETADATA` window and catches the same mutation, so I don't think anything is lost — but say the word if you'd rather see the row variant and I'll extend the harness.

## Description

Added an "Observable behaviour change beyond errors" section, and Breaking Changes no longer says "None" — it now names the `SQL_SUCCESS` → `SQL_SUCCESS_WITH_INFO` shift and the after-fetch-diagnostics consequence, since as you say that's the part which changes for queries that were already working.

On the 8153 example: I've kept it as the illustration in the test but didn't assert it's what SQL Server emits pre-row, matching your own caveat. The test pins the structural behaviour, not that specific message's placement on the wire.

Also noted your point about `pending_fetch_info` in `release_busy_if_row_exhausted` — that's a good pointer, and it's the reason posting directly is right here: `finish_execute` has a success return to carry the record, whereas that peek doesn't.

Theekshna (commented):
Unattended run — no new findings. This PR has already had exhaustive review (Copilot bot, David-Engel, Vahid-b, shiwanigupta0809) covering the core mechanism, msodbcsql parity, mutation-verified tests, and process-safety concerns; all four review threads are resolved. I re-derived the key claims independently rather than trusting the prior passes.

**Verified independently:**
- The core mechanism in `mssql-tds`: `peek_past_current_row` → `next_row_cursor` → `handle_row_read_token`. `Tokens::Error` (line 6207) records the error, drains the stream, and returns `Err`, which propagates out through `next_row_cursor`/`peek_past_current_row` to `finish_execute`'s `if let Err(e) = ... { return fail_with_tds(...) }` — so a pre-row server error genuinely surfaces from execute now. `Tokens::Info` (line 6236) captures and returns `Ok(None)`, so the loop continues past it — confirming the PR's claim that INFO tokens in that window are drained during execution, not just errors.
- **msodbcsql parity, checked directly against the reference tree** (not taking Vahid-b's citation on trust): `sqlctokn.cpp:3614-3627` — the non-`SQLALTROW` branch clears `TDS_RESULTSPENDING`, sets `TDS_ROWSPENDING`, and returns `S_PENDING` at line 3626, parking the parse loop at the first row exactly as described. The `COLMETADATA` handler at `sqlctokn.cpp:1965` only sets `TDS_RESULTSPENDING` and continues. Citation is accurate.
- The ordering fix is present at this head: `exec_common.rs`'s peek runs before `client.take_info_messages()`, which is what makes the info-message capture land under the current execute rather than a later fetch.
- `cargo check -p mssqlodbc -p mssql-tds --lib --tests --offline` compiles clean at `1b5e2eaf`.

**Deferred items** (SQLMoreResults pre-row-error gap, zero-row busy-release discard, stale-handle cleanup on SQL_ERROR) are all documented in the PR's "Deferred / to file" section with work-item references (AB#47771, AB#47508, AB#47510) — legitimate deferrals per the repo's pre-existing-crate-wide-shape convention, not silently dropped findings.

**Not re-run:** live e2e / mssql-python suite (no live SQL Server in this environment) — relying on the reported CI runs, consistent with the PR's own stated verification. `gh pr checks 464` shows the Azure DevOps validation pipeline still running against the latest commit (build 171913, all legs pending, none failed) at review time.

msodbcsql parity: verified above. Test sufficiency: both new unit tests were independently confirmed by the author to be mutation-guarded (documented failure output in the PR discussion), and the newest commit specifically closes the info-message coverage gap that shiwanigupta0809 flagged. Divergences: none open — this closes a divergence rather than introducing one. PR description: current, includes the "Observable behaviour change" and "Breaking Changes" sections added in response to review. No AI slop in the new doc comments — the `peek_past_current_row` rewrite and the new test doc comments each explain a real invariant (parking contract, why the test would silently pass under a regression), not restatement.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

shiwanigupta0809 (approved):

David-Engel:
Re-checked as you asked, and **you were right — my result was a stale-driver artifact. I withdraw the claim that the `mssql-python` case was still broken.**

The hashes in my own worktree:

```
libmssqlodbc.so   Sep 1 23:06   6be900723fd2a23a6e585be2a1d8fa01   <- the #464 build
mssqlodbc.so      Sep 1 17:55   706fd5c1bacbfb35e2194669d7ffd450   <- pre-#464 main
```

`706fd5c1…` is exactly the hash I reported as "md5-verified as the `.so` the test container actually loads". So the container loaded it faithfully — it just was not your fix. I ran `cargo build --release -p mssqlodbc` for the #464 build but only ran `finalize-artifact.sh` on the earlier main build, so `mssqlodbc.so` was never refreshed. Precisely the trap you described.

Worth naming the reasoning error, because the hash check made it look rigorous: I verified *the container loads the file I copied*, not *the file I copied is the new build*. Those are different claims and only the second one mattered. A hash is only evidence if it is compared against something independently known to be correct — comparing a file to itself proves nothing. The right check was `md5sum` against a freshly finalized artifact, or simply asserting the new symbol is present.

That also retires the "something above the ODBC layer" theory: there is nothing left to explain. Your direct TDS probe showing both `sp_prepexec` and `sp_execute` emitting COLMETADATA before the ERROR token, and `peek_past_current_row` surfacing 1222 on both, is the correct account of the prepared path.

Two things stand from my side:

- The coverage gap was real and `PreparedExecuteReturnsServerLockTimeout` is the right answer to it — the fix-disabled/fix-enabled table is exactly the evidence I would have wanted.
- The stale-handle bug (`(None,)` on a reused statement handle after a mid-result-set error) is unaffected by the driver mix-up: I saw it on the pre-#464 build, which is where it belongs. I will take it on AB#47510 and re-confirm it against current `main` before writing anything, on a freshly finalized artifact this time.

Sorry for the noise on a PR that was correct.

PR #476, opened 2026-09-02
Opt mssql-odbc into the workspace lint configuration

### LINKED ISSUE #475: mssql-odbc does not inherit [workspace.lints] — it has no [lints] section
### Problem statement

`mssql-odbc/Cargo.toml` has **no `[lints]` section at all**, so it does not inherit `[workspace.lints]` from the root `Cargo.toml`. Cargo lint inheritance is opt-in per crate — a workspace member gets nothing unless it declares:

```toml
[lints]
workspace = true
```

Every other workspace member does this. `mssql-odbc` does not:

| Crate | `[lints] workspace = true` |
|---|---|
| `mssql-tds` | ✅ |
| `mssql-js` | ✅ |
| `mssql-tds-cli` | ✅ |
| `mssql-mock-tds` | ✅ |
| `mssql-tds-bench` | ✅ |
| **`mssql-odbc`** | ❌ **no `[lints]` section** |
| `mssql-tds/fuzz` | ❌ (not a workspace member) |
| `mssql-mock-tds-py` | ❌ (excluded from workspace) |

So `mssql-odbc` — the crate that holds the entire unsafe C FFI surface, roughly 154 `unsafe fn`s — is the one crate exempt from centrally-configured lints.

**Why this is easy to miss.** Running `cargo clippy -p mssqlodbc` still prints plenty of warnings, but they come from `mssql-tds` being rebuilt as a path dependency. Confirmed by attributing warnings to their source file: with `missing_docs_in_private_items = "warn"` temporarily added to `[workspace.lints.clippy]`, `cargo clippy -p mssqlodbc` reported 1089 warnings and **all 1089 were in `mssql-tds/` files**. `mssql-odbc` itself produced zero. A raw warning count looks healthy and hides the gap completely.

**Impact today** is limited but not nil. `[workspace.lints.rust]` sets `unexpected_cfgs`, and `[workspace.lints.clippy]` has three *active* entries — `await_holding_lock`, `await_holding_refcell_ref`, and `large_futures` — with the remaining 35 commented out. `mssql-odbc` is exempt from all four. `cargo bclippy` still passes `-D warnings` on the command line, so general clippy defaults do apply.

The three active ones are precisely the async-correctness rules (deadlock and oversized-future detection), and `mssql-odbc` is a heavily async crate that boxes future state deliberately — so it is an unfortunate crate to have exempt.

**Impact tomorrow** is the real concern. The commented-out block in the root `Cargo.toml` is clearly staged for adoption:

```toml
[workspace.lints.clippy]
# Don't Panic - prevent panics from unwraps and unsafe slicing or indexing
# string_slice = "warn"
# indexing_slicing = "warn"
# unwrap_used = "warn"
...
```

Whenever those are uncommented, they will silently skip `mssql-odbc`. Rules named "Don't Panic — prevent panics from unwraps and unsafe slicing or indexing" not applying to the driver's FFI layer is close to the opposite of the intent.

### Root cause

Not an override or a suppression — simply the absence of an opt-in, which Cargo treats as "inherit nothing".

`[workspace.lints]` is a *definition* table, not an *application* table. Declaring lints at the root does nothing on its own; it only makes them available for members to inherit. This mirrors `[workspace.dependencies]`, which this repo already uses the same way — define centrally, then opt in per crate with `{ workspace = true }`.

The timeline explains how it happened:

| When | What |
|---|---|
| 2026-04-30 | `523cf5f7` "Add workspace clippy lint infrastructure with recommended lints" adds `[workspace.lints.*]` and opts in **every crate that existed at the time**: `mssql-tds`, `mssql-js`, `mssql-tds-cli`, `mssql-mock-tds`. The members list was exactly those four. |
| 2026-07-08 | `5d375afc` "Merge OdbcOnRust to github main (#94)" imports `mssql-odbc` from a separate codebase, ~10 weeks later. Its manifest was authored independently and never picked up the convention. |

`git log -S'[lints]' -- mssql-odbc/Cargo.toml` returns nothing, confirming the section was never present rather than removed later.

Nothing in Cargo or CI flags a member that skips the opt-in, so this stayed invisible. `mssql-tds-bench`, also added after the infrastructure landed, *does* opt in — so the convention is understood, it was just missed on the imported crate.

Preventing recurrence is worth a thought as part of the fix: any crate added to `members` should carry `[lints] workspace = true`, and that is currently enforced only by reviewer memory.

### Proposed solution

Add to `mssql-odbc/Cargo.toml`:

```toml
[lints]
workspace = true
```

**This is free right now.** Verified against `a128ac78`:

```
cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings
→ exit 0
```

No new warnings, because the only active workspace lint is `unexpected_cfgs` and `mssql-odbc` does not trip it. Doing it now — while the cost is zero — means the staged clippy rules apply everywhere the moment they are switched on, instead of someone discovering the gap later with a backlog attached.

**Measured cost of the eventual full adoption.** Since the obvious follow-up question is "what happens when the other 35 are switched on", here are the numbers for `mssql-odbc` with all 38 staged lints enabled, measured against `a128ac78`:

| Scope | Warnings |
|---|---:|
| Today, `[lints] workspace = true` with current config | **0** |
| All 38 staged lints, production code (`--lib`) | **682** |
| All 38 staged lints, including tests (`--all-targets`) | **3,351** |

Breakdown of the full 3,351:

| Count | Lint |
|---:|---|
| 1,786 | `undocumented_unsafe_blocks` |
| 999 | `unwrap_used` |
| 284 | `indexing_slicing` |
| 46 | `panic` |
| 45 | `allow_attributes_without_reason` |
| 44 | `allow_attributes` |
| 42 | `assertions_on_result_states` |
| 38 | `multiple_unsafe_ops_per_block` |
| 33 | `cast_sign_loss` |
| 15 | `map_err_ignore` |
| 13 | `let_underscore_must_use` |
| 4 | `unnecessary_safety_comment` |
| 2 | `string_slice` |

The gap between 682 and 3,351 is almost entirely test code: **every one** of the 999 `unwrap_used`, 46 `panic`, and 42 `assertions_on_result_states` hits is in tests, as is 1,333 of the 1,786 `undocumented_unsafe_blocks`. If tests get a relaxed profile — which is normal for these lints — the real production backlog is 682, concentrated in `get_data.rs`, `fetch_scroll.rs`, and the conversion modules.

None of this blocks the fix proposed here, since the cost today is zero. It is recorded so whoever enables the staged lints later can size the work per lint rather than switching them all on at once.

Also worth deciding deliberately, rather than by omission:

- `mssql-tds/fuzz` is a separate non-member crate and inherits nothing.
- `mssql-mock-tds-py` and `mssql-py-core` are in the workspace `exclude` list, so they cannot inherit `[workspace.lints]` at all. They need their own lint configuration, which is consistent with them already needing separate `fmt`/`clippy` invocations (see `scripts/bclippy.ps1`).

### Affected crate

mssql-odbc

### Alternatives considered

- **Leave it and add `[lints] workspace = true` at the time the clippy rules are uncommented.** Workable, but it merges two unrelated problems: whoever enables `unwrap_used` would suddenly face the whole `mssql-odbc` backlog in the same PR. Landing the opt-in now, at zero cost, separates them cleanly.
- **Repeat the lint list in `mssql-odbc/Cargo.toml` instead of inheriting.** Defeats the purpose of `[workspace.lints]` and will drift.
- **Rely on `-D warnings` in `cargo bclippy`.** That escalates clippy's *default* lints but cannot enable non-default ones like `unwrap_used` or `indexing_slicing`, which is exactly what the commented-out block is for.

### Additional context

Found while investigating #474 (enforcing `# Safety` docs on private `unsafe fn`s). The two are related but need different fixes, and the distinction is worth recording:

- `[workspace.lints]` in `Cargo.toml` sets lint **levels** and requires per-crate opt-in — which is what this issue is about.
- `clippy.toml` sets lint **configuration options**, is resolved by walking up the directory tree, and needs no opt-in. Verified: a root `clippy.toml` with `check-private-items = true` produced 50 warnings in `mssql-odbc` files even though `mssql-odbc` has no `[lints]` section.

So #474 is not blocked by this issue, and this issue is worth fixing on its own merits.

saurabh500:
## Description

`mssql-odbc` was the only workspace member without a `[lints]` section, so it inherited none of the workspace lint configuration. This adds the two lines that opt it in:

```toml
[lints]
workspace = true
```

**Why it was missing.** The crate predates the lint setup. `523cf5f7` (2026-04-30) added `[workspace.lints]` and opted in the four members that existed at the time; `mssql-odbc` arrived later, on 2026-07-08 via `5d375afc`, as an import from a separate codebase. `git log -S'[lints]' -- mssql-odbc/Cargo.toml` returns nothing — the section was never there.

`[workspace.lints]` is a *definition* table, not an *application* table. Members inherit only via an explicit `[lints] workspace = true`, the same opt-in pattern as `[workspace.dependencies]`.

**What it was exempt from.** `unexpected_cfgs`, plus the three currently-active clippy lints — `await_holding_lock`, `await_holding_refcell_ref` and `large_futures`. Those are the async-correctness rules (deadlock and oversized-future detection), and `mssql-odbc` is a heavily async crate that boxes future state deliberately, so it is an unfortunate one to have exempt.

**Why it went unnoticed.** `cargo clippy -p mssqlodbc` still emits plenty of warnings, but they all originate in `mssql-tds` as a path dependency. A raw warning count looks healthy and hides the gap completely — attributing warnings by file path via `--message-format=json` is what surfaces it.

## Verification

Confirmed the lints now actually reach the crate, by inspecting the rustc invocation (`cargo clippy -p mssqlodbc -v`):

```
--warn=clippy::await_holding_lock
--warn=clippy::await_holding_refcell_ref
--warn=clippy::large_futures
--warn=unexpected_cfgs
```

None of these were passed before this change. **Zero new warnings result**, so this is purely closing the gap rather than a behaviour change — no source files are touched.

## Follow-up, deliberately not in this PR

The other 35 lints in `[workspace.lints.clippy]` remain commented out workspace-wide. For sizing, `mssql-odbc` alone would emit **682** warnings in production code (`--lib`) and **3,351** including tests if all 38 were switched on. `undocumented_unsafe_blocks` (1,786) and `unwrap_used` (999) are 83% of that, and every `unwrap_used` hit is in test code. Those numbers are recorded in #475 so the staged lints can be adopted per-lint rather than all at once.

## Related Issues

Fixes #475

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes — 3,116/3,120 unit tests pass, all `mssqlodbc` tests green. The 4 failures are the pre-existing `certificate_validator` fixture tests (they need `mssql-tds/tests/test_certificates/generate_certs.sh`); integration suites need a `.env` with a live server. Both are environmental and fail identically on `main`.
- [x] New/changed functionality has tests — n/a, build configuration only
- [x] Public API changes are documented — n/a, no API change

### DIFF
--- a/mssql-odbc/Cargo.toml
+++ b/mssql-odbc/Cargo.toml
@@ -8,6 +8,9 @@ publish = false
 name = "mssqlodbc"
 crate-type = ["cdylib"]
 
+[lints]
+workspace = true
+
 [dependencies]
 mssql-tds = { path = "../mssql-tds" }
 tokio = { version = "1", features = ["rt-multi-thread", "sync"] }


Theekshna (approved):

Theekshna (commented):
Unattended run — no findings. This is a 3-line `Cargo.toml`-only change (`[lints]\nworkspace = true` for `mssql-odbc`), no source touched.

**Verified:**
- `git diff` against the merge-base matches the PR's own reported stats exactly (3/0/1 file).
- The lint set the PR claims mssql-odbc was missing — `unexpected_cfgs`, `await_holding_lock`, `await_holding_refcell_ref`, `large_futures` — matches `[workspace.lints]` in the root `Cargo.toml` exactly; every other clippy lint listed there is commented out, so nothing else is newly enabled.
- Ran `cargo clippy -p mssqlodbc --all-targets --offline` at this head: zero warnings, confirming the PR's "zero new warnings" claim rather than trusting it.
- No merge-conflict markers; `baseRefName` (`main`) matches the PR's target.

msodbcsql parity: N/A, no ODBC behavior touched. Test sufficiency: N/A for a lint-config change — the clippy run above is the applicable verification and it passed clean. Divergences: N/A. PR description: matches the diff exactly, linked to #475, checklist claims consistent with the change's scope. No AI slop — no new comments or prose added.

CI: Azure DevOps validation still running on this head (build 171909) at review time; `copilot-pull-request-reviewer[bot]` already recommended approval.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

shiwanigupta0809 (approved):

PR #477, opened 2026-09-02
Require an invocation before a review claims a capability is unavailable

Vahid-b:
## Description

The code-review skill told attended reviews that Azure DevOps was unreachable, so they skipped the work-item cross-check without ever calling the tool. A recent review closed with *"ADO is not reachable non-interactively from this session; I took the work-item reference at face value per the skill's fail-open rule."* The reviewer followed the skill correctly — the instruction was defective.

This is a **scope leak, not a stale fact**, which matters because the fix differs. A stale fact gets corrected; a scope leak has to be re-scoped, or it returns the next time someone re-verifies the tool and edits the wrong sentence. Three defects compounded:

1. **The hazard was attributed to the tool, not the context.** The step 6 bullet sits under *"An unattended run still:"*, where treating interactive auth as unavailable is correct. But it read `The Azure DevOps MCP server is the known trap`, a standing claim about the server — so it travelled into attended sessions, where a human can answer any prompt.
2. **"Fail open" was never scoped.** It sits outside the unattended-run bullet list and reads as a general rule, with nothing bounding "unreachable" to a call that was attempted and failed.
3. **Step 1 imported both into every review.** Step 1 is universal and ended with `See step 6 for handling ADO in an unattended run`.

A fourth, independent issue: step 5 described cross-platform coverage as CI-only and never mentioned that `cargo check` / `cargo clippy --target` type-check `cfg`-gated code from any host without linking — making "my host is Linux, so the CI matrix owns Windows" the default.

### Verified 2026-09-02, by invocation

- Azure DevOps MCP `wit_work_item` (`action: get`) returned a work item **immediately** — no interactive prompt, no hang.
- `cargo clippy --target aarch64-pc-windows-msvc -p mssql-tds --all-targets` exits 0 in ~26 s and **does** compile the `cfg(windows)` paths — `--all-targets`, not `--lib`: confirmed by mutating a `#[cfg(windows)]` test in `datasource_parser.rs`, which `--lib` missed entirely and `--all-targets` caught.
- Running a Windows-target test binary genuinely *is* unavailable here (no MSVC linker), and cross-target `clippy` can't stand in for it — it type-checks, it never executes a test, so it says nothing about whether mutations can be run outside Linux either. Those are the two caveats of four that survived testing; only ADO access and type-checking Windows-gated code without a Windows host were actually disproven by invocation.

### What changed

| # | Location | Change |
|---|---|---|
| 1 | Step 1 | Attempt the lookup before treating it as out of reach |
| 2 | Step 6 bullet | Re-scoped to the *context*, not one server; records the non-interactive behaviour with date and evidence |
| 3 | Fail open | Retitled *"after a failed call, not instead of one"* |
| 4 | Step 5 | Cross-target type-checking for platform-gated code |
| 5 | Principles | *"An unavailability is a claim, and it carries the same burden as a defect"* |

Edit 5 is the one that generalises — the others fix ADO and Windows; that one covers the next capability nobody has probed.

The interactive-auth warning is **kept**, not deleted: it remains correct for pipeline runs, and removing it would trade one over-application for a real hang later. The "known trap" history is left visible so a reviewer who remembers the old wording can see the change was deliberate.

**Out of scope:** using the Windows SDK ODBC headers as the authority for symbol/type questions. Useful, but Windows-host-specific — it belongs in a personal agent, not a shared skill that macOS and Linux teammates read.

### Review corrections

Review found several issues in the initial version, all fixed on this head (see resolved threads and commit history for detail):

- A timeout/no-response failure couldn't satisfy the Principles bullet's "quote what came back" — recording the outcome is now the requirement, quoting only when a response exists (Copilot).
- Step 1's ADO claim and the Principles citation were undated, against the file's own preamble (David-Engel).
- `wit_query` was claimed verified alongside `wit_work_item`, but only the latter was actually invoked (Copilot).
- No outcome existed for "the tool isn't exposed in this session at all" — added, and mirrored into steps 1 and 6 (Copilot).
- `cargo check` was shown without `--target`, silently defeating the check it documents (Copilot).
- The local cross-target check read as a requirement; reframed as optional, with a CI-job-verification fallback (saurabh500).
- The exemplar used `-p mssqlodbc --lib`, which skips `#[cfg(test)]` entirely — most of this repo's platform-gated code lives in inline test modules, so the example never validated what it claimed to. Switched to `-p mssql-tds --all-targets`, the crate where the platform-gated surface actually is (shiwanigupta0809, Theekshna).
- The Principles bullet's "(2026-09-02, [AB#47807](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47807))" cited this PR's own work item instead of the review it described — the caveats are from a 2026-09-02 review of #459, tracked under [AB#47509](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47509) (Theekshna).
- The `openssl-sys` build-script caveat attributed the dependency solely to `native-tls` and scoped it to Linux; `mssql-tds` also pulls `openssl` directly for Always Encrypted, and the gate is non-Windows/non-macOS generally (David-Engel).
- "Neither approach can run a Windows test binary" sat next to a sentence about CI and read as if CI couldn't run Windows tests either — false, CI's Windows job runs them via `nextest`. Reworded to name `cargo check`/`cargo clippy` specifically (Copilot).
- The Principles bullet claimed cross-target `clippy` disproved "Linux-only mutation scope", but `clippy` only type-checks — it never executes a test, so it can't speak to mutation-test execution. That caveat survives alongside the Windows-test-binary one; only two of the four caveats were actually disproven by invocation (Copilot).

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47807

[AB#47807](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47807)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes — see note
- [ ] New/changed functionality has tests — n/a, documentation only
- [ ] Public API changes are documented — n/a, no API change

**Note on `cargo btest`:** it fails on this workstation, and it fails **identically on `origin/main`**. I ran the same invocation on a clean `origin/main` worktree: `mssql-tds --test transaction` reports `16 tests run: 0 passed, 16 failed` on both. The failures are the transaction, connectivity and bench suites, all of which need a live SQL Server that is not configured here. This branch changes exactly one file — `.github/skills/code-review/SKILL.md` — and no Rust, build script or manifest, so it cannot affect them. CI is the authority on that box.

Coverage: `pr-code-coverage.yml`'s `paths-ignore` covers `.github/**`, so no Code Coverage Report comment is expected on this PR.

### DIFF
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -95,21 +95,22 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
-   - **Does the platform-gated half still compile?** `cargo check` / `cargo clippy
-     --target <triple>` type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]`
-     code without needing a matching host, because neither links (`rustup target add
-     aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). It cannot run a
-     Windows test binary, so runtime behaviour there still rests on the CI matrix —
-     but "my host is Linux, so the matrix owns this" is the wrong default when
-     reviewing platform-gated code, and a mutation applied to an arm your host does
-     not select proves nothing on its own. A failure here can still come from a
-     dependency's build script rather than the reviewed code: targeting a Linux
-     triple pulls `openssl-sys` in through `native-tls`, and its build script fails
-     before rustc ever sees the crate if it can't find a target OpenSSL sysroot —
-     confirmed on `mssql-odbc`, where the Windows-target direction above needs no
-     OpenSSL at all (`native-tls` routes through Schannel on Windows, Security.framework
-     on macOS — only Linux targets hit the sysroot lookup). That is an environment gap,
-     not a type error; read what actually failed before attributing it to the diff.
+   - **Does the platform-gated half still compile?** `cargo check --target <triple>` /
+     `cargo clippy --target <triple>` type-checks `#[cfg(windows)]` and
+     `#[cfg(target_os = ...)]` code without needing a matching host, because neither
+     links (`rustup target add aarch64-pc-windows-msvc`; ~30 s warm for
+     `-p mssqlodbc --lib`). It cannot run a Windows test binary, so runtime behaviour
+     there still rests on the CI matrix — but "my host is Linux, so the matrix owns
+     this" is the wrong default when reviewing platform-gated code, and a mutation
+     applied to an arm your host does not select proves nothing on its own. A failure
+     here can still come from a dependency's build script rather than the reviewed
+     code: targeting a Linux triple pulls `openssl-sys` in through `native-tls`, and
+     its build script fails before rustc ever sees the crate if it can't find a target
+     OpenSSL sysroot — confirmed on `mssql-odbc`, where the Windows-target direction
+     above needs no OpenSSL at all (`native-tls` routes through Schannel on Windows,
+     Security.framework on macOS — only Linux targets hit the sysroot lookup). That is
+     an environment gap, not a type error; read what actually failed before
+     attributing it to the diff.
 
    ```bash
    cargo nextest run -p <affected-crate> --lib --no-fail-fast   # or `cargo btest`
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -22,8 +22,8 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    is worth it when you can: it catches a PR that drifts from what its work item
    asked, or one still open against a closed item. Attempt the lookup before treating
    it as out of reach — on an already-authenticated host, the Azure DevOps MCP server
-   answers `wit_work_item` without an interactive prompt. See step 6 for what to do
-   when a call actually fails, and for unattended runs.
+   answers `wit_work_item` without an interactive prompt (verified 2026-09-02). See
+   step 6 for what to do when a call actually fails, and for unattended runs.
 2. **Check the PR out locally.** A diff alone is not enough to review this codebase —
    most defects here turn on unchanged code (the other implementer of a trait, the
    caller three layers up, the `#[cfg]` variant of a constant). Use a dedicated
@@ -385,11 +385,11 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
   "I could not check X" only after invoking the thing and recording what happened —
   quote the response when one comes back, or state that none did (a timeout after a
   bounded wait, a hard error) when it doesn't; a silent hang is itself an outcome, not
-  an excuse to skip the record. Three of the four caveats closing one recent review —
-  no ADO access, no Windows host, Linux-only mutation scope — did not survive being
-  tested. Keep two categories apart: *the environment cannot do this* needs a failed
-  invocation, while *the defect is inherently untestable* (process teardown, a TOCTOU
-  window) needs only an argument and stays valid.
+  an excuse to skip the record. Three of the four caveats closing one review
+  (2026-09-02, AB#47807) — no ADO access, no Windows host, Linux-only mutation scope —
+  did not survive being tested. Keep two categories apart: *the environment cannot do
+  this* needs a failed invocation, while *the defect is inherently untestable* (process
+  teardown, a TOCTOU window) needs only an argument and stays valid.
 - If a change is correct, don't invent problems. An empty severity group means "none
   found" — say so briefly.
 - Reviewing is not merging. The PR author owns the merge — never merge someone
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -21,9 +21,9 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    Azure DevOps work item — flag a PR that has neither. Resolving an `AB#` reference
    is worth it when you can: it catches a PR that drifts from what its work item
    asked, or one still open against a closed item. Attempt the lookup before treating
-   it as out of reach — the Azure DevOps MCP server answers `wit_work_item` without
-   an interactive prompt. See step 6 for what to do when a call actually fails, and
-   for unattended runs.
+   it as out of reach — on an already-authenticated host, the Azure DevOps MCP server
+   answers `wit_work_item` without an interactive prompt. See step 6 for what to do
+   when a call actually fails, and for unattended runs.
 2. **Check the PR out locally.** A diff alone is not enough to review this codebase —
    most defects here turn on unchanged code (the other implementer of a trait, the
    caller three layers up, the `#[cfg]` variant of a constant). Use a dedicated
@@ -96,12 +96,19 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      show the suite still passes.
    - **Does the platform-gated half still compile?** `cargo check` / `cargo clippy
      --target <triple>` type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]`
-     code from any host, because neither links (`rustup target add
+     code without needing a matching host, because neither links (`rustup target add
      aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). It cannot run a
      Windows test binary, so runtime behaviour there still rests on the CI matrix —
      but "my host is Linux, so the matrix owns this" is the wrong default when
      reviewing platform-gated code, and a mutation applied to an arm your host does
-     not select proves nothing on its own.
+     not select proves nothing on its own. A failure here can still come from a
+     dependency's build script rather than the reviewed code: targeting a non-Windows
+     triple pulls `openssl-sys` in through `native-tls`, and its build script fails
+     before rustc ever sees the crate if it can't find a target OpenSSL sysroot —
+     confirmed on `mssql-odbc`, where the Windows-target direction above needs no
+     OpenSSL at all (`native-tls` routes through Schannel there). That is an
+     environment gap, not a type error; read what actually failed before attributing
+     it to the diff.
 
    ```bash
    cargo nextest run -p <affected-crate> --lib --no-fail-fast   # or `cargo btest`
@@ -375,12 +382,14 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
 - Distinguish facts (verified in code) from concerns (worth checking). Don't state
   guesses as defects. Say what you ran and what you read.
 - **An unavailability is a claim, and it carries the same burden as a defect.** Write
-  "I could not check X" only after invoking the thing and quoting what came back.
-  Three of the four caveats closing one recent review — no ADO access, no Windows
-  host, Linux-only mutation scope — did not survive being tested. Keep two categories
-  apart: *the environment cannot do this* needs a failed invocation, while *the defect
-  is inherently untestable* (process teardown, a TOCTOU window) needs only an argument
-  and stays valid.
+  "I could not check X" only after invoking the thing and recording what happened —
+  quote the response when one comes back, or state that none did (a timeout after a
+  bounded wait, a hard error) when it doesn't; a silent hang is itself an outcome, not
+  an excuse to skip the record. Three of the four caveats closing one recent review —
+  no ADO access, no Windows host, Linux-only mutation scope — did not survive being
+  tested. Keep two categories apart: *the environment cannot do this* needs a failed
+  invocation, while *the defect is inherently untestable* (process teardown, a TOCTOU
+  window) needs only an argument and stays valid.
 - If a change is correct, don't invent problems. An empty severity group means "none
   found" — say so briefly.
 - Reviewing is not merging. The PR author owns the merge — never merge someone
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -20,8 +20,10 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    missing or doesn't match the diff. This repo requires a linked GitHub issue or
    Azure DevOps work item — flag a PR that has neither. Resolving an `AB#` reference
    is worth it when you can: it catches a PR that drifts from what its work item
-   asked, or one still open against a closed item. See step 6 for handling ADO in an
-   unattended run.
+   asked, or one still open against a closed item. Attempt the lookup before treating
+   it as out of reach — the Azure DevOps MCP server answers `wit_work_item` without
+   an interactive prompt. See step 6 for what to do when a call actually fails, and
+   for unattended runs.
 2. **Check the PR out locally.** A diff alone is not enough to review this codebase —
    most defects here turn on unchanged code (the other implementer of a trait, the
    caller three layers up, the `#[cfg]` variant of a constant). Use a dedicated
@@ -92,6 +94,14 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
+   - **Does the platform-gated half still compile?** `cargo check` / `cargo clippy
+     --target <triple>` type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]`
+     code from any host, because neither links (`rustup target add
+     aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). It cannot run a
+     Windows test binary, so runtime behaviour there still rests on the CI matrix —
+     but "my host is Linux, so the matrix owns this" is the wrong default when
+     reviewing platform-gated code, and a mutation applied to an arm your host does
+     not select proves nothing on its own.
 
    ```bash
    cargo nextest run -p <affected-crate> --lib --no-fail-fast   # or `cargo btest`
@@ -120,18 +130,28 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    - notes in the body that it came from an unattended run, so the author knows the
      findings were not checked by a human first.
    - never merges, and never resolves a thread it did not open.
-   - treats every interactive authentication path as unavailable, and prefers a tool
-     that fails loudly over one that waits politely. The Azure DevOps MCP server is the
-     known trap: its OAuth flow blocks on a browser nobody will open, and the run keeps
-     reporting itself as healthy while it hangs. Use whatever non-interactive ADO access
-     you have instead, bound it with a timeout, and mark ADO unavailable for the rest of
-     the run on the first failure rather than retrying per PR.
+   - treats every *interactive* authentication path as unavailable, and prefers a tool
+     that fails loudly over one that waits politely. The hazard belongs to the context
+     rather than to any one server: a flow that would open a browser has nobody to
+     answer it, and the run can keep reporting itself as healthy while it hangs. Bound
+     each such call with a timeout, and on a real failure mark that dependency
+     unavailable for the rest of the run instead of retrying per PR.
 
-   **Fail open.** ADO is context, not a gate: it confirms a PR does what its work item
-   asked. When it is unreachable, take an `AB#<number>` at face value as satisfying the
-   linked-work-item requirement in step 1 and review normally. Report the skipped
-   cross-check in the run log, not in the PR — a reviewer's infrastructure trouble is
-   not the author's problem.
+     This does **not** describe the Azure DevOps MCP server in ordinary use. It answers
+     `wit_work_item` and `wit_query` non-interactively on an already-authenticated
+     host — verified 2026-09-02, an `action: get` on a work item returned immediately
+     with no prompt and no hang. Earlier revisions of this file named it as a known
+     trap; that has since been read as a standing fact about the tool and has cost
+     real cross-checks. Call it, then decide.
+
+   **Fail open — after a failed call, not instead of one.** ADO is context, not a
+   gate: it confirms a PR does what its work item asked. When a lookup *actually*
+   fails, take an `AB#<number>` at face value as satisfying the linked-work-item
+   requirement in step 1, review normally, and report the skipped cross-check in the
+   run log rather than in the PR — a reviewer's infrastructure trouble is not the
+   author's problem. Reaching for this paragraph without having attempted the lookup
+   is the failure mode it exists to bound, and it silently drops the one check that
+   catches a PR drifting from what its work item asked.
 7. Ground yourself in reference code and public/private documentation/specifications.
    If you don't know the codebase, or which references to use, ask for context before
    reviewing.
@@ -354,6 +374,13 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
   that contradiction *is* the finding.
 - Distinguish facts (verified in code) from concerns (worth checking). Don't state
   guesses as defects. Say what you ran and what you read.
+- **An unavailability is a claim, and it carries the same burden as a defect.** Write
+  "I could not check X" only after invoking the thing and quoting what came back.
+  Three of the four caveats closing one recent review — no ADO access, no Windows
+  host, Linux-only mutation scope — did not survive being tested. Keep two categories
+  apart: *the environment cannot do this* needs a failed invocation, while *the defect
+  is inherently untestable* (process teardown, a TOCTOU window) needs only an argument
+  and stays valid.
 - If a change is correct, don't invent problems. An empty severity group means "none
   found" — say so briefly.
 - Reviewing is not merging. The PR author owns the merge — never merge someone
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -111,18 +111,19 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      Either way, confirm the specific CI job for that platform actually ran and passed
      on the PR's head commit rather than assuming "the matrix owns this" without
      looking; a local check only adds a faster answer to the same question CI already
-     gates. Neither approach can run a Windows test binary, and a mutation applied to
-     an arm your invocation doesn't select — the wrong target, or `--lib` over
-     `--all-targets` — proves nothing on its own; that still rests on CI. A local
-     check can also fail in a dependency's build script rather than the reviewed
-     code: targeting a non-Windows, non-macOS triple pulls in `openssl-sys` — through
-     `native-tls`'s backend choice and, in `mssql-tds`, through a direct `openssl`
-     dependency for the Always Encrypted primitives — and its build script fails
-     before rustc ever sees the crate if it can't find a target OpenSSL sysroot.
-     Confirmed on `mssql-tds`: the Windows-target direction above needs none at all
-     (`native-tls` routes through Schannel on Windows, Security.framework on macOS).
-     That is an environment gap, not a type error; read what actually failed before
-     attributing it to the diff.
+     gates, since neither `cargo check` nor `cargo clippy` executes a compiled test
+     binary — only CI's Windows job actually runs the `#[cfg(windows)]` tests. A
+     mutation applied to an arm your invocation doesn't select — the wrong target, or
+     `--lib` over `--all-targets` — proves nothing on its own. A local check can also
+     fail in a dependency's build script rather than the reviewed code: targeting a
+     non-Windows, non-macOS triple pulls in `openssl-sys` — through `native-tls`'s
+     backend choice and, in `mssql-tds`, through a direct `openssl` dependency for the
+     Always Encrypted primitives — and its build script fails before rustc ever sees
+     the crate if it can't find a target OpenSSL sysroot. Confirmed on `mssql-tds`:
+     the Windows-target direction above needs none at all (`native-tls` routes
+     through Schannel on Windows, Security.framework on macOS). That is an
+     environment gap, not a type error; read what actually failed before attributing
+     it to the diff.
 
    ```bash
    cargo nextest run -p <affected-crate> --lib --no-fail-fast   # or `cargo btest`
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -99,21 +99,29 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      available, `cargo check --target <triple>` / `cargo clippy --target <triple>`
      type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code without a
      matching host, because neither links (`rustup target add
-     aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). This is a
-     convenience, not a requirement — installing a target and building locally isn't
-     something every reviewer wants or needs to do. Either way, confirm the specific
-     CI job for that platform actually ran and passed on the PR's head commit rather
-     than assuming "the matrix owns this" without looking; a local check only adds a
-     faster answer to the same question CI already gates. Neither approach can run a
-     Windows test binary, and a mutation applied to an arm your host does not select
-     proves nothing on its own — that still rests on CI. A local check can also fail
-     in a dependency's build script rather than the reviewed code: targeting a Linux
-     triple pulls `openssl-sys` in through `native-tls`, and its build script fails
-     before rustc ever sees the crate if it can't find a target OpenSSL sysroot —
-     confirmed on `mssql-odbc`, where the Windows-target direction above needs no
-     OpenSSL at all (`native-tls` routes through Schannel on Windows,
-     Security.framework on macOS — only Linux targets hit the sysroot lookup). That is
-     an environment gap, not a type error; read what actually failed before
+     aarch64-pc-windows-msvc`; ~26 s warm for `-p mssql-tds --all-targets`). Use
+     `--all-targets`, not `--lib`: most of this repo's platform-gated code lives
+     inside inline `#[cfg(test)] mod tests` blocks, which `--lib` compiles without the
+     `test` cfg and so skips silently, `--target` notwithstanding — confirmed by
+     mutating a `#[cfg(windows)]` test in `datasource_parser.rs`, which `--lib` missed
+     and `--all-targets` caught. `mssql-tds` is also the crate to reach for here, not
+     `mssql-odbc`: it carries roughly 120 `cfg(windows)`/`cfg(target_os)` sites against
+     `mssql-odbc`'s ~18. This is a convenience, not a requirement — installing a
+     target and building locally isn't something every reviewer wants or needs to do.
+     Either way, confirm the specific CI job for that platform actually ran and passed
+     on the PR's head commit rather than assuming "the matrix owns this" without
+     looking; a local check only adds a faster answer to the same question CI already
+     gates. Neither approach can run a Windows test binary, and a mutation applied to
+     an arm your invocation doesn't select — the wrong target, or `--lib` over
+     `--all-targets` — proves nothing on its own; that still rests on CI. A local
+     check can also fail in a dependency's build script rather than the reviewed
+     code: targeting a non-Windows, non-macOS triple pulls in `openssl-sys` — through
+     `native-tls`'s backend choice and, in `mssql-tds`, through a direct `openssl`
+     dependency for the Always Encrypted primitives — and its build script fails
+     before rustc ever sees the crate if it can't find a target OpenSSL sysroot.
+     Confirmed on `mssql-tds`: the Windows-target direction above needs none at all
+     (`native-tls` routes through Schannel on Windows, Security.framework on macOS).
+     That is an environment gap, not a type error; read what actually failed before
      attributing it to the diff.
 
    ```bash
@@ -393,10 +401,10 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
   response when one comes back, state that none did (a timeout after a bounded wait, a
   hard error) when it doesn't, or say the tool isn't exposed in this session's
   inventory at all when there was nothing to call — a silent hang or a missing tool is
-  an outcome, not an excuse to skip the record. Three of the four caveats closing one
-  review (2026-09-02, AB#47807) — no ADO access, type-checking Windows-gated code
-  without a Windows host, Linux-only mutation scope — did not survive being tested;
-  running a Windows test binary did. Keep two categories apart: *the environment
+  an outcome, not an excuse to skip the record. Three of the four caveats closing a
+  2026-09-02 review of #459 (AB#47509) — no ADO access, type-checking Windows-gated
+  code without a Windows host, Linux-only mutation scope — did not survive being
+  tested; running a Windows test binary did. Keep two categories apart: *the environment
   cannot do this* needs a failed invocation or a confirmed absence, while *the defect
   is inherently untestable* (process teardown, a TOCTOU window) needs only an argument
   and stays valid.
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -95,19 +95,23 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
-   - **Does the platform-gated half still compile?** `cargo check --target <triple>` /
-     `cargo clippy --target <triple>` type-checks `#[cfg(windows)]` and
-     `#[cfg(target_os = ...)]` code without needing a matching host, because neither
-     links (`rustup target add aarch64-pc-windows-msvc`; ~30 s warm for
-     `-p mssqlodbc --lib`). It cannot run a Windows test binary, so runtime behaviour
-     there still rests on the CI matrix — but "my host is Linux, so the matrix owns
-     this" is the wrong default when reviewing platform-gated code, and a mutation
-     applied to an arm your host does not select proves nothing on its own. A failure
-     here can still come from a dependency's build script rather than the reviewed
-     code: targeting a Linux triple pulls `openssl-sys` in through `native-tls`, and
-     its build script fails before rustc ever sees the crate if it can't find a target
-     OpenSSL sysroot — confirmed on `mssql-odbc`, where the Windows-target direction
-     above needs no OpenSSL at all (`native-tls` routes through Schannel on Windows,
+   - **Does the platform-gated half still compile?** Where the toolchain is already
+     available, `cargo check --target <triple>` / `cargo clippy --target <triple>`
+     type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code without a
+     matching host, because neither links (`rustup target add
+     aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). This is a
+     convenience, not a requirement — installing a target and building locally isn't
+     something every reviewer wants or needs to do. Either way, confirm the specific
+     CI job for that platform actually ran and passed on the PR's head commit rather
+     than assuming "the matrix owns this" without looking; a local check only adds a
+     faster answer to the same question CI already gates. Neither approach can run a
+     Windows test binary, and a mutation applied to an arm your host does not select
+     proves nothing on its own — that still rests on CI. A local check can also fail
+     in a dependency's build script rather than the reviewed code: targeting a Linux
+     triple pulls `openssl-sys` in through `native-tls`, and its build script fails
+     before rustc ever sees the crate if it can't find a target OpenSSL sysroot —
+     confirmed on `mssql-odbc`, where the Windows-target direction above needs no
+     OpenSSL at all (`native-tls` routes through Schannel on Windows,
      Security.framework on macOS — only Linux targets hit the sysroot lookup). That is
      an environment gap, not a type error; read what actually failed before
      attributing it to the diff.
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -23,7 +23,8 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    asked, or one still open against a closed item. Attempt the lookup before treating
    it as out of reach — on an already-authenticated host, the Azure DevOps MCP server
    answers `wit_work_item` without an interactive prompt (verified 2026-09-02). See
-   step 6 for what to do when a call actually fails, and for unattended runs.
+   step 6 for what to do when a call actually fails or isn't exposed in this session
+   at all, and for unattended runs.
 2. **Check the PR out locally.** A diff alone is not enough to review this codebase —
    most defects here turn on unchanged code (the other implementer of a trait, the
    caller three layers up, the `#[cfg]` variant of a constant). Use a dedicated
@@ -102,13 +103,13 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      but "my host is Linux, so the matrix owns this" is the wrong default when
      reviewing platform-gated code, and a mutation applied to an arm your host does
      not select proves nothing on its own. A failure here can still come from a
-     dependency's build script rather than the reviewed code: targeting a non-Windows
+     dependency's build script rather than the reviewed code: targeting a Linux
      triple pulls `openssl-sys` in through `native-tls`, and its build script fails
      before rustc ever sees the crate if it can't find a target OpenSSL sysroot —
      confirmed on `mssql-odbc`, where the Windows-target direction above needs no
-     OpenSSL at all (`native-tls` routes through Schannel there). That is an
-     environment gap, not a type error; read what actually failed before attributing
-     it to the diff.
+     OpenSSL at all (`native-tls` routes through Schannel on Windows, Security.framework
+     on macOS — only Linux targets hit the sysroot lookup). That is an environment gap,
+     not a type error; read what actually failed before attributing it to the diff.
 
    ```bash
    cargo nextest run -p <affected-crate> --lib --no-fail-fast   # or `cargo btest`
@@ -145,20 +146,21 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      unavailable for the rest of the run instead of retrying per PR.
 
      This does **not** describe the Azure DevOps MCP server in ordinary use. It answers
-     `wit_work_item` and `wit_query` non-interactively on an already-authenticated
-     host — verified 2026-09-02, an `action: get` on a work item returned immediately
-     with no prompt and no hang. Earlier revisions of this file named it as a known
-     trap; that has since been read as a standing fact about the tool and has cost
-     real cross-checks. Call it, then decide.
+     `wit_work_item` non-interactively on an already-authenticated host — verified
+     2026-09-02, an `action: get` on a work item returned immediately with no prompt
+     and no hang (`wit_query` hasn't been invoked, so it isn't claimed here). Earlier
+     revisions of this file named it as a known trap; that has since been read as a
+     standing fact about the tool and has cost real cross-checks. Call it, then decide.
 
    **Fail open — after a failed call, not instead of one.** ADO is context, not a
    gate: it confirms a PR does what its work item asked. When a lookup *actually*
-   fails, take an `AB#<number>` at face value as satisfying the linked-work-item
-   requirement in step 1, review normally, and report the skipped cross-check in the
-   run log rather than in the PR — a reviewer's infrastructure trouble is not the
-   author's problem. Reaching for this paragraph without having attempted the lookup
-   is the failure mode it exists to bound, and it silently drops the one check that
-   catches a PR drifting from what its work item asked.
+   fails, or the tool isn't exposed in this session's inventory at all so there is
+   nothing to call, take an `AB#<number>` at face value as satisfying the
+   linked-work-item requirement in step 1, review normally, and report the skipped
+   cross-check in the run log rather than in the PR — a reviewer's infrastructure
+   trouble is not the author's problem. Reaching for this paragraph while the tool is
+   available but unattempted is the failure mode it exists to bound, and it silently
+   drops the one check that catches a PR drifting from what its work item asked.
 7. Ground yourself in reference code and public/private documentation/specifications.
    If you don't know the codebase, or which references to use, ask for context before
    reviewing.
@@ -382,14 +384,17 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
 - Distinguish facts (verified in code) from concerns (worth checking). Don't state
   guesses as defects. Say what you ran and what you read.
 - **An unavailability is a claim, and it carries the same burden as a defect.** Write
-  "I could not check X" only after invoking the thing and recording what happened —
-  quote the response when one comes back, or state that none did (a timeout after a
-  bounded wait, a hard error) when it doesn't; a silent hang is itself an outcome, not
-  an excuse to skip the record. Three of the four caveats closing one review
-  (2026-09-02, AB#47807) — no ADO access, no Windows host, Linux-only mutation scope —
-  did not survive being tested. Keep two categories apart: *the environment cannot do
-  this* needs a failed invocation, while *the defect is inherently untestable* (process
-  teardown, a TOCTOU window) needs only an argument and stays valid.
+  "I could not check X" only after recording what happened when you tried: quote the
+  response when one comes back, state that none did (a timeout after a bounded wait, a
+  hard error) when it doesn't, or say the tool isn't exposed in this session's
+  inventory at all when there was nothing to call — a silent hang or a missing tool is
+  an outcome, not an excuse to skip the record. Three of the four caveats closing one
+  review (2026-09-02, AB#47807) — no ADO access, type-checking Windows-gated code
+  without a Windows host, Linux-only mutation scope — did not survive being tested;
+  running a Windows test binary did. Keep two categories apart: *the environment
+  cannot do this* needs a failed invocation or a confirmed absence, while *the defect
+  is inherently untestable* (process teardown, a TOCTOU window) needs only an argument
+  and stays valid.
 - If a change is correct, don't invent problems. An empty severity group means "none
   found" — say so briefly.
 - Reviewing is not merging. The PR author owns the merge — never merge someone
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -94,17 +94,18 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
-   - **Does the platform-gated half still compile?** `cargo check --target <triple>`
-     / `cargo clippy --target <triple> --all-targets` type-checks `#[cfg(windows)]`
-     and `#[cfg(target_os = ...)]` code from any host, since neither links. Use
-     `--all-targets`, not `--lib`: a substantial share of this repo's platform-gated
-     code — including platform-gated tests — lives inside inline `#[cfg(test)] mod
-     tests` blocks, which `--lib` skips silently (a different `--lib` than the
-     `nextest run --lib` below, which does run those tests). This is optional and
-     doesn't *run* anything either — still confirm the platform's CI job actually
-     passed on the head commit rather than assuming the matrix covers it. A
-     non-Windows/non-macOS triple also pulls in `openssl-sys`, whose build script can
-     fail without a target sysroot; that's an environment gap, not a finding.
+   - **Does the platform-gated half still compile?** `cargo check --target <triple>
+     --all-targets` / `cargo clippy --target <triple> --all-targets` type-checks
+     `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code from any host, since
+     neither links. Use `--all-targets`, not `--lib`: a substantial share of this
+     repo's platform-gated code — including platform-gated tests — lives inside
+     inline `#[cfg(test)] mod tests` blocks, which `--lib` skips silently (a
+     different `--lib` than the `nextest run --lib` below, which does run those
+     tests). This is optional and doesn't *run* anything either — still confirm the
+     platform's CI job actually passed on the head commit rather than assuming the
+     matrix covers it. A non-Windows/non-macOS triple also pulls in `openssl-sys`,
+     whose build script can fail without a target sysroot; that's an environment
+     gap, not a finding.
 
    ```bash
    cargo nextest run -p <affected-crate> --lib --no-fail-fast   # or `cargo btest`
--- a/.github/skills/code-review/SKILL.md
+++ b/.github/skills/code-review/SKILL.md
@@ -104,10 +104,13 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      inside inline `#[cfg(test)] mod tests` blocks, which `--lib` compiles without the
      `test` cfg and so skips silently, `--target` notwithstanding — confirmed by
      mutating a `#[cfg(windows)]` test in `datasource_parser.rs`, which `--lib` missed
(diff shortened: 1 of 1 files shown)


.github/skills/code-review/SKILL.md
```diff
@@ -20,8 +20,10 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    missing or doesn't match the diff. This repo requires a linked GitHub issue or
    Azure DevOps work item — flag a PR that has neither. Resolving an `AB#` reference
    is worth it when you can: it catches a PR that drifts from what its work item
-   asked, or one still open against a closed item. See step 6 for handling ADO in an
-   unattended run.
+   asked, or one still open against a closed item. Attempt the lookup before treating
+   it as out of reach — the Azure DevOps MCP server answers `wit_work_item` without
```
David-Engel: Step 6 qualifies this claim ("on an already-authenticated host", dated, with evidence). Step 1 does not — and step 1 is the universal step that unattended runs read too, so an unauthenticated runner meets an unconditional fact about the server before it ever reaches the qualifier. That is defect #1 from the description with the polarity flipped: the hazard was over-generalised, now the capability is. It is also undated, which the preamble argues against.

Suggested:

```markdown
   asked, or one still open against a closed item. Attempt the lookup before treating
   it as out of reach — on an authenticated host the Azure DevOps MCP server answers
   `wit_work_item` without an interactive prompt (verified 2026-09-02). See step 6 for
   what to do when a call actually fails, and for unattended runs.
```

.github/skills/code-review/SKILL.md
```diff
@@ -354,6 +374,13 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
   that contradiction *is* the finding.
 - Distinguish facts (verified in code) from concerns (worth checking). Don't state
   guesses as defects. Say what you ran and what you read.
+- **An unavailability is a claim, and it carries the same burden as a defect.** Write
+  "I could not check X" only after invoking the thing and quoting what came back.
+  Three of the four caveats closing one recent review — no ADO access, no Windows
```
David-Engel: Nit: the step 6 paragraph models this convention well — dated, with what was invoked and what came back. "one recent review" leaves a future reader unable to locate the incident and judge whether the lesson still applies. A date or PR link would cost a few words.

David-Engel (commented):

Documentation-only change: one file, `.github/skills/code-review/SKILL.md`, no Rust, build script, manifest or workflow. The diagnosis in the description holds up against the old text — the `known trap` bullet really was a standing claim about a server sitting inside an unattended-run list, and `Fail open` really was unbounded. The re-scoping is the right shape, and keeping the interactive-auth warning rather than deleting it is the right call.

Two comments, both non-blocking. No blocking findings.

### Suggestion

**Step 1 re-introduces the same defect shape it is fixing** (`SKILL.md:23-26`). Step 6 carefully qualifies the claim — *"on an already-authenticated host"*, dated, with the evidence attached. Step 1 does not: it states flatly that "the Azure DevOps MCP server answers `wit_work_item` without an interactive prompt". Step 1 is the universal step, read by attended and unattended runs alike, so an unauthenticated pipeline runner now meets an unconditional tool fact before it ever reaches step 6's qualifier. That is defect #1 from your own description with the polarity flipped: the hazard was over-generalised, and now the capability is. It is also undated, which the file's preamble specifically argues against ("every concrete number... is a dated observation, not a standing truth").

The imperative is the part that matters and it survives the fix:

```markdown
   asked, or one still open against a closed item. Attempt the lookup before treating
   it as out of reach — on an authenticated host the Azure DevOps MCP server answers
   `wit_work_item` without an interactive prompt (verified 2026-09-02). See step 6 for
   what to do when a call actually fails, and for unattended runs.
```

### Nit

**"one recent review" is unanchored** (`SKILL.md:379`). The new step 6 paragraph models the convention well — dated, with what was invoked and what came back. The Principles bullet leans on an incident a future reader cannot locate to judge whether the lesson still applies. A date or PR link would cost a few words.

### What was checked

- Read the full diff against the merge base (`acafe831..2e206a19`) and the surrounding sections of the file; grepped `.github/` for other copies of the ADO guidance — there are none, so no doc is left contradicting the new text.
- Verified the new step 5 example names a real target: `mssql-odbc/Cargo.toml` declares `name = "mssqlodbc"`, so `-p mssqlodbc --lib` is correct, and the crate is `crate-type = ["cdylib"]`, which `cargo check` / `cargo clippy` do not link — the cross-target claim is sound as written.
- `gh pr checks`: all required checks pass; the ADO pipeline reports `skipping`, consistent with the `paths-ignore` note in the description.
- Not run: the cross-target `clippy` invocation itself, and no build or test run. A single-file Markdown change cannot affect them, and CI is green — so this is a stated scope limit, not a capability I found unavailable.

.github/skills/code-review/SKILL.md
```diff
@@ -354,6 +374,13 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
   that contradiction *is* the finding.
 - Distinguish facts (verified in code) from concerns (worth checking). Don't state
   guesses as defects. Say what you ran and what you read.
+- **An unavailability is a claim, and it carries the same burden as a defect.** Write
+  "I could not check X" only after invoking the thing and quoting what came back.
```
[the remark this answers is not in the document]
Vahid-b: Agreed — as written, the requirement to "quote what came back" is unsatisfiable for exactly the timeout/hang failure mode step 6 tells unattended runs to bound and treat as unavailable. Fixed in 3e80e97e: the bullet now requires recording the invocation's outcome, quoting a response only when one exists, and calling out that a silent hang is itself a recordable outcome.

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Agreed — fixed in 06e52be2: "one recent review" now reads "one review (2026-09-02, AB#47807)", matching the dated-citation convention the neighboring step 6 paragraph already uses.

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Agreed — step 1 stated the ADO claim unconditionally while step 6 correctly hedged and dated it, reintroducing defect #1's shape with the polarity flipped. Fixed in 06e52be2: step 1 now carries the same `(verified 2026-09-02)` citation step 6 has, matching your suggested wording (kept "already-authenticated" for consistency with step 6's exact phrase).

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -120,18 +137,28 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
-   not the author's problem.
+   - treats every *interactive* authentication path as unavailable, and prefers a tool
+     that fails loudly over one that waits politely. The hazard belongs to the context
+     rather than to any one server: a flow that would open a browser has nobody to
+     answer it, and the run can keep reporting itself as healthy while it hangs. Bound
+     each such call with a timeout, and on a real failure mark that dependency
+     unavailable for the rest of the run instead of retrying per PR.
+
+     This does **not** describe the Azure DevOps MCP server in ordinary use. It answers
+     `wit_work_item` and `wit_query` non-interactively on an already-authenticated
```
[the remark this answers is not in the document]
Vahid-b: Agreed — the PR's own verification record only covers `wit_work_item` (`action: get`); bundling `wit_query` in as also-verified was exactly the untested-capability-claim this PR exists to prevent. Fixed in 583faca3: dropped `wit_query` from the verified claim and noted explicitly that it hasn't been invoked.

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +94,21 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+   - **Does the platform-gated half still compile?** `cargo check` / `cargo clippy
+     --target <triple>` type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]`
+     code without needing a matching host, because neither links (`rustup target add
+     aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). It cannot run a
+     Windows test binary, so runtime behaviour there still rests on the CI matrix —
+     but "my host is Linux, so the matrix owns this" is the wrong default when
+     reviewing platform-gated code, and a mutation applied to an arm your host does
+     not select proves nothing on its own. A failure here can still come from a
+     dependency's build script rather than the reviewed code: targeting a non-Windows
+     triple pulls `openssl-sys` in through `native-tls`, and its build script fails
```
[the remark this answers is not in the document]
Vahid-b: Confirmed — `native-tls` uses Security.framework on macOS, not OpenSSL (`mssql-odbc/plan.md:110`), so "non-Windows" overstated which targets actually hit the sysroot lookup. Fixed in 583faca3: scoped the caveat to Linux targets specifically and named Security.framework as the macOS path.

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -354,6 +381,15 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
   that contradiction *is* the finding.
 - Distinguish facts (verified in code) from concerns (worth checking). Don't state
   guesses as defects. Say what you ran and what you read.
+- **An unavailability is a claim, and it carries the same burden as a defect.** Write
+  "I could not check X" only after invoking the thing and recording what happened —
```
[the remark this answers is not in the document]
Vahid-b: Good catch — a session with no ADO tool in its inventory has nothing to invoke, so the rule as written had no satisfiable path there. Fixed in 583faca3: the Principles bullet now names "the tool isn't exposed in this session's inventory at all" as its own recordable outcome, and I mirrored the carve-out into step 1's pointer and step 6's fail-open paragraph so none of the three require an impossible call.

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,21 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
+   - **Does the platform-gated half still compile?** `cargo check` / `cargo clippy
```
saurabh500: Why the target triple ?

shiwanigupta0809:
Reviewed the diff and the current head, and checked the technical advice by running it against this repo rather than reading it. Not re-raising anything from the existing threads — the step 1 hedge, the `wit_query` over-claim, the macOS/Security.framework scoping, the missing-tool outcome, and the undated citation are all settled.

The change is well-aimed. Separating "scope leak" from "stale fact" is the right diagnosis, and edit 5 is the one that will still be doing work a year from now. The PR also practices what it adds: the `cargo btest` checklist note records the invocation, the result, and the `origin/main` control, which is exactly the burden the new Principles bullet asks for.

### Suggestion

**The new step 5 command doesn't type-check the platform-gated code the bullet is about — `--lib` excludes `#[cfg(test)]`, which is where most of this repo's Windows-only code lives.**

The bullet promises that `cargo check` / `cargo clippy --target <triple>` "type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code without needing a matching host", and exemplifies with `-p mssqlodbc --lib`. `--lib` selects only the library target and compiles it without `cfg(test)`, so every `#[cfg(windows)]` item inside a `#[cfg(test)] mod tests` block is not compiled at all.

That is not a corner case here. `mssql-tds/src/connection/datasource_parser.rs` has roughly twenty Windows-only test functions after its `#[cfg(test)]` marker at line 864 — `test_named_pipe_default_instance`, `test_lpc_local_only`, `test_localdb_parsing`, `test_action_chain_shared_memory`, and so on — plus `#[cfg(not(windows))]` counterparts asserting different action-chain lengths. `client_context.rs` and `connection_actions.rs` have the same shape. Inline `#[cfg(test)]` modules are the dominant test style in this codebase, which is why the repo's own alias is

```
cargo bclippy = cargo clippy --workspace --frozen --all-features --all-targets -- -D warnings
```

with `--all-targets`, not `--lib`.

So a reviewer who follows this bullet, sees it exit 0, and concludes the Windows half compiles has verified strictly less than they think — and the failure is invisible, because the skipped code produces no output. That is the same trap the bullet's own closing sentence warns about ("a mutation applied to an arm your host does not select proves nothing on its own"), one level down: here the arm isn't selected because the *target* wasn't built, not because the host didn't match.

`--all-targets` closes it. Worth re-timing if you do — the quoted "~30 s warm for `-p mssqlodbc --lib`" will go up once the test targets are in, and an understated number is its own small trap for the next reader deciding whether this is cheap enough to bother with.

Two smaller things that follow from the same line, both optional:

- **The exemplar crate is the thinner one.** `mssql-tds` carries the platform-gated surface — named pipes, shared memory, LocalDB, `win_tls`, roughly 118 `cfg(windows)` / `cfg(target_os)` sites — while `mssql-odbc`'s is essentially the `auth` module. Naming `mssql-tds` in the example would put a reviewer on the crate where this actually pays.
- **The verification record inherits the gap.** "`cargo clippy --target aarch64-pc-windows-msvc -p mssqlodbc --lib` exits 0 in ~26 s and **does** compile the `cfg(windows)` paths" is true of the lib's paths and not of the test ones, so by the standard this PR itself introduces the claim is slightly wider than what was invoked. Cheap to make exact, and doing so is a nice demonstration of the rule in the same PR that adds it.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,21 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     --target <triple>` type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]`
+     code without needing a matching host, because neither links (`rustup target add
+     aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). It cannot run a
```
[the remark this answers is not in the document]
Vahid-b: Good catch — `cargo check` was shown bare while `cargo clippy` had `--target` attached, so the check-only path silently defaulted to the host target and would only exercise the Linux-selected arm. Fixed in 291763: both commands now explicitly carry `--target <triple>`.

> [!NOTE]

shiwanigupta0809 (commented):
_Reviewed — see https://github.com/microsoft/mssql-rs/pull/477#issuecomment-5514364606_

.github/skills/code-review/SKILL.md
saurabh500: Will this require additional software installation on windows and linux on dev machines? if, yes, then please use a different strategy. 
Also, why put the burden on the developer machine? This is what CI is supposed to do

.github/skills/code-review/SKILL.md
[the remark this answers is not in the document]
Vahid-b: Good catch — `cargo check` was shown bare while `cargo clippy` had `--target` attached, so the check-only path silently defaulted to the host target and would only exercise the Linux-selected arm. Fixed in 29176345: both commands now explicitly carry `--target <triple>`.

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,21 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     Windows test binary, so runtime behaviour there still rests on the CI matrix —
+     but "my host is Linux, so the matrix owns this" is the wrong default when
+     reviewing platform-gated code, and a mutation applied to an arm your host does
+     not select proves nothing on its own. A failure here can still come from a
+     dependency's build script rather than the reviewed code: targeting a Linux
+     triple pulls `openssl-sys` in through `native-tls`, and its build script fails
```
saurabh500: We are asking the reviewer to build and test locally, which is what the CI gates. Can we do away with this requirement? 

The problem is that these instructions may require software to be installed which may not be desirable for all developer scenarios. 

Perhaps the fallback should be to add a review note to validate a certain stage in the pipeline is failing or not.

.github/skills/code-review/SKILL.md
Vahid-b: Because without it, `cargo check`/`clippy` only evaluates your own host's cfg selection — on a Linux host, `#[cfg(windows)]` items are never even considered, so a bare `cargo check` proves nothing about them. `--target <triple>` is what makes rustc actually evaluate the *other* platform's cfg-gated arms.

That said, per your point below, I've reframed this in 9fe1973e: it's now explicitly a convenience when the toolchain's already at hand, not a requirement.

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Agreed — reframed in 9fe1973e: this is now explicitly a convenience available when the toolchain is already at hand, not a requirement to install anything. The fallback is to confirm the specific CI job for that platform actually ran and passed on the PR's head commit, rather than assuming the matrix covers it without checking either way.

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,26 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
+   - **Does the platform-gated half still compile?** Where the toolchain is already
+     available, `cargo check --target <triple>` / `cargo clippy --target <triple>`
+     type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code without a
+     matching host, because neither links (`rustup target add
+     aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). This is a
```
Theekshna: [Suggestion] The exemplar command doesn't type-check the platform-gated code this bullet is about — restating and independently confirming shiwanigupta0809's still-open top-level comment (posted 18:25:23Z, before the last push at 18:40:45Z, and not addressed by it).\n\n--lib selects only the library target and does not set cfg(test), so a #[cfg(windows)] item inside a #[cfg(test)] mod tests block is never compiled by cargo check --lib / cargo clippy --lib, regardless of --target. I verified this is not a corner case here: mssql-tds/src/connection/datasource_parser.rs has a #[cfg(test)] module starting at line 864, and roughly twenty #[cfg(windows)]/#[cfg(not(windows))] items after it (901, 913, 924, 957, 969, 995, 1075, ... 1597) — all inside the test module, none reachable from a --lib build. client_context.rs and connection_actions.rs have the same shape (10 and 27 cfg(windows)/cfg(target_os) sites respectively). The repo's own cargo bclippy alias uses --all-targets for exactly this reason.\n\nSo a reviewer following this bullet as written verifies less than the surrounding prose promises, silently — the skipped code produces no output either way. --all-targets (and, per shiwanigupta0809, mssql-tds rather than mssqlodbc as the exemplar crate, since that's where the platform-gated surface actually is) would close it.

.github/skills/code-review/SKILL.md
```diff
@@ -354,6 +388,18 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
   that contradiction *is* the finding.
 - Distinguish facts (verified in code) from concerns (worth checking). Don't state
   guesses as defects. Say what you ran and what you read.
+- **An unavailability is a claim, and it carries the same burden as a defect.** Write
+  "I could not check X" only after recording what happened when you tried: quote the
+  response when one comes back, state that none did (a timeout after a bounded wait, a
+  hard error) when it doesn't, or say the tool isn't exposed in this session's
+  inventory at all when there was nothing to call — a silent hang or a missing tool is
+  an outcome, not an excuse to skip the record. Three of the four caveats closing one
+  review (2026-09-02, AB#47807) — no ADO access, type-checking Windows-gated code
```
Theekshna: [Suggestion] This citation points at the wrong work item, undermining the exact convention this bullet asks for (SKILL.md's own preamble: dated, locatable evidence).\n\nThe caveats described — no ADO access, type-checking Windows-gated code without a Windows host, Linux-only mutation scope, contrasted with a Windows test binary being the one that survived — are Vahid-b's own 2026-09-02 review of **PR #459** (mssql-odbc: shut down the ENV runtime without joining its worker thread), which is tracked under **AB#47509** (linked in that PR's description and body). AB#47807 is *this* PR's own work-item number (used in every commit trailer here and in this PR's own 'Related Issues' section) — a different item entirely.\n\nA future reader trying to locate 'the review with three of four caveats' by searching AB#47807 won't find it; AB#47509 (or a link to PR #459) would. Worth an easy fix given the bullet's own subject is exactly this kind of checkable, locatable citation.

Theekshna (commented):
Unattended review. Documentation-only change to `.github/skills/code-review/SKILL.md` — diagnosis (a scope leak, not a stale fact) is sound, the re-scoping of the ADO/interactive-auth hazard from "server" to "context" is correct, and all 9 inline review threads (David-Engel, saurabh500, Copilot bot) are resolved. Two open items remain from the top-level discussion / my own read, both non-blocking.

| Category | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 2 |
| Nit | 0 |

Note on how ADO is handled in this review itself: per this run's own operating instructions, I do not invoke the Azure DevOps MCP server at all in unattended sweeps and fail open on `AB#47807` at face value — independent of what this PR proposes for attended reviews. That does not bear on whether the PR's content is correct.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,26 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     aarch64-pc-windows-msvc`; ~30 s warm for `-p mssqlodbc --lib`). This is a
+     convenience, not a requirement — installing a target and building locally isn't
+     something every reviewer wants or needs to do. Either way, confirm the specific
+     CI job for that platform actually ran and passed on the PR's head commit rather
+     than assuming "the matrix owns this" without looking; a local check only adds a
+     faster answer to the same question CI already gates. Neither approach can run a
+     Windows test binary, and a mutation applied to an arm your host does not select
+     proves nothing on its own — that still rests on CI. A local check can also fail
+     in a dependency's build script rather than the reviewed code: targeting a Linux
+     triple pulls `openssl-sys` in through `native-tls`, and its build script fails
```
David-Engel: **Nit — attribution.** `openssl-sys` also arrives here through a direct `openssl` dependency, not only via `native-tls`: `mssql-tds/Cargo.toml:93-94` declares it under `cfg(all(not(windows), not(target_os = "macos")))` for the Always Encrypted primitives. `cargo tree --offline -p mssqlodbc --target x86_64-unknown-linux-gnu -i openssl-sys --depth 1` shows both parents (`native-tls v0.2.18` and `openssl v0.10.81`). So the sysroot lookup isn't a `native-tls` artifact, and it applies to every non-Windows/non-macOS target (musl, the BSDs), not Linux specifically. The bullet's actual point — build-script failure is an environment gap, not a type error — still holds.

David-Engel (commented):

## Summary

Documentation-only change to `.github/skills/code-review/SKILL.md` (+60/−14, one file). The re-scoping is done correctly and completely: the interactive-auth hazard now attaches to the unattended *context* rather than to the ADO MCP server, fail-open is bounded to "after a failed call", and step 1 carries the same dated hedge step 6 does. I grepped the whole file for leftover ADO/availability guidance — there is no stale copy of the old "known trap" or unconditional fail-open wording anywhere else, so the scope leak really is closed rather than moved.

No blocking issues. One nit on a dependency-attribution detail in the new step 5 bullet. The two open threads from @shiwanigupta0809 / @Theekshna (`--lib` vs `--all-targets` for the platform-gated exemplar, and the `AB#47807` citation) are still unaddressed on this head and are not re-filed here.

## Verification

- Diffed `origin/main...9fe1973e` and read the full changed file, plus the unchanged sections around each hunk.
- Read all 12 review threads and the top-level comments before writing, to avoid re-filing settled findings.
- `-p mssqlodbc` is correct: `mssql-odbc/Cargo.toml` declares `name = "mssqlodbc"`, so the exemplar invocation is valid as written.
- Checked the CI-fallback advice against reality: ADO does surface per-job statuses on this repo (`mssql-rs Pull request validation (Build Stage Build Windows)` / `(Build Windows ARM)` on PRs #455 and #463), so "confirm the specific CI job for that platform actually ran and passed" is satisfiable from `gh pr checks` alone. That fallback is sound.
- `cargo tree --offline -p mssqlodbc --target x86_64-unknown-linux-gnu -i openssl-sys --depth 1` → `openssl-sys v0.9.117` with **two** parents: `native-tls v0.2.18` *and* `openssl v0.10.81`. See the nit.
- Required checks are green on this head (`Analyze`, `CodeQL`, `check`, `license/cla`); the ADO pipeline is `skipping`, consistent with the `.github/**` path filter the PR body cites.
- No Rust, manifest, or build-script surface changed, so no build/test run was warranted beyond the dependency-graph query above.

## Blocking

None.

## Suggestion

None.

## Nit

**1. `.github/skills/code-review/SKILL.md:110-115` — `openssl-sys` does not arrive on a Linux target only through `native-tls`.** `mssql-tds/Cargo.toml:93-94` declares `openssl` directly under `[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]` for the Always Encrypted crypto primitives, independent of TLS. The `cargo tree` above confirms both edges. Two small consequences for the sentence as written: the sysroot lookup would survive a move off `native-tls`, and it applies to every non-Windows/non-macOS target (musl/Alpine, the BSDs), not just Linux specifically. Something like "targeting a non-Windows, non-macOS triple pulls in `openssl-sys` — through `native-tls` and, in `mssql-tds`, through a direct `openssl` dependency for the Always Encrypted primitives" keeps the guidance and the attribution both accurate. The actionable point of the bullet — a build-script failure is an environment gap, not a type error in the diff — is unaffected either way.

.github/skills/code-review/SKILL.md
Vahid-b: Confirmed — you're right, those caveats trace to Vahid-b's 2026-09-02 review of PR #459 (its "What I could not check" section matches exactly: ADO reachability, the Windows repro, Linux-only mutation scope), tracked under AB#47509, not this PR's own AB#47807. Fixed in 3b61d1bf: now cites "a 2026-09-02 review of #459 (AB#47509)".

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Confirmed and fixed in 3b61d1bf. I mutated a `#[cfg(windows)]` test in `datasource_parser.rs` (a bogus method call) and verified directly: `--lib` compiled clean (never saw it), `--all-targets` caught it immediately with E0599. Switched the exemplar to `-p mssql-tds --all-targets` — also swapping the crate per shiwanigupta0809's point below, since `mssql-tds` carries ~120 platform-gated sites against `mssql-odbc`'s ~18. Retimed at ~26s warm.

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Confirmed via `mssql-tds/Cargo.toml:93-94` and matches your `cargo tree` output. Fixed in 3b61d1bf: now credits both paths — `native-tls`'s backend choice and the direct `openssl` dependency for Always Encrypted — and broadened the gate from "Linux" to "non-Windows, non-macOS".

> [!NOTE]

Vahid-b:
@shiwanigupta0809 thanks — confirmed and fixed in 3b61d1bf. I mutated a `#[cfg(windows)]` test in `datasource_parser.rs` to check this concretely: `--lib` compiled clean (never saw the mutation), `--all-targets` caught it immediately. Switched the exemplar to `-p mssql-tds --all-targets` and to `mssql-tds` as the crate (~120 platform-gated sites vs `mssql-odbc`'s ~18), retimed at ~26s warm. Also updated the PR's own "Verified" line and the mutation-on-unselected-arm sentence to cover the `--lib`/`--all-targets` axis, not just target-triple mismatch — appreciated the precise repro.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,34 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     `test` cfg and so skips silently, `--target` notwithstanding — confirmed by
+     mutating a `#[cfg(windows)]` test in `datasource_parser.rs`, which `--lib` missed
+     and `--all-targets` caught. `mssql-tds` is also the crate to reach for here, not
+     `mssql-odbc`: it carries roughly 120 `cfg(windows)`/`cfg(target_os)` sites against
+     `mssql-odbc`'s ~18. This is a convenience, not a requirement — installing a
+     target and building locally isn't something every reviewer wants or needs to do.
+     Either way, confirm the specific CI job for that platform actually ran and passed
+     on the PR's head commit rather than assuming "the matrix owns this" without
+     looking; a local check only adds a faster answer to the same question CI already
+     gates. Neither approach can run a Windows test binary, and a mutation applied to
```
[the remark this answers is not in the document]
Vahid-b: Confirmed — `validation-stages.yml`'s `Build_Windows` job does run real test binaries on Windows via `build-template.yml:121-124`'s `cargo llvm-cov nextest --all-targets`. Fixed in 6107f5d2: "Neither approach" is now spelled out as `cargo check`/`cargo clippy` specifically (neither executes a binary, local or in CI), and the sentence states outright that CI's Windows job is what actually runs the `#[cfg(windows)]` tests.

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,35 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code without a
+     matching host, because neither links (`rustup target add
+     aarch64-pc-windows-msvc`; ~26 s warm for `-p mssql-tds --all-targets`). Use
+     `--all-targets`, not `--lib`: most of this repo's platform-gated code lives
+     inside inline `#[cfg(test)] mod tests` blocks, which `--lib` compiles without the
+     `test` cfg and so skips silently, `--target` notwithstanding — confirmed by
+     mutating a `#[cfg(windows)]` test in `datasource_parser.rs`, which `--lib` missed
+     and `--all-targets` caught. `mssql-tds` is also the crate to reach for here, not
+     `mssql-odbc`: it carries roughly 120 `cfg(windows)`/`cfg(target_os)` sites against
+     `mssql-odbc`'s ~18. This is a convenience, not a requirement — installing a
```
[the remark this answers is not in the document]
Vahid-b: Confirmed — this contradicted the existing crate-selection rule at lines 132-134 ("pick the crate the diff touches"). Fixed in 6ea53212: parenthesized as an explanation of why *this measurement* uses `mssql-tds`, with an explicit pointer back to picking whichever crate the diff actually touches.

> [!NOTE]

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot on behalf of @David-Engel. This is **not an approval** and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back where it is wrong.

## Summary

Documentation-only change to `.github/skills/code-review/SKILL.md` (+73/-14). No findings.

The core fix is sound: the old step 6 stated "the Azure DevOps MCP server is the known trap" as a standing property of the tool, and the fail-open paragraph sat right underneath it with no precondition. That pairing let a run reach for fail-open without ever calling the tool, which is exactly the failure the PR describes. Moving the hazard from the *server* to the *context* ("a flow that would open a browser has nobody to answer it"), and retitling fail-open to "**after a failed call, not instead of one**", closes it at the right level — the invariant now generalizes to any interactive dependency rather than hard-coding one server's reputation. The forward reference added in step 1 and the new "unavailability is a claim" bullet in the Principles section make the rule reachable from both entry points.

## Verification

Documentation change, so no build or test run. I checked the diff's factual claims against the worktree instead:

- **`~120` vs `~18` cfg sites.** Matches. `cfg(windows)`/`cfg(target_os)` occurrences under `mssql-tds/src` come out at 123; `mssql-odbc/src` at 18. The framing ("not because it's a crate to default to") also resolves the crate-selection concern raised on `c1e38834` — line 111 now points explicitly at the `-p mssql-tds --lib` rule further down, which is genuinely below it.
- **`--all-targets` vs `--lib`.** `datasource_parser.rs` has ~23 `#[cfg(windows)]` items inside its `#[cfg(test)] mod tests` block starting at line 864, so it is a real exemplar for the point that `--lib` skips them silently.
- **Direct `openssl` dependency.** Confirmed: `mssql-tds/Cargo.toml` declares `openssl = "0.10.66"` under `[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]` for the Always Encrypted primitives. The build-script caveat is accurate, and correctly scoped to non-Windows/non-macOS triples — the Windows-target direction the bullet recommends does not pull it in.
- **Internal consistency.** No leftover "known trap" framing survives outside the deliberate historical mention at line 169. Step 1's forward pointer, step 6's bullet, the fail-open paragraph and the Principles bullet all agree on the same precondition. Indentation renders correctly as continuation paragraphs within their list items.
- **CI.** `Analyze`, `CodeQL`, `check` and `license/cla` pass on `6ea5321`. `mssql-rs Pull request validation` reports `skipping`, which is the expected ADO path-exclude behavior for a `.github/**`-only diff and is itself documented at line 275.

## Blocking

None.

## Suggestions

None.

## Nits

None blocking. One optional readability note, take it or leave it: the closing clause at lines 409-412 — "since neither invoking `wit_work_item` nor cross-target `clippy` executes a test" — has to carry both halves of the four-caveat split at once, and the `wit_work_item` half is not really about test execution. It parses, but a reader has to back up once. Not worth a push on its own.

.github/skills/code-review/SKILL.md
```diff
@@ -20,8 +20,11 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
    missing or doesn't match the diff. This repo requires a linked GitHub issue or
    Azure DevOps work item — flag a PR that has neither. Resolving an `AB#` reference
    is worth it when you can: it catches a PR that drifts from what its work item
-   asked, or one still open against a closed item. See step 6 for handling ADO in an
-   unattended run.
+   asked, or one still open against a closed item. Attempt the lookup before treating
+   it as out of reach — on an already-authenticated host, the Azure DevOps MCP server
+   answers `wit_work_item` without an interactive prompt (verified 2026-09-02). See
+   step 6 for what to do when a call actually fails or isn't exposed in this session
+   at all, and for unattended runs.
```
David-Engel: The ADO verification now appears three times in this file — here, again at lines 165-170, and once more in the Principles bullet at 408-412, each carrying its own "verified 2026-09-02." One canonical statement is enough.

Suggest keeping the full explanation in step 6 and shortening this to a pointer:

> Resolving an `AB#` reference is worth it when you can: it catches a PR that drifts from what its work item asked, or one still open against a closed item. Attempt the lookup before treating it as out of reach — see step 6.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,38 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     `--lib` over `--all-targets` — proves nothing on its own. A local check can also
+     fail in a dependency's build script rather than the reviewed code: targeting a
+     non-Windows, non-macOS triple pulls in `openssl-sys` — through `native-tls`'s
+     backend choice and, in `mssql-tds`, through a direct `openssl` dependency for the
+     Always Encrypted primitives — and its build script fails before rustc ever sees
+     the crate if it can't find a target OpenSSL sysroot. Confirmed on `mssql-tds`:
+     the Windows-target direction above needs none at all (`native-tls` routes
+     through Schannel on Windows, Security.framework on macOS). That is an
+     environment gap, not a type error; read what actually failed before attributing
+     it to the diff.
```
David-Engel: This is 32 lines, and I don't think it belongs in this PR. The stated goal is "require an invocation before a review claims a capability is unavailable" — cross-compilation mechanics are a second, independent topic that happens to share an origin story.

It also works against the rule in the file header: "Every concrete number, constant, and known-failure list below is a dated observation, not a standing truth. Prefer the command that re-derives a fact over the value written here." The `~26 s warm` timing, the `~120` vs `~18` cfg-site counts, and the `native-tls` backend routing details are all going to rot, and nobody will notice when they do.

The load-bearing content is about three sentences. Something like:

> - **Does the platform-gated half still compile?** `cargo clippy --target <triple> --all-targets` type-checks `#[cfg(windows)]` code from any host, since nothing links. Use `--all-targets`, not `--lib`: this repo's platform-gated code mostly lives in inline `#[cfg(test)] mod tests`, which `--lib` compiles without the `test` cfg and silently skips. This is optional, and it does not *run* the tests — still confirm the platform's CI job passed on the head commit. Non-Windows/macOS triples pull in `openssl-sys`, whose build script fails without a target sysroot; that's an environment gap, not a finding.

Happy to see the longer version land as its own PR if you think the detail earns its place.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,38 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
+   - **Does the platform-gated half still compile?** Where the toolchain is already
+     available, `cargo check --target <triple>` / `cargo clippy --target <triple>`
+     type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code without a
+     matching host, because neither links (`rustup target add
+     aarch64-pc-windows-msvc`; ~26 s warm for `-p mssql-tds --all-targets`). Use
```
David-Engel: Two nits in this bullet:

`aarch64-pc-windows-msvc` is presumably your host's natural pairing, but `x86_64-pc-windows-msvc` is the target most reviewers here will want.

And "use `--all-targets`, not `--lib`" sits a few lines above the code block at 132 that recommends `nextest run -p <crate> --lib`. Different commands with different reasoning, but the reader has to work that out on their own. Worth a clause reconciling them.

.github/skills/code-review/SKILL.md
```diff
@@ -120,18 +155,29 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     answer it, and the run can keep reporting itself as healthy while it hangs. Bound
+     each such call with a timeout, and on a real failure mark that dependency
+     unavailable for the rest of the run instead of retrying per PR.
+
+     This does **not** describe the Azure DevOps MCP server in ordinary use. It answers
+     `wit_work_item` non-interactively on an already-authenticated host — verified
+     2026-09-02, an `action: get` on a work item returned immediately with no prompt
+     and no hang (`wit_query` hasn't been invoked, so it isn't claimed here). Earlier
+     revisions of this file named it as a known trap; that has since been read as a
+     standing fact about the tool and has cost real cross-checks. Call it, then decide.
```
David-Engel: "Earlier revisions of this file named it as a known trap; that has since been read as a standing fact about the tool and has cost real cross-checks" is the rationale for the PR, not guidance for a reviewer. It belongs in the commit message or the PR description — a future reviewer reading this file doesn't need the history of how the instruction got here.

Same category as the parenthetical at 107-111 explaining why `mssql-tds` was chosen as the example crate: that's an answer to a prior review comment that got preserved in the artifact.

Condensed, folding this into the bullet above:

> - treats every *interactive* auth path as unavailable, and prefers a tool that fails loudly over one that waits politely: a browser flow has nobody to answer it, and the run can report itself healthy while it hangs. Bound such calls with a timeout and, on a real failure, mark that dependency unavailable for the rest of the run rather than retrying per PR. This is not a reason to skip the ADO MCP server — it answers `wit_work_item` non-interactively on an authenticated host (verified 2026-09-02). Call it, then decide.

.github/skills/code-review/SKILL.md
```diff
@@ -120,18 +155,29 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+
+   **Fail open — after a failed call, not instead of one.** ADO is context, not a
+   gate: it confirms a PR does what its work item asked. When a lookup *actually*
+   fails, or the tool isn't exposed in this session's inventory at all so there is
+   nothing to call, take an `AB#<number>` at face value as satisfying the
+   linked-work-item requirement in step 1, review normally, and report the skipped
+   cross-check in the run log rather than in the PR — a reviewer's infrastructure
+   trouble is not the author's problem. Reaching for this paragraph while the tool is
+   available but unattempted is the failure mode it exists to bound, and it silently
+   drops the one check that catches a PR drifting from what its work item asked.
```
David-Engel: These last two sentences say the same thing the heading already says ("after a failed call, not instead of one"). The heading change is the substantive bit; the closing explanation of the failure mode reads as defending the rule rather than stating it.

Trimming to the rule alone would keep the force and lose ~4 lines.

.github/skills/code-review/SKILL.md
```diff
@@ -354,6 +400,19 @@ than `gh pr review`, diff-hunk anchoring, `--paginate` when verifying — are in
...
+  hard error) when it doesn't, or say the tool isn't exposed in this session's
+  inventory at all when there was nothing to call — a silent hang or a missing tool is
+  an outcome, not an excuse to skip the record. Two of the four caveats closing a
+  2026-09-02 review of #459 (AB#47509) — no ADO access, type-checking Windows-gated
+  code without a Windows host — did not survive being tested; Linux-only mutation
+  scope and running a Windows test binary both did, since neither invoking
+  `wit_work_item` nor cross-target `clippy` executes a test. Keep two categories
+  apart: *the environment cannot do this* needs a failed invocation or a confirmed
+  absence, while *the defect is inherently untestable* (process teardown, a TOCTOU
+  window) needs only an argument and stays valid.
```
David-Engel: Good rule, and generalizing it into Principles is the right call. But 13 lines is a lot, and most of them are the #459 case study — which is a third restatement of the ADO finding plus a recap of the cross-target result from step 5.

The rule stands without the worked example:

> - **An unavailability is a claim and carries a defect's burden of proof.** Write "I could not check X" only after recording the attempt: the error, the timeout after a bounded wait, or that the tool isn't in this session's inventory at all. *The environment cannot do this* needs a failed invocation or a confirmed absence; *the defect is inherently untestable* (process teardown, a TOCTOU window) needs only an argument and stays valid.

Every other bullet in this section is 2-5 lines, so this would also match the surrounding density.

David-Engel (commented):
The behavior change is right — the old "ADO is a known trap" wording turned a per-run environment condition into a standing fact about the tool, and "attempt the call, then decide" fixes a real failure mode. Elevating *unavailability is a claim* into Principles is a good generalization of it.

My concern is density. The file grows 3,452 → 4,686 words (+36%) for what is essentially two rules, in a file that gets loaded into every review's context. A few things also read as answers to earlier review comments that got preserved in the artifact rather than as guidance a future reviewer needs.

All comments below are non-blocking suggestions.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,38 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     aarch64-pc-windows-msvc`; ~26 s warm for `-p mssql-tds --all-targets`). Use
+     `--all-targets`, not `--lib`: most of this repo's platform-gated code lives
+     inside inline `#[cfg(test)] mod tests` blocks, which `--lib` compiles without the
+     `test` cfg and so skips silently, `--target` notwithstanding — confirmed by
+     mutating a `#[cfg(windows)]` test in `datasource_parser.rs`, which `--lib` missed
+     and `--all-targets` caught. (This example measures `mssql-tds`, not `mssql-odbc`,
+     because it carries far more platform-gated surface to demonstrate against —
+     roughly 120 `cfg(windows)`/`cfg(target_os)` sites against `mssql-odbc`'s ~18 —
+     not because it's a crate to default to: check whichever crate the diff actually
+     touches, per the crate-selection rule below.) This is a convenience, not a
```
saurabh500: This resolves my main objection. The local cross-target build is now framed as a convenience — "not something every reviewer wants or needs to do" — rather than a step that gates the review or forces a toolchain install. That's the change I was asking for.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,38 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
+     requirement — installing a target and building locally isn't something every
+     reviewer wants or needs to do.
+     Either way, confirm the specific CI job for that platform actually ran and passed
```
saurabh500: And this is the fallback I suggested: confirm the platform's CI job actually ran and passed on the head commit instead of assuming the matrix owns it. Combined with the `openssl-sys` build-script caveat below, the local path is clearly optional and its failure modes are called out. Resolving my request.

Theekshna (commented):
**Automated unattended review** — generated by an authorized unattended run. This is a comment-only review and was not checked by a human before posting.

## Summary

This PR correctly re-scopes the Azure DevOps guidance in `.github/skills/code-review/SKILL.md`: fail-open is now bounded to an actual failed lookup, and the cross-target review guidance now says `--all-targets` where the earlier draft silently skipped most platform-gated test code. I reviewed the current head, the full prior inline/review/top-level discussion, the PR template, the coverage workflow, and the supporting repo files the new text cites; I did not find any new actionable issues beyond the already-existing discussion.

| Severity | Count |
|---|---:|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

Blocking: none.

Suggestion: none.

Nit: none.

Checks I specifically re-verified for this pass: the PR description still matches the `.github/**`-only diff and links `AB#47807`; the checklist is not contradicted by `gh pr checks` (`Analyze`, `CodeQL`, `check`, and `license/cla` pass, while the ADO validation pipeline is skipped consistently with `.github/workflows/pr-code-coverage.yml`'s `.github/**` path ignore); `mssql-odbc` parity is N/A because this PR does not touch ODBC surface; and there are no code changes or new tests in the diff, so the test-sufficiency / mutation-check requirement does not produce a new PR-local finding here.

Theekshna (approved):

saurabh500 (commented):
## Re-review — my pushback on the local cross-target build is addressed

Re-reviewing after the reframing in `9fe1973e` and follow-ups. My three threads all pushed the same point: don't require reviewers to install a cross-target toolchain and build locally — that's what CI gates — and give a fallback that checks the pipeline instead.

**Verdict: addressed.** The cross-target check is now explicitly a convenience rather than a required step, and the CI-stage fallback I asked for is in place. Inline confirmations below.

**Nit (optional, your call):** the step-5 addition runs ~25 lines on a single optional convenience check, which is dense for a process list a reviewer skims top-to-bottom. It reads well but could be tightened against the repo's conciseness convention. Not blocking.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +95,38 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
+   - **Does the platform-gated half still compile?** Where the toolchain is already
+     available, `cargo check --target <triple>` / `cargo clippy --target <triple>`
+     type-checks `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code without a
+     matching host, because neither links (`rustup target add
+     aarch64-pc-windows-msvc`; ~26 s warm for `-p mssql-tds --all-targets`). Use
+     `--all-targets`, not `--lib`: most of this repo's platform-gated code lives
```
saurabh500: [Suggestion] This quantifier is currently false, and it is a checkable claim of exactly the kind this file's own preamble asks us to re-derive. At head `6ea53212`, of the 120 `cfg(windows)`/`cfg(not(windows))`/`cfg(target_os)` sites in `mssql-tds/src`, **85 are in non-test code and only ~35 fall inside (or after the first) `#[cfg(test)]` marker** — so most platform-gated code is *not* in test modules. The `--all-targets`-over-`--lib` guidance itself is still correct (the `datasource_parser.rs` mutation proves `--lib` skips cfg-gated *test* code), so only the "most" framing needs fixing, e.g.:

> Use `--all-targets`, not `--lib`: a substantial share of this repo's platform-gated code — including much of the platform-gated *test* code — lives inside inline `#[cfg(test)] mod tests` blocks, which `--lib` compiles without the `test` cfg and so skips silently.

saurabh500 (commented):
_Unattended review sweep — automated, posted without a human checking the findings first. Event is COMMENT only; this is not an approval and gates nothing._

**Verdict:** No blocking findings. Docs-only change to the review skill; every factual claim I could re-derive at head `6ea53212` holds, with one exception noted inline (a "most" quantifier that is currently false). The remaining open items are the scope/length/redundancy suggestions in David-Engel's latest review — I independently agree they are reasonable and am deliberately not restating them as new threads.

**What I re-derived locally (head `6ea53212`):**
- `mssql-tds` 120 vs `mssql-odbc` 18 platform-cfg sites, matching "~120 vs ~18" — though those totals only reach 120/18 by counting `cfg(not(windows))` too; strict `cfg(windows)`+`cfg(target_os)` is 107/8. This is part of the rot-prone-numbers class David-Engel already flagged, so I am not opening a separate thread on it.
- The direct `openssl` dependency is real and gated exactly as described: `mssql-tds/Cargo.toml:93` — `[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]`.
- `datasource_parser.rs` has its `#[cfg(test)]` module at line 864 with `#[cfg(windows)]` items inside it, so the "`--lib` skips cfg-gated test code" rationale is grounded (the mutation proof already in the thread confirms it).

I could not verify, from this unattended run: the ADO non-interactive claim, the ~26s timing, or that `cargo check --target aarch64-pc-windows-msvc` succeeds on a Linux host. Those are the author's dated invocation records and I took them at face value.

**Blocking:** None.

**Suggestion:** One, inline at the "most ... lives inside inline `#[cfg(test)] mod tests`" line — that quantifier is false at this head (85 of 120 platform-cfg sites are in non-test code). No others new; David-Engel's open threads already cover the ADO-stated-three-times redundancy, the section's fit for this PR, and the rot-prone numbers.

**Nit:** None new.

**Checklist:** the unchecked `cargo btest` box is legitimate for a docs-only change — it fails identically on `origin/main` here (the transaction/connectivity suites need a live server), and `pr-code-coverage.yml`'s `paths-ignore` covers `.github/**`, so the absent Code Coverage Report comment is expected rather than a CI miss.

Theekshna (commented):
Unattended run found no new findings. Covered: msodbcsql parity (N/A — doc-only change, no ODBC surface), test sufficiency (N/A — no code), divergence docs (N/A), PR description currency (matches diff, AB#47807 linked, cargo btest checkbox note verified consistent with the file's own documented live-server caveat), and evidence audit — independently re-derived the two numeric claims motivating the crate choice: mssql-tds carries 133 cfg(windows)/cfg(target_os) sites (broad pattern incl. 
ot(windows)) against mssql-odbc's 18, both consistent with the PR's `roughly 120` / `~18`, and confirmed genuine #[cfg(windows)] tests exist inside datasource_parser.rs's inline mod tests, supporting the --lib vs --all-targets mutation claim. My prior two findings (doc-state contradiction) are resolved on this head. Did not duplicate the open, unresolved human threads already tracking real issues here (David-Engel's trim/restructure suggestions, saurabh500's still-valid catch that the `most... lives inside #[cfg(test)]` framing overstates the actual ~29% share) — those remain outstanding for the author, not re-filed.

.github/skills/code-review/SKILL.md
saurabh500: Still open — the head is unchanged at `6ea53212` (the commit I flagged), so line 103 still reads verbatim: "most of this repo's platform-gated code lives inside inline `#[cfg(test)] mod tests` blocks."

Re-derived directly against this head's tree: of the 127 `cfg(windows)`/`cfg(not(windows))`/`cfg(target_os)` attribute sites in `mssql-tds/src` (the crate this bullet measures against), **36 (28%) sit inside `#[cfg(test)]` modules and 91 (72%) do not** — the same ~70/30 split as my earlier 85/35-of-120 count. So "most … lives inside `#[cfg(test)]`" is unsupported. The `--all-targets`-over-`--lib` guidance itself is still correct and doesn't depend on "most": the `datasource_parser.rs` mutation proves `--lib` skips cfg-gated *test* code regardless of the ratio.

One forward note: David-Engel's proposed condensation of this section carries the same "this repo's platform-gated code mostly lives in inline `#[cfg(test)] mod tests`" wording, so whichever text lands still needs the fix. My proposed reword stands:

> Use `--all-targets`, not `--lib`: a substantial share of this repo's platform-gated code — including much of the platform-gated *test* code — lives inside inline `#[cfg(test)] mod tests` blocks, which `--lib` compiles without the `test` cfg and so skips silently.

.github/skills/code-review/SKILL.md
saurabh500: Head is unchanged since my acceptance (`6ea53212`), so lines 111-113 still carry the "convenience, not a requirement — installing a target and building locally isn't something every reviewer wants or needs to do" framing that resolved this. Still addressed.

.github/skills/code-review/SKILL.md
saurabh500: Head is unchanged since my acceptance (`6ea53212`), so lines 114-118 still carry the "confirm the specific CI job for that platform actually ran and passed on the PR's head commit" fallback that resolved this. Still addressed.

saurabh500 (commented):
_Unattended review sweep — these findings were not checked by a human before posting. Comment-only._

**Verdict: COMMENT — one non-blocking suggestion remains (an existing open thread).** This doc-only change to `.github/skills/code-review/SKILL.md` is sound: the ADO/interactive-auth guidance is correctly re-scoped from "server" to "context," and the `--all-targets`-over-`--lib` cross-target guidance closes a real blind spot. My two earlier objections — framing the local cross-target build as a convenience, and confirming the platform's CI job on the head commit — are addressed and remain so at this head (`6ea53212`).

**Blocking:** none.

**Suggestion:** one of my earlier inline threads (`SKILL.md:103`) is still open — the "most … lives inside `#[cfg(test)]`" quantifier remains inaccurate at this head. The evidence and proposed reword are in that thread; I've replied there rather than restating them here.

**Nit:** none.

David-Engel (changes requested):
I'd prefer to see the repetition reduced and unneeded history of the file removed before this PR is merged.

.github/skills/code-review/SKILL.md
Vahid-b: Fair, and agreed on the specifics rotting. Cut to ~11 lines in e0ceeef4: dropped the timing, the `~120` vs `~18` site counts, the mutation-test citation, and the crate-choice tangent, keeping just the rule (`--all-targets` not `--lib`, why, that it's optional, the CI fallback, and the `openssl-sys` build-script caveat).

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Agreed, cut in e0ceeef4 — the paragraph now ends at "a reviewer's infrastructure trouble is not the author's problem," dropping the two sentences that just restated the heading.

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Agreed — fixed in e0ceeef4. Step 1 now just points to step 6 ("see step 6 for what the Azure DevOps MCP server actually does...") instead of restating the fact with its own date, so `verified 2026-09-02` now appears once, in step 6.

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Agreed — that's the PR's own rationale, not reviewer guidance. Dropped in e0ceeef4; the bullet now states the rule and the one canonical `wit_work_item` fact (verified 2026-09-02) without the "earlier revisions" narration. Same fix covers the `mssql-tds` exemplar parenthetical you flagged in the same category (removed along with the rest of the step-5 trim).

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Confirmed and reworded in e0ceeef4 (as part of the broader step-5 trim). Independently recounted `mssql-tds/src` and got the same split you did: 85/120 (71%) of `cfg(windows)`/`cfg(target_os)` sites are outside test modules. "Most...lives inside `#[cfg(test)]`" is now "a substantial share...including platform-gated tests...lives inside `#[cfg(test)]`."

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Agreed — dropped the #459 case study entirely in e0ceeef4. The bullet is now 6 lines and states the rule and the two-categories distinction without a worked example, matching the 2-5 line density of its siblings.

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Both fixed in e0ceeef4 — dropped the concrete `aarch64-pc-windows-msvc` example entirely (now a generic `<triple>`, consistent with the placeholder already used earlier in the same sentence), which sidesteps the aarch64-vs-x86_64 question. Also added a parenthetical reconciling this `--lib` with the `nextest run --lib` example below: different command, different semantics (nextest's `--lib` does run the `cfg(test)` code; check/clippy's doesn't).

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Thanks for confirming. The CI-fallback also survived the latest trim (e0ceeef4), just shorter.

> [!NOTE]

.github/skills/code-review/SKILL.md
Vahid-b: Thanks for confirming. This framing survived the latest trim (e0ceeef4) — the bullet still opens with the optional/convenience wording, just shorter.

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +94,17 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
      `$BASE` and compare failure sets; only the difference belongs in the review.
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
+   - **Does the platform-gated half still compile?** `cargo check --target <triple>`
```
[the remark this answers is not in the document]
Vahid-b: Confirmed — the trim in e0ceeef4 reintroduced the exact bug fixed earlier: `--all-targets` sat only on the `clippy` example. Fixed in 55aac39d: both `cargo check --target <triple> --all-targets` and `cargo clippy --target <triple> --all-targets` now carry it.

> [!NOTE]

.github/skills/code-review/SKILL.md
```diff
@@ -120,18 +134,23 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
-   linked-work-item requirement in step 1 and review normally. Report the skipped
-   cross-check in the run log, not in the PR — a reviewer's infrastructure trouble is
-   not the author's problem.
+   - treats every *interactive* authentication path as unavailable, and prefers a tool
+     that fails loudly over one that waits politely. The hazard belongs to the context
+     rather than to any one server: a flow that would open a browser has nobody to
+     answer it, and the run can keep reporting itself as healthy while it hangs. Bound
+     each such call with a timeout, and on a real failure mark that dependency
+     unavailable for the rest of the run instead of retrying per PR. This is not a
+     reason to skip the Azure DevOps MCP server: it answers `wit_work_item`
```
David-Engel, line 143: **Suggestion — the capability claim carries a precondition the bullet gives no way to check, and the trim dropped the fallback that used to cover it.**

The verified fact is scoped precisely: `wit_work_item` answers non-interactively *on an already-authenticated host*. But this bullet lives under "An unattended run still:", and an unattended runner is exactly the context least likely to satisfy that precondition — a fresh pipeline agent has no cached ADO credential, so "Call it, then decide" is the path back into the browser-OAuth hang the same sentence warns about.

The stated mitigation is "Bound each such call with a timeout," which works for a shell invocation but generally does not for an MCP tool call: the agent has no timeout control over it, and the hang presents as a healthy-looking run. The previous revision also offered a concrete escape — *"Use whatever non-interactive ADO access you have instead"* — and that clause is gone, so what remains tells an unattended run to make the call with a mitigation it cannot apply.

One clause restores the escape without reopening anything already settled:

```markdown
     it, then decide — after confirming the precondition holds. An unattended runner
     that cannot bound the call (an MCP invocation offers no timeout control) should
     verify the host is already authenticated, or use whatever non-interactive ADO
     access it has instead, before calling.
```

Non-blocking: the fail-open paragraph directly below still yields a correct outcome if the call does hang and someone kills the run — this is about not walking into it in the first place.

.github/skills/code-review/SKILL.md
```diff
@@ -92,6 +94,18 @@ you happen to be reviewing. Skill maintenance is not that author's problem.
...
    - **Is this coverage gap real?** Introduce the bug the missing test would catch and
      show the suite still passes.
+   - **Does the platform-gated half still compile?** `cargo check --target <triple>
+     --all-targets` / `cargo clippy --target <triple> --all-targets` type-checks
+     `#[cfg(windows)]` and `#[cfg(target_os = ...)]` code from any host, since
+     neither links. Use `--all-targets`, not `--lib`: a substantial share of this
+     repo's platform-gated code — including platform-gated tests — lives inside
+     inline `#[cfg(test)] mod tests` blocks, which `--lib` skips silently (a
+     different `--lib` than the `nextest run --lib` below, which does run those
+     tests). This is optional and doesn't *run* anything either — still confirm the
```
David-Engel, line 104: **Nit — the `rustup target add <triple>` prerequisite was lost in the trim, along with the framing that resolved @saurabh500's objection.**

At `6ea53212` this bullet read "(`rustup target add <triple>` ...)" and closed with "a convenience, not a requirement — installing a target and building locally isn't something every reviewer wants or needs to do." That is the wording @saurabh500 accepted ("This resolves my main objection"). After `e0ceeef4` both are gone; "This is optional and doesn't *run* anything either" now reads as *an optional step*, not as *an optional install*.

The practical cost is small but real: `--target <triple>` needs that target's std installed, so a reviewer on a default toolchain gets `error[E0463]: can't find crate for core` before reaching the one environment caveat the bullet does keep (`openssl-sys`) — and would reasonably read a hard error as a finding.

Four words carry it:

```markdown
     tests). This is optional — it needs `rustup target add <triple>` and doesn't
     *run* anything either — still confirm the platform's CI job actually passed on
```

Not blocking, and no objection to the trim itself; the 12-line version is a clear improvement over 32.

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot in an unattended run. This is **not** an approval and does not satisfy the human review requirement. Findings were not checked by a human first; push back on anything that looks off.

Documentation-only change: one file, `.github/skills/code-review/SKILL.md`. No Rust, build script, manifest or workflow, so nothing here can move CI behavior — `Analyze`, `CodeQL` and `check` are green on `55aac39d`, and the ADO pipeline is correctly `skipping` for a `.github/**`-only diff.

What I verified rather than read: the `openssl-sys` gate is real and correctly scoped — `mssql-tds/Cargo.toml` declares `openssl` under `[target.'cfg(all(not(windows), not(target_os = "macos")))'.dependencies]`, so "non-Windows/non-macOS" is right and the earlier Linux-only phrasing would have been too narrow. The `--all-targets` vs `--lib` distinction also holds against this tree: `mssql-tds/src/connection/{client_context,connection_actions,datasource_parser}.rs` and `connection_provider/tds_connection_provider.rs` all carry platform-`cfg` items inside their inline `#[cfg(test)]` regions, which a `cargo check --lib` never compiles, and the hedged "a substantial share ... including platform-gated tests" now matches what is actually there.

The re-scoping is the right shape, and nothing in the rest of the repo contradicts the new wording — no other skill, prompt or instruction file repeats the old fail-open rule, so there is no second copy to drift.

I am not re-raising anything settled in the existing threads (the step 1 hedge, the `wit_query` over-claim, the missing-tool outcome, `--all-targets` on both commands, the `#459`/`AB#47509` citation, the CI-runs-Windows-tests correction). Both comments below are about text the last trim removed, not about the original diff.

**Blocking:** 0. **Suggestions:** 1. **Nits:** 1.

David-Engel (approved):

.github/skills/code-review/SKILL.md
David-Engel: Good enough as-is.

.github/skills/code-review/SKILL.md
David-Engel: Good enough as-is.

Issue #465, opened 2026-09-01
PR comments cancel in-flight code coverage runs and leave a false red check


David-Engel:
## Problem

Commenting on a PR cancels that PR's in-flight code coverage run and leaves a red `coverage-report` check, even though nothing is wrong with the PR.

`.github/workflows/pr-code-coverage.yml` triggers on both `pull_request` and `issue_comment` (the latter for the `/coverage` escape hatch), and both share one concurrency group:

```yaml
concurrency:
  group: pr-coverage-${{ github.event.pull_request.number || github.event.issue.number }}
  cancel-in-progress: true
```

Concurrency is evaluated at the **workflow** level, before the job-level `if:` gate. So an ordinary PR comment:

1. starts an `issue_comment` run,
2. that run joins group `pr-coverage-<n>` and cancels the in-flight `pull_request` run,
3. its own `coverage-report` job is then skipped by the `if:` (no `/coverage` in the body).

Net result: the comment cancels coverage and contributes nothing in its place.

## Evidence

Observed twice on #459 while replying to review feedback:

| Time (UTC) | Run | Event | Result |
| --- | --- | --- | --- |
| 20:25:24 | `33555221213` | `pull_request` | **cancelled** |
| 20:25:33 | `33555236763` | `issue_comment` | skipped |

The `pull_request` run was cancelled 11s in, 9s after the comment landed. The same pairing appears at 19:00:22 (`33546961046`, cancelled) / 19:02:21 (`33547161229`, skipped).

Re-running by commenting `/coverage` works, but you have to notice the red check and know why it's red first — and the natural reading is "my change broke coverage."

## Impact

- Any PR comment — a review reply, a status note, a question — can cancel coverage mid-flight.
- The cancellation surfaces as a `fail` check on the PR, which is misleading.
- More comments while the re-run is in flight cancel it again.
- It penalizes exactly the review conversation the PR workflow asks for.

## Suggested fix

Keep the cancel-in-progress behavior for pushes, but stop non-`/coverage` comments from entering the group. Options, roughly in order of preference:

1. **Put the event in the group key** so comment runs can't cancel push runs:
   ```yaml
   group: pr-coverage-${{ github.event.pull_request.number || github.event.issue.number }}-${{ github.event_name }}
   ```
   Simplest change; a `/coverage` re-run no longer cancels an in-flight push run, which is arguably also more correct.

[...]

Issue #266, opened 2026-08-13
Any PR comment silently cancels that PR's in-flight coverage report


saurabh500:
## Summary

`.github/workflows/pr-code-coverage.yml` triggers on `issue_comment: created` in addition to `pull_request`, and both event types share one `cancel-in-progress` concurrency group. As a result **any comment on a PR — not just `/coverage` — starts a new run in that group, cancels the in-flight `pull_request` run, and then immediately skips itself.** The PR is left showing a `CANCELLED` coverage check with no report.

## Root cause

Three pieces interact:

```yaml
on:
  pull_request:
    branches: [main]
  issue_comment:
    types: [created]          # <-- fires on EVERY comment

concurrency:
  group: pr-coverage-${{ github.event.pull_request.number || github.event.issue.number }}
  cancel-in-progress: true    # <-- shared group across both event types
```

and the job-level guard:

```yaml
if: >
  (github.event_name == 'pull_request' && ...) ||
  (github.event_name == 'issue_comment' &&
   github.event.issue.pull_request != null &&
   startsWith(github.event.comment.body, '/coverage') && ...)
```

Concurrency is evaluated **before** the job-level `if`. So a non-`/coverage` comment:

1. resolves to the same group `pr-coverage-<n>`,
2. cancels the running `pull_request` job (which may be mid-`Wait for ADO build`, a long step),
3. then evaluates `startsWith(body, '/coverage')` → false → skips.

Net effect: the comment destroys the report and produces nothing in its place.

## Observed on #264

| Time (UTC) | Event |
|---|---|
| 07:16:00 | `pull_request` run created ([run 31677060976](https://github.com/microsoft/mssql-rs/actions/runs/31677060976)) |
| 07:17:11 | benchmark comment posted on the PR |
| 07:17:14 | `issue_comment` run created → `completed/skipped` |
| 07:17:30 | original run → `cancelled` at step 6, *Download Cobertura coverage artifact* |

Steps 7–9 (`Compute overall coverage`, `Run diff-cover`, `Comment coverage summary`) never ran.

## Impact

- **Not a merge blocker.** `diff-cover` runs without `--fail-under`, so this is informational only.
- But it is **silent and repo-wide**: every PR that receives a comment while its coverage job is running loses that report, and the surviving red `CANCELLED` check costs a reviewer time to diagnose. The window is wide because the job polls the ADO REST API waiting for the build artifact — normally many minutes.
[...]

PR #478, opened 2026-09-02
Stop PR comments and /coverage runs cancelling in-flight coverage reports

### LINKED ISSUE #279: PR coverage run is cancelled by any PR comment, including comments the job skips
### Describe the bug

Any comment posted on a PR cancels that PR's in-flight `PR Code Coverage` run — **including comments the coverage job explicitly declines to act on**.

`.github/workflows/pr-code-coverage.yml` triggers on `issue_comment: types: [created]` with no filter at the trigger level (lines 40–42). The `/coverage` filter lives only in the job-level `if:` (line 68):

```yaml
startsWith(github.event.comment.body, '/coverage')
```

But the workflow-level `concurrency` block (lines 48–50) is keyed on the PR number alone and evaluated at **run** creation, before any job `if:` runs:

```yaml
concurrency:
  group: pr-coverage-${{ github.event.pull_request.number || github.event.issue.number }}
  cancel-in-progress: true
```

So an ordinary review comment creates a run, that run joins group `pr-coverage-<n>`, `cancel-in-progress` kills the real `pull_request` run, and only *then* does the job evaluate its `if:` and skip — having done no work but destroyed the run that was doing the work.

The guard is correct; it is just in the wrong place to prevent the cancellation.

### Steps to reproduce

1. Open a same-repo PR so the `pull_request` coverage run starts. It begins a long wait for the ADO Cobertura artifact (the workflow allows up to 1h15m; ADO builds take ~45 min).
2. While that run is waiting, post any normal comment on the PR — a review reply, not `/coverage`.
3. Observe the coverage run is cancelled within seconds, at the `Download Cobertura coverage artifact` step.

Observed on PR #264 (`4c64fd2f`):

| Time (UTC) | Event |
|---|---|
| 19:32:15 | `pull_request` run [`31736332629`](https://github.com/microsoft/mssql-rs/actions/runs/31736332629) starts on `dev/saurabh/row-decode-dispatch-perf` @ `4c64fd2f` |
| 19:35:06 | review reply comment `5285489791` posted — does **not** start with `/coverage` |
| 19:35:09 | `issue_comment` run [`31736577259`](https://github.com/microsoft/mssql-rs/actions/runs/31736577259) created on `main`, joins group `pr-coverage-264` |
| 19:35:26 | run `31736332629` **cancelled** at step 6 of 9 |
| 19:35:27 | run `31736577259` concludes **`skipped`** |

The canceller's own conclusion is `skipped`, which is the proof that concurrency is resolved before the job `if:` — a run whose only job never executed still cancelled a sibling.

Note that `issue_comment` runs execute against `main`, not the PR head. Listing runs filtered to the PR branch shows only the cancelled victim and none of the cause, which makes this awkward to diagnose from the PR page:

```
gh run list --branch dev/saurabh/row-decode-dispatch-perf --workflow "PR Code Coverage"
# -> 31736332629  pull_request  cancelled  4c64fd2f     (1 run; canceller not shown)
```

### Expected behavior

A comment that does not start with `/coverage` should have no effect on a running coverage report. Only a new push to the PR should cancel and supersede an in-flight run.

### Actual behavior

The coverage run is cancelled mid-wait and reports as a failed check on the PR (`gh pr checks` renders the `cancelled` conclusion as `fail`), even though nothing about the code or the coverage changed. Re-running it is the only recovery, and it is cancelled again by the next comment.

Because coverage waits up to 1h15m and review rounds consist of posting comments, essentially any PR under active review loses its coverage run.

### Suggested fix

Keep real runs sharing the group — a new push *should* supersede a stale wait — and shunt no-op comment runs into their own group so they cannot cancel anything:

```yaml
concurrency:
  group: >-
    pr-coverage-${{ github.event.pull_request.number || github.event.issue.number }}${{
      github.event_name == 'issue_comment'
      && !startsWith(github.event.comment.body, '/coverage')
      && format('-noop-{0}', github.run_id) || '' }}
  cancel-in-progress: true
```

This duplicates the `/coverage` predicate in two places, so an alternative is to drop `cancel-in-progress` for `issue_comment` runs, or to gate the trigger itself. Either is fine; the requirement is only that a run which will skip must not share a cancelling group with a run that will work.

### Version

`main` @ `445459bb`; workflow file `.github/workflows/pr-code-coverage.yml` unchanged since its introduction.

### Affected crate

N/A — this is CI infrastructure (`.github/workflows/pr-code-coverage.yml`), not a crate. It affects coverage reporting for every crate.

### Environment

GitHub Actions, `ubuntu-latest`. Not environment-specific.

### Additional context

Reporting-only: `diff-cover` runs without `--fail-under`, so this never blocks a merge. The cost is a red check that looks like a coverage regression, plus the loss of the diff-coverage number for the round.

Found while confirming CI on #264 — the cancellation there was caused by my own review replies.

David-Engel:
## Description

Commenting on a PR cancels that PR's in-flight code coverage run and leaves a red `coverage-report` check that reads as a coverage regression.

`concurrency` is resolved at **run creation**, before the job-level `if:` is evaluated. So an ordinary review comment starts an `issue_comment` run, that run joins the shared `pr-coverage-<n>` group, `cancel-in-progress` kills the real `pull_request` run mid-wait, and only then does the job evaluate its `if:` and skip. The canceller contributes nothing and destroys the run that was doing the work. Because the artifact wait is up to 1h15m and review rounds consist of posting comments, essentially any PR under active review loses its coverage report.

This splits the group two ways:

1. **Runs the job will skip get a unique `-noop-<run_id>` suffix**, so they share a group with nothing and cancel nothing. The predicate mirrors the full job `if:`, so forked-PR events and non-maintainer comments are covered too, not just the `/coverage` prefix.
2. **The event name joins the group key**, so the automatic `pull_request` path and the manual `/coverage` path cannot pre-empt each other.

Point 2 is worth spelling out, because it is the part that is easy to miss. `/coverage` comments and ordinary comments are *both* `issue_comment` events, and a `/coverage` run is a "real" run — so a no-op shunt alone still leaves it sharing a group with an in-flight `pull_request` run. Only the `pull_request` run's check is attached to the PR head SHA, so letting `/coverage` cancel it re-creates the same red check this guard exists to prevent, at exactly the moment someone is manually recovering the report.

Within each path `cancel-in-progress` still applies: a new push supersedes a stale wait, and a second `/coverage` supersedes the first.

## Relationship to #300

**#300 and this PR are alternatives — only one should land.** #300 (draft, opened 2026-08-15) implements point 1, and its no-op predicate is the basis for the one here. This PR adds point 2, which #300 leaves open:

| Scenario | #300 | This PR |
| --- | --- | --- |
| Ordinary comment cancels in-flight `pull_request` run | fixed | fixed |
| Forked-PR / non-maintainer no-op cancels a real run | fixed | fixed |
| **`/coverage` cancels in-flight `pull_request` run** | **still occurs** | fixed |
| New push supersedes a stale `pull_request` run | preserved | preserved |
| Second `/coverage` supersedes the first | preserved | preserved |

Happy for the maintainers to instead take #300 and add the event key there, if keeping that PR's history is preferable.

## Related Issues

Fixes #279

Consolidated #266 and #465 into #279 as duplicates; all three reported this same bug.

## Testing

No Rust code is touched, so the `cargo` checks are not applicable. Verified three ways:

**1. YAML parses and the folded scalar collapses to a single line** (a stray newline in a group key would be a silent behaviour change):

```
PARSE OK
cancel-in-progress: True
has newline: False
```

**2. `Analyze (actions)` passes** on this PR, which is CodeQL's workflow analysis over the changed file.

**3. Simulated GitHub's expression semantics** (`a && b` -> `a` if falsy else `b`; `a || b` -> `a` if truthy else `b`) over every trigger path:

| Scenario | Job runs? | Resulting group |
| --- | --- | --- |
| push to same-repo PR | yes | `pr-coverage-464-pull_request` |
| push to same-repo PR (again) | yes | `pr-coverage-464-pull_request` |
| push to forked PR | no | `pr-coverage-464-pull_request-noop-3` |
| ordinary review comment | no | `pr-coverage-464-issue_comment-noop-4` |
| another ordinary comment | no | `pr-coverage-464-issue_comment-noop-5` |
| `/coverage` by maintainer | yes | `pr-coverage-464-issue_comment` |
| `/coverage` by maintainer (2nd) | yes | `pr-coverage-464-issue_comment` |
| `/coverage` by non-maintainer | no | `pr-coverage-464-issue_comment-noop-8` |
| comment on a plain issue | no | `pr-coverage-464-issue_comment-noop-9` |

All eight derived assertions pass, including the two that must *not* regress (a new push still supersedes a stale push run; a second `/coverage` still supersedes the first).

### Why this PR cannot demonstrate the fix on itself

Two reasons, both worth stating plainly rather than implying a green check here means anything:

- `paths-ignore` excludes `.github/**`, so this PR triggers no coverage run at all.
- More fundamentally, `issue_comment`-triggered workflows always execute the **default branch** copy of the workflow file. So until this merges, comments anywhere — including here — are still governed by `main`'s current group expression. The fix cannot take effect from a PR branch.

Post-merge verification is one comment on any PR with a coverage run in flight: the run should survive, and the comment's own run should appear as `skipped` in a `-noop-` group.

## Breaking Changes / Migration Notes

None. Skipped runs still appear as `skipped` exactly as before; only group membership changes.

## Checklist

- [x] `cargo bfmt` passes (N/A — no Rust changes)
- [x] `cargo bclippy` passes (N/A — no Rust changes)
- [x] `cargo btest` passes (N/A — no Rust changes)
- [x] New/changed functionality has tests (N/A — workflow config; verified by simulation above)
- [x] Public API changes are documented (no public API changes)

### DIFF
--- a/.github/workflows/pr-code-coverage.yml
+++ b/.github/workflows/pr-code-coverage.yml
@@ -45,8 +45,33 @@ on:
 # Each job opts into only the scopes it needs.
 permissions: {}
 
+# Concurrency is resolved at run creation, before the job `if:` is evaluated, so
+# a run that will skip (an ordinary PR comment) would otherwise cancel the real
+# in-flight run and leave a red, misleading check behind. Two things keep runs
+# apart:
+#
+#   1. Runs that the `coverage-report` job will skip get a unique
+#      `-noop-<run_id>` suffix, so they share a group with nothing and cancel
+#      nothing. The predicate mirrors the job `if:` below — keep them in sync.
+#   2. The event name is part of the group, so the automatic `pull_request` path
+#      and the manual `/coverage` path never cancel each other. Both are "real"
+#      runs, but only the `pull_request` run's check is attached to the PR head
+#      SHA; letting a `/coverage` run cancel it would re-create the same red
+#      check this guard exists to prevent — precisely when someone is manually
+#      recovering a report.
+#
+# Within each path `cancel-in-progress` still applies, so a new push supersedes
+# a stale wait and a second `/coverage` supersedes the first.
 concurrency:
-  group: pr-coverage-${{ github.event.pull_request.number || github.event.issue.number }}
+  group: >-
+    pr-coverage-${{ github.event.pull_request.number || github.event.issue.number }}-${{ github.event_name }}${{
+    !((github.event_name == 'pull_request' &&
+    github.event.pull_request.head.repo.full_name == github.repository) ||
+    (github.event_name == 'issue_comment' &&
+    github.event.issue.pull_request != null &&
+    startsWith(github.event.comment.body, '/coverage') &&
+    contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)))
+    && format('-noop-{0}', github.run_id) || '' }}
   cancel-in-progress: true
 
 env:


shiwanigupta0809:
Read the diff, the full workflow on this branch, the job `if:` it mirrors, and the `post-coverage-comment` action the two paths now share. The diagnosis is right and the fix is correct as written — I confirmed the predicate matches the job `if:` character for character, walked the expression semantics on both branches (`!(true) && format(...) || ''` → `''`; `!(false) && format(...) || ''` → `-noop-<id>`), and checked that point 2 is genuinely load-bearing rather than belt-and-braces: `/coverage` and an ordinary comment are both `issue_comment`, so without the no-op shunt an ordinary comment would cancel an in-flight `/coverage` run, and without the event key a `/coverage` run would cancel an in-flight `pull_request` run. Both halves are needed.

Also checked the consequence the event split newly permits — a `/coverage` run and a `pull_request` run in flight simultaneously — and it's benign: `post-coverage-comment` uses `sticky-pull-request-comment` with `header: Code Coverage Report`, so both upsert the same marked comment rather than posting duplicates, and both resolve the same build by querying `refs/pull/<n>/merge` with `queueTimeDescending`. Worth knowing it was considered; nothing to change.

### Suggestion

**Job-level `concurrency` may make the duplicated predicate unnecessary — worth testing, because the duplication is the one durable hazard this change introduces.**

The comment in the diff names it: *"The predicate mirrors the job `if:` below — keep them in sync."* That is six lines of non-obvious logic, including the maintainer allowlist, existing in two places with nothing enforcing agreement. Drift has a concrete consequence rather than a cosmetic one: tighten the job `if:` (say, drop `COLLABORATOR`) and forget the concurrency copy, and a non-maintainer's `/coverage` comment is classified "real" by the group expression while being skipped by the job — so it joins `pr-coverage-<n>-issue_comment` and cancels a maintainer's in-flight `/coverage` run. That is the same class of bug this PR is fixing, reintroduced by a one-sided edit.

The PR's own diagnosis suggests the escape. The problem is stated as:

> `concurrency` is resolved at **run creation**, before the job-level `if:` is evaluated.

That is a property of *workflow-level* concurrency specifically. `jobs.<job_id>.concurrency` is documented as a separate key, evaluated in job context — so a job skipped by its `if:` should never acquire the group at all, and the shunt has nothing left to do:

```yaml
  coverage-report:
    runs-on: ubuntu-latest
    if: >-
      ...unchanged...
    concurrency:
      group: pr-coverage-${{ github.event.pull_request.number || github.event.issue.number }}-${{ github.event_name }}
      cancel-in-progress: true
```

Point 2 is preserved verbatim; point 1 falls out of the `if:` that already exists; the predicate exists once.

**What I verified and what I did not.** I confirmed from the workflow-syntax reference that `jobs.<job_id>.concurrency` exists, takes the same `group`/`cancel-in-progress` shape, and allows the `github` context the group needs. I did **not** verify that a job skipped by `if:` declines to acquire its concurrency group — the docs describe job-level concurrency without stating the skip interaction, and I have no way to run a workflow from here. So this is a question about a mechanism, not a claim that the current fix is wrong. If the skip does acquire the group, the suggestion collapses and the shunt is required exactly as written.

It is cheap to settle: a scratch repo with an `issue_comment` trigger, a job-level group, an `if:` that skips, and a long-running sibling run answers it in one comment. Given the fix cannot be demonstrated from a PR branch anyway (per your own note about `issue_comment` always executing the default-branch copy), that experiment is also the only way either design gets tested before it is depended upon.

If it does work, one more thing changes for the better: the group would no longer need `github.run_id`, so every skipped comment run stops creating a distinct single-member concurrency group. Minor, but a burst of review comments on an active PR currently produces one throwaway group per comment.

Not blocking either way — if the answer is "job-level acquires on skip too", or you would simply rather ship the known-good shape and revisit, the comment already documents the hazard, which is the important part.

shiwanigupta0809 (commented):
_Reviewed — see https://github.com/microsoft/mssql-rs/pull/478#issuecomment-5514613426_

David-Engel:
> Job-level `concurrency` may make the duplicated predicate unnecessary — worth testing, because the duplication is the one durable hazard this change introduces.

Thanks for checking both cancellation paths and the shared sticky-comment behavior. I’m keeping the current workflow-level design in this PR because the critical skipped-job concurrency behavior is not documented or verified; the mirrored predicate comment makes the maintenance constraint explicit, and the job-level approach remains a useful follow-up experiment rather than changing this known-correct fix.

shiwanigupta0809 (approved):

PR #479, opened 2026-09-02
Support bit, float, GUID and datetime parameters (AB#47500)

Theekshna:
## Description

This extends `SQLBindParameter` coverage to the fixed-width scalar C types -
bit, both float widths, GUID, and the three datetime structs - and corrects
these conversion behaviours:

- temporal parameters are declared at the maximum fractional-seconds scale, so
  `SQL_DESC_SCALE` reads 7 whatever `ColumnSize` and `DecimalDigits` say;
- a dropped datetime fraction is `22008` on `time`, `datetime2` and
  `datetimeoffset` alike, rather than splitting by target;
- `SQL_C_FLOAT` is range-checked for underflow and infinity, not just overflow;
- two state corrections: an illegal timezone on an over-precise
  `datetimeoffset` is `22007`, and a decimal literal that drops more than 38
  fractional digits is `22001`;
- two `unreachable!` calls in the `sql_variant` serializer are now protocol
  errors;
- a decimal literal whose only excess digits are trailing fraction zeros stays
  exact. `parse_decimal_literal` trims them and reduces the scale to match, so
  `"1" x 38 + ".00"` no longer overflows the exact mantissa, falls back to the
  `f64` approximation, and reaches a `decimal(38,0)` target as a different
  number under `SQL_SUCCESS`.

`DecimalDigits` still bounds the value even though it no longer reaches the
declaration, so a fraction the application asked not to carry is still an
error. Character, integer and binary pairings, data-at-execution sequencing,
and the `SQL_C_DEFAULT` resolution rules are unchanged.

## Shared with the fetch path

Both directions now go through one set of conversion primitives rather than a
copy each:

- `conversion/datetime.rs` is new and holds the calendar - `DateTimeParts` plus
  the `days_since_0001_from_civil` / `civil_from_days_since_0001` inverse pair -
  lifted out of the fetch path so parameters and fetch cannot disagree about
  what day a date is.
- `narrow_f64_to_f32` in `conversion/numeric.rs` is now the single `real` range
  check for both directions. The fetch path carried its own overflow-only copy,
  which is the pre-existing bug below.
- `SqlTime` counts 100 ns ticks; the parameter path writes them and
  `hms_from_ticks_100ns` reads them back, so the tick convention is stated once.

## One pre-existing bug fixed

`SQL_C_FLOAT` on the **fetch** path checked only overflow, so a magnitude below
`FLT_MIN` was written out silently instead of being reported as 22003 - exactly
[...]

PR #458, opened 2026-09-01
Bundle mssql-odbc in mssql-python-rs Wheels

gargsaumya:
## Description

This pull request adds support for optionally building and injecting the native `mssql-odbc` driver into Python wheels as part of the CI pipeline, across Windows, Linux (glibc + musl), and macOS (x64 and ARM64) platforms. The implementation introduces new pipeline parameters, conditional build steps, and a new build template for macOS, as well as updates to existing templates to control artifact publishing.

**Key changes:**

### Pipeline parameterization and documentation

- Added a new `buildOdbcNative` boolean parameter to the main pipeline YAMLs (`.pipeline/OneBranch/NonOfficialPythonWheelsPublish.yml`, `.pipeline/OneBranch/stages.yml`) to enable or disable building and injecting the native ODBC driver. Also updated documentation in `.pipeline/OneBranch/stages.yml` to describe the new workflow. [[1]](diffhunk://#diff-c2969227d3ff035fa63ad0fddb2934fd73e18ae6861cf959f0a9fed433936e0cR34-R38) [[2]](diffhunk://#diff-a39d2f7607b7f2a138fc01be791fcb78293d61ea85b15c98c7a29014c0a5f50eR9-R16) [[3]](diffhunk://#diff-a39d2f7607b7f2a138fc01be791fcb78293d61ea85b15c98c7a29014c0a5f50eR35-R39) [[4]](diffhunk://#diff-c2969227d3ff035fa63ad0fddb2934fd73e18ae6861cf959f0a9fed433936e0cR78)

### Conditional ODBC native driver build and injection

[...]

PR #482, opened 2026-09-02
Report truncation on a zero-length SQL_C_BINARY probe [AB#47537]

David-Engel:
## Description

`test_004_cursor_arrow.py::test_arrow_lob_wide` hard-crashed the Python interpreter with SIGSEGV against this driver. The crash is ours, not mssql-python's.

mssql-python's `arrow_batch` switches the whole result set to `SQLGetData` as soon as any column is a MAX type, and then reads *every* column that way — including the fixed `binary(9)` column in that test. Its `GetDataVar` helper starts each `SQL_C_BINARY` read with an empty `std::vector`, so the first call arrives with `TargetValuePtr = NULL` and `BufferLength = 0`, and it only grows the buffer and retries while the driver reports there is more to come.

This driver answered that probe with `SQL_SUCCESS` and an indicator of 9. To the caller that means "the whole value is in your buffer", so it stopped retrying and `memcpy`'d 9 bytes out of the buffer it never grew.

The fix reports `01004` / `SQL_SUCCESS_WITH_INFO` whenever the probe leaves bytes undelivered (including `SQL_NO_TOTAL`, where bytes remain but the count is unknown). An empty value *is* fully delivered by the probe, so that case keeps `SQL_SUCCESS` and consumes the column — matching msodbcsql, which answers `SQL_NO_DATA` on a repeat.

Measured against the CI-pinned reference (`msodbcsql 18.6.2.1`, `SQL_DRIVER_VER` `18.06.0002`), `SQLGetData(col, SQL_C_BINARY, NULL, 0, &ind)`:

| column | msodbcsql 18.6.2.1 | mssql-odbc (before) | mssql-odbc (after) |
|---|---|---|---|
| `binary(9)` | `SQL_SUCCESS_WITH_INFO`, `01004`, ind 9 | `SQL_SUCCESS`, ind 9 | `SQL_SUCCESS_WITH_INFO`, `01004`, ind 9 |
| `varbinary(8)` (empty) | `SQL_SUCCESS`, ind 0, then `SQL_NO_DATA` | `SQL_SUCCESS`, ind 0, forever | `SQL_SUCCESS`, ind 0, then `SQL_NO_DATA` |
| `nvarchar(10)` | `SQL_SUCCESS_WITH_INFO`, `01004`, ind 6 | `SQL_SUCCESS`, ind 6 | `SQL_SUCCESS_WITH_INFO`, `01004`, ind 6 |
| `varbinary(max)` | `SQL_SUCCESS_WITH_INFO`, `01004`, ind 3 | `HYC00` ([AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239)) | `HYC00` ([AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239)) |
| `int` / `datetime2(3)` | `SQL_ERROR`, `22003` | `SQL_SUCCESS` | `SQL_SUCCESS_WITH_INFO`, `01004` |

Actually delivering the bytes on the retry is still [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239), so `test_arrow_lob_wide` now fails with a `DatabaseError` (`HYC00`) instead of killing the process — which is what [AB#47537](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47537) asks for.

The fixed-source-type row (`int`, `decimal`, `datetime2`, ...) is a deliberate deviation, source-verified and recorded in `.github/instructions/mssql-odbc.instructions.md`: msodbcsql splits on `IsFixedSqlType()` over the **SQL** type (which explicitly excludes `SQL_BINARY`, so `binary(9)` sits on the variable side) and treats a short buffer for a fixed source type as a data overflow. Matching its `22003` would break mssql-python's `sql_variant` support, which depends on that same zero-length probe succeeding.

## The probe is answered in two places

Review caught that `main` moved under this branch: **#446** added a *second* zero-length `SQL_C_BINARY` probe on the buffered fast path in `sql_get_data_safe`, which `return`s before `write_captured_column` is reached. Since `arrow_batch` fetches a single row with nothing bound — exactly what makes a row buffered — that new site is the only one the crash path touches.

The merge was textually clean (the two blocks are ~750 lines apart in different functions) and every unit test kept passing, because they drive `write_captured_column` directly rather than the reachable entry point. `test_arrow_lob_wide` segfaulted again on the merge.

Both sites now route through one `answer_binary_probe` helper whose **return code carries the disposition** — `SQL_SUCCESS` means fully delivered and the caller must retire the column, `SQL_SUCCESS_WITH_INFO` means bytes remain and the value stays resident — so neither caller re-derives the rule and there is exactly one `available == 0` comparison in the file.

## Verification

Reproduced and verified end to end with the real mssql-python suite (`ubuntu:24.04` container, bundled `libmsodbcsql-18.6.so.2.1` swapped for our `mssqlodbc.so`, SQL Server 2025 in Docker), against a baseline driver built from **plain `main`** so the comparison is measured rather than asserted:

| suite | plain `main` (incl. #446) | + this PR |
|---|---|---|
| `test_004_cursor_arrow.py` | **SIGSEGV, core dumped** | 45 passed / 4 failed, no crash |
| `test_004_cursor.py` | 381 passed / 142 failed | identical |
| `test_019_sql_variant.py` | 22 passed / 7 failed | identical |

Also:
- `cargo bfmt`, `cargo bclippy` clean; **1350** `mssqlodbc` unit tests pass.
- Full ODBC e2e suite: 38/38 binaries pass.
- The six new e2e tests pass against **both** mssql-odbc and a real msodbcsql registration, so they need no `SKIP_IF_COMPARING_MSODBCSQL()` and hold on the parity leg.
- **The probe e2e tests are load-bearing, not incidentally green:** run against a driver built from unpatched `main`, they fail; with the fix, they pass. `BinaryColumnBesideALobStaysWithinItsBuffer` fails on `main` with *"delivered 0 bytes but claimed success"*.
- The `SQL_NO_TOTAL` unit test was likewise verified by mutation — loosening `available == 0` to `<= 0` fails that test and only that test.

> Note on CI: every Linux/Alpine test stage is `skipped` on PR builds, so `Test MacOS` is the only leg that runs the ODBC e2e suite — and it has been failing repo-wide in `Install and start Colima-based Docker` (28% of recent PR builds). The e2e results above were therefore produced locally against the exact merged tree; see the PR comments for the infra diagnosis.

## Related Issues

Fixes https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47537
Fixes https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47482

[AB#47482](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47482) (*Investigate SIGSEGV in mssql-python test_004_cursor_arrow*) is the investigation item for this same crash — same test, same signal 11. It is a consequence of the binary-probe gap fixed here rather than an independent bug, so it closes with this PR instead of as separate work; `BinaryColumnBesideALobStaysWithinItsBuffer` pins its crash shape.

Deliberately **not** closed here: [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239) (binary *data* delivery for `SQL_C_BINARY`). `test_arrow_lob_wide` still fails — now as a clean `DatabaseError` rather than a segfault — because of that remaining gap.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


mssql-odbc/src/api/get_data.rs
```diff
@@ -410,9 +410,22 @@ fn write_captured_column(
...
+        // empty buffer it passed, crashing the interpreter (AB#47537).
+        // `SQL_NO_TOTAL` is a truncation too — bytes remain, the count is just
+        // not known. Measured identical on msodbcsql 18.6.2.1: a `binary(9)`
+        // column answers `SQLGetData(SQL_C_BINARY, NULL, 0)` with
+        // `SQL_SUCCESS_WITH_INFO`, `01004` and indicator 9.
+        let available = binary_length(value);
+        unsafe { write_if_some(strlen_or_ind_ptr, available) };
+        if available == 0 {
+            return SQL_SUCCESS;
+        }
```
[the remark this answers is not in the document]
David-Engel: Good catch — fixed in f1fa32dd. The empty-value branch now clears `last_captured` / `partial_text_offset` and sets `current_row_last_col`, so a repeat reports `SQL_NO_DATA` and a final-column read reaches `finish_get_data`'s completion path.

I measured the reference driver to confirm this is the right disposition rather than assuming it, since the non-empty branch deliberately does the opposite. msodbcsql 18.6.2.1 (`SQL_DRIVER_VER` `18.06.0002`), repeated `SQLGetData(col, SQL_C_BINARY, NULL, 0, &ind)`:

| column | probe #1 | probe #2 |
|---|---|---|
| `varbinary(8)` (empty) | `SQL_SUCCESS`, ind 0 | `SQL_NO_DATA` |
| `binary(9)` | `SQL_SUCCESS_WITH_INFO`, `01004`, ind 9 | `SQL_SUCCESS_WITH_INFO`, `01004`, ind 9 |

So consuming on empty and staying resident when bytes remain is exactly msodbcsql's split.

The empty-value unit test now asserts the follow-up `SQL_NO_DATA` as you suggested, and I added the mirror assertion to the non-empty test so the two halves can't drift. The e2e test `ZeroLengthBinaryProbeOnEmptyValueSucceedsAndConsumesColumn` covers it live and passes against both this driver and a real msodbcsql registration, so it holds on the parity leg without a `SKIP_IF_COMPARING_MSODBCSQL()`.

> [!NOTE]

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -123,6 +123,20 @@ authoritative parity reference for this crate. Its source lives in the
     and delivers the bytes. Pre-existing for an explicit `SQL_C_BINARY` bind;
     deferred resolution makes it reachable without the application naming the C
     type.
+  - A zero-length `SQL_C_BINARY` `SQLGetData` on a **fixed-width** column
+    (`int`, `datetime2`, ...) reports `01004` / `SQL_SUCCESS_WITH_INFO` with the
+    byte count, where msodbcsql reports `22003` / `SQL_ERROR`. Measured against
```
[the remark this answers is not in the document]
David-Engel: Correct — my wording was wrong. `binary_length` returns a byte count only for the variants it has an encoding for (`int` → 4) and falls through to `SQL_NO_TOTAL` for every temporal variant, which my own measurement showed (`datetime2(3)` reports indicator `-4`). Reworded in f1fa32dd to say the indicator carries a byte count where `binary_length` knows one and `SQL_NO_TOTAL` where it does not, and why.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
```diff
@@ -410,9 +410,32 @@ fn write_captured_column(
...
+        // call is needed: `SQL_SUCCESS` claims its zero-length buffer now holds
+        // the value, and mssql-python then copies `indicator` bytes out of the
+        // empty buffer it passed, crashing the interpreter (AB#47537).
+        // `SQL_NO_TOTAL` is a truncation too — bytes remain, the count is just
+        // not known. Measured identical on msodbcsql 18.6.2.1: a `binary(9)`
+        // column answers `SQLGetData(SQL_C_BINARY, NULL, 0)` with
+        // `SQL_SUCCESS_WITH_INFO`, `01004` and indicator 9.
+        let available = binary_length(value);
+        unsafe { write_if_some(strlen_or_ind_ptr, available) };
+        if available == 0 {
```
saurabh500: Could we add a focused test for the `SQL_NO_TOTAL` path? The current tests cover zero and positive lengths, but treating `SQL_NO_TOTAL` like an empty value still leaves the suite green; that would consume the column and return `SQL_SUCCESS` instead of the intended truncation. Use a captured temporal value that `binary_length` maps to `SQL_NO_TOTAL`, then assert:

```rust
assert_eq!(ret, SQL_SUCCESS_WITH_INFO);
assert_eq!(ind, SQL_NO_TOTAL);
assert_eq!(state.diag_records.last().unwrap().sql_state, SQLSTATE_01004);
```

Repeat the probe and assert `SQL_SUCCESS_WITH_INFO` again so the test also proves that the value remains resident.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -1616,6 +1616,58 @@ TEST_F(GetDataLiveTest, NullVarbinaryMaxToBinaryTargetReportsNull) {
...
+
+    SQLCloseCursor(stmt_);
+}
+
+// The other half of the contract: an empty value has nothing left to deliver, so
+// the same probe is a plain success, and because that call delivered the whole
+// value the column is consumed -- a repeat reports SQL_NO_DATA. Measured
+// identically on msodbcsql 18.6.2.1 (SQL_SUCCESS with indicator 0 and no
+// diagnostic, then SQL_NO_DATA).
+TEST_F(GetDataLiveTest, ZeroLengthBinaryProbeOnEmptyValueSucceedsAndConsumesColumn) {
```
saurabh500, line 1659: Could we also cover the exact empty `sql_variant` sequence affected by this change? This test repeats `SQL_C_BINARY`, while mssql-python probes, queries `SQL_CA_SS_VARIANT_TYPE`, and then reads using the reported C type. A focused parity test could select `CAST(CAST('' AS VARBINARY(8)) AS SQL_VARIANT)`, perform the zero-length probe, assert `NumericAttr(stmt_, 1, SQL_CA_SS_VARIANT_TYPE) == SQL_C_BINARY`, and then assert that the actual read returns `SQL_NO_DATA`. Running that test on the msodbcsql leg would protect the interaction between consuming the empty value and retaining `last_variant_base`.

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -123,6 +123,23 @@ authoritative parity reference for this crate. Its source lives in the
     and delivers the bytes. Pre-existing for an explicit `SQL_C_BINARY` bind;
     deferred resolution makes it reachable without the application naming the C
     type.
+  - A zero-length `SQL_C_BINARY` `SQLGetData` on a **fixed-width** column
```
saurabh500: This parity record includes the runtime measurement, but not the msodbcsql source citation required by lines 23-30 of this file. Please add the relevant source path and describe what its zero-length fixed-width conversion does. If the source branch could not be identified, explicitly label this as a black-box 18.6.2.1 measurement rather than source-verified parity.

saurabh500 (approved):
Approved. The zero-length `SQL_C_BINARY` probe now correctly signals truncation when bytes remain and consumes truly empty values, preventing the caller-side out-of-bounds read without broadening the deferred binary-delivery work. The inline comments are non-blocking coverage and documentation follow-ups.

mssql-odbc/src/api/get_data.rs
David-Engel: Good catch — that gap was real. Added `get_data_binary_probe_on_unknown_length_reports_truncation` in f0f0c4d9, using a `DateTime2` capture, asserting `SQL_SUCCESS_WITH_INFO` / `SQL_NO_TOTAL` / `01004` and then a repeated probe to prove the value stays resident, as you specified.

I also verified the test is actually load-bearing rather than just green, by making the exact mutation you described — `available == 0` → `available <= 0`:

```
FAIL  get_data_binary_probe_on_unknown_length_reports_truncation
PASS  get_data_binary_probe_reports_length_without_consuming
PASS  get_data_binary_probe_on_bytes_reports_truncation
PASS  get_data_binary_probe_on_empty_value_succeeds_and_consumes_the_column
PASS  get_data_binary_probe_reports_null
```

One test caught it and only one, which is exactly the hole you identified. Reverted after confirming.

The test also guards its own premise — it asserts `binary_length(value) == SQL_NO_TOTAL` up front, so if a future change gives temporal values a real binary encoding the test fails loudly instead of quietly testing the wrong branch.

> [!NOTE]

mssql-odbc/tests/e2e/tests/get_data_test.cpp
David-Engel: Added in f0f0c4d9 as `ColAttributeLiveTest.EmptyVariantProbeConsumesValueButKeepsBaseType` — the sequence you described: probe, `SQL_CA_SS_VARIANT_TYPE`, then the read. I put it in `col_attribute_test.cpp` rather than `get_data_test.cpp` so it sits next to `VariantUnderlyingTypeAfterProbe` and can reuse the `NumericAttr` helper.

You were right that the interaction was uncovered — `last_captured` (which my change now clears) and `last_variant_base` are separate state, so nothing would have caught a future change clearing both together. I measured it on both drivers through the DM before writing the assertions:

| step | mssql-odbc | msodbcsql 18.6.2.1 |
|---|---|---|
| zero-length probe | `SQL_SUCCESS`, ind 0 | `SQL_SUCCESS_WITH_INFO`/`01004`, ind 0 |
| `SQL_CA_SS_VARIANT_TYPE` | `SQL_C_BINARY` | `SQL_C_BINARY` |
| re-read | `SQL_NO_DATA` | `SQL_NO_DATA` |

So the base type does survive the probe consuming the value, and the re-read behaves — both agree with msodbcsql. The test passes on the msodbcsql leg, so it's a real parity assertion and needs no `SKIP_IF_COMPARING_MSODBCSQL()`.

One thing worth flagging that I found while measuring: the probe's own return code diverges for the *variant* case — msodbcsql answers `01004` where we answer `SQL_SUCCESS` (a bare empty `varbinary(8)` with no variant wrapper is `SQL_SUCCESS` on both). I used `ASSERT_SQL_OK`, which accepts either, so the test asserts the parts that matter without pinning that difference. It's invisible to mssql-python, whose probe is gated on `SQL_SUCCEEDED`. I've registered it in the deviations list rather than leaving it undocumented — happy to chase it separately if you'd rather it match exactly.

> [!NOTE]

.github/instructions/mssql-odbc.instructions.md
David-Engel: Fair — that was a black-box measurement wearing a parity claim's clothes. Found the source and rewrote the note in f0f0c4d9.

It also **corrected my hypothesis**, which is a good argument for the rule. I had assumed the split was "fixed-width *C* type"; it is actually the *SQL* type, via `IsFixedSqlType()` (`odbc/sqlcprot.h`):

```cpp
inline BOOL IsFixedSqlType(SWORD swSqlType)
{
    return !IsPartialLenType(swSqlType) &&
           (swSqlType != SQL_VECTOR_MAPPED && swSqlType != SQL_CHAR &&
            swSqlType != SQL_WCHAR && swSqlType != SQL_BINARY);
}
```

`SQL_BINARY` is explicitly excluded, so `binary(9)` is on the *variable* side — which is why it truncates rather than overflowing, something my original framing got wrong.

`ColDataRetriever<>::GetColData` (`odbc/sqlcdata.h`) uses that to pick a policy class. Fixed source type → `BinaryOutputWithFixedLengthSqlType`, where delivery is all-or-nothing (the source itself comments that such data "is fetched in one call"), so a short buffer is an overflow: `if (... == BinaryWithFixedLengthSqlType && (SIZE_T)cbBuf < cbDataAvail) { wError = IDS_22_003; }` for `int`, and `Error = CVT_PREC` (`#define CVT_PREC IDS_22_003`) out of `ConvertToBinary` for `datetime2`. `IDS_22_003` → `22003` in `cli_common/src/clntcomn.cpp`. Variable source type takes `if (!IsFixedOrBinaryWithFixedServerType() && cbDataAvail) wError = IDS_01_004;` with the full length in the indicator.

That also gave me the source-level answer for the empty-value case in the sibling thread: `if (!cbDataAvail) goto Return3;` short-circuits ahead of both checks, and the column is then marked `STMT_ST_GETDATA_DONE` because `FIsLenRemaining()` is false — which is exactly the consume-on-empty behavior, now documented with its citation.

Two caveats I've written into the note rather than papering over:

1. **I omitted line numbers deliberately.** The checkout is `master` (`7a0c3d59`), not the 18.6.2.1 release branch, so line numbers could be wrong for the version CI actually compares against. File + function + condition are the durable part.
2. I verified each cited symbol exists with the quoted condition (`IsFixedSqlType`, `CVT_PREC`, the `IDS_22_003` and `IDS_01_004` branches, `FIsLenRemaining`, `STMT_ST_GETDATA_DONE`, the SQLSTATE map entry) rather than taking them on trust.

The runtime measurement stays alongside it, now labelled as corroborating the source reading rather than standing in for it.

> [!NOTE]

saurabh500 (approved):
Re-approved on `f0f0c4d9`. All three follow-ups landed well, and the third one materially improved the change.

**Verified locally** (detached checkout at `f0f0c4d9`, tree restored afterward):

- `cargo nextest run -p mssqlodbc --lib` — **1226 passed, 1 skipped**.
- Reproduced your mutation independently: `available == 0` → `available <= 0` fails `get_data_binary_probe_on_unknown_length_reports_truncation` and **only** that test. The new test is genuinely load-bearing, and asserting `binary_length(value) == SQL_NO_TOTAL` up front is the right way to keep it from quietly drifting onto the wrong branch.
- Confirmed the state-independence the variant test pins: `last_variant_base` is written only at capture (`get_data.rs:605`) and read at `col_attribute.rs:158`; the empty-probe branch clears `last_captured` and `partial_text_offset` but not the base type. So the e2e assertion is checking a real invariant, not a coincidence.

On the source citation — this is the part worth calling out. Correcting the boundary from "fixed-width C type" to `IsFixedSqlType()` on the **SQL** type, with `SQL_BINARY` explicitly excluded, is not a cosmetic rewording: it changes which side `binary(9)` sits on and therefore why it truncates instead of overflowing. Omitting line numbers because the checkout is `master` rather than the 18.6.2.1 release branch is the right call; file + function + condition is the durable citation.

On the residual divergence you surfaced (a `sql_variant` wrapping an empty value: `01004` on msodbcsql, `SQL_SUCCESS` here) — agreed this should not block. There are genuinely zero bytes, so `SQL_SUCCESS` is a truthful answer and is not the AB#47537 failure mode, which was claiming delivery while bytes remained. Both codes pass `SQL_SUCCEEDED`, so it is invisible to mssql-python. Documenting it and letting `ASSERT_SQL_OK` accept either is better than pinning a difference you would rather revisit under AB#47239.

No blocking issues. Merge once the re-triggered validation goes green — the merge is yours.

shiwanigupta0809 (commented):
Read the diff, the resulting `write_captured_column`, and the current state of `main`. Both prior threads are resolved well — I am not re-raising the empty-value consumption, the `SQL_NO_TOTAL` test, the variant e2e sequence, or the `IsFixedSqlType` citation.

**Blocking:** 1 (inline). **Suggestions:** none.

The finding is not about the change itself, which is correct. It is that `main` moved underneath this branch about an hour ago and the fix no longer covers the path the bug travels on.

### The short version

PR #446 ("Optimize row-wise `SQLGetData`") **merged**. It added a second zero-length `SQL_C_BINARY` probe — in the buffered fast path, ahead of the one this PR fixes — which returns bare `SQL_SUCCESS` with the byte count in the indicator. That is the AB#47537 shape exactly, and on current `main` it is the branch that a `SQLFetch` + unbound `SQLGetData` result set takes.

`git` will merge this cleanly: the two probes are ~750 lines apart and do not textually conflict. So the merge is silent, CI stays green, and the crash comes back on the path the repro actually uses.

Details, line numbers and the reason the two probes disagree are inline on the fixed block.

### Verified

**The fix itself is right, and the empty/non-empty split matches the reference.** `available == 0` → consume and `SQL_SUCCESS`; otherwise stay resident and `SQL_SUCCESS_WITH_INFO` / `01004`. That mirrors the measured msodbcsql behaviour in both directions, and `SQL_NO_TOTAL` landing on the truncation side is right — bytes remain, only the count is unknown.

**Consuming on empty is safe for `finish_get_data`.** Setting `current_row_last_col = col_index` is what lets `ready` become true for a last-column read, so the busy-release path is reachable; the non-empty branch deliberately leaves it unadvanced, so a resident value cannot release the connection early. The two arms are consistent with what `finish_get_data` expects of them.

**Clearing `last_captured` does not disturb the variant answer.** `last_variant_base` is separate state written at capture and read by `col_attribute.rs`, so the empty-probe branch clearing `last_captured` leaves `SQL_CA_SS_VARIANT_TYPE` answerable — which is what the new e2e test pins.

mssql-odbc/src/api/get_data.rs
```diff
@@ -410,9 +410,32 @@ fn write_captured_column(
...
+        // reads it for real on a following call. Bytes left undelivered make
+        // this a truncation, and saying so is what tells the caller a second
+        // call is needed: `SQL_SUCCESS` claims its zero-length buffer now holds
+        // the value, and mssql-python then copies `indicator` bytes out of the
+        // empty buffer it passed, crashing the interpreter (AB#47537).
+        // `SQL_NO_TOTAL` is a truncation too — bytes remain, the count is just
+        // not known. Measured identical on msodbcsql 18.6.2.1: a `binary(9)`
+        // column answers `SQLGetData(SQL_C_BINARY, NULL, 0)` with
+        // `SQL_SUCCESS_WITH_INFO`, `01004` and indicator 9.
+        let available = binary_length(value);
```
shiwanigupta0809, line 1025: **Blocking** — this is now the *second* zero-length `SQL_C_BINARY` probe in this file, and on current `main` it is no longer the one that runs first. PR #446 merged today and added an earlier probe that still has the AB#47537 behaviour.

On `main`, `get_data.rs` now has both:

```rust
// main:244 — inside sql_get_data_safe, buffered fast path (added by #446)
let binary_probe = target_type == SQL_C_BINARY && buffer_length == 0;
if binary_probe
    && let Some(value) = row.values.get(col_index - 1).and_then(Option::as_ref)
    && !matches!(value, ColumnValues::Null)
{
    let length = binary_length(value);
    let variant_base = row.variant_bases.get(col_index - 1).copied().flatten();
    stmt_state.last_variant_base = variant_base.map(|base| (col_index, base));
    unsafe { write_if_some(strlen_or_ind_ptr, length) };
    return SQL_SUCCESS;          // <-- bare SQL_SUCCESS, indicator = 9
}
```

```rust
// main:990 — write_captured_column, the block this PR rewrites
if binary_probe {
    unsafe { write_if_some(strlen_or_ind_ptr, binary_length(value)) };
    return SQL_SUCCESS;
}
```

The first one returns before `write_captured_column` is ever reached. It is a `return` out of `sql_get_data_safe`, not a fallthrough — so this PR's `01004` never fires for any column served from `buffered_get_data_row`.

**Why that is the path the bug uses.** #446's capture is gated on `row_array_size == 1 && bindings.is_empty()` (`fetch_scroll.rs`, `get_data_fetch`) — which is precisely mssql-python's `arrow_batch` shape as described in this PR's own summary: it switches the whole result set to `SQLGetData` and binds nothing. With a MAX column present, `current_result_supports_row_into()` is false, so it takes the *prefix* capture path and materialises the inline columns ahead of the first PLP. `binary(9)` in `test_arrow_lob_wide` is one of those. So the value is in `row.values`, non-`Null`, and line 245 answers it.

Net effect if this merges as-is: `SQL_SUCCESS` with indicator 9 into a zero-length buffer, `GetDataVar` stops retrying and `memcpy`s 9 bytes out of an empty `std::vector` — the same SIGSEGV, on the new path.

**This will not surface as a conflict.** The two blocks are ~750 lines apart and touch different functions, so the merge is textually clean and CI stays green. The new unit tests all drive `write_captured_column` directly, so they keep passing while the reachable path regresses.

What I would suggest, roughly in order of preference:

1. **Merge `main` in and apply the same rule at both sites.** The cleanest version is probably to factor the decision into one helper — something like `fn binary_probe_result(available: SqlLen, ...) -> SqlReturn` — so the two probes cannot drift again. The buffered site would need `post_diag(WARN_STRING_TRUNCATION)` and the `SQL_SUCCESS_WITH_INFO` return; note it currently has no `else` for the empty case at all, since it falls through to the generic path when `values[i]` is `None`, so the consume-on-empty half needs thinking about separately there.
2. Failing that, **at minimum add an e2e regression that goes through `SQLFetch` + unbound `SQLGetData` on a `binary(n)` column with a MAX column in the same result set**, so the buffered path is covered rather than only the captured one. The two new e2e tests here are valuable, but if they run on a result set that takes the buffered capture they may be passing for a different reason than the unit tests suggest — worth confirming which branch they actually exercise on top of current `main`.

Also worth a look while reconciling: #446's site sets `last_variant_base` before returning, and this PR's empty-value branch clears `last_captured` / `partial_text_offset` and advances `current_row_last_col`. Those are different dispositions of the surrounding state for the same logical operation, so whichever way they get unified, the empty-`sql_variant` sequence pinned by `EmptyVariantProbeConsumesValueButKeepsBaseType` is the case to re-check — it depends on exactly this interaction.

Nothing wrong with the change as written; it was correct against the `main` it was branched from. This is purely that the ground moved.

shiwanigupta0809:
⚠️ **Please hold the merge** — posted a review with a blocking inline finding: https://github.com/microsoft/mssql-rs/pull/482#pullrequestreview-5100127111

The change itself is correct. The problem is that `main` moved under this branch: **#446 merged today** and added a *second* zero-length `SQL_C_BINARY` probe in the buffered fast path (`main`, `get_data.rs:244-256`) that returns bare `SQL_SUCCESS` with the byte count in the indicator — the AB#47537 shape.

It `return`s out of `sql_get_data_safe` before `write_captured_column` is reached, so this PR's `01004` never fires for a column served from `buffered_get_data_row`. That capture is gated on `row_array_size == 1 && bindings.is_empty()`, which is exactly mssql-python's `arrow_batch` shape — so it is the path `test_arrow_lob_wide` travels.

The two blocks are ~750 lines apart in different functions, so this merges **cleanly with no conflict** and CI stays green. The new unit tests drive `write_captured_column` directly, so they keep passing while the reachable path regresses.

Details and suggested reconciliation inline.

mssql-odbc/src/api/get_data.rs
David-Engel: Blocking catch confirmed and fixed in 8855510b — thank you, this would have shipped as a "fixed" crash. Everything in your analysis held up when I checked it.

**Verified the regression before fixing it.** Merged `main` (clean, as you predicted) and re-ran the repro:

```
tests/test_004_cursor_arrow.py::test_arrow_lob_wide
timeout: the monitored command dumped core
PYTEST_RC=139
```

I also built a driver from **plain `main`** to be sure this isn't specific to my branch — `main` is crashing right now, independent of this PR:

| suite | plain `main` (incl. #446) | + this PR |
|---|---|---|
| `test_004_cursor_arrow.py` | **SIGSEGV, core dumped** | 44 passed / 5 failed, no crash |
| `test_004_cursor.py` | 381 / 142 | 381 / 142 (identical) |
| `test_019_sql_variant.py` | 22 / 7 | 22 / 7 (identical) |

**Took option 1.** Both sites now route through `answer_binary_probe`, and the return code carries the disposition so neither caller re-derives the rule: `SQL_SUCCESS` means fully delivered and the caller must retire the column, `SQL_SUCCESS_WITH_INFO` means bytes remain and the value stays resident. There is now exactly one `available == 0` comparison in the file.

On the consume-on-empty half you flagged for the buffered site: an empty non-`Null` value *does* enter that branch (`values[i]` is `Some`), it just never retired the column. It now uses the same idiom as the neighbouring complete-delivery arms — `row.consumed = row.consumed.max(col_index)` plus `current_row_last_col` / `partial_text_offset` — and I kept `last_variant_base` being set on both dispositions, since the variant probe depends on it.

**On your point about which branch the e2e tests were really exercising** — you were right to push on that, and it's the part I'd have gotten wrong by inference. Rather than reason about it, I ran all four probe e2e tests against the unpatched `main` driver:

```
[  FAILED  ] ZeroLengthBinaryProbeReportsTruncationWhenBytesRemain
[  FAILED  ] ZeroLengthBinaryProbeOnEmptyValueSucceedsAndConsumesColumn
[  FAILED  ] ZeroLengthBinaryProbeReportsTruncationOnBufferedRow
[  FAILED  ] ZeroLengthBinaryProbeOnEmptyBufferedValueConsumesColumn
[  PASSED  ] 0 tests.
```

All four fail without the fix and pass with it, so they're load-bearing rather than passing for the wrong reason. The two new ones use `test_arrow_lob_wide`'s actual shape — `binary(9)` alongside `nvarchar(max)`, read unbound — so the buffered probe is covered directly. I also added a `buffered_binary_probe_...` unit test that drives `sql_get_data_safe`, since as you noted the existing unit tests call `write_captured_column` directly and prove nothing about reachability.

`EmptyVariantProbeConsumesValueButKeepsBaseType` still passes on both legs, so the interaction you singled out survives the unification.

One pre-existing test needed its expectation updated: `buffered_get_data_preserves_variant_base_for_probe` asserted the old bare `SQL_SUCCESS`. Its subject — variant base surviving, value staying re-readable — is untouched, so I moved the return code and *added* indicator and `01004` assertions rather than loosening anything.

Full re-validation: 1293 unit tests, clean fmt/clippy, 38/38 e2e binaries, and all five new e2e tests green against a real msodbcsql registration too.

I've also written the general lesson into the ODBC instructions rather than just fixing the instance — that a protocol rule answered in two places needs one shared function, and that a test calling an inner function directly proves nothing about which branch production traffic reaches. That's the failure mode that let this through, and it wasn't specific to this probe.

> [!NOTE]

David-Engel:
Hold addressed in 8855510b — the blocking thread is resolved with details.

You were right on every point. I reproduced the regression before fixing it (merged main, re-ran the repro: `core dumped`, `RC=139`), and built a driver from **plain main** to confirm it is not branch-specific — main is crashing right now, independent of this PR:

| suite | plain `main` (incl. #446) | + this PR |
|---|---|---|
| `test_004_cursor_arrow.py` | **SIGSEGV, core dumped** | 44 passed / 5 failed, no crash |
| `test_004_cursor.py` | 381 / 142 | 381 / 142 (identical) |
| `test_019_sql_variant.py` | 22 / 7 | 22 / 7 (identical) |

Took your option 1: both probes now route through one `answer_binary_probe` helper whose return code carries the disposition, so neither caller re-derives the rule. Also added the buffered-path e2e coverage you asked for — and rather than assume it exercises the right branch, I ran all four probe e2e tests against the unpatched `main` driver: **4 failed, 0 passed**, then all four pass with the fix.

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -125,6 +125,63 @@ authoritative parity reference for this crate. Its source lives in the
     `SQL_C_CHAR` default was already unsupported, so the new mapping does not
     regress observable fetch behavior. msodbcsql resolves all three to
     `SQL_C_BINARY` and delivers the bytes.
+  - A zero-length `SQL_C_BINARY` `SQLGetData` on a column whose **source SQL
+    type is fixed-length** (`int`, `datetime2`, ...) reports `01004` /
+    `SQL_SUCCESS_WITH_INFO`, where msodbcsql reports `22003` / `SQL_ERROR`. The
+    indicator carries a byte count where `binary_length` knows one (`int` → 4)
+    and `SQL_NO_TOTAL` where it does not — every temporal variant falls through
```
Theekshna: [Suggestion] This says the SQL_NO_TOTAL class is "every temporal variant", but `binary_length`'s catch-all (get_data.rs:2338) also covers `ColumnValues::Decimal`/`ColumnValues::Numeric`, which aren't temporal. I checked the reference tree: `SQL_DECIMAL`/`SQL_NUMERIC` are not in `IsPartialLenType`'s case list (sqlcprot.h:1421-1436) and aren't in `IsFixedSqlType`'s exclusion set (sqlcprot.h:1581-1585), so msodbcsql classifies them as fixed-source-type too -- the same divergence class as the measured `int`/`datetime2` row (`22003`/SQL_ERROR there vs. `01004`/SQL_SUCCESS_WITH_INFO + SQL_NO_TOTAL here). This is a source read, not a msodbcsql measurement of an actual `decimal`/`numeric` column -- flagging as an unmeasured neighbor of the measured claim, not asserting the behavior. Worth measuring `decimal(10,2)`/`numeric` alongside int/datetime2(3) and folding the result into this note.

Theekshna (commented):

## Summary
This fixes AB#47537 (SIGSEGV from a zero-length `SQL_C_BINARY` probe reporting bytes as delivered) by unifying both probe sites (`sql_get_data_safe`'s buffered fast path and `write_captured_column`) through one `answer_binary_probe` helper. The review history here is already unusually thorough: a Copilot bot pass, two rounds from @saurabh500 (including independent mutation verification of the `SQL_NO_TOTAL` test and source-citation review), and a real blocking catch from @shiwanigupta0809 (main moved under the branch mid-review, reopening the exact bug via a second probe site added by #446) that was fixed and re-verified with a plain-`main` crash repro. I re-verified rather than re-trusting: confirmed `IsFixedSqlType`/`IsPartialLenType` in the reference tree match the PR's citation exactly (`sqlcprot.h:1581-1585`, `:1421-1436`), re-read the unified `answer_binary_probe` and both call sites, and diffed the divergence documentation against the code.

| Severity | Count |
| --- | ---: |
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 0 |

**Suggestion** (inline on `.github/instructions/mssql-odbc.instructions.md:132`): the new divergence note characterizes the `SQL_NO_TOTAL` class by example as "every temporal variant," but `binary_length`'s catch-all (`get_data.rs:2338`) also covers `ColumnValues::Decimal`/`ColumnValues::Numeric`, which are not temporal. I checked the reference tree: `SQL_DECIMAL`/`SQL_NUMERIC` are not in `IsPartialLenType`'s case list and aren't in `IsFixedSqlType`'s exclusion set, so msodbcsql classifies them as fixed-source-type too — same divergence class as the measured `int`/`datetime2` row (`22003`/`SQL_ERROR` there vs. `01004`/`SQL_SUCCESS_WITH_INFO` + `SQL_NO_TOTAL` here). This is a source-read observation, not a msodbcsql measurement of an actual `decimal`/`numeric` column, so I'm not asserting the behavior — just that it's an unmeasured neighbor of the measured claim and the doc's phrasing undersells it. Worth a follow-up to measure `decimal(10,2)`/`numeric` alongside `int`/`datetime2(3)` and fold it into this same note (or file a tracking item if that's deferred).

Everything else held up: parity fix is source-verified in both directions (empty-value consumption via `sqlcdata.h`'s `Return3` short-circuit, and the fixed/variable split via `IsFixedSqlType`); the four probe e2e tests and the `SQL_NO_TOTAL` unit test were independently mutation-verified by prior reviewers (I re-read but did not re-run the mutation, given it's already evidenced twice in-thread); PR description/checklist match the diff and `gh pr checks` (all green); divergence is documented with source citations and a labelled residual gap (empty-`sql_variant` `01004` vs. `SQL_SUCCESS`); no AI slop — the new doc comments and `answer_binary_probe` doc explain *why* (the AB#47537 mechanism, the drift risk of two answer sites) rather than restating code.

.github/instructions/mssql-odbc.instructions.md
David-Engel: You're right on both counts, and the "unmeasured neighbor" framing was the right call — fixed in 9ecf31c2.

Confirmed the code claim first: `binary_length`'s catch-all is `_ => return SQL_NO_TOTAL`, so `Decimal`/`Numeric` land there alongside the temporal variants. "Every temporal variant" was describing a subset as if it were the whole class.

Then measured rather than extending the source reading, since a parity claim here needs one. Both drivers, 18.6.2.1 (`SQL_DRIVER_VER` `18.06.0002`):

| column | msodbcsql | mssql-odbc |
|---|---|---|
| `int`, `money`, `smallmoney`, `uniqueidentifier` | `SQL_ERROR` / `22003` | `01004`, ind 4 / 8 / 4 / 16 |
| `decimal(10,2)`, `decimal(38,10)`, `numeric(18,4)` | `SQL_ERROR` / `22003` | `01004`, ind `SQL_NO_TOTAL` |
| `date`, `time(7)`, `datetime2(3)`, `datetimeoffset(3)` | `SQL_ERROR` / `22003` | `01004`, ind `SQL_NO_TOTAL` |

So your source read was correct: `decimal`/`numeric` are classified fixed-source by msodbcsql and land in the same divergence class as `int`/`datetime2` — same `22003` vs `01004`, and on our side with `SQL_NO_TOTAL` rather than a byte count.

The note now carries that table and says the class is "decimal and numeric as well as every temporal type," and I added the `IsPartialLenType` / `IsFixedSqlType` observation you made to the source-verified paragraph, since it's the reason `decimal` sits on the fixed side.

Two things I did beyond the immediate correction:

**Pinned the class in code, not just prose.** `binary_length_covers_the_value_kinds` now includes `Decimal`, `Numeric`, `Date`, `Time` and `DateTime2` asserting `SQL_NO_TOTAL`, so the doc is no longer the only record and the two can't drift. I also added `money` and `smallmoney` asserting **8** and **4** — they're fixed-source types like the rest but *do* have explicit arms, so they're the counter-examples that keep the boundary honest rather than letting someone conclude "fixed-source ⇒ `SQL_NO_TOTAL`."

**Measured `uniqueidentifier` too**, which wasn't in the note at all: `22003` on msodbcsql, `01004` with indicator 16 here. Same class, now recorded.

Full re-validation after the change: 1297 unit tests, clean fmt/clippy.

> [!NOTE]

Theekshna (commented):

## Summary
This adds coverage for the fixed-source-type side of the AB#47537 fix (`int`/`decimal`/`numeric`/temporal columns answering the zero-length `SQL_C_BINARY` probe) and, per the follow-up note in this branch's history, extends the divergence doc's measurement table to cover `decimal`/`numeric` explicitly instead of folding them into "temporal" — resolving the one Suggestion I raised on the prior commit at this exact spot. I independently re-verified rather than re-trusting the fix: read `IsFixedSqlType`/`IsPartialLenType` (`sqlcprot.h:1421-1444`, `:1579-1585`) and the two error branches in `sqlcdata.h` (`IDS_22_003` at the fixed-length `BinaryWithFixedLengthSqlType` check, `IDS_01_004` at the `!IsFixedOrBinaryWithFixedServerType()` check) against the reference tree, and confirmed they match the PR's citations exactly. `SQL_NO_TOTAL = -4` in `odbc_types.rs`, so `available == 0` correctly excludes it from "fully delivered" — consistent with the claimed mutation test (I could not re-run it: this environment's Rust toolchain fails to codegen `crc32fast`'s AVX512 path unrelated to this PR, confirmed on two separate `RUSTFLAGS` attempts, so I verified the boundary condition by reading rather than by running).

| Severity | Count |
| --- | ---: |
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 0 |

**[Suggestion]** (`mssql-odbc/tests/e2e/tests/get_data_test.cpp`): the new e2e tests cover the *agreement* row (`binary(9)`/`varbinary`/`nvarchar`, where this driver matches msodbcsql) on both probe sites, and the sql_variant/empty-value interaction in `col_attribute_test.cpp`. The *divergence* row — a fixed-source-type column (`int`, `decimal`, `datetime2`, ...) answering `01004`/`SQL_SUCCESS_WITH_INFO` here where msodbcsql answers `22003`/`SQL_ERROR` — is only unit-tested with synthetic captured values (`get_data_binary_probe_on_unknown_length_reports_truncation`), not exercised against a live server. That's the row most worth an e2e regression guard, precisely because it's the one place this driver deliberately disagrees with the reference and the doc's own measurement table is the record of that disagreement — a live-server test would catch a future accidental convergence or regression that a unit test built from a synthetic `ColumnValues` cannot. Not blocking: the classification mechanism (`IsFixedSqlType`) is source-verified and type-agnostic, so the unit coverage is sound: this is a completeness gap in the e2e suite, not a correctness question.

Everything else held up: the two probe sites route through one `answer_binary_probe`, so they cannot re-diverge the way #446 caused them to before; the `SQL_NO_TOTAL`/decimal/numeric doc note and table now match `binary_length`'s actual catch-all and were pinned with new unit test cases; PR description and checklist match the diff and `gh pr checks` (all green, including a genuine 23m50s Windows build — no repeat of the anomaly I flagged on #486's e2e build); no AI slop in the new doc comments (they explain the drift mechanism and cite reference lines, not restate the diff).

David-Engel:
Good suggestion — implemented in 9d03926f.

You were right that the divergence row was the gap: the agreement row had e2e coverage on both probe sites, but the row where this driver *deliberately* disagrees with msodbcsql was only exercised through synthetic `ColumnValues`. Added `ZeroLengthBinaryProbeOnFixedSourceTypesReportsTruncation`, covering both indicator sub-classes — `int` (explicit `binary_length` arm → 4) and `decimal(10,2)` / `datetime2(3)` (catch-all → `SQL_NO_TOTAL`).

Since the test asserts behaviour msodbcsql does not share, it needs `SKIP_IF_COMPARING_MSODBCSQL()` — and per the repo rule that such a guard must be added only against a compare run that actually fails, I verified rather than assumed. With the guard temporarily removed, the msodbcsql leg fails exactly as expected:

```
Expected equality of these values:
    Which is: -1          <- SQL_ERROR, vs expected SQL_SUCCESS_WITH_INFO
    Which is: "01004"
    Which is: "22003"     <- msodbcsql
    Which is: 4
    Which is: 0           <- indicator untouched
```

With the guard restored: reference leg skips, mssql-odbc leg passes, full e2e stays 38/38.

On your note about not re-running the mutation because of the `crc32fast` AVX512 codegen failure — thanks for saying so explicitly rather than implying you'd run it. For what it's worth the boundary is now also pinned from the other direction: `binary_length_covers_the_value_kinds` asserts `SQL_NO_TOTAL` for `Decimal`/`Numeric`/`Date`/`Time`/`DateTime2` *and* asserts `money` → 8 / `smallmoney` → 4 as counter-examples, so "fixed-source ⇒ `SQL_NO_TOTAL`" can't be inferred by accident.

One unrelated thing this shook out: a comment line in the e2e file had been joined to the following `TEST_F` declaration, silently commenting out `ZeroLengthBinaryProbeOnEmptyBufferedValueConsumesColumn`. Caught it because the compare-leg build broke; fixed in the same commit, and both buffered-path tests are confirmed running again.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -1616,10 +1616,180 @@ TEST_F(GetDataLiveTest, NullVarbinaryMaxToBinaryTargetReportsNull) {
...
+
+        SQLCloseCursor(stmt_);
+    }
+}
+
 // An integer column delivered to its natural fixed-width C target, rather than
 // being rendered as text.
-TEST_F(GetDataLiveTest, IntColumnToSlongTarget) {
-    ASSERT_SQL_OK(ExecDirect("SELECT CAST(-2000000 AS INT) AS c1"), SQL_HANDLE_STMT,
+TEST_F(GetDataLiveTest, IntColumnToSlongTarget) {    ASSERT_SQL_OK(ExecDirect("SELECT CAST(-2000000 AS INT) AS c1"), SQL_HANDLE_STMT,
```
Theekshna: [Nit] This commit's edit dropped the newline between the opening brace and the first statement: `TEST_F(GetDataLiveTest, IntColumnToSlongTarget) {    ASSERT_SQL_OK(...)` is now on one line. Compiles fine (braces don't need a newline), but it's a diff/readability regression introduced by this commit, not pre-existing. Split it back onto its own line.

Theekshna (commented):
Unattended re-review since the last marker. The new commit (`9d03926f`) adds a live e2e guard (`ZeroLengthBinaryProbeOnFixedSourceTypesReportsTruncation`) for the AB#47537 divergence row previously flagged as untested — closes that Suggestion. Re-derived the row: `binary_length()` in `get_data.rs` gives `int` -> 4 (explicit arm) and lets `decimal`/`datetime2` fall through to the `SQL_NO_TOTAL` catch-all, matching the test's expected indicators. Cross-checked the `IsFixedSqlType`/`IsPartialLenType` parity citation in `.github/instructions/mssql-odbc.instructions.md` against `sqlcprot.h` in the reference tree — the exclusion set (`SQL_BINARY`/`SQL_CHAR`/`SQL_WCHAR` plus the `IsPartialLenType` list) matches as described, so `decimal`/`numeric` are correctly on the fixed side. All 6 prior review threads are resolved. One new Nit from this commit's own edit, below. ADO PR validation has not yet reported a status for this exact head commit (GH Actions checks are all green) — left for CI/a human, not re-triggered.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 1 |

mssql-odbc/tests/e2e/tests/get_data_test.cpp
David-Engel: Fixed in 4757358d. You're right that it was introduced by this branch, not pre-existing.

This is the second casualty of the same pair of careless edits — the first joined a comment line to the following `TEST_F` and silently commented out `ZeroLengthBinaryProbeOnEmptyBufferedValueConsumesColumn`. So rather than just fixing the line you flagged, I audited the branch's whole diff for the same damage:

- No other brace-joined-to-statement lines in either e2e file.
- No added line over 100 columns.
- `gtest --gtest_list_tests` now registers **62** cases in `get_data_test` and **27** in `col_attribute_test`, with all six new tests listed by name.

That last one is the check that actually matters here: a commented-out `TEST_F` doesn't fail, it simply ceases to exist, so the suite stays green with fewer tests and neither CI nor a passing run would tell you. Counting registered cases is what catches it, and it's what I should have done when I first noticed the joined line rather than fixing only the instance in front of me.

Full e2e re-run after the change: 38/38.

> [!NOTE]

David-Engel:
### `Test MacOS` failures here are a repo-wide infra flake, not this PR

Recording the diagnosis so it is not re-derived. This PR's code has never failed a check — the four red runs are all one setup task.

**Scope.** Across the last 25 PR validation builds, **7 (28%)** failed on `Install and start Colima-based Docker` — PRs #482, #493 (x3), #458, #498 (x2). Every other leg on this PR passes, including the `mssql-python suite on mssql-odbc driver (cross-repo)` job that this fix actually targets.

**Why it times out.** The symptom is ADO's `The task has timed out`, *not* the script's own `##[error]colima did not start:`. That means the step is killed before `start-colima-macos.sh` reaches its retry/budget logic. The arithmetic:

| | value |
|---|---|
| step `timeoutInMinutes` (`sql-setup-template.yml`) | 13 min / 780s |
| `COLIMA_BUDGET_SECONDS` | 480s |
| `COLIMA_START_TIMEOUT_SECONDS` (first attempt) | 540s |
| `brew update` + `brew install docker colima` | **unbounded** |

A slow Homebrew plus a single full-length first `colima start` exceeds 780s on its own, so the step dies before the script can retry or report. The script's header documents the boot flake as *"~3% of runs"*; the measured rate is now ~28%, so either the agents or Homebrew have degraded materially since it was tuned.

**Plausible fixes** (someone who owns the macOS pool should pick — deliberately *not* doing this here, since patching shared pipeline infra through an unrelated feature PR is how a 'temporary' workaround becomes permanent):
- bound `brew update` / `brew install` with `run_bounded` so they are inside the budget rather than outside it;
- lower `COLIMA_START_TIMEOUT_SECONDS` so `brew` + attempt 1 fits inside 13 min;
- or raise the step `timeoutInMinutes` to cover the script's real worst case.

David-Engel:
**Update — worsening, and I am stopping the retry loop.**

Build 172522 failed the same way, but it has now spread to a **second** macOS job: both `Test MacOS` *and* `Build mssql-python macOS (cross-repo)` timed out on `Install and start Colima-based Docker`. Both templates invoke the same script with the same 13-minute step budget (`sql-setup-template.yml`, `test-mssql-python-macos-template.yml`).

That is 5 failed runs on this PR with no convergence. I am not re-triggering again: each `/azp run` queues the full matrix, so if the macOS pool is capacity-starved, retrying is plausibly making it worse rather than better.

Deferring to whoever owns the macOS pool — diagnosis and three candidate fixes are in the comment above. This PR's own code has not failed a check at any point; the Linux, Windows, ARM, Kerberos and the `mssql-python suite on mssql-odbc driver (cross-repo)` legs are all green on this head.

Theekshna (commented):
Unattended re-review since the last marker (`9d03926f`→`e4cd6eb9`).

The intervening commits are almost entirely a merge from `main` (pulling in unrelated PRs #483/#477/#487) plus two genuine content changes on top: `4757358d` restores the line break after `IntColumnToSlongTarget`'s opening brace, and the merge's comment reword in `get_data_test.cpp` changes the `SQL_NUMERIC_STRUCT`-not-implemented note from "tracked gap (AB#47816), retarget when it closes" to "permanent non-goal." Diffing merge-base-to-head at both the old and new review points shows `get_data.rs` and `col_attribute_test.cpp` byte-identical to the last review; only `get_data_test.cpp` moved (174→171 lines), confirming no new production logic since `9d03926f`.

Both changes are closures, not new surface: the brace fix is exactly the Nit I filed last round (verified restored, matches source). The reworded comment now agrees with `.github/instructions/mssql-odbc.instructions.md`'s "Known divergences" table, which already records this as a **permanent** divergence (`SQL_C_CHAR` for exact numerics) — the old wording was stale relative to that table, so this is a real (if small) documentation-consistency fix, not slop.

Checked both of my prior threads on this PR via GraphQL: both resolved. The decimal/numeric "unmeasured neighbor" Suggestion from two rounds ago was closed with a genuine re-measurement (`9ecf31c2`) that also proactively measured `uniqueidentifier`, which hadn't been asked for — good practice, not re-flagging.

Six-check items: msodbcsql parity — audited the whole `binary_length` match in `get_data.rs` in a prior round (fixed vs. variable classification cross-checked against `IsFixedSqlType`/`IsPartialLenType` in `sqlcprot.h`); no arms changed since. Divergence docs — the fixed-width `int`/`datetime2` deviation and the `SQL_C_NUMERIC` gap are both recorded in `.github/instructions/mssql-odbc.instructions.md`. PR description — the evidence tables (zero-length probe matrix, mssql-python suite comparison) are unchanged since my prior verification pass and still match the diff; linked to AB#47537 (and cross-references AB#47239, AB#47816 by number only); checklist all checked, consistent with `gh pr checks` (no failures). No new AI slop. Two stale `saurabh500` `APPROVED` reviews predate the merge commit and both content changes — not treated as a live approval.

No findings this round.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

David-Engel:
Thanks for the clean pass. One factual correction for the record, since it attributes a change to this branch that it does not contain.

> the merge's comment reword in `get_data_test.cpp` changes the `SQL_NUMERIC_STRUCT`-not-implemented note from "tracked gap (AB#47816), retarget when it closes" to "permanent non-goal."

This branch did not reword that comment. On `e4cd6eb9` the line is unchanged from `main`:

```
mssql-odbc/tests/e2e/tests/get_data_test.cpp:783
// SQL_NUMERIC_STRUCT is not implemented, recorded as a tracked gap (AB#47816)
```

`git grep "permanent non-goal"` returns nothing anywhere in the branch, and `git diff --stat origin/main -- mssql-odbc/tests/e2e/tests/get_data_test.cpp` is **171 insertions, 0 deletions** — the file is purely additive here, so no existing comment was reworded by this PR or by the merge.

Doesn't change your conclusion (0 findings, and you judged the wording fine either way) — just flagging so the review record doesn't carry a change this PR didn't make.

Your other observations match what I see: both prior threads resolved, `get_data.rs` and `col_attribute_test.cpp` byte-identical since `9d03926f`, and the two `saurabh500` approvals are indeed stale relative to the merge — which is why the PR still reads `REVIEW_REQUIRED`.

David-Engel:
**Local e2e verification of the #479 merge** — because CI structurally cannot provide it right now.

On PR builds every Linux/Alpine test stage is `skipped` (confirmed on build 172554: `Test Linux amd64`, `Test Linux arm64`, `Test Alpine`, `Test Alpine arm64` all skipped). `Test MacOS` is therefore the **only** leg that runs the ODBC e2e suite — and it has died in `Install and start Colima-based Docker` on every run since the #479 merge, so the tests never execute. That left the merge with no e2e coverage in CI at all.

Closed that gap locally against the exact merged tree (`137171f4`'s content on top of `0ed0d415`):

| check | result |
|---|---|
| ODBC e2e suite (`run_e2e.sh --release`) | **38/38 passed** |
| `mssql_python` `test_004_cursor_arrow.py` | 45 passed / 4 failed, **no crash** (`RC=1`, not `139`) |
| `cargo nextest -p mssqlodbc --lib` | **1350 passed** |
| `cargo fmt --check` / `cargo clippy -D warnings` | clean |

The arrow suite is one better than before the merge (45 vs 44 passing), and the AB#47537 crash remains fixed — `test_arrow_lob_wide` fails cleanly with `HYC00` from the deferred binary-delivery gap (AB#47239) rather than SIGSEGV.

Re-triggering validation once for this head. To be clear about the earlier commitment not to re-run: that was about repeatedly retrying *the same* commit, which was not converging. This is a new head after a merge that hit the flake once, and macOS passed unaided on the previous head (`e4cd6eb9`), so a single run here is not the loop I stopped.

David-Engel:
**Correcting my own earlier diagnosis** — and I am leaving CI red now, as I said I would.

Build 172603 shows the two macOS jobs failing `Install and start Colima-based Docker` in **two different ways**:

| job | failure |
|---|---|
| `Test MacOS` | `The task has timed out.` (13-min step budget) |
| `Build mssql-python macOS (cross-repo)` | `Bash exited with code '1'` |

The second one matters. Exit 1 is `start-colima-macos.sh`'s **own** give-up path — it reached the end of the retry loop and printed `colima did not start: N of 3 configured attempt(s)`. So that job did *not* run out of wall clock; it burned all three `colima start` attempts inside the budget and the VM still never came up.

That partly invalidates what I suggested earlier. Bounding `brew` or resizing `COLIMA_START_TIMEOUT_SECONDS` addresses the *timeout* variant, but it will not help a VM that cannot boot at all in three tries. The script's own header already anticipates this ("the lima hostagent either misses its 5s startup window or never emits the `running` event") and treats it as ~3% flake; at the rate now observed, `COLIMA_START_ATTEMPTS=3` is simply not enough headroom, or the hosted image has regressed.

So the useful next step is probably to read `C:\Users\davidengel/.colima/_lima/colima/ha.stderr.log` — the script already tails 50 lines of it on each failed attempt, so the real boot error should be in the job log — rather than tuning timeouts blind.

Not re-triggering again. I said last round that if this head failed twice I would stop rather than keep spending full matrices on a strained pool, and it has. The PR's own code is unaffected: every non-macOS leg is green, including `mssql-python suite on mssql-odbc driver (cross-repo)`, and the e2e suite plus the arrow regression were verified locally against this exact tree (38/38; 45 passed / no crash).

David-Engel:
**Resolved — CI is fully green (19/19).** Closing out my earlier macOS comments so they do not read as still-open.

#500 (*Install the newest bottled docker CLI instead of source-building it*) fixed it, and it landed on the root cause rather than the symptom. My diagnosis had the right mechanism — `brew update` / `brew install docker colima` running **outside** the step's time budget — but understated *why* they were slow: `brew install docker` was **source-building** the CLI. #500 wraps the install in `run_bounded` so it is inside the budget, and uses `--force-bottle` (with a newest-bottle fallback) so it never builds from source.

That also retires the second failure mode I flagged, where one job exhausted all three `colima start` attempts. With the install consuming the clock, those retries were being squeezed rather than the VM genuinely failing to boot — so my correction was right that timeout tuning alone would not have fixed it, but the real fix removes the cost instead of budgeting for it.

Merged into this branch at `9869b838`. All three macOS legs now pass:

| leg | result |
|---|---|
| `Build Stage Test MacOS` | pass |
| `Build Stage Build MacOS` | pass |
| `Build mssql-python macOS (cross-repo)` | pass |

Worth noting what that unlocks: `Test MacOS` is the only PR leg that runs the ODBC e2e suite (every Linux/Alpine test stage is `skipped` on PRs), so this is the **first CI-side execution of the six new e2e tests in this PR**. Everything I had been verifying locally is now confirmed by the pipeline.

The only remaining item is a fresh approval — the earlier ones went stale across five merges from `main`.

saurabh500 (approved):

PR #483, opened 2026-09-02
Report SQL_C_NUMERIC and SQL_C_UTINYINT for sql_variant values (AB#47702)

David-Engel:
## Description

`SQLColAttributeW(SQL_CA_SS_VARIANT_TYPE)` answered `SQL_C_CHAR` for a `sql_variant` holding `decimal`, `numeric` or `money`. mssql-python routes **solely** on that answer, so those values reached Python as `str` instead of `decimal.Decimal`.

The divergence was previously recorded as *permanent*, on the reasoning that reporting `SQL_C_NUMERIC` would make callers request a `SQL_NUMERIC_STRUCT` this driver does not emit. That turns out not to describe the consumer: `ddbc_bindings.cpp`'s `MapVariantCTypeToSQLType` maps `SQL_C_NUMERIC` → `SQL_NUMERIC`, and the `SQL_DECIMAL`/`SQL_NUMERIC` fetch case then **re-fetches with `SQL_C_CHAR`** and wraps the text in `Decimal`. The character delivery path this driver already has is exactly what serves it — no struct is ever requested.

Review then surfaced a second, independent bug in the same function: `tinyint` answered `SQL_C_TINYINT` (signed) where the server type is unsigned 0-255, so a caller fetching 200 into a signed `char` would read -56. That is also fixed here.

## Measurement

All 24 base types a `sql_variant` can carry, measured on both drivers loaded directly (no Driver Manager), replaying mssql-python's exact `SQL_C_BINARY` probe → `SQL_CA_SS_VARIANT_TYPE` sequence. Reference: retail **msodbcsql18 `SQL_DRIVER_VER` 18.06.0001**.

**Fixed by this PR**

| base type | msodbcsql | before | after |
| --- | --- | --- | --- |
| `decimal` / `numeric` / `money` / `smallmoney` | `SQL_C_NUMERIC` (2) | `SQL_C_CHAR` (1) ❌ | `SQL_C_NUMERIC` ✅ |
| `tinyint` | `SQL_C_UTINYINT` (-28) | `SQL_C_TINYINT` (-6) ❌ | `SQL_C_UTINYINT` ✅ |

**Already matching** — `bit`, `smallint`, `int`, `bigint`, `real`, `float`, `char`, `varchar`, `nchar`, `nvarchar`, `binary`, `varbinary`, `uniqueidentifier`.

**Deliberate, now recorded in the divergence table**

| base type | msodbcsql | mssql-odbc | why |
| --- | --- | --- | --- |
| `date` / `smalldatetime` / `datetime` / `datetime2` | `SQL_C_DATE` (9) / `SQL_C_TIMESTAMP` (11) | `SQL_C_TYPE_DATE` (91) / `SQL_C_TYPE_TIMESTAMP` (93) | The canonicalization direction documented in `api/type_rules.rs` — this driver canonicalizes toward 3.x, msodbcsql toward its internal 2.x. Measured under both `SQL_OV_ODBC3` and `SQL_OV_ODBC3_80`; not app-version gated. |
| `time` / `datetimeoffset` | `SQL_C_BINARY` under `SQL_OV_ODBC3`, `SQL_C_SS_*` under 3.80 | `SQL_C_SS_TIME2` / `SQL_C_SS_TIMESTAMPOFFSET` always | Both agree under 3.80. Answering the real type regardless of declared version gives the caller strictly more information. |

**Net: 20 of 24 base types measured identical; the remaining 4 are deliberate.**

## Verification

Loading the built `mssqlodbc.dll` directly — the same way mssql-python `LoadLibrary`/`dlopen`s it — against a live SQL Server, applying mssql-python's own `MapVariantCTypeToSQLType` routing:

**Before**

```
[FAIL] decimal (row 7)            variantCType=   1 -> str      (expected Decimal)
[FAIL] numeric (row 8)            variantCType=   1 -> str      (expected Decimal)
[FAIL] implicit numeric (row 25)  variantCType=   1 -> str      (expected Decimal)
[FAIL] money                      variantCType=   1 -> str      (expected Decimal)
```

**After**

```
[PASS] decimal (row 7)            variantCType=   2 -> Decimal  value=999.9900
[PASS] numeric (row 8)            variantCType=   2 -> Decimal  value=888.88
[PASS] implicit numeric (row 25)  variantCType=   2 -> Decimal  value=45.67
[PASS] money                      variantCType=   2 -> Decimal  value=12.3400
[PASS] tinyint (row 2)            variantCType= -28 -> int
[PASS] int / smallint / bigint / real / bit / varchar / nvarchar
```

Confirmed in CI: the cross-repo `mssql-python suite on mssql-odbc driver` job runs `tests/test_019_sql_variant.py` against this driver, and all four tests named in [AB#47702](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47702) pass.

## Tests

- `VariantExactNumericsReportNumeric` — `decimal`/`numeric`/`money`/`smallmoney`/implicit numeric.
- `VariantBaseTypesMatchMsodbcsql` — all 18 measured-parity base types, exhaustive by construction so the next divergence cannot hide behind whichever type someone thought to add.
- `VariantDateTimeBaseTypesUseTheThreeXSpellings` — pins the deliberate 2.x/3.x difference.
- `VariantDecimalStillDeliversAsCharacter` — the `SQL_C_CHAR` read that actually serves mssql-python.

The first, second and fourth were run against msodbcsql18 **through the Driver Manager** and pass there, so they are genuine parity assertions rather than self-fulfilling ones, and carry no `SKIP_IF_COMPARING_MSODBCSQL()`.

## Also in this change

- `variant_read_c_type` in the benchmark harness now accepts the 2.x `SQL_C_DATE`/`SQL_C_TIMESTAMP` spellings. msodbcsql answers those for temporal variants, and they were falling through to the `SQL_C_CHAR` default — so the harness would have benchmarked a character fetch on one driver against a typed fetch on the other, the exact asymmetry its own doc comment says the fold exists to prevent. Unreachable today, since `make_variant` builds no temporal column.
- Stale references retired: `get_data_test.cpp` called `SQL_NUMERIC_STRUCT` a permanent non-goal, and the benchmark harness and its README justified skipping decimal variants by a divergence that no longer exists.

## Not in scope

`SQL_C_NUMERIC` is still not an accepted `SQLGetData`/`SQLBindCol` target — any source into it returns `HYC00`. mssql-python never asks for it, but a caller that trusts the reported type would. The divergence row is reclassified from "deliberate and permanent" to a tracked gap, with the msodbcsql ARD-vs-variant precision/scale behavior measured and recorded on the follow-up item.

## Guardrail

The root cause was a divergence justified by an assumption about a *consumer's* code that was never read. `.github/instructions/mssql-odbc.instructions.md` now requires the same standard of evidence for a consumer claim as for a msodbcsql claim, and forbids marking a divergence permanent on an unread consumer.

## Related Issues

- Fixes https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47702
- Follow-up filed: https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47816

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


mssql-odbc/tests/e2e/tests/col_attribute_test.cpp
```diff
@@ -493,3 +494,61 @@ TEST_F(ColAttributeLiveTest, VariantTypeBeforeProbeIsSequenceError) {
...
+// This compares against msodbcsql: the value is the whole point of the test.
+TEST_F(ColAttributeLiveTest, VariantExactNumericsReportNumeric) {
+    struct Case {
+        const char* label;
+        const char* expr;
+    };
+    const Case cases[] = {
+        {"decimal", "CAST(999.99 AS DECIMAL(18, 4))"},
+        {"numeric", "CAST(888.88 AS NUMERIC(10, 2))"},
+        {"money", "CAST(12.34 AS MONEY)"},
```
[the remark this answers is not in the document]
David-Engel: Good catch — the `Money4` arm changed but only `MONEY` was measured. Added a `SMALLMONEY` case to this test and a `Money4` row to `variant_c_type_covers_the_base_types`.

Measured first rather than assumed: retail msodbcsql18 `SQL_DRIVER_VER` 18.06.0001 answers `SQL_C_NUMERIC` (2) for a `sql_variant` holding `smallmoney`, same as `decimal`/`numeric`/`money`. The whole test (all five cases) now passes against msodbcsql18 through the Driver Manager, so it stays a real parity assertion rather than expanding past the measurement.

> [!NOTE]

mssql-odbc/docs/typed-columnar-fetch-plan.md
```diff
@@ -103,8 +103,8 @@ These were found by reading `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp` while reviewi
...
+| Any source into `SQL_C_NUMERIC` | converts, per Appendix D | `HYC00` | Gap — Task [47816](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47816). Previously recorded as permanent on the belief that mssql-python asks for decimals as character data; it does, but only *after* keying on a `SQL_C_NUMERIC` answer from the row below, so the target still has to become real for any caller that trusts the reported type. Anchored by `UnsupportedCTypeReturnsHyc00ThenValueReadable`. |
+| `SQL_CA_SS_VARIANT_TYPE` for a variant holding `decimal` / `numeric` / `money` | `SQL_C_NUMERIC` | `SQL_C_NUMERIC` | Parity, since AB#47702. Measured against retail msodbcsql18 `18.06.0001`: `decimal`, `numeric` and `money` all answer `SQL_C_NUMERIC` (`2`). This reports the type of the value; it does not imply `SQL_NUMERIC_STRUCT` is a usable target yet (row above). mssql-python maps the answer to `SQL_NUMERIC` and then re-fetches with `SQL_C_CHAR`, so the character delivery path is what actually serves it. |
```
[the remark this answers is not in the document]
David-Engel: You're right, and all three were stale. Updated in this PR:

- `tests/e2e/tests/get_data_test.cpp` — the anchor comment now describes `SQL_C_NUMERIC` as a tracked gap (AB#47816) and says to retarget the test at another unimplemented C type when that closes, rather than deleting the `HYC00` coverage.
- `mssql-odbc-bench/src/odbc_bench.cpp` and `mssql-odbc-bench/README.md` — the exclusion was justified by a divergence that no longer exists. Both now record that the drivers agree (`variant_read_c_type`'s `default` arm folds `SQL_C_NUMERIC` onto `SQL_C_CHAR` on either driver), and that the column set is deliberately unchanged so existing perf baselines stay comparable.

I grepped for the remaining phrasings (`permanent non-goal`, `not scheduled to become supported`, `deliberately reports SQL_C_CHAR`) and there are no other occurrences left in the repo. Adding a decimal variant column to the benchmark is now technically possible but I left it out — it would shift the table shape and invalidate the recorded baselines, which is outside this fix.

> [!NOTE]

Theekshna (commented):
> Unattended automated review: this review draft was prepared by automation and was not pre-screened by a human.

I checked the changed sql_variant metadata path in mssql-odbc/src/api/col_attribute.rs against the existing probe flow in mssql-odbc/src/api/get_data.rs, the updated divergence documentation, the prior Copilot discussion, the passing CI/coverage report, and mssql-python's MapVariantCTypeToSQLType routing. The landed diff matches the PR description, includes linked work items, and I did not find any additional actionable issues.

| Severity | Count |
| --- | ---: |
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

No additional blocking findings, suggestions, or nits.

The previously raised Copilot points about SMALLMONEY parity coverage, stale divergence notes, and the missing verification-index entry are already addressed in the current diff.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -513,17 +513,19 @@ fn variant_c_type(base: TdsDataType) -> SqlSmallInt {
...
         | TdsDataType::DecimalN
         | TdsDataType::Numeric
         | TdsDataType::NumericN
         | TdsDataType::Money
         | TdsDataType::Money4
-        | TdsDataType::MoneyN => SQL_C_CHAR,
+        | TdsDataType::MoneyN => SQL_C_NUMERIC,
         TdsDataType::DateN => SQL_C_TYPE_DATE,
         TdsDataType::TimeN => SQL_C_SS_TIME2,
         TdsDataType::DateTime | TdsDataType::DateTim4 | TdsDataType::DateTimeN => {
```
Theekshna: Not a problem with this change, but worth catching while this function is open: three of the neighbouring arms diverge from msodbcsql in exactly the way this PR exists to fix, and none of them is in the divergence table or compared by any e2e test.

| variant base | mssql-odbc | msodbcsql | path through the msodbcsql source |
| --- | --- | --- | --- |
| `Int1` (L509) | `SQL_C_TINYINT` (-6) | `SQL_C_UTINYINT` (-28) | `rgbSRV2CTYPE[SQLINT1]` |
| `DateN` | `SQL_C_TYPE_DATE` (91) | `SQL_C_DATE` (9) | `rgbSRV2SQLTYPE[SQLDATEN]` = `SQL_DATE` -> `fColCType = SQL_NATIVE_DATE` (`sqlcdata.cpp:1865`) -> `sqlcdesc.cpp:6468` |
| `DateTime` / `DateTim4` / `DateTimeN` / `DateTime2N` | `SQL_C_TYPE_TIMESTAMP` (93) | `SQL_C_TIMESTAMP` (11) | `rgbSRV2SQLTYPE` = `SQL_TIMESTAMP` -> `SQL_NATIVE_*` (`sqlcdata.cpp:1846-1859`) -> `sqlcdesc.cpp:6465` |

The corroborating detail is `SQL_CA_SS_VARIANT_SQL_TYPE` at `sqlcdesc.cpp:6504`, which explicitly fixes 9/10/11 up to 91/92/93. The `SQL_CA_SS_VARIANT_TYPE` case does no such fixup, so it looks deliberate that it returns the ODBC 2.x spellings.

`TimeN` and `DateTimeOffsetN` do match -- both reach `SQL_C_SS_TIME2` / `SQL_C_SS_TIMESTAMPOFFSET` for a 3.8 app.

Two caveats on this, in fairness:

1. **This is a source read, not a measurement.** The standard this PR is amending the instructions file to reinforce is that a msodbcsql claim gets measured against the retail driver. Someone should do that before any of this is written into the divergence table.
2. **It is harmless for mssql-python today.** `MapVariantCTypeToSQLType` (`ddbc_bindings.cpp:3275-3294`) accepts both spellings of date and timestamp, and all three spellings of tinyint. So there is no live bug here.

What makes it worth recording anyway is the shape of it: the measurement table in the description covers decimal, int, real, float, bit, varchar and nvarchar, and `VariantUnderlyingTypeAfterProbe` compares only int and varchar against msodbcsql. Date, datetime and tinyint variants have never been measured or compared, which is the same blind spot that let the `SQL_C_CHAR` answer survive. A follow-up work item alongside AB#47816 seems like the right home -- I would not hold this PR for it.

mssql-odbc-bench/src/odbc_bench.cpp
```diff
@@ -533,10 +533,11 @@ ColumnSpec make_lob_text(std::vector<ColumnSpec>& columns, bool wide,
     return column;
 }
 
-// Build one sql_variant column. Only base types whose SQL_CA_SS_VARIANT_TYPE
-// answer is unambiguous are used: mssql-odbc deliberately reports SQL_C_CHAR
-// where msodbcsql reports SQL_C_NUMERIC for decimal/money variants, and a
-// benchmark that depended on that difference would not be the same work on both.
+// Build one sql_variant column. The exact numerics are no longer excluded for a
+// parity reason: since AB#47702 both drivers answer SQL_C_NUMERIC for
+// decimal/money variants, which `variant_read_c_type` folds onto SQL_C_CHAR on
```
Theekshna: Small thing that rides along with the other comment. `variant_read_c_type` (L1146) accepts `SQL_C_TYPE_DATE` and `SQL_C_TYPE_TIMESTAMP` but not `SQL_C_DATE` (9) or `SQL_C_TIMESTAMP` (11), so those fall to the `default` arm and become `SQL_C_CHAR`. Its own doc comment says both spellings are accepted precisely so "a benchmark that only understood one spelling would fail on the other driver rather than measure it".

If msodbcsql really does answer the 2.x spellings for date/datetime variants, then adding a date or datetime variant column here would silently measure a typed fetch on mssql-odbc against a character fetch on msodbcsql -- the exact asymmetry this rewritten comment is reasoning about.

Purely latent right now: `make_variant` only builds `variant_integer`, `variant_bigint` and the nvarchar arm, so nothing hits it. Just noting it since it is the mechanism by which the divergence would leak into a benchmark later.

mssql-odbc/src/api/col_attribute.rs
Theekshna, line 574: Unrelated to this change, but worth catching while the function is open: the `Int1` arm (L509) answers `SQL_C_TINYINT` (-6) where msodbcsql answers `SQL_C_UTINYINT` (-28), via `rgbSRV2CTYPE[SQLINT1]`.

This one is not the 2.x/3.x spelling question -- both are current ODBC 3.x identifiers and they differ in *signedness*, not vintage. `tinyint` is unsigned 0-255 on the server, so `SQL_C_UTINYINT` is the accurate answer; a caller that trusts `SQL_C_TINYINT` and fetches into a signed `char` reads 255 as -1.

It is also internally inconsistent with this driver's own view: `type_rules.rs:314` resolves `SQL_TINYINT` -> `SQL_C_UTINYINT` for a default binding. So the same value gets described two different ways depending on whether it arrives through a variant or a column.

Two caveats, in fairness:

1. **This is a source read, not a measurement.** The standard this PR is amending the instructions file to reinforce is that a msodbcsql claim gets measured against the retail driver. Worth confirming that way before anything is written into the divergence table.
2. **It is harmless for mssql-python today** -- `MapVariantCTypeToSQLType` accepts all three tinyint spellings (`ddbc_bindings.cpp:3291-3294`).

What makes it worth recording is the shape rather than the impact: the measurement table in the description covers decimal, int, real, float, bit, varchar and nvarchar, and `VariantUnderlyingTypeAfterProbe` compares only int and varchar against msodbcsql. A `tinyint` variant has never been measured or compared, which is the same blind spot that let the `SQL_C_CHAR` answer survive. A follow-up work item alongside AB#47816 seems like the right home -- I would not hold this PR for it.

(For completeness, since it is the neighbouring question and the answer is *not* a bug: `DateN` -> `SQL_C_TYPE_DATE` and the datetime arms -> `SQL_C_TYPE_TIMESTAMP` also differ from msodbcsql, which answers the 2.x `SQL_C_DATE` (9) / `SQL_C_TIMESTAMP` (11) here -- ungated by app version, unlike the adjacent `SQL_TIME2_MAPPED` arm. But that is the documented, deliberate canonicalization-direction difference in `type_rules.rs:13-20`: this driver canonicalizes toward the 3.x spellings, msodbcsql toward its internal 2.x ones. Correct as written.)

mssql-odbc-bench/src/odbc_bench.cpp
Theekshna, line 538: Latent, and adjacent to the comment being rewritten here rather than caused by it. `variant_read_c_type` (L1146) accepts `SQL_C_TYPE_DATE` (91) and `SQL_C_TYPE_TIMESTAMP` (93) but not `SQL_C_DATE` (9) or `SQL_C_TIMESTAMP` (11), so those fall to the `default` arm and become `SQL_C_CHAR`.

That matters because of the canonicalization-direction difference in `type_rules.rs:13-20`: this driver answers the 3.x spellings for a date or datetime variant, msodbcsql answers the 2.x ones (`sqlcdesc.cpp:6465-6469`, ungated by app version). This is by design on both sides and permanent -- it is not something a future fix removes.

So if a date or datetime variant column were ever added here, the harness would bind a typed fetch on mssql-odbc and a character fetch on msodbcsql, which is exactly the asymmetry `variant_read_c_type`'s own doc comment says the both-spellings handling exists to prevent ("a benchmark that only understood one spelling would fail on the other driver rather than measure it"). Note this is a harness reading a driver's answer, not the driver accepting 2.x identifiers -- the 3.x-only scope does not cover it.

Purely hypothetical right now: `make_variant` only builds `variant_integer`, `variant_bigint` and the nvarchar arm, so nothing reaches it. Fine to leave; noting it so the gap is on record next to the comment that reasons about this exact class of difference.

Theekshna (commented):
Reviewed this against the msodbcsql source and against mssql-python's own source rather than taking the description on trust. The core change is correct and the evidence holds up. One small follow-up below; not a blocker.

**Verified independently**

- **Parity is real.** `sqlcmisc.cpp`'s `rgbSRV2CTYPE[]` maps `SQLDECIMAL`, `SQLDECIMALN`, `SQLNUMERIC`, `SQLNUMERICN`, `SQLMONEY`, `SQLMONEY4` and `SQLMONEYN` all to `SQL_C_NUMERIC`. Money is confirmed twice over: `sqlcdata.cpp:1861` forces `SQL_NATIVE_MONEY` for the three money types, and `sqlcdesc.cpp:6466` maps that back to `SQL_C_NUMERIC`. All seven changed arms are right, smallmoney included.
- **The per-value model matches.** msodbcsql reaches the answer through `GetVariantBaseTypeInfo` -> `SetKnownInformation(bTDSType)`, so it answers from the base type of the value just probed -- the same thing `last_variant_base` does here.
- **The consumer claim holds.** In `ddbc_bindings.cpp`, `MapVariantCTypeToSQLType` maps `SQL_C_NUMERIC` -> `SQL_NUMERIC`, and the `SQL_DECIMAL`/`SQL_NUMERIC` fetch case re-fetches with `SQL_C_CHAR` into a 64-byte buffer and wraps the text in `Decimal`. `SQL_C_NUMERIC` appears in that file *only* on the input-parameter binding path, never as a `SQLGetData`/`SQLBindCol` target -- so the "Not in scope" section is accurate for this consumer, and 64 bytes is ample for `decimal(38, x)` and for money.
- **No hidden coupling.** `variant_c_type` has exactly one caller. Default-C-type resolution for binds lives in `type_rules.rs::resolve_default_c_type` and is untouched, so `SQL_C_DEFAULT` behavior does not move. That was the obvious place a regression could have hidden.

The docs reclassification and the instructions-file guardrail are both well-argued, and the smallmoney case added in response to the earlier comment pins a genuinely distinct wire type rather than padding the test.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Confirmed by measurement, and fixed here rather than deferred — it turned out to be a one-token change in the function this PR already touches.

**Measured** (retail msodbcsql18 `18.06.0001`, replaying mssql-python's probe-then-attribute sequence against both drivers loaded directly):

```
tinyint    msodbcsql: -28 SQL_C_UTINYINT    mssql-odbc: -6 SQL_C_TINYINT
```

So your read of `rgbSRV2CTYPE[SQLINT1]` holds up against the retail binary. `Int1` now answers `SQL_C_UTINYINT`, which also removes the inconsistency you spotted with `type_rules.rs` — `resolve_default_c_type` (L314) and `effective_param_c_type` (L103) both already mapped `SQL_TINYINT` to `SQL_C_UTINYINT`, so the variant path was the odd one out.

**On the blind spot, which I agree was the real finding.** Rather than add a `tinyint` row, I measured all 24 base types a `sql_variant` can carry, on both drivers. `VariantBaseTypesMatchMsodbcsql` now compares all 18 that are measured to agree — bit, tinyint, smallint, int, bigint, real, float, decimal, numeric, money, smallmoney, char, varchar, nchar, nvarchar, binary, varbinary, uniqueidentifier — against msodbcsql with no skip guard. It passes. Exhaustive by construction, so the next divergence can't hide behind whichever type someone thought to add.

**Your parenthetical was right too, and it cost me an assumption.** `date`/`smalldatetime`/`datetime`/`datetime2` do differ (2.x `9`/`11` vs our 3.x `91`/`93`), and it is *not* app-version gated — I measured under both `SQL_OV_ODBC3` and `SQL_OV_ODBC3_80` and msodbcsql answers the 2.x form either way. Correct as written, now pinned by `VariantDateTimeBaseTypesUseTheThreeXSpellings` (skips the reference leg) and recorded in the divergence table.

One thing the sweep turned up that neither of us had: **`time`/`datetimeoffset` are app-version dependent.** Under `SQL_OV_ODBC3` msodbcsql answers `SQL_C_BINARY` while we answer `SQL_C_SS_TIME2`/`SQL_C_SS_TIMESTAMPOFFSET`; under `SQL_OV_ODBC3_80` both agree. That is your `SQL_TIME2_MAPPED` gating showing up from the outside. Recorded as a row, and deliberately left out of the parity test since it would assert the fixture's declared version rather than the driver.

Net: 20 of 24 base types measured identical, 4 deliberate.

> [!NOTE]

mssql-odbc-bench/src/odbc_bench.cpp
David-Engel: Fixed rather than left, since it was two case labels and the measurement backs it up.

I measured the temporal variants against retail msodbcsql18 `18.06.0001` while chasing the `tinyint` thread, and your reading is exactly right:

```
date            msodbcsql: 9  SQL_C_DATE       mssql-odbc: 91 SQL_C_TYPE_DATE
datetime        msodbcsql: 11 SQL_C_TIMESTAMP  mssql-odbc: 93 SQL_C_TYPE_TIMESTAMP
smalldatetime   msodbcsql: 11 SQL_C_TIMESTAMP  mssql-odbc: 93 SQL_C_TYPE_TIMESTAMP
datetime2       msodbcsql: 11 SQL_C_TIMESTAMP  mssql-odbc: 93 SQL_C_TYPE_TIMESTAMP
```

And ungated by app version, as you said — same answers under `SQL_OV_ODBC3` and `SQL_OV_ODBC3_80`. So `variant_read_c_type` really would have bound a typed fetch on one driver and a character fetch on the other, which is the precise failure its own doc comment claims to prevent. `SQL_C_DATE` and `SQL_C_TIMESTAMP` now fold onto the 3.x forms alongside the spellings already there, and the comment explains why with the measured values rather than leaving it as a symmetry argument.

Still unreachable — `make_variant` builds no temporal column — so this changes no benchmark today. It just means the trap isn't waiting for whoever adds one.

Worth noting this is the only one of the divergences that is *permanent*: the exact-numeric one was a bug, but the 2.x/3.x split is the deliberate canonicalization direction in `type_rules.rs:13-20` and no future fix removes it. That's now a row in the divergence table too, so the harness comment and the table agree.

> [!NOTE]

mssql-odbc/tests/e2e/tests/col_attribute_test.cpp
```diff
@@ -493,3 +497,160 @@ TEST_F(ColAttributeLiveTest, VariantTypeBeforeProbeIsSequenceError) {
...
+                  SQL_HANDLE_STMT, stmt_);
+    EXPECT_STREQ("999.9900", reinterpret_cast<const char*>(text));
+
+    SQLCloseCursor(stmt_);
+}
+
+// Every base type whose answer is measured to agree with msodbcsql, in one
+// place. Spot-checking a couple of types is what let the exact-numeric answer
+// (AB#47702) and the `tinyint` signedness answer both survive unnoticed, so the
+// table is exhaustive by construction rather than by whichever type someone
```
Theekshna: **[Nit]** The claim is stronger than the mechanism behind it. `variant_c_type` still ends in `_ => SQL_C_CHAR` (`col_attribute.rs:554`), and both this list (18 cases) and `variant_c_type_covers_the_base_types` (21 cases) are hand-written — so the table is exhaustive by *enumeration*, not by construction. Add a `TdsDataType` variant tomorrow and it silently answers `SQL_C_CHAR`, with nothing failing anywhere.

Harmless today, and not worth holding the PR for. What I would change is the comment rather than the code: as worded it tells the next person they are protected by a mechanism that does not exist, which is close to the failure mode this PR is cleaning up — a claim carried forward without being re-checked.

Either fix is fine:

- Replace the wildcard with the remaining variants explicitly. Adding a type then becomes a compile error, and you actually have the property being described. This is the one I would pick.
- Or soften to "covers every base type a variant can carry" and drop "by construction".

Supporting detail, not a bug: that wildcard is why `IntN` is unhandled while `BitN`, `FltN`, `MoneyN`, `DecimalN`, `NumericN` and `DateTimeN` all are. Per MS-TDS a variant carries integers as `INT1/2/4/8TYPE` and never `INTNTYPE`, so nothing reaches it — but that asymmetry is exactly what an enumerated list fails to surface. Source read, not a measurement.

Theekshna (commented):
Re-reviewed at `06b1c434`. Both findings from my earlier review are addressed, and measured rather than deferred — the `tinyint` signedness fix is confirmed against retail msodbcsql18, and the response went past the instance to close the blind spot by measuring all 24 base types.

Two things I checked independently, since the fix's safety rests on them:

- **The consumer is unaffected.** `ddbc_bindings.cpp:3291-3294` maps `SQL_C_TINYINT` / `SQL_C_UTINYINT` / `SQL_C_STINYINT` alike to `SQL_TINYINT`, so changing the reported spelling does not move mssql-python's routing.
- **The driver can deliver what it now reports.** `fetch_convert.rs:200` writes `SQL_C_UTINYINT` as a real target and `type_rules.rs:188,219` accept it — so this is not a second instance of the `SQL_C_NUMERIC` "reports a type it cannot service" gap tracked in AB#47816.

`VariantBaseTypesMatchMsodbcsql` carries no `SKIP_IF_COMPARING_MSODBCSQL()`, so it is a real parity assertion rather than a self-fulfilling one, and `CAST(200 AS TINYINT)` is the value that actually distinguishes signed from unsigned. The date/time test correctly does skip the reference leg. 18 compared + 4 deliberate + 2 version-dependent reconciles to 24.

One Nit inline, on a comment rather than on code. Nothing blocking.

| Severity | Count |
| --- | --- |
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 1 |

mssql-odbc/tests/e2e/tests/col_attribute_test.cpp
David-Engel: You're right, and the framing is the part that stings — I wrote a claim about a protection mechanism without checking that the mechanism existed, in a PR whose whole point is that someone did exactly that. Took your first option.

`variant_c_type` no longer ends in `_ => SQL_C_CHAR`. All twelve remaining variants are enumerated — `IntN`, `Image`, `Text`, `NText`, `Xml`, `Json`, `Vector`, `Udt`, `SqlTable`, `SsVariant`, `Void`, `None` — so the property is now the one being described rather than an aspiration.

**Verified rather than asserted**, since that was the whole failure here. I added a throwaway `TdsDataType` variant and rebuilt:

```
error[E0004]: non-exhaustive patterns: `&TdsDataType::ProbeOnly` not covered
   |                          ^^^^ pattern `&TdsDataType::ProbeOnly` not covered
```

Then reverted it. Adding a type is now a compile error at this match, not a silent character answer.

Your `IntN` detail earned its own comment in the code rather than a passing mention — it is the one arm whose absence looks like an oversight next to `BitN`/`FltN`/`MoneyN`/`DecimalN`/`NumericN`/`DateTimeN`, and "a variant carries integers as `INT1/2/4/8TYPE`, never `INTNTYPE`" is exactly the kind of thing that gets rediscovered the hard way. It is cited to MS-TDS 2.2.5.5.2 and flagged as a source read, not a measurement, per your own caveat.

The e2e comment now points at that match as the enforcement instead of claiming the property for itself, and says plainly that the list is hand-written. Behavior is unchanged — every one of those twelve still answers `SQL_C_CHAR`, and none can reach it today. This is purely about what the next person is told.

> [!NOTE]

shiwanigupta0809:
Read the full diff, `variant_c_type` at head, `type_rules::resolve_default_c_type`, `is_valid_c_type`, the `last_variant_base` lifecycle, the benchmark harness fold, and both doc locations. Checked the parity claims against the msodbcsql 18.6 source. Not re-raising smallmoney coverage, the stale non-goal notes, the index entry, the tinyint signedness, the exhaustive-match framing, or the temporal fold in the harness — that thread is thorough and all six are resolved.

### Verified

**The staleness question closes cleanly.** This change raises the cost of a stale answer — `SQL_C_NUMERIC` vs `SQL_C_SLONG` now decides `Decimal` vs `int` in mssql-python, where before a stale answer mostly produced a differently-shaped string. So I traced the lifetime rather than assuming: `last_variant_base` is written only in `resume_row_to_column` (`get_data.rs:582`), keyed to its column, filtered on that column at read (`col_attribute.rs:157`), and cleared by `reset_row_stream()` (`stmt.rs:929`), which `begin_row()` calls on every fetched row. A variant column holding `int` in row 1 and `decimal` in row 2 cannot answer `SQL_C_SLONG` for row 2.

**The driver can service `SQL_C_UTINYINT` on this path.** The earlier thread cited `fetch_convert.rs`, which is the bound-fetch converter — a different path from the one a variant read takes. Confirming via the right one: `SQLGetData` reaches `convert_typed_c` (`get_data.rs:1287`), which routes `is_integer_c_target` to `convert_integer_c`, and `SQL_C_UTINYINT` is in `is_integer_c_type`. Same conclusion, so the finding stands — just checked where it actually happens.

**`SQL_C_SS_TIME2` is not version-gated at the delivery gate.** `is_valid_c_type` lists it unconditionally (`type_rules.rs:202-203`), so reporting it to an ODBC 3.0 application does not create a second "reports what it cannot deliver". That was my first guess at a problem and it is not one.

### Suggestion

**The tinyint arm was fixed on the stated principle of internal consistency with `resolve_default_c_type`. Two arms below, `TimeN` / `DateTimeOffsetN` still break that same consistency — and the new divergence row justifies the gap with exactly the kind of unevidenced consumer claim this PR's new guardrail forbids.**

The tinyint fix is justified in its own comment as:

> consistent with this driver's own `type_rules::resolve_default_c_type`, which resolves `SQL_TINYINT` to `SQL_C_UTINYINT` for a column binding.

That is the right standard, and it is the standard the temporal arms fail. `resolve_default_c_type` gates this exact pair of types on the declared version:

```rust
SQL_SS_TIME2 if is_3_80 => SQL_C_SS_TIME2,
SQL_SS_TIME2 => SQL_C_BINARY,
SQL_SS_TIMESTAMPOFFSET if is_3_80 => SQL_C_SS_TIMESTAMPOFFSET,
SQL_SS_TIMESTAMPOFFSET => SQL_C_BINARY,
```

— `type_rules.rs:324-327`. `variant_c_type` does not:

```rust
TdsDataType::TimeN => SQL_C_SS_TIME2,                          // col_attribute.rs:536
TdsDataType::DateTimeOffsetN => SQL_C_SS_TIMESTAMPOFFSET,      // col_attribute.rs:541
```

So under `SQL_OV_ODBC3`, one `time` value is described two ways depending on how it arrives — `SQL_C_BINARY` as a column resolved from `SQL_C_DEFAULT`, `SQL_C_SS_TIME2` as a variant. That is the same sentence that was used to justify changing `Int1`, and msodbcsql gates both consistently, which is why its variant answer moves with the declared version at all.

There is a structural reason the arm cannot currently do better, and it is worth naming: `col_attribute.rs` never reads `OdbcVersion` — no import, no ENV access anywhere in the file — so `variant_c_type` has no way to give a version-dependent answer even if it wanted to. Notably, PR #481 adds precisely that ENV-before-STMT read to `get_data.rs` for the same need, so the pattern and its lock-ordering rationale already exist to copy.

What I would actually change is smaller than the code, though. The divergence row reads:

> Answering the real type regardless of the declared version gives the caller strictly more information, and a 3.x application that does not know the `SQL_C_SS_*` types will not ask for them.

The second clause is a claim about how an unnamed consumer behaves, offered without a citation — and this PR's own new guardrail, twelve lines added to the instructions file, says:

> **A divergence justified by "the consumer does not need parity here" is a claim about the consumer's code, and needs the same evidence as a claim about msodbcsql: cite the file and function you read.**

I do not think the conclusion is wrong; answering the real type probably is the better behaviour, and the impact is small since a 3.0-declaring application receiving `-154` will fall into its own unknown-type branch. But the row states it as settled on reasoning the same PR just ruled insufficient, and it omits that the driver contradicts itself here under 3.0. Either dropping the consumer clause and resting the row on "additive, and the type is deliverable at any declared version" (which I verified above, so it is defensible on evidence the PR already has), or noting the internal split with a follow-up, would leave the row meeting the bar the PR is setting.

The rest holds up well. Measuring all 24 base types rather than adding the one row that was asked for, and enumerating the match so the next type is a compile error rather than a silent `SQL_C_CHAR`, are both the version of the fix that stops the class instead of the instance.

shiwanigupta0809 (commented):
_Reviewed - see https://github.com/microsoft/mssql-rs/pull/483#issuecomment-5523323492_

David-Engel:
> the row states it as settled on reasoning the same PR just ruled insufficient

That one lands. I wrote a guardrail forbidding divergences justified by uncited claims about consumer behaviour, and then justified a divergence with an uncited claim about consumer behaviour, twelve lines later. Second time this review round that I have been caught doing the thing the PR is about — the pattern is apparently easier to describe than to avoid.

I verified all three of your technical claims rather than taking them on trust, since that is the lesson:

- `resolve_default_c_type` does gate both types on the declared version — `type_rules.rs:324-327`, `SQL_SS_TIME2 if is_3_80 => SQL_C_SS_TIME2` else `SQL_C_BINARY`.
- `is_valid_c_type` does accept both `SQL_C_SS_*` types unconditionally — `type_rules.rs:202-203`.
- `col_attribute.rs` has no version access at all — grepped for `OdbcVersion`, `odbc_version`, `EnvHandle`, `is_3_80`; zero hits.

So the internal split is real, and so is the structural reason it cannot be fixed in that arm today.

**I took both of your options rather than choosing between them**, because each fixes a different defect. The row now rests on `is_valid_c_type` accepting both types unconditionally — the driver will service the target it reports, which is evidence already in the tree rather than a guess about callers. The uncited clause is gone. And the row no longer omits the contradiction: it states plainly that `resolve_default_c_type` gates what `variant_c_type` does not, so under `SQL_OV_ODBC3` the same `time` value is described one way as a variant and another as a column resolved from `SQL_C_DEFAULT`.

Reclassified from **Deliberate** to a tracked gap, **AB#47830**, which carries the measurement, the `is_valid_c_type` finding, and your note that PR #481 is already landing the ENV-before-STMT read this needs. The two arms in `variant_c_type` now carry the same note, so the split is visible where someone would actually hit it rather than only in the table.

I did not attempt the version gating here. It needs ENV access from `col_attribute.rs` with the lock ordering #481 is establishing, and doing it independently would risk a second, divergent implementation of the same pattern.

Also noted your correction on the delivery path — I had cited `fetch_convert.rs`, the bound-fetch converter, where a variant read actually goes through `convert_typed_c` in `get_data.rs`. Same conclusion, wrong citation; worth having on the record given this PR's subject.

Theekshna (commented):
Unattended sweep review. No new findings — this PR (and its stacked exchange between Theekshna and David-Engel already on the thread) covers the six extra checks in full: msodbcsql parity independently re-derived against `rgbSRV2CTYPE` in the reference tree (`SQLDECIMALN`/`SQLNUMERICN`/`SQLMONEYN` → `SQL_C_NUMERIC`, matching this PR's core change), the full 24-base-type variant table audited not just the touched arms, divergence docs (instructions.md + typed-columnar-fetch-plan.md + get_data_test.cpp comment) now consistent with the tracked-gap reclassification, and no AI slop in the enumerated-match rewrite. Some CI jobs (Build Linux, Test MacOS, coverage-report, overall PR validation) were still pending at review time; everything that has reported so far is green.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -506,30 +506,47 @@ fn num_prec_radix(meta: &ColumnMetadata) -> SqlLen {
...
+        // This is the type of the *value*, not a promise that SQL_NUMERIC_STRUCT
+        // is an available SQLGetData target - it is not yet (AB#47816).
         TdsDataType::Decimal
         | TdsDataType::DecimalN
         | TdsDataType::Numeric
         | TdsDataType::NumericN
         | TdsDataType::Money
         | TdsDataType::Money4
-        | TdsDataType::MoneyN => SQL_C_CHAR,
+        | TdsDataType::MoneyN => SQL_C_NUMERIC,
```
saurabh500, line 563: Non-blocking, and the direction here is right — reporting `SQL_C_NUMERIC` is what msodbcsql does, and the old `SQL_C_CHAR` was the actual divergence. Two things worth naming, though, because this line newly creates a report/service mismatch on the variant path that did not exist before:

- The driver reports `SQL_C_NUMERIC` but does not yet accept it as a `SQLGetData` / `SQLBindCol` target (`HYC00`, AB#47816). Before this PR the variant answer (`SQL_C_CHAR`) was always serviceable.
- Microsoft's ODBC data-type-mapping workflow tells a caller to bind the returned C type. mssql-python is safe because it re-fetches numerics as `SQL_C_CHAR` regardless (verified in `ddbc_bindings.cpp`), but a generic SQL-Server-aware consumer that follows that workflow and requests `SQL_C_NUMERIC` would get `HYC00` here where msodbcsql succeeds.

The divergence-table row already acknowledges this ("the target still has to become real for any caller that trusts the reported type"), so this is a question about sequencing, not correctness: was gating the reported type on AB#47816 considered, so report and service land together? If shipping the parity-correct report first is intended, consider making this comment say plainly that a caller binding the returned type gets `HYC00` until AB#47816 lands, rather than the softer "not a promise ... is an available SQLGetData target".

saurabh500 (commented):
_Unattended sweep review: prepared by automation, not pre-screened by a human. Event is COMMENT only._

**Verdict: no blocking issues — the core change is correct and unusually well-verified. One non-blocking suggestion about sequencing, left inline.**

I re-derived the parity claim independently rather than taking the description or prior reviews on trust:

- **msodbcsql is the contract, and every changed arm matches it.** Read the reference driver's `rgbSRV2CTYPE[]` (`Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp`): `SQLDECIMAL(N)` / `SQLNUMERIC(N)` / `SQLMONEY` / `SQLMONEY4` / `SQLMONEYN` all map to `SQL_C_NUMERIC`, and `SQLINT1` maps to `SQL_C_UTINYINT`. The numeric and tinyint-signedness fixes are exactly the reference answers; the date/datetime 3.x spellings (91/93 vs msodbcsql's 2.x 9/11) are the deliberate, documented canonicalization divergence.
- **Consumer safety holds for the real consumer.** Traced mssql-python's `ddbc_bindings.cpp`: a variant reporting `SQL_C_NUMERIC` maps via `MapVariantCTypeToSQLType` to `SQL_NUMERIC`, then re-fetches with `SQL_C_CHAR` on all three fetch paths (row-wise `SQLGetData`, bound-column `SQLBindCol`, Arrow). It never requests `SQL_C_NUMERIC` as a target, so the AB#47816 gap does not bite it.
- **The runnable unit test bites.** Reverting the numeric arm to `SQL_C_CHAR` fails `variant_c_type_covers_the_base_types` at the `Numeric` case (`left: 1, right: 2`). The enumerated match (no `_` arm) makes adding a `TdsDataType` a compile error, which is the property the comment now claims.
- CI checks report passing, including the cross-repo mssql-python suite. The e2e parity tests need a live server + retail msodbcsql driver and were not re-run locally.

**Blocking:** none.

**Suggestion:** one, inline at `mssql-odbc/src/api/col_attribute.rs` — the variant path now reports a C type the driver does not yet service, which is harmless for mssql-python but worth a sequencing/wording note for a generic consumer.

**Nit:** none.

mssql-odbc/src/api/col_attribute.rs
David-Engel: Both points are fair, and the second is the sharper one — this line does newly create a report/service mismatch that did not exist while the answer was `SQL_C_CHAR`. Done what you asked: the comment now names the failure instead of gesturing at it.

**On sequencing — yes, it was considered explicitly, and report-first was chosen deliberately.** Implementing `SQL_C_NUMERIC` output in this PR was on the table and was weighed against shipping the report alone. The reasoning:

Keeping `SQL_C_CHAR` is not a safe holding position. It isn't a narrower answer, it's the *wrong* answer, and it fails **silently** — mssql-python routes on this value alone and handed back `str` for a decimal with nothing anywhere to indicate the type was wrong. That is AB#47702. Trading a silent wrong-shape bug affecting the primary consumer for a defined `HYC00` that only reaches a caller asking for a struct this driver has never offered on *any* path seemed like the right trade, especially since `HYC00` is precisely "optional feature not implemented" and is discoverable.

Gating the report on AB#47816 would have meant leaving `Decimal` broken in mssql-python until `SQL_NUMERIC_STRUCT` output landed — a much larger change needing ARD precision/scale plumbing.

**What I changed.** The comment previously said the answer was "not a promise that `SQL_NUMERIC_STRUCT` is an available `SQLGetData` target," which is softer than the truth. It now states that a caller binding the returned type gets `HYC00` until AB#47816 lands where msodbcsql succeeds, that the mismatch is new here, and why the order was chosen. Both divergence-table rows carry it too — the `SQL_C_NUMERIC` row is reclassified from "a missing target" to "a report/service mismatch," which is a different and more urgent thing.

I also updated **AB#47816** itself to record your point: the generic-consumer risk, the ODBC workflow it comes from, and that mssql-python's `SQL_C_CHAR` re-fetch means no shipping consumer is broken today. So the follow-up now carries the argument for its own priority rather than reading as routine conformance work.

One measurement while I was in there, since it bounds the exposure: `is_valid_c_type` accepts `SQL_C_NUMERIC` (`type_rules.rs:185`), so the rejection is in the conversion layer rather than the validation gate — which is what makes AB#47816 an additive change to `convert_typed_c` rather than a redesign.

> [!NOTE]

Theekshna (approved):

Theekshna (commented):
Unattended re-review at 59b1d9c9. Head moved past my last marker only by a merge from main (nine already-merged, unrelated commits, e.g. #490/#489/#461/#485/#444/#446/#459) plus one genuine new commit on this branch: `ccd6f07` "State the SQL_C_NUMERIC report/service mismatch plainly." I diffed against the merge-base with main (`f8c75d9a`) to isolate that from the noise.

That commit is comment/doc-only — it expands the code comment on ariant_c_type's decimal/numeric/money arm and the two matching divergence-table rows in 	yped-columnar-fetch-plan.md, to say plainly that binding the newly-reported SQL_C_NUMERIC currently gets HYC00 until AB#47816 lands, and to record why report-first was chosen anyway (a defined "not implemented" beats silently mis-typed data). No logic changed; ariant_c_type's match arms are byte-identical to the version I audited at 77a056b.

| Severity | Count |
| --- | ---: |
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

Checked against my own bar for check 5: this is exactly the kind of comment SKILL.md exempts from "slop" — it records a *why* (a deliberate tradeoff and an ordering constraint against AB#47816), not a restatement of the code — so I'm not flagging its length. Divergence docs and code comment now say the same thing, consistently. No new msodbcsql-parity surface, no new test-relevant branch, and the PR description (measurement tables, MapVariantCTypeToSQLType routing, AB#47702/AB#47816 links) still matches the current diff. gh pr checks shows nothing failing.

Theekshna (approved):

PR #444, opened 2026-09-01
mssql-odbc: buffer and transcode data-at-execution parameters with mismatched wideness

Vahid-b:
## Summary

Fixes `HYC00` ("Parameter conversion not yet implemented") when a data-at-execution character parameter's C type and declared SQL type disagree on wideness — e.g. `SQL_C_WCHAR` streamed against a narrow `SQL_VARCHAR`/`SQL_LONGVARCHAR`. mssql-python binds every character parameter's data-at-execution path as `SQL_C_WCHAR`, including narrow (ASCII) values, once the value exceeds its ~4000-character inline threshold, so this made any such bound string fail outright.

## Background

Found while validating [AB#47709](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47709) ("PLP ColumnSize 0 causes NVARCHAR(MAX) and XML SQLGetData failures"), which turned out to already be fixed by #396 before this investigation began. Comparing the before/after CI logs for that item's 23 listed failing tests showed 21/23 now pass, but 2 (`test_encoding_decoding_boundary_conditions`, `test_encoding_decoding_stress_test_comprehensive`) still failed — on a different, unrelated signature: `HYC00 Parameter conversion not yet implemented` raised from an `INSERT` binding a 4001-character parameter, not from `SQLGetData`. Filed separately as #443.

## Root cause

`dae_placeholder_type` required the C type and the SQL type to agree on wideness before accepting a stream at all — `SQL_C_WCHAR` only streamed against a wide SQL type, `SQL_C_CHAR` only against a narrow one. `SQLPutData` wrote each chunk straight to the wire untranscoded, so a mismatched pairing really would corrupt the value if streamed as-is (tracked as [AB#47590](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47590)).

mssql-python's `param_detect.hpp` always declares `SQL_C_WCHAR` for a streamed character parameter — even for ASCII values it also declares as the *narrow* `SQL_VARCHAR` — a documented convention in its own source ("a long-standing alias in the Python layer"). That's exactly the pairing this driver refused.

## Fix

[...]

PR #485, opened 2026-09-02
Enforce safety docs for private unsafe functions

### LINKED ISSUE #474: Enforce # Safety docs on private unsafe fns via clippy check-private-items
### Problem statement

`clippy::missing_safety_doc` is on by default, but it **only checks publicly-exported items**. Every `unsafe fn` that is private or `pub(crate)` can ship with no `# Safety` section and CI stays green.

That is most of the unsafe surface in `mssql-odbc`. The crate exports its ODBC entry points through `#[unsafe(no_mangle)] pub extern "C"` wrappers, so nearly all of the interesting unsafe helpers — the ones that actually write into application-supplied buffers — sit behind them as private or `pub(crate)` functions and are invisible to the lint.

This was found while reviewing #446. Four new `unsafe fn`s were added to `mssql-odbc/src/api/get_data.rs` (`try_write_complete_buffered_string`, `try_write_complete_buffered_decimal`, `try_write_exact_buffered_scalar`, `try_write_direct_captured_string_chunk`) with no documented safety contract, and clippy said nothing. They are documented now, but only because someone read the diff.

The contracts are not cosmetic. These helpers take a raw `TargetValuePtr` plus a length the driver cannot validate, and two different length conventions sit right next to each other: `buffer_length` is a **byte** count for both `SQL_C_CHAR` and `SQL_C_WCHAR` per the ODBC spec, while `try_write_direct_captured_string_chunk` takes `buf_elements`, an **element** count. A caller confusing those on the wide path overruns the application's buffer by 2x. That is exactly what a `# Safety` section is for.

### Proposed solution

Add a `clippy.toml` at the repository root:

```toml
check-private-items = true
```

This option (stable since Rust 1.76; the repo pins 1.95) extends `missing_safety_doc` and `unnecessary_safety_doc` to private items, so every `unsafe fn` must state its caller obligations regardless of visibility.

**Measured cost of turning it on.** Against `49f2ccb4`, `cargo clippy --workspace --all-features --all-targets`:

- **52 sites**, all in `mssql-odbc`. `mssql-tds`, `mssql-js`, `mssql-tds-cli`, and `mssql-mock-tds` are already clean.
- `clippy::missing_safety_doc` is the **only** lint that newly fires workspace-wide. No other lint category is affected, so this does not turn into an open-ended cleanup.

| Count | File |
|---:|---|
| 7 | `src/api/catalog.rs` |
| 3 | `src/api/describe_col.rs` |
| 3 | `src/api/put_data.rs` |
| 3 | `src/api/set_connect_attr.rs` |
| 2 | `src/api/bind_param.rs` |
| 2 | `src/api/close_cursor.rs` |
| 2 | `src/api/col_attribute.rs` |
| 2 | `src/api/describe_param.rs` |
| 2 | `src/api/driver_connect.rs` |
| 2 | `src/api/get_data.rs` |
| 2 | `src/api/set_stmt_attr.rs` |
| 1 each | `bind_col.rs`, `cancel.rs`, `connect.rs`, `disconnect.rs`, `end_tran.rs`, `exec_direct.rs`, `execute.rs`, `fetch_scroll.rs`, `get_connect_attr.rs`, `get_desc_field.rs`, `get_diag.rs`, `get_env_attr.rs`, `get_functions.rs`, `get_info.rs`, `get_type_info.rs`, `more_results.rs`, `num_result_cols.rs`, `param_data.rs`, `prepare.rs`, `row_count.rs`, `set_desc_field.rs`, `set_env_attr.rs` |

Most are the `sql_*` / `sql_*_impl` pairs behind each ODBC entry point, which share one contract shape — the handle must be valid, and each output pointer must be writable for its stated length. They should document quickly and consistently.

Suggested sequencing, so this doesn't land as one large unreviewable commit:

1. Add `clippy.toml` with the lint allowed at crate level in `mssql-odbc`, so the option is active and the other crates are protected from regressing immediately.
2. Document the 52 sites in a few reviewable batches, grouped by area.
3. Remove the crate-level allow.

`mssql-py-core` is excluded from the workspace and needs its own clippy pass; it should be checked separately as part of this.

### Affected crate

mssql-odbc

### Alternatives considered

- **Leave it as-is and rely on review.** This is the status quo, and #446 is the evidence it does not hold — four undocumented `unsafe fn`s on the hot `SQLGetData` path reached a PR with green CI.
- **Setting it in `[workspace.lints.clippy]` in the root `Cargo.toml`.** Not possible — this was tried and does not work. Cargo's `[lints]` table sets lint *levels*; `check-private-items` is a lint *configuration option*, and there is no Cargo syntax for those. Both forms fail:
  - `check-private-items = true` → `error: invalid type: boolean 'true', expected a string or map`
  - `check-private-items = { level = "warn" }` → Cargo accepts the syntax but forwards it as `-W clippy::check_private_items`, giving `warning[E0602]: unknown lint`, because it is treated as a lint name rather than an option.

  Configuration options only come from `clippy.toml` (or `.clippy.toml`), which is why the file has to be added even though the repo already centralizes lint levels in `Cargo.toml`. Worth noting that `clippy.toml` is resolved from the crate being linted upward, so a single file at the repository root covers the whole workspace.

- **`clippy::undocumented_unsafe_blocks`.** Requires a `// SAFETY:` comment on every `unsafe {}` block rather than a `# Safety` section on every `unsafe fn`. It is a `restriction` lint and much noisier — this codebase has ~154 `unsafe fn`s in `mssql-odbc` and many more blocks. Worth considering later, but `check-private-items` is the higher value-per-unit-of-churn change and should come first.
- **`missing_panics_doc` / `missing_errors_doc`.** Also governed by `check-private-items`, but both are `pedantic` and off by default, so enabling the option does not pull them in. Out of scope here.

### Additional context

Documenting these is a genuine correctness aid rather than box-ticking, because the driver sits behind a C ABI where nothing can be validated at runtime. The obligations originate at the FFI boundary and are propagated inward by `unsafe fn`, so the place to record them is on each function that carries them.

Config reference: <https://rust-lang.github.io/rust-clippy/master/index.html#missing_safety_doc>

Follow-up to #446.

saurabh500:
## Description

Private and crate-visible `unsafe fn`s were outside Clippy's `missing_safety_doc` coverage, leaving the ODBC driver's raw-handle and application-buffer contracts dependent on manual review.

This change:

- enables `check-private-items` from a repository-root `clippy.toml`;
- documents every resulting ODBC safety obligation, including deferred parameter and rowset buffers;
- removes unnecessary `unsafe fn` signatures from six Windows CNG helpers and documents the four boundaries that remain unsafe;
- uses an unaligned read while scanning `SQL_C_WCHAR` data supplied to `SQLPutData`, matching ODBC's unaligned-buffer contract.

No lint allows or exemptions are added.

## Related Issues

Closes #474

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes (`./scripts/bclippy.sh`)
- [ ] `cargo btest` passes - full live-server integration coverage requires a configured SQL Server; the targeted `mssqlodbc` suite passes (1,223 tests)
- [x] New/changed functionality has tests - Clippy is the regression check for private unsafe functions; existing ODBC tests cover the pointer-read path
- [x] Public API changes are documented - N/A, no public API change

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,0 +1,1 @@
+check-private-items = true
--- a/mssql-odbc/src/api/catalog.rs
+++ b/mssql-odbc/src/api/catalog.rs
@@ -691,6 +691,10 @@ pub(crate) unsafe fn sql_tables_w(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. Each
+/// non-null name pointer must be readable for its paired length in UTF-16 code
+/// units, or through a NUL terminator when that length is `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 unsafe fn sql_tables_w_impl(
     statement_handle: SqlHandle,
@@ -834,6 +838,10 @@ pub(crate) unsafe fn sql_columns_w(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. Each
+/// non-null name pointer must be readable for its paired length in UTF-16 code
+/// units, or through a NUL terminator when that length is `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 unsafe fn sql_columns_w_impl(
     statement_handle: SqlHandle,
@@ -978,6 +986,10 @@ pub(crate) unsafe fn sql_primary_keys_w(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. Each
+/// non-null name pointer must be readable for its paired length in UTF-16 code
+/// units, or through a NUL terminator when that length is `SQL_NTS`.
 unsafe fn sql_primary_keys_w_impl(
     statement_handle: SqlHandle,
     catalog_name: *const SqlWChar,
@@ -1113,6 +1125,10 @@ pub(crate) unsafe fn sql_foreign_keys_w(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. Each
+/// non-null name pointer must be readable for its paired length in UTF-16 code
+/// units, or through a NUL terminator when that length is `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 unsafe fn sql_foreign_keys_w_impl(
     statement_handle: SqlHandle,
@@ -1307,6 +1323,10 @@ pub(crate) unsafe fn sql_statistics_w(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. Each
+/// non-null name pointer must be readable for its paired length in UTF-16 code
+/// units, or through a NUL terminator when that length is `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 unsafe fn sql_statistics_w_impl(
     statement_handle: SqlHandle,
@@ -1471,6 +1491,10 @@ pub(crate) unsafe fn sql_special_columns_w(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. Each
+/// non-null name pointer must be readable for its paired length in UTF-16 code
+/// units, or through a NUL terminator when that length is `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 unsafe fn sql_special_columns_w_impl(
     statement_handle: SqlHandle,
@@ -1655,6 +1679,10 @@ pub(crate) unsafe fn sql_procedures_w(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. Each
+/// non-null name pointer must be readable for its paired length in UTF-16 code
+/// units, or through a NUL terminator when that length is `SQL_NTS`.
 unsafe fn sql_procedures_w_impl(
     statement_handle: SqlHandle,
     catalog_name: *const SqlWChar,
--- a/mssql-odbc/src/api/catalog.rs
+++ b/mssql-odbc/src/api/catalog.rs
@@ -651,7 +651,9 @@ fn apply_catalog_metadata(
 /// Implements `SQLTablesW`.
 ///
 /// # Safety
-/// Each name pointer must be null or reference `*_len` readable UTF-16 units.
+/// Each name pointer must be null, reference its paired length in readable
+/// UTF-16 units, or be readable through a NUL terminator when that length is
+/// `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 pub(crate) unsafe fn sql_tables_w(
     statement_handle: SqlHandle,
@@ -798,7 +800,9 @@ fn sql_tables_w_safe(
 /// Implements `SQLColumnsW`.
 ///
 /// # Safety
-/// Each name pointer must be null or reference `*_len` readable UTF-16 units.
+/// Each name pointer must be null, reference its paired length in readable
+/// UTF-16 units, or be readable through a NUL terminator when that length is
+/// `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 pub(crate) unsafe fn sql_columns_w(
     statement_handle: SqlHandle,
@@ -953,7 +957,9 @@ fn sql_columns_w_safe(
 /// Implements `SQLPrimaryKeysW`.
 ///
 /// # Safety
-/// Each name pointer must be null or reference `*_len` readable UTF-16 units.
+/// Each name pointer must be null, reference its paired length in readable
+/// UTF-16 units, or be readable through a NUL terminator when that length is
+/// `SQL_NTS`.
 pub(crate) unsafe fn sql_primary_keys_w(
     statement_handle: SqlHandle,
     catalog_name: *const SqlWChar,
@@ -1073,7 +1079,9 @@ fn sql_primary_keys_w_safe(
 /// Implements `SQLForeignKeysW`.
 ///
 /// # Safety
-/// Each name pointer must be null or reference `*_len` readable UTF-16 units.
+/// Each name pointer must be null, reference its paired length in readable
+/// UTF-16 units, or be readable through a NUL terminator when that length is
+/// `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 pub(crate) unsafe fn sql_foreign_keys_w(
     statement_handle: SqlHandle,
@@ -1283,7 +1291,9 @@ fn sql_foreign_keys_w_safe(
 /// Implements `SQLStatisticsW`.
 ///
 /// # Safety
-/// Each name pointer must be null or reference `*_len` readable UTF-16 units.
+/// Each name pointer must be null, reference its paired length in readable
+/// UTF-16 units, or be readable through a NUL terminator when that length is
+/// `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 pub(crate) unsafe fn sql_statistics_w(
     statement_handle: SqlHandle,
@@ -1448,7 +1458,9 @@ fn sql_statistics_w_safe(
 /// Implements `SQLSpecialColumnsW`.
 ///
 /// # Safety
-/// Each name pointer must be null or reference `*_len` readable UTF-16 units.
+/// Each name pointer must be null, reference its paired length in readable
+/// UTF-16 units, or be readable through a NUL terminator when that length is
+/// `SQL_NTS`.
 #[allow(clippy::too_many_arguments)]
 pub(crate) unsafe fn sql_special_columns_w(
     statement_handle: SqlHandle,
@@ -1646,7 +1658,9 @@ fn sql_special_columns_w_safe(
 /// Implements `SQLProceduresW`.
 ///
 /// # Safety
-/// Each name pointer must be null or reference `*_len` readable UTF-16 units.
+/// Each name pointer must be null, reference its paired length in readable
+/// UTF-16 units, or be readable through a NUL terminator when that length is
+/// `SQL_NTS`.
 pub(crate) unsafe fn sql_procedures_w(
     statement_handle: SqlHandle,
     catalog_name: *const SqlWChar,
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -40,6 +40,12 @@ use mssql_tds::query::metadata::PlpEncoding;
 /// - Supports `SQL_C_CHAR` and `SQL_C_WCHAR` for text retrieval.
 /// - Supports incremental row resume and chunked PLP retrieval via
 ///   `read_active_plp_chunk`.
+///
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`.
+/// `target_value_ptr`, when non-null, must be writable for `buffer_length`
+/// bytes, including when `target_type` is `SQL_C_WCHAR`. `strlen_or_ind_ptr`,
+/// when non-null, must be writable for one `SqlLen`.
 pub(crate) unsafe fn sql_get_data(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
@@ -70,6 +76,11 @@ pub(crate) unsafe fn sql_get_data(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`.
+/// `target_value_ptr`, when non-null, must be writable for `buffer_length`
+/// bytes, including when `target_type` is `SQL_C_WCHAR`. `strlen_or_ind_ptr`,
+/// when non-null, must be writable for one `SqlLen`.
 unsafe fn sql_get_data_impl(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -44,8 +44,10 @@ use mssql_tds::query::metadata::PlpEncoding;
 /// # Safety
 /// `statement_handle` must be null or point to a live `StmtHandle`.
 /// `target_value_ptr`, when non-null, must be writable for `buffer_length`
-/// bytes, including when `target_type` is `SQL_C_WCHAR`. `strlen_or_ind_ptr`,
-/// when non-null, must be writable for one `SqlLen`.
+/// bytes for variable-width targets, including `SQL_C_WCHAR`, where the length
+/// is still measured in bytes. For a fixed-width target it must be writable for
+/// the full size of `target_type`, even when `buffer_length` is zero or smaller.
+/// `strlen_or_ind_ptr`, when non-null, must be writable for one `SqlLen`.
 pub(crate) unsafe fn sql_get_data(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
@@ -79,8 +81,10 @@ pub(crate) unsafe fn sql_get_data(
 /// # Safety
 /// `statement_handle` must be null or point to a live `StmtHandle`.
 /// `target_value_ptr`, when non-null, must be writable for `buffer_length`
-/// bytes, including when `target_type` is `SQL_C_WCHAR`. `strlen_or_ind_ptr`,
-/// when non-null, must be writable for one `SqlLen`.
+/// bytes for variable-width targets, including `SQL_C_WCHAR`, where the length
+/// is still measured in bytes. For a fixed-width target it must be writable for
+/// the full size of `target_type`, even when `buffer_length` is zero or smaller.
+/// `strlen_or_ind_ptr`, when non-null, must be writable for one `SqlLen`.
 unsafe fn sql_get_data_impl(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
--- a/mssql-odbc/src/api/put_data.rs
+++ b/mssql-odbc/src/api/put_data.rs
@@ -451,8 +451,8 @@ mod tests {
 
     #[test]
     fn nts_byte_count_reads_a_misaligned_wide_buffer() {
-        let mut storage = [0u8; 8];
-        let ptr = unsafe { storage.as_mut_ptr().add(1) };
+        let mut storage = [0u16; 4];
+        let ptr = unsafe { storage.as_mut_ptr().cast::<u8>().add(1) };
         assert_ne!(
             ptr as usize % std::mem::align_of::<u16>(),
             0,
--- a/mssql-odbc/src/api/put_data.rs
+++ b/mssql-odbc/src/api/put_data.rs
@@ -29,11 +29,11 @@ use crate::handles::{HandleType, StmtHandle, handle_from_raw};
 /// - `data_ptr`, when `strlen_or_ind` is a positive byte count, must be
 ///   readable for that many bytes and must remain valid for the duration of
 ///   this call.
-/// - `data_ptr`, when `strlen_or_ind` is `SQL_NTS`, must be non-null, aligned
-///   for the bound parameter's C type, and NUL-terminated within an allocation
-///   it owns: the terminator search reads `u16` units for `SQL_C_WCHAR` and
-///   `u8` units otherwise, and runs off the end of the allocation if no
-///   terminator is present.
+/// - `data_ptr`, when `strlen_or_ind` is `SQL_NTS`, must be non-null and
+///   NUL-terminated within an allocation it owns: the terminator search
+///   reads potentially unaligned `u16` units for `SQL_C_WCHAR` and `u8` units
+///   otherwise, and runs off the end of the allocation if no terminator is
+///   present.
 pub(crate) unsafe fn sql_put_data(
     statement_handle: SqlHandle,
     data_ptr: SqlPointer,
@@ -50,6 +50,11 @@ pub(crate) unsafe fn sql_put_data(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. For a
+/// positive `strlen_or_ind`, `data_ptr` must be readable for that many bytes.
+/// For `SQL_NTS`, it must be non-null and readable through a NUL terminator as
+/// the bound C type.
 unsafe fn sql_put_data_impl(
     statement_handle: SqlHandle,
     data_ptr: SqlPointer,
@@ -70,11 +75,15 @@ unsafe fn sql_put_data_impl(
     unsafe { sql_put_data_safe(statement_handle, stmt, data_ptr, strlen_or_ind) }
 }
 
+/// # Safety
+/// `data_ptr` must be non-null and point to an allocation readable through a
+/// NUL terminator. The allocation is read as potentially unaligned `u16` units
+/// when `c_type` is `SQL_C_WCHAR`, and as `u8` units otherwise.
 unsafe fn nts_byte_count(data_ptr: SqlPointer, c_type: i16) -> usize {
     if c_type == SQL_C_WCHAR {
         let ptr = data_ptr as *const u16;
         let mut units = 0usize;
-        while unsafe { *ptr.add(units) } != 0 {
+        while unsafe { ptr.add(units).read_unaligned() } != 0 {
             units += 1;
         }
         units * std::mem::size_of::<u16>()
@@ -88,6 +97,11 @@ unsafe fn nts_byte_count(data_ptr: SqlPointer, c_type: i16) -> usize {
     }
 }
 
+/// # Safety
+/// `statement_handle` must identify the live `stmt`. For a positive
+/// `strlen_or_ind`, `data_ptr` must be readable for that many bytes. For
+/// `SQL_NTS`, it must be non-null and readable through a NUL terminator as the
+/// currently bound C type.
 unsafe fn sql_put_data_safe(
     statement_handle: SqlHandle,
     stmt: &StmtHandle,
--- a/mssql-odbc/src/api/put_data.rs
+++ b/mssql-odbc/src/api/put_data.rs
@@ -449,6 +449,23 @@ mod tests {
         assert_eq!(unsafe { nts_byte_count(ptr, 0) }, 1);
     }
 
+    #[test]
+    fn nts_byte_count_reads_a_misaligned_wide_buffer() {
+        let mut storage = [0u8; 8];
+        let ptr = unsafe { storage.as_mut_ptr().add(1) };
+        assert_ne!(
+            ptr as usize % std::mem::align_of::<u16>(),
+            0,
+            "test pointer must be misaligned"
+        );
+        unsafe {
+            ptr.cast::<u16>().write_unaligned(0x0041);
+            ptr.add(2).cast::<u16>().write_unaligned(0x0000);
+        }
+
+        assert_eq!(unsafe { nts_byte_count(ptr.cast(), SQL_C_WCHAR) }, 2);
+    }
+
     /// A binding can disappear under an open sequence via
     /// `SQLFreeStmt(SQL_RESET_PARAMS)`, leaving no C type to size `SQL_NTS`
     /// with. The call is rejected instead of guessing.
--- a/mssql-odbc/src/api/bind_col.rs
+++ b/mssql-odbc/src/api/bind_col.rs
@@ -55,6 +55,10 @@ pub(crate) unsafe fn sql_bind_col(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. When
+/// non-null, `target_value_ptr` and `strlen_or_ind_ptr` must remain valid for
+/// the bound rowset until the column is unbound or the statement is freed.
 unsafe fn sql_bind_col_impl(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
--- a/mssql-odbc/src/api/bind_col.rs
+++ b/mssql-odbc/src/api/bind_col.rs
@@ -35,7 +35,10 @@ use crate::handles::{DescHandle, HandleType, StmtHandle, handle_from_raw};
 /// elements of `buffer_length` bytes for a character or binary target, or of
 /// the full C type size for a fixed-width target, even when `buffer_length` is
 /// zero or smaller. `strlen_or_ind_ptr`, when non-null, must be writable for
-/// `SQL_ATTR_ROW_ARRAY_SIZE` `SqlLen` values.
+/// `SQL_ATTR_ROW_ARRAY_SIZE` `SqlLen` values. When
+/// `SQL_ATTR_ROW_BIND_OFFSET_PTR` is non-null, these bound-buffer extents begin
+/// at the base plus the pointed-to byte offset, so each allocation must also
+/// cover that leading displacement.
 pub(crate) unsafe fn sql_bind_col(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
@@ -68,7 +71,9 @@ pub(crate) unsafe fn sql_bind_col(
 /// or binary target, or of the full C type size for a fixed-width target, even
 /// when `buffer_length` is zero or smaller. `strlen_or_ind_ptr`, when non-null,
 /// must remain valid and be writable for `SQL_ATTR_ROW_ARRAY_SIZE` `SqlLen`
-/// values.
+/// values. When `SQL_ATTR_ROW_BIND_OFFSET_PTR` is non-null, these bound-buffer
+/// extents begin at the base plus the pointed-to byte offset, so each allocation
+/// must also cover that leading displacement.
 unsafe fn sql_bind_col_impl(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
--- a/mssql-odbc/src/api/bind_col.rs
+++ b/mssql-odbc/src/api/bind_col.rs
@@ -30,7 +30,12 @@ use crate::handles::{DescHandle, HandleType, StmtHandle, handle_from_raw};
 ///
 /// # Safety
 /// `statement_handle` must be a valid `StmtHandle` or null. The buffers must
-/// stay valid until the column is unbound or the statement is freed.
+/// stay valid until the column is unbound or the statement is freed. At each
+/// fetch, `target_value_ptr` must be writable for `SQL_ATTR_ROW_ARRAY_SIZE`
+/// elements of `buffer_length` bytes for a character or binary target, or of
+/// the full C type size for a fixed-width target, even when `buffer_length` is
+/// zero or smaller. `strlen_or_ind_ptr`, when non-null, must be writable for
+/// `SQL_ATTR_ROW_ARRAY_SIZE` `SqlLen` values.
 pub(crate) unsafe fn sql_bind_col(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
@@ -57,8 +62,13 @@ pub(crate) unsafe fn sql_bind_col(
 
 /// # Safety
 /// `statement_handle` must be null or point to a live `StmtHandle`. When
-/// non-null, `target_value_ptr` and `strlen_or_ind_ptr` must remain valid for
-/// the bound rowset until the column is unbound or the statement is freed.
+/// non-null, `target_value_ptr` must remain valid until the column is unbound or
+/// the statement is freed. At each fetch, it must be writable for
+/// `SQL_ATTR_ROW_ARRAY_SIZE` elements of `buffer_length` bytes for a character
+/// or binary target, or of the full C type size for a fixed-width target, even
+/// when `buffer_length` is zero or smaller. `strlen_or_ind_ptr`, when non-null,
+/// must remain valid and be writable for `SQL_ATTR_ROW_ARRAY_SIZE` `SqlLen`
+/// values.
 unsafe fn sql_bind_col_impl(
     statement_handle: SqlHandle,
     column_number: SqlUSmallInt,
--- a/mssql-odbc/src/api/bind_param.rs
+++ b/mssql-odbc/src/api/bind_param.rs
@@ -69,6 +69,10 @@ pub(crate) unsafe fn sql_bind_parameter(
     })
 }
 
+/// # Safety
+/// `statement_handle` must point to a live `StmtHandle`.
+/// `parameter_value_ptr` and `strlen_or_ind_ptr`, when non-null, must remain
+/// valid until the parameter is unbound or the statement is freed.
 #[allow(clippy::too_many_arguments)]
 unsafe fn sql_bind_parameter_impl(
     statement_handle: SqlHandle,
@@ -376,6 +380,8 @@ pub(crate) unsafe fn sql_free_stmt_reset_params(statement_handle: SqlHandle) -> 
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`.
 unsafe fn sql_free_stmt_reset_params_impl(statement_handle: SqlHandle) -> SqlReturn {
     if statement_handle.is_null() {
         error!("SQLFreeStmt(SQL_RESET_PARAMS): statement_handle is null");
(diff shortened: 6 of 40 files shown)


clippy.toml
```diff
@@ -0,0 +1 @@
+check-private-items = true
```
saurabh500, line 1: Why is this not in workspace Cargo.toml. Consider moving it there, so that all clippy rules are in the same place.

saurabh500: `check-private-items` is a Clippy configuration option, not a lint level, so Cargo's `[workspace.lints.clippy]` cannot represent it. Putting it there makes Cargo pass `clippy::check_private_items` as a lint name, which Clippy rejects as unknown; a root `clippy.toml` is the supported workspace-wide configuration path.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
```diff
@@ -40,6 +40,12 @@ use mssql_tds::query::metadata::PlpEncoding;
 /// - Supports `SQL_C_CHAR` and `SQL_C_WCHAR` for text retrieval.
 /// - Supports incremental row resume and chunked PLP retrieval via
 ///   `read_active_plp_chunk`.
+///
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`.
+/// `target_value_ptr`, when non-null, must be writable for `buffer_length`
+/// bytes, including when `target_type` is `SQL_C_WCHAR`. `strlen_or_ind_ptr`,
+/// when non-null, must be writable for one `SqlLen`.
```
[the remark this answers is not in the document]
saurabh500: Updated the exported, crate-visible, and private `SQLGetData` contracts to require the full C type size for fixed-width targets even when `BufferLength` is zero or smaller. The variable-width byte-count rule, including `SQL_C_WCHAR`, remains explicit.

> [!NOTE]

David-Engel (commented):

## Summary

This closes a real enforcement gap rather than just adding prose: `missing_safety_doc` only covered exported items, so every `pub(crate)`/private `unsafe fn` in the ODBC FFI surface was documented by convention alone. Enabling `check-private-items` makes that a build failure, and the one behavior change — `read_unaligned` in `nts_byte_count` — is the correct fix for ODBC's unaligned-buffer contract. I found nothing blocking.

## Verification

Run on `f302c94` in a dedicated worktree, diffed against `origin/main` (39 files, +358/-26).

- **The new config is load-bearing, not decorative.** `mssql-odbc` is a workspace member with its own `[lints] workspace = true`, so it was worth confirming the root `clippy.toml` actually reaches it. I removed the `# Safety` block you added above `sql_cancel_impl` and re-ran clippy: `warning: unsafe function's docs are missing a # Safety section` on `mssql-odbc/src/api/cancel.rs`. Restored afterwards. The lint fires on private items exactly as intended.
- `cargo clippy --all-features --all-targets` clean for `mssql-tds` (Windows), which confirms the six `unsafe fn` → `fn` conversions in `cng.rs` left no `unused_unsafe` behind. The enclosing `unsafe` blocks in `from_pem` and `to_pkcs8_pem` are still justified by the `decode_object`/`encode_object` calls that remain unsafe, so those blocks stay warranted.
- `cargo clippy --all-targets` clean for `mssql-odbc`.
- **The `SQLGetData` contract correction is accurate.** `write_captured_column` calls `convert_typed_c(value, target_type, target_value_ptr, strlen_or_ind_ptr)` — `buffer_length` is not passed at all, so a fixed-width target genuinely writes the full C-type width regardless. Your revision matches `convert_typed_c`'s own pre-existing `# Safety` note, and it resolves the bot's suppressed comment correctly.
- **Documented pointer types match the implementations**, which is where this class of doc tends to drift: `param_bind_offset` reads `*const SqlLen` (docs say `SqlLen`), `read_bind_offset` reads `*mut SqlULen` (docs say `SqlULen`), and `sql_get_env_attr_safe` writes `*mut u32` (docs say one `u32`). I spot-checked these three because the signed/unsigned split between the two bind-offset attributes is easy to get backwards.
- **Dropping the alignment clause from `sql_put_data`'s contract is correct on both paths.** The `SQL_NTS` path now uses `read_unaligned`, and the positive-length path reads through `std::slice::from_raw_parts(data_ptr as *const u8, byte_count)`, which carries no alignment obligation. The `u8` branch of `nts_byte_count` is trivially aligned.
- `mssql-py-core` is outside the workspace but still inherits the root `clippy.toml`; it contains no `unsafe fn` at all, so the new config is a no-op there.
- CI: green, with the ADO coverage rollup still in flight. Diff coverage 100% (9 lines), overall 93.4%.

## Suggestions

**`mssql-odbc/src/api/put_data.rs:86` — the alignment fix has coverage but no guard.**

Diff coverage reports 100% for this file, but that only means the existing NTS tests execute the line; every one of them passes a `[u16; 3]` that is already `u16`-aligned, so they pass identically with `*ptr.add(units)` restored. The change is right, and nothing pins it against a future revert.

There is a precedent for exactly this test in the same crate — `param_bind_offset_accepts_a_misaligned_application_pointer` in `set_stmt_attr.rs:1548` — which builds a deliberately misaligned pointer and asserts the misalignment before reading. The same shape works here:

```rust
#[test]
fn nts_byte_count_reads_a_misaligned_wide_buffer() {
    let mut storage = [0u8; 8];
    let ptr = unsafe { storage.as_mut_ptr().add(1) };
    assert_ne!(ptr as usize % std::mem::align_of::<u16>(), 0, "test pointer must be misaligned");
    unsafe { ptr.cast::<u16>().write_unaligned(0x0041) };
    unsafe { ptr.add(2).cast::<u16>().write_unaligned(0x0000) };
    assert_eq!(unsafe { nts_byte_count(ptr.cast(), SQL_C_WCHAR) }, 2);
}
```

This won't trap misalignment UB on x86 by itself, but it documents the contract, keeps the case reachable under Miri or a sanitizer run, and makes a revert of `read_unaligned` visible as an intent change rather than a silent one.

## Nits

**`mssql-odbc/src/api/get_functions.rs:57` — name the constant like the exported doc does.** The private contract says "or for the ODBC 3 function bitmap when requesting `SQL_API_ODBC3_ALL_FUNCTIONS`", while the exported doc above it says `SQL_API_ODBC3_ALL_FUNCTIONS_SIZE` words. Since the whole point of this PR is that private contracts should be checkable on their own terms, naming `SQL_API_ODBC3_ALL_FUNCTIONS_SIZE` here costs nothing and matches the `SQL_API_ALL_FUNCTIONS_SIZE` phrasing already in the same sentence.

I did not re-file the `clippy.toml` placement question — you already answered it, and the answer is right: `check-private-items` is a configuration key, not a lint level, so `[workspace.lints.clippy]` cannot express it.

David-Engel (commented):

## Summary

Re-review at `a7cb22d`. The two commits since `f302c94` are exactly the follow-ups from the prior pass: a misalignment regression test for `nts_byte_count`, and the `SQL_API_ODBC3_ALL_FUNCTIONS_SIZE` naming fix in the `sql_get_functions_impl` contract. Nothing else in the 39-file diff moved. Still nothing blocking.

## Verification

Run in a dedicated worktree against `origin/main` (merge base `d7ba300`, +375/-26).

- `cargo nextest run -p mssqlodbc --lib -E 'test(nts_byte_count)'` — 2 passed, including the new `nts_byte_count_reads_a_misaligned_wide_buffer`.
- `cargo clippy -p mssqlodbc -p mssql-tds --all-targets -- -D warnings` — clean, so the new test's `unsafe` blocks don't trip the tightened config.
- The new test now pins `read_unaligned` against a silent revert, which was the one gap in the previous pass. The write offsets stay in bounds: `ptr` is `storage[1]`, the two `u16` writes cover bytes 1..3 and 3..5 of an 8-byte array, and the expected count of 2 matches one non-terminator unit.
- `sql_get_data`'s fixed-width clause is intact on both the exported and the private `_impl` contract, so the bot's suppressed comment stays resolved.
- Required CI: all checks green on this head, including the ADO validation run and the coverage rollup (diff coverage 100%).

## Nits

**`mssql-odbc/src/api/put_data.rs:451` — the misalignment isn't guaranteed by the type.** `[u8; 8]` has `align_of == 1`, so the language makes no promise about where the compiler places `storage`. If the backing slot ever lands on an odd address, `as_mut_ptr().add(1)` is `u16`-aligned and the `assert_ne!` guard fails rather than the read being exercised. Every real toolchain aligns an 8-byte stack slot to at least 2, so this won't fire in practice — but the guard exists precisely because you didn't want to rely on that.

The precedent you followed, `param_bind_offset_accepts_a_misaligned_application_pointer` (`set_stmt_attr.rs:1548`), sidesteps this by starting from `[SqlLen; 2]`, where the base alignment *is* guaranteed. The same trick works here:

```rust
let mut storage = [0u16; 4];
let ptr = unsafe { storage.as_mut_ptr().cast::<u8>().add(1) };
```

That makes `ptr` unconditionally odd, and the rest of the test is unchanged.

David-Engel (commented):

## Summary

Re-review at `0b0d083`. The only change since `a7cb22d` is the one I asked for last pass: `nts_byte_count_reads_a_misaligned_wide_buffer` now starts from `[0u16; 4]`, so `ptr` is unconditionally odd rather than relying on where the compiler happens to place a `[u8; 8]`. Nothing else in the 39-file diff moved. Nothing blocking.

## Verification

Run in a dedicated worktree against `origin/main` (merge base `d7ba300`, +375/-26).

- **The new test is not vacuous — and it is stronger than I gave it credit for last pass.** I reverted `ptr.add(units).read_unaligned()` back to `*ptr.add(units)` and re-ran `cargo nextest run -p mssqlodbc --lib -E 'test(nts_byte_count)'`. It does not silently pass on x86: the test aborts with `misaligned pointer dereference: address must be a multiple of 0x2 but is 0x...` from the `debug_assertions` alignment check, which nextest runs in the dev profile. So this pins `read_unaligned` against a revert on ordinary CI, not only under Miri. I retract the "won't trap misalignment UB on x86 by itself" caveat from my previous review. Mutation reverted afterwards; the worktree is clean.
- The new pointer is guaranteed misaligned: `[0u16; 4]` has `align_of == 2`, so `as_mut_ptr().cast::<u8>().add(1)` is always odd and the `assert_ne!` guard cannot mask a skipped read. The two `write_unaligned` calls cover bytes 1..3 and 3..5 of an 8-byte allocation, in bounds.
- `cargo clippy --workspace --frozen --all-features --all-targets -- -D warnings` — clean on this head, so the tightened `check-private-items` config is satisfied everywhere it applies and the `cng.rs` `unsafe fn` → `fn` conversions left no `unused_unsafe`.
- `mssql-py-core` is outside the workspace but inherits the root `clippy.toml`; it still contains no `unsafe fn` at all, so `./scripts/bclippy.*` is unaffected.
- Non-Windows `unsafe fn`s are covered too, not just the Windows ones I can compile here: `security/unix/gssapi_ffi.rs` already carried `# Safety` on every export, and the `#[cfg(not(windows))] sql_set_connect_attr` / `_impl` pair got contracts in this PR.
- The `get_functions.rs` naming nit from last pass is addressed — the private contract now says `SQL_API_ODBC3_ALL_FUNCTIONS_SIZE`.
- Required CI: `gh pr checks 485` is green across the board on this head, including the ADO validation run and the coverage rollup (diff coverage 100%, 20 lines).

## Suggestions

**`mssql-odbc/src/api/exports.rs:623` and `:675`, `fetch.rs:25`, `fetch_scroll.rs`, `bind_col.rs:60` — the bound-column contracts have the same fixed-width hole you just closed for `SQLGetData`.**

The new rowset contracts say each bound buffer "must be writable for the configured rowset according to its C type and `BufferLength`". For a fixed-width target the `and BufferLength` half is not just imprecise, it is inert: `element_stride` (`fetch_scroll.rs:1241`) ignores `buffer_length` entirely for every typed C target and returns the C type's own size, falling back to `buffer_length` only for character and binary. Its own test spells this out — `element_stride_falls_back_to_the_c_type_size` asserts `element_stride(SQL_C_SLONG, 0) == 4`. So a caller who binds `SQL_C_SLONG` with `BufferLength = 0` and sizes the allocation from the documented extent is still written through for `SQL_ATTR_ROW_ARRAY_SIZE * 4` bytes.

This is exactly the gap the bot raised on `get_data.rs`, and you fixed it there with wording worth reusing:

```text
/// - Every active bound-column data buffer must be writable for the configured
///   rowset: `SQL_ATTR_ROW_ARRAY_SIZE` elements of `BufferLength` bytes for a
///   character or binary target, or of the full C type size for a fixed-width
///   target, even when `BufferLength` is zero or smaller.
```

Doc-only, no behavior change, and it keeps the two retrieval paths saying the same thing about the same conversion layer.

## Nits

None.

David-Engel (commented):

## Summary

Re-review at `38347e2`. The only change since `0b0d083` is the rowset-buffer wording I suggested last pass: the bound-column contracts on `SQLBindCol`, `SQLFetch`, `SQLFetchScroll`, `sql_bind_col`/`_impl`, `sql_fetch`, and `sql_fetch_scroll`/`_impl` now spell out the fixed-width case instead of deferring to "according to its C type and `BufferLength`". That closes the hole. Nothing else in the 39-file diff moved. Nothing blocking.

## Verification

Run in a dedicated worktree against `origin/main` (merge base `d7ba300`, +403/-29).

- **The new wording matches the implementation.** `element_stride` (`fetch_scroll.rs:1245`) returns the C type size for every typed target and falls back to `buffer_length.max(0)` only for character and binary, and `element_stride_falls_back_to_the_c_type_size` / `fixed_width_stride_ignores_the_buffer_length` pin both halves. So "`SQL_ATTR_ROW_ARRAY_SIZE` elements of `BufferLength` bytes for a character or binary target, or of the full C type size for a fixed-width target, even when `BufferLength` is zero or smaller" is now literally what the write path does.
- **The per-element wording is only sound because row-wise binding is rejected**, which I checked rather than assumed: `sql_fetch_scroll_impl` returns `HYC00` when `row_bind_type != SQL_BIND_BY_COLUMN` (`fetch_scroll.rs:683`), so a contiguous column-wise array is the only shape a caller can reach. If row-wise binding ever lands, all seven of these contracts need revisiting together.
- `cargo clippy -p mssqlodbc --all-targets -- -D warnings` — clean, so the tightened `check-private-items` config is still satisfied.
- `cargo nextest run -p mssqlodbc --lib -E 'test(nts_byte_count) + test(bind_offset) + test(element_stride) + test(fixed_width_stride)'` — 14 passed, including `nts_byte_count_reads_a_misaligned_wide_buffer` and `the_bind_offset_displaces_both_bases`.
- Required CI: green on this head across GitHub checks and the ADO validation stages; only the coverage rollup is still in flight.

## Suggestions

**`mssql-odbc/src/api/bind_col.rs:33` and `:65`, `exports.rs:648`, `:682`, `:623`, `fetch.rs:25`, `fetch_scroll.rs:80` and `:105` — the extent still omits the bind-offset displacement.**

Each contract now says the data buffer must be writable for `SQL_ATTR_ROW_ARRAY_SIZE` elements and the indicator/octet-length arrays for `SQL_ATTR_ROW_ARRAY_SIZE` `SqlLen` values, measured from the bound base. But the write does not start at the base. `deliver_bound` computes the slot as `(target_value_ptr as *mut u8).add(bind_offset + row_index * stride)` (`fetch_scroll.rs:1771`), and `displaced_len_ptr` (`:1351`) shifts the indicator and octet-length bases by the same `bind_offset` before indexing by row. `the_bind_offset_displaces_both_bases` asserts exactly that, and `bind_offset` comes straight from `read_bind_offset(row_bind_offset_ptr)` — an application-controlled `SqlULen`, read fresh on every fetch (`:963`).

So a caller who reads these contracts literally, sets `SQL_ATTR_ROW_BIND_OFFSET_PTR` to a non-zero value, and sizes the allocation for `SQL_ATTR_ROW_ARRAY_SIZE` elements gets an out-of-bounds write. The required extent is `*SQL_ATTR_ROW_BIND_OFFSET_PTR + SQL_ATTR_ROW_ARRAY_SIZE * stride` bytes from the base.

The `SQLFetch`/`SQLFetchScroll` docs do mention the attribute, but only to say the pointer itself must be readable for one `SqlULen` — that constrains the pointer, not the buffers it displaces. And `SQLBindCol`/`sql_bind_col` don't mention it at all, even though that is where the buffer the offset displaces is registered. One clause on the buffer sentence covers it:

```text
/// ... even when `BufferLength` is zero or smaller. When
/// `SQL_ATTR_ROW_BIND_OFFSET_PTR` is non-null, every extent above is measured
/// from the base displaced by the pointed-to byte count, so each buffer must
/// additionally be writable for that many leading bytes.
```

Doc-only, no behavior change. I'm raising it because it is the same class of gap as the two you already fixed — a contract that is satisfiable while the call still writes out of bounds — and this is the last one I can find on the bound-column path.

## Nits

None.

David-Engel (commented):

Posted from an unattended run.

## Summary

Re-review at `f4645ff`. The single new commit since `38347e2` is exactly the bind-offset wording I asked for last pass, and it landed in all eight places I listed: `sql_bind_col`/`_impl`, `sql_fetch`, `sql_fetch_scroll`/`_impl`, and the `SQLBindCol` / `SQLFetch` / `SQLFetchScroll` exports each now say the bound-buffer extents "begin at the base plus the pointed-to byte offset, so each allocation must also cover that leading displacement." That closes the column-side gap.

One suggestion remains, and it is the mirror image of the one just fixed: the **parameter** side has the same displacement, and the parameter contracts this PR introduced still describe their extents from the undisplaced base. Doc-only, non-blocking.

## Verification

Dedicated worktree against `origin/main`, merge base `d7ba300`, 39 files, +426/-29.

- **The misalignment test is not vacuous — I proved it.** I reverted `nts_byte_count` to `unsafe { *ptr.add(units) }` and re-ran `nts_byte_count_reads_a_misaligned_wide_buffer`: it aborts with `misaligned pointer dereference: address must be a multiple of 0x2 but is 0xd0efbff311`. Rust's debug-assertion alignment check catches this in the test profile, so the test genuinely guards the `read_unaligned` fix rather than passing incidentally on x86. Restored afterwards.
- **The `unsafe fn` removals in `cng.rs` are Windows-only code, so I linted them on Windows** rather than trusting the Linux leg. `cargo clippy --workspace --all-features --all-targets -- -D warnings` and `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` are both clean, so no `unused_unsafe` was left at any of the eight rewritten call sites (`sha256` ×2, `aes_256_cbc` ×2, `import_capi`, `export_capi`, `pem_to_der`, `der_to_pem`), and `check-private-items` is satisfied.
- **`mssql-py-core` is excluded from the workspace but still sits under the new root `clippy.toml`**, which is the easy way for a repo-wide Clippy config to break something CI-adjacent. Linted it separately: clean.
- **`SQL_ATTR_CURRENT_CATALOG` really is the only character attribute** on the `SQLGetConnectAttrW` path (every other arm returns a `u32`), so the new "`buffer_length` bytes for `SQL_ATTR_CURRENT_CATALOG`, one fixed-width value otherwise" contract matches the code.
- **The character-vs-byte asymmetry in the descriptor docs is correct**, which is the kind of thing that is usually wrong: `SQLGetDescRecW`'s `name` is documented in UTF-16 code units and `SQLGetDescFieldW`'s `value_ptr` in bytes, matching the ODBC spec's differing `BufferLength` units for those two calls.
- `cargo nextest run -p mssqlodbc --lib -E 'test(nts_byte_count) + test(bind_offset) + test(put_data)'` — 27 passed.
- `gh pr checks 485` — green on this head, including all ADO validation stages and the cross-repo `mssql-python` suites.

I also confirmed the driver is internally consistent on unaligned wide reads: `read_wchar_bytes` (`param_buffer.rs:244`) already read unit-by-unit with `read_unaligned` and says why, so `put_data` was the lone outlier this PR corrects — not a one-off patch over a broader pattern.

## Suggestions

**`mssql-odbc/src/api/execute.rs:37` and `:48`, `exec_direct.rs:41` and `:66`, `exports.rs:535` and `:567` (plus `bind_param.rs:72`) — the parameter contracts omit the bind-offset displacement, the same gap you just closed on the column side.**

The new wording is:

> For each non-data-at-execution parameter, the currently bound value, indicator, and octet-length buffers must remain readable according to the bound C type and lengths. `SQL_ATTR_PARAM_BIND_OFFSET_PTR`, when set, must remain readable for one `SqlLen`.

The second sentence constrains the *pointer*; it does not constrain the buffers the pointer displaces. But the read does not start at the bound base. `exec_common.rs:548` reads `stmt_state.inert_attrs.param_bind_offset()`, and `BoundParam::with_bind_offset` (`bound_param.rs:69`) then shifts all three pointers by that byte count:

```rust
self.parameter_value_ptr = self.parameter_value_ptr.wrapping_byte_offset(offset);
self.strlen_or_ind_ptr   = self.strlen_or_ind_ptr.wrapping_byte_offset(offset);
self.octet_length_ptr    = self.octet_length_ptr.wrapping_byte_offset(offset);
```

`param_bind_offset()` returns `isize` read fresh from the application's `SQLLEN` at execute time (`stmt.rs:519`), and `param_bind_offset_is_read_through_the_stored_pointer` asserts it round-trips a **negative** value (`-8`), so the displacement is signed in both directions. A caller who reads these contracts literally, sets `SQL_ATTR_PARAM_BIND_OFFSET_PTR`, and sizes the allocation for the bound C type and length alone gets an out-of-bounds read.

Same one-clause fix you used on the column side:

```text
/// ... according to the bound C type and lengths. When
/// `SQL_ATTR_PARAM_BIND_OFFSET_PTR` is non-null, these extents begin at the
/// base plus the pointed-to byte offset, which may be negative, so each
/// allocation must also cover that displacement.
```

`SQLBindParameter` / `sql_bind_parameter` are worth including for the same reason `SQLBindCol` was: that is where the buffer the offset displaces is registered, and its doc currently only promises the buffers "remain valid until the statement is executed."

For the record, `SqlLen` (not `SqlULen`) is the right type in the existing sentence — it matches `stmt.rs:522`'s `*const SqlLen` and the signed `isize` return. The row-side counterpart correctly says `SqlULen`. That asymmetry is real, not a typo.

## Nits

None.

mssql-odbc/src/api/put_data.rs
```diff
@@ -29,11 +29,11 @@ use crate::handles::{HandleType, StmtHandle, handle_from_raw};
 /// - `data_ptr`, when `strlen_or_ind` is a positive byte count, must be
 ///   readable for that many bytes and must remain valid for the duration of
 ///   this call.
-/// - `data_ptr`, when `strlen_or_ind` is `SQL_NTS`, must be non-null, aligned
-///   for the bound parameter's C type, and NUL-terminated within an allocation
-///   it owns: the terminator search reads `u16` units for `SQL_C_WCHAR` and
-///   `u8` units otherwise, and runs off the end of the allocation if no
-///   terminator is present.
+/// - `data_ptr`, when `strlen_or_ind` is `SQL_NTS`, must be non-null and
```
David-Engel, line 32: The `SQL_NTS` obligation stops at this layer. The exported entry point in `exports.rs` (the `SQLPutData` `# Safety` block, ~lines 608-611) still documents only the positive-length case: "`data_ptr`, when `strlen_or_ind` is a positive byte count, must be readable for that many bytes." Nothing about `SQL_NTS` — which is the one shape where the driver performs an unbounded scan (`SQLPutData` -> `sql_put_data` -> `sql_put_data_safe:249` -> `nts_byte_count`).

That is the same class of gap the bot caught on `SQLGetData`: an application author can satisfy the exported contract completely and still hand us a buffer we read off the end of. Since this PR is specifically about making these obligations explicit, and it already carried the bind-offset clauses up into `exports.rs` for `SQLBindParameter` / `SQLExecute` / `SQLExecDirectW`, this one looks like the odd one out. Suggest adding the same sentence there:

```rust
/// - `data_ptr`, when `strlen_or_ind` is `SQL_NTS`, must be non-null and
///   NUL-terminated within an allocation it owns: the terminator search reads
///   potentially unaligned `u16` units for `SQL_C_WCHAR` and `u8` units
///   otherwise, and runs off the end of the allocation if no terminator is
///   present.
```

mssql-odbc/src/api/catalog.rs
```diff
@@ -691,6 +691,10 @@ pub(crate) unsafe fn sql_tables_w(
     })
 }
 
+/// # Safety
+/// `statement_handle` must be null or point to a live `StmtHandle`. Each
+/// non-null name pointer must be readable for its paired length in UTF-16 code
+/// units, or through a NUL terminator when that length is `SQL_NTS`.
```
David-Engel, line 700: This `SQL_NTS` clause is new and correct — `opt_arg` scans to the terminator when `len` is `SQL_NTS` — but it only landed on the private `_impl`. The two layers above still carry the older wording: `sql_tables_w` at `catalog.rs:654` and the exported `SQLTablesW` at `exports.rs:867` both say only "Each name pointer must be null or reference `*_len` readable UTF-16 units", which describes nothing when `*_len` is `SQL_NTS` (-3). The same holds for the other six catalog pairs you documented in this file (`sql_columns_w`, `sql_primary_keys_w`, `sql_foreign_keys_w`, `sql_statistics_w`, `sql_special_columns_w`, `sql_procedures_w`).

`check-private-items` will not flag this, since it only requires a `# Safety` section to exist, so the private layer is now strictly more accurate than the public one. Worth propagating the "or through a NUL terminator when that length is `SQL_NTS`" half up to the `pub(crate)` wrappers and the exports so the FFI-facing contract is the complete one.

David-Engel (commented):

## Summary

The two newest commits add the `SQL_ATTR_PARAM_BIND_OFFSET_PTR` and `SQL_ATTR_ROW_BIND_OFFSET_PTR` extent clauses. I checked the wording against the code rather than the spec, because the two attributes genuinely differ: `InertStmtAttrs::param_bind_offset` returns `isize` (and `bound_param.rs` has a `bind_offset_may_be_negative` test), so "signed byte offset, which may be negative" is correct on the parameter side; `read_bind_offset` returns `usize` from a `*mut SqlULen`, so the unsigned row-side wording is correct too. The rowset extents also match `element_stride`, which ignores `BufferLength` for a fixed-width target, and row-wise binding is refused outright at `fetch_scroll.rs:689`, so "`SQL_ATTR_ROW_ARRAY_SIZE` elements of `BufferLength` bytes ... or of the full C type size" describes what the code actually writes.

What is left is a layering gap rather than an error. Several obligations were added only to the private `_impl` function, while the `pub(crate)` wrapper and the exported `extern "C"` entry point above it keep the older, weaker sentence. `check-private-items` cannot catch that — it flags a *missing* `# Safety` section, never an incomplete one — so these do not resurface on their own. Two concrete cases inline; both are doc-only.

## Verification

- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean on Windows, which is the leg that actually compiles the `#[cfg(windows)]` CNG code whose `unsafe fn` markers you removed.
- `cargo nextest run -p mssqlodbc -E 'test(nts_byte_count) or test(bind_offset)'` — 12 passed, including the new `nts_byte_count_reads_a_misaligned_wide_buffer`.
- Read the doc-only hunks against the code they describe: `element_stride`, `displaced_len_ptr`, `read_bind_offset`, `clear_stale_null_indicator` (`fetch_scroll.rs`), `param_bind_offset` (`stmt.rs`), `opt_arg` (`catalog.rs`), and the `SQL_NTS` path in `put_data.rs`.
- Also confirmed the `SQL_NTS` path already rejects a null `data_ptr` with `ERR_INVALID_NULL_POINTER` before `nts_byte_count`, so the new "must be non-null" clause is a contract statement and not a live null deref.

## Blocking

None.

## Suggestions

Two, both inline and both doc-only:

1. `mssql-odbc/src/api/put_data.rs` — the exported `SQLPutData` never mentions `SQL_NTS`, which is the one shape that makes the driver scan for a terminator.
2. `mssql-odbc/src/api/catalog.rs` — the new `SQL_NTS` clause on the seven `*_impl` functions is not reflected on the `pub(crate)` wrappers or the exported catalog entry points.

## Nits

None.

mssql-odbc/src/api/put_data.rs
saurabh500: Propagated the `SQL_NTS` allocation and terminator-scan obligation to the exported `SQLPutData` safety contract, including the unaligned `SQL_C_WCHAR` read behavior.

> [!NOTE]

mssql-odbc/src/api/catalog.rs
saurabh500: Updated all seven catalog `pub(crate)` wrappers and exported entry points to cover NUL-terminated input when the paired length is `SQL_NTS`, matching their private shims.

> [!NOTE]

Theekshna (commented):

## Summary
This PR enables Clippy's `check-private-items` (root `clippy.toml`), documents the ODBC safety contracts that lint now requires across ~38 `mssql-odbc/src/api/*.rs` files, drops `unsafe fn` from 6 `mssql-tds` CNG helpers that no longer need it, and switches `nts_byte_count`'s `SQL_C_WCHAR` terminator scan from a plain dereference to `read_unaligned()`.

This PR already went through 8 rounds of thorough automated review (`David-Engel`, itself another unattended-run identity) that mutation-tested the alignment fix, traced the fixed-width/bind-offset doc gaps against the actual write paths, and verified CNG/clippy correctness on Windows. The only two open suggestions from that chain — `SQLPutData`'s exported contract not mentioning `SQL_NTS`, and the catalog `SQL_NTS` wording not propagated from the private `_impl`s to their `pub(crate)`/exported callers — are exactly and only what the final commit (`8ed8f8b7`) changes, in `catalog.rs` and `exports.rs`. I diffed `a29b4c9..8ed8f8b7` directly and confirmed both are resolved correctly with no side effects.

## The 5 checks

| Check | Result |
|---|---|
| msodbcsql parity | Read `Sql/Ntdbms/sqlncli/odbc/sqlccmd.cpp` (`C:\REPOS\Drivers\ODBC_2`, read-only). msodbcsql's own `SQL_C_WCHAR` NTS length calc casts the caller's raw pointer straight to `wchar_t*` and calls `wcslen`, with no unaligned-read defense of its own. This PR's `read_unaligned()` is strictly more robust than the reference driver's equivalent, not a gap or a behavioral divergence — output is identical for valid input, only the internal read mechanism changed. |
| Test sufficiency | The mutation check this needs was already done and shown, not asserted: the prior review reverted `read_unaligned()` to a plain deref and quoted the actual `cargo nextest` failure (`misaligned pointer dereference: address must be a multiple of 0x2...`) — Rust's debug-assertion alignment check catches it in the test profile, so `nts_byte_count_reads_a_misaligned_wide_buffer` genuinely pins the fix. I tried to reproduce this myself and hit an unrelated local toolchain issue (this machine's Rust build fails to codegen `crc32fast`'s AVX-512/PCLMULQDQ intrinsics) before I could rebuild with the mutation, so I'm relying on the quoted panic from the prior round rather than re-deriving it. Flagging as a caveat on my own verification, not a gap in the PR. |
| Divergences documented | None needed — see parity row; same externally observable behavior as msodbcsql, just a more defensive implementation. |
| PR description currency | Matches the diff. Closes #474. Checklist is honest about the one unchecked box (`cargo btest` — full e2e needs a live SQL Server; the local `mssqlodbc` suite is reported passing). |
| Verbose slop | None. Spot-checked the doc-only files (`bind_col.rs`, `fetch_scroll.rs`, `catalog.rs`, `exports.rs`, `cng.rs`) against their implementations — every added `# Safety` clause states a specific, verified obligation (e.g. `SQL_ATTR_ROW_BIND_OFFSET_PTR`/`SQL_ATTR_PARAM_BIND_OFFSET_PTR` displacement, signed vs. unsigned split between the two), not generic boilerplate. |

CI: all GitHub Actions checks green; ADO validation mostly green with a few build legs (Linux, Linux ARM, macOS test, mssql-python macOS) still in progress at review time — no failures observed, not re-triggered.

| Severity | Count |
| --- | ---: |
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

No findings.

David-Engel (commented):

Posted from an unattended run.

## Summary

Re-review at `8ed8f8b7`. The single commit since `a29b4c90` is exactly the propagation I asked for last pass: the `SQL_NTS` clause is now on `SQLPutData`'s exported contract, and the seven catalog `pub(crate)` entry points plus their eight exported `*W` wrappers carry the same "null, paired length, or readable through a NUL terminator when that length is `SQL_NTS`" wording the private `*_impl`s already had. I traced all three layers down to `util.rs:105` `read_utf16_long`, which is the function that actually scans for the terminator, and the wording matches what it does.

Nothing else in the diff moved, so this pass re-verified the two parts that are not doc-only and read the contracts I had not yet checked against their implementations.

## Verification

- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — clean on `x86_64-pc-windows-msvc`, so the six de-`unsafe`d CNG helpers leave no `unused_unsafe` behind. Every remaining call site (`sha256` at `cng.rs:574` and `:616`, `pem_to_der:344`, `import_capi:362`, `export_capi:415`, `der_to_pem:441`) still sits inside an enclosing `unsafe` block that is needed for other operations, which is why none of them show up in the diff.
- `cargo nextest run -p mssqlodbc --lib -E 'test(put_data) or test(nts_byte_count)'` — 17 passed.
- Differential probe on the one behavior change. I reverted `nts_byte_count`'s `ptr.add(units).read_unaligned()` back to `*ptr.add(units)` in the worktree and re-ran: `nts_byte_count_reads_a_misaligned_wide_buffer` aborts with `misaligned pointer dereference: address must be a multiple of 0x2 but is 0x9cff4fee51`. Rust's debug-profile alignment assertion catches it, so that test is a real regression guard rather than intent documentation — worth recording explicitly, since the repo has no Miri or sanitizer job to fall back on. Reverted immediately; worktree is clean.
- Checked that the unaligned fix closes the class rather than one instance. Every other application-buffer read on the parameter and rowset paths already uses `read_unaligned`: `conversion/param_buffer.rs:78,87,161-169,252,263`, `exec_common.rs:575`, `handles/stmt.rs:526`, `fetch_scroll.rs:1279,3157,3164`. `put_data.rs` was the last plain dereference. The `SQLWCHAR*`-typed catalog and connect paths in `util.rs` legitimately keep aligned reads, since alignment there is implied by the declared C parameter type rather than left to an untyped `SQLPOINTER`.
- Confirmed `check-private-items` cannot drag in unrelated noise. It also widens `missing_errors_doc` and `missing_panics_doc`, but both are `pedantic` and this workspace's `[workspace.lints.clippy]` leaves them off, so `missing_safety_doc` is the only lint whose scope actually changes.
- Spot-checked the contracts I had not previously traced. `SQL_ATTR_CURRENT_CATALOG` really is the only character-valued connection attribute in `get_connect_attr.rs` (every other arm writes `*mut u32`), and `set_current_catalog` does treat `string_length` as a byte count, so both of those new clauses are accurate.

CI: no failures. All GitHub checks green; two ADO legs still pending at review time.

## Blocking

None.

## Suggestions

None.

## Nits

**`mssql-odbc/src/api/set_connect_attr.rs`** — the new contract on `sql_set_connect_attr_w_impl` says the access-token struct opens with "a four-byte little-endian length", but `decode_access_token` reads it with `u32::from_ne_bytes` (`set_connect_attr.rs:484`), its own doc says "a 4-byte native-endian length `n` (an `unsigned int`)", and the inline comment adds "written by the caller in host byte order". Same field, same file, two different answers.

It makes no observable difference on any target this driver ships to, but in a change whose whole point is that the safety contracts are authoritative, it is the kind of discrepancy a later reader resolves in the wrong direction — by "correcting" `from_ne_bytes` to `from_le_bytes` so the code matches the contract. Suggest matching the wording that is already there:

```rust
/// `connection_handle` must be null or point to a live `DbcHandle`. For
/// `SQL_COPT_SS_ACCESS_TOKEN`, `value_ptr` must point to a four-byte
/// native-endian length followed by that many readable token bytes. For
/// `SQL_ATTR_CURRENT_CATALOG`, it must be readable for `string_length` bytes of
/// UTF-16, or through a NUL terminator when `string_length` is `SQL_NTS`.
```

Theekshna (approved):

David-Engel (commented):

Posted from an unattended run.

## Summary

Re-review at `abf5a08`, which merges `main` (`#444` DAE wideness transcoding, `#446` row-wise `SQLGetData`) into the branch. I diffed the PR's own contribution against the new merge base `4382290` and confirmed it is byte-for-byte the same set of changes reviewed at `6c083ae` — the merge only moved the base and shifted hunk contexts. The last outstanding finding from `8ed8f8b` is fixed: the access-token length prefix is now described as native-endian in all four places (`exports.rs` W and ANSI entry points, `sql_set_connect_attr_w_impl`, and the `make_token_struct` test helper, which switched to `to_ne_bytes`). That matches `decode_access_token`'s `u32::from_ne_bytes` at `set_connect_attr.rs:484` and is the right reading of the msodbcsql `ACCESSTOKEN` struct, whose `dataSize` is a native `DWORD`. Keeping the token payload described as UTF-16**LE** while the length prefix is native-endian is the correct asymmetry, not a leftover.

I re-checked the two behavioral changes against the merged code rather than assuming the earlier verification still held, because both landed in files `main` touched:

The `read_unaligned` fix in `nts_byte_count` is still the only aligned read on the application pointer in the `SQL_NTS` path. `#444` added a second consumer — the buffered `needs_transcode` branch at `put_data.rs:383` — and it builds its slice as `*const u8`, same as the direct write at `put_data.rs:425`. So the relaxed "potentially unaligned" contract in the doc is honored end to end, and there is no residual place where a misaligned `SQLWCHAR*` would be reinterpreted as an aligned `u16`. The new `nts_byte_count_reads_a_misaligned_wide_buffer` test starts from `[0u16; 4]` and offsets by one byte, so the pointer is unconditionally odd and the `assert_ne!` guard cannot pass vacuously on some other target.

Removing `unsafe` from the six CNG helpers (`sha256`, `aes_256_cbc`, `import_capi`, `export_capi`, `pem_to_der`, `der_to_pem`) is correct: each one owns its raw pointers entirely and imposes nothing on callers, so the `unsafe fn` marker was only widening the audit surface. The four that stay `unsafe` — `open_alg`, `get_u32_property`, `decode_object`, `encode_object` — each take a caller-supplied `PCWSTR`/`PCSTR` or handle, which is a genuine obligation. Because these are `#[cfg(windows)]`, `unused_unsafe` at the call sites would only surface on a Windows compile; my clippy run below is that leg.

I also spot-verified the added contracts against the code they describe rather than reading them as prose. `sql_get_env_attr_impl`'s "writable for one `u32`" matches the sole `write_if_some(value_ptr as *mut u32, ..)` arm. `sql_get_functions_impl`'s three-way extent matches the `SQL_API_ALL_FUNCTIONS` / `SQL_API_ODBC3_ALL_FUNCTIONS` `copy_nonoverlapping` sizes, and now names `SQL_API_ODBC3_ALL_FUNCTIONS_SIZE` explicitly. The signed-vs-unsigned split between the two bind offsets is real and correctly reflected: `StmtState::row_bind_offset_ptr` is `*mut SqlULen` and the docs say `SqlULen`, while `InertStmtAttrs::param_bind_offset` reads `*const SqlLen` and returns `isize`, and the docs say signed and "may be negative". The ANSI `SQLSetConnectAttr` doc correctly omits "of UTF-16" for `SQL_ATTR_CURRENT_CATALOG` where the W variant includes it, matching the `CStr`/`from_raw_parts` byte handling in `sql_set_connect_attr_impl`.

One thing I checked and was wrong about, recorded so it does not get re-raised: `nts_byte_count` is not reachable with a null `data_ptr`. The `SQL_NTS` branch null-checks and posts `ERR_INVALID_NULL_POINTER` at `put_data.rs:223` before resolving the C type, independently of the later `data_ptr.is_null() && byte_count > 0` guard.

Putting `check-private-items` in a root `clippy.toml` is the only way to express it — it is a configuration key, not a lint level, so `[workspace.lints.clippy]` cannot carry it. Root placement also reaches the two workspace-excluded crates (`mssql-py-core`, `mssql-mock-tds-py`) through Cargo's ancestor lookup, and no competing `clippy.toml` exists lower in the tree to shadow it.

## Verification

- `cargo clippy -p mssqlodbc -p mssql-tds --all-features --all-targets -- -D warnings` on the merged head — clean. This is the Windows leg, so it is the one that actually compiles the `#[cfg(windows)]` CNG code whose `unsafe fn` markers were removed, and it confirms `check-private-items` is satisfied against `main`'s newly merged code and not just against the pre-merge branch.
- `cargo nextest run -p mssqlodbc --all-features` on the merged head — 1287 passed, 1 skipped, 0 failed.
- Read the full 40-file diff against the code it documents, and re-read `put_data.rs`, `get_data.rs`, `fetch_scroll.rs`, and `exec_common.rs` post-merge because `#444` and `#446` rewrote parts of those files.

## Blocking

None.

## Suggestions

None.

## Nits

None.

Nothing left on my side. This looks ready for human review, though this comment is not an approval and does not satisfy the human review requirement. Worth noting that PR validation for `abf5a08` was still pending when I posted, so the merge commit itself has not been green-lit by CI yet.

Theekshna (approved):

PR #487, opened 2026-09-03
Round legacy DATETIME fetches to milliseconds

saurabh500:
## Description

Legacy SQL Server `DATETIME` stores time in 1/300-second ticks, while ODBC exposes the fraction at millisecond precision. Round to the nearest millisecond before filling timestamp structs or rendering character values, matching `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp::ConvertDateTime` in msodbcsql. Unit coverage checks all 300 tick remainders, and E2E coverage verifies a rounding-sensitive `.127` value through `SQLGetData`, `SQLBindCol`, and `SQL_C_CHAR` rendering against both drivers.

The CI-equivalent local mssql-python run used the same containerized driver-swap and per-file pytest strategy as the pipeline. `test_parse_datetime` now passes. Failed/error outcomes fell from 290 in linked test run 3301980 to 276 locally; the upstream suite advanced from `3774fbb` to `08cd24c` and grew from 2,152 to 2,327 tests between those runs.

## Related Issues

[AB#47703](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47703)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

### DIFF
--- a/mssql-odbc/src/conversion/fetch_convert.rs
+++ b/mssql-odbc/src/conversion/fetch_convert.rs
@@ -466,14 +466,15 @@ pub(crate) fn extract_datetime_parts(value: &ColumnValues) -> Option<DateTimePar
             // `datetime` time is counted in 1/300-second ticks since midnight.
             let ticks = u64::from(dt.time);
             let secs = ticks / 300;
-            let fraction_ns = ((ticks % 300) * 1_000_000_000 / 300) as u32;
+            // ODBC exposes the legacy type rounded to millisecond precision.
+            let fraction_ms = ((dt.time % 300) * 1_000 + 150) / 300;
             p.year = y;
             p.month = m;
             p.day = day;
             p.hour = (secs / 3600) as u16;
             p.minute = ((secs % 3600) / 60) as u16;
             p.second = (secs % 60) as u16;
-            p.fraction_ns = fraction_ns;
+            p.fraction_ns = fraction_ms * 1_000_000;
             // `datetime` always renders 3 fractional digits.
             p.scale = 3;
             p.has_date = true;
@@ -2307,6 +2308,30 @@ mod tests {
         assert_eq!(out.month, 1);
         assert_eq!(out.day, 1);
         assert_eq!(out.second, 1);
+    }
+
+    #[test]
+    fn datetime_legacy_fraction_rounds_to_milliseconds() {
+        use mssql_tds::datatypes::column_values::SqlDateTime;
+
+        for (ticks, expected_fraction) in [(37, 123_000_000), (38, 127_000_000), (299, 997_000_000)]
+        {
+            let mut out = SqlTimestampStruct::default();
+            let mut ind: SqlLen = 0;
+            unsafe {
+                convert_datetime_c(
+                    &ColumnValues::DateTime(SqlDateTime {
+                        days: 0,
+                        time: ticks,
+                    }),
+                    SQL_C_TYPE_TIMESTAMP,
+                    (&mut out as *mut SqlTimestampStruct).cast(),
+                    &mut ind,
+                )
+            }
+            .unwrap();
+            assert_eq!(out.fraction, expected_fraction, "ticks: {ticks}");
+        }
     }
 
     #[test]
--- a/mssql-odbc/src/conversion/fetch_convert.rs
+++ b/mssql-odbc/src/conversion/fetch_convert.rs
@@ -2314,8 +2314,8 @@ mod tests {
     fn datetime_legacy_fraction_rounds_to_milliseconds() {
         use mssql_tds::datatypes::column_values::SqlDateTime;
 
-        for (ticks, expected_fraction) in [(37, 123_000_000), (38, 127_000_000), (299, 997_000_000)]
-        {
+        for ticks in 0u32..300 {
+            let expected_fraction = ((ticks * 20 + 3) / 6) * 1_000_000;
             let mut out = SqlTimestampStruct::default();
             let mut ind: SqlLen = 0;
             unsafe {
--- a/mssql-odbc/tests/e2e/tests/datetime_types_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/datetime_types_test.cpp
@@ -274,6 +274,58 @@ TEST_F(DateTimeTypesLiveTest, Datetime2ToTimestampTargetViaBoundFetch) {
     SQLCloseCursor(stmt_);
 }
 
+TEST_F(DateTimeTypesLiveTest, LegacyDatetimeRoundsFractionToMillisecondsOnBothPaths) {
+    const std::string query =
+        "SELECT CAST('2024-05-20T12:34:56.127' AS DATETIME) AS c1";
+
+    ASSERT_SQL_OK(ExecDirect(query), SQL_HANDLE_STMT, stmt_);
+    ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);
+
+    SQL_TIMESTAMP_STRUCT viaGetData{};
+    SQLLEN getInd = 0;
+    ASSERT_SQL_OK(
+        SQLGetData(stmt_, 1, SQL_C_TYPE_TIMESTAMP, &viaGetData, sizeof(viaGetData), &getInd),
+        SQL_HANDLE_STMT, stmt_);
+    EXPECT_EQ(2024, viaGetData.year);
+    EXPECT_EQ(5, viaGetData.month);
+    EXPECT_EQ(20, viaGetData.day);
+    EXPECT_EQ(12, viaGetData.hour);
+    EXPECT_EQ(34, viaGetData.minute);
+    EXPECT_EQ(56, viaGetData.second);
+    EXPECT_EQ(127000000u, viaGetData.fraction);
+    EXPECT_EQ(static_cast<SQLLEN>(sizeof(SQL_TIMESTAMP_STRUCT)), getInd);
+    SQLCloseCursor(stmt_);
+
+    ASSERT_SQL_OK(ExecDirect(query), SQL_HANDLE_STMT, stmt_);
+    ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);
+
+    constexpr char expectedText[] = "2024-05-20 12:34:56.127";
+    char text[sizeof(expectedText)]{};
+    SQLLEN textInd = 0;
+    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_CHAR, text, sizeof(text), &textInd), SQL_HANDLE_STMT,
+                  stmt_);
+    EXPECT_STREQ(expectedText, text);
+    EXPECT_EQ(static_cast<SQLLEN>(sizeof(expectedText) - 1), textInd);
+    SQLCloseCursor(stmt_);
+
+    ASSERT_SQL_OK(ExecDirect(query), SQL_HANDLE_STMT, stmt_);
+
+    SQL_TIMESTAMP_STRUCT bound{};
+    SQLLEN boundInd = 0;
+    ASSERT_SQL_OK(SQLBindCol(stmt_, 1, SQL_C_TYPE_TIMESTAMP, &bound, sizeof(bound), &boundInd),
+                  SQL_HANDLE_STMT, stmt_);
+    ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);
+    EXPECT_EQ(2024, bound.year);
+    EXPECT_EQ(5, bound.month);
+    EXPECT_EQ(20, bound.day);
+    EXPECT_EQ(12, bound.hour);
+    EXPECT_EQ(34, bound.minute);
+    EXPECT_EQ(56, bound.second);
+    EXPECT_EQ(127000000u, bound.fraction);
+    EXPECT_EQ(static_cast<SQLLEN>(sizeof(SQL_TIMESTAMP_STRUCT)), boundInd);
+    SQLCloseCursor(stmt_);
+}
+
 // A NULL date/time column must report SQL_NULL_DATA and leave the buffer alone,
 // on both paths. Covered once here rather than per type: the NULL path is in
 // the delivery layer, not per-conversion.


mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -2309,6 +2310,30 @@ mod tests {
         assert_eq!(out.second, 1);
     }
 
+    #[test]
+    fn datetime_legacy_fraction_rounds_to_milliseconds() {
```
saurabh500, line 2314: This could use a E2E ODBC test as well. PLease add one

David-Engel (commented):

No findings. The change is a one-line arithmetic correction in `extract_datetime_parts` plus a unit test, and it holds up under inspection.

The rounding is correct and total. `((dt.time % 300) * 1_000 + 150) / 300` is round-half-up on a `u32` where the intermediate maxes at `299 * 1_000 + 150 = 299_150` and the result maxes at `997`, so neither the multiply nor the later `fraction_ms * 1_000_000` (max `997_000_000`) can overflow `u32`, and the fraction can never reach `1_000` and roll into `second`. Ties are unreachable in practice: `r * 10 / 3` has a fractional part of only `.0`, `.333`, or `.667`, so half-up versus half-even never diverges. The endpoints line up with what SQL Server renders — `r = 1` → `.003`, `r = 2` → `.007`, `r = 299` → `.997` — which is the observable contract the linked work item is about.

Verified the test actually guards the change rather than restating it: dropping the `+ 150` back to truncation makes `datetime_legacy_fraction_rounds_to_milliseconds` fail on `ticks: 38` with `left: 126000000, right: 127000000`. The full `-p mssqlodbc --lib -E 'test(datetime)'` selection is 21 passed / 0 failed on the unmodified head, so the neighbouring legacy-`datetime` cases (`datetime_legacy_epoch_and_ticks`, the `SQL_C_TYPE_DATE` / `SQL_C_TYPE_TIME` truncation-warning cases) are unaffected. That last point is worth stating explicitly since it is the easiest place for a rounding change to leak: `fraction_ms` is zero exactly when `dt.time % 300 == 0`, the same condition as before, so no truncation-warning arm changes its verdict.

Blast radius is contained. `extract_datetime_parts` is the only site in `mssql-odbc` that converts 1/300-second ticks, and it feeds both `convert_datetime_c` and the text path in `api/get_data.rs` via `format_datetime_parts`, so `SQL_C_CHAR`/`SQL_C_WCHAR` fetches of a legacy `datetime` pick up the same fix — `.126` becomes `.127` there too. No other crate duplicates the tick math, and the inbound parameter path does not go through this function, so nothing is left half-converted.

Two notes rather than findings. The E2E coverage question is already open on your own inline thread, so I am deliberately not re-filing it. And the PR checklist still has `cargo bclippy` and `cargo btest` unchecked while the coverage report shows 100% diff coverage and no required check is failing — worth ticking before you mark this ready, since the boxes are what a human reviewer reads first.

mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -466,14 +466,15 @@ pub(crate) fn extract_datetime_parts(value: &ColumnValues) -> Option<DateTimePar
...
+            // ODBC exposes the legacy type rounded to millisecond precision.
+            let fraction_ms = ((dt.time % 300) * 1_000 + 150) / 300;
             p.year = y;
             p.month = m;
             p.day = day;
             p.hour = (secs / 3600) as u16;
             p.minute = ((secs % 3600) / 60) as u16;
             p.second = (secs % 60) as u16;
-            p.fraction_ns = fraction_ns;
+            p.fraction_ns = fraction_ms * 1_000_000;
```
saurabh500, line 477: The description says this lands "before filling timestamp structs", but it changes the character path too. `api/get_data.rs:1451-1458` routes `ColumnValues::DateTime` through `extract_datetime_parts` and then `format_datetime_parts`, which takes `fraction_ns / 100`, formats it to 7 digits and truncates to `scale`. At 38 ticks that used to give `"1266666"[..3]` = `.126`; it now gives `"1270000"[..3]` = `.127`.

That second change is also correct — msodbcsql feeds the same rounded `msec` to its `fChar` branch at `sqlccnvt.cpp:278` — so this is just a request to name it in the description, since it is a second user-visible behavior change and someone bisecting a rendered-string difference later will want to find it.

mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -2309,6 +2310,30 @@ mod tests {
         assert_eq!(out.second, 1);
     }
 
+    #[test]
+    fn datetime_legacy_fraction_rounds_to_milliseconds() {
+        use mssql_tds::datatypes::column_values::SqlDateTime;
+
+        for (ticks, expected_fraction) in [(37, 123_000_000), (38, 127_000_000), (299, 997_000_000)]
```
saurabh500: Optional, and a stronger form of what is already here: compare all 300 remainders against the reference expression rather than sampling three.

```rust
for r in 0u32..300 {
    // msodbcsql sqlccnvt.cpp:272
    let expected = (r * 20 + 3) / 6 * 1_000_000;
    // ...convert and assert out.fraction == expected, "ticks: {r}"
}
```

That pins parity with `ConvertDateTime` exactly instead of at three points, and it picks up the zero-fraction case — the one that should stay quiet — which none of `37`, `38`, `299` covers today.

The sampled table is already sufficient to catch the regression: I deleted `+ 150` from the conversion and this test failed at ticks=38 with `126000000` vs `127000000`. So this is a nicety, not a gap.

saurabh500 (commented):
## Summary

Reviewed at `45ced7d9`. The rounding is correct and the new test genuinely guards it. **No blocking findings.**

### What I verified

- **Exact parity with the reference driver.** `msodbcsql` `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp:272` computes `msec = ((ThreeHundredthsSec*20+3)/6)`, and line 288 writes `msec*1000000UL` into `SQL_TIMESTAMP_STRUCT::fraction`. This PR's `(t*1000 + 150)/300` is that same expression scaled by 50 — `1000t + 150 = 50(20t + 3)` and `300 = 50 * 6` — so with positive scaling and no overflow the two agree on every input, not just the sampled ones.
- **No overflow, no carry.** `ticks % 300 <= 299` bounds `fraction_ms` at 997, so `fraction_ms * 1_000_000 <= 997_000_000`, comfortably inside `u32`. Rounding can never reach 1000 ms, so there is no roll-into-the-next-second case that needs handling.
- **Truncation reporting is unchanged.** `fraction_ms == 0` iff `ticks % 300 == 0`, which is the same predicate as the old `fraction_ns == 0`, so the 01S07 arms at lines 708 and 727 behave exactly as before. `tests/e2e/tests/fetch_error_semantics_test.cpp:215-227` already exercises that path for a legacy `DATETIME` carrying a non-zero fraction.
- **The test is not vacuous.** I deleted `+ 150`, leaving plain truncation, and it failed at ticks=38 with `126000000` vs `127000000`. Restored afterwards.
- **This was the last path still truncating.** `mssql-js/lib/decode.ts:306` and `mssql-py-core/src/types.rs:1359` already round half-up, the latter to microseconds because DB-API exposes µs. `extract_datetime_parts` is the only tick conversion in `mssql-odbc`, and `mssql-tds` carries `(days, time)` raw through decode, serialize, bulk copy and Always Encrypted. There is also no `SQL_C_TYPE_TIMESTAMP` parameter path yet (`params/conversion_matrix.rs:55-60`), so no round-trip is affected.

Ran locally: `cargo nextest run -p mssqlodbc --lib -E 'test(datetime)'` gives 21/21 passing, and `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` is clean. Both are package-scoped, so they do not stand in for the workspace `cargo bclippy` / `cargo btest` aliases.

### Suggestion — checklist and readiness

`cargo bclippy` and `cargo btest` are unticked while the PR is out of draft, and ADO validation is still pending on this commit. `.github/instructions/pr-workflow.instructions.md` asks for validation green before ready-for-review.

### E2E test

I replied on your own line-2314 thread with a concrete shape rather than opening a competing one. Short version: the suite does already have legacy `DATETIME` cases, but the only fractional value any of them uses is `.010`, which is 3 ticks and yields exactly 10 ms under both the old and the new formula — so nothing currently in E2E can tell the two apart.

---

_Two inline notes below, both optional._

mssql-odbc/src/conversion/fetch_convert.rs
saurabh500: Concrete shape for this, and one correction to what I first assumed.

The E2E suite does already cover legacy `DATETIME` — `fetch_error_semantics_test.cpp:200,217,231,376` and `col_attribute_test.cpp:220,254,272`. What is missing is a case that can tell this fix apart from the old behavior. The only fractional value any of them uses is `.010` at `fetch_error_semantics_test.cpp:217`, which is 3 ticks and yields exactly 10 ms under both formulas, and none of them asserts `fraction` at all.

So the value matters more than the placement. `.127` is the one to use:

```cpp
TEST_F(DateTimeTypesLiveTest, DatetimeFractionRoundsToMillisecond) {
    ASSERT_SQL_OK(ExecDirect("SELECT CAST('2023-06-15 12:34:56.127' AS DATETIME) AS c1"),
                  SQL_HANDLE_STMT, stmt_);
    ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);

    SQL_TIMESTAMP_STRUCT ts{};
    SQLLEN ind = 0;
    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_TYPE_TIMESTAMP, &ts, sizeof(ts), &ind),
                  SQL_HANDLE_STMT, stmt_);
    EXPECT_EQ(127000000u, ts.fraction);
    SQLCloseCursor(stmt_);
}
```

`.127` stores as 38 ticks, where the old code produced 126 ms and plain truncation would also produce 126 — so it is the only shape that fails against both. `.123` is 37 ticks and rounds down, so it passes with or without the `+ 150`, and would be a test that guards nothing.

Worth pairing it with a `SQL_C_CHAR` fetch of the same column asserting `"2023-06-15 12:34:56.127"`, since `format_datetime_parts` is the other consumer this fix corrects and it has no E2E coverage on a rounding-sensitive value either.

mssql-odbc/src/conversion/fetch_convert.rs
saurabh500: Updated the PR description to call out character rendering as a user-visible part of the fix, along with the new text-path E2E coverage.

> [!NOTE]

mssql-odbc/src/conversion/fetch_convert.rs
saurabh500: Expanded the unit test to compare all 300 tick remainders against msodbcsql's reference expression, including the zero-fraction case.

> [!NOTE]

mssql-odbc/src/conversion/fetch_convert.rs
saurabh500: Added a rounding-sensitive `.127` E2E case covering timestamp `SQLGetData`, bound fetch, and `SQL_C_CHAR` rendering. The targeted case passes against both mssql-odbc and pinned msodbcsql 18.6.2.1.

> [!NOTE]

David-Engel (commented):

No findings. This is a re-review of `a5c7e695`; the delta since the last automated pass at `45ced7d9` is the all-300-remainder unit loop and the new E2E case, and both hold up.

**The E2E case genuinely guards the fix.** `CAST('2024-05-20T12:34:56.127' AS DATETIME)` lands on tick 38 (`.127 * 300 = 38.1`, SQL Server rounds to 38), which is exactly a remainder where old and new behavior diverge: truncation gave `fraction_ns = 38 * 1e9 / 300 = 126_666_666`, rounding gives `127_000_000`. So `EXPECT_EQ(127000000u, ...)` fails on `main` rather than passing either way. The text assertion is equally load-bearing — `format_datetime_parts` does `fraction_ns / 100` formatted to 7 digits and sliced to `scale`, so `"1266666"[..3]` = `.126` before and `"1270000"[..3]` = `.127` after. Picking the `T`-separated literal for the input while asserting the space-separated form on output is the right call: the `T` form is language-neutral for `datetime`, the space form is what the driver emits.

**All three arms exercise distinct code.** I traced the bound path rather than assuming it: `write_datetime` (`api/fetch_scroll.rs:404`) hands `ColumnValues::DateTime` to the generic `write_value` → `deliver_bound`, not to a `write_temporal` fast path like `write_datetime2` uses, so it lands back in `convert_datetime_c` → `extract_datetime_parts`. The `SQL_C_CHAR` arm reaches the same function via `api/get_data.rs:1456`. One fix, three surfaces, all asserted.

**The exact-fit text buffer is fine.** `char text[sizeof(expectedText)]` is 24 bytes for a 23-char rendering, and `ASSERT_SQL_OK` accepts `SQL_SUCCESS_WITH_INFO`, so a truncation would not fail there — but `textInd` carries the *untruncated* length under ODBC, and asserting it equals 23 alongside `EXPECT_STREQ` closes that gap. No headroom needed.

**Unit loop.** `((ticks * 20 + 3) / 6)` is msodbcsql's `sqlccnvt.cpp:272` expression and is algebraically the same as the implementation's `((t * 1000 + 150) / 300)` — dividing both terms by 50 maps one onto the other exactly, floor division included. That makes it a restatement rather than an independent oracle, but it is the *reference driver's* restatement, which is the contract that matters here, and it still fails if the `+ 150` is dropped. Ran `-p mssqlodbc --lib -E 'test(datetime)'` on this head: 21 passed, 0 failed, so the neighbouring legacy-`datetime` and truncation-warning cases are unaffected. `fraction_ms` is still zero exactly when `dt.time % 300 == 0`, so no truncation-warning arm changes verdict.

`SmallDateTime` is minute-resolution and never touches the tick math, and `extract_datetime_parts` remains the only 1/300-second conversion in the crate, so the blast radius is still contained.

One note, not a finding, and the same one as last time: the checklist still has `cargo bclippy` and `cargo btest` unchecked while all five ADO build stages and the cross-repo `mssql-python` suite are green on this head. Worth ticking before a human picks this up, since the boxes are the first thing they read.

Posted by an unattended run — these findings were not checked by a human before posting.

Theekshna (commented):
Unattended run found no new findings. Covered: msodbcsql parity for the rounding formula (independently re-derived `(r*1000+150)/300` against `sqlccnvt.cpp:272`'s `((ThreeHundredthsSec*20+3)/6)` — algebraically identical, confirmed no overflow/carry), the e2e `.127`-tick mutation check (confirmed tick 38 is the genuine old-vs-new divergence point), blast radius (`extract_datetime_parts` is the only 1/300s tick conversion site), divergence docs (N/A — this fixes a bug, not a divergence), and AI-slop in the diff — none found. The unchecked `cargo bclippy`/`cargo btest` checklist boxes and a `coverage-report` GH Actions timeout (75 min, no artifact found) were already flagged by prior reviewers / are CI infra noise given all ADO validation stages are green; not re-filing.

| Severity | Count |
| --- | ---: |
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

David-Engel (approved):

PR #489, opened 2026-09-03
Report NUMERIC result columns as SQL_NUMERIC

saurabh500:
## Description

`SQLDescribeColW` now reports SQL Server `NUMERIC` result columns, including uncast decimal literals that SQL Server infers as `NUMERIC`, as `SQL_NUMERIC` while preserving `SQL_DECIMAL` for `DECIMALN` and money types. For full msodbcsql parity, the legacy fixed `SQLDECIMAL` token also reports `SQL_NUMERIC`. This restores exact SQL-type output-converter dispatch in mssql-python.

The mapping matches `Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp` in msodbcsql. The bundled msodbcsql 18.6.2.1 reference driver (`SQL_DRIVER_VER=18.06.0002`) and the fixed Rust driver both pass the two affected mssql-python tests. Unit coverage anchors the legacy and nullable decimal, numeric, and money mappings; the live e2e case covers explicit decimal/numeric columns and uncast decimal literals.

Validation:
- `cargo bfmt`
- `cargo bclippy`
- `cargo btest -E 'not binary(connectivity)'` (3,760 tests passed)
- `cargo nextest run -p mssqlodbc decimal_numeric_and_money_types_match_msodbcsql`
- The two affected `tests/test_003_connection.py` pytest cases (2 passed)

The unfiltered `cargo btest` run has four unrelated `mssql-tds::connectivity` failures that require Azure identity or certificate-specific test conditions.

## Related Issues

[AB#47818](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47818)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes
- [x] Public API changes are documented (N/A - no public API change)

### DIFF
--- a/mssql-odbc/src/api/describe_col.rs
+++ b/mssql-odbc/src/api/describe_col.rs
@@ -196,8 +196,9 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
             8 => SQL_DOUBLE,
             _ => SQL_UNKNOWN_TYPE,
         },
-        TdsDataType::Decimal | TdsDataType::DecimalN => SQL_DECIMAL,
-        TdsDataType::Numeric | TdsDataType::NumericN => SQL_NUMERIC,
+        TdsDataType::DecimalN => SQL_DECIMAL,
+        // msodbcsql's rgbSRV2SQLTYPE maps the legacy fixed SQLDECIMAL token to SQL_NUMERIC.
+        TdsDataType::Decimal | TdsDataType::Numeric | TdsDataType::NumericN => SQL_NUMERIC,
         TdsDataType::Money | TdsDataType::Money4 | TdsDataType::MoneyN => SQL_DECIMAL,
         TdsDataType::DateN => SQL_TYPE_DATE,
         // SQL Server's `time` supports up to 7-digit fractional seconds; SQL_TYPE_TIME
--- a/mssql-odbc/src/api/describe_col.rs
+++ b/mssql-odbc/src/api/describe_col.rs
@@ -315,9 +315,8 @@ pub(crate) fn decimal_digits(meta: &mssql_tds::query::metadata::ColumnMetadata) 
     }
 }
 
-// Unit tests cover the validation/error paths only. The metadata-driven mapping
-// helpers (`odbc_sql_type`, `column_size`, `decimal_digits`) cannot be exercised
-// here because `mssql_tds::ColumnMetadata::type_info_variant` is `pub(crate)`
+// Precision- and scale-dependent branches in `column_size` and `decimal_digits` cannot
+// be exercised here because `mssql_tds::ColumnMetadata::type_info_variant` is `pub(crate)`
 // and there is no public constructor — those branches are covered end-to-end by
 // `tests/e2e/tests/describe_col_test.cpp` against a live SQL Server.
 #[cfg(test)]
@@ -354,6 +353,28 @@ mod tests {
     fn null_handle_returns_invalid_handle() {
         let rc = unsafe { describe(ptr::null_mut(), 1) };
         assert_eq!(rc, SQL_INVALID_HANDLE);
+    }
+
+    #[test]
+    fn decimal_numeric_and_money_types_match_msodbcsql() {
+        let mut columns = int_columns(1);
+        let meta = &mut columns[0];
+
+        for (tds_type, length, sql_type) in [
+            (TdsDataType::Decimal, 17, SQL_NUMERIC),
+            (TdsDataType::DecimalN, 17, SQL_DECIMAL),
+            (TdsDataType::Numeric, 17, SQL_NUMERIC),
+            (TdsDataType::NumericN, 17, SQL_NUMERIC),
+            (TdsDataType::Money, 8, SQL_DECIMAL),
+            (TdsDataType::Money4, 4, SQL_DECIMAL),
+            (TdsDataType::MoneyN, 8, SQL_DECIMAL),
+        ] {
+            meta.data_type = tds_type;
+            meta.type_info.tds_type = tds_type;
+            meta.type_info.length = length;
+
+            assert_eq!(odbc_sql_type(meta), sql_type, "{tds_type:?}");
+        }
     }
 
     #[test]
--- a/mssql-odbc/src/api/describe_col.rs
+++ b/mssql-odbc/src/api/describe_col.rs
@@ -9,10 +9,10 @@ use tracing::{debug, error};
 use crate::api::odbc_types::{
     SQL_BIGINT, SQL_BINARY, SQL_BIT, SQL_CHAR, SQL_DECIMAL, SQL_DOUBLE, SQL_ERROR, SQL_GUID,
     SQL_INTEGER, SQL_INVALID_HANDLE, SQL_LONGVARBINARY, SQL_LONGVARCHAR, SQL_NO_NULLS,
-    SQL_NULLABLE, SQL_REAL, SQL_SMALLINT, SQL_SS_TIME2, SQL_SS_TIMESTAMPOFFSET, SQL_SS_VARIANT,
-    SQL_SUCCESS, SQL_SUCCESS_WITH_INFO, SQL_TINYINT, SQL_TYPE_DATE, SQL_TYPE_TIMESTAMP,
-    SQL_UNKNOWN_TYPE, SQL_VARBINARY, SQL_VARCHAR, SQL_WCHAR, SQL_WLONGVARCHAR, SQL_WVARCHAR,
-    SqlHandle, SqlReturn, SqlSmallInt, SqlUSmallInt, SqlWChar,
+    SQL_NULLABLE, SQL_NUMERIC, SQL_REAL, SQL_SMALLINT, SQL_SS_TIME2, SQL_SS_TIMESTAMPOFFSET,
+    SQL_SS_VARIANT, SQL_SUCCESS, SQL_SUCCESS_WITH_INFO, SQL_TINYINT, SQL_TYPE_DATE,
+    SQL_TYPE_TIMESTAMP, SQL_UNKNOWN_TYPE, SQL_VARBINARY, SQL_VARCHAR, SQL_WCHAR, SQL_WLONGVARCHAR,
+    SQL_WVARCHAR, SqlHandle, SqlReturn, SqlSmallInt, SqlUSmallInt, SqlWChar,
 };
 use crate::api::sqlstate::{
     ERR_FUNCTION_SEQUENCE, ERR_INVALID_DESCRIPTOR_INDEX, WARN_STRING_TRUNCATION, post_diag,
@@ -196,10 +196,8 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
             8 => SQL_DOUBLE,
             _ => SQL_UNKNOWN_TYPE,
         },
-        TdsDataType::Decimal
-        | TdsDataType::DecimalN
-        | TdsDataType::Numeric
-        | TdsDataType::NumericN => SQL_DECIMAL,
+        TdsDataType::Decimal | TdsDataType::DecimalN => SQL_DECIMAL,
+        TdsDataType::Numeric | TdsDataType::NumericN => SQL_NUMERIC,
         TdsDataType::Money | TdsDataType::Money4 | TdsDataType::MoneyN => SQL_DECIMAL,
         TdsDataType::DateN => SQL_TYPE_DATE,
         // SQL Server's `time` supports up to 7-digit fractional seconds; SQL_TYPE_TIME
--- a/mssql-odbc/tests/e2e/tests/describe_col_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/describe_col_test.cpp
@@ -7,7 +7,7 @@
 //   3.  BasicMetadata                       - INT/VARCHAR/NVARCHAR cols → name + type + size
 //   4.  NameTruncationReturnsInfo           - short name buf → SUCCESS_WITH_INFO + 01004 + full length
 //   5.  InvalidColumnOrdinal                - column 0 (bookmark) and past-end → 07009
-//   6.  DecimalPrecisionAndScale            - DECIMAL(10,2) → SQL_DECIMAL, colSize=10, decDigits=2
+//   6.  DecimalAndNumericPrecisionAndScale  - DECIMAL/NUMERIC preserve distinct SQL type codes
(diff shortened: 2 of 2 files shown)


mssql-odbc/src/api/describe_col.rs
```diff
@@ -196,10 +196,8 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
             8 => SQL_DOUBLE,
             _ => SQL_UNKNOWN_TYPE,
         },
-        TdsDataType::Decimal
-        | TdsDataType::DecimalN
-        | TdsDataType::Numeric
-        | TdsDataType::NumericN => SQL_DECIMAL,
+        TdsDataType::Decimal | TdsDataType::DecimalN => SQL_DECIMAL,
```
shiwanigupta0809: **Suggestion** — `TdsDataType::Decimal` (the legacy fixed `SQLDECIMAL`, `0x37`) does not map to `SQL_DECIMAL` in the table the PR description cites. It maps to `SQL_NUMERIC`.

From `rgbSRVTYPE2SQLTYPE` in `Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp`, the four relevant rows:

```cpp
SQL_DECIMAL,        // SQLDECIMALN     (0x6A)
SQL_NUMERIC,        // SQLNUMERICN     (0x6C)
...
SQL_NUMERIC,        // SQLDECIMAL      (0x37)   <-- this one
...
SQL_NUMERIC,        // SQLNUMERIC      (0x3F)
```

Cross-checked the TDS values against `sqldatatypes.rs:49,63,71,73` so the labels line up: `Decimal = 0x37`, `Numeric = 0x3F`, `DecimalN = 0x6A`, `NumericN = 0x6C`.

So three of the four arms match msodbcsql and `Decimal` is the odd one out — msodbcsql answers `SQL_NUMERIC` for it, this line answers `SQL_DECIMAL`.

**This is almost certainly unreachable**, which is why I am raising it as a suggestion rather than a bug: TDS 7.x sends `DECIMALN` for every `decimal` and `numeric` column, so the fixed-length `0x37` never arrives from a modern SQL Server. Nothing in the mssql-python scenario this fixes goes near it.

What makes it worth a line is the claim rather than the behaviour. The PR body says "The mapping matches `Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp` in msodbcsql", and the next person to touch this arm will read that as covering all four. Two options, either is fine:

- Match the table — `TdsDataType::DecimalN => SQL_DECIMAL` and fold `Decimal` in with `Numeric | NumericN => SQL_NUMERIC`. Odd-looking, so it would want the one-line reason.
- Keep the current grouping, which is arguably the more coherent answer, and note that `SQLDECIMAL` deliberately diverges because the legacy fixed type is unreachable on TDS 7.x — the same treatment the other recorded deviations get.

Worth knowing either way that `Money`/`Money4`/`MoneyN → SQL_DECIMAL` on the line below is confirmed correct against the same table (`SQLMONEYN`, `SQLMONEY4` and `SQLMONEY` all map to `SQL_DECIMAL`), so the money arm did not need to move with this change.

shiwanigupta0809 (commented):
Read the diff, traced every consumer of `odbc_sql_type`, and checked the mapping against `sqlcmisc.cpp` rather than taking the description's claim. Only Copilot has reviewed this so far, and it generated no comments.

**Blocking:** none. **Suggestions:** 1 (inline).

### Verified

**No downstream consumer is broken by the split.** This is the thing that would make a metadata change like this expensive, so I enumerated it rather than spot-checking. Every place that keys on the exact numeric SQL type already treats the pair together:

- `type_rules.rs:138` — `SQL_DECIMAL | SQL_NUMERIC => 1..=SQL_PREC_NUMERIC`
- `type_rules.rs:157` — `parameter_size_is_precision`
- `type_rules.rs:258-259` — the supported-type list
- `type_rules.rs:312` — `resolve_default_c_type` → `SQL_C_CHAR`
- `col_attribute.rs:566-567` — `SQL_DESC_UNSIGNED`
- `get_type_info.rs:256-257` — `SQLGetTypeInfo` already emits both, so an application that reads `SQL_NUMERIC` from `SQLDescribeCol` and looks it up gets a row back

`column_size`, `decimal_digits`, `display_size`, `octet_length`, `searchable` and `num_prec_radix` all match on `TdsDataType` directly, so they never saw the ODBC type at all. And `type_name` (`col_attribute.rs:594-595`) already answered `"decimal"` / `"numeric"` separately — so before this change `SQL_DESC_TYPE_NAME` and `SQL_DESC_CONCISE_TYPE` disagreed with each other for a `numeric` column. This change makes them consistent, which is a second thing it fixes beyond the mssql-python dispatch.

**The e2e test is a real parity assertion.** No `SKIP_IF_COMPARING_MSODBCSQL()`, and msodbcsql maps `SQLDECIMALN → SQL_DECIMAL` / `SQLNUMERICN → SQL_NUMERIC`, so both legs genuinely agree rather than the test being self-fulfilling.

### Suggestion

One entry of the four does not match the table the description cites. Left inline at the exact arm.

shiwanigupta0809:
Reviewed — posted as a review with one inline suggestion on \describe_col.rs\: https://github.com/microsoft/mssql-rs/pull/489#pullrequestreview-5100012584

Short version: no downstream consumer breaks (traced all of them — every site already pairs \SQL_DECIMAL | SQL_NUMERIC\), and this also fixes a pre-existing disagreement between \SQL_DESC_TYPE_NAME\ and \SQL_DESC_CONCISE_TYPE\ for \
umeric\ columns. One of the four arms does not match the \sqlcmisc.cpp\ table the description cites — details inline.

mssql-odbc/src/api/describe_col.rs
```diff
@@ -196,10 +196,8 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
...
+        TdsDataType::Numeric | TdsDataType::NumericN => SQL_NUMERIC,
```
Theekshna: **Missing Rust unit test — the one thing I would like addressed before merge.**

The comment at lines 317-321 says the metadata-driven mapping helpers "cannot be exercised here" because `ColumnMetadata::type_info_variant` is `pub(crate)`. That holds for `column_size` and `decimal_digits`, which need precision/scale — but not for `odbc_sql_type`, which reads only `meta.data_type` and `meta.type_info.length`, both public. `col_attribute.rs:884 retype_column` already does exactly this against `int_columns()`.

So as it stands the only thing guarding this change is an e2e test that needs a live SQL Server and does not run under `cargo btest`.

The reported 100% diff coverage is incidental rather than a regression anchor: both changed lines are executed by `col_attribute`'s `unsigned_is_false_only_for_signed_numerics`, which retypes a column to `DecimalN`/`NumericN` and then only asserts `SQL_DESC_UNSIGNED` — it would pass whether this mapping returned 2 or 3.

Could you add a table-driven unit test over `Decimal` / `DecimalN` / `Numeric` / `NumericN` (plus the money arm, since it sits in the same block), and narrow the 317-321 comment to the two helpers it actually applies to?

mssql-odbc/src/api/describe_col.rs
saurabh500: Fixed in aff5354b: the legacy `SQLDECIMAL` token now maps to `SQL_NUMERIC`, matching msodbcsql's `rgbSRV2SQLTYPE`; `DecimalN` remains `SQL_DECIMAL`.

> [!NOTE]

mssql-odbc/src/api/describe_col.rs
```diff
@@ -196,10 +196,9 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
             8 => SQL_DOUBLE,
             _ => SQL_UNKNOWN_TYPE,
         },
-        TdsDataType::Decimal
-        | TdsDataType::DecimalN
-        | TdsDataType::Numeric
-        | TdsDataType::NumericN => SQL_DECIMAL,
+        TdsDataType::DecimalN => SQL_DECIMAL,
+        // msodbcsql's rgbSRV2SQLTYPE maps the legacy fixed SQLDECIMAL token to SQL_NUMERIC.
+        TdsDataType::Decimal | TdsDataType::Numeric | TdsDataType::NumericN => SQL_NUMERIC,
```
Theekshna, line 201: **Missing Rust unit test — my only pre-merge ask.**

The comment at lines 318-322 says the metadata-driven mapping helpers "cannot be exercised here" because `ColumnMetadata::type_info_variant` is `pub(crate)`. That holds for `column_size` and `decimal_digits`, which need precision/scale — but not for `odbc_sql_type`, which reads only `meta.data_type` and `meta.type_info.length`, both public. `col_attribute.rs:884 retype_column` already does exactly this against `int_columns()`.

So the only thing guarding this mapping is an e2e test that needs a live SQL Server and does not run under `cargo btest`.

The reported 100% diff coverage is incidental rather than a regression anchor: the changed lines are executed by `col_attribute`'s `unsigned_is_false_only_for_signed_numerics`, which retypes a column to `DecimalN`/`NumericN` and then only asserts `SQL_DESC_UNSIGNED` — it would pass whether this mapping returned 2 or 3.

This matters a little more after `aff5354b`, since the arm now encodes a deliberate, non-obvious split (`DecimalN` alone on `SQL_DECIMAL`, the legacy `Decimal` token grouped with the numerics) that a future reader could easily "tidy up" with nothing in the default test run to stop them.

Could you add a table-driven unit test over `Decimal` / `DecimalN` / `Numeric` / `NumericN` (plus the money arm, since it sits in the same block), and narrow the 318-322 comment to the two helpers it actually applies to?

mssql-odbc/tests/e2e/tests/describe_col_test.cpp
```diff
@@ -142,8 +142,8 @@ TEST_F(DescribeColLiveTest, InvalidColumnOrdinal) {
     SQLCloseCursor(stmt_);
 }
 
-TEST_F(DescribeColLiveTest, DecimalPrecisionAndScale) {
-    ExecDirect("SELECT CAST(3.14 AS DECIMAL(10,2)) AS d");
+TEST_F(DescribeColLiveTest, DecimalAndNumericPrecisionAndScale) {
+    ExecDirect("SELECT CAST(3.14 AS DECIMAL(10,2)) AS d, CAST(6.28 AS NUMERIC(12,3)) AS n");
```
Theekshna: Scope note rather than a defect, and not something I saw raised elsewhere: I believe this change reaches further than "NUMERIC result columns" implies. SQL Server types bare decimal literals as `numeric`, so `SELECT 1.5` should now report `SQL_NUMERIC` where it previously reported `SQL_DECIMAL` — affecting every un-cast decimal literal, not only explicitly declared `NUMERIC` columns.

That is the msodbcsql-matching outcome (it maps 0x6C to `SQL_NUMERIC`), so the driver was previously diverging on decimal literals too — which strengthens the case for this change rather than arguing against it.

**Caveat: I could not verify this against a live server**, so please treat it as unconfirmed. Adding an un-cast `1.5` as a third column here would both confirm and pin it, and it is worth a line in the PR description so downstream consumers with `SQL_DECIMAL`-keyed dispatch understand the blast radius.

Theekshna (commented):
Reviewed at `aff5354b`, after @shiwanigupta0809. I traced the same ground independently — every consumer of `odbc_sql_type`, and the mapping against `rgbSRV2SQLTYPE` in `Sql/Ntdbms/sqlncli/odbc/sqlcmisc.cpp` — and reached the same conclusions, so I am not restating them. Her downstream-consumer enumeration and the `SQL_DESC_TYPE_NAME` / `SQL_DESC_CONCISE_TYPE` inconsistency point both match what I found. The legacy `SQLDECIMAL` (0x37) divergence I had also flagged is resolved by `aff5354b`, so I dropped my duplicate of it rather than re-raising it.

I re-checked `aff5354b` specifically: folding `Decimal` in with `Numeric | NumericN` changes nothing downstream, since every consumer already treats `SQL_DECIMAL` and `SQL_NUMERIC` as a pair, and 0x37 is unreachable on TDS 7.x either way.

Two things I did not see raised, neither blocking the mapping change.

### 1. No Rust unit test (inline on the mapping arm)
The only guard on this mapping is an e2e test that needs a live SQL Server and does not run under `cargo btest`. That is more fixable than the comment at `describe_col.rs:318-322` suggests — details inline. This is my only pre-merge ask.

### 2. The change is broader than "NUMERIC result columns"
`SELECT 1.5` should now report `SQL_NUMERIC` as well, since SQL Server types bare decimal literals as `numeric`. Flagged inline on the test; I could not confirm it against a live server, so please treat it as unverified.

### One addition to the parity verification
Beyond the numeric arms and money, I also confirmed that nothing treats the result of `odbc_sql_type` as a C type, so the `SQL_NUMERIC` / `SQL_C_NUMERIC` value collision (both are `2`) is unreachable: `fetch_scroll.rs:899` routes through `resolve_default_c_type`, and `col_attribute.rs:516` deliberately answers `SQL_C_CHAR` for a variant holding a decimal. And `describe_col_test.cpp:146` is the only e2e test in the suite that casts to `NUMERIC`, so nothing existing regresses on the new mapping.

**Blocking:** none.

mssql-odbc/tests/e2e/tests/describe_col_test.cpp
saurabh500: Added uncast `1.5` coverage in d39a9ed7 with `SQL_NUMERIC`, precision 2, and scale 1 assertions, and documented the literal scope in the PR description. Those expectations follow SQL Server’s documented minimum-precision `numeric` inference for decimal-point constants.

> [!NOTE]

mssql-odbc/src/api/describe_col.rs
saurabh500: Added table-driven unit coverage in d39a9ed7 for `Decimal`, `DecimalN`, `Numeric`, `NumericN`, and all three money token variants. I also narrowed the comment to only the precision/scale-dependent helpers.

> [!NOTE]

Theekshna (commented):
Unattended sweep review. No new findings — both asks from my prior review (Rust unit test for the `Decimal`/`DecimalN`/`Numeric`/`NumericN`/money mapping arm, and confirming the un-cast-literal case against a live server) were addressed by the latest commit. Independently re-derived the full `rgbSRV2SQLTYPE` mapping for all seven wire types in `sqlcmisc.cpp` — all match. Divergence docs, description and checklist are current; no AI slop. CI still has several jobs pending at review time; everything reported so far is green.

Theekshna (approved):

PR #490, opened 2026-09-03
Report CLR UDT columns as SQL_SS_UDT

saurabh500:
## Description

SQL Server CLR UDT columns were exposed by `SQLDescribeCol` as `SQL_VARCHAR`, causing consumers such as mssql-python to select character bindings for binary UDT values.

Map `TdsDataType::Udt` to the SQL Server-specific `SQL_SS_UDT` type and report the wire-declared maximum byte size for bounded UDTs. Live parity coverage checks `SQLDescribeCol` type, size, scale, and nullability plus `SQLColAttribute` fields `SQL_DESC_TYPE`, `SQL_DESC_CONCISE_TYPE`, `SQL_DESC_LENGTH`, `SQL_DESC_PRECISION`, `SQL_DESC_OCTET_LENGTH`, and `SQL_DESC_DISPLAY_SIZE` for `geography`, `geometry`, and `hierarchyid`. This matches retail msodbcsql: the unbounded spatial types report zero size, while `hierarchyid` reports 892 for length, precision, and octet length and zero display size.

`odbc_sql_type` also drives fetch-side `SQL_C_DEFAULT` resolution, so a UDT now resolves to `SQL_C_BINARY` instead of `SQL_C_CHAR`. Observable bound-fetch behavior remains `HYC00` before and after because binary PLP delivery is not implemented yet ([AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239)); actual UDT value delivery remains outside this work item.

Vector deliberately retains its existing `SQL_VARCHAR` mapping. Vector values currently have a working character-rendering path, while result delivery for `SQL_C_SS_VECTOR` is not implemented; changing the advertised type here would regress `SQL_C_DEFAULT` bound fetches to `HYC00` and needs dedicated vector result-delivery coverage.

**Validation**

- `cargo bfmt`
- `cargo bclippy`
- Focused Rust UDT metadata and descriptor-size unit tests
- `cargo btest -E "not (test(connectivity))"`: 3,761 passed, 40 skipped
- Full C++ ODBC suite: 38/38 binaries passed against both the Rust driver and retail msodbcsql 18.6.2.1 (`SQL_DRIVER_VER=18.06.0002`), with zero parity divergences
- Targeted mssql-python metadata tests: 3 failures before, 3 passes after
- Full mssql-python suite: failures/errors reduced from 253 to 250; only the geography, geometry, and hierarchyid metadata tests were resolved, with no new or missing tests

## Related Issues

Fixes [AB#47541](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47541)

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47541

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

### DIFF
--- a/mssql-odbc/src/api/col_attribute.rs
+++ b/mssql-odbc/src/api/col_attribute.rs
@@ -408,6 +408,8 @@ fn display_size(meta: &ColumnMetadata) -> SqlLen {
 /// representation, which for the temporal types is the C struct the driver
 /// hands back, not the TDS payload width.
 pub(super) fn octet_length(meta: &ColumnMetadata) -> SqlLen {
+    // A bounded UDT is still PLP; use its byte limit for binary transfer while
+    // display_size remains zero because the opaque value has no text rendering.
     if meta.data_type == TdsDataType::Udt {
         return desc_length(meta);
     }
--- a/mssql-odbc/src/api/col_attribute.rs
+++ b/mssql-odbc/src/api/col_attribute.rs
@@ -335,7 +335,7 @@ fn binary_precision(meta: &ColumnMetadata) -> Option<SqlSmallInt> {
 /// digits) for a column size of 10, a GUID needs 36, and binary renders as two
 /// hex characters per byte.
 fn display_size(meta: &ColumnMetadata) -> SqlLen {
-    // `*(max)`, xml and json are unbounded; ODBC reports zero.
+    // `*(max)`, xml, json, and opaque UDTs have no fixed character display size.
     if meta.is_plp() {
         return 0;
     }
@@ -408,6 +408,9 @@ fn display_size(meta: &ColumnMetadata) -> SqlLen {
 /// representation, which for the temporal types is the C struct the driver
 /// hands back, not the TDS payload width.
 pub(super) fn octet_length(meta: &ColumnMetadata) -> SqlLen {
+    if meta.data_type == TdsDataType::Udt {
+        return desc_length(meta);
+    }
     if meta.is_plp() {
         return 0;
     }
@@ -632,8 +635,7 @@ fn type_name(meta: &ColumnMetadata) -> &'static str {
     }
 }
 
-// Only `int` column metadata can be built outside the decoder (`int_columns`),
-// so the per-type mapping tables are covered end-to-end by
+// Most per-type mapping tables are covered end-to-end by
 // `tests/e2e/tests/col_attribute_test.cpp` against a live SQL Server.
 #[cfg(test)]
 mod tests {
@@ -646,7 +648,7 @@ mod tests {
     use crate::api::sqlstate::ERR_INVALID_DESCRIPTOR_FIELD;
     use crate::test_support::TestHandles;
     use mssql_tds::datatypes::sqldatatypes::TypeInfo;
-    use mssql_tds::test_client_support::int_columns;
+    use mssql_tds::test_client_support::{int_columns, udt_column};
 
     /// A statement positioned on a result set of `n` nullable `int` columns.
     fn stmt_with_int_columns(h: &TestHandles, n: usize) {
@@ -1387,6 +1389,18 @@ mod tests {
                 .expect("varchar(max) is a PLP type");
         }
         assert_eq!(numeric(&h, 1, SQL_DESC_OCTET_LENGTH), 0);
+    }
+
+    #[test]
+    fn udt_size_attributes_match_bounded_and_unbounded_metadata() {
+        for (max_byte_size, expected_size) in [(u16::MAX, 0), (892, 892)] {
+            let meta = udt_column(max_byte_size);
+
+            assert_eq!(desc_length(&meta), expected_size);
+            assert_eq!(SqlLen::from(precision(&meta)), expected_size);
+            assert_eq!(octet_length(&meta), expected_size);
+            assert_eq!(display_size(&meta), 0);
+        }
     }
 
     /// A `decimal` carries its own precision on the wire, which takes
--- a/mssql-odbc/src/api/describe_col.rs
+++ b/mssql-odbc/src/api/describe_col.rs
@@ -228,12 +228,19 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
         // reporting the column as character data hides the variant entirely.
         TdsDataType::SsVariant => SQL_SS_VARIANT,
         TdsDataType::Udt => SQL_SS_UDT,
+        // Result delivery does not support SQL_C_SS_VECTOR yet; retaining the
+        // character type keeps SQL_C_DEFAULT fetches on the working text path.
         TdsDataType::Vector => SQL_VARCHAR,
         _ => SQL_UNKNOWN_TYPE,
     }
 }
 
 pub(crate) fn column_size(meta: &mssql_tds::query::metadata::ColumnMetadata) -> u64 {
+    // CLR UDT metadata is PLP even when MAX_BYTE_SIZE is bounded. Only 0xFFFF
+    // carries the unbounded convention used by geography and geometry.
+    if meta.data_type == TdsDataType::Udt && meta.type_info.length != usize::from(u16::MAX) {
+        return meta.type_info.length as u64;
+    }
     // PLP / `*(max)` / xml / json: ColumnSize is "unbounded". Report 0 per ODBC spec
     if meta.is_plp() {
         return 0;
@@ -323,7 +330,7 @@ mod tests {
 
     use super::*;
     use crate::test_support::TestHandles;
-    use mssql_tds::test_client_support::int_columns;
+    use mssql_tds::test_client_support::{int_columns, udt_column};
 
     /// Calls `sql_describe_col_w` with default-ish out pointers. Intended for
     /// error-path tests where the values of the out params are irrelevant.
@@ -354,14 +361,15 @@ mod tests {
     }
 
     #[test]
-    fn udt_column_reports_sql_ss_udt() {
-        let mut metadata = int_columns(1);
-        let meta = metadata.first_mut().unwrap();
-        meta.data_type = TdsDataType::Udt;
-        meta.type_info.tds_type = TdsDataType::Udt;
-        meta.type_info.length = usize::from(u16::MAX);
+    fn udt_column_reports_sql_ss_udt_metadata() {
+        for (max_byte_size, expected_column_size) in [(u16::MAX, 0), (892, 892)] {
+            let meta = udt_column(max_byte_size);
 
-        assert_eq!(odbc_sql_type(meta), SQL_SS_UDT);
+            assert!(meta.is_plp());
+            assert_eq!(odbc_sql_type(&meta), SQL_SS_UDT);
+            assert_eq!(column_size(&meta), expected_column_size);
+            assert_eq!(decimal_digits(&meta), 0);
+        }
     }
 
     #[test]
--- a/mssql-odbc/src/api/describe_col.rs
+++ b/mssql-odbc/src/api/describe_col.rs
@@ -9,10 +9,10 @@ use tracing::{debug, error};
 use crate::api::odbc_types::{
     SQL_BIGINT, SQL_BINARY, SQL_BIT, SQL_CHAR, SQL_DECIMAL, SQL_DOUBLE, SQL_ERROR, SQL_GUID,
     SQL_INTEGER, SQL_INVALID_HANDLE, SQL_LONGVARBINARY, SQL_LONGVARCHAR, SQL_NO_NULLS,
-    SQL_NULLABLE, SQL_REAL, SQL_SMALLINT, SQL_SS_TIME2, SQL_SS_TIMESTAMPOFFSET, SQL_SS_VARIANT,
-    SQL_SUCCESS, SQL_SUCCESS_WITH_INFO, SQL_TINYINT, SQL_TYPE_DATE, SQL_TYPE_TIMESTAMP,
-    SQL_UNKNOWN_TYPE, SQL_VARBINARY, SQL_VARCHAR, SQL_WCHAR, SQL_WLONGVARCHAR, SQL_WVARCHAR,
-    SqlHandle, SqlReturn, SqlSmallInt, SqlUSmallInt, SqlWChar,
+    SQL_NULLABLE, SQL_REAL, SQL_SMALLINT, SQL_SS_TIME2, SQL_SS_TIMESTAMPOFFSET, SQL_SS_UDT,
+    SQL_SS_VARIANT, SQL_SUCCESS, SQL_SUCCESS_WITH_INFO, SQL_TINYINT, SQL_TYPE_DATE,
+    SQL_TYPE_TIMESTAMP, SQL_UNKNOWN_TYPE, SQL_VARBINARY, SQL_VARCHAR, SQL_WCHAR, SQL_WLONGVARCHAR,
+    SQL_WVARCHAR, SqlHandle, SqlReturn, SqlSmallInt, SqlUSmallInt, SqlWChar,
 };
 use crate::api::sqlstate::{
     ERR_FUNCTION_SEQUENCE, ERR_INVALID_DESCRIPTOR_INDEX, WARN_STRING_TRUNCATION, post_diag,
@@ -227,7 +227,8 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
         // mssql-python keys its sql_variant handling off this exact type, so
         // reporting the column as character data hides the variant entirely.
         TdsDataType::SsVariant => SQL_SS_VARIANT,
-        TdsDataType::Vector | TdsDataType::Udt => SQL_VARCHAR,
+        TdsDataType::Udt => SQL_SS_UDT,
+        TdsDataType::Vector => SQL_VARCHAR,
         _ => SQL_UNKNOWN_TYPE,
     }
 }
@@ -316,11 +317,6 @@ pub(crate) fn decimal_digits(meta: &mssql_tds::query::metadata::ColumnMetadata) 
     }
 }
 
-// Unit tests cover the validation/error paths only. The metadata-driven mapping
-// helpers (`odbc_sql_type`, `column_size`, `decimal_digits`) cannot be exercised
-// here because `mssql_tds::ColumnMetadata::type_info_variant` is `pub(crate)`
-// and there is no public constructor — those branches are covered end-to-end by
-// `tests/e2e/tests/describe_col_test.cpp` against a live SQL Server.
 #[cfg(test)]
 mod tests {
     use std::ptr;
@@ -355,6 +351,17 @@ mod tests {
     fn null_handle_returns_invalid_handle() {
         let rc = unsafe { describe(ptr::null_mut(), 1) };
         assert_eq!(rc, SQL_INVALID_HANDLE);
+    }
+
+    #[test]
+    fn udt_column_reports_sql_ss_udt() {
+        let mut metadata = int_columns(1);
+        let meta = metadata.first_mut().unwrap();
+        meta.data_type = TdsDataType::Udt;
+        meta.type_info.tds_type = TdsDataType::Udt;
+        meta.type_info.length = usize::from(u16::MAX);
+
+        assert_eq!(odbc_sql_type(meta), SQL_SS_UDT);
     }
 
     #[test]
--- a/mssql-odbc/src/api/describe_param.rs
+++ b/mssql-odbc/src/api/describe_param.rs
@@ -538,9 +538,9 @@ fn describe_tds_type(
         }
         TdsDataType::Image => (SQL_LONGVARBINARY, parameter_length(length, false)?, 0),
         TdsDataType::SsVariant => (SQL_SS_VARIANT, 8000, 0),
-        // Unbounded types report a size of 0, the same "unbounded" convention
-        // `describe_col::column_size` already uses for PLP. A table type has no
-        // meaningful column size at all.
+        // Parameter discovery does not include the CLR MAX_BYTE_SIZE carried by
+        // COLMETADATA, so UDT size remains 0. XML is unbounded; a table type has
+        // no meaningful column size at all.
         TdsDataType::Udt => (SQL_SS_UDT, 0, 0),
         TdsDataType::Xml => (SQL_SS_XML, 0, 0),
         TdsDataType::SqlTable => (SQL_SS_TABLE, 0, 0),
--- a/mssql-odbc/src/api/describe_param.rs
+++ b/mssql-odbc/src/api/describe_param.rs
@@ -538,10 +538,11 @@ fn describe_tds_type(
         }
         TdsDataType::Image => (SQL_LONGVARBINARY, parameter_length(length, false)?, 0),
         TdsDataType::SsVariant => (SQL_SS_VARIANT, 8000, 0),
-        // Parameter discovery does not include the CLR MAX_BYTE_SIZE carried by
-        // COLMETADATA, so UDT size remains 0. XML is unbounded; a table type has
-        // no meaningful column size at all.
+        // Although `suggested_tds_length` carries CLR MAX_BYTE_SIZE, retail
+        // msodbcsql 18.6.2.1 reports ParameterSize 0 for bounded and unbounded
+        // UDT parameters. Result columns report the bounded size.
         TdsDataType::Udt => (SQL_SS_UDT, 0, 0),
+        // XML is unbounded; a table type has no meaningful column size at all.
         TdsDataType::Xml => (SQL_SS_XML, 0, 0),
         TdsDataType::SqlTable => (SQL_SS_TABLE, 0, 0),
         // Neither of the next two arms fires against a shipping server. On SQL
@@ -973,6 +974,24 @@ mod tests {
 
         for (row, expected) in cases {
             assert_eq!(parse_parameter_row(&row, 1, true).unwrap().1, expected);
+        }
+    }
+
+    #[test]
+    fn udt_parameter_size_matches_msodbcsql() {
+        for length in [892, -1, i32::from(u16::MAX)] {
+            let (_, description) =
+                parse_parameter_row(&row(1, TdsDataType::Udt, length, 0, 0), 1, true).unwrap();
+
+            assert_eq!(
+                description,
+                ParameterDescription {
+                    data_type: SQL_SS_UDT,
+                    parameter_size: 0,
+                    decimal_digits: 0,
+                    nullable: SQL_NULLABLE,
+                }
+            );
         }
     }
 
--- a/mssql-odbc/tests/e2e/tests/describe_col_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/describe_col_test.cpp
@@ -11,12 +11,17 @@
 //   7.  NullableFlag                        - sys.objects.name (NOT NULL) vs principal_id (NULL)
 //   8.  DateTime2AndDateTimeOffset          - datetime2 → SQL_TYPE_TIMESTAMP, datetimeoffset → -155
 //   9.  NVarCharColumnSizeIsCharCount       - NVARCHAR(100) → colSize=100 (chars, not bytes)
+//  10.  ClrUdtColumnsReportSqlSsUdt          - geography/geometry/hierarchyid → SQL_SS_UDT
 
 #include "odbc_test_fixture.h"
 
 // SQL Server-specific types not in standard <sqlext.h>.
 #ifndef SQL_SS_TIMESTAMPOFFSET
 #define SQL_SS_TIMESTAMPOFFSET (-155)
+#endif
+
+#ifndef SQL_SS_UDT
+#define SQL_SS_UDT (-151)
 #endif
 
 // SQLDescribeColW with SQL_NULL_HSTMT — DM rejects before driver.
@@ -224,6 +229,22 @@ TEST_F(DescribeColLiveTest, NVarCharColumnSizeIsCharCount) {
     ASSERT_SQL_OK(rc, SQL_HANDLE_STMT, stmt_);
     EXPECT_EQ(SQL_WVARCHAR, dataType);
     EXPECT_EQ(100u, colSize);
+
+    SQLCloseCursor(stmt_);
+}
+
+TEST_F(DescribeColLiveTest, ClrUdtColumnsReportSqlSsUdt) {
+    ExecDirect("SELECT CAST(NULL AS geography) AS geography_col, "
+               "CAST(NULL AS geometry) AS geometry_col, "
+               "CAST(NULL AS hierarchyid) AS hierarchyid_col");
+
+    for (SQLUSMALLINT column : {SQLUSMALLINT{1}, SQLUSMALLINT{2}, SQLUSMALLINT{3}}) {
+        SQLSMALLINT dataType = 0;
+        SQLRETURN rc = SQLDescribeCol(
+            stmt_, column, nullptr, 0, nullptr, &dataType, nullptr, nullptr, nullptr);
+        ASSERT_SQL_OK(rc, SQL_HANDLE_STMT, stmt_);
+        EXPECT_EQ(SQL_SS_UDT, dataType) << "column=" << column;
+    }
 
     SQLCloseCursor(stmt_);
 }
--- a/mssql-odbc/tests/e2e/tests/describe_col_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/describe_col_test.cpp
@@ -11,7 +11,7 @@
 //   7.  NullableFlag                        - sys.objects.name (NOT NULL) vs principal_id (NULL)
 //   8.  DateTime2AndDateTimeOffset          - datetime2 → SQL_TYPE_TIMESTAMP, datetimeoffset → -155
 //   9.  NVarCharColumnSizeIsCharCount       - NVARCHAR(100) → colSize=100 (chars, not bytes)
-//  10.  ClrUdtColumnsReportSqlSsUdt          - geography/geometry/hierarchyid → SQL_SS_UDT
+//  10.  ClrUdtColumnsReportSqlSsUdt          - CLR UDT type, size, scale, and nullability
 
 #include "odbc_test_fixture.h"
 
@@ -238,12 +238,36 @@ TEST_F(DescribeColLiveTest, ClrUdtColumnsReportSqlSsUdt) {
                "CAST(NULL AS geometry) AS geometry_col, "
                "CAST(NULL AS hierarchyid) AS hierarchyid_col");
 
-    for (SQLUSMALLINT column : {SQLUSMALLINT{1}, SQLUSMALLINT{2}, SQLUSMALLINT{3}}) {
+    struct UdtColumn {
+        SQLUSMALLINT ordinal;
+        SQLULEN size;
+    };
+    const UdtColumn columns[] = {
+        {1, 0},
+        {2, 0},
+        {3, 892},
+    };
+
+    for (const auto& column : columns) {
         SQLSMALLINT dataType = 0;
+        SQLULEN colSize = 0;
+        SQLSMALLINT decDigits = -1;
+        SQLSMALLINT nullable = SQL_NULLABLE_UNKNOWN;
         SQLRETURN rc = SQLDescribeCol(
-            stmt_, column, nullptr, 0, nullptr, &dataType, nullptr, nullptr, nullptr);
+            stmt_,
+            column.ordinal,
+            nullptr,
+            0,
+            nullptr,
+            &dataType,
+            &colSize,
+            &decDigits,
+            &nullable);
         ASSERT_SQL_OK(rc, SQL_HANDLE_STMT, stmt_);
-        EXPECT_EQ(SQL_SS_UDT, dataType) << "column=" << column;
+        EXPECT_EQ(SQL_SS_UDT, dataType) << "column=" << column.ordinal;
+        EXPECT_EQ(column.size, colSize) << "column=" << column.ordinal;
+        EXPECT_EQ(0, decDigits) << "column=" << column.ordinal;
+        EXPECT_EQ(SQL_NULLABLE, nullable) << "column=" << column.ordinal;
     }
 
     SQLCloseCursor(stmt_);
--- a/.github/instructions/mssql-odbc.instructions.md
+++ b/.github/instructions/mssql-odbc.instructions.md
@@ -116,13 +116,13 @@ authoritative parity reference for this crate. Its source lives in the
     `BufferLength` 0 is exempt — the documented idiom for a fixed-width target,
     carrying no width claim. Whether these should instead report `01004` with a
     truncated value, closer to msodbcsql, is open and untracked.
-  - A `varbinary` / `image` column bound `SQL_C_DEFAULT` resolves to
-    `SQL_C_BINARY` (`describe_col.rs` → `SQL_VARBINARY` / `SQL_LONGVARBINARY`,
-    then `resolve_default_c_type`), which bound delivery does not implement yet
-    (AB#47239), so it fails per row with `HYC00`. msodbcsql resolves identically
-    and delivers the bytes. Pre-existing for an explicit `SQL_C_BINARY` bind;
-    deferred resolution makes it reachable without the application naming the C
-    type.
+  - A `varbinary` / `image` / CLR UDT column bound `SQL_C_DEFAULT` resolves to
+    `SQL_C_BINARY` (`describe_col.rs` → `SQL_VARBINARY` / `SQL_LONGVARBINARY` /
+    `SQL_SS_UDT`, then `resolve_default_c_type`), which bound delivery does not
+    implement yet (AB#47239), so it fails per row with `HYC00`. msodbcsql
+    resolves identically and delivers the bytes. Pre-existing for an explicit
+    `SQL_C_BINARY` bind; deferred resolution makes it reachable without the
+    application naming the C type.
   - `SQL_C_CHAR` is **UTF-8** in both directions; the driver never reads or
     writes the client code page. msodbcsql uses the client code page -
     `dwClientCodePage = SystemLocale::Singleton().AnsiCP()`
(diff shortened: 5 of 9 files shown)


mssql-odbc/src/api/describe_col.rs
```diff
@@ -227,7 +227,8 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
         // mssql-python keys its sql_variant handling off this exact type, so
         // reporting the column as character data hides the variant entirely.
         TdsDataType::SsVariant => SQL_SS_VARIANT,
-        TdsDataType::Vector | TdsDataType::Udt => SQL_VARCHAR,
+        TdsDataType::Udt => SQL_SS_UDT,
```
Theekshna, line 241: Correct, and verified against msodbcsql: a result column with `fSqlType == SQL_UDT_MAPPED` is reported as `SQL_SS_UDT` in both `SQLDescribeCol` (`sqlcdesc.cpp:7736`) and `SQL_DESC_TYPE`/`SQL_DESC_CONCISE_TYPE` (`sqlcdesc.cpp:6292`). It also resolves an internal inconsistency, since `describe_param.rs:534` already reported `Udt` as `SQL_SS_UDT`.

One side effect worth a line in the PR description, since it isn't purely metadata reporting: `odbc_sql_type` feeds `SQL_C_DEFAULT` resolution on the fetch path (`fetch_scroll.rs:899`), so a UDT column bound with `SQL_C_DEFAULT` now resolves to `SQL_C_BINARY` instead of `SQL_C_CHAR` (`type_rules.rs:329`).

I traced it and the observable outcome is unchanged — `deliver_bound_plp`'s compatibility matrix (`fetch_scroll.rs:1758`) rejects both `SQL_C_CHAR` and `SQL_C_BINARY` against `PlpEncoding::Binary`, so the fetch is HYC00 before and after (AB#47239). So it's benign, but it's benign by accident rather than by construction, and the description currently reads as though the fetch path isn't touched at all.

mssql-odbc/src/api/describe_col.rs
```diff
@@ -357,6 +353,17 @@ mod tests {
...
 
+    #[test]
+    fn udt_column_reports_sql_ss_udt() {
+        let mut metadata = int_columns(1);
+        let meta = metadata.first_mut().unwrap();
+        meta.data_type = TdsDataType::Udt;
+        meta.type_info.tds_type = TdsDataType::Udt;
+        meta.type_info.length = usize::from(u16::MAX);
+
+        assert_eq!(odbc_sql_type(meta), SQL_SS_UDT);
```
Theekshna: This test constructs metadata that no real UDT column can have, and asserts less than it looks like it does.

- `int_columns` builds `TypeInfo::fixed_len(Int4)`, and nothing here replaces `type_info_variant`. The result is a `Udt` column with `is_plp() == false`, which the decoder explicitly rejects as a protocol error (`decoder.rs:3090`, "UDT column metadata is not partially-length-prefixed").
- `meta.type_info.tds_type` and `meta.type_info.length` are dead stores. `odbc_sql_type` matches on `meta.data_type` alone, so neither line affects the assertion.

Building the metadata with `TypeInfo::partial_len` (the way `mixed_lob_columns` does for `nvarchar(max)`) would make it a realistic UDT column, and would let the test also assert `column_size(meta)` and `decimal_digits(meta)` — the other two values `SQLDescribeCol` returns for the column, and currently untested anywhere. That's also where the `ColumnSize` question in my e2e comment lives, so a realistic fixture here would pin it at the unit level too.

A helper alongside `int_columns` / `mixed_lob_columns` in `test_client_support` seems like the natural home.

mssql-odbc/tests/e2e/tests/describe_col_test.cpp
```diff
@@ -227,3 +232,19 @@ TEST_F(DescribeColLiveTest, NVarCharColumnSizeIsCharCount) {
...
+    ExecDirect("SELECT CAST(NULL AS geography) AS geography_col, "
+               "CAST(NULL AS geometry) AS geometry_col, "
+               "CAST(NULL AS hierarchyid) AS hierarchyid_col");
+
+    for (SQLUSMALLINT column : {SQLUSMALLINT{1}, SQLUSMALLINT{2}, SQLUSMALLINT{3}}) {
+        SQLSMALLINT dataType = 0;
+        SQLRETURN rc = SQLDescribeCol(
+            stmt_, column, nullptr, 0, nullptr, &dataType, nullptr, nullptr, nullptr);
+        ASSERT_SQL_OK(rc, SQL_HANDLE_STMT, stmt_);
+        EXPECT_EQ(SQL_SS_UDT, dataType) << "column=" << column;
```
Theekshna: This asserts only `dataType`; `colSize`, `decDigits` and `nullable` are all `nullptr`. Since this suite establishes parity by comparing the same assertions across both drivers under `parity_report.py`, the "zero parity divergences" claim in the description currently covers the type code and nothing else.

That matters here because `ColumnSize` looks like a genuine divergence. In msodbcsql, `FIsVarMaxType` (`tds/TdsParser.h:1941`) returns TRUE only for `SQLXML`, or for `(bTDSType & 0xb1) == 0xa1 && cbMaxLen == VARMAX_INDICATOR`. `SQLUDT` is `0xF0`, and `0xF0 & 0xb1 == 0xb0`, so a UDT column never takes the varmax path. `SQLDescribeCol` then falls to the else branch (`sqlcdesc.cpp:~7793`) and returns `lpcolinfo->cbPrecision`, the UDT max byte size from COLMETADATA.

On our side `column_size` short-circuits on `is_plp()` and returns 0, and UDT TYPE_INFO always parses to `PartialLen` (`sqldatatypes.rs:943`), so every UDT column reports 0.

I did not trace what `cbPrecision` holds when the wire `MAX_BYTE_SIZE` is `0xFFFF`, so I can't tell you the expected number. Adding `colSize` to this test settles it in a single parity run: either it matches and the parity claim gets stronger, or it doesn't and we document the divergence.

(To be fair this is arguably pre-existing — `describe_param.rs:534` maps `Udt` to size 0 and cites `describe_col`'s PLP convention — but this PR is what makes UDT metadata something consumers will actually read.)

Theekshna (commented):
The mapping change is correct, but I have one likely parity divergence to settle and two test gaps.

**Parity checks I ran against msodbcsql (`Sql/Ntdbms/sqlncli`)**

- Confirmed match: a result column with `fSqlType == SQL_UDT_MAPPED` is reported as `SQL_SS_UDT` by both `SQLDescribeCol` (`sqlcdesc.cpp:7736`) and `SQL_DESC_TYPE`/`SQL_DESC_CONCISE_TYPE` (`sqlcdesc.cpp:6292`).
- Confirmed match, no divergence: `SQL_DESC_TYPE_NAME`. msodbcsql forces `colusertype = 0` for CLR UDTs (`sqlcdesc.cpp:8823`) and falls through to `g_rgServerTypeInfo[SQLUDT] = L"udt"`, which is exactly what `col_attribute.rs:618` already returns.
- Confirmed match: default C type. msodbcsql's `Sql2CDefault` gives `SQL_C_BINARY` for UDT, and `type_rules.rs:329` already agrees.
- Likely divergence: `ColumnSize`. See the comment on the e2e test.

**Coverage gap not visible in this diff**

`odbc_sql_type` also drives `SQLColAttributeW` (`col_attribute.rs:274`/`283`) and the IRD (`ird.rs:69`), so `SQL_DESC_TYPE` and `SQL_DESC_CONCISE_TYPE` for a UDT column now return -151 too. Neither entry point got a test. `col_attribute_test.cpp:413` already has the `SQL_SS_VARIANT` precedent to copy, and it would cover the actual behaviour change through a second door.

**Follow-up work item (not this PR)**

Advertising `SQL_SS_UDT` is precisely what invites a consumer to ask for `SQL_CA_SS_UDT_CATALOG_NAME` / `SCHEMA_NAME` / `TYPE_NAME` / `ASSEMBLY_TYPE_NAME` (`SQL_CA_SS_BASE+18..21` = 1218-1221), which msodbcsql supports and `col_attribute.rs` does not — it implements only `SQL_CA_SS_VARIANT_TYPE` (1215). The TDS layer already parses all four names into `UdtInfoInColMetadata`, so this should be cheap. Worth filing and referencing from the PR description.

I could not run the test suite locally: my toolchain fails codegen on `crc32fast` (`E9012 llvm.x86.pclmulqdq.256`), unrelated to this change. Relying on your reported `cargo btest` run and the ADO validation.

mssql-odbc/src/api/describe_col.rs
```diff
@@ -227,7 +227,8 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
         // mssql-python keys its sql_variant handling off this exact type, so
         // reporting the column as character data hides the variant entirely.
         TdsDataType::SsVariant => SQL_SS_VARIANT,
-        TdsDataType::Vector | TdsDataType::Udt => SQL_VARCHAR,
+        TdsDataType::Udt => SQL_SS_UDT,
+        TdsDataType::Vector => SQL_VARCHAR,
```
shiwanigupta0809, line 244: **Suggestion** — `Vector` has the same defect this PR is fixing for `Udt`, and the driver already has everything needed to fix it. The description says the change "preserv[es] the existing vector mapping", so this looks deliberate; I would like to understand why, because from here it reads like the other half of the same bug.

The case for them being the same shape:

| | `Udt` (fixed here) | `Vector` (left as-is) |
|---|---|---|
| msodbcsql answers | `SQL_SS_UDT` via `SQL_UDT_MAPPED` (`sqlsrv.h:191`) | `SQL_SS_VECTOR` (`-156`, `msodbcsql.h:335`) via `SQL_VECTOR_MAPPED` — same remap pattern |
| constant exists in this driver | `odbc_types.rs:386` | `odbc_types.rs:389` |
| `classify_parameter_sql_type` | Supported (`type_rules.rs:275`) | Supported (`type_rules.rs:278`) |
| `resolve_default_c_type` | `=> SQL_C_BINARY` (`:329`) | `=> SQL_C_SS_VECTOR` (`:331`), with a unit test at `:421-422` |
| `SQL_DESC_TYPE_NAME` | `"udt"` (`col_attribute.rs:618`) | `"vector"` (`col_attribute.rs:616`) |
| `odbc_sql_type` | `SQL_SS_UDT` after this PR | **`SQL_VARCHAR`** |

Two consequences follow from that last row, both of which are the reasons given for the UDT half:

- **`SQL_SS_VECTOR` is unreachable as a result-column type.** `resolve_default_c_type(SQL_SS_VECTOR) => SQL_C_SS_VECTOR` is only reachable through `SQLBindParameter`, where the application names the SQL type. On the fetch side, `resolve_default_bindings` reads its SQL type from `odbc_sql_type`, which never returns `SQL_SS_VECTOR` — so a `SQL_C_DEFAULT`-bound vector column resolves through `SQL_VARCHAR` to `SQL_C_CHAR`. The `SQL_C_SS_VECTOR` arm and its test pass, but nothing on the fetch path can reach them.
- **The driver contradicts itself.** `SQL_DESC_TYPE_NAME` says `"vector"` while `SQL_DESC_CONCISE_TYPE` says `SQL_VARCHAR` (12) for the same column. That is precisely the disagreement PR #489 is fixing for `numeric`, and the mssql-python failure mode in this PR's own description — a consumer keying on the concise type and selecting a character binding for a non-character value.

If vector delivery is not ready and reporting `SQL_SS_VECTOR` would expose an unimplemented path, that is a perfectly good reason to hold it — but then it is worth a line here saying so and a work item, the way the `varbinary` → `SQL_C_BINARY` gap is handled at `fetch_scroll.rs:713-718`. Right now the split reads as though `Vector` was examined and found correct.

One thing that makes me suspect it is closer to ready than it looks: `column_value_to_text` has a real `ColumnValues::Vector(vec) => Ok(format_vector(vec))` arm (`get_data.rs:2458`), so vectors do render today — unlike UDT bytes, which have no arm at all. So the two are not in the same state, and whichever way you go, the reason is worth recording.

shiwanigupta0809 (commented):
Read the diff, traced `odbc_sql_type`'s consumers, and checked the mapping against msodbcsql's own tables. Only Copilot has reviewed so far, and it generated no comments.

**Blocking:** none. **Suggestions:** 1 (inline).

### Verified

**The msodbcsql parity claim holds.** `rgbSRVTYPE2SQLTYPE` in `sqlcmisc.cpp` maps `SQLUDT` to `SQL_UDT_MAPPED`, and `sqlsrv.h:191` defines that as `(SQL_VARCHAR+2)` with the comment `// Mapped to SQL_SS_UDT` — the same internal-remap pattern `SQL_SS_VARIANT` uses. So `-151` is genuinely what msodbcsql answers, and the new e2e test carrying no `SKIP_IF_COMPARING_MSODBCSQL()` is a real assertion rather than a self-fulfilling one.

**"Limited to metadata reporting" is accurate — I checked, because it easily might not have been.** `odbc_sql_type` feeds `SQL_C_DEFAULT` resolution (`fetch_scroll.rs:896-900` → `resolve_default_bindings` → `resolve_default_c_type`), and `type_rules.rs:329` maps `SQL_SS_UDT => SQL_C_BINARY` where the old `SQL_VARCHAR` mapped to `SQL_C_CHAR`. So a `SQL_C_DEFAULT`-bound UDT column does silently change which C type it resolves to.

It makes no observable difference, for two reasons that both had to hold:

- A UDT column is PLP, so it lands in `deliver_bound_plp`. Its `compatible` gate (`fetch_scroll.rs:1758-1764`) admits only the four character/encoding pairings, and `PlpEncoding::Binary` is in none of them — so `SQL_C_CHAR` was already refused there before this change, exactly as `SQL_C_BINARY` is now. Both answer `RowIssue::Unsupported`.
- Even off the PLP path, `column_value_to_text` (`get_data.rs:2428-2469`) has no `ColumnValues::Bytes` arm, so the old `SQL_C_CHAR` resolution fell to `_ => Err(TextError::Unsupported)` anyway.

Worth recording somewhere, though: the `resolve_default_bindings` doc comment at `fetch_scroll.rs:713-718` enumerates which columns resolve to an undeliverable `SQL_C_BINARY`, and lists only `varbinary` / `image`. UDT now joins that set. The behaviour is unchanged, but that comment is now incomplete.

### Suggestion

The line this PR splits has a second half with the same bug. Left inline.

shiwanigupta0809:
Reviewed — posted as a review with one inline suggestion on the \Vector\ arm: https://github.com/microsoft/mssql-rs/pull/490#pullrequestreview-5100062308

Short version: the msodbcsql parity holds (\SQL_UDT_MAPPED\ is \SQL_VARCHAR+2\, remapped out to \SQL_SS_UDT\), and I verified that "limited to metadata reporting" is true even though \odbc_sql_type\ does feed \SQL_C_DEFAULT\ resolution — the PLP bound path refused both the old and new target. The suggestion is that \Vector\, the other half of the line being split, has the same defect with all the plumbing already in place.

David-Engel (commented):
Automated review (unattended run — these findings were not checked by a human before posting). Reviewed the full diff at `099ccc5`, traced every consumer of `odbc_sql_type`, and read the three existing reviews on this SHA before writing anything.

**Blocking: 0. Suggestions: 0. Nits: 0.** Nothing new to add beyond the threads already open from @Theekshna and @shiwanigupta0809 — I deliberately did not re-file those.

### What I verified that the prior reviews could not

@Theekshna noted a local toolchain failure and deferred to the reported `cargo btest`. I have a working build, so I spent the run on the one check this repo's review skill calls highest-yield: does the new test fail when the change is reverted?

```
# baseline
PASS [0.035s] mssqlodbc api::describe_col::tests::udt_column_reports_sql_ss_udt

# mutate describe_col.rs:230 back to `TdsDataType::Udt => SQL_VARCHAR`
FAIL [0.039s] mssqlodbc api::describe_col::tests::udt_column_reports_sql_ss_udt
  assertion `left == right` failed
    left: 12
   right: -151

# full crate, unmutated
Summary [18.020s] 1287 tests run: 1287 passed, 1 skipped
```

So `udt_column_reports_sql_ss_udt` is not vacuous — it is pinned to the exact constant this PR changes. That is orthogonal to @Theekshna's point about the fixture being unrealistic (`int_columns` leaves `type_info_variant` as `fixed_len(Int4)`, so `is_plp()` is false for a column the decoder requires to be PLP). Both are true: the test guards the mapping, and it would guard more if it were built from `TypeInfo::partial_len`. Worth doing, not worth blocking on.

### Derived attributes: checked, no collateral change

The one class of regression I looked for and did not find: attributes computed *from* the reported SQL type rather than from `data_type`. `column_size`, `decimal_digits`, `display_size`, `desc_length`, `binary_precision`, `octet_length`, `type_name` and `searchable` all match on `meta.data_type` directly, so none of them shift when the type code moves from `12` to `-151`. `ird.rs:69` and `col_attribute.rs:274`/`283` do carry the new value through to `SQL_DESC_CONCISE_TYPE`/`SQL_DESC_TYPE` — correctly, and matching the `sqlcdesc.cpp:6292` behavior @Theekshna cited — and `datetime_interval_code_for(-151)` stays `0` since UDT is not in the timestamp family. The change is contained to the type code, as the description says.

Also consistent with `describe_param.rs:534`, which has reported `Udt` as `SQL_SS_UDT` all along, so this closes an internal divergence rather than opening one.

### CI

`gh pr checks 490`: no failures. Windows/Linux/macOS/ARM builds, both cross-repo `mssql-python` jobs, Kerberos, CodeQL and `check` all green. `Test MacOS`, `coverage-report` and the overall validation roll-up were still pending at review time.

### Label

Left unchanged. The required validation roll-up has not finished on this SHA, and there are open reviewer threads (the `Vector` arm question and the `ColumnSize` parity question) that are the author's call rather than mine.

mssql-odbc/src/api/describe_col.rs
saurabh500: Kept Vector on `SQL_VARCHAR` deliberately and documented why in code and the PR description: changing it now would resolve `SQL_C_DEFAULT` to unsupported `SQL_C_SS_VECTOR` result delivery, regressing the working text path to `HYC00`. The vendor type should land with dedicated vector result-delivery coverage rather than in this UDT fix.

> [!NOTE]

mssql-odbc/tests/e2e/tests/describe_col_test.cpp
saurabh500: Extended the live test to assert type, size, scale, and nullability. A retail msodbcsql 18.6 probe established sizes 0/0/892 for geography/geometry/hierarchyid; the Rust driver now matches those values, and the full 38-binary parity run has zero divergences.

> [!NOTE]

mssql-odbc/src/api/describe_col.rs
saurabh500: Updated the PR description and `resolve_default_bindings` documentation to state that `SQL_C_DEFAULT` now resolves UDTs to `SQL_C_BINARY`, while observable bound delivery remains `HYC00` until AB#47239 is implemented.

> [!NOTE]

mssql-odbc/src/api/describe_col.rs
saurabh500: Replaced the mutated `Int4` fixture with a shared `TypeInfo::partial_len(Udt, ...)` helper. The unit test now verifies PLP shape, `SQL_SS_UDT`, scale, and both unbounded and 892-byte column sizes.

> [!NOTE]

Theekshna (commented):
Unattended sweep review. No new findings — the latest commit addresses every open item from prior review threads: the `ColumnSize` parity divergence for CLR UDT columns (now matches msodbcsql's `cbPrecision`-fallback behavior, measured 0/0/892 for geography/geometry/hierarchyid against retail msodbcsql18 with zero divergences across the 38-binary parity suite), the unrealistic unit-test fixture (now built from a real PLP `udt_column` helper), and the undocumented `SQL_C_DEFAULT`→`SQL_C_BINARY` side effect (now recorded in both the PR description and the `resolve_default_bindings` doc comment). The `Vector` arm is deliberately left on `SQL_VARCHAR` with an inline comment explaining why (avoids regressing a working text-fetch path to HYC00); this is consistent with the driver's own state and not a bug. CI: all completed jobs green; several still pending at review time. No AI slop, description/checklist current.

David-Engel (commented):

## Summary

Reviewed the full diff at `83255b1` against `main`, then read every prior review body, inline thread and PR-level comment before writing anything. The two open items from the earlier round are genuinely closed: the `SQL_C_DEFAULT` → `SQL_C_BINARY` side effect is now stated in both the description and `fetch_scroll.rs`, and @Theekshna's `ColumnSize` parity question is answered by the retail 18.6 probe plus the new `colSize` assertions. The `Vector` arm being held back is documented in code and in the description, with the reason (`SQL_C_SS_VECTOR` result delivery is unimplemented) that makes it a deliberate split rather than half a fix. I did not re-file any of those.

The mapping itself is correct and the fixture rework is a real improvement — `udt_column` builds `TypeInfo::partial_len(Udt, ...)`, so the unit test now exercises a column shape the decoder would actually accept, and it asserts the two values (`column_size`, `decimal_digits`) that were previously untested anywhere. `test_client_support` stays behind `#[cfg(any(test, feature = "test-util"))]`, so the new `pub fn` is not a public surface addition.

One new observation below, about the `SQLColAttribute` fields that are *derived* from `column_size` and did not move with it.

**Blocking: 0. Suggestions: 1. Nits: 0.**

## Verification

- `cargo nextest run -p mssqlodbc --lib -E 'test(udt)'` — 2 passed (`describe_col::tests::udt_column_reports_sql_ss_udt_metadata`, `get_type_info::tests::udt_data_type_returns_hyc00`).
- Traced `column_size`'s consumers rather than reading them: `desc_length`, `display_size`, `octet_length` and `precision` in `col_attribute.rs`, plus `ird.rs:73-74`.
- Ran a temporary probe against `udt_column(892)` and `udt_column(u16::MAX)` through the real `SQLColAttribute` path (added to `col_attribute.rs`'s test module, then reverted — the worktree is clean). Results below.
- `gh pr checks 490`: no failures. All ADO build/test legs, both cross-repo `mssql-python` jobs, Kerberos, CodeQL and `check` are green; `coverage-report`, `Merge Coverage` and the validation roll-up were still pending at review time.

## Findings

### Suggestion — the bounded-UDT size now disagrees with itself across `SQLColAttribute` fields

`column_size` returning `MAX_BYTE_SIZE` for a bounded UDT is right for `SQLDescribeCol` — the 892 for `hierarchyid` matches the retail probe you cited. But three other attributes read through the same helper and only two of them followed, because `display_size` and `octet_length` still short-circuit on `is_plp()` (`col_attribute.rs:327` and `:399`), which is `true` for every UDT.

Measured through `SQLColAttribute`:

```
udt_column(892):    SQL_DESC_LENGTH=892  SQL_DESC_PRECISION=892  SQL_DESC_OCTET_LENGTH=0  SQL_DESC_DISPLAY_SIZE=0
udt_column(0xFFFF): SQL_DESC_LENGTH=0    SQL_DESC_PRECISION=0    SQL_DESC_OCTET_LENGTH=0  SQL_DESC_DISPLAY_SIZE=0
```

Before this PR all four were `0` for both, so they were at least self-consistent. Now a `hierarchyid` column reports a length of 892 and a transfer size of 0 — and `ird.rs:73-74` carries exactly that pair into the IRD.

I am asking rather than asserting, because I could not check msodbcsql's `SQL_DESC_OCTET_LENGTH` / `SQL_DESC_DISPLAY_SIZE` / `SQL_DESC_PRECISION` for a UDT column and the ODBC spec is not the contract here. Two things make it worth settling before this merges:

1. `ClrUdtTypeFields` asserts only `SQL_DESC_TYPE` and `SQL_DESC_CONCISE_TYPE`, so the size-bearing attributes are outside the parity comparison for UDT columns. The "zero parity divergences" result therefore does not cover them — the same gap @Theekshna raised against the `SQLDescribeCol` test, which you closed by adding `colSize`. Adding `SQL_DESC_LENGTH`, `SQL_DESC_OCTET_LENGTH`, `SQL_DESC_DISPLAY_SIZE` and `SQL_DESC_PRECISION` to that loop settles all four in one parity run, the same way the `colSize` addition settled `ColumnSize`.
2. If they *should* move, `SQL_DESC_OCTET_LENGTH` is the one that matters operationally: it is what an application sizes a `SQL_C_BINARY` buffer from once AB#47239 lands, and 0 is the "unbounded, use `SQLGetData`" answer for a column that is in fact capped at 892 bytes.

If the probe shows msodbcsql reports 0 for those and 892 only for `ColumnSize`/`SQL_DESC_LENGTH`, then the current behavior is correct as written and a one-line note next to the `column_size` early return saying so would keep the next reader from "fixing" it.

Nothing else in the diff looks off to me. To be explicit: this is not an approval, and the `Vector` question, the parity-run scope and anything above are the author's call — a human reviewer should still look.

mssql-odbc/src/api/describe_col.rs
```diff
@@ -238,12 +238,20 @@ pub(crate) fn odbc_sql_type(meta: &mssql_tds::query::metadata::ColumnMetadata) -
...
+        // character type keeps SQL_C_DEFAULT fetches on the working text path.
+        TdsDataType::Vector => SQL_VARCHAR,
         _ => SQL_UNKNOWN_TYPE,
     }
 }
 
 pub(crate) fn column_size(meta: &mssql_tds::query::metadata::ColumnMetadata) -> u64 {
+    // CLR UDT metadata is PLP even when MAX_BYTE_SIZE is bounded. Only 0xFFFF
+    // carries the unbounded convention used by geography and geometry.
+    if meta.data_type == TdsDataType::Udt && meta.type_info.length != usize::from(u16::MAX) {
```
David-Engel, line 252: This makes `SQLDescribeCol` and `SQLDescribeParam` disagree about the same UDT. `describe_param.rs:541-544` still returns `(SQL_SS_UDT, 0, 0)`, and its comment justifies the `0` as "the same 'unbounded' convention `describe_col::column_size` already uses for PLP" — which is precisely the convention this branch stops following for a bounded UDT.

Not a bug in this diff (I don't think `sp_describe_undeclared_parameters` surfaces a MAX_BYTE_SIZE for a UDT parameter anyway), but the cross-reference is now stale and points at code that no longer behaves that way. Either mirror the max-byte-size there or reword that comment, so the next person doesn't take it at face value.

mssql-odbc/src/api/col_attribute.rs
```diff
@@ -408,6 +408,9 @@ fn display_size(meta: &ColumnMetadata) -> SqlLen {
 /// representation, which for the temporal types is the C struct the driver
 /// hands back, not the TDS payload width.
 pub(super) fn octet_length(meta: &ColumnMetadata) -> SqlLen {
+    if meta.data_type == TdsDataType::Udt {
```
David-Engel, line 413: Worth a line here on why UDT gets an early return but deliberately does *not* get one in `display_size`. After this change a `hierarchyid` reports LENGTH / PRECISION / OCTET_LENGTH of 892 but DISPLAY_SIZE of 0, which reads like an oversight until you know it's intentional (an opaque UDT has no character rendering, so there's no display width to claim). The new unit test pins the asymmetry but not the reason, and this is exactly the kind of thing a later "consistency fix" would quietly break.

Separate nit on the same values: `desc_length` widens through `SqlLen` (i64), while `precision` bottoms out at `SqlSmallInt::try_from(column_size(..)).unwrap_or(SqlSmallInt::MAX)` on line 467. A MAX_BYTE_SIZE in `32768..=65534` would therefore report 32767 for PRECISION and the true value for LENGTH. Unreachable for a real CLR UDT — SQL Server caps `MaxByteSize` at 8000 — so purely theoretical, and not worth changing unless you're touching that fallback for another reason.

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot on behalf of @David-Engel. This is **not an approval** and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back where I've got it wrong.

Reviewed the full diff at `d3ba08e`, traced every consumer of `odbc_sql_type` / `column_size` / `desc_length` / `octet_length`, and read the delivery paths in `get_data.rs` and `fetch_scroll.rs`. **No blocking findings.** Two non-blocking notes inline, plus one correction and one follow-up below.

**What I verified**

- The mapping is right: `SQL_SS_UDT` is `-151` in `odbc_types.rs`, matching the `#define` guards added to both e2e files. `TdsDataType::Udt` already decodes to `ColumnValues::Bytes` (`decoder.rs:3090`), so nothing downstream was keying off the old `SQL_VARCHAR`. Splitting `Vector` out of the shared arm with an explicit rationale is a good call — it keeps that type on the working text path.
- Size fields behave as the tests claim. `cargo test -p mssqlodbc --lib -- udt` passes locally: bounded (892) reports LENGTH / PRECISION / OCTET_LENGTH 892 and column size 892; unbounded (`0xFFFF`) reports 0 everywhere, and `is_plp()` holds in both cases.
- CI is green at this head. All required checks pass; only `Test MacOS` and `coverage-report` are still pending, neither of which is affected by a metadata-only change. The new e2e cases run inside the passing Build Windows / Build Linux stages, so the `892` / `0` expectations have been exercised against a real server rather than only asserted.

**One correction to the reworded notes**

The `fetch_scroll.rs` doc comment and the matching bullet in `.github/instructions/mssql-odbc.instructions.md` now fold CLR UDT into the `varbinary` / `image` sentence: "deferred resolution makes it reachable without the application naming the C type." That framing is accurate for `varbinary` / `image`, but overstated for UDT. `column_value_to_text` sends `ColumnValues::Bytes` to `TextError::Unsupported` (`get_data.rs:2483`), so the *old* `SQL_VARCHAR` → `SQL_C_CHAR` default already failed a UDT column with `HYC00` — in `deliver_bound` and in `SQLGetData` alike. The new `SQL_C_BINARY` default fails identically. Nothing became newly broken here, and no application regresses from this PR. Worth a word tweak so a future reader chasing AB#47239 doesn't go looking for a regression that isn't there.

**Follow-up worth filing: identifying *which* UDT**

Applications can now detect `SQL_SS_UDT`, and the natural next call is `SQL_DESC_TYPE_NAME` or `SQL_CA_SS_UDT_TYPE_NAME`. Today `type_name` returns the literal `"udt"` (`col_attribute.rs:633`) and `SQL_CA_SS_UDT_CATALOG_NAME` / `SCHEMA_NAME` / `TYPE_NAME` (1041–1043) aren't handled at all, so those come back `HY091`. The data is already on the wire and already parsed — `UdtInfoInColMetadata { db_name, schema_name, type_name, assembly_qualified_name }` in `sqldatatypes.rs`, currently carrying `#[allow(dead_code)]`. Reporting the real type name is what turns this correct type code into something an application can act on, and the plumbing is mostly there. Out of scope for this PR, but a cheap and high-value next step.

mssql-odbc/src/api/col_attribute.rs
saurabh500: Documented the deliberate asymmetry in `32180881`: bounded UDTs use their byte limit for binary transfer, while display size remains 0 because an opaque UDT has no text rendering. I left the theoretical precision clamp unchanged since SQL Server caps real CLR UDT `MaxByteSize` at 8000.

> [!NOTE]

mssql-odbc/src/api/describe_col.rs
saurabh500: Reworded this in `32180881`: parameter discovery does not carry the CLR `MAX_BYTE_SIZE` from `COLMETADATA`, so its UDT size remains 0 without claiming that it mirrors result-column sizing.

> [!NOTE]

David-Engel (commented):

## Summary

Reviewed the full diff at `3218088` against `main`, and read every prior review body, inline thread and PR-level comment before writing. **Blocking: 0. Suggestions: 0. Nits: 0.** Nothing new to file — the two notes I left on `d3ba08e` are both closed by this commit, and I deliberately did not re-raise anything from @Theekshna's or @shiwanigupta0809's threads.

## What changed since the last round

`3218088` is comment-only, and it lands exactly where the open notes were:

- `describe_param.rs:541` no longer cites `describe_col::column_size`'s PLP convention as its justification for `0`. I checked the replacement claim rather than taking it: `describe_tds_type` is fed `suggested_tds_length` out of `sp_describe_undeclared_parameters` (`describe_param.rs:426`), and the RPC-side UDT metadata is `UdtInfoInRpc` (`sqldatatypes.rs:690`), which carries no `max_byte_size` at all — only `UdtInfoInColMetadata` does. The new wording is accurate, and the stale cross-reference is gone.
- `col_attribute.rs:411` now states why UDT gets an early return in `octet_length` but not in `display_size`. That is the asymmetry a later "consistency fix" would have quietly reverted.

## Verification

- **Mutation check on the newest substantive branch.** Deleted the `TdsDataType::Udt => desc_length(meta)` early return from `octet_length` and re-ran: `udt_size_attributes_match_bounded_and_unbounded_metadata` fails with `left: 0, right: 892`. The test is pinned to the behavior it names, not vacuous. Worktree restored afterwards.
- `cargo nextest run -p mssqlodbc --lib -E 'test(udt)'` — 3 passed (`describe_col::tests::udt_column_reports_sql_ss_udt_metadata`, `col_attribute::tests::udt_size_attributes_match_bounded_and_unbounded_metadata`, `get_type_info::tests::udt_data_type_returns_hyc00`).
- **Traced the `MAX_BYTE_SIZE` provenance end to end.** `read_type_info`'s `PartialLengthType::Udt` arm (`sqldatatypes.rs:925-956`) reads the `USHORT` and stores it in *both* `TypeInfo.length` and `UdtInfoInColMetadata.max_byte_size`, so `column_size`'s `meta.type_info.length != usize::from(u16::MAX)` guard reads the right field with the right sentinel. `udt_column` reproduces that shape via `TypeInfo::partial_len`, so the fixture matches what the parser actually produces.
- **Re-derived the size fan-out** rather than trusting the earlier round: `desc_length` → `binary_precision` (`None` for UDT) → `column_size`; `precision` → no temporal match, no binary precision, `get_precision()` `None` → `column_size`; `octet_length` → the new UDT arm → `desc_length`; `display_size` → `is_plp()` → 0. Bounded reports 892/892/892/0, unbounded 0/0/0/0, which is what both the unit test and `ClrUdtDescriptorFields` assert.
- **`SQLGetData` is untouched by the mapping change.** `get_data.rs` has no `SQL_C_DEFAULT` arm at all, so the type code never reaches it; the `SQL_C_DEFAULT` → `SQL_C_BINARY` shift is confined to `resolve_default_bindings` on the bound-fetch path, exactly as the reworded doc comment now says.
- `gh pr checks 490`: **all green**, including `coverage-report`, `Merge Coverage`, `Test MacOS` and the ADO validation roll-up that were still pending last round. Diff coverage 97%, the one uncovered line being the `Vector => SQL_VARCHAR` arm that is deliberately unchanged.

## Notes, not findings

Both are already recorded in the threads above and need no action here — repeating only so the follow-up work is easy to find:

- `SQL_CA_SS_UDT_CATALOG_NAME` / `SCHEMA_NAME` / `TYPE_NAME` / `ASSEMBLY_TYPE_NAME` still return `HY091`, and `SQL_DESC_TYPE_NAME` still answers the literal `"udt"`. The data is parsed and sitting in `UdtInfoInColMetadata` behind `#[allow(dead_code)]`. Advertising `SQL_SS_UDT` is what makes an application ask for those, so it is the natural next work item.
- The `Vector` arm stays on `SQL_VARCHAR` with a stated reason. Holding it back is a judgement call I agree with, but it means `SQL_C_SS_VECTOR` remains unreachable from the fetch path.

To be explicit: this is not an approval, and a human reviewer should still look.

mssql-odbc/src/api/describe_param.rs
```diff
@@ -538,9 +538,9 @@ fn describe_tds_type(
         }
         TdsDataType::Image => (SQL_LONGVARBINARY, parameter_length(length, false)?, 0),
         TdsDataType::SsVariant => (SQL_SS_VARIANT, 8000, 0),
-        // Unbounded types report a size of 0, the same "unbounded" convention
-        // `describe_col::column_size` already uses for PLP. A table type has no
-        // meaningful column size at all.
+        // Parameter discovery does not include the CLR MAX_BYTE_SIZE carried by
+        // COLMETADATA, so UDT size remains 0. XML is unbounded; a table type has
+        // no meaningful column size at all.
         TdsDataType::Udt => (SQL_SS_UDT, 0, 0),
```
Theekshna, line 544: This comment isn't accurate, and it hides the same bug you just fixed for the result-set path.

msodbcsql *does* report the UDT max byte size from `SQLDescribeParam`:

1. `sqlcdesc.cpp:9263` reads `suggested_tds_length` into `FRS_Format.lMaxLen`.
2. `tds.h:1609` — `GETTER4S(USHORT, YukonUDTType, usMaxLen, lMaxLen)` — so `get_YukonUDTType_usMaxLen()` *is* `(USHORT)lMaxLen`.
3. `GetColInfoFromColFmt` is a template (`sqlctokn.cpp:748`) explicitly instantiated for `DESC_FIRST_RS_FORMAT` (`sqlctokn.cpp:1498`), so the parameter path runs the same `case SQLUDT` body → `cbPrecision = usMaxLen`.
4. `ParamInfoFromColInfo` (`sqlcdesc.cpp:8128`) memcpys `cbPrecision` into `param.cbColDef`, which is what `SQLDescribeParam` returns as `ParameterSizePtr`.

So `suggested_tds_length` is exactly the CLR MAX_BYTE_SIZE, and `describe_tds_type` already receives it as the `length` argument — it's just discarded here. Same shape of fix as `column_size`, mirroring the `USHORT` truncation so both `-1` and `65535` collapse to 0 like `VARMAX_INDICATOR`.

Fix it or file it, but the comment as written asserts a limitation that doesn't exist and will stop the next person from looking.

Theekshna (commented):
Re-reviewed the descriptor size commits against msodbcsql. The result-set side is correct: `SQL_DESC_LENGTH`, `SQL_DESC_PRECISION`, `SQL_DESC_OCTET_LENGTH` and `SQL_DESC_DISPLAY_SIZE` all match for both the bounded and `0xFFFF` cases, and `precision()` correctly falls through to `column_size` because `get_precision()` is `None` for `PartialLen`. The `SQL_C_DEFAULT` prose rewrites are accurate. My earlier `octet_length` finding is resolved.

One comment inline on `describe_param.rs`. Separately: did `SQL_CA_SS_UDT_*` (1218-1221) get a work item? The TDS layer already parses all four names into `UdtInfoInColMetadata`.

Caveat: I can't build locally (`crc32fast` codegen failure on this toolchain), so this is source reading only.

mssql-odbc/src/api/describe_param.rs
saurabh500: I measured this directly against the pinned retail msodbcsql 18.6.2.1: it returns `ParameterSize=0` for geography, geometry, and hierarchyid; changing hierarchyid to 892 produced a real parity failure. Commit `83561dbf` keeps the shipped behavior, corrects the rationale, and adds unit plus live comparison coverage for the distinction.

> [!NOTE]

David-Engel (commented):

## Summary

Reviewed the full diff at `83561db` against `main`, and read every prior review body, inline thread and PR-level comment before writing. **Blocking: 0. Suggestions: 0. Nits: 0.** Nothing new to file, and I did not re-raise anything from @Theekshna's or @shiwanigupta0809's threads.

`83561db` is the only change since my last pass, and it answers the one open thread — @Theekshna's `describe_param` finding, which traced msodbcsql source (`cbPrecision = usMaxLen` → `cbColDef`) to argue `SQLDescribeParam` should report the CLR `MAX_BYTE_SIZE`. Resolving that against a measurement of retail 18.6.2.1 rather than against the source trace is the right call for this crate: the shipping driver is the parity contract, and a source path that exists is not evidence that it is reached for this input. The commit keeps the shipped `0`, replaces the previous "parameter discovery does not carry MAX_BYTE_SIZE" rationale (which was inaccurate) with the measured one, and pins both halves — unit test on the parameter side, live `DescribesClrUdtMetadata` on the e2e side. Since neither UDT e2e test carries `SKIP_IF_COMPARING_MSODBCSQL`, the asymmetry between column size 892 and parameter size 0 is now asserted against both drivers on every parity run rather than resting on a one-off probe.

## Verification

- **Mutation check on the new test, using the exact change the thread proposed.** Rewrote `describe_param.rs:544` to `TdsDataType::Udt => (SQL_SS_UDT, parameter_length(length, false)?, 0)` and re-ran: `udt_parameter_size_matches_msodbcsql` fails with `parameter_size: 892` vs `0`. So the test is pinned to the decision it documents, not vacuous. Worktree restored afterwards.
- `cargo nextest run -p mssqlodbc --lib -E 'test(udt)'` — 4 passed (`describe_col::tests::udt_column_reports_sql_ss_udt_metadata`, `col_attribute::tests::udt_size_attributes_match_bounded_and_unbounded_metadata`, `describe_param::tests::udt_parameter_size_matches_msodbcsql`, `get_type_info::tests::udt_data_type_returns_hyc00`).
- **Re-traced the size fan-out** rather than trusting the earlier round: `column_size` (new UDT branch, `length != 0xFFFF`) → `desc_length` → `precision` (no temporal match, `binary_precision` `None`, `get_precision()` `None` for `PartialLen`) and → `octet_length` (new UDT arm); `display_size` still short-circuits on `is_plp()`. Bounded reports 892/892/892/0, unbounded 0/0/0/0 — what both the unit test and `ClrUdtDescriptorFields` assert.
- **Checked the one derived attribute nobody has mentioned.** `is_unsigned` (`col_attribute.rs:574`) tests the *ODBC* type rather than the TDS type, so it was worth confirming the `12` → `-151` move does not flip it. `SQL_SS_UDT` is not in the signed-numeric set and neither was `SQL_VARCHAR`, so `SQL_DESC_UNSIGNED` stays `SQL_TRUE`. Every other size- or name-bearing attribute matches on `meta.data_type` directly.
- **Confirmed the fixture still matches the parser.** `read_type_info`'s `PartialLengthType::Udt` arm (`sqldatatypes.rs:925-956`) writes the `USHORT` into both `TypeInfo.length` and `UdtInfoInColMetadata.max_byte_size`, so `column_size`'s `length != usize::from(u16::MAX)` guard reads the right field with the right sentinel, and `udt_column` reproduces that shape. An Always Encrypted column cannot reach the new branch either: `data_type` / `type_info` describe the ciphertext type there, so an encrypted UDT presents as `varbinary`.
- **IRD carry-through is metadata only.** `ird.rs:73-74` propagates the new length / octet length, and the only readers of those record fields are `SQLGetDescField` / `SQLGetDescRec` — nothing sizes a buffer from them.
- `gh pr checks 490`: all green, including both cross-repo `mssql-python` jobs, Kerberos, CodeQL and the ADO validation roll-up.

## Findings

None. This looks ready for human review — but to be explicit, this is not an approval, and a human reviewer should still look.

saurabh500:
@Theekshna @David-Engel I opened [AB#47844](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47844) as a child of [AB#46375](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46375) to track support for the `SQL_CA_SS_UDT_*` identity metadata.

Theekshna (approved):

PR #492, opened 2026-09-03
Honor TVP column precision and scale overrides

### LINKED ISSUE #364: TVP row serialization ignores column precision and scale overrides
### Describe the bug

**Description**

[TvpColumnDef] supports explicit [precision] and [scale] metadata for [Decimal], [Numeric], [Time], [DateTime2], and [DateTimeOffset]. These overrides are applied when writing TVP TYPE_INFO, but they are not applied when deriving [TdsTypeContext] for row serialization.
In [write_tvp_rows], context is derived only from [column.column_type]:
let (_, ctx) = column
    .column_type
    .to_column_value_and_context(db_collation);

For a value-free decimal template such as [SqlType::Decimal(None)], this produces a precision-38 context. If the column advertises [decimal(9,2)], row values are therefore serialized using the precision-38 length of 17 bytes instead of the expected 5 bytes.
Temporal scale overrides are similarly absent from the row context, although temporal values may currently carry their scale internally.

### Steps to reproduce

Below test scenarios need to be covered by TVP - 
**Suggested Tests**

Cover TVPs containing:

- Empty row sets
- NULL rows
- Non-NULL rows
- [decimal(9,2)] and numeric(9,2)
- [time(3)]
- datetime2(3)
- datetimeoffset(3)

### Expected behavior

**Expected Behavior**
TVP row serialization should derive its context from the complete [TvpColumnDef], including:

- [precision]for [Decimal] and [Numeric] 
- [scale] for numeric and temporal types 
- Correct decimal value length based on declared precision.

The row encoding must match the metadata emitted in TVP TYPE_INFO.

**Example:**
let mut column = TvpColumnDef::new(SqlType::Decimal(None));
column.precision = Some(9);
column.scale = Some(2);

Expected row encoding for [decimal(9,2)]:

- Length byte: 5
- One sign byte
- Four value bytes

Current row context uses precision 38:
- Length byte: 17
- One sign byte
- Sixteen value bytes

### Actual behavior

**Actual Behavior**
TVP metadata and row encoding can disagree:

- Metadata advertises the requested precision and scale.
- Row context is derived from the bare [SqlType] template.
- Decimal templates default to precision 38 and produce 17-byte values.

### Version

mssql-tds v0.1.0

### Affected crate

mssql-tds

### Environment

Ubuntu 24.04.3 LTS, x86_64

### Additional context

**Suggested Fix**

Add a TVP-specific context derivation path, for example:
let (_, mut ctx) = column
    .column_type
    .to_column_value_and_context(db_collation);

ctx.precision = column.precision.or(ctx.precision);
ctx.scale = column.scale.or(ctx.scale);

Ensure [max_size] and any temporal length calculations remain consistent with the overridden metadata.

**Suggested Tests**

Cover TVPs containing:

- Empty row sets
- NULL rows
- Non-NULL rows
- [decimal(9,2)] and numeric(9,2)
- [time(3)]
- datetime2(3)
- datetimeoffset(3)

Assert both emitted TYPE_INFO and row payload lengths/values.

**Impact**
Py-core cannot safely support TVP numeric and temporal metadata overrides until row serialization honors [TvpColumnDef]. A py-core-only fix would still emit incompatible non-NULL decimal row data.

saurabh500:
## Description

Align TVP row serialization with the declared column precision and scale:

- preserve the msodbcsql-compatible 17-byte decimal/numeric value format while losslessly rescaling values to the column scale
- serialize `time`, `datetime2`, and `datetimeoffset` rows at the column scale rather than the cell scale
- reject invalid metadata, precision overflow, and conversions that would discard fractional data
- retain the existing defaults (`decimal(1,0)` and temporal scale 7); values incompatible with omitted decimal precision/scale now fail locally with `UsageError` instead of a server-side arithmetic overflow
- add byte-level unit coverage and a live SQL Server round-trip regression test, including NULL values

## Related Issues

Closes #364

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes

### DIFF
--- a/mssql-tds/src/datatypes/sql_tvp.rs
+++ b/mssql-tds/src/datatypes/sql_tvp.rs
@@ -562,6 +562,7 @@ pub(crate) async fn write_tvp_rows(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::datatypes::column_values::{SqlDateTime2, SqlDateTimeOffset};
     use crate::io::packet_writer::PacketWriter;
     use crate::io::packet_writer::tests::MockNetworkWriter;
     use crate::message::messages::PacketType;
@@ -830,17 +831,38 @@ mod tests {
 
     #[test]
     fn test_validate_unrepresentable_temporal_value() {
-        let mut column = TvpColumnDef::new(SqlType::Time(None));
-        column.scale = Some(3);
-        let data = TvpTableData::new(
-            vec![column],
-            vec![vec![SqlType::Time(Some(SqlTime {
-                time_nanoseconds: 10_000_001,
-                scale: 7,
-            }))]],
-        );
+        let time = SqlTime {
+            time_nanoseconds: 10_000_001,
+            scale: 7,
+        };
+        let cases = [
+            (SqlType::Time(None), SqlType::Time(Some(time.clone()))),
+            (
+                SqlType::DateTime2(None),
+                SqlType::DateTime2(Some(SqlDateTime2 {
+                    days: 1,
+                    time: time.clone(),
+                })),
+            ),
+            (
+                SqlType::DateTimeOffset(None),
+                SqlType::DateTimeOffset(Some(SqlDateTimeOffset {
+                    datetime2: SqlDateTime2 {
+                        days: 1,
+                        time: time.clone(),
+                    },
+                    offset: 60,
+                })),
+            ),
+        ];
 
-        assert!(matches!(data.validate(), Err(Error::UsageError(_))));
+        for (column_type, value) in cases {
+            let mut column = TvpColumnDef::new(column_type);
+            column.scale = Some(3);
+            let data = TvpTableData::new(vec![column], vec![vec![value]]);
+
+            assert!(matches!(data.validate(), Err(Error::UsageError(_))));
+        }
     }
 
     #[test]
@@ -866,6 +888,59 @@ mod tests {
             )))]],
         );
         assert!(matches!(data.validate(), Err(Error::UsageError(_))));
+    }
+
+    #[test]
+    fn test_apply_decimal_metadata_rejects_scale_factor_overflow() {
+        let mut value = DecimalParts::new(true, 38, u8::MAX, 1);
+
+        let error = apply_decimal_metadata(&mut value, 38, 0).unwrap_err();
+
+        assert!(matches!(
+            error,
+            Error::UsageError(message)
+                if message == "TVP decimal/numeric value scale 255 cannot be converted to scale 0"
+        ));
+    }
+
+    #[test]
+    fn test_apply_decimal_metadata_rejects_magnitude_overflow() {
+        let mut value = DecimalParts::new(true, 38, 0, u128::MAX);
+
+        let error = apply_decimal_metadata(&mut value, 38, 1).unwrap_err();
+
+        assert!(matches!(
+            error,
+            Error::UsageError(message)
+                if message == "TVP decimal/numeric value overflows precision 38 at scale 1"
+        ));
+    }
+
+    #[test]
+    fn test_apply_decimal_metadata_preserves_zero() {
+        let mut value = DecimalParts::new(false, 4, 2, 0);
+
+        apply_decimal_metadata(&mut value, 4, 4).unwrap();
+
+        assert_eq!(value, DecimalParts::new(false, 4, 4, 0));
+    }
+
+    #[test]
+    fn test_apply_temporal_scale_rejects_invalid_scale() {
+        let mut value = ColumnValues::Time(SqlTime {
+            time_nanoseconds: 0,
+            scale: DEFAULT_VARTIME_SCALE,
+        });
+        let invalid_scale = DEFAULT_VARTIME_SCALE + 1;
+
+        let error = apply_temporal_scale(&mut value, invalid_scale).unwrap_err();
+
+        assert!(matches!(
+            error,
+            Error::UsageError(message)
+                if message
+                    == format!("TVP temporal value cannot be represented at scale {invalid_scale}")
+        ));
     }
 
     #[tokio::test]
--- a/mssql-tds/src/datatypes/sql_tvp.rs
+++ b/mssql-tds/src/datatypes/sql_tvp.rs
@@ -234,7 +234,7 @@ fn apply_decimal_metadata(value: &mut DecimalParts, precision: u8, scale: u8) ->
         let scale_factor = 10_u128.pow(u32::from(scale - value.scale));
         magnitude = magnitude.checked_mul(scale_factor).ok_or_else(|| {
             Error::UsageError(format!(
-                "TVP decimal/numeric value overflows precision {precision} at scale {scale}"
+                "TVP decimal/numeric value magnitude is too large to rescale to scale {scale}"
             ))
         })?;
     }
@@ -278,6 +278,8 @@ fn temporal_value_fits_scale(value: &SqlTime, scale: u8) -> bool {
         return false;
     }
 
+    // Despite the name, `time_nanoseconds` counts 100-ns ticks (scale 7), so a
+    // value survives scale N only when it is a multiple of 10^(7-N) ticks.
     let scale_factor = 10_u64.pow(u32::from(DEFAULT_VARTIME_SCALE - scale));
     value.time_nanoseconds.is_multiple_of(scale_factor)
 }
@@ -390,6 +392,9 @@ impl TvpTableData {
                 if let Some((precision, scale)) = column.decimal_metadata()
                     && let SqlType::Decimal(Some(value)) | SqlType::Numeric(Some(value)) = cell
                 {
+                    // Rescale a copy and discard it: this only pre-flights the
+                    // conversion that `write_tvp_rows` performs for real, so a
+                    // rejectable value fails before any bytes reach the wire.
                     let mut value = *value;
                     apply_decimal_metadata(&mut value, precision, scale)?;
                 }
@@ -549,6 +554,9 @@ pub(crate) async fn write_tvp_rows(
             // the declared metadata; the value bytes come from the cell.
             let (_, ctx) = column.column_type.to_column_value_and_context(db_collation);
             let (mut value, _) = cell.to_column_value_and_context(db_collation);
+            // Callers must have run `TvpTableData::validate` first: this can
+            // reject a value after `TVP_ROW_TOKEN` and earlier cells are already
+            // buffered, which would leave a half-written TVP on the wire.
             apply_column_metadata(column, &mut value)?;
             TdsValueSerializer::serialize_value(packet_writer, &value, &ctx).await?;
         }
@@ -912,7 +920,8 @@ mod tests {
         assert!(matches!(
             error,
             Error::UsageError(message)
-                if message == "TVP decimal/numeric value overflows precision 38 at scale 1"
+                if message
+                    == "TVP decimal/numeric value magnitude is too large to rescale to scale 1"
         ));
     }
 
--- a/mssql-tds/src/datatypes/sql_tvp.rs
+++ b/mssql-tds/src/datatypes/sql_tvp.rs
@@ -18,6 +18,8 @@
 use bitflags::bitflags;
 
 use crate::core::TdsResult;
+use crate::datatypes::column_values::{ColumnValues, DEFAULT_VARTIME_SCALE, SqlTime};
+use crate::datatypes::decoder::DecimalParts;
 use crate::datatypes::sqltypes::SqlType;
 use crate::datatypes::tds_value_serializer::TdsValueSerializer;
 use crate::error::Error;
@@ -119,10 +121,12 @@ pub struct TvpColumnDef {
     pub column_type: SqlType,
     /// Per-column flags. `NULLABLE` is always set during serialization.
     pub flags: TvpColumnFlags,
-    /// Precision override for `Decimal`/`Numeric` columns.
+    /// Precision override for `Decimal`/`Numeric` columns. Values that exceed
+    /// this precision are rejected.
     pub precision: Option<u8>,
     /// Scale override for `Decimal`/`Numeric`/`Time`/`DateTime2`/
-    /// `DateTimeOffset` columns.
+    /// `DateTimeOffset` columns. Row values are converted to this scale;
+    /// conversions that would discard fractional data are rejected.
     pub scale: Option<u8>,
 }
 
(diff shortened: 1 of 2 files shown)


mssql-tds/src/datatypes/sql_tvp.rs
```diff
@@ -137,6 +141,159 @@ impl TvpColumnDef {
...
+
+    fn decimal_metadata(&self) -> Option<(u8, u8)> {
+        match &self.column_type {
+            SqlType::Decimal(value) | SqlType::Numeric(value) => Some((
+                self.precision
+                    .or_else(|| value.as_ref().map(|value| value.precision))
+                    .unwrap_or(1),
+                self.scale
+                    .or_else(|| value.as_ref().map(|value| value.scale))
+                    .unwrap_or(0),
```
saurabh500, line 153: These defaults match `write_type_info`, so metadata and rows can never disagree on the fallback. Good.

Worth spelling out in the PR description though, because it is user-visible: a caller who forgets to set `precision`/`scale` on a decimal column now gets a client-side `UsageError` for essentially any value with a fractional part or more than one digit. Before, that same call shipped `DECIMAL(1,0)` metadata and came back with an arithmetic-overflow from the server after a round trip. Better error locality either way, but anyone matching on the old failure shape sees something different now.

mssql-tds/src/datatypes/sql_tvp.rs
```diff
@@ -137,6 +141,159 @@ impl TvpColumnDef {
...
+                 without truncation",
+                value.scale
+            )));
+        }
+        magnitude /= scale_factor;
+    } else if value.scale < scale {
+        let scale_factor = 10_u128.pow(u32::from(scale - value.scale));
+        magnitude = magnitude.checked_mul(scale_factor).ok_or_else(|| {
+            Error::UsageError(format!(
+                "TVP decimal/numeric value overflows precision {precision} at scale {scale}"
```
saurabh500: This is a `u128` overflow, not a precision overflow — the multiply blew past `u128::MAX`, which is far beyond anything precision 38 could describe. Directionally true, but it points the reader at the wrong bound. `precision` is still used at line 247, so dropping it here does not leave an unused binding.

```suggestion
                "TVP decimal/numeric value magnitude is too large to rescale to scale {scale}"
```

mssql-tds/src/datatypes/sql_tvp.rs
```diff
@@ -137,6 +141,159 @@ impl TvpColumnDef {
...
+    Ok(())
+}
+
+fn temporal_value_fits_scale(value: &SqlTime, scale: u8) -> bool {
+    if scale > DEFAULT_VARTIME_SCALE {
+        return false;
+    }
+
+    let scale_factor = 10_u64.pow(u32::from(DEFAULT_VARTIME_SCALE - scale));
+    value.time_nanoseconds.is_multiple_of(scale_factor)
```
saurabh500, line 284: `10^(7 - scale)` is right, but only because `time_nanoseconds` counts 100-ns ticks rather than nanoseconds — the field name says otherwise. Worth a line so the next reader does not "fix" it.

```suggestion
    // Despite the name, `time_nanoseconds` counts 100-ns ticks (scale 7), so a
    // value survives scale N only when it is a multiple of 10^(7-N) ticks.
    let scale_factor = 10_u64.pow(u32::from(DEFAULT_VARTIME_SCALE - scale));
    value.time_nanoseconds.is_multiple_of(scale_factor)
```

mssql-tds/src/datatypes/sql_tvp.rs
```diff
@@ -210,6 +371,28 @@ impl TvpTableData {
...
+                        "TVP row {row_idx} column {col_idx} value cannot be represented at \
+                         temporal scale {scale}"
+                    )));
+                }
+
+                if let Some((precision, scale)) = column.decimal_metadata()
+                    && let SqlType::Decimal(Some(value)) | SqlType::Numeric(Some(value)) = cell
+                {
+                    let mut value = *value;
+                    apply_decimal_metadata(&mut value, precision, scale)?;
```
saurabh500, line 399: This copies the value, rescales it, and throws the result away purely to surface the error. It works, but two things follow that are not written down anywhere.

First, the same power-of-ten arithmetic now runs twice per decimal cell — once here and once in `write_tvp_rows`. Fine for typical TVPs, measurable at 10k+ rows.

Second, and more important: this check and the one in `apply_column_metadata` have to stay in lockstep. They agree today. If they ever drift, `write_tvp_rows` starts rejecting values after `TVP_ROW_TOKEN` is already in the buffer (see line 552). Naming the intent makes that dependency visible.

```suggestion
                    // Rescale a copy and discard it: this only pre-flights the
                    // conversion that `write_tvp_rows` performs for real, so a
                    // rejectable value fails before any bytes reach the wire.
                    let mut value = *value;
                    apply_decimal_metadata(&mut value, precision, scale)?;
```

mssql-tds/src/datatypes/sql_tvp.rs
```diff
@@ -365,7 +548,8 @@ pub(crate) async fn write_tvp_rows(
             // Context from the column definition keeps the encoding aligned with
             // the declared metadata; the value bytes come from the cell.
             let (_, ctx) = column.column_type.to_column_value_and_context(db_collation);
-            let (value, _) = cell.to_column_value_and_context(db_collation);
+            let (mut value, _) = cell.to_column_value_and_context(db_collation);
+            apply_column_metadata(column, &mut value)?;
```
saurabh500, line 560: `apply_column_metadata` can return `Err` here after `TVP_ROW_TOKEN` and any earlier cells of the row are already in the packet buffer, leaving a half-written TVP behind.

Unreachable today, and I checked rather than assumed: `SqlType::serialize_table` calls `TvpTableData::validate()` three lines before `write_tvp_rows`, the decimal arm matches the same `Decimal`/`Numeric` pairs that `to_column_value_and_context` produces, the temporal arm shares `temporal_value_fits_scale`, and `scale > 7` is already caught by `TvpColumnDef::validate`. The two sets are identical.

But nothing in this function's signature or docs says the caller must validate first, and the row-length check above carries the same unstated requirement. Pin it:

```suggestion
            // Callers must have run `TvpTableData::validate` first: this can
            // reject a value after `TVP_ROW_TOKEN` and earlier cells are already
            // buffered, which would leave a half-written TVP on the wire.
            apply_column_metadata(column, &mut value)?;
```

saurabh500 (commented):
## Verdict: sound fix, at the right layer. Nothing blocking.

Self-review notes. The inline comments are polish plus one behavior change that belongs in the PR description.

### What this changes

`TvpColumnDef.precision`/`.scale` used to reach exactly one place: `write_type_info`, which writes the **column metadata**. `write_tvp_rows` then built its `ctx` from the `None`-valued column template and its value from the raw cell, consulting the column definition for neither. So the metadata said one thing and the row bytes said another. Two distinct failures fell out of that:

**Decimal/Numeric — silently wrong numbers.** `serialize_decimal` writes the raw magnitude and never looks at scale (`ctx.precision` is pinned to 38, so the payload is always 17 bytes; `ctx.scale` is set to `Some(0)` and unused). The server decodes that magnitude using the *metadata* scale. A cell holding `123.4500` (magnitude `1_234_500`, scale 4) declared as `DECIMAL(9,2)` arrived as `12345.00`. No error — just off by `10^(value_scale - column_scale)`.

**Time/DateTime2/DateTimeOffset — a desynced stream.** `serialize_time` derives *both* the divisor and the byte length from `value.scale`. Metadata `TIME(3)` tells the server to read 4 bytes; a cell carrying `scale: 7` wrote 5. That extra byte shifts every field after it and corrupts the remainder of the row set.

`apply_column_metadata` now rewrites each value into the column's declared metadata immediately before serialization, and `validate()` pre-flights every cell so bad data fails before any bytes reach the wire. Lossy conversions are rejected rather than truncated, which matches what `DecimalParts::from_string` already does.

### Verification

- `cargo clippy --package mssql-tds --all-features --all-targets -- -D warnings` — clean.
- All 29 `sql_tvp` unit tests pass.
- **The tests pin the mechanism, not just the fix.** Drop `apply_column_metadata` from `write_tvp_rows` and `test_temporal_rows_use_column_scale` sees `5, 0x80, 0x96, ...` where it expects `4, 0xE8, 0x03, ...`, while `test_decimal_and_numeric_rows_use_odbc_fixed_width` sees `1_234_500` for `12_345`. Both fail for the right reason.
- Every pre-existing TVP integration test sets value scale equal to column scale, so none of them could have caught this — which is why the bug survived. The new integration test deliberately mismatches the two.
- **What did not run:** the integration test needs a live server; I only ran the unit tests. The error paths inside `write_tvp_rows` are unreachable in production (see the note on line 552), so they are covered through `validate()` only.

### Correctness spot checks that came back clean

- `decimal_metadata()` defaults of `(1, 0)` and `temporal_scale()`'s default of 7 both match `write_type_info`, so metadata and rows cannot disagree on the fallback.
- Overflow: the scale-down `checked_pow` handles a `value.scale` above 38; the scale-up `10_u128.pow` is bounded because `scale <= precision <= 38` is validated first, and `10^38 < 2^127`. `temporal_value_fits_scale` guards its own exponent with the early `scale > 7` return.
- NULL cells fall through both arms untouched, and the temporal scale-up path (`time.scale = scale` with `time_nanoseconds` left alone) is correct because the field always holds 100-ns ticks regardless of scale.
- `validate()` and `apply_column_metadata` cover an identical set of rejections — I walked both arms.

mssql-tds/src/datatypes/sql_tvp.rs
saurabh500: Documented that the copied decimal is deliberately rescaled and discarded to preflight the conversion before serialization writes any bytes.

> [!NOTE]

mssql-tds/src/datatypes/sql_tvp.rs
saurabh500: Updated the PR description to call out the retained `decimal(1,0)`/temporal scale 7 defaults and the new client-side `UsageError` behavior when omitted decimal metadata is incompatible with a value.

> [!NOTE]

mssql-tds/src/datatypes/sql_tvp.rs
saurabh500: Added the caller invariant at the serialization site, including why validation must precede buffering the row token and cells.

> [!NOTE]

mssql-tds/src/datatypes/sql_tvp.rs
saurabh500: Changed the diagnostic to identify magnitude overflow while rescaling, and updated the unit assertion accordingly.

> [!NOTE]

mssql-tds/src/datatypes/sql_tvp.rs
saurabh500: Added the comment explaining that `time_nanoseconds` stores 100-ns ticks and why representability requires a multiple of `10^(7-N)`.

> [!NOTE]

Theekshna (commented):
Unattended hourly sweep (first automated review of this PR). Reviewed the full diff against `origin/main` merge-base (`mssql-tds/src/datatypes/sql_tvp.rs` + `mssql-tds/tests/test_tvp.rs`, 618 lines).

The fix threads `TvpColumnDef.precision`/`.scale` into row serialization via a new `apply_column_metadata` seam, replacing the bare-`SqlType` context that caused issue #364 (row context defaulting to precision-38/17-byte decimals and cell-scale temporals regardless of the declared column metadata). Verified by hand:
- **Byte math**: re-derived `test_decimal_and_numeric_rows_use_odbc_fixed_width` — decimal(9,4)=1,234,500 rescaled to column scale 2 (÷100, exact) → magnitude 12,345; numeric(9,1)=1,234 rescaled to scale 2 (×10) → 12,340. Both match the asserted output bytes.
- **msodbcsql parity**: the PR's core claim — TVP decimal/numeric row values are always serialized as the fixed 17-byte (1 sign + 16 value) form regardless of declared precision, with precision instead governing a local digit-count rejection — matches the reference driver. `sqlccmd.cpp:2755` (`C:\REPOS\Drivers\ODBC_2\Sql\Ntdbms\sqlncli\odbc`): `serializedSize = SQL_MAX_NUMERIC_LEN + 1 /*sign*/; // This is always 17 bytes.`, with an adjoining comment citing TDS spec §2.2.5.5.1.6 for the same invariant. No divergence.
- **No duplicated/drifting logic**: `TvpTableData::validate()`'s pre-flight decimal/temporal checks call the exact same `apply_decimal_metadata` / `temporal_value_fits_scale` functions that `write_tvp_rows` uses for real, so the two can't drift apart. The sole production caller (`sqltypes.rs::serialize_table`) calls `data.validate()?` before `write_tvp_column_metadata`/`write_tvp_rows`, so the "half-written TVP on the wire" risk the code comments flag is actually closed, not just documented.
- **NULL handling**: `apply_column_metadata` matches on `ColumnValues::Decimal`/`Numeric`/temporal variants, which a NULL cell doesn't produce, so NULLs correctly bypass rescaling — exercised by both the unit test (`test_temporal_rows_use_column_scale`'s second row) and the live round-trip test's second row.
- **Test sufficiency**: `test_tvp_column_scale_controls_row_encoding` is a live-server round-trip (not mocked) that creates a real `decimal(9,2)`/`numeric(9,2)`/`time(3)`/`datetime2(3)`/`datetimeoffset(3)` table type, inserts values at a *larger* precision/scale than declared, and asserts the server-returned values match the rescaled form — this is the strongest available check for a wire-format claim. Issue #364's suggested scenarios (empty table, NULL rows, non-NULL rows, decimal/numeric(9,2), time/datetime2/datetimeoffset(3)) are all covered, including a pre-existing `test_tvp_empty_table`.

Could not execute this locally: this host's `ms-prod-1.95` codegen backend fails to compile `crc32fast`'s AVX-512/VPCLMULQDQ intrinsics (`E9012`/`E9013`) for every workspace crate regardless of `-C target-feature`/`target-cpu` overrides — confirmed pre-existing and unrelated to this PR by reproducing the identical failure against `mssql-tds` alone. `gh pr checks` for this head SHA has all fast checks green; the ADO `mssql-rs Pull request validation` pipeline (which runs the live-server integration suite, including `test_tvp_column_scale_controls_row_encoding`) and its macOS test stage are still `pending` at review time — the PR's own `cargo btest` checklist box is correctly left unchecked pending that result, not a contradiction.

No findings.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

mssql-tds/src/datatypes/sql_tvp.rs
```diff
@@ -137,6 +141,161 @@ impl TvpColumnDef {
...
+                    .unwrap_or(DEFAULT_VARTIME_SCALE),
+            ),
+            _ => None,
+        }
+    }
+
+    fn validate(&self) -> TdsResult<()> {
+        if let Some((precision, scale)) = self.decimal_metadata() {
+            validate_decimal_metadata(precision, scale)?;
+        }
```
Theekshna: Checked the msodbcsql parity claim against `C:\REPOS\Drivers\ODBC_2`. Two findings, both supporting this change:

**17-byte decimal rows.** `sqlccmd.cpp:4344-4348` computes `stLen = sizeof(SQL_NUMERIC_STRUCT) - offsetof(sign)` = 1 sign + 16 value, regardless of declared precision. Matches pinning `ctx.precision = 38` here. Corroborated at `:2755`.

**Rejecting non-representable temporals on input is what the reference driver does.** `sqlccnvt.cpp:4351-4356` returns `CVT_SCALE_FRAC_MISMATCH` gated on `fConversionDirection == TOSERVER` -> `IDS_22_008_01` -> SQLSTATE 22008, whose text is *"Datetime field overflow. Fractional second precision exceeds the scale specified in the parameter binding."* So the `UsageError` here is parity, not extra strictness.

No verified divergence. I could not trace where msodbcsql populates the `SQL_NUMERIC_STRUCT` for a TVP cell, so I have no basis to compare its decimal *rescaling* policy to `apply_decimal_metadata` - that one is unverified, not divergent.

mssql-tds/src/datatypes/sql_tvp.rs
```diff
@@ -137,6 +141,161 @@ impl TvpColumnDef {
...
+    value.time_nanoseconds.is_multiple_of(scale_factor)
+}
+
+fn apply_column_metadata(column: &TvpColumnDef, value: &mut ColumnValues) -> TdsResult<()> {
+    if let Some((precision, scale)) = column.decimal_metadata()
+        && let ColumnValues::Decimal(value) | ColumnValues::Numeric(value) = value
+    {
+        apply_decimal_metadata(value, precision, scale)?;
+    }
+
```
Theekshna: Verified this seam rather than assuming it. Three things I checked:

**Defaults can't drift.** `decimal_metadata()` / `temporal_scale()` reproduce `write_type_info`'s `override -> value -> (1,0) / 7` chain for all four cases - decimal at `sqltypes.rs:762-775`, Time/DateTime2/DateTimeOffset at `:811-835`. Metadata and rows can't disagree on the fallback.

**Overflow arithmetic holds.** Scale-down `checked_pow` covers `value.scale > 38`; scale-up `10_u128.pow` is bounded by the pre-validated `scale <= precision <= 38` with `10^38 < 2^127`; `checked_mul` catches the rest.

**The half-written-row hazard flagged further down is actually closed, not just documented.** `sqltypes.rs:1124` runs `data.validate()?` before `write_tvp_rows`, `validate()` calls these same two functions, and the `discriminant` check at line 369 stops a `Decimal` cell reaching a `Numeric` column.

mssql-tds/src/datatypes/sql_tvp.rs
```diff
@@ -608,6 +821,281 @@ mod tests {
...
+    #[test]
+    fn test_apply_decimal_metadata_rejects_scale_factor_overflow() {
+        let mut value = DecimalParts::new(true, 38, u8::MAX, 1);
+
+        let error = apply_decimal_metadata(&mut value, 38, 0).unwrap_err();
+
+        assert!(matches!(
+            error,
+            Error::UsageError(message)
+                if message == "TVP decimal/numeric value scale 255 cannot be converted to scale 0"
```
Theekshna: Nit: this and `test_apply_decimal_metadata_rejects_magnitude_overflow` below assert on exact error strings (`if message == "..."`). Any reword breaks both for zero signal. `message.contains("scale 255")` would pin the same behavior.

The tests themselves are worth keeping - `DecimalParts::new` is `pub` and `from_string("0", 38, 255)` hits the zero early-return and yields scale 255, so these states are reachable, not hypothetical.

mssql-tds/tests/test_tvp.rs
```diff
@@ -400,6 +400,133 @@ mod tvp_tests {
         .await;
     }
 
+    #[tokio::test]
+    async fn test_tvp_column_scale_controls_row_encoding() {
```
Theekshna: This is the strongest thing in the PR - a live round-trip that declares `decimal(9,2)`/`time(3)`/etc. and feeds cells at a *larger* scale, which is exactly the mismatch every pre-existing TVP test avoided. Re-derived the arithmetic by hand: `1_234_500 / 100 -> 12_345`, `1_234 * 10 -> 12_340`, `time(3)` of 1s -> `0x03E8` in 4 bytes. All consistent.

Worth noting in the description: the blast radius is narrower than the diff suggests. `mssql-py-core` routes TVP decimal cells through `types.rs:397 -> hinted_decimal -> from_string(value, precision, hint.scale)`, and `from_string` already rejects `input_scale > scale` and emits at the hint scale, so `apply_column_metadata` is a no-op there. `mssql-js` and `mssql-tds-cli` have no TVP surface at all. Only direct `mssql-tds` Rust callers were affected - and only they see the new client-side `UsageError`.

Caveat on my validation: `cargo btest` doesn't build on my host (`crc32fast` fails `E9012`/`E9013` on the `ms-prod-1.95` backend - pre-existing, reproduces against `mssql-tds` alone). I checked the byte assertions by reading, not by running. The ADO live-server stage is the only thing that will actually execute this test, so leaving the `cargo btest` box unchecked until it's green is the right call.

mssql-tds/src/datatypes/sql_tvp.rs
Theekshna: **Please document that this rescaling deliberately diverges from msodbcsql** — a doc comment here or a line in the PR description. I traced the reference driver's TVP decimal path and it does *not* do this, so a future reader chasing "parity" could delete `apply_decimal_metadata` and reintroduce #364.

What msodbcsql actually does for a TVP decimal column:
- **Metadata** scale/precision comes from the *binding* — `SetTvpColMetadata` (`sqlcmisc.cpp:5411-5412`) reads `pTvpCol->cbPrecision` / `pTvpCol->ibScale`, written once before any row.
- **Row value** is the app's `SQL_NUMERIC_STRUCT` copied **verbatim**. `SQLDECIMALN`/`SQLNUMERICN` are `fIsFastRPCType = TRUE` (`sqlcmisc.cpp:442,444`), so `ProcessRPCData` dispatches to `FastDescribeRPCParam`, whose decimal arm (`sqlcmisc.cpp:7006-7016`) just `WriteToExtBuffer`s the source struct. No conversion anywhere in that function.
- The one place it reconciles the two — the `// Do Fix-Up work so that no conversion will be attempted` block at `sqlcmisc.cpp:6630-6637`, which overwrites the declared prec/scale *from the value* — is nested inside the `fIsVariant` branch opened at `:6583`, so it never runs for a plain decimal column.

Net: msodbcsql sends binding-derived metadata alongside an unscaled value and makes the **application** responsible for pre-scaling; mismatched input is silently corrupted. This PR inverts that — value follows metadata, lossy conversions are rejected. Both reach the same invariant (metadata scale == value scale) by opposite means, and given `DecimalParts` carries its own scale (unlike ODBC, where scale lives in the descriptor), rescaling is the correct choice here. Just worth writing down.

mssql-tds/src/datatypes/sql_tvp.rs
Theekshna: Nit: this and `test_apply_decimal_metadata_rejects_magnitude_overflow` below assert exact error strings (`if message == "..."`). Any reword breaks both for zero signal — `message.contains("scale 255")` pins the same behavior.

Keep the tests themselves: `DecimalParts::new` is `pub` and `from_string("0", 38, 255)` hits the zero early-return and yields scale 255, so these states are reachable rather than hypothetical.

Theekshna (approved):

PR #493, opened 2026-09-03
[py-core] Implementing ExecuteMany

subrata-ms:
## Description
This pull request introduces significant improvements to the async cursor and row fetching logic in the codebase, primarily to support efficient buffering and iteration over multiple result sets, especially for batched execution scenarios like `executemany`. It adds new internal structures for buffering results, exposes the `rowcount` property, and refactors how parameter binding is handled for batched execution.

Key changes include:

**Buffered Results and Row Fetching Enhancements:**

* Introduced `BufferedResults` and `BufferedRowSet` in `async_fetch.rs` to enable efficient buffering of multiple result sets, allowing for batched row fetching and correct handling of `executemany` scenarios. New methods support advancing between result sets and draining rows efficiently. [[1]](diffhunk://#diff-6efdd76294f1dfe8ce03399d24e06da1e5e1e8e851ed540c6577009e7a630b4cR70-R111) [[2]](diffhunk://#diff-6efdd76294f1dfe8ce03399d24e06da1e5e1e8e851ed540c6577009e7a630b4cR120-R134) [[3]](diffhunk://#diff-6efdd76294f1dfe8ce03399d24e06da1e5e1e8e851ed540c6577009e7a630b4cR143-R144)
* Refactored the fetch logic to check for buffered results and use specialized handling (`fetch_buffered` and `next_buffered_set`) when present, ensuring correct fetch and result set iteration semantics. [[1]](diffhunk://#diff-6efdd76294f1dfe8ce03399d24e06da1e5e1e8e851ed540c6577009e7a630b4cR385-R401) [[2]](diffhunk://#diff-6efdd76294f1dfe8ce03399d24e06da1e5e1e8e851ed540c6577009e7a630b4cR526-R591) [[3]](diffhunk://#diff-6efdd76294f1dfe8ce03399d24e06da1e5e1e8e851ed540c6577009e7a630b4cR662-R676) [[4]](diffhunk://#diff-6efdd76294f1dfe8ce03399d24e06da1e5e1e8e851ed540c6577009e7a630b4cR801-R871)

**Async Cursor API and State:**

* Extended `PyAsyncCursor` to include `rowcount` (as an atomic integer) and `buffered_results` fields, exposing a `rowcount` getter to Python and ensuring these states are managed and passed through fetch and execute operations. [[1]](diffhunk://#diff-85fa845be93caab797168826f6e49584a85ce9577adb715c7dc051dd4919a179R182-R183) [[2]](diffhunk://#diff-85fa845be93caab797168826f6e49584a85ce9577adb715c7dc051dd4919a179R214-R215) [[3]](diffhunk://#diff-85fa845be93caab797168826f6e49584a85ce9577adb715c7dc051dd4919a179R241-R242) [[4]](diffhunk://#diff-85fa845be93caab797168826f6e49584a85ce9577adb715c7dc051dd4919a179R263-R264) [[5]](diffhunk://#diff-85fa845be93caab797168826f6e49584a85ce9577adb715c7dc051dd4919a179R401-R406)
* Added a new `executemany` method to `PyAsyncCursor`, supporting batched execution with proper buffering and row count aggregation.

**Parameter Binding Improvements:**

* Added a `ParameterBindingPlan` struct in `async_parameters.rs` to encapsulate SQL rewriting and placeholder metadata, optimizing repeated parameter binding for `executemany` by reusing the plan for each row.

**Other Notable Fixes:**

* Improved datetime to SQL type conversion by using a helper function for day extraction, making the logic more robust.

These changes collectively improve performance, correctness, and maintainability for batched and multi-result-set operations in the async MSSQL Python driver.

## Related Issues

ADO work item: https://dev.azure.com/SqlClientDrivers/mssql-python/_workitems/edit/47195
## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


mssql-py-core/src/async_execute.rs
```diff
@@ -429,11 +790,237 @@ pub(crate) fn execute<'py>(
...
+        );
+        let request = ExecuteManyRequest {
+            operation,
+            parameter_sets,
+            use_prepare,
+            timeout,
+            autocommit,
+        };
+        let claim = session_state
+            .claim_execute(cursor_id)
```
subrata-ms, line 908: **Blocking ? Recheck cursor lifecycle after asynchronous preflight**

execute_resources() checks closed before the awaitable is created, but ind_parameter_sets(...).await runs and yields before this claim. During a long preflight, wait cursor.close() can claim the idle session, finish cleanup, and set the cursor closed. Once preflight completes, this claim sees an open/idle connection and executes using the captured resources, so SQL can run through a cursor that has already closed. Please either claim the cursor operation before preflight with cancellation-safe release, or carry/recheck the cursor's closed state immediately before this claim (the session claim then arbitrates an in-progress close). Add a regression test using a yielding/slow input iterable that closes the cursor before preflight completes and asserts the pending executemany cannot execute.

subrata-ms (commented):
## Summary

This PR adds async executemany, buffered result-set navigation, row-count tracking, reusable parameter binding plans, and corrected DATETIMEOFFSET normalization. The implementation and coverage are substantial, but one cursor-lifecycle race must be fixed before merge.

## Blocking

- The Related Issues section is empty. Repository policy requires linking a GitHub issue (Fixes #...) or Azure DevOps work item before this PR is ready for review.

- One lifecycle bug is called out inline.

mssql-py-core/src/async_execute.rs
subrata-ms: fixed

David-Engel (commented):
Reviewed the `executemany` implementation end to end: preflight binding, sequential execution, result buffering, and the buffered `fetch*()`/`nextset()` fast paths. The session-state transitions look right to me — `finish_execute(produced_rows)` / `finish_fetch(exhausted, !exhausted || has_next)` line up with `AsyncConnectionState`, the guard drop ordering (`ExecuteManyInterruptionGuard` before `SessionOperationGuard`) produces the log-then-mark-broken sequence you want, and the buffered nextset/fetch state machine handles the zero-row and skip-final-set cases correctly. Test coverage is genuinely thorough. `cargo clippy --all-features --all-targets -- -D warnings` is clean against `mssql-py-core`.

Nothing blocking. Two things worth fixing, plus two nits.

## Suggestions

**1. The binding plan is rebuilt for every parameter row (`async_execute.rs`, `bind_parameter_chunk`)**

```rust
let plan = state
    .plan
    .get_or_insert(ParameterBindingPlan::new(&state.operation, row_is_named)?);
```

`Option::get_or_insert` takes its argument by value, so `ParameterBindingPlan::new(...)` is evaluated on every iteration and the result is discarded once `state.plan` is `Some`. That means the full placeholder lexer runs per row, and so does `rpc_prefix()`, which allocates a lowercase copy of the entire SQL and does a substring scan per candidate prefix. This is exactly the work `ParameterBindingPlan` was introduced to hoist out of the loop ("Rewritten SQL and placeholder metadata reused for every ExecuteMany row"), so today the type is doing the allocation but not the saving. It scales as O(rows x sql_len), which is most visible on the large batches `executemany` is for.

`get_or_insert_with` can't carry the `?`, so the straightforward fix is:

```rust
if state.plan.is_none() {
    state.plan = Some(ParameterBindingPlan::new(&state.operation, row_is_named)?);
}
let plan = state.plan.as_ref().expect("plan initialized above");
```

The mixed-shape check above already guarantees `row_is_named` matches the cached plan, so this is behavior-preserving.

**2. The DATETIMEOFFSET UTC normalization has two side effects beyond the intended fix (`types.rs`)**

The normalization itself is correct — `cursor.rs` decodes `DateTimeOffset` by building the datetime in `timezone.utc` and then calling `astimezone`, so the encoder does need to send UTC. Good catch, and it fixes a real round-trip bug. Two knock-on effects to consider:

- **`InputSqlType::DateTime` hint path** (line ~412): `SqlType::DateTimeOffset(value) => Ok(SqlType::DateTime2(value.map(|v| v.datetime2)))` now receives the UTC-normalized `datetime2`. So binding a tz-aware datetime with `setinputsizes([SQL_TYPE_TIMESTAMP])` silently shifts the stored value by the UTC offset, where it previously stored the wall clock the caller passed. That may be what you want, but it's a behavior change that isn't mentioned anywhere and isn't covered by a test — the new unit test only exercises the `DateTimeOffset` branch.
- **`py_to_column_value` metadata-aware path** (line ~1014, the `SqlDbType::DateTimeOffset` branch used by bulk copy) still writes local wall-clock time plus the offset. Before this change both paths agreed; now the same tz-aware `datetime` lands differently depending on whether it goes through parameter binding or bulk copy.

Also worth calling out in the PR description / CHANGELOG, since it's a wire-format behavior change riding along in an ExecuteMany PR.

## Nits

**3. `read_buffered_row_set` yields on a bare `256`** while the module defines `EXECUTEMANY_PREFLIGHT_CHUNK_SIZE` and `EXECUTEMANY_EXECUTION_YIELD_INTERVAL` right above. A named constant (or reusing the existing one) keeps the three yield intervals tunable together.

**4. Inconsistent `rows_affected` overflow handling in `execute_many_on_client`.** `execute_on_client` maps an out-of-`i64` count to `-1`, which the `count >= 0` check then drops from the total. The drain loop maps the same condition to `i64::MAX` via `unwrap_or(i64::MAX)` and counts it as a known value. Practically unreachable, but the two branches should agree.

David-Engel (commented):
Reviewed the full diff against `origin/main` (merge base `4382290`): the `executemany` preflight/execute/buffer pipeline, the buffered `fetch*()`/`nextset()` fast paths, `ParameterBindingPlan`, and the DATETIMEOFFSET UTC normalization. Required CI is green on `316ec49`, including the ADO validation matrix and the cross-repo `mssql-python` legs.

The state machine holds up. `finish_execute(produced_rows)` / `finish_fetch(exhausted, !exhausted || has_next)` line up with `AsyncConnectionState`, `drain_previous: row_index == 0 && claim.drain_previous` correctly drains only once per batch, `reset_cursor: row_index == 0` lets the prepared handle be reused across rows, and draining the buffered deque to empty leaves `fetch_state == Exhausted` so later `fetchall()`/`nextset()` calls fall back to the streaming path safely. Moving the implicit BEGIN to `begin_transaction_with_options(..., options)` is a real fix — it retires the `TODO(mssql-tds)` and bounds the prologue by the statement timeout. Test coverage is thorough and correctly uses `#temp` tables.

Nothing blocking. Four suggestions and two nits.

## Suggestions

**1. `ParameterBindingPlan` is rebuilt for every parameter row (`async_execute.rs`, `bind_parameter_chunk`)**

```rust
let plan = state
    .plan
    .get_or_insert(ParameterBindingPlan::new(&state.operation, row_is_named)?);
```

`Option::get_or_insert` takes its argument **by value**, so `ParameterBindingPlan::new(...)` is evaluated on every iteration and the result is thrown away once `state.plan` is `Some`. The placeholder lexer and `rpc_prefix()` (which lowercases a copy of the whole SQL) therefore run once per row, which is exactly the work the new type was introduced to hoist out of the loop. It scales O(rows × sql_len), so the cost is largest on the big batches `executemany` targets. `get_or_insert_with` can't carry the `?`, so:

```rust
if state.plan.is_none() {
    state.plan = Some(ParameterBindingPlan::new(&state.operation, row_is_named)?);
}
let plan = state.plan.as_ref().expect("plan initialized above");
```

The mixed-shape check just above already guarantees `row_is_named` matches the cached plan, so this is behavior-preserving. (Also flagged in the earlier human review and still open on this SHA — repeating it because it's the one item with real throughput impact.)

**2. `close()` never releases the buffered rows (`async_cursor.rs`)**

`execute()` clears the buffer synchronously (`buffered_results.replace(VecDeque::new())`), but `close()` doesn't. After an `executemany` that produces result sets, every buffered `PyRowWriter` for every set stays resident until the `PyAsyncCursor` object itself is collected — even though the cursor is closed and `fetch_resources()` will reject any further use. That's fine for `async with conn.cursor()`, but a cursor held in a list/attribute, or captured by an exception traceback frame, pins the entire buffer. Clearing `buffered_results` in `CursorCleanup::run` (or right where `closed` is set) makes the release deterministic and matches the `execute()` path.

**3. The terminal buffered `nextset()` skips the session-liveness check (`async_fetch.rs`, `next_buffered_set`)**

```rust
if fetch_state.status() == FetchStatus::Exhausted && !buffered_results.has_next() {
    buffered_results.advance();
    description_state.replace(None);
    return pyo3_async_runtimes::tokio::future_into_py(py, async { Ok(false) });
}
```

Every sibling path validates the session first: `fetch_buffered`'s `Exhausted` early return calls `session_state.ensure_open()`, and the non-buffered `nextset` only short-circuits to `Ok(false)` *after* `claim_fetch` returns `ClaimError::NoResultSet`. Here the last `nextset()` over a buffered batch returns `False` on a connection that has since been closed or marked broken, instead of raising `Connection is closed` / `Connection is broken`. An `ensure_open().map_err(map_claim_error)?` before the early return restores parity. Same path also skips the `rowcount.store(-1, ...)` the non-early branch performs — harmless today since buffered `executemany` already leaves `rowcount` at `-1`, but it's an easy divergence to trip over later.

**4. Buffering is unbounded in two places, and the docstring doesn't say so**

`bind_parameter_sets` materializes `RpcParameter`s for the *entire* input iterable before a single row executes, and `execute_many_on_client` then buffers *every* row of *every* result set before the awaitable resolves. Both are deliberate (preflight validation; result-set boundary preservation), and `read_buffered_row_set`'s `yield_now()` keeps the loop cooperative — but neither has a ceiling. The `executemany` docstring documents the sequencing and transaction semantics carefully; adding a sentence that peak memory is proportional to the full input plus the full result set would set the right expectation for the "stream 500k rows through a generator" case, where callers may reasonably assume the iterable is consumed lazily.

## Nits

**5. `read_buffered_row_set` yields on a bare `256`** while `EXECUTEMANY_PREFLIGHT_CHUNK_SIZE` and `EXECUTEMANY_EXECUTION_YIELD_INTERVAL` are defined at the top of the same module. A third named constant keeps the yield intervals tunable together and makes the magic number self-documenting.

**6. `operation.clone()` per parameter row (`execute_many_on_client`).** `ExecuteRequest.operation` is only consumed when the prepared statement is created — i.e. `row_index == 0` in the `use_prepare` path — yet the full rewritten SQL is cloned for every row. `Arc<str>` (or borrowing in `ExecuteRequest`) would drop a per-row allocation that grows with statement size.

Also, `total_rows_affected` handles the out-of-`i64` case two different ways in the same function: `execute_on_client` maps it to `-1` (which the `count >= 0` guard then drops from the total), while the drain loop maps it to `i64::MAX` via `unwrap_or(i64::MAX)` and counts it. Unreachable in practice, but the branches should agree.

saurabh500 (commented):
_Unattended review sweep — these findings were not checked by a human before posting, and I did **not** run the integration reproduction (the `mssql-py-core` suite needs a maturin build plus a live SQL Server; the `executemany` tests are `@pytest.mark.integration`). Everything below is traced from source at head `316ec49`. Comment-only._

**Verdict: COMMENT — no new blocking issue; ship-worthy once the open items are triaged.** This is a substantial, well-tested feature and the state machine lines up with `AsyncConnectionState`. CI is fully green (ADO matrix plus cross-repo mssql-python) and diff coverage is ~94%. David-Engel's two reviews and the Copilot pass already surface every line-specific issue I found, so I'm adjudicating them independently rather than opening duplicate inline threads. The item I'd prioritize is a `nextset()` inconsistency after a DML-only `executemany` (already filed by Copilot) — real but low-severity, with the exact fix below.

**Blocking:** None. Nothing here is data-losing or reachable on the common path; the maintainer review reached the same conclusion.

**Suggestions**

1. **DML-only `executemany` makes `nextset()` raise instead of return `False` — an internal inconsistency** (`async_execute.rs:959`; Copilot's open thread at `:960`). On a sequential call, a pure-DML batch (`INSERT` with no `OUTPUT`) buffers nothing, so `produced_rows` is false → `fetch_state` becomes `NoResultSet` and `finish_execute(false)` clears `active_operation`. `nextset()` then gets `ClaimError::NoResultSet` from `claim_fetch`, but its early-return (`async_fetch.rs:680`) only converts `Exhausted | TerminalNoRows` to `Ok(false)`, so it falls through to `ProgrammingError: "No active result set"`. A plain `execute("INSERT")` instead sets `TerminalNoRows` (`async_fetch.rs:734`) and `nextset()` returns `False`, and the multi-set `executemany` tests assert `nextset() is False` at the end — so the DML `executemany` path breaks the driver's own contract. (Per PEP 249's letter, raising after a no-result-set execute is actually permissible, so this is a consistency defect rather than a spec violation — but the driver has deliberately chosen the friendlier `False` everywhere else.) It's uncovered: no test calls `nextset()` after a pure-DML `executemany`. Fix — mirror the single-execute path for the non-empty case while leaving empty-input semantics untouched:

```rust
fetch_state.set(if produced_rows {
    FetchStatus::Ready
} else if batch_count > 0 {
    FetchStatus::TerminalNoRows
} else {
    FetchStatus::NoResultSet
});
```

The `batch_count > 0` guard matters: empty-input `executemany` also reaches this line, and you don't want an empty batch to start reporting a terminal result set. Worth a test for both non-empty DML (`nextset()` → `False`) and empty input (unchanged).

2. **`close()` never releases the buffered rows** (`async_cursor.rs`; Copilot `:183`, David-Engel #2). `execute()` clears the buffer via `buffered_results.replace(VecDeque::new())`, but `CursorCleanup` does not, so a closed-but-still-referenced cursor (held in a list, an attribute, or a traceback frame) pins every buffered `PyRowWriter` until the Python object is collected. Clearing `buffered_results` in `CursorCleanup::run` makes the release deterministic. Confirmed against both paths.

3. **Terminal buffered `nextset()` skips the liveness check** (`async_fetch.rs:813`; Copilot `:817`, David-Engel #3). The early return at 813–816 does not call `ensure_open()`, unlike `fetch_buffered` (`:540`) and the non-buffered `nextset`. So the final `nextset()` over a buffered batch returns `False` on a since-closed or broken connection instead of raising. Adding `ensure_open().map_err(map_claim_error)?` before that early return restores parity. Confirmed.

4. **`ParameterBindingPlan` is rebuilt every row** (`async_execute.rs:625`; David-Engel #1, and the author's own review). `Option::get_or_insert` evaluates its argument by value each iteration, so `ParameterBindingPlan::new` — the placeholder lexer plus `rpc_prefix()` lowercasing a copy of the whole SQL — runs per row and is discarded once `plan` is `Some`. This is the one item with a real `O(rows × sql_len)` throughput cost, exactly the work the type was meant to hoist out of the loop. The `is_none()`-then-`as_ref()` rewrite David-Engel suggested is behavior-preserving. Confirmed.

5. **DATETIMEOFFSET UTC normalization has two knock-on behavior changes** (`types.rs`; David-Engel #2). The core fix looks right, but the `InputSqlType::DateTime` hint path and the bulk-copy `SqlDbType::DateTimeOffset` path now handle the same tz-aware value differently. Worth a test on the hint path and a line in the description, since it's effectively a wire-format change riding along in this PR. Deferring to David-Engel's analysis on the exact expected values.

6. **Unbounded buffering, undocumented** (David-Engel #4). Preflight materializes every `RpcParameter` and execution buffers every row of every result set before the awaitable resolves, so peak memory scales with full input plus full result set. A sentence in the `executemany` docstring would set the right expectation for the "stream 500k rows" case.

**Nits**

- Bare `256` yield interval in `read_buffered_row_set` (`async_fetch.rs:542`) while the `EXECUTEMANY_*` constants sit at the top of the module (David-Engel #5).
- `operation.clone()` per row, and disagreeing overflow handling in `execute_many_on_client`: `execute_on_client` maps out-of-`i64` rows to `-1` (dropped by the `count >= 0` guard) while the drain loop maps it to `i64::MAX` and counts it (David-Engel #6). Unreachable in practice, but the two branches should agree.

**Tests:** Coverage is genuinely thorough and correctly uses `#temp` tables. The one gap is two integration tests named for yielding that assert only final state and never observe scheduler progress, so they cannot distinguish the `yield_now()` they target being present from absent except through incidental timing (Copilot flagged both): `test_executemany_yields_at_execution_interval` (`:774`, asserts only `rowcount == 256`) and `test_executemany_yields_while_buffering_large_result_set` (`:1011`, asserts only `fetchall()`). A controllable ticker or interleaving counter would make the named behavior actually observable.

Theekshna (commented):
Automated unattended run found nothing new to add. Covered: the executemany preflight/execute/buffer state machine, ParameterBindingPlan, DATETIMEOFFSET UTC normalization, mssql-py-core fmt/clippy (run separately, both clean), and the six extra checks (ODBC parity N/A, no undocumented benchmark claims to audit). All substantive issues here (DML-only nextset() inconsistency, close() buffer leak, terminal-nextset liveness check, ParameterBindingPlan per-row rebuild, DATETIMEOFFSET hint-path side effect) are already filed by @David-Engel and @saurabh500 and remain open/unresolved -- not re-filing.

mssql-py-core/src/async_cursor.rs
```diff
@@ -179,6 +179,8 @@ pub struct PyAsyncCursor {
     closed: Arc<AtomicBool>,
     fetch_state: Arc<FetchState>,
     description_state: Arc<DescriptionState>,
+    rowcount: Arc<AtomicI64>,
+    buffered_results: Arc<BufferedResults>,
```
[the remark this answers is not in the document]
subrata-ms: CursorCleanup now clears unread buffered ExecuteMany results, resets the fetch state to NoResultSet, and clears the cursor description when close() completes. The same cleanup is performed when the parent connection is already closing or closed. This prevents a long-lived reference to a closed cursor from retaining potentially large buffered result sets.

A regression test now buffers 512 unread rows, closes the cursor while retaining the Python reference, verifies that result metadata is cleared, and confirms the connection remains usable from another cursor. The pre-existing test expecting description metadata to survive cursor closure was updated to reflect the new cleanup semantics.

mssql-py-core/src/async_execute.rs
```diff
@@ -429,11 +794,241 @@ pub(crate) fn execute<'py>(
...
+                        }
+                    };
+                    let description_materialization_ms = description_started.elapsed().as_millis();
+                    buffered_results.replace(results);
+                    description_state.replace(description);
+                    fetch_state.set(if produced_rows {
+                        FetchStatus::Ready
+                    } else {
+                        FetchStatus::NoResultSet
+                    });
```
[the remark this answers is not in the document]
subrata-ms: Implemented and validated the fix.

Changes
Non-empty DML-only executemany() now transitions to TerminalNoRows, matching ordinary DML execute() behavior:
Fetch methods still raise ProgrammingError.
nextset() consistently returns False, including repeated calls.
Aggregated rowcount remains unchanged.
Empty ExecuteMany input remains NoResultSet, so nextset() continues raising ProgrammingError.
Added regression coverage for both paths in test_async_executemany.py:560-568 and test_async_executemany.py:760-779.
No changes were made to mssql-tds or ImplementationDoc.md.
Validation
Rust formatting: passed
Strict Clippy with all features and targets: passed
Native extension rebuild: passed
Focused tests: 3 passed
Complete ExecuteMany suite: 53 passed
Full standard Python suite: 980 passed, 10 skipped, 3 deselected
Reviewer response
Thanks for identifying this distinction. The issue was valid: successful non-empty DML-only executemany() operations were being left in NoResultSet, causing nextset() to raise instead of returning False as it does after ordinary DML execute().

The final state now distinguishes the two cases:

Non-empty DML-only ExecuteMany transitions to TerminalNoRows.
Empty ExecuteMany input remains NoResultSet, since no statement was executed.
Regression tests verify that DML fetch methods continue to raise, repeated nextset() calls return False, aggregated rowcount is preserved, and empty-input nextset() still raises. The full standard test suite passes with 980 tests passed.

mssql-py-core/src/async_fetch.rs
```diff
@@ -644,6 +799,77 @@ pub(crate) fn nextset<'py>(
...
+    fetch_state: Arc<FetchState>,
+    description_state: Arc<DescriptionState>,
+    buffered_results: Arc<BufferedResults>,
+    rowcount: Arc<AtomicI64>,
+) -> PyResult<Bound<'py, PyAny>> {
+    if fetch_state.status() == FetchStatus::Exhausted && !buffered_results.has_next() {
+        buffered_results.advance();
+        description_state.replace(None);
+        return pyo3_async_runtimes::tokio::future_into_py(py, async { Ok(false) });
+    }
```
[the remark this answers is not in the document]
subrata-ms: Thanks for catching this. The terminal buffered nextset() fast path bypassed the connection lifecycle check and could return False after the connection had entered closing, closed, or broken state.

The path now calls session_state.ensure_open() before advancing buffered results or clearing the cursor description. This aligns it with buffered fetch*() and normal nextset() behavior.

Regression coverage verifies both closing and closed connection states and confirms that rejected calls do not mutate the cursor description. The complete ExecuteMany suite and full standard Python suite pass.

mssql-py-core/tests/test_async_executemany.py
```diff
@@ -0,0 +1,1397 @@
...
+            )
+            assert await cursor.fetchall() == [(1, "one"), (2, "two"), (3, "three")]
+        finally:
+            await conn.close()
+
+    asyncio.run(run())
+
+
+@pytest.mark.integration
+def test_executemany_yields_at_execution_interval(client_context):
```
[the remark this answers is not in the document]
subrata-ms: Thanks for identifying this. Both existing tests only validated the resulting row count or buffered rows, so they did not observe cooperative scheduling and continued to pass when the explicit yields were removed.

The two output-only tests have been replaced with a deterministic current-thread Tokio test. Both ExecuteMany execution and result buffering now use a shared interval-aware yield helper. The test schedules a waiting task, verifies it does not run before the interval, and verifies it runs when the 256-item boundary is reached.

A mutation check replacing yield_now() with an immediately ready future caused the new test to fail, confirming that it guards the intended fairness behavior. Formatting, strict Clippy, all Rust tests, the ExecuteMany suite, and the full standard Python suite pass.

David-Engel (commented):

Re-reviewed at head `0356cd1`, focused on the delta from `316ec49` (the previously reviewed SHA) and re-checked the full diff against merge base `4382290`. Built `mssql-py-core` locally and ran the crate's Rust unit tests; the four resolved Copilot threads are genuinely fixed, and I ran mutation checks against the two new Rust tests. ADO work-item cross-check was skipped this run (infrastructure, not your problem). Required CI on this head is still pending — nothing failing.

**Verified fixes on this head**

- `nextset()` after a non-empty DML-only `executemany` now returns `False`: `fetch_state.set(...)` picks `TerminalNoRows` when `batch_count > 0`, which `nextset`'s `ClaimError::NoResultSet` arm (`async_fetch.rs:680-687`) converts to `Ok(false)`. Empty input still lands on `NoResultSet` and still raises. Both paths have tests.
- `next_buffered_set`'s terminal fast path now calls `ensure_open().map_err(map_claim_error)?` before `advance()`, and `test_executemany_exhausted_nextset_checks_connection_lifecycle` asserts both the `Closing` and `Closed` raise **and** that `description` is not mutated. Good guard.
- `CursorCleanup::clear_result_state()` is wired into both the normal close and the `Closing | Closed` claim branch.
- The two output-only "yields" integration tests were replaced with `executemany_yields_at_execution_and_buffer_boundaries`. I mutated `yield_at_interval` to `std::future::ready(()).await` and the test failed (`left: 0, right: 1`), so the helper is really guarded. I also mutated the DATETIMEOFFSET normalization back to `py_obj` and `datetimeoffset_normalizes_datetime_to_utc_and_preserves_offset` failed. Both new tests earn their keep. (For anyone reading the assertion and doing the arithmetic: `SqlTime::time_nanoseconds` is 100 ns ticks despite the name, so `360_000_000_000` is the correct 10:00:00 UTC value, not a 100x error.)

Nothing blocking. Three of the items below are carried over from the earlier reviews and are still open in the code on this head; I'm restating them compactly rather than re-filing inline, since the commit message says review comments were addressed and these were not.

## Suggestions

**1. `ParameterBindingPlan` is still rebuilt for every parameter row** (`async_execute.rs`, `bind_parameter_chunk`)

```rust
let plan = state
    .plan
    .get_or_insert(ParameterBindingPlan::new(&state.operation, row_is_named)?);
```

`Option::get_or_insert` takes its argument **by value**, so `ParameterBindingPlan::new(...)` is evaluated on every iteration and thrown away once `state.plan` is `Some`. The placeholder lexer and `rpc_prefix()` (which lowercases a copy of the whole SQL) therefore run per row — exactly the work the new type was introduced to hoist out of the loop. It scales `O(rows × sql_len)`, which is worst on the large batches `executemany` targets. `get_or_insert_with` can't carry the `?`, so:

```rust
if state.plan.is_none() {
    state.plan = Some(ParameterBindingPlan::new(&state.operation, row_is_named)?);
}
let plan = state.plan.as_ref().expect("plan initialized above");
```

The mixed-shape check immediately above already guarantees `row_is_named` matches the cached plan, so this is behavior-preserving. This is the one open item with measurable throughput impact.

**2. The DATETIMEOFFSET UTC normalization still has two knock-on effects** (`types.rs`)

The core fix is right — MS-TDS stores DATETIMEOFFSET as UTC date/time plus the offset in minutes, and `cursor.rs` decodes by building in `timezone.utc` then calling `astimezone`, so the encoder has to send UTC. Two side effects remain:

- `InputSqlType::DateTime` (`types.rs:414-416`) does `SqlType::DateTimeOffset(value) => Ok(SqlType::DateTime2(value.map(|v| v.datetime2)))`, and that `datetime2` is now UTC-normalized. Binding a tz-aware datetime with `setinputsizes([SQL_TYPE_TIMESTAMP])` therefore stores a value shifted by the UTC offset, where it previously stored the wall clock the caller passed. No test covers that branch — `datetimeoffset_normalizes_datetime_to_utc_and_preserves_offset` only exercises the `DateTimeOffset` arm.
- The bulk-copy path (`py_to_column_value`, `SqlDbType::DateTimeOffset` at `types.rs:1014`) still writes local wall clock plus offset. I confirmed the divergence empirically: with the normalization reverted, `arrow_bulkcopy::tests::tz_aware_timestamp_to_datetimeoffset_ok` still passes, so the two paths are fully independent and now disagree for the same tz-aware `datetime`.

Worth a test on the hint path and a line in the PR description, since it's a wire-format behavior change riding along in an ExecuteMany PR.

**3. `test_executemany_close_clears_unread_buffered_result_state` doesn't guard what it's named for**

The test buffers 512 rows, closes the cursor, and asserts `cursor.description is None`. But `clear_result_state()` does three independent things, and `description_state.replace(None)` alone satisfies that assertion — delete the `self.buffered_results.replace(Default::default())` line and the test stays green. Since the buffer isn't observable from Python on a closed cursor (`fetch_resources()` rejects first), the release can only be pinned in Rust. Extracting the body as a free function and unit-testing it is cheap:

```rust
fn clear_result_state(
    fetch_state: &FetchState,
    description_state: &DescriptionState,
    buffered_results: &BufferedResults,
) { /* ... */ }
```

then assert `!buffered_results.has_current()` after calling it on a populated `BufferedResults`. Keep the integration test as-is for the description/connection-reuse half.

## Nits

**4. `description` is now cleared on `close()` for *every* cursor, not just buffered ones.** `test_description_survives_fetch_empty_result_exhaustion_and_close` was renamed and its assertion flipped from `== description` to `is None`, which is the tell: this is a user-visible change to a closed cursor's `.description` that isn't mentioned in the PR description. It's what Copilot's thread asked for, so I'm not arguing against it — but the cross-repo `mssql-python` legs are the thing that would catch a consumer reading `description` after `close()`, and those are still pending on this head. Worth confirming they go green before merge, and worth a line in the description.

**5. `operation.clone()` per parameter row** (`execute_many_on_client`). `ExecuteRequest.operation` is only consumed when the prepared statement is created (`row_index == 0` in the `use_prepare` path, or on a signature change), yet the full rewritten SQL is cloned for every row. `Arc<str>` would drop a per-row allocation that grows with statement size.

**6. `rows_affected` overflow is still handled two ways in the same function.** `execute_on_client` maps an out-of-`i64` count to `-1` (which the `count >= 0` guard then drops from the total), while the drain loop maps the same condition to `i64::MAX` via `unwrap_or(i64::MAX)` and counts it as known. Unreachable in practice, but the branches should agree.

**7. The new yield test guards the helper, not the call sites.** `assert_yields_at_boundary` exercises `yield_at_interval` directly, so deleting the `yield_at_interval(...)` call from `read_buffered_row_set` or from the `execute_many_on_client` loop leaves it green — the two integration tests that (weakly) covered those call sites were removed. The helper being correct is the important half and is now solid; just noting the call sites are unpinned. `test_executemany_preflight_yields_to_event_loop` (4096 rows, mock client, observable ticker) is a good model if you want one for the execution loop.

David-Engel (commented):
> **Automated review** — unattended run, not checked by a human before posting. This is **not** an approval and does not satisfy the human review requirement. Push back on anything that looks wrong.

Re-reviewed at head `1bc714b` against merge base `4382290`, focused on the delta from `0356cd1` (the previously reviewed SHA). Built `mssql-py-core` locally and ran its unit suite (230 passed), plus `cargo fmt -- --check` and `cargo clippy --all-features --all-targets -- -D warnings` against the excluded crate separately — all clean. Required CI on this head is pending, nothing failing. ADO work item 47195 was not cross-checked this run (reviewer infrastructure, not your problem).

**Verified fixes on this head**

- `ParameterBindingPlan` is now built once. `get_or_create_binding_plan` replaces the `Option::get_or_insert` call whose argument was evaluated per row. I mutated it back to `Ok(plan.get_or_insert(ParameterBindingPlan::new(operation, named)?))` and `executemany_reuses_parameter_binding_plan` failed, so the new test genuinely guards the fix — it catches the eager evaluation through the second call's `TypeError` rather than through pointer identity, but it does catch it.
- `rowcount_from_rows_affected` unifies the out-of-`i64` handling: both `execute_on_client` and the drain loop now map it to `-1`, which the `count >= 0` guard drops from the total. `rows_affected_overflow_is_unknown` covers `None`, `i64::MAX`, `i64::MAX + 1` and `u64::MAX`.
- `ExecuteRequest::operation` is a `Cow`, so `execute_many_on_client` borrows instead of cloning the rewritten SQL per row. The remaining `into_owned()` calls are on the non-prepared branches, where `TdsClient` takes an owned `String` anyway.
- `clear_result_state` is now a free function with a direct Rust unit test (`clear_result_state_discards_buffered_results`) that asserts `!buffered_results.has_current()`, so the buffer release is pinned rather than being incidentally satisfied by the `description is None` assertion.
- The `InputSqlType::DateTime` hint path no longer inherits the UTC normalization. Binding a tz-aware `datetime` with `setinputsizes([SQL_TYPE_TIMESTAMP])` stores the wall clock the caller passed, matching pre-PR behavior, and the strengthened test pins both `days` (local date, not the UTC date) and `time_nanoseconds == 110_450_000_060` under a +14:00 offset. Passing a `datetime.date` under that hint still errors — `py_time` fails on `getattr("hour")` — so the removed `py_datetime_to_sql_type` round-trip did not silently become a midnight default.
- The `executemany` docstring now states that peak memory scales with the full input plus all buffered rows.

I also re-walked the buffered state machine and the lifecycle guards and did not find a new problem. `claim_fetch` moves the phase to `FetchingRow` and `claim_cursor_close` only tolerates `Fetching`, so an in-flight `fetch_buffered` cannot race a `close()` that clears the deque — the `take_rows(...).expect(...)` invariant holds. Draining a buffered set to empty leaves `fetch_state == Exhausted` with the set still queued, so the terminal `nextset()` pops it and later `fetch*()` calls fall through to the streaming path and return empty, matching the non-buffered contract.

Nothing blocking. One carried-over suggestion and two nits.

## Suggestion

**1. The bulk-copy DATETIMEOFFSET path still disagrees with parameter binding** (`types.rs`, `py_to_column_value`, `SqlDbType::DateTimeOffset` branch around line 1011). This is the half of the earlier DATETIMEOFFSET item that is still open on this head — the `InputSqlType::DateTime` half was fixed, this one was not. `py_datetime_to_sql_type` now calls `astimezone(timezone.utc)` and sends UTC date/time plus the offset, which is right for MS-TDS and for the `cursor.rs` decoder that builds in `timezone.utc` then calls `astimezone`. The bulk-copy branch still writes `year`/`month`/`day` and `hour`/`minute`/`second` straight off the caller's object plus `utcoffset()`, so the same tz-aware `datetime` now lands differently depending on whether it goes through `executemany`/`execute` parameter binding or through bulk copy. The bulk-copy behavior predates this PR, so fixing it here is optional — but the divergence is new, and it is worth either a follow-up issue or a line in the PR description so it is not discovered from a round-trip mismatch later.

## Nits

**2. The terminal buffered `nextset()` does not reset `rowcount`** (`async_fetch.rs`, `next_buffered_set`). The early return now correctly calls `ensure_open().map_err(map_claim_error)?` before `advance()` — good — but it skips the `rowcount.store(-1, Ordering::Release)` that the non-early branch performs. Harmless today, because a buffered `executemany` already leaves `rowcount` at `-1` whenever any result set was produced, so the two paths cannot currently disagree. It is still an easy divergence to trip over if the aggregation rule ever changes.

**3. `setinputsizes()` hints survive an empty-input `executemany`.** `consume_input_sizes(&cursor, input_sizes_generation)` is guarded by `if batch_count > 0`, while `execute()` consumes the hints unconditionally on success. Not consuming them when nothing executed is defensible, but it is an undocumented difference between the two entry points; a one-line comment saying the guard is deliberate would save the next reader the trip through both call sites.

Also worth watching before merge: `test_description_survives_fetch_empty_result_exhaustion_and_close` was renamed and its assertion flipped to `description is None`, which makes a closed cursor's `.description` user-visibly different. The cross-repo `mssql-python` legs are the thing that would catch a consumer reading `description` after `close()`, and they are still pending on this head.

subrata-ms:
@David-Engel , the bulk-copy DATETIMEOFFSET change/fix is not in the scope of this PR. I will raise a separate bug for that.

mssql-py-core/tests/test_async_executemany.py
```diff
@@ -0,0 +1,1488 @@
...
+            await cursor.executemany(
+                "SELECT CAST(? AS int) AS value",
+                [(1,)],
+                use_prepare=False,
+            )
+            assert await cursor.fetchall() == [(1,)]
+            description = cursor.description
+
+            close_awaitable = conn.close()
+            with pytest.raises(RuntimeError, match="Connection is closing"):
```
David-Engel: The `Connection is closing` expectation here is a race, and it is what is failing required CI on this head (Build Linux and the cross-repo leg, build 172473): `Expected regex: 'Connection is closing' / Actual message: 'Connection is closed'`.

`PyAsyncConnection::close` (`async_connection.rs:300-335`) calls `begin_close()` synchronously and then hands teardown to `future_into_py`, which spawns onto the tokio runtime immediately rather than waiting for the Python `await`. The `Closing` window therefore lasts only as long as the TDS logout plus transport teardown, and nothing synchronizes the test against it — by the time the interpreter reaches `cursor.nextset()` the spawned task can already have called `mark_closed()`.

The `Closed` half below is deterministic and is what actually pins the new `ensure_open()` guard in `next_buffered_set`; keep it. For this half, either accept both states, since the guarantee is "rejected without mutating cursor state" rather than which lifecycle error surfaces:

```suggestion
            with pytest.raises(RuntimeError, match="Connection is (closing|closed)"):
```

or force the observation with something that holds the close task before `mark_closed()`, in the spirit of the mock-client ticker in `test_executemany_preflight_yields_to_event_loop`. Note this test passed in build 172296 and fails in 172473 with no relevant code change between them, so a pipeline re-run may hide it rather than fix it.

David-Engel (commented):

Re-reviewed at head `245a4a3` against merge base `4382290`. The delta from `1bc714b` (the previously reviewed SHA) is two lines, and both are the nit fixes from that round: `next_buffered_set` now stores `rowcount = -1` on the terminal buffered early-return path so it no longer diverges from the non-early branch, and the `if batch_count > 0` guard around `consume_input_sizes` now carries a comment saying the retention is deliberate. Both are correct and I have nothing further on them. The carried-over bulk-copy DATETIMEOFFSET suggestion is answered — you said you would file it separately, which is the right call since that path predates this PR.

I did not rebuild locally this round. The Build Linux leg of build `172473` compiled `mssql-py-core` and ran its full suite on this exact head, which is stronger evidence than anything I could reproduce here, so I read the CI logs instead. ADO work item 47195 was not cross-checked (reviewer infrastructure, not your problem).

## Verification

Required CI is **failing on this head**, and one of the failures is real. Build `172473` fails on `Build Stage Build Linux` and on `Build mssql-python (cross-repo)`, and both fail for the same reason: `tests/test_async_executemany.py::test_executemany_exhausted_nextset_checks_connection_lifecycle` fails with `AssertionError: Regex pattern did not match. Expected regex: 'Connection is closing' / Actual message: 'Connection is closed'` at `test_async_executemany.py:947`. Linux reported `1 failed, 972 passed, 11 skipped, 12 deselected`; the cross-repo leg reported `1 failed, 972 passed` and then failed `Build mssql-python wheel` with `ls: cannot access 'mssql_python/*.so'`, which is a cascade of the test step exiting non-zero rather than an independent problem. Fix the test and the wheel leg should clear with it.

The macOS failures look like infrastructure, not yours: `Install and start Colima-based Docker` timed out, which left `Test MacOS` with no database, and the two `Save code coverage ...` tasks and `Publish Cobertura coverage artifact` fail downstream of that (`Path does not exist: .../target/cobertura.xml`). `Build MacOS`, `Build Windows`, `Build Windows ARM`, `Build Linux ARM`, the Kerberos leg, and the `mssql-python suite on mssql-odbc driver` leg all pass. I would not chase the macOS legs.

Differential check on the failing test: the same test passed in build `172296` (head `1bc714b`, `partiallySucceeded`) and fails in `172473`. The only code change between those heads is the `rowcount.store(-1, ...)` line, which sits *after* the `ensure_open()` early return and cannot affect the error path this test exercises. So this is a pre-existing flake in the new test rather than a regression introduced by the last commit — which also means re-running the pipeline may well go green and hide it.

## Blocking

**1. `test_executemany_exhausted_nextset_checks_connection_lifecycle` races the connection teardown and is failing required CI** (`mssql-py-core/tests/test_async_executemany.py:946-949`). `PyAsyncConnection::close` (`async_connection.rs:300-335`) calls `session_state.begin_close()` synchronously and then hands the teardown to `pyo3_async_runtimes::tokio::future_into_py`, which spawns the task on the multi-threaded runtime immediately — it does not wait for the Python `await`. So the `Closing` state is only observable for as long as the TDS logout plus transport teardown takes, and nothing in the test synchronizes against that. Between `close_awaitable = conn.close()` returning and the interpreter reaching `cursor.nextset()`, the spawned task can already have run `close_client(...)` and `mark_closed()`, at which point `map_claim_error` yields `Connection is closed` and the `match="Connection is closing"` expectation fails. Against the containerized `sql1` server the round trip is fast enough that this lands on the wrong side regularly. This is also the only test in the suite that asserts on the `Closing` window, so there is no precedent to lean on.

The `Closed` half of the test is deterministic and worth keeping — it is what actually pins the `ensure_open()` guard you added to `next_buffered_set`. For the `Closing` half, either drop it or make it not depend on timing. The cheap fix is to accept both states for the pre-await call, since what the guard promises is "rejected, and cursor state untouched", not which of the two lifecycle errors surfaces:

```python
close_awaitable = conn.close()
with pytest.raises(RuntimeError, match="Connection is (closing|closed)"):
    await cursor.nextset()
assert cursor.description == description
```

If you want to keep asserting `Closing` specifically, the observation has to be forced rather than hoped for — the test needs something that holds the close task before `mark_closed()` runs, in the spirit of the mock-client ticker you already use in `test_executemany_preflight_yields_to_event_loop`. Whichever way you go, please re-run validation and confirm the Linux and cross-repo legs are green, rather than relying on a re-run that happens to win the race.

## Suggestions

None on this head.

## Nits

None on this head.

sylvesterkaczmarek (commented):
One DB-API rowcount issue: `executemany()` ignores executions whose `rows_affected` is unknown and still returns the sum of the known rows. If one row reports unknown and another reports 1, the cursor reports 1 even though the total is indeterminate. Could any unknown per-row count keep the aggregate `rowcount` at -1?

mssql-py-core/tests/test_async_executemany.py
subrata-ms: fixed

David-Engel (commented):
> **Automated review** — unattended run, not checked by a human before posting. This is **not an approval** and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

Re-reviewed at head `0c56fca` against merge base `4382290`, with a full re-read of the diff (`async_execute.rs`, `async_fetch.rs`, `async_cursor.rs`, `async_parameters.rs`, `types.rs`) and a focused look at the delta from `245a4a3`.

The delta is the two-line fix for the blocking finding from the last round, and it is the right fix. `test_executemany_exhausted_nextset_checks_connection_lifecycle` now matches `r"Connection is (?:closing|closed)"` with a comment explaining why the state is not deterministic. That is exactly what the guard promises — the call is rejected and `cursor.description` is untouched — without pinning which lifecycle error wins the race against the spawned close task. The `Closed` half of the test, which is what actually pins the `ensure_open()` guard you added to `next_buffered_set`, is unchanged and still deterministic. Nothing else moved.

I re-walked the rest and did not find a new blocking issue. `execute_many_on_client` reuses the prepared statement correctly (`reset_cursor: row_index == 0`, `drain_previous: row_index == 0 && claim.drain_previous`), the `while client.has_open_batch()` loop buffers trailing result sets from multi-statement batches, and the rowcount aggregate (`results.is_empty() && has_known_count`) collapses to `-1` whenever anything produced rows. `begin_transaction_with_options(..., options.clone())` still routes BEGIN failures through `ExecuteFailure::broken`, so replacing the TODO did not weaken the conservative poisoning. The DATETIMEOFFSET normalization is right: `astimezone(utc)` before `datetime_days`/`py_time` matches the TDS wire format (UTC date/time plus offset minutes), and the new unit test pins it correctly — `2023-10-30T00:00+14:00` → 2023-10-29, 10h as `360_000_000_000` 100ns ticks, offset `840`. The `InputSqlType::DateTime` hint path keeps wall-clock semantics, which is what it did before this PR.

## Verification

Built and checked `mssql-py-core` locally (it is excluded from the workspace, so run separately): `cargo clippy --all-features --all-targets -- -D warnings` is clean, and `cargo nextest run --all-features` reports **230 passed, 1 skipped**, including the new `clear_result_state_discards_buffered_results`, `executemany_yields_at_execution_and_buffer_boundaries`, `executemany_reuses_parameter_binding_plan`, `rows_affected_overflow_is_unknown`, and `datetimeoffset_normalizes_datetime_to_utc_and_preserves_offset`. I did not run the Python integration suite — it needs a maturin build plus a live SQL Server, and the `executemany` tests are `@pytest.mark.integration`.

Required CI on this head is **not green yet**. `Build Stage Test MacOS` is failing at 14m40s in build `172484`, which matches the Colima/Docker startup timeout that failed the same leg on the previous head — infrastructure, not your change, and I would not chase it. More importantly, `Build Stage Build Linux`, `Build Stage Build Linux ARM`, and the aggregate `mssql-rs Pull request validation` are still **pending**, and Linux is the leg that ran the failing test last round. Please confirm it goes green on `0c56fca` before marking this ready — that is the whole point of this commit.

## Blocking

None.

## Suggestions

**1. `fetch_buffered` can panic if the buffer is cleared between the dispatch check and the spawned task** (`mssql-py-core/src/async_fetch.rs:566-568`). `fetch()` decides to take the buffered path on `buffered_results.has_current()` while holding the GIL, but `take_rows(limit)` runs later on a tokio worker and `.expect("buffered fetch requires a current result set")` assumes the buffer is still populated. Almost every writer is excluded by the claim state machine — `claim_execute` and `claim_cursor_close` both reject a cursor whose phase is `FetchingRow`, so an in-flight fetch blocks them — but `PyAsyncCursor::close`'s `ClaimError::Closing | Closed` arm (`async_cursor.rs:566-570`) is the exception: it calls `cleanup.clear_result_state()` with no claim at all. Since `PyAsyncConnection::close` (`async_connection.rs:300-310`) flips the lifecycle to `Closing` synchronously and unconditionally, this sequence is reachable:

```python
fut = cursor.fetchall()   # buffered path; claim_fetch OK, task spawned
conn.close()              # begin_close() -> Closing, no busy check
cursor.close()            # ClaimError::Closing -> clear_result_state() empties the buffer
await fut                 # take_rows() -> None -> expect() panics
```

The window is small — the worker usually reaches `take_rows` within microseconds — and it needs the caller to create the awaitable without awaiting it, which is unusual but not exotic (your own tests hold `close_awaitable = conn.close()` across statements). The blast radius is a `PanicException` surfacing through the FFI boundary plus a `FetchGuard` dropped without `complete()`, rather than anything unsound, so this is a hardening suggestion rather than a merge blocker. Degrading gracefully costs one line and preserves the same observable behavior as an exhausted buffer:

```rust
let (rows, exhausted, has_next) = buffered_results
    .take_rows(limit)
    .unwrap_or_else(|| (Vec::new(), true, false));
```

If you would rather keep the invariant explicit, the alternative is to have the `Closing | Closed` arm of `close()` leave `buffered_results` alone and let the `Drop for PyAsyncCursor` path release it — but the `unwrap_or_else` is cheaper and keeps the memory-release fix from the earlier round intact.

## Nits

None on this head.

subrata-ms:
> One DB-API rowcount issue: `executemany()` ignores executions whose `rows_affected` is unknown and still returns the sum of the known rows. If one row reports unknown and another reports 1, the cursor reports 1 even though the total is indeterminate. Could any unknown per-row count keep the aggregate `rowcount` at -1?

Thanks for catching this issue @[Sylvester Kaczmarek](https://github.com/sylvesterkaczmarek). Updated ExecuteMany row-count aggregation so that any unknown per-execution count makes the final aggregate indeterminate and therefore reports DB-API rowcount == -1; known counts are no longer returned as a partial total. Added regression coverage for both orderings—unknown followed by known and known followed by unknown—and mutation-verified that restoring the previous behavior fails both cases. The full non-long-haul Python suite passes with 1165 tests.

mssql-py-core/src/async_execute.rs
```diff
@@ -85,14 +119,29 @@ fn should_replace_prepared_statement(
...
+    rows_affected
+        .and_then(|count| i64::try_from(count).ok())
+        .unwrap_or(-1)
+}
+
+fn accumulate_rowcount(total: &mut Option<i64>, count: i64) {
+    if count < 0 {
+        *total = None;
+    } else if let Some(total) = total {
+        *total = total.saturating_add(count);
```
David-Engel: **Suggestion — the new aggregation rule has no Rust unit test.**

`accumulate_rowcount` carries the whole contract you just changed: unknown latches, it never recovers once latched, and known counts add. Its sibling `rowcount_from_rows_affected` got `rows_affected_overflow_is_unknown` two rounds ago, but this one is only exercised through `test_executemany_unknown_rowcount_makes_aggregate_unknown`, which is `@pytest.mark.integration` and so only runs on the legs that have a live SQL Server. A pure unit test costs six lines and runs everywhere:

```rust
#[test]
fn unknown_rowcount_latches_the_aggregate() {
    let mut total = Some(0_i64);
    accumulate_rowcount(&mut total, 2);
    assert_eq!(total, Some(2));
    accumulate_rowcount(&mut total, -1);
    assert_eq!(total, None);
    accumulate_rowcount(&mut total, 5);
    assert_eq!(total, None, "a later known count must not revive the aggregate");
}
```

The third assertion is the one that matters — it is the ordering @sylvesterkaczmarek asked about, and it is the part an unwary refactor to `total = Some(count)` would break silently.

**Nit, same line — `saturating_add` reports an overflowed aggregate as known.** With `total == Some(i64::MAX - 1)` and `count == 5` this yields `Some(i64::MAX)`, which the caller then stores as a real row count. A single out-of-`i64` count maps to `-1` (unknown) via `rowcount_from_rows_affected`, so the two levels now disagree in the same way the drain loop and `execute_on_client` did before you unified them. `checked_add(count).map_or(None, Some)` — or simply `*total = total.checked_add(count)` on the `Some` arm — keeps them consistent. Unreachable in practice; raising it only because it is the exact class of inconsistency the earlier round fixed one level down.

mssql-py-core/src/async_cursor.rs
```diff
@@ -427,6 +489,30 @@ impl PyAsyncCursor {
...
 
+    /// Execute T-SQL once for each parameter row and return this cursor.
+    ///
+    /// Rows execute sequentially after the complete input iterable is validated.
+    /// Positional rows use `?`; mapping rows use `%(name)s`. A SQL error stops
+    /// execution and reports its zero-based parameter-row index. Earlier rows may
+    /// already be committed when autocommit is enabled; an explicit transaction
+    /// remains open for the caller to commit or roll back.
+    ///
+    /// DML row counts are aggregated. Row-producing results set `rowcount` to
```
David-Engel: **Suggestion — the docstring still describes the pre-`accumulate_rowcount` contract.**

"DML row counts are aggregated" was accurate when a partial sum was returned; now any execution that reports an unknown count drives the whole aggregate to `-1`. That is the behavior @sylvesterkaczmarek asked for and it is the right call, but it is exactly the kind of thing a caller writing `if cursor.rowcount != len(rows): raise` needs told. Something like:

```
/// DML row counts are aggregated, and the aggregate is `-1` when any
/// execution reports an unknown count. Row-producing results set
/// `rowcount` to `-1`, are buffered, and retain their boundaries for
/// `fetch*()` and `nextset()`.
```

**And a question on the granularity.** `accumulate_rowcount` is called per *statement result*, not per parameter row, so the latch fires on any `StatementResult::NoRows { rows_affected: None }` anywhere in the batch — not just on a DML statement whose count was suppressed. `mssql-tds` surfaces a countless DONE that way for `SET`, `PRINT` and DDL (`tds_client.rs` asserts `NoRows { rows_affected: None }` for the `PRINT`/assignment cases in `execute_surfaces_print_after_assignment_counts` and `execute_surfaces_message_but_skips_select_count`, and `take_dml_result_counts`'s doc names `SET NOCOUNT ON` and DDL as producing no qualifying count). So `executemany("SET NOCOUNT OFF; INSERT ...", rows)` or `executemany("CREATE TABLE ...; INSERT ...", rows)` now reports `-1` even though every DML statement reported a count.

Conservative is defensible and I am not asking you to change it — the driver cannot distinguish "this statement never reports a count" from "this DML statement's count was suppressed", and `execute()` has the same blind spot. Just worth deciding deliberately and saying so in the docstring, because the common single-statement `INSERT` case is unaffected and the difference will otherwise only show up in a multi-statement batch.

mssql-py-core/tests/test_async_executemany.py
```diff
@@ -0,0 +1,1516 @@
...
+                use_prepare=False,
+            )
+            await cursor.executemany(
+                "IF ? = 1 SET NOCOUNT ON ELSE SET NOCOUNT OFF; "
+                "INSERT INTO #async_executemany_unknown_count VALUES (?)",
+                [(no_count, value) for value, no_count in enumerate(no_count_values)],
+                use_prepare=False,
+            )
+
+            assert cursor.rowcount == -1
```
David-Engel, line 824: **Suggestion — this test may not pin the mechanism it names, and the two parametrizations may be the same case.**

The intent per your PR comment is "unknown followed by known" and "known followed by unknown". But the leading `IF ? = 1 SET NOCOUNT ON ELSE SET NOCOUNT OFF` is itself a statement whose DONE carries no count, and `mssql-tds` surfaces that as `StatementResult::NoRows { rows_affected: None }` → `rowcount_from_rows_affected` → `-1` → the latch fires. If that is what happens against a real server, then both `(True, False)` and `(False, True)` are just "the leading SET statement reported no count", the ordering coverage isn't exercised, and the assertion holds for a broader reason than the one in the test name. It still fails on the old code (the `NOCOUNT OFF` row's `INSERT` reported `1`, which the old branch would have returned), so it is a valid regression guard — it just may not be guarding the `NOCOUNT`-suppression path.

A control case settles it in one extra `executemany` and is worth having either way:

```python
await cursor.executemany(
    "SET NOCOUNT OFF; INSERT INTO #async_executemany_unknown_count VALUES (?)",
    [(10,), (11,)],
    use_prepare=False,
)
assert cursor.rowcount == 2  # fails if the SET statement is what latches the aggregate
```

If that assertion fails, the leading `SET` is doing the work and the `NOCOUNT`-suppressed-`INSERT` path needs its own shape (a single-statement operation against a session where `NOCOUNT` is already `ON`, set by a preceding `execute`).

Minor, same test: it asserts only `rowcount == -1` and never checks that the rows landed, so it cannot distinguish "aggregate is unknown" from "the batch did far less than expected" beyond the absence of an exception. A `SELECT COUNT(*)` after the `executemany` asserting `2` costs one statement and makes the `-1` mean something.

mssql-py-core/src/async_fetch.rs
```diff
@@ -65,6 +67,57 @@ impl FetchState {
...
+        self.0.lock().unwrap_or_else(|error| error.into_inner())
+    }
+}
+
+fn take_buffered_rows(
+    buffered_results: &BufferedResults,
+    limit: usize,
+) -> (Vec<PyRowWriter>, bool, bool) {
+    buffered_results
+        .take_rows(limit)
```
David-Engel: The fallback is right and it closes the panic window from the last round — `(Vec::new(), true, false)` is observationally identical to draining an already-empty buffered set, and `fetch_guard.complete(true, false)` releases the operation exactly as an ordinary exhausted fetch does.

**Nit only:** `cleared_buffer_is_treated_as_exhausted` pins this helper, not the call site — putting `.expect(...)` back in `fetch_buffered` leaves the test green, since the helper would still be compiled and tested in isolation. The helper being correct is the important half and is now covered, so I am not asking for more; just noting the call site is unpinned, same as the `yield_at_interval` note from an earlier round.

David-Engel (commented):

Re-reviewed at head `0ff5c0f` against merge base `4382290`, focused on the delta from `0c56fca` (the previously reviewed SHA) with a re-read of the full diff.

The delta is two changes and both are the right shape:

- **`take_buffered_rows`** replaces the `.expect("buffered fetch requires a current result set")` in `fetch_buffered` with `unwrap_or_else(|| (Vec::new(), true, false))`, which closes the panic window I raised last round (`cursor.fetchall()` awaitable created → `conn.close()` flips the lifecycle to `Closing` → `cursor.close()`'s `Closing | Closed` arm calls `clear_result_state()` with no claim → the spawned task reaches `take_rows` on an emptied deque). I re-walked the fallback: `fetch_guard.complete(true, false)` → `finish_fetch(id, true, false)` clears `active_operation` exactly as an ordinary exhausted fetch does, and the trailing `fetch_state.set(Exhausted)` on an already-closed cursor is unobservable because `fetch_resources()` rejects on `closed` first. `next_buffered_set` was already `Option`-safe through `advance()`, so the same race there degrades to `Ok(false)` rather than a panic. No remaining `expect`/`unwrap` on the buffered path.
- **`accumulate_rowcount`** answers @sylvesterkaczmarek's question correctly: `total: Option<i64>` latches to `None` on the first unknown count and never recovers, so a mixed batch reports `-1` instead of a partial sum. The empty-input case is unaffected — `batch_count == 0` still overrides `total_rows_affected` to `0` before the store, so the `Some(0)` seed does not change empty-`executemany` semantics.

Nothing blocking. Two suggestions and two nits, all on the new aggregation code.

## Verification

- Built `mssql-py-core` locally (it is excluded from the workspace, so run separately): `cargo fmt -- --check` clean, `cargo clippy --all-features --all-targets -- -D warnings` clean, `cargo nextest run --all-features` reports **231 passed, 1 skipped** — one more than the previous head, the new `cleared_buffer_is_treated_as_exhausted`.
- I did not run the Python integration suite (needs a maturin build plus a live SQL Server, and the `executemany` tests are `@pytest.mark.integration`). CI covers it better than I can here.
- **Required CI is green where it matters on this head.** `Build Stage Build Linux` now **passes** (37m13s) in build `172503`, which is the leg that failed the previous two heads on `test_executemany_exhausted_nextset_checks_connection_lifecycle` — that fix holds, and the new `test_executemany_unknown_rowcount_makes_aggregate_unknown` and the pre-existing `test_executemany_inserts_rows_and_aggregates_rowcount` both ran against a live server there, so the aggregation change does not break the known-count path. `Build mssql-python (cross-repo)`, the `mssql-python suite on mssql-odbc driver` leg, `Build Linux ARM`, `Build Windows`, `Build Windows ARM` and the Kerberos leg also pass. `Build MacOS` / `Test MacOS` / `Build mssql-python macOS` and the aggregate check are still **pending**; nothing is failing. ADO work item 47195 was not cross-checked (reviewer infrastructure, not your problem).

## Blocking

None.

## Suggestions

Both are on inline threads:

1. `accumulate_rowcount` is the one piece of the new rowcount contract with no Rust unit test, while its sibling `rowcount_from_rows_affected` has `rows_affected_overflow_is_unknown`. Six lines would pin the latch-and-never-recover rule so it runs everywhere instead of only on the live-server legs.
2. The `executemany` docstring still describes only the old rule ("DML row counts are aggregated"), and the new behavior is finer-grained than "one execution reported no count" — it latches on *any* statement result that carries no count, so a multi-statement operation containing a `SET` / `PRINT` / DDL statement now reports `rowcount == -1` even when every DML statement reported a count. Worth a docstring line, and worth confirming the granularity is what you intended.

## Nits

1. The aggregate uses `saturating_add`, so an aggregate that overflows `i64` is reported as a *known* `i64::MAX` — the same asymmetry you unified in `rowcount_from_rows_affected` two rounds ago, now reintroduced one level up. Unreachable in practice; noted for consistency.
2. `cleared_buffer_is_treated_as_exhausted` pins the helper, not the call site: restoring `.expect(...)` inside `fetch_buffered` leaves the test green. The helper is the important half and is now covered, so this is just a note that the call site is unpinned — same shape as the `yield_at_interval` note from an earlier round.

mssql-py-core/src/async_execute.rs
subrata-ms: Thanks for both suggestions. I added direct Rust unit coverage for the row-count aggregation contract: known counts accumulate, an unknown count latches the aggregate as unknown, and later known counts cannot revive it. I also replaced saturating addition with checked addition so an aggregate exceeding i64 is reported as unknown, consistent with individual out-of-range row counts. Added a separate overflow regression test as well. Both tests were mutation-verified, and formatting, strict Clippy, and the complete Rust suite pass.

mssql-py-core/src/async_cursor.rs
subrata-ms: Thanks for pointing this out. I updated the ExecuteMany docstring to document the new conservative aggregation behavior: the aggregate is -1 when any surfaced statement result has an unknown row count. I used “surfaced statement result” deliberately because countless intermediate DDL or SET boundaries may be collapsed by the protocol layer, while surfaced message-only or suppressed-count results make the aggregate indeterminate. Formatting, strict Clippy, and the complete Rust suite pass.

mssql-py-core/tests/test_async_executemany.py
subrata-ms: Thanks for the suggestion. I strengthened the integration test with a control ExecuteMany using SET NOCOUNT OFF; INSERT ... and assert that it reports rowcount == 2. This confirms that the leading countless SET boundary is collapsed and does not latch the aggregate, so the two parameterizations genuinely exercise unknown-then-known and known-then-unknown ordering. I also added a final SELECT COUNT(*) assertion to verify both target rows were inserted, making it clear that rowcount == -1 represents an indeterminate aggregate rather than incomplete execution. Both orderings pass against SQL Server, along with formatting, strict Clippy, and the complete Rust suite.

mssql-py-core/src/async_fetch.rs
subrata-ms: Thanks for noting the call-site gap. I removed it structurally by moving the cleared-buffer fallback into BufferedResults::take_rows() itself. The method now always returns (rows, exhausted, has_next), with a missing current buffer represented as (empty, true, false), and fetch_buffered() calls that method directly. The unit test therefore covers the exact production operation rather than a detached wrapper. Restoring the old .expect(...) at the call site now fails to compile because take_rows() no longer returns an Option. Formatting, strict Clippy, and the complete Rust suite pass.

David-Engel (commented):

Reviewed the full diff against `origin/main` (merge base `4382290`) at head `bbebaa89`, plus the incremental changes since the last automated pass on `0ff5c0fb`. Verified locally against `mssql-py-core`: `cargo clippy --all-features --all-targets -- -D warnings` is clean and `cargo nextest run --lib` passes 233/233, including the four new helper tests. Required CI is still running on this head; nothing has reported a failure.

The items raised in earlier rounds are all addressed on this SHA. `get_or_create_binding_plan` now builds `ParameterBindingPlan` once instead of per row, `CursorCleanup::run` and the `Closing`/`Closed` short-circuit both call `clear_result_state()`, the terminal buffered `nextset()` calls `ensure_open()` before short-circuiting, `fetch_state` becomes `TerminalNoRows` for a non-empty DML batch (with `assert await cursor.nextset() is False` pinning it), the yield intervals are named constants, and `accumulate_rowcount` now uses `checked_add` so an overflowing aggregate latches to unknown the same way `rowcount_from_rows_affected` does. I re-walked the buffered state machine across the zero-row, single-set, multi-set, partial-drain and skip-final-set paths and the `finish_fetch(exhausted, !exhausted || has_next)` transitions release the session exactly once in each.

I also confirmed the new `SET NOCOUNT OFF; INSERT` assertion should hold: `advance_to_result_boundary` collapses a DONE with neither a count nor a message (`tds_client.rs:4206`), so the `SET` statement never surfaces as its own `NoRows { rows_affected: None }` and cannot latch the aggregate to unknown.

Nothing blocking. Two suggestions and two nits, none of them previously raised.

## Suggestions

**1. Preflight errors other than `TypeError` lose the parameter-row index (`async_execute.rs:659`)**

```rust
.map_err(|error| {
    if error.is_instance_of::<PyTypeError>(py) {
        PyTypeError::new_err(format!("executemany parameter row {row_index}: {error}"))
    } else {
        error
    }
})
```

Only `TypeError` gets annotated. A missing mapping key raises `PyKeyError::new_err(source_name)` from `ParameterBindingPlan::bind_row`, and any non-`TypeError` raised by `rpc_parameter` passes through unchanged, so the caller sees a bare `KeyError: 'name'` with no indication of which row produced it. That is the one diagnostic `executemany` uniquely needs: the docstring promises "A SQL error stops execution and reports its zero-based parameter-row index", and the execution path delivers it via `"PyAsyncCursor.executemany failed while executing parameter row {}"`, but preflight only delivers it for one error class. On a 100k-row batch the difference is between a one-line fix and a bisect.

`test_executemany_named_parameter_error_and_extra_key_parity:402` currently pins the un-annotated form (`pytest.raises(KeyError, match="name")`), so it would need updating alongside. Annotating every preflight error while preserving its type — rather than re-raising everything as `PyTypeError` — keeps `except KeyError` working for callers.

**2. `ParameterBindingPlan::bind_row` duplicates the `bind_named` / `bind_positional` contract, and only the copy is untested in Rust (`async_parameters.rs:400`)**

`bind_row` re-implements, line for line, the same four rules as the two existing binders: the "SQL contains no parameter markers, but N parameters were supplied" guard, the "SQL contains N parameter markers, but M parameters were supplied" guard, `PyKeyError` on a missing dict key, and `validate_hint_count`. The two copies already differ in a small way — `bind_named` validates the no-marker case before `validate_hint_count`, `bind_row` validates hints first — so a bad-hint-count call against marker-less SQL now reports a different error depending on which entry point you used. That is harmless today, but it is exactly the kind of drift that duplication produces.

`bind_positional` and `bind_named` are `ParameterBindingPlan::new(...)` followed by `bind_row(...)`, so reimplementing them as thin wrappers would collapse the duplication and give the shared path the coverage the old ones already have.

Failing that, `bind_row` needs Rust unit tests. `async_parameters.rs` has none for it, and the only `ParameterBindingPlan` unit test — `executemany_reuses_parameter_binding_plan` — asserts pointer identity, not binding behavior. Everything else that exercises `bind_row` is one of the 42 `@pytest.mark.integration` tests in `test_async_executemany.py`, which need a live SQL Server. `async_parameters.rs` already has a Python-only test pattern that could cover the marker-count, mixed-shape and missing-key cases without a server.

## Nits

**3. The terminal buffered `nextset()` mutates state before the awaitable exists and does not restore it on failure (`async_fetch.rs:813-818`)**

```rust
session_state.ensure_open().map_err(map_claim_error)?;
buffered_results.advance();
description_state.replace(None);
rowcount.store(-1, Ordering::Release);
return pyo3_async_runtimes::tokio::future_into_py(py, async { Ok(false) });
```

If `future_into_py` fails — the "no running event loop" case that `test_executemany_buffered_awaitable_creation_failure_restores_fetch_state` deliberately provokes — the front set has already been popped and `description` cleared, but the caller gets a `RuntimeError` rather than `False`. The claim branch twelve lines below goes to the trouble of restoring `previous_fetch_status` and `previous_description` in exactly this situation, and `fetch_buffered`'s `Exhausted` early return avoids the problem entirely by mutating nothing before the awaitable is built. The observable damage is small (a retry falls through to the streaming path and still returns `False`), but the asymmetry is worth closing, and that test only covers the non-terminal branch.

**4. `nextset()` re-derives the rows-affected conversion instead of sharing `rowcount_from_rows_affected` (`async_fetch.rs:737-741`)**

```rust
StatementResult::NoRows { rows_affected } => rows_affected
    .and_then(|count| i64::try_from(count).ok())
    .unwrap_or(-1),
```

That is `rowcount_from_rows_affected` from `async_execute.rs:135`, which has its own unit test (`rows_affected_overflow_is_unknown`) pinning the `u64`-overflow-means-unknown rule. Two copies of a rule that is only tested in one place will eventually disagree; making the helper `pub(crate)` and calling it here keeps the contract in one spot.

sumitmsft (approved):

PR #498, opened 2026-09-04
Match variant metadata and fix descriptor test corruption

Theekshna:
## Description

Match `sql_variant` metadata with msodbcsql:

- Report legacy `SQL_C_DATE` / `SQL_C_TIMESTAMP` codes (9/11) for variant `date` and datetime-family values at every declared ODBC version. This changes the previous ODBC 3.8 result from the ODBC 3 codes 91/93 to 9/11.
- Report `SQL_C_SS_TIME2` / `SQL_C_SS_TIMESTAMPOFFSET` for variant `time` and `datetimeoffset` only under ODBC 3.8; ODBC 2 and 3 receive `SQL_C_BINARY`.
- Cover date, smalldatetime, datetime, datetime2, time, and datetimeoffset through connected ODBC 2, 3, and 3.8 fixtures.

Also complete [AB#47811](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47811) by fixing both wrong-width `SQL_DESC_OCTET_LENGTH` reads: the C++ E2E helper and the Rust unit-test helper now use `SQLLEN` / `SqlLen` instead of `SQLSMALLINT` / `SqlSmallInt`. This prevents an 8-byte descriptor write into a 2-byte stack slot. The remaining descriptor helper reads were audited for other wide fields.

`SetDescRecLiveTest.EquivalentToSetDescFieldSequence` was guarded by a `GTEST_SKIP()` for the crash this fix removes. Because `GTEST_SKIP()` returns before the test body runs, the corrected `GetLen` reads never executed and the test reported SKIPPED. The skip and its now-stale comment are removed, so the test actually exercises the fixed read; the E2E comparison run on this branch is the first to do so against a live server.

Pre-3.8 binary metadata does not implement buffered binary retrieval, which remains tracked by [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239).

## Related Issues

- Fixes https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47830
- Fixes https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47811

## Validation

- `cargo bfmt`
- `cargo bclippy`
- `cargo nextest run -p mssqlodbc --all-targets --no-fail-fast` (1296 passed, 1 skipped before the final test-helper-only fix)
- `cargo nextest run -p mssqlodbc -E 'test(writes_all_eight_fields)'` after the Rust `SqlLen` fix
- Windows E2E comparison passed against this branch's mssql-odbc driver and retail msodbcsql 18.6.2.1, including connected ODBC 2/3/3.8 temporal variant coverage. That run predates removing the `GTEST_SKIP()`, so `EquivalentToSetDescFieldSequence` was reported SKIPPED rather than passing; the comparison is being re-run now that the test body actually executes.
- Full `cargo btest`: 3503 passed, 399 SQL-backed tests failed because `SQL_PASSWORD` is unavailable

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


mssql-odbc/src/api/col_attribute.rs
```diff
@@ -541,13 +565,13 @@ fn variant_c_type(base: TdsDataType) -> SqlSmallInt {
...
         | TdsDataType::Money4
         | TdsDataType::MoneyN => SQL_C_CHAR,
-        TdsDataType::DateN => SQL_C_TYPE_DATE,
-        TdsDataType::TimeN => SQL_C_SS_TIME2,
-        TdsDataType::DateTime | TdsDataType::DateTim4 | TdsDataType::DateTimeN => {
-            SQL_C_TYPE_TIMESTAMP
-        }
-        TdsDataType::DateTime2N => SQL_C_TYPE_TIMESTAMP,
-        TdsDataType::DateTimeOffsetN => SQL_C_SS_TIMESTAMPOFFSET,
+        TdsDataType::DateN => SQL_C_DATE,
```
David-Engel, line 584: The `SQL_C_SS_*` gate below is cited (`IS351ORLESSAPP`, `sqlcdesc.cpp:6474`), but this arm and the `SQL_C_TIMESTAMP` arms are the ones that look wrong at a glance: they report the ODBC 2.x codes `9`/`11` at *every* declared version, including 3.8, where this crate previously reported `91`/`93`. That justification currently lives only in the PR description. A short comment citing the msodbcsql source that produces `9`/`11` here would keep a future reader from "correcting" this back to `SQL_C_TYPE_DATE`.

Related: the doc comment on `uses_3_80_variant_types` says earlier versions "use the binary fallback", which reads as covering all temporal variants — it only covers `time` and `datetimeoffset`.

David-Engel (commented):

This came from an unattended run, so nothing below was checked by a human first.

## Summary

The change makes `SQL_CA_SS_VARIANT_TYPE` version-aware for temporal `sql_variant` values and fixes two wrong-width `SQL_DESC_OCTET_LENGTH` reads in test helpers. The implementation matches the shape this crate already uses for version-dependent behavior, the new tests are non-vacuous, and I found nothing blocking. One suggestion about sourcing the parity claim, below.

## Verification

Reviewed the full `merge-base..HEAD` diff (`e64242c..2184095`, 5 files, +227/-42) plus the surrounding code, then:

- **Retrieval path consistency.** The most load-bearing question here is whether an application that reads `9`/`11` back from this attribute can then use it. It can: `get_data.rs:169` validates through `canonical_c_type`, which folds `SQL_C_DATE..=SQL_C_TIMESTAMP` onto the `SQL_C_TYPE_*` spellings (`type_rules.rs:66-72`), and `fetch_convert.rs:503-509,695,745` accepts both spellings directly. `bind_col.rs:663-665` pins the same fold. So the new metadata does not point callers at a target type the driver rejects.
- **Single mapping site.** `SQL_CA_SS_VARIANT_TYPE` is only answered in `col_attribute.rs`; there is no second `SQLGetDescField`/IRD path left on the old mapping.
- **Lock ordering.** The new env read matches the established pattern in `bind_param.rs:141-148`, `exec_common.rs:511-518`, `fetch_scroll.rs:817-824` and `get_type_info.rs:104-111` — parent lock taken and released before the stmt lock, guard dropped at the end of the block. No lock is held across the stmt acquisition.
- **E2E fixtures.** `ODBCTest::SetUp` (`odbc_test_fixture.cpp:13-25`) only allocates the ENV and sets the version, so the new `ColAttributeOdbc2LiveTest` / `ColAttributeOdbc3LiveTest` override before any connection handle exists — valid against a real Driver Manager. The `SQL_C_SS_TIME2` / `SQL_C_SS_TIMESTAMPOFFSET` fallback `#define`s (`0x4000`/`0x4001`) match `odbc_types.rs:425-426`.
- **Mutation check.** Forcing `variant_uses_3_80_types` to `true` unconditionally makes `temporal_variant_type_honors_the_declared_odbc_version` fail (`left: 16384, right: -2`), so the test guards the plumbing and not just the pure helper. Reverted afterwards; the worktree is clean.
- **Ran locally** (`CARGO_TARGET_DIR` redirected, worktree only): `cargo nextest run -p mssqlodbc --lib -E 'test(variant_c_type) + test(variant_type) + test(temporal_variant) + test(extended_variant_types) + test(writes_all_eight_fields)'` → 7 passed; `cargo clippy -p mssqlodbc --frozen --all-features --all-targets -- -D warnings` → clean.
- **Not verified:** the msodbcsql comparison legs. I have no msodbcsql source or a live server in this run, so the parity claims for the E2E cases rest on your reported comparison run, not on anything I re-derived.

## Blocking

None found.

## Suggestion

**`col_attribute.rs:568` — the surprising half of the parity change has no source citation in the code.**

`uses_3_80_variant_types` carries a good citation (`IS351ORLESSAPP`, `sqlcdesc.cpp:6474`) for the `SQL_C_SS_*` gate, but that covers the half a reader would guess. The unguarded half — `DateN` → `SQL_C_DATE` and the datetime family → `SQL_C_TIMESTAMP` at *every* declared version, including 3.8, where this crate previously reported `91`/`93` — is the part that looks wrong at a glance, and its justification lives only in the PR description. Per the repo's own review guidance, msodbcsql is the contract and parity assertions should cite the file and line. A one-line comment on those arms pointing at whatever msodbcsql code produces `9`/`11` here would stop the next reader (or the next reviewer) from "fixing" it back to `SQL_C_TYPE_DATE`.

Secondary, same function: the doc comment's "earlier or unset versions use the binary fallback" reads as if it governs all temporal variants, when it only governs `time` and `datetimeoffset`. Worth narrowing the wording now that `date`/`datetime` follow a different, version-independent rule.

## Nits

None.

## CI

`gh pr checks 498` is red, but both failures trace to macOS agent infrastructure rather than this diff. From the ADO timeline for build 172511: `Install and start Colima-based Docker` → *"The task has timed out"*, after which `Save code coverage in Cobertura format` exits 101 and `Publish Cobertura coverage artifact` cannot find `target/cobertura.xml`. The GitHub `coverage-report` job then fails at its only failing step, `Download Cobertura coverage artifact`. Every build job, the Kerberos job and the cross-repo mssql-python suite passed. Worth a re-run to get a green record before this merges; I did not re-run anything.

mssql-odbc/src/api/col_attribute.rs
Theekshna: Addressed in bd828473. I added the `GetIRDField` citation for the version-independent `SQL_C_DATE` / `SQL_C_TIMESTAMP` mappings (`sqlcdesc.cpp:6462-6469`) and narrowed the helper documentation to say the binary fallback applies only to variant `time` and `datetimeoffset`. Verified with `cargo bfmt`.

David-Engel (commented):

Re-review of `bd82847` (the previous automated pass covered `2184095`). Nothing below was checked by a human first.

## Summary

The only change since the last pass is `bd82847`, which is comment-and-doc-only: it adds the `GetIRDField` citation (`sqlcdesc.cpp:6462-6469`) on the version-independent `SQL_C_DATE` / `SQL_C_TIMESTAMP` arms and narrows the `uses_3_80_variant_types` doc to say the binary fallback covers only variant `time` and `datetimeoffset`. That addresses the one suggestion from the previous pass, and I have no new findings.

## Verification

Re-reviewed the full `merge-base..HEAD` diff (`e64242c..bd82847`, 5 files, +230/-42) and re-ran the checks that mattered, plus two things the previous pass had left open.

- **Downstream consumer of the changed codes.** The load-bearing risk in switching variant `date`/`datetime` from `91`/`93` to `9`/`11` is the primary consumer. `MapVariantCTypeToSQLType` in `mssql-python`'s `ddbc_bindings.cpp:3263` already accepts both spellings — `case SQL_C_DATE: case SQL_C_TYPE_DATE:` and `case SQL_C_TIMESTAMP: case SQL_C_TYPE_TIMESTAMP:` — so the new codes land on the existing branches, and the presence of those legacy cases is itself corroboration that msodbcsql returns `9`/`11` here. The cross-repo `mssql-python suite on mssql-odbc driver` job passes on this commit.
- **Not a retrieval dead end at ODBC 2/3.** Variant `time`/`datetimeoffset` now report `SQL_C_BINARY` below 3.8, where binary delivery is still `HYC00` (Task 47239, documented in the diff). That is metadata-only: `get_data.rs:679` still serves `(ColumnValues::Time, SQL_C_SS_TIME2)` with no version gate, matching msodbcsql, where `rgbTRANSTYPE380` gates only the *default* resolution and not an explicit ask. An ODBC 3 application that names the type still gets its value.
- **Descriptor-width audit holds.** The remaining `GetSmallInt` reads in `set_desc_rec_test.cpp` (`SQL_DESC_COUNT`, `SQL_DESC_CONCISE_TYPE`, `SQL_DESC_DATETIME_INTERVAL_CODE`) and the remaining `get_small_int` reads in `set_desc_rec.rs` are all genuine `SQLSMALLINT` fields. `SQL_DESC_OCTET_LENGTH` was the only wide one, and both legs now use `SQLLEN` / `SqlLen`.
- **Single dispatch point.** `SQL_CA_SS_VARIANT_TYPE` is answered only in `col_attribute.rs`, and `exports.rs:790` exports only `SQLColAttributeW` — the Driver Manager funnels the ANSI and ODBC 2 `SQLColAttributes` spellings into it, so there is no second path on the old mapping.
- **Probe helper matches the house pattern.** `ExpectVariantType` uses the same `SQLGetData(col, SQL_C_BINARY, &probe, 0, &indicator)` shape as the existing variant tests at lines 512/519/580/602, so it is not exercising a different code path than the tests it sits next to. The ODBC 2/3 fixtures override `SQL_ATTR_ODBC_VERSION` after `ODBCTest::SetUp` sets 3.80 but before any DBC exists, which is legal; `ColAttributeLiveTest` inherits the 3.80 default, so the 3.8 leg is meaningful.
- **Ran locally** (worktree only, `CARGO_TARGET_DIR` redirected): `cargo nextest run -p mssqlodbc --lib -E 'test(variant_c_type) + test(variant_type) + test(temporal_variant) + test(extended_variant_types) + test(writes_all_eight_fields)'` → 7 passed; `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` → clean; `cargo fmt -- --check` → clean.
- **Not verified:** the msodbcsql comparison legs. Without msodbcsql source or a live server in this run, the `sqlcdesc.cpp` citations rest on your reported comparison run rather than anything I re-derived — though the `mssql-python` switch above is independent corroboration of the `9`/`11` half.

## Blocking

None found.

## Suggestions

None.

## Nits

None.

## CI

`gh pr checks 498` is red on `Build Stage Test MacOS`, and it is the same macOS agent infrastructure failure as the previous run, not the diff. The ADO timeline for build 172525 shows `Install and start Colima-based Docker` → *"The task has timed out"*, after which both `Save code coverage` steps exit 101 and `Publish Cobertura coverage artifact` cannot find `target/cobertura.xml`. Every build job, Kerberos, and the cross-repo `mssql-python` suite passed. Because required CI is not green I have not applied the `ready for human review` label; a re-run to get a clean record would be worth it before merge. I did not re-run anything.

saurabh500 (commented):
_Unattended review sweep — GitHub Copilot on behalf of @saurabh500. Nothing below was checked by a human first, and this is **not an approval**; push back on anything that looks wrong._

## Verdict

No blocking findings. The temporal `sql_variant` metadata now matches msodbcsql exactly, and the descriptor-width test-helper fixes are correct. This is **metadata parity**, not end-to-end retrieval — see the caveat below. Recommend a human approve once CI is green (the current red leg is macOS Colima/Docker infrastructure, not this diff).

## What I verified

Reviewed the full `e64242c..bd82847` diff (5 files) in a worktree and checked the parity claim against the msodbcsql reference source rather than the ODBC spec, per this repo's review guidance. The contract is `GetIRDField`'s `SQL_CA_SS_VARIANT_TYPE` arm (`Sql/Ntdbms/sqlncli/odbc/sqlcdesc.cpp:6454-6488`):

- `SQL_NATIVE_DATE` → `SQL_C_DATE` (9), every version
- `SQL_NATIVE_DATETIME` / `SQL_NATIVE_SMALLDATETIME` / `SQL_NATIVE_LEGACY_DATETIME` → `SQL_C_TIMESTAMP` (11), every version
- `SQL_TIME2_MAPPED` → `IS351ORLESSAPP ? SQL_C_BINARY : SQL_C_SS_TIME2`
- `SQL_TIMESTAMPOFFSET_MAPPED` → `IS351ORLESSAPP ? SQL_C_BINARY : SQL_C_SS_TIMESTAMPOFFSET`

`SetKnownInformation` (`sqlcdata.cpp:1846-1885`) assigns those `fColCType` values from a variant's inner TDS type, and SQL Server `time`/`datetimeoffset` classify as `SQL_TIME2_MAPPED`/`SQL_TIMESTAMPOFFSET_MAPPED` — not the legacy `SQL_NATIVE_TIME` (confirmed by the comment at `sqlcprot.h:2174`: "TDS type is SQLTIMEN ... receives c type as SQL_TIME2_MAPPED"). So every new arm in `variant_c_type` lines up one-for-one with the reference, and modelling only `Odbc3_80` as "extended" is equivalent to the driver's `IS351ORLESSAPP` gate across the four `OdbcVersion` values this crate carries. `Unset → SQL_C_BINARY` is a safe defensive fallback (a DM-created statement is connected with a declared version before a result set exists).

- **Non-vacuous unit test.** The 3.80 leg is pinned by `with_env` setting `SQL_OV_ODBC3_80` (`test_support.rs:58-68`) and the ODBC 3 leg by the explicit env mutation, so forcing `uses_3_80_variant_types` true breaks the second assertion (`SQL_C_SS_TIME2` vs `SQL_C_BINARY`). Confirmed by reading the assertions against the helper defaults.
- **Lock ordering / poison handling** on the new env read matches the established `bind_param.rs:141-148` pattern (parent lock taken and released before the stmt lock; `error!` + `SQL_ERROR` with no diag on a poisoned env mutex).
- **Descriptor fix is real.** `SQL_DESC_OCTET_LENGTH` is a `SQLLEN` (pointer-width — 8 bytes on the 64-bit targets), so reading it into a 2-byte `SQLSMALLINT` / `SqlSmallInt` slot let the driver's wide write overrun the caller's stack. `get_len` / `GetLen` fix both comparison legs. There is nothing to mutation-test here — deliberately reproducing the undersized write is itself the UB being removed.

**Not verified:** no live SQL Server or msodbcsql binary in this run, so the connected E2E legs rest on the author's reported comparison against retail msodbcsql 18.6.2.1, not on anything I re-ran.

## Caveat (not a finding — already disclosed by the PR)

For ODBC 2/3 this now advertises `SQL_C_BINARY` for `time`/`datetimeoffset` variants, but a real (non-probe) binary read of a variant is still `HYC00` today (`get_data.rs:957-971`: `SQL_C_BINARY` with a non-zero buffer is not a deliverable target). That retrieval gap is pre-existing, tracked by AB#47239, and called out in `typed-columnar-fetch-plan.md` in this same PR — so the change is correct metadata parity, just don't read it as making those values retrievable end-to-end on pre-3.8.

## Blocking

None.

## Suggestion

None new. The earlier request to cite the msodbcsql source for the version-independent `9`/`11` arms was addressed in bd82847 — the `GetIRDField` citation (`sqlcdesc.cpp:6462-6469`) on the date/datetime arms and the narrowed `uses_3_80_variant_types` doc comment both match the reference source I read.

## Nit

None.

mssql-odbc/tests/e2e/tests/set_desc_rec_test.cpp
```diff
@@ -166,8 +173,8 @@ TEST_F(SetDescRecLiveTest, EquivalentToSetDescFieldSequence) {
 
     EXPECT_EQ(GetSmallInt(via_rec, 1, SQL_DESC_CONCISE_TYPE),
               GetSmallInt(via_field, 1, SQL_DESC_CONCISE_TYPE));
-    EXPECT_EQ(GetSmallInt(via_rec, 1, SQL_DESC_OCTET_LENGTH),
-              GetSmallInt(via_field, 1, SQL_DESC_OCTET_LENGTH));
+    EXPECT_EQ(GetLen(via_rec, 1, SQL_DESC_OCTET_LENGTH),
+              GetLen(via_field, 1, SQL_DESC_OCTET_LENGTH));
```
David-Engel, line 174: Blocking: these fixed `GetLen` reads never run. Line 131 of this file still has `GTEST_SKIP() << "crashes on a wrong-width SQL_DESC_OCTET_LENGTH read - AB#47811";` at the top of the test body, and `GTEST_SKIP()` returns immediately — so this whole body is dead code and the test reports SKIPPED rather than PASSED.

Re-enabling this test is the deliverable of AB#47811, and the PR description claims it "now passes on both comparison legs", which cannot be true while the skip stands. The skip's comment (lines 126-131) is also stale now: it describes the crash this PR just fixed.

Please delete lines 126-131 and re-run the Windows E2E comparison, since that is the first run that actually exercises the fixed read against a live server.

David-Engel (commented):

Re-review at `249cb6a` (previous automated passes covered `2184095` and `bd82847`, both pre-rebase). Nothing below was checked by a human first.

## Summary

The `sql_variant` metadata work looks right. Reading `SQL_ATTR_ODBC_VERSION` off the parent ENV before taking the STMT lock never holds both mutexes at once, so it cannot deadlock against any existing ordering, and `parent_dbc()`/`parent_env()` are the established, safety-documented accessors. The env read is correctly scoped to `SQL_CA_SS_VARIANT_TYPE` only, so it costs nothing on the ordinary attribute paths. `Unset` falling into the binary branch matches `IS351ORLESSAPP`, and the added case pins it.

On the `9`/`11` question: returning the deprecated `SQL_C_DATE` / `SQL_C_TIMESTAMP` spellings is safe for round-tripping, because every C-type entry point (`SQLGetData`, `SQLBindCol`, `SQLBindParameter`, `SQLSetDescField`) runs the value through `type_rules::canonical_c_type` before `is_valid_c_type` sees it. An application that feeds the attribute result straight back into `SQLGetData` still works.

The descriptor fix itself is correct in both places. `SQLGetDescField` ignores `BufferLength` for fixed-size attributes, so reading `SQL_DESC_OCTET_LENGTH` — a `SQLLEN`/`SqlLen` field — into a 2-byte slot really was an 8-byte write into 2 bytes, and the new `GetLen` / `get_len` helpers are the right shape. I audited the sibling descriptor reads and found no other narrow-slot cases: every remaining `GetSmallInt` / `get_small_int` call targets `SQL_DESC_COUNT`, `SQL_DESC_CONCISE_TYPE`, or `SQL_DESC_DATETIME_INTERVAL_CODE`, which are genuinely `SQLSMALLINT`.

One blocking issue: the E2E half of AB#47811 is not actually finished.

## Verification

Reviewed the full PR diff (5 files, +237/-102) in an isolated worktree at `249cb6a`, plus `type_rules.rs`, `odbc_types.rs`, `handles/stmt.rs`, `handles/dbc.rs`, `get_data.rs`, and the E2E fixture harness. Ran `cargo clippy -p mssqlodbc --all-targets --all-features` (clean) and `cargo nextest run -p mssqlodbc --all-features -E 'test(col_attribute) or test(variant) or test(desc_rec)'` (52 passed, 0 failed), which covers `variant_c_type_covers_the_base_types`, `temporal_variant_type_honors_the_declared_odbc_version`, `only_odbc_3_80_uses_extended_variant_types`, and `writes_all_eight_fields`. No SQL Server was reachable, so the connected ODBC 2/3/3.8 fixtures were read rather than run. Grepped the E2E harness to confirm `GTEST_SKIP` is not redefined anywhere.

## Findings

### Blocking

**`SetDescRecLiveTest.EquivalentToSetDescFieldSequence` is still skipped, so the E2E fix is unreachable.** `mssql-odbc/tests/e2e/tests/set_desc_rec_test.cpp:131` still carries `GTEST_SKIP() << "crashes on a wrong-width SQL_DESC_OCTET_LENGTH read - AB#47811";` at the top of the test body. `GTEST_SKIP()` returns immediately, so everything after it — including the new `GetLen` reads this PR adds at lines 176-177 — is dead code that never executes. The test reports SKIPPED, not PASSED.

That makes the PR description inaccurate on two points: "The previously mis-typed read in `SetDescRecLiveTest.EquivalentToSetDescFieldSequence` now passes on both comparison legs" cannot be true while the skip is in place, and "complete AB#47811" is not met, since the work item asks specifically to re-enable this test. The skip's own comment is now stale too: it explains a crash that this PR has fixed.

The fix is to delete the comment block at lines 126-131. Please re-run the Windows E2E comparison afterward and confirm the test genuinely passes on both legs, since removing the skip is what first exercises the fixed read against a live server, and the original comment notes that a crash here takes down every descriptor test after it.

The Rust-side half (`set_desc_rec.rs`) is complete and correct — this is only about the C++ E2E side.

## CI

`mssql-rs Pull request validation (Build mssql-python Build mssql-python macOS (cross-repo))` is failing at this head. It looks unrelated to this diff (cross-repo macOS build), but worth confirming before merge. Coverage and the overall validation stage are still pending.

mssql-odbc/tests/e2e/tests/set_desc_rec_test.cpp
Theekshna: You're right, and thanks for catching it — the `GTEST_SKIP()` returned before the body ran, so the fixed `GetLen` reads at lines 176-177 never executed and the test reported SKIPPED rather than PASSED.

Removed in 3324de2a: deleted the skip and the stale comment describing the crash this PR fixes, and dropped the matching `(disabled - AB#47811)` note from the file's header index at line 17, which was stale for the same reason.

I verified the body is now safe to run before re-enabling it: `SQL_DESC_OCTET_LENGTH` (lines 176-177) is the only `SQLLEN`-width field read in the test and it goes through `GetLen`, which uses an 8-byte `SQLLEN` slot; the two `SQL_DESC_CONCISE_TYPE` reads at 174-175 are genuinely `SQLSMALLINT` and correctly stay on `GetSmallInt`. So there's no remaining wrong-width read to crash the binary.

The Windows E2E comparison is running now against this push — that's the first run to actually exercise the read against a live server on both legs, so it's the real verification. I also corrected the PR description's "now passes on both comparison legs" claim, which as you noted could not have been true while the skip stood.

> [!NOTE]

David-Engel (commented):

Re-review at `3324de2` (earlier automated passes covered `2184095`, `bd82847`, and `249cb6a`). Nothing below was checked by a human first.

## Summary

The blocking finding from the `249cb6a` pass is fixed. `GTEST_SKIP()` is gone from `SetDescRecLiveTest.EquivalentToSetDescFieldSequence`, the stale crash comment has been replaced with a note explaining why `GetLen` is required, and the summary header no longer marks variation 6 as disabled. The `GetLen` / `get_len` helpers are the right shape on both the C++ and Rust sides, and the version-gated `variant_c_type` work still looks correct.

One new blocking finding: the parity ledger this PR edits now contradicts the code it ships.

## Verification

Reviewed the full diff (5 files, +241/-109) in an isolated worktree at `3324de2`, plus the E2E fixture (`odbc_test_fixture.cpp`), `odbc_types.rs`, and the surrounding descriptor and variant tests. Confirmed `ODBCTest::SetUp` only allocates the ENV and sets `SQL_OV_ODBC3_80`, so the new `ColAttributeOdbc2LiveTest` / `ColAttributeOdbc3LiveTest` fixtures legitimately override the version before `Connect()`. Ran `cargo clippy -p mssqlodbc --all-targets --all-features` (clean) and `cargo nextest run -p mssqlodbc --all-features -E 'test(variant) or test(desc_rec) or test(col_attribute)'` (52 passed, 0 failed). No SQL Server was reachable, so the live ODBC 2/3/3.8 fixtures and the re-enabled descriptor test were read rather than executed.

## Findings

### Blocking

**`docs/typed-columnar-fetch-plan.md` rows 109 and 110 now describe the opposite of what this PR ships, and one of them cites a test the PR deletes.** The PR updates the narrative further down the same file but leaves the parity table untouched:

- Row 109 (`SQL_CA_SS_VARIANT_TYPE` for `date` / `smalldatetime` / `datetime` / `datetime2`) still records the answer as a **deliberate divergence**: "msodbcsql `SQL_C_DATE` (9) / `SQL_C_TIMESTAMP` (11), mssql-odbc `SQL_C_TYPE_DATE` (91) / `SQL_C_TYPE_TIMESTAMP` (93)". This PR changes the driver to answer 9/11, so the row is now inverted — a reader consulting the ledger would conclude the new behavior is a regression away from a documented decision. The row also says it is "Anchored by `VariantDateTimeBaseTypesUseTheThreeXSpellings`", which this PR removes from `col_attribute_test.cpp`, leaving a dangling anchor.
- Row 110 (`time` / `datetimeoffset` under `SQL_OV_ODBC3`) is still marked **Gap — Task 47830** and states that "`variant_c_type` cannot currently do better — `col_attribute.rs` has no access to the environment's `OdbcVersion`". That is precisely the constraint this PR removes, and the row's closing rationale for excluding these types from `VariantBaseTypesMatchMsodbcsql` no longer holds either, since the new `Odbc2/Odbc3/Odbc38TemporalVariantTypes` tests do assert them per declared version.

Both rows should move to parity and re-anchor on the new tests, and AB#47830 looks like it can be closed by this change. Since these are the only two rows that gate the temporal variant behavior, leaving them stale is the difference between the ledger documenting the driver and the ledger misdocumenting it.

### Nits

**Whitespace noise in `col_attribute_test.cpp`.** `git diff --check` flags a new blank line at EOF (line 750), left behind by the deleted `VariantDateTimeBaseTypesUseTheThreeXSpellings`. Separately, the blank line between the close of `VariantExactNumericsReportNumeric` and the comment block introducing `VariantDecimalStillDeliversAsCharacter` was removed, so the comment now abuts the previous test's closing brace — the rest of the file keeps that separation.

**The summary header list no longer matches file order.** Entries 24–26 (`Odbc2/Odbc3/Odbc38TemporalVariantTypes`) are listed after 23 `ClrUdtDescriptorFields`, but in the file the three tests sit immediately after `VariantUnderlyingTypeAfterProbe` (entry 21), ahead of `VariantTypeBeforeProbeIsSequenceError` and `ClrUdtDescriptorFields`. The list has been a reliable index of this file; worth keeping it in step.

## CI

No failing checks at this head. `coverage-report`, the macOS build/test legs, and the cross-repo mssql-python macOS build are still in progress, so the run is not green yet either.

Theekshna:
Thanks — all three findings from the `3324de2` pass are addressed. Blocking one in `35b0560`, nits in `b2ac1e5`.

> **Blocking:** `docs/typed-columnar-fetch-plan.md` rows 109 and 110 now describe the opposite of what this PR ships, and one of them cites a test the PR deletes.

Correct, and it got worse before it got better: `main` moved under this branch while it sat, so once I merged, row 109 was both inverted *and* pointing at `VariantDateTimeBaseTypesUseTheThreeXSpellings`, which no longer exists anywhere in the tree. Fixed in the merge commit:

- **Row 109** now reads parity, credits AB#47830, and cites `GetIRDField` at `sqlcdesc.cpp:6462-6469` as the reason the legacy `9`/`11` codes are answered at every declared version. It keeps the part worth keeping — that the asymmetry is msodbcsql's own, since the adjacent `SQL_CA_SS_VARIANT_SQL_TYPE` case *does* apply the `SQL_DATE` → `SQL_TYPE_DATE` fix-up — so the next reader has the citation rather than a bare assertion.
- **Row 110** moves off Gap/AB#47830 and now records the `IS351ORLESSAPP` gate (`sqlcdesc.cpp:6474`). I left one genuine divergence documented rather than claiming clean parity: msodbcsql snapshots the version into `wStatus` at `SQLAllocHandle(SQL_HANDLE_DBC)`, while this driver reads the environment live per call. Equivalent through a Driver Manager, which forbids changing the version once a connection exists; observably different only for a caller that loads the driver directly and re-declares mid-connection.

Both re-anchor on `Odbc2/Odbc3/Odbc38TemporalVariantTypes`.

> **Whitespace noise in `col_attribute_test.cpp`.**

Both fixed; `git diff --check` against the merge base is clean.

> **The summary header list no longer matches file order.**

The three temporal entries are now 22–24, immediately after `VariantUnderlyingTypeAfterProbe`, with `VariantTypeBeforeProbeIsSequenceError` and `ClrUdtDescriptorFields` shifted down.

One thing I deliberately did **not** fix, so it is on the record rather than silently left: `ClrUdtDescriptorFields` is still out of order — listed 27th but sitting 7th in the file. That drift is pre-existing on `main` (listed 23rd there, same physical position), not something this PR introduced, and correcting it renumbers the entire list. Same for three tests `main`'s index omits outright: `DatetimeSubtypeAccompaniesTheVerboseType`, `NullVariantDoesNotDisturbTheFollowingColumn` and `VariantTypeIsPerColumn`. I did add the index entry for `EmptyVariantProbeConsumesValueButKeepsBaseType`, since #482 landed it without one and the merge put it adjacent to this PR's tests. Verified everything else now matches file order.

Two items for @Theekshna, both outside what I can do here:

1. **AB#47830** — agreed it looks closeable by this change, but I can't close ADO work items.
2. **The E2E comparison run still hasn't covered this.** Worth being blunt: no run so far has exercised the re-enabled descriptor test against a live server, and the merged `col_attribute_test.cpp` has never been compiled locally — `cargo nextest` can't run in this environment (the msrustup toolchain fails codegen on `crc32fast`), so my verification of the merge is inspection and brace balance only. CI's E2E leg is the first compiler to see that file.

David-Engel (commented):

Re-review at `b2ac1e5d` (earlier automated passes covered `2184095`, `bd82847`, `249cb6a`, and `3324de2`). Nothing below was checked by a human first.

## Summary

Both open items from the `3324de2` pass are closed. The merge commit `35b0560` re-anchored the two `typed-columnar-fetch-plan.md` divergence rows on parity, re-cited them against `GetIRDField` / `IS351ORLESSAPP`, and replaced the dangling `VariantDateTimeBaseTypesUseTheThreeXSpellings` anchor with the three new version-gated fixtures — the row for `time` / `datetimeoffset` under `SQL_OV_ODBC3` now matches what `type_rules::resolve_default_c_type` actually does (`SQL_SS_TIME2 if is_3_80 => SQL_C_SS_TIME2`, else `SQL_C_BINARY`), so the internal-inconsistency claim in the row holds. `b2ac1e5d` then cleared the two whitespace/index nits.

The merge itself is the only thing in this range that was not previously reviewed, and it resolves cleanly: `EmptyVariantProbeConsumesValueButKeepsBaseType` from #482 survives byte-for-byte, main's other additions to both conflicted files are intact, and the summary index picked up the entry main had left unlisted. `ASSERT_SQL_OK` wraps `SQL_SUCCEEDED`, so #482's new `SQL_SUCCESS_WITH_INFO` / `01004` answer on a zero-length `SQL_C_BINARY` probe does not break the new `ExpectVariantType` helper.

No new findings.

## Verification

Reviewed the full `e1620ace..b2ac1e5d` diff (5 files, +246/-113) in the isolated worktree, plus `type_rules.rs`, `get_desc_field.rs`, `env.rs`, and the E2E fixture headers. Specifically:

- **Merge integrity.** `git merge-tree 3324de2a e1620ace` reproduces the real conflict in `col_attribute_test.cpp`; compared the merged file against both parents and confirmed nothing from main was dropped and the docs edits in the merge commit are additive.
- **Descriptor width audit.** `SQL_DESC_OCTET_LENGTH` is `FieldValue::Len` and `SQL_DESC_COUNT` / `SQL_DESC_CONCISE_TYPE` / `SQL_DESC_DATETIME_INTERVAL_CODE` are `FieldValue::SmallInt` in `get_desc_field.rs`, so the remaining `GetSmallInt` reads in `set_desc_rec_test.cpp` are correctly sized and `GetLen` (`sizeof(SQLLEN)`) is the right helper for the one wide field. `NumericAttr` in `col_attribute_test.cpp` was already `SQLLEN`.
- **Whitespace.** `git diff --check e1620ace..b2ac1e5d` is clean.
- **Ran locally** (`CARGO_TARGET_DIR` redirected, worktree only): `cargo clippy -p mssqlodbc --all-targets --all-features` clean, and `cargo nextest run -p mssqlodbc --lib --all-features` → 1370 passed, 0 failed. The `b2ac1e5d` delta is C++ comments and blank lines only, so those results carry forward from `35b0560`.
- **Not verified:** the msodbcsql comparison legs and the re-enabled `EquivalentToSetDescFieldSequence`. No live server or msodbcsql source in this run, so the ODBC 2/3/3.8 parity claims still rest on your reported Windows E2E run.

## Findings

None. This looks ready for human review, but it is **not approved** — a human still needs to sign off, and the ADO validation run is still in progress at this head.

David-Engel (commented):
**Automated review** — generated by GitHub Copilot on behalf of @David-Engel. This is **not an approval** and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

## Summary

Two independent changes, both well-scoped.

`SQL_CA_SS_VARIANT_TYPE` now answers the temporal base types the way `GetIRDField` does: `SQL_C_DATE`/`SQL_C_TIMESTAMP` at every declared version, and `SQL_C_SS_TIME2`/`SQL_C_SS_TIMESTAMPOFFSET` only under `SQL_OV_ODBC3_80` with `SQL_C_BINARY` as the fallback. That closes AB#47830 and, as the doc update notes, removes the internal split with `type_rules::resolve_default_c_type`, which already gated the same two types on `is_3_80`.

The environment read is correct. `variant_uses_3_80_types` is computed and the `EnvHandle` guard dropped *before* `stmt.inner.lock()`, so no lock is nested and there's no ordering hazard with the `stmt → env` reads in `get_data.rs` / `fetch_scroll.rs`. `parent_dbc()`/`parent_env()` are the same accessors the existing version-sensitive paths use, and the DM lifetime contract documented on `parent_dbc` covers the raw-pointer deref.

The reported values are serviceable: `canonical_c_type` folds `SQL_C_DATE`/`SQL_C_TIMESTAMP` (9/11) onto `SQL_C_TYPE_DATE`/`SQL_C_TYPE_TIMESTAMP` before `is_valid_c_type`, so a caller that binds what it was told still works. `SQL_C_SS_TIME2 = 0x4000` in `odbc_types.rs` matches the `SQL_C_TYPES_EXTENDED` definition the test's `#ifndef` fallback uses, so the new e2e expectations line up on both legs.

The `set_desc_rec` fix is the right one. `SQL_DESC_OCTET_LENGTH` is a `SQLLEN` field and `SQLGetDescField` legitimately ignores `BufferLength` for fixed-size attributes, so the driver writing 8 bytes was correct and the 2-byte `GetSmallInt` slot was the bug — the test, not the driver, needed to change. I grepped the rest of `tests/e2e/tests/` and found no other `SQLLEN` descriptor field read through a `SQLSMALLINT` slot, so this looks like the only instance of that crash class.

## Verification

- `cargo clippy --all-features --all-targets -- -D warnings` (from `mssql-odbc/`) — clean; no leftover unused imports from the `SQL_C_TYPE_*` swap.
- `cargo test --lib -- api::col_attribute api::set_desc_rec` — 38 passed, 0 failed, including the new `only_odbc_3_80_uses_extended_variant_types`, `temporal_variant_type_honors_the_declared_odbc_version`, and the re-enabled `equivalent_to_the_same_sequence_of_set_desc_field_calls`.
- `gh pr checks 498` — all green, including the ADO validation matrix and the cross-repo mssql-python suite on the mssql-odbc driver.
- Read `type_rules::canonical_c_type`/`is_valid_c_type`, `handles/stmt.rs::parent_dbc`, `handles/dbc.rs::parent_env`, and the `odbc_test_fixture` `SetUp`/`ASSERT_SQL_OK` definitions to confirm the new ODBC 2/3 fixtures set the version before any DBC exists (base `SetUp` allocates only the ENV, `Connect()` is called by the override) and that `ASSERT_SQL_OK` tolerates `SQL_SUCCESS_WITH_INFO` from the zero-length probe.

I did not run the e2e suite — it needs a live server and the msodbcsql comparison leg — so the parity claims in the doc table are taken as measured.

### Blocking

None.

### Suggestion

**`SQL_C_BINARY` is metadata parity that this driver can't yet honor.** For an app declaring `SQL_OV_ODBC3`, a variant `time`/`datetimeoffset` used to report `SQL_C_SS_TIME2`/`SQL_C_SS_TIMESTAMPOFFSET`, which `is_valid_c_type` accepts and the driver can actually deliver. It now reports `SQL_C_BINARY`, which a real buffer still rejects with `HYC00` until AB#47239 lands. So a caller following the ODBC data-type-mapping workflow goes from a working fetch to a defined-but-dead end. The doc update says this plainly and it is the same trade already made for `SQL_C_NUMERIC` under AB#47816, and the cross-repo mssql-python leg is green, so I don't think it blocks — but it is worth a human confirming no ODBC-3-declaring consumer was relying on the old answer, and it strengthens the case for sequencing AB#47239 soon after this.

### Nit

**Two places now encode "3.80 gates the SS temporal C types."** `col_attribute::uses_3_80_variant_types` is `odbc_version == OdbcVersion::Odbc3_80`, and `type_rules.rs:307` spells the identical predicate inline as `let is_3_80 = ...`. Since the whole point of this change is that those two gates must agree, a single shared predicate in `type_rules` (or on `OdbcVersion` itself) would make drift impossible rather than merely unlikely. The dedicated unit test would then live next to it.

**The env mutex is taken for every `SQL_CA_SS_VARIANT_TYPE` call**, including the ones that go on to fail with `HY113` (not a variant column) or `HY010` (no probe yet). Hoisting it is the right call for lock ordering, so this is only a note that the cheap path pays for the version read; not worth restructuring unless profiling says otherwise.

**`ColAttributeOdbc2LiveTest` and `ColAttributeOdbc3LiveTest` are byte-identical apart from the version constant**, and the three `*TemporalVariantTypes` bodies differ only in the last two expectations. A small fixture template parameterized on the declared version, or a helper taking `(version, expected_time, expected_dto)`, would collapse roughly sixty lines and make the next version-sensitive case a one-liner.

Theekshna:
Thanks. Took the first nit in `45af334a`; pushing back on the other two, and the suggestion needs a human call I can't make.

### Nit 1 — two places encoding the same 3.8 gate — **taken**

Agreed, and your framing is the reason: agreement between those two gates is the *point* of this change, so "unlikely to drift" isn't good enough when the whole justification in the plan-doc row is that a variant `time` should be described the same way whether it arrives as a column resolved from `SQL_C_DEFAULT` or as `SQL_CA_SS_VARIANT_TYPE`.

Put it on `OdbcVersion::uses_3_80_types()` rather than in `type_rules`, since both consumers already depend on the enum and neither then depends on the other's module. `uses_3_80_variant_types` stays as the named seam in `col_attribute` — it carries the `IS351ORLESSAPP` citation and the existing unit test — but now delegates. Behavior-preserving by construction: both sites resolve to the same comparison that was previously written out twice.

I did consider arguing the other way, since msodbcsql reaches these two answers through *different* mechanisms — `rgbTRANSTYPE380` for default resolution versus `IS351ORLESSAPP` for the variant metadata — so "they must always agree" is a property of this driver rather than something the spec guarantees. But that's an argument for documenting the coupling, not for duplicating it, and the shared predicate is where the comment now lives.

### Nit 2 — env mutex taken before the `HY113` / `HY010` failures — **declining**

Agreed as written, and you already named the reason it stays: hoisting is what keeps parent-before-child lock ordering intact, and inverting that to save a lock on the failure path would trade a correctness property for an unmeasured micro-optimization. The cheap path pays for one uncontended lock on the ENV mutex, only when `field_identifier == SQL_CA_SS_VARIANT_TYPE`. Leaving it.

### Nit 3 — fixture/body duplication across the three version cases — **declining, deliberately**

This is the one I'd push back on hardest. Those three bodies are the PR's parity evidence, and they're currently the *only* thing that has actually been verified against msodbcsql on a live server — build 172805's Windows leg reports `39 parity, 0 divergence(s), 0 shared failure(s), 0 skipped`. Collapsing sixty lines of explicit per-version expectations into a helper parameterized on `(version, expected_time, expected_dto)` buys concision and costs legibility exactly where a future reader most needs to see the literal expected value next to the literal declared version. I'd rather the next person read three flat tables than reconstruct them from a template. Not worth re-opening verified test code on a green PR for a line count.

### Suggestion — `SQL_C_BINARY` is metadata parity this driver can't yet honor — **agreed, and it needs a human**

You've stated it correctly and I don't want to soften it: for an app declaring `SQL_OV_ODBC3`, a variant `time`/`datetimeoffset` moves from an answer the driver can deliver to one that returns `HYC00` until AB#47239 lands. That is a real regression in serviceability for that specific caller shape, traded for metadata parity.

Two things that bound it, neither of which makes it a non-issue: the ODBC-3 caller can still name `SQL_C_SS_TIME2` explicitly and get its value — `get_data.rs` serves `(ColumnValues::Time, SQL_C_SS_TIME2)` with no version gate, matching msodbcsql, where the 3.8 gate governs only *default* resolution — so this is a dead end only for a caller that mechanically binds whatever the attribute reports. And the cross-repo mssql-python leg is green, which covers the one consumer we know routes on this value.

What I can't do is confirm no *other* ODBC-3-declaring consumer relied on the old answer. @Theekshna — that's your call, and it does strengthen the case for sequencing AB#47239 soon after this.

One correction to the record while I'm here: my earlier caveat that the e2e parity legs had never run is now stale. Build 172805 compiled `col_attribute_test.cpp` under MSVC and ran both legs; `col_attribute_test` and `set_desc_rec_test` each report `PASS PASS parity`. The remaining gap is narrower than I'd stated — the pipeline publishes results per binary, not per gtest case, so I still can't point at a line proving `EquivalentToSetDescFieldSequence` individually reported PASSED rather than SKIPPED. What is verifiable is that the `GTEST_SKIP()` is gone from the source at head and the binary passed.

David-Engel (commented):

Re-review at `45af334a` (earlier automated passes covered `2184095`, `bd82847`, `249cb6a`, `3324de2`, and `b2ac1e5d`). Nothing below was checked by a human first.

## Summary

The only change since the last pass is `45af334a`, which takes the shared-predicate nit: `OdbcVersion::uses_3_80_types()` now backs both `col_attribute::uses_3_80_variant_types` and the `is_3_80` in `type_rules::resolve_default_c_type`. Putting it on the enum rather than in `type_rules` is the right call — neither consumer picks up a dependency on the other's module, and the `IS351ORLESSAPP` citation now lives on the single definition. Behavior-preserving, and I confirmed it is guarded rather than merely refactored. No new findings.

## Verification

Reviewed the full `e1620ace..45af334a` diff (7 files, +259/-114) in the isolated worktree, plus `handles/env.rs`, `handles/stmt.rs`, `get_desc_field.rs`, and the descriptor helpers in the sibling e2e suites.

- **The shared predicate is non-vacuous at both call sites.** Mutated `OdbcVersion::uses_3_80_types` to `self != OdbcVersion::Odbc2` and ran the full `mssqlodbc --lib` suite: `col_attribute::tests::only_odbc_3_80_uses_extended_variant_types`, `col_attribute::tests::temporal_variant_type_honors_the_declared_odbc_version` and `fetch_scroll::tests::a_default_binding_follows_the_declared_odbc_version` all fail. So the refactor is covered from the variant-metadata side *and* the default-resolution side — which is the property the change exists to protect. Reverted afterwards; `git status` clean.
- **No other site spells the gate inline.** Grepped `Odbc3_80` across `mssql-odbc/src`: the only two production comparisons were the two this commit collapsed. Everything else is a test fixture or the `TryFrom`/`get_env_attr` mapping.
- **Descriptor-width audit re-run at this head.** Every remaining `GetSmallInt` / `get_small_int` read across `set_desc_rec_test.cpp`, `set_desc_field_test.cpp`, `set_desc_rec.rs` and the other e2e suites targets `SQL_DESC_COUNT`, `SQL_DESC_CONCISE_TYPE`, `SQL_DESC_DATETIME_INTERVAL_CODE`, `SQL_DESC_ALLOC_TYPE`, `SQL_DESC_NULLABLE`, `SQL_DESC_TYPE`, `SQL_DESC_PRECISION` or `SQL_DESC_SCALE`, all genuinely `SQLSMALLINT`. `SQL_DESC_OCTET_LENGTH` is the only `FieldValue::Len` field read back in a test, and both legs now use `SQLLEN` / `SqlLen`. The `GTEST_SKIP()` is gone from the source at head.
- **Constants line up.** `SQL_C_DATE = 9`, `SQL_C_TIMESTAMP = 11`, `SQL_C_SS_TIME2 = 0x4000`, `SQL_C_SS_TIMESTAMPOFFSET = 0x4001` in `odbc_types.rs` match the e2e `#ifndef` fallbacks, and no `SQL_C_TYPE_*` reference survives in `col_attribute.rs` after the import swap.
- **Ran locally** (`CARGO_TARGET_DIR` redirected, worktree only): `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` clean; `cargo fmt -- --check` clean; `git diff --check` against the merge base clean; `cargo nextest run -p mssqlodbc --lib --all-features -E 'test(variant) + test(desc_rec) + test(col_attribute) + test(3_80)'` → 53 passed, and the full `--lib` run → 1370 passed on the unmutated tree.
- **Not verified:** the msodbcsql comparison legs. No live server or msodbcsql source in this run, so the ODBC 2/3/3.8 parity claims and the individual PASS/SKIP status of `EquivalentToSetDescFieldSequence` still rest on the reported Windows E2E run, as you noted. The `45af334a` delta does not touch the e2e sources, so build 172805's result carries forward.

## Findings

None.

## CI

`gh pr checks 498` is fully green at this head — all 19 checks pass, including every ADO build/test leg, Kerberos, the cross-repo `mssql-python suite on mssql-odbc driver`, and `coverage-report` (97% diff coverage; the only uncovered lines are the poisoned-env-mutex arm).

Applying `ready for human review`. This is still **not an approval** — a human needs to sign off, and the open question about whether any other ODBC-3-declaring consumer relied on the old `SQL_C_SS_*` answer is theirs to close.

David-Engel (commented):

## Summary

Re-review at head `45af334a`. The delta since the previously reviewed head (`b2ac1e5d`) is exactly the shared-gate change: `OdbcVersion::uses_3_80_types()` now lives on the enum in `handles/env.rs`, and both `type_rules::resolve_default_c_type` and `col_attribute` route through it. That was the substantive nit from the last pass and it is properly addressed — I grepped every `Odbc3_80` occurrence under `mssql-odbc/src` and the only remaining equality comparisons are the enum declaration, `TryFrom`, `get_env_attr`'s reverse mapping, and test arguments. No production site spells the predicate out separately anymore, so the two gates that must agree can no longer drift.

Re-verifying the rest of the PR at this head, the earlier conclusions hold:

- **Variant temporal metadata.** `variant_c_type` now reports `SQL_C_DATE`/`SQL_C_TIMESTAMP` (9/11) at every declared version and `SQL_C_SS_TIME2`/`SQL_C_SS_TIMESTAMPOFFSET` only under `SQL_OV_ODBC3_80`, with `SQL_C_BINARY` otherwise. That matches the cited `GetIRDField` behavior and removes the internal split with `resolve_default_c_type`.
- **The reported values stay serviceable for the date/datetime family.** `bind_col.rs:194` and `get_data.rs:205` both run `canonical_c_type` before `is_valid_c_type`, and `canonical_c_type` folds `9..=11` onto `91..=93`, so a caller that binds exactly what `SQL_CA_SS_VARIANT_TYPE` reported still works.
- **Lock ordering.** The env read is hoisted above `stmt.inner.lock()` and the guard is dropped inside the `if` block, so nothing is nested. It matches the existing `stmt.parent_dbc().parent_env()` pattern in `bind_param.rs`, `exec_common.rs`, `fetch_scroll.rs`, `get_data.rs`, and `catalog.rs`.
- **New e2e fixtures.** `ODBCTest::SetUp` allocates only the ENV and declares 3.80; the `ColAttributeOdbc2LiveTest`/`ColAttributeOdbc3LiveTest` overrides re-declare the version before `Connect()` allocates any DBC, which is the legal window for that attribute.
- **Descriptor fix.** `SQL_DESC_OCTET_LENGTH` is a `SQLLEN` field and `SQLGetDescField` ignores `BufferLength` for fixed-size attributes, so the 2-byte slot was the bug and the test was the right thing to change. The `GTEST_SKIP()` that made the earlier fix dead code is gone, so the corrected `GetLen` reads actually execute now. I re-audited both helpers' remaining callers: the only fields still read through the small-int helper are `SQL_DESC_COUNT`, `SQL_DESC_CONCISE_TYPE`, and `SQL_DESC_DATETIME_INTERVAL_CODE`, all genuinely `SQLSMALLINT`.

## Verification

- `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` — clean.
- `cargo nextest run -p mssqlodbc --lib -E 'test(col_attribute) + test(set_desc_rec) + test(type_rules) + test(env)'` — 86 passed, 0 failed.
- `cargo nextest run -p mssqlodbc --lib -E 'test(temporal_variant_type_honors) + test(only_odbc_3_80_uses) + test(variant_c_type_covers) + test(writes_all_eight_fields)'` — 4 passed, confirming the new version-sensitive tests and the re-widened `SqlLen` read run.
- `gh pr checks 498` — every check passing, including the ADO validation matrix and the cross-repo mssql-python suite on the mssql-odbc driver.
- Read `type_rules::canonical_c_type`/`is_valid_c_type`, `handles/dbc.rs::parent_env`, `handles/stmt.rs::parent_dbc`, and `tests/e2e/lib/odbc_test_fixture.cpp::SetUp` to confirm the claims above.

I did not run the C++ e2e suite; it needs a live server and the msodbcsql comparison leg, so the measured-parity claims in `typed-columnar-fetch-plan.md` are taken as reported. The PR description notes the comparison run is being re-executed now that `EquivalentToSetDescFieldSequence` actually has a body — worth confirming that came back green before merge, since this head is the first where that test can pass rather than skip.

### Blocking

None.

### Suggestion

None new. The `SQL_C_BINARY`-under-ODBC-3 serviceability trade-off raised on the previous head still stands as written and is unchanged by this delta; it does not need re-litigating here.

### Nit

**`col_attribute::uses_3_80_variant_types` is now a pure passthrough.** After this commit its whole body is `odbc_version.uses_3_80_types()`, and its doc comment restates what the method on `OdbcVersion` already documents. Calling `env_state.odbc_version.uses_3_80_types()` directly at the one call site would remove the indirection and the duplicated citation. The `only_odbc_3_80_uses_extended_variant_types` test would then belong next to the predicate in `handles/env.rs`, where it covers the shared gate for both consumers instead of only the variant one. Purely cosmetic — the behavior is right either way.

Theekshna:
> **`col_attribute::uses_3_80_variant_types` is now a pure passthrough.**

Taken in `5139a23c`. You're right that it was carrying its weight only as long as it held the comparison itself — once the body became a forward, the doc comment was a second copy of the `IS351ORLESSAPP` / `sqlcdesc.cpp:6474` citation that already sits on `OdbcVersion::uses_3_80_types`, and a duplicated citation is exactly the thing that goes stale. The single call site now calls the method directly and the wrapper is gone.

I also took the second half of the suggestion and moved the test to `handles/env.rs` next to the predicate, renamed `only_odbc_3_80_uses_the_extended_c_types` since it is no longer variant-specific. That matters more than the tidying: the earlier automated pass demonstrated by mutation that this gate is load-bearing on *both* the variant-metadata and default-binding-resolution sides, so the test reads better as a guard on the shared predicate than as a variant-only assertion. Its doc comment now states the failure mode it protects against — the two sides disagreeing, so a variant column reports a C type the driver would then refuse in a bind.

One consequence worth noting: `OdbcVersion` became unused in the `col_attribute` lib target (clippy `-D warnings` caught it), so the import moved into that file's `mod tests`, which still uses it to declare a version in a fixture.

`cargo fmt --check` and `cargo clippy -p mssqlodbc --all-targets --all-features -- -D warnings` are both clean.

On your open item — whether the msodbcsql comparison legs were re-run now that `EquivalentToSetDescFieldSequence` has a body — that is already closed, and I have better evidence than the carry-forward you assumed. ADO build 172844 ran at `45af334a`, not just 172805 at the older head: all 19 checks SUCCESS, `Build_Windows` 3934/3934, and the E2E console log reports `39 parity, 0 divergence(s), 0 shared failure(s), 0 skipped` with zero `[  SKIPPED ]` lines across all 39 binaries. So that test genuinely executed and passed on both comparison legs rather than skipping. Worth flagging that ADO publishes ODBC E2E results per *binary*, not per gtest case, so the per-test status is only visible in the console log.

mssql-odbc/docs/typed-columnar-fetch-plan.md
```diff
@@ -130,7 +130,9 @@ Reading a `sql_variant` column takes three things, not one, and mssql-python nee
...
 2. `SQLGetData(col, SQL_C_BINARY, NULL, 0, &indicator)` must succeed. This is a length/NULL probe, not a data read; it is admitted while binary delivery stays unimplemented (a real buffer is still `HYC00`, tracked as Task [47239](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47239)).
-3. `SQLColAttribute(SQL_CA_SS_VARIANT_TYPE)` returns the C type of the value just probed.
+3. `SQLColAttribute(SQL_CA_SS_VARIANT_TYPE)` returns the C type of the value just probed. For ODBC 2 and 3, variant `time` and `datetimeoffset` values report `SQL_C_BINARY` because their SQL Server-specific C types require ODBC 3.8. This is metadata parity only: retrieving either value into a real binary buffer still returns `HYC00` under Task [47239](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47239).
+
+Variant `date` and datetime-family values deliberately report the legacy C type codes `SQL_C_DATE` / `SQL_C_TIMESTAMP` (9/11), matching msodbcsql at every declared ODBC version, while `SQLDescribeCol` and `SQL_DESC_CONCISE_TYPE` continue to report the ODBC 3 SQL type codes `SQL_TYPE_DATE` / `SQL_TYPE_TIMESTAMP` (91/93).
```
David-Engel: Nit: the second half of this sentence contradicts item 1 four lines above it. For a `sql_variant` column, `SQLDescribeCol` and `SQL_DESC_CONCISE_TYPE` both report `SQL_SS_VARIANT` (`-150`) — `describe_col.rs:240` maps `TdsDataType::SsVariant => SQL_SS_VARIANT`, and `SQLColAttribute`'s concise-type arm reuses the same `odbc_sql_type`. `91`/`93` only appear for a plain, non-variant `date`/`datetime` column. As written, this section — which is specifically about how mssql-python reads a variant column — reads as if the variant column itself describes as `SQL_TYPE_DATE`.

Suggest scoping it, e.g.: "... while a plain (non-variant) `date` / `datetime` column continues to describe as `SQL_TYPE_DATE` / `SQL_TYPE_TIMESTAMP` (91/93). The variant column itself still describes as `SQL_SS_VARIANT`."

The same phrase in the row-109 note ("and `SQLDescribeCol` / `SQL_DESC_CONCISE_TYPE` still report the ODBC 3 SQL type codes `91` / `93`") carries the same ambiguity, though there the surrounding text is clearly about msodbcsql's internal asymmetry, so it matters less.

David-Engel (commented):
**Automated review** — generated by GitHub Copilot on behalf of @David-Engel, unattended. This is **not an approval** and does not satisfy the human review requirement. Push back on anything that looks wrong.

Reviewed the net diff against the merge base (`e1620ac`) at head `5139a23`. The variant-metadata change and the descriptor-width fix both look right, and the earlier rounds' blocking items are all resolved in the tree I read. One documentation nit, nothing blocking.

### What I checked beyond the earlier rounds

- **The `9`/`11` serviceability claim holds, but the earlier justification stopped one layer short.** `get_data.rs:205` only runs `canonical_c_type` for the *validation* check; the delivery dispatch downstream keeps the raw `TargetType`, and the fixed-width fast path `try_write_exact_buffered_scalar` (`get_data.rs:737-799`) has arms for `SQL_C_TYPE_DATE`/`SQL_C_TYPE_TIMESTAMP` only, with `_ => false`. What makes the reported codes serviceable is the general path: `conversion/fetch_convert.rs:609` and `:659` match `SQL_C_TYPE_DATE | SQL_C_DATE` and `SQL_C_TYPE_TIMESTAMP | SQL_C_TIMESTAMP` explicitly, and `bind_col.rs:663-665` folds them on the bind side. So a caller that feeds the attribute result straight back into `SQLGetData` does get its value — just via the conversion path rather than the exact-scalar one. Recording the mechanism so the next reader does not have to re-derive it.
- **The AB#47811 audit is complete for the descriptor suites.** In `set_desc_rec_test.cpp` every remaining `GetSmallInt` call reads a genuinely `SQLSMALLINT` field (`SQL_DESC_COUNT`, `SQL_DESC_CONCISE_TYPE`, `SQL_DESC_DATETIME_INTERVAL_CODE`), and `SQL_DESC_OCTET_LENGTH` is the only `SQLLEN`-width read. I also checked the sibling suites for the same class: `set_desc_field_test.cpp`'s `GetSmallInt` callers are `SQL_DESC_CONCISE_TYPE`/`PRECISION`/`SCALE`/`COUNT`, and `alloc_free_desc_test.cpp` / `execute_test.cpp` / `get_desc_field_test.cpp` read only `SQL_DESC_ALLOC_TYPE`, `COUNT`, `CONCISE_TYPE`, `NULLABLE` and `SQL_DESC_DATA_PTR` (into a `SQLPOINTER`). No other wrong-width read is left in the e2e tree. The Rust `get_len` helper mirrors the existing `get_small_int` shape, including the `BufferLength = 0` argument, which is correct for a fixed-size field.
- **The shared gate has no stragglers.** After the `uses_3_80_types()` extraction, `Odbc3_80` no longer appears in any production comparison outside `handles/env.rs`; the remaining hits are test fixtures and the `TryFrom<u32>`/`get_env_attr` mappings. The env lock in `col_attribute.rs:135-144` is scoped to the block and released before `stmt.inner.lock()`, so parent-before-child ordering holds with no overlap.
- **The new e2e fixtures are sound.** `ODBCTest::SetUp` allocates only the ENV and sets `SQL_OV_ODBC3_80`, so overriding `SQL_ATTR_ODBC_VERSION` before `Connect()` is legal — no DBC exists yet. `SQL_C_SS_TIME2 = 0x4000` in the `#ifndef` fallback matches `odbc_types.rs:425-426`.

### Blocking

None.

### Suggestion

None. The `SQL_C_BINARY`-under-ODBC-3 serviceability trade-off was raised and answered in an earlier round; I have nothing to add beyond agreeing it argues for sequencing AB#47239 soon.

### Nit

One, inline on `docs/typed-columnar-fetch-plan.md`.

### Not verified

No live SQL Server or msodbcsql binary in this run, so the connected legs and the msodbcsql parity citations rest on ADO build 172940 and the author's reported comparison, not on anything I re-ran. Build 172940's Windows, macOS-build, Windows-ARM, SQL-host and cross-repo mssql-python jobs report pass at this head; Build Linux, Build Linux ARM and Test macOS were still pending when I looked.

mssql-odbc/docs/typed-columnar-fetch-plan.md
Theekshna: Correct on both counts, and I verified them rather than taking them on faith: `describe_col.rs:240` maps `TdsDataType::SsVariant => SQL_SS_VARIANT`, and `col_attribute.rs:303` has `concise_type` forwarding straight to the same `odbc_sql_type`. So a variant column reports `-150` from both entry points and never 91/93 — the sentence was simply wrong in the context it sat in, not merely ambiguous. Fixed in `8d4875e6`.

I went a bit further than the suggested wording, because the sentence had a second job: it was trying to say "this is the *C* type of the value, not the column's SQL type," and burying that in a subordinate clause is what let it drift into contradicting item 1. The replacement leads with that distinction, then states the variant column still describes as `SQL_SS_VARIANT` and explicitly ties it back to item 1's dependency, so the two paragraphs now reinforce each other instead of appearing to disagree.

I also fixed the row-109 clause rather than leaving it. You're right that the surrounding text makes the intent clearer there, but "matters less" still leaves a false statement in the parity ledger — which is the one document a future reader will trust over the code. The real asymmetry that row is documenting is between the two adjacent msodbcsql cases, so I dropped the `SQLDescribeCol` mention entirely and made it say that the same value reports `91`/`93` through `SQL_CA_SS_VARIANT_SQL_TYPE` while reporting the 2.x C code here. That is the actual claim, and it is true.

Documentation-only change; no code touched.

> [!NOTE]

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot, unattended. This is **not an approval** and does not satisfy the human review requirement. Nothing below was checked by a human first; push back on anything that looks wrong.

Reviewed the net diff against the merge base (`e1620ac`) at head `8d4875e6`. No blocking findings, no suggestions, no nits. The delta since the previously reviewed head (`5139a23`) is documentation only — the row-109 and item-3 wording that now distinguishes the *C* type of a variant value from the *SQL* type of the column — and it is accurate: `describe_col.rs:240` maps `TdsDataType::SsVariant => SQL_SS_VARIANT`, so a variant column does describe as `-150`, and the `91`/`93` spellings belong to a plain non-variant `date`/`datetime` column, which this change does not touch.

## What I verified rather than assumed

**The `9`/`11` answer is serviceable, not just parity theater.** This was the part most worth checking, because reporting a deprecated code is only defensible if a caller can turn around and bind it. `type_rules::canonical_c_type` maps `SQL_C_DATE..=SQL_C_TIMESTAMP` through `ODBC2_DATETIME_OFFSET` before `is_valid_c_type` sees it, so a caller that binds the reported `9` or `11` lands on `SQL_C_TYPE_DATE`/`SQL_C_TYPE_TIMESTAMP` and is served. The report and the bind agree.

**The shared gate is load-bearing on both sides.** Mutated `OdbcVersion::uses_3_80_types` from `self == Odbc3_80` to `self != Odbc2` and re-ran the affected tests: five failed, spanning both consumers — `only_odbc_3_80_uses_the_extended_c_types`, `temporal_variant_type_honors_the_declared_odbc_version` (variant metadata side) and `ss_datetime_defaults_depend_on_the_odbc_version`, `get_data_default_follows_the_declared_odbc_version`, `a_default_binding_follows_the_declared_odbc_version` (default-resolution side). The new tests are not vacuous, and collapsing the two call sites onto one predicate is now guarded rather than merely tidy.

**The lock hoist cannot deadlock.** `env_state` is a named binding scoped to the `if` block, so the ENV guard drops before `stmt.inner.lock()` is taken — the two are never held simultaneously, and `stmt.parent_dbc().parent_env()` is the same accessor `bind_param.rs:142`, `get_data.rs:155`, `exec_common.rs:545` and `fetch_scroll.rs:819` already use. The extra uncontended lock is confined to `field_identifier == SQL_CA_SS_VARIANT_TYPE`.

**The AB#47811 audit is complete for the touched files.** Every surviving `GetSmallInt` read in `set_desc_rec_test.cpp` (`SQL_DESC_COUNT`, `SQL_DESC_CONCISE_TYPE`, `SQL_DESC_DATETIME_INTERVAL_CODE`) is a genuine `SQLSMALLINT` field; the same holds for the remaining `get_small_int` calls in `set_desc_rec.rs`. `SQL_DESC_OCTET_LENGTH` was the only `SQLLEN` field being read through a 2-byte slot, and fixing it test-side is the right call — ODBC has the driver ignore `BufferLength` for fixed-size attributes, so the buffer width is the application's obligation, not something the driver should defensively clamp.

**The version fixtures are correctly constructed.** `ODBCTest::SetUp` allocates the ENV and declares `SQL_OV_ODBC3_80` without connecting, so the two new subclasses re-declare before `Connect()` rather than after — the order a Driver Manager requires.

## State at review time

All three earlier review threads are resolved and their findings are addressed at this head; I did not re-file any of them. `gh pr checks 498` is green on every required leg, including the Windows msodbcsql comparison and the cross-repo mssql-python suite; only the non-blocking `coverage-report` job is still pending. Diff coverage is 97%, with the two uncovered lines being the ENV mutex-poisoned branch — consistent with how the sibling poisoned-lock paths in this file are treated.

The `SQL_C_BINARY` answer for ODBC-2/3 variant `time`/`datetimeoffset` remains a real serviceability trade (metadata parity ahead of buffered binary retrieval, AB#47239), but it was raised, argued and accepted earlier in this PR and is documented in the plan doc. Recording it as context, not re-opening it.

Local verification: `cargo nextest run -p mssqlodbc --lib` on the five affected tests, clean before the mutation and failing after it, reverted afterwards.

David-Engel (commented):
**Automated review** — generated by GitHub Copilot on behalf of @David-Engel from an unattended run. This is **not an approval** and does not satisfy branch protection; the findings below were not checked by a human first.

## Verdict: no findings

I reviewed the full diff against `origin/main` at head `8d4875e6`, plus the surrounding code the change depends on, and found nothing to raise. Earlier rounds' findings (the stale `GTEST_SKIP()`, the doc rows inverted by the merge from `main`, the passthrough wrapper, and the `91`/`93` scoping in the plan doc) are all addressed at this head, so I have not re-filed any of them.

### What I verified

- **Version gate is wired correctly and is load-bearing.** `variant_c_type(base, uses_3_80_types)` is reached from exactly one call site, and `last_variant_base` has exactly one producer (`get_data.rs`) and one consumer (`col_attribute.rs`), so there is no second path that still answers the old codes. Mutating the gate to `TdsDataType::TimeN => SQL_C_SS_TIME2` (unconditional) fails both `variant_c_type_covers_the_base_types` and `temporal_variant_type_honors_the_declared_odbc_version` — the new tests are not vacuous.
- **The reported legacy codes are serviceable.** `SQL_C_DATE` (9) / `SQL_C_TIMESTAMP` (11) are folded by `type_rules::canonical_c_type` before `is_valid_c_type`, and `get_data.rs:205` canonicalizes on entry, so a caller that mechanically binds the reported type still lands on the `SQL_C_TYPE_*` conversion arms (`fetch_convert.rs:609/659`). This is unchanged behavior for `date`/`datetime`, since 9/11 and 91/93 canonicalize to the same value.
- **Lock ordering.** Reading the ENV version before taking the STMT lock matches the established parent-before-child pattern in `bind_param.rs`, `exec_common.rs`, `get_type_info.rs`, `catalog.rs` and `fetch_scroll.rs`; the guard drops at the end of the `if` block, so the two locks are never held together.
- **E2E fixtures.** `ODBCTest::SetUp` declares `SQL_OV_ODBC3_80` and does not connect, so the new `ColAttributeOdbc2LiveTest` / `ColAttributeOdbc3LiveTest` overrides legitimately re-declare the version before `Connect()`, and `Odbc38TemporalVariantTypes` on the base fixture genuinely exercises the 3.8 leg. No leftover assertion of `91`/`93` for a variant remains anywhere under `mssql-odbc/tests`.
- **Descriptor fix.** `SQL_DESC_OCTET_LENGTH` is the only `SQLLEN`-width field read in `EquivalentToSetDescFieldSequence`; the two `SQL_DESC_CONCISE_TYPE` reads are genuinely `SQLSMALLINT` and correctly stay on `GetSmallInt`. Same audit holds for the Rust `get_len` helper.
- **Targeted tests.** `cargo nextest run -p mssqlodbc --lib -E 'test(col_attribute) + test(env)'` — 66 passed. `writes_all_eight_fields` passes as well.

### Non-blocking notes (not findings)

- The `coverage-report` GitHub check is red, but the log shows it timed out waiting 75 minutes for the ADO Cobertura artifact ("Coverage artifact not found within 75 minutes") — an infrastructure failure, not a code one. The posted coverage comment (97% diff / 93.6% overall) is from build 172844 at `45af334a`; the only uncovered lines are the poisoned-ENV-mutex arm, consistent with every sibling call site. ADO PR validation is green at this head.
- The ODBC 2/3 serviceability trade for variant `time`/`datetimeoffset` (`SQL_C_BINARY` reported, `HYC00` on a real buffer until AB#47239) was raised and answered in an earlier round; I am recording it here only so it stays visible for the human reviewer, not as a new finding.

This looks ready for human review. It is **not approved** — a human still owns the approval and the merge.

saurabh500 (approved):

Issue #499, opened 2026-09-04
macOS PR validation: brew source-builds the docker CLI, timing out Test MacOS


David-Engel:
### Describe the bug

`Test MacOS` in PR validation is failing roughly 56% of the time. 92% of the failures are the `Install and start Colima-based Docker` step ending in `The task has timed out.` at ~790s — its 13-minute pipeline step timeout.

Colima is not the cause. Homebrew ships **no bottle for the docker CLI on Intel macOS** as of 29.8.0, so `brew install docker` compiles it from source, which means building Go first:

```
==> ./make.bash                                    468s
🍺 /usr/local/Cellar/go/1.27.1: 15,577 files, 245.2MB, built in 8 minutes
==> go build -ldflags=... github.com/docker/cli
```

Measured across 147 recent runs of the step, split by how each agent's Homebrew resolved the formula:

| docker install mode | n | brew phase | failed |
| --- | --- | --- | --- |
| Bottle 29.7.2 | 83 | 29s median, 138s max | 1% |
| Source-build 29.8.0 | 64 | **441s median, 774s max** | **36%** |

The registry confirms the asymmetry — 29.8.0 publishes `arm64_sequoia`, `arm64_sonoma`, `arm64_tahoe` and Linux bottles, but no x86_64 macOS bottle at any macOS version. 29.7.2 did publish `sonoma`.

### Steps to reproduce

1. Open any PR against `main` and let PR validation run.
2. Inspect `Test MacOS` → `Install and start Colima-based Docker`.
3. Roughly one run in two, the step times out at ~790s having spent most of it in `brew install`.

### Expected behavior

Installing the docker CLI should take seconds from a prebuilt binary, leaving the step budget for actually booting the VM.

### Actual behavior

Homebrew builds Go and the docker CLI from source, consuming the entire 780s step budget. Colima is then killed mid-boot while healthy — the logs show it having satisfied all three lima requirements with the guest agent running, seconds from ready, when the timeout fires. Downstream coverage conversion and publication steps then fail because the test workload never ran; those are symptoms, not separate causes.

### Why this looked intermittent

Agents differ only in which formula version their Homebrew resolves, and that flipped over a single day:

| Date | resolved 29.7.2 (bottled) | resolved 29.8.0 (unbottled) |
| --- | --- | --- |
| 2026-09-03 | 82 / 82 | 0 |
| 2026-09-04 | 2 / 68 | 66 (97%) |

[...]

PR #501, opened 2026-09-04
DO NOT MERGE: validate the vendored macOS docker toolchain on a real agent

David-Engel:
## ⚠️ DO NOT MERGE — validation only

This PR exists to prove the vendored macOS toolchain works on a real agent. It **replaces the entire PR validation stage set** with a single smoke test while `RunToolchainSmoke` is true, so a run costs one Linux job and one macOS job instead of the full matrix.

Two things must be reverted before any of this merges:
- `RunToolchainSmoke` default back to `false`
- the stage selection in `validation-pipeline.yml`

Both are confined to `validation-pipeline.yml`; `validation-stages.yml` is untouched.

Relates to #499 and builds on #500.

## What it proves

The producer (`build-macos-docker-toolchain.py`) runs on Linux and can only check Mach-O headers. It cannot show that:

- the packaged binaries actually **run** on macOS (a truncated download passes a header check)
- seeding colima's cache genuinely **stops** the ~350 MB guest image download
- lima's guest agent — which lives in `share/`, outside `bin/`, and **was missing from an earlier revision of this payload** — is complete enough to boot a VM

That third one is the reason this PR exists. The first version of the producer extracted only `bin/`, dropping `share/lima/lima-guestagent.Linux-x86_64.gz` (8.1 MB) and `libexec/lima/` (7.6 MB). It would have published cleanly, installed cleanly, and failed at VM boot. It is fixed, but the fix is verified only structurally.

## How

**Linux job** builds both arch payloads and publishes them as pipeline artifacts — not Universal Packages, to keep the loop fast and avoid throwaway versions in the feed. Feed publish and download sizing is a separate question, deliberately not answered here.

**macOS job** then:

1. downloads the x86_64 payload and runs `docker --version`, `colima version`, `limactl --version`
2. seeds `~/Library/Caches/colima/caches/<sha256-of-image-url>` from the payload, recording the file's mtime
3. `colima start`, then **asserts the mtime is unchanged** — if colima had re-downloaded the image, the file would have been replaced
4. runs the real `start-sql-server-macos.sh` and queries `@@VERSION`

Step 4 is the point: a passing run means a container actually ran on a vendored toolchain, not merely that the docker daemon answered.

## What a pass tells us

| Question | Answered |
| --- | --- |
| Do the packaged binaries run on macOS? | yes |
| Is the lima payload complete enough to boot a VM? | yes |
| Does cache seeding remove the 358 MB download? | yes, via the mtime assertion |
| Does a container actually run? | yes, SQL Server starts and answers a query |
| Will a ~415 MB Universal Package publish and download? | **no — separate test** |

## Timings to watch

The macOS job logs artifact download time and `colima start` duration. Current baselines to compare against: brew install ~29s on the bottled path, colima boot p50 265s, guest image fetch+decompress p50 21s.

### DIFF
--- a/.pipeline/macos-docker-toolchain-pipeline.yml
+++ b/.pipeline/macos-docker-toolchain-pipeline.yml
@@ -1,0 +1,79 @@
+# Publishes the macOS docker toolchain (docker CLI, colima, lima and the guest
+# disk image colima boots) as a Universal Package per architecture, so macOS CI
+# jobs install it from our own feed instead of reaching Homebrew, ghcr.io and
+# github.com at job time.
+#
+# Background: Homebrew stopped bottling the docker CLI for Intel macOS at 29.8.0,
+# so `brew install docker` began source-building it (and Go with it, ~8 min),
+# timing out Test MacOS. Colima also downloads a ~350 MB guest image from a
+# personal GitHub repo on every run, which has independently failed on DNS.
+#
+# Manual trigger by design: bumping the toolchain should be a deliberate act with
+# a reviewed version, not something an agent's Homebrew state decides.
+#
+# The build runs on Linux -- it only fetches, verifies and repacks; nothing in
+# the payload is executed here. build-macos-docker-toolchain.py checks the Mach-O
+# architecture of every binary it packs, so a mis-resolved bottle fails the build
+# rather than shipping.
+
+trigger: none
+pr: none
+
+parameters:
+# Versioning is about the payload, not the tools inside it -- manifest.json is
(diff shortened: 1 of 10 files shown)


David-Engel:
Closing: this was validation-only and it did its job. The smoke test passed on a real macOS agent — the packaged binaries ran, the seeded cache made colima skip the 358 MB guest image download (proved by unchanged mtime), and SQL Server 2025 started and answered a query on a fully vendored toolchain.

The keeper work is rebased onto main (which now includes #500) and pushed as `david/macos-docker-toolchain`, without the throwaway pipeline takeover.

PR #505, opened 2026-09-05
Fix nvarchar(max) to SQL_C_CHAR PLP buffer overflow on surrogate straddle

### LINKED ISSUE #211: mssql-odbc: nvarchar(max) → SQL_C_CHAR PLP transcode can overflow the caller buffer on surrogate-straddle, causing silent data loss (and a debug-build panic)
### Describe the bug

On the `nvarchar(max)` → `SQL_C_CHAR` chunked PLP delivery path in `mssql-odbc/src/api/get_data.rs`, the `max_read` sizing under-budgets the transcoded output when a UTF-16 high surrogate is carried across a chunk boundary. The transcoded UTF-8 can exceed the caller's buffer capacity; `copy_with_nul` then truncates mid-UTF-8-sequence. Because the wire position has already advanced, the dropped bytes are lost from the stream (silent corruption + an over-reported `StrLen_or_Ind`). In a debug build the `debug_assert!(!truncated, "transcoded PLP chunk overflowed caller buffer")` turns this into a panic across the FFI boundary.

This code is in `main` (it landed via the column-wise fetch / incremental PLP work, PR #153 — `git blame` points at commit `7e9dfadc`); it is **not** introduced by PR #146. Filing separately so it is tracked against the owning path. Discovered while reviewing #146, which builds on top of this path.

**Root cause.** `max_read` is sized as `((payload_capacity / 3) * 2) & !1` on the assumption "each UTF-16 code unit read from the wire this chunk expands to ≤3 UTF-8 bytes." But `utf16le_chunk_to_utf8` can emit **one more unit than it read this chunk**: a high surrogate carried in from the previous chunk (`pending_high_surrogate`) is prepended to `units` before decoding. When that carried surrogate pairs off (→ a 4-byte astral char) and the remainder of the chunk is 3-byte BMP characters, total output exceeds `payload_capacity`.

### Steps to reproduce

Value `"A𐐷你"` = `A` (1 byte) + U+10437 (surrogate pair `D801 DC37`) + U+4F60 你 (3-byte BMP), read as `SQL_C_CHAR` with a 7-byte buffer:

```sql
SELECT N'A' + NCHAR(0xD801) + NCHAR(0xDC37) + N'你';
```

```
SQLGetData(col, SQL_C_CHAR, buf, /*buffer_length*/ 7, &ind);  // repeat to drain
```

Standalone deterministic repro of the sizing math (mirrors `utf16le_chunk_to_utf8` + `max_read`):

```
buffer_length=7  payload_capacity=6  max_read=4 bytes (2 units/chunk)
chunk 0: read 4 wire bytes -> 1 utf8 byte  "A"    fits_in_6? YES
chunk 1: read 4 wire bytes -> 7 utf8 bytes "𐐷你"  fits_in_6? *** OVERFLOW ***
```

The existing `NvarcharMaxToCharChunkedAstralRoundTrip` e2e test does **not** catch this: it uses pure emoji (U+1F600), which expands to only 2 UTF-8 bytes/unit and so never exceeds the 3-byte budget regardless of surrogate carry.

### Expected behavior

Deliver whole characters that fit and continue on the next call without loss — matching msodbcsql, which truncates at a byte boundary and keeps the stream intact. No overflow, no lost bytes, correct `StrLen_or_Ind`.

### Actual behavior

Transcoded chunk exceeds the buffer; `copy_with_nul` truncates mid-sequence; the truncated bytes are dropped from the stream (silent value corruption). In debug builds, `debug_assert!(!truncated, ...)` panics across FFI.

### Suggested fix

Minimal stopgap — reserve headroom for the possibly-carried surrogate so a full pair plus a following BMP char always fits:

```rust
} else if transcode_utf16_to_utf8 {
    // A high surrogate carried from the previous chunk adds one emitted unit
    // beyond what we read this chunk. Reserve 3 bytes of headroom before
    // dividing so the worst case (carried surrogate pairs off, remainder is
    // 3-byte BMP text) stays within payload_capacity.
    ((payload_capacity.saturating_sub(3) / 3) * 2) & !1
} else {
```

…and add a mixed astral+BMP e2e case (small buffer, e.g. `N'A' + NCHAR(0xD801)+NCHAR(0xDC37) + N'你'` drained in 7–16 byte chunks) so the regression is covered.

Preferred long-term fix — the `pending_utf8` tail-buffer already sketched in the `TODO(convergence)` comment in `stream_active_plp_chunk`: deliver whole UTF-8 bytes that fit and carry the unflushed tail across calls. That removes both this overflow **and** the sub-minimal-buffer `HY090` divergence from msodbcsql in one move (see the skipped `PlpZeroCapacityBufferDoesNotSpin` test).

### Version

Commit `d6a2557` (PR #146 head); underlying code from `7e9dfadc` on `main`.

### Affected crate

mssql-odbc

### Environment

- OS: Windows
- SQL Server: 2022 (16.x)
- Path: `mssql-odbc/src/api/get_data.rs` — `stream_active_plp_chunk` `max_read` sizing (`transcode_utf16_to_utf8` branch) and `utf16le_chunk_to_utf8`.

saurabh500:
## Description

On the `nvarchar(max)` -> `SQL_C_CHAR` chunked PLP delivery path, `max_read` was sized on the assumption that each UTF-16 code unit read this chunk expands to at most 3 UTF-8 bytes: `((payload_capacity / 3) * 2) & !1`. That assumption breaks when a high surrogate is carried across a chunk boundary: the carried surrogate pairs off into a 4-byte astral sequence, so `utf16le_chunk_to_utf8` can emit one more unit than it read, and the transcoded output overruns the caller's buffer. `copy_with_nul` then truncated mid-sequence, and because the wire position had already advanced the dropped bytes were lost from the stream (silent corruption with an over-reported `StrLen_or_Ind`). In debug builds the `debug_assert!(!truncated, ...)` turned this into a panic across the FFI boundary.

The fix adopts the long-term approach the issue recommended: a `pending_utf8` tail buffer on `ActivePlpStream`. Each chunk is transcoded in full, only the bytes that fit are delivered, and the remainder is retained and flushed on later calls. The stream stays alive past wire EOF until the tail drains, mirroring the existing `pending_units` widening path (`varchar(max)` -> `SQL_C_WCHAR`).

Because output is now byte-addressable, this also removes the sub-minimal-buffer `HY090` divergence from msodbcsql: a `SQL_C_CHAR` buffer with any payload room now makes progress one byte per call instead of being rejected. Read sizing (`utf16le_max_read`) subtracts the carried tail and keeps a 2-code-unit floor so a surrogate pair can always assemble in one call. The transcode path still reports `SQL_NO_TOTAL` rather than a byte count, unchanged.

Coverage:
- Two Rust unit tests for the sizing/carry helpers: the surrogate-straddle overflow retains its UTF-8 tail across calls, and multibyte output drains one byte at a time from a 1-byte buffer.
- A C++ e2e regression (`NvarcharMaxToCharSurrogateStraddleRetainsUtf8Tail`) reproducing the exact `N'A' + NCHAR(0xD801)+NCHAR(0xDC37) + NCHAR(0x4F60)` case in a 7-byte buffer.
- The former `PlpSubMinimalBufferIsRejectedButProbeIsAnswered` test is updated to `PlpSubMinimalBufferDrainsWithoutLossAndProbeIsAnswered`, asserting a 2-byte buffer now drains losslessly and runs on the msodbcsql comparison leg for the ASCII case.

## Related Issues

Fixes #211

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (all 1345 mssql-odbc unit tests, including the two new ones, pass locally; the added C++ e2e case and mssql-tds integration tests run in CI against a live SQL Server)
- [ ] Public API changes are documented (N/A - no public API changes; new helpers are private and `pending_utf8` is `pub(crate)`)

### DIFF
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -181,6 +181,7 @@ fn sql_get_data_safe(
         let prepared_stream = (
             active_plp.encoding,
             active_plp.pending_units.len(),
+            active_plp.pending_utf8.len(),
             active_plp.narrow_to_wide.is_some(),
         );
         return stream_active_plp_chunk(
@@ -1465,7 +1466,7 @@ fn stream_active_plp_chunk<'a>(
     buffer_length: SqlLen,
     strlen_or_ind_ptr: *mut SqlLen,
     starting_new_stream: bool,
-    prepared_stream: Option<(PlpEncoding, usize, bool)>,
+    prepared_stream: Option<(PlpEncoding, usize, usize, bool)>,
     mut retained_stmt_state: Option<MutexGuard<'a, StmtState>>,
 ) -> SqlReturn {
     if target_type != SQL_C_CHAR && target_type != SQL_C_WCHAR {
@@ -1487,8 +1488,12 @@ fn stream_active_plp_chunk<'a>(
         return SQL_ERROR;
     }
 
-    let (plp_encoding, widen_carry_len) = if let Some((encoding, widen_carry_len, widening_ready)) =
-        prepared_stream
+    let (plp_encoding, widen_carry_len, utf8_carry_len) = if let Some((
+        encoding,
+        widen_carry_len,
+        utf8_carry_len,
+        widening_ready,
+    )) = prepared_stream
     {
         let compatible = match (target_type, encoding) {
             (SQL_C_WCHAR, PlpEncoding::Utf16Text) => true,
@@ -1517,7 +1522,7 @@ fn stream_active_plp_chunk<'a>(
             }
             return SQL_ERROR;
         }
-        (Some(encoding), widen_carry_len)
+        (Some(encoding), widen_carry_len, utf8_carry_len)
     } else {
         let mut stmt_state = match retained_stmt_state.take() {
             Some(state) => state,
@@ -1621,6 +1626,7 @@ fn stream_active_plp_chunk<'a>(
         let stream_state = (
             stream.map(|s| s.encoding),
             stream.map_or(0, |s| s.pending_units.len()),
+            stream.map_or(0, |s| s.pending_utf8.len()),
         );
         retained_stmt_state = Some(stmt_state);
         stream_state
@@ -1675,12 +1681,7 @@ fn stream_active_plp_chunk<'a>(
         // Whole UTF-16 code units only.
         payload_capacity & !1
     } else if transcode_utf16_to_utf8 {
-        // One BMP UTF-16 code unit expands to at most 3 UTF-8 bytes, so read at
-        // most (cap / 3) code units per chunk. Keeping the byte count even means
-        // a code unit is never split mid-read; surrogate pairs that straddle a
-        // chunk boundary are carried explicitly. This conservative sizing
-        // guarantees the transcoded output always fits the caller's buffer.
-        ((payload_capacity / 3) * 2) & !1
+        utf16le_max_read(payload_capacity, utf8_carry_len)
     } else {
         payload_capacity
     };
@@ -1694,18 +1695,10 @@ fn stream_active_plp_chunk<'a>(
     // available and 01004 are reported as usual. msodbcsql answers this shape
     // the same way.
     //
-    // Payload room too small to carry one whole character is a different case:
-    // the buffer is not a probe, and the conservative sizing above cannot use
-    // the room it has, so every retry would report truncation without
-    // consuming anything and an application looping on an unchanged buffer
-    // would never terminate. That stays HY090.
-    //
-    // Reached by a SQL_C_CHAR buffer of 2-3 bytes transcoding from UTF-16,
-    // where one character needs up to 3, and by a SQL_C_WCHAR buffer of 1 or 3
-    // bytes: one cannot hold even the terminator, the other has a spare byte
-    // that no whole code unit fits in. msodbcsql instead delivers one byte per
-    // call here; matching that needs an unflushed-tail buffer in
-    // `ActivePlpStream` and is tracked separately.
+    // Payload room too small to carry one whole wide character is a different
+    // case: a SQL_C_WCHAR buffer of 1 or 3 bytes cannot make progress and is
+    // rejected with HY090. UTF-8 output has a byte carry, so any SQL_C_CHAR
+    // buffer with payload room can make progress.
     //
     // A probe is exactly two shapes: a zero-length buffer, and one sized for
     // the terminator alone. Everything else that cannot make progress is an
@@ -1716,6 +1709,8 @@ fn stream_active_plp_chunk<'a>(
     let is_length_probe = buffer_length == 0 || buffer_length as usize == terminator_bytes;
     let makes_no_progress = if widen_narrow_to_utf16 {
         widen_out_units == 0
+    } else if transcode_utf16_to_utf8 {
+        max_read == 0 && utf8_carry_len == 0
     } else {
         max_read == 0
     };
@@ -2039,35 +2034,41 @@ fn stream_active_plp_chunk<'a>(
         unsafe { write_if_some(strlen_or_ind_ptr, usable as SqlLen) };
     } else if transcode_utf16_to_utf8 {
         // NVARCHAR PLP wire bytes are UTF-16LE; transcode to UTF-8 for
-        // SQL_C_CHAR, carrying a split code unit or surrogate pair across the
-        // chunk boundary so the value is never corrupted.
-        let utf8 = {
+        // SQL_C_CHAR, carrying split input and output across calls.
+        {
             let Ok(mut ss) = stmt.inner.lock() else {
                 return SQL_ERROR;
             };
             let Some(stream) = ss.active_plp.as_mut() else {
                 return SQL_ERROR;
             };
-            utf16le_chunk_to_utf8(
+            let ActivePlpStream {
+                pending_byte,
+                pending_high_surrogate,
+                pending_utf8,
+                ..
+            } = stream;
+            let emit = transcode_utf16le_into_pending(
                 &payload[..read],
                 reached_end,
-                &mut stream.pending_byte,
-                &mut stream.pending_high_surrogate,
-            )
+                pending_byte,
+                pending_high_surrogate,
+                pending_utf8,
+                payload_capacity,
+            );
+            unsafe {
+                copy_with_nul(
+                    target_value_ptr as *mut u8,
+                    buffer_length as usize,
+                    &pending_utf8[..emit],
+                );
+                write_if_some(
+                    strlen_or_ind_ptr,
+                    SqlLen::try_from(emit).unwrap_or(SqlLen::MAX),
+                );
+            }
+            pending_utf8.drain(..emit);
         };
-        let utf8_bytes = utf8.as_bytes();
-        let truncated = unsafe {
-            copy_with_nul(
-                target_value_ptr as *mut u8,
-                buffer_length as usize,
-                utf8_bytes,
-            )
-        };
-        // Conservative max_read sizing guarantees the transcoded chunk fits.
-        debug_assert!(!truncated, "transcoded PLP chunk overflowed caller buffer");
-        unsafe {
-            write_if_some(strlen_or_ind_ptr, utf8_bytes.len() as SqlLen);
-        }
     } else {
         // SQL_C_CHAR delivery of a non-UTF-16 text PLP column: the wire bytes are
         // copied verbatim. `SingleByteText` and `Utf8Text` have identical bodies
@@ -2121,15 +2122,14 @@ fn stream_active_plp_chunk<'a>(
         }
     };
 
-    // The wire being exhausted is not the same as the value being delivered: the
-    // widening path can still hold decoded units the caller's buffer had no room
-    // for. Ending the stream here would drop them and report success.
-    let widen_units_still_held = stmt_state
+    // The wire being exhausted is not the same as the value being delivered:
+    // converted output can remain after the caller's buffer fills.
+    let converted_data_still_held = stmt_state
         .active_plp
         .as_ref()
-        .is_some_and(|s| !s.pending_units.is_empty());
+        .is_some_and(|s| !s.pending_units.is_empty() || !s.pending_utf8.is_empty());
 
-    if reached_end && !widen_units_still_held {
+    if reached_end && !converted_data_still_held {
         if let Some(mut stream) = stmt_state.active_plp.take() {
             let buffer = stream.take_prefetch_buffer();
             if buffer.capacity() > stmt_state.plp_prefetch_scratch.capacity() {
@@ -2234,6 +2234,33 @@ pub(crate) fn widen_into_pending(
         pending.truncate(base + written);
     }
     out_units.min(pending.len())
+}
+
+/// Chooses an even UTF-16LE wire read size for the available UTF-8 output room.
+///
+/// Two code units are the minimum non-zero read so a surrogate pair can produce
+/// output in one call. Surplus UTF-8 bytes remain in `pending_utf8`.
+fn utf16le_max_read(payload_capacity: usize, pending_utf8_len: usize) -> usize {
+    let remaining = payload_capacity.saturating_sub(pending_utf8_len);
+    if remaining == 0 {
+        0
+    } else {
+        remaining.div_ceil(3).saturating_mul(2).max(4)
+    }
+}
+
+/// Appends one transcoded UTF-16LE chunk and returns the UTF-8 prefix that fits.
+fn transcode_utf16le_into_pending(
+    new_bytes: &[u8],
+    reached_end: bool,
+    pending_byte: &mut Option<u8>,
+    pending_high_surrogate: &mut Option<u16>,
+    pending_utf8: &mut Vec<u8>,
+    out_bytes: usize,
+) -> usize {
+    let utf8 = utf16le_chunk_to_utf8(new_bytes, reached_end, pending_byte, pending_high_surrogate);
+    pending_utf8.extend_from_slice(utf8.as_bytes());
+    out_bytes.min(pending_utf8.len())
 }
 
 /// Transcodes a chunk of UTF-16LE PLP wire bytes to UTF-8 for SQL_C_CHAR
@@ -3014,6 +3041,96 @@ mod tests {
         let second = utf16le_chunk_to_utf8(low, true, &mut pb, &mut ph);
         assert_eq!(second, "😀");
         assert!(pb.is_none() && ph.is_none());
+    }
+
+    #[test]
+    fn utf16_transcode_retains_surrogate_straddle_overflow() {
+        let wire = utf16le("A\u{10437}你");
+        let mut pending_byte = None;
+        let mut pending_high = None;
+        let mut pending_utf8 = Vec::new();
+        let mut delivered = Vec::new();
+        let payload_capacity = 6;
+
+        let first_read = utf16le_max_read(payload_capacity, pending_utf8.len());
+        assert_eq!(first_read, 4);
+        let first_emit = transcode_utf16le_into_pending(
+            &wire[..first_read],
+            false,
+            &mut pending_byte,
+            &mut pending_high,
+            &mut pending_utf8,
+            payload_capacity,
+        );
+        delivered.extend_from_slice(&pending_utf8[..first_emit]);
+        pending_utf8.drain(..first_emit);
+        assert_eq!(delivered, b"A");
+
+        let second_read = utf16le_max_read(payload_capacity, pending_utf8.len());
+        assert_eq!(second_read, 4);
+        let second_emit = transcode_utf16le_into_pending(
+            &wire[first_read..first_read + second_read],
+            true,
+            &mut pending_byte,
+            &mut pending_high,
+            &mut pending_utf8,
+            payload_capacity,
+        );
+        assert_eq!(pending_utf8.len(), 7);
+        assert_eq!(second_emit, payload_capacity);
+        delivered.extend_from_slice(&pending_utf8[..second_emit]);
+        pending_utf8.drain(..second_emit);
+        assert_eq!(pending_utf8.len(), 1);
+
+        let final_emit = transcode_utf16le_into_pending(
+            &[],
+            true,
+            &mut pending_byte,
+            &mut pending_high,
+            &mut pending_utf8,
+            payload_capacity,
+        );
+        delivered.extend_from_slice(&pending_utf8[..final_emit]);
+        pending_utf8.drain(..final_emit);
+
+        assert_eq!(String::from_utf8(delivered).unwrap(), "A\u{10437}你");
+        assert!(pending_utf8.is_empty());
+    }
+
+    #[test]
+    fn utf16_transcode_drains_multibyte_output_one_byte_at_a_time() {
+        let wire = utf16le("你");
+        let mut pending_byte = None;
+        let mut pending_high = None;
+        let mut pending_utf8 = Vec::new();
+        let mut delivered = Vec::new();
+
+        let emit = transcode_utf16le_into_pending(
+            &wire,
+            true,
+            &mut pending_byte,
+            &mut pending_high,
+            &mut pending_utf8,
+            1,
+        );
+        delivered.extend_from_slice(&pending_utf8[..emit]);
+        pending_utf8.drain(..emit);
+
+        while !pending_utf8.is_empty() {
+            assert_eq!(utf16le_max_read(1, pending_utf8.len()), 0);
+            let emit = transcode_utf16le_into_pending(
+                &[],
+                true,
+                &mut pending_byte,
+                &mut pending_high,
+                &mut pending_utf8,
+                1,
+            );
+            delivered.extend_from_slice(&pending_utf8[..emit]);
+            pending_utf8.drain(..emit);
+        }
+
+        assert_eq!(String::from_utf8(delivered).unwrap(), "你");
     }
 
     /// A dangling half code unit at true end-of-stream is genuinely malformed
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -2243,11 +2243,14 @@ pub(crate) fn widen_into_pending(
 
 /// Picks how many UTF-16LE wire bytes to read for the UTF-8 room still free.
 ///
-/// Even, so a read never splits a code unit, and never below two code units so
-/// a surrogate pair (both halves are needed to form one character) can
-/// transcode in a single call. This is only a target: a pair yields a 4-byte
-/// character, so the output can still overrun the room and spill into
-/// `pending_utf8`.
+/// Two wire bytes make one UTF-16 code unit, and a Basic Multilingual Plane
+/// code unit is at most three UTF-8 bytes, so about `room / 3` code units --
+/// two wire bytes each -- is the most that fits. Kept even so a read never
+/// splits a code unit, and never below two code units so a surrogate pair (the
+/// two code units UTF-16 uses for characters outside that plane, such as emoji)
+/// can transcode in one call. Still only a target: a surrogate pair is four
+/// UTF-8 bytes, one past what the `/ 3` sizing assumes, so output can overrun
+/// the room and spill into `pending_utf8`.
 fn utf16le_max_read(payload_capacity: usize, pending_utf8_len: usize) -> usize {
     let remaining = payload_capacity.saturating_sub(pending_utf8_len);
     if remaining == 0 {
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -2033,8 +2033,13 @@ fn stream_active_plp_chunk<'a>(
         // writable for one `SqlLen`.
         unsafe { write_if_some(strlen_or_ind_ptr, usable as SqlLen) };
     } else if transcode_utf16_to_utf8 {
-        // NVARCHAR PLP wire bytes are UTF-16LE; transcode to UTF-8 for
-        // SQL_C_CHAR, carrying split input and output across calls.
+        // The wire carries nvarchar as UTF-16; the caller asked for UTF-8
+        // (SQL_C_CHAR). A code unit or a surrogate pair split across a chunk
+        // boundary is carried on the input side (pending_byte /
+        // pending_high_surrogate). Output can overrun too: a surrogate pair
+        // becomes a 4-byte character, so a chunk may transcode to more UTF-8
+        // than the buffer holds. Transcode the whole chunk, copy only what
+        // fits, and keep the rest in pending_utf8 for the next call.
         {
             let Ok(mut ss) = stmt.inner.lock() else {
                 return SQL_ERROR;
@@ -2236,10 +2241,13 @@ pub(crate) fn widen_into_pending(
     out_units.min(pending.len())
 }
 
-/// Chooses an even UTF-16LE wire read size for the available UTF-8 output room.
+/// Picks how many UTF-16LE wire bytes to read for the UTF-8 room still free.
 ///
-/// Two code units are the minimum non-zero read so a surrogate pair can produce
-/// output in one call. Surplus UTF-8 bytes remain in `pending_utf8`.
+/// Even, so a read never splits a code unit, and never below two code units so
+/// a surrogate pair (both halves are needed to form one character) can
+/// transcode in a single call. This is only a target: a pair yields a 4-byte
+/// character, so the output can still overrun the room and spill into
+/// `pending_utf8`.
 fn utf16le_max_read(payload_capacity: usize, pending_utf8_len: usize) -> usize {
     let remaining = payload_capacity.saturating_sub(pending_utf8_len);
     if remaining == 0 {
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -2244,19 +2244,22 @@ pub(crate) fn widen_into_pending(
 /// Picks how many UTF-16LE wire bytes to read for the UTF-8 room still free.
 ///
 /// Two wire bytes make one UTF-16 code unit, and a Basic Multilingual Plane
-/// code unit is at most three UTF-8 bytes, so about `room / 3` code units --
-/// two wire bytes each -- is the most that fits. Kept even so a read never
-/// splits a code unit, and never below two code units so a surrogate pair (the
-/// two code units UTF-16 uses for characters outside that plane, such as emoji)
-/// can transcode in one call. Still only a target: a surrogate pair is four
-/// UTF-8 bytes, one past what the `/ 3` sizing assumes, so output can overrun
-/// the room and spill into `pending_utf8`.
+/// code unit is at most three UTF-8 bytes, so `room / 3` code units (floored,
+/// two wire bytes each) is the most whose output still fits -- which keeps whole
+/// characters inside a single returned buffer instead of splitting one across
+/// calls. Never below two code units, so a surrogate pair (the two code units
+/// UTF-16 uses for characters outside that plane, such as emoji) can assemble in
+/// one call even when the room is a byte or two.
+///
+/// `pending_utf8` still absorbs the two shapes that overshoot the room anyway: a
+/// surrogate pair is four UTF-8 bytes, one past the three the `/ 3` floor budgets
+/// per unit, and the two-unit floor over-reads any room under six bytes.
 fn utf16le_max_read(payload_capacity: usize, pending_utf8_len: usize) -> usize {
     let remaining = payload_capacity.saturating_sub(pending_utf8_len);
     if remaining == 0 {
         0
     } else {
-        remaining.div_ceil(3).saturating_mul(2).max(4)
+        (remaining / 3).saturating_mul(2).max(4)
     }
 }
 
--- a/mssql-odbc/src/api/get_data.rs
+++ b/mssql-odbc/src/api/get_data.rs
@@ -2073,6 +2073,14 @@ fn stream_active_plp_chunk<'a>(
                 );
             }
             pending_utf8.drain(..emit);
+            // Forward progress, mirroring the widening branch's guard. The
+            // extra read > 0 term is the UTF-16 difference: a chunk can be a
+            // lone carried high surrogate (the #211 straddle), consuming wire
+            // input while emitting nothing yet -- progress, not a stall.
+            debug_assert!(
+                emit > 0 || read > 0 || reached_end || payload_capacity == 0,
+                "UTF-16 to UTF-8 transcode made no forward progress"
+            );
         };
     } else {
         // SQL_C_CHAR delivery of a non-UTF-16 text PLP column: the wire bytes are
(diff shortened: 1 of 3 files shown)


mssql-odbc/src/api/get_data.rs
saurabh500, line 2067: Why this deletion ?

saurabh500: That assert encoded the exact assumption this issue (#211) disproves. It read as "conservative `(cap / 3) * 2` sizing guarantees the transcoded chunk fits the caller buffer," so any overflow (`truncated == true`) was treated as impossible.

That only holds while every UTF-16 code unit stays within 3 UTF-8 bytes. A surrogate pair straddling a chunk boundary -- a high surrogate carried into the next read, then completed by its low half -- transcodes to a single 4-byte character, one byte past the 3-per-unit budget. So a chunk can produce more UTF-8 than the buffer holds: `truncated` becomes true, the assert panics in debug builds, and the old release path silently dropped the overflow (the data loss in the issue).

The fix deliberately allows the transcoded output to exceed the buffer. It transcodes the whole chunk into `pending_utf8`, copies only what fits, and carries the remainder to later calls. `copy_with_nul` now only ever receives `pending_utf8[..emit]`, which fits by construction, so there is no `truncated` condition left to assert on -- keeping the assert would fire on exactly the case this PR fixes.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
```diff
@@ -2236,6 +2241,39 @@ pub(crate) fn widen_into_pending(
...
+/// two code units UTF-16 uses for characters outside that plane, such as emoji)
+/// can transcode in one call. Still only a target: a surrogate pair is four
+/// UTF-8 bytes, one past what the `/ 3` sizing assumes, so output can overrun
+/// the room and spill into `pending_utf8`.
+fn utf16le_max_read(payload_capacity: usize, pending_utf8_len: usize) -> usize {
+    let remaining = payload_capacity.saturating_sub(pending_utf8_len);
+    if remaining == 0 {
+        0
+    } else {
+        remaining.div_ceil(3).saturating_mul(2).max(4)
```
David-Engel: Switching from floor to `div_ceil` does more than make room for the surrogate carry: it makes the read deliberately over-fill in the ordinary case, so multi-byte characters now split across `SQLGetData` calls for buffer sizes that previously stayed aligned. Concretely, with an 8192-byte buffer (payload room 8191) over 3-byte CJK text: old `(8191 / 3) * 2 = 5460` bytes -> 8190 bytes out, fits exactly; new `8191.div_ceil(3) * 2 = 5462` -> 8193 bytes out, 2 bytes carried, so every chunk ends mid-character. Same at 100 and 10 bytes of room. No data is lost, and byte-level splitting is legal for `SQL_C_CHAR`, but each returned buffer is no longer valid UTF-8 on its own — worth confirming the cross-repo mssql-python leg decodes the accumulated bytes rather than per chunk before relying on that.

Since `pending_utf8` now exists, the ceiling isn't needed for correctness. `((remaining / 3) * 2).max(4)` fixes the reported bug just as well (the spill still absorbs the carried-surrogate overflow and the tiny-buffer `max(4)` over-read), while preserving the old chunk alignment for any buffer with 6+ bytes of room. If the over-read is intentional, it'd be worth saying so in the doc comment — it currently reads as if `/ 3` were still the guiding bound.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -1296,37 +1317,20 @@ TEST_F(GetDataLiveTest, VarcharMaxAstralToWcharSurrogatePairBuffer) {
...
-// SQL_C_CHAR to the client codepage, and the codepage measured here is
-// single-byte, so every character is one byte (and unrepresentable ones are
-// best-fit away, losing data). mssql-odbc delivers UTF-8, where a character is
-// 1-4 bytes. Matching it needs an unflushed-tail buffer in ActivePlpStream;
-// tracked separately. Hence the skip on the comparison leg.
-TEST_F(GetDataLiveTest, PlpSubMinimalBufferIsRejectedButProbeIsAnswered) {
-    SKIP_IF_COMPARING_MSODBCSQL();
+// Benefits-from-mock-tds: assert that alternate calls drain the UTF-8 tail
+// without consuming more wire data.
+TEST_F(GetDataLiveTest, PlpSubMinimalBufferDrainsWithoutLossAndProbeIsAnswered) {
```
David-Engel, line 1334: Dropping `SKIP_IF_COMPARING_MSODBCSQL()` here is a parity claim ("both drivers deliver one byte per call"), and per `.github/instructions/mssql-odbc.instructions.md` that needs a measurement against a named build, not just reasoning — the compare leg pins retail msodbcsql 18.6.2.1. Worth waiting for the comparison leg to come back green on this head before merging, since a 2-byte `SQL_C_CHAR` buffer against `nvarchar(max)` is exactly the shape the old comment said msodbcsql handled differently.

David-Engel (commented):

The underlying defect and the fix both check out. With the old floor sizing `((cap / 3) * 2) & !1`, a chunk's output was bounded by `cap` **only when nothing was carried in** — a `pending_high_surrogate` from the previous read completes into a 4-byte character that was never in this read's budget (2 code units budgeted at 6 bytes can emit 4 + 3 = 7), so `copy_with_nul` silently truncated in release and the tail was lost. Carrying the surplus in `pending_utf8` and draining it on later calls is the right shape, and it mirrors the existing `pending_units` carry on the widening path, including the `reached_end && !converted_data_still_held` deferral of stream teardown.

I traced the drain loop to termination: when `max_read == 0` (carry ≥ payload room), `read_prefetched_wire` / `try_read_buffered_plp` with a zero-length `out` consumes no payload, and `try_ensure_active_buffered_chunk` still reports `reached_end` for an exhausted stream, so the final call that empties `pending_utf8` correctly tears the stream down and returns `SQL_SUCCESS` rather than an extra spurious `01004`. Dropping HY090 for tiny `SQL_C_CHAR` buffers is safe for the same reason — a 2-byte buffer reads 4 wire bytes once and then emits one byte per call, so it always advances.

Ran `cargo nextest run -p mssqlodbc --lib` filtered to the PLP/UTF-16 tests (32 passed), including the two new ones.

Two non-blocking points below. Verdict: **suggestions only, no blocking findings**.

mssql-odbc/src/api/get_data.rs
saurabh500: Done -- switched to the floor in 74dc6d3. `utf16le_max_read` is now `(remaining / 3) * 2` (keeping the `.max(4)` tiny-buffer floor), so a buffer with six or more bytes of room reads a whole number of BMP characters again and each returned `SQL_C_CHAR` buffer stays valid UTF-8 on its own. `pending_utf8` still absorbs the two shapes that overshoot -- the 4-byte surrogate pair and the sub-6-byte `max(4)` over-read -- so the reported overflow stays fixed.

I also rewrote the doc comment so `/ 3` reads as the actual (floored) bound and names the two overshoot cases, rather than implying an over-read. Good catch on the mssql-python decode-per-chunk risk: the cross-repo suite leg is green on this head, and floor sizing only reduces mid-character splits, so it is strictly safer there.

> [!NOTE]

mssql-odbc/tests/e2e/tests/get_data_test.cpp
saurabh500: Agreed -- the non-skip is a parity claim, so it needs the named-build measurement rather than reasoning. I strengthened the comment above the test (74dc6d3) to name the comparison leg's pinned build (retail msodbcsql 18.6.2.1, `SQL_DRIVER_VER` 18.06.0002) and frame it as measured, not assumed, matching the convention on `ZeroLengthBinaryProbeOnFixedSourceTypesReportsTruncation`.

The comparison leg is that measurement: it runs this test against 18.6.2.1 and checks both drivers drain a 2-byte `SQL_C_CHAR` buffer over `nvarchar(max)` one byte per call, delivering the whole value without loss -- which is exactly the behavior the fix implements to match. I'll confirm the Test/comparison legs are green on this head before the PR merges; the author owns the merge and it will not go in on red checks.

> [!NOTE]

David-Engel (commented):

Re-reviewed at `74dc6d3` after the switch back to floor sizing. Both points from the previous round are resolved, and I found nothing blocking on this head.

**Sizing math checks out.** With a carried-in high surrogate the worst case for `n` code units read is `4 + 3*(n-1) = 3n + 1` UTF-8 bytes, so `(remaining / 3) * 2` bounds a chunk at `remaining + 1` — one byte over, which is exactly the overshoot `pending_utf8` exists to absorb. The `.max(4)` floor is what keeps `remaining` of 1-2 from collapsing to a zero read (the old HY090 shape), and the `remaining == 0 -> 0` guard is what stops the carry from growing without bound and keeps a `buffer_length` of 0 or 1 a non-consuming probe. For any buffer with six or more payload bytes over BMP text nothing is carried at all, so each returned `SQL_C_CHAR` buffer is standalone-valid UTF-8 again — that was the concern last round.

**Bounds and teardown.** `emit = min(payload_capacity, pending_utf8.len())` with `payload_capacity = buffer_length - 1`, so `copy_with_nul` can never truncate and `&pending_utf8[..emit]` / `drain(..emit)` are always in range. `max_read` can now exceed `payload_capacity` via the `.max(4)` floor, but only on the transcode arm, which is excluded from `direct_wire_output`, so the `write_bytes(target_value_ptr, 0, max_read)` SAFETY precondition still holds for the arms that rely on it. On the drain calls (`max_read == 0`) `active_row_read_state` stays `PlpPaused` after the value is exhausted, so `active_plp_reached_end()` keeps reporting `true` and the call that empties `pending_utf8` tears the stream down with `SQL_SUCCESS` rather than an extra spurious `01004`. `utf16le_chunk_to_utf8` is idempotent under repeated `reached_end` because it `take()`s both carries, so no stray `U+FFFD` on the tail calls.

**Tests and CI.** Ran the PLP/UTF-16 slice of `mssqlodbc --lib` (38 passed) including both new helper tests. ADO validation ran on this head (`19:42Z` -> `20:20Z`, not the duplicate-skip path) and is green, so the `SKIP_IF_COMPARING_MSODBCSQL()` removal in `PlpSubMinimalBufferDrainsWithoutLossAndProbeIsAnswered` is a measured parity claim against the pinned retail build, not an assumed one. `coverage-report` is still pending.

One residual, not worth a change: the `.max(4)` floor also applies for `remaining` of 3-5, where `(remaining / 3) * 2 = 2` would already make progress, so a 4-6 byte `SQL_C_CHAR` buffer over CJK still splits a character across calls. It is legal for `SQL_C_CHAR`, matches msodbcsql, and the doc comment on `utf16le_max_read` already calls it out.

Verdict: **no findings**.

David-Engel (commented):

## Summary

This replaces the input-side-only expansion budget on the `nvarchar(max)` -> `SQL_C_CHAR` PLP path with a real output tail (`pending_utf8`), so a surrogate pair carried across a chunk boundary can no longer transcode to more UTF-8 than the caller's buffer holds. The approach matches the existing `pending_units` widening path, and the stream now correctly stays alive past wire EOF until the tail drains (`converted_data_still_held`). `Fixes #211` is present, all required checks are green, and the delivered-byte accounting still reports `SQL_NO_TOTAL` on truncated calls, so the indicator semantics are unchanged.

I verified the change by reading the full diff against merge base `5c41291`, reading the surrounding PLP delivery path (`stream_active_plp_chunk`, `read_prefetched_wire`, `PlpColumnStream::read_into`, `copy_with_nul`), and running the mssql-odbc unit tests (`cargo nextest run -p mssqlodbc --lib -E 'test(/utf16/)'` — 25 passed, including both new cases).

Two things I checked specifically and believe are correct. First, `pending_utf8` cannot grow without bound: each call adds at most 7 bytes (a carried high surrogate pairing into 4, plus one 3-byte BMP unit) and `utf16le_max_read` returns 0 as soon as the carry reaches `payload_capacity`, so the tail settles at roughly `payload_capacity + 7`. Second, the `.max(4)` floor is load-bearing rather than cosmetic: without it a 2- or 3-byte `SQL_C_CHAR` buffer computes `max_read == 0` with an empty carry and falls into the `HY090` branch that this PR just declared unreachable for `SQL_C_CHAR`. It might be worth saying that in the doc comment, since the current text only motivates the floor by surrogate-pair assembly. Relatedly, the `HY090` branch is now genuinely unreachable on the transcode path (`max_read == 0 && utf8_carry_len == 0` implies `payload_capacity == 0` implies `buffer_length <= 1`, which `is_length_probe` already catches) — the rewritten comment correctly scopes itself to `SQL_C_WCHAR`, so nothing to change, just confirming the reasoning holds.

### Blocking

None.

### Suggestion

**`mssql-odbc/src/api/get_data.rs:1684`** — when `utf16le_max_read` returns 0 because the carry already fills the caller's room, the call still runs the entire wire-read path: it takes the dbc lock, calls `try_read_active_plp_chunk` with an empty slice, and on `Pending` falls through to `block_on(read_active_plp_chunk(&mut []))`, where `read_into` can issue a real network read for the next chunk header. Every byte that call will deliver is already sitting in `pending_utf8`. So a transient read error or a query timeout landing on one of these drain-only calls returns `SQL_ERROR` and discards output that was already transcoded successfully — and with a 2-byte buffer over a 200-byte value there are roughly 200 such trips. Short-circuiting the read when `max_read == 0 && utf8_carry_len > 0` would make the drain purely local; it needs `reached_end` persisted on `ActivePlpStream` so the `reached_end && !converted_data_still_held` teardown still fires on the final drain. I'd treat this as a follow-up rather than merge-blocking: the widening path has the same shape today (it uses 1 instead of 0), so this is not a regression introduced here. It just becomes a common path now that the byte carry makes small `SQL_C_CHAR` buffers legal.

### Nit

**`mssql-odbc/tests/e2e/tests/get_data_test.cpp:1020`** — `SKIP_IF_COMPARING_MSODBCSQL()` here has no stated reason, which breaks the convention every other skip in this file follows (`BackwardColumnRejectedRereadIsNoData`, `VarbinaryMaxToCharReturnsHyc00`, and the test this one replaces all explain themselves). One line — msodbcsql best-fits U+10437 into the client codepage, so a byte-for-byte comparison cannot hold — keeps the next reader from having to re-derive it.

**`mssql-odbc/src/api/get_data.rs:3133`** — the `while !pending_utf8.is_empty()` drain loop has no forward-progress guard, so a regression that makes `emit` come back 0 hangs the test instead of failing it. `drain_widening` a few hundred lines below guards exactly this with `assert!(per_call.len() < 10_000, "widening made no forward progress")`; mirroring that here costs one line.

Nothing above blocks merge from my side, but this is not an approval and does not satisfy the human review requirement.

mssql-odbc/src/api/get_data.rs
```diff
@@ -1675,12 +1681,7 @@ fn stream_active_plp_chunk<'a>(
         // Whole UTF-16 code units only.
         payload_capacity & !1
     } else if transcode_utf16_to_utf8 {
-        // One BMP UTF-16 code unit expands to at most 3 UTF-8 bytes, so read at
-        // most (cap / 3) code units per chunk. Keeping the byte count even means
-        // a code unit is never split mid-read; surrogate pairs that straddle a
-        // chunk boundary are carried explicitly. This conservative sizing
-        // guarantees the transcoded output always fits the caller's buffer.
-        ((payload_capacity / 3) * 2) & !1
+        utf16le_max_read(payload_capacity, utf8_carry_len)
```
David-Engel, line 1684: Suggestion (non-blocking): when this returns 0 because the carry already fills the caller's room, the call still runs the whole wire-read path — dbc lock, `try_read_active_plp_chunk` with an empty slice, and on `Pending` a `block_on(read_active_plp_chunk(&mut []))` that can issue a real network read for the next chunk header. Every byte the call will deliver is already in `pending_utf8`, so a transient read error or a query timeout on one of these drain-only calls returns `SQL_ERROR` and throws away output that was transcoded successfully. With a 2-byte buffer over a 200-byte value that's ~200 round trips through the connection lock.

Short-circuiting when `max_read == 0 && utf8_carry_len > 0` would make the drain purely local. It needs `reached_end` persisted on `ActivePlpStream` so the `reached_end && !converted_data_still_held` teardown still fires on the last drain call. Fine as a follow-up — the widening path has the same shape (it uses 1 rather than 0), so this isn't a regression here, it just becomes a common path now that small `SQL_C_CHAR` buffers are legal.

mssql-odbc/src/api/get_data.rs
```diff
@@ -3016,6 +3057,96 @@ mod tests {
...
+            true,
+            &mut pending_byte,
+            &mut pending_high,
+            &mut pending_utf8,
+            1,
+        );
+        delivered.extend_from_slice(&pending_utf8[..emit]);
+        pending_utf8.drain(..emit);
+
+        while !pending_utf8.is_empty() {
```
David-Engel, line 3145: Nit: no forward-progress guard here, so a regression that makes `emit` come back 0 hangs the test rather than failing it. `drain_widening` further down guards the same shape with `assert!(per_call.len() < 10_000, "widening made no forward progress")` — worth mirroring.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -1010,6 +1010,27 @@ TEST_F(GetDataLiveTest, NvarcharMaxToCharChunkedAstralRoundTrip) {
...
 }
 
+// A carried high surrogate can make the next UTF-16LE read emit more UTF-8
+// bytes than its input-only expansion budget. A 7-byte buffer leaves 6 payload
+// bytes: the second read transcodes U+10437 plus U+4F60 to 7 bytes.
+//
+// Benefits-from-mock-tds: force the exact PLP wire chunks and assert that the
+// final UTF-8 tail remains pending after the wire is exhausted.
+TEST_F(GetDataLiveTest, NvarcharMaxToCharSurrogateStraddleRetainsUtf8Tail) {
+    SKIP_IF_COMPARING_MSODBCSQL();
```
David-Engel, line 1024: Nit: this skip has no stated reason, unlike every other `SKIP_IF_COMPARING_MSODBCSQL()` in this file (and unlike the test it replaces). One line — msodbcsql best-fits U+10437 into the client codepage, so a byte-for-byte comparison can't hold — saves the next reader from re-deriving it.

mssql-odbc/src/api/get_data.rs
saurabh500: Good catch, and agreed it's worth doing. I've filed #513 to track it and kept it out of this PR: the widening path has the identical shape today, so the short-circuit should land for both paths together (with the persisted `reached_end` and probe/prefetch edge cases you called out), rather than a transcode-only half-fix bolted onto the #211 change. Deferring per your "fine as a follow-up".

> [!NOTE]

mssql-odbc/src/api/get_data.rs
saurabh500: Done in cae3356e — mirrored `drain_widening` with an iteration counter and `assert!(iterations < 10_000, "transcode drain made no forward progress")` so a regression that stops emitting fails fast instead of hanging.

> [!NOTE]

mssql-odbc/tests/e2e/tests/get_data_test.cpp
saurabh500: Done in cae3356e — added the reason to the test's doc comment: msodbcsql on Windows best-fits U+10437 into the client ANSI codepage (no representation there), so a byte-for-byte comparison against our UTF-8 output can't hold.

> [!NOTE]

David-Engel (commented):

## Summary

Clean from my side at `cae3356e`. The `pending_utf8` output tail is the right shape for this bug: the old `(payload_capacity / 3) * 2` sizing budgeted 3 UTF-8 bytes per code unit read *this* chunk, which a carried high surrogate defeats — the pair emits 4 bytes for one unit read, `copy_with_nul` truncated, and the wire had already advanced, so the dropped bytes were gone. Transcoding whole and retaining the overflow fixes that at the layer where the mismatch actually lives, and it mirrors the existing `pending_units` widening path rather than inventing a second pattern.

I verified by reading the full diff against merge base `5c41291`, reading the surrounding delivery path (`stream_active_plp_chunk`, `read_prefetched_wire`, `PlpColumnStream::read_into` / `ensure_active_chunk`, `copy_with_nul`), and running the mssql-odbc unit tests against this head — `cargo nextest run -p mssqlodbc --lib -E 'test(/utf16/) or test(/transcode/) or test(/plp/)'`, 38 passed including both new cases. All required checks are green and `Fixes #211` is present.

Specific things I checked and believe hold:

- **No caller-buffer overrun from the `.max(4)` floor.** `direct_wire_output` excludes `(SQL_C_CHAR, Utf16Text)`, so the 4-byte minimum read lands in `inline_payload`/`heap_payload`, never in the application's 2-byte buffer. Delivery is separately bounded by `emit = min(payload_capacity, pending_utf8.len())`, and `payload_capacity == buffer_length - 1`, so `copy_with_nul` cannot truncate and the terminator always fits.
- **`pending_utf8` stays bounded.** Growth requires `remaining > 0`, and per-call output is at most `remaining` for `remaining >= 6` and at most 6 bytes under the floor, so the residual after each drain settles within a few bytes of `payload_capacity` rather than accumulating over a long value.
- **Surviving past wire EOF is safe.** `converted_data_still_held` correctly keeps the stream alive until the tail drains, and a later read on an exhausted stream is inert — `ensure_active_chunk` returns `false` on `reached_end` before touching the reader, so no stray chunk header is consumed from the next column.
- **Indicator semantics unchanged.** Truncated calls still land on the `transcode_utf16_to_utf8` arm of `remaining_indicator` and report `SQL_NO_TOTAL`; only the final call reports delivered bytes.
- **Zero-progress branch.** `max_read == 0 && utf8_carry_len == 0` now implies `payload_capacity == 0`, which `is_length_probe` already catches, so `HY090` is genuinely unreachable on this path — matching the rewritten comment, which correctly scopes itself to `SQL_C_WCHAR`.

The `HY090` → one-byte-per-call change for sub-minimal `SQL_C_CHAR` buffers is a user-visible ODBC behavior change, but it moves toward msodbcsql rather than away, and the description calls it out.

Both nits from the previous round are addressed: the drain loop now has the `iterations < 10_000` forward-progress guard mirroring `drain_widening`, and the `SKIP_IF_COMPARING_MSODBCSQL()` on the surrogate-straddle test states its reason. The earlier drain-only wire-read suggestion is tracked in #513 and is not a regression here — the widening path has the same shape today.

### Blocking

None.

### Suggestion

None.

### Nit

None.

Nothing blocks merge from my side, but this is not an approval and does not satisfy the human review requirement.

Theekshna (commented):
Unattended hourly review sweep: no new findings. Prior review discussion (David-Engel) already covered the transcode sizing floor, the removed debug_assert, and the forward-progress guard on the drain loop; all threads are resolved at this head and the full validation pipeline (including the msodbcsql 18.6.2.1 comparison leg) is green at cae3356e.

Covered: correctness trace of utf16le_max_read/transcode_utf16le_into_pending/pending_utf8 carry invariants, msodbcsql parity against .github/instructions/mssql-odbc.instructions.md (SQL_C_CHAR-is-UTF-8 is a pre-existing documented divergence, not new), PR description/issue #211 linkage, and AI-slop scan of the added doc comments. Could not run a local mutation test in this environment (pre-existing crc32fast/AVX-512 codegen failure in the toolchain, unrelated to this diff); relied on the green CI run instead.

mssql-odbc/src/api/get_data.rs
Theekshna, line 1684: When this returns 0 the call still goes through the wire path. Usually cheap — `try_ensure_active_buffered_chunk` returns early while `chunk_remaining > 0` (decoder.rs:626) — but at a chunk boundary the header isn't buffered, so it falls through `try_read_tds_packet` to `CursorPoll::Pending` and a `block_on(read_active_plp_chunk(&mut []))` packet read. A timeout or transient error there returns SQL_ERROR and throws away bytes that were already converted.

msodbcsql short-circuits this exact shape: `sqlcdata.h:594-596`, `if (!cbBuf) { wError = IDS_01_004; goto Return3; }` — it returns without touching the parser.

Fine to leave for #513, but please put that citation on the issue so it's tracked as a measured divergence rather than a perf nit.

mssql-odbc/src/api/get_data.rs
```diff
@@ -3016,6 +3060,102 @@ mod tests {
...
+    fn utf16_transcode_retains_surrogate_straddle_overflow() {
+        let wire = utf16le("A\u{10437}你");
+        let mut pending_byte = None;
+        let mut pending_high = None;
+        let mut pending_utf8 = Vec::new();
+        let mut delivered = Vec::new();
+        let payload_capacity = 6;
+
+        let first_read = utf16le_max_read(payload_capacity, pending_utf8.len());
+        assert_eq!(first_read, 4);
```
Theekshna, line 3081: This assertion can't tell the floor from the `div_ceil` version it replaced. `utf16le_max_read` is asserted at exactly three points across both new tests — `(6, 0) -> 4` twice and `(1, >=1) -> 0` — and both are blind to the change:

- `(6, 0)`: floor `(6/3)*2 = 4`; ceil `6.div_ceil(3)*2 = 4`. Identical.
- `(1, >=1)`: `remaining == 0` short-circuits before the arithmetic either way.

So reverting 74dc6d3 would still pass everything here. Worth a small table test over remaining = 1/3/5/6/7/8191 to pin the floor and the `.max(4)` behavior the doc comment now spends a paragraph on.

mssql-odbc/src/api/get_data.rs
```diff
@@ -2038,36 +2033,47 @@ fn stream_active_plp_chunk<'a>(
...
+                    target_value_ptr as *mut u8,
+                    buffer_length as usize,
+                    &pending_utf8[..emit],
+                );
+                write_if_some(
+                    strlen_or_ind_ptr,
+                    SqlLen::try_from(emit).unwrap_or(SqlLen::MAX),
+                );
+            }
+            pending_utf8.drain(..emit);
```
Theekshna, line 2075: Nit: the widening branch guards forward progress with `debug_assert!(emitted > 0 || reached_end || widen_out_units == 0, ...)`. cae3356e added the equivalent iteration guard to the unit test but not here. Worth mirroring it now that small `SQL_C_CHAR` buffers are a supported shape.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -1296,37 +1321,22 @@ TEST_F(GetDataLiveTest, VarcharMaxAstralToWcharSurrogatePairBuffer) {
...
-    // which cannot hold a complete transcoded character.
-    SQLCHAR tiny[2] = {0xFF, 0xFF};
-    SQLLEN ind = 0;
-    SQLRETURN rc = SQLGetData(stmt_, 1, SQL_C_CHAR, tiny, sizeof(tiny), &ind);
-    EXPECT_EQ(SQL_ERROR, rc) << "a buffer that cannot make progress must not report truncation";
-    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "HY090");
-
-    // The rejection consumed nothing, so the value still reads back in full.
-    EXPECT_EQ(RepeatToken("abcd", 50), ReadCharDataInChunks(stmt_, 1, 8192));
+    EXPECT_EQ(RepeatToken("abcd", 50), ReadCharDataInChunks(stmt_, 1, 2));
```
Theekshna: `StrLen_or_Ind` isn't asserted anywhere on this path — `ReadCharDataInChunks` has a `final_ind` out-param and grep shows zero callers pass it.

The behavior is right as written (`finish_get_data` never writes the indicator, so the last call keeps `emit` while truncated calls get overwritten with `SQL_NO_TOTAL`), but nothing pins it, and the PR description leans on "still reports SQL_NO_TOTAL, unchanged". The non-transcoding PLP test already does `EXPECT_NE(ind, SQL_NO_TOTAL)`, so this is a couple of lines: pass `&ind` here and assert the final call's indicator is the byte count.

Theekshna (commented):
Reviewed the fix and checked the parity claims against the msodbcsql source directly.

**Parity confirmed** — the design matches msodbcsql, it isn't a divergence:
- Byte-granular converted tail: `sqlsrv.h:1816-17` (`SIZE_T cbTruncatedCharsInConvBuf`, `BYTE rgbConvBuffer[]`) and `sqlcdata.h:879` (`memcpy_s(lpbBuf, cbBuf, rgbTmp, cbBuf)` fills the buffer exactly). msodbcsql does split a multi-byte destination character at a byte boundary across calls, so `pending_utf8` and the "each buffer may not be valid UTF-8 alone" consequence are on-parity.
- Dropping HY090 for a 2-byte `SQL_C_CHAR` buffer: `sqlcdata.h:562` (`cbBuf -= sizeof(SQLCHAR); if (cbBuf < 0) ...`) leaves 1 payload byte and makes progress. Correct to remove.
- `SQL_NO_TOTAL` on the converting path: `sqlcdata.h:1239-1250`, for `wDestCType == SQL_C_CHAR && wSrcCType == SQL_C_WCHAR`. The in-code citation is accurate.

No correctness bug found. Four comments inline: two are small test gaps I'd like closed before merge, two are follow-ups.

mssql-odbc/src/api/get_data.rs
saurabh500: Added the `sqlcdata.h:594-596` citation to #513 (https://github.com/microsoft/mssql-rs/issues/513#issuecomment-5569147124) so it's tracked as a measured divergence — the `if (!cbBuf) { wError = IDS_01_004; goto Return3; }` short-circuit is now recorded as the parity target for the drain-only local delivery. Leaving the change itself to that follow-up as you noted.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
saurabh500: Fixed in eb5261ab. Added `utf16le_max_read_floors_to_whole_characters`, a table test over remaining = 1/3/5/6/7/8191 (and a couple of carry cases). The `(7, 0) -> 4` and `(8191, 0) -> 5460` rows are the ones that pin floor vs `div_ceil` (ceil would read 6 and 5462), and the small rows pin the `.max(4)` two-unit floor. Reverting 74dc6d3 now fails here.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
saurabh500: Done in eb5261ab. Mirrored the widening guard on the transcode arm with one deliberate difference: `debug_assert!(emit > 0 || read > 0 || reached_end || payload_capacity == 0, ...)`. The extra `read > 0` term is load-bearing for UTF-16 — a chunk can be a lone carried high surrogate (the #211 straddle), which consumes wire input while emitting nothing yet. That's real input-side progress, so without the term the guard would false-fire on the exact case this PR exists to handle. It still catches the genuine stall (`emit == 0 && read == 0 && !reached_end && payload_capacity > 0`).

> [!NOTE]

mssql-odbc/tests/e2e/tests/get_data_test.cpp
saurabh500: Done in eb5261ab. `PlpSubMinimalBufferDrainsWithoutLossAndProbeIsAnswered` now passes `&final_ind` into `ReadCharDataInChunks` and asserts `EXPECT_NE(SQL_NO_TOTAL, final_ind)` plus `EXPECT_EQ(1, final_ind)` — the final 2-byte call delivers one payload byte and keeps that count (finish_get_data leaves the indicator alone), while the truncated calls stay on SQL_NO_TOTAL. Since this test doesn't skip the msodbcsql leg, the assertion pins the parity on both drivers.

> [!NOTE]

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot on behalf of @David-Engel. This is not an approval and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

## Summary

Clean at `eb5261ab`. Re-reviewed the full diff against merge base `5c41291`, not just the delta from `cae3356e`.

The fix is the right shape: the old `(payload_capacity / 3) * 2` sizing budgeted 3 UTF-8 bytes per code unit read *this* chunk, which a carried `pending_high_surrogate` defeats — the pair emits 4 bytes for one unit read, so `copy_with_nul` truncated after the wire had already advanced and the tail was gone. Moving the overflow into a `pending_utf8` output tail fixes it at the layer where the mismatch lives, and mirrors the existing `pending_units` widening carry rather than inventing a second pattern.

Sizing math re-derived: with a carried-in high surrogate, `n` code units read produce at most `4 + 3*(n-1) = 3n + 1` UTF-8 bytes, so `(remaining / 3) * 2` bounds a chunk at `remaining + 1` — one byte over, exactly the overshoot `pending_utf8` absorbs. Delivery is separately bounded by `emit = min(payload_capacity, pending_utf8.len())` with `payload_capacity == buffer_length - 1`, so `copy_with_nul` cannot truncate and `&pending_utf8[..emit]` / `drain(..emit)` are always in range. The `.max(4)` floor can push `max_read` above `payload_capacity`, but the transcode arm is excluded from `direct_wire_output`, so that read lands in `inline_payload`/`heap_payload` and never in the application's buffer — the `write_bytes(target_value_ptr, 0, max_read)` SAFETY precondition is untouched.

The three changes since `cae3356e` all land correctly:

- **`utf16le_max_read_floors_to_whole_characters`** closes the gap Theekshna flagged — the previous assertions were blind to floor-vs-`div_ceil`.
- **The transcode-arm `debug_assert!`** mirrors the widening guard. The extra `read > 0` term is load-bearing rather than a loosening: a chunk that is a lone carried high surrogate consumes wire input while emitting nothing, which is the #211 straddle itself. It cannot mask a real stall — when `carry >= payload_capacity` then `max_read == 0` and `emit == payload_capacity > 0`, and when `carry < payload_capacity` then `max_read >= 4`, so `emit == 0 && read == 0 && !reached_end && payload_capacity > 0` is only reachable from a genuinely stalled stream.
- **The e2e `final_ind` assertion** pins what the description leans on: truncated calls are overwritten with `SQL_NO_TOTAL` at the `remaining_indicator` write, and only the terminal `SQL_SUCCESS` call keeps `emit` because `finish_get_data` returns before that write. Tracing a 2-byte buffer over `REPLICATE(N'abcd', 50)`: read-4-wire/emit-1 alternating with a max_read-0 local drain of 1, so the last call delivers exactly one byte — `EXPECT_EQ(1, final_ind)` is the right constant, not an arbitrary one.

Also re-confirmed on this head: the `HY090` branch is genuinely unreachable for `SQL_C_CHAR` (`max_read == 0 && utf8_carry_len == 0` implies `payload_capacity == 0` implies `buffer_length <= 1`, which `is_length_probe` already catches), and the rewritten comment correctly scopes itself to `SQL_C_WCHAR`. `converted_data_still_held` keeps the stream alive past wire EOF, and a later read on an exhausted stream is inert because `ensure_active_chunk` returns on `reached_end` before touching the reader.

## Verification

- Read the full diff plus the surrounding delivery path: `stream_active_plp_chunk`, `read_prefetched_wire`, `PlpColumnStream::read_into` / `ensure_active_chunk`, `copy_with_nul`, and the `ReadCharDataInChunks` helper's pre-existing `final_ind` out-param.
- `cargo nextest run -p mssqlodbc --lib -E 'test(/utf16/) or test(/transcode/) or test(/plp/)'` — 39 passed at this head.
- **Mutation-tested the new table test**, since its whole purpose is to discriminate a change the old assertions could not see. Both mutations fail it, so it is not vacuous:
  - `(remaining / 3)` → `remaining.div_ceil(3)`: `utf16le_max_read(7, 0)` left `6`, right `4`.
  - dropping `.max(4)`: `utf16le_max_read(1, 0)` left `0`, right `4`.
- `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` — clean.
- All required checks green or pending on this head; none failing. `Fixes #211` is present, and the follow-up carved out of review (drain-only calls still traversing the wire path) is tracked in #513 with the `sqlcdata.h:594-596` parity citation attached.

### Blocking

None.

### Suggestion

None.

### Nit

None.

Nothing blocks merge from my side, but this is not an approval and does not satisfy the human review requirement.

Theekshna (approved):

Issue #507, opened 2026-09-06
mssql-odbc: single-row bound fetch regressed ~17% against the pinned perf baseline


saurabh500:
### Describe the bug

The ODBC perf pipeline (pipeline 2325) flagged a confirmed regression against the pinned `mssql-odbc` baseline in run `20260905.1` (buildId 172669). The regression is isolated to rowset size 1: the fixed per-call setup cost of `SQLFetchScroll` grew, and at rowset 1 that cost is paid once per row instead of being amortized across 64 or 1000 rows.

Three changes landed between the last green run (`49f2ccb4`, run 171863) and the regressed run, and each added fixed work to every fetch:

1. Descriptor-backed bindings ([AB#47437](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47437)) made every `SQLFetchScroll`, `SQLBindCol`, and `SQLFreeStmt(SQL_UNBIND)` call `handles::live_type(ard)`, which takes the process-global `LIVE_HANDLES` mutex and does a hash lookup. The `bind_cycle` workload pays this 17 times per row.
2. Deferred `SQL_C_DEFAULT` resolution (#460) acquires the ENV mutex and heap-allocates a `Vec<SqlSmallInt>` of per-column SQL types on every fetch, whether or not any binding actually uses `SQL_C_DEFAULT`.
3. The PLP metadata cache added in `e8c4ffa4` was lost in the revert at `49f2ccb4`, so `fill_rowset` re-locks the STMT and rebuilds per-column PLP encoding info on every call.

### Steps to reproduce

1. Build `mssql-odbc` at `5c412918` (or any commit at or after the merge of #460).
2. Run the perf-lab benchmark suite against it with the pinned baseline commit from `mssql-odbc-bench/perf-lab/baseline-commit.txt`:
   - `mssql-odbc-bench/perf-lab/run-benchmarks.sh` (or `run-benchmarks.ps1` on Windows)
3. Compare the `fetch/rowset_100k_c15_fixed/bound_rowset_1` and `fetch/rowset_20k_c15_fixed/bind_cycle_rowset_1` medians against the baseline.

### Expected behavior

Single-row fetch wall time stays at or below the pinned baseline. Run 171863 had it 17.71% *faster* than baseline on `bound_rowset_1` and 15.55% faster on `bind_cycle_rowset_1`, so that is the level the driver had already reached.

### Actual behavior

[...]

PR #508, opened 2026-09-06
Restore single-row bound fetch performance

saurabh500:
## Description

Perf pipeline run `20260905.1` (buildId 172669) flagged a confirmed regression against the pinned `mssql-odbc` baseline, isolated to rowset size 1: `bound_rowset_1` was 16.7% slower and `bind_cycle_rowset_1` 13.2% slower, while the rowset 64 and 1000 variants stayed flat. The reported `execute_ms` and `metadata_bind_ms` phases were unchanged, which places the cost in the fixed per-call setup of `SQLFetchScroll` -- work that is amortized across a wide rowset but paid once per row at rowset 1.

This PR removes two such fixed costs. Both were previously paid on every fetch regardless of the result set's shape, including on the all-fixed-width benchmarks where neither had anything to do.

**Make `SQL_C_DEFAULT` resolution conditional.** Deferred `SQL_C_DEFAULT` resolution acquired the ENV mutex and heap-allocated a `Vec<SqlSmallInt>` of per-column SQL types on every fetch, whether or not any binding was actually left at `SQL_C_DEFAULT`. Both now happen only when some binding needs them. The original code did this unconditionally on the grounds that gating on the bindings would invert the lock order, but it does not: the ARD lock is released before the ENV and STMT locks are taken, so no lock is held while acquiring another and the STMT-before-DESC rule is preserved.

**Cache PLP metadata per result set.** `fill_rowset` re-locked the STMT and rebuilt a `Vec<Option<PlpColumnInfo>>` on every call, ungated, so a 15-column fixed-width result set allocated and filled a 15-element vector of `None` per fetch. This restores the cache that `e8c4ffa4` added and the revert at `49f2ccb4` lost, now storing the resolved text encoding too so `get_encoding_type` is hoisted out of the hot path. It is populated in `refresh_metadata_caches` only when the result set actually has PLP columns, and cleared in `clear_result_metadata`.

## Results

Measured by perf run `20260906.2` (buildId 172991) against the pinned baseline. The regression gate passed: no confirmed benchmark had at least 5% higher wall time.

| Benchmark | Before (run 172669) | After (run 172991) |
| --- | --- | --- |
| `fetch/rowset_100k_c15_fixed/bound_rowset_1` | +16.7% | +4.17% |
| `fetch/rowset_20k_c15_fixed/bind_cycle_rowset_1` | +13.2% | +6.22%, not confirmed (2/4) |

No other benchmark regressed; the rowset 64/1000 and `getdata/*` families are flat or improved.

### What this PR does not recover

An earlier revision also skipped the `LIVE_HANDLES` registry lookup when the ARD is the statement's own implicit descriptor. Review showed that to be unsafe and it has been withdrawn: `SQLDisconnect` is a call on the connection rather than the statement, so nothing serializes it against an in-flight fetch, and `disconnect.rs:96-99` documents that it frees statements while operations on them are still running. Since `StmtHandle::new` registers all four implicit descriptors and `StmtHandle::drop` removes them from `LIVE_HANDLES`, the lookup genuinely does narrow that window for implicit descriptors, exactly as it does for explicit ones. The check is unconditional again and `bind_col.rs` is untouched by this PR.

That is where most of the `bind_cycle` gain lived -- that workload pays 17 lookups per row at rowset size 1 -- which is why `bind_cycle_rowset_1` remains the weakest benchmark here. The follow-up worth doing is making `LIVE_HANDLES` itself cheap rather than skipping it: it is a single process-global `Mutex<HashMap<usize, HandleType>>` hashing pointer keys with SipHash, so a faster hasher and/or an `RwLock` would cut the cost while leaving the guard exactly as strict. I would rather do that separately than widen this PR.

## Related Issues

Related to #507

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes -- `cargo nextest run -p mssqlodbc` is green at 1344/1344, but the full workspace run cannot complete locally: the `mssql-tds` integration tests need a live SQL Server. Relying on PR validation, which is green on this head.
- [x] New/changed functionality has tests -- `plp_column_cache_does_not_outlive_its_result_set` covers the cache lifecycle including the PLP-to-non-PLP transition; mutation-checked by removing the reset from `clear_result_metadata` and confirming it fails.
- [x] Public API changes are documented -- no public API change. `PlpColumnInfo` and `StmtState::plp_columns` are `pub(crate)`.

### DIFF
--- a/mssql-odbc/src/api/bind_col.rs
+++ b/mssql-odbc/src/api/bind_col.rs
@@ -116,7 +116,7 @@ fn sql_bind_col_safe(
     // `free_desc` already walks DBC→STMT in the other direction to reset a
     // freed descriptor's associations, and holding both here in the opposite
     // order would be a classic ABBA deadlock.
-    let (ard, canonical_type) = {
+    let (ard, ard_is_explicit, canonical_type) = {
         let Ok(mut stmt_state) = stmt.inner.lock() else {
             error!("SQLBindCol: stmt mutex poisoned");
             return SQL_ERROR;
@@ -161,8 +161,9 @@ fn sql_bind_col_safe(
         // the column from the row cursor and run validation msodbcsql never reaches.
         if target_value_ptr.is_null() {
             let ard = stmt_state.effective_ard(stmt);
+            let ard_is_explicit = stmt_state.active_ard.is_some();
             drop(stmt_state);
-            let Ok(()) = unbind_ard_column(ard, column_number) else {
+            let Ok(()) = unbind_ard_column(ard, ard_is_explicit, column_number) else {
                 error!("SQLBindCol: ard mutex poisoned; unbind failed");
                 if let Ok(mut stmt_state) = stmt.inner.lock() {
                     post_sql_error(
@@ -198,7 +199,11 @@ fn sql_bind_col_safe(
             return SQL_ERROR;
         }
 
-        (stmt_state.effective_ard(stmt), canonical_type)
+        (
+            stmt_state.effective_ard(stmt),
+            stmt_state.active_ard.is_some(),
+            canonical_type,
+        )
     };
 
     let binding = ColumnBinding {
@@ -214,7 +219,7 @@ fn sql_bind_col_safe(
         // SQLSetDescFieldW/SQLSetDescRec.
         octet_length_ptr: strlen_or_ind_ptr,
     };
-    let Ok(()) = bind_ard_column(ard, binding) else {
+    let Ok(()) = bind_ard_column(ard, ard_is_explicit, binding) else {
         error!("SQLBindCol: ard mutex poisoned or missing record after growth");
         if let Ok(mut stmt_state) = stmt.inner.lock() {
             post_sql_error(
@@ -245,8 +250,12 @@ fn sql_bind_col_safe(
 /// conversion here is not expected to fail in practice — but this still
 /// reports it as an error rather than panicking or silently truncating to the
 /// wrong record.
-fn bind_ard_column(ard: SqlHandle, binding: ColumnBinding) -> Result<(), ()> {
-    if crate::handles::live_type(ard) != Some(HandleType::Desc) {
+fn bind_ard_column(
+    ard: SqlHandle,
+    ard_is_explicit: bool,
+    binding: ColumnBinding,
+) -> Result<(), ()> {
+    if ard_is_explicit && crate::handles::live_type(ard) != Some(HandleType::Desc) {
         return Err(());
     }
     let desc = unsafe { handle_from_raw::<DescHandle>(ard) };
@@ -275,8 +284,12 @@ fn bind_ard_column(ard: SqlHandle, binding: ColumnBinding) -> Result<(), ()> {
 /// descriptor: reporting `SQL_SUCCESS` here would tell the application an
 /// unbind happened when it didn't, and a stale bound column would keep
 /// writing through a possibly-freed application buffer on the next fetch.
-fn unbind_ard_column(ard: SqlHandle, column_number: SqlUSmallInt) -> Result<(), ()> {
-    if crate::handles::live_type(ard) != Some(HandleType::Desc) {
+fn unbind_ard_column(
+    ard: SqlHandle,
+    ard_is_explicit: bool,
+    column_number: SqlUSmallInt,
+) -> Result<(), ()> {
+    if ard_is_explicit && crate::handles::live_type(ard) != Some(HandleType::Desc) {
         return Err(());
     }
     let desc = unsafe { handle_from_raw::<DescHandle>(ard) };
@@ -318,7 +331,7 @@ pub(crate) unsafe fn sql_free_stmt_unbind(statement_handle: SqlHandle) -> SqlRet
 }
 
 fn sql_free_stmt_unbind_safe(stmt: &StmtHandle) -> SqlReturn {
-    let ard = {
+    let (ard, ard_is_explicit) = {
         let Ok(mut stmt_state) = stmt.inner.lock() else {
             error!("SQLFreeStmt(SQL_UNBIND): stmt mutex poisoned");
             return SQL_ERROR;
@@ -329,10 +342,13 @@ fn sql_free_stmt_unbind_safe(stmt: &StmtHandle) -> SqlReturn {
             post_diag(&mut stmt_state, ERR_FUNCTION_SEQUENCE);
             return SQL_ERROR;
         }
-        stmt_state.effective_ard(stmt)
+        (
+            stmt_state.effective_ard(stmt),
+            stmt_state.active_ard.is_some(),
+        )
     };
 
-    if crate::handles::live_type(ard) != Some(HandleType::Desc) {
+    if ard_is_explicit && crate::handles::live_type(ard) != Some(HandleType::Desc) {
         error!("SQLFreeStmt(SQL_UNBIND): ard freed concurrently; unbind failed");
         if let Ok(mut stmt_state) = stmt.inner.lock() {
             post_sql_error(
@@ -890,7 +906,7 @@ mod tests {
             strlen_or_ind_ptr: ptr::null_mut(),
             octet_length_ptr: ptr::null_mut(),
         };
-        assert!(bind_ard_column(explicit_ard, binding).is_err());
+        assert!(bind_ard_column(explicit_ard, true, binding).is_err());
     }
 
     /// Same race, same guard, for `unbind_ard_column`.
@@ -899,6 +915,6 @@ mod tests {
         let mut h = TestHandles::with_env_dbc_stmt();
         let explicit_ard = h.alloc_explicit_desc();
         assert_eq!(h.free_explicit_desc(explicit_ard), SQL_SUCCESS);
-        assert!(unbind_ard_column(explicit_ard, 1).is_err());
+        assert!(unbind_ard_column(explicit_ard, true, 1).is_err());
     }
 }
--- a/mssql-odbc/src/api/bind_col.rs
+++ b/mssql-odbc/src/api/bind_col.rs
@@ -116,7 +116,7 @@ fn sql_bind_col_safe(
     // `free_desc` already walks DBC→STMT in the other direction to reset a
     // freed descriptor's associations, and holding both here in the opposite
     // order would be a classic ABBA deadlock.
-    let (ard, ard_is_explicit, canonical_type) = {
+    let (ard, canonical_type) = {
         let Ok(mut stmt_state) = stmt.inner.lock() else {
             error!("SQLBindCol: stmt mutex poisoned");
             return SQL_ERROR;
@@ -161,9 +161,8 @@ fn sql_bind_col_safe(
         // the column from the row cursor and run validation msodbcsql never reaches.
         if target_value_ptr.is_null() {
             let ard = stmt_state.effective_ard(stmt);
-            let ard_is_explicit = stmt_state.active_ard.is_some();
             drop(stmt_state);
-            let Ok(()) = unbind_ard_column(ard, ard_is_explicit, column_number) else {
+            let Ok(()) = unbind_ard_column(ard, column_number) else {
                 error!("SQLBindCol: ard mutex poisoned; unbind failed");
                 if let Ok(mut stmt_state) = stmt.inner.lock() {
                     post_sql_error(
@@ -199,11 +198,7 @@ fn sql_bind_col_safe(
             return SQL_ERROR;
         }
 
-        (
-            stmt_state.effective_ard(stmt),
-            stmt_state.active_ard.is_some(),
-            canonical_type,
-        )
+        (stmt_state.effective_ard(stmt), canonical_type)
     };
 
     let binding = ColumnBinding {
@@ -219,7 +214,7 @@ fn sql_bind_col_safe(
         // SQLSetDescFieldW/SQLSetDescRec.
         octet_length_ptr: strlen_or_ind_ptr,
     };
-    let Ok(()) = bind_ard_column(ard, ard_is_explicit, binding) else {
+    let Ok(()) = bind_ard_column(ard, binding) else {
         error!("SQLBindCol: ard mutex poisoned or missing record after growth");
         if let Ok(mut stmt_state) = stmt.inner.lock() {
             post_sql_error(
@@ -250,12 +245,8 @@ fn sql_bind_col_safe(
 /// conversion here is not expected to fail in practice — but this still
 /// reports it as an error rather than panicking or silently truncating to the
 /// wrong record.
-fn bind_ard_column(
-    ard: SqlHandle,
-    ard_is_explicit: bool,
-    binding: ColumnBinding,
-) -> Result<(), ()> {
-    if ard_is_explicit && crate::handles::live_type(ard) != Some(HandleType::Desc) {
+fn bind_ard_column(ard: SqlHandle, binding: ColumnBinding) -> Result<(), ()> {
+    if crate::handles::live_type(ard) != Some(HandleType::Desc) {
         return Err(());
     }
     let desc = unsafe { handle_from_raw::<DescHandle>(ard) };
@@ -284,12 +275,8 @@ fn bind_ard_column(
 /// descriptor: reporting `SQL_SUCCESS` here would tell the application an
 /// unbind happened when it didn't, and a stale bound column would keep
 /// writing through a possibly-freed application buffer on the next fetch.
-fn unbind_ard_column(
-    ard: SqlHandle,
-    ard_is_explicit: bool,
-    column_number: SqlUSmallInt,
-) -> Result<(), ()> {
-    if ard_is_explicit && crate::handles::live_type(ard) != Some(HandleType::Desc) {
+fn unbind_ard_column(ard: SqlHandle, column_number: SqlUSmallInt) -> Result<(), ()> {
+    if crate::handles::live_type(ard) != Some(HandleType::Desc) {
         return Err(());
     }
     let desc = unsafe { handle_from_raw::<DescHandle>(ard) };
@@ -331,7 +318,7 @@ pub(crate) unsafe fn sql_free_stmt_unbind(statement_handle: SqlHandle) -> SqlRet
 }
 
 fn sql_free_stmt_unbind_safe(stmt: &StmtHandle) -> SqlReturn {
-    let (ard, ard_is_explicit) = {
+    let ard = {
         let Ok(mut stmt_state) = stmt.inner.lock() else {
             error!("SQLFreeStmt(SQL_UNBIND): stmt mutex poisoned");
             return SQL_ERROR;
@@ -342,13 +329,10 @@ fn sql_free_stmt_unbind_safe(stmt: &StmtHandle) -> SqlReturn {
             post_diag(&mut stmt_state, ERR_FUNCTION_SEQUENCE);
             return SQL_ERROR;
         }
-        (
-            stmt_state.effective_ard(stmt),
-            stmt_state.active_ard.is_some(),
-        )
+        stmt_state.effective_ard(stmt)
     };
 
-    if ard_is_explicit && crate::handles::live_type(ard) != Some(HandleType::Desc) {
+    if crate::handles::live_type(ard) != Some(HandleType::Desc) {
         error!("SQLFreeStmt(SQL_UNBIND): ard freed concurrently; unbind failed");
         if let Ok(mut stmt_state) = stmt.inner.lock() {
             post_sql_error(
@@ -906,7 +890,7 @@ mod tests {
             strlen_or_ind_ptr: ptr::null_mut(),
             octet_length_ptr: ptr::null_mut(),
         };
-        assert!(bind_ard_column(explicit_ard, true, binding).is_err());
+        assert!(bind_ard_column(explicit_ard, binding).is_err());
     }
 
     /// Same race, same guard, for `unbind_ard_column`.
@@ -915,6 +899,6 @@ mod tests {
         let mut h = TestHandles::with_env_dbc_stmt();
         let explicit_ard = h.alloc_explicit_desc();
         assert_eq!(h.free_explicit_desc(explicit_ard), SQL_SUCCESS);
-        assert!(unbind_ard_column(explicit_ard, true, 1).is_err());
+        assert!(unbind_ard_column(explicit_ard, 1).is_err());
     }
 }
--- a/mssql-odbc/src/api/fetch_scroll.rs
+++ b/mssql-odbc/src/api/fetch_scroll.rs
@@ -971,6 +971,12 @@ fn fetch_scroll_safe(
                 error!("SQLFetchScroll: env mutex poisoned");
                 if let Ok(mut stmt_state) = stmt.inner.lock() {
                     stmt_state.clear_state(STMT_STATE_FETCH_IN_PROGRESS);
+                    post_sql_error(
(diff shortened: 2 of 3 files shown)


mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -956,13 +926,15 @@ fn fetch_scroll_safe(
...
-    let bindings: Vec<ColumnBinding> = {
+    let mut bindings: Vec<ColumnBinding> = {
         // `ard` can be an explicit descriptor resolved under the STMT lock,
         // already dropped by now — re-check liveness right before
         // dereferencing to narrow (not fully close) the race against a
         // concurrent `SQLFreeHandle(SQL_HANDLE_DESC)` on that same
         // descriptor.
-        if crate::handles::live_type(ard) != Some(crate::handles::HandleType::Desc) {
+        if ard_is_explicit
+            && crate::handles::live_type(ard) != Some(crate::handles::HandleType::Desc)
```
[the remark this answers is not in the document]
saurabh500: You are right, and my reasoning was wrong. I checked `SQLFreeHandle`, which rejects implicit descriptors with HY017, and statement-level `SQLFreeHandle`, which the Driver Manager serializes -- but I missed that `SQLDisconnect` is a call on the connection, not the statement, so nothing serializes it against an in-flight fetch. `disconnect.rs:96-99` says so outright.

I also confirmed the registry lookup genuinely helps here rather than being incidentally true only for explicit descriptors: `StmtHandle::new` registers all four implicit descriptors through `handle_to_raw` (`handles/stmt.rs:994-1009`), and `StmtHandle::drop` frees them via `free_handle` (`handles/stmt.rs:1083`), which does `live_handles().remove(...)`. So after a racing disconnect, `live_type(ard)` really does return `None` for an implicit ARD and the fetch bails instead of dereferencing freed memory.

Restored the unconditional lookup at all three sites and removed the `ard_is_explicit` plumbing entirely in 0da952f3. I widened the comment here to name the `SQLDisconnect` path too, since that was the case the old wording did not cover.

The other two optimizations in this PR (conditional `SQL_C_DEFAULT` resolution and the per-result-set PLP metadata cache) do not touch handle lifetimes and are unchanged.

mssql-odbc/src/api/bind_col.rs
```diff
@@ -245,8 +250,12 @@ fn sql_bind_col_safe(
...
 /// reports it as an error rather than panicking or silently truncating to the
 /// wrong record.
-fn bind_ard_column(ard: SqlHandle, binding: ColumnBinding) -> Result<(), ()> {
-    if crate::handles::live_type(ard) != Some(HandleType::Desc) {
+fn bind_ard_column(
+    ard: SqlHandle,
+    ard_is_explicit: bool,
+    binding: ColumnBinding,
+) -> Result<(), ()> {
+    if ard_is_explicit && crate::handles::live_type(ard) != Some(HandleType::Desc) {
```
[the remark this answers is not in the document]
saurabh500: Agreed, and this is the same root cause as the `fetch_scroll.rs` thread: `SQLDisconnect` can free the statement and its four implicit descriptors while a call on that statement is still in flight, so the implicit ARD is not immune to concurrent destruction and the registry lookup was doing real work for it.

Reverted `bind_col.rs` wholesale in 0da952f3 -- every change in this file was part of that optimization, so all three call sites (`bind_ard_column`, `unbind_ard_column`, and `sql_free_stmt_unbind_safe` at the lines you flagged) are back to the unconditional check, and the `ard_is_explicit` parameter is gone along with the two test updates it required.

Worth noting for the PR's purpose: this file was where the measured win lived, since `bind_cycle` pays 17 of these lookups per row at rowset size 1. Correctness wins, so I am taking the loss here. If the remaining two optimizations do not close the gap, the follow-up is to make `LIVE_HANDLES` itself cheaper (it is a single process-global `Mutex<HashMap>`) rather than to skip it -- that keeps the guard while removing the contention. I will file that separately rather than widening this PR.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -989,11 +957,38 @@ fn fetch_scroll_safe(
...
+        // These locks are needed only for deferred default bindings and are
+        // acquired sequentially, after the ARD snapshot has been released.
+        let odbc_version = {
+            let env = stmt.parent_dbc().parent_env();
+            let Ok(env_state) = env.inner.lock() else {
+                error!("SQLFetchScroll: env mutex poisoned");
+                if let Ok(mut stmt_state) = stmt.inner.lock() {
+                    stmt_state.clear_state(STMT_STATE_FETCH_IN_PROGRESS);
+                }
+                return SQL_ERROR;
```
David-Engel, line 981: This exit returns `SQL_ERROR` without posting a diagnostic, while the two paths just above it (the `live_type` check and the ARD-poison branch) both post `HY000 "Internal error reading column bindings"`. `free_errors` already cleared the record list at line 821, so the application sees `SQL_ERROR` and then `SQL_NO_DATA` from `SQLGetDiagRec`. Before this change the ENV read happened before the statement was claimed, so this state was not reachable. You already hold the stmt lock here to clear the guard — posting the same `HY000` there would keep the three failure paths uniform.

mssql-odbc/src/handles/stmt.rs
```diff
@@ -1052,12 +1061,35 @@ impl StmtState {
...
+                    })
+                    .collect::<Arc<[_]>>()
+            });
     }
 
     /// Clears result metadata and every cache derived from it.
     pub(crate) fn clear_result_metadata(&mut self) {
         self.column_metadata.clear();
         self.column_names_utf16.clear();
+        self.plp_columns = None;
```
David-Engel, line 1092: This reset plus the rebuild above is the whole safety argument for the cache, and a stale entry surviving into a different result set would mis-decode a bound LOB silently rather than failing. The nearest existing test, `clearing_result_metadata_clears_cached_column_names`, uses `int_columns(1)`, so adding a `plp_columns.is_none()` assertion there would pass vacuously. Worth a test built on PLP metadata (`mixed_lob_columns` in `fetch_scroll.rs`'s test module is the right shape) that asserts `Some` after `begin_result_set`, `None` after `clear_result_metadata`, and `None` after a following `begin_result_set` with a non-PLP set — that last transition is the one where a surviving cache would actually be consulted.

David-Engel (commented):
**Automated review** — generated by GitHub Copilot on behalf of @David-Engel. This is **not** an approval and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

## Summary

Reviewed at `0da952f3` against merge base `5c412918`. The diff at this head is two files, not three: `bind_col.rs` and the implicit-ARD registry skip were reverted in `0da952f3` after the Copilot threads, so what remains is (a) gating `SQL_C_DEFAULT` resolution — and the ENV lock plus the `Vec<SqlSmallInt>` it needed — on some binding actually being `SQL_C_DEFAULT`, and (b) hoisting per-column PLP encoding info into a `StmtState::plp_columns` cache rebuilt in `refresh_metadata_caches`.

Both remaining changes look behavior-preserving to me.

On the lock-ordering question raised in the description: the reasoning holds. `desc_state` is a block-local guard dropped when the `let mut bindings = { ... }` block ends, so by the time the `SQL_C_DEFAULT` branch runs no DESC lock is held, and the ENV and STMT locks inside it are acquired and released sequentially rather than nested. That satisfies both `.github/instructions/mssql-odbc.instructions.md` rules — parent-before-child and never holding STMT while taking DESC. Gating on `bindings` is also exactly equivalent to the unconditional form, since `resolve_default_bindings` short-circuits on any binding whose `target_type != SQL_C_DEFAULT` and indexes `column_sql_types` with `.get()`.

On the PLP cache: I traced every writer of `column_metadata` reachable outside `#[cfg(test)]`. `begin_result_set` / `begin_batch` are the only production assignment sites and both go through `refresh_metadata_caches`; `catalog.rs` and `get_type_info.rs::rename_type_info_columns` mutate names/flags in place and then call `refresh_metadata_caches` explicitly; `get_type_info.rs::clear_type_info_nullable` only clears a nullable flag, which feeds neither `plp_encoding()` nor `get_encoding_type()`. `clear_result_metadata` resets the cache. The remaining `s.column_metadata = ...` sites are all inside test modules. `column_count` and `plp_columns` are read under the same lock acquisition, so the in-range invariant the `deliver_bound_plp` guard comment relies on is preserved. Collapsing "no PLP columns" to `None` is equivalent to a vector of all-`None`, since `plp_encoding()` returns `None` for exactly the columns `is_plp()` rejects.

## Verification

`cargo clippy --workspace --all-features --all-targets -- -D warnings` clean, and `cargo nextest run -p mssqlodbc --lib --no-fail-fast` green at 1359/1359 on this head, in an isolated worktree with a dedicated `CARGO_TARGET_DIR`. No live SQL Server, so the bound-fetch behavior itself was verified by reading rather than executing. Required CI is green on `0da952f3`, including `GH-mssql-odbc Linux performance regression`; `coverage-report` is still pending.

## Findings

### Suggestion

**`fetch_scroll.rs:971-975` — the new ENV-poison exit returns `SQL_ERROR` with no diagnostic record.** Its two immediate neighbours, the `live_type` failure and the ARD-poison exit, both post `HY000 "Internal error reading column bindings"` before returning. This one clears `STMT_STATE_FETCH_IN_PROGRESS` and returns bare. Because `free_errors` already ran at line 821, the application gets `SQL_ERROR` followed by `SQL_NO_DATA` from `SQLGetDiagRec`. Before this PR the ENV read happened up front and could not reach this state at all, so the diagnostic-less failure is new. Posting the same `HY000` in the `if let Ok(mut stmt_state)` block you already open would make the three paths uniform. (The `stmt.inner.lock()` failure just below genuinely cannot post, and matches the pre-existing precedent at line 994, so that one reads fine.)

**`handles/stmt.rs:1064-1092` — the cache-invalidation argument is load-bearing but untested against real PLP metadata.** The correctness of the whole PLP change rests on `plp_columns` never surviving into a result set it does not describe, and a stale entry would mis-decode a bound LOB silently rather than failing. The nearest existing test, `clearing_result_metadata_clears_cached_column_names`, builds `int_columns(1)`, so extending it to assert `plp_columns.is_none()` would pass vacuously. A unit test built on PLP metadata — `mixed_lob_columns` in `fetch_scroll.rs`'s test module is already the right shape — asserting `Some` after `begin_result_set`, `None` after `clear_result_metadata`, and `None` after a subsequent `begin_result_set` with a non-PLP set would guard it. The PLP-to-non-PLP transition is the one worth pinning, since that is the case where a surviving cache would be used rather than ignored.

**The PR description no longer matches the diff.** It still presents the implicit-ARD registry skip as the first of three changes, still says "This PR removes all three", still claims two `bind_col` tests were updated for an `ard_is_explicit` parameter, and the "Worth a careful look" section asks reviewers to scrutinise reasoning that `0da952f3` withdrew. It also says the perf pipeline has not been re-run, but `GH-mssql-odbc Linux performance regression` has since passed on this head. Worth refreshing before this leaves draft, both because it is what a human reviewer reads first and because the remaining win is now smaller than the description implies — by your own note, `bind_col.rs` was where most of the `bind_cycle` gain lived.

### Nit

**`fetch_scroll.rs:1689`** — the guard comment on `deliver_bound_plp` still says `plp_columns` "snapshots the statement's copy, whose length also fixes `column_count`, so the lookup is always in range". That is now only literally true when the cache is `Some`; the `None` case is in range trivially rather than by that argument. The conclusion is unchanged, so this is purely wording.

mssql-odbc/src/api/fetch_scroll.rs
saurabh500: Good catch, and your read on why it became reachable is exactly right: gating the `SQL_C_DEFAULT` setup on the bindings is what moved the ENV read from before validation to after the statement is claimed, so this path did not previously exist in a state where the application could observe it.

Fixed in c5d23c42 -- it now posts the same `HY000 "Internal error reading column bindings"` alongside clearing the guard, so all three failures in this block are uniform.

I left the `stmt.inner.lock()` failure just below it (the `column_sql_types` read) returning bare `SQL_ERROR`, since posting a diagnostic there requires the very lock that just failed. That one predates this change in shape and has nowhere to post.

mssql-odbc/src/handles/stmt.rs
saurabh500: Agreed on both the risk and the vacuousness problem -- `int_columns(1)` would have made a `plp_columns.is_none()` assertion there pass no matter what the code did.

Added `plp_column_cache_does_not_outlive_its_result_set` in c5d23c42, built on `mixed_lob_columns(2)` as you suggested. It asserts all three transitions: `Some` of length 3 after `begin_result_set` with the int columns mapping to `None` and the `nvarchar(max)` column to `Utf16Text`, `None` after `clear_result_metadata`, and `None` again after a following `begin_result_set(int_columns(2))`.

I mutation-checked it rather than assuming: removing `self.plp_columns = None` from `clear_result_metadata` makes it fail on the second assertion, so it is actually load-bearing. The third transition is additionally covered by construction, since `refresh_metadata_caches` only builds the cache under `.any(ColumnMetadata::is_plp)` and so stores `None` for a non-PLP set.

David-Engel (commented):
**Automated review** — generated by GitHub Copilot on behalf of @David-Engel. This is **not** an approval and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

## Summary

Reviewed at `8b280d81` against merge base `5c412918`. **No findings.** The three items from the previous automated pass at `0da952f3` are all addressed, and I re-verified the change independently rather than only checking that something changed.

The diff is two files and two behavior-preserving optimizations: gating `SQL_C_DEFAULT` resolution (and the ENV lock plus the `Vec<SqlSmallInt>` it needs) on some binding actually being `SQL_C_DEFAULT`, and hoisting per-column PLP wire/text encodings into a `StmtState::plp_columns` cache rebuilt in `refresh_metadata_caches`.

## What I verified

**The `SQL_C_DEFAULT` gate is exactly equivalent to the unconditional form.** `resolve_default_bindings` (`fetch_scroll.rs:764-772`) `continue`s on any binding whose `target_type != SQL_C_DEFAULT` and indexes `column_sql_types` with `.get()`, so skipping the call when no binding is `SQL_C_DEFAULT` cannot change an outcome.

**Lock ordering holds.** `desc_state` is a block-local guard dropped when the `let mut bindings = { ... }` block ends, so no DESC lock is held when the ENV and STMT locks are taken, and those two are acquired and released sequentially rather than nested. That satisfies both rules in `.github/instructions/mssql-odbc.instructions.md:525-542` — parent-before-child, and never holding a STMT lock while acquiring a DESC lock. Since the thread never holds two locks at once in this block, there is no new cycle to order against.

**The PLP cache cannot go stale.** Outside `#[cfg(test)]`, `column_metadata` is assigned only at `stmt.rs:1050` (`begin_result_set`, which `begin_batch` delegates to) and mutated in place only by `catalog.rs::apply_catalog_metadata` and `get_type_info.rs::rename_type_info_columns`, both of which call `refresh_metadata_caches` immediately after; `clear_type_info_nullable` touches only `flags`, which feeds neither `plp_encoding()` nor `get_encoding_type()`. `clear_result_metadata` resets it. Collapsing "no PLP columns" to `None` is equivalent to an all-`None` vector, since `plp_encoding()` returns `Some` for exactly the columns `is_plp()` accepts (`mssql-tds/src/query/metadata.rs:95-98`). `column_count` and `plp_columns` are read under the same lock acquisition, so the in-range invariant the corrected `deliver_bound_plp` comment relies on is preserved.

**The new test is load-bearing, and so is the `SQL_C_DEFAULT` gate.** Mutation-checked both, independently of your own check:
- `.any(ColumnMetadata::is_plp)` → `.any(|_| false)`: `plp_column_cache_does_not_outlive_its_result_set` fails. (Your check was removing the reset from `clear_result_metadata`; this one covers the build side too.)
- `.any(|binding| binding.target_type == SQL_C_DEFAULT)` → `.any(|_| false)`: three `fetch_scroll` tests fail (`a_default_binding_follows_the_declared_odbc_version`, `a_default_binding_set_through_the_descriptor_api_also_resolves`, `default_binding_resolves_from_current_result_metadata`).

`mixed_lob_columns(2)` is 2 ints plus one `nvarchar(max)` (`mssql-tds/src/test_client_support.rs:450-463`), so the test's `len() == 3` and `[2] == Utf16Text` assertions are the right shape rather than incidentally true.

**The perf numbers in the description were measured on the post-revert code.** Worth stating explicitly because the revert landed the same day: the `GH-mssql-odbc Linux performance regression` check on `13079f83` (which still had the `bind_col.rs` optimization) was **cancelled**, and buildId 172991 / `Build #20260906.2` — the run the results table cites — ran on `0da952f3`, where `bind_col.rs` was already back to unconditional. So the +4.17% / +6.22% figures reflect what this PR actually ships. The two commits since (`c5d23c42`, `8b280d81`) add a diagnostic on a poisoned-mutex path, a unit test, and a comment, none of which is on the measured path, so re-running perf is not needed for those.

## Verification

`cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` clean, `cargo nextest run -p mssqlodbc --lib --no-fail-fast` green at 1360/1360, plus the two mutation runs above — in the assigned worktree with a dedicated `CARGO_TARGET_DIR`. No live SQL Server here, so the bound-fetch behavior itself was verified by reading plus the cross-repo `mssql-python suite on mssql-odbc driver` job, which is green. All 19 checks pass on `8b280d81`, including `coverage-report` (84% diff coverage; the 13 uncovered lines are the two poisoned-mutex exits, and the report notes diff coverage is reported, not enforced).

The description now matches the diff, and `#507` is linked.

*This came from an unattended run; the findings were not checked by a human first.*

David-Engel (commented):

**Summary**
This PR does the right two optimizations: it defers `SQL_C_DEFAULT` resolution until it is actually needed and caches PLP metadata per result set so the hot fetch loop avoids re-deriving it on every call. I reviewed the changed fetch path, the new statement-state cache, the associated tests, and the PR checks.

**Verification**
- Reviewed the diff in a dedicated detached worktree for `8b280d81c3abec5055f6f531f8e98220e6ad65a2`.
- Read the changed paths in `mssql-odbc/src/api/fetch_scroll.rs` and `mssql-odbc/src/handles/stmt.rs`, plus the surrounding call sites and cache lifecycle.
- `gh pr checks 508 --repo microsoft/mssql-rs` passed.

**Blocking**
- `mssql-odbc/src/api/fetch_scroll.rs:987` — the new second `stmt.inner.lock()` can fail after `STMT_STATE_FETCH_IN_PROGRESS` has already been set, and this branch returns `SQL_ERROR` without clearing the guard. If the mutex is poisoned, the statement stays stuck in fetch-in-progress until the handle is torn down. Please clear the state and post the diagnostic before returning here, or avoid the second lock by deriving `column_sql_types` from the snapshot before releasing the statement lock.

saurabh500:
> `mssql-odbc/src/api/fetch_scroll.rs:987` — the new second `stmt.inner.lock()` can fail after `STMT_STATE_FETCH_IN_PROGRESS` has already been set, and this branch returns `SQL_ERROR` without clearing the guard.

Confirmed and fixed in 5a7e0c62. The finding is correct: that second acquisition was new in this PR, and the code it replaced had specifically snapshotted the SQL types under the first lock with a comment saying it did so "rather than re-locking the statement." I reintroduced the re-lock when I gated the `SQL_C_DEFAULT` work, and its poison branch did leave the guard set.

One correction on the first of the two suggested remedies:

> Please clear the state and post the diagnostic before returning here

That one isn't reachable. The lock that failed there *is* the statement lock, so the `if let Ok(mut stmt_state) = stmt.inner.lock()` pattern used by the neighbouring ENV-poison branch would fail too — clearing would need `PoisonError::into_inner`, which changes poison semantics for a path that only exists because the mutex is already poisoned. (The ENV branch can clear because only the ENV mutex is poisoned there and the STMT one is typically healthy.)

So I took the second suggestion, which is strictly better than either:

`column_sql_types` is now built in `refresh_metadata_caches` next to `column_names_utf16` and `plp_columns`, and an `Arc` of it is carried out in the snapshot that already exists. The second acquisition and its unreachable recovery path both disappear, and the types are derived once per result set instead of once per fetch — so this is also slightly faster than the unconditional `Vec` build it originally replaced.

Equivalence argument for the cache: `odbc_sql_type` reads only `data_type` and `type_info.length`. The three non-test sites that take `&mut column_metadata` (`catalog.rs:633`, `get_type_info.rs:324`, `get_type_info.rs:341`) mutate only `column_name` and `flags`, and two of them already call `refresh_metadata_caches()`. So the cached value cannot go stale against what the old inline derivation would have produced.

Added `column_sql_type_cache_does_not_outlive_its_result_set`, mutation-checked both ways: dropping the clear in `clear_result_metadata` fails it, and dropping the rebuild in `refresh_metadata_caches` fails it plus the three pre-existing `default_binding` tests — which also confirms the cache is genuinely on the `SQL_C_DEFAULT` resolution path rather than dead weight.

`cargo bfmt`, `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings`, and 1345/1345 `mssqlodbc` tests are green.

Note for anyone reading the file: the same "return `SQL_ERROR` on statement poison without clearing the guard" shape still exists at the `spare_get_data_row` acquisition a few lines below. That one is pre-existing, not introduced here, and has the identical unreachable-clear problem, so I've left it alone rather than widen this PR.

David-Engel (commented):
Automated review of `5a7e0c6` (full diff vs `origin/main`). No blocking issues, no suggestions, no nits.

What I verified beyond reading the diff:

- **Cache freshness.** `column_sql_types` and `plp_columns` are only correct if every path that mutates `column_metadata` rebuilds them. The production mutation sites are `begin_result_set`/`begin_batch` (which call `refresh_metadata_caches`), `apply_catalog_metadata` in `catalog.rs` and `rename_type_info_columns` in `get_type_info.rs` (both call it explicitly), and `clear_type_info_nullable`, which does not — but it only clears `COLMETA_NULLABLE_FLAG`, and `odbc_sql_type` reads `data_type`/`type_info.length` while `plp_encoding`/`get_encoding_type` read the type info and collation, so neither cache can go stale there. `clear_result_metadata` resets both.
- **Lock ordering.** The ENV acquisition moved after the STMT lock is released, which reads as an ordering change but is not one: the STMT guard is dropped at the end of the snapshot block and the ARD guard at the end of the bindings block, so the ENV lock is taken with nothing held. The `let ... else` poison branches hold no ENV guard when they re-take the STMT lock, so the parent-before-child rule in `mssql-odbc.instructions.md` is still satisfied. The new poison branch also clears `STMT_STATE_FETCH_IN_PROGRESS` and posts a diagnostic, matching its two sibling early returns.
- **Gate equivalence.** `resolve_default_bindings` is a no-op for any binding whose `target_type != SQL_C_DEFAULT`, so gating the whole block on `bindings.iter().any(...)` cannot change the resolved bindings. The only observable difference is that a poisoned ENV mutex no longer fails a fetch with no deferred defaults, which is strictly better.
- **PLP lookup.** `plp_columns` is `Some` only when a column is PLP, and then its length is `column_metadata.len()`, so the `column - 1` lookup stays in range; when it is `None`, no column is classified `PlpStreaming` and the `deliver_bound_plp` guard is unreachable as documented.
- **Tests.** `cargo nextest run -p mssqlodbc` — 1361/1361 pass locally. The two new cache-lifecycle tests cover the transitions that matter (PLP → non-PLP, wider → narrower), and the existing `default_binding_*` tests still exercise the now-conditional resolution path.

Required CI is green on this head, including the ADO validation run.

PR #511, opened 2026-09-07
Add mssql-odbc fuzzing and fix datetimeoffset fetch overflow

saurabh500:
## Description

Adds coverage-guided (libFuzzer) fuzzing for the `mssql-odbc` crate, mirroring the existing `mssql-tds/fuzz` setup, wires the new targets into the existing Azure DevOps fuzz pipeline, and fixes a bug the new targets surfaced.

**Fuzz infrastructure**
- New `#[cfg(fuzzing)]` `fuzz_support` module re-exposes internal parse/convert helpers as safe wrappers. The `fuzzing` cfg is set only by cargo-fuzz, so the shipped driver's FFI surface is unchanged.
- New `mssql-odbc/fuzz/` crate registering eight libFuzzer targets:
  - `fuzz_connection_string`, `fuzz_numeric_parse`, `fuzz_datetime_parse`, `fuzz_read_utf16`, `fuzz_transcode`
  - `fuzz_fetch_convert` — the `SQLGetData` conversion matrix (`ColumnValues` → every fixed-width `SQL_C_*` target)
  - `fuzz_bound_param` — the `SQLExecute` parameter read/convert path
  - `fuzz_ffi_execute` — the FFI-layer target (see below)
- The `fuzz_fetch_convert`, `fuzz_bound_param` and `fuzz_ffi_execute` C-type / SQL-type target lists are derived at runtime from the production routing (the `fetch_convert` C-target predicates and the `conversion_matrix` SQL-target tables) rather than hand-copied, so they exercise the full accepted conversion matrix and pick up new converters automatically as the driver adds them.
- The lib now builds as `["cdylib", "rlib"]` so the fuzz crate can link it as an ordinary Rust dependency; the shipped dynamic library is unaffected.

**FFI-layer target — `fuzz_ffi_execute`**
The other seven targets fuzz individual helpers. `fuzz_ffi_execute` instead drives the driver's real `extern "C"` entry points end to end: it allocates a fresh ENV+DBC+STMT through the exported `SQLAllocHandle`/`SQLSetEnvAttr`, installs an in-memory `TdsClient` whose transport replays the fuzzer's bytes as the server's TDS response, then runs `SQLExecDirectW` → `SQLNumResultCols` → `SQLFetch` → `SQLGetData` (first input byte selects the `SQLGetData` target C type). This crosses the C ABI the shipped driver actually exposes and drives the statement/cursor state machine over a hostile/corrupt server stream, not just a leaf converter.

Every entry point wraps its body in a `catch_unwind` (`ffi_entry!`) that converts a panic to `SQL_ERROR` in production. This does **not** hide logic panics from the fuzzer: libfuzzer-sys installs a panic hook that runs at the panic site — before any unwinding, so before that `catch_unwind` — and `abort()`s, which libFuzzer records as a crash. Verified empirically by injecting a panic behind `ffi_entry!` and confirming libFuzzer reports a deadly signal (then reverted). A 30-minute local run reached ~6.8k coverage points with no crashes/timeouts/OOMs.

**Pipeline wiring**
- `fuzz-template.yml` gains a `fuzzDir` parameter (default `mssql-tds/fuzz`), so the reusable run template is no longer hardcoded to the tds crate. Existing tds jobs are unchanged.
- The scheduled `Fuzz` stage gains two odbc jobs: `Fuzz_Odbc_Ffi_Linux` (runs `fuzz_ffi_execute` for 30 min) and `Fuzz_Odbc_Additional_Targets_Linux` (the seven leaf targets, 5 min each, serial).
- The per-PR fuzz compile check now also builds the odbc fuzz crate (`build-template.yml` + `containerized-build.sh`), and `dev/build-fuzz.sh` builds both crates locally.

**Bug fix — datetimeoffset fetch overflow (DoS)**
`datetimeoffset_parts` in `conversion/fetch_convert.rs` computed the UTC tick total in `i64`. A corrupt or hostile server can decode a `time_nanoseconds` far outside a single day — the TDS decoder derives the time-field width from the row's length byte and the scale from the column metadata without cross-checking them, so `scale_time_value` can inflate a 5-byte wire value past the `i64` range. Applying the stored offset then overflowed the `i64` tick addition and **panicked across the `SQLGetData` FFI boundary** (a DoS under debug assertions; a silently wrong value on release builds).

The fix computes the intermediate in `i128`, so the existing `0..=MAX_DAYS_SINCE_0001` day-range check rejects such values as an out-of-range conversion (`None` → normal ODBC conversion error) instead of overflowing. `i128` yields identical results for all valid inputs. Found by `fuzz_fetch_convert`, deterministically reproduced, and re-verified fixed (the saved crash input now returns cleanly). The sibling paths (`time_parts`, `datetime2_parts`) have no offset add, so `fuzz_fetch_convert` finds no panic in them — but that only rules out panics, not wrong values. `hms_from_ticks_100ns` truncates an out-of-range tick count with `as u16` rather than erroring, so an out-of-day-range time is still returned wrong under `SQL_SUCCESS`. That is a pre-existing correctness bug, out of scope for this PR and tracked in #526.

## Related Issues

[AB#45636](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/45636) — https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/45636

Follow-ups filed during review: #526 (pre-existing out-of-range time truncation in `hms_from_ticks_100ns`) and #529 (SQLSTATE `07006` vs. msodbcsql's `22008` for datetimeoffset overflow — the `ConvError::Restricted` → `07006` mapping is pre-existing on `main`; remapping needs a distinct out-of-range temporal variant + per-target cross-driver verification).

## Checklist

- [x] `cargo bfmt` passes — run for `mssql-odbc` + the fuzz crate
- [x] `cargo bclippy` passes — run for `mssql-odbc` (`-p mssqlodbc -D warnings`); fuzz crate proven clean via `cargo fuzz build`
- [x] `cargo btest` passes — `mssql-odbc` suite (1404 tests), including the new regression test
- [x] New/changed functionality has tests — `datetimeoffset_overflowing_time_is_rejected_not_panicked` (mechanism-break verified: reverting to `i64` makes it panic while the other datetimeoffset tests pass)
- [ ] Public API changes are documented — N/A; no public API change (`fuzz_support` is `#[cfg(fuzzing)]`-only)

<sub>Validation was run targeted for the affected crate (in WSL, nightly + cargo-fuzz). Full-workspace `cargo bfmt` / `cargo bclippy` / `cargo btest` (and the separate `mssql-py-core` runs) will be run before marking ready-for-review. A fuzz README and seed corpora are intentionally left as follow-ups.</sub>


mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,384 @@
...
+    SQL_TINYINT,
+    SQL_SMALLINT,
+    SQL_INTEGER,
+    SQL_BIGINT,
+    SQL_CHAR,
+    SQL_VARCHAR,
+    SQL_WVARCHAR,
+    SQL_BINARY,
+    SQL_VARBINARY,
+];
```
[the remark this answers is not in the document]
saurabh500: Fixed in 754fb35a. `PARAM_SQL_TYPES` is now built from `conversion_matrix`'s own target rows (`CHARACTER_SQL_TARGETS` ++ `BINARY_SQL_TARGETS` ++ `INTEGER_SQL_TARGETS`, now `pub(crate)`), so it tracks the implemented set instead of a hand-copied list — that adds `SQL_LONGVARCHAR`, `SQL_WCHAR`, `SQL_WLONGVARCHAR`, and `SQL_LONGVARBINARY`. I also completed `PARAM_C_TYPES` with the legacy `SQL_C_TINYINT`/`SQL_C_SHORT`/`SQL_C_LONG` integer spellings.

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,384 @@
...
+            int_val: i32::from_le_bytes(cur.arr()),
+        }),
+        11 => {
+            let enc = build_encoding(cur);
+            ColumnValues::String(SqlString::new(cur.rest(), enc))
+        }
+        12 => {
+            let days = u32::from_le_bytes(cur.arr()) % 3_652_059;
+            ColumnValues::Date(SqlDate::create(days).unwrap())
+        }
```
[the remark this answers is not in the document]
saurabh500: Fixed in 754fb35a — case 12 now does `SqlDate::create(days).map_or(ColumnValues::Null, ColumnValues::Date)`, matching the `build_decimal` map_or pattern already used in cases 7/8. A constructor failure yields `Null` instead of panicking.

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,384 @@
...
+    fn take(&mut self, n: usize) -> Vec<u8> {
+        let end = self.pos.saturating_add(n).min(self.data.len());
+        let mut v = self.data.get(self.pos..end).unwrap_or(&[]).to_vec();
+        self.pos = end;
+        v.resize(n, 0);
+        v
+    }
+
+    fn arr<const N: usize>(&mut self) -> [u8; N] {
+        self.take(N).try_into().unwrap()
```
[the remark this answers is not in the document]
saurabh500: Fixed in 754fb35a — `arr()` now falls back to `[0u8; N]` via `unwrap_or` instead of `unwrap()`, so there's no panic site. It stays unreachable today (`take(N)` always resizes to exactly N), but the fallback is encoded now rather than assumed.

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,384 @@
...
+    SQL_C_SBIGINT,
+    SQL_C_UBIGINT,
+    SQL_C_BIT,
+    SQL_C_FLOAT,
+    SQL_C_DOUBLE,
+    SQL_C_GUID,
+    SQL_C_TYPE_DATE,
+    SQL_C_TYPE_TIME,
+    SQL_C_TYPE_TIMESTAMP,
+];
```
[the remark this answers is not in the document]
saurabh500: Fixed in 754fb35a. The reachable set is now derived at run time by filtering `FETCH_TARGET_CANDIDATES` through the same predicates `convert_typed_c` routes on (`is_integer_c_target || is_float_c_target || == SQL_C_GUID || is_datetime_c_target`), so it can't fall behind production. That pulls in the legacy `SQL_C_TINYINT`/`SQL_C_SHORT`/`SQL_C_LONG` spellings, the legacy date/time spellings, and both SS temporal targets including `SQL_C_SS_TIMESTAMPOFFSET` — the 8 fixed-width arms that were unreachable. The FFI `SQLGetData` set reuses that routed set plus the char/wchar/binary targets. `SQL_C_NUMERIC` is listed as a candidate but filtered out until a converter lands.

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,556 @@
...
+    unsafe { SQLFreeHandle(SQL_HANDLE_ENV, env) };
+}
+
+/// Installs an in-memory, already-connected `TdsClient` over `server_bytes` on
+/// `dbc`, so the execute path reads its response from the fuzz input instead of
+/// a socket.
+fn install_fuzz_client(dbc: SqlHandle, server_bytes: &[u8]) {
+    let client = create_fuzz_tds_client(FuzzPacketReader::from_data(server_bytes), 4096);
+    let dbc_ref = unsafe { handle_from_raw::<DbcHandle>(dbc) };
+    let mut state = dbc_ref.inner.lock().unwrap();
```
saurabh500: **Suggestion** — this is the one `.unwrap()` of the class Copilot flagged twice that survived the fix round.

`#[cfg(fuzzing)]` is not `#[cfg(test)]`, so the crate rule still applies: `.github/instructions/mssql-odbc.instructions.md` L359-364 says never `.unwrap()` outside `#[cfg(test)]`, and L364 names the canonical `Mutex` form explicitly. That file also explains how it got missed — `cargo bclippy` never compiles this module (no `--cfg fuzzing`), and the CI fuzz gate runs `cargo +nightly check`, not clippy, so nothing lints it at all.

The cost isn't a driver defect: this DBC was allocated a few lines up and is never shared, so poisoning is unreachable today. The cost is misattribution. If anything in the harness ever panics while holding that lock, the *next* input's `install_fuzz_client` aborts and libFuzzer files a crash against the driver for a harness bug.

```rust
let Ok(mut state) = dbc_ref.inner.lock() else {
    return;
};
```

Returning early is safe here — `run_exec_fetch` on an unconnected DBC just takes the not-connected error path.

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,556 @@
...
+    SQL_C_TIMESTAMP,
+    SQL_C_SS_TIMESTAMPOFFSET,
+    // Implemented C type with no fetch converter yet — filtered out below until
+    // production starts routing it.
+    SQL_C_NUMERIC,
+];
+
+/// Exactly the targets `get_data::convert_typed_c` hands to a fixed-width
+/// converter, spelled with the same predicates it routes on.
+fn routes_fixed_width_fetch(target: SqlSmallInt) -> bool {
```
saurabh500: **Suggestion** — this is character-for-character `is_typed_c_target` from `api/get_data.rs:2571-2576`, which is the predicate `SQLGetData` actually gates `convert_typed_c` on (`get_data.rs:416`, `:1181`).

The doc comment is right that `get_data` is private, but that is a one-word change: `api/mod.rs:26` is `mod get_data;` while eight of its siblings in the same block (`fetch`, `fetch_scroll`, `free_handle`, `get_connect_attr`, `odbc_types`, `set_connect_attr`, `set_desc_field`, `set_env_attr`) are already `pub(crate) mod`.

Worth doing because it is the same drift the last round set out to close. `FETCH_TARGET_CANDIDATES` was reworked so the fuzzer could not fall behind production's routing table — but the routing *predicate* is still a parallel copy. Add a fifth arm to `is_typed_c_target` and the candidate list picks the new type up while this filter silently drops it, so the target quietly stops covering the thing that just changed.

Either:

- make it `pub(crate) mod get_data;` and call `is_typed_c_target` directly, or
- move `is_typed_c_target` down into `fetch_convert.rs` next to the three predicates it is composed of (all three are already imported here) and have `get_data` use it from there.

Either way `routes_fixed_width_fetch` goes away.

mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -355,16 +355,23 @@ pub(crate) fn datetime2_parts(datetime: &SqlDateTime2) -> DateTimeParts {
...
+        + i128::from(datetime.offset) * 60 * 10_000_000;
+    let days = i128::from(datetime.datetime2.days) + utc_ticks.div_euclid(ticks_per_day);
+    if !(0..=i128::from(MAX_DAYS_SINCE_0001)).contains(&days) {
         return None;
     }
 
-    let date = civil_from_days_since_0001(days);
-    let t = hms_from_ticks_100ns(utc_ticks.rem_euclid(TICKS_PER_DAY) as u64);
+    let date = civil_from_days_since_0001(days as i64);
+    let t = hms_from_ticks_100ns(utc_ticks.rem_euclid(ticks_per_day) as u64);
```
saurabh500, line 374: **Suggestion / follow-up** — the fix is right and the day-range check does the work. What I want to flag is what the clean fuzz run on the siblings does and does not tell you.

The description says the `datetime2` and `time` paths "were fuzzed without incident". True for panics — and it cannot be more than that, because `hms_from_ticks_100ns` truncates with `as u16` / `as u8` instead of overflowing, so a fuzzer has nothing to trip over.

Verified locally with a throwaway test (since removed):

```rust
ColumnValues::DateTime2(SqlDateTime2 {
    days: 0,
    time: SqlTime { time_nanoseconds: i64::MAX as u64, scale: 7 },
})
```

targeting `SQL_C_TYPE_TIMESTAMP` returns `Ok(Exact)` — i.e. `SQL_SUCCESS` — with:

```
SqlTimestampStruct { year: 1, month: 1, day: 1, hour: 24554, minute: 48, second: 5, fraction: 477580700 }
```

`hour: 24554`, handed back in a `SQLUSMALLINT` under a success code. `time_parts` and `datetime2_parts` never check `time_nanoseconds < TICKS_PER_DAY`, so the exact decoder confusion your comment above describes lands there as a wrong value rather than a panic — which is the case the instructions' "a success code is a promise about the caller's buffer" rule exists for.

Pre-existing and not introduced here, so I would leave the code out of this PR — but a follow-up issue seems right, and I would soften the description's claim about the siblings so it is not read as coverage.

(Same class, much narrower, on the path you did fix: a `time_nanoseconds` that is an exact multiple of a day still passes the day check and silently shifts the date. Also pre-existing.)

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,556 @@
...
+static FFI_GETDATA_TARGETS: LazyLock<Vec<SqlSmallInt>> = LazyLock::new(|| {
+    let mut targets = ROUTED_FETCH_TARGETS.clone();
+    targets.extend_from_slice(&[SQL_C_CHAR, SQL_C_WCHAR, SQL_C_BINARY]);
+    targets
+});
+
+/// Drive the result-fetch conversion matrix: a decoded column value written to
+/// a fixed-width `SQL_C_*` target. Mirrors `get_data::convert_typed_c`'s routing
+/// (integer / float / GUID / datetime) without needing the private `get_data`
+/// module. The 64-byte target overtops every fixed C struct (the widest is the
```
saurabh500: **Nit** — `SqlSsTimestampoffsetStruct` (`api/odbc_types.rs:714-724`) is 20 bytes, not 16: eight 2-byte fields plus a 4-byte `fraction`. It is in `ROUTED_FETCH_TARGETS`, so it is in scope for this buffer.

64 is still comfortably above it, so nothing is wrong with the code — but this is the sentence that justifies the constant, so it is worth being right.

saurabh500 (commented):
## Summary

Solid PR. The harness mirrors the existing `mssql-tds/fuzz` layout closely enough that the two read as one system rather than a bolted-on second one, and the `datetimeoffset` fix is a real reachable-from-the-wire panic across the FFI boundary, not a theoretical one. All four Copilot findings from the first round were properly closed in 754fb35a — I re-derived `FETCH_TARGET_CANDIDATES` against `is_integer_c_type` / `is_float_c_target` / `is_datetime_c_target` and it is now exactly their union, with `SQL_C_NUMERIC` correctly excluded since production gates `convert_typed_c` on `is_typed_c_target`.

**Verdict: no blocking issues.** Three suggestions and a nit inline; none need to hold the merge, and the `datetime2`/`time` one is pre-existing and probably a follow-up issue rather than this PR's job.

### What I ran

- `cargo nextest run -p mssqlodbc --lib -E 'test(datetimeoffset)'` → 8 passed.
- **Mechanism-break on the new regression test.** Reverted `datetimeoffset_parts` from `i128` back to `i64` and re-ran: `datetimeoffset_overflowing_time_is_rejected_not_panicked` failed with `attempt to add with overflow` at `fetch_convert.rs:366`, while the other seven datetimeoffset tests stayed green. So the test pins the mechanism it names rather than passing for some unrelated reason — that is the part worth knowing about a new regression test, and it holds.
- A throwaway probe on the sibling `datetime2` path, removed afterwards — see the inline comment on `fetch_convert.rs`.

I did not exercise the fuzz targets themselves (no nightly + cargo-fuzz here); that is on the ADO legs.

### On the red check

`coverage-report` failed with `❌ Timeout: coverage artifact not found within 75 minutes`. That is the job waiting on the still-pending ADO validation build, not a coverage regression — the posted report is 100% diff / 93.6% overall. Worth re-running once ADO finishes rather than reading it as a real failure.

One process note: the description says full-workspace `bfmt`/`bclippy`/`btest` "will be run before marking ready-for-review", but the PR is already ready while ADO is still pending on `f42ef1b`. Worth confirming that green before pulling in human reviewers.

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,556 @@
...
+    SQL_C_LONG,
+    SQL_C_ULONG,
+    SQL_C_SBIGINT,
+    SQL_C_UBIGINT,
+];
+
+/// Every SQL target `params::conversion_matrix::is_supported_conversion` can
+/// reach, taken straight from the matrix's own target rows so the fuzzed set
+/// follows the implemented conversions instead of a hand-copied list.
+static PARAM_SQL_TYPES: LazyLock<Vec<SqlSmallInt>> = LazyLock::new(|| {
```
David-Engel, line 414: **Suggestion** — the last round closed the "hand-copied list falls behind production" gap on the *fetch* side by deriving `ROUTED_FETCH_TARGETS` from the routing predicates. The same gap is still open on the *param* side, and the doc comments claim it isn't.

This says "Every SQL target `is_supported_conversion` can reach, taken straight from the matrix's own target rows." It's three of the six rows. `conversion_matrix.rs` also has `DECIMAL_SQL_TARGETS`, `TEMPORAL_SQL_TARGETS`, and `CHARACTER_PAYLOAD_SQL_TARGETS`, plus the inline `SQL_SS_VARIANT` / `SQL_BIT` / `SQL_REAL|SQL_FLOAT|SQL_DOUBLE` / `SQL_GUID` rows — 14 reachable SQL targets that `fuzz_bound_param` can never draw. `PARAM_C_TYPES` above has the mirror problem: its doc says "Every C type `params::conversion_matrix` has a row for", but it omits `SQL_C_BIT`, `SQL_C_FLOAT`, `SQL_C_DOUBLE`, `SQL_C_GUID`, and all five temporal C types, each of which has its own arm.

The cost isn't hypothetical, and the repo's own tests pin it. `SQL_C_CHAR` and `SQL_C_WCHAR` *are* in `PARAM_C_TYPES`, so the SQL side is the only thing blocking these pairings — and `params::conversion_matrix::tests::a_character_c_type_reaches_the_temporal_sql_types` and `::only_a_character_c_type_reaches_decimal_xml_or_variant` both pass today, asserting exactly the pairings this target cannot select. That leaves these arms of `bound_param_to_value` (`conversion/param_convert.rs:186-245`) unreachable:

```
(NarrowText | WideText, SqlFamily::DateTime) => datetime_from_text(..)
(NarrowText | WideText, SqlFamily::Decimal)  => decimal_from_text(..)
(NarrowText | WideText, SqlFamily::Xml)      => ..
(NarrowText | WideText, SqlFamily::Variant)  => ..
```

`datetime_from_text` and `decimal_from_text` are the input-side twins of the path this PR just fixed a panic in: attacker- or app-controlled *text* parsed into a temporal/numeric wire value, with the same scale/width arithmetic. Those are the two I'd most want a fuzzer sitting on, and they're the two it can't reach.

The fix is the same shape as the one already applied: make `DECIMAL_SQL_TARGETS`, `TEMPORAL_SQL_TARGETS`, and `CHARACTER_PAYLOAD_SQL_TARGETS` `pub(crate)` alongside their three siblings and concat all six here (plus `SQL_SS_VARIANT`), and extend `PARAM_C_TYPES` with the scalar and temporal C types. Unsupported pairings are already handled — `sql_family` / the `_ =>` arm rejects them — so widening both lists only adds reachable arms, it doesn't manufacture impossible states.

Two smaller things in the same function while you're here, both optional:

- `ind_mode` only produces `SQL_NULL_DATA`, `SQL_NTS`, `fuzz_len`, and `fuzz_len / 2`. It never produces a data-at-execution indicator (`SQL_DATA_AT_EXEC`, or anything `<= SQL_LEN_DATA_AT_EXEC_OFFSET`) or an otherwise-invalid negative length, so `read_indicator`'s rejection paths stay unfuzzed.
- If you'd rather not widen the lists in this PR, please at least soften the two doc comments — "every target the matrix can reach" reads as a coverage guarantee, and the next person deriving something from it will trust it.

David-Engel (commented):

## Summary

Reviewed the full diff against merge base `058aaf46`. The `datetimeoffset_parts` fix is correct — `i128` is wide enough that no `u64` `time_nanoseconds` plus any `i16` offset can overflow, `rem_euclid` keeps the `as u64` in range, and the `as i64` on `days` is guarded by the range check above it. The harness mirrors `mssql-tds/fuzz` closely and the pipeline wiring is a clean parameterization of the existing template with no target-name collisions between the two fuzz crates.

I read the earlier Copilot review (including the collapsed suppressed finding) and the two author self-review rounds, and I'm not re-filing anything already raised there — the `install_fuzz_client` `.unwrap()`, the `routes_fixed_width_fetch` / `is_typed_c_target` duplication, the `datetime2`/`time` truncation follow-up, and the `SqlSsTimestampoffsetStruct` size nit are all already on the record.

**One new finding, non-blocking.** Details inline.

### What I ran

- `cargo nextest run -p mssqlodbc --lib -E 'test(datetimeoffset) or test(conversion_matrix)'` → 21 passed, including the new `datetimeoffset_overflowing_time_is_rejected_not_panicked`. I did not repeat the mechanism-break; it's already documented in the self-review.
- I deliberately did not build the `fuzz/` crate — it pulls a dependency tree this review didn't audit, and every claim below is decidable by reading. The ADO legs cover it.

### On CI

`coverage-report` is red but the ADO validation build for `f42ef1b` is still pending, so that failure is the artifact-wait timeout, not a coverage regression. The three new fuzz jobs are also still pending — worth confirming `Fuzz_Odbc_Ffi_Linux` and `Fuzz_Odbc_Additional_Targets_Linux` actually go green before this gets human reviewers, since neither has ever run on a real agent.

mssql-odbc/src/fuzz_support.rs
saurabh500: Fixed — `install_fuzz_client` now does `let Ok(mut state) = dbc_ref.inner.lock() else { return; };` instead of `.unwrap()`, so a poisoned lock from a harness panic can't get filed as a driver crash. Good catch on the `cfg(fuzzing)` lint blind spot; I validated the change compiles under `cargo +nightly fuzz build`.

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
saurabh500: Fixed at the root: moved `is_typed_c_target` into `fetch_convert.rs` next to the three component predicates it's built from, deleted the fuzzer's copy, and `ROUTED_FETCH_TARGETS` now filters `FETCH_TARGET_CANDIDATES` through that same production predicate. Add a fifth arm and both the candidate list and the filter pick it up — no parallel copy to drift. Updated the `get_data` and `fetch_scroll` imports to the new home.

> [!NOTE]

mssql-odbc/src/conversion/fetch_convert.rs
saurabh500: Agreed — the clean run only proves no-panic, and it can't prove more because the truncation swallows the out-of-range value instead of tripping. Softened the description to say exactly that and to name the mechanism: `hms_from_ticks_100ns` wraps the hour with `as u16` and still returns `SQL_SUCCESS`. Reproduced the `hour: 24554` case locally. It's pre-existing and out of scope for this PR, so I filed #526 to fix it (range-check the ticks before the fields are trusted).

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
saurabh500: Fixed the comment — 20-byte `SqlSsTimestampoffsetStruct`, not 16. It turned out to matter beyond the wording: the same 20-byte width drove a padding fix in `fuzz_bound_param` (the value buffer was only padding 8 bytes — see the param-buffer thread).

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
saurabh500: Fixed both sides. `PARAM_C_TYPES` now carries every row — `SQL_C_BIT`/`SQL_C_FLOAT`/`SQL_C_DOUBLE`/`SQL_C_GUID` and all five temporal C structs (14→23) — and `PARAM_SQL_TYPES` is now derived by scanning `is_supported_conversion` over the full `i16` range instead of concatenating three of the six target consts, so it reaches every SQL target the matrix accepts and can't fall behind it. Reverted the three now-unused target consts to private, widened `ind_mode` to hit `read_indicator`'s data-at-exec/default-param/negative-length rejection paths, and fixed both drifted doc comments.

Widening the C types also caught a real harness bug: the value buffer only padded 8 bytes, but the added wide structs read up to 20 (datetimeoffset), so ASan flagged an over-read on short inputs. Padded to 32 so the buffer models a conforming application buffer for the widest fixed-width C type; re-fuzzed clean.

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,582 @@
...
+        };
+    }
+}
+
+/// Every C type `params::conversion_matrix::is_supported_conversion` has a row
+/// for: the character and binary buffers, the scalar bit/float/GUID spellings,
+/// all five date/time C structs, and every `type_rules::is_integer_c_type`
+/// variant (including the legacy `SQL_C_TINYINT`/`SQL_C_SHORT`/`SQL_C_LONG`
+/// spellings). `sql_type` is drawn independently, so unsupported pairings still
+/// exercise the bind-time rejection path.
```
David-Engel: **Suggestion** — this sentence, and its twin on `PARAM_SQL_TYPES` below, say drawing `sql_type` independently means "unsupported pairings still exercise the bind-time rejection path". They don't: nothing bind-time runs in this harness. The only production caller of `is_supported_conversion` outside tests is `api/bind_param.rs:236`, inside `SQLBindParameter`, and `fuzz_bound_param` builds a `BoundParam` literal and calls `bound_param_to_value` directly — the post-bind path. (`param_convert.rs:2927` also calls the matrix, but that is inside `#[cfg(test)] mod tests`, which starts at line 1345.)

What actually absorbs an unsupported pairing here is `bound_param_to_value`'s own `sql_family` / `_ =>` fallback, which is a different thing to claim coverage of. Two smaller consequences worth naming rather than leaving to the next reader:

- A pairing production rejects at bind time is not reachable in a real driver, so a panic found on one of those combinations would be a manufactured-impossible state — the same reasoning that correctly keeps `EncodingType::DelayedSet` out of `build_encoding`.
- The wording is the exact drift the last two rounds closed on `FETCH_TARGET_CANDIDATES` and `PARAM_SQL_TYPES`: a doc that reads as a coverage guarantee and isn't.

Cheapest fix is the wording, in both places — say the pairing falls through `bound_param_to_value`'s unsupported-conversion arm, not the bind-time gate. If you would rather keep the claim true, gating the draw with `is_supported_conversion(c_type, sql_type)` (falling back to a supported pair) confines the fuzzer to states `SQLBindParameter` can actually produce; genuinely fuzzing the bind-time rejection would mean driving `SQLBindParameter` itself, which is a separate target.

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,582 @@
...
+            time: u16::from_le_bytes(cur.arr()),
+        }),
+        18 => ColumnValues::Bytes(cur.rest()),
+        19 => ColumnValues::Uuid(uuid::Uuid::from_bytes(cur.arr())),
+        _ => ColumnValues::Null,
+    }
+}
+
+/// Fixed-width `SQL_C_*` fetch targets offered to the conversion fuzzers. The
+/// reachable subset is derived at run time by [`routes_fixed_width_fetch`] — the
```
David-Engel: **Nit** — stale after this commit: `routes_fixed_width_fetch` was deleted here in favor of the production `is_typed_c_target`, so this intra-doc link points at nothing (and `cargo doc` would flag it as a broken link). Suggest matching the wording already used on `ROUTED_FETCH_TARGETS` below:

```rust
/// reachable subset is derived at run time by [`is_typed_c_target`] — the exact
/// predicate `get_data::convert_typed_c` routes on — so the fuzzer tracks
```

mssql-odbc/fuzz/fuzz_targets/fuzz_bound_param.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Fuzz the bind-parameter read + convert path (`bound_param_to_value`) that
+//! `SQLExecute` runs over an application's value and length/indicator buffers.
+//! The wrapper backs the value buffer with the input plus eight trailing zero
```
David-Engel: **Nit** — this says "eight trailing zero bytes", but the padding was widened to 32 in this PR's head commit to cover the 20-byte `SqlSsTimestampoffsetStruct` (`fuzz_support.rs:445`). The wrapper's own doc comment was updated; this one was not. "plus trailing zero padding wide enough for the widest fixed-width C target" keeps the two in step without pinning a number that will drift again.

David-Engel (commented):

## Summary

Re-reviewed at head `4ca532ae` (the previous automated pass covered `f42ef1b9`). The new commit does what the last round's threads asked: `is_typed_c_target` moved from `api/get_data.rs` into `conversion/fetch_convert.rs` next to the three predicates it composes, with `get_data.rs:416`/`:1181` and `fetch_scroll.rs` importing it from the new home, so `ROUTED_FETCH_TARGETS` now filters through the production predicate instead of a parallel copy. `PARAM_C_TYPES` covers all 23 rows the matrix has, `PARAM_SQL_TYPES` is derived by scanning `is_supported_conversion` over the `i16` range (strictly better than concatenating the six target consts, and it let the three consts go back to private), the value buffer padding matches the 20-byte `SqlSsTimestampoffsetStruct`, and the `install_fuzz_client` lock is no longer `.unwrap()`ed. I confirmed every C type in `PARAM_C_TYPES` has a row in `conversion_matrix.rs:73-101`, and that `SQL_C_DEFAULT` is not among them, so the `debug_assert_ne!` at the top of `is_supported_conversion` can't fire during the scan.

The `datetimeoffset_parts` fix is unchanged and still correct: `u64` ticks plus an `i16` offset in minutes cannot leave `i128` range, the day check runs before the `as i64`, and `rem_euclid` keeps the `as u64` in range.

**No blocking issues.** One suggestion and two nits, all doc-accuracy in the fuzz harness — the same drift class this PR's last two rounds closed.

### What I ran

- `cargo clippy -p mssqlodbc --lib` with `RUSTFLAGS="--cfg fuzzing"` on Windows → clean, zero warnings from `mssql-odbc`. Worth doing because `cargo bclippy` never compiles this module (no `--cfg fuzzing`) and the PR's own CI gate is `cargo +nightly fuzz build` / `cargo +nightly check` on Linux, so nothing else lints `fuzz_support.rs` on a Windows host.
- `cargo nextest run -p mssqlodbc --lib -E 'test(datetimeoffset) or test(conversion_matrix)'` → 21 passed, including `datetimeoffset_overflowing_time_is_rejected_not_panicked`. I did not repeat the mechanism-break; it is documented in the self-review and the code under it is unchanged since.
- I did not build the `fuzz/` crate; it pulls a dependency tree this review did not audit, and every claim below is decidable by reading. The ADO legs cover it.

### On CI

Required checks are pending, not failing, for this head: the ADO validation build (`173533`) and `coverage-report` are both still running, and the Kerberos leg has already passed. The three new fuzz jobs are scheduled-stage jobs, so they will not report here — `Fuzz_Odbc_Ffi_Linux` and `Fuzz_Odbc_Additional_Targets_Linux` have still never run on a real agent, which is worth confirming before this gets human reviewers. The job shape itself matches the existing tds fuzz jobs (`CARGO_TARGET_DIR`, pool demands, serial template repeats) and the eight odbc target names do not collide with the nine tds ones, so the per-target crash artifact names stay unique.

mssql-odbc/src/fuzz_support.rs
saurabh500: You're right — there's no bind-time gate in this harness. Corrected both doc comments to say an unsupported pairing falls through `bound_param_to_value`'s own post-bind arm: an unrecognized `sql_type` fails `sql_family`, and a recognized one with no matching C type hits the `_ =>` `ConversionNotImplemented`.

I kept the independent draw rather than gating it with `is_supported_conversion`. The only panic-prone code an unsupported pairing reaches — `read_indicator` and `read_param_value` — is driven by `c_type` + indicator + bytes, none of which depend on `sql_type`, so it can't manufacture an impossible read that `SQLBindParameter` would have blocked. Meanwhile the draw keeps exercising the `sql_family`/`_ =>` rejection itself, confirming it returns `Err` instead of panicking. Fuzzing the real bind-time gate would mean driving `SQLBindParameter`, which is a separate target, as you note.

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
saurabh500: Fixed — pointed the link at `is_typed_c_target` (and dropped "predicates" to the singular "predicate"), matching the wording on `ROUTED_FETCH_TARGETS` so `cargo doc` no longer has a dangling link.

> [!NOTE]

mssql-odbc/fuzz/fuzz_targets/fuzz_bound_param.rs
saurabh500: Fixed — the target-file doc now reads "trailing zero padding wide enough for the widest fixed-width C target" instead of pinning eight bytes, so it stays in step with the wrapper's 32-byte padding.

> [!NOTE]

mssql-odbc/src/fuzz_support.rs
```diff
@@ -0,0 +1,587 @@
...
+///
+/// The value buffer is the fuzz input plus trailing zero padding wide enough to
+/// cover the widest fixed-width C target the matrix routes (the 20-byte
+/// `SqlSsTimestampoffsetStruct`); that padding also guarantees a NUL terminator
+/// for the `SQL_NTS` scans, so every in-bounds contract `read_param_value`
+/// documents holds and a crash reflects a real conversion defect, not
+/// harness-induced UB. The indicator spans the valid lengths plus
+/// `read_indicator`'s rejection markers (data-at-execution, default-parameter,
+/// and negative lengths), all of which are refused before the value buffer is
+/// read, so they stay in-bounds by construction.
```
David-Engel: **Nit** — this says the rejection markers are "all of which are refused before the value buffer is read, so they stay in-bounds by construction." That holds for arms 4, 5 and 6, but not for arm 7 (`-6 - (fuzz_len & 0x3f)`).

`read_indicator` only rejects a plain negative length when `indicator_is_a_length(param.c_type)` (`param_buffer.rs:64-68` — `SQL_C_CHAR`/`SQL_C_WCHAR`/`SQL_C_BINARY`/`SQL_C_SS_VECTOR`). For the other 19 entries in `PARAM_C_TYPES` the comment at `param_buffer.rs:124-126` says the leftover value is ignored rather than validated, so it falls through to `Ok(Indicator::Length(ind))` and `read_param_value` does read the buffer. The repo's own `param_buffer::tests::indicator_is_not_a_length_for_fixed_width_c_types` pins exactly that.

No safety problem — the 32-byte padding is what keeps that read in-bounds, which is the better justification anyway, and reaching the read on a negative indicator is extra coverage rather than a defect. It's just that the sentence names a mechanism that doesn't apply to a quarter of the arms. Something like "the negative-length arm is refused only for the character/binary C types; for the fixed-width ones it falls through to a struct read that the padding keeps in-bounds" would keep the claim true.

Same drift class the last two rounds closed, which is the only reason it seems worth a line.

David-Engel (commented):

## Summary

Re-reviewed at head `54392056`; the previous automated passes covered `f42ef1b9` and `4ca532ae`. This head commit is doc-comment-only relative to `4ca532ae`, and it closes all three items from the last round: the `[`routes_fixed_width_fetch`]` intra-doc link now points at `is_typed_c_target` (which is imported in this module, so the link resolves), the `fuzz_bound_param.rs` module doc no longer pins "eight trailing zero bytes" against the 32-byte padding, and both `PARAM_C_TYPES` / `PARAM_SQL_TYPES` docs now correctly attribute an unsupported pairing to `bound_param_to_value`'s post-bind `sql_family` / `_ =>` arms rather than the `SQLBindParameter` bind-time gate. I confirmed that attribution against the code: the only non-test production caller of `is_supported_conversion` is `api/bind_param.rs`, which this harness does not drive.

The rest of the diff is unchanged since the last pass and still reads correctly. `datetimeoffset_parts` in `i128` cannot overflow for any `u64` ticks plus an `i16` offset, the day-range check runs before the `as i64`, and `rem_euclid` keeps the `as u64` in range. `is_typed_c_target` living in `fetch_convert.rs` next to its three component predicates is a behavior-preserving move — `get_data.rs` and `fetch_scroll.rs` import it from the new home and nothing else referenced it. The pipeline wiring is a clean parameterization of `fuzz-template.yml` via `fuzzDir`, and the new `Build ODBC Fuzz Targets` step sits after the existing `Install Fuzz Dependencies` step that provides nightly + `cargo-fuzz`.

I read the earlier Copilot review (including the collapsed suppressed finding) and every prior round on this PR, and I'm not re-filing anything already answered there.

**No blocking issues.** One nit inline.

### What I ran

- `cargo clippy -p mssqlodbc --lib` with `RUSTFLAGS="--cfg fuzzing"` on Windows → clean; every warning in the output came from `mssql-tds`'s pre-existing `fuzz_support` missing-docs, none from `mssql-odbc`. Worth repeating each round because `cargo bclippy` never compiles this module (no `--cfg fuzzing`) and the CI gate is `cargo +nightly fuzz build` / `cargo +nightly check`, not clippy.
- `cargo nextest run -p mssqlodbc --lib -E 'test(datetimeoffset) or test(conversion_matrix) or test(read_indicator) or test(param_buffer)'` → 41 passed, including `datetimeoffset_overflowing_time_is_rejected_not_panicked`. I did not repeat the mechanism-break; it is documented in the self-review and the code under it is unchanged.
- I did not build the `fuzz/` crate; it pulls a dependency tree this review did not audit, and every claim above is decidable by reading. The ADO legs cover it.

All required checks are green at this head, including the ADO validation build (`173541`) and `coverage-report`. The two new fuzz jobs are scheduled-stage jobs and do not report on the PR, so `Fuzz_Odbc_Ffi_Linux` and `Fuzz_Odbc_Additional_Targets_Linux` still have not run on a real agent — worth confirming once before relying on them as a gate.

mssql-odbc/src/fuzz_support.rs
saurabh500: You're right — I traced arm 7 (`-6 - (fuzz_len & 0x3f)`, so −6..−69): it clears the null, default-param, and data-at-exec checks, then `read_indicator` only rejects it via `indicator_is_a_length`, which is the four character/binary/vector types. For the other 19 fixed-width entries it falls through to `Ok(Indicator::Length(ind))` and `read_param_value` does read the buffer — kept in-bounds by the 32-byte padding, not by rejection. Reworded both the function doc and the inline arm comment to say exactly that: arms 4–6 are refused before the read, arm 7's negative length is refused only for the character/binary C types and otherwise falls through to a padded struct read.

> [!NOTE]

mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -441,6 +448,16 @@ pub(crate) fn is_datetime_c_target(target_type: SqlSmallInt) -> bool {
     )
 }
 
+/// Returns `true` if `target_type` is one of the fixed-width C targets
+/// [`convert_typed_c`] routes: the integer, float, GUID, and date/time structs.
```
David-Engel: **Nit** — broken intra-doc link, new in this diff. `convert_typed_c` lives in `api/get_data.rs` and is not imported into `fetch_convert.rs`, so rustdoc can't resolve the path:

```
warning: unresolved link to `convert_typed_c`
   --> mssql-odbc/src/conversion/fetch_convert.rs:452:7
452 | /// [`convert_typed_c`] routes: the integer, float, GUID, and date/time structs.
    |       ^^^^^^^^^^^^^^^ no item named `convert_typed_c` in scope
```

This is the same class as the `routes_fixed_width_fetch` link you fixed last round, just in the opposite direction: at merge base `is_typed_c_target` lived next to `convert_typed_c`, so the name would have resolved — moving the function here left the link dangling. It's the only one of the ten `cargo doc` warnings on this crate that doesn't reproduce on unmodified files.

No CI gate on rustdoc warnings today, so purely cosmetic. Easiest fix is to drop the link and keep the code span:

```suggestion
/// `get_data::convert_typed_c` routes: the integer, float, GUID, and date/time structs.
```

David-Engel (commented):

## Summary

Re-reviewed at head `88cc88e9`; earlier automated passes covered `f42ef1b9`, `4ca532ae`, and `54392056`. This head commit is doc-comment-only relative to `54392056` and closes the one nit from the last round: the `fuzz_bound_param` doc and the inline arm comment no longer claim every out-of-range indicator arm is refused before the value buffer is read. The new wording matches what `read_indicator` actually does — `indicator_is_a_length` only rejects a plain negative length for the character/binary C types, and the fixed-width entries in `PARAM_C_TYPES` fall through to a struct read the 32-byte padding keeps in-bounds. That is the correct attribution.

I re-read the whole diff against merge-base `058aaf46` rather than just the increment. The production change is still the small part and it still reads correctly:

- `datetimeoffset_parts` in `i128` cannot overflow for any `u64` tick count plus an `i16` offset (worst case ≈ 1.8e19 + 2e13, far inside `i128`), the `0..=MAX_DAYS_SINCE_0001` check runs before the `as i64` narrowing, and `rem_euclid(TICKS_PER_DAY)` keeps the `as u64` in range. The regression test is not vacuous: `time_nanoseconds = i64::MAX as u64` with `offset = 60` lands ~10.7M days out, well past `MAX_DAYS_SINCE_0001`, and the old `i64` addition overflows on the same input.
- Moving `is_typed_c_target` from `api/get_data.rs` into `conversion/fetch_convert.rs` is behavior-preserving. I confirmed `is_datetime_c_target` has no other reference left in `get_data.rs` (so the dropped import is correct) and that `fetch_scroll.rs:1732`/`:2115` are the only other call sites.
- `crate-type = ["cdylib", "rlib"]` and the `#[cfg(fuzzing)] pub mod fuzz_support` gate leave the shipped driver's FFI surface unchanged; `unexpected_cfgs` in the workspace `Cargo.toml` already whitelists `cfg(fuzzing)`.
- Pipeline wiring is a clean parameterization: `fuzzDir` defaults to `mssql-tds/fuzz` so the existing tds jobs are byte-identical in behavior, and the two new odbc jobs mirror `Fuzz_TokenStream_Linux` (same pool, same `CARGO_TARGET_DIR`, same template order). The new `Build ODBC Fuzz Targets` step sits after `Install Fuzz Dependencies`, which provides nightly + `cargo-fuzz`.

I read the earlier Copilot review (including the collapsed suppressed finding) and every prior thread on this PR; all thirteen are resolved and I'm not re-filing anything already answered.

**No blocking issues.** One nit inline — new in this PR, same class as the `routes_fixed_width_fetch` link nit from the last round.

### Verification

- `cargo clippy -p mssqlodbc --lib` with `RUSTFLAGS="--cfg fuzzing"` on Windows → clean for `mssql-odbc`; every warning in the output came from `mssql-tds`'s pre-existing `fuzz_support` missing-docs. Worth repeating each round because `cargo bclippy` never compiles this module (no `--cfg fuzzing`) and the CI gate is `cargo +nightly fuzz build` / `cargo +nightly check`, not clippy.
- `cargo nextest run -p mssqlodbc --lib -E 'test(datetimeoffset) or test(is_typed_c_target)'` → 8 passed, including `datetimeoffset_overflowing_time_is_rejected_not_panicked`.
- `cargo doc -p mssqlodbc --no-deps --document-private-items` → 10 warnings, 9 of which reproduce on unmodified files (`exports.rs`, `odbc_authentication_transformer.rs`, `stmt.rs`, `connection/mod.rs`). The tenth is the inline nit below and is introduced by this diff.
- I did not build the `fuzz/` crate; it pulls a dependency tree this review did not audit, and every claim above is decidable by reading. The ADO legs cover it.

All required checks are green at this head, including the ADO validation build (`173558`), `coverage-report`, and CodeQL. The two new fuzz jobs are scheduled-stage jobs and do not report on the PR, so `Fuzz_Odbc_Ffi_Linux` and `Fuzz_Odbc_Additional_Targets_Linux` still have not run on a real agent — worth confirming once before relying on them as a gate.

This is not an approval.

mssql-odbc/src/conversion/fetch_convert.rs
saurabh500: Fixed — dropped the link and kept the code span as `get_data::convert_typed_c`. You're right it's the mirror of last round's `routes_fixed_width_fetch` fix: moving `is_typed_c_target` out of `get_data` left the intra-doc link with no `convert_typed_c` in scope. Confirmed with `cargo doc -p mssqlodbc --document-private-items` that this specific warning is gone; the other unresolved-link warnings are all pre-existing on unmodified files.

> [!NOTE]

David-Engel (commented):

## Summary

Re-reviewed at head `78c298d2`; earlier automated passes covered `f42ef1b9`, `4ca532ae`, `54392056`, and `88cc88e9`. This head is a one-line doc change relative to `88cc88e9` and it closes the last round's nit: the `is_typed_c_target` doc now spells `get_data::convert_typed_c` as a plain code span instead of an intra-doc link, which is the right call because `convert_typed_c` is not imported into `fetch_convert.rs` and rustdoc could not resolve the path. I verified the warning is actually gone rather than moved — see Verification.

I re-read the full diff against merge-base `058aaf46` rather than only the increment, and the production surface still reads correctly:

- `datetimeoffset_parts` in `i128` cannot overflow for any `u64` tick count plus an `i16` offset (worst case ≈ 1.8e19 + 2e13, far inside `i128`), the `0..=MAX_DAYS_SINCE_0001` check runs before the `as i64` narrowing, and `rem_euclid(TICKS_PER_DAY)` keeps the `as u64` in range. The new test is not vacuous — `time_nanoseconds = i64::MAX as u64` with `offset = 60` lands ~10.7M days out — and its `ConvError::Restricted` assertion matches the shape `datetimeoffset_out_of_range_after_offset_is_rejected` already used on `main`, so the fix joins existing behavior rather than inventing a new SQLSTATE mapping.
- `write_fixed` uses `write_unaligned` (`fetch_convert.rs:75`), so `fuzz_fetch_convert`'s plain `[u8; 64]` backing is a valid target, as its doc claims.
- `PARAM_C_TYPES`' 23 entries are exactly the C types `is_supported_conversion` has an arm for (`conversion_matrix.rs:71-99`, counting the `is_integer_c_type` catch-all), and none of them is `SQL_C_DEFAULT`, so the `debug_assert_ne!` at the top of that function cannot fire during the `PARAM_SQL_TYPES` scan.
- Moving `is_typed_c_target` into `fetch_convert.rs` next to its three component predicates is behavior-preserving; `get_data.rs` and `fetch_scroll.rs` import it from the new home and the dropped `is_datetime_c_target` import in `get_data.rs` has no remaining use.
- `crate-type = ["cdylib", "rlib"]` plus the `#[cfg(fuzzing)] pub mod fuzz_support` gate leave the shipped driver's FFI surface unchanged.
- Pipeline wiring is a clean parameterization: `fuzzDir` defaults to `mssql-tds/fuzz`, so the four existing tds fuzz jobs are behaviorally identical, and the two new odbc jobs mirror `Fuzz_TokenStream_Linux` / `Fuzz_Additional_Targets_Linux` (same pool demands, same `CARGO_TARGET_DIR`, same absent `timeoutInMinutes`). The eight odbc target names do not collide with the nine tds ones, so `fuzz-crashes-<target>-$(System.JobName)` stays unique.

I read the original Copilot review (including the collapsed suppressed finding) and every prior thread on this PR; all fifteen are resolved and I am not re-filing anything already answered.

## Verification

- `cargo doc -p mssqlodbc --no-deps --document-private-items` → 9 warnings, down from the 10 at `88cc88e9`. Every remaining one points at a file this diff does not touch (`api/exports.rs` ×4, `connection/odbc_authentication_transformer.rs` ×3, `handles/stmt.rs`, `connection/mod.rs`), so the diff now introduces no rustdoc warning of its own. That is the specific thing this head commit claims to fix.
- `cargo clippy -p mssqlodbc --lib` with `RUSTFLAGS="--cfg fuzzing"` on Windows → zero warnings from `mssql-odbc`; the 94 in the output are `mssql-tds`'s pre-existing `fuzz_support` missing-docs. Worth repeating each round because `cargo bclippy` never compiles this module (no `--cfg fuzzing`) and the CI gate is `cargo +nightly fuzz build` / `cargo +nightly check`, not clippy.
- `cargo nextest run -p mssqlodbc --lib -E 'test(datetimeoffset) or test(is_typed_c_target) or test(conversion_matrix) or test(supported_conversion)'` → 22 passed, including `datetimeoffset_overflowing_time_is_rejected_not_panicked`. I did not repeat the mechanism-break; it is documented in the self-review and the code under it is unchanged since.
- `cargo metadata --manifest-path mssql-odbc/fuzz/Cargo.toml --no-deps` resolves, and the ADO `Build ODBC Fuzz Targets` step (`cargo +nightly fuzz build`) is green on this head, which compiles all eight targets. I did not build the `fuzz/` crate myself; it pulls a dependency tree this review did not audit, and every claim above is decidable by reading.

## Findings

**Blocking** — none.

**Suggestion** — none.

**Nit** — none.

One standing note, unchanged from the last two rounds and not a finding against the diff: `Fuzz_Odbc_Ffi_Linux` and `Fuzz_Odbc_Additional_Targets_Linux` live in the scheduled `Fuzz` stage, so they do not report on the PR and have still never run on a real agent. Worth confirming once before treating them as a gate.

Theekshna (commented):

First-pass review. No prior discussion needed carrying over. Fuzzing harness for `mssql-odbc` and the `datetimeoffset` fetch-panic fix both look sound on their own terms; the one finding below is a msodbcsql parity gap in the *replacement* SQLSTATE, not a correctness regression.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 0 |

**Six required checks**
1. msodbcsql parity: mismatch found — see inline comment 1. A related pre-existing time/datetime2 wrap-vs-reject divergence is already tracked separately (issue #526), not new here.
2. Test sufficiency: mostly adequate — the fuzz crate is compile-checked in PR validation and scheduled in `.pipeline/fuzz-pipeline.yml`; the new overflow regression test is mutation-verified (reverting the `i128` widening in `datetimeoffset_parts` back to `i64` makes `datetimeoffset_overflowing_time_is_rejected_not_panicked` fail with an overflow panic), but it only pins the panic being gone, not the surfaced SQLSTATE.
3. Divergences documented: no — found no `.github/instructions/mssql-odbc.instructions.md` entry recording the `07006` vs. reference `22008` behavior as a deliberate choice.
4. PR description currency: mostly current — diff, AB#45636, and green checks line up; minor drift is that the description frames this as a normal out-of-range conversion while the code/test currently surface `07006` specifically.
5. Verbose slop: none found.
6. Audit of PR's own evidence: re-derived the overflow boundary — `i64::MAX` ticks plus a 60-minute offset is ~10,675,199 days, well past `MAX_DAYS_SINCE_0001 = 3,652,058`, consistent with the panic this PR fixes; the neighboring underflow case is already covered by an existing test.

mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -2154,6 +2171,37 @@ mod tests {
...
+                    },
+                    offset: 60,
+                }),
+                SQL_C_SS_TIMESTAMPOFFSET,
+                (&mut out as *mut SqlSsTimestampoffsetStruct).cast(),
+                &mut ind,
+            )
+        }
+        .unwrap_err();
+        assert_eq!(err, ConvError::Restricted);
```
Theekshna, line 2202: [Suggestion] This regression test locks the datetimeoffset-overflow-on-fetch case to `ConvError::Restricted`, which `finish_typed_conv` maps to SQLSTATE `07006` (see `api/get_data.rs` and `api/sqlstate.rs`). The panic is correctly fixed, but the replacement SQLSTATE still diverges from the reference driver: msodbcsql's fetch path validates the decoded `datetimeoffset` struct and reports a 22008-class temporal-overflow error instead of a restricted-conversion error (msodbcsql `sqlccnvt.cpp` temporal validation, plus `cli_common/clntcomn.cpp` SQLSTATE mapping). Worth confirming whether `22008` is the intended surfaced state here, and if `07006` is deliberate, recording it as a documented divergence in `.github/instructions/mssql-odbc.instructions.md`.

saurabh500: You're right that `07006` diverges from msodbcsql here — I confirmed against the reference source that its temporal path (`sqlccnvt.cpp`, incl. `SQL_SS_TIMESTAMPOFFSET`) surfaces `22008` for temporal overflow. Two things, though: the mapping is pre-existing, not introduced here — it comes from `extract_datetime_parts(value).ok_or(ConvError::Restricted)?` at `fetch_convert.rs:481`, and the sibling test `datetimeoffset_out_of_range_after_offset_is_rejected` already on `main` asserts the same `Restricted`. This PR's regression test only pins that existing behavior while proving the former i64-overflow panic is gone.

Remapping to `22008` is cross-cutting: that one `ok_or(Restricted)` covers every datetime C target and both value- and character-derived inputs, so it needs a distinct out-of-range temporal `ConvError` variant and per-target verification against msodbcsql — out of scope for a fuzzing + DoS-fix PR. Filed #529 to track it (with the msodbcsql evidence and the documented-divergence alternative you suggested); keeping the test as-is here since it correctly asserts current behavior.

> [!NOTE]

David-Engel (commented):

## Summary

Re-reviewed at head `636d30c8`. The only change since the last clean automated pass at `78c298d2` is the merge of `origin/main` (`63cb4b4e`, PR #523 "Deliver SQL_C_BINARY results for binary, UDT and LOB fetch"), which lands in the same two files this PR edits — `api/get_data.rs` and `api/fetch_scroll.rs`. So this pass focused on the merge resolution, plus a fresh mechanism-break check on the regression test. No findings.

## What I checked

**Merge resolution.** The `is_typed_c_target` move survived the merge intact and there is exactly one definition, in `conversion/fetch_convert.rs:454`. All four production call sites resolve to it — `get_data.rs:422`, `get_data.rs:1128`, `fetch_scroll.rs:1730`, `fetch_scroll.rs:2126` — and the import lists in both files were updated in step (`is_datetime_c_target` dropped from `get_data`'s list because it is now only reached through the moved predicate). Main's new `SQL_C_BINARY` delivery path does not reference the old `get_data::is_typed_c_target`, so nothing was left dangling. `cargo clippy -p mssqlodbc --all-targets` is clean on the merged tree.

**`cfg(fuzzing)` code after the merge.** Worth calling out because ordinary clippy does not compile it, so a merge could break `fuzz_support.rs` without any local or PR-gated check noticing — the odbc fuzz compile check in `containerized-build.sh` runs only on the Linux container leg. I ran `RUSTFLAGS="--cfg fuzzing" cargo check -p mssqlodbc --lib` against the merged tree: clean, zero warnings from `mssqlodbc` (the 93 missing-doc warnings are pre-existing `mssql-tds/src/fuzz_support.rs` ones, unchanged by this PR). `mssql_tds::fuzz_support::{FuzzPacketReader, create_fuzz_tds_client}` are gated on `cfg(fuzzing)` only, not on the `test-util` feature, so the odbc fuzz crate's plain `mssqlodbc = { path = ".." }` dependency is sufficient.

**The `datetimeoffset` fix and its test.** Mechanism-break verified rather than taken on faith: reverting `datetimeoffset_parts` to the `i64` arithmetic makes `datetimeoffset_overflowing_time_is_rejected_not_panicked` fail with `attempt to add with overflow` at `fetch_convert.rs:366`, while the other seven `datetimeoffset` tests stay green. Restored, and all eight pass on the merged tree. The `i128` widening is value-identical for in-range inputs: `div_euclid`/`rem_euclid` over the same `TICKS_PER_DAY`, and the `days as i64` cast is downstream of the `0..=MAX_DAYS_SINCE_0001` guard, so it can't truncate.

**Pipeline wiring.** `Fuzz_Odbc_Ffi_Linux` and `Fuzz_Odbc_Additional_Targets_Linux` mirror the existing `Fuzz_TokenStream_Linux` / `Fuzz_Additional_Targets_Linux` shape exactly — same pool, same `imageOverride`, same `CARGO_TARGET_DIR`, same repeated-template-per-target structure — and sit in the `Fuzz` stage gated on `RunFuzz == 'true'`, so they don't add per-PR time. The `fuzzDir` parameter defaults to `mssql-tds/fuzz`, leaving the three existing tds jobs unchanged.

## Adjudicating the open thread

@Theekshna's SQLSTATE finding on `fetch_convert.rs:2202` (that `ConvError::Restricted` → `07006` diverges from msodbcsql's `22008` for temporal overflow) — the divergence is real, but the deferral is right. `datetimeoffset_out_of_range_after_offset_is_rejected` already asserts `ConvError::Restricted` for the same class on `main` at `63cb4b4e`, so the mapping is pre-existing and this PR only routes one more input into it. Fixing it here would leave the sibling temporal paths inconsistent; it belongs in its own change, alongside #526.

## Findings

**Blocking** — none.

**Suggestion** — none.

**Nit** — none.

Earlier rounds' findings (the `cfg(fuzzing)` `unwrap` sites, the hand-copied target lists on both the fetch and param sides, the buffer-padding and indicator-arm doc drift, the two broken intra-doc links) are all addressed in the current head and I re-read each fix rather than trusting the reply.

Theekshna (commented):

Re-review triggered by a head-SHA change since the last pass. Verified with `git patch-id` on both sides that the PR's own diff against its merge-base is unchanged in substance — the only first-parent commit since the last review is `Merge remote-tracking branch 'origin/main'`, which pulled in an unrelated main PR (#523, binary/UDT/LOB fetch) touching the same import blocks in `fetch_scroll.rs`/`get_data.rs`. No new author commits, no re-review of already-settled content.

The one previously-filed finding (SQLSTATE `07006` vs. msodbcsql's `22008` on datetimeoffset overflow) was answered in full: pre-existing behavior, not introduced here, tracked in #529, out of scope for a fuzzing+DoS-fix PR. Not re-filing it.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 0 |

**Six required checks**
1. msodbcsql parity: no new surface since last review; prior finding resolved (see above).
2. Test sufficiency: unchanged from last review — `datetimeoffset_overflowing_time_is_rejected_not_panicked` still mutation-verified (reverting `i128`→`i64` in `datetimeoffset_parts` still panics). `gh pr checks 511` is all green, including the odbc fuzz compile check and `mssqlodbc` suite.
3. Divergences documented: the `07006`/`22008` gap is tracked via issue #529 rather than an instructions.md entry — an acceptable alternative per the author's own reasoning (cross-cutting, needs a new `ConvError` variant and per-target verification).
4. PR description currency: see Suggestion below — the description names #526 as the deferred follow-up but not #529, both filed during review of this same PR.
5. Verbose slop: none found; no new prose since last pass.
6. Audit of PR's own evidence: unchanged from last review (i64::MAX-tick overflow boundary re-derivation still holds; no new evidence to audit this pass).

**Suggestion**
- PR description ("Related Issues" section): lists issue #526 (out-of-range time truncation) as the review-time follow-up but omits #529 (SQLSTATE `07006` vs. `22008` divergence), filed in the same review round on this PR. A reader of the description would miss that a second, related gap was deliberately deferred rather than fixed. Worth a one-line addition alongside the existing #526 reference.

saurabh500:
Addressed the description-currency suggestion: the "Related Issues" section now lists **#529** (the `07006` vs. `22008` datetimeoffset-overflow SQLSTATE divergence) alongside the existing #526, so both review-time deferrals are on the record.

Theekshna (commented):

Re-review triggered by both prior review threads on this PR now being resolved (marker criterion c) — no new commits since the last review (head SHA unchanged). Verified both outstanding items are actually closed out, not just marked resolved: the `07006`/`22008` SQLSTATE divergence thread has saurabh500's reply confirming it's pre-existing, deferred to #529 with msodbcsql evidence cited; the PR description now lists both #526 and #529 under "Related Issues" (confirmed via `gh pr view`), closing the description-currency suggestion from the prior pass. `gh pr checks 511` is fully green across all jobs, including the Kerberos, cross-repo `mssql-python`, and coverage-merge stages. Nothing new to review since the code is byte-identical to the last pass.

David-Engel (approved):

PR #515, opened 2026-09-07
Add ESRP crate publishing pipeline

### LINKED ISSUE #514: Publish Rust crate artifacts through ESRP
### Problem statement

The official and non-official OneBranch pipelines do not produce `.crate` artifacts for the publishable Rust packages, and the official release pipeline cannot publish those packages to crates.io through ESRP.

### Proposed solution

Package `mssql-tds` and `mssql-mock-tds` in both build pipelines, retain them as a dedicated pipeline artifact with release metadata and hashes, and add opt-in `EsrpRelease@12` steps to the official release pipeline. Publish `mssql-tds` first, wait for it to become available on crates.io, and then publish `mssql-mock-tds`.

### Affected crate

Not applicable / Multiple

### Alternatives considered

Publishing directly with `cargo publish` would require a crates.io token and bypass the required ESRP release workflow. Submitting both crates to ESRP together would not preserve dependency ordering.

### Additional context

Internal guidance: https://eng.ms/docs/microsoft-security/identity/trust-and-security-services/tss-release-distribute/tss-release-esrp-parent/oss-publishing/releasing-open-source/cratesio

saurabh500:
## Description

- Package `mssql-tds` and `mssql-mock-tds` in both official and non-official OneBranch builds.
- Publish a dedicated Rust crate artifact with ordered release metadata and SHA-256 hashes.
- Add opt-in `EsrpRelease@12` publishing for each crate, with preflight checks, dependency ordering, crates.io availability polling, and independent switches for partial-run recovery.
- Add a `validateCratesOnly` mode that runs artifact validation and selected-crate preflights without invoking ESRP or publishing immutable crate versions; the existing `mssql-py-core` tag flow remains independent.
- Add the required registry version to the local `mssql-tds` dependency so `mssql-mock-tds` can be packaged for crates.io, then compile the packaged mock against the packaged core before emitting the artifact.
- Add source-repository and license-file metadata to both crates and exclude `.env*`, integration tests, and benchmarks from the immutable `mssql-tds` package.

## Related Issues

Fixes #514

## Validation

- `cargo bfmt`
- `cargo bclippy`
- `cargo nextest run --frozen --package mssql-tds --package mssql-mock-tds --lib --no-fail-fast --profile ci` (2,131 passed)
- Official and non-official `.crate` creation, release-manifest validation, normalized dependency metadata, and crates.io preflight behavior exercised locally
- [Current-head OneBranch run 20260909.8](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=173691) succeeded against `1d66060b`: `mssql-tds` passed package verification, the exact packaged mock passed `cargo check` against the exact packaged core, and the emitted artifact retained valid hashes and normalized dependency metadata
- Both final `.crate` files contain repository metadata and root MIT license text; `.env*`, `tests/**`, and `benches/**` are absent
- Focused regression checks cover sandbox dependency upserts, transient crates.io responses, packaged dependency validation, and count-limited version stamping
- Full `cargo btest` was attempted but requires the live SQL Server integration environment unavailable on this machine

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes

### DIFF
--- a/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
+++ b/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
@@ -45,7 +45,7 @@ parameters:
     default: false
 
   - name: tagRelease
-    displayName: 'Create git tag and release branch'
+    displayName: 'Create mssql-py-core git tag and release branch'
     type: boolean
     default: false  # Safety: default to false
 
@@ -327,17 +327,12 @@ extends:
               displayName: 'Verify mssql-mock-tds on crates.io'
 
     ###########################################################################
-    # Stage 3: Tag the source commit and create release branch
+    # Stage 3: Tag the mssql-py-core release source
     ###########################################################################
-    - ${{ if and(eq(parameters.tagRelease, true), ne(parameters.validateCratesOnly, true)) }}:
+    - ${{ if eq(parameters.tagRelease, true) }}:
       - stage: Tag
-        displayName: 'Tag Release and Create Branch'
-        ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
-          dependsOn:
-          - Release
-          - ReleaseCrates
-        ${{ else }}:
-          dependsOn: Release
+        displayName: 'Tag mssql-py-core Release'
+        dependsOn: Release
         jobs:
         - job: TagAndBranch
           displayName: 'Git tag + release branch'
--- a/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
+++ b/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
@@ -329,7 +329,7 @@ extends:
     ###########################################################################
     # Stage 3: Tag the source commit and create release branch
     ###########################################################################
-    - ${{ if eq(parameters.tagRelease, true) }}:
+    - ${{ if and(eq(parameters.tagRelease, true), ne(parameters.validateCratesOnly, true)) }}:
       - stage: Tag
         displayName: 'Tag Release and Create Branch'
         ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
--- a/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
+++ b/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
@@ -230,12 +230,12 @@ extends:
             value: '$(Build.ArtifactStagingDirectory)'
           steps:
           - download: officialBuild
-            artifact: 'drop_Build_RustCrates_rust_crates'
+            artifact: 'drop_Build_RustCrates'
             displayName: 'Download Rust crate artifacts'
 
           - pwsh: >
               .pipeline/scripts/validate-crate-release-artifact.ps1
-              -ArtifactDirectory '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates_rust_crates'
+              -ArtifactDirectory '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates'
             displayName: 'Validate Rust crate release artifact'
 
           - ${{ if eq(parameters.publishMssqlTds, true) }}:
@@ -268,7 +268,7 @@ extends:
                 intent: 'PackageDistribution'
                 contenttype: 'Rust'
                 contentsource: 'Folder'
-                folderlocation: '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates_rust_crates\mssql-tds'
+                folderlocation: '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates\mssql-tds'
                 waitforreleasecompletion: true
                 owners: '$(owner)'
                 approvers: '$(approver)'
@@ -304,7 +304,7 @@ extends:
                 intent: 'PackageDistribution'
                 contenttype: 'Rust'
                 contentsource: 'Folder'
-                folderlocation: '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates_rust_crates\mssql-mock-tds'
+                folderlocation: '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates\mssql-mock-tds'
                 waitforreleasecompletion: true
                 owners: '$(owner)'
                 approvers: '$(approver)'
--- a/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
+++ b/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
@@ -39,6 +39,11 @@ parameters:
     type: boolean
     default: false  # Safety: default to false to prevent accidental releases
 
+  - name: validateCratesOnly
+    displayName: 'Validate Rust crate artifacts without publishing'
+    type: boolean
+    default: false
+
   - name: tagRelease
     displayName: 'Create git tag and release branch'
     type: boolean
@@ -212,7 +217,7 @@ extends:
     ###########################################################################
     # Stage 2: Publish Rust crates through ESRP
     ###########################################################################
-    - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
+    - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true), eq(parameters.validateCratesOnly, true)) }}:
       - stage: ReleaseCrates
         displayName: 'Release Rust Crates'
         dependsOn: Release
@@ -229,6 +234,8 @@ extends:
           - name: ob_outputDirectory
             value: '$(Build.ArtifactStagingDirectory)'
           steps:
+          - checkout: self
+
           - download: officialBuild
             artifact: 'drop_Build_RustCrates'
             displayName: 'Download Rust crate artifacts'
@@ -238,7 +245,7 @@ extends:
               -ArtifactDirectory '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates'
             displayName: 'Validate Rust crate release artifact'
 
-          - ${{ if eq(parameters.publishMssqlTds, true) }}:
+          - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.validateCratesOnly, true)) }}:
             - pwsh: >
                 .pipeline/scripts/test-cratesio-package.ps1
                 -CrateName 'mssql-tds'
@@ -247,7 +254,7 @@ extends:
                 -MaxAttempts 1
               displayName: 'Preflight mssql-tds version'
 
-          - ${{ if eq(parameters.publishMssqlMockTds, true) }}:
+          - ${{ if or(eq(parameters.publishMssqlMockTds, true), eq(parameters.validateCratesOnly, true)) }}:
             - pwsh: >
                 .pipeline/scripts/test-cratesio-package.ps1
                 -CrateName 'mssql-mock-tds'
--- a/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
+++ b/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
@@ -1,8 +1,8 @@
 #################################################################################
 #                        OneBranch Pipelines - Official Release                 #
-# Downloads NuGet wheel artifacts from the Official Build pipeline and         #
-# publishes them to Azure Artifacts. Also tags the source commit and           #
-# creates a release branch.                                                     #
+# Downloads artifacts from the Official Build pipeline, publishes the wheel    #
+# NuGet package to Azure Artifacts, and can publish Rust crates to crates.io.   #
+# It can also tag the source commit and create a release branch.                #
 #                                                                               #
 # This pipeline does NOT build anything — it consumes artifacts produced by    #
 # OfficialBuild.yml.                                                            #
@@ -26,6 +26,16 @@ pr: none
 parameters:
   - name: publishNuGet
     displayName: 'Publish NuGet package to Azure Artifacts'
+    type: boolean
+    default: false  # Safety: default to false to prevent accidental releases
+
+  - name: publishMssqlTds
+    displayName: 'Publish mssql-tds to crates.io'
+    type: boolean
+    default: false  # Safety: default to false to prevent accidental releases
+
+  - name: publishMssqlMockTds
+    displayName: 'Publish mssql-mock-tds to crates.io'
     type: boolean
     default: false  # Safety: default to false to prevent accidental releases
 
@@ -200,12 +210,127 @@ extends:
           displayName: 'Verify NuGet package'
 
     ###########################################################################
-    # Stage 2: Tag the source commit and create release branch
+    # Stage 2: Publish Rust crates through ESRP
+    ###########################################################################
+    - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
+      - stage: ReleaseCrates
+        displayName: 'Release Rust Crates'
+        dependsOn: Release
+        jobs:
+        - job: PublishRustCrates
+          displayName: 'Publish Rust crates to crates.io'
+          templateContext:
+            type: releaseJob
+            isProduction: true
+          pool:
+            type: windows
+          variables:
+          - group: 'ESRP Federated Creds (AME)'
+          - name: ob_outputDirectory
+            value: '$(Build.ArtifactStagingDirectory)'
+          steps:
+          - download: officialBuild
+            artifact: 'drop_Build_RustCrates_rust_crates'
+            displayName: 'Download Rust crate artifacts'
+
+          - pwsh: >
+              .pipeline/scripts/validate-crate-release-artifact.ps1
+              -ArtifactDirectory '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates_rust_crates'
+            displayName: 'Validate Rust crate release artifact'
+
+          - ${{ if eq(parameters.publishMssqlTds, true) }}:
+            - pwsh: >
+                .pipeline/scripts/test-cratesio-package.ps1
+                -CrateName 'mssql-tds'
+                -Version '$(mssqlTdsCrateVersion)'
+                -ExpectedState Absent
+                -MaxAttempts 1
+              displayName: 'Preflight mssql-tds version'
+
+          - ${{ if eq(parameters.publishMssqlMockTds, true) }}:
+            - pwsh: >
+                .pipeline/scripts/test-cratesio-package.ps1
+                -CrateName 'mssql-mock-tds'
+                -Version '$(mssqlMockTdsCrateVersion)'
+                -ExpectedState Absent
+                -MaxAttempts 1
+              displayName: 'Preflight mssql-mock-tds version'
+
+          - ${{ if eq(parameters.publishMssqlTds, true) }}:
+            - task: EsrpRelease@12
+              displayName: 'Publish mssql-tds to crates.io'
+              inputs:
+                connectedservicename: 'ESRP Managed Identity federated auth - AME tenant-mssql-rs'
+                usemanagedidentity: true
+                keyvaultname: '$(AuthAKVName)'
+                signcertname: '$(AuthSignCertName)'
+                clientid: '$(EsrpClientId)'
+                intent: 'PackageDistribution'
+                contenttype: 'Rust'
+                contentsource: 'Folder'
+                folderlocation: '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates_rust_crates\mssql-tds'
+                waitforreleasecompletion: true
+                owners: '$(owner)'
+                approvers: '$(approver)'
+                serviceendpointurl: 'https://api.esrp.microsoft.com'
+                mainpublisher: 'ESRPRELPACMAN'
+                domaintenantid: '975f013f-7f24-47e8-a7d3-abc4752bf346'
+
+            - pwsh: >
+                .pipeline/scripts/test-cratesio-package.ps1
+                -CrateName 'mssql-tds'
+                -Version '$(mssqlTdsCrateVersion)'
+                -ExpectedState Available
+              displayName: 'Wait for mssql-tds on crates.io'
+
+          - ${{ if and(eq(parameters.publishMssqlMockTds, true), ne(parameters.publishMssqlTds, true)) }}:
+            - pwsh: >
+                .pipeline/scripts/test-cratesio-package.ps1
+                -CrateName 'mssql-tds'
+                -Version '$(mssqlTdsCrateVersion)'
+                -ExpectedState Available
+                -MaxAttempts 1
+              displayName: 'Verify mssql-tds dependency is published'
+
+          - ${{ if eq(parameters.publishMssqlMockTds, true) }}:
+            - task: EsrpRelease@12
+              displayName: 'Publish mssql-mock-tds to crates.io'
+              inputs:
+                connectedservicename: 'ESRP Managed Identity federated auth - AME tenant-mssql-rs'
+                usemanagedidentity: true
+                keyvaultname: '$(AuthAKVName)'
+                signcertname: '$(AuthSignCertName)'
+                clientid: '$(EsrpClientId)'
+                intent: 'PackageDistribution'
+                contenttype: 'Rust'
+                contentsource: 'Folder'
+                folderlocation: '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates_rust_crates\mssql-mock-tds'
+                waitforreleasecompletion: true
+                owners: '$(owner)'
+                approvers: '$(approver)'
+                serviceendpointurl: 'https://api.esrp.microsoft.com'
+                mainpublisher: 'ESRPRELPACMAN'
+                domaintenantid: '975f013f-7f24-47e8-a7d3-abc4752bf346'
+
+            - pwsh: >
+                .pipeline/scripts/test-cratesio-package.ps1
+                -CrateName 'mssql-mock-tds'
+                -Version '$(mssqlMockTdsCrateVersion)'
+                -ExpectedState Available
+              displayName: 'Verify mssql-mock-tds on crates.io'
+
+    ###########################################################################
+    # Stage 3: Tag the source commit and create release branch
     ###########################################################################
     - ${{ if eq(parameters.tagRelease, true) }}:
       - stage: Tag
         displayName: 'Tag Release and Create Branch'
-        dependsOn: Release
+        ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
+          dependsOn:
+          - Release
+          - ReleaseCrates
+        ${{ else }}:
+          dependsOn: Release
         jobs:
         - job: TagAndBranch
           displayName: 'Git tag + release branch'
--- a/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
+++ b/.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
@@ -40,7 +40,7 @@ parameters:
     default: false  # Safety: default to false to prevent accidental releases
 
   - name: validateCratesOnly
-    displayName: 'Validate Rust crate artifacts without publishing'
+    displayName: 'Validate selected Rust crates without publishing'
     type: boolean
     default: false
 
@@ -245,7 +245,7 @@ extends:
               -ArtifactDirectory '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates'
             displayName: 'Validate Rust crate release artifact'
 
-          - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.validateCratesOnly, true)) }}:
+          - ${{ if eq(parameters.publishMssqlTds, true) }}:
             - pwsh: >
                 .pipeline/scripts/test-cratesio-package.ps1
                 -CrateName 'mssql-tds'
@@ -254,7 +254,7 @@ extends:
                 -MaxAttempts 1
               displayName: 'Preflight mssql-tds version'
 
-          - ${{ if or(eq(parameters.publishMssqlMockTds, true), eq(parameters.validateCratesOnly, true)) }}:
+          - ${{ if eq(parameters.publishMssqlMockTds, true) }}:
             - pwsh: >
                 .pipeline/scripts/test-cratesio-package.ps1
                 -CrateName 'mssql-mock-tds'
@@ -263,7 +263,7 @@ extends:
                 -MaxAttempts 1
               displayName: 'Preflight mssql-mock-tds version'
 
-          - ${{ if eq(parameters.publishMssqlTds, true) }}:
+          - ${{ if and(eq(parameters.publishMssqlTds, true), ne(parameters.validateCratesOnly, true)) }}:
             - task: EsrpRelease@12
               displayName: 'Publish mssql-tds to crates.io'
               inputs:
@@ -299,7 +299,7 @@ extends:
                 -MaxAttempts 1
               displayName: 'Verify mssql-tds dependency is published'
 
-          - ${{ if eq(parameters.publishMssqlMockTds, true) }}:
+          - ${{ if and(eq(parameters.publishMssqlMockTds, true), ne(parameters.validateCratesOnly, true)) }}:
             - task: EsrpRelease@12
               displayName: 'Publish mssql-mock-tds to crates.io'
               inputs:
--- a/.pipeline/scripts/package-rust-crates.ps1
+++ b/.pipeline/scripts/package-rust-crates.ps1
@@ -32,16 +32,17 @@ $manifestEntries = @()
 Push-Location $RepositoryRoot
 try {
     foreach ($crateName in $crateNames) {
-        # The dependent crate cannot be verified until mssql-tds reaches crates.io.
-        # Workspace validation runs separately, so package without a generated lockfile.
         $arguments = @(
             'package'
             '--package', $crateName
-            '--no-verify'
             '--allow-dirty'
             '--exclude-lockfile'
             '--target-dir', $TargetDirectory
         )
+        if ($crateName -eq 'mssql-mock-tds') {
+            # Verification requires its mssql-tds version to exist on crates.io.
+            $arguments += '--no-verify'
+        }
 
         & cargo @arguments
         if ($LASTEXITCODE -ne 0) {
--- a/.pipeline/scripts/package-rust-crates.ps1
+++ b/.pipeline/scripts/package-rust-crates.ps1
@@ -1,0 +1,84 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+<#
+.SYNOPSIS
+  Creates the Rust crate artifacts consumed by the official release pipeline.
+#>
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory = $true)][string]$OutputDirectory,
+    [Parameter(Mandatory = $true)][string]$Version,
+    [string]$TargetDirectory = (Join-Path ([System.IO.Path]::GetTempPath()) 'mssql-rs-crate-target'),
+    [string]$RepositoryRoot = (Resolve-Path (Join-Path $PSScriptRoot '../..'))
+)
+
+$ErrorActionPreference = 'Stop'
+$crateNames = @('mssql-tds', 'mssql-mock-tds')
+
+if (Test-Path $OutputDirectory) {
+    $existing = @(Get-ChildItem $OutputDirectory -Force)
+    if ($existing.Count -ne 0) {
+        throw "Crate output directory is not empty: $OutputDirectory"
+    }
+}
+else {
+    New-Item -ItemType Directory -Path $OutputDirectory -Force | Out-Null
+}
+
+New-Item -ItemType Directory -Path $TargetDirectory -Force | Out-Null
+$manifestEntries = @()
+
+Push-Location $RepositoryRoot
+try {
+    foreach ($crateName in $crateNames) {
+        # The dependent crate cannot be verified until mssql-tds reaches crates.io.
+        # Workspace validation runs separately, so package without a generated lockfile.
+        $arguments = @(
+            'package'
+            '--package', $crateName
+            '--no-verify'
+            '--allow-dirty'
+            '--exclude-lockfile'
+            '--target-dir', $TargetDirectory
+        )
+
+        & cargo @arguments
+        if ($LASTEXITCODE -ne 0) {
+            throw "cargo package failed for $crateName with exit code $LASTEXITCODE"
+        }
+
+        $fileName = "$crateName-$Version.crate"
+        $source = Join-Path $TargetDirectory "package/$fileName"
+        if (-not (Test-Path $source -PathType Leaf)) {
+            throw "cargo package did not create the expected file: $source"
+        }
+
+        $crateDirectory = Join-Path $OutputDirectory $crateName
+        New-Item -ItemType Directory -Path $crateDirectory -Force | Out-Null
+        $destination = Join-Path $crateDirectory $fileName
+        Copy-Item $source $destination
+
+        $manifestEntries += [ordered]@{
+            name = $crateName
+            version = $Version
+            file = "$crateName/$fileName"
+            sha256 = (Get-FileHash $destination -Algorithm SHA256).Hash.ToLowerInvariant()
+        }
+    }
+}
+finally {
+    Pop-Location
+}
+
+$releaseManifest = [ordered]@{
+    schemaVersion = 1
+    crates = $manifestEntries
+}
+$manifestPath = Join-Path $OutputDirectory 'release-manifest.json'
+$releaseManifest | ConvertTo-Json -Depth 4 | Set-Content $manifestPath
+
+Write-Host 'Created Rust crate artifacts:'
+Get-ChildItem $OutputDirectory -Recurse -File | ForEach-Object {
+    Write-Host "  $($_.FullName)"
+}
--- a/.pipeline/scripts/package-rust-crates.ps1
+++ b/.pipeline/scripts/package-rust-crates.ps1
@@ -67,6 +67,49 @@ try {
             sha256 = (Get-FileHash $destination -Algorithm SHA256).Hash.ToLowerInvariant()
         }
     }
+
+    $verificationDirectory = Join-Path $TargetDirectory 'package-verification'
+    New-Item -ItemType Directory -Path $verificationDirectory -Force | Out-Null
+    foreach ($crateName in $crateNames) {
+        $cratePath = Join-Path $TargetDirectory "package/$crateName-$Version.crate"
+        & tar -xzf $cratePath -C $verificationDirectory
+        if ($LASTEXITCODE -ne 0) {
+            throw "Could not extract $cratePath for verification"
+        }
+    }
+
+    $tdsPackageDirectory = Join-Path $verificationDirectory "mssql-tds-$Version"
+    $mockPackageDirectory = Join-Path $verificationDirectory "mssql-mock-tds-$Version"
+    $mockManifestPath = Join-Path $mockPackageDirectory 'Cargo.toml'
+    $mockManifest = Get-Content $mockManifestPath -Raw
+    $dependency = [regex]::Match(
+        $mockManifest,
+        '(?ms)^\[dependencies\.mssql-tds\]\s*$(?<body>.*?)(?=^\[|\z)'
+    )
+    if (-not $dependency.Success) {
+        throw 'Packaged mssql-mock-tds has no mssql-tds dependency'
+    }
+    if ($dependency.Groups['body'].Value -match '(?m)^\s*path\s*=') {
+        throw 'Packaged mssql-mock-tds unexpectedly contains an mssql-tds path dependency'
+    }
+
+    $relativeTdsPath = [System.IO.Path]::GetRelativePath(
+        $mockPackageDirectory,
+        $tdsPackageDirectory
+    ).Replace('\', '/')
+    $patchedDependency = $dependency.Value.TrimEnd() + "`npath = `"$relativeTdsPath`"`n`n"
+    $mockManifest = $mockManifest.Substring(0, $dependency.Index) +
+        $patchedDependency +
+        $mockManifest.Substring($dependency.Index + $dependency.Length)
+    Set-Content $mockManifestPath $mockManifest -NoNewline
+
+    & cargo check `
+        --manifest-path $mockManifestPath `
+        --target-dir (Join-Path $TargetDirectory 'package-verification-target')
+    if ($LASTEXITCODE -ne 0) {
+        throw "Packaged mssql-mock-tds failed cargo check with exit code $LASTEXITCODE"
+    }
+    Write-Host 'Verified packaged mssql-mock-tds against packaged mssql-tds.'
 }
 finally {
     Pop-Location
--- a/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
+++ b/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
@@ -84,9 +84,9 @@ foreach ($entry in @(
     }
     else {
         $replacement = "{ $key = `"$value`", ".Replace('$', '$$')
-        $dependency = [regex]::Replace(
+        $openBracePattern = [regex]'\{\s*'
+        $dependency = $openBracePattern.Replace(
             $dependency,
-            '\{\s*',
             $replacement,
             1
         )
--- a/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
+++ b/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
@@ -6,10 +6,8 @@
   SANDBOX / TEST-ONLY helper. Pins the mssql-tds path dependency to a version.
 
 .DESCRIPTION
-  The mssql-tds path dependency in mssql-mock-tds/Cargo.toml has no version, which
-  cargo publish rejects ("all dependencies must have a version"). This adds the
-  stamped version while keeping the path (cargo uses the version on publish and the
-  path for local verify builds).
+  Updates the mssql-tds dependency to the stamped version while keeping the path
+  (cargo uses the version on publish and the path for local verify builds).
 
   A version alone is not enough: cargo records a dependency with no registry as
   coming from crates.io, where mssql-tds does not exist. Consumers of the feed then
@@ -58,21 +56,46 @@ $index = if ($IndexUrl -match '^sparse\+') { $IndexUrl } else { "sparse+$IndexUr
 
 $path = 'mssql-mock-tds/Cargo.toml'
 $c = Get-Content $path -Raw
-$c = $c -replace 'mssql-tds\s*=\s*\{\s*path\s*=\s*"\.\./mssql-tds"',
-                 "mssql-tds = { path = `"../mssql-tds`", version = `"$Version`", registry-index = `"$index`""
 
-# -replace changes nothing when the dependency line drifts from the shape above
-# (keys reordered, workspace = true), so without this the script would exit 0 having
-# done nothing and the run would fail much later, in cargo publish -p mssql-mock-tds
-# -- after cargo-publish-mock.ps1 has already put mssql-tds@$Version on the feed.
-# Feed versions are immutable, so that burns the version and the retry cannot
-# succeed. Counting rather than merely testing for presence also catches a second
-# run appending a duplicate set of keys, which TOML rejects.
-$dep = [regex]::Match($c, '(?m)^\s*mssql-tds\s*=\s*\{[^}]*\}')
-if (-not $dep.Success) {
-    Write-Error "No mssql-tds dependency line found in $path"
+$dependencyPattern = [regex]'(?m)^\s*mssql-tds\s*=\s*\{[^}\r\n]*\}\s*$'
+$dependencyMatches = $dependencyPattern.Matches($c)
+if ($dependencyMatches.Count -ne 1) {
+    Write-Error "Expected one mssql-tds dependency in ${path}, found $($dependencyMatches.Count)"
     exit 1
 }
+
+$dependency = $dependencyMatches[0].Value
+foreach ($entry in @(
+    @{ Key = 'version'; Value = $Version },
+    @{ Key = 'registry-index'; Value = $index }
+)) {
+    $key = $entry.Key
+    $value = $entry.Value
+    $keyPattern = [regex]"(?<![\w-])$key\s*=\s*`"[^`"]*`""
+    $keyMatches = $keyPattern.Matches($dependency)
+    if ($keyMatches.Count -gt 1) {
+        Write-Error "Expected at most one '$key' in the mssql-tds dependency of ${path}, found $($keyMatches.Count)"
+        exit 1
+    }
+
+    if ($keyMatches.Count -eq 1) {
+        $dependency = $keyPattern.Replace($dependency, "$key = `"$value`"", 1)
+    }
+    else {
+        $dependency = [regex]::Replace(
+            $dependency,
+            '\{\s*',
+            "{ $key = `"$value`", ",
+            1
+        )
+    }
+}
+
+$c = $c.Substring(0, $dependencyMatches[0].Index) +
+    $dependency +
+    $c.Substring($dependencyMatches[0].Index + $dependencyMatches[0].Length)
+
+$dep = $dependencyPattern.Match($c)
 foreach ($key in 'version', 'registry-index') {
     $n = ([regex]::Matches($dep.Value, "(?<![\w-])$key\s*=")).Count
     if ($n -ne 1) {
--- a/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
+++ b/.pipeline/scripts/pin-mssql-tds-dep-version.ps1
@@ -79,13 +79,15 @@ foreach ($entry in @(
     }
 
     if ($keyMatches.Count -eq 1) {
-        $dependency = $keyPattern.Replace($dependency, "$key = `"$value`"", 1)
+        $replacement = "$key = `"$value`"".Replace('$', '$$')
+        $dependency = $keyPattern.Replace($dependency, $replacement, 1)
     }
     else {
+        $replacement = "{ $key = `"$value`", ".Replace('$', '$$')
         $dependency = [regex]::Replace(
             $dependency,
             '\{\s*',
-            "{ $key = `"$value`", ",
+            $replacement,
             1
         )
     }
--- a/.pipeline/scripts/stamp-crate-versions.ps1
+++ b/.pipeline/scripts/stamp-crate-versions.ps1
@@ -82,12 +82,12 @@ function Set-MssqlTdsDependencyVersion {
     )
 
     $pattern = [regex]'(?m)^(?<prefix>\s*mssql-tds\s*=\s*\{[^\r\n}]*?\bversion\s*=\s*)"[^"]+"(?<suffix>[^\r\n}]*\}\s*)$'
-    $matches = $pattern.Matches($Content)
-    if ($matches.Count -ne 1) {
-        throw "Expected one versioned mssql-tds dependency in ${Path}, found $($matches.Count)"
+    $depMatches = $pattern.Matches($Content)
+    if ($depMatches.Count -ne 1) {
+        throw "Expected one versioned mssql-tds dependency in ${Path}, found $($depMatches.Count)"
     }
 
-    $match = $matches[0]
+    $match = $depMatches[0]
     $replacement = $match.Groups['prefix'].Value + "`"$Version`"" + $match.Groups['suffix'].Value
     return $Content.Substring(0, $match.Index) +
         $replacement +
--- a/.pipeline/scripts/stamp-crate-versions.ps1
+++ b/.pipeline/scripts/stamp-crate-versions.ps1
@@ -57,17 +57,16 @@ function Set-PackageVersion {
         throw "No [package] section found in $Path"
     }
 
-    $versionMatches = [regex]::Matches($packageSection.Value, '(?m)^version\s*=\s*"[^"]+"')
+    $versionMatches = [regex]::Matches($packageSection.Value, '(?m)^(version\s*=\s*)"[^"]+"')
     if ($versionMatches.Count -ne 1) {
         throw "Expected one package version in ${Path}, found $($versionMatches.Count)"
     }
 
-    $updatedSection = [regex]::Replace(
-        $packageSection.Value,
-        '(?m)^(version\s*=\s*)"[^"]+"',
-        "`${1}`"$Version`"",
-        1
-    )
+    $versionMatch = $versionMatches[0]
+    $replacement = $versionMatch.Groups[1].Value + "`"$Version`""
+    $updatedSection = $packageSection.Value.Substring(0, $versionMatch.Index) +
+        $replacement +
+        $packageSection.Value.Substring($versionMatch.Index + $versionMatch.Length)
 
     return $Content.Substring(0, $packageSection.Index) +
         $updatedSection +
--- a/.pipeline/scripts/stamp-crate-versions.ps1
+++ b/.pipeline/scripts/stamp-crate-versions.ps1
@@ -1,97 +1,147 @@
-# stamp-crate-versions.ps1
-# Patches Cargo.toml version fields for publishable crates before cargo publish.
-# Uses the same versioning scheme as the NuGet wheel packaging.
-#
-# Usage (in pipeline):
-#   pwsh .pipeline/scripts/stamp-crate-versions.ps1 -BuildReason "$(Build.Reason)" -BuildId "$(Build.BuildId)" -IsOfficial $false
-#
-# Usage (local testing):
-#   pwsh .pipeline/scripts/stamp-crate-versions.ps1 -BuildReason "IndividualCI" -BuildId "99999" -IsOfficial $false -WhatIf
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
 
+<#
+.SYNOPSIS
+  Stamps publishable crate manifests with the version for the current build.
+
+.DESCRIPTION
+  Official builds retain the source version. Non-official scheduled builds use a
+  nightly version, while other non-official builds use a unique dev version.
+  mssql-mock-tds is kept aligned with mssql-tds, including its dependency version.
+#>
+[CmdletBinding()]
 param(
-    [Parameter(Mandatory=$true)]
+    [Parameter(Mandatory = $true)]
     [string]$BuildReason,
 
-    [Parameter(Mandatory=$true)]
+    [Parameter(Mandatory = $true)]
     [string]$BuildId,
 
-    [Parameter(Mandatory=$false)]
-    [bool]$IsOfficial = $false,
+    [string]$IsOfficial = 'False',
 
-    [Parameter(Mandatory=$false)]
-    [string[]]$Crates = @("mssql-tds", "mssql-mock-tds"),
+    [string[]]$Crates = @('mssql-tds', 'mssql-mock-tds'),
 
-    [Parameter(Mandatory=$false)]
-    [switch]$WhatIf
+    [switch]$WhatIf,
+
+    [string]$RepositoryRoot = (Resolve-Path (Join-Path $PSScriptRoot '../..'))
 )
 
 $ErrorActionPreference = 'Stop'
 
-# Read base version from the first crate
-$cargoTomlPath = Join-Path $PSScriptRoot "../../$($Crates[0])/Cargo.toml"
-$cargoToml = Get-Content $cargoTomlPath -Raw
-# Match the first `version = "..."` at the start of a line (multiline mode).
-# This targets the [package] version, which appears before any dependency versions.
-if ($cargoToml -match '(?m)^version\s*=\s*"([^"]+)"') {
-    $baseVersion = $Matches[1]
-} else {
-    Write-Error "Could not extract version from $cargoTomlPath"
-    exit 1
+function Get-PackageVersion {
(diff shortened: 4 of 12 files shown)


mssql-mock-tds/Cargo.toml
```diff
@@ -21,7 +21,7 @@ default = []
 vendored-openssl = ["openssl/vendored"]
 
 [dependencies]
-mssql-tds = { path = "../mssql-tds", default-features = false }
+mssql-tds = { path = "../mssql-tds", version = "0.1.0", default-features = false }
```
David-Engel, line 26: **Blocking.** Confirming the Copilot finding from `ac98cc82` — it is still unaddressed on `92f6bc9f`, and I reproduced it rather than taking it on faith.

`.pipeline/OneBranch/PublishFeeds-Sandbox.yml:600` runs `pin-mssql-tds-dep-version.ps1` on every sandbox publish *and* every dry-run. That script prepends its own `version` key:

```powershell
$c = $c -replace 'mssql-tds\s*=\s*\{\s*path\s*=\s*"\.\./mssql-tds"',
                 "mssql-tds = { path = `"../mssql-tds`", version = `"$Version`", registry-index = `"$index`""
```

The prefix it matches is still present after your change, so the existing `version = "0.1.0"` survives in the tail and you get two. Its own duplicate-key guard at lines 76-80 then fails the step. Running the script against this head's manifest:

```
Write-Error: Expected exactly one 'version' in the mssql-tds dependency of
mssql-mock-tds/Cargo.toml, found 2: mssql-tds = { path = "../mssql-tds",
version = "0.1.0-sandbox.1", registry-index = "sparse+https://...",
version = "0.1.0", default-features = false }
ExitCode=1
```

The step ordering doesn't save it: `stamp-crate-versions-sandbox.ps1` (line 598) rewrites only `[package].version`, so the dependency's `version` is untouched when `pin` runs.

The good news is that it fails loudly and *before* `cargo-publish-mock.ps1`, so no feed version gets burned. The bad news is that this pipeline isn't part of PR validation, so the green checks on this head say nothing about it — the first sandbox run after merge is where it surfaces.

Suggested fix: make `pin` upsert instead of prepend, mirroring the `Set-MssqlTdsDependencyVersion` helper you just added to `stamp-crate-versions.ps1`:

```powershell
# replace an existing version, or insert one if absent
if ($c -match '(?m)^\s*mssql-tds\s*=\s*\{[^\r\n}]*\bversion\s*=') {
    $c = $c -replace '(?m)^(\s*mssql-tds\s*=\s*\{[^\r\n}]*?\bversion\s*=\s*)"[^"]+"', "`${1}`"$Version`""
} else {
    $c = $c -replace 'mssql-tds\s*=\s*\{\s*path\s*=\s*"\.\./mssql-tds"',
                     "mssql-tds = { path = `"../mssql-tds`", version = `"$Version`""
}
# registry-index still needs its own insert-if-absent pass
```

Keep the duplicate-key count check as-is — it's the thing that caught this, and it should keep catching the next drift.

.pipeline/scripts/package-rust-crates.ps1
```diff
@@ -0,0 +1,84 @@
...
+
+Push-Location $RepositoryRoot
+try {
+    foreach ($crateName in $crateNames) {
+        # The dependent crate cannot be verified until mssql-tds reaches crates.io.
+        # Workspace validation runs separately, so package without a generated lockfile.
+        $arguments = @(
+            'package'
+            '--package', $crateName
+            '--no-verify'
```
David-Engel: **Suggestion.** The comment above is accurate for `mssql-mock-tds` but not for `mssql-tds`, and the loop applies `--no-verify` to both.

`mssql-tds` has no unpublished dependencies, so `cargo package -p mssql-tds` can build the packaged source and would catch the classic packaging failures — a file missed by `include`/`exclude`, a `path`-only dev-dependency, a `build.rs` input that never made it into the tarball. crates.io versions are immutable, so a `.crate` that doesn't compile is permanent; you burn the version and ship a broken one.

The PR description says you ran `cargo check` on the packaged source by hand for both crates, which is exactly the right check — this is just about moving it into the pipeline so the release gate enforces it every time rather than relying on someone remembering.

```powershell
$arguments = @(
    'package'
    '--package', $crateName
    '--allow-dirty'
    '--exclude-lockfile'
    '--target-dir', $TargetDirectory
)
# mssql-mock-tds cannot be verified until its mssql-tds dependency
# reaches crates.io; mssql-tds itself has no such constraint.
if ($crateName -eq 'mssql-mock-tds') {
    $arguments += '--no-verify'
}
```

The `Package Rust crates` job ran in 2m27s on the non-official pipeline, so there's room in the budget for the verify build.

.pipeline/scripts/test-cratesio-package.ps1
```diff
@@ -0,0 +1,59 @@
...
+        }
+        throw "crates.io returned HTTP $statusCode for $uri"
+    }
+
+    if ($statusCode -eq 200) {
+        Write-Host "$CrateName@$Version is available on crates.io."
+        return
+    }
+    if ($statusCode -ne 404) {
+        throw "crates.io returned HTTP $statusCode for $uri"
```
David-Engel, line 55: **Suggestion.** In the `Available` polling path, any status that isn't 200 or 404 throws immediately and kills the loop. That's the right call for the `Absent` preflight (`MaxAttempts 1`, fail fast before publishing), but this branch runs *after* `EsrpRelease` has already pushed the crate, across up to 40 requests over ~10 minutes. crates.io rate-limits with 429 and does have 5xx blips; a single one aborts `ReleaseCrates` on a crate that published fine.

The blast radius is real but bounded: `Wait for mssql-tds on crates.io` failing takes the `mssql-mock-tds` publish and the `Tag` stage with it. Your independent `publishMssqlTds` / `publishMssqlMockTds` switches make that recoverable by hand, which is good design — this just avoids needing them for a blip.

```powershell
if ($statusCode -ne 404) {
    if ($statusCode -lt 500 -and $statusCode -ne 429) {
        throw "crates.io returned HTTP $statusCode for $uri"
    }
    Write-Host "Transient HTTP $statusCode from crates.io; retrying."
}
```

Worth keeping the hard throw for 4xx other than 429/404 — those won't fix themselves.

.pipeline/scripts/validate-crate-release-artifact.ps1
```diff
@@ -0,0 +1,89 @@
...
+    $expectedFileName = "$expectedName-$($entry.version).crate"
+    if ([System.IO.Path]::GetFileName($cratePath) -ne $expectedFileName) {
+        throw "Expected crate file $expectedFileName, found $([System.IO.Path]::GetFileName($cratePath))"
+    }
+
+    $actualHash = (Get-FileHash $cratePath -Algorithm SHA256).Hash.ToLowerInvariant()
+    if ($actualHash -ne $entry.sha256) {
+        throw "SHA-256 mismatch for $expectedFileName"
+    }
+    $expectedFiles += $cratePath
```
David-Engel, line 68: **Suggestion.** The validator checks that both manifest entries carry the same *package* version, but never looks at the `mssql-tds` dependency version recorded inside the `mssql-mock-tds` `.crate` — which is the one field that, if wrong, produces a published crate nobody can resolve.

`Set-MssqlTdsDependencyVersion` does throw when the dependency isn't in the expected shape, so this isn't currently reachable. But that check runs on the build agent hours earlier, in a different stage; this script is the last gate before an irreversible publish, and it already opens and hashes the file. Confirming the dependency is cheap and closes the loop:

```powershell
if ($expectedName -eq 'mssql-mock-tds') {
    $staging = Join-Path ([System.IO.Path]::GetTempPath()) ([guid]::NewGuid())
    New-Item -ItemType Directory -Path $staging -Force | Out-Null
    try {
        tar -xzf $cratePath -C $staging
        $packaged = Get-Content (Join-Path $staging "$expectedName-$($entry.version)/Cargo.toml") -Raw
        $dep = [regex]::Match($packaged, '(?ms)^\[dependencies\.mssql-tds\].*?^version\s*=\s*"([^"]+)"')
        if (-not $dep.Success -or $dep.Groups[1].Value -ne $entry.version) {
            throw "mssql-mock-tds records mssql-tds dependency version '$($dep.Groups[1].Value)', expected '$($entry.version)'"
        }
        if ($packaged -match '(?ms)^\[dependencies\.mssql-tds\][^\[]*^path\s*=') {
            throw 'mssql-mock-tds still carries a path dependency on mssql-tds'
        }
    }
    finally { Remove-Item $staging -Recurse -Force -ErrorAction SilentlyContinue }
}
```

I ran the extraction path against a locally produced `.crate` on this head and it yields `[dependencies.mssql-tds] / version = "0.1.0"` with `path` stripped, so the assertion matches what cargo actually emits. Pick whichever half you think is worth the lines — the version equality is the one that matters.

.pipeline/scripts/stamp-crate-versions.ps1
```diff
@@ -1,97 +1,147 @@
...
+
+function Set-MssqlTdsDependencyVersion {
+    param(
+        [Parameter(Mandatory = $true)][string]$Content,
+        [Parameter(Mandatory = $true)][string]$Version,
+        [Parameter(Mandatory = $true)][string]$Path
+    )
+
+    $pattern = [regex]'(?m)^(?<prefix>\s*mssql-tds\s*=\s*\{[^\r\n}]*?\bversion\s*=\s*)"[^"]+"(?<suffix>[^\r\n}]*\}\s*)$'
+    $matches = $pattern.Matches($Content)
```
David-Engel: **Nit.** `$matches` is a PowerShell automatic variable, populated by `-match`. Assigning to it works, but it shadows the built-in inside this function and reads as a bug to anyone scanning for a `-match` nearby. `$depMatches` (or `$found`) costs nothing and removes the double-take — the `$match = $matches[0]` on line 90 is where it's most confusing.

David-Engel (commented):
## Summary

Adds a `RustCrates` build job that stamps, packages and hashes `mssql-tds` + `mssql-mock-tds` into a `drop_Build_RustCrates` artifact, plus an opt-in `ReleaseCrates` stage that validates that artifact and publishes each crate to crates.io through `EsrpRelease@12`, in dependency order, with crates.io preflight and post-publish polling. The shape is sound: publish switches default to `false`, the two crates publish in the right order with a gate on `mssql-tds` landing first, `mssql-mock-tds` can be republished on its own after a partial run, and the artifact validator checks names, order, filenames, SHA-256 and that no extra `.crate` files rode along (including a path-escape check on `entry.file`). The diff matches the description, and `Fixes #514` is present.

I verified the core claim locally: `cargo package --package mssql-mock-tds --no-verify --allow-dirty --exclude-lockfile` succeeds on this head, and the `Cargo.toml` inside the resulting `.crate` normalizes to `[dependencies.mssql-tds] / version = "0.1.0"` with `path` stripped — which is exactly what the manifest change is for. I also confirmed `--exclude-lockfile` is stable (cargo 1.95.0), that neither crate name is taken on crates.io, and that `GET /api/v1/crates/{name}/{version}` returns 404 for an absent version, so `test-cratesio-package.ps1`'s state mapping is right. PR validation is green on this head; note that `PublishFeeds-Sandbox.yml` is not part of it, which matters for the finding below.

One blocking item, which is a confirmation of the Copilot bot's finding rather than a new one — it was raised on `ac98cc82` and is still live on `92f6bc9f`. I reproduced it rather than restating it, and the mechanism holds exactly as described.

*Posted by an unattended review run. These findings have not been checked by a human first.*

### Blocking

1. `mssql-mock-tds/Cargo.toml:24` — the added `version` key breaks the sandbox publish pipeline. Details inline.

### Suggestion

2. `.pipeline/scripts/package-rust-crates.ps1:40` — `--no-verify` is applied to `mssql-tds` as well, where it isn't needed.
3. `.pipeline/scripts/test-cratesio-package.ps1:50` — a transient crates.io 429/503 aborts the post-publish wait.
4. `.pipeline/scripts/validate-crate-release-artifact.ps1:68` — the validator never checks the `mssql-tds` dependency version recorded inside the `mssql-mock-tds` `.crate`.

### Nit

5. `.pipeline/scripts/stamp-crate-versions.ps1:85` — `$matches` shadows the automatic variable.

One observation, not a finding: the checklist leaves `New/changed functionality has tests` unchecked for ~410 lines of new PowerShell with non-trivial regex. There is no `*.Tests.ps1` anywhere in the repo, so there is no existing pattern to follow and I'm not asking you to invent one — but points 1 and 4 are both the kind of thing a five-line Pester case would have caught, if you ever want a place to start.

No issues found in the ESRP task wiring, the stage dependency graph, the `Tag` stage's conditional `dependsOn`, or the artifact naming.

.pipeline/scripts/stamp-crate-versions.ps1
saurabh500: Renamed the local collection to `$depMatches` so it no longer shadows PowerShell's automatic `$Matches` variable.

> [!NOTE]

.pipeline/scripts/package-rust-crates.ps1
saurabh500: Removed `--no-verify` for `mssql-tds`, so `cargo package` now verifies its packaged source in every build. The flag remains only for `mssql-mock-tds`, whose unpublished dependency prevents verification before release.

> [!NOTE]

mssql-mock-tds/Cargo.toml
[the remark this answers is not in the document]
saurabh500: Updated `pin-mssql-tds-dep-version.ps1` to upsert both `version` and `registry-index` instead of prepending duplicate keys. I exercised it against both the current versioned manifest and the legacy version-less shape.

> [!NOTE]

.pipeline/scripts/validate-crate-release-artifact.ps1
saurabh500: The release validator now reads the normalized `Cargo.toml` directly from the `.crate`, requires the `mssql-tds` dependency version to match, and rejects residual path or registry overrides. A deliberately corrupted archive was rejected in validation.

> [!NOTE]

mssql-mock-tds/Cargo.toml
saurabh500: Fixed the sandbox regression by making the pin script replace existing values or insert missing ones while retaining the duplicate-key guard. Both manifest shapes now produce exactly one `version` and one `registry-index`.

> [!NOTE]

.pipeline/scripts/test-cratesio-package.ps1
saurabh500: The post-publish path now retries request failures, HTTP 429, and HTTP 5xx responses, while still failing immediately for other 4xx statuses. The preflight `Absent` mode remains fail-fast.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -200,12 +210,127 @@ extends:
...
+    ###########################################################################
+    - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
+      - stage: ReleaseCrates
+        displayName: 'Release Rust Crates'
+        dependsOn: Release
+        jobs:
+        - job: PublishRustCrates
+          displayName: 'Publish Rust crates to crates.io'
+          templateContext:
+            type: releaseJob
```
David-Engel, line 228: **Suggestion — a question, not an assertion.** This job declares `templateContext.type: releaseJob` / `isProduction: true` and then runs three repo scripts by relative path:

```yaml
- pwsh: >
    .pipeline/scripts/validate-crate-release-artifact.ps1
    -ArtifactDirectory '$(Pipeline.Workspace)\officialBuild\drop_Build_RustCrates'
```

A `pwsh` step resolves that against `$(System.DefaultWorkingDirectory)`, so it needs the sources synced into the agent. Every other job in this file that touches the repo either is a plain job reading `$(Build.SourcesDirectory)` (`PublishRelease`, line 137) or declares the checkout explicitly (`TagAndBranch`, line 342 — `- checkout: self` with `persistCredentials: true`). This one does neither.

My understanding is that 1ES release jobs don't sync sources by default, precisely so a release stage consumes only signed artifacts. That's internal guidance and I couldn't fetch a citable source for it, which is why this is a question — I'd rather flag it than assert it. The cost is asymmetric either way: an explicit checkout is a no-op if sources were already present, and the entire `ReleaseCrates` stage fails on its first real run if they weren't.

```yaml
          steps:
          - checkout: self

          - download: officialBuild
            artifact: 'drop_Build_RustCrates'
```

This can't be settled by inspection — see the dry-run point on the stage condition above, which is what would let you answer it without risking a crates.io version.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -200,12 +210,127 @@ extends:
           displayName: 'Verify NuGet package'
 
     ###########################################################################
-    # Stage 2: Tag the source commit and create release branch
+    # Stage 2: Publish Rust crates through ESRP
+    ###########################################################################
+    - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
+      - stage: ReleaseCrates
```
David-Engel, line 221: **Suggestion.** The stage is gated on `or(publishMssqlTds, publishMssqlMockTds)`, so with both switches false it doesn't exist at all. The only way to run `validate-crate-release-artifact.ps1` and the two crates.io preflights in their real environment is therefore to arm an `EsrpRelease@12` publish in the same run — and crates.io versions are immutable, so a surprise there has no undo.

Everything upstream of the ESRP tasks is safe to repeat: the validator is read-only, and both preflights assert *absence*. Splitting the gate gives you a rehearsal for the stage, which is also how you'd answer the `checkout` question below without putting a version at risk:

```yaml
  - name: validateCratesOnly
    displayName: 'Validate crate artifact without publishing'
    type: boolean
    default: false
```

```yaml
    - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true), eq(parameters.validateCratesOnly, true)) }}:
      - stage: ReleaseCrates
```

Both `EsrpRelease` tasks and the post-publish waits already carry their own `${{ if }}` on the two publish switches, so they stay off in that mode and the job reduces to download + validate. Leave the `Tag` stage's `dependsOn` as it is — a validation-only run isn't a release, so it shouldn't gate the tag.

Not blocking: the design is already conservative — both switches default to `false`, the preflights fail closed, and the independent switches make a partial run recoverable. This is about making the safe half reachable on its own.

David-Engel (commented):
## Summary

Re-review of `91073f94`, which addresses all five findings from the pass on `92f6bc9f`. I verified each fix by running the scripts rather than reading them, and all five hold. No blocking items remain.

What I ran on this head, from a dedicated worktree with an isolated target dir:

- `pin-mssql-tds-dep-version.ps1` against **both** manifest shapes — the new versioned line and the legacy version-less one. Each produces exactly one `version` and one `registry-index`, and the duplicate-key guard is retained. That was the blocking finding; it is fixed.
- `package-rust-crates.ps1` end to end. `mssql-tds` now packages **with** verification (`Finished dev profile ... in 41s`, exit 0). That check matters because the flag change is new on this head — the OneBranch run cited in the description ran against `92f6bc9f`, where `--no-verify` still applied to both crates, so the verify path had never executed anywhere. It works: cargo strips the path-only dev-dependencies (`mssql-mock-tds` and the `test-util` self dep) and the verify step builds lib/bins only, so neither blocks the packaged build.
- `validate-crate-release-artifact.ps1` against the artifact that produced. Passes, sets both `##vso` version variables, and the new packaged-manifest assertion reads `[dependencies.mssql-tds] / version = "0.1.0"` with `path` stripped — exactly the field that, if wrong, ships an unresolvable crate.
- `test-cratesio-package.ps1` — read only. The `Available` branch now retries 429 and 5xx and still hard-fails other 4xx; `Absent` stays fail-fast at `MaxAttempts 1`. Correct.
- Confirmed the `stamp-crate-versions.ps1` rewrite is safe to make: the sandbox path calls `stamp-crate-versions-sandbox.ps1`, a separate script that only touches `[package].version`, so the modified script has exactly one caller — the new `RustCrates` job.

I also checked the artifact-name question I'd otherwise have raised: explicit `PublishPipelineArtifact@1` with a `drop_Build_*` name is the established pattern for every job in `stages.yml`, so `RustCrates` follows it. No finding there.

Two suggestions below, both about the release stage rather than the scripts, plus one nit. None blocks.

*Posted by an unattended review run. These findings have not been checked by a human first.*

### Blocking

None.

### Suggestion

1. `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml:223` — `PublishRustCrates` is a `releaseJob` with no `checkout: self`, yet its steps invoke repo scripts by relative path. Question, not an assertion — details inline.
2. `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml:216` — the stage only exists when a publish switch is armed, so the validator and preflights can't be rehearsed without also arming an irreversible publish.

### Nit

3. `.pipeline/scripts/pin-mssql-tds-dep-version.ps1:82` — `$` is a substitution metacharacter in a `Regex.Replace` replacement string.

On the unchecked `New/changed functionality has tests` box: I'm not re-raising it. There is still no `*.Tests.ps1` anywhere in the repo, and the two fixes I'd have pointed at as Pester candidates are both now guarded by the scripts themselves — the pin script's duplicate-key check and the validator's packaged-dependency assertion.

CI note: `Build Stage Test MacOS` and `Build mssql-python macOS (cross-repo)` are red on this head. This PR changes only pipeline YAML, four PowerShell scripts and one `Cargo.toml` dependency key, so I see no mechanism connecting them — but I did not reproduce them, and I'm not applying `ready for human review` while required validation is failing. Worth a look before you mark it ready.

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
```diff
@@ -58,21 +56,46 @@ $index = if ($IndexUrl -match '^sparse\+') { $IndexUrl } else { "sparse+$IndexUr
...
+    $value = $entry.Value
+    $keyPattern = [regex]"(?<![\w-])$key\s*=\s*`"[^`"]*`""
+    $keyMatches = $keyPattern.Matches($dependency)
+    if ($keyMatches.Count -gt 1) {
+        Write-Error "Expected at most one '$key' in the mssql-tds dependency of ${path}, found $($keyMatches.Count)"
+        exit 1
+    }
+
+    if ($keyMatches.Count -eq 1) {
+        $dependency = $keyPattern.Replace($dependency, "$key = `"$value`"", 1)
```
David-Engel: **Nit.** `Regex.Replace` treats `$` in the *replacement* string as a substitution metacharacter (`$1`, `$&`, `$$`, `${name}`), and both branches interpolate caller-supplied values straight into it — this line and the insert at lines 85-90.

Neither input can carry a `$` today: `$Version` is a semver string and `$IndexUrl` is an ADO feed URL. I ran the script against both the versioned and the version-less manifest shapes and confirmed the output is correct in each case, so this is cosmetic rather than a live bug — flagging it only because the upsert rewrite is exactly the kind of code someone reuses with a less well-behaved input later.

```powershell
$dependency = $keyPattern.Replace($dependency, "$key = `"$value`"".Replace('$', '$$'), 1)
```

The insert branch wants the same treatment. Your call — the current form is fine for the inputs it has.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Added a default-off `validateCratesOnly` parameter. It enables download, artifact validation, and both absence preflights while keeping ESRP tasks and post-publish waits disabled; validation-only runs do not affect the tag dependency graph.

> [!NOTE]

.pipeline/scripts/pin-mssql-tds-dep-version.ps1
saurabh500: Escaped replacement-string dollar signs in both upsert branches before passing values to `Regex.Replace`. I also exercised the script with a dollar-bearing registry URL to verify the value remains literal.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Added an explicit `checkout: self` before the artifact download so the release job can reliably invoke the repository scripts regardless of release-job checkout defaults.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -200,12 +215,129 @@ extends:
...
+          - ${{ if or(eq(parameters.publishMssqlMockTds, true), eq(parameters.validateCratesOnly, true)) }}:
+            - pwsh: >
+                .pipeline/scripts/test-cratesio-package.ps1
+                -CrateName 'mssql-mock-tds'
+                -Version '$(mssqlMockTdsCrateVersion)'
+                -ExpectedState Absent
+                -MaxAttempts 1
+              displayName: 'Preflight mssql-mock-tds version'
+
+          - ${{ if eq(parameters.publishMssqlTds, true) }}:
```
David-Engel: **Suggestion.** `validateCratesOnly` widens the stage condition and the two preflight conditions with `or(...)`, but nothing downstream consults it — so it reads as a mode while behaving as an extra way to switch the stage on. Two combinations fall out of that:

- `validateCratesOnly=true` **and** `publishMssqlTds=true`: the switch labelled *"Validate Rust crate artifacts without publishing"* is checked, and the `EsrpRelease@12` task on this line still runs. crates.io versions are immutable, so that one has no undo — the same hazard the `default: false` comments elsewhere in this file are guarding against.
- `validateCratesOnly=true` **and** `publishMssqlMockTds=true` with `publishMssqlTds=false`: this is exactly the mock-only republish the independent switches exist for, and it can't pass. `Preflight mssql-tds version` (line 248, `-ExpectedState Absent`) and `Verify mssql-tds dependency is published` (line 292, `-ExpectedState Available`) both run against `$(mssqlTdsCrateVersion)` in the same job, and the absence check fails first.

The smallest fix that makes this a real dry run is to let the publish switches keep expressing *intent* and have `validateCratesOnly` strip only the irreversible half — the two `EsrpRelease@12` tasks and their post-publish waits:

```yaml
          - ${{ if and(eq(parameters.publishMssqlTds, true), ne(parameters.validateCratesOnly, true)) }}:
            - task: EsrpRelease@12
```

with the same treatment on the `publishMssqlMockTds` block, and the two preflight conditions reverted to the plain `eq(parameters.publish*, true)` they had on `91073f94`. An operator then arms the switches for the release they intend, adds `validateCratesOnly`, and gets download + validate + the exact preflight set that run would use.

The trade-off is that `validateCratesOnly` alone, with both publish switches off, would run the validator but no preflights. If you'd rather keep today's shape, the narrower fix clears both symptoms too: add `ne(parameters.validateCratesOnly, true)` to the two ESRP gates and to the `Verify mssql-tds dependency is published` condition on line 292.

Not blocking — the first case needs two switches set at once, and the second fails closed.

David-Engel (commented):

## Summary

Re-review of `dc6485be`, which adds a `validateCratesOnly` dry-run switch, an explicit `checkout: self` on the release job, and `$`-escaping in the pin script's `Regex.Replace` calls. All three findings from the pass on `91073f94` are addressed, and every required check is green on this head — including `Build Stage Test MacOS` and `Build mssql-python macOS (cross-repo)`, which were red last time. No blocking items.

## Verification

Run from a dedicated worktree with an isolated target dir:

- `pin-mssql-tds-dep-version.ps1` with an index URL containing `$feed$`. The value survives verbatim into the emitted `registry-index`, and a second run against the already-pinned manifest still leaves exactly one `version` and one `registry-index` key. That was the nit on `91073f94`; it is fixed, and the duplicate-key guard is intact.
- `stamp-crate-versions.ps1` against copied manifests. It stamps `0.1.0` → `0.1.0-dev.<date>.<buildId>` in both `[package]` sections, rewrites the mock's `mssql-tds` dependency to match, and the re-stamp guard fires on a second run.
- Read the two YAML fixes against the file rather than the diff: `checkout: self` is in place ahead of the artifact download, and the stage condition now admits a run with both publish switches off.
- `gh pr checks 515` — all 19 checks pass on this head. Diff coverage reports 100% / "no lines with coverage information", which is expected for a YAML + PowerShell diff.

I did not re-verify the packaging path end to end; `package-rust-crates.ps1` and `validate-crate-release-artifact.ps1` are unchanged since the pass that ran them.

### Blocking

None.

### Suggestion

1. `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml:266` — `validateCratesOnly` is additive rather than exclusive, so it neither suppresses a publish nor makes the mock-only path rehearsable. Details inline.

### Nit

2. The PR description still describes only the packaging and ESRP publishing from earlier commits. `validateCratesOnly` is the operator-facing knob most likely to be hunted for later — worth a line in the description before you mark this ready.

3. `.pipeline/OneBranch/stages.yml:75` — the `RustCrates` job sets `CARGO_TARGET_DIR: C:\cargo_target_dir`, matching the other Windows jobs, but the job's only cargo invocation is `package-rust-crates.ps1`, which passes `--target-dir '$(Agent.TempDirectory)\rust-crate-target'`. `--target-dir` takes precedence, so one of the two is dead config. Harmless either way; just pick one so the next reader doesn't have to work out which wins.

Nothing new in the ESRP task wiring, the artifact shape (`drop_Build_RustCrates` root holds `release-manifest.json` plus one folder per crate, which is what the validator and both `folderlocation` inputs expect), or the `Tag` stage's conditional `dependsOn`.

On the unchecked `New/changed functionality has tests` box: still not re-raising it. There is no `*.Tests.ps1` anywhere in the repo, and the behaviors I'd have wanted covered are now guarded by the scripts' own checks.

This looks ready for human review, but this is not an approval.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Made `validateCratesOnly` exclusive for irreversible operations: selected-crate preflights still run, while both ESRP tasks and their post-publish waits are omitted. Mock-only validation now performs the published-dependency check without an incompatible absence preflight for `mssql-tds`.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -200,12 +215,129 @@ extends:
...
+              displayName: 'Verify mssql-mock-tds on crates.io'
+
+    ###########################################################################
+    # Stage 3: Tag the source commit and create release branch
     ###########################################################################
     - ${{ if eq(parameters.tagRelease, true) }}:
       - stage: Tag
         displayName: 'Tag Release and Create Branch'
-        dependsOn: Release
+        ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
```
David-Engel: **Suggestion.** Now that `validateCratesOnly` strips the ESRP tasks, this `dependsOn` fork lets a dry run produce a real release tag.

`tagRelease=true` + `validateCratesOnly=true` + either publish switch: `ReleaseCrates` still exists (line 220 admits `validateCratesOnly`), runs download → validate → preflight, publishes nothing, and succeeds. `Tag` then depends on it and `TagAndBranch` pushes `v$version` and `release/$version` (lines 381-384) for a version that never reached crates.io. Before this commit that combination actually published, so the tag matched reality; now it cannot.

It is recoverable — `git push origin --delete` both refs — but the cost lands later: the real release run fails at `git tag -a` (line 373) because the tag already exists, after the crates have gone out and while someone is watching a red release stage.

```yaml
    - ${{ if and(eq(parameters.tagRelease, true), ne(parameters.validateCratesOnly, true)) }}:
      - stage: Tag
```

That leaves the `dependsOn` fork exactly as written and just makes the whole stage inert in validation mode, matching the treatment you gave the two ESRP gates at lines 266 and 302.

Not blocking: it takes two switches set together, and checking `tagRelease` on a rehearsal is arguably operator error. Worth the one-line guard anyway, since the failure surfaces during the release it was meant to rehearse.

David-Engel (commented):

## Summary

Re-review of `104601de`, which makes `validateCratesOnly` exclusive rather than additive: the two `EsrpRelease@12` tasks and their post-publish waits are now gated on `ne(parameters.validateCratesOnly, true)`, the preflights are back to the plain publish switches, and the dead `CARGO_TARGET_DIR` on the `RustCrates` job is gone. That resolves all three items from the pass on `dc6485be`, including the mock-only case that previously could not pass. Every required check is green on this head. No blocking items.

## Verification

Run from the assigned worktree with an isolated target dir:

- Walked the `ReleaseCrates` condition matrix against the file rather than the diff. `validateCratesOnly` + both publish switches: validate + both `Absent` preflights, no ESRP, no waits. `validateCratesOnly` + `publishMssqlMockTds` only: mock `Absent` preflight plus `Verify mssql-tds dependency is published` (line 296, `Available`), which is the right precondition for a mock-only republish and is exactly the combination that could not pass before. `validateCratesOnly` alone: validate only. The real release paths are unchanged from the pass that checked them.
- Parsed all four changed YAML files with a YAML loader — no structural damage from the condition edits.
- Confirmed the `CARGO_TARGET_DIR` removal is a no-op for behavior: `package-rust-crates.ps1` passes `--target-dir` explicitly, and `--target-dir` takes precedence over the environment variable, so the packaging job runs against the same directory it did in the cited OneBranch run. Nothing else under `.pipeline/templates` reads that variable for this job.
- The four PowerShell scripts are byte-identical to `dc6485be`, which is the head where `pin-mssql-tds-dep-version.ps1` and `stamp-crate-versions.ps1` were executed, and `package-rust-crates.ps1` / `validate-crate-release-artifact.ps1` were executed on `91073f94`. I did not re-run them.
- `gh pr checks 515` — all 19 checks pass on this head, including the ADO PR validation matrix and both macOS jobs. Diff coverage reports "no lines with coverage information", expected for a YAML + PowerShell diff.
- Sanity-checked the publish metadata the release depends on: both manifests carry `description` + `license`, `mssql-tds/README.md` exists for its `readme` key, and the tracked content is 4.8 MB / 0.16 MB raw, comfortably inside the crates.io 10 MB limit once gzipped.

### Blocking

None.

### Suggestion

1. `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml:335` — with ESRP now stripped in validation mode, `tagRelease=true` + `validateCratesOnly=true` tags and pushes a release for a version that was never published. Details inline.

### Nit

2. The OneBranch evidence in the description is [run 20260908.8 against `92f6bc9f`](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=173508), which is three commits back. Since then the `RustCrates` job gained a real verify build for `mssql-tds` (`--no-verify` now applies only to the mock) and lost its `CARGO_TARGET_DIR`. Neither changes what cargo is told to do, but the verify build has still never compiled inside the OneBranch Windows image — a fresh non-official run on this head would close that gap cheaply, and it is the packaging half you cannot rehearse from the release pipeline.

Nothing new in the ESRP task wiring, the artifact shape, the validator's packaged-dependency assertion, or the stamping/pinning scripts. Still not re-raising the unchecked `New/changed functionality has tests` box: there is no `*.Tests.ps1` anywhere in the repo, and the behaviors worth covering are guarded by the scripts' own checks.

This looks ready for human review once you have decided on the tag gate, but this is not an approval.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Guarded the entire `Tag` stage with `ne(parameters.validateCratesOnly, true)`. Validation-only runs can no longer create release tags or branches, even if `tagRelease` is also selected.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -200,12 +215,129 @@ extends:
...
+    ###########################################################################
+    # Stage 3: Tag the source commit and create release branch
     ###########################################################################
-    - ${{ if eq(parameters.tagRelease, true) }}:
+    - ${{ if and(eq(parameters.tagRelease, true), ne(parameters.validateCratesOnly, true)) }}:
       - stage: Tag
         displayName: 'Tag Release and Create Branch'
-        dependsOn: Release
+        ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
+          dependsOn:
```
David-Engel: **Suggestion.** This fork is what makes the git tag mean "the crates went out too" — but the tag's version comes from a different crate.

`TagAndBranch` reads `mssql-py-core/Cargo.toml` (line 351) and pushes `v$version` / `release/$version`. On this head that is `0.1.10`, while the crates this stage publishes are `mssql-tds` and `mssql-mock-tds` at `0.1.0` (both `[package] version = "0.1.0"`, and `stamp-crate-versions.ps1` keeps the source version verbatim for `isOfficial`). So a full release run tags `v0.1.10` and `release/0.1.10` for a crates.io publish of `0.1.0`, and nothing in the repo records which crate versions that tag corresponds to.

That was harmless while the tag only covered the wheel NuGet package. Now that `ReleaseCrates` gates `Tag`, the tag is the only durable artifact of the crate release, and it names the wrong number.

Cheapest fix that keeps the existing scheme: put the crate version in the annotation, since the validator already publishes it as a job variable.

```yaml
              git tag -a $tagName $commitSha -m "Release $version (crates: $(mssqlTdsCrateVersion))"
```

(`mssqlTdsCrateVersion` is set by `validate-crate-release-artifact.ps1` in the `ReleaseCrates` job, so it would need re-deriving or passing across stages — a stage output variable, or just re-reading `mssql-tds/Cargo.toml` at the tagged commit.)

Alternative, if you would rather not couple them at all: leave `dependsOn: Release` and let the crate publish stand on its own. Your call — the point is only that today's tag silently claims a version the crates do not have.

.pipeline/scripts/stamp-crate-versions.ps1
```diff
@@ -1,97 +1,147 @@
...
     }
-    default {
-        $crateVersion = "$baseVersion-dev.$dateStamp.$BuildId"
-        Write-Host "CI build detected (reason: $BuildReason)"
+
+    $updatedSection = [regex]::Replace(
+        $packageSection.Value,
+        '(?m)^(version\s*=\s*)"[^"]+"',
+        "`${1}`"$Version`"",
+        1
```
David-Engel: **Nit.** The trailing `1` here is not a replacement count. `Regex` has no static `Replace(string, string, string, int)` overload — the only 4-arg static overload takes `RegexOptions`, so `1` binds to `RegexOptions.IgnoreCase` and the call replaces *every* match, case-insensitively. Confirmed against the overload list:

```
static string Replace(string input, string pattern, string replacement)
static string Replace(string input, string pattern, string replacement, RegexOptions options)
static string Replace(string input, string pattern, string replacement, RegexOptions options, timespan matchTimeout)
```

`[regex]::Replace('a{b{c', '\{', 'X', 1)` returns `aXbXc`, not `aXb{c`.

No live bug: the `$versionMatches.Count -ne 1` guard on line 61 means there is only ever one match to replace, and TOML keys are case-sensitive so the accidental `IgnoreCase` has nothing to catch. But the count-limited form is exactly what the code reads as, and you use the *correct* one seven lines away in the sibling script (`pin-mssql-tds-dep-version.ps1:83`, `$keyPattern.Replace($dependency, $replacement, 1)` — the instance overload, where the third argument really is a count).

```powershell
    $updatedSection = [regex]::new('(?m)^(version\s*=\s*)"[^"]+"').Replace(
        $packageSection.Value,
        "`${1}`"$Version`"",
        1
    )
```

Same treatment applies to `pin-mssql-tds-dep-version.ps1:87`, which is the other static-4-arg call; there the input is a single-line inline table with one `{`, so it is likewise inert today.

While you are in this function: `$Version` is interpolated straight into a `Regex.Replace` replacement string, so a `$` in it would be read as a substitution metacharacter. It can't be — the value is `<semver>[-dev.<date>.<buildid>]` — but it is the same class you escaped in the pin script on `dc6485be`, and escaping it here keeps the two consistent.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -200,12 +215,129 @@ extends:
...
+                -CrateName 'mssql-mock-tds'
+                -Version '$(mssqlMockTdsCrateVersion)'
+                -ExpectedState Available
+              displayName: 'Verify mssql-mock-tds on crates.io'
+
+    ###########################################################################
+    # Stage 3: Tag the source commit and create release branch
     ###########################################################################
-    - ${{ if eq(parameters.tagRelease, true) }}:
+    - ${{ if and(eq(parameters.tagRelease, true), ne(parameters.validateCratesOnly, true)) }}:
```
David-Engel: **Nit.** This guard is the right fix for the tag-without-publish case, and I am not asking you to undo it — just to note the side effect it now has on the *wheel* half of the pipeline.

`validateCratesOnly` is documented as "Validate selected Rust crates without publishing" (line 41), but it now suppresses the `Tag` stage unconditionally. So `publishNuGet=true` + `tagRelease=true` + `validateCratesOnly=true` publishes the NuGet package to the feed and then silently skips the tag and release branch — a real release with no tag, and no failed step to notice. The two knobs read as independent, and nothing in the parameter description hints they are not.

A comment on the parameter is enough to make the coupling discoverable:

```yaml
  - name: validateCratesOnly
    displayName: 'Validate selected Rust crates without publishing'
    type: boolean
    default: false
    # Also suppresses the Tag stage entirely: a rehearsal must not tag a
    # release. Do not combine with publishNuGet + tagRelease.
```

David-Engel (commented):

## Summary

Re-review of `ac2cb8b7`, which guards the whole `Tag` stage with `ne(parameters.validateCratesOnly, true)` — the one suggestion left open on `104601de`. That resolves it: a validation-only run can no longer produce `v$version` / `release/$version` for crates that were never published. All 34 checks pass on this head, including the non-official OneBranch run whose `Package Rust crates` job exercises the `mssql-tds` verify build in the real Windows image (the gap flagged as a nit last time — now closed by [run 173571](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_build/results?buildId=173571)). No blocking items.

## Verification

This diff is entirely pipeline definitions plus new PowerShell that runs inside release automation, so this pass was **read-only by policy** — I did not execute any script from the diff. What I did check, from the assigned worktree:

- Walked the full stage/step condition matrix against the file rather than the diff. The `Tag` stage's `dependsOn` fork can never name a stage that does not exist: `ReleaseCrates` is admitted by `publishMssqlTds || publishMssqlMockTds || validateCratesOnly`, and the fork only selects it under the first two. Real-release, mock-only-republish, and all three `validateCratesOnly` combinations resolve to the step sets you would want.
- Confirmed the manifest change is safe workspace-wide: `mssql-mock-tds` is the only crate that pins a version on the `mssql-tds` path dependency (`git grep '^\s*mssql-\(tds\|mock-tds\)\s*='` across all `Cargo.toml`); `mssql-tds-cli`, `mssql-odbc`, `mssql-js`, `mssql-py-core`, `mssql-tds-bench` and the fuzz target are all path-only, so stamping `mssql-tds` to a `-dev.*` version cannot break workspace resolution.
- Confirmed `stamp-crate-versions.ps1` still has exactly one caller (`stages.yml:86`); the sandbox path uses `stamp-crate-versions-sandbox.ps1` + `pin-mssql-tds-dep-version.ps1`, so the rewrite's blast radius is the new job only.
- Verified `install-dependencies.yml` and `cargo-authenticate-template.yml` actually declare the parameters the new `RustCrates` job passes (`osType`, `skipPythonSetup`, `skipRustupInstall`, `enableJsDeps`), and that its pool/`ob_outputDirectory`/explicit-`PublishPipelineArtifact` shape matches `Windows_x64`.
- Re-read the pin script's upsert against the current manifest line by hand: `version` is found once and replaced, `registry-index` is absent and inserted, and the trailing duplicate-key guard sees exactly one of each. The blocking finding from `92f6bc9f` stays fixed.
- `gh pr checks 515` — 34/34 pass. Coverage reports "no lines with coverage information", expected for a YAML + PowerShell diff.
- The only thing I ran was a standalone .NET probe of `Regex.Replace` overload resolution (no repo code involved), which is the basis for nit 2.

### Blocking

None.

### Suggestion

1. `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml:336` — the tag now also marks a crate release, but it is computed from a different crate's version. Details inline.

### Nit

2. `.pipeline/scripts/stamp-crate-versions.ps1:69` — `[regex]::Replace(input, pattern, replacement, 1)` binds `1` as `RegexOptions.IgnoreCase`, not a replacement count. Same shape at `pin-mssql-tds-dep-version.ps1:87`. Harmless today; details inline.

3. `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml:332` — `validateCratesOnly` now suppresses tagging for the *wheel* release too. Details inline.

Nothing new in the ESRP task wiring, the artifact shape, the validator's packaged-dependency assertion, the crates.io preflight/poll state machine, or the packaging script. Still not re-raising the unchecked `New/changed functionality has tests` box: there is no `*.Tests.ps1` anywhere in the repo, and the behaviors worth covering are guarded by the scripts' own checks.

This looks ready for human review, but this is not an approval.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Decoupled the existing `mssql-py-core` tag stage from `ReleaseCrates`: it again depends only on the NuGet release stage and is explicitly named as an `mssql-py-core` tag. Crate publishing now stands on its own instead of implicitly attaching mismatched crate versions to that tag.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Removed the coupling rather than documenting it: validation mode no longer suppresses or controls the independent `mssql-py-core` tag stage. The tag stage is again governed solely by `tagRelease` and the NuGet release stage.

> [!NOTE]

.pipeline/scripts/stamp-crate-versions.ps1
saurabh500: Replaced both static four-argument calls with unambiguous count-limited logic: package stamping now splices the single validated match, and dependency insertion uses an instance `Regex.Replace(..., 1)`. The focused stamping test confirms the package key and unrelated dependency versions are preserved.

> [!NOTE]

mssql-mock-tds/Cargo.toml
David-Engel, line 26: **Suggestion.** Neither manifest sets `repository`, so both crates.io pages would render with no link back to the source. `cargo package` says it directly for this crate: `warning: manifest has no documentation, homepage or repository`.

`mssql-tds/Cargo.toml` has `description`, `license`, `readme`, `keywords` and `categories` but no `repository` either; this manifest additionally has no `keywords`/`categories`, which is cosmetic (and `README.md` is picked up automatically without a `readme` key).

Package metadata is baked into the published version and cannot be edited afterwards, so this is fixable now or by burning a version later.

```toml
repository = "https://github.com/microsoft/mssql-rs"
```

David-Engel (commented):

## Summary

Re-review of `ee2fd0e7`, which decouples the `Tag` stage from `ReleaseCrates` — it is governed by `tagRelease` alone again, depends only on `Release`, and is named as an `mssql-py-core` tag — and replaces both static four-argument `Regex.Replace` calls with count-correct forms. All three items from the pass on `ac2cb8b7` are resolved, and all 34 checks pass on this head. No blocking items.

Net against `main`, the `Tag` stage's gating and `dependsOn` are now identical to `main`; only display strings changed. So the tag-naming-a-different-crate's-version problem and the `validateCratesOnly`-silently-suppresses-the-wheel-tag side effect are both gone rather than documented around.

Two new suggestions, neither a regression, both about what the *published* crate contains rather than about the pipeline. They are worth settling before the first publish because a crates.io version is immutable.

## Verification

From the assigned worktree with an isolated target dir:

- `stamp-crate-versions.ps1` against copied manifests via `-RepositoryRoot`. It stamps `0.1.0` → `0.1.0-dev.20260908.999` in both `[package]` sections and rewrites the mock's `mssql-tds` dependency to match. `Compare-Object` against the originals shows exactly those lines changed — in particular the unrelated `version = "1.19.0"` line further down `mssql-tds/Cargo.toml` is untouched, which is the case the new splice had to get right. The re-stamp guard throws on a second run.
- `pin-mssql-tds-dep-version.ps1` against **both** manifest shapes with an index URL containing `$feed$`. The versioned line is replaced and `registry-index` inserted; the legacy version-less line gets both keys inserted at the `{`; the `$` survives verbatim; a second run still leaves exactly one of each key; the `~force-auth` guard still rejects.
- `cargo package --package mssql-mock-tds --no-verify --allow-dirty --exclude-lockfile` on this head. The packaged `Cargo.toml` normalizes to `[dependencies.mssql-tds]` / `version = "0.1.0"` with `path` stripped, and `.cargo_vcs_info.json` records `sha1 = ee2fd0e78f57ded975492b9891f067c1c66f1445`. The crate carries its own source-commit provenance, so decoupling the git tag costs no traceability for the crate release.
- Grepped for dangling references after the stage edit: nothing outside `OfficialPythonWheelsRelease.yml` names `ReleaseCrates`, and `Release` is unconditional, so `dependsOn: Release` always resolves in every parameter combination.
- `gh pr checks 515` — 34/34 pass, nothing pending or failing.

### Blocking

None.

### Suggestion

1. **`mssql-tds` would publish a `.env` inside the `.crate`.** `cargo package --list --package mssql-tds` returns 207 files and `.env` is one of them: `mssql-tds/.env` is tracked (`git ls-files mssql-tds/.env` returns it), so the `**/.env` rule in `.gitignore` does not keep it out of the package. Today's content is benign — `DB_USERNAME`, `DB_HOST`, `DB_PORT`, `ENABLE_TEST_TRACE`, `TRUST_SERVER_CERTIFICATE`, no password — which is why this is a suggestion rather than blocking. But this PR is what turns that file into a permanent, immutable artifact on crates.io, and anything credential-shaped added to it later ships with the next publish. The 50 `tests/` files ride along for the same reason.

   Verified fix, in `mssql-tds/Cargo.toml`. Adding this takes the list from 207 files to 153 with `.env`, `tests/` and `benches/` gone (I reverted it afterwards; the worktree is clean):

   ```toml
   exclude = [".env", "tests/**", "benches/**"]
   ```

   `tests/**` is safe for the verify step, which builds lib and bins only. `mssql-mock-tds` needs nothing here — its package is already 12 files with no `.env`.

2. **Neither crate sets `repository`.** Details inline on `mssql-mock-tds/Cargo.toml:24`.

### Nit

3. `.pipeline/scripts/stamp-crate-versions-sandbox.ps1:40` still uses the static four-argument `[regex]::Replace($m.Value, pattern, replacement, 1)` shape this head just fixed in its two siblings, and unlike the code you replaced it has no `Count -ne 1` guard in front of it. Pre-existing and outside this diff, and inert in practice because `[package]` carries one `version` line — but it is the same script family, reached from `PublishFeeds-Sandbox.yml`, and it is a one-line change if you want the class gone entirely.

Nothing new in the ESRP task wiring, the artifact shape, the validator's packaged-dependency assertion, the crates.io preflight/poll state machine, or the packaging script. Still not re-raising the unchecked `New/changed functionality has tests` box: there is no `*.Tests.ps1` anywhere in the repo, and the behaviors worth covering are guarded by the scripts' own checks.

This looks ready for human review, but this is not an approval.

mssql-mock-tds/Cargo.toml
saurabh500: Added `repository = "https://github.com/microsoft/mssql-rs"` to both publishable manifests. I also rebuilt and inspected both `.crate` archives to verify the normalized metadata carries the source link.

> [!NOTE]

mssql-tds/Cargo.toml
```diff
@@ -5,8 +5,10 @@ edition = "2024"
 description = "Rust implementation of the TDS (Tabular Data Stream) protocol for SQL Server"
 license = "MIT"
 readme = "README.md"
+repository = "https://github.com/microsoft/mssql-rs"
```
David-Engel, line 9: **Suggestion.** `license = "MIT"` gives crates.io the SPDX field, but the license *text* never reaches the tarball. `LICENSE` is at the repo root, outside both package directories, so cargo cannot pick it up: `tar -tzf` on the two `.crate` files I built from this head returns no `LICEN*` entry for either crate. MIT asks that the notice accompany copies of the software, and a crates.io version is immutable — same reason `repository` and the `.env` exclusion were worth settling before the first publish rather than after.

Verified fix, in both `mssql-tds/Cargo.toml` and `mssql-mock-tds/Cargo.toml`:

```toml
license-file = "../LICENSE"
```

Cargo copies a `license-file` that lives outside the package root into the tarball root, so no per-crate copy of the file is needed and nothing else has to move. I added it to `mssql-mock-tds` and repackaged: 12 files -> 13, with `mssql-mock-tds-0.1.0/LICENSE` present, and cargo printed no warning about carrying both `license` and `license-file`. Reverted afterwards; the worktree is clean.

mssql-tds/Cargo.toml
```diff
@@ -5,8 +5,10 @@ edition = "2024"
...
 keywords = ["sql-server", "tds", "mssql", "database", "async"]
 categories = ["database", "network-programming"]
+exclude = [".env", "tests/**", "benches/**"]
```
David-Engel: **Nit.** `.env` is an exact path, not a pattern. The reason `mssql-tds/.env` was in the package at all is that it is force-tracked despite `.gitignore:31` (`**/.env`) — cargo packages tracked files — so a future `.env.test` or `.env.local` would have to be force-added the same way before it could ship. That makes this narrow rather than wrong, and today's list produces exactly the package you want (I confirmed the 153-file result).

Broadening it closes the class instead of the instance, at no cost:

```toml
exclude = ["**/.env*", "tests/**", "benches/**"]
```

I ran `cargo package --list --package mssql-tds` with that form on this head: same file set, `.env` still excluded, no `tests/` or `benches/` entries. Reverted afterwards.

David-Engel (commented):

## Summary

Re-review of `610cc4a8`, which adds `exclude = [".env", "tests/**", "benches/**"]` and `repository` to the publishable manifests and replaces the last static four-argument `Regex.Replace` — in `stamp-crate-versions-sandbox.ps1` — with a count-guarded splice. All three items from the pass on `ee2fd0e7` are resolved, and I verified each by running the code rather than reading it. No blocking items, and nothing is failing in CI.

The one thing this head changes that is worth a second look is the packaging behavior, because `exclude` interacts with the `[[bench]]`/`[[test]]` targets declared in `mssql-tds/Cargo.toml`. It is fine: cargo strips those target sections along with the excluded files, and the verify build proves the packaged source still compiles.

## Verification

From the assigned worktree with an isolated target dir:

- `package-rust-crates.ps1` end to end. `mssql-tds` packages **with** verification (`Finished dev profile ... in 18.55s`), `mssql-mock-tds` with `--no-verify`, and the manifest/hashes are written. The `mssql-tds` tarball is 153 files and contains only `src/`, `Cargo.toml`, `Cargo.toml.orig`, `README.md` and `.cargo_vcs_info.json` — no `.env`, no `tests/`, no `benches/`.
- Read the normalized manifest out of both `.crate` files. Cargo drops the three `[[bench]]` and 50-odd `[[test]]` sections whose files are no longer packaged (`warning: ignoring benchmark 'perf' as benches\perf.rs is not included in the published package`), so a consumer never sees a target pointing at a missing file. `repository` is present in both. Also confirmed there is no `src/**/tests/` directory in `mssql-tds`, so the root-anchored `tests/**` pattern drops only the integration suite.
- `validate-crate-release-artifact.ps1` against that artifact: passes, emits both `##vso` version variables, and the packaged-dependency assertion reads `[dependencies.mssql-tds]` / `version = "0.1.0"` with `path` stripped and `default-features = false` preserved.
- The sandbox pair on copied manifests, since `stamp-crate-versions-sandbox.ps1` is what changed here. It stamps both `[package]` versions to `0.1.0-dev.<date>.777` and the new `Count -ne 1` guard keeps the edit inside `[package]` — `[dependencies.uuid]` / `version = "1.19.0"` is untouched. `pin-mssql-tds-dep-version.ps1` then replaces the dependency `version` and inserts `registry-index`, leaving exactly one of each.
- `gh pr checks 515` — 23 pass, 6 pending, none failing. `Build Package Rust crates` passed in 3m19s on this head ([run 173608](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_build/results?buildId=173608)), so the verify build has now run in the OneBranch Windows image with the new `exclude` in place — the gap I nitted two passes ago is closed for the packaging half.

### Blocking

None.

### Suggestion

1. `mssql-tds/Cargo.toml:8` — neither published crate carries the MIT license text; `LICENSE` lives at the repo root, outside both package directories. Verified fix inline.

### Nit

2. `mssql-tds/Cargo.toml:11` — `.env` is an exact match rather than a pattern. Details inline.

Nothing new in the ESRP task wiring, the artifact shape, the crates.io preflight/poll state machine, the release-stage condition matrix, or the official stamping script — all unchanged since the passes that checked them. Still not re-raising the unchecked `New/changed functionality has tests` box: there is no `*.Tests.ps1` anywhere in the repo, and the behaviors worth covering are guarded by the scripts' own checks.

This looks ready for human review, but this is not an approval.

mssql-tds/Cargo.toml
saurabh500: Added `license-file = "../LICENSE"` to both publishable manifests. Rebuilt both archives and verified each contains a byte-identical `LICENSE` at its package root while retaining the MIT SPDX metadata.

> [!NOTE]

mssql-tds/Cargo.toml
saurabh500: Broadened the package exclusion to `**/.env*`. The rebuilt `mssql-tds` archive contains no environment files, integration tests, or benchmarks.

> [!NOTE]

David-Engel (commented):

## Summary

Re-review of `e18eba18`, which is a two-file delta on top of `610cc4a8`: `license-file = "../LICENSE"` added to both publishable manifests, and `exclude`'s `.env` entry widened to `**/.env*`. Both items from the previous pass are resolved, and I verified each by packaging rather than reading. No blocking items, and nothing is failing in CI.

The interesting question on this head was whether an out-of-package `license-file` actually survives packaging, since `LICENSE` lives at the repo root and neither crate directory contains it. It does: cargo relocates the file into the archive root and rewrites the normalized manifest to `license-file = "LICENSE"`. I confirmed that on cargo 1.95.0, which is the channel `rust-toolchain.toml` pins, so the behavior matches what `ms-prod-1.95` will do on the agent.

## Verification

From the assigned worktree with an isolated target dir:

- `package-rust-crates.ps1` end to end. `mssql-tds` packages **with** verification (`Finished dev profile ... in 46.71s`), `mssql-mock-tds` with `--no-verify --exclude-lockfile`, and the manifest/hashes are written. Both `.crate` files now contain a top-level `LICENSE` whose first lines are `MIT License` / `Copyright (c) Microsoft Corporation.`, and both normalized manifests read `license = "MIT"` plus `license-file = "LICENSE"` — the `../` is gone, so nothing dangles for a consumer.
- Re-checked the exclusions on the pipeline-produced archives: no path in either crate matches `.env`, `tests/`, or `benches/`. `mssql-tds` is `src/`, `Cargo.toml`, `Cargo.toml.orig`, `README.md`, `LICENSE`, `.cargo_vcs_info.json`; `mssql-mock-tds` adds `FLEXIBLE_QUERY_EXECUTION.md`. The widened `**/.env*` pattern still drops only `mssql-tds/.env` — that is the sole `.env*` file in the tree — and now also covers `.env.example`-style siblings if one is ever added.
- `validate-crate-release-artifact.ps1` against that artifact: passes, emits both `##vso` version variables, and the packaged-dependency assertion still reads `[dependencies.mssql-tds]` / `version = "0.1.0"` with `path` stripped. The extra `LICENSE` file does not perturb it — the script only counts `*.crate` files.
- `stamp-crate-versions.ps1` on copies of both new manifests. The count-guarded splice still lands on the `[package]` `version` line in each; `license`, `license-file`, `readme`, `repository`, and `exclude` are untouched.
- `cargo publish --dry-run -p mssql-tds --allow-dirty` — packages, verifies, and aborts at upload as expected. It does surface one warning, noted below.
- `gh pr checks 515` — 16 pass, 2 pending, none failing.

### Blocking

None.

### Suggestion

1. `.pipeline/scripts/package-rust-crates.ps1` — this head changes packaging behavior only, and packaging is not exercised by GitHub PR validation: there is no `Package Rust crates` job among the 18 checks here, because `buildRustCrates` only turns that job on in the OneBranch official/non-official pipelines. The Validation section still cites [run 20260909.4 against `ee2fd0e7`](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=173590), and the last OneBranch packaging evidence I can find is run 173608 on `610cc4a8` — both predate the `license-file` and `**/.env*` changes. My local run above is on the pinned 1.95 channel and should be representative, but a OneBranch build on `e18eba18` is the only thing that proves it on the `RUST-W22-SQL25` image with `ms-prod-1.95`. Worth queueing one and refreshing the Validation bullets before this goes to human review, since the next consumer of a stale link is whoever approves the immutable publish.

### Nit

2. `mssql-tds/Cargo.toml:6-7` and `mssql-mock-tds/Cargo.toml:6-7` — carrying both `license` and `license-file` makes cargo emit, during `cargo publish` verification:

   ```
   warning: only one of `license` or `license-file` is necessary
   `license` should be used if the package license can be expressed with a standard SPDX expression.
   `license-file` should be used if the package uses a non-standard license.
   ```

   This is cosmetic for the pipeline — `EsrpRelease@12` uploads the prebuilt `.crate` from `folderlocation` and never runs `cargo publish`, and crates.io prefers `license` when both are present, so the SPDX expression is what renders — but anyone publishing by hand will see it. If you want the license text shipped without the warning, the more common idiom is a `LICENSE` copy inside each crate directory and no `license-file` key at all; cargo packages any non-excluded file in the crate root, so the text still lands. That trades the warning for a duplicated file in the repo, so keeping it as-is is a defensible call — flagging it so the choice is deliberate.

Nothing else on this head. The ESRP task wiring, artifact shape, crates.io preflight/poll state machine, release-stage condition matrix, `validateCratesOnly` exclusivity, `Tag`-stage decoupling, and the pin/sandbox stamping scripts are all byte-identical to `610cc4a8` and were checked in earlier passes. Still not re-raising the unchecked `New/changed functionality has tests` box: there is no `*.Tests.ps1` harness anywhere in the repo, and the behaviors worth covering are guarded by the scripts' own `throw`s.

This looks close to ready for human review, but this is not an approval.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -200,11 +215,123 @@ extends:
...
+          - ${{ if and(eq(parameters.publishMssqlMockTds, true), ne(parameters.publishMssqlTds, true)) }}:
+            - pwsh: >
+                .pipeline/scripts/test-cratesio-package.ps1
+                -CrateName 'mssql-tds'
+                -Version '$(mssqlTdsCrateVersion)'
+                -ExpectedState Available
+                -MaxAttempts 1
+              displayName: 'Verify mssql-tds dependency is published'
+
+          - ${{ if and(eq(parameters.publishMssqlMockTds, true), ne(parameters.validateCratesOnly, true)) }}:
```
Theekshna, line 302: [Suggestion] `.pipeline/scripts/package-rust-crates.ps1` special-cases `mssql-mock-tds` to package with `--no-verify`, and this release stage goes straight from verifying the `mssql-tds` dependency is published to publishing `mssql-mock-tds` to crates.io, with no build/check step against the packaged artifact in between. `mssql-mock-tds` is the one crate this pipeline never automatically compiles from its packaged form before publishing — a missing file or bad packaged source would only surface after burning an immutable crates.io version. Consider unpacking the `.crate` file and running `cargo check` (or `cargo publish --dry-run`) against the extracted package before the `EsrpRelease@12` step.

Theekshna (commented):

First-pass review. Earlier sandbox/tagging/license threads on this PR are already resolved and I verified those fixes are real against the current diff. One new non-blocking gap remains in the release validation path.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 0 |

**Six required checks**
1. msodbcsql parity: N/A — changed files are `.pipeline/**` plus `mssql-tds/Cargo.toml` and `mssql-mock-tds/Cargo.toml`; nothing under `mssql-odbc/**`.
2. Test sufficiency: partial — manifest/ordering/hash preflight validation exists, but the packaged `mssql-mock-tds` artifact's compile path is still untested in automation, see inline comment.
3. Divergences documented: N/A — no separate release-process doc exists to diverge from beyond `.github/instructions/pr-workflow.instructions.md`, which this doesn't touch.
4. PR description currency: mostly current — diff matches description and links `Fixes #514`; checklist doesn't contradict green checks, but the OneBranch run cited in the body predates the current head SHA.
5. Verbose slop: none found; pipeline comments are dense but consistent with existing style.
6. Audit of PR's own evidence: re-derived crate publish coverage — workspace members are `mssql-tds`, `mssql-js`, `mssql-tds-cli`, `mssql-mock-tds`, `mssql-odbc`, `mssql-tds-bench`, with `mssql-py-core`/`mssql-mock-tds-py` excluded and the fuzz target unpublished. Only `mssql-tds` and `mssql-mock-tds` lack `publish = false`, and the new packaging/publish flow covers exactly those two — no gap there. `mssql-py-core` remains intentionally independent, consistent with the workspace exclusion.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Added packaged-source verification to `package-rust-crates.ps1`, where Rust is already installed. The job extracts both exact `.crate` files, points the extracted mock manifest at the extracted core package, and requires `cargo check` to pass before the artifact is emitted; the release-stage hash validation then guarantees those same bytes are published.

> [!NOTE]

David-Engel (commented):

## Summary

Re-review of `1d66060b`. Against `e18eba18` this head is a single-file, 43-line delta: `package-rust-crates.ps1` now extracts both `.crate` files, repoints the packaged `mssql-mock-tds` at the packaged `mssql-tds` via a temporary `path` key, and `cargo check`s it. That closes the last open item on the PR — the `--no-verify` hole for the mock — and it is the only thing that changed. No blocking items, no suggestions, no nits, and nothing is failing in CI.

The question worth answering on this head is whether the new gate is real or decorative, since a check that silently resolves `mssql-tds` from somewhere other than the packaged tarball would pass unconditionally. It is real, and I proved it by breaking it rather than by reading it.

## Verification

From the assigned worktree, cargo 1.95.0 (the channel `rust-toolchain.toml` pins), isolated target dir:

- `package-rust-crates.ps1` end to end. `mssql-tds` packages **with** verification, `mssql-mock-tds` with `--no-verify`, then the new block runs and prints `Verified packaged mssql-mock-tds against packaged mssql-tds.` Both `.crate` files and `release-manifest.json` land with matching SHA-256s.
- **The check compiles the packaged tds, not the repo's.** Cargo's own output names it: `Checking mssql-tds v0.1.0 (...\package-verification\mssql-tds-0.1.0)`, and every diagnostic path is rooted under `package-verification\`. The `--manifest-path` is outside the workspace, so the repo's `Cargo.toml`/`Cargo.lock` never enter the resolve.
- **Mutation test.** Appended a syntax error to `package-verification\mssql-tds-0.1.0\src\lib.rs` and re-ran the same `cargo check`: `error: could not compile mssql-tds (lib)`, exit code **101**, which the script's `$LASTEXITCODE` guard turns into a throw. The gate fails when the packaged source is broken, so it is not vacuous.
- The `.crate` artifacts are untouched by the verification. Hashes are computed on the copies in `-OutputDirectory` before extraction, and the `path` patch is applied only to the extracted tree under `-TargetDirectory`. Re-listed both tarballs afterwards to confirm: `mssql-tds` is `src/`, `Cargo.toml`, `Cargo.toml.orig`, `README.md`, `LICENSE`, `.cargo_vcs_info.json`; `mssql-mock-tds` adds `FLEXIBLE_QUERY_EXECUTION.md`. No `.env`, no `tests/`, no `benches/`, and `path` is still stripped from the shipped `[dependencies.mssql-tds]`.
- Feature coverage of the two gates composes correctly. The mock pulls `mssql-tds` with `default-features = false`, so this check exercises the reduced graph — but `cargo package -p mssql-tds` already builds the packaged tds with its default features, so between the two, both configurations of the packaged source are compiled before anything reaches ESRP.
- Re-ran `validate-crate-release-artifact.ps1` against the produced artifact: passes, emits both `##vso` version variables, packaged-dependency assertion still reads `version = "0.1.0"` with `path`/`registry`/`registry-index` absent.
- `gh pr checks 515` — none failing. **[Non-official run 173691](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_build/results?buildId=173691) is green on this head**, including `Build Package Rust crates` (3m35s), so the new block has now executed on the `RUST-W22-SQL25` image against a stamped `-dev` version — not just the `0.1.0` case I ran locally. That also retires the stale-evidence suggestion from the `e18eba18` pass; the Validation section's run link now points at a completed, passing run of the current head.

## Findings

### Blocking

None.

### Suggestion

None.

### Nit

None.

The rest of the diff is byte-identical to `e18eba18` and was checked in earlier passes: the ESRP task wiring and `folderlocation`s, the `drop_Build_RustCrates` artifact shape, the crates.io preflight/poll state machine, the `ReleaseCrates` condition matrix and `validateCratesOnly` exclusivity, the `Tag`-stage decoupling, `checkout: self` on the release job, the count-guarded stamping/pin scripts, and the manifest metadata (`repository`, `license-file`, `exclude`). All review threads on this PR are resolved, and I confirmed each earlier fix still holds in the tree rather than assuming it. Still not re-raising the unchecked `New/changed functionality has tests` box: there is no `*.Tests.ps1` harness anywhere in the repo, and every behavior worth covering is guarded by the scripts' own `throw`s — which this head extends rather than weakens.

This looks ready for human review, but this is not an approval.

Theekshna (approved):

PR #516, opened 2026-09-07
Reserve the mssql-python-rs Project on PyPI

gargsaumya:
### Summary

Adds a dedicated OneBranch pipeline and a placeholder package to reserve the `mssql-python-rs` distribution name on PyPI. This is the first step toward shipping the Rust-based native provider (TDS core + ODBC driver) as a PyPI package consumed by mssql-python.

- `.pipeline/OneBranch/PyPIRelease.yml` — OneBranch **Official** pipeline that publishes to PyPI via ESRP (same connected service / `MainPublisher: ESRPRELPACMAN` mssql-python uses). `publishToPyPI` defaults to **false** (dry run: verify only, no publish).
- `.pipeline/pypi-stub/` — a minimal `mssql-python-rs` `0.0.0` placeholder, with a **prebuilt, twine-checked sdist committed** under `dist/`.

### Why publish a committed sdist instead of building in the pipeline

The sealed OneBranch `type: windows` container has **no Python and no outbound network** (`UsePythonVersion` can't download; `pip` hits `WinError 10013`). For a tiny static reservation stub, the pipeline therefore builds nothing: it publishes the committed sdist straight from source. 
### Validation

- Stub sdist builds locally and passes `twine check`.
- Dry run (`publishToPyPI = false`) is green on the sealed OneBranch pool, confirming the pipeline compiles and the `ESRP Federated Creds (AME)` variable group authorizes.

### Follow-ups (not in this PR)

- Classify the pipeline **Production** in the product catalog (required for the Official template to run / publish).
- Run `publishToPyPI = true` from `main` to reserve the name; yank `0.0.0` once the real release ships.

[AB#47922](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47922)


David-Engel (commented):

## Summary

Adds a manual-only OneBranch **Official** pipeline that ESRP-publishes a committed `mssql-python-rs` `0.0.0` sdist to public PyPI to reserve the name, plus the stub sources. No Rust/JS/Python product code changes. I checked the pipeline against `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml`, `.pipeline/OneBranch/stages.yml`, `.pipeline/templates/build-python-wheels-template.yml`, and the closest reference — `microsoft/mssql-python` `OneBranchPipelines/official-release-pipeline.yml`. The `EsrpRelease@9` inputs (`connectedservicename`, `usemanagedidentity`, `Intent`, `ContentType: PyPI`, `ContentSource: Folder`, `Owners: $(owner)`, `Approvers: $(approver)`, `MainPublisher: ESRPRELPACMAN`) match that reference exactly, so the ESRP wiring looks right. The one thing I'd want fixed before this can publish is that nothing ties the committed binary artifact to the sources reviewers actually read.

## Verification

- Extracted `.pipeline/pypi-stub/dist/mssql_python_rs-0.0.0.tar.gz` in an isolated worktree and compared its `pyproject.toml`, `README.md`, and `mssql_python_rs/__init__.py` against the tracked files: identical apart from line endings (the archive was built with CRLF; the repo is `* text=auto eol=lf`). The archive contains only `PKG-INFO`, `README.md`, `pyproject.toml`, `setup.cfg`, `mssql_python_rs/__init__.py`, and `*.egg-info` metadata — nothing unexpected. sha256 `82d07f7f7ae5206a1c435e0d09c388a3dfe4a692235e227d0da0f72a6ccf1704`.
- No Rust code in the diff, so no `cargo bclippy` / `cargo btest` run. I did not execute the pipeline.
- Required CI is passing or pending; nothing failing that contradicts the PR description.

## Findings

### Blocking

**1. Nothing ties the published artifact to the reviewed sources** — `.pipeline/OneBranch/PyPIRelease.yml:94-101`, `:136`

`VerifyStub` only globs `mssql_python_rs-*.tar.gz` and asserts a count of 1. It never opens the archive, so it cannot detect that the tarball's version, metadata, or payload differ from `.pipeline/pypi-stub/pyproject.toml` and `mssql_python_rs/__init__.py`. `PublishToPyPI` then hands ESRP the whole folder (`FolderLocation: .../pypi-stub/dist`), not the file `VerifyStub` inspected. A `.tar.gz` is opaque in the PR diff, so a future commit can change what gets published to public PyPI under Microsoft's ESRP identity without that change being visible in review. I confirmed the current tarball does match the sources — the problem is that nothing keeps that true.

Two ways to close it:

- **Preferred — build in the pipeline.** The "sealed `type: windows` container has no Python" premise doesn't match this repo: `.pipeline/OneBranch/stages.yml:59-92` runs the `Windows_x64` job on `pool: type: windows` with the same `WindowsContainerImage`, and `build-python-wheels-template.yml:166-192` does `UsePythonVersion@0` followed by `pip install maturin wheel` there — including under `OfficialPythonWheelsBuild.yml`, which extends the same `v2/OneBranch.Official.CrossPlat.yml`. The mssql-python release pipeline solves the same problem with an explicit Python-capable pool (`pool: {type: windows, isCustom: true, name: Python-1ES-pool, demands: imageOverride -equals PYTHON-1ES-MMS2022}`). If a specific pool/agent failed for you, it's worth naming that failure in the header comment rather than the general claim, since the repo already contradicts it.
- **Otherwise — verify the committed artifact.** Have `VerifyStub` expand the tarball (`tar.exe` ships in ltsc2022), assert `PKG-INFO` `Name`/`Version` equal the values in `pyproject.toml` and `__init__.py.__version__`, assert the payload files match the tracked copies after newline normalization, compare sha256 against a committed `.sha256`, then copy only that one verified file into a fresh staging folder and point `FolderLocation` at it. That also removes the "whole folder gets published" exposure if a second artifact ever lands in `dist/`.

### Suggestions

**1. Reinforcing the branch gate raised in [the existing thread on `:110`](https://github.com/microsoft/mssql-rs/pull/516#discussion_r3947072054).** I agree with the mechanism and want to add context rather than restate it: the mssql-python reference pipeline has no `Build.SourceBranch` allow-list either, so this isn't a repo-convention break. But the exposure is higher here, because that pipeline publishes artifacts produced by a build pipeline run, while this one publishes a file committed on whatever ref the operator queues. A `Build.SourceBranch` condition on `PublishToPyPI` is cheap and composes with the verification above.

**2. The `0.0.1` text is baked into the committed artifact, not just the README** — related to [the README thread](https://github.com/microsoft/mssql-rs/pull/516#discussion_r3947072121). `PKG-INFO` inside `mssql_python_rs-0.0.0.tar.gz` carries the rendered README verbatim, so it currently says "This `0.0.1` release is a non-functional placeholder" and PyPI would render that on the project page. Editing `README.md` alone won't change it — the sdist has to be rebuilt and re-committed. This is exactly the drift the content check in the blocking finding would have caught.

### Nit

**1. No `LICENSE` in the sdist** — `.pipeline/pypi-stub/pyproject.toml:11`. Metadata declares MIT but the archive ships no license text (`SOURCES.txt` lists only README, pyproject, and `__init__.py`). Adding a `LICENSE` file next to `pyproject.toml` plus `license-files = ["LICENSE"]` puts the text on the PyPI page, which is the usual bar for a Microsoft-published distribution.

.pipeline/OneBranch/PyPIRelease.yml
```diff
@@ -0,0 +1,147 @@
...
+            Copy-Item $sdists[0].FullName -Destination "$(Build.ArtifactStagingDirectory)/" -Force
+          displayName: 'Verify committed stub sdist'
+
+      #####################################################################
+      # Job 2 (gated): publish the committed stub sdist to PyPI via ESRP. The
+      # repo is checked out, so ESRP reads the sdist straight from source — no
+      # Python, pip, build, or download needed on the sealed pool. Runs ONLY
+      # when publishToPyPI is true.
+      #####################################################################
+      - ${{ if eq(parameters.publishToPyPI, true) }}:
```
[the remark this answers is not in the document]
gargsaumya: Fixed in 53ab9b8f. The production publish job now has a runtime condition requiring Build.SourceBranch == refs/heads/main, in addition to the publishToPyPI parameter gate.

.pipeline/OneBranch/PyPIRelease.yml
```diff
@@ -0,0 +1,147 @@
...
+# version 0.0.0) to CREATE and RESERVE the PyPI project name and prove the ESRP  #
+# -> PyPI path end to end.                                                       #
+#                                                                               #
+# PHASE 2 (later): the build stage below is replaced to collect the real,       #
+# signed wheels from the Official Python Wheels Build and publish those. The     #
+# ESRP release stage stays as-is. The placeholder 0.0.0 version is yanked once   #
+# the first real version ships.                                                 #
+#                                                                               #
+# Manual trigger only. `publishToPyPI` defaults to false: a default run is a    #
+# dry run (build + `twine check`) with NO publish. Flip it to true only for     #
```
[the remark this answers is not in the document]
gargsaumya: Fixed in 53ab9b8f. The header now describes the dry run accurately: it verifies and stages the committed sdist; it no longer claims the pipeline builds the package or runs twine check.

.pipeline/OneBranch/PyPIRelease.yml
```diff
@@ -0,0 +1,148 @@
...
+            inputs:
+              connectedservicename: '$(ESRPConnectedServiceName)'
+              usemanagedidentity: true
+              keyvaultname: '$(AuthAKVName)'
+              signcertname: '$(AuthSignCertName)'
+              clientid: '$(EsrpClientId)'
+              Intent: 'PackageDistribution'
+              ContentType: 'PyPI'
+              ContentSource: 'Folder'
+              FolderLocation: '$(Build.SourcesDirectory)/.pipeline/pypi-stub/dist'
```
saurabh500: **Tie the published artifact to the reviewed sources, and constrain the release set.** `VerifyStub` only asserts one path matching `mssql_python_rs-*.tar.gz` and copies it to `$(Build.ArtifactStagingDirectory)` — a job-local dir this job never reads — while ESRP is handed the whole checked-out `dist/` folder here (`ContentSource: Folder`). So nothing opens the archive to check its version/metadata/payload against the tracked `pyproject.toml` / `__init__.py`, and the archive is opaque in the PR diff (GitHub shows it changed, not its contents). The current tarball does match the sources — I extracted it — the gap is that the pipeline doesn't keep that true across future commits.

Because `VerifyStub` and this job don't share a filesystem (`dependsOn` passes status/outputs, not files), do the check where the publish happens. Inside this job, right before `EsrpRelease@9`:

```powershell
$dist  = "$(Build.SourcesDirectory)/.pipeline/pypi-stub/dist"
$stage = "$(Agent.TempDirectory)/pypi-publish"
New-Item -ItemType Directory -Force -Path $stage | Out-Null
$sdist = @(Get-ChildItem $dist -Filter 'mssql_python_rs-*.tar.gz' -File)
if ($sdist.Count -ne 1) { throw "expected exactly one stub sdist, found $($sdist.Count)" }
# extract (tar.exe ships in ltsc2022); assert PKG-INFO Name/Version == pyproject.toml + __init__.__version__;
# assert archive members match the tracked files (newline-normalized), then:
Copy-Item $sdist[0].FullName $stage
```

then point `FolderLocation` at `$stage`. On ESRP folder semantics: `ContentType: PyPI` may enumerate only supported distributions, so a stray non-dist file might be ignored or fail validation rather than be published — but staging only the one verified file removes the ambiguity either way. (Same substance as the existing blocking finding in the on-behalf-of-@David-Engel review; filed as a suggestion here because the current artifact checks out.)

.pipeline/OneBranch/PyPIRelease.yml
```diff
@@ -0,0 +1,148 @@
...
+            Copy-Item $sdists[0].FullName -Destination "$(Build.ArtifactStagingDirectory)/" -Force
+          displayName: 'Verify committed stub sdist'
+
+      #####################################################################
+      # Job 2 (gated): publish the committed stub sdist to PyPI via ESRP. The
+      # repo is checked out, so ESRP reads the sdist straight from source — no
+      # Python, pip, build, or download needed on the sealed pool. Runs ONLY
+      # when publishToPyPI is true.
+      #####################################################################
+      - ${{ if eq(parameters.publishToPyPI, true) }}:
```
saurabh500: **A green dry run doesn't exercise the publish path.** This compile-time `${{ if }}` removes the entire `PublishToPyPI` job on a default run, so a green dry run proves the variable group is readable and the committed sdist is present — but not the `Build.SourceBranch == refs/heads/main` condition (line 114), the `EsrpRelease@9` step, ESRP auth / PyPI connectivity, or that this `releaseJob` even checks out the repo. That last point matters: the publish reads `$(Build.SourcesDirectory)/.pipeline/pypi-stub/dist`, which is empty if the release job doesn't check out sources — and nothing has run this job yet. Confirm the release job checks out the repo before relying on publish-from-source; folding verify+publish into one job (see the `FolderLocation` comment) also de-risks it. Minor related point: a publish queued from a non-`main` ref is silently *skipped* by the line-114 condition, not failed, and the stage can still look green because `VerifyStub` passed.

.pipeline/pypi-stub/pyproject.toml
```diff
@@ -0,0 +1,29 @@
...
+requires = ["setuptools>=68", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "mssql-python-rs"
+version = "0.0.0"
+description = "Reserved name for the Rust-based native provider (TDS core + ODBC driver) consumed by mssql-python. Placeholder release; not yet functional."
+readme = "README.md"
+requires-python = ">=3.10"
+license = { text = "MIT" }
```
saurabh500, line 11: **No LICENSE in the sdist.** Metadata declares MIT but the archive ships no license text (`SOURCES.txt` lists only README, pyproject, and `__init__.py`). Simplest fix: add a `LICENSE` file next to `pyproject.toml` and rebuild the sdist — setuptools picks it up via its default license-file glob, no config needed. (The `[project] license-files` key is PEP 639 and needs setuptools ≥77, above this project's `setuptools>=68`.) Puts the license text on the PyPI project page, the usual bar for a Microsoft-published distribution. Overlaps the earlier review's nit.

saurabh500 (commented):
> 🤖 **Unattended review sweep.** Generated by an automated sweep on behalf of @saurabh500 and **not checked by a human before posting**. Event `COMMENT`, not an approval — push back on anything that looks wrong.

## Verdict

Not an approval, but **no blocking findings** from this pass. I re-reviewed at HEAD (`53ab9b8f`) and confirmed the fix commit cleared the two earlier blockers: the publish job is now gated to `refs/heads/main` (line 114), and the committed sdist was rebuilt so its `PKG-INFO` and embedded README read `0.0.0` (I extracted `dist/mssql_python_rs-0.0.0.tar.gz`; the three payload files match the tracked sources after newline normalization). What remains are suggestions and a nit.

On severity: the earlier on-behalf-of-@David-Engel review flagged the artifact↔source provenance gap as **blocking**. Since I verified the current artifact matches the tracked sources, there is no present-tense defect, so I file it below as a strong suggestion — it becomes genuinely important once phase 2 reuses this pipeline to publish real wheels.

## Blocking

None from this pass.

## Suggestions

Two are inline — on `FolderLocation` and on the `publishToPyPI` gate. One more isn't tied to a diff line:

**PyPI name-reservation policy (PEP 541).** PyPI's [PEP 541 "Invalid projects"](https://peps.python.org/pep-0541/#invalid-projects) list includes *"project is name squatting (package has no functionality or is empty)"* — which is how this package describes itself. In practice this is unlikely to be actioned given Microsoft is a reachable owner with a genuine forthcoming project, but it means the reservation isn't guaranteed durable. Worth confirming the approach with the PyPI/ESRP release owner, or publishing a minimal *functional* prerelease instead of an explicitly non-functional placeholder.

## Nit

One inline, on `pyproject.toml`.

David-Engel (commented):

## Summary

Re-reviewed at HEAD `53ab9b8f`. The two earlier blockers are resolved: `PublishToPyPI` is gated on `Build.SourceBranch == refs/heads/main` (`.pipeline/OneBranch/PyPIRelease.yml:114`), and the committed sdist was rebuilt so `PKG-INFO` and the embedded README read `0.0.0`. I listed and streamed the archive members without extracting: it contains only `PKG-INFO`, `README.md`, `pyproject.toml`, `setup.cfg`, `mssql_python_rs/__init__.py`, and `*.egg-info` metadata, and the payload matches the tracked sources.

I'm not restating the open threads on `FolderLocation`/artifact provenance, dry-run coverage, or the missing `LICENSE` — I agree with them as filed. Everything below is new ground: the reserved name doesn't line up with anything this repo builds, and the repo's release documentation doesn't know this pipeline exists. **No blocking findings from this pass.**

## Verification

Read-only. The diff adds a production ESRP release pipeline, so per review policy I did not execute pipelines, builds, or the packaging toolchain; I read the YAML and inspected the committed archive with `tar -t` / `tar -xO` (no extraction to disk, no Python invoked). No Rust/JS changes, so no `cargo bclippy` / `cargo btest`. Required CI is green (`coverage-report` pending); nothing contradicts the PR description.

## Suggestions

**1. The reserved distribution name doesn't match anything this repo produces — worth settling before the name is locked in.**

`.pipeline/OneBranch/PyPIRelease.yml:10-13` says phase 2 "collect[s] the real, signed wheels from the Official Python Wheels Build and publish[es] those" through this same ESRP stage. Those wheels are `mssql_py_core-*.whl`: `mssql-py-core/pyproject.toml` declares `[project] name = "mssql-py-core"` with `[tool.maturin] module-name = "mssql_py_core"`, and that naming is baked into `docs/release-management.md:281-290`, `scripts/test_inject_odbc.py`, and `.pipeline/templates/private-link-smoke-template.yml:23`. `mssql-python-rs` appears nowhere in the repo outside this PR.

PyPI rejects an upload whose metadata `Name` doesn't match the target project, so phase 2 cannot publish today's wheels under the reserved name without first renaming `[project].name` in `mssql-py-core/pyproject.toml` (and deciding whether the import module follows — the stub reserves `mssql_python_rs`, the extension is `mssql_py_core`). Three coherent outcomes; this PR should pick one explicitly:

- rename the maturin distribution to `mssql-python-rs` in a follow-up, and say so in the phase-2 comment so the dependency is visible;
- reserve `mssql-py-core` instead (or in addition), if that's the name that ships; or
- confirm `mssql-python-rs` is intentionally a *different* distribution from `mssql-py-core` (e.g. the mssql-python-facing wrapper), in which case the phase-2 comment is misleading, because it points at the Official Python Wheels Build output.

Same divergence in the metadata: the stub declares `requires-python = ">=3.10"` while `mssql-py-core/pyproject.toml` declares `>=3.8`. Harmless on its own (`Requires-Python` is per-release), but it's a second signal that the stub metadata was written independently of the package it's reserving for. Reservations are cheap to make and awkward to unwind — Microsoft-owned squatted names can't simply be handed to a different project later.

## Nits

**1. `docs/release-management.md` isn't updated.** Its "Pipeline Files Reference" table (`:264-273`) enumerates every pipeline file in the repo, and its release-activity scenarios describe how wheels reach consumers. A new production publish path — the first one in this repo that pushes to a *public* feed rather than Azure Artifacts — should appear there, along with the two manual follow-ups from the PR description (Production classification in the product catalog, and yanking `0.0.0`). Otherwise the next person doing a release has no way to discover this pipeline exists or that it must eventually be re-pointed.

**2. No recorded recipe for regenerating the committed sdist.** The reviewed artifact is a checked-in binary that must be rebuilt by hand whenever the stub sources change — the `0.0.1` README drift caught earlier in this PR is exactly that failure mode. `README.md` can't hold the instructions since its contents render on the PyPI project page, so put the exact command (`python -m build --sdist`, plus a note to rebuild after *any* edit under `.pipeline/pypi-stub/`) in the `PyPIRelease.yml` header comment or a separate non-packaged note. Complements the pipeline-side verification proposed in the open `FolderLocation` thread rather than duplicating it: one keeps a bad artifact from shipping, the other keeps a good one easy to reproduce.

.pipeline/OneBranch/PyPIRelease.yml
gargsaumya: Fixed in a135f934. The release job now explicitly checks out self, extracts the single sdist, verifies PKG-INFO name/version against pyproject.toml and __version__, compares LICENSE/README/pyproject/__init__ payloads with the tracked sources (newline-normalized), copies only that verified archive to a clean staging directory, and points ESRP at that directory.

.pipeline/OneBranch/PyPIRelease.yml
gargsaumya: Fixed in a135f934 by folding dry-run verification and publishing into one always-running releaseJob with an explicit checkout. A dry run now exercises the release-job checkout plus full archive verification/staging. When publishToPyPI=true, a separate step fails loudly unless Build.SourceBranch is refs/heads/main, followed by ESRP. ESRP auth/PyPI connectivity remain intentionally unexercised until the real publish.

.pipeline/pypi-stub/pyproject.toml
gargsaumya: Fixed in a135f934. Added the MIT LICENSE beside pyproject.toml and rebuilt the committed sdist. The archive now includes LICENSE and passes twine check.

David-Engel (commented):

## Summary

Re-reviewed at HEAD `a135f934`. The hardening commit does what the open threads asked: verify and publish are folded into one `releaseJob` with an explicit `- checkout: self` (`.pipeline/OneBranch/PyPIRelease.yml:90`), the publish is gated by a step that *throws* unless `Build.SourceBranch` is `refs/heads/main` (`:92-97`), the archive is now opened and its `PKG-INFO` `Name`/`Version` plus four payload files are compared against the tracked sources (`:118-142`), only the one verified file is staged, and `FolderLocation` points at that staging directory (`:144`, `:161`) instead of the checked-out `dist/`. `LICENSE` is now in the archive (`License-File: LICENSE` in `PKG-INFO`). That closes the provenance gap that was blocking, and I'm not re-filing the still-open items from earlier passes (the `mssql-python-rs` vs `mssql-py-core` name divergence, PEP 541 durability, `docs/release-management.md`) — I agree with them as filed. **No blocking findings from this pass.** What's below is residual coverage in the new verification step.

## Verification

Extracted `.pipeline/pypi-stub/dist/mssql_python_rs-0.0.0.tar.gz` in the isolated review worktree and ran the pipeline's verification block verbatim against the tracked sources on PowerShell 7: it passes (`expectedName=mssql-python-rs`, `expectedVersion=0.0.0`, `initVersion=0.0.0`, archive metadata matches, all four payload files byte-identical after newline normalization). Archive members are exactly `PKG-INFO`, `README.md`, `LICENSE`, `pyproject.toml`, `setup.cfg`, `mssql_python_rs/__init__.py` and `*.egg-info` metadata — nothing unexpected. No Rust/JS/Python product code in the diff, so no `cargo bclippy` / `cargo btest`. I did not run the pipeline or ESRP.

## Suggestions

**1. The text PyPI actually renders isn't covered by the payload comparison** — `.pipeline/OneBranch/PyPIRelease.yml:137`

`$trackedFiles` compares the sdist's `README.md`, but PyPI renders the long description from `PKG-INFO`'s body, not from the `README.md` member. Those are separate copies inside the archive: `PKG-INFO` here carries the full README text after `Dynamic: license-file`. So a rebuilt-or-edited archive whose `README.md` matches the tracked file while its `PKG-INFO` description is stale would pass every current check and still publish the wrong text to the project page. That is precisely the drift that already happened once in this PR — the `0.0.1` string that reached the archive was in the rendered description. Cheap fix in the same step: split `PKG-INFO` at the first blank line after the headers and compare the remainder to the normalized tracked `README.md`.

**2. The archive's member set is unconstrained** — `.pipeline/OneBranch/PyPIRelease.yml:137-142`

The loop proves the four named files match, but nothing asserts the archive contains *only* those files plus expected metadata. An extra member would ship unreviewed even though all four comparisons pass — and the one that matters is `setup.py`: `setuptools.build_meta` (the backend declared in the verified `pyproject.toml`) executes `setup.py` when it is present, so an added member is code execution at install/build time, not just dead weight. Same for a second module dropped under `mssql_python_rs/`, which `packages = ["mssql_python_rs"]` would install. Since the whole point of this step is that the `.tar.gz` is opaque in a PR diff, enumerate and assert:

```powershell
$expectedMembers = @('PKG-INFO','README.md','LICENSE','pyproject.toml','setup.cfg','mssql_python_rs/__init__.py')
$members = @(tar.exe -tzf $sdists[0].FullName) |
  ForEach-Object { $_ -replace '^[^/]+/', '' } |
  Where-Object { $_ -and -not $_.EndsWith('/') -and $_ -notlike 'mssql_python_rs.egg-info/*' }
$unexpected = $members | Where-Object { $expectedMembers -notcontains $_ }
if ($unexpected) { throw "Unexpected sdist members: $($unexpected -join ', ')" }
```

**3. The "dry run is green" evidence predates this structure** — `.pipeline/OneBranch/PyPIRelease.yml:79-90`

The PR description's validation bullet describes a green dry run, but that run exercised the old two-job layout where verification lived in an ordinary build job. At HEAD the verification runs inside `templateContext: {type: releaseJob, isProduction: true}` with an explicit `- checkout: self`, and the header comment (`:52-53`) states the pipeline must be classified Production for the Official template to run. Three things a fresh dry run at HEAD would settle that the old one cannot: that a OneBranch release job honors `checkout: self` (the concern raised in the thread on the previous head, answered by adding the step rather than by a run), that `tar.exe` is present in the sealed `ltsc2022/vse2022` image, and whether an `isProduction` release job is even queueable before the catalog classification lands — if it isn't, the dry-run path is unusable until then, which is worth knowing before this is the only pre-publish check. Re-running with `publishToPyPI = false` at HEAD and updating the description costs one queue.

## Nit

**1. A dry run leaves nothing behind** — `.pipeline/OneBranch/PyPIRelease.yml:87`, `:144`

`ob_outputDirectory` is still declared as `$(Build.ArtifactStagingDirectory)`, but the restructure dropped the `Copy-Item ... $(Build.ArtifactStagingDirectory)` that the old `VerifyStub` job used to surface the sdist; the verified file now only lands in `$(Agent.TempDirectory)/pypi-publish`. Copying it to the output directory as well would give a dry run an auditable artifact of exactly what was verified, at no cost to the publish path.

Theekshna (commented):
Unattended hourly review sweep (first automated review of this PR from this login). Checked out `pr516-review` at `a135f934` (verified against `gh pr view --json headRefOid` immediately before diffing and again immediately before posting), diffed against merge-base `465c3cb0` with `origin/main`. No new findings beyond what's already filed and open (David-Engel's bot review at this exact head: 0 blocking / 3 suggestions / 1 nit; saurabh500's earlier PEP 541 durability suggestion) — I independently re-verified rather than re-filing.

Covered: `mssql-odbc` parity is N/A (diff touches only `.pipeline/OneBranch/PyPIRelease.yml` and `.pipeline/pypi-stub/**`, no ODBC surface). Extracted the committed `mssql_python_rs-0.0.0.tar.gz` with `tar.exe` and confirmed its 12 members are exactly what's expected (no stray `setup.py` or extra module), that `PKG-INFO`'s rendered body is byte-identical to the tracked `README.md` (no `0.0.1` staleness), and independently ran `twine check` against it — PASSED, matching the PR's own claim. Confirmed via `git log -p` that `a135f934` restructured the pipeline from two jobs into a single `releaseJob`/`isProduction: true` job, which is why David-Engel's Suggestion #3 (the PR description's "dry run is green" claim predates this structure) is correct and still open — not re-filing it. Confirmed `docs/release-management.md`'s "Pipeline Files Reference" table has no entry for this pipeline, matching David-Engel's nit. No AI slop in the stub package or pipeline comments — the header comments justify a non-obvious decision (why the sdist is committed instead of built in-pipeline) rather than restating code. PR description: no checklist to contradict, AB#47922 linked (taken at face value per policy).

This is an unattended run; findings were not checked by a human first.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

David-Engel (commented):

## Summary

Adds a manual-trigger OneBranch Official pipeline (`.pipeline/OneBranch/PyPIRelease.yml`) that publishes a committed `mssql-python-rs` `0.0.0` sdist to PyPI via ESRP, plus the stub package sources under `.pipeline/pypi-stub/`. No Rust or product code is touched. The rework in `a135f934` is a real improvement over the earlier two-job shape: verify and publish now live in one `releaseJob` with an explicit `checkout: self`, the branch gate fails loudly instead of silently skipping, `FolderLocation` points at a staged directory containing only the one verified archive, and the sdist now carries `LICENSE`.

## Verification

Read-only plus local probing in an isolated worktree; nothing was built or executed from the pipeline.

- Extracted the committed `dist/mssql_python_rs-0.0.0.tar.gz` and replicated the pipeline's checks by hand: `PKG-INFO` reports `Name: mssql-python-rs` / `Version: 0.0.0`, matching `pyproject.toml` and `__init__.__version__`, and all four newline-normalized payload comparisons (`LICENSE`, `README.md`, `pyproject.toml`, `mssql_python_rs/__init__.py`) match the tracked sources. The archive also has `License-File: LICENSE` and `Description-Content-Type: text/markdown`, so the PyPI page will render the README and show the license — the earlier license finding is genuinely closed.
- Confirmed `dist/` is not covered by `.gitignore`, so the archive is normally tracked rather than force-added.
- Read all six existing review threads (all resolved); the findings below are residual gaps, not restatements.
- `gh pr checks 516`: all required checks passing on `a135f934`.

## Findings

### Suggestion

**1. Archive verification is an allow-list, not a membership check — `.pipeline/OneBranch/PyPIRelease.yml`, "Verify and stage stub sdist"**

`$trackedFiles` covers four paths, but the archive actually ships eight: it also contains `setup.cfg`, `PKG-INFO`, and `mssql_python_rs.egg-info/{PKG-INFO,SOURCES.txt,dependency_links.txt,top_level.txt}`, none of which are compared against anything. Nothing asserts the *absence* of unexpected members. That matters specifically because the tarball is an opaque binary in the PR diff — GitHub shows "Bin 0 -> 2481 bytes", not contents — so the pipeline check is the only line of defense, and it currently answers "do the four files I named match?" rather than "is this archive exactly what the tracked sources produce?". A future commit could add, say, a `setup.py` to the tarball and every existing check would still pass; pip executes the build backend when installing from an sdist, so extra files in an sdist are not inert. Enumerating members (`tar.exe -tzf`) and comparing against an expected set closes the gap for a few extra lines:

```powershell
$expected = @(
  'LICENSE','PKG-INFO','README.md','pyproject.toml','setup.cfg',
  'mssql_python_rs/__init__.py',
  'mssql_python_rs.egg-info/PKG-INFO','mssql_python_rs.egg-info/SOURCES.txt',
  'mssql_python_rs.egg-info/dependency_links.txt','mssql_python_rs.egg-info/top_level.txt'
) | Sort-Object
$actual = @(tar.exe -tzf $sdists[0].FullName |
  ForEach-Object { $_ -replace '^mssql_python_rs-0\.0\.0/','' } |
  Where-Object { $_ -and -not $_.EndsWith('/') }) | Sort-Object
if (Compare-Object $expected $actual) { throw 'Archive members differ from the expected set.' }
```

**2. Four ESRP variables are used that nothing in this repo defines or exercises — lines 152–169**

`$(ESRPConnectedServiceName)`, `$(DomainTenantId)`, `$(owner)`, and `$(approver)` appear nowhere else in `.pipeline/`; the only group variables this repo demonstrably consumes are `EsrpClientId`, `AuthAKVName`, and `AuthSignCertName` (see `build-python-wheels-template.yml` and `build-odbc-windows-template.yml`). The lowercase `$(owner)` / `$(approver)` also break from the PascalCase convention of the others, which reads like a copy from a different pipeline. The risk is the failure mode: ADO expands an undefined macro to its own literal text, so a missing group variable ships `Owners: '$(owner)'` to ESRP instead of failing fast — and because `EsrpRelease@9` is behind the `${{ if eq(parameters.publishToPyPI, true) }}` compile-time gate, no dry run can surface it. This is a one-shot production publish; a cheap guard in the always-running verify step would catch it before it matters:

```powershell
foreach ($v in 'ESRPConnectedServiceName','AuthAKVName','AuthSignCertName','EsrpClientId','DomainTenantId','owner','approver') {
  $value = (Get-Item "env:$($v.ToUpper())" -ErrorAction SilentlyContinue).Value
  if (-not $value -or $value -match '^\$\(') { throw "ESRP variable '$v' is unset or unexpanded." }
}
```

(Adjust to however you prefer to read the group values; the point is asserting non-empty and not-a-literal-`$(...)` on every run, not just the publish run.)

**3. `0.0.0` is a final release, so `pip install mssql-python-rs` will resolve to the non-functional stub — `.pipeline/pypi-stub/pyproject.toml` line 6**

The PR plans to yank `0.0.0` only "once the real release ships", which leaves an open-ended window where the name is reserved *and* installable, handing anyone who guesses or typo-hits the name a package that does nothing. A PEP 440 pre-release version — `0.0.0.dev0` or `0.0.1a0` — reserves the PyPI project name identically but is skipped by pip's default resolution, so the window closes without depending on a follow-up yank. It would mean rebuilding the sdist and updating `__version__` to match.

### Nit

**4. Nothing re-checks the committed archive on PRs.** `PyPIRelease.yml` is `trigger: none` / `pr: none`, so the tracked-vs-archive consistency check only runs when someone manually queues the release pipeline. A PR that edits `pyproject.toml` or `__init__.py` without rebuilding `dist/` will pass all validation and only fail at release time. If the stub is expected to live for a while, a small step in PR validation over `.pipeline/pypi-stub/` would catch drift where it is cheap to fix.

**5. `license = { text = "MIT" }` uses the deprecated table form.** `Metadata-Version: 2.4` in the archive means this was built with setuptools ≥ 77, where the table form is deprecated and scheduled for removal in favor of the PEP 639 `license = "MIT"` + `license-files` spelling. Not a problem for the committed archive, but whoever regenerates the sdist on a newer setuptools may hit it. It would require raising `requires = ["setuptools>=68"]`.

saurabh500 (approved):

PR #517, opened 2026-09-07
Add character C type to temporal SQL type parameter conversions

Theekshna:
## Description

`SQLBindParameter` rejected a character C type bound to a temporal SQL type with `HYC00` "Parameter conversion not yet implemented". `mssql-python` binds a `datetime.time` as **isoformat text** against `SQL_TYPE_TIME`, not as a `SQL_C_TYPE_TIME` struct. Because `#pytest_all_data_types` has a `TIME` column, that one parameter failed the whole 13-parameter insert, the table stayed empty, and `fetchone`/`fetchmany`/`fetchall` correctly reported zero rows. The fetch path was never at fault.

### Conversions added

`SQL_C_CHAR` and `SQL_C_WCHAR` now reach all five temporal SQL types:

| C type | `TYPE_DATE` | `TYPE_TIME` | `SS_TIME2` | `TYPE_TIMESTAMP` | `SS_TIMESTAMPOFFSET` |
| --- | :---: | :---: | :---: | :---: | :---: |
| `SQL_C_CHAR` | new | new | new | new | new |
| `SQL_C_WCHAR` | new | new | new | new | new |
| `SQL_C_TYPE_DATE` | existing | - | - | - | - |
| `SQL_C_TYPE_TIME` / `SQL_C_SS_TIME2` | - | existing | existing | - | - |
| `SQL_C_TYPE_TIMESTAMP` | - | - | - | existing | - |
| `SQL_C_SS_TIMESTAMPOFFSET` | - | - | - | - | existing |

The scalar temporal rows are deliberately **not** widened: the matrix's contract is that a pairing accepted at bind time is one execute can convert, and nothing converts e.g. `SQL_C_TYPE_TIMESTAMP` to `SQL_TYPE_DATE` yet ([AB#47790](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47790)).

Accepted forms: `2024-05-20`, `12:34:56`, `12:34`, `12:34:56.123456` (up to 9 fractional digits), `2024-05-20 12:34:56.123`, and `2024-05-20 12:34:56+05:30`. A literal that omits the offset against `SQL_SS_TIMESTAMPOFFSET` takes `+00:00`, matching `CONVERT(datetimeoffset, '2024-05-20 12:34:56')`.

### Semantics, each verified against the msodbcsql source

| Behaviour | Reference |
| --- | --- |
| A valid offset folds to UTC and is cleared for every non-`datetimeoffset` target | `ParseDateTime`, `sqlccnvt.cpp:4820-4849` - guard is `fTypeOut != ...TIMESTAMPOFFSET`; `TOSERVER` calls `ConvertOffsetToUTC`, then zeroes the tz fields |
| The offset is bounded on its **total**, not per component | `IsValidTimezoneOffsetValue`, `dataconv.cpp:110-137` |
| A post-parse failure from a character source is `22018`, not `22007` | `CVT_CAST_ERROR` = `IDS_22_005`, `sqlcprot.h:956` |
| Date-only into a `time` target is `22018` | `sqlccnvt.cpp:4801-4806` |
| A timestamp into a `time` target drops the date silently | `SQL_TIME2_MAPPED` arm has no date check, `sqlccnvt.cpp:4482-4497` |

Folding runs before the target's own rules, so an offset can move a value onto another day, and can push a midnight literal off midnight and make a `date` target report `22008`.

### Measured divergences

| Case | msodbcsql | Here | Status |
| --- | --- | --- | --- |
| ISO `T` separator; unpadded fields (`1:01:01`); trailing empty fraction (`12:00:00.`) | `22018` | converts | Deliberate, pre-existing in the shared parser. [AB#47246](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47246) |
| Time-only literal into a timestamp target | fills the current local date | `22018` | Gap - needs a platform local-date helper. [AB#47247](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47247) |
| `YYYY/MM/DD` and `{d '...'}` / `{ts '...'}` escapes | accepted | `22018` | Gap, and the one direction where msodbcsql is *more* permissive. [AB#47246](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47246) |

The first two carry `SKIP_IF_COMPARING_MSODBCSQL()`, each backed by a source reading plus, for the `T` separator, a compare-leg measurement against retail 18.6.2.1.

### Implementation

- `conversion/datetime.rs` - `parse_datetime_literal` and `is_valid_timezone_offset` moved here from `fetch_convert.rs` / `param_convert.rs` and made `pub(crate)`; this module is already the shared date/time model, as `numeric.rs` is for `decimal_from_text`. One offset rule, one parser, both directions.
- `conversion/param_convert.rs` - `datetime_from_text` and `fold_offset_to_utc`, plus `(NarrowText | WideText, SqlFamily::DateTime)` arms. Parsing and folding happen at the layer msodbcsql folds at (`ParseDateTime`); target rules stay in the shared `convert_datetime_sql`. Only whole minutes move in the fold, so seconds and fraction are carried across rather than rebuilt from a tick count, which would round away a sub-tick fraction that must still be rejected as `22008`.
- `params/conversion_matrix.rs` - `TEMPORAL_SQL_TARGETS` added to the two character rows.
- `docs/parameters_plan.md` - records the above for the parameter direction.

No public API change, and no parameter-side behaviour change for any pairing that already bound.

One parameter-side diagnostic did move, for a pairing that did not bind before. A character C type bound with `SQL_LEN_DATA_AT_EXEC` against a temporal target was refused by `SQLBindParameter`; the bind is accepted now and the same `HYC00` is raised by `SQLExecute` instead, because the data-at-execution indicator is only read while the parameter list is built. The streamed path itself is unchanged - a temporal wire value is fixed-length, so it could never have been chunked. This is the same bind-to-execute move the integer, xml and variant targets made in P5, so it is recorded alongside them in `parameters_plan.md` and pinned by `cross_family_dae_is_rejected` and `CrossFamilyDataAtExecutionIsRejectedAtExecute`.

**The fetch direction did get two behaviour changes, both tightenings of the shared parser.** Neither shape can be rendered by SQL Server, so only a character column can carry one to that path.

1. The offset was bounded as `hh > 14 || mm > 59`, so `+14:30` converted; sharing `is_valid_timezone_offset` bounds the total instead, and a character column holding `2023-01-01 12:34:56+14:30` read into `SQL_C_TYPE_TIMESTAMP` now reports `22018`. Pinned by `an_out_of_range_offset_is_rejected_for_non_offset_targets`.
2. A separator with no time after it (`2024-05-20T`) split into a date and an empty time, and the empty half was filtered out before reaching `parse_time_literal`, so the literal parsed as date-only. It is now `22018`. Pinned by `malformed_literals_are_rejected` and `a_dangling_separator_is_22018_for_every_temporal_target`. Not a divergence in either direction: msodbcsql's length-dispatched grammar has no 11-character template, so it rejects this shape too.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47851

[AB#47851](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47851)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes - 1398/1398 in `mssqlodbc`. The workspace run also reports failures in `mssql-tds::transaction` and `bench/perf`, which panic with `SQL_PASSWORD environment variable not set`: live-server integration targets, and `mssql-tds` does not depend on `mssqlodbc`.
- [x] Public API changes are documented - none


mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -944,6 +951,18 @@ fn decimal_of(sql_type: SqlSmallInt, value: DecimalParts) -> SqlType {
...
+/// direction's literal parser so both directions accept exactly the same forms.
+///
+/// A literal the target cannot hold is left to [`convert_datetime_sql`], which
+/// applies the same component and fraction rules the struct path uses; only an
+/// unparsable literal is settled here, as `22018`.
+fn datetime_from_text(param: &BoundParam, text: AppText) -> Result<TypedValue, ParamBuildError> {
+    let parts = parse_datetime_literal(&text.into_string())
+        .ok_or(ParamBuildError::Value(ConvError::InvalidCharacterValue))?;
+    convert_datetime_sql(param.sql_type, param.decimal_digits, parts)
+}
```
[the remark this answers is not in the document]
Theekshna: Good catch, and confirmed - this is a real hole. `parse_datetime_literal` bounded the components independently (`hh > 14 || mm > 59`), so `+14:01`..`+14:59` parsed, while `+15:00` did not. `convert_datetime_sql` then applied `is_valid_timezone_offset` only under `sql_type == SQL_SS_TIMESTAMPOFFSET`, so the four newly reachable targets discarded the offset and converted the malformed literal silently. Two spellings of the same error also got two different answers, which is the part that makes it a defect rather than a judgement call.

Fixed in 4607a526, but at the parser rather than in `datetime_from_text` as suggested:

- The parser already *tried* to bound the offset; the check was just incomplete. Fixing it there means one rule in one place.
- Validating in `datetime_from_text` would leave `parse_datetime_literal` still handing an illegal offset to the fetch direction, which shares it. Fetch never validates the offset - it only renders it or writes it into `SQL_SS_TIMESTAMPOFFSET_STRUCT` - so a character column holding `+14:30` would still have produced an illegal offset there.

`is_valid_timezone_offset` moved from `param_convert.rs` to `datetime.rs` next to the parser and is now called by both paths, so the two cannot drift. It is the port of msodbcsql's `IsValidTimezoneOffsetValue` (`dataconv.cpp:118`), which bounds the total at +/-14:00 - so this now matches msodbcsql rather than merely being self-consistent.

Tests added:
- `an_offset_is_bounded_on_the_total` - `+/-14:00` accepted, `+14:01` / `+14:30` / `+14:59` / `-14:01` / `-14:30` rejected. My original `an_offset_is_signed_and_bounded` only probed `+15:00` and `+05:60`, which is exactly why this slipped through.
- `an_out_of_range_offset_is_rejected_for_every_temporal_target` - all five targets, asserting `22018` and that `+14:00` still converts.
- E2E `AnOutOfRangeOffsetLiteralIs22018ForEveryTarget`.

1389/1389 unit tests pass; `cargo bfmt` and `cargo bclippy` clean.

David-Engel (commented):
Automated review — no human checked these findings before posting. Reviewed at head `525c0dd`, merge base `6712b80`.

## Summary

Adds `SQL_C_CHAR` / `SQL_C_WCHAR` → the five temporal SQL types on the parameter path, by lifting `parse_datetime_literal` and `is_valid_timezone_offset` out of `fetch_convert.rs` / `param_convert.rs` into `conversion/datetime.rs` and routing character buffers through the existing `convert_datetime_sql`. The layering matches the `decimal_from_text` precedent already in this file, the divergences from msodbcsql are each cited to a source location, and the deliberate ones carry `SKIP_IF_COMPARING_MSODBCSQL()`. No blocking issues found.

**What I verified**

- `cargo nextest run -p mssqlodbc --lib` on head: 1395/1395 pass.
- Mutation check, because "has a test" is not the question. Flipping the fold direction in `fold_offset_to_utc` (`- offset_minutes` → `+`) *and* loosening the bound in `is_valid_timezone_offset` (`14 * 60` → `15 * 60`) fails 8 tests, including `a_valid_offset_is_folded_to_utc_for_non_offset_targets`, `folding_an_offset_can_cross_a_date_boundary`, `an_offset_is_bounded_on_the_total` and `an_out_of_range_offset_is_rejected_for_every_temporal_target`. The new tests do guard the new behaviour.
- Read the parser against multi-byte input: the offset tail is matched on `s.as_bytes()`, and `s[..s.len() - 6]` only runs after `tail[0]` is confirmed to be `+`/`-`, which cannot be a UTF-8 continuation byte, so the slice is always on a char boundary. `String::from_utf8_lossy` turns malformed narrow input into U+FFFD, which then fails the parse as `22018` rather than panicking.
- Adjudicated the Copilot thread on `param_convert.rs:975` (independent hour/minute bounds admitting `+14:01`..`+14:59`): the mechanism was real, and the fix in `4607a52` is correct and better placed than the one suggested — moving the check into the shared parser closes the fetch direction too, which validating in `datetime_from_text` would not have. Confirmed by mutation above.
- `gh pr checks 517`: no failures; `Test MacOS` and `coverage-report` still pending. Diff coverage 98%.

## Blocking

None.

## Suggestions

None.

## Nits

**1. The fetch direction got tighter too, and that isn't named in the description.** `mssql-odbc/src/conversion/fetch_convert.rs` previously bounded the offset as `hh > 14 || mm > 59`; the moved parser now applies `is_valid_timezone_offset`, so a *character column* holding `2023-01-01 12:34:56+14:30` read into `SQL_C_TYPE_TIMESTAMP` goes from converting to `22018`. That is the same correctness argument as the parameter side and no SQL Server `datetimeoffset` can render such a value, so this reads right — but "no behaviour change for any pairing that already bound" in the description is a parameter-side statement, and the fetch change is only implied by "One offset rule, one parser, both directions." Worth a sentence in the description, and there is currently no fetch-side test pinning the new rejection (the existing offset tests in `fetch_convert.rs` use `+05:30`).

**2. A character literal with no offset against `SQL_SS_TIMESTAMPOFFSET` silently becomes `+00:00`.** `parse_datetime_literal` leaves `has_tz` false with `tz_hour`/`tz_minute` at 0, `datetime_from_text` skips the fold, and `convert_datetime_sql` passes `is_valid_timezone_offset(0, 0)` and emits `offset: 0`. That matches `CAST('2024-05-20 12:34:56' AS datetimeoffset)`, so it is very likely the right answer — but every test on that target supplies an explicit offset (`a_character_datetimeoffset_literal_keeps_its_offset`, `CharDatetimeoffsetLiteralKeepsItsOffset`, `CharParamDeclaresTheTemporalParameterType`), so the defaulted case is unpinned. One line in `both_character_widths_reach_the_same_temporal_value` or its own case would fix that.

Theekshna:
Both nits addressed in `fccf8c1`.

> **1.** The fetch direction got tighter too, and that isn't named in the description. [...] there is currently no fetch-side test pinning the new rejection

Right, and the description sentence was a parameter-side claim doing duty as a global one. Added a paragraph naming the change explicitly: a character column holding `2023-01-01 12:34:56+14:30` read into `SQL_C_TYPE_TIMESTAMP` goes from converting to `22018`. Pinned by `an_out_of_range_offset_is_rejected_for_non_offset_targets` in `fetch_convert.rs`, sitting next to `offset_is_ignored_for_non_offset_targets` so the pair reads as one rule. Non-vacuous by construction: `+14:30` passes the old `hh > 14 || mm > 59` bound and fails only the total.

> **2.** A character literal with no offset against `SQL_SS_TIMESTAMPOFFSET` silently becomes `+00:00`. [...] the defaulted case is unpinned.

Pinned twice rather than once, because "very likely the right answer" is worth actually measuring:

- `a_character_datetimeoffset_literal_without_an_offset_defaults_to_utc` asserts `offset == 0` **and** that the wall clock is carried unfolded.
- `CharDatetimeoffsetLiteralKeepsItsOffset` gained a second case expecting `2024-05-20 12:34:56.0000000 +00:00`. It carries no `SKIP_IF_COMPARING_MSODBCSQL()`, so the compare leg adjudicates the default against retail rather than leaving it on a source reading.

`cargo nextest run -p mssqlodbc --lib`: 1397/1397. Checklist count updated.

David-Engel (commented):
Automated review — no human checked these findings before posting. Reviewed at head `fccf8c1`, merge base `6712b80`. Not an approval.

## Summary

Re-review at the new head. The two nits from the `525c0dd` pass are both closed, and the fixes are the right shape:

- **Fetch-side tightening is now named in the description and pinned.** `an_out_of_range_offset_is_rejected_for_non_offset_targets` in `fetch_convert.rs` sits next to `offset_is_ignored_for_non_offset_targets`, and `+14:30` is chosen deliberately — it passes the old `hh > 14 || mm > 59` bound and fails only the total, so the test cannot pass against the pre-move parser.
- **The defaulted `+00:00` is pinned twice.** `a_character_datetimeoffset_literal_without_an_offset_defaults_to_utc` asserts `offset == 0` *and* that the wall clock is carried unfolded (not just the offset), and the second case in `CharDatetimeoffsetLiteralKeepsItsOffset` carries no `SKIP_IF_COMPARING_MSODBCSQL()`, so the compare leg adjudicates the default against retail rather than leaving it on a source reading.

Everything else in the delta since `525c0dd` is test and comment only.

## Verification

- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` at `fccf8c1`: **1397/1397 pass**.
- **Mutation probe 1** — widened the offset bound in `is_valid_timezone_offset` from `14 * 60` to `15 * 60`: 7 failures, including the new fetch-side test and `an_out_of_range_offset_is_rejected_for_every_temporal_target`. The total-vs-component rule is genuinely guarded in both directions.
- **Mutation probe 2** — replaced `fold_offset_to_utc` with a bare `has_tz = false` (the discarded-offset reading): 4 failures across `a_valid_offset_is_folded_to_utc_for_non_offset_targets`, both date-boundary tests and the `date`-target overflow test. The fold is guarded, not merely exercised.
- Re-derived the surrounding facts rather than trusting the diff: temporal `ParameterType`s fall through `parameter_column_size_is_valid`'s `_ => return true`, so the newly reachable bind path is not gated on `ColumnSize`; `every_accepted_pairing_is_convertible` already covers the new matrix rows against `bound_param_to_value`; `dae_placeholder_type` still rejects a character-C-type-to-temporal data-at-execution binding at execute (`sql_family` mismatch → `HYC00`), which is the same state and the same timing as every other non-streamable pairing, so widening the matrix does not open a silent DAE path.
- Diff coverage 98% per the automated report; the eight missing lines are `panic!` arms inside test `match`es plus one unreachable `return None`.

## Notes, none blocking

- The `is_valid_timezone_offset` mixed-sign arms are unreachable from `parse_datetime_literal` — the parser applies one sign to both components — but they remain live for the `SQL_SS_TIMESTAMPOFFSET_STRUCT` path, which is where `IsValidTimezoneOffsetValue`'s rule actually earns its keep. Correct as written.
- The `p.scale` a literal carries is deliberately not consulted by `convert_datetime_sql`; `DecimalDigits` bounds the value on both the struct and character paths, and `MsodbcsqlOverPreciseTimeLiteralsAre22008` adjudicates that on both legs.
- The permissive-shape and divergence tables in `docs/parameters_plan.md` line up with `the_permissive_shapes_stay_accepted` and the `SKIP_IF_COMPARING_MSODBCSQL()` cases; the doc claims are checkable and check out.

No findings. This looks ready for human review — but it is **not** approved, and an automated pass is not a substitute for one.

mssql-odbc/src/conversion/datetime.rs
```diff
@@ -124,6 +124,165 @@ pub(crate) fn hms_from_ticks_100ns(ticks: u64) -> (u16, u16, u16, u32) {
...
+    };
+
+    if let Some(d) = date_str {
+        let (y, m, day) = parse_date_literal(d)?;
+        p.year = y;
+        p.month = m;
+        p.day = day;
+        p.has_date = true;
+    }
+    if let Some(t) = time_str.filter(|t| !t.is_empty()) {
```
saurabh500: **Nit:** a trailing `T` separator with no time after it is silently accepted as a date-only value. At the split below, `"2024-05-20T"` yields `("2024-05-20", "")`, and the empty time is then dropped by `.filter(|t| !t.is_empty())` (line 267), so the literal parses to a `date`-only `DateTimeParts` with `has_time = false`:

```
PROBE 2024-05-20T => DateTimeParts { year: 2024, month: 5, day: 20, has_date: true, has_time: false, .. }
```

The effect is target-dependent: `SQL_TYPE_TIME` / `SQL_SS_TIME2` already reject it (`!has_time` → `22018` at `param_convert.rs:1064`), but `date` and `timestamp` targets accept the malformed literal (timestamp fills midnight) instead of failing. `docs/parameters_plan.md` enumerates the `T`-*separator* between a real date and time (`CharTimestampAcceptsTheIsoSeparator`) and the trailing empty fraction (`12:00:00.`), but not a `T` with nothing after it — so I can't tell whether this shape is intended. If it isn't, an empty-RHS arm before the existing one rejects just this case without touching any accepted shape:

```rust
let (date_str, time_str) = match s.split_once(['T', ' ']) {
    Some((_, t)) if t.trim().is_empty() => return None, // dangling separator, no time
    Some((d, t)) => (Some(d), Some(t.trim())),
    None if s.contains(':') => (None, Some(s)),
    None => (Some(s), None),
};
```

(Surrounding whitespace like `"2024-05-20 "` is unaffected — the outer `trim()` at line 226 removes it before this match, and that form is already pinned as accepted. This nit is only about a literal `T`/space with an empty right-hand side.)

saurabh500 (commented):
🤖 _Automated, unattended PR-review sweep — no human vetted this before it posted. Treat it as a second opinion, not a gate._

**Verdict: Comment — no blocking findings.**

The change is well-scoped and unusually well-documented. The character C-type → temporal routing, the shared `parse_datetime_literal` / `is_valid_timezone_offset` move into `conversion/datetime.rs`, and the offset-fold semantics all line up, and the divergence catalog in `docs/parameters_plan.md` is thorough. All reported checks on `fccf8c1` were green when I looked (macOS, Windows, Linux, cross-repo mssql-python, Kerberos, CodeQL, coverage).

I re-ran the temporal conversion tests independently (`cargo nextest run -p mssqlodbc --lib conversion::param_convert conversion::datetime conversion::fetch_convert`, 216 passing). I also mutation-probed the `22007`→`22018` remap in `datetime_from_text`: with only that remap removed, the same command fails two tests — `a_date_only_literal_cannot_fill_a_time_target` and `a_time_only_literal_cannot_fill_a_timestamp` — so that remap is genuinely covered rather than incidentally green.

**Blocking:** None.

**Suggestion:** None.

**Nit:**
- `DateTimeParts::scale` (`datetime.rs:41`) is documented as `(0-7)` — true for a real source column — but the newly added `parse_datetime_literal` writes `p.scale = scale` (`datetime.rs:273`) where `parse_time_literal` yields up to `9` (`the_accepted_literal_shapes_parse` asserts `(123_456_789, 9)`). The parser-produced value is ignored on the parameter path (`convert_datetime_sql` uses `decimal_digits`, not `parts.scale`), though the field is still load-bearing for fetch-side rendering, so this is only a doc gap for parsed character literals — consider noting both meanings (a column is 0–7; a parsed literal can reach 9) or capping the stored value. (Reported here rather than inline because line 41 sits outside the diff hunks.)
- One inline Nit at `datetime.rs:267` — a trailing `T` with no time is accepted as date-only for date/timestamp targets.

For context: one prior automated sweep already reviewed this at the current head (`fccf8c1`) and found nothing blocking; this pass adds only the two nits above.

mssql-odbc/src/conversion/datetime.rs
Theekshna: Confirmed and fixed in `f170225` — thanks, this was a real hole and the target-dependence is exactly what made it easy to miss.

One correction to the suggested fix, though: the extra match arm isn't needed, because the `.filter(|t| !t.is_empty())` on line 267 *is* the whole defect. `parse_time_literal("")` already returns `None` — `s.split(':')` on an empty string yields one empty item, so the `minute_s` `?` bails. The filter was reaching in front of the parser and swallowing the failure. Deleting it lets the empty half reach `parse_time_literal`, which rejects it.

I verified that rather than assuming it. With your arm added but the filter left in place, the guard is dead: I stubbed the arm to `&& false` and all 1398 tests still passed, which is what pointed at the filter. Keeping both would have left a redundant branch shadowing the real cause.

So the change is one deleted line plus a comment at the split naming why the empty right-hand side has to reach the parser.

Pinned in two places, and mutation-checked against the original code — restoring the `.filter()` fails both:

- `malformed_literals_are_rejected` gains `"2024-05-20T"` and `"2024-05-20T   "` (the second because the outer `trim()` at line 226 collapses it to the first, which is worth pinning so nobody reintroduces the shape via whitespace handling).
- `a_dangling_separator_is_22018_for_every_temporal_target` loops all five targets, since your point was that `date` and `timestamp` were the ones actually accepting it.

On intent: this shape wasn't deliberate, and it isn't in the divergence tables because it isn't a divergence — msodbcsql's length-dispatched grammar has no 11-character template, so `2024-05-20T` fails there too. This moves us toward parity rather than away, so no `docs/parameters_plan.md` entry. `cargo nextest run -p mssqlodbc --lib`: 1398/1398.

> [!NOTE]

David-Engel (commented):

## Summary

Widens the parameter conversion matrix so `SQL_C_CHAR` / `SQL_C_WCHAR` reach all five temporal SQL types, hoists `parse_datetime_literal` and `is_valid_timezone_offset` into `conversion/datetime.rs` so both directions share one parser and one offset rule, and adds `datetime_from_text` + `fold_offset_to_utc` on the parameter side.

I read the full diff and traced the parser by hand against the literal shapes the new e2e table asserts (`10:12:59 +00:00`, `-11:01:21`, `1800-02-02 13:59:59.0000000000 +03:00`, `12:00:00:`, `12:00:00.....`, `32768:01:01`) — each lands on the rejection path the tests claim. `fold_offset_to_utc` is correct on the arithmetic I checked: whole-minute offsets mean seconds and fraction can never affect the day rollover, `..p` carries `second`/`fraction_ns`/`scale`/`has_date`/`has_time` across the fold so the sub-tick `22008` check still fires, and `days * 1440` cannot overflow `i64` at `MAX_DAYS_SINCE_0001`. The two prior review threads (offset bound, dangling `T` separator) are both genuinely fixed and pinned by tests in each direction.

The `22007` → `22018` remap in `datetime_from_text` is correctly scoped: `datetime_metadata` returns `InvalidDecimalDigits` (HY104), not `InvalidDateTime`, so a bad `DecimalDigits` is not swallowed by the remap.

Documentation quality here is unusually high — `parameters_plan.md` and the doc comments record the divergences, the reasoning, and the work items rather than leaving them to be rediscovered.

## Verification

- Read the complete diff against merge base `6712b80`, plus `sql_family`, `AppText::into_string`, `convert_datetime_sql`, `datetime_metadata`, `reject_fraction_past_scale`, `dae_placeholder_type`, and the `ParamBuildError` → SQLSTATE mapping.
- `cargo nextest run -p mssqlodbc --lib` in the isolated worktree/target: **1398 passed, 0 failed**.
- E2E (`param_conversions_test.cpp`) needs a live SQL Server and the msodbcsql compare leg; not run here — read only.
- Required CI: nothing failing at review time (Windows/ARM build, Kerberos, and the mssql-python cross-repo suite all green; remaining legs still pending).

## Findings

### Suggestion

**1. Data-at-execution bindings now fail later instead of at bind time.** `dae_placeholder_type` (`param_convert.rs:329`) rejects on family mismatch:

```rust
if sql_family(sql_type) != Some(c_family) {
    return Err(ParamBuildError::ConversionNotImplemented);
}
```

`SQL_C_CHAR`/`SQL_C_WCHAR` map to `SqlFamily::Character`, but the five temporal types map to `SqlFamily::DateTime`. Before this PR, an application that bound `SQL_C_CHAR` / `SQL_TYPE_TIME` with `SQL_LEN_DATA_AT_EXEC(n)` got `HYC00` from `SQLBindParameter`, because the matrix rejected the pairing outright. Now the matrix accepts it, so the bind succeeds and the same `HYC00` surfaces from `SQLExecute` instead. Same SQLSTATE, later call, and no test covers the path.

That's a small diagnostic-timing regression rather than a correctness one, and streaming a date literal is exotic — but it is newly reachable surface. Worth either a line in `parameters_plan.md` under the AB#47590 DAE gaps, or a case in the e2e/unit tests pinning that the pairing is `HYC00` at execute, so the next person touching `dae_placeholder_type` knows this row exists on purpose. (Not blocking; happy to be told it belongs in AB#47790 instead.)

### Nit

**2. `is_valid_timezone_offset`'s mixed-sign arms are unreachable from the parser.** `parse_datetime_literal` applies one `sign` to both `hh` and `mm`, so `tz_hour` and `tz_minute` can never disagree in sign on that path — the parser only ever exercises the `±840` total bound. The mixed-sign rules stay live for the `SQL_C_SS_TIMESTAMPOFFSET` struct path, which is why they still belong in the shared helper. A half-line on the function saying "the mixed-sign arms are reachable only from the struct path" would stop a future reader from assuming `an_offset_is_bounded_on_the_total` covers them.

Theekshna:
Both points from the latest review are addressed in 4fd312ee.

> the same `HYC00` now surfaces at `SQLExecute` rather than `SQLBindParameter`, and nothing pins it

Right, and it was already the established pattern here rather than anything new: the integer/xml/variant targets made the same bind-to-execute move in P5. So the temporal targets just join that set instead of getting their own story. Pinned in all three places that set is recorded:

- `cross_family_dae_is_rejected` (`param_convert.rs`) — added the five temporal targets across both char C types.
- `CrossFamilyDataAtExecutionIsRejectedAtExecute` (`param_conversions_test.cpp`) — the loop now runs `(c_type, sql_type)` pairs so `date` and `datetime2` are exercised at the real ODBC boundary, not just the integer target.
- `parameters_plan.md` — a line under the AB#47590 DAE gaps noting the temporal targets joined the set, and why the streamed path itself is unchanged (a temporal wire value is fixed-length, so it could never have been chunked; only the call that reports the refusal moved).

I mutation-probed the unit rows rather than trusting that they pass: relaxing the family guard to let `SqlFamily::DateTime` through makes `cross_family_dae_is_rejected` fail on exactly the new rows (`Ok(DaeStream { sql_type: VarcharMax, .. })` where an `Err` was expected), so they are load-bearing.

> a reader could assume `an_offset_is_bounded_on_the_total` covers the mixed-sign arms

Fair — it does not, and the arms are unreachable from `parse_datetime_literal` at all, since that applies one sign to both components. Added a note on `is_valid_timezone_offset` saying the mixed-sign rules are reachable only from the `SQL_C_SS_TIMESTAMPOFFSET` struct path, so nobody reads the literal tests as covering them.

Local run on this head: 1398/1398 in `mssqlodbc`, fmt and clippy clean.

On CI: the failure on `f170225` was not this branch. `Build Stage Build MacOS` never got an agent (still `queued` while the parent build reported failure), so the merged `CoberturaCoverage` artifact never published and `coverage-report` hit its 75-minute poll deadline. Every other leg passed, including the macOS *test* leg, and the preceding head `fccf8c1` was fully green on the same pipeline. I deliberately did not touch the poll deadline to make it go green — this push re-triggers the pipeline on real work instead.

David-Engel (commented):

## Summary

Re-review at head `4fd312e`, merge base `6712b80`. The delta since the last pass (`f170225`) is the one commit that closes the data-at-execution suggestion from that review: a doc paragraph in `parameters_plan.md`, the reachability note on `is_valid_timezone_offset`, five new rows in `cross_family_dae_is_rejected`, and the `CrossFamilyDataAtExecutionIsRejectedAtExecute` table gaining two temporal cases. No production-behaviour change in this commit. **No findings.**

All four prior review threads are now closed and each fix is pinned in the direction it applies to:

- Copilot's independent hour/minute offset bound → fixed at the parser (`is_valid_timezone_offset`), so the fetch direction is closed too.
- saurabh500's dangling `T` separator → fixed by deleting the `.filter(|t| !t.is_empty())` that was shadowing `parse_time_literal`'s own rejection, which is the more precise fix than the suggested extra match arm.
- The DAE bind-to-execute diagnostic move → recorded and pinned by this commit.
- The mixed-sign reachability note on `is_valid_timezone_offset` → added.

## Verification

- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` at `4fd312e` in an isolated worktree and target dir: **1398 passed, 0 failed**. (Note the crate's package name is `mssqlodbc`, not `mssql-odbc`.)
- **Mutation probe on the new pin**, because a test added to close a review comment is exactly the kind that can be vacuous. Relaxed the family gate in `dae_placeholder_type` to `sql_family(sql_type) != Some(c_family) && sql_family(sql_type) != Some(SqlFamily::DateTime)` — the five new rows then return `Ok(DaeStream { sql_type: VarcharMax, needs_transcode: false })` and `cross_family_dae_is_rejected` is the only failure in the 17-test `dae` selection. The pin guards the behaviour it names; reverted afterwards.
- Traced the newly reachable path end to end rather than trusting the diff: `bound_param_to_value` → `datetime_from_text` → `fold_offset_to_utc` → `convert_datetime_sql`, plus `sql_family`, `read_indicator`/`typed_null` (a NULL temporal parameter never reaches the parser), `reject_fraction_past_scale` and `datetime_metadata`.
- Hand-checked the msodbcsql-derived e2e rejection table against the parser: `10:12:59 +00:00` (offset with no date to anchor it), `-11:01:21` (non-digit hour), `1800-02-02 13:59:59.0000000000 +03:00` (10 fractional digits), `12:00:00:` (four colon-separated fields), `12:00:00.....`, `32768:01:01` — each lands on the rejection the test claims, and none of them lands there by accident.
- Required CI on this head: all green, including the cross-repo mssql-python suite on the mssql-odbc driver, macOS/Windows/Linux/ARM builds, Kerberos and `coverage-report`.

## Findings

### Blocking

None.

### Suggestions

None.

### Nits

**1. `DateTimeParts::scale`'s doc comment is still scoped to a column (already raised by @saurabh500, still open).** `datetime.rs:41` documents the field as "Declared fractional-seconds scale (0-7) of the source column", but `parse_datetime_literal` writes `parse_time_literal`'s digit count, which reaches 9 — `the_accepted_literal_shapes_parse` asserts `(123_456_789, 9)` directly.

Worth adding that this is benign, which the original nit left open: the only consumer is `format_datetime_parts` (`fetch_convert.rs:583-587`), and it clamps with `frac[..usize::from(p.scale).min(frac.len())]` over a 7-character string, so a scale of 8 or 9 renders as 7 rather than panicking on the slice. It is also unreachable in practice — `format_datetime_parts` serves a character C target, which never routes through `parse_datetime_literal`, and the parameter path reads `DecimalDigits` instead of `parts.scale`. So this is a doc gap only, and a one-line amendment ("a parsed literal can carry up to 9; renderers clamp to 7") closes it. Flagging only because it is the one open comment on the PR, not because it needs to block anything.

---

The description matches the implementation, including the parameter-side DAE diagnostic move and both fetch-side tightenings, and AB#47851 lines up with what shipped. Ready for human review — but this is not an approval.

Theekshna:
Addressed the nit in 9e30981.

> `DateTimeParts::scale`'s doc comment is still scoped to a column

Correct, and worth fixing rather than waving through. `parse_time_literal` bounds the fraction at `frac_digits.len() > 9`, not 7, and returns that count as the scale — `the_accepted_literal_shapes_parse` asserts `(123_456_789, 9)` outright. So the field's stated `(0-7)` range was simply false for anything the literal parser produced, and it named only one of the field's two sources.

Your benign analysis checks out: `format_datetime_parts` is the only reader, and its `frac[..usize::from(p.scale).min(frac.len())]` over the 7-character `format!("{hundred_ns:07}")` clamps rather than panicking. It is also unreachable today, since that renderer serves a character C target and the parameter path reads `DecimalDigits` instead of `parts.scale`.

That is exactly why I did not leave it as-is, though. A doc that understates a field's range is the kind of thing a later change reads as an invariant — `&frac[..scale]` without the clamp is a natural-looking edit that would panic on a 9-digit literal. The doc now names both sources, gives the real bound, and says outright not to treat the field as an index into a 7-digit string, so the clamp in `format_datetime_parts` reads as load-bearing instead of incidental.

Doc-only, no behaviour change. `cargo bfmt` and `cargo bclippy` clean and 1398/1398 in `mssqlodbc` on this head.

One correction on the framing: this was not an open thread. Both review threads on the PR (`PRRT_kwDOPLFXwM6f3qKw`, `PRRT_kwDOPLFXwM6gAt5G`) are resolved, and a paginated `reviewThreads` query returns no unresolved ones — the point survived only in a review body, which is presumably why it looked open.

Separately, the local toolchain problem I mentioned earlier was my own misdiagnosis, not a broken toolchain: `rust-toolchain.toml` pins channel `1.95`, which msrustup cannot resolve (its toolchains are all `ms-` prefixed), so every `cargo` alias failed before reaching a compiler. With `MSRUSTUP_TOOLCHAIN=ms-prod-1.95` set, `cargo bfmt` and `cargo bclippy` both pass on the pinned toolchain with no lint suppressions. `cargo btest` still cannot run locally, but for an unrelated and genuine reason: it is the only one of the three that needs codegen, and `crc32fast` 1.5.1 sets `--cfg stable_vpclmulqdq` for rustc >= 1.95 while the MS backend has no codegen for `llvm.x86.pclmulqdq.256/512` or 64-byte `simd_xor`. That is a dependency/toolchain issue with no bearing on this PR, so it is not being patched into this branch.

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot on behalf of @David-Engel. Unattended run; no human vetted these findings before posting. This is **not an approval** and does not satisfy the human review requirement. Push back on anything that looks off.

## Summary

Re-review at head `9e30981`, merge base `6712b80`. The delta since `4fd312e` is doc-only — `DateTimeParts::scale` now names both of the field's sources, gives the real bound, and says outright not to treat it as an index into a 7-digit string. That closes the last open nit, and the framing correction is right: `parse_time_literal` bounds the fraction at 9 digits, so the old `(0-7)` was simply false for anything the literal parser produced.

I re-read the complete diff rather than only the delta. The change holds up. `SQL_C_CHAR` / `SQL_C_WCHAR` reach all five temporal `ParameterType`s; `parse_datetime_literal` and `is_valid_timezone_offset` live in `conversion/datetime.rs` so the two directions cannot drift on the offset rule; and `fold_offset_to_utc` carries `second` / `fraction_ns` / `scale` across via `..p` rather than rebuilding them from a tick count, so a sub-tick fraction still reaches the `22008` check instead of being rounded away. `days * 1440` cannot overflow `i64` at `MAX_DAYS_SINCE_0001`, and the offset tail is matched on `s.as_bytes()` with the `str` slice taken only after `tail[0]` is confirmed ASCII, so no multi-byte payload can land mid-character.

## Verification

- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` at `9e30981`, isolated worktree and target dir: **1398 passed, 0 failed**.
- **Mutation probe on the fold direction.** Flipped `- offset_minutes` to `+ offset_minutes` in `fold_offset_to_utc`: 3 failures — `a_valid_offset_is_folded_to_utc_for_non_offset_targets`, `folding_an_offset_can_cross_a_date_boundary`, `an_out_of_range_offset_is_rejected_for_every_temporal_target`. Reverted.
- **Mutation probe on the `SQL_TYPE_DATE` non-zero-time arm.** Replaced the condition with `false`: 3 failures, two of them this PR's new tests (`a_character_literal_still_obeys_the_target_rules`, `folding_an_offset_can_make_a_date_target_overflow`). Reverted. This is also the evidence for the nit below.
- Hand-traced the parser against the msodbcsql-derived e2e rejection table rather than trusting the assertions: `10:12:59 +00:00` fails the offset-needs-a-date-and-a-time guard, `-11:01:21` fails the plain-digit check on the hour, `32768:01:01` overflows the `u16` hour parse, `1800-02-02 13:59:59.0000000000 +03:00` fails the 9-digit fraction bound, `12:00:00:` yields four colon fields, `12:00:00.....` fails the fraction digit check. None of them lands on the rejection by accident.
- Checked the shapes that *must* keep working: a bare `+05:30` with nothing before it collapses to an empty date and is rejected by `parse_date_literal`; `2024-05-20`'s internal hyphen is never read as an offset sign because `tail[3] != b':'`; the outer `trim()` is what makes `"2024-05-20T   "` fold onto the pinned `"2024-05-20T"` case.
- `gh pr checks 517`: nothing failing on this head — CodeQL, `check` and the duplicate-evaluation leg pass; the ADO build/test legs and `coverage-report` are still pending.
- E2E (`param_conversions_test.cpp`) needs a live SQL Server and the msodbcsql compare leg; read only, not run here.

## Findings

### Blocking

None.

### Suggestions

None.

### Nit

**1. `convert_datetime_sql`'s `SQL_TYPE_DATE` arm still declares the branch unreachable, and this PR is what makes it reachable.** `mssql-odbc/src/conversion/param_convert.rs:1054`:

```rust
// Unreachable through the API: the conversion matrix has no
// `SQL_C_TYPE_TIMESTAMP` -> `SQL_TYPE_DATE` row, so no binding can
// carry a time here yet (AB#47790). msodbcsql accepts the pairing.
    return Err(truncated);
}
```

The premise is still true — there is no `SQL_C_TYPE_TIMESTAMP` → `SQL_TYPE_DATE` row — but the conclusion drawn from it is not. `TEMPORAL_SQL_TARGETS` now puts `SQL_C_CHAR` / `SQL_C_WCHAR` → `SQL_TYPE_DATE` in the matrix, and a literal like `2024-05-20 12:00:00` carries a time straight into this branch. The PR's own tests land here: `a_character_literal_still_obeys_the_target_rules`, `folding_an_offset_can_make_a_date_target_overflow`, and the e2e `ACharLiteralStillObeysTheTargetRules` / `FoldingAnOffsetCanMakeADateTargetOverflow` — neither of which carries `SKIP_IF_COMPARING_MSODBCSQL()`, so the `22008` is adjudicated against retail 18.6.2.1 rather than left on a source reading. The `if false` probe above confirms the arm is load-bearing for two new unit tests, where before this PR only the direct-call test `a_timestamp_with_a_time_component_cannot_become_a_date` reached it.

The behaviour is right and well covered; only the comment is stale. It is worth fixing rather than waving through for the same reason the `scale` doc was: an "unreachable through the API" note is exactly what makes a later reader comfortable deleting a guard, and this one now sits in front of a check that two application-visible paths depend on. Two lines naming the character path — and that the state is now measured on the compare leg — would close it, and the AB#47790 reference can stay, since the `SQL_C_TYPE_TIMESTAMP` row it tracks is still absent.

---

Everything else checks out. `parameters_plan.md`'s divergence and permissive-shape tables line up with `the_permissive_shapes_stay_accepted` and the `SKIP_IF_COMPARING_MSODBCSQL()` cases, the DAE bind-to-execute move is recorded and pinned in all three places, and the description matches what shipped. Ready for human review — but this is not an approval.

Theekshna:
Fixed in 73cd90b.

> `convert_datetime_sql`'s `SQL_TYPE_DATE` arm still declares the branch unreachable, and this PR is what makes it reachable

Correct, and a better catch than the `scale` one. The comment's premise and its conclusion had come apart: there is still no `SQL_C_TYPE_TIMESTAMP` -> `SQL_TYPE_DATE` row, but "no binding can carry a time here" stopped following from that the moment `TEMPORAL_SQL_TARGETS` put the two character C types in front of `SQL_TYPE_DATE`. `2024-05-20 12:00:00` now walks straight into that guard.

This is the worse of the two stale comments, for the reason you name. A field doc that understates a range invites a bad edit; an explicit "unreachable through the API" sitting directly above a guard invites deleting the guard outright — and the failure mode there is silent truncation of a time to a date, not a panic. Your `if false` probe is the proof it is load-bearing: two of this PR's own unit tests fail without it.

The comment now leads with the character path and AB#47851, keeps the AB#47790 note for the still-absent `SQL_C_TYPE_TIMESTAMP` row, and records that the `22008` is adjudicated on the compare leg rather than asserted only here. I checked that last claim rather than repeating it from your review: `ACharLiteralStillObeysTheTargetRules` and `FoldingAnOffsetCanMakeADateTargetOverflow` both exist in `param_conversions_test.cpp` and neither carries `SKIP_IF_COMPARING_MSODBCSQL()`, so retail 18.6.2.1 does adjudicate them.

Comment-only, no behaviour change. `cargo bfmt` and `cargo bclippy` clean on the pinned toolchain, 1398/1398 in `mssqlodbc`.

Worth noting for anyone reading later: this is the second stale-comment nit in a row, and both were about code whose *behaviour* is right and well covered. Neither would have been caught by tests, since a test cannot fail on a comment. The rest of the diff was re-read for the same class of problem while making this change, and the remaining reachability claims hold - `datetime.rs`'s `!has_date && !has_time` guard is genuinely unreachable, and `is_valid_timezone_offset`'s mixed-sign arms are reachable only from the struct path, which is now stated where a reader will hit it.

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot on behalf of @David-Engel. Unattended run; no human vetted these findings before posting. This is **not an approval** and does not satisfy the human review requirement. Push back on anything that looks off.

## Summary

Re-review at head `73cd90b`, merge base `6712b80`. The delta since `9e30981` is the single comment-only commit that closes the last open nit: the `SQL_TYPE_DATE` arm in `convert_datetime_sql` no longer claims to be unreachable through the API, names the character path that now reaches it, points at the two compare-leg tests that measure the `22008`, and keeps the AB#47790 reference for the still-absent `SQL_C_TYPE_TIMESTAMP` row. That is exactly the correction that was asked for, and no production behaviour moved.

I re-read the full diff rather than only the delta. All five prior review threads are closed and each fix is pinned in the direction it applies to:

- Copilot's independent hour/minute offset bound → fixed at the shared parser, so the fetch direction is closed too.
- @saurabh500's dangling `T` separator → fixed by deleting the `.filter(|t| !t.is_empty())` that was shadowing `parse_time_literal`'s own rejection.
- The DAE bind-to-execute diagnostic move → recorded in `parameters_plan.md` and pinned in `cross_family_dae_is_rejected` / `CrossFamilyDataAtExecutionIsRejectedAtExecute`.
- The `is_valid_timezone_offset` mixed-sign reachability note → added.
- The `DateTimeParts::scale` doc and the `SQL_TYPE_DATE` reachability comment → both corrected.

**No blocking findings.**

## Verification

- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` at `73cd90b` in an isolated worktree and target dir: **1398 passed, 0 failed**.
- **Mutation probe on the fold guard**, which no earlier pass had probed. Relaxed `if parts.has_tz && param.sql_type != SQL_SS_TIMESTAMPOFFSET` to `if parts.has_tz`, so a `datetimeoffset` target folds and loses its offset: exactly one failure, `a_character_datetimeoffset_literal_keeps_its_offset`. The guard is genuinely pinned, not incidentally green. Reverted. (`both_character_widths_reach_the_same_temporal_value` cannot catch this one — both widths fold identically — so that single test is carrying it.)
- Traced the parser by hand rather than trusting the assertions. The offset tail is matched on `s.as_bytes()`, and `s[..s.len() - 6]` runs only after `tail[0]` is confirmed to be `+`/`-`, which cannot be a UTF-8 continuation byte, so the slice is always on a char boundary; `String::from_utf8_lossy` turns malformed narrow input into U+FFFD, which then fails the parse as `22018` rather than panicking. A lowercase `t` separator (`2024-05-20t12:34:56`) and a `Z` suffix both land on the non-digit check rather than sneaking through. A bare `+05:30` collapses to an empty date and is rejected by `parse_date_literal`.
- Re-derived the bind-time surface: temporal `ParameterType`s fall through `parameter_column_size_is_valid`'s `_ => return true` (`api/type_rules.rs:139`), so the newly reachable rows are not gated on `ColumnSize`; `datetime_metadata` returns `InvalidDecimalDigits` (HY104), which the `22007`→`22018` remap in `datetime_from_text` correctly does not swallow.
- Hand-checked the msodbcsql-derived e2e rejection table against the parser: `01:01:01.0000000000` (10 fractional digits), `24:00:00`, `32768:01:01` (`u16` overflow), `1e8:01:01`, `12:00:00:` (four colon fields), `10:12:59 +00:00` (offset with no date to anchor it) each land on the rejection the test claims.
- `gh pr checks 517`: all required checks green on this head, including the cross-repo mssql-python suite on the mssql-odbc driver, all four build legs, Kerberos, CodeQL and `coverage-report`.
- E2E (`param_conversions_test.cpp`) needs a live SQL Server and the msodbcsql compare leg; read only, not run here.

## Findings

### Blocking

None.

### Suggestions

None.

### Nit

**1. The one state in this PR that is neither cited to msodbcsql nor measured on the compare leg is the fold that leaves the representable range.** `fold_offset_to_utc` (`param_convert.rs:70`) answers `22018` when `utc_days` falls outside `0..=MAX_DAYS_SINCE_0001`:

```rust
let utc_days = total_minutes.div_euclid(1440);
if !(0..=MAX_DAYS_SINCE_0001).contains(&utc_days) {
    return Err(invalid); // ConvError::InvalidCharacterValue -> 22018
}
```

`folding_an_offset_can_cross_a_date_boundary` asserts that for `0001-01-01 00:00:00+05:30`, and the `SQL_SS_TIMESTAMPOFFSET` arm of `convert_datetime_sql` has the same shape at line 1109. Everything else in this change is either pinned to a source location (`ConvertOffsetToUTC`, `IsValidTimezoneOffsetValue`, `CVT_CAST_ERROR` = `IDS_22_005`) or adjudicated against retail 18.6.2.1 by an e2e case without `SKIP_IF_COMPARING_MSODBCSQL()` — this one is neither. `FoldingAnOffsetCanCrossADateBoundary` covers only the successful crossing, so the underflow never reaches the compare leg.

There is also an internal argument the other way. The PR's own rule is "`22008` passes through: a dropped component is an overflow, not a parse failure" — and a valid literal pushed off the end of the representable range by folding reads more like an overflow than a failed cast. The literal parsed fine; the target range is what it lost.

I am **not** claiming `22018` is wrong — the "character source ⇒ failed cast" rule makes it defensible, and the struct path's `22007` remaps to it consistently. The gap is that this is the one state left on an assertion rather than a measurement. An e2e case binding `0001-01-01 00:00:00+05:30` (and the `9999-12-31 23:59:59-05:30` mirror) against `SQL_TYPE_TIMESTAMP` without the skip would settle it on the compare leg for the cost of a few lines; if retail disagrees, that is a one-line change plus a `parameters_plan.md` row rather than something to discover later.

---

Everything else checks out. The divergence and permissive-shape tables in `docs/parameters_plan.md` line up with `the_permissive_shapes_stay_accepted` and the `SKIP_IF_COMPARING_MSODBCSQL()` cases, `ATemporalTextParamBindsAlongsideOtherParameters` reproduces the actual mssql-python failure shape rather than only the per-parameter one, and the description matches what shipped. Ready for human review — but this is not an approval.

Theekshna:
> the `22018` returned when a fold pushes a value outside the representable range ... rests on a unit assertion only

Acted on this, and probing it turned up something worse than an unmeasured state: the guard was **untested outright**. Deleting its upper arm left all 1398 unit tests green.

The reason is that for every date-bearing target the folded year gets rejected a second time downstream by `days_since_0001_from_civil`, with the same `22018`. So an assertion bound to `SQL_TYPE_TIMESTAMP` cannot distinguish the guard from that downstream check — it passes either way. `SQL_TYPE_TIME` discards the date, which makes it the only target where the fold's own guard is the sole thing rejecting the value.

`a2f843d1`:

- `folding_an_offset_can_cross_a_date_boundary` now asserts both range exits (`0001-01-01 00:00:00+05:30`, `9999-12-31 23:59:59-05:30`) against `SQL_TYPE_TIME` as well as `SQL_TYPE_TIMESTAMP`. Stubbing the guard to `if false` now fails the test; before this commit it did not.
- `FoldingAnOffsetCanCrossADateBoundary` carries both cases against `SQL_TYPE_TIMESTAMP` with **no** `SKIP_IF_COMPARING_MSODBCSQL()`, so retail 18.6.2.1 adjudicates the SQLSTATE rather than the driver agreeing with its own unit test.

Flagging the trade deliberately: that is an unskipped compare-leg assertion added to a currently-green PR. If retail answers `22008` here the leg goes red before merge — which is the outcome worth having, since your point that discovering it post-merge is worse is the right one. I'd then flip the state and record the divergence in `parameters_plan.md` rather than re-skip the case.

I did **not** extend the compare leg to the `SQL_TYPE_TIME` rows. Retail's behaviour for a time target whose fold leaves the range is unverified in the reference source, and pinning it live would be guessing.

`cargo bfmt` and `cargo bclippy` clean; 1398/1398 unit tests pass.

mssql-odbc/src/conversion/datetime.rs
```diff
@@ -124,6 +129,174 @@ pub(crate) fn hms_from_ticks_100ns(ticks: u64) -> (u16, u16, u16, u32) {
     )
 }
 
+/// Parses `YYYY-MM-DD`.
+fn parse_date_literal(s: &str) -> Option<(i16, u16, u16)> {
```
saurabh500: It is hard to understand what the Tuples return. Can we make this return a named struct instead? 
Similar suggestion for parse_time_literal(s: &str) -> Option<(u16, u16, u16, u32, u8)>

Theekshna: Agreed — done in `7492dd32`. I took it wider than the two you named, since `civil_from_days_since_0001` and `hms_from_ticks_100ns` had the same problem and are the inverse/decoder partners of these:

- `CivilDate { year, month, day }` — returned by both `civil_from_days_since_0001` and `parse_date_literal`.
- `TimeOfDay { hour, minute, second, fraction_ns }` — returned by `hms_from_ticks_100ns`.
- `ParsedTime { time: TimeOfDay, scale }` — returned by `parse_time_literal`.

The one judgement call: `scale` sits on `ParsedTime` rather than on `TimeOfDay`. It describes how many fractional digits the *literal* wrote, not anything about the instant, so folding it into `TimeOfDay` would have forced `hms_from_ticks_100ns` to invent a value it has no basis for.

Pure refactor, no behaviour change: `cargo bfmt` and `cargo bclippy` clean, 1398/1398 unit tests pass unmodified apart from two assertions that destructured the old tuples.

> [!NOTE]

David-Engel (commented):
> **Automated review** — generated by GitHub Copilot on behalf of @David-Engel. Unattended run; no human vetted these findings before posting. This is **not an approval** and does not satisfy the human review requirement. Push back on anything that looks off.

## Summary

Re-review at head `7492dd3`, merge base `6712b80`. Two commits since the last pass (`73cd90b`), and neither moves production behaviour:

- `a2f843d` closes my last nit by pinning the fold's range guard at **both** ends and, critically, on `SQL_TYPE_TIME` — the one target where the guard itself is what rejects the value. It also takes the underflow and overflow cases to the compare leg without `SKIP_IF_COMPARING_MSODBCSQL()`, so the `22018` is adjudicated against retail rather than against this driver's own unit test.
- `7492dd3` is @saurabh500's tuple-to-named-struct refactor, taken wider than the two functions he named: `CivilDate`, `TimeOfDay` and `ParsedTime` now cover `civil_from_days_since_0001` and `hms_from_ticks_100ns` as well as the two literal parsers. Keeping `scale` on `ParsedTime` rather than on `TimeOfDay` is the right call — it describes the literal's digit count, and `hms_from_ticks_100ns` has no basis to invent one.

**No findings.** All six prior review threads are closed, and each fix is pinned in the direction it applies to.

## Verification

- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` at `a2f843d` and again at `7492dd3`, isolated worktree and target dir: **1398 passed, 0 failed** on both. `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` and `cargo fmt -- --check` clean at head.
- **Mutation probe on the new range-guard pin.** Neutered the guard in `fold_offset_to_utc` (`if false && !(0..=MAX_DAYS_SINCE_0001).contains(&utc_days)`): exactly one failure, and it is the *new* `SQL_TYPE_TIME` iteration at `param_convert.rs:2136` — `9999-12-31 23:59:59-05:30` came back `Ok(Time(SqlTime { time_nanoseconds: 197_990_000_000, .. }))`, i.e. 05:29:59 on a date that does not exist. The `SQL_TYPE_TIMESTAMP` iteration still passed, because `days_since_0001_from_civil` rejects the folded year downstream with the same `22018`. That confirms the claim in the commit: before `a2f843d` the guard was not merely unmeasured on the compare leg, it was untested outright. Reverted.
- **Mutation probe on the refactor**, since a mechanical rename of same-typed fields is exactly where a silent swap hides. Swapped `minute: t.minute` / `second: t.second` in `time_parts` (`fetch_convert.rs`): 3 failures — `datetime2_to_timestamp_struct`, `time_to_ss_time2_struct`, `datetime2_into_time_target_reports_fraction_truncation`. The field mapping is guarded by the existing fetch tests, not just by the compiler's type check. Reverted.
- Read the complete diff against the merge base again, not only the delta. The refactor is behaviour-preserving at every call site; the local `date` in `date_parts` shadows the `&SqlDate` argument only after its last use.
- Re-derived the surrounding surface rather than trusting the diff: output and input-output bindings are still refused at `SQLBindParameter` (`bind_param.rs:258`), so widening the matrix opens no output-parameter path that would need a temporal-to-character write-back; `every_accepted_pairing_is_convertible` walks the new rows and would catch a matrix row the converter cannot serve; the ODBC 2.x concise `SQL_DATE`/`SQL_TIME`/`SQL_TIMESTAMP` spellings are not supported parameter types at all (`set_desc_field.rs:1012`), so adding only the 3.x ids leaves no half-open row behind.
- `gh pr checks 517`: nothing failing at this head. CodeQL, `check`, `Analyze` and the duplicate-evaluation leg pass; the ADO build/test legs and `coverage-report` re-queued after the push and are still pending.
- E2E (`param_conversions_test.cpp`) needs a live SQL Server and the msodbcsql compare leg; read only, not run here.

## Findings

### Blocking

None.

### Suggestions

None.

### Nits

None.

---

The divergence and permissive-shape tables in `docs/parameters_plan.md` still line up with `the_permissive_shapes_stay_accepted` and the `SKIP_IF_COMPARING_MSODBCSQL()` cases, the DAE bind-to-execute move stays recorded and pinned in all three places, and the description matches what shipped. Ready for human review — but this is not an approval.

saurabh500 (approved):

Issue #470, opened 2026-09-02
mssql-odbc: SQLBindCol/SQLBindParameter leave stale ARD/APD length/precision/scale behind on a differently-shaped rebind


Vahid-b:
## Context

Follow-up from [PR #436](https://github.com/microsoft/mssql-rs/pull/436) ([AB#47437](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47437) — descriptor records as the binding/metadata source of truth), flagged by an automated code review's non-blocking question.

## Problem

`ColumnBinding::write_to_record` (ARD, `SQLBindCol`) and the APD half of `BoundParam::write_to_records` (`SQLBindParameter`) only write `SQL_DESC_CONCISE_TYPE`, `SQL_DESC_DATETIME_INTERVAL_CODE`, `SQL_DESC_DATA_PTR`, `SQL_DESC_OCTET_LENGTH`, `SQL_DESC_INDICATOR_PTR`, and `SQL_DESC_OCTET_LENGTH_PTR`. Neither touches `SQL_DESC_LENGTH`, `SQL_DESC_PRECISION`, or `SQL_DESC_SCALE` on that record.

Sequence that reproduces the gap:
1. `SQLBindCol`/`SQLBindParameter` binds ordinal N with a C type whose precision/scale get set to some value.
2. A later `SQLBindCol`/`SQLBindParameter` call rebinds the *same* ordinal to a different C type.
3. `SQLGetDescField`/`SQLGetDescRecW` on `SQL_DESC_LENGTH`/`SQL_DESC_PRECISION`/`SQL_DESC_SCALE` for that record still reports the *previous* binding's value, not a value appropriate to the new C type.

This is metadata-introspection only — execution reads parameter type/size from the IPD (which *is* fully reset on every `SQLBindParameter`; `write_to_records` sets `precision`/`length`/`scale` unconditionally there) and reads column data via the ARD's `concise_type`/`data_ptr`/lengths (also unaffected). No query result or parameter value is wrong; only a direct `SQLGetDescField`/`SQLGetDescRecW` read of those three specific APD/ARD fields after a differently-shaped rebind can observe stale data.

## Verified against msodbcsql

[...]

PR #521, opened 2026-09-08
Support SQL_C_NUMERIC input parameters

Theekshna:
## Description

- Support `SQL_C_NUMERIC` input parameters using APD precision/scale and msodbcsql-compatible conversion semantics.
- Match retail diagnostics for prepared, direct, and data-at-execution paths, including no numeric fractional-truncation diagnostic once execution enters DAE.
- Reset APD numeric defaults and binding state on descriptor type changes so stale pointers and metadata cannot be reused.
- Add focused unit tests and retail-parity e2e coverage for boundaries, descriptor handling, diagnostics, rebinding, and DAE.

## Related Issues

Fixes - https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47946

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes (attempted; unrelated SQL-backed `mssql-tds` tests failed across the local suite after retries)

## Validation

- `cargo nextest run -p mssqlodbc`: 1,420 passed
- `cargo bfmt`: passed
- `cargo bclippy`: passed
- Affected C++ e2e targets: built successfully
- msodbcsql 18 results from build 173584 confirmed the corrected expectations: bare `(38, 0)` binding returns `12`, and numeric truncation produces no diagnostic when execution enters DAE
- Live Rust-driver e2e rerun is delegated to CI because the local SQL Server credentials fail authentication


mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -925,11 +1001,20 @@ fn decimal_from_text(param: &BoundParam, text: AppText) -> Result<TypedValue, Pa
...
         // back to `OutOfRange` here would report 22003 where every smaller
         // literal of the same shape reports 22001.
-        match 10i128.checked_pow(source_scale - target_scale) {
-            Some(divisor) if mantissa % divisor == 0 => mantissa / divisor,
-            Some(_) => return Err(ParamBuildError::StringTruncation),
+        let exponent = u32::try_from(source_scale - target_scale)
+            .map_err(|_| ParamBuildError::Value(ConvError::OutOfRange))?;
+        match 10i128.checked_pow(exponent) {
+            Some(divisor) => {
+                if mantissa % divisor != 0 {
```
David-Engel, line 1042: This arm changes behaviour for the pre-existing `SQL_C_CHAR` → `decimal` path, not just the new numeric one. On `main` a non-zero remainder returned `StringTruncation` here, before the `magnitude >= 10^precision` check below; now the truncated value reaches that check, so a value that overflows the target precision *and* drops a non-zero digit reports `22003` where it used to report `22001`.

Measured at this head via the existing `convert_decimal` helper:

```
convert_decimal(SQL_DECIMAL, 3, 1, "1234.55") -> Err(Value(OutOfRange))   // was StringTruncation on main
convert_decimal(SQL_DECIMAL, 5, 1, "1.55")    -> Err(StringTruncation)    // unchanged
```

Nothing in the description flags this, and no test covers the overlap. Worth confirming which state retail reports when both conditions hold; if it is `22001`, checking `outcome == ConvOk::Truncated` before the precision bound inside `decimal_from_text` restores the old answer without affecting the numeric path.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -911,10 +929,68 @@ fn decimal_from_text(param: &BoundParam, text: AppText) -> Result<TypedValue, Pa
...
+    let magnitude = u128::from_le_bytes(source.val);
+    // msodbcsql's FastDescribeRPCParam copies a matching non-NULL
+    // SQL_NUMERIC_STRUCT whole, including its wire precision and scale
+    // (`sqlcmisc.cpp:7014`).
+    if usize::try_from(param.app_precision) == Ok(param.column_size)
+        && param.app_scale == param.decimal_digits
+    {
+        let metadata = RpcTypeMetadata {
+            precision: Some(source.precision),
+            scale: Some(source.scale as u8),
```
David-Engel, line 970: This path writes the application's raw struct bytes into the NUMERICN type preamble, bypassing the `1..=38` precision / `0 <= scale <= precision` bounds that `decimal_metadata` enforces on every other route. `source.scale` is `i8`, so a negative embedded scale becomes `255` — the new `a_numeric_with_matching_descriptors_keeps_the_struct_bytes` test asserts `scale: Some(u8::MAX)` — and a zero-initialised struct yields precision `0`. `SqlType::serialize` writes both as raw bytes (`sqltypes.rs:774`), which is a protocol-level violation rather than the clean server-side `42000` that `NumericStructMatchingMetadataKeepsEmbeddedMetadata` pins for `precision = 1`.

It is also easier to reach than it looks: `SQLBindParameter` leaves APD precision/scale at `0` and sets `ipd_record.precision` from `ColumnSize`, so `SQLBindParameter(..., SQL_C_NUMERIC, SQL_DECIMAL, 0, 0, &num, 0, NULL)` with no `SQLSetDescField` call satisfies this condition — a binding `decimal_metadata` would otherwise reject as `InvalidParameterSize(0)`.

Suggest validating the embedded precision/scale here (or falling back to the APD-derived metadata when they are out of range), with e2e cases for `ColumnSize = 0` and a negative embedded scale to record retail's actual answer.

David-Engel (commented):
> _Automated, unattended review sweep — no human vetted these findings before posting. This is **not** an approval and does not satisfy the human review requirement._

**Verdict: COMMENT.** Two findings, neither of them a restatement of the three Copilot bot threads already on this PR (DAE truncation warning dropped on the `NeedData` path, negative IPD scale now accepted through the shared `set_scale` helper, stale APD precision/scale after a rebind). Those three are real and are left to the existing threads.

Reviewed `604b15c4` against `origin/main` in a dedicated worktree, after reading all three prior-discussion channels. The core of the change — reading `SQL_NUMERIC_STRUCT` unaligned, rescaling from the APD scale, the `10^precision` digit-count bound, negative-zero normalisation — checks out, and the retail-parity e2e cases are the right way to pin the msodbcsql-specific parts.

**What I ran:** `cargo nextest run -p mssqlodbc --lib` against a temporary probe test in `param_convert.rs`'s test module (reverted afterwards; the worktree is clean) to settle finding 1 empirically rather than by reading. CI is not green at this head yet: `Build Stage Build Linux ARM` is red and several ADO stages are still pending, so nothing here is a merge-readiness statement.

---

## Suggestions

**1. `decimal_from_scaled` changes the SQLSTATE of the pre-existing `SQL_C_CHAR` → `decimal` path (`param_convert.rs:1006`).**

On `main`, a non-zero dropped digit returned `StringTruncation` *before* the `magnitude >= 10^precision` check ran. Now the truncated value falls through to that check, so a value that both overflows the target precision and drops a non-zero digit reports `22003` instead of `22001`.

Measured at this head with a throwaway test calling the existing `convert_decimal` helper:

```
convert_decimal(SQL_DECIMAL, 3, 1, "1234.55") -> Err(Value(OutOfRange))    // 22003; main returns StringTruncation (22001)
convert_decimal(SQL_DECIMAL, 3, 1, "1234.50") -> Err(Value(OutOfRange))    // unchanged
convert_decimal(SQL_DECIMAL, 5, 1, "1.55")    -> Err(StringTruncation)     // unchanged
```

This isn't mentioned in the PR description and no existing test covers the overlap, so it looks incidental to the refactor rather than intended. `decimal_from_text`'s own doc comment cites `sqlccnvt.cpp:7823` / `sqlcfunc.cpp:3348` for the `22001` mapping; the question that comment doesn't answer is which state retail reports when *both* conditions hold. Worth confirming against retail and pinning it with a test either way — if `22001` is correct, checking `outcome == Truncated` before the precision bound in `decimal_from_text` restores it without disturbing the numeric path.

**2. The matching-descriptor fast path forwards unvalidated struct precision/scale into the TDS type preamble (`param_convert.rs:948-960`).**

Every other route into `decimal`/`numeric` goes through `decimal_metadata`, which enforces precision ∈ `1..=38` and `0 <= scale <= precision` and otherwise reports `InvalidParameterSize` / `InvalidDecimalDigits`. The fast path takes `source.precision` and `source.scale as u8` straight from the application's bytes, and those bytes are never validated anywhere:

- `source.scale` is `i8`. A negative embedded scale becomes `255` — the new `a_numeric_with_matching_descriptors_keeps_the_struct_bytes` test asserts exactly that (`scale: Some(u8::MAX)`), and `SqlType::serialize`'s NUMERICN preamble writes it as a raw byte (`sqltypes.rs:774`).
- `source.precision` of `0` (a zero-initialised struct) writes precision `0`.

Both are outside TDS's legal `1..=38` range, which is a different failure class from the e2e-verified `precision = 1` case: that one is a well-formed preamble the server rejects cleanly with `42000`, whereas an out-of-range precision/scale byte is a protocol violation and can cost the connection rather than producing a diagnostic. `NumericStructMatchingMetadataKeepsEmbeddedMetadata` doesn't cover either shape.

Related, and the reason this is easier to hit than it looks: `SQLBindParameter` leaves APD precision/scale at `0` (`bind_param.rs:281`, and `write_to_records` doesn't touch them), while `ipd_record.precision` is set from `ColumnSize` (`bound_param.rs:123`). So a plain `SQLBindParameter(..., SQL_C_NUMERIC, SQL_DECIMAL, 0, 0, &num, 0, NULL)` with no `SQLSetDescField` call satisfies `app_precision == column_size && app_scale == decimal_digits` and enters the fast path — a binding that for every other C type is rejected by `decimal_metadata` as `InvalidParameterSize(0)`.

Suggest either validating the struct's precision/scale on the fast path (`HY104`/`HY105`, or `22003`, whichever retail reports) or falling back to the APD-derived metadata when they're out of range, plus an e2e case for `ColumnSize = 0` and one for a negative embedded scale to record what retail actually does.

---

## Nits

None.

mssql-odbc/src/api/exec_common.rs
```diff
@@ -116,6 +117,8 @@ fn unwind_dae_inner(
 /// Parks the streaming client on the statement so `SQLParamData` / `SQLPutData`
 /// can drive the sequence, and enters the ODBC "Need Data" state. The DBC keeps
 /// `active_stmt` set, so the connection stays busy for the duration.
+/// msodbcsql's `SQLParamData` clears prior statement errors on entry
+/// (`sqlccmd.cpp:6818`), including pre-stream conversion warnings.
```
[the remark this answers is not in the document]
Theekshna: Confirmed as a genuine gap and fixed in c2eb0bb6. Verified against msodbcsql's `AddRPCUserParameters` (sqlccmd.cpp, ~9807-9930): the per-parameter RPC loop posts a truncated non-DAE parameter's `01S07` via `PostSQLError2` inline (the `CHKRC_SERR_GOTO` macro only traps hard `SQL_ERROR`, not `SQL_SUCCESS_WITH_INFO`) and keeps scanning until it hits the streamed parameter that makes the call return `SQL_NEED_DATA` — so the diagnostic genuinely is observable via `SQLGetDiagRec` immediately after the initial call, before `SQLParamData` is ever invoked.

`park_dae_client` now takes a `fractional_truncated: bool` and posts `WARN_FRACTIONAL_TRUNCATION` before returning `SQL_NEED_DATA`. Added a unit test (`park_dae_client_posts_numeric_fractional_truncation_before_param_data`) and updated both e2e tests in `execute_test.cpp` to assert `01S07` right after the initial `SQLExecute`/`SQLExecDirectW` call, in addition to the existing post-`SQLParamData` empty-diag assertion.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -530,7 +530,7 @@ pub(super) fn set_scale(
         .record(record_number)
         .map(|r| (r.concise_type, r.precision));
     if let Some((SQL_C_NUMERIC, precision)) = record_info
-        && (scale < 0 || scale > precision)
+        && scale > precision
```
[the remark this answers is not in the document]
Theekshna: Confirmed and fixed in c2eb0bb6. `SQL_C_NUMERIC` and `SQL_NUMERIC` do share the same integer value (2), and msodbcsql's IPD path (`CheckSqlPrecScale<FALSE>`, sqlcdesc.cpp:11511-11536) casts scale to `UWORD*` before comparing to precision — any negative scale is always "too large" there, returning `IDS_S1_094` (HY094) unconditionally for a 3.x app. That's a separate, stricter rule from the AD-only `CheckADDescRecConsistency` (sqlcdesc.cpp:11380-11394, upper-bound-only) this code was modeling.

`set_scale` now gates the negative-scale allowance on `kind.is_application()`, so an IPD `SQL_NUMERIC` record rejects a negative scale while AD records keep the existing behavior. `set_data_ptr` needed no change — confirmed `SQL_DESC_DATA_PTR` is writable only for AD kinds (`classify_field`'s `is_ad` gate), so it's never reachable for IPD. Added a unit test (`set_desc_field_ipd_numeric_scale_may_not_be_negative`) covering the IPD rejection alongside a still-valid scale.

> [!NOTE]

mssql-odbc/src/params/bound_param.rs
```diff
@@ -199,6 +201,8 @@ impl BoundParam {
             sql_type,
             column_size,
             decimal_digits,
+            app_precision: apd_record.precision,
+            app_scale: apd_record.scale,
```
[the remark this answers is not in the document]
Theekshna: This turned out to be more severe than the rebind framing here suggests, and is fixed in c2eb0bb6.

Checked msodbcsql's actual default: `SQLBindParameter`'s non-TVP path calls `SetADRecBP` (sqlcdesc.cpp:3145), which unconditionally does `ZeroMemory(lpbindinfo, ...); SetTypeDefaults((DESCRECBASE*)lpbindinfo, fCType);` (sqlcdesc.cpp:2883) on *every* call. `SetTypeDefaults`'s `SQL_NUMERIC`/`SQL_C_NUMERIC` case (sqlcdesc.cpp:12344) sets `cbPrecision = SQL_PREC_NUMERIC` (38), `ibScale = 0` — not `(0, 0)`.

`write_to_records` here never touched the APD's precision/scale at all, so `app_precision`/`app_scale` stayed at `DescRecord::default_for`'s `(0, 0)` seed — not just on rebind, but on the *first* bind too. Since `decimal_from_numeric`'s fast/slow decision compares `app_precision`/`app_scale` against the IPD's `column_size`/`decimal_digits`, this meant: a bare `SQLBindParameter(..., SQL_C_NUMERIC, SQL_DECIMAL, 38, 0, ...)` with no `SQLSetDescFieldW` call at all wrongly took the slow (rescale) path where msodbcsql fast-paths (since msodbcsql's real default `(38,0)` matches that target, `(0,0)` doesn't) — and once in that slow path, a struct with a nonzero embedded scale gets silently reinterpreted as scale `0`, corrupting the value. A later rebind at the same ordinal could independently pick up a scale left by a differently-shaped prior binding.

Fixed by resetting the APD's precision/scale to `(SQL_PREC_NUMERIC, 0)` for `SQL_C_NUMERIC` in `write_to_records`, on every call. Added a unit test (`write_to_records_resets_numeric_apd_precision_and_scale_to_msodbcsql_default`) plus two live e2e regression tests (`NumericStructWithoutDescriptorFieldWritesUsesTheStructsOwnScale`, `NumericRebindDoesNotInheritAPreviousBindsStaleApdScale`) that fail without the fix and pass with it.

> [!NOTE]

mssql-odbc/src/api/exec_common.rs
```diff
@@ -134,6 +146,9 @@ pub(super) fn park_dae_client(
         return SQL_ERROR;
     };
     stmt_state.dae = Some(DaeState::new(client, prepared, orphaned, dae_params));
+    if fractional_truncated {
+        post_diag(&mut stmt_state, WARN_FRACTIONAL_TRUNCATION);
```
saurabh500, line 143: **Suggestion — 01S07 fractional-truncation is dropped when a DAE parameter precedes a truncating numeric parameter.**

`build_named_params` converts every non-DAE parameter up front and folds all truncations into one `fractional_truncated` bool (it OR's `outcome == ConvOk::Truncated` per parameter, regardless of ordinal). `park_dae_client` posts `01S07` here on the `SQL_NEED_DATA` return, and the flag is not stored in `DaeState`. The first `SQLParamData` calls `free_errors` (`param_data.rs`), wiping that record, and completion runs plain `finish_execute` — so it is never re-posted.

That is fine for the tested ordering (`SELECT CONVERT(VARCHAR(32), ?) + ':' + ?`, param 1 numeric, param 2 streamed): the warning is genuinely pre-stream, observable at `SQLExecute`, gone after `SQLParamData` — matching msodbcsql. But msodbcsql's RPC loop is *lazy*: `AddRPCUserParameters` walks parameters in ordinal order via `ProcessRPCData` and `break`s at the first DAE parameter, resuming later with `AddRPCUserParameters(lpstmt, iParamStart)` (`sqlccmd.cpp`). So for the reverse ordering — DAE at ordinal 1, truncating `SQL_C_NUMERIC` at ordinal 2 — msodbcsql posts nothing at `SQLExecute` (it has not reached param 2), then posts `01S07` while resuming past the DAE parameter, surfacing it as `SQL_SUCCESS_WITH_INFO` on the final `SQLParamData`.

This driver instead posts `01S07` at `SQLExecute` (premature for that ordering), loses it after the first `SQLParamData`, and returns `SQL_SUCCESS` at completion — no `01S07` where msodbcsql returns `SQL_SUCCESS_WITH_INFO`.

Concrete case, not currently covered (both e2e tests bind the numeric before the DAE parameter):
- Param 1: `SQL_DATA_AT_EXEC` streamed varchar
- Param 2: `SQL_C_NUMERIC` `1.55` into a scale-1 target (drops a fractional digit)
- Expected (msodbcsql): `SQLExecute` → `SQL_NEED_DATA`, **no** 01S07; after streaming param 1, final `SQLParamData` → `SQL_SUCCESS_WITH_INFO`, `01S07`.

To match, the truncation needs to be tied to ordinal position rather than a single pre-park bool: truncations on parameters before the first DAE parameter belong to the initiating `SQLExecute`; a truncation on a parameter after a DAE parameter belongs to the `SQLParamData` transition that advances past it (carry it in `DaeState`, post it when that parameter is finalized). A DAE → truncating-numeric → DAE case is worth a test too.

I could not run this on a live server (no SQL Server in this environment; the crate's `nextest` is also blocked by the pre-existing crc32fast AVX-512 issue you noted), so I am filing it as a Suggestion rather than Blocking — but the mechanism is verified against both this diff and the msodbcsql source. Confidence ~7/10 on the observable divergence.

saurabh500 (commented):
🤖 _Posted by an unattended automated review sweep, on behalf of @saurabh500. Event is **COMMENT** only — no approval or change request is implied._

**Verdict:** Close. The two descriptor fixes in `c2eb0bb6` are correct and match the msodbcsql contract. The 01S07 fix is right for the ordering the new tests cover (a truncating numeric parameter *before* the DAE parameter), but it looks like it drops and mis-times the warning for the reverse ordering — a DAE parameter *before* a truncating `SQL_C_NUMERIC` parameter. One inline suggestion; nothing blocking.

**Blocking:** none.

**Suggestion:** one, inline in `exec_common.rs` — the 01S07 fractional-truncation warning is lost when a truncating numeric parameter follows a DAE parameter in ordinal order.

**Nit:** none.

_Adjudication of prior discussion (not re-filing):_
- I checked `c2eb0bb6` against the local msodbcsql source rather than the ODBC spec: `SetTypeDefaults` sets `SQL_NUMERIC`/`SQL_DECIMAL` to `(SQL_PREC_NUMERIC, 0)`, and `CheckSqlPrecScale` compares the scale as `UWORD`, so a negative IPD scale returns HY094 (`IDS_S1_094`) for a 3.x app. Both descriptor fixes land where msodbcsql lands.
- @David-Engel's two `param_convert.rs` threads (the `22001`→`22003` SQLSTATE change on the `SQL_C_CHAR`→decimal path, and the fast-path forwarding of the struct's embedded precision/scale) are still open at this head — `param_convert.rs` is untouched by `c2eb0bb6`. Not duplicating them.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Confirmed as a genuine regression — thanks for catching this. On `main` (`09b44d87`), `decimal_from_text`'s scale-down branch returns `Err(StringTruncation)` immediately on a nonzero rescale remainder, before ever reaching the magnitude/precision check. The shared `decimal_from_scaled` helper this PR introduced ran the magnitude check unconditionally afterward, so `convert_decimal(SQL_DECIMAL, 3, 1, "1234.55")` now returned `22003` instead of `22001`.

Fixed in `90fadb36` by splitting the helper into `rescale_mantissa` (rescale only) and `decimal_from_magnitude` (the precision/magnitude check). `decimal_from_text` now returns `StringTruncation` immediately on a truncated rescale, matching main's precedence, before calling `decimal_from_magnitude`; `decimal_from_numeric` still runs both unconditionally, preserving its existing tested behavior (a numeric struct's precision/scale bytes aren't a "dropped fraction" input in the same sense). Added a regression test (`a_dropped_fraction_is_22001_even_when_the_result_also_overflows`) covering both the truncation-wins case and the overflow-without-truncation case.

> [!NOTE]

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Verified this is intentional parity with msodbcsql, not a divergence. In retail's `sqlcfunc.cpp` (~3163-3172), the `SQL_C_NUMERIC` binding path only takes the "slow"/validating branch when `(fSqlType != SQL_NUMERIC && fSqlType != SQL_DECIMAL) || lpparam->cbColDef != lpbindinfo->cbPrecision || lpparam->ibScale != lpbindinfo->ibScale)` — i.e. when the APD precision/scale don't match the IPD's declared values. When they *do* match (the fast-path condition our code mirrors), retail does nothing further: `FastDescribeRPCParam` (`sqlcmisc.cpp` ~6975-7030, `case SQLNUMERICN/SQLDECIMALN`) does an unconditional `WriteToExtBuffer` of the caller's raw `SQL_NUMERIC_STRUCT` bytes straight to the wire, with no validation of precision/scale. Our `decimal_from_numeric` fast path (forwarding `source.precision`/`source.scale` when `app_precision == column_size && app_scale == decimal_digits`) is a faithful port of that exact behavior — the SQL Server engine is retail's last line of defense against malformed bytes here too, same as ours. No code change needed; resolving.

> [!NOTE]

mssql-odbc/src/api/exec_common.rs
Theekshna: Confirmed as a genuine bug — good catch. Our `build_named_params` does a single eager ordinal pass over all parameters (unlike msodbcsql's lazy break-and-resume `AddRPCUserParameters`), so it correctly *detects* truncation both before and after the first DAE parameter in one pass, but was posting all of it immediately via `fractional_truncated`. Since `SQLParamData` clears prior statement diagnostics on entry (`sqlccmd.cpp:6818`, already noted in an existing comment), a warning for a parameter ordinally *after* the first DAE parameter got wiped before the app's completing `SQLParamData` call ever ran, silently downgrading `01S07` to plain success.

Fixed in `90fadb36`: truncation is now split into `fractional_truncated` (before the first DAE param — posted immediately, unchanged) and a new `trailing_truncated` (at/after the first DAE param — parked on `DaeState` and posted by the completing `SQLParamData` call instead, via the same `finish_execute_with_param_warning` path already used for the non-DAE case). Added a unit test for the detection/segmentation logic (`build_named_params_truncation_after_first_dae_param_is_deferred`) and an e2e regression test (`NumericTruncationAfterDataAtExecutionIsReportedAtCompletion`) covering the reverse ordering end-to-end.

Note: this fix only tracks a before/after split relative to the *first* DAE parameter, not per-DAE-segment tracking for statements with multiple DAE parameters interspersed with truncating ones — that would require mirroring `AddRPCUserParameters`' segment-by-segment resumption more precisely. I judged that out of scope for this fix given it correctly handles the reported case; flagging it here in case it's worth a follow-up work item.

> [!NOTE]

David-Engel (commented):

## Summary

Reviewed `90fadb36` against the merge base with `origin/main` (`09b44d87`) in a dedicated worktree, after reading all three prior-discussion channels. The five threads open at earlier heads (DAE `01S07` dropped on the `NeedData` path, negative IPD scale, stale APD numeric metadata, the `22001`/`22003` precedence regression on the `SQL_C_CHAR` → decimal path, and the ordinal-relative deferral of a trailing truncation) are all answered at this head; I am not re-filing any of them.

The core of the change holds up: `SQL_NUMERIC_STRUCT` is read unaligned and whole, the APD scale drives the rescale, the `10^precision` digit-count bound is applied to the rescaled magnitude rather than to the mantissa width, negative zero normalises positive, and the `rescale_mantissa` / `decimal_from_magnitude` split restores `main`'s truncation-before-overflow precedence for text sources while leaving the numeric source on the unconditional-overflow rule the comment cites. The `fractional_truncated` / `trailing_truncated` split is a faithful model of `AddRPCUserParameters`' break-and-resume, and the deferred half survives `SQLParamData`'s entry-time diagnostics reset because it is parked on `DaeState` and re-posted by `finish_execute_with_param_warning`.

Nothing blocking. One suggestion about a doc comment that claims more than the code does, and two nits.

## Verification

- `cargo clippy -p mssqlodbc --all-targets -- -D warnings` — clean.
- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` — 1418 passed, 0 failed.
- Mutation checks on the two behaviours this head is about, to confirm the new tests are not vacuous. Both caught it:
  - Disabling the early `ConvOk::Truncated` return in `decimal_from_text` → `a_dropped_fraction_is_22001_even_when_the_result_also_overflows` fails.
  - Forcing `build_named_params` to always set `fractional_truncated` (dropping the `dae_params.is_empty()` split) → `build_named_params_truncation_after_first_dae_param_is_deferred` fails with "the truncation is ordinally after the DAE param, not before it".
- A throwaway probe test in `bound_param.rs` for the suggestion below (reverted; the worktree is clean).
- No live SQL Server here, so none of the e2e cases in `execute_test.cpp` / `param_conversions_test.cpp` were run — the retail-parity claims rest on your measurements plus the CI compare leg.

**CI is not green at this head**, so nothing here is a merge-readiness statement: `Build Linux`, `Build Linux ARM`, `Build Windows` and `Build Windows ARM` are all red on build `173463` while `Build MacOS` passed. That asymmetry plus a clean local clippy and a green 1418-test `mssqlodbc` lib run on Windows suggests the failure is not a plain Rust compile error in the changed crate; worth reading the ADO logs before assuming it is infrastructure.

---

## Blocking

None.

## Suggestions

**1. The `write_to_records` doc comment claims more than the reset actually does (`params/bound_param.rs:102-118`).**

Two statements in the new paragraph do not hold:

- *"every other C type's APD precision/scale stays at whatever `DescRecord::default_for` seeded (`0, 0`)"* — only true when the application never wrote `SQL_DESC_PRECISION` / `SQL_DESC_SCALE` itself. The reset is inside `if self.c_type == SQL_C_NUMERIC`, and nothing else in `write_to_records` touches those two APD fields, so a rebind to a different C type keeps whatever the previous binding left there.
- *"Resetting on every call (not just the first) also closes the APD half of the rebind-staleness gap this comment used to track as [#470]"* — it closes it only for rebinds *whose new C type is* `SQL_C_NUMERIC`. A rebind at the same ordinal from `SQL_C_NUMERIC` to any other C type still inherits the stale pair.

Verified with a throwaway test: seeding an APD record with `(precision: 38, scale: 5)` — the shape a prior `SQL_C_NUMERIC` bind plus an explicit `SQLSetDescFieldW` leaves behind — and calling `write_to_records` for a `SQL_C_CHAR` / `SQL_VARCHAR` binding leaves the record at `(38, 5)`, which is neither msodbcsql's `SetTypeDefaults` answer for `SQL_C_CHAR` nor the `(0, 0)` the comment predicts.

Functionally this is still harmless today — `app_precision` / `app_scale` are read only by `decimal_from_numeric` — so I am not asking for a behaviour change here. The concern is that #470 gets closed on the strength of this comment when the general APD case is still open. Suggest narrowing the wording to "closes the `SQL_C_NUMERIC` case of the APD half" and keeping the #470 pointer for the rest, or extending the reset to a real C-type-keyed `SetTypeDefaults` port.

## Nits

**1. The defensive `Complete` arm drops `trailing_truncated` (`api/execute.rs:315-322`, `api/exec_direct.rs:261-268`).**

Both DAE paths forward only `fractional_truncated` into `finish_execute_with_param_warning` on the `StreamedParamStatus::Complete` arm, so a truncation ordinally after the first DAE parameter is lost there. The arm is the "shouldn't happen" one and already logs at `error!`, so the blast radius is a dropped warning on a path that is anomalous anyway — but `fractional_truncated || trailing_truncated` costs nothing and removes the asymmetry with the `NeedData` arm right below it.

**2. `BoundParam::app_precision` / `app_scale` are the only fields on the struct without a doc comment (`params/bound_param.rs:44-45`).**

Every neighbour has one, and these two are easy to confuse with `column_size` / `decimal_digits` sitting immediately above them — a one-liner saying they are the APD's (source) pair, as opposed to the IPD's (target) pair, would carry its weight.

saurabh500:
There are test failures.

Theekshna:
Thanks. Build 173584 exposed two incorrect retail-parity expectations: numeric truncation was expected to produce `01S07` around DAE, but msodbcsql enters/completes DAE without that diagnostic; and a bare `(38, 0)` numeric binding was expected to return `12.345`, while retail returns `12`.

Both are corrected in 38276730. That commit also resets descriptor binding provenance on application type changes so stale data/indicator pointers cannot be reconstructed after unbinding. Local `cargo bfmt`, `cargo bclippy`, all 1,420 `mssqlodbc` nextest tests, and the affected C++ builds pass. The refreshed CI run will validate the live Rust-driver e2e legs.

David-Engel (commented):

## Summary

Reviewed `38276730` against the merge base with `origin/main` (`09b44d87`) in a dedicated worktree, after re-reading all three prior-discussion channels. The five threads answered at earlier heads are not re-filed, and neither are the two suggestions from my `90fadb36` pass.

This head reverses two behaviours on the strength of retail measurements from build 173584: the pre-DAE `01S07` is no longer posted, and a bare `(38, 0)` numeric bind is now expected to yield `12`. Build `173606` — the validation run on this exact head, which runs the same e2e binaries against msodbcsql as well as against this driver — contradicts both. That is the whole of my blocking feedback; the rest of the change still reads correctly to me, including the `data_bound` split, the IPD-vs-AD scale gate, and the `rescale_mantissa` / `decimal_from_magnitude` precedence from the previous head.

## Verification

- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` — 1420 passed, 0 failed, so the checklist's unit-test claim holds; the red legs are e2e-only.
- ADO build `173606` timeline plus the raw logs for `Build Linux` (log 203) and `Build Windows` (log 273). No live SQL Server here, so all e2e evidence below is CI's rather than mine.
- Parity summaries at this head: Linux `Summary: 38 parity, 1 divergence(s), 0 shared failure(s)`; Windows `Summary: 37 parity, 2 divergence(s), 0 shared failure(s)`.

## Blocking

1. **`NumericStructWithoutDescriptorFieldWritesUsesDefaultApdScale` fails against this driver on all four red legs while msodbcsql passes the same test** — the bare bind takes the matching-descriptor fast path and ships `12.345` where retail ships `12`. Inline on `conversion/param_convert.rs`.
2. **The pre-DAE `01S07` removal matches Linux retail but is contradicted by Windows retail**, which still reports `01S07` at the initiating call and now fails `execute_test` in the parity leg. Inline on `tests/e2e/tests/execute_test.cpp`.

## Suggestions

1. `set_type`'s new unbind is not scoped to parameter descriptors and reaches the ARD/fetch path. Inline on `api/set_desc_field.rs`.

## Nits

1. The literal `false` left behind in `SQLParamData`'s completion arm. Inline on `api/param_data.rs`.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -911,10 +929,82 @@ fn decimal_from_text(param: &BoundParam, text: AppText) -> Result<TypedValue, Pa
...
+
+fn decimal_from_numeric(
+    param: &BoundParam,
+    source: SqlNumericStruct,
+) -> Result<(TypedValue, ConvOk), ParamBuildError> {
+    let magnitude = u128::from_le_bytes(source.val);
+    // msodbcsql's FastDescribeRPCParam copies a matching non-NULL
+    // SQL_NUMERIC_STRUCT whole, including its wire precision and scale
+    // (`sqlcmisc.cpp:7014`).
+    if usize::try_from(param.app_precision) == Ok(param.column_size)
```
David-Engel: **Blocking — this gate now fires for a bare `SQLBindParameter`, and the parity leg on this head shows retail does not fast-path there.**

`write_to_records` seeds the APD at `(SQL_PREC_NUMERIC, 0)` on every `SQL_C_NUMERIC` bind (`params/bound_param.rs:139-142`), so `SQLBindParameter(..., SQL_C_NUMERIC, SQL_DECIMAL, 38, 0, &num, 0, NULL)` with no `SQLSetDescFieldW` call satisfies `app_precision == column_size && app_scale == decimal_digits` and lands here. The branch forwards the struct's own `(5, 3)` into the NUMERICN preamble, so `12.345` reaches the server as `decimal(5,3)` and `CONVERT(VARCHAR(64), ?)` returns `12.345`.

That is exactly what build `173606` measured at this head, on all four failing legs (`param_conversions_test.cpp:995`):

```
Expected equality of these values:
  "12"
  ExecuteAndReadBack()
    Which is: "12.345"
```

The same run executes the same binary against msodbcsql and it passes — Linux parity report: `param_conversions_test  FAIL  PASS  divergence - investigate`. So the corrected `12` expectation from build 173584 is right and this driver is the side that diverges: retail rescales onto the target scale `0` for this binding instead of forwarding the embedded metadata.

The `(38, 0)` default itself is confirmed only for the `SQLSetDescFieldW(SQL_DESC_TYPE)` path — the new `ChangingTypeToNumericResetsDefaultsAndUnbindsData` passes on both drivers. Nothing in the run confirms retail leaves `(38, 0)` on the APD after a plain `SQLBindParameter`, which is the assumption this gate now rests on. The cheapest way to settle which half is wrong is a parity e2e that does the bare bind and reads `SQL_DESC_PRECISION` / `SQL_DESC_SCALE` back off the APD on both drivers:

- if retail reports something other than `(38, 0)` there, the reset in `write_to_records` is what needs narrowing;
- if it does report `(38, 0)`, then this condition needs a third term — the application having actually written the APD pair — rather than the two-field comparison, since retail plainly reaches a different branch with the same descriptor contents.

mssql-odbc/tests/e2e/tests/execute_test.cpp
```diff
@@ -168,6 +200,81 @@ TEST_F(PrepareExecuteLiveTest, DataAtExecutionInterleavesWithBoundParams) {
...
+                  SQL_HANDLE_STMT, stmt_);
+    ASSERT_SQL_OK(SQLBindParameter(stmt_, 2, SQL_PARAM_INPUT, SQL_C_CHAR,
+                                   SQL_VARCHAR, 0, 0, &streamed_token, 0,
+                                   &streamed_ind),
+                  SQL_HANDLE_STMT, stmt_);
+
+    SQLRETURN rc = SQLExecute(stmt_);
+    ASSERT_EQ(SQL_NEED_DATA, rc)
+        << ODBCTestUtils::GetDiagMessage(SQL_HANDLE_STMT, stmt_);
+    EXPECT_EQ("", ODBCTestUtils::GetDiagState(SQL_HANDLE_STMT, stmt_));
```
David-Engel, line 231: **Blocking — this expectation holds on the Linux retail driver but not on the Windows one, and the Windows parity leg fails on it at this head.**

Build `173606`'s Windows leg runs this same binary against `C:\Windows\system32\msodbcsql18.dll`:

```
execute_test.cpp(225): error: Expected equality of these values:
  ""
  ODBCTestUtils::GetDiagState(3, stmt_)
    Which is: "01S07"
```

Both `NumericTruncationBeforeDataAtExecutionIsNotReported` and `ExecDirectNumericTruncationBeforeDataAtExecutionIsNotReported` fail there, and the parity report records `execute_test  PASS  FAIL  divergence - investigate` (`Summary: 37 parity, 2 divergence(s)`). The Linux leg, which apt-installs `msodbcsql18 18.6.2.1` in the job, passes both.

So retail's answer is not uniform, and dropping `fractional_truncated` from `park_dae_client` plus `DaeState::trailing_truncated` picked the Linux one. This is not simply "the earlier fix was wrong" — restoring `c2eb0bb6`'s behaviour would match Windows and fail Linux instead. Worth checking the Windows DLL's file version before deciding: if it predates 18.6.2.1 this is a retail behaviour change between versions rather than a platform split, and the newer answer is presumably the contract, in which case the expectation needs to say so (a version guard, or a comment pinning it to 18.6.2.1) so the Windows leg stops reporting a divergence the driver cannot fix.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -442,6 +442,14 @@ pub(super) fn set_type(
     write_record_field(state, record_number, |r| {
         r.concise_type = resolved;
         r.datetime_interval_code = datetime_interval_code_for(resolved);
+        if kind.is_application() {
```
David-Engel: **Suggestion — this unbind also fires for `AppRow`, which puts a parameter-binding fix on the fetch path.**

`DescKind::is_application()` is `AppRow | AppParam | Ad` (`handles/desc.rs:106-108`), so `SQLSetDescFieldW(ard, n, SQL_DESC_TYPE, ...)` now nulls the column's `SQL_DESC_DATA_PTR`, and `ColumnBinding::from_record` treats a null `data_ptr` as unbound (`handles/stmt.rs:266-269`). Retyping an already-bound column through the ARD — `SQLBindCol`, then a `SQL_DESC_TYPE` write — therefore drops the binding at fetch time rather than retyping it.

The new e2e (`ChangingTypeToNumericResetsDefaultsAndUnbindsData`) covers the APD only, and it passes on both drivers there, so the parameter half is parity-verified; the ARD half is not covered by anything in this PR. Either narrow the reset to the kinds this change is about (`AppParam` / `Ad`), or add the matching ARD parity case, so the fetch path doesn't inherit an unverified behaviour change.

mssql-odbc/src/api/param_data.rs
```diff
@@ -325,7 +327,14 @@ fn sql_param_data_safe(
...
             }
 
-            finish_execute(dbc, stmt, statement_handle, client, "SQLParamData")
+            finish_execute_with_param_warning(
+                dbc,
+                stmt,
+                statement_handle,
+                client,
+                "SQLParamData",
+                false,
```
David-Engel, line 341: **Nit — the literal `false` reads as a leftover rather than a decision.**

With `trailing_truncated` gone, `finish_execute_with_param_warning`'s last argument is now unconditionally `false` from `SQLParamData`. A one-line comment saying the DAE completion path deliberately posts no fractional-truncation warning — which is the point of this head's change — would stop the next reader from "restoring" it.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Confirmed and fixed in 04e1ebc6. Added `DescRecord::precision_scale_explicit`, set only by `set_precision`/`set_scale` (i.e. an actual `SQLSetDescFieldW`/`SQLSetDescRec` write), and added as a third condition to the fast-path gate. `write_to_records` (SQLBindParameter) and `set_type`'s `SQL_C_NUMERIC` default-fill both reset it to `false`, mirroring msodbcsql's `SetADRecBP`/`SetTypeDefaults` reset-on-every-bind. Also had to fix the slow (rescale) path: it was unconditionally trusting `app_scale` as the source scale even when never explicitly written, which would have misread a bare bind's struct at the *defaulted* scale (0) instead of its own embedded scale — now it falls back to the struct's own scale when `precision_scale_explicit` is false, which is what produces retail's "12" (not "12345") for the exact scale-3-struct-into-NUMERIC(38,0) case. Added unit test coverage for both the gate and the fallback (out-of-range and in-bounds cases).

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Theekshna: Agreed, fixed in 04e1ebc6 by taking the narrower option: scoped the `data_ptr`/`data_bound`/`precision_scale_explicit` reset in `set_type` to `AppParam`/`Ad` only, excluding `AppRow`. That reset mirrors msodbcsql's APD-side `SetTypeDefaults` behavior on a parameter-binding concept with no verified ARD analogue here, so `AppRow` retyping through `SQLSetDescFieldW` no longer touches `data_ptr`/`data_bound` and won't unbind an already-bound fetch column.

> [!NOTE]

mssql-odbc/src/api/param_data.rs
Theekshna: Added the comment in 04e1ebc6: the literal `false` is intentional — a DAE completion never carries a fractional-truncation warning here since the truncation check runs once against the fully-streamed value, not per `SQLPutData` chunk.

> [!NOTE]

mssql-odbc/tests/e2e/tests/execute_test.cpp
Theekshna: Root-caused this: it isn't a msodbcsql version/platform split at all. unixODBC's `function_return_ex` (`DriverManager/__info.c`) only extracts driver diagnostics for `SQL_SUCCESS_WITH_INFO`, `SQL_ERROR`, and `SQL_NO_DATA` — not `SQL_NEED_DATA`. So on Linux, neither driver's `01S07` posted before `SQL_NEED_DATA` is ever visible to the application through `SQLGetDiagRec`, regardless of what the driver itself posts. Windows' driver manager doesn't have that filter, so `01S07` is visible there.

This driver now posts the diagnostic unconditionally on every platform (matching `AddRPCUserParameters`/`sqlccmd.cpp`, pinned by `park_dae_client_posts_numeric_fractional_truncation`), and the e2e expectations are gated with `#ifdef __linux__` to assert no visible diagnostic there (verified against retail msodbcsql 18.6.2.1 in build 173710 — both drivers show the same platform split). Documented in `mssql-odbc.instructions.md` with the citation and AB#47946. Truncation after the first DAE parameter remains silent per the existing tracking.

> [!NOTE]

mssql-odbc/tests/e2e/tests/execute_test.cpp
```diff
@@ -168,6 +200,88 @@ TEST_F(PrepareExecuteLiveTest, DataAtExecutionInterleavesWithBoundParams) {
...
+                  SQL_HANDLE_STMT, stmt_);
+    ASSERT_SQL_OK(SQLBindParameter(stmt_, 2, SQL_PARAM_INPUT, SQL_C_CHAR,
+                                   SQL_VARCHAR, 0, 0, &streamed_token, 0,
+                                   &streamed_ind),
+                  SQL_HANDLE_STMT, stmt_);
+
+    SQLRETURN rc = SQLExecute(stmt_);
+    ASSERT_EQ(SQL_NEED_DATA, rc)
+        << ODBCTestUtils::GetDiagMessage(SQL_HANDLE_STMT, stmt_);
+#ifdef __linux__
```
David-Engel: **Suggestion — this guard splits on OS, but the behavior splits on driver manager; macOS lands in the Windows branch.**

The suppression you root-caused is unixODBC's `function_return_ex`, and macOS builds of this suite link unixODBC as well — `tests/e2e/CMakeLists.txt:36` says "On Linux / macOS, prefer unixODBC; fall back to iODBC on macOS". With `#ifdef __linux__`, a macOS run takes the `#else` branch and asserts `01S07`, which the very same unixODBC filter will hide, so the test fails there for a reason that has nothing to do with the driver. Every other platform split in this suite keys off `_WIN32` (`connection_pool_test.cpp:29`, `driver_connect_test.cpp:381`, `dll_unload_stress_test.cpp:46`).

Inverting the guard keeps the Windows expectation and makes the unixODBC case the default:

```cpp
#ifdef _WIN32
    EXPECT_SQLSTATE(SQL_HANDLE_STMT, stmt_, "01S07");
#else
    // unixODBC's function_return_ex does not extract driver diagnostics for
    // SQL_NEED_DATA, so neither driver's 01S07 is application-visible here.
    EXPECT_EQ("", ODBCTestUtils::GetDiagState(SQL_HANDLE_STMT, stmt_));
#endif
```

Same for the copy at line 773, and the new instructions entry (item 14) reads "On Linux, unixODBC's `function_return_ex` ..." — "on unixODBC platforms (Linux and macOS)" would describe the actual rule.

Not urgent: CI doesn't build the e2e binaries on macOS today (`build-odbc-macos-template.yml:11` — "Driver only - no C++ gtest e2e binaries"), so this only bites a local macOS run or a future macOS e2e leg.

David-Engel (commented):

## Summary

Reviewed `2848cf44` against the merge base with `origin/main` in a dedicated worktree, re-reading the four prior discussion threads plus the four commits since `38276730` (`04e1ebc6`, `e0aefbb6`, `83a867dd`, `2848cf44`). Ran `cargo nextest run -p mssqlodbc --lib` at this head: 1,423 passed, 0 failed. All required checks are green, including both parity legs.

The earlier blocking items look genuinely resolved:

- The `precision_scale_explicit` gate keeps a bare `SQLBindParameter` off the struct-metadata fast path, and the slow path now falls back to the struct's own embedded scale when the APD was never written explicitly — `a_numeric_without_explicit_apd_precision_scale_still_rescales` and its in-bounds sibling pin both halves.
- Retaining the flag across a same-`ValueType` rebind (`e0aefbb6`) while still resetting the values matches the `SQLBindParameter` rebind rule, and `write_to_records_keeps_explicit_flag_across_a_same_type_rebind` / `..._resets_numeric_apd_precision_and_scale_to_msodbcsql_default` cover both directions.
- Scoping the `set_type` unbind to `AppParam`/`Ad` keeps the ARD fetch path out of it.
- `park_dae_client` now posts `01S07` before `SQL_NEED_DATA`, and `build_named_params` correctly folds only pre-first-DAE truncations into the flag (`build_named_params_ignores_truncation_after_first_dae_param`).
- `data_bound` is set on every path that establishes a binding (`write_to_records`, `set_data_ptr`, and therefore `SQLSetDescRec`) and is seeded `false` by `DescRecord::default_for`, so `SQL_RESET_PARAMS`-then-regrow can't resurrect a stale record.

Two remaining items, neither blocking.

### Nit — the PR description's second bullet no longer matches this head

"Match retail diagnostics ... including no numeric fractional-truncation diagnostic once execution enters DAE" was accurate at `604b15c4`, but since `83a867dd` the driver posts `01S07` before `SQL_NEED_DATA` on every platform; only unixODBC's driver-manager filter hides it from the application. Worth rewording (and pointing at the new `.github/instructions/mssql-odbc.instructions.md` item 14) before marking ready for review — the repo workflow asks for the description to reflect the final change set.

The remaining suggestion is inline on `mssql-odbc/tests/e2e/tests/execute_test.cpp`.

mssql-odbc/tests/e2e/tests/execute_test.cpp
Theekshna: Good catch — you're right, the split is unixODBC vs. driver-manager, not Linux specifically. Inverted the guard to `_WIN32` (Windows expects `01S07`, everything else — Linux and macOS, both on unixODBC — expects no visible diagnostic) at both call sites, and reworded the instructions entry to say "on unixODBC platforms (Linux and macOS)". Pushed in `e9902f73`.

> [!NOTE]

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -911,10 +929,102 @@ fn decimal_from_text(param: &BoundParam, text: AppText) -> Result<TypedValue, Pa
...
+    // (`sqlcmisc.cpp:7014`).
+    // Additionally requires an application-authored APD precision/scale
+    // (`SQLSetDescField`/`SQLSetDescRec`), not merely a coincidental match
+    // with this driver's own default-fill (`SQLBindParameter`'s
+    // `SQL_PREC_NUMERIC`/scale-0 reset, or a `SQL_DESC_TYPE` rewrite):
+    // msodbcsql's parity-comparison harness shows a bare `SQLBindParameter`
+    // whose IPD happens to be `(precision, scale=0)` does NOT take this fast
+    // path in retail — it still rescales the embedded struct to the IPD's
+    // scale, unlike an app that explicitly wrote matching APD fields.
+    if param.precision_scale_explicit
```
David-Engel, line 964: **Suggestion — the retained `precision_scale_explicit` flag makes the same bare bind fast-path or not, depending on descriptor history.**

`write_to_records` keeps the flag across a same-`ValueType` rebind while resetting the values to `(38, 0)` (`params/bound_param.rs:149-160`, pinned by `write_to_records_keeps_explicit_flag_across_a_same_type_rebind`). So at an ordinal that once had an explicit APD precision/scale write, a plain `SQLBindParameter(..., SQL_C_NUMERIC, SQL_DECIMAL, 38, 0, &num, 0, NULL)` satisfies all three terms of this gate and forwards the struct's embedded metadata — while the identical call on a fresh record fails the flag term and rescales.

Measured at this head with a throwaway test calling `decimal_from_numeric` directly, struct `(precision 5, scale 3, magnitude 12345)` into a `(38, 0)` target:

```
precision_scale_explicit = false -> Decimal(38, 0, 12)      // "12" — what NumericStructWithoutDescriptorFieldWritesUsesDefaultApdScale pins
precision_scale_explicit = true  -> Decimal(5, 3, 12345)    // "12.345" — the fast path
```

`NumericRebindDoesNotInheritAPreviousBindsStaleApdScale` doesn't reach this: its second bind targets `(10, 2)`, so the numeric comparison fails before the flag matters. Retail has no analogue of this flag — `SetADRecBP`/`SetTypeDefaults` reset the record on every call — so whichever branch retail takes for a bare `(38, 0)` bind, it should presumably take the same one regardless of what happened at that ordinal earlier.

Worth an e2e that repeats the bare `(38, 0)` bind at an ordinal whose APD precision/scale were explicitly written earlier, run against both drivers. If retail rescales there too, note that dropping the retention alone would regress `NumericRebindDoesNotInheritAPreviousBindsStaleApdScale` — its second bind would then fall back to the struct's embedded scale `9` instead of the freshly-reset APD scale `0` — so the non-explicit fallback rule below (line 1000) would have to move with it.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -526,11 +554,12 @@ pub(super) fn set_scale(
         return SQL_ERROR;
     };
 
+    let is_application = kind.is_application();
     let record_info = state
         .record(record_number)
         .map(|r| (r.concise_type, r.precision));
     if let Some((SQL_C_NUMERIC, precision)) = record_info
-        && (scale < 0 || scale > precision)
+        && (scale > precision || (!is_application && scale < 0))
```
David-Engel, line 567: **Nit — the IPD half of this gate covers `SQL_NUMERIC` but not `SQL_DECIMAL`.**

`record_info` matches the concise type against `SQL_C_NUMERIC` (`2`), which doubles as `SQL_NUMERIC` on an IPD record — that is what makes the new `!is_application && scale < 0` arm reachable. An IPD record typed `SQL_DECIMAL` (`3`) isn't matched at all, so it still accepts a negative `SQL_DESC_SCALE` and answers `SQL_SUCCESS` where the `SQL_NUMERIC` spelling of the same target now answers `HY094`.

Low impact and not a regression: the same gap exists on `main`, and the bad scale is still caught at execute time by `decimal_metadata` as `InvalidDecimalDigits` → `HY104` (`conversion/param_convert.rs:1401-1404`), so nothing malformed reaches the wire. Flagging it only because this PR is what introduces the IPD-vs-AD split here — if `CheckSqlPrecScale<FALSE>` treats `SQL_DECIMAL` the same way it treats `SQL_NUMERIC`, widening the match is a one-line follow-up.

David-Engel (commented):

## Summary

Reviewed `e9902f73` against the merge base with `origin/main` (`09b44d87`) in a dedicated worktree, after re-reading all three prior-discussion channels. The head commit is exactly the `_WIN32` inversion asked for last pass, applied at both call sites plus the instructions entry, and it now matches every other platform split in this suite (`connection_pool_test.cpp:29`, `driver_connect_test.cpp:381`, `dll_unload_stress_test.cpp:46`). Every thread open at an earlier head reads as answered at this one, and none is re-filed.

Nothing blocking. One suggestion about an untested descriptor-history interaction in the `SQL_C_NUMERIC` fast-path gate, and one pre-existing nit next to a line this PR changes.

## Verification

- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` at this head — 1,423 passed, 0 failed.
- A throwaway probe test in `param_convert.rs` to settle the suggestion empirically rather than by reading (reverted; the worktree is clean).
- `gh pr checks 521`: ADO build `173830` — Build Linux, Build Linux ARM, Build Windows, Build Windows ARM, both cross-repo mssql-python legs and the Kerberos leg are green; the three macOS legs were still pending when I looked. No live SQL Server here, so all e2e and parity evidence is CI's rather than mine.

## Blocking

None.

## Suggestions

1. A same-`ValueType` rebind retains `precision_scale_explicit`, so an otherwise-identical bare `SQLBindParameter` takes the fast path or the rescale path depending on descriptor history. Inline on `conversion/param_convert.rs`.

## Nits

1. The new IPD negative-scale gate covers `SQL_NUMERIC` but not `SQL_DECIMAL`. Inline on `api/set_desc_field.rs`.

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Agreed this is a real gap worth pinning down precisely rather than guessing. I traced both `write_to_records` and `decimal_from_numeric` and confirmed the scenario as described: a same-C-type bare rebind retains `precision_scale_explicit` from a prior explicit write, so a subsequent bare `(38,0)` bind can take the fast path even though it was never itself authored via `SQLSetDescField`. I ran a differential probe against the locally installed retail ODBC Driver 18 to try to settle it empirically, but it surfaced additional subtlety in the plain bare-bind case that I couldn't fully reconcile with our existing pinned tests in the time available — this needs a careful e2e test using the existing `BindNumericRaw`/`SQLSetDescFieldW` helpers (not an ad hoc script) to get a trustworthy answer, and any fix would also need to account for `NumericRebindDoesNotInheritAPreviousBindsStaleApdScale`, which depends on the flag surviving a same-type rebind in a different sub-case. Filed as microsoft/mssql-rs#536 to track with full repro details, so it doesn't block this PR.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Theekshna: Good catch, and confirmed against msodbcsql: `CheckSqlPrecScale<FALSE>` (sqlcdesc.cpp) has a single `case SQL_NUMERIC: case SQL_DECIMAL:` block applying the identical unsigned-scale bound to both. Widened the `set_scale` match to `SQL_C_NUMERIC | SQL_DECIMAL`, updated the doc comment to cite the retail switch, and added a test (`set_desc_field_ipd_decimal_scale_may_not_be_negative`) mirroring the existing `SQL_NUMERIC` one.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -439,9 +439,26 @@ pub(super) fn set_type(
...
     };
 
+    // Scoped to AppParam/Ad, not AppRow: this mirrors msodbcsql's APD-side
+    // SetTypeDefaults reset (SQL_C_NUMERIC's (SQL_PREC_NUMERIC, 0) default),
+    // which is a parameter-binding concept with no ARD analogue verified
+    // here. Including AppRow would drop an already-bound fetch column's
+    // `data_ptr` the moment its C type is retyped through the ARD, turning a
+    // legitimate rebind into a silent unbind with no e2e coverage for that
+    // half.
+    let is_application_param = matches!(kind, DescKind::AppParam | DescKind::Ad);
```
David-Engel: **Blocking — `DescKind::Ad` is the explicit-descriptor kind, and an explicit descriptor can be the ARD, so this reset still silently unbinds fetch columns.**

The previous pass flagged that `is_application()` pulled `AppRow` into this reset and dropped an already-bound fetch column's `SQL_DESC_DATA_PTR`. Narrowing to `AppParam | Ad` fixes the implicit ARD, but `Ad` is exactly the kind `SQLAllocHandle(SQL_HANDLE_DESC, ...)` produces (`DescRecord::default_for` groups `AppRow | AppParam | Ad`), and `validate_descriptor_association` (`api/set_stmt_attr.rs:517-543`) accepts any explicit descriptor on the same connection for `SQL_ATTR_APP_ROW_DESC`. So the ARD-side half of that finding is still reachable — just through the explicit-descriptor route instead of the implicit one.

I reproduced it at this head with a throwaway test: allocate an explicit descriptor, `SQLSetStmtAttrW(SQL_ATTR_APP_ROW_DESC, desc)`, `SQLBindCol(stmt, 1, SQL_C_CHAR, buf, ...)`, then `SQLSetDescFieldW(desc, 1, SQL_DESC_TYPE, SQL_C_LONG, 0)`. `data_ptr` reads back as `0x0`, and `ColumnBinding::from_record` (`handles/stmt.rs:266-269`) treats a null `data_ptr` as unbound, so the next `SQLFetch` writes nothing to `buf` and reports no diagnostic. On `main` this sequence retypes the column and keeps the binding, so it is a regression rather than a pre-existing gap.

The `is_explicit()` route also means the same descriptor can be the APD on one statement and the ARD on another, so kind alone can't tell you the role here. The role-accurate signal is which slot the descriptor is currently associated to, not `DescKind`, which suggests either resolving the effective role at the call site or keying this reset on `AppParam` only until the ARD half is understood and covered. Either way it would be worth an e2e mirroring `ChangingTypeToNumericResetsDefaultsAndUnbindsData` on an explicit ARD, since nothing in the current suite exercises that pairing.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -925,22 +1035,37 @@ fn decimal_from_text(param: &BoundParam, text: AppText) -> Result<TypedValue, Pa
...
+            Some(divisor) => {
+                if mantissa % divisor != 0 {
+                    outcome = ConvOk::Truncated;
+                }
+                mantissa / divisor
+            }
             None if mantissa == 0 => 0,
-            None => return Err(ParamBuildError::StringTruncation),
+            None => {
+                outcome = ConvOk::Truncated;
```
David-Engel, line 1049: **Suggestion — on the `SQL_C_NUMERIC` path this arm turns an unvalidated source scale into a silent `0` plus a warning, where the text path raises an error.**

The comment above this arm reasons from `decimal_from_text`, where `source_scale` came out of a parsed literal and the caller converts `ConvOk::Truncated` into a hard `ParamBuildError::StringTruncation` — so reporting `22001` instead of `22003` is the whole point and the value never leaves the driver. `decimal_from_numeric` deliberately does not do that conversion: it lets `decimal_from_magnitude` run and returns the outcome, so `Truncated` becomes `01S07` and the `0` is what actually goes on the wire.

That matters because the non-explicit branch just above (`let source_scale = if param.precision_scale_explicit { ... } else { i64::from(source.scale) }`) takes the scale straight from the application's `SQL_NUMERIC_STRUCT` with no bounds check — unlike the explicit branch, where `set_scale`/`set_data_ptr` have already constrained the APD scale to `..= precision <= 38`. A bare `SQLBindParameter(..., SQL_C_NUMERIC, SQL_DECIMAL, 10, 2, &num, 0, NULL)` where the struct carries `scale = 45` lands here: `10i128.checked_pow(43)` is `None`, so the parameter is sent as `0` and the statement succeeds with `01S07`, silently discarding the value.

The struct scale is garbage in that case, so rejecting it seems more defensible than rounding it to zero — either `ParamBuildError::Value(ConvError::OutOfRange)` for `22003`, or an up-front bound on `source.scale` in `decimal_from_numeric`'s non-explicit branch before it reaches the rescale. Worth confirming against the parity harness which one retail produces, since the answer also settles what happens for a struct scale in the merely-unusual `39..=127` range.

mssql-odbc/src/api/set_desc_field.rs
```diff
@@ -526,11 +556,12 @@ pub(super) fn set_scale(
         return SQL_ERROR;
     };
 
+    let is_application = kind.is_application();
     let record_info = state
         .record(record_number)
         .map(|r| (r.concise_type, r.precision));
-    if let Some((SQL_C_NUMERIC, precision)) = record_info
-        && (scale < 0 || scale > precision)
+    if let Some((SQL_C_NUMERIC | SQL_DECIMAL, precision)) = record_info
```
David-Engel, line 566: **Nit — the upper bound reads `precision` as already-written, so it is order-dependent on a fresh IPD record.**

`DescRecord::default_for(ImpParam)` seeds `precision: 0` (`handles/desc.rs:284`), and `set_type` does not seed a type-appropriate precision on the IPD side. So `SQLSetDescFieldW(ipd, 1, SQL_DESC_TYPE, SQL_DECIMAL)` followed by `SQL_DESC_SCALE = 2` now fails `HY094` on `2 > 0`, where `main` returned `SQL_SUCCESS` for the same sequence. Writing `SQL_DESC_PRECISION` first — which both the new tests and `SQLSetDescRec`'s fixed field order do — avoids it.

Not a reason to hold the PR: the same order sensitivity already exists on `main` for the `SQL_NUMERIC` spelling, this only widens it to `SQL_DECIMAL`, and precision-before-scale is the natural order. Mentioning it because the retail rule being modeled here is a consistency check over a complete record, and reproducing it as a per-field write check makes the result depend on write order — so it may be worth confirming that `CheckSqlPrecScale<FALSE>` really does reject a scale written before its precision, rather than only running once the record is complete.

David-Engel (commented):

## Summary

Reviewed `413f757c` against the merge base with `origin/main` (`09b44d87`) in a dedicated worktree, after re-reading every review thread on this PR. The head commit is the `SQL_DECIMAL` widening asked for last pass, and it is scoped correctly: an application descriptor can never hold concise type `3` (`set_type` rejects it through `is_valid_c_type`), so the widened arm only reaches an IPD record, which is what `CheckSqlPrecScale<FALSE>` covers.

One blocking finding, and it is a residual half of an earlier thread rather than a new regression in this commit: the `set_type` reset was narrowed from `is_application()` to `AppParam | Ad`, which removes the implicit-ARD case but leaves the explicit-descriptor-used-as-ARD case live. I reproduced it locally. Two smaller items below.

## Verification

- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` at this head with the review `CARGO_TARGET_DIR` — 1,424 passed, 0 failed.
- A throwaway probe test in `api/set_stmt_attr.rs` that allocates an explicit descriptor, associates it as `SQL_ATTR_APP_ROW_DESC`, binds column 1 with `SQLBindCol`, then writes `SQL_DESC_TYPE` on that descriptor. `SQL_DESC_DATA_PTR` reads back as `0x0` afterwards, so the column is unbound at fetch time. Probe reverted; the worktree is clean.
- `gh pr checks 521`: ADO build `173864` — Build Windows ARM, both cross-repo mssql-python legs and the Kerberos leg are green; Build Windows, Build Linux, Build Linux ARM, both macOS legs and the coverage report were still pending when I looked. No live SQL Server here, so all e2e and parity evidence is CI's rather than mine.

## Blocking

1. The narrowed `set_type` unbind still fires for `DescKind::Ad`, so retyping a column bound through an explicitly-allocated ARD silently unbinds it. Inline on `api/set_desc_field.rs`.

## Suggestions

1. `rescale_mantissa`'s no-representable-divisor arm turns an oversized source scale into a silent zero plus `01S07` on the `SQL_C_NUMERIC` path, where the sibling text path raises `22001`. Inline on `conversion/param_convert.rs`.

## Nits

1. The widened IPD scale gate is order-dependent on a fresh record, since `precision` defaults to `0`. Inline on `api/set_desc_field.rs`.

mssql-odbc/src/api/set_desc_field.rs
Theekshna: Confirmed against retail source, and you're right that the narrowing was itself the mistake. `SQLSetDescField`'s `SQL_DESC_TYPE`/`SQL_DESC_CONCISE_TYPE` handler (`sqlcdesc.cpp:1736-1745`) resets `rgbValue` to `NOT_BOUND` and calls `SetTypeDefaults` for any record where `ObjectType == SQL_HANDLE_AD` — and per `sqlsrv.h:540-542` that's a single object type shared by the implicit ARD, implicit APD, and any explicit descriptor, with no role distinction at this layer at all. `SetTypeDefaults` itself is likewise fully role-agnostic. So retail's reset is unconditional on "is this an AD," not "is this specifically an APD" — which also means kind alone was never the problem; the gate just needed to cover all three `is_application()` kinds instead of trying to special-case one of them out.

Reverted to `kind.is_application()` in `0eb1f4bb`, closing both the implicit-ARD and explicit-desc-as-ARD gaps you found. Added `changing_ard_type_to_numeric_resets_defaults_and_unbinds_data` (unit test, implicit ARD) and `changing_explicit_desc_type_to_numeric_unbinds_data_when_used_as_ard` (unit test) plus `ChangingTypeToNumericThroughArdUnbindsData` and `ChangingTypeToNumericThroughExplicitDescUsedAsArdUnbindsData` (e2e, mirroring your suggested repro exactly) in the same commit.

> [!NOTE]

mssql-odbc/src/conversion/param_convert.rs
Theekshna: Checked this against retail's `SQL_C_NUMERIC`→native-`NUMERICN` conversion in `sqlccnvt.cpp` (the function with the `case SQL_C_NUMERIC:` arm, falling into the shared rescale block ~line 7254-7330). Retail always compares the struct's own claimed scale against the target's declared scale via a digit-at-a-time loop (`numeric_div10`/`numeric_shift10add`), with no upfront bound on the source scale. A dropped nonzero digit while shifting right produces a soft `CVT_FRACT_TRUNC`/`01S07` unless the loop's *current* scale is `<= 0` — which is unreachable here since the target scale is always non-negative and validated. A hard `22003` only comes from a *separate*, later check (`NumeLenToPrec[...] > precision`), driven by the resulting digit count, not the source scale itself.

That's mathematically equivalent to what `rescale_mantissa` does today: when `10i128.checked_pow` overflows, the result is forced to `0` (unless already exactly `0`) and reported `Truncated`/`01S07` — the same answer a digit-at-a-time division to 0 would reach. So the current silent-truncation-to-warning behavior is verified retail parity; rejecting the value upfront with a hard `22003` (as suggested) would actually be a divergence from retail, not a fix. Leaving this as-is. Appreciate the sharp catch on the asymmetry with `decimal_from_text` — that asymmetry is intentional and mirrors retail's own text-vs-struct paths.

> [!NOTE]

mssql-odbc/src/api/set_desc_field.rs
Theekshna: Confirmed — `CheckSqlPrecScale<FALSE>` is invoked from `CheckIPDDescRecConsistency` (`sqlcdesc.cpp:~11437-11455`), which runs against the IPD's complete, current fields as a full-record consistency check (at bind/execute time), not inlined into the per-field `SQL_DESC_SCALE`/`SQL_DESC_PRECISION` write handlers. So retail genuinely doesn't reject a scale-before-precision write at the moment of the call; this driver's current field-at-a-time check is order-dependent by design, and that's a real (if narrow) divergence.

As you noted it's pre-existing for `SQL_NUMERIC` and not introduced by this PR's `SQL_DECIMAL` widening, and precision-before-scale is the natural/expected write order (matching `SQLSetDescRec`'s fixed field order and the tests here). Given your own framing that this isn't a reason to hold the PR, and that fixing it properly would mean deferring this validation to a full-record check at bind/execute time — a real design change, not a nit-sized fix — I'm leaving it as a known, accepted limitation rather than opening a new issue for it.

> [!NOTE]

David-Engel (commented):

## Summary

Reviewed `a6c05c10` against the merge base with `origin/main` (`ea1dfe61`) in a dedicated worktree, after re-reading every review thread on this PR. **No new findings.**

The delta since the last pass (`413f757c`) is two commits:

- `0eb1f4bb` reverts the `set_type` unbind gate from `AppParam | Ad` back to `kind.is_application()`, which is what the previous blocking finding asked for. The justification checks out on both halves: `DescKind::Ad` is the kind an explicit descriptor carries whichever role it is plugged into, so the narrowed gate genuinely left the explicit-descriptor-as-ARD case live, and widening to `AppRow` is consistent with retail resetting `rgbValue` for any `ObjectType == SQL_HANDLE_AD`. The two new unit tests and two new e2e tests cover the implicit-ARD and explicit-desc-as-ARD paths that the earlier thread reproduced.
- `a6c05c10` merges `main` and repairs `fuzz_support.rs` after #511 landed. The harness now fills the three new `BoundParam` fields with `SQLBindParameter`'s pre-explicit-write defaults and drives `bound_param_to_rpc` instead of the now-`#[cfg(test)]` `bound_param_to_value`. Fuzz-only; no shipped behavior moves.

Two things I checked specifically because the ARD widening could plausibly have broken them, and both are fine:

- `SQLSetDescRec` orders `set_type` **before** `set_precision`/`set_scale` (`api/set_desc_rec.rs:192-217`), so the `SQL_C_NUMERIC` `(38, 0)` reset inside `set_type` cannot silently discard a caller-supplied precision/scale in the same call.
- `ColumnBinding::from_record` keys "unbound" off a null `data_ptr` (`handles/stmt.rs:266-269`), so the retyped ARD record really is skipped by the fetch loop — the new `EXPECT_EQ(7, old_value)` assertion has something behind it rather than passing by accident.

The widening is also a net improvement on a path nobody flagged: before `0eb1f4bb`, `SQLSetDescFieldW(ard, n, SQL_DESC_TYPE, SQL_C_NUMERIC)` left the implicit ARD's precision at `0`, so the following `SQL_DESC_DATA_PTR` write failed `set_data_ptr`'s `1..=38` bound with `HY094`. It now seeds `(38, 0)` first and the write succeeds.

Everything else in the diff is unchanged from the previously reviewed heads, and all seven prior threads are answered — including the two closed as verified retail parity (the fast path's unvalidated struct bytes, and `rescale_mantissa`'s overflow-to-zero + `01S07`) and the one deferred to #536.

## Verification

- `cargo clippy -p mssqlodbc --all-targets --all-features` at this head with the review `CARGO_TARGET_DIR` — clean, no warnings.
- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` — **1,442 passed, 0 failed**.
- `gh pr checks 521`: ADO build `173970` is still running; the cross-repo mssql-python suite on mssql-odbc has passed, the rest of the matrix is pending. No live SQL Server here, so all e2e and parity evidence is CI's rather than mine.

## CI note

`coverage-report` is red, but it is not reporting anything about this head's code. The job's "Wait for ADO build to start" step picked up the *previous* ADO build for this PR (`173959`, completed/failed) and exited 1 on `❌ Azure DevOps build 173959 failed; cannot report coverage.` before ever fetching an artifact. Build `173970` is the one for `a6c05c10` and is still in progress. Worth re-running the coverage job once `173970` finishes, and worth confirming `173959`'s own failure was infrastructure rather than a real break — I could not read the ADO logs from here.

Because a required check is currently failing, this review does not add the `ready for human review` label.

gargsaumya (approved):

Vahid-b (approved):

PR #523, opened 2026-09-08
Deliver SQL_C_BINARY results for binary, UDT and LOB fetch (AB#47239)

David-Engel:
Resolves [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239).

## Problem

`SQL_C_BINARY` answered only the zero-length length probe. Any read that actually asked for bytes returned `HYC00`, through `SQLGetData` and through bound columns alike. Since `varbinary`, `image`, CLR UDT and `SQL_SS_TIME2`-below-3.8 all resolve to `SQL_C_BINARY` by default, that made binary, UDT and LOB fetch unusable without the application naming a different C type — and it is what mssql-python's binary/LOB/Arrow/sql_variant suites were failing on.

This is a restore, not a new feature: binary delivery existed in P1 and was removed in `5de38575` as collateral when chunked-offset streaming was ceded to #153. PR #482 restored the probe half; this restores delivery, re-expressed on the wire-stream model rather than an offset over a materialized value.

## What changed

Delivery on all four paths that can produce a binary result:

| Path | Change |
| --- | --- |
| Materialized value, `SQLGetData` | New `deliver_captured_binary`, chunked across repeated calls using the existing partial-read offset |
| PLP stream, `SQLGetData` | Opened the target gates in `stream_active_plp_chunk`, plus a raw-bytes write branch with no terminator |
| Bound non-PLP column | New `SQL_C_BINARY` branch in `deliver_bound` |
| Bound PLP column | Gate, capacity, and final copy in `deliver_bound_plp` |

Three things were less obvious than expected and are worth a reviewer's attention:

- **PLP streaming has two independent compatibility gates** — one admits the first chunk, a different one admits the resumed chunk. Opening only the first made single-chunk reads pass and chunked reads fail on the second call.
- **Binary capacity is the full stride**, not `stride - 1`. The text paths reserve a terminator slot; binary has none, so reusing their arithmetic drops the last byte of every full buffer.
- **Binary must not share the character accumulator's tail handling.** A bound binary read of a UTF-8 encoded PLP column fell through to the `Utf8Text` arm, which trims a truncated tail back to a whole UTF-8 character — silently dropping bytes. Caught in review; `ABoundBinaryReadOfAUtf8ColumnKeepsTheTruncatedTail` fails as a divergence against msodbcsql if the trim is restored.

## Scope

Only kinds whose byte form is fixed by the value alone are delivered: `binary` / `varbinary` / `image`, the character types as their wire bytes, `xml`, `json`, and the PLP UDTs (`hierarchyid`, `geometry`, `geography`). That is exactly what the removed P1 implementation covered — its `column_value_to_bytes` handled `Bytes`/`Xml`/`Json` — plus the character types.

Fixed-width scalars (`int`, `bigint`, `money`, `uniqueidentifier`, `date`, `datetime2`, `decimal`, ...) still return `HYC00`, as they did before this change. `FixedWidthKindsAreStillRefused` pins that boundary. msodbcsql does convert them, but inconsistently enough that each needs its own decision rather than a guessed table:

| Type | msodbcsql delivers |
| --- | --- |
| `int` | 4 wire bytes (`02010000` for 258) |
| `money` | 8 wire bytes, high dword first |
| `uniqueidentifier` | 16 bytes, mixed-endian |
| `date` | 6 bytes — a `SQL_DATE_STRUCT`, not the 3-byte wire value |
| `datetime2(7)` | 16 bytes — a `SQL_TIMESTAMP_STRUCT` |
| `decimal(10,2)` | `SQL_ERROR` / `22003` |

`column_value_to_bytes` records these so the follow-up does not have to re-derive them.

## The contract

Measured from msodbcsql; the new e2e tests run on both drivers rather than against hardcoded expectations:

- The indicator reports bytes **remaining before the call**, not bytes written. A `binary(9)` read into a 4-byte buffer reports 9, then 5, then 1.
- Chunked reads return `SQL_SUCCESS_WITH_INFO` + `01004` until the final chunk, which is plain `SQL_SUCCESS`; a further call is `SQL_NO_DATA`.
- No null terminator — the entire buffer is payload.
- An empty `varbinary` is `SQL_SUCCESS` with indicator 0.
- Bound deliveries report their length on `SQL_DESC_OCTET_LENGTH_PTR`, not the indicator.

## Testing

- New `binary_fetch_test.cpp`: 13 cases, every expectation parity-checked against msodbcsql.
- Three tests that pinned the old "still unimplemented" posture are rewritten to assert the delivered behaviour, and no longer skip the msodbcsql leg now that both drivers agree.
- Full e2e suite: **40 parity, 0 divergences, 0 shared failures**.
- `cargo test -p mssqlodbc`: 1391 passed. `cargo clippy --all-targets`: clean.

### mssql-python impact

Measured with this driver swapped in, against SQL Server 2025:

| Suite | Before | After |
| --- | --- | --- |
| `test_004_cursor.py` (binary/LOB subset) | 23 failed / 15 passed | 12 failed / 26 passed |
| `test_004_cursor_arrow.py` | 4 failed + a crash | 49 passed, 1 skipped |
| `test_019_sql_variant.py` | failing | 29 passed |
| `test_017_spatial_types.py` | failing | 1 failed / 36 passed |
| `test_013_encoding_decoding.py` | failing | 8 failed / 181 passed |

Every residual failure above — including the ones with `fetch` in their name, which abort in their `executemany` setup — is `SQL_ATTR_PARAMSET_SIZE > 1 is not supported`. That is parameter-array binding, a separate gap. No binary *fetch* failure remains.


mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -1797,6 +1793,8 @@ unsafe fn deliver_bound_plp(
             | (SQL_C_CHAR, PlpEncoding::SingleByteText)
             | (SQL_C_CHAR, PlpEncoding::Utf8Text)
             | (SQL_C_CHAR, PlpEncoding::Utf16Text)
+            // Binary delivers the wire bytes whatever the column holds.
+            | (SQL_C_BINARY, _)
```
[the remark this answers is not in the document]
David-Engel: Confirmed and fixed in 28875e73. The dispatch chain reached the `Utf8Text` arm for binary because none of the earlier conditions match `SQL_C_BINARY`, so `trim_partial_utf8` dropped the tail. The arm is now `target != SQL_C_BINARY && matches!(encoding, PlpEncoding::Utf8Text)`, letting binary fall through to the raw byte copy.

Pinned by `ABoundBinaryReadOfAUtf8ColumnKeepsTheTruncatedTail`, a bound 9-byte read of a JSON column of 2-byte characters. I verified the test has teeth by reintroducing the bug: byte 8 came back `0x00` instead of `0xC3` and the suite reported it as a divergence, since msodbcsql does not trim.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -2127,9 +2135,25 @@ unsafe fn deliver_bound(
         };
     }
 
+    if binding.target_type == SQL_C_BINARY {
+        // Binary is not a string: no terminator, and the indicator carries the
+        // untruncated byte count exactly as the character targets do.
+        let Some(bytes) = column_value_to_bytes(value) else {
+            return RowOutcome::Error(RowIssue::Unsupported);
+        };
+        unsafe { write_if_some(indicator, bytes.len() as SqlLen) };
```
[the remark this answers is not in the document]
David-Engel: Correct, fixed in 28875e73 — it now writes `octet_length`. Every other delivery in `deliver_bound` (`convert_typed_c`, the WCHAR and CHAR branches, the fixed-width path) and `deliver_bound_plp` uses `octet_length`; the binary branch was the only one on `indicator`.

Added `bound_binary_reports_its_length_on_octet_length_not_the_indicator`, which builds a `ColumnBinding` with the two pointers at distinct arrays — a `SQLBindCol`-based test cannot tell them apart, which is why this went unnoticed.

mssql-odbc/src/api/get_data.rs
```diff
@@ -993,10 +993,9 @@ fn finish_get_data(
 /// Refusing costs nothing an application had before: the same call answered
 /// `HYC00` prior to AB#47815.
 ///
-/// A `varbinary` / `image` column resolves to `SQL_C_BINARY`, which neither
-/// delivery path implements yet (AB#47239), so a real read keeps returning
-/// `HYC00` through the resolved target. That matches the posture the bound path
-/// took in AB#47481; msodbcsql resolves identically and delivers the bytes.
+/// A `varbinary` / `image` column resolves to `SQL_C_BINARY`, which both
+/// delivery paths now implement (AB#47239), so a real read through the resolved
+/// target returns the bytes, matching msodbcsql.
```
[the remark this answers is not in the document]
David-Engel: Right, the bullets contradicted the paragraph above them. Rewritten in 28875e73 to say the probe and the read both work on both paths, and to state the actual limit: only kinds whose byte form is fixed by the value alone are delivered.

mssql-odbc/src/api/get_data.rs
```diff
@@ -2510,10 +2553,33 @@ pub(crate) enum TextError {
...
+/// is measured (AB#47239 follow-up).
+pub(crate) fn column_value_to_bytes(value: &ColumnValues) -> Option<&[u8]> {
+    match value {
+        ColumnValues::Bytes(b) => Some(b),
+        // The wire bytes, which is what msodbcsql delivers here: UTF-16LE for
+        // the wide types and codepage bytes for the narrow ones.
+        ColumnValues::String(s) => Some(&s.bytes),
+        ColumnValues::Xml(x) => Some(&x.bytes),
+        ColumnValues::Json(j) => Some(&j.bytes),
+        _ => None,
```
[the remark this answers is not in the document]
David-Engel: The PR description was wrong, not the code — thank you for catching it. Those types are what *msodbcsql* delivers; I conflated the measured reference behaviour with what this change implements.

I checked the scope against the work item rather than guessing. AB#47239 is a restore of binary delivery removed in `5de38575`, and the P1 `column_value_to_bytes` it replaces handled `Bytes`/`Xml`/`Json` only. This implementation is a superset (it adds the character types), so the restore is complete as scoped, and the fixed-width conversions were never part of it.

I have narrowed the description to the byte-backed variants and kept the measured msodbcsql table as an explicit follow-up, since `date` gives a `SQL_DATE_STRUCT`, `datetime2` a `SQL_TIMESTAMP_STRUCT`, and `decimal` reports `22003` — inconsistent enough that each deserves its own decision. `FixedWidthKindsAreStillRefused` pins the boundary.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
```diff
@@ -1098,11 +1098,10 @@ TEST_F(FetchScrollLiveTest, ABoundNvarcharMaxDoesNotSplitASurrogatePair) {
...
-    SKIP_IF_COMPARING_MSODBCSQL();
-    // Two rows and a trailing scalar: the refused target takes the drain path
-    // rather than the fill loop, so proving the row ended is not enough --
-    // the value after it, and the row after that, have to decode correctly.
+TEST_F(FetchScrollLiveTest, ABoundVarbinaryMaxDeliversAcrossARowset) {
+    // Two rows and a trailing scalar: a bound LOB is drained into the caller's
+    // buffer, and the value after it -- and the row after that -- still have to
+    // decode, which is what a mis-sized drain would break.
     ExecDirect(
         "SELECT n, REPLICATE(CAST(0x41 AS VARBINARY(MAX)), 5000) AS lob, n * 11 AS tail "
```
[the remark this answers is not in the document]
David-Engel: Good catch — the value was materialized and delivered by `deliver_bound`, so `deliver_bound_plp` was untested. Raised to 1,100,000 bytes in 28875e73, and the test now checks all 32 bytes of the slot rather than just the first, which is what pins capacity at `stride` instead of `stride - 1`.

David-Engel (commented):
Reviewed all seven changed files at 28875e73, AB#47239, and the prior review discussions. The earlier findings are addressed, and retaining the fixed-width scalar limitation matches the work item's scope. One correctness issue to fix before merge and one test suggestion are detailed inline.

Validation: ran a focused read-only comparison against the PR build and msodbcsql 18.06.0001; did not rerun the full suite. The reference measurement is not against CI's pinned 18.6.2.1 build. At review time, CI had no reported failures but several platform checks were still in progress; the existing 96% diff-coverage report came from an earlier build.

mssql-odbc/src/api/get_data.rs
```diff
@@ -2555,6 +2612,88 @@ fn binary_length(value: &ColumnValues) -> SqlLen {
...
+    }
+    unsafe {
+        write_if_some(
+            strlen_or_ind_ptr,
+            SqlLen::try_from(remaining).unwrap_or(SqlLen::MAX),
+        )
+    };
+
+    if take < remaining {
+        stmt_state.partial_text_offset = Some((col_index, offset + take));
```
David-Engel, line 2688: **Blocking [P2]: Account for the binary offset in subsequent length probes.**

This new partial-read offset is honored by later data reads, but the zero-length probe in `write_captured_column` still uses `binary_length(value)` without subtracting it. On `SELECT CAST(0x010203040506070809 AS BINARY(9))`, a 4-byte `SQL_C_BINARY` read followed by a zero-length probe reports indicators **9, 9** on this PR, versus **9, 5** on msodbcsql (`SQL_DRIVER_VER = 18.06.0001`). The next 4-byte read reports 5 on both, and this PR's `varbinary(max)` probe already reports 5 correctly.

That overstates what remains for callers that probe between chunks. Please make the captured-value probe offset-aware for the current column's binary continuation, preserving `SQL_NO_TOTAL` and existing scalar-probe behavior. Add a partial-read -> probe -> resumed-read regression test that checks the remaining count and confirms the probe consumes no bytes.

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
```diff
@@ -0,0 +1,276 @@
...
+    SQLCloseCursor(stmt_);
+}
+
+// ---------------------------------------------------------------------------
+// PLP: varbinary(max) and the UDT types, which arrive as a wire stream rather
+// than a materialized value.
+// ---------------------------------------------------------------------------
+
+TEST_F(BinaryFetchLiveTest, VarbinaryMaxDeliversWholeValue) {
+    FetchOne("SELECT REPLICATE(CAST(0x41 AS VARBINARY(MAX)), 10)");
```
David-Engel: **Suggestion: Cast the result of `REPLICATE` to `VARBINARY(MAX)`.**

SQL Server converts the binary argument to character data inside `REPLICATE`, so this query returns `varchar(max)`, not `varbinary(max)`. I measured `SQL_DESC_CONCISE_TYPE = 12` (`SQL_VARCHAR`) on both the PR build and msodbcsql 18.06.0001. The same expression appears in `VarbinaryMaxChunksAcrossCalls`, `BoundVarbinaryMaxDeliversWholeValue`, and `ABoundVarbinaryMaxDeliversAcrossARowset` in `fetch_scroll_test.cpp`. These still exercise binary-target delivery, but not the binary-source metadata their names imply.

Cast the whole result, for example `SELECT CAST(REPLICATE(CAST(0x41 AS VARBINARY(MAX)), 10) AS VARBINARY(MAX))`, and assert the resulting SQL type so implicit conversion cannot silently change what the tests cover.

mssql-odbc/src/api/get_data.rs
David-Engel: Updated captured binary probes to report the remaining byte count from the current continuation offset while preserving `SQL_NO_TOTAL` and fixed-width scalar probe behavior. Added both an in-process regression and an E2E partial-read → probe → resumed-read test that verifies the probe does not consume bytes.

> [!NOTE]

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
David-Engel: Cast each affected `REPLICATE` result to `VARBINARY(MAX)` and added `SQLDescribeCol` assertions for `SQL_VARBINARY`, including the rowset streaming case.

> [!NOTE]

David-Engel:
Picked up `5604bdb4` and kept it — the `remaining_binary_length` guard is better than what I had written independently. Mine subtracted whenever the length was non-negative, which would have mis-reported a scalar probe following a `SQL_C_CHAR` partial read on the same column; declining to subtract unless `column_value_to_bytes` is `Some` avoids that. Three follow-ups pushed on top in `630257d9`:

**The two new probe tests were passing a null `TargetValuePtr`.** The Driver Manager rejects that with `HY009` before the call reaches a driver, so `FixedBinaryProbeReportsRemainingBytesWithoutConsumingThem` failed on msodbcsql *and* on this driver — a shared failure rather than a divergence, which meant the remaining-count assertions never actually ran. Both now pass a real one-byte buffer at length 0, the way the existing probe tests in `get_data_test.cpp` do. With that corrected they exercise the contract and the suite is 40 parity / 0 divergences.

**Restored `answer_binary_probe`'s doc block.** My earlier commit inserted `deliver_captured_binary` between that doc and the function it documents, so the one warning standing against the two probe sites drifting had ended up on the delivery routine instead.

**Recorded why the buffered fast path deliberately does not subtract the offset.** I started to "fix" it there for symmetry and it would have been wrong: a binary continuation keeps its value in `last_captured`, so it takes the `already_captured` branch and never reaches that site, and any offset still set for the column there belongs to a *text* continuation and is not in bytes. Left the code alone and wrote the reasoning down, since the symmetry argument points the other way.

Also added the PLP counterpart to the probe test, which derives its remainder from the stream rather than from `remaining_binary_length`.

Verified: 40 parity / 0 divergences, 1393 unit tests, clippy and fmt clean, mssql-python suites unchanged.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
```diff
@@ -1098,14 +1098,22 @@ TEST_F(FetchScrollLiveTest, ABoundNvarcharMaxDoesNotSplitASurrogatePair) {
 // Bound binary delivery is unimplemented for every type, not just the max ones
 // (AB#47239), so this asserts our own answer rather than parity -- msodbcsql
 // delivers it.
-TEST_F(FetchScrollLiveTest, ABoundVarbinaryMaxIsStillUnsupported) {
-    SKIP_IF_COMPARING_MSODBCSQL();
-    // Two rows and a trailing scalar: the refused target takes the drain path
-    // rather than the fill loop, so proving the row ended is not enough --
-    // the value after it, and the row after that, have to decode correctly.
+TEST_F(FetchScrollLiveTest, ABoundVarbinaryMaxDeliversAcrossARowset) {
```
saurabh500, line 1100: Nit: the three-line comment directly above this renamed test still reads "Bound binary delivery is unimplemented for every type, not just the max ones (AB#47239) ... msodbcsql delivers it." The test right below it now asserts this driver delivers the bytes too — byte-for-byte across the rowset — so the comment contradicts what it documents. Suggest replacing it with something like: "Bound VARBINARY(MAX) delivery across a rowset with a trailing scalar, so a mis-sized drain would corrupt the following column and row."

saurabh500 (commented):
_This review was produced by an unattended PR-review sweep (event: comment only) — no human has eyeballed it yet._

**Verdict:** No blocking findings. The SQL_C_BINARY delivery reads as correct and mirrors the existing text/PLP chunking model faithfully: the offset/length bookkeeping in `remaining_binary_length`/`binary_length`, the deliberate choice not to set `current_row_last_col` on a truncated binary read (so the column stays re-readable through the `already_captured` branch), and the two gates opened in `stream_active_plp_chunk` all line up with the char path they copy. I mutated `remaining_binary_length` to ignore the partial-read offset and the two binary probe unit tests went red, so the new unit coverage has teeth. Cross-platform CI (including the msodbcsql parity legs and mssql-python) is green; I could not run the live or parity legs locally, so those rest on CI. Everything below is test-quality only.

**Blocking:** none.

**Suggestion:**

1. `BinaryColumnBesideALobStaysWithinItsBuffer` (mssql-odbc/tests/e2e/tests/get_data_test.cpp, ~line 1863) is the test purpose-built for the `GetDataVar` probe-then-grow path this PR fixes — a fixed-width `BINARY(9)` beside an `nvarchar(max)` LOB — but after this change it no longer pins the new behavior. Its step-2 real read keeps the pre-fix `else` branch:
   ```cpp
   if (SQL_SUCCEEDED(read_rc)) {
       EXPECT_LE(read_ind, static_cast<SQLLEN>(buf.size()));
       EXPECT_EQ(9, read_ind);
   } else {
       // Binary *data* delivery is still AB#47239, so a refusal is expected today.
       EXPECT_EQ(SQL_ERROR, read_rc);
   }
   ```
   Two gaps on the exact path this PR touches: (a) a regression back to `HYC00`/`SQL_ERROR` here would take the `else` branch and still pass, so the test would stay green through the very failure it is named for; (b) the success branch never checks the delivered bytes, so a driver reporting `read_ind == 9` with wrong or untouched buffer contents also passes. Since delivery is now the contract on this path, drop the `else` and verify the payload:
   ```cpp
   ASSERT_TRUE(SQL_SUCCEEDED(read_rc)) << "rc=" << read_rc;
   EXPECT_EQ(9, read_ind);
   EXPECT_EQ(0, std::memcmp(buf.data(), "\x01\x02\x03\x04\x05\x06\x07\x08\x09", 9));
   ```
   This block sits outside every diff hunk, so it is here rather than inline.

**Nit:**

1. Stale comment at mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp:1098-1100 — flagged inline.

2. mssql-odbc/tests/e2e/tests/get_data_test.cpp:1598-1600 and 1617-1619 — both NULL-leg tests still carry "binary *data* delivery is still unimplemented (AB#47239), which is why only the NULL case is covered here." Delivery is implemented by this PR and now covered by the new `...DeliversTheBytes` tests, so the "only the NULL case, because delivery is unimplemented" rationale is stale. Suggest rewording to say these cover the NULL leg and pointing at the delivery tests for the rest. Both are outside diff hunks, hence here. The identical stale line inside the `else` branch at ~1868 is covered by the Suggestion above.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
David-Engel: Updated the comment to describe the bound `VARBINARY(MAX)` rowset delivery and trailing-scalar integrity that the test now covers. I also strengthened `BinaryColumnBesideALobStaysWithinItsBuffer` to require successful delivery and assert all nine payload bytes, and refreshed the related stale NULL-leg comments.

> [!NOTE]

saurabh500 (commented):
## Verdict: approve — no blocking findings

Restores `SQL_C_BINARY` delivery on all four paths (materialized `SQLGetData`, PLP `SQLGetData`, bound non-PLP, bound PLP). I checked out the branch and reviewed against the merge base `7ecdef4`; everything I traced holds up.

**What I verified**

- **Buffer bounds on all four paths.** `deliver_bound` (`bytes.len().min(stride)` into a `stride`-byte slot), `deliver_bound_plp` (`out_bytes` capped at `capacity_elements`), `deliver_captured_binary` (`remaining.min(capacity)`), and the PLP `direct_wire_output` write (bounded by `max_read <= buffer_length`). No path can overrun.
- **Binary capacity is the full stride, not `stride - 1`.** This is the detail most likely to be wrong, and `ABoundVarbinaryMaxDeliversAcrossARowset` pins it by asserting all 32 bytes of the slot are payload — a terminator would cost the last one.
- **Indicator semantics.** The per-target `write_if_some(strlen_or_ind_ptr, read)` in the binary branch looked inconsistent at first, but it is overwritten by `remaining_indicator` on the truncation path; the only path where `read` survives is the `reached_end && !converted_data_still_held` early return, where `read == remaining`. Consistent with the measured "bytes remaining before the call" contract.
- **`trim_partial_utf8` exclusion.** `target != SQL_C_BINARY &&` is the right gate — binary has no character boundaries to preserve. Good call verifying `ABoundBinaryReadOfAUtf8ColumnKeepsTheTruncatedTail` fails as a parity divergence if the trim is reintroduced; that is what makes the test worth having.
- **No spurious `HY090`.** `terminator_bytes = 0` for binary makes `is_length_probe` and `makes_no_progress` coincide at `buffer_length == 0`, so the probe shape reaches the probe branch rather than the error.
- **Both PLP compatibility gates** (first chunk vs. resumed) opened consistently; the resumed gate's `Some(_)` correctly still refuses the no-active-stream case.
- **`deliver_captured_binary` discarding the captured column index** is safe — `write_captured_column` is only reachable via the `already_captured` branch, which already checked `*c == col_index`.
- **Skipping `finish_get_data` on the probe-truncation return** is equivalent to calling it: `current_row_last_col` is deliberately not advanced there, so `ready` would be false anyway.
- **`SqlString.bytes` is the raw wire form**, so `column_value_to_bytes` handing UTF-16LE to a `SQL_C_BINARY` read of an `nvarchar` column is correct, not a transcoding gap.
- **Scope.** Limiting to kinds whose byte form is fixed by the value alone (`Bytes`, `String`, `Xml`, `Json`) is the right line to draw, and `FixedWidthKindsAreStillRefused` pins the other side of it so the remaining `HYC00` cases can't drift silently.

**What I ran**

`cargo nextest run -p mssqlodbc --no-fail-fast` — 1409 tests, all pass. CI is green on `1dd0330` across the full matrix (Windows/Linux/macOS/ARM, Kerberos, the cross-repo mssql-python builds), with 100% diff coverage.

I did not re-raise the earlier Copilot findings or the stale-comment nit — all confirmed fixed in `28875e7` and the follow-ups. One non-blocking note inline.

Leaving this as a comment rather than a formal approval so the approving click stays yours.

mssql-odbc/src/api/get_data.rs
```diff
@@ -2220,6 +2255,16 @@ fn stream_active_plp_chunk<'a>(
                 "UTF-16 to UTF-8 transcode made no forward progress"
             );
         };
+    } else if target_type == SQL_C_BINARY {
+        // Binary delivery is a straight byte copy with no terminator, whatever
+        // the column's encoding. Every binary read that moves bytes qualifies for
+        // `direct_wire_output` -- the gates above guarantee `plp_encoding` is
+        // `Some` here, so it is false only when there is no room or no buffer,
+        // and then `read` is 0 -- meaning the payload already landed in the
+        // caller's buffer and only the length is left to report.
```
saurabh500: **Suggestion** — the justification here is half true, and the false half is the one that would hurt later.

`direct_wire_output` is false in two situations, and the comment claims `read` is 0 in both. For binary, `terminator_bytes` is 0, so `payload_capacity == buffer_length`, and `max_read` falls to the plain `payload_capacity` arm (`widen_narrow_to_utf16` and `transcode_utf16_to_utf8` are both false). `max_read` therefore never looks at `target_value_ptr`:

- **No room** (`buffer_length == 0`) → `max_read == 0` → `read == 0`. The claim holds.
- **No buffer** (`target_value_ptr` null, `buffer_length > 0`) → `max_read > 0`, so the read lands in `inline_payload`/`heap_payload` and `read` can be non-zero. This branch then reports `read` on the indicator and copies it nowhere — bytes consumed off the PLP stream and dropped, with `SQL_SUCCESS_WITH_INFO` claiming they were delivered.

Not reachable today: the Driver Manager rejects a null `TargetValuePtr` with `HY009` before the call arrives, which is exactly why the e2e probes pass a 1-byte buffer. So this is not a live defect and I would not hold the PR for it.

The reason I am raising it at all is that it is a stated invariant rather than a stylistic point, sitting next to an `unsafe` block, and the text arm below does not depend on it — `copy_verbatim` routes the non-direct case through `copy_with_nul`, so it degrades quietly where this arm degrades silently. If the DM check ever stops being the thing standing in front of this, the comment says the case cannot happen.

Cheapest fix is to stop claiming the second case and assert what actually has to hold:

```rust
} else if target_type == SQL_C_BINARY {
    // Binary delivery is a straight byte copy with no terminator, whatever
    // the column's encoding. The gates above guarantee `plp_encoding` is
    // `Some`, so `direct_wire_output` is false only for a zero-length buffer
    // (`read == 0`) or a null `target_value_ptr`, which the Driver Manager
    // rejects with HY009 before the call reaches the driver. Either way the
    // payload already landed in the caller's buffer and only the length is
    // left to report.
    debug_assert!(
        direct_wire_output || read == 0,
        "binary chunk read {read} bytes without a caller buffer to put them in"
    );
    unsafe {
        write_if_some(strlen_or_ind_ptr, read as SqlLen);
    }
}
```

That keeps the fast path identical and turns a silent drop into a debug failure if the assumption ever stops holding.

David-Engel: Corrected the comment so it no longer claims a null caller buffer forces `read == 0`, and added the proposed debug assertion to catch any future path that reads binary PLP bytes without a destination buffer.

> [!NOTE]

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    unsigned char buf[16] = {};
+    SQLLEN ind = 0;
+    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_BINARY, buf, sizeof(buf), &ind), SQL_HANDLE_STMT,
+                  stmt_);
+    EXPECT_EQ(9, ind);
+    const unsigned char expected[9] = {1, 2, 3, 4, 5, 6, 7, 8, 9};
+    EXPECT_EQ(0, std::memcmp(buf, expected, sizeof(expected)));
+    SQLCloseCursor(stmt_);
+}
```
Theekshna, line 59: No test payload here uses byte value `0x00`. Binary sits next to nul-terminated `SQL_C_CHAR`/`SQL_C_WCHAR` delivery (`copy_with_nul`) -- a regression that treats binary as a C string would stop at the first zero and still pass everything here. Add a case like `CAST(0x01000203 AS BINARY(4))` asserting all 4 bytes including the zero.

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    unsigned char buf[8];
+    std::memset(buf, 0xEE, sizeof(buf));
+    SQLLEN ind = -1;
+    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_BINARY, buf, sizeof(buf), &ind), SQL_HANDLE_STMT,
+                  stmt_);
+    EXPECT_EQ(0, ind);
+    EXPECT_EQ(0xEE, buf[0]) << "an empty value must not disturb the buffer";
+    SQLCloseCursor(stmt_);
+}
```
Theekshna, line 160: Only the PLP NULL leg is covered (`NullVarbinaryMaxToBinaryTargetReportsNull`). No NULL case for a non-PLP `binary(n)`/`varbinary(n)` read as `SQL_C_BINARY` -- different code path in `write_captured_column`. Add `CAST(NULL AS VARBINARY(8))` asserting `SQL_NULL_DATA`.

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
```diff
@@ -0,0 +1,336 @@
...
+    AssertColumnSqlType(SQL_VARBINARY);
+
+    unsigned char buf[64] = {};
+    SQLLEN ind = 0;
+    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_BINARY, buf, sizeof(buf), &ind), SQL_HANDLE_STMT,
+                  stmt_);
+    EXPECT_EQ(10, ind);
+    for (int i = 0; i < 10; ++i) EXPECT_EQ(0x41, buf[i]) << "byte " << i;
+    SQLCloseCursor(stmt_);
+}
```
Theekshna, line 191: All `VarbinaryMax*` tests use a non-empty value. An empty PLP stream (`CAST('' AS VARBINARY(MAX))`) is a different wire shape than PLP-NULL or a non-empty stream -- the empty-first-chunk branch in `stream_active_plp_chunk` for `SQL_C_BINARY` is untested. Add a real (nonzero-buffer) read of it.

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
```diff
@@ -0,0 +1,336 @@
...
+    unsigned char buf[4] = {};
+    SQLLEN ind = 0;
+    ASSERT_SQL_OK(SQLBindCol(stmt_, 1, SQL_C_BINARY, buf, sizeof(buf), &ind), SQL_HANDLE_STMT,
+                  stmt_);
+    EXPECT_EQ(SQL_SUCCESS_WITH_INFO, SQLFetch(stmt_));
+    EXPECT_EQ(9, ind) << "the untruncated length, as the character targets report";
+    EXPECT_EQ(0, std::memcmp(buf, "\x01\x02\x03\x04", 4));
+    SQLFreeStmt(stmt_, SQL_UNBIND);
+    SQLCloseCursor(stmt_);
+}
```
Theekshna, line 305: No `Bound*` test here uses a NULL value. `deliver_bound`'s and `deliver_bound_plp`'s NULL branches are unexercised for `SQL_C_BINARY`. Add a bound NULL case for both non-PLP and PLP (varbinary(max)).

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
```diff
@@ -0,0 +1,336 @@
...
+    unsigned char buf[64] = {};
+    SQLLEN ind = 0;
+    ASSERT_SQL_OK(SQLBindCol(stmt_, 1, SQL_C_BINARY, buf, sizeof(buf), &ind), SQL_HANDLE_STMT,
+                  stmt_);
+    ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);
+    EXPECT_EQ(10, ind);
+    for (int i = 0; i < 10; ++i) EXPECT_EQ(0x41, buf[i]) << "byte " << i;
+    SQLFreeStmt(stmt_, SQL_UNBIND);
+    SQLCloseCursor(stmt_);
+}
```
Theekshna, line 322: Two untested cases for bound binary buffer bounds:
1. Two adjacent bound `SQL_C_BINARY` columns in one row -- only binary-beside-LOB-text is tested; a wrong stride here corrupts a neighboring binary slot instead of tripping a text-decode error.
2. `SQL_ATTR_ROW_ARRAY_SIZE > 1` with bound `SQL_C_BINARY` -- array binding is tested for other C types, and binary is tested across separate `SQLFetch` calls, but never both together. The new `row_index * stride` math is untested with `row_index > 0`.

Theekshna (commented):
Reviewed the delivery paths and prior review threads — no blocking issues; all earlier findings from Copilot bot and saurabh500 are addressed with tests that have teeth. Test coverage is solid for the paths it targets, but comparing against edge cases the code itself opens up (embedded zero bytes, NULL legs, array binding), a few gaps remain:

| # | Scenario | Gap |
|---|----------|-----|
| 1 | Embedded `0x00` byte in a binary value | No payload uses byte value 0; a nul-termination regression from the adjacent text path would go unnoticed |
| 2 | NULL on non-PLP `binary(n)`/`varbinary(n)` via `SQL_C_BINARY` | Only the PLP NULL leg is tested |
| 3 | Real read of an empty `VARBINARY(MAX)` (PLP) | Only empty non-PLP and PLP-NULL are tested; empty-PLP is a distinct wire shape |
| 4 | NULL on a *bound* `SQL_C_BINARY` column | No bound test uses NULL, non-PLP or PLP |
| 5 | Two adjacent bound `SQL_C_BINARY` columns | Only binary-beside-LOB-text is tested |
| 6 | `SQL_ATTR_ROW_ARRAY_SIZE > 1` with bound `SQL_C_BINARY` | New `row_index * stride` math untested with `row_index > 0` |

Inline comments below point at scenarios 1-6 individually.

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
David-Engel: Added adjacent bound binary-column coverage with separate four-byte slots, plus a two-row column-wise array binding that verifies the second row lands at the binary stride and both payloads remain exact.

> [!NOTE]

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
David-Engel: Added a real nonzero-buffer read of an empty `VARBINARY(MAX)` value, asserting a zero indicator and an untouched destination buffer.

> [!NOTE]

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
David-Engel: Added bound NULL coverage for both `VARBINARY(8)` and `VARBINARY(MAX)` in one fetch, verifying both indicators and untouched buffers.

> [!NOTE]

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
David-Engel: Added a non-PLP `VARBINARY(8)` NULL read that verifies `SQL_NULL_DATA` and confirms the binary buffer remains untouched.

> [!NOTE]

mssql-odbc/tests/e2e/tests/binary_fetch_test.cpp
David-Engel: Added `FixedBinaryPreservesEmbeddedZero`, which reads `0x01000203` as `SQL_C_BINARY` and verifies all four bytes including the embedded zero.

> [!NOTE]

Theekshna:
Tests are failing

David-Engel:
> Tests are failing

Fixed the binary-fetch failure in 8a585c3e. The first PLP binary read now stages through temporary storage so an empty VARBINARY(MAX) reports length zero without overwriting the caller's buffer; later chunks still stream directly into the caller buffer.

Theekshna (approved):

PR #530, opened 2026-09-09
Render binary columns as hex for character targets (AB#47240)

David-Engel:
Resolves [AB#47240](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47240). The work item's title and scope now explicitly include bound fetch as well as `SQLGetData`.

## Problem

ODBC Appendix D lists binary-to-character as a mandatory conversion, and msodbcsql renders binary columns as upper-case hex. `ColumnValues::Bytes` had no arm in `column_value_to_text`, so `binary` / `varbinary` / `image` read as `SQL_C_CHAR` or `SQL_C_WCHAR` returned `HYC00`.

## What changed

| Path | Change |
| --- | --- |
| Materialized `SQLGetData` | `ColumnValues::Bytes` renders upper-case hex without a `0x` prefix; chunk capacity preserves whole hex pairs |
| Streamed `SQLGetData` (`varbinary(max)`, UDTs) | Compatibility gates admit binary-to-character conversion, reads are sized for hex expansion, and wide hex bypasses raw UTF-16 delivery |
| Bound fetch (`SQLBindCol` + `SQLFetch` / `SQLFetchScroll`) | Both `deliver_bound` and `deliver_bound_plp` render hex, preserve complete pairs, and report the converted byte count |

The work item scoped the max-length half as overlapping [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239). That machinery is now in place, so both halves are here: `varbinary(max)`, `image`, `hierarchyid`, `geometry` and `geography` all render.

Review follow-up `a8840251` fixes three edge cases: buffered bound outputs no longer split a hex pair; tiny streamed hex buffers return `01004` without consuming input instead of `HY090`; and `SQL_C_WCHAR` binary streams reach hex conversion rather than the raw UTF-16 branch. The regressions cover narrow/wide targets, zero-length and undersized probes, odd buffer sizes, exact fits, chunk continuation, guard bytes, and draining a truncated bound value before the following column.

`128ccf49` adds measured non-PLP zero-length probes, bounds the test readers so missing termination causes an assertion rather than an out-of-bounds read, and removes redundant capacity rounding from the bound streaming loop.

## Whole-pair truncation

A byte's two hex characters are atomic. msodbcsql fills whole pairs and leaves an odd trailing payload slot empty rather than splitting a pair.

| `SQL_C_CHAR` buffer length | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| Characters written | 0 | 0 | 2 | 2 | 4 | 4 | 6 |

Handing the hex string to an unrestricted character writer would produce `"010"` at buffer length 4. Materialized `SQLGetData` and buffered bound delivery therefore use `hex_buffer_elements`, which reserves the terminator and rounds payload capacity down to complete pairs. Bound PLP delivery appends both digits only when `filled + 2` fits. Streamed `SQLGetData` reads at most half (`SQL_C_CHAR`) or a quarter (`SQL_C_WCHAR`) of the available payload capacity in wire bytes. A pair cannot straddle two chunks.

Length indicators are byte counts in the target representation: two bytes per wire byte for `SQL_C_CHAR`, four for `SQL_C_WCHAR`. Truncated calls report the remaining count before that call; completing calls report the bytes delivered. Zero-length output buffers remain untouched; a nonempty value reports `01004` and remains available for a subsequent read.

## Measured contract

The 23 `bin_to_char_test.cpp` cases pass against the Rust driver and the installed retail msodbcsql 18.6.1.1 (`SQL_DRIVER_VER=18.06.0001`). CI pins msodbcsql 18.6.2.1 for its comparison legs. The cases include:

- `0x0102AB` becomes `"0102AB"`, indicator 6; `SQL_C_WCHAR` gives the same text with indicator 12.
- Chunked at 4 narrow bytes: `01` / `02` / `AB` / `40`, indicators 8, 6, 4, 2, then `SQL_NO_DATA`. Wide chunks report twice those byte counts.
- Empty binary becomes `""`, indicator 0. Repeated zero-length probes on nonempty `binary(3)` and `varbinary(8)` return `01004` with indicators 6/12 for CHAR/WCHAR, preserve sentinel-filled buffers, and leave the whole value readable. PLP probes cover the same behavior and undersized buffers.
- A 1.1 MB `varbinary(max)` reports 2,200,000 narrow bytes or 4,400,000 wide bytes, including bound delivery.
- Bound `binary(3)`, `varbinary(8)` and `varbinary(max)` preserve whole pairs across buffer sizes 0-7 in both target widths.
- `hierarchyid` renders `"58"`; `geometry` renders its 22 wire bytes as 44 characters. `uniqueidentifier` keeps its GUID string form.
- `geography::Point(47.651, -122.349, 4326)` renders `"E6100000010C17D9CEF753D347407593180456965EC0"`, indicator 44. The geography regression was added in `3b882725` after measuring the reference driver.

## Tests anchored on the old refusal

- `UnsupportedColumnTypeHyc00PreservesValue` becomes `UnsupportedTargetHyc00PreservesValue`, using an unsupported **target** (`SQL_C_NUMERIC`). With binary converting, no non-PLP column type remains unconvertible in this helper.
- `VarbinaryMaxToCharReturnsHyc00` is removed; `bin_to_char_test` covers that column positively.
- Adding the `Bytes` arm makes `column_value_to_text` exhaustive, so its unreachable catch-all is gone.
- `ab922220` adds standalone unit tests for `hex_buffer_elements` (0-8 and `usize::MAX` boundaries), `hex_pair` (all 256 byte values), and `bytes_to_hex` (empty input, byte order, zero padding, upper-case output, no prefix). These require no SQL Server.

## Deliberately not addressed

A NULL binary column read as characters writes a terminator where msodbcsql leaves the caller's buffer untouched. That predates this change: a NULL never enters the conversion. The deferred gap is tracked in #555; it is not an approved permanent deviation. `NullVarbinaryMaxReportsNull` deliberately runs without a comparison skip because it asserts only the indicator both drivers agree on.

The missing `SQLGetInfo` conversion masks, including `SQL_CONVERT_BINARY`, `SQL_CONVERT_VARBINARY` and `SQL_CONVERT_LONGVARBINARY`, are already explicitly tracked in [AB#47996](https://dev.azure.com/SqlClientDrivers/mssql-rs/_workitems/edit/47996). Completing conversion discovery is outside this PR.

## Verification

- `128ccf49` local runs: **159 targeted Rust unit tests and 149 e2e cases passed** (23 binary-to-character, 76 SQLGetData, 50 bound-fetch). All 23 binary-to-character cases also passed against retail msodbcsql 18.6.1.1. Workspace formatting and scoped clippy passed. The focused local coverage report covers **98.9% of changed lines**, above the last completed CI overall coverage of 93.8%.
- `32d43f08`, after integrating #533 from `main`: **169 targeted Rust unit tests and 147 e2e cases passed** (21 binary-to-character, 76 SQLGetData, 50 bound-fetch), covering both sides of the resolved hex/collation conversion merge.
- Completed CI evidence for `3b882725`: [PR validation build 174075](https://dev.azure.com/SqlClientDrivers/public/_build/results?buildId=174075) passed its reported checks. Its [coverage report](https://dev.azure.com/SqlClientDrivers/public/_build/results?buildId=174075&view=codecoverage-tab) shows **100% diff coverage (127/127 changed lines)** versus **93.8% overall**. CI reruns for `128ccf49`; these completed CI numbers are explicitly from `3b882725`.


mssql-odbc/src/api/get_data.rs
```diff
@@ -1667,6 +1679,8 @@ fn stream_active_plp_chunk<'a>(
             // the encoding. The two must agree or a chunked read is admitted on
             // its first call and refused on its second.
             (SQL_C_BINARY, _) => true,
+            // A binary column read as characters is rendered as hex (AB#47240).
+            (SQL_C_CHAR | SQL_C_WCHAR, PlpEncoding::Binary) => true,
```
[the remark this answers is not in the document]
David-Engel: Real bug, fixed in `bb450f93`. `deliver_bound_plp` now admits `(SQL_C_CHAR | SQL_C_WCHAR, PlpEncoding::Binary)` and emits hex.

One correction to the diagnosis, from measuring it: the bound path returned `HYC00` for a **10-byte** value as well as a 1.1 MB one, so a bound `varbinary(max)` reaches `deliver_bound_plp` whatever its size rather than only once it grows large enough to stream. The consequence you identified is the same and is the real defect — the two delivery paths disagreed about the same column — but it wasn't size-dependent in the way described, so I widened the test to pin both sizes rather than just the streaming one.

Three details worth flagging for review:

- The hex arm has to sit **ahead of** the `target == SQL_C_WCHAR` arm in the accumulator chain, which would otherwise claim it and emit the raw UTF-16 path.
- `capacity_elements` rounds down to an even count, so a byte's two characters are never split — the same pair-atomicity rule the `SQLGetData` side follows.
- The reported total is scaled to hex units by scaling `wire_total` before `plp_indicator`, which keeps that function's existing three-case logic intact.

`hex_pair` is now shared with `get_data` rather than duplicating a digit table.

`BoundVarbinaryMaxRendersHexWhetherBufferedOrStreamed` covers both sizes and **passed on msodbcsql before the fix**, so it is a genuine divergence test rather than a restatement of our own behaviour.

mssql-odbc/src/api/get_data.rs
```diff
@@ -2880,6 +2944,18 @@ fn strip_sub_one_leading_zero(s: String) -> String {
     }
 }
 
+/// Upper-case hex, two characters per byte and no `0x` prefix, matching what
+/// msodbcsql renders for a binary column read as characters.
+fn bytes_to_hex(bytes: &[u8]) -> String {
```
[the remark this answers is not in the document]
David-Engel: Updated in `bb450f93`. The P1 bullet now records that `SQL_C_BINARY` was restored in AB#47239 and that binary → character was added in AB#47240 on both the `SQLGetData` and bound-fetch paths, and it points at the renamed test (`GetDataLiveTest.UnsupportedTargetHyc00PreservesValue`) instead of describing the old `VARBINARY(8)` anchor as an outstanding dependency.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -809,22 +809,6 @@ TEST_F(GetDataLiveTest, UnsupportedCTypeReturnsHyc00ThenValueReadable) {
 // HYC00 rather than corrupt the stream. The reference msodbcsql driver supports
 // binary-to-char (hex) conversion, so this is mssql-odbc-specific — skip it on
 // the msodbcsql comparison leg.
```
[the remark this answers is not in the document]
David-Engel: Deleted in `bb450f93` — the four stranded lines are gone and the PLP test below now carries only its own header.

mssql-odbc/tests/e2e/tests/get_data_test.cpp
```diff
@@ -1934,25 +1918,28 @@ TEST_F(GetDataLiveTest, Datetime2ToTimestampTargetKeepsFraction) {
 // non-PLP type with no character rendering. If binary→hex is ever implemented,
 // re-point this again, or assert the recovery via the target-type HYC00 path (an
 // unsupported SQL_C target) with a type that will stay unsupported.
-TEST_F(GetDataLiveTest, UnsupportedColumnTypeHyc00PreservesValue) {
+// Anchored on an unsupported *target* rather than an unsupported column type.
+// It used to read a VARBINARY(8) as characters, but AB#47240 made that a real
+// conversion, and no non-PLP column type is unconvertible any more. SQL_C_NUMERIC
+// is the durable stand-in: the refusal comes from the target side, which is where
+// a soft HYC00 has to keep the value addressable.
```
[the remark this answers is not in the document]
David-Engel: Collapsed into one block in `bb450f93`. The older paragraph described the `VARBINARY` source and the maintenance note that predicted this change; both are obsolete now that binary converts. The replacement states the durable point: the refusal comes from the target side, which cannot be closed by adding a source conversion, which is why `SQL_C_NUMERIC` is the stand-in.

David-Engel:
> Bound-fetch paths remain inconsistent, and small streamed buffers incorrectly return `HY090`.

Addressed both suppressed findings in a8840251. Buffered and streamed bound delivery now share the whole-hex-pair capacity rule with `SQLGetData`; tiny hex buffers return `01004` without consuming the value. Added narrow/wide regression cases for zero-length, undersized, odd-byte, and exact-fit buffers. These also exposed and fixed a separate `SQLGetData` dispatch bug that sent binary streams through the raw UTF-16 branch.

The final focused local run covers 94.79% of the PR's changed lines, above the currently reported 93.7% overall coverage. The published coverage report will be refreshed by CI for this commit.

saurabh500 (commented):
This review came from an unattended sweep; the findings were not checked by a human first.

**Verdict: approve in substance; no blocking findings.** This is a comment-only review, not a GitHub approval.

The binary-to-character paths now render uppercase hex while keeping each byte's pair together. I checked materialized SQLGetData, streamed PLP and bound-fetch delivery against the msodbcsql source and its buffer-sizing callers: `sqlccnvt.cpp:2692-2709,7342-7349`, `sqlcdata.cpp:1065-1071,1127-1131`, and `sqlcdata.h:550-570,694-700` under `Sql/Ntdbms/sqlncli/odbc`. The earlier comments about bound delivery, tiny buffers and pair splitting are addressed at this head.

### Blocking
No blocking findings.

### Suggestion
No suggestions.

### Nit
No nits.

Changing the shared hex digit table to lowercase made the PR-modified Rust scalar-rendering test fail; restoring it made that focused test pass. This establishes uppercase sensitivity, not mutation coverage of every delivery branch.

The independent rubber-duck critique did not add or remove any findings. It prompted a wording clarification, verified against the source and review records: the reference comparison above is source inspection, not a new runtime parity measurement.

Theekshna (commented):

First pass from this automation on PR #530 (no prior marker). Two qualified reviews already ran (a Copilot bot pass and saurabh500's automated sweep with msodbcsql-source citations) and all 4 review threads are resolved. I read all three discussion endpoints plus `gh pr checks` before drafting anything; CI is fully green at this head (all legs pass, 100% diff coverage reported). I am not re-filing the bound-fetch capacity inconsistency or the streaming small-buffer `HY090` bug — both real, both already found and fixed in `a8840251`.

**Verified, not just read:**
- Traced `a8840251`'s fix directly: `hex_buffer_elements` is now the single shared rounding helper used by the materialized path (`write_captured_column`), the streaming path's error-suppression gate (`makes_no_progress && !is_length_probe && !hex_stream`), and both bound-fetch entry points (`deliver_bound_plp`'s `capacity_elements`, `deliver_bound`'s `buf_elements`). Worked through the arithmetic for `SQL_C_CHAR` buffer lengths 2 and 4 by hand against `hex_buffer_elements`/`max_read` and confirmed they reproduce the PR's own contract table (0 and 2 characters respectively) instead of the previous `HY090`.
- Confirmed the second fix: `target_type == SQL_C_WCHAR && !hex_stream` now excludes hex delivery from the raw-UTF-16 dispatch arm at `get_data.rs:2219`, matching the "exposed and fixed a separate SQLGetData dispatch bug" note in David-Engel's top-level comment.
- Re-derived the central msodbcsql claim myself: `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp` around line 2692 (`SQL_C_BINARY` case) does write two hex characters per byte only while `cbOutTmp` has room for more than one output byte, i.e. it never emits a lone half-pair — confirming the "whole pairs, odd trailing slot left empty" behavior the PR's contract table asserts. Also spot-checked `sqlcdata.h` around line 550 for the general even-rounding/terminator convention cited by the other reviewers; it's there and consistent.
- All 6 changed files reviewed against `origin/main`; no unrelated code touched.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 2 |
| Nit | 0 |

**Six required checks**
1. msodbcsql parity: re-derived the hex whole-pair rule directly from `sqlccnvt.cpp` (above), independent of the citations already in the PR/prior reviews. Audited the full `hex_stream`/`transcode_utf16_to_utf8`/`widen_narrow_to_utf16`/verbatim arms in both `get_data.rs` and `fetch_scroll.rs` — the compatibility gates in both files admit exactly `(SQL_C_CHAR | SQL_C_WCHAR, PlpEncoding::Binary)` for hex, and no other arm was left unaudited.
2. Test sufficiency: e2e coverage is extensive (`bin_to_char_test.cpp`, 15+ cases parity-checked, plus `a8840251`'s 174 new lines covering the two fixed bugs) and CI is green. Gap: `hex_buffer_elements`, `hex_pair` and `bytes_to_hex` (`get_data.rs:2936-2960`) are pure, zero-dependency functions with no entry in the `#[cfg(test)]` module, unlike their sibling `utf16le_max_read` a few hundred lines below, which has exactly this kind of dedicated unit test (`utf16le_max_read_floors_to_whole_characters`). See Suggestion below.
3. Divergences documented: N/A — no new divergence from msodbcsql behavior is introduced; the PR's own "Deliberately not addressed" section already documents the one known gap (NULL-terminator-on-NULL-binary) with a comment and a passing test that pins both drivers' agreement on the indicator.
4. PR description currency: gap found, see Suggestion below.
5. Verbose slop: none found. The `hex_buffer_elements` doc comment and the inline comments at the `hex_stream`/`makes_no_progress` sites explain non-obvious invariants (why hex shapes bypass `HY090`, why pairs can't straddle chunks) rather than restating code.
6. Audit of PR's own evidence: re-derived the whole-pair claim from `sqlccnvt.cpp` myself (above) rather than accepting the citation. The PR's own contract table (buffer lengths 1-7 -> 0,0,2,2,4,4,6 characters) is fully explained by `hex_buffer_elements`'s rounding and I did not find an unmeasured neighbor in that set — it's an exhaustive small-integer table, not a sampled one.

**Suggestion**
- Test sufficiency: add a fast unit test for `hex_buffer_elements` (and ideally `hex_pair`/`bytes_to_hex`) pinning the boundary table already in the PR description (buffer lengths 1-7 -> 0,0,2,2,4,4,6 payload characters). These are pure functions reachable without a live server or e2e harness, and the file already has the convention for exactly this shape of helper (`utf16le_max_read_floors_to_whole_characters`). Currently their correctness rests entirely on e2e, which is real coverage but slower to run and not available in every dev loop.
- PR description currency: the description's "Verification" section (1393 Rust tests, 41 e2e parity cases, `cargo test`/`clippy` output) predates `a8840251`, which fixed two real bugs found in review — the bound-fetch (`deliver_bound`/`deliver_bound_plp`) capacity inconsistency and the streaming small-buffer `HY090` misfire — and added 174 lines of new regression coverage for them. Neither fix, nor the corrected test/coverage numbers, is reflected in the description; a reader would not know either shipped in this PR.

Theekshna (commented):

First pass from this automation on PR #530 (no prior marker). Two qualified reviews already ran (a Copilot bot pass and saurabh500's automated sweep with msodbcsql-source citations) and all 4 review threads are resolved. I read all three discussion endpoints plus `gh pr checks` before drafting anything; CI is fully green at this head (all legs pass, 100% diff coverage reported). I am not re-filing the bound-fetch capacity inconsistency or the streaming small-buffer `HY090` bug — both real, both already found and fixed in `a8840251`.

**Verified, not just read:**
- Traced `a8840251`'s fix directly: `hex_buffer_elements` is now the single shared rounding helper used by the materialized path (`write_captured_column`), the streaming path's error-suppression gate (`makes_no_progress && !is_length_probe && !hex_stream`), and both bound-fetch entry points (`deliver_bound_plp`'s `capacity_elements`, `deliver_bound`'s `buf_elements`). Worked through the arithmetic for `SQL_C_CHAR` buffer lengths 2 and 4 by hand against `hex_buffer_elements`/`max_read` and confirmed they reproduce the PR's own contract table (0 and 2 characters respectively) instead of the previous `HY090`.
- Confirmed the second fix: `target_type == SQL_C_WCHAR && !hex_stream` now excludes hex delivery from the raw-UTF-16 dispatch arm at `get_data.rs:2219`, matching the "exposed and fixed a separate SQLGetData dispatch bug" note in David-Engel's top-level comment.
- Re-derived the central msodbcsql claim myself: `Sql/Ntdbms/sqlncli/odbc/sqlccnvt.cpp` around line 2692 (`SQL_C_BINARY` case) does write two hex characters per byte only while `cbOutTmp` has room for more than one output byte, i.e. it never emits a lone half-pair — confirming the "whole pairs, odd trailing slot left empty" behavior the PR's contract table asserts. Also spot-checked `sqlcdata.h` around line 550 for the general even-rounding/terminator convention cited by the other reviewers; it's there and consistent.
- All 6 changed files reviewed against `origin/main`; no unrelated code touched.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 2 |
| Nit | 0 |

**Six required checks**
1. msodbcsql parity: re-derived the hex whole-pair rule directly from `sqlccnvt.cpp` (above), independent of the citations already in the PR/prior reviews. Audited the full `hex_stream`/`transcode_utf16_to_utf8`/`widen_narrow_to_utf16`/verbatim arms in both `get_data.rs` and `fetch_scroll.rs` — the compatibility gates in both files admit exactly `(SQL_C_CHAR | SQL_C_WCHAR, PlpEncoding::Binary)` for hex, and no other arm was left unaudited.
2. Test sufficiency: e2e coverage is extensive (`bin_to_char_test.cpp`, 15+ cases parity-checked, plus `a8840251`'s 174 new lines covering the two fixed bugs) and CI is green. Gap: `hex_buffer_elements`, `hex_pair` and `bytes_to_hex` (`get_data.rs:2936-2960`) are pure, zero-dependency functions with no entry in the `#[cfg(test)]` module, unlike their sibling `utf16le_max_read` a few hundred lines below, which has exactly this kind of dedicated unit test (`utf16le_max_read_floors_to_whole_characters`). See Suggestion below.
3. Divergences documented: N/A — no new divergence from msodbcsql behavior is introduced; the PR's own "Deliberately not addressed" section already documents the one known gap (NULL-terminator-on-NULL-binary) with a comment and a passing test that pins both drivers' agreement on the indicator.
4. PR description currency: gap found, see Suggestion below.
5. Verbose slop: none found. The `hex_buffer_elements` doc comment and the inline comments at the `hex_stream`/`makes_no_progress` sites explain non-obvious invariants (why hex shapes bypass `HY090`, why pairs can't straddle chunks) rather than restating code.
6. Audit of PR's own evidence: re-derived the whole-pair claim from `sqlccnvt.cpp` myself (above) rather than accepting the citation. The PR's own contract table (buffer lengths 1-7 -> 0,0,2,2,4,4,6 characters) is fully explained by `hex_buffer_elements`'s rounding and I did not find an unmeasured neighbor in that set — it's an exhaustive small-integer table, not a sampled one.

**Suggestion**
- Test sufficiency: add a fast unit test for `hex_buffer_elements` (and ideally `hex_pair`/`bytes_to_hex`) pinning the boundary table already in the PR description (buffer lengths 1-7 -> 0,0,2,2,4,4,6 payload characters). These are pure functions reachable without a live server or e2e harness, and the file already has the convention for exactly this shape of helper (`utf16le_max_read_floors_to_whole_characters`). Currently their correctness rests entirely on e2e, which is real coverage but slower to run and not available in every dev loop.
- PR description currency: the description's "Verification" section (1393 Rust tests, 41 e2e parity cases, `cargo test`/`clippy` output) predates `a8840251`, which fixed two real bugs found in review — the bound-fetch (`deliver_bound`/`deliver_bound_plp`) capacity inconsistency and the streaming small-buffer `HY090` misfire — and added 174 lines of new regression coverage for them. Neither fix, nor the corrected test/coverage numbers, is reflected in the description; a reader would not know either shipped in this PR.

David-Engel:
> Test sufficiency: add a fast unit test for `hex_buffer_elements` (and ideally `hex_pair`/`bytes_to_hex`).
> PR description currency: neither fix, nor the corrected test/coverage numbers, is reflected in the description.

Added standalone tests for all three helpers in ab922220, covering the capacity boundary table, `usize::MAX`, all 256 byte values, and empty/zero-padded output. Updated the description with the bound-capacity, tiny-buffer, and wide-streaming fixes and revision-specific evidence, distinguishing the latest local results from the completed CI coverage on a8840251. This addresses both identical review summaries.

Theekshna (commented):
Automated review (unattended hourly sweep): re-reviewed at `ab922220`. Both Suggestions from my prior pass on this PR are resolved by this commit -- unit tests for `hex_buffer_elements`/`hex_pair`/`bytes_to_hex` are added (hand-verified against the rounding formula; matches the described buffer-length contract table) and the commit is test-only (46 lines, one file), and the description now reflects `a8840251`/`ab922220`. No new findings. Covered: the delta since my last review (diff-only test additions), parity conclusions carried over unchanged since no production code moved, and description currency. This review was produced by an unattended scheduled run; findings were not checked by a human before posting.

mssql-odbc/src/api/get_data.rs
```diff
@@ -2381,7 +2440,7 @@ fn stream_active_plp_chunk<'a>(
         SQL_NO_TOTAL
     } else if let Some(total) = known_total {
         let consumed_before = total_read.saturating_sub(read) as u64;
-        total.saturating_sub(consumed_before) as SqlLen
+        (total.saturating_sub(consumed_before) as SqlLen).saturating_mul(hex_scale)
```
shiwanigupta0809: **Suggestion (integration, not a defect in this PR):** this line is one of three places where #530 and #533 edit the same source line, from the same base blob, in ways neither PR's review has seen together.

Both branch from identical blobs — `get_data.rs` `9d6cb8b02`, `fetch_scroll.rs` `c5133bf0a` — and both restructure `stream_active_plp_chunk` and `deliver_bound_plp`. The overlaps I can see:

| site | #530 | #533 |
|---|---|---|
| this `remaining_indicator` arm | wraps it in `.saturating_mul(hex_scale)` | wraps it in `.saturating_add(held_converted_bytes)` |
| `deliver_bound_plp`'s first delivery arm | replaces `if let Some(decoder) = narrow_decoder…` with `if hex_stream { … } else if let Some(decoder)…` | replaces the **same line** with `if transcode_narrow_to_utf8 { … } else if let Some(decoder)…` |
| `direct_wire_output` | adds a comment above the expression | rewrites the expression into `first_binary_read` + `wire_shaped_output` |
| `max_read`, `makes_no_progress` | adds a `hex_stream` arm / `&& !hex_stream` | adds a `transcode_narrow_to_utf8` arm / `\|\| transcode_narrow_to_utf8` |

The good news, which I checked rather than assumed: the *arithmetic* is safe in any merge order. `hex_scale != 1` requires `PlpEncoding::Binary` and `held_converted_bytes != 0` requires `SingleByteText`, so the two terms are mutually exclusive and a merged `(wire_remaining + held) * hex_scale` is correct however it is written. Same for the two `deliver_bound_plp` arms and the two `max_read` arms — `Binary` and `SingleByteText` cannot both hold.

The part I would not leave to a conflict resolution is the **compatibility gate**. #530 adds `(SQL_C_CHAR | SQL_C_WCHAR, PlpEncoding::Binary) => true` to *both* gates — the `prepared_stream` one and the fresh-stream one. #533 rewrites the `prepared_stream` gate wholesale: the tuple changes from `(PlpEncoding, usize, usize, bool)` to `(PlpEncoding, usize, usize, Option<&'static Encoding>)` and `widening_ready` becomes `decoder_ready`. A resolver taking #533's rewritten gate can drop #530's new arm without the rest of the file complaining, and the symptom is exactly what that gate's own comment warns about: *"The two must agree or a chunked read is admitted on its first call and refused on its second."* A chunked `varbinary(max)` read as `SQL_C_CHAR` would succeed on call 1 and return `HYC00` on call 2.

CI would catch that specific case — `bin_to_char_test.cpp`'s `01 / 02 / AB / 40` continuation drives calls 2 through 4 through the prepared gate — but only on a live-server e2e leg, and only if the run happens *after* the merge rather than on the pre-merge head, which is where both PRs' green runs are today.

Concretely, what I would ask for:

1. Sequence them explicitly and say which lands first, rather than letting merge order decide.
2. On the second one, rebase onto the merged `main` and re-run the **other** PR's e2e file (`bin_to_char_test` or the new `fetch_scroll`/`get_data` collation cases) before merging — a green run on the pre-merge head does not cover the resolved chain.
3. When resolving `deliver_bound_plp`'s arm chain, keep both leading arms and leave a one-line note that `Binary` and `SingleByteText` make them exclusive, so the next person does not have to re-derive it.

Not re-filing this on #533 — it is one finding about a pair, and this is the smaller diff of the two.

shiwanigupta0809 (commented):
Reviewed the hex rendering end to end and re-derived the buffer contract independently. **No defects found in this PR.** One integration finding, inline, that only shows up when this PR is read next to #533.

What I checked rather than took on trust:

- **`hex_buffer_elements` matches the table in the description exactly.** `((n - 1) & !1) + 1` over n = 1..7 gives 1, 1, 3, 3, 5, 5, 7, i.e. payload capacities 0, 0, 2, 2, 4, 4, 6 — the msodbcsql column in your table. `usize::MAX` is a fixed point rather than an overflow.
- **The three capacity computations agree.** Streamed `max_read = payload_capacity / 2` (CHAR) and `/ 4` (WCHAR), bound `hex_buffer_elements(buf_elements) - 1`, and materialized `hex_buffer_elements(buf_elements)` all land on the same even-payload rule, so a pair cannot split at a chunk boundary *or* at a slot boundary. For `SQL_C_WCHAR`, `copy_with_nul`'s `buffer_length / 2` element count is always at least one greater than `units.len()`, so the NUL has room without ever clipping a pair.
- **`hex_scale` is consistent in both directions.** `produced_bytes` for the bound path (`out_units.len() * 2` = wire × 4 for WCHAR, `out_bytes.len()` = wire × 2 for CHAR) matches the scaled `wire_total` fed to `plp_indicator`, so the fits case and the truncated case report the same unit.
- **The new branch does not shadow an existing one.** `hex_stream` requires `PlpEncoding::Binary`, and `transcode_utf16_to_utf8` requires `SQL_C_CHAR` on a `Utf16Text` column, so the `else if hex_stream` arm sitting after it is unreachable-by-construction rather than order-dependent. `direct_wire_output` correctly omits both hex pairings.
- **The `!hex_stream` bypass of the `HY090` guard is not a new looseness.** The driver only ever raised `HY090` for odd `SQL_C_WCHAR` lengths that happened to drive `max_read` to 0 (1 and 3); a 7-byte wide text buffer already slips through on `main`. Treating a sub-pair hex buffer as a probe is consistent with that, not a regression.

The `NullVarbinaryMaxReportsNull` carve-out is the right call — a NULL never reaches the conversion, so folding it in here would have widened the diff for an unrelated path.

mssql-odbc/src/api/get_data.rs
David-Engel: #533 landed first in `182af926`; I merged that `main` into this reviewed branch in `32d43f08` (preserving review history rather than rebasing), retained the binary arms in both compatibility gates along with both conversion paths and the combined indicator arithmetic, and reran all 21 `bin_to_char`, 76 `get_data`, and 50 `fetch_scroll` e2e cases successfully against the resolved code. `bb31d15d` adds the requested one-line `Binary`/`SingleByteText` exclusivity note above the bound-delivery chain; CI is running against the integrated branch.

> [!NOTE]

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
```diff
@@ -0,0 +1,506 @@
...
+    SQLLEN ind = -1;
+    std::string text;
+    ReadChar(64, &rc, &ind, &text);
+    EXPECT_EQ(SQL_SUCCESS, rc);
+    EXPECT_EQ(2, ind);
+    EXPECT_EQ("58", text);
+    SQLCloseCursor(stmt_);
+}
+
+TEST_F(BinToCharLiveTest, GeometryRendersItsWireBytesAsHex) {
```
Theekshna, line 393: [Suggestion] The PR description claims `varbinary(max), image, hierarchyid, geometry and geography all render`, but this file only has a sibling test for hierarchyid and geometry — there's no geography case. Nothing in mssql-tds/mssql-odbc distinguishes geometry from geography (both are generic UDT/varbinary wire payloads through the same ColumnValues::Bytes path), so this is very likely to just work — but it's an unmeasured neighbor of a claim the PR's own evidence table makes, and a one-test addition (e.g. SELECT geography::Point(-122.349, 47.651, 4326), asserting the hex rendering) would close it out rather than leaving the claim unverified.

Theekshna (commented):
Unattended hourly sweep re-review at `bb31d15d` (prior sweep at `ab922220` found 0). Since then: a merge from `main` (unrelated content, confirmed the merge-base diff is unchanged — still the same 6 files/~723 lines) and a 1-line clarifying comment in `fetch_scroll.rs`. The mechanism (hex rendering for `ColumnValues::Bytes`, whole-pair truncation via `hex_buffer_elements`, streamed and bound paths) is unchanged from the prior clean pass.

Checks run this pass: `column_value_to_text`'s other match arms (Int/Money/Decimal/Uuid/Xml/Json/Vector) are pre-existing and untouched by this PR — not re-auditing them exhaustively against the legacy reference tree within this pass's budget, no msodbcsql divergence found in what I did check (Uuid upper-casing is already handled correctly per the existing comment). Re-derived one evidence-table row independently: `geometry::Point(1,2,0)` wire payload is 22 bytes -> 44 hex chars, matching the test's asserted indicator and string. Doc currency: `typed-columnar-fetch-plan.md` accurately reflects the restored `SQL_C_BINARY` + new hex-rendering state. Local Rust unit-test run blocked by an unrelated toolchain codegen error (`pclmulqdq`/AVX512 in this environment's custom `ms-prod-1.95` toolchain) unconnected to this diff; relying on the CI run this PR's description cites (151 unit + e2e cases, including the `hex_buffer_elements`/`hex_pair`/`bytes_to_hex` unit tests) rather than fighting the local build.

One test-coverage gap found, filed inline.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 0 |

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
David-Engel: Added `GeographyRendersItsWireBytesAsHex` in `3b882725`, using `geography::Point(47.651, -122.349, 4326)` (latitude first) and asserting the exact 44-character uppercase hex output plus its indicator. Measured the output against retail msodbcsql 18.6.1.1 before adding the assertion; all 22 binary-to-character cases pass on both drivers.

> [!NOTE]

Theekshna (commented):
Unattended hourly sweep, no new findings. Since my last review at `bb31d15d`, the only production-relevant addition is the `GeographyRendersItsWireBytesAsHex` test (`3b882725`) requested in my prior Suggestion; I independently re-derived its wire payload (SRID little-endian `E6100000`=4326, doubles `47.651`/`-122.349` via `BitConverter`) and it matches. The other file in this delta (`get_info_test.cpp`) is unrelated `main` content (#539) pulled in by a merge, not part of this PR's own diff against merge-base. All prior review threads (mine and others') are resolved, the PR description now reflects `3b882725`, and CI is fully green.

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
```diff
@@ -0,0 +1,519 @@
...
+    void FetchOne(const std::string& sql) {
+        ASSERT_SQL_OK(ExecDirect(sql), SQL_HANDLE_STMT, stmt_);
+        ASSERT_SQL_OK(SQLFetch(stmt_), SQL_HANDLE_STMT, stmt_);
+    }
+    // Reads column 1 as SQL_C_CHAR into a buffer of exactly `bufLen` bytes.
+    void ReadChar(SQLLEN bufLen, SQLRETURN* rc, SQLLEN* ind, std::string* text) {
+        char buf[512];
+        std::memset(buf, 0x7E, sizeof(buf));
+        *rc = SQLGetData(stmt_, 1, SQL_C_CHAR, buf, bufLen, ind);
+        *text = SQL_SUCCEEDED(*rc) ? std::string(buf) : std::string();
```
Vahid-b: **Suggestion.** `buf` is filled with `0x7E` and has no sentinel terminator, then `std::string(buf)` scans for a NUL.

Every case asserted today has the driver write a terminator, so this is safe at this head. But a regression that *drops* the terminator turns what should be a clean `EXPECT_EQ` failure into an out-of-bounds read past `buf[511]` - the test would crash or produce garbage instead of telling you what broke. It also blocks adding the `bufLen == 0` case (see my comment on line 95), because at `BufferLength == 0` the driver writes nothing at all and this line would read all 512 sentinel bytes and keep going.

One line fixes both:

```cpp
char buf[512];
std::memset(buf, 0x7E, sizeof(buf));
buf[sizeof(buf) - 1] = '\0';   // a missing driver terminator must fail the assert, not run off the end
```

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
```diff
@@ -0,0 +1,519 @@
...
+
+// The pair-atomicity rule, across every interesting buffer size. A buffer of n
+// bytes holds floor((n-1)/2)*2 characters: one is reserved for the terminator and
+// the remainder rounds down to a whole number of bytes.
+TEST_F(BinToCharLiveTest, HexNeverSplitsAByteAcrossTheBufferEdge) {
+    struct Case {
+        SQLLEN bufLen;
+        const char* expected;
+    };
+    const Case cases[] = {{1, ""}, {2, ""},     {3, "01"},   {4, "01"},
```
Vahid-b, line 103: **Suggestion.** The non-PLP whole-pair table starts at `bufLen = 1`; `BufferLength == 0` is absent.

The PLP sibling `StreamedHexPairSizedProbesDoNotConsumeValue` iterates `bufBytes` from `0`, so the two paths are covered asymmetrically at exactly the size with the most parity history: `.github/instructions/mssql-odbc.instructions.md` records that `BufferLength = 0` *is* a length probe in msodbcsql, and that raising it as a question "found a real divergence". Right now the materialized path's zero-length probe for a binary column is pinned by nothing on either leg.

Adding `{0, ""}` costs one entry. Needs the `ReadChar` fix on line 40 first, or the new case reads past the buffer.

Grading this honestly: I expect `{0, ""}` is correct by symmetry with the PLP test, but that is **inference, not measurement** - I could not run the e2e suite here. If it turns out msodbcsql answers something other than `SQL_SUCCESS_WITH_INFO` / `ind=6` / empty buffer at `BufferLength = 0`, then this is a latent divergence the suite does not cover rather than a missing row, which would make it a larger finding than filed.

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
```diff
@@ -0,0 +1,519 @@
...
+TEST_F(BinToCharLiveTest, NullVarbinaryMaxReportsNull) {
+    FetchOne("SELECT CAST(NULL AS VARBINARY(MAX))");
+
+    char buf[16];
+    std::memset(buf, 0x7E, sizeof(buf));
+    SQLLEN ind = -1;
+    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_CHAR, buf, sizeof(buf), &ind), SQL_HANDLE_STMT,
+                  stmt_);
+    EXPECT_EQ(SQL_NULL_DATA, ind);
+    // Buffer contents are deliberately not asserted: this driver writes a
```
Vahid-b: **Suggestion.** This divergence is recorded only in a comment - no work item, no registry entry.

Leaving the fix out of this PR is the right call; a NULL never enters the conversion, so folding it in would widen the diff for an unrelated path. The gap is that `grep -rn 'AB#4724'` finds no item for it, and it is absent from the "Deliberate deviations are listed below" list in `.github/instructions/mssql-odbc.instructions.md` - which is the registry a later parity pass actually reads. That file's own convention is to record a deviation "in code comments **and the tracking work item**". Without both, the next person doing a parity sweep re-derives this from scratch, or "fixes" it toward msodbcsql without knowing it was considered.

Two concrete asks:

1. File an item for the NULL-binary terminator behavior and reference it here, plus a bullet in the deviations list.
2. Add one line saying this test deliberately carries **no** `SKIP_IF_COMPARING_MSODBCSQL()` because it asserts only the indicator both drivers agree on. Without that, someone tightening it to assert buffer contents turns the parity leg red and will not understand why.

Note the divergence itself is the author's measurement, not mine - I could not run the suite on this host, so I am flagging the bookkeeping, not disputing the behavior.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -1836,10 +1840,15 @@ unsafe fn deliver_bound_plp(
...
+    let hex_stream =
+        matches!(encoding, PlpEncoding::Binary) && (target == SQL_C_CHAR || target == SQL_C_WCHAR);
     // Room for the payload. Character targets always write a terminator; binary
-    // is not a string, so the whole slot is payload.
+    // is not a string, so the whole slot is payload. A hex rendering rounds down
+    // to an even count so a byte's two characters are never split.
     let capacity_elements = if target == SQL_C_BINARY {
         stride
+    } else if hex_stream {
+        hex_buffer_elements(buf_elements).saturating_sub(1)
```
Vahid-b: **Nit.** This arm is provably a no-op on this path, and the comment above it credits the wrong mechanism.

The comment says the rounding is what stops a byte's two characters being split. On `deliver_bound_plp` it is not - `filled + 2 > capacity_elements` at line 1886 already does, because `filled` is always even. Verified rather than reasoned:

```python
def hex_buffer_elements(n): return 0 if n == 0 else ((n - 1) & ~1) + 1
def max_filled(cap):
    f = 0
    while f + 2 <= cap: f += 2          # mirrors the break at fetch_scroll.rs:1886
    return f
[b for b in range(0, 4097)
 if max_filled(hex_buffer_elements(b) - 1 if b else 0) != max_filled(max(b - 1, 0))]
# -> []
```

Identical output for every input in 0..4096, so no test can discriminate this arm from the plain `buf_elements.saturating_sub(1)` below it.

Worth contrasting with the two places the same helper *is* load-bearing: `deliver_bound` (line 2271) and `write_captured_column` (`get_data.rs:1281`) both hand the rounded count to `copy_with_nul`, an unrestricted character writer with no pair awareness, so dropping it there really would emit `"010"`.

Either drop the arm here, or keep it as belt-and-braces and move the atomicity comment down to the `+2` check - otherwise a later reader trusting this comment could simplify away the check that is actually doing the work.

Vahid-b (commented):
## Summary

Adds the mandatory binary-to-character conversion as upper-case hex across all four delivery paths (materialized `SQLGetData`, streamed `SQLGetData`, `deliver_bound`, `deliver_bound_plp`), with whole-hex-pair truncation. I re-derived the three capacity computations independently and they agree; the indicator scaling is consistent in both directions; and the msodbcsql parity leg runs this binary unskipped and is green. **No blocking findings.** Six findings, all in test-harness code, documentation, or redundancy.

## Verification

Worktree at `3b882725`, `BASE=0a16128d`, `CARGO_TARGET_DIR` redirected so the main worktree's cache is untouched.

Ran:

- `cargo fmt -- --check` clean; `cargo clippy -p mssqlodbc --all-features --all-targets -- -D warnings` clean.
- `cargo nextest run -p mssqlodbc --lib --no-fail-fast` -> **1437 passed, 0 failed** (2.0s).
- **Mutation, to confirm the new unit tests are not vacuous:** replaced `((buf_elements - 1) & !1) + 1` with `buf_elements` in `hex_buffer_elements`. `hex_buffer_elements_reserves_terminator_and_whole_pairs` **fails** (`left: 2, right: 1`). Restored. The test guards what it names.
- **Confirmed the parity leg actually covers this file** rather than assuming it: `.pipeline/scripts/containerized-odbc-e2e.sh:43` passes `--compare-with-msodbcsql`, `validation-pipeline.yml:44` pins `msodbcsqlVersion: '18.6.2.1'`, and `run_e2e.sh:91` discovers binaries by `find -name '*_test'`. Since `bin_to_char_test.cpp` carries zero `GTEST_SKIP` / `SKIP_IF_COMPARING_MSODBCSQL`, all 22 cases ran on both legs. So the parity claim is CI-validated at 18.6.2.1, not only at the 18.6.1.1 the description cites.
- Resolved AB#47240. Its "trap to handle in the same change" (re-point `UnsupportedColumnTypeHyc00PreservesValue`) is done. See the note on scope below.

Read all three discussion endpoints in full before drafting. I am **not** re-filing anything already raised and resolved: the `deliver_bound_plp` gate (`bb450f93`), the bound-capacity / `HY090` / wide-dispatch bugs (`a8840251`), the helper unit tests (`ab922220`), the #533 merge integration (`32d43f08` / `bb31d15d`), the geography case (`3b882725`), or the doc and comment staleness.

**Not verified:** I have no unixODBC toolchain on this host (`cmake`, `odbcinst`, `/usr/include/sql.h` all absent), so I could not build or run the e2e suite itself. Every e2e assertion below rests on the CI legs. The two msodbcsql expectations in findings 2 and 3 are **inference, not measurement** - each names below exactly what would settle it.

## Findings not tied to a diff line

**Nit - work-item scope.** AB#47240's title scopes it to `SQLGetData`. This PR also changes `deliver_bound` and `deliver_bound_plp`. The widening is correct and is described in the PR body, but the work item still reads as `SQLGetData`-only; worth a line there so the bound-fetch half is discoverable from the item later.

**Deferred / to file - `SQLGetInfo` never advertises the new conversion.** `grep -rn "SQL_CONVERT_\|SQL_CVT_" mssql-odbc/src/` returns zero hits, and `get_info.rs:356` maps every unhandled InfoType to `SQL_ERROR` + `HY096`. An application probing `SQL_CONVERT_BINARY` / `SQL_CONVERT_VARBINARY` / `SQL_CONVERT_LONGVARBINARY` before attempting the conversion still gets an error. This is a pre-existing crate-wide gap and correctly out of scope here - but this PR is the first change that makes the discovery surface observably behind the implementation for a specific conversion, so it is worth a tracking item rather than nothing.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 3 |
| Nit | 3 |

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
```diff
@@ -0,0 +1,519 @@
...
+    FetchOne("SELECT CAST(0x0102AB AS BINARY(3))");
+
+    SQLWCHAR wbuf[64];
+    std::memset(wbuf, 0, sizeof(wbuf));
+    SQLLEN ind = -1;
+    ASSERT_SQL_OK(SQLGetData(stmt_, 1, SQL_C_WCHAR, wbuf, sizeof(wbuf), &ind), SQL_HANDLE_STMT,
+                  stmt_);
+    EXPECT_EQ(12, ind) << "characters are reported in bytes for a wide target";
+    std::string narrow;
+    for (int i = 0; wbuf[i] && i < 64; ++i) narrow.push_back(static_cast<char>(wbuf[i]));
```
Vahid-b: **Nit.** Reversed short-circuit order: `wbuf[i]` is evaluated *before* the bound test, so an unterminated buffer dereferences `wbuf[64]` on a 64-element array.

```cpp
for (int i = 0; i < 64 && wbuf[i]; ++i) narrow.push_back(static_cast<char>(wbuf[i]));
```

Same pattern on line 168. Benign today for the same reason as `ReadChar` - the driver always terminates - but it is the same class of harness bug: a driver regression produces UB instead of a readable assertion failure.

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
David-Engel: Fixed in `128ccf49`: `ReadChar` now searches only within `BufferLength`, explicitly fails when the terminator is missing, and constructs the string with a bounded length. Zero-length reads do not scan the buffer; I also bounded the narrow bound-fetch assertion.

> [!NOTE]

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
David-Engel: Recorded the remeasured NULL-output difference in #555 and linked it from the indicator-only test, which now explicitly explains why it runs without a comparison skip (`128ccf49`). I did not add a deliberate-deviation entry: that registry explicitly excludes deferred implementation gaps, and this is not an approved decision to preserve the difference.

> [!NOTE]

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
David-Engel: Fixed both loops in `128ccf49` to check the bound before reading the element. The wide buffers now start with nonzero sentinels, so a missing terminator produces an assertion failure instead of being masked by zero initialization.

> [!NOTE]

mssql-odbc/tests/e2e/tests/bin_to_char_test.cpp
David-Engel: Measured on retail msodbcsql 18.6.1.1: a zero-length read returns `01004` and indicator 6/12 for CHAR/WCHAR but leaves the buffer untouched, rather than writing an empty string. `128ccf49` adds a dedicated non-PLP test for both widths and both `binary(3)`/`varbinary(8)`, checking unchanged sentinels, repeatable probes, and complete delivery afterward.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Removed the redundant bound-PLP rounding arm in `128ccf49` and moved the whole-pair explanation to the `filled + 2` guard. The materialized and buffered-bound writers retain the shared rounding helper they depend on.

> [!NOTE]

David-Engel:
> Nit - work-item scope. AB#47240's title scopes it to SQLGetData.
> Deferred / to file - SQLGetInfo never advertises the new conversion.

Updated AB#47240's title and appended an explicit scope paragraph covering `deliver_bound` and `deliver_bound_plp`. The conversion-discovery gap is already tracked in [AB#47996](https://dev.azure.com/SqlClientDrivers/mssql-rs/_workitems/edit/47996), whose conversion-mask list explicitly includes `SQL_CONVERT_BINARY`, `SQL_CONVERT_VARBINARY`, and `SQL_CONVERT_LONGVARBINARY`; I linked that existing item in the PR description rather than creating a duplicate.

Theekshna (commented):
Unattended hourly sweep re-review at `128ccf49` (prior sweep at `3b882725` found 0). Since then, one commit ("Harden binary hex tests and cover zero-length probes"): it removes the `hex_buffer_elements()` rounding from `deliver_bound_plp`'s hex-stream capacity and tightens `bin_to_char_test.cpp`'s buffer/terminator assertions.

I did not accept "redundant capacity rounding" from the commit/PR description at face value -- I re-derived it. The hex loop only ever checks `filled + 2 > capacity_elements` before appending a whole pair, so `filled` after the loop is always `2*floor(capacity/2)` regardless of whether `capacity_elements` itself was pre-rounded to even. I hand-verified `hex_buffer_elements(n)-1` and `n-1` produce the same `2*floor(x/2)` for n=0..9 (the only place the rounding could matter is when `n-1` is odd, i.e. n even, and even there the post-loop `filled` is identical either way) -- the removal is a true no-op, not a hidden off-by-one. The new e2e cases exercise this via a 65-byte buffer (odd, so it doesn't itself probe the even-`n` boundary), but the boundary is exhaustively unit-tested elsewhere (`hex_buffer_elements_reserves_terminator_and_whole_pairs`, `get_data.rs`) and my own derivation covers the parity this specific commit touches.

Also checked: the new `#555` reference (NULL-terminator divergence deferral) is a real, open issue matching the description. `gh pr checks` is fully green (including 100% diff coverage on the two touched files). All review threads (mine and others') are resolved, and the description already documents this exact commit's rationale accurately.

No new findings.

Theekshna (approved):

PR #531, opened 2026-09-09
Align Python wheel NuGet versions

gargsaumya:
## Description

Make the `mssql-python-rs-wheels` NuGet transport follow the `mssql-python-rs` Python distribution version from `mssql-py-core/pyproject.toml` instead of the independently versioned `mssql-py-core` Rust crate.

With the current sources, the pipelines now produce:

- Dev: `0.1.0-dev.<YYYYMMDD>.<BuildId>`
- Nightly: `0.1.0-nightly.<YYYYMMDD>`
- Official: `0.1.0`

The embedded Rust component remains independently versioned at `0.1.10`. Release documentation now records this ownership model.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47953

## Validation

- `cargo bfmt`
- Focused version-source assertions for dev, nightly, and official package forms
- YAML diagnostics for both changed pipeline files
- `git diff --check`

`cargo bclippy` and `cargo btest` are blocked before compilation on current `main` because `Cargo.lock` requires an update while both repository aliases enforce `--frozen`. This PR does not change Cargo manifests or the lockfile.

## Checklist

- [x] `cargo bfmt` passes
- [ ] `cargo bclippy` passes
- [ ] `cargo btest` passes


.pipeline/OneBranch/stages.yml
```diff
@@ -515,12 +515,13 @@ stages:
       - pwsh: |
           $ErrorActionPreference = 'Stop'
 
-          # Extract version from mssql-py-core/Cargo.toml
-          $cargoToml = Get-Content "$(Build.SourcesDirectory)/mssql-py-core/Cargo.toml" -Raw
-          if ($cargoToml -match '(?ms)^\[package\].*?^version\s*=\s*"([^"]+)"') {
+          # The NuGet package transports the Python distribution, so it follows
+          # the wheel version rather than the independently-versioned Rust crate.
+          $pyprojectToml = Get-Content "$(Build.SourcesDirectory)/mssql-py-core/pyproject.toml" -Raw
```
David-Engel, line 521: **Suggestion** — the published NuGet version moves *backwards* here. The package id stays `mssql-python-rs-wheels`, but its version goes from the Cargo lineage (`0.1.10-*`, per the pre-change docs example) to `0.1.0-*`. Under NuGet SemVer 2.0 ordering the already-published `0.1.10-*` packages sort **higher** than everything this pipeline will produce from now on, so a consumer resolving "latest prerelease" rather than a pinned base version keeps resolving stale `0.1.10` packages and silently stops picking up new builds. An official `0.1.0` would also sort below existing `0.1.10-dev.*` prereleases.

I can't see the `mssql-rs/mssql-rs` feed or `mssql-python`'s package reference from this repo, so this is a question rather than a defect: has the `mssql-python` side been checked for a floating/latest-prerelease reference, and is there a plan for the existing `0.1.10-*` packages? If `mssql-python` pins an explicit base version (as the examples in `docs/release-management.md` do), this is a non-issue — but a one-line "version lineage reset" note in the Version Scheme section would save the next operator the same investigation.

David-Engel (commented):
## Summary

*Unattended automated review — these findings were not checked by a human first.*

This flips the `mssql-python-rs-wheels` NuGet version source from `mssql-py-core/Cargo.toml` to `[project].version` in `mssql-py-core/pyproject.toml` on both the build (`stages.yml`) and official release (`OfficialPythonWheelsRelease.yml`) paths, and rewrites the release runbook around the split ownership. The mechanism is correct and internally consistent: `stages.yml:502` already read `pyproject.toml` for wheel validation, so the packaging step now agrees with it, and the `Tag` stage still reads `$metadata.Version` (Cargo) at `OfficialPythonWheelsRelease.yml:364`, matching the documented "the tag is Cargo-versioned" model. No blocking issues.

What I ran in an isolated worktree at `5bef165`:

- `python -m pytest scripts/test_release_pipeline.py -k "nonofficial or wheel_validation"` — 6 passed.
- **Mutation check on the new test.** Reverted the `stages.yml` step to `Cargo.toml` / `[package]` and all three new cases failed (`0.1.10-dev.20260910.12345` vs the expected `0.1.0-dev...`). The test genuinely guards the change, and the fixture deliberately using differing Cargo (`0.1.10`) and pyproject (`0.1.0`) versions is exactly the right shape — a same-version fixture would have been vacuous here.
- Rendered the two changed ordered lists through GitHub's `/markdown` API rather than eyeballing them; see the `docs/release-management.md` comments.

Findings: 0 blocking, 2 suggestions, 4 nits — all inline.

Context, not a finding: I reproduced the `--frozen` failure your Validation section describes. `Cargo.lock` is gitignored and untracked (`.gitignore:27`), so `--frozen` fails on any fresh checkout until a lock file exists — `cargo generate-lockfile` once, then `cargo bclippy`, works. That is pre-existing repo behavior rather than anything this PR did, and this PR touches no Rust, so leaving those boxes unchecked with the explanation reads fine to me.

.pipeline/OneBranch/stages.yml
```diff
@@ -515,12 +515,13 @@ stages:
...
+          if ($pyprojectToml -match '(?ms)^\[project\]\s*$.*?^version\s*=\s*"([^"]+)"') {
```
David-Engel: **Nit** — this inline regex is now the second copy of the same pattern in this file (the wheel-validation step at line 503 has it too), and neither bounds the table: with `(?s)`, `^\[project\]\s*$.*?^version\s*=` will match a `version` from a *later* table if `[project]` ever loses its own, silently packaging the wrong version instead of failing.

`Get-TomlField` in `.pipeline/scripts/get-python-release-metadata.ps1:36-46` already solves this with `(?=^\[|\z)` and throws a clear message. Not a live bug — `[project].version` exists today and I confirmed both steps extract `0.1.0` — but folding these two steps onto that helper (or copying its section-bounded pattern) would stop the three version readers from drifting apart.

docs/release-management.md
```diff
@@ -213,30 +219,36 @@ Sprint end:
...
+- [ ] `[project].version` in `mssql-py-core/pyproject.toml` is correct (e.g., `0.2.1`)
+- [ ] Any Rust crate selected for release has the intended independent version in its `Cargo.toml`
 - [ ] No outstanding breaking changes without coordination
 
 ### Release build
 
-1. **Trigger the Official pipeline manually** with `isOfficial: true`
+1. **Run the Official Python Wheels Build** for the source commit to release
+2. **Trigger the Official release pipeline** with that build selected and `publishNuGet: true`
   - This produces a clean semver NuGet: `mssql-python-rs-wheels.0.2.1`
```
David-Engel: **Suggestion** — these sub-bullets are indented 2 spaces under a `2. ` marker, which needs 3. I ran this block through GitHub's `/markdown` API: it closes the `<ol>` after step 2 and emits the three bullets as a separate top-level `<ul>`, so the runbook reads as though they are steps in their own right rather than details of step 2.

```suggestion
   - This produces a clean semver NuGet: `mssql-python-rs-wheels.0.2.1`
```

docs/release-management.md
```diff
@@ -292,21 +304,14 @@ missing ODBC payloads, and exact-source metadata failures.
...
-1. Cherry-pick fix to `release/0.2.1`
-2. Bump `mssql-py-core/Cargo.toml` to `0.2.2`
-3. Trigger Official pipeline from `release/0.2.1` branch
-4. Tag `v0.2.2`, update `mssql-python` to use `0.2.2`
+1. Cherry-pick the fix to the appropriate maintenance branch
+2. Bump `[project].version` in `mssql-py-core/pyproject.toml` for the Python hotfix (for example, `0.2.1` → `0.2.2`)
+3. Bump `[package].version` in `mssql-py-core/Cargo.toml` only if the Rust crate also needs a new independent version
+4. Run the Official Python Wheels Build for the hotfix commit
+5. Trigger the Official release pipeline with `publishNuGet: true`
+  - This publishes `mssql-python-rs-wheels.0.2.2`
```
David-Engel: **Suggestion** (same issue as line 230, worse here) — the 2-space bullet splits the hotfix list into three blocks. GitHub renders it as `<ol start="1">` for steps 1-5, a stray top-level `<ul>` for this line, then `<ol start="6">` for steps 6-7. Three spaces keeps it attached to step 5.

```suggestion
   - This publishes `mssql-python-rs-wheels.0.2.2`
```

docs/release-management.md
```diff
@@ -292,21 +304,14 @@ missing ODBC payloads, and exact-source metadata failures.
...
+6. Enable `tagRelease` only when creating the separate Cargo-versioned mssql-py-core tag and release branch
+7. Update `mssql-python` to use `mssql-python-rs==0.2.2`
```
David-Engel, line 319: **Nit** — this names the PyPI distribution, but the post-release step at line 254 and the architecture diagram both have `mssql-python` consuming the NuGet transport (`mssql-python-rs-wheels@0.2.1`). The numbers happen to match now that both follow `pyproject.toml`, but the package name doesn't. Copilot raised this as a suppressed low-confidence comment; I checked it against lines 14 and 254 and agree.

```suggestion
7. Update `mssql-python` to use `mssql-python-rs-wheels@0.2.2`
```

docs/release-management.md
```diff
@@ -74,7 +75,7 @@ NuGet SemVer 2.0 ordering: `dev` < `nightly` < release (no suffix).
 1. Schedule triggers at 2 AM UTC
 2. Pipeline builds 34 wheels across all platforms
 3. Publish stage:
-   - Extracts version from mssql-py-core/Cargo.toml (e.g., 0.2.0)
+  - Extracts the Python distribution version from `mssql-py-core/pyproject.toml` (e.g., `0.2.0`)
```
David-Engel: **Nit** — this line sits inside the fenced block that opens at line 73, so the new backticks render literally rather than as code, and it's the only line in the block using them. The indentation also dropped to 2 spaces while its siblings at lines 79 and 81 stay at 3.

```suggestion
   - Extracts the Python distribution version from mssql-py-core/pyproject.toml (e.g., 0.2.0)
```

scripts/test_release_pipeline.py
```diff
@@ -104,6 +105,71 @@ def test_release_defaults_are_safe():
     assert pipeline["resources"]["pipelines"][0]["source"] == "Official Python Wheels Build"
 
 
+@pytest.mark.parametrize(
+    ("build_reason", "expected_version"),
+    [
+        ("Schedule", "0.1.0-nightly.20260910"),
+        ("IndividualCI", "0.1.0-dev.20260910.12345"),
+        ("Manual", "0.1.0-dev.20260910.12345"),
```
David-Engel: **Nit** — the set covers `Schedule` / `IndividualCI` / `Manual`-non-official, but not the `Manual` + `isOfficial: true` arm of the same `switch`, which is the one producing the permanent clean `0.1.0`. Since `expand` already takes `isOfficial` as a flag, a third parameter plus a `("Manual", True, "0.1.0")` case would close out the switch. Optional — the official *release* path is covered separately by `test_wheel_validation_and_optional_nuspec` against `OfficialPythonWheelsRelease.yml`; this would just cover the official arm of the build pipeline.

saurabh500 (commented):
Unattended review sweep; these findings were not checked by a human before posting.

**Verdict: no new blocking findings.**

Both nonofficial and official NuGet packaging now take the Python distribution version, while optional mssql-py-core source tags continue using the independent Cargo version. I traced both paths through metadata extraction, wheel validation, and nuspec generation, and read the existing inline comments, review bodies, and PR-level discussion. Previously raised points remain in their existing threads rather than being duplicated here; this is not a claim that downstream feed migration has been verified.

### Blocking
No new blocking findings.

### Suggestion
No new suggestions beyond the existing discussion.

### Nit
No new nits beyond the existing discussion.

The six focused version/wheel-validation cases passed. Temporarily changing both production version sources back to Cargo made all five version-sensitive cases fail with 0.1.10 instead of 0.1.0; the missing-wheel rejection case remained green as expected. Restoring the sources restored all six passes. These local cases do not exercise live NuGet publication or downstream package resolution.

An independent rubber-duck critique found no corrections to incorporate; the verdict and findings were unchanged.

docs/release-management.md
gargsaumya: Keeping this as `mssql-python-rs==0.2.2` is intentional. microsoft/mssql-python#768 changes the published `mssql-python` wheel to declare an exact direct runtime dependency on `mssql-python-rs`; NuGet remains the internal transport used by build/Conda workflows. Those workflows separately read the pinned `eng/versions/mssql-python-rs.version`, so changing this hotfix step to a NuGet package name would describe the build transport rather than the customer dependency.

docs/release-management.md
```diff
@@ -213,30 +224,36 @@ Sprint end:
...
+- [ ] `[project].version` in `mssql-py-core/pyproject.toml` is correct (e.g., `0.2.1`)
+- [ ] Any Rust crate selected for release has the intended independent version in its `Cargo.toml`
 - [ ] No outstanding breaking changes without coordination
 
 ### Release build
 
-1. **Trigger the Official pipeline manually** with `isOfficial: true`
+1. **Run the Official Python Wheels Build** for the source commit to release
+2. **Trigger the Official release pipeline** with that build selected and `publishNuGet: true`
   - This produces a clean semver NuGet: `mssql-python-rs-wheels.0.2.1`
```
David-Engel: **Suggestion** (previously raised, thread resolved without a change) — this sub-bullet is indented 2 spaces under a `2. ` marker, which needs 3. Re-confirmed against GitHub's `/markdown` API at this head: the `<ol>` closes after step 2 and all three bullets render as a separate top-level `<ul>`, so they read as steps in their own right rather than details of step 2.

```suggestion
   - This produces a clean semver NuGet: `mssql-python-rs-wheels.0.2.1`
```

docs/release-management.md
```diff
@@ -292,21 +309,14 @@ missing ODBC payloads, and exact-source metadata failures.
...
-1. Cherry-pick fix to `release/0.2.1`
-2. Bump `mssql-py-core/Cargo.toml` to `0.2.2`
-3. Trigger Official pipeline from `release/0.2.1` branch
-4. Tag `v0.2.2`, update `mssql-python` to use `0.2.2`
+1. Cherry-pick the fix to the appropriate maintenance branch
+2. Bump `[project].version` in `mssql-py-core/pyproject.toml` for the Python hotfix (for example, `0.2.1` → `0.2.2`)
+3. Bump `[package].version` in `mssql-py-core/Cargo.toml` only if the Rust crate also needs a new independent version
+4. Run the Official Python Wheels Build for the hotfix commit
+5. Trigger the Official release pipeline with `publishNuGet: true`
+  - This publishes `mssql-python-rs-wheels.0.2.2`
```
David-Engel: **Suggestion** (previously raised, thread resolved without a change) — same 2-space bullet issue, and it splits the hotfix list into three blocks. `/markdown` at this head returns `<ol>` for steps 1-5, a stray top-level `<ul>` for this line, then `<ol start="6">` for steps 6-7. Three spaces keeps it attached to step 5.

```suggestion
   - This publishes `mssql-python-rs-wheels.0.2.2`
```

David-Engel (commented):
## Summary

*Unattended automated review — these findings were not checked by a human first.*

Re-review at `bdabc318`. The follow-up commit addresses the substantive items from the previous pass: the TOML extraction is now bounded per-table in both `stages.yml` steps, the parametrized test covers the `Manual` + `isOfficial: true` arm, and the Version Scheme section documents the `0.1.10-*` → `0.1.0-*` lineage move. Mechanism still looks right — `stages.yml` packaging now agrees with the wheel-validation step that already read `pyproject.toml`, and the `Tag` stage keeps using `$metadata.Version` (Cargo), matching the documented split.

What I ran in an isolated worktree at `bdabc318`:

- `python -m pytest scripts/test_release_pipeline.py` — 46 passed.
- **Mutation check.** Reverted both `stages.yml` steps to `Cargo.toml` / `[package]`: all four `test_nonofficial_nuget_versions_follow_python_distribution` cases failed with `0.1.10*` instead of `0.1.0*`, including the new official arm. Restored, green again.
- **Regex check.** Against a `pyproject.toml` whose `[project]` table has no `version` but a later `[tool.maturin]` does, the old inline pattern silently returned the later table's value; the new bounded pattern fails and throws. That is a real correctness fix, not just a style change.
- Re-rendered the two runbook lists through GitHub's `/markdown` API rather than eyeballing them.

Findings: 0 blocking, 2 suggestions — both inline, and both are the *same* markdown-rendering issue I raised last pass. Those threads were resolved without a change and without a reply, so I am re-raising them once against the current head rather than letting them disappear. If the intent is to leave them as-is, a reply saying so is enough and I will not raise them again.

Note on the `mssql-python-rs==0.2.2` thread: your explanation about microsoft/mssql-python#768 and `eng/versions/mssql-python-rs.version` makes sense to me — the runtime dependency and the build transport are different things. Not re-raising.

Still outside what I can verify locally: live NuGet feed behavior and downstream resolution in `mssql-python`. The new Version Scheme paragraph tells consumers to pin, which is the right mitigation to document, but someone should confirm the `mssql-python` side is actually pinned before the first `0.1.0-*` publish.

Theekshna (commented):
Unattended hourly sweep, no new findings. Covered: msodbcsql parity — N/A (no ODBC surface touched); re-ran `scripts/test_release_pipeline.py` (46 passed, independently confirming the mutation-tested version-source assertions already reported by prior reviews at this SHA); verified the two remaining 2-space-indent markdown nits (lines 235, 317) are already tracked in an open, unresolved thread from a prior review at this same head, so not re-filed; checklist (bclippy/btest unchecked with frozen-lockfile explanation) is consistent with the diff touching no Rust files; the `coverage-report` Action job failure is an artifact-wait timeout (75 min), unrelated to this docs/pipeline-only diff.

docs/release-management.md
gargsaumya: Fixed in c9046483 — indented to 3 spaces so the sub-bullet attaches to its ordered-list item.

docs/release-management.md
gargsaumya: Fixed in c9046483 — indented to 3 spaces so the sub-bullet attaches to its ordered-list item.

shiwanigupta0809 (commented):
## Summary

Repointing the `mssql-python-rs-wheels` NuGet version at `[project].version` is the right call — the package transports wheels, so it should carry the wheel version, and the Cargo version was only ever incidental to it. The split-ownership model is now coherent across both pipelines, the runbook, and the tests.

**Blocking: 0 · Suggestion: 0 · Nit: 1.**

I reviewed the full `origin/main...c9046483` diff and deliberately went looking for a missed version source rather than re-reading what earlier passes already covered. I'm not re-filing anything from the Copilot, David-Engel, saurabh500 or Theekshna threads.

## What I verified independently

Five things that could plausibly have been wrong here, all of which hold:

- **The release path can't drift from the wheels it ships.** `$metadata.PythonVersion` resolves through `Get-BuildSourceFile`, which is `git show ${CommitSha}:mssql-py-core/pyproject.toml` against the *selected build's* commit — not the release pipeline's own checkout. So the nuspec version is read from the same tree the wheels were built from, and a `pyproject.toml` bump landing on `main` between build and release cannot mislabel the package. This was the failure mode I most expected to find and it's already closed.

- **No version source was left behind.** `stages.yml` has zero remaining `Cargo.toml` reads on the wheel path — both the validation step (502) and the nuspec step (521) now read `pyproject.toml`. The `Cargo.toml` read in `get-python-release-metadata.ps1:48` survives, correctly, only to feed `tagRelease`.

- **Validation and packaging agree by construction.** `verify-python-wheels.ps1 -ExpectedVersion $pythonVersion` and the nuspec `<version>` derive from the same extraction in both pipelines, so a wheel whose filename version disagrees with the package version fails the build rather than shipping.

- **The `tagRelease` behaviour matches the new prose.** `OfficialPythonWheelsRelease.yml:364` uses `$version = $metadata.Version` (Cargo), not the `releaseVersion` variable this PR just switched to the Python version. The new "Optional mssql-py-core source tag" section is accurate — worth stating explicitly, because reusing `$(releaseVersion)` there is exactly the kind of thing a change like this drags along by accident, and it didn't.

- **The bounded TOML regex is correct, including the greedy-`\s*$` case.** `(?ms)^\[project\]\s*$(?<body>...)` looks like it could over-consume, since `\s` matches newlines under `(?s)`. It can't produce a wrong answer: backtracking makes `$` settle on the last blank line after the header, so `body` always still contains the `version` key. It also correctly stops at `[project.optional-dependencies]`, where a `version` would not be the distribution version.

## Nit

**`docs/release-management.md:277` — the switch-reference table no longer identifies a single version.**

```
| `tagRelease` | Tag and branch the selected build's `mssql-py-core` version. This does not tag Rust crate versions. |
```

This line is untouched by the PR, and it was unambiguous before it: `mssql-py-core` had one version, and NuGet and the tag shared it. This change gives `mssql-py-core` two — `[package].version` `0.1.10` and `[project].version` `0.1.0` — so "the selected build's `mssql-py-core` version" now names two different values, and the row reads as though it might tag the one the release just published.

The second sentence makes it worse rather than better. Read against the neighbouring rows it plausibly means "doesn't tag `mssql-tds` / `mssql-mock-tds`", which is true. Read plainly it says `tagRelease` doesn't tag a Rust crate version, which is the opposite of what line 242 of this same file now says and of what the pipeline does at `OfficialPythonWheelsRelease.yml:364`.

Why this is only a nit: the authoritative explanation is 35 lines above in the section this PR added, and the switch table is a quick reference rather than the runbook step an operator follows. Why it's worth fixing anyway: the table is what someone scans mid-release, and the ambiguity is one this PR introduced.

Something like:

```
| `tagRelease` | Tag and branch the `mssql-py-core` **Cargo** version (`[package].version`) from the selected build. Independent of the NuGet/wheel version, and does not tag `mssql-tds` or `mssql-mock-tds`. |
```

I'd have left this inline, but line 277 falls between diff hunks so it isn't commentable.

## On the checklist

Not a finding, just confirming the state for whoever merges: `cargo bclippy` and `cargo btest` are still ticked while the Validation section says both are blocked by the `--frozen` lockfile. Copilot raised this and the thread is open. Given the diff touches no Rust — two YAML files, one Markdown file, and a pytest module — the honest resolution is probably to unticked them and say the Rust gates are N/A, rather than to chase a lockfile fix that has nothing to do with this change.

.pipeline/OneBranch/stages.yml
```diff
@@ -515,14 +516,16 @@ stages:
...
-          # Extract version from mssql-py-core/Cargo.toml
-          $cargoToml = Get-Content "$(Build.SourcesDirectory)/mssql-py-core/Cargo.toml" -Raw
-          if ($cargoToml -match '(?ms)^\[package\].*?^version\s*=\s*"([^"]+)"') {
-            $version = $Matches[1]
-          } else {
-            Write-Error "Could not extract version from Cargo.toml"
+          # The NuGet package transports the Python distribution, so it follows
+          # the wheel version rather than the independently-versioned Rust crate.
+          $pyprojectToml = Get-Content "$(Build.SourcesDirectory)/mssql-py-core/pyproject.toml" -Raw
+          $projectMatch = [regex]::Match($pyprojectToml, '(?ms)^\[project\]\s*$(?<body>.*?)(?=^\[|\z)')
```
David-Engel, line 522: **Suggestion** — this bounded `[project]` extraction is now the third verbatim copy in the pipeline. `Get-TomlField` in `.pipeline/scripts/get-python-release-metadata.ps1:32-45` already implements the identical pair of patterns:

```powershell
'(?ms)^\[{0}\]\s*$(?<body>.*?)(?=^\[|\z)'
'(?m)^\s*{0}\s*=\s*"([^"]+)"'
```

and this PR adds two more inline (line 503 and here). All three are correct today, which is exactly why they will drift: a fix to one — say, tolerating `dynamic = ["version"]`, or a table whose body contains a line starting with `[` — has no reason to find the other two, and the release path and the build path would then disagree about the package version without any test noticing.

`stages.yml` already reaches into that directory (`$(Build.SourcesDirectory)/.pipeline/scripts/verify-python-wheels.ps1`, line 509), so a small `.pipeline/scripts/Get-TomlField.ps1` dot-sourced by both steps and by `get-python-release-metadata.ps1` would collapse this to one implementation. It would also give the pattern a place to be unit-tested directly, instead of only through the two pipeline-expansion tests.

Not blocking — the duplication is correct and the change is otherwise self-consistent. Worth doing while the three copies are still identical.

David-Engel (commented):
## Summary

*Unattended automated review — these findings were not checked by a human first.*

Reviewed the full `origin/main...c9046483` diff (merge base `ea1dfe61`). The change is coherent: the NuGet package transports the wheels, so it should carry the wheel version, and both the build (`stages.yml`) and release (`OfficialPythonWheelsRelease.yml`) paths now read `[project].version` while `tagRelease` keeps reading `[package].version`.

**Blocking: 0 · Suggestion: 1 · Nit: 0.**

I read the Copilot, David-Engel, saurabh500, Theekshna and shiwanigupta0809 threads first and am not re-filing anything from them — including the version-downgrade discussion, the switch-table `tagRelease` wording, and the remaining 2-space markdown indents. I went looking for evidence prior passes asserted but did not show.

## What I verified independently

**1. The new test is not vacuous.** I mutated `stages.yml:528` from `$version = $versionMatch.Groups[1].Value` to a hardcoded `"0.1.10"` (the Cargo lineage this PR moves away from) and re-ran the suite. All four parametrized cases fail:

```
FAILED test_nonofficial_nuget_versions_follow_python_distribution[Schedule-False-0.1.0-nightly.20260910]
FAILED test_nonofficial_nuget_versions_follow_python_distribution[IndividualCI-False-0.1.0-dev.20260910.12345]
FAILED test_nonofficial_nuget_versions_follow_python_distribution[Manual-False-0.1.0-dev.20260910.12345]
FAILED test_nonofficial_nuget_versions_follow_python_distribution[Manual-True-0.1.0]
```

Writing `Cargo.toml` at `0.1.10` and `pyproject.toml` at `0.1.0` in the fixture is what makes this work — the assertion can only pass if the step reads the Python version. Worktree restored afterwards. Unmutated: `46 passed`.

**2. The bounded TOML regex holds on adversarial input, and its failure modes are fail-closed.** Prior passes reasoned about the greedy-`\s*$` backtracking; I ran the exact pattern pair from `stages.yml:522-523` against seven `pyproject.toml` shapes:

| Input shape | Result |
|---|---|
| CRLF, decoy `version` in `[build-system]` *and* `[tool.maturin]` | `0.1.0` |
| `[project.optional-dependencies]` subtable follows | `0.1.0` |
| Blank / whitespace-only lines directly after `[project]` | `0.1.0` |
| Commented-out `# version = "9.9.9"` decoy inside `[project]` | `0.1.0` |
| `[project]` is the last table in the file | `0.1.0` |
| `dynamic = ["version"]`, no `[project].version` | throws |
| Array continuation line starting with `[` before `version` | throws |

The two failures are the only ways to derail it, and both hit the `-not $versionMatch.Success` arm rather than silently producing a wrong package version. That is the property that matters here, since a silently-wrong version ships.

**3. Nothing in this repo resolves the package version, so the `0.1.10-*` → `0.1.0-*` downgrade has no in-repo consumer to mislead.** Grepped `.pipeline`, `scripts`, `docs` and `.github` for `mssql-python-rs-wheels` and `nugetVersion`: the only non-doc hits are the two nuspec-producing steps and the `NuGetCommand@2` pack tasks. The cross-repo `Build mssql-python` job consumes wheel *artifacts*, not the feed. The ordering hazard is real but entirely external, and the new paragraph at `release-management.md:64-67` is the right place for it.

**4. `tagRelease` did not get dragged along.** `OfficialPythonWheelsRelease.yml:364` still uses `$version = $metadata.Version`, while `releaseVersion` (180) now carries `$pythonVersion`. The two are genuinely decoupled, matching the new "Optional mssql-py-core source tag" section.

## Suggestion

**The bounded extraction now exists in three places.** See the inline comment on `.pipeline/OneBranch/stages.yml`.

## Notes for whoever merges

**Checklist.** Concrete evidence for the open Copilot thread rather than a re-file: `git ls-files Cargo.lock` returns nothing — the lockfile is not tracked at all, so `--frozen` fails with `cannot create the lock file ... because --frozen was passed` on any clean checkout. That is repo-wide and predates this PR, and confirms the author's Validation note. Since the diff touches two YAML files, one Markdown file and one pytest module, unticking `cargo bclippy` / `cargo btest` and marking them N/A is accurate rather than a workaround.

**CI.** `mssql-rs Pull request validation (Build Stage Build Linux)` is failing at this head. The `stages.yml` diff is confined to two steps in the `Publish` stage (lines 500-530); the Linux build job is untouched, so this is very unlikely to be the PR. Worth a rerun or a glance at the log before merge — I have no ADO access to confirm, and it is why I am leaving the `ready for human review` label alone rather than managing it.

Theekshna (commented):

Re-reviewed at `c9046483`, superseding my prior clean pass at `bdabc318`. The only change since then is a 2-line whitespace fix in `docs/release-management.md` (list-continuation indent for the two `- This produces/publishes a NuGet...` sub-bullets, now 3 spaces instead of 2) — this is exactly the markdown nit an earlier pass in this thread had flagged as tracked-but-unresolved; it's now fixed and I confirm the new indentation correctly nests under the `2.`/`5.` ordered items per CommonMark. No production code, pipeline logic, or test assertions changed in this delta.

| Category | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

**Six required checks**
1. msodbcsql parity: N/A — no ODBC surface touched (Python wheel/NuGet packaging and release docs only).
2. Test sufficiency: ran `scripts/test_release_pipeline.py` locally (`pwsh` + `pytest` available in this environment) — all 49 tests pass at this head. Mutated the PR's own fix (`$version = $versionMatch.Groups[1].Value` -> a hardcoded literal in `stages.yml`) and confirmed all 4 parametrized cases of `test_nonofficial_nuget_versions_follow_python_distribution` fail as expected, then restored the file — the new test genuinely guards the version-source change rather than passing vacuously.
3. Divergences documented: N/A — no msodbcsql/SqlClient/mssql-jdbc behavior involved.
4. PR description currency: matches the diff; the checklist's unchecked `cargo bclippy`/`cargo btest` boxes are consistent with this PR touching no Rust files, and `gh pr checks` shows the ADO validation legs passing (two legs still pending at review time, not failing).
5. Verbose slop: none — the two new inline comments (`stages.yml:518-519`, `OfficialPythonWheelsRelease.yml:154-155`) each state a non-obvious rationale (NuGet follows the wheel/distribution version, not the independently-versioned Rust crate) rather than restating the code.
6. Audit of PR's own evidence: re-derived the Dev/Nightly/Official version-format claims from the pipeline script directly via the mutation test above rather than accepting the description's table at face value; all three forms reproduce as stated. No unmeasured neighbor found — the parametrized test in `test_release_pipeline.py` covers all four `(build_reason, is_official)` combinations the pipeline branches on.

**Note:** one review thread from the PR author (David-Engel, `stages.yml:522`, self-review) remains open — a non-blocking suggestion to de-duplicate the `[project]`-table regex extraction into a shared script. Not re-filed; it's the author's own thread, explicitly marked not blocking, and outside this delta's scope.

David-Engel (approved):

.pipeline/OneBranch/stages.yml
gargsaumya: Agreed this is worth doing. Holding the shared `Get-TomlField.ps1` extraction as a focused follow-up rather than folding it into this version-alignment PR, since #532 is in flight touching both `stages.yml` and `get-python-release-metadata.ps1`. Will consolidate all three call sites once that lands.

David-Engel (commented):

## Summary

Reviewed the full `origin/main...c9046483` diff (merge base `ea1dfe61`). Repointing `mssql-python-rs-wheels` at `[project].version` is the right model — the package transports wheels, so it should carry the wheel version — and both the build (`stages.yml`) and release (`OfficialPythonWheelsRelease.yml`) paths now agree while `tagRelease` deliberately keeps `[package].version`.

**Blocking: 0 · Suggestion: 1 · Nit: 1.**

I read every existing thread first and am **not** re-filing: the `0.1.10-*` -> `0.1.0-*` ordering hazard, the third copy of the bounded TOML regex, the `tagRelease` switch-table wording, the remaining 2-space indent inside the fenced block at `docs/release-management.md:83`, or the `--frozen`/checklist discussion. All of those are already on record. Both findings below are new.

## Verification

- Ran `python -m pytest scripts/test_release_pipeline.py -k "nonofficial or wheel_validation"` in this worktree: **7 passed**. `pwsh` and `pyyaml` were available, so the new parametrized cases genuinely execute the extracted `stages.yml` step rather than pattern-matching the YAML.
- Confirmed `get-python-release-metadata.ps1:56` populates `PythonVersion` only under `-IncludeWheelMetadata`, and that the release pipeline passes that switch (`OfficialPythonWheelsRelease.yml:150`) — so `$packageVersion = $pythonVersion` cannot resolve to `$null` and silently pack an empty `<version>`.
- Grepped the repo for `mssql-python-rs-wheels` and `nugetVersion`: only the two nuspec-producing steps, the docs and the tests. No in-repo consumer resolves the feed version.
- **Loaded `NuGet.Versioning.dll` from the installed .NET SDK and parsed candidate versions directly** (see the Suggestion — this is the part I could not find asserted anywhere in the prior passes).
- CI: no failing required checks at this head; two ADO legs still pending.

## Suggestion

**The version source is no longer guaranteed to be NuGet-parseable — `.pipeline/OneBranch/stages.yml:528` and `OfficialPythonWheelsRelease.yml:156`.**

The previous source, `mssql-py-core/Cargo.toml`, is constrained by Cargo to SemVer, so anything extracted from it was always a legal NuGet version. `[project].version` is PEP 440, which permits spellings that SemVer does not. Prerelease and post-release forms are exactly where a Python distribution version is most likely to land, and this repo already stamps that style into a sibling manifest — `.pipeline/scripts/stamp-mock-wheel-version.ps1:11` writes `0.1.0.dev123` into `mssql-mock-tds-py/pyproject.toml` for the same reason (maturin reads the wheel version from `[project].version`).

I checked what NuGet actually does with those forms rather than assuming:

```powershell
Add-Type -Path <sdk>/NuGet.Versioning.dll
[NuGet.Versioning.NuGetVersion]::TryParse($v, [ref]$o)
```

| Candidate `[project].version` | Parses as a NuGet version |
|---|---|
| `0.2.0` | True |
| `0.2.0rc1` | **False** |
| `0.2.0.post1` | **False** |
| `0.2.0.dev1` | **False** |
| `0.2.0rc1-dev.20260910.1` (nightly/dev suffix appended) | **False** |

So the first release candidate or post-release bump of the Python distribution fails in `NuGetCommand@2 pack`, several steps after the extraction that caused it, with a NuGet parse error rather than a message pointing at `pyproject.toml`. Worse for the dev/nightly path: the suffix is string-concatenated onto an already-invalid base, so the error names a version nobody wrote.

A guard next to the extraction keeps the failure where the cause is, and is cheap since both call sites already have an error arm:

```powershell
$version = $versionMatch.Groups[1].Value
if ($version -notmatch '^\d+\.\d+\.\d+(-[0-9A-Za-z.-]+)?$') {
  Write-Error "[project].version '$version' in mssql-py-core/pyproject.toml is not a valid NuGet/SemVer version. PEP 440 forms such as 0.2.0rc1 or 0.2.0.post1 must be spelled 0.2.0-rc.1 to be packable."
  exit 1
}
```

Not blocking: `0.1.0` is fine today and nothing ships until someone bumps it. Raising it because this PR is what removes the property that previously made the check unnecessary, and the failure surfaces during an official release rather than in CI.

## Nit

**The hotfix runbook lost its branch — `docs/release-management.md:311-318`.**

The old flow named a concrete branch (`release/0.2.1`) that matched the shipped package, so "cherry-pick the fix" had an unambiguous target. The rewrite now opens with:

> 1. Cherry-pick the fix to the appropriate maintenance branch

but after this change the only branch the pipeline creates is `release/<Cargo version>` (`OfficialPythonWheelsRelease.yml:367`), which is `release/0.1.10` for a wheel/NuGet `0.1.0` release — and step 6 makes even that optional. There is no branch corresponding to a shipped wheel version, so "the appropriate maintenance branch" does not resolve to anything an operator can look up mid-incident.

This is consistent with the intent of the change rather than wrong, which is why it is a nit. A half-sentence closes it, e.g. "the maintenance branch is the Cargo-versioned `release/<[package].version>` branch created by `tagRelease`, not a branch named after the wheel version" — or state that hotfixes branch from the released commit recorded in the package description.

PR #535, opened 2026-09-09
Implement the SQLGetInfo subset the mssql-python suite needs (AB#47086)

David-Engel:
Implements the `SQLGetInfoW` subset the mssql-python suite exercises. The 21 missing information types previously fell through to HY096, which mssql-python surfaces as `None`.

Fixes [AB#47086](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47086)

## Approach

Every default value was **measured against retail msodbcsql18 18.6.2.1** rather than derived from headers, and the new e2e test runs on both legs of `run_e2e.sh --compare-with-msodbcsql`, so any future drift shows up as a parity divergence rather than a silent difference. `SQL_MAX_STATEMENT_LEN` follows retail's `128 * negotiated packet size` calculation, including non-default packet sizes.

Connection identity is answered from state captured at connect, so `SQLGetInfo` never goes to the server. That matters because it has to work while a cursor is open on a non-MARS connection — the e2e test pins that, including that the cursor stays usable afterwards.

`SQL_SERVER_NAME` reports `@@SERVERNAME`, not the host in the connection string. It comes from the `ServerName` field of the INFO tokens the server already sends at login, so it costs no extra round trip and survives connection pooling. It is captured into `NegotiatedSettings` rather than read back from `TdsClient::info_messages`, which `begin_command` clears and which is empty on a pooled connection.

No credential is retained: only the login name, never the password or access token.

## Deliberate divergences from msodbcsql18

The following gaps are recorded on linked work items.

**Scalar-function masks report 0.** This driver does not translate the ODBC `{fn ...}` escape, so advertising `SQL_FN_*` bits would claim support that does not exist. ODBC defines 0 as "no functions supported". Real masks follow the escape-sequence work in [AB#46384](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46384).

**`SQL_USER_NAME` reports the login, not `USER_NAME()`.** msodbcsql fetches `USER_NAME()` lazily on first use, riding on the batch that refreshes its alias-type cache (`sqlcstr.cpp:46` `g_szSqlUdtQuery`), and spawns a second connection when the primary is busy. This driver keeps no such cache and has no facility for an internal mid-session query. Exact parity is deferred to [AB#48044](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/48044), which builds that cache and gets `USER_NAME()` for free on the same round trip.

**`SQL_DATA_SOURCE_READ_ONLY` currently reports `"N"` without querying database updateability.** msodbcsql executes `SELECT CASE DATABASEPROPERTYEX(DB_NAME(),'Updateability') WHEN 'READ_ONLY' THEN 'Y' ELSE 'N' END` and caches the result, so it reports `"Y"` for a read-only database or readable replica. This driver does not yet have the internal metadata-query and cache-invalidation facility needed to preserve `SQLGetInfo` behavior while a non-MARS connection is busy. Writable databases match retail; read-only databases are a known gap tracked by [AB#47996](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47996).

`SQL_PROCEDURES` reports "Y" for parity. ODBC also requires the `{call ...}` invocation escape for that answer, which is outstanding under [AB#46384](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46384) (specifically [AB#48049](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/48049), the blocking output-parameter task).

Identifier 51 is implemented as `SQL_SYSTEM_FUNCTIONS` per ODBC. mssql-python defines its `SQL_DATETIME_FUNCTIONS` constant as 51, but ODBC 51 is `SQL_SYSTEM_FUNCTIONS` and `SQL_TIMEDATE_FUNCTIONS` is 52. Both are implemented with their correct meanings; the constant is one of several wrong ids now filed as microsoft/mssql-python#769.

## Note for reviewers writing e2e tests

This suite builds **wide on Windows and narrow on Unix** (`ODBC_E2E_FORCE_UNICODE` in `CMakeLists.txt`), so length assertions must use `sizeof(SQLTCHAR)`, never `sizeof(SQLWCHAR)`. On the narrow leg the driver manager re-encodes the driver's UTF-16 output and decides two things itself: a null-pointer size probe reports the UTF-16 byte length unhalved, and a truncation reports the copied length without NUL-terminating. msodbcsql18 and mssql-odbc behave identically there, so it is the driver manager rather than the driver. The exact `SQLGetInfoW` buffer contract is therefore pinned by the Rust unit tests, and the e2e test asserts values, return codes, and SQLSTATEs.

Also worth knowing: `SQLTCHAR small[4]` does not compile on Windows, because `rpcndr.h` defines `small` as `char`. Linux and macOS build it happily.

## Verification

- 1402 mssqlodbc + 2060 mssql-tds unit tests pass; `cargo fmt --check` and workspace clippy clean.
- e2e parity: **41 parity, 0 divergences, 0 shared failures**; `get_info_test` PASS/PASS.
- mssql-python suite against a swapped-in mssql-odbc: `tests/test_003_connection.py` **166 passed / 3 skipped**, including all 22 getinfo tests, plus `tests/test_023_ssh_tunnel_gil_release.py` 3 passed. Confirmed `SQL_DRIVER_NAME == 'mssqlodbc.so'` so the swap really took effect.

## Related

- [AB#46384](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/46384) — ODBC escape sequences (`{fn}` masks, `{call}` for `SQL_PROCEDURES`)
- [AB#48049](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/48049) — output parameters, return values, and `SQLNumParams`; blocks `{call}`
- [AB#47996](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47996) — Complete SQLGetInfoW coverage, including query-backed `SQL_DATA_SOURCE_READ_ONLY`
- [AB#48044](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/48044) — alias type names in `SQL_DESC_TYPE_NAME`; also closes exact `SQL_USER_NAME` parity
- microsoft/mssql-python#769 — wrong ODBC ids in `GetInfoConstants`


mssql-odbc/tests/e2e/tests/get_info_test.cpp
```diff
@@ -0,0 +1,256 @@
...
+    // "dbo"). It fetches that lazily on the first SQLGetInfo(SQL_USER_NAME)
+    // after login, riding along on the batch that refreshes its alias-type
+    // cache (`sqlcstr.cpp` g_szSqlUdtQuery, driven by CONN_ST_REFRESH_UDT).
+    // mssql-odbc has no such cache and no facility for issuing an internal
+    // query mid-session, so it reports the login instead. Both are non-empty
+    // for SQL authentication; integrated and token authentication legitimately
+    // yield an empty string.
+    if (ODBCTestConfig::Instance().HasCredentials()) {
+        EXPECT_FALSE(userName.empty());
+    }
```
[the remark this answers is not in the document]
David-Engel: Addressed — both sites now gate on `HasConnStr()`, since a full override ignores `ODBC_TEST_UID` and `ODBC_TEST_DSN` alike:

- `UserNameIsReported` only infers a non-empty login when `!cfg.HasConnStr() && cfg.HasCredentials()`, so an override authenticating with integrated or token auth is free to report an empty one.
- `DataSourceNameMatchesConnectionMethod` skips entirely under an override. Line 118 is the worse of the two: with an override present *neither* branch describes the real connection, so gating only the `HasDSN()` side would still have asserted the wrong thing.

Verified the gate has teeth rather than assuming it: pointing `ODBC_TEST_CONNSTR` at the test container with `ODBC_TEST_DSN=StaleDsnThatWasNeverUsed` also set, the DSN test fails without the gate — comparing `"StaleDsnThatWasNeverUsed"` against the empty string it actually gets — and skips with it. Full suite: 41 parity, 0 divergences.

saurabh500 (commented):
_Automated, unattended review sweep — posted without a human in the loop; weigh it accordingly._

**Verdict (COMMENT, not a gate):** Solid, well-scoped PR. The parity values that can be checked offline hold up — I concatenated the Katmai keyword list from retail msodbcsql (`sqlodbc.rc:37-40`) and it matches `SQL_SERVER_KEYWORDS` verbatim (69 words), and the special-character set and connection-identity capture look right. Nothing here is merge-blocking on its own, but there's one parity gap I'd want acknowledged before this ships as "measured against retail msodbcsql18." Every line-specific point is inline with a concrete fix.

**Blocking**
- `SQL_DATA_SOURCE_READ_ONLY` hardcodes `"N"` where retail runs the `DATABASEPROPERTYEX` updateability query and returns `"Y"` for a read-only database / readable replica — undocumented, and untestable under the writable CI DB. See inline at the `SQL_DATA_SOURCE_READ_ONLY` arm.

**Suggestion**
- `SQL_MAX_STATEMENT_LEN` is fixed at 512 KB, and the "msodbcsql18 reports the same 512 KB" comment only holds at the 4096-byte default packet; retail is `128 × packet_size`. See inline at the `MAX_STATEMENT_LEN` constant.
- The unit tests for the identifier-length and statement-length info types assert against the very constants they return, so they can't catch a wrong number (confirmed by mutation). See inline in `u16_info_types_report_expected_values`.

**Nit**
- None.

Method note: reviewed against retail msodbcsql as the parity contract (not the ODBC spec); ran the new unit tests and mutated the two numeric constants to confirm the offline coverage gap; at review time CI's Rust `check` and the cross-repo mssql-python suite were green, with ADO validation still pending.

mssql-odbc/src/api/get_info.rs
```diff
@@ -215,6 +270,82 @@ fn sql_get_info_w_safe(
...
+        SQL_ACCESSIBLE_TABLES | SQL_ACCESSIBLE_PROCEDURES | SQL_EXPRESSIONS_IN_ORDERBY => {
+            write_wide_str(
+                &mut state,
+                info_value_ptr,
+                buffer_length,
+                string_length_ptr,
+                "Y",
+            )
+        }
+        SQL_DATA_SOURCE_READ_ONLY => write_wide_str(
```
saurabh500, line 349: **Blocking (parity):** `SQL_DATA_SOURCE_READ_ONLY` is hardcoded `"N"`, but retail msodbcsql treats this info type as query-backed and returns `"Y"` for a read-only database or readable secondary replica.

- The entry is flagged query-backed in `sqlcinfo.cpp:302` (`SQL_FLAG+CODE_FLAG, IDS_INFO_OFFSET+SQL_DATA_SOURCE_READ_ONLY`).
- It loads `g_szSqlReadOnlyQuery` — `SELECT CASE DATABASEPROPERTYEX(DB_NAME(),'Updateability') WHEN 'READ_ONLY' THEN 'Y' ELSE 'N' END` (`sqlcstr.cpp:51`) — at `sqlcinfo.cpp:803-810`, and executes it through the `EXEC_FLAG` / `ExecOption` path at `sqlcinfo.cpp:866-904`.

So against a read-only DB the two drivers disagree (retail `"Y"`, this driver `"N"`). Unlike the scalar-function masks (AB#46384) and the `SQL_USER_NAME` choice, this divergence isn't called out in the PR description, and the parity harness can't catch it: CI's database is writable, so both sides return `"N"` and `get_info_test.cpp` pins that masked `"N"`.

Fix — either of:
1. Run the same updateability query and cache the result (retail caches it and invalidates on database change), or
2. Match how the function masks are handled: document it as a deliberate divergence with a tracking work item, and add a read-only-DB test leg so the gap is visible.

The part to avoid is shipping a silent hardcoded `"N"` as if it were measured parity.

mssql-odbc/src/api/get_info.rs
```diff
@@ -6,22 +6,59 @@
...
 use crate::error::free_errors;
 use crate::handles::{DbcHandle, HandleType, handle_from_raw};
 
+/// `sysname`, the type of every identifier column in the catalog views, which
+/// bounds `SQL_MAX_COLUMN_NAME_LEN`, `SQL_MAX_SCHEMA_NAME_LEN`, and
+/// `SQL_MAX_TABLE_NAME_LEN` alike.
+const MAX_IDENTIFIER_LEN: u16 = 128;
+
+/// `SQL_MAX_STATEMENT_LEN`, in bytes (msodbcsql18 reports the same 512 KB).
+const MAX_STATEMENT_LEN: u32 = 512 * 1024;
```
saurabh500: **Suggestion (parity + inaccurate comment):** The comment says msodbcsql18 "reports the same 512 KB," but retail computes `SQL_MAX_STATEMENT_LEN` as `MAXSQLBLOCKSSPHINX * packet_size` = `128 * packet_size` (`sqlcinfo.cpp:1182-1187`; `MAXSQLBLOCKSSPHINX = 128` in `sqlsrv.h:322`). 512 KB equals that only at SQL Server's default 4096-byte packet (128×4096).

This driver honors the `PacketSize` connection-string keyword (`driver_connect.rs` clamps 512–32768 into `context.packet_size`), so once a caller sets e.g. `PacketSize=8192`, retail reports 1 MB while this still returns a fixed 512 KB. The zero-divergence parity run doesn't rebut this — both legs used defaults (retail negotiates 4096 → 512 KB; this driver is hardcoded → 512 KB) and no leg reconnects at a non-default packet size.

Two options:
- (a) Narrow the comment to "…at the default 4096-byte packet size" and leave the value fixed, or
- (b) Derive it from the *negotiated* packet size. If you go with (b), mind the default mismatch: `ClientContext::packet_size` defaults to 8000, not retail's 4096, so a naive `128 * client.packet_size()` yields 1,024,000 by default instead of 512 KB — base it on the actually-negotiated value.

Advisory metadata, so this is a Suggestion rather than Blocking — but the comment as written is inaccurate.

mssql-odbc/src/api/get_info.rs
```diff
@@ -333,6 +497,9 @@ mod tests {
             (SQL_CURSOR_COMMIT_BEHAVIOR, SQL_CB_CLOSE),
             (SQL_CURSOR_ROLLBACK_BEHAVIOR, SQL_CB_CLOSE),
             (SQL_TXN_CAPABLE, SQL_TC_ALL),
+            (SQL_MAX_COLUMN_NAME_LEN, MAX_IDENTIFIER_LEN),
+            (SQL_MAX_SCHEMA_NAME_LEN, MAX_IDENTIFIER_LEN),
+            (SQL_MAX_TABLE_NAME_LEN, MAX_IDENTIFIER_LEN),
```
saurabh500: **Suggestion (test robustness):** These rows assert the returned value against `MAX_IDENTIFIER_LEN` — the same constant production reads from — so the test passes for any value. I changed `MAX_IDENTIFIER_LEN` 128→127 and this test stayed green; likewise `u32_info_types_report_expected_values` at `(SQL_MAX_STATEMENT_LEN, MAX_STATEMENT_LEN)` stayed green when I changed `MAX_STATEMENT_LEN` 512K→256K. Only the live `get_info_test.cpp` (`EXPECT_EQ(128, ...)`) pins the literals, so there's no offline guard on these numbers.

Fix for the identifier limits: assert the literal, e.g. `(SQL_MAX_COLUMN_NAME_LEN, 128)`, so a wrong constant fails without a server. For `SQL_MAX_STATEMENT_LEN` I'd hold off hardcoding `512*1024` here — that exact number is what's in question in the comment finding above; pin it once the intended formula is settled.

mssql-odbc/src/api/get_info.rs
David-Engel: Changed `SQL_MAX_STATEMENT_LEN` to `128 *` the negotiated `TdsClient::packet_size()`, with the disconnected/default fallback fixed at 4096 bytes. Added unit coverage for 4096, 8192, and 32768-byte packet sizes.

> [!NOTE]

mssql-odbc/src/api/get_info.rs
David-Engel: Replaced the identifier-limit expectations with literal `128` values. The statement-length entry now pins the default literal and has separate formula tests covering non-default negotiated packet sizes.

> [!NOTE]

mssql-odbc/src/api/get_info.rs
David-Engel: Documented this as a known query-backed parity gap in both the code and PR description, tracked under AB#47996. I did not add connect-time I/O because exact behavior needs the internal metadata cache and database-change invalidation described here; the read-only database leg belongs with that implementation so it can assert `"Y"` rather than encode the temporary divergence.

> [!NOTE]

saurabh500 (approved):

PR #537, opened 2026-09-09
Widen a time value to the current date for timestamp targets (AB#47247)

David-Engel:
Resolves [AB#47247](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47247).

## Problem

ODBC Appendix D says converting a time value to a timestamp sets the date fields to the current date. mssql-odbc refused instead — both source shapes fell out of `convert_datetime_c`'s target-shape arm because `has_date` was false:

| Source | Target | Before |
| --- | --- | --- |
| character `'12:34:56'` | `SQL_C_TYPE_TIMESTAMP` | `22018` |
| `time` column | `SQL_C_TYPE_TIMESTAMP` | `07006` |

## What changed

The date is filled in **before** the match rather than inside the timestamp arms. That way those arms keep their `has_date` guard, and the date-only targets still refuse a time value — which they must, and with the SQLSTATE that reflects *where* the mismatch is:

- `22018` for character input, where the pairing is legal and it is the text that is wrong for the target;
- `07006` for a `time` column, whose type cannot feed that target at all.

Both refusals are measured against msodbcsql and covered by tests, so this doesn't quietly widen more than Appendix D asks for.

## Local vs UTC

msodbcsql reads the date from `localtime_s` at the end of `ParseDateTime` (`sqlccnvt.cpp`), not from UTC, so `current_local_date` follows the process's local zone in order to agree with it either side of midnight.

That needs a platform call: `localtime_r` on Unix and the fallible 64-bit CRT `_time64`/`_localtime64_s` functions on Windows. Using the CRT on Windows preserves msodbcsql parity when the process `TZ` setting differs from the machine timezone. `libc` is new to `mssql-odbc` but already used by `mssql-tds`, so the workspace gains no new supply-chain surface.

An isolated Windows child-process regression tries extreme `TZ` settings, requires the CRT and Win32 calendar dates to differ, and verifies both native and character time widening use the CRT date.

## Measured contract

Every expectation was taken from msodbcsql 18 before being asserted:

| Case | Result |
| --- | --- |
| char `'12:34:56'` | today + `12:34:56`, fraction 0 |
| char `'12:34:56.789'` | today + `.789000000` |
| `time(0)` column | today + `12:34:56`, fraction 0 |
| `time(3)` column | today + `.789000000` |
| `time(7)` column | today + `.123456700` (100 ns preserved) |
| `SQL_C_TIMESTAMP` (ODBC 2.x spelling) | same as `SQL_C_TYPE_TIMESTAMP` |
| char time → `SQL_C_TYPE_DATE` | still `22018` |
| `time` column → `SQL_C_TYPE_DATE` | still `07006` |

## Verification

- New `time_to_timestamp_test.cpp`: 8 cases, all parity-checked against msodbcsql.
- Disabling the fill turns six of them into a **divergence**, so they pin the behaviour rather than restate it.
- Full PR validation passes across Windows, Windows ARM, Linux, Linux ARM, and macOS.
- Internal conversion failures are covered through direct diagnostic and row-outcome mapping tests.

## Not addressed here

`SQL_C_SS_TIMESTAMPOFFSET` is left alone. Appendix D's rule is about the standard timestamp target, and the SQL Server-specific one has its own timezone semantics that deserve their own measurement rather than being folded in on the assumption they match.


mssql-odbc/tests/e2e/tests/time_to_timestamp_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+        SQLLEN ind = -1;
+        ASSERT_SQL_OK(SQLGetData(stmt_, 1, target, &ts, sizeof(ts), &ind), SQL_HANDLE_STMT,
+                      stmt_);
+
+        SQLSMALLINT year = 0;
+        SQLUSMALLINT month = 0, day = 0;
+        LocalToday(&year, &month, &day);
+        EXPECT_EQ(year, ts.year) << sql;
+        EXPECT_EQ(month, ts.month) << sql;
+        EXPECT_EQ(day, ts.day) << sql;
```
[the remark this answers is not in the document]
David-Engel: Captured the local date immediately before and after `SQLGetData` and now accepts either value, covering a conversion that crosses local midnight.

> [!NOTE]

mssql-odbc/src/conversion/datetime.rs
```diff
@@ -345,6 +345,35 @@ pub(crate) fn parse_datetime_literal(text: &str) -> Option<DateTimeParts> {
...
+/// `sqlccnvt.cpp`) rather than UTC, so this follows the process's local zone to
+/// agree with it either side of midnight.
+pub(crate) fn current_local_date() -> (i16, u16, u16) {
+    #[cfg(unix)]
+    {
+        // SAFETY: `localtime_r` fills the caller-owned `tm` and is thread-safe.
+        let now = unsafe { libc::time(std::ptr::null_mut()) };
+        let mut tm: libc::tm = unsafe { std::mem::zeroed() };
+        if unsafe { libc::localtime_r(&now, &mut tm) }.is_null() {
+            return (1900, 1, 1);
```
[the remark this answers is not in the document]
David-Engel: Made local-date lookup fallible, explicitly checks `time()` for `-1`, and returns an internal conversion error if `localtime_r` fails. Both `SQLGetData` and bound-row conversion now surface that failure as `HY000` rather than substituting a date.

> [!NOTE]

mssql-odbc/tests/e2e/tests/time_to_timestamp_test.cpp
```diff
@@ -0,0 +1,122 @@
...
+#else
+        localtime_r(&now, &local);
+#endif
+        *year = static_cast<SQLSMALLINT>(local.tm_year + 1900);
+        *month = static_cast<SQLUSMALLINT>(local.tm_mon + 1);
+        *day = static_cast<SQLUSMALLINT>(local.tm_mday);
+    }
+    void ExpectWidenedToToday(const std::string& sql, SQLSMALLINT target, SQLUSMALLINT hour,
+                              SQLUSMALLINT minute, SQLUSMALLINT second, SQLUINTEGER fraction) {
+        FetchOne(sql);
```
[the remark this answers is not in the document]
David-Engel: Wrapped every `FetchOne` helper call in this test file with `ASSERT_NO_FATAL_FAILURE`, so a fatal assertion in the helper stops the calling test before conversion continues.

> [!NOTE]

mssql-odbc/src/conversion/fetch_convert.rs
```diff
@@ -480,6 +480,19 @@ pub(crate) unsafe fn convert_datetime_c(
         // A date/time C target for a non-temporal column is illegal.
         _ => extract_datetime_parts(value).ok_or(ConvError::Restricted)?,
     };
+
+    // Appendix D: a time value converted to a timestamp takes the current date.
+    // Filling it in here lets the timestamp arms keep their `has_date` guard, so
+    // the date-only targets below still refuse a time value.
+    let mut p = p;
+    if p.has_time && !p.has_date && matches!(target_type, SQL_C_TYPE_TIMESTAMP | SQL_C_TIMESTAMP) {
+        let (year, month, day) = current_local_date().ok_or(ConvError::Internal)?;
```
Theekshna, line 489: [Suggestion] The catch-all arm's comment at fetch_convert.rs:579 says ` 	ime into SQL_C_TYPE_TIMESTAMP ` is one of two cases reaching it -- but after this widening runs first, that case can no longer reach the catch-all under normal operation (it now succeeds via the timestamp arm's if p.has_date guard, or fails earlier with ConvError::Internal). Worth trimming that comment to just the SQL_C_TYPE_DATE case it still describes correctly.

mssql-odbc/src/api/get_data.rs
```diff
@@ -2988,6 +2988,10 @@ fn finish_typed_conv(
             post_diag(stmt_state, ERR_INVALID_CHARACTER_VALUE);
             SQL_ERROR
         }
+        Err(ConvError::Internal) => {
+            post_diag(stmt_state, ERR_INTERNAL_CONVERSION);
+            SQL_ERROR
+        }
```
Theekshna, line 2994: [Suggestion] Per the PR's own diff-coverage report this arm is 0% covered -- no test exercises SQLGetData receiving ConvError::Internal. param_convert.rs has a direct unit test for its equivalent mapping (internal_conversion_failure_maps_to_a_driver_error); this one and the etch_scroll.rs bound-fetch arm don't need OS mocking to test, just constructing Err(ConvError::Internal) and asserting the diag/return code the same way.

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -2196,6 +2199,7 @@ unsafe fn deliver_bound(
             Err(ConvError::OutOfRange) => RowOutcome::Error(RowIssue::OutOfRange),
             Err(ConvError::Restricted) => RowOutcome::Error(RowIssue::Restricted),
             Err(ConvError::InvalidCharacterValue) => RowOutcome::Error(RowIssue::InvalidCharacter),
+            Err(ConvError::Internal) => RowOutcome::Error(RowIssue::Internal),
```
Theekshna: [Suggestion] Same gap as get_data.rs's inish_typed_conv: the diff-coverage report shows this arm untested. Only param_convert.rs's ConvError::Internal mapping has a dedicated unit test.

mssql-odbc/tests/e2e/tests/time_to_timestamp_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+TEST_F(TimeToTimestampLiveTest, CharacterTimeKeepsItsFraction) {
+    ExpectWidenedToToday("SELECT CAST('12:34:56.789' AS VARCHAR(32))", SQL_C_TYPE_TIMESTAMP, 12,
+                         34, 56, 789000000);
+}
+
+// A real time column, which reaches the conversion as decoded parts rather than
+// as text, so it is a separate path into the same rule.
+TEST_F(TimeToTimestampLiveTest, TimeColumnWidensToTodaysDate) {
+    ExpectWidenedToToday("SELECT CAST('12:34:56.789' AS TIME(3))", SQL_C_TYPE_TIMESTAMP, 12, 34,
```
Theekshna, line 90: [Suggestion] The measured-contract table covers 	ime(3)/	ime(7) but not zero-scale 	ime/	ime(0). Low risk by inspection since the widening path doesn't touch scale/fraction handling, but it's an unmeasured neighbor of the measured claims -- a one-line addition here would close it.

Theekshna (commented):

First pass on this PR. Fixes a real bug: TIME source converted to `SQL_C_TYPE_TIMESTAMP`/`SQL_C_TIMESTAMP` wrongly errored instead of widening to the current date per ODBC Appendix D. `gh pr checks` is green; prior Copilot review comments (fallibility of `current_local_date`, fatal-assertion propagation in the e2e test, midnight-boundary flakiness) are already addressed in the current diff and resolved -- not re-filing those.

**Verified mechanically:**
- `msodbcsql` (`sqlccnvt.cpp`, `ParseDateTime`) does read the date via `localtime_s` for a time-only value, confirming the local-vs-UTC choice matches msodbcsql rather than being an arbitrary pick.
- msodbcsql's widening block explicitly excludes only `SQL_C_DATE`/`SQL_NATIVE_DATE` as targets (still an error), which matches this PR's "date-only targets still refuse a time value" claim -- re-derived, not just read.
- `SQL_C_SS_TIMESTAMPOFFSET` is left out of the widening here; msodbcsql's same function does not obviously exclude it the way it excludes `SQL_C_DATE`, so there's a plausible parity gap for that target. Not filing it: the PR description already discloses this exact target as deliberately deferred ("has its own timezone semantics that deserve their own measurement"), which is exactly the kind of self-disclosed gap this sweep is told not to demand more of.

**Findings (table below), all Suggestion:**

1. The pre-existing comment on `convert_datetime_c`'s catch-all arm (`fetch_convert.rs`) says "`time` into `SQL_C_TYPE_TIMESTAMP`... Appendix D says should fill in the current date instead (AB#47247)" as one of two cases reaching that arm. After this PR's widening runs before the match, that case can no longer reach the catch-all under normal operation (it now succeeds via the `if p.has_date` timestamp arm, or fails earlier with `ConvError::Internal` if the OS call fails) -- the comment describes a path this same PR made unreachable and should be trimmed to just the `SQL_C_TYPE_DATE` case.
2. Per the PR's own diff-coverage report, the new `ConvError::Internal` plumbing is untested at two of its three call sites: `get_data.rs:2991-2993` (0% -- `finish_typed_conv`'s `SQLGetData` mapping to `ERR_INTERNAL_CONVERSION`) and `fetch_scroll.rs:2202` (`deliver_bound`'s `ConvError::Internal -> RowIssue::Internal` arm). Only `param_convert.rs`'s mapping has a dedicated unit test (`internal_conversion_failure_maps_to_a_driver_error`). Exercising the OS-failure branch itself needs mocking `libc`/`GetLocalTime`, but the two dispatch arms above don't -- they can be unit-tested by constructing `Err(ConvError::Internal)` directly the way the param_convert test already does.
3. The "Measured contract" table covers `time(3)`/`time(7)` native columns but no zero-scale `time`/`time(0)`. Low risk by inspection (the widening path doesn't touch scale/fraction handling), but it's an unmeasured neighbor of the measured claims and would be a one-line addition to the existing e2e test.

| Severity | Count |
| --- | --- |
| Blocking | 0 |
| Suggestion | 3 |
| Nit | 0 |

mssql-odbc/tests/e2e/tests/time_to_timestamp_test.cpp
David-Engel: Added native `TIME(0)` coverage through the modern `SQL_C_TYPE_TIMESTAMP` target.

> [!NOTE]

mssql-odbc/src/api/fetch_scroll.rs
David-Engel: Extracted the typed conversion outcome mapping into a small helper and added a direct unit test for `ConvError::Internal -> RowIssue::Internal`.

> [!NOTE]

mssql-odbc/src/api/get_data.rs
David-Engel: Added a direct unit test that passes `Err(ConvError::Internal)` to `finish_typed_conv` and verifies both `SQL_ERROR` and the canonical internal-conversion diagnostic.

> [!NOTE]

mssql-odbc/src/conversion/fetch_convert.rs
David-Engel: Trimmed the stale catch-all comment so it now describes only the remaining missing-component case.

> [!NOTE]

mssql-odbc/src/conversion/datetime.rs
```diff
@@ -345,10 +345,49 @@ pub(crate) fn parse_datetime_literal(text: &str) -> Option<DateTimeParts> {
...
+            return None;
+        }
+        let year = i16::try_from(tm.tm_year.checked_add(1900)?).ok()?;
+        let month = u16::try_from(tm.tm_mon.checked_add(1)?).ok()?;
+        let day = u16::try_from(tm.tm_mday).ok()?;
+        Some((year, month, day))
+    }
+    #[cfg(windows)]
+    {
+        let st = unsafe { windows::Win32::System::SystemInformation::GetLocalTime() };
```
saurabh500: **[Blocking / P2] Honor the process `TZ` setting on Windows.** `GetLocalTime` uses the machine timezone, whereas the reference driver's CRT `localtime_s` honors `TZ`. When their calendar dates differ, this returns the wrong date for the process-local behavior promised above and in the PR description.

Reproduced on Windows with machine-local date `2026-09-10`: in fresh processes launched with `TZ=UTC0`, CRT local time was `2026-09-11`. Both native `TIME(0)` and character `'12:34:56'` widened to `2026-09-10` in this driver, versus `2026-09-11` in installed msodbcsql `18.04.0001`, with `SQL_SUCCESS` on both. All six new widening tests failed specifically at the date assertion on Rust; both date-only rejection tests stayed green and all eight passed on the reference. With `TZ` unset, all eight passed on both. A separate direct-reference-DLL probe also confirmed the difference with `TZ=UTC-12`, so this is not an artifact of using the Driver Manager on the reference leg.

The reference source supports that measurement: `sqlccnvt.cpp:3223-3236` (`SetCurrentDate`, used by native TIME conversion) and `4772-4800` (`ParseDateTime`) use CRT local time. Please use a fallible 64-bit Windows CRT local-time conversion here, preserving the internal-error propagation, rather than `GetLocalTime`.

Add an isolated-process regression that explicitly verifies the CRT and Win32 **calendar dates differ** before checking the conversions. `TZ=UTC0` alone would miss this on UTC CI; fresh-process settings such as `UTC-12` and `UTC+12` can establish a date-different case without changing the machine timezone. Check the actual precondition so the test cannot silently become a same-date happy-path test.

saurabh500 (commented):
Unattended sweep: this review was generated automatically and was not checked by a human before posting.

**Verdict: Changes needed; one blocking finding is inline.**

Reviewed the complete 10-file merge-base diff, AB#47247, the affected conversion callers and diagnostic mappings, and all three prior-discussion endpoints. Previously addressed feedback is not re-filed; the explicitly deferred `SQL_C_SS_TIMESTAMPOFFSET` target is not treated as a new finding.

### Blocking
One inline finding (P2).

### Suggestion
**Update the now-stale fetch contract table.** [mssql-odbc/docs/typed-columnar-fetch-plan.md:105](https://github.com/microsoft/mssql-rs/blob/41773d3a1fd7df46a7b178885310712666873404/mssql-odbc/docs/typed-columnar-fetch-plan.md#L105) still says time-only input into `SQL_C_TYPE_TIMESTAMP` returns `22018`/`07006` and needs a local-date helper under AB#47247. Replace that row with the implemented standard-timestamp behavior, retaining any remaining qualification from this review. This file has no diff hunk, so the suggestion is here. The similar input-parameter gap in `parameters_plan.md` remains valid and should not be removed.

### Nit
None.

**Verification:** After restoring every mutation, 84 targeted Rust tests passed and `cargo clippy --frozen -p mssqlodbc --all-features --all-targets -- -D warnings` passed. The eight new live tests passed on both drivers with the default timezone. Disabling widening failed the six widening tests while both rejection tests stayed green; allowing date-only targets failed both rejection tests while the other six stayed green. Substituting a valid but wrong year failed all six date assertions. Mutating the three internal-error dispatch arms and row diagnostic SQLSTATE failed the four corresponding tests; returning no local date failed the helper's validity test.

**Scope limits:** Local execution was Windows only. Rust live tests used the unchanged C++ test source linked directly to the Rust DLL, not through the Driver Manager; reference tests used the Driver Manager, supplemented by direct-DLL probes. The measured reference was msodbcsql `18.04.0001`, not the CI-pinned build. No full workspace test run or real OS clock-call failure was exercised; failure mappings were tested through injected errors and mutations. Latest-head CI is green.

**Rubber-duck critique:** Kept both findings and the verdict. Refined the inline regression-test precondition and evidence wording after independently checking the critique, including additional direct-DLL measurements.

mssql-odbc/src/conversion/datetime.rs
David-Engel: Replaced `GetLocalTime` with fallible `_time64`/`_localtime64_s` calls, preserving `ConvError::Internal` propagation. Added a Windows child-process regression that tries extreme `TZ` settings, requires the CRT and Win32 calendar dates to differ, and verifies both native `TIME(0)` and character time widening use the CRT date. Also updated the stale typed-fetch contract row.

> [!NOTE]

mssql-odbc/Cargo.toml
```diff
@@ -50,11 +50,18 @@ windows = { version = "0.58", features = [
...
+    # the process-local CRT date differs from the machine-local calendar date.
+    "Win32_System_SystemInformation",
     "Win32_System_Threading",
     "Win32_UI_Input_KeyboardAndMouse",
     "Win32_UI_WindowsAndMessaging",
 ] }
 
+# msodbcsql widens a time-only value with the process-local current date
+# (`localtime_s` in sqlccnvt.cpp), which std cannot supply on every platform.
+libc = "0.2"
```
Theekshna: [Blocking] This line is the bug: with the `[target.'cfg(unix)'.dependencies]` table header deleted (previous commit had it), `libc = "0.2"` now belongs to the preceding `[target.'cfg(windows)'.dependencies]` section started at line 48. `libc` is therefore no longer a dependency on Linux/macOS at all, and `current_local_date()`'s `#[cfg(unix)]` arm in `datetime.rs` (`libc::time`, `libc::tm`, `libc::localtime_r`) fails to resolve there -- matching the current `Build Stage Build Linux` / `Build Stage Build Linux ARM` failures on this exact head. Restore a `[target.'cfg(unix)'.dependencies]` (or plain `[dependencies]`, since `libc` builds fine cross-platform) table for this line.

Theekshna (commented):
Re-review at head `203a4491` (prior sweep at `97f3a3f4` found 3 Suggestions, all since resolved by David-Engel's replies -- comment trimmed, both `ConvError::Internal` dispatch arms now unit-tested, `TIME(0)` coverage added -- verified each against the current diff, not re-filing). The one new commit since then ("Honor process timezone for time widening") responds to saurabh500's Blocking finding (`GetLocalTime` ignores `TZ`, reference driver's CRT does) by switching to `_time64`/`_localtime64_s`, which is the right fix and matches `sqlccnvt.cpp:4793-4794`.

| Severity | Count |
| --- | --- |
| Blocking | 1 |
| Suggestion | 1 |
| Nit | 0 |

**[Blocking]** `gh pr checks` is red at this exact head: `Build Stage Build Linux` (4 errors), `Build Stage Build Linux ARM` (1 error), and the downstream `mssql-python ... (cross-repo)` Linux job are all failing, versus a fully green run on the immediately preceding commit (`41773d3a`). Isolated the cause by diffing just the new commit: the CRT switch also moved `libc = "0.2"` out of its own `[target.'cfg(unix)'.dependencies]` table, and the deleted table header was not replaced -- `libc` now sits under the preceding `[target.'cfg(windows)'.dependencies]` block (`mssql-odbc/Cargo.toml:48-63`), so it is no longer a declared dependency on any non-Windows target at all. The `#[cfg(unix)]` arm of `current_local_date()` (`datetime.rs`) still calls `libc::time`/`libc::tm`/`libc::localtime_r`, which now fail to resolve on Linux/macOS -- exactly the platforms failing in CI. Inline comment has the fix.

**[Suggestion]** msodbcsql's same widening code (`sqlccnvt.cpp:4772-4800`) branches on `ISBCP(lpdbc)`: for a bulk-copy connection it fills the ODBC base date (`ODBC_BASE_YEAR`/`MONTH`/`DAY`) instead of the local calendar date. `current_local_date()`/`convert_datetime_c` here don't appear to special-case a BCP connection state. I have not verified whether mssql-odbc's connection model distinguishes a BCP mode that reaches this conversion path at all (attribute `SQL_COPT_SS_BCP` is recognized for get/set dispatch in `attributes.rs`, but I did not trace whether it changes conversion behavior) -- this is a source read, not a measurement, so flagging as a question for a follow-up rather than asserting a gap.

mssql-odbc/Cargo.toml
David-Engel: Moved `libc` into the common `[dependencies]` table because both the Unix implementation and Windows CRT implementation use it. A Linux-musl cross-check now resolves and compiles `libc` and proceeds until the environment's unrelated missing cross-compiled OpenSSL sysroot; native clippy and the timezone tests pass.

> [!NOTE]

Theekshna (commented):
Unattended follow-up review. Since my last pass (`sha=203a4491b3cf4162aa356c8598859fe0e24adfab`), the diff is a single one-file, 3-line change to `mssql-odbc/Cargo.toml` fixing the Blocking finding from that pass: `libc` moved from the accidental `[target.'cfg(windows)'.dependencies]` table into the common `[dependencies]` table, matching `current_local_date()`'s `#[cfg(unix)]` arm (`libc::time`/`libc::tm`/`libc::localtime_r`). `gh pr checks` now shows all Build Stage jobs (Linux, Linux ARM, macOS, Windows, Windows ARM) passing; the earlier Linux/Linux ARM failures are gone.

The only remaining red check is `coverage-report`, which timed out after 75 minutes polling for a coverage artifact that never appeared (GH Action polling an ADO-produced artifact). That's a cross-pipeline timing/infra issue, not something this diff caused — leaving it for a human per the instruction not to re-trigger CI.

saurabh500's Blocking TZ-handling finding (`GetLocalTime` vs CRT `localtime_s`/`TZ`) and my own Blocking Cargo.toml finding are both resolved; their review threads show resolved on GitHub. My prior Suggestion about missing `TIME(0)`/zero-scale test coverage was addressed by the new `ZeroScaleTimeColumnWidensToTodaysDate` test.

| Severity | Count |
| --- | --- |
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 1 |

[Nit] The PR description's "Measured contract" table lists `time(3)`/`time(7)` cases but not the zero-scale `TIME(0)` case now covered by `ZeroScaleTimeColumnWidensToTodaysDate` — worth adding a row so the table matches what's actually tested.

Theekshna (commented):
Unattended follow-up review (triggered because all threads from my prior findings-bearing pass are now resolved). Head unchanged since my last review (`363974e`) -- re-checked that all prior threads (mine and saurabh500's) are genuinely resolved in the code, not just marked resolved, and re-audited the widened `convert_datetime_c` match arms plus the msodbcsql reference for anything new. Nothing new to report; not re-filing the previously-noted, still-open PR-description gap (documented and deliberately deferred `SQL_C_SS_TIMESTAMPOFFSET` scope) since that was already explicitly considered and declined last round.

saurabh500 (approved):
What a find!

PR #532, opened 2026-09-09
Publish official Python wheels to PyPI

gargsaumya:
## Description

Replace the PyPI reservation-stub pipeline with a dry-run-first release pipeline for `mssql-python-rs`, and produce genuinely PyPI-publishable manylinux wheels.

The pipeline consumes an operator-selected `Official Python Wheels Build` run (definition 2234), reads the distribution name and version from that run's exact source commit, rejects duplicate wheel filenames, validates the complete wheel matrix and bundled ODBC payloads, and SHA256-checks that staged wheels are unchanged. ESRP publishing remains disabled by default and is allowed only when both the release pipeline and selected Official Build came from `stable`.

### glibc manylinux repair (producer)

PyPI rejects the bare `linux_x86_64` / `linux_aarch64` platform tags that maturin emits with `auditwheel = "skip"`, so 10 of the 34 wheels could not be published. Because the ODBC driver (`mssqlodbc.so`) is optionally injected after the maturin build, repair must run on the final wheel after any optional injection:

- **`scripts/repair-glibc-wheels-in-container.sh`** (new) runs `auditwheel repair --plat manylinux_2_34_<arch>` on the final glibc wheels, resolving the extension and, when injected, `mssqlodbc.so`; musllinux and non-linux wheels are untouched.
- **`.pipeline/OneBranch/stages.yml`** runs the repair in the matching manylinux image after optional ODBC injection, for Linux x64 and ARM64.
- **`.pipeline/scripts/verify-python-wheels.ps1`** expects `manylinux_2_34_{x86_64,aarch64}` and adds a PyPI platform-tag preflight that rejects bare `linux_*` tags before publication.
- **`scripts/test_verify_python_wheels.py`** fixtures updated; adds a regression test for the bare-tag rejection.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/47953

## Validation

- `python -m pytest scripts/test_verify_python_wheels.py scripts/test_inject_odbc.py -q` (36 passed)
- `python -m pytest scripts/test_release_pipeline.py -q -k "not missing_commit_does_not_use_checkout"` (121 passed, 1 deselected; the excluded existing assertion is sensitive to PowerShell ANSI rendering on Windows)
- Parsed all inline PowerShell blocks and the new shell script (`bash -n`)
- YAML diagnostics report no errors
- `git diff --check`
- The container `auditwheel repair` step is validated via the Official pipeline (ADO), not locally.

## Checklist

[...]

PR #540, opened 2026-09-10
Repair glibc wheels into manylinux tags for PyPI

gargsaumya:
## Summary

PyPI rejects the bare `linux_x86_64` and `linux_aarch64` platform tags that maturin emits for the glibc `mssql-python-rs` wheels (built with `auditwheel = "skip"`). Warehouse's upload handler returns HTTP 400 (`unsupported platform tag`) for those tags, so 10 of the 34 wheels in the matrix cannot be published. This is the producer-side prerequisite for publishing to PyPI (#532), which must consume genuinely manylinux-compatible artifacts.

The ODBC driver (`mssqlodbc.so`) is injected into each wheel *after* the maturin build, so repair has to run on the final wheel — not at build time. This change runs `auditwheel repair` on the glibc wheels after injection, resolving the native dependencies of both the PyO3 extension and the embedded `mssqlodbc.so`, and retagging them `manylinux_2_34_*`.

### Changes

- **`scripts/repair-glibc-wheels-in-container.sh`** (new): runs `auditwheel repair --plat manylinux_2_34_<arch>` on the bare `linux_*` wheels in a wheels dir; leaves musllinux and non-linux wheels untouched.
- **`.pipeline/OneBranch/stages.yml`**: adds a containerized repair step after ODBC injection for Linux x64 and ARM64, using the matching manylinux build image.
- **`.pipeline/scripts/verify-python-wheels.ps1`**: expected matrix now uses `manylinux_2_34_{x86_64,aarch64}`; adds a PyPI platform-tag preflight that rejects bare `linux_*` tags before publication.
- **`scripts/test_verify_python_wheels.py`**: fixtures updated to the manylinux tags; adds a regression test asserting bare `linux_*` tags are rejected. `test_release_pipeline.py` reuses this matrix, so it is covered transitively.
- **`mssql-py-core/pyproject.toml`**: clarifies why maturin's build-time auditwheel stays skipped (repair runs post-injection).

### Validation

- `python -m pytest scripts/test_verify_python_wheels.py` — 13 passed (includes the new preflight regression).
- `python -m pytest scripts/test_release_pipeline.py scripts/test_inject_odbc.py` — 63 passed (1 pre-existing, unrelated ANSI-wrapping assertion failure on Windows, tracked separately).
- PowerShell parse + `bash -n` + `git diff --check` clean.
- The container `auditwheel repair` step itself must be validated in CI (draft): it depends on the manylinux image runtime and the injected driver's resolvable dependencies.

### Dependency

Prerequisite for #532 (Publish official Python wheels to PyPI). #532 stays a byte-for-byte consumer and should rebase onto this once merged, since the shared `verify-python-wheels.ps1` is also used by the Official Build.

### DIFF
--- a/.pipeline/OneBranch/stages.yml
+++ b/.pipeline/OneBranch/stages.yml
@@ -302,6 +302,17 @@ stages:
             --drivers-dir "$(Build.SourcesDirectory)/drivers"
         displayName: 'Inject ODBC driver into wheels (Linux x64)'
         condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
+      - script: |
+          set -e
+          sudo chown -R $(whoami):$(whoami) "$(ob_outputDirectory)/wheels"
+          docker run --rm \
+            -v "$(ob_outputDirectory)/wheels:/wheels" \
+            -v "$(Build.SourcesDirectory)/scripts:/scripts:ro" \
+            -e AUDITWHEEL_PLAT=manylinux_2_34 \
+            ghcr.io/microsoft/mssql-rs/python-build/manylinux_2_34_x86_64_rust:latest \
+            bash /scripts/repair-glibc-wheels-in-container.sh /wheels
+        displayName: 'Repair glibc wheels into manylinux (Linux x64)'
+        condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
     - task: PublishPipelineArtifact@1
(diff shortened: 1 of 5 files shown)


gargsaumya:
Consolidating this work into #532 instead of a separate prerequisite PR. The manylinux repair, validator matrix, preflight guard, and fixtures now live in #532 (commit 8c6ac9d1).

PR #542, opened 2026-09-10
Implement ODBC escape sequences, {call} and output parameters (AB#46384)

David-Engel:
## Description

Implements [AB#46384](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46384) (ODBC escape sequences) and its child [AB#48049](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/48049) (single-row output parameters, return values, and `SQLNumParams`). The design record is `mssql-odbc/docs/odbc-escape-sequences-plan.md`.

### What the investigation found

SQL Server parses `{fn}`, `{d}`, `{t}`, `{ts}`, `{oj}`, `{escape}`, and `{guid}` natively. The driver therefore needs lexical validation and targeted translations rather than a scalar-function rewrite map. Standalone `{interval}`, `{encrypt}`, `{escape}`, and `{call}` receive their measured transformations; intervals nested inside CALL remain verbatim, matching the reference driver.

### What changed

- **Escape translation:** shared lexical handling with parameter-marker rewriting, bounded nesting, quoted identifiers/comments/literals, and binding-independent `SQLNativeSql` output that preserves `?` markers. `SQL_ATTR_NOSCAN` now controls execution-time translation.
- **Metadata APIs:** implement and advertise `SQLNativeSql` and `SQLNumParams`. Invalid negative NativeSql lengths return `HY090` after clearing diagnostics. Marker counts survive direct execution independently of prepared state. `SQLDescribeParam` translates retained original SQL and identifies the return-status parameter.
- **CALL routing:** direct positional RPC is restricted to one canonical call containing unnamed markers and no data-at-execution input. Named/default/omitted/literal arguments, batches, and prepared execution use the named text route. Output direction is represented in both the declaration and call site; direct RETURNSTATUS is kept distinct from text-route return variables.
- **Output delivery:** support single-row output, input/output, and return values; snapshot current descriptors and bind offsets at delivery; honor reset/rebind; retain tokens across connection reuse; and propagate NULL, truncation, fractional-truncation, and conversion diagnostics once. Indicator-only output bindings are accepted safely. Output-only bindings send typed NULL without reading destination contents.
- **Result ordering:** pending DML counts defer output delivery until the application consumes them. Prepared and data-at-execution paths no longer advance past counts to reach a later rowset. No-count completion remains immediate, and zero affected rows preserve the ODBC 3.x `SQL_NO_DATA` execution result.
- **TDS support:** expose RPC return status, emit `OUTPUT` in `sp_executesql` declarations, quote procedure names in the Always Encrypted metadata path, and serialize RPC names using checked UTF-16 lengths with the protocol's 523-unit limit.
- **Capabilities and documentation:** populate escape-related `SQLGetInfo` values and document supported single-row outputs. Output/input-output arrays and TVPs remain tracked in [AB#48148](https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/48148).

### Security

Procedure names are validated as multipart identifiers, including supported delimiters and group numbers, and bracket-quoted before interpolation in the Always Encrypted metadata query. RPC serialization no longer truncates UTF-8 byte lengths into an incorrect prefix: it counts UTF-16 units and rejects over-limit names before writing. These shared TDS paths already had application-supplied callers outside ODBC; the fixes protect the newly exposed ODBC path without claiming the earlier callers were constant-only.

### Testing

Verified locally on the changes pushed in `17240a9b`:

- Formatting and Clippy scripts pass for the workspace and excluded `mssql-py-core` crate.
- **3,795 ODBC/TDS library tests pass** with `cargo nextest run --frozen -p mssqlodbc -p mssql-tds --lib`; temporary TLS fixtures were removed afterward.
- **189 Rust-driver E2E tests pass** across API contracts, DML outputs, call routing, escape sequences, output parameters, execute, fetch-scroll, and more-results suites.
- **61 focused reference E2E tests pass on each platform:** Windows retail `SQL_DRIVER_VER=18.06.0001` and CI-pinned Linux package `18.6.2.1-1`. These are focused parity runs, not claims that the entire CI matrix has passed.
- The Linux output-warning failures were reproduced with the pinned driver. Tests now drain the output tail while the final rowset is nonempty (`SQL_ATTR_ROW_ARRAY_SIZE=2`), so unixODBC exposes the warning rather than hiding diagnostics attached to `SQL_NO_DATA`. Strict return-code/SQLSTATE assertions remain; no comparison skips were added.
- Reordered named-call tests retain pure canonical calls and set matching IPD `SQL_DESC_NAME` values on both drivers. Without descriptor names, both measured retail builds discard inline names on their canonical RPC path; Rust intentionally preserves the SQL names via text routing.
- Fuzz invariants cover lexical marker preservation and NOSCAN identity. The nightly fuzz-specific compile check passed earlier; runtime fuzz iterations did not run because the Windows fuzz executable failed to link.

Fresh CI validation is required for `17240a9b`. Earlier build 174657 had Linux parity failures and retry-related duplicate coverage artifact failures; the local results above do not imply those earlier runs were green.

### Measurement baseline

The original escape golden tables were verified against pinned msodbcsql 18.6.2.1 (`SQL_DRIVER_VER=18.06.0002`) and SQL Server 2025 (`17.00.4006`). Review regressions were additionally measured against Windows retail 18.06.0001 and pinned Linux 18.6.2.1 on SQL Server 2022. The design record distinguishes these baselines and cites reference-source behavior.

### Breaking changes

No exported ODBC signatures change. Observable changes include acceptance of single-row output/return bindings, correct direct-execution marker counts, strict invalid NativeSql length handling, deferred output availability through DML counts, preserved inline named arguments, and rejection of over-limit RPC names instead of malformed wire output. `sp_executesql` now declares by-reference parameters with `OUTPUT`. Streamed output, output arrays, and TVPs remain unsupported.

## Related Issues

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/46384

https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/48049

Remaining output arrays and TVPs: https://sqlclientdrivers.visualstudio.com/mssql-rs/_workitems/edit/48148

## Checklist

- [x] `cargo bfmt` passes (broader `scripts/bfmt.ps1` run)
- [x] `cargo bclippy` passes (broader `scripts/bclippy.ps1` run)
- [ ] `cargo btest` full workspace/coverage run passes on the latest commit in CI; local targeted nextest result: 3,795/3,795
- [x] New/changed functionality has unit and E2E coverage, plus focused Windows/Linux reference comparisons
- [x] Public behavior and remaining gaps are documented (`mssql-odbc/docs/odbc-escape-sequences-plan.md`, `mssql-odbc/docs/parameters_plan.md`)


mssql-odbc/docs/odbc-escape-sequences-plan.md
```diff
@@ -0,0 +1,459 @@
...
+
+## 5. Design
+
+### 5.1 One lexer, not two
+
+New module `mssql-odbc/src/sql/escape.rs` (or `api/escape.rs` to keep the flat `api` layout):
+
+```rust
+pub(crate) struct ScanOutcome {
+    pub sql: String,            // translated text, `?` already rewritten to @P1..@Pn
```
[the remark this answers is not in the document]
David-Engel: Good catch — this was a real design bug, not a wording problem. Fixed in `0548790c`.

§5.1 now splits the work into two phases over one shared lexer: `translate_escapes` handles escapes only and leaves every `?` exactly as written, and the existing `rewrite_param_markers` stays a separate, execution-only phase. There is a caller table spelling out which phases each entry point runs — `SQLNativeSql` runs phase 1 only, `SQLExecDirectW` / `SQLPrepareW` run phase 1 (unless `SQL_ATTR_NOSCAN` is on) and then phase 2. Both phases share one span-classifying lexer so they cannot disagree about what a comment or literal is.

§8 adds the corresponding unit test: `translate_escapes` must leave every `?` untouched, and phase 2 alone must still match today's `rewrite_param_markers` output.

mssql-odbc/docs/odbc-escape-sequences-plan.md
```diff
@@ -0,0 +1,459 @@
...
+| `{[?=]call proc(…)}` | RPC, or textual `EXEC`, §5.5 |
+| anything else | `42000`, statement not sent |
+
+Whitespace: match msodbcsql — a translated escape is emitted as `" " + text + " "`.
+
+### 5.2 `SQL_ATTR_NOSCAN`
+
+Promote out of `INERT_STMT_ATTRS` (`mssql-odbc/src/handles/stmt.rs:588`) into a real statement field.
+Default stays `SQL_NOSCAN_OFF` (scanning on). `SQLNativeSql` and the `SQLDescribeParam` path ignore
+it, matching `DoSubstitutions`' `lpstmt == NULL` behaviour.
```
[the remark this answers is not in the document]
David-Engel: Agreed, and it falls out of the same root cause. Fixed in `0548790c`.

New §5.2 requires the statement to retain the original SQL text. `PreparedPlan` (`mssql-odbc/src/handles/stmt.rs:1015`) keeps only the marker-rewritten text today, which is lossy exactly as you describe: with `SQL_ATTR_NOSCAN` on, `{? = call proc(?)}` would be stored as `{@P1 = call proc(@P2)}` and the canonical return-status marker is unrecoverable by any later scan.

§5.8 now states that `SQLDescribeParam` translates the retained original text, always, ignoring `SQL_ATTR_NOSCAN` — which matches `AutoFillIPD`'s `lpstmt == NULL` behaviour in msodbcsql (`sqlcdesc.cpp:9355`). Retaining the original is folded into commit 4 of the §6 sequence, alongside making the attribute functional.

mssql-odbc/docs/odbc-escape-sequences-plan.md
```diff
@@ -0,0 +1,459 @@
...
+### 5.5 `{call}` routing
+
+Two paths, mirroring msodbcsql:
+
+- **RPC path** — the trimmed statement is exactly one `{[?=]call name[(args…)]}` and every argument
+  is `?`, `DEFAULT`, or empty. Dispatch through `execute_stored_procedure(name, …)` with parameters
+  from an extended `build_named_params`.
+- **Text path** — everything else (a call inside a batch, literal or nested-escape arguments,
+  multiple calls): rewrite to `EXEC name @P1,DEFAULT,…` and run through the existing `sp_executesql`
+  route, which already supports OUTPUT parameters.
```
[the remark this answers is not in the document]
David-Engel: Confirmed in source, and this corrected a wrong claim elsewhere in the doc. Fixed in `0548790c`.

`build_parameter_list_string_impl` (`mssql-tds/src/message/parameters/rpc_parameters.rs:678`) emits `"{param_name} {param_type_name} "` and never appends `OUTPUT`, even when the parameter carries `StatusFlags::BY_REF_VALUE` — despite `RpcParameter::is_output()` existing at `:529`. My earlier §4.1 claimed output parameters via `sp_executesql` were supported; they are not. Only the RPC path works today, where `BY_REF_VALUE` on the wire is sufficient.

Changes made:

- §4.1 now lists this as one of three mssql-tds gaps this PR closes, with the offending line quoted.
- §5.5 shows both places `OUTPUT` has to appear — the `@params` declaration and the call-site argument — and notes that omitting either silently drops the value.
- §5.7 item 2 makes the mssql-tds fix a prerequisite for the text path rather than a nice-to-have, and ties it to AB#41343.
- §6 gives it its own commit (7), with a test that round-trips an output value through `sp_executesql`.
- §8 requires e2e coverage of an output parameter through the text fallback and the RPC path separately.

mssql-odbc/docs/odbc-escape-sequences-plan.md
```diff
@@ -0,0 +1,459 @@
...
+| `{d '…'}`, `{t '…'}`, `{ts '…'}` | yes |
+| `{oj a LEFT OUTER JOIN b ON …}` | yes |
+| `… LIKE 'a\_b' {escape '\'}` | yes |
+| `{guid '…'}` | yes |
+| `{interval '1' DAY}` | **no** — rejected |
+
+### 2.2 Measured against msodbcsql 18
+
+`SQLNativeSqlW` through the unixODBC driver manager, driver `ODBC Driver 18 for SQL Server`
+**[measured]** — this is the parity contract the new code must reproduce:
```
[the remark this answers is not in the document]
David-Engel: Fair, and it mattered more than it first looked. Fixed in `0548790c`.

My box had msodbcsql 18.6.1.1 while CI pins 18.6.2.1 (`msodbcsqlVersion` in `.pipeline/validation-pipeline.yml`, installed by `.pipeline/scripts/install-msodbcsql.sh`), so the golden tables were measured against a different build from the one the parity leg compares against. I re-ran both probe programs against the pinned build in a container: the `SQLNativeSql`, `SQLNumParams` and `SQLDescribeParam` output is byte-for-byte identical, so the tables hold.

The doc now opens with a Measurement baseline section naming the driver (18.6.2.1, `SQL_DRIVER_VER` `18.06.0002`) and the server (SQL Server 2025, `17.00.4006`), tags the tables `[measured, 18.6.2.1]`, and states that anything added later must be re-measured on the pinned version rather than on whatever the developer box happens to have.

The same run also produced the escape-related `SQLGetInfo` values that #535 left at zero and pointed at this work item (`odbc_types.rs:286-291`), so §7 is now a measured table rather than a to-do.

saurabh500 (commented):
I found correctness issues in escape translation, canonical-call routing, and output-parameter delivery that should be addressed before merging. Details and suggested fixes are inline.

The parser findings were checked with an isolated harness importing the unchanged source. The output-buffer and execution-route findings are based on tracing the affected code and reference-driver source, not fresh live-server or retail-driver parity runs. I checked the existing review discussion; the missing call-site OUTPUT finding identifies an implementation gap in a requirement already acknowledged there.

mssql-odbc/src/api/escape.rs
```diff
@@ -0,0 +1,1784 @@
...
+    let mut at = 0usize;
+    while let Some((open, close)) = find_escape(body, at) {
+        let inner = &body[open + 1..close - 1];
+        let (kind, _) = classify(inner)?;
+        if inside_call && kind == EscapeKind::Fn {
+            return Err(EscapeError::syntax(
+                "The {fn ...} escape is not allowed inside a canonical procedure call",
+            ));
+        }
+        validate_nested(inner, inside_call)?;
```
saurabh500: **Blocking: bound escape nesting before recursive descent.** Each nested escape adds another `validate_nested` stack frame, with no depth check. Deeply nested application SQL can overflow the stack before execution reaches SQL Server, terminating the process hosting the driver rather than returning an ODBC error. The isolated parser check reproduced the stack overflow.

Please replace this recursion with an iterative traversal, or enforce a conservative nesting limit before descending and return an `EscapeError`. A subprocess regression test should verify that excessive nesting produces a controlled result instead of terminating the host.

mssql-odbc/src/api/output_params.rs
```diff
@@ -0,0 +1,240 @@
...
+    stmt_state: &mut StmtState,
+    return_values: &[ReturnValue],
+    return_status: Option<i32>,
+) {
+    let returns_status = stmt_state.call_returns_status;
+    let bound: Vec<(usize, BoundParam)> = stmt_state
+        .bound_params
+        .iter()
+        .enumerate()
+        .filter_map(|(i, p)| p.map(|p| (i, p)))
```
saurabh500: **Blocking: resolve the current output bindings, not the execute-time snapshot.** `StmtState::bound_params` is populated when execution starts. `SQLFreeStmt(SQL_RESET_PARAMS)` clears the APD/IPD records without clearing this snapshot, and rebinding updates the descriptors rather than these saved pointers. If bindings change while procedure results remain pending, this writeback can target an old buffer, including one the application has released after unbinding.

Please obtain the current effective descriptor bindings before delivery, respecting the existing descriptor/statement lock ordering. Resetting parameters must prevent later writes through the former bindings, and rebinding must redirect delivery to the new buffers. The reference driver's `GetReturnValue` reads the current APD and binding rather than an execution snapshot.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -168,7 +181,15 @@ pub(crate) unsafe fn bound_param_to_rpc(
     param: &BoundParam,
 ) -> Result<(RpcParameter, ConvOk), ParamBuildError> {
     let ((value, type_metadata), outcome) = unsafe { bound_param_to_value_with_outcome(param) }?;
-    let parameter = RpcParameter::new(name.into(), StatusFlags::NONE, value);
+    // BY_REF_VALUE is what makes the server send a RETURNVALUE token back for
+    // this parameter; without it an OUTPUT binding would be sent as a plain
+    // input and silently produce nothing.
+    let status = if is_output_direction(param.input_output_type) {
```
saurabh500, line 205: **Blocking: output-only parameters must not go through input-buffer conversion.** `bound_param_to_value_with_outcome(param)` has already read and converted the application buffer before this new direction check. For `SQL_PARAM_OUTPUT`, that storage need not contain an initialized input value. In particular, a character buffer with no length indicator takes the `SQL_NTS` path and can be scanned beyond its declared capacity; existing indicator contents can also trigger conversion failures or data-at-execution handling.

Please construct a typed NULL for output-only bindings before reading their value/length buffers or checking for data-at-execution. Keep input conversion for input and input/output bindings. I checked the apparent reference-driver exception: although canonical-call processing promotes an internal status flag, `sqlcmisc.cpp` still checks the original output-only `fParamType` when deciding to send NULL without reading application data.

mssql-odbc/src/api/more_results.rs
```diff
@@ -308,9 +310,16 @@ fn sql_more_results_safe(statement_handle: SqlHandle, stmt: &StmtHandle) -> SqlR
...
             let info_messages = client.take_info_messages();
             post_tds_info_messages(&mut stmt_state, &info_messages);
+            // ODBC makes output parameters and the return status readable only
+            // once every result set the procedure produced has been consumed,
+            // which is exactly here. Writing back earlier would let an
+            // application read a value the spec says is not available yet.
+            let return_values = client.get_return_values();
+            let return_status = client.get_return_status();
+            unsafe {
+                write_back_output_params(&mut stmt_state, &return_values, return_status);
```
saurabh500: **Blocking: the fetch-exhausted path bypasses this output delivery.** `release_busy_if_row_exhausted` calls `complete_current_result`, which can drain the RPC's RETURNVALUE/RETURNSTATUS tail, release the connection, and set `stmt_state.batch_exhausted`. The next `SQLMoreResults` then takes the early `batch_exhausted` branch and never reaches this `StatementResult::End` arm. A procedure returning rows plus output parameters therefore loses its outputs when the application fetches the rows normally. Another statement can also reuse the released client and clear its saved return tokens.

Please capture pending return values/status on the statement before releasing the client and deliver them from the exhausted fast path as well as this path. Cover a row-returning procedure whose rows are fetched to exhaustion before `SQLMoreResults`, including connection reuse by another statement.

mssql-odbc/src/api/output_params.rs
```diff
@@ -0,0 +1,240 @@
...
+unsafe fn write_value(param: &BoundParam, value: &ColumnValues) -> Result<bool, RowIssue> {
+    let binding = ColumnBinding {
+        column_number: 1,
+        target_type: param.c_type,
+        target_value_ptr: param.parameter_value_ptr,
+        buffer_length: param.buffer_length,
+        strlen_or_ind_ptr: param.strlen_or_ind_ptr,
+        octet_length_ptr: param.octet_length_ptr,
+    };
+    unsafe { deliver_bound_value(&binding, value) }
```
saurabh500, line 147: **Blocking: apply the parameter bind offset to output writes.** Input staging displaces the application pointers using `SQL_ATTR_PARAM_BIND_OFFSET_PTR`, but `bound_params` retains the base pointers. This adaptation copies those undisplaced pointers, and `deliver_bound_value` passes a zero offset to the column writer. With a nonzero parameter bind offset, input is read from the selected record while output values and indicators overwrite the base record.

Please use the existing parameter-positioning helper to apply the effective parameter bind offset to the value, indicator, and octet-length pointers before delivery. Test that only the displaced record changes and the base record remains untouched.

mssql-odbc/src/api/exec_common.rs
```diff
@@ -701,6 +701,52 @@ pub(super) fn snapshot_bound_params(
...
+        stmt_state.bound_params.iter().skip(skip).copied().collect();
+    match unsafe {
+        build_named_params_for_row(
+            &bound,
+            marker_count.saturating_sub(skip),
+            bind_offset,
+            crate::api::odbc_types::SQL_BIND_BY_COLUMN,
+            0,
+            false,
+        )
```
saurabh500, line 741: **Blocking: preserve DEFAULT argument positions in canonical RPCs.** `CallSite::is_rpc_eligible` accepts `CallArg::Default`, but this builder sees only `marker_count` and the bound markers, not the parsed argument sequence. Consequently `{call p(?,DEFAULT,?)}` sends two positional arguments instead of three argument slots, so the second bound value goes to the wrong procedure parameter and the trailing parameter takes its default.

Please build the RPC vector from `CallSite.args`, inserting default-valued RPC slots where required. A smaller interim fix is to exclude calls containing default arguments from RPC eligibility so the existing text form preserves their positions.

mssql-odbc/src/api/exec_direct.rs
```diff
@@ -174,11 +185,27 @@ fn sql_exec_direct_w_safe(
...
-                Err(rc) => return rc,
-            };
+        // A canonical call binds its parameters by position; everything else
+        // binds them by the `@P1..@Pn` names the rewritten text declares.
+        let rpc_call = call.as_ref().filter(|c| c.is_rpc_eligible());
+        let named_params = match rpc_call {
+            Some(c) => {
+                let skip = usize::from(c.returns_status);
+                match unsafe {
+                    build_positional_params(&mut stmt_state, marker_count, skip, "SQLExecDirectW")
```
saurabh500, line 224: **Blocking: select the data-at-execution route before constructing positional parameters.** This creates unnamed parameters for an RPC-eligible canonical call, but a data-at-execution parameter later sends the operation through `begin_sp_executesql` or the deferred text path. The rewritten SQL references `@P1`, etc., while the TDS declaration builder skips parameters whose name is `None`. Return-status calls also have their first binding removed before that route change.

Please decide the actual execution route before parameter construction. If streaming uses the text route, retain the complete named parameter list, including any return variable; otherwise add streaming support to the stored-procedure RPC path. Cover canonical calls with data-at-execution input both with and without a return-status binding.

mssql-odbc/src/api/exec_common.rs
```diff
@@ -1052,6 +1098,18 @@ pub(super) fn finish_execute(
...
+        // drain above, so this is where its output parameters and return status
+        // become available -- SQLMoreResults never runs for it.
+        let return_values = client.get_return_values();
+        let return_status = client.get_return_status();
+        unsafe {
+            crate::api::output_params::write_back_output_params(
+                &mut stmt_state,
+                &return_values,
+                return_status,
+            );
```
saurabh500: **Blocking: propagate the output-delivery outcome to the API return code.** `write_back_output_params` returns `()` even when it posts conversion errors or truncation diagnostics. This caller still chooses success based only on server INFO messages, and the `SQLMoreResults` caller likewise returns `SQL_NO_DATA` without considering writeback failure. For example, an output value that does not fit the bound integer type records an error while execution reports success.

Please return an aggregate delivery outcome and propagate errors/warnings through both callers. Preserve the original warning kind too: `deliver_bound_value` currently reduces fractional and string truncation to the same boolean, so fractional truncation is subsequently reported as string truncation.

mssql-odbc/src/api/exec_direct.rs
```diff
@@ -192,11 +219,13 @@ fn sql_exec_direct_w_safe(
         stmt_state.prepared = None;
         stmt_state.parameter_metadata.clear();
         stmt_state.clear_state(STMT_STATE_PREPARED);
+        stmt_state.call_returns_status = call.as_ref().is_some_and(|c| c.returns_status);
```
saurabh500, line 252: **Blocking: record the return-value source for each execution route, and reset it on every execution.** This flag is true for any recovered `{?=call ...}`, including a call sent through the text fallback because it contains a literal argument. There, the procedure's return value belongs to the generated output variable; the raw RETURNSTATUS is for the enclosing `sp_executesql`. `write_back_output_params` selects the latter and skips the output variable. Separately, this field is not initialized by `SQLPrepare`/`SQLExecute`, so a previous direct return-status call can make a later prepared execution misclassify its first output binding.

Please distinguish direct-RPC RETURNSTATUS from text-route output-variable delivery, and initialize that mapping for every execute. Cover both a literal-containing return-status call and reuse of the same statement for a subsequent prepared output operation.

mssql-odbc/src/api/escape.rs
```diff
@@ -0,0 +1,1784 @@
...
+
+        let body = &sql[open + 1..close - 1];
+        let (kind, remainder) = classify(body)?;
+        escape_count += 1;
+
+        match kind {
+            EscapeKind::Fn | EscapeKind::PassThrough => {
+                validate_nested(body, false)?;
+                // Native to SQL Server: emit the escape exactly as written.
+                out.push_str(&sql[open..close]);
```
saurabh500: **Blocking: translate nested escapes before preserving the native outer escape.** This branch validates nested tags but then copies the entire original outer escape unchanged. For example, `SELECT {fn CONCAT({interval '1' DAY},'x')}` still contains the untranslated interval escape after phase 1. SQL Server cannot handle that inner escape natively, even though the outer function escape is supported.

Please translate inner escapes before reconstructing a passed-through outer escape; validation alone is insufficient. The isolated parser regression assertion fails on this input.

mssql-odbc/src/api/escape.rs
```diff
@@ -0,0 +1,1784 @@
...
+        text.push_str("?=");
+    }
+    text.push_str(&call.proc_name);
+    text.push(' ');
+    for (i, arg) in call.args.iter().enumerate() {
+        if i > 0 {
+            text.push(',');
+        }
+        match arg {
+            CallArg::Marker { text: t, .. } | CallArg::Text(t) => text.push_str(t),
```
saurabh500: **Blocking: the text execution form still needs call-site OUTPUT annotations.** For an output-bound marker, `{call p(1,?)}` becomes `EXEC p 1,@P1`, not `EXEC p 1,@P1 OUTPUT`. The PR now adds `OUTPUT` to the enclosing RPC parameter declaration, but that does not propagate the value out of the inner procedure call. Prepared canonical calls also use the rendered text.

This was acknowledged in the earlier design-review thread, but the call-site half remains missing from the implementation. Please add a binding-aware execution rendering step that emits `OUTPUT` for output/input-output arguments while keeping `SQLNativeSql` translation binding-independent. Add a round-trip test through the text fallback, not only the direct canonical RPC path.

mssql-odbc/src/fuzz_support.rs
```diff
@@ -58,6 +60,39 @@ pub fn fuzz_connection_string(input: &str) {
...
+
+    if let Ok(translated) = translate_escapes(input) {
+        // Phase 1 must leave every `?` exactly where it was: SQLNativeSql
+        // returns markers untranslated.
+        assert_eq!(
+            translated.sql.matches('?').count(),
+            markers,
+            "escape translation changed the marker count: {input:?} -> {:?}",
+            translated.sql
+        );
```
saurabh500: **Suggestion: count parameter markers lexically rather than counting every question mark.** `{encrypt N'?'}` legitimately transforms its quoted literal into hexadecimal, so the raw question-mark count changes even though no parameter marker was lost. The current assertion reports a fuzz failure for this valid translation; the isolated check reproduced it.

Please compare executable marker counts from the lexer/marker-rewriting phase before and after translation. Keep this quoted-question-mark case as a regression test for the invariant itself.

mssql-odbc/src/api/escape.rs
```diff
@@ -0,0 +1,1784 @@
...
+/// a procedure name goes, and that name reaches `execute_stored_procedure` and,
+/// on Always Encrypted connections, a T-SQL `EXEC` string. Only a multi-part
+/// identifier is accepted — each part either a regular identifier or a
+/// `[bracketed]` / `"quoted"` one — plus msodbcsql's optional `;n` procedure
+/// group number (`sqlcmisc.cpp:8215`). Anything else is a syntax error, so no
+/// separator, comment or statement terminator can ride along.
+pub(crate) fn validate_procedure_name(raw: &str) -> Result<String, EscapeError> {
+    let invalid = || EscapeError::syntax(format!("Invalid procedure name '{raw}' in {{call ...}}"));
+
+    let (name, group) = match raw.split_once(';') {
```
saurabh500: **Suggestion: recognize procedure-group semicolons only outside identifier delimiters.** `split_once(';')` treats a semicolon inside a quoted identifier as a group separator, so a valid name such as `[p;q]` is rejected before execution. The new `quote_procedure_name` helper in `tds_client.rs` makes the same split.

Please identify the end of the multipart identifier before parsing an optional `;n` suffix, and apply the same handling in both places. The existing `mssql-tds::sql_identifier` parsing/escaping helpers are worth reusing for the identifier portion. The quoted-semicolon parser check currently fails.

saurabh500:
Separately from the correctness fixes, I suggest opening a tracking item to unify the SQL-lexing machinery used by `mssql-odbc` and `mssql-py-core`.

ODBC's `CodeScan` (`mssql-odbc/src/api/escape.rs`) and Python's `rewrite_placeholders` (`mssql-py-core/src/async_parameters.rs`) independently recognize strings, quoted/bracketed identifiers, comments, and executable parameter markers. This duplication predates this PR; the PR usefully shares the scanner within ODBC, but not across the two bindings. Both already share TDS execution through `mssql-tds`.

The tracking item should cover a small shared span-based lexer, potentially in `mssql-tds`, with explicit behavior policies. It must preserve the intentional differences: Python supports nested block comments, `%(name)s` placeholders, and collision-free generated names; ODBC preserves msodbcsql's non-nesting comment behavior and canonical-extension quirks. Keep placeholder naming/binding, ODBC escape translation, and Python/ODBC value conversion in their respective layers. Reuse both test suites and add explicit cases where the two policies are expected to disagree.

I would handle that extraction as a separate behavior-preserving refactor after the correctness fixes, rather than expanding this PR. The smaller reuse opportunity for the existing multipart-identifier helpers in `mssql-tds/src/sql_identifier.rs` can be considered here.

mssql-odbc/src/api/output_params.rs
David-Engel: Delivery now snapshots the current effective APD/IPD bindings with the existing lock ordering rather than using execute-time pointers. Live regressions verify that resetting parameters prevents pending writes and rebinding redirects them to the new buffers (4a6ac0f5).

mssql-odbc/src/api/exec_common.rs
David-Engel: Took the smaller safe fix: DEFAULT and omitted arguments are no longer eligible for direct RPC, so their positions are preserved in the text form. Live coverage checks both `{call #route(?,DEFAULT,?)}` and `{call #route(?,,?)}` (4a6ac0f5).

mssql-odbc/src/api/exec_common.rs
David-Engel: Output delivery now returns an aggregate outcome that reaches execution and SQLMoreResults return codes, including the exhausted fast path. Regression coverage checks conversion errors and distinguishes string truncation (01004) from fractional truncation (01S07), with delivery reported only once (4a6ac0f5).

mssql-odbc/src/api/escape.rs
David-Engel: Native outer escapes now reconstruct their contents after translating nested escapes, rather than copying untranslated inner text. Added parser regressions for nested translations, including the interval-inside-CONCAT case (4a6ac0f5).

mssql-odbc/src/api/escape.rs
David-Engel: Added binding-aware execution rendering using global lexical marker ordinals; output/input-output call arguments receive OUTPUT while return assignments and SQLNativeSql remain unaffected. Direct text fallback, embedded calls, prepared calls, and direction changes on a cached prepared statement now have live round-trip coverage (4a6ac0f5).

mssql-odbc/src/api/output_params.rs
David-Engel: Output delivery now uses the parameter-positioning helper with the current bind offset for the value, indicator, and octet-length pointers. A live descriptor test verifies that all displaced destinations are updated while the base record remains untouched (4a6ac0f5).

mssql-odbc/src/conversion/param_convert.rs
David-Engel: Output-only bindings now produce typed NULL before any value/indicator read, and DAE detection skips their destination indicators. Added unit coverage for uninitialized buffers and retained type metadata, plus a live test with nonterminated storage, a stale SQL_DATA_AT_EXEC indicator, and no indicator pointer (4a6ac0f5).

mssql-odbc/src/api/escape.rs
David-Engel: Added a 64-level nesting preflight before recursive translation; deeper input returns 42000. A subprocess regression exercises 10,000 nested escapes and verifies controlled rejection instead of host termination (4a6ac0f5).

mssql-odbc/src/api/more_results.rs
David-Engel: Return values/status are now retained on the statement before fetch exhaustion releases the client, and SQLMoreResults delivers them once from its exhausted fast path. The live regression fetches to exhaustion, reuses the connection through another statement, and then checks the original outputs (4a6ac0f5).

mssql-odbc/src/api/exec_direct.rs
David-Engel: DAE is detected from positioned input bindings before route selection and parameter construction, so both streaming paths retain the complete named list and return-assignment variable. Live regressions cover streamed VARCHAR and buffered INTEGER input, each with and without a return binding (4a6ac0f5).

mssql-odbc/src/api/escape.rs
David-Engel: Both procedure-name paths now recognize group suffixes outside identifier delimiters and reuse the existing SQL identifier parsing/quoting helpers. Tests cover quoted semicolons such as `[p;q]`, valid group suffixes, and malformed names (4a6ac0f5).

mssql-odbc/src/api/exec_direct.rs
David-Engel: The return-status flag now identifies only the selected direct RPC route and is reset for prepare/prepared execution; text routes use the named output variable instead of wrapper RETURNSTATUS. Live tests cover literal-containing return calls and reuse of a direct-return statement for a prepared output call (4a6ac0f5).

mssql-odbc/src/fuzz_support.rs
David-Engel: The fuzz invariant now compares executable marker counts from marker rewriting, not raw question marks. Added a regression for `{encrypt N'?'}` so transformation of quoted text does not register as a lost parameter (4a6ac0f5).

mssql-odbc/src/api/escape.rs
```diff
@@ -0,0 +1,1962 @@
...
+}
+
+impl CallSite {
+    /// True when every argument can be carried as an RPC parameter, i.e. the
+    /// call needs no `EXEC` text at all.
+    /// DEFAULT stays textual so omission cannot shift later positional bindings.
+    pub(crate) fn is_rpc_eligible(&self) -> bool {
+        self.args
+            .iter()
+            .all(|a| matches!(a, CallArg::Marker { .. }))
```
saurabh500: **Preserve named arguments by keeping them on the text route.** This accepts `CallArg::Marker { name: Some(_), .. }`, but `SQLExecDirect` subsequently calls `build_positional_params`, which sends unnamed parameters and discards the parsed argument names. For `{call p(@b=?,@a=?)}` against a procedure declared as `p(@a int, @b int)`, the first bound value therefore goes to `@a` instead of the requested `@b`. The prepared/text path preserves the intended mapping.

Restrict positional RPC eligibility to unnamed markers. Named arguments will then use the existing text renderer, which already preserves their names and adds `OUTPUT` where needed:

```suggestion
            .all(|a| matches!(a, CallArg::Marker { name: None, .. }))
```

Please add direct/prepared regression tests with reordered named arguments, including an output binding. The focused parser probe reproduces the incorrect eligibility decision; the downstream positional dispatch was verified in source, not through a fresh live-server run.

Theekshna (commented):

First pass from this automation on this PR (no prior marker). This is a large change (32 files, ~5.3k lines: escape-sequence lexer/translator, `{call}` RPC routing, output parameters, `SQLNativeSql`/`SQLNumParams`). `saurabh500` already ran a deep review (2026-09-11 21:14) that filed 10 Blocking findings; all 10 were fixed in `4a6ac0f5` with matching replies and new regression coverage, which I spot-checked rather than took on faith (e.g. the `escape.rs:412` recursion-depth fix, the `output_params.rs:46` snapshot-vs-live-binding fix, and the `more_results.rs:320` fetch-exhaustion output-loss fix all read as claimed against the current diff).

**One Blocking thread from that review remains open and unanswered as of this head** -- `saurabh500`'s `escape.rs:127` comment (filed 2026-09-12T07:52, the same timestamp as this PR's last push) that `is_rpc_eligible` still accepts named markers (`CallArg::Marker { name: Some(_), .. }`), while `build_positional_params` (`exec_common.rs:723`) calls `build_named_params_for_row(..., named: false)`, discarding the parsed name. I independently re-verified this in the current source: `escape.rs:127` is unchanged since that comment, and `build_named_params_for_row` (`exec_common.rs:809-831`) sets `name = named.then(|| parameter_name(i))`, which is `None` under the positional path -- so `{call p(@b=?,@a=?)}` still routes by position, not name. Not re-filing since it's already an open, correctly-targeted thread; noting it here so it isn't missed as a merge blocker.

**New finding from this pass:**

### Verify Before You Claim / evidence audit
Re-derived one of the PR's own parity claims against the reference tree rather than accepting it: the premise that SQL Server parses `{fn}`/`{d}`/`{t}`/`{ts}`/`{oj}`/`{escape}`/`{guid}` natively but not `{interval}` is consistent with msodbcsql having dedicated `ECODE_INTERVAL` translation code (`sqlcmisc.cpp:4711-4719`, `:7883-7920`) rather than passing that escape through untouched -- the one escape family the design doc says needs driver-side translation is the one the reference driver also special-cases. Did not re-derive the full golden tables (out of scope for the time available); flagging only that the specific claim checked holds.

| Severity | Count |
|---|---|
| Blocking | 1 |
| Suggestion | 0 |
| Nit | 0 |

**[Blocking]** `gh pr checks` at this head (`a61e843d`) shows `mssql-rs Pull request validation (Build Stage Build Linux)`, `(Build Stage Build Windows)`, `(Build Stage Test MacOS)`, and the downstream `coverage-report` (which times out waiting on that pipeline's artifacts) all failing -- confirmed against this exact SHA via `commits/{sha}/check-runs`, not a stale/superseded run. This directly contradicts the PR checklist's `cargo bfmt`/`cargo bclippy`/`cargo btest` and "1504 unit tests / 43/43 e2e pass" claims, which predate this state. `cargo check -p mssqlodbc --all-targets` compiles clean locally, so the Rust workspace itself isn't the obvious culprit -- the last commit (`a61e843d`) only touches `output_params_test.cpp`/`README.md` (e2e C++), which is a more likely place to look. I don't have non-interactive access to the Azure Pipelines build logs to confirm root cause (fail-open per this run's ADO restrictions), so this needs a human to look at the actual build output before merging.

mssql-odbc/src/api/escape.rs
Theekshna: Confirmed still live at HEAD (`a61e843d`). Traced the call sites: `is_rpc_eligible()` is only consumed in `exec_direct.rs` (`build_positional_params(..., named=false)`), so this is a `SQLExecDirect`-only bug — `SQLExecute`'s prepared path always uses `build_named_params` and is unaffected. But for `SQLExecDirect`, `{call p(@b=?,@a=?)}` still sends values positionally, silently swapping `@a`/`@b`. Please land the suggested fix (`name: None` in the eligibility check) plus a regression test with reordered named arguments before merge.

Theekshna:
Re-checked CI at current HEAD (`a61e843d`): `mssql-rs Pull request validation` is still failing on Build Linux, Build Windows, and Test MacOS, and the downstream `coverage-report` times out waiting on it. This contradicts the PR description''s test-pass claims. Please investigate/fix before this goes further — the last commit only touched `output_params_test.cpp`/`README.md`, which is the more likely place to look first.

mssql-odbc/docs/odbc-escape-sequences-plan.md
```diff
@@ -0,0 +1,582 @@
...
+Omitting either one silently drops the output value, which is why §5.7 item 2 is a prerequisite for
+this path rather than a nice-to-have.
+
+**Procedure-name validation is mandatory before either path.** Accept only
+`[db.][schema.]name` where each part is a regular identifier or a `[bracketed]` / `"quoted"`
+identifier, plus an optional `;N` group number; reject everything else with `42000`. Never
+concatenate the name into T-SQL without bracket-quoting it. This is the injection-relevant surface
+of the change and must be reviewed as such.
+
+### 5.6 Output parameters, return values, availability ordering
```
Theekshna, line 442: `parameters_plan.md` still lists output/input-output parameters as a "driver-wide gap" and under "Deferred features" (Divergences table row 5). That's now stale for this PR's single-row case — please update it:

**Supported by this PR:**
- Single-row `SQL_PARAM_OUTPUT` / `SQL_PARAM_INPUT_OUTPUT` / return-status, delivered via RPC or `EXEC ... OUTPUT` text, matched by name then ordinal.
- Correct delivery across fetch-exhaustion, connection reuse, bind-offset, and rebind/`SQL_RESET_PARAMS`.
- Truncation/conversion diagnostics on writeback (`01004`, `01S07`, `22018`/`22002`).

**Not supported (tracked in AB#48148):**
- Output/input-output params in a parameter array (`SQL_ATTR_PARAMSET_SIZE > 1`) — row 5 should say "arrays only," not a blanket gap.
- Table-valued parameters (TVP).
- The still-open named-argument `{call}` routing bug (escape.rs:127 thread) that can misdirect an output binding.

Please add AB#48148 alongside the other AB references in the divergences table and "Remaining work."

mssql-odbc/src/api/num_params.rs
```diff
@@ -0,0 +1,97 @@
...
+        "SQLNumParams: handle is not a STMT"
+    );
+
+    let Ok(mut state) = stmt.inner.lock() else {
+        error!("SQLNumParams: stmt mutex poisoned");
+        return SQL_ERROR;
+    };
+    free_errors(&mut state);
+
+    let Some(plan) = state.prepared.as_ref() else {
```
shiwanigupta0809: **Blocking:** `SQLNumParams` derives its result only from `state.prepared`, but `SQLExecDirectW` clears that state. As a result, calling `SQLNumParams` after a valid direct execution always returns `HY010`, even though the direct-execution case is part of AB#48049. Please retain the current direct statement's marker count independently of `PreparedPlan`, consult it here, and cover direct statements both with markers and with zero markers.

mssql-odbc/src/api/native_sql.rs
```diff
@@ -0,0 +1,246 @@
...
+        "SQLNativeSqlW: handle is not a DBC"
+    );
+
+    // The DM rejects a null input pointer before the driver sees it.
+    debug_assert!(
+        !in_statement_text.is_null(),
+        "SQLNativeSqlW: in_statement_text is null — DM should have rejected this"
+    );
+
+    let sql = unsafe { read_utf16_long(in_statement_text, text_length1) };
```
shiwanigupta0809: **Blocking:** Validate `text_length1` before passing it to `read_utf16_long`. That helper intentionally maps negative values other than `SQL_NTS` to an empty string, so an invalid length such as `-2` can be processed successfully instead of returning `SQL_ERROR` with `HY090`. Please acquire and clear the diagnostic state first, reject `text_length1 < 0 && text_length1 != SQL_NTS`, and add a direct API regression test.

shiwanigupta0809:
## Review summary

This PR is not ready to merge yet. I found two new blocking API-contract issues and posted them inline:

- `SQLNumParams` loses the marker count after `SQLExecDirect`, so the supported direct-execution sequence returns `HY010`.
- `SQLNativeSqlW` accepts invalid negative input lengths instead of reporting `HY090`.

The named-call RPC issue is also still blocking: named markers are accepted for the positional RPC route, which can silently swap input or output bindings. That issue is already covered by the existing unresolved thread, so I did not duplicate it.

The targeted `mssqlodbc` library suite passes all 1,584 tests. Current CI is not green: the macOS parity stage has two diagnostic-expectation failures, while the Linux/Windows failures are in coverage-artifact publishing. AB#46384 and AB#48049 are active and align with the intended scope.

mssql-odbc/src/api/exec_direct.rs
```diff
@@ -323,7 +381,21 @@ fn sql_exec_direct_w_safe(
...
-    let exec_result: Result<(), mssql_tds::error::Error> = if marker_count > 0 {
+    let exec_result: Result<(), mssql_tds::error::Error> = if let Some(call) = call.as_ref() {
+        // A statement that is nothing but `{call proc(?)}` goes out as a TDS
+        // RPC rather than as text, which is what makes output parameters and
+        // the return status available. Anything less strict — a call inside a
+        // batch, or with a literal argument — took the EXEC text form during
+        // translation and runs through sp_executesql below.
+        dbc.runtime
+            .block_on(client.execute_stored_procedure(
+                call.proc_name.clone(),
```
Vahid-b, line 393: **Suggestion: the RPC procedure-name length prefix is byte-counted, and this is the first path that feeds it application text.**

This is the same widening argument the PR already makes for the `EXEC` sink in `tds_client.rs` — `{call ...}` turns the procedure name into text lifted out of an application's SQL — but it applies to a second sink that wasn't hardened.

`call.proc_name` reaches `RpcType::Named`, and `SqlRpc::write_proc` (`mssql-tds/src/message/rpc.rs:172-178`) serialises it as:

```rust
packet_writer.write_i16_async((stored_proc_name.len() as u8).into()).await?;
packet_writer.write_string_unicode_async(stored_proc_name.as_str()).await?;
```

`NameLenProcID` is a `US_VARCHAR`, so the prefix has to be the count of UTF-16 code units, but `str::len()` is UTF-8 **bytes** and `write_string_unicode_async` writes `value.encode_utf16()` (`io/packet_writer.rs:598-600`). Two ways that diverges, and `validate_procedure_name` accepts both inputs:

- **Non-ASCII names.** `is_identifier_start`/`is_identifier_part` use `char::is_alphabetic`/`is_alphanumeric`, so `{call café_proc(?)}` validates. `"café_proc".len()` is 10, `encode_utf16().count()` is 9 — the server is told to read one extra UTF-16 unit and consumes two bytes of the following `OptionFlags`/parameter data. Bracketed parts (`{call [proc ✓](?)}`) reach the same place.
- **Names longer than 255 bytes.** `validate_procedure_name` caps the part *count* at 4 but not the length, so `as u8` wraps silently.

Either one is a malformed RPC rather than a clean ODBC diagnostic. Note the encoding defect is pre-existing (`mssql-js` already passes an app-supplied name), so this isn't a regression — but this PR is what makes it reachable from ODBC, and the cheap containment is on this side: reject in `validate_procedure_name` when `raw.encode_utf16().count() > 255`, and either fix `write_proc` to use `encode_utf16().count()` or reject non-ASCII there too. A `{call}` e2e case with a non-ASCII procedure name would pin it.

mssql-odbc/src/api/bind_param.rs
```diff
@@ -253,19 +254,43 @@ fn sql_bind_parameter_safe(
...
+
+        // An output-only parameter has no value to send, so its indicator is
+        // not read at execute time; ODBC still requires a value buffer to write
+        // the result back into.
+        if matches!(input_output_type, SQL_PARAM_OUTPUT | SQL_RETURN_VALUE)
+            && parameter_value_ptr.is_null()
+        {
+            error!("SQLBindParameter: an output parameter needs a value buffer");
+            post_diag(&mut stmt_state, ERR_INVALID_NULL_POINTER);
             return SQL_ERROR;
```
Vahid-b: **Suggestion / question: this rejects a null `ParameterValuePtr` more aggressively than the DM does.**

ODBC's documented `HY009` condition for `SQLBindParameter` fires when `ParameterValuePtr` **and** `StrLen_or_IndPtr` are both null *and* `InputOutputType` is not `SQL_PARAM_OUTPUT` — i.e. the spec deliberately carves out output bindings that supply only an indicator (an application that wants to learn NULL-ness or the available length, and then uses `SQLGetData`-style sizing). Here the value pointer alone being null is fatal regardless of the indicator.

I'm raising this as a question rather than a defect because msodbcsql is the contract in this crate and I don't have its source to cite; the repo's own review guidance is explicit that a parity claim argued from the spec gets overturned about as often as it's upheld. So: what does msodbcsql 18.6.2.1 do for `SQLBindParameter(1, SQL_PARAM_OUTPUT, SQL_C_SLONG, SQL_INTEGER, 0, 0, NULL, 0, &ind)`?

Worth resolving either way, because the layer underneath already tolerates it — `output_only_ignores_length_contents_and_retains_null_metadata` in `param_convert.rs` exercises exactly this shape with a null value pointer — so the restriction lives only at the API edge. If msodbcsql accepts it, `write_value` also needs a null-target check so delivery reports through the indicator instead of writing. If msodbcsql refuses it too, a one-line comment citing the file/line would keep the next reader from re-opening this.

mssql-odbc/src/api/odbc_types.rs
```diff
@@ -379,6 +399,10 @@ pub const SQL_PARAM_TYPE_UNKNOWN: SqlSmallInt = 0;
 pub const SQL_PARAM_INPUT: SqlSmallInt = 1;
 pub const SQL_PARAM_INPUT_OUTPUT: SqlSmallInt = 2;
 pub const SQL_PARAM_OUTPUT: SqlSmallInt = 4;
+/// The `{? = call ...}` return-status parameter. ODBC 3.5+.
+pub const SQL_PARAM_INPUT_OUTPUT_STREAM: SqlSmallInt = 8;
```
Vahid-b, line 402: Nit: this doc comment describes `SQL_RETURN_VALUE` (the `{? = call ...}` return-status direction, value 5) but sits on `SQL_PARAM_INPUT_OUTPUT_STREAM` (value 8, ODBC 3.8 streamed output). Moving it down two lines onto `SQL_RETURN_VALUE` would match — and the two streamed constants are the ones that could use a note, since they exist here only to be refused with `HYC00`.

Vahid-b (commented):
## Automated review (unattended run)

Reviewed the full merge-base diff (`be0ea47b..a61e843d`, 32 files / ~5.3k lines) in a dedicated worktree. Three reviewers have already been through this (`copilot-pull-request-reviewer`, `saurabh500` with 10 Blocking findings now fixed, `Theekshna`, `shiwanigupta0809`), so I adjudicated rather than re-filed: I re-checked each open thread against the current source and only raised things nobody has. **No new Blocking findings.**

### What I verified rather than assumed

- **Local validation of the checklist claims.** In a clean worktree at `a61e843d`: `cargo fmt -- --check` clean; `cargo clippy -p mssqlodbc -p mssql-tds --all-features --all-targets -- -D warnings` clean; `cargo nextest run -p mssqlodbc -p mssql-tds --lib` = 3767 tests, 3760 passed. The 7 failures are all `certificate_validator` / `win_tls::validate` pinned-cert cases that need `tests/test_certificates/*.pem` fixtures — the known clean-tree baseline, unrelated to this PR. So the `bfmt`/`bclippy`/`btest` boxes hold locally.
- **The lexer refactor is behaviour-preserving.** `rewrite_param_markers` was replaced by `CodeScan`; I diffed the two state machines line by line, including the `--(*` / `*)--` canonical-extension carve-out and the `prev1`/`prev2` bookkeeping (both only ever record the *first* char of a step, in old and new). They agree, and `code_scan_reproduces_the_marker_lexer_quirks` pins it.
- **The `EXEC` injection sink is the only one.** Grepped `mssql-tds` for every interpolation of a procedure name: `tds_client.rs:6117` is the single site, and it is now routed through `quote_procedure_name`. `RpcType::Named` at `:3328` is a wire field, not T-SQL.
- **No double delivery of output parameters.** `finish_execute`'s no-result-set branch writes back and then clears `STMT_STATE_CURSOR_OPEN`, so a following `SQLMoreResults` short-circuits on `if !cursor_open` before it can reach the `StatementResult::End` write-back. Traced, not assumed.
- **`described_count = marker_count - return_status` cannot underflow.** `return_status` is only true for `{? = call ...}`, whose return marker is itself counted at prepare time under both `NOSCAN` settings, so `marker_count >= 1` whenever it is set.
- **e2e registration.** All three new test binaries (`output_params_test`, `call_routing_test`, `escape_sequence_test`) are wired into `tests/e2e/CMakeLists.txt`.
- **`SQLGetInfo` type codes.** 38 / 48 / 109 / 110 / 113 / 115 are the correct ODBC codes and none collides with an existing arm. I did not re-derive the measured msodbcsql bitmasks.

### Open threads — still live at this head, not re-filed

I re-read each against the current source; all four are still accurate and none is answered:

- `escape.rs:127` (`saurabh500`, Blocking) — `is_rpc_eligible` still accepts `CallArg::Marker { name: Some(_) }` while `exec_direct.rs` builds positional params, so `{call p(@b=?,@a=?)}` still binds by position on the `SQLExecDirect` route.
- `num_params.rs:52` and `native_sql.rs:68` (`shiwanigupta0809`, Blocking).
- `odbc-escape-sequences-plan.md:434` (`Theekshna`, doc staleness).

### CI

Re-confirmed at `a61e843d`: `coverage-report`, `Build Stage Build Linux`, `Build Stage Build Windows` and `Build Stage Test MacOS` are failing. Already reported by `Theekshna` and characterised by `shiwanigupta0809` (macOS parity diagnostics + coverage-artifact publishing), so I am not re-filing it — noting only that my local Rust-side runs above are green, which is consistent with the failure being outside the workspace build.

### Findings

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 2 |
| Nit | 1 |

Details inline at `exec_direct.rs:392`, `bind_param.rs:294` and `odbc_types.rs:403`.

This review was produced by an unattended run; the findings were not checked by a human before posting.

shiwanigupta0809 (commented):
## Summary

This PR implements ODBC escape translation, canonical-call routing, output parameters, `SQLNativeSqlW`, and `SQLNumParams`. The targeted 176 unit tests passed, but I found one new output-parameter ordering bug and one msodbcsql parity mismatch. The PR is not ready to merge; existing unresolved review threads and required CI checks also remain open.

## Blocking

One new finding: output parameters are exposed before outstanding DML result counts are consumed. See the inline comment on `exec_common.rs`.

## Suggestion

One new parity finding: nested interval escapes inside canonical calls are translated even though msodbcsql passes them through. See the inline comment on `escape.rs`.

## Nit

None.

**Existing review status:** Previously raised findings are not duplicated. Unresolved threads still include the named-argument positional-RPC bug, `SQLNumParams` after direct execution, invalid negative `SQLNativeSqlW` input lengths, and procedure-name wire-length handling. Required checks remain red at head `a61e843d`: Linux and Windows report coverage-artifact publication failures, macOS reports test failures, and the dependent coverage check failed.

**Linked work items:** `AB#46384` and child `AB#48049` are Active. The implemented scope generally matches them, but the output-delivery timing issue conflicts with `AB#48049`’s ordering requirement.

mssql-odbc/src/api/exec_common.rs
```diff
@@ -1047,11 +1099,37 @@ pub(super) fn finish_execute(
...
+        // A procedure that returns no result set is already exhausted by the
+        // drain above, so this is where its output parameters and return status
+        // become available -- SQLMoreResults never runs for it.
+        let return_values = client.get_return_values();
+        let return_status = client.get_return_status();
+        let output_rc = if let Ok(bound_params) = bound_params {
+            // Fresh descriptor snapshot; execute-time input pointers may have
+            // been reset or rebound while the results were being consumed.
+            unsafe {
+                crate::api::output_params::write_back_output_params(
```
shiwanigupta0809, line 1135: The no-row path writes output parameters back immediately after draining the wire, even when `take_dml_result_counts()` produced update-count results that the application has not consumed. Those counts remain observable through `SQLRowCount` and `SQLMoreResults`, but the output buffer has already changed when `SQLExecDirectW`/`SQLExecute` returns. For a procedure that performs an `UPDATE` with `NOCOUNT` off and then assigns an output parameter, this violates AB#48049’s requirement that outputs become available only after all procedure results are consumed.

Please retain `RETURNVALUE`/`RETURNSTATUS` in `pending_output_params` whenever at least one DML count was collected, and deliver it only on the `SQLMoreResults` call after the final count returns `SQL_NO_DATA`. Add an e2e regression with `SET NOCOUNT OFF`, an update count, and an output parameter, asserting the destination remains unchanged until the terminating `SQLMoreResults`.

mssql-odbc/src/api/escape.rs
```diff
@@ -0,0 +1,1962 @@
...
+                let nested = translate_escapes_inner(body, inside_call, outputs, marker)?;
+                out.push('{');
+                out.push_str(&nested.sql);
+                out.push('}');
+            }
+            EscapeKind::Escape => {
+                push_translated(&mut out, &translate_escape_clause(remainder));
+                *marker += executable_marker_count(remainder);
+            }
+            EscapeKind::Interval => {
```
shiwanigupta0809, line 479: `EscapeKind::Interval` is translated even when nested inside a canonical `{call ...}`. msodbcsql instead passes date/time/interval escapes through while `fNestedInCall` is set (`sqlcmisc.cpp:4711-4716`) and restores the protected braces when rendering the text route (`sqlcmisc.cpp:8469-8495`). As a result, `SQLNativeSqlW("{call p({interval '1' DAY})}")` diverges from the reference driver.

Please preserve the interval escape when `inside_call` is true and add a golden `SQLNativeSqlW` test for an interval nested in `{call}`.

mssql-odbc/src/api/native_sql.rs
David-Engel: Fixed in 17240a9b: SQLNativeSql clears diagnostics and rejects negative input lengths other than SQL_NTS with HY090 before decoding. Added direct API and live regressions for invalid lengths, replacement diagnostics, and valid calls clearing prior errors; both retail reference builds pass the same live cases.

mssql-odbc/docs/odbc-escape-sequences-plan.md
David-Engel: Updated both design documents in 17240a9b to describe supported single-row output/return delivery, current bindings and offsets, reset/rebind behavior, and conversion diagnostics. The remaining output gap is arrays only; output arrays and TVPs now reference AB#48148, and the named-call routing defect is fixed.

mssql-odbc/src/api/odbc_types.rs
David-Engel: Moved the return-status documentation onto SQL_RETURN_VALUE in 17240a9b; the streamed-direction constants are no longer mislabeled.

mssql-odbc/src/api/num_params.rs
David-Engel: Fixed in 17240a9b by retaining the accepted direct statement's marker count separately from PreparedPlan. Unit and live regressions cover zero/nonzero direct counts, prepare/direct reuse, reset, and recovery after execution errors; the API suite passes on Rust, Windows retail 18.06.0001, and pinned Linux 18.6.2.1.

mssql-odbc/src/api/bind_param.rs
David-Engel: Both Windows retail 18.06.0001 and pinned Linux 18.6.2.1 accept this binding: integer output reports length 4 and NULL reports SQL_NULL_DATA; a null character destination has zero capacity even with positive BufferLength. This matches sqlcdesc.cpp's SQLBindParameter/SetIPDRec path and sqlctokn.cpp's GetReturnValue; 17240a9b removes the rejection and safely handles null-target delivery, with unit and live regressions.

mssql-odbc/src/api/exec_direct.rs
David-Engel: Fixed the shared RPC serializer in 17240a9b to prefix the checked UTF-16 unit count and reject over-limit names before serialization, preserving Unicode support. MS-TDS 2.2.6.6 permits 1046 bytes (523 UTF-16 units), rather than 255; boundary/serializer tests and Unicode direct/prepared E2Es pass.

mssql-odbc/src/api/exec_common.rs
David-Engel: Fixed in 17240a9b: drained RETURNVALUE/RETURNSTATUS tokens remain pending while any DML count is still observable, and delivery occurs after the final count is consumed. Also removed premature rowset advancement from prepared and data-at-execution paths; 14 direct/prepared live cases cover counts (including zero), unchanged NOCOUNT delivery, reset/rebind, and one-time diagnostics, passing Rust and both reference builds.

mssql-odbc/src/api/escape.rs
David-Engel: Fixed in 17240a9b by preserving INTERVAL escapes while inside CALL, matching the reference's fNestedInCall behavior. Added parser and SQLNativeSql golden regressions; the live case passes Rust, Windows retail 18.06.0001, and pinned Linux 18.6.2.1.

mssql-odbc/src/api/escape.rs
David-Engel: Fixed in 17240a9b: only unnamed markers qualify for positional RPC; named arguments retain their names on the text route. Reordered input/output regressions pass for direct and prepared calls on Rust and both retail reference builds; the parity tests set matching IPD names because msodbcsql's canonical RPC otherwise discards inline names.

David-Engel:
> This contradicts the PR description''s test-pass claims.

Corrected the description to distinguish local results from CI and identify the actual driver builds tested. Commit 17240a9b fixes the API/routing review findings and the Linux output-warning test sequencing without skips: unixODBC hid warnings attached to SQL_NO_DATA, so the tests now drain the output tail while the final rowset is nonempty and still require the exact warning/error status and SQLSTATE. The 61 focused parity cases pass on each of Windows retail 18.06.0001 and pinned Linux 18.6.2.1; local Rust validation passes 3,795 library tests and 189 E2Es. Fresh build 175114 is pending; no full-CI pass is claimed. The retry-related duplicate coverage artifacts in the older build were not patched into this feature branch.

Theekshna (commented):
Unattended hourly sweep found no findings at this head. Covered: re-verification of all previously-open review threads' fixes (named-call RPC routing, deferred output delivery, interval-in-call, `SQLNumParams`/`SQLNativeSqlW` validation, null-pointer output binding, RPC proc-name UTF-16 length), msodbcsql parity on the touched escape-translation table, divergence documentation, PR description currency, AI-slop scan, and an evidence spot-check against the reference tree; all 26 review threads are resolved and CI is fully green at this SHA.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

mssql-odbc/src/api/exec_direct.rs
```diff
@@ -174,11 +191,50 @@ fn sql_exec_direct_w_safe(
...
         publish_scalar_processed(&stmt_state);
-        let named_params =
-            match unsafe { build_named_params(&mut stmt_state, marker_count, "SQLExecDirectW") } {
-                Ok(params) => params,
-                Err(rc) => return rc,
+        let bind_offset = unsafe { stmt_state.inert_attrs.param_bind_offset() };
+        let mut has_dae = false;
+        for index in 0..marker_count {
+            let Some(Some(bound)) = stmt_state.bound_params.get(index) else {
+                post_diag(&mut stmt_state, ERR_UNBOUND_PARAMETER);
```
Vahid-b, line 198: **The unbound case returns `07002`, but this PR''s own design record says msodbcsql returns `07001`.**

`odbc-escape-sequences-plan.md` §3.3 states the contract this block implements:

> `?=` forces the first bound parameter to OUTPUT; `HY105` if bound `SQL_PARAM_INPUT` (`sqlcmisc.cpp:8310`), **`07001` if unbound**.

The `HY105` half is implemented exactly (line 205 below, and `execute.rs:934`). The unbound half is not: parameter 1 of `{? = call proc(?)}` falls through the generic `ERR_UNBOUND_PARAMETER` arm here, which is `07002` (`sqlstate.rs:175`). `07001` does not appear anywhere in the crate. `execute.rs:930` added a branch specifically for this case on the prepared route and made the same choice, so this reads as an oversight rather than a deliberate deviation.

`07001` ("wrong number of parameters") and `07002` ("COUNT field incorrect") are distinguishable by any application that switches on SQLSTATE, so this is observable.

Either match the doc, or record the deviation with a measurement — the doc line is a source reading (`sqlcmisc.cpp:8310`), not one of the entries tagged **[measured]**, so it is worth confirming against the pinned build before matching it.

**Separately: this validation has no test at all, on either route.** I changed `ERR_INVALID_PARAMETER_TYPE` to `ERR_FUNCTION_SEQUENCE` at both `exec_direct.rs:205` and `execute.rs:934` and all 1608 `mssqlodbc` unit tests still passed; grepping `HY105` across `mssql-odbc/` turns up only `execute_test.cpp:2144`, which is the unrelated Windows-DM streamed-direction case. Two `EscapeSequenceLiveTest` cases would close it and would run on the compare leg, settling the `07001`/`07002` question at the same time:

- `{? = call proc(?)}` with parameter 1 bound `SQL_PARAM_INPUT` → `SQL_ERROR` + `HY105`
- `{? = call proc(?)}` with parameter 1 unbound → `SQL_ERROR` + whichever state the reference driver reports

Both direct and prepared, since `exec_direct.rs` and `execute.rs` reach the check through separate code.

mssql-odbc/src/conversion/param_convert.rs
```diff
@@ -168,7 +199,15 @@ pub(crate) unsafe fn bound_param_to_rpc(
...
+        StatusFlags::BY_REF_VALUE
```
Vahid-b, line 206: **`BY_REF_VALUE` is the one bit that makes output parameters work, and nothing below the live-server e2e suite pins it.**

The comment two lines up is right that without this flag "an OUTPUT binding would be sent as a plain input and silently produce nothing". I replaced the whole conditional with `let status = StatusFlags::NONE;` and **all 1608 `mssqlodbc` unit tests still passed** — the failure only surfaces against a real server.

`output_only_does_not_read_uninitialized_buffers` (line 5607) calls `bound_param_to_rpc` and asserts on the returned `ConvOk` and the decoded `SqlType`, but never inspects the `RpcParameter`''s status flags, so it passes either way. The same is true of `output_only_indicators_never_stage_data_at_execution` in `exec_common.rs`, which asserts `built.params.len() == 1` but not what is in it.

This is behavior the e2e suite does cover (and that leg is green in CI), so it is not a correctness finding — it is that a one-line regression in the wire encoding costs a full live-server round trip to detect instead of a unit run. Two cheap options:

- Assert the status flag directly in `output_only_does_not_read_uninitialized_buffers`, e.g. that the `RpcParameter` returned for `SQL_PARAM_OUTPUT` / `SQL_PARAM_INPUT_OUTPUT` / `SQL_RETURN_VALUE` carries `BY_REF_VALUE` and that `SQL_PARAM_INPUT` does not.
- Or exercise it through `mssql-mock-tds` and assert the byte on the wire — `call_routing_test.cpp:199` already carries a `Benefits-from-mock-tds:` tag for the neighbouring RPC-name length, so the convention is in use here.

The DAE branch in `exec_common.rs` sets the same flag from `is_output_direction` and is equally unpinned, so one assertion covering both construction sites would be ideal.

Vahid-b (commented):
**Automated review** — generated by GitHub Copilot on behalf of @Vahid-b. This is **not** an approval and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

Reviewed the full merge-base diff (`88fee8ba..68d3b8da`, 38 files / ~6.7k lines) in a detached worktree. The previous automated pass on this PR was at `a61e843d`; this one covers the two commits added since (`17240a9b`, `68d3b8da`). All 26 review threads are resolved and I have not re-filed anything already raised.

Lenses that activated: correctness/reliability, security, protocol/wire-format, driver architecture, FFI/memory safety, spec/ABI conformance, test integrity, design/process.

**What I checked and did not find a problem with**

- `write_back_output_params` ordinal/name matching, the `index - usize::from(returns_status)` subtraction (unreachable underflow: the `?=` marker is always handled by the RETURNSTATUS branch first), and the `pending_output_params` lifecycle against `clear_exhaustion_state` ordering in both `finish_execute` and `release_busy_if_row_exhausted`.
- The `CodeScan` extraction: I diffed the new lexer against the deleted inline state machine in `rewrite_param_markers` character by character (`--(*` / `*)--` lookahead, doubled delimiters, non-nesting block comments, `prev1`/`prev2` update points) and they are equivalent.
- Null-destination delivery through `deliver_bound`: the `SQL_C_BINARY` and typed-C paths are safe because `write_value` forces `buffer_length` to 0, which makes `element_stride` 0 and `take` 0; `convert_typed_c` documents a null target.
- `SQLDescribeParam`'s `marker_count - usize::from(return_status)` cannot underflow, since a recovered `{? = call …}` always rewrites at least one marker on both the NOSCAN and translated paths.
- `quote_procedure_name` / `validate_procedure_name` bracket-escaping on the Always Encrypted `EXEC` sink, and the RPC `ProcName` UTF-16 unit count.

**Test integrity (differential, not read-only)**

I mutated the PR's own logic and re-ran `cargo nextest run -p mssqlodbc --lib` each time, restoring after every run:

| Mutation | Result |
| --- | --- |
| `no_row_execute_return`: disable the `row_count == 0` → `SQL_NO_DATA` branch | **2 tests fail** — real guard |
| `more_results`: drop `&& pending_row_counts.is_empty()` from the exhausted fast path | **2 tests fail** — real guard |
| `exec_direct`/`execute`: change the `{? = call …}` direction diagnostic `HY105` → `HY010` | **1608/1608 pass** — no guard |
| `param_convert`: replace `StatusFlags::BY_REF_VALUE` with `StatusFlags::NONE` for output directions | **1608/1608 pass** — no unit guard |

The first two are genuinely load-bearing. The last two are the two inline comments below.

**Validation run**

- `cargo nextest run -p mssqlodbc -p mssql-tds --lib --no-fail-fast` → 3788/3795 pass. The 7 failures are the pre-existing `certificate_validator` / `win_tls::validate` fixture failures on a clean tree that `.github/skills/code-review/SKILL.md` documents; they are unrelated to this diff.
- `cargo bclippy` → clean.
- `gh pr checks` → all green, including the containerized ODBC e2e leg with `ODBC_E2E_COMPARE=1` against pinned msodbcsql 18.6.2.1.

Neither finding is blocking. CI and the msodbcsql compare leg are green, and the design record matches the implementation everywhere else I checked.

mssql-odbc/src/api/exec_direct.rs
David-Engel: Added separate direct/prepared unit regressions and two unskipped live cases in 672fa2fd, with parameter 2 correctly bound to isolate the return marker. INPUT returns fail with HY105 on both routes; production diagnostics remain unchanged. The document's 07001 came from reading the resource name rather than clntcomn.cpp::GetSqlStateFromOdbcError: IDS_07_001 maps to 07001 for ODBC 2.x and 07002 for ODBC 3.x. Measurements on Windows retail 18.06.0001 and pinned Linux 18.6.2.1-1 agree: an unbound return with parameter 2 bound yields HY105 direct (descriptor-default direction validation), but 07002 prepared. Section 3.3 now records that distinction and Rust's consistent 07002; the full 22-case escape suite passes on Rust and both reference builds.

mssql-odbc/src/conversion/param_convert.rs
David-Engel: Added focused flag assertions in 672fa2fd for input/output/input-output/return directions, both named and positional scalar parameters, and streamed/buffered DAE construction plus buffered rebuild. A test-util-only accessor inspects the actual flags without widening the production API. Replacing BY_REF_VALUE with NONE at both construction sites now fails four focused unit tests; restored code passes those tests and all 3,801 ODBC/TDS library tests.

saurabh500 (approved):

PR #548, opened 2026-09-10
Move crate registry checks outside isolated release containers

saurabh500:
## Description

**Crates only.** Move existing crates.io preflight and availability checks out of intentionally network-isolated OneBranch containers into read-only Windows Azure Pipelines jobs. This fixes the socket-permission failure in [release run 174224](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=174224&view=logs) without weakening isolation.

- Validate original crate artifacts, check selected versions are absent, publish the core through governed ESRP, confirm core availability, then publish and confirm the mock. Core-only, mock-only, and validation-only selections retain their appropriate gates; validation-only never waits for an unpublished proposed core.
- Each crate job downloads and validates the original manifest/archive hashes from the same selected `officialBuild` resource. Publishers never consume files produced by registry jobs. Shared steps are explicitly referenced as `/.pipeline/templates/validate-release-crates.yml@self` so GovernedTemplates resolves them in this repository.
- Preserve ESRP@12, managed identity, service connection, completion waits, production context, policy validation, and publishing isolation. Network jobs receive no ESRP variable group or publishing connection. Existing registry script behavior is unchanged.

**Python/NuGet behavior is unchanged from the original base.** The earlier NuGet preflight, wheel-stage refactor, extra wheel revalidation, and conflict-policy change have been removed. The original `PublishRelease` job, wheel validation, optional NuGet packaging/publication, `continueOnConflict: true`, and tagging are restored exactly. No changes to the separate PyPI pipeline.

There is already an S360 work item in ADO tracking the release-job classification concern raised in review.

### Validation and limits

Current GitHub commit: `389a5ee6e2345f7316bd5e1f6ba297d06ca64127`. The latest review follow-up changes only comments, documentation, and tests; parsed pipeline configuration is identical to the native-previewed `f48da721893ab59d1bd7ecf0f14e55cfa6fdd102`.

- Latest `python -m pytest scripts/test_release_pipeline.py -q`: **108 passed**, including all 32 flag combinations, both-crate and mock-only failure/cancellation/skipping, existing registry HTTP/transport behavior, explicit core and mock archive corruption/missing cases, original Python behavior, and a regression requiring all six crate step references to be root-qualified with `@self`. The mock-only propagation cases reject an in-memory mutation that bypasses `RegistryPreflight`.
- The broader release/wheel/ODBC packaging suite passed **124 tests** before the template-binding fix. `cargo bfmt` and `git diff --check` pass for the latest changes. Rust sources/dependencies are unchanged; broad clippy and live-server suites were not rerun.
- Compared parsed pipeline configuration against the original base: everything outside `ReleaseCrates` is identical, including NuGet settings and tagging.
- **Correction to earlier evidence:** earlier previews inlined the crate step template into `yamlOverride`. They checked the job graph but did **not** validate repository/template resolution. A native preview of commit `6cb0bbec` reproduced the reported StageList error: the relative template was incorrectly sought in GovernedTemplates.
- After qualifying all six references, **six native ADO Preview API calls** for definition 2236 passed using actual committed Azure DevOps source, pinned to `f48da721893ab59d1bd7ecf0f14e55cfa6fdd102`, with **no `yamlOverride` and no inlining**. Combinations: full publication, all-off, core-only, mock-only, both-crate validation-only, and mock-only validation. Expanded assertions verified real template resolution, one checkout per job, policy checks, crate HTTP tasks without container/target/isolation injection, ESRP restricted to `windows_build_container`, and unchanged NuGet publication behavior. Later comments/docs/tests-only commits were not re-previewed; their parsed pipeline configuration is unchanged.
- Subsequent live runs on `dummyofficial` at ADO merge commit `d34d446a65161d9d122f56ac9de26f660cf9eed8` succeeded: [174309](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=174309) exercised core-only preflight, ESRP publication, and availability; [174316](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=174316) exercised mock-only preflight, ESRP publication, and availability. Both had NuGet publication and tagging disabled. These provide runtime evidence for all three registry-job types without `ob_outputDirectory`; the combined both-crate path has preview/local coverage rather than live evidence cited here.

GovernedTemplates' `Windows.Custom.Job.yml` supports the selected read-only custom-job context. Its artifact publication/post-job paths are conditional on publishing outputs, which the registry jobs do not do. No live run, publication, tag, release-branch write, or policy change was initiated during this review follow-up.

### ADO delivery

[ADO PR 8291](https://sqlclientdrivers.visualstudio.com/mssql-rs/_git/mssql-rs/pullrequest/8291) delivered `f48da721893ab59d1bd7ecf0f14e55cfa6fdd102` into `dummyofficial` and deleted its source branch on completion. Subsequent GitHub commits change only comments, documentation, and tests; the deleted ADO source branch has not been recreated.

## Related Issues

Related to #514. Follow-up to merged #515 and #538; their branches and PRs are unchanged.

## Checklist

- [x] `cargo bfmt` passes
- [ ] `cargo bclippy` passes
- [ ] `cargo btest` passes

No public API changes. Crate release execution and isolation are documented in `docs/release-management.md`.


docs/release-management.md
```diff
@@ -292,15 +314,18 @@ ESRP, Git tags, or release-branch writes.
...
+Custom registry jobs have one explicit `checkout: self`, since their template
+does not inject a checkout.
 
 For release-pipeline changes, use the ADO Preview API first to inspect expanded
-gates and OneBranch policy/checkout settings without queuing a run. Once a live
-run is authorized, select a known successful Official Build and leave all switches
+gates and OneBranch policy/checkout settings without queuing a run. Preview proves
+the selected agent context, not runtime pool authorization or registry reachability.
+Once a live run is authorized, select a known successful Official Build and leave all switches
 off to exercise genuine artifacts in the governed container without publishing.
```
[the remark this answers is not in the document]
saurabh500: Documented the non-publishing crate preflight configuration: set `validateCratesOnly: true` plus the intended crate selections, with `publishNuGet` and `tagRelease` false. Also clarified that all-off exercises only wheels, selected crate versions must be unpublished, and mock-only validation requires an existing core. No live run was queued.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -227,53 +227,78 @@ extends:
...
+            - checkout: self
+            - template: /.pipeline/templates/validate-release-crates.yml@self
+
+            - ${{ if eq(parameters.publishMssqlTds, true) }}:
+              - pwsh: >
+                  & "$(Build.SourcesDirectory)/.pipeline/scripts/test-cratesio-package.ps1"
+                  -CrateName 'mssql-tds'
+                  -Version '$(mssqlTdsCrateVersion)'
+                  -ExpectedState Absent
+                  -MaxAttempts 1
```
David-Engel, line 268: Worth reconsidering `-MaxAttempts 1` now that this actually reaches the network.

`test-cratesio-package.ps1` treats the `Absent` path as strictly single-shot: any exception (transport/DNS/TLS) throws immediately regardless of `MaxAttempts`, and any status that is not 404/200 throws on the first response — including 429 and 5xx. The new `("Absent", [429], False, 1)` and `("Absent", [503], False, 1)` cases in `test_release_pipeline.py` pin that behavior deliberately.

That was harmless while the step could never complete a request. On the `Azure Pipelines` hosted pool the egress IP is shared with a lot of other tenants, so a crates.io 429 or a transient blip is a realistic way for a release to abort before publishing anything. Retrying the `Absent` check is safe in the direction that matters: a retry can never turn "published" into "absent", since 200 still fails immediately and only 429/5xx/transport errors would be retried. Fail-fast on 200/401/403 should stay.

Smallest version of this: pass `-MaxAttempts 3 -DelaySeconds 5` here (and on the other two preflight steps) and let the script retry transient classes on the `Absent` path too. Not blocking — just cheap insurance against a preflight-only release failure that a human then has to re-queue.

.pipeline/templates/validate-release-crates.yml
```diff
@@ -0,0 +1,15 @@
...
+# Re-download the immutable selected build artifact in every job, never an
+# intermediate artifact or a directory supplied by the registry-check agent.
+steps:
+- download: officialBuild
+  artifact: 'drop_Build_RustCrates'
+  displayName: 'Download Rust crate artifacts'
+
+- pwsh: >
+    & "$(Build.SourcesDirectory)/.pipeline/scripts/validate-crate-release-artifact.ps1"
+    -ArtifactDirectory "$(Pipeline.Workspace)/officialBuild/drop_Build_RustCrates"
```
David-Engel, line 14: This step does double duty and the second job is the non-obvious one: `validate-crate-release-artifact.ps1` ends by emitting `##vso[task.setvariable variable=mssqlTdsCrateVersion]` / `mssqlMockTdsCrateVersion`. That is why `CoreAvailable` and `MockAvailable` need the download at all — they consume those job-scoped variables, not the files.

As written, someone optimizing the post-publish jobs would reasonably drop this template ("the crate is already published, why re-hash it?") and the wait would then query `https://crates.io/api/v1/crates/mssql-tds/$(mssqlTdsCrateVersion)` literally, 404 for 40 attempts, and fail ~10 minutes later with a misleading "was not available" message.

One line in the existing header comment would prevent that, e.g. "...and publishes the manifest versions as `mssqlTdsCrateVersion`/`mssqlMockTdsCrateVersion` for later steps in the same job."

scripts/test_release_pipeline.py
```diff
@@ -74,6 +77,9 @@ def expand(value, parameters):
                 result.append(expanded)
         return result
     if isinstance(value, dict):
+        if value.get("template") == "/.pipeline/templates/validate-release-crates.yml@self":
```
David-Engel, line 80: This inlines exactly one hard-coded template path. Any other `- template:` reference added to the release pipeline later falls through to the generic dict branch and is returned as an opaque step, so the per-job assertions below (`len(downloads) == 1`, `job_names.index("Validate Rust crate release artifact")`, `validation_index == (2 if isCustom else 1)`) silently stop covering its contents rather than failing.

Consider making the unknown case explicit, e.g. `assert "template" not in value, f"Unhandled template reference: {value['template']}"` on the fallback path, so a new template forces a conscious update to the harness. Same spirit as `test_crate_templates_resolve_in_self_repository`, which already hard-codes the count `6` — both will need touching together, and an explicit failure message is friendlier than a confusing index error.

David-Engel (commented):

## Summary

Release-pipeline-only change: the crates.io preflight and availability checks move out of the network-isolated OneBranch containers into read-only custom Windows jobs, while ESRP publication stays governed. I walked all 32 switch combinations of the expanded job graph and it is behaviorally equivalent to the previous single-job flow:

- `ValidateCrates -> RegistryPreflight -> PublishCore -> CoreAvailable -> PublishMock -> MockAvailable`, with `PublishMock` correctly falling back to `dependsOn: RegistryPreflight` in the mock-only case. Every `dependsOn` target exists in every reachable combination — `CoreAvailable` only guards `PublishMock` when `publishMssqlTds` is true and `validateCratesOnly` is false, which is exactly when it is emitted.
- The reason every crate job re-includes `validate-release-crates.yml` is not just artifact integrity: `validate-crate-release-artifact.ps1` emits `##vso[task.setvariable]` for `mssqlTdsCrateVersion`/`mssqlMockTdsCrateVersion`, so each job derives the version it checks from the immutable manifest rather than from a prior job. That makes the "no cross-job state" claim in the docs hold, and it is why the availability jobs need the download too.
- ESRP inputs, `templateContext`, the ESRP variable group scoping, `waitforreleasecompletion`, NuGet settings (`continueOnConflict: true`), the `Release` job, and tagging are unchanged. Custom jobs receive no ESRP group and no release connection; the `@self`-qualified template paths match the convention already used by `stages.yml@self` and `PublishFeeds-Sandbox.yml`.

No Rust sources or dependencies changed. Nothing blocking; three smaller items below, the first being the one I would actually act on.

## Verification

- Read the full diff (4 files) against merge base `0d431b71`, plus the unchanged `test-cratesio-package.ps1` and `validate-crate-release-artifact.ps1` they orchestrate.
- Ran `python -m pytest scripts/test_release_pipeline.py -q` in an isolated worktree: **91 passed**, matching the PR claim. That includes the new graph, failure-propagation, HTTP-state, and artifact-integrity cases.
- `git diff --check` clean. No cargo runs (no Rust changes). No ADO run was queued and no pipeline was previewed from here — runtime pool authorization and actual agent-to-crates.io reachability remain unverified by this review, as the PR body already notes.

## Findings

### Suggestion

1. Preflight `Absent` checks fail fast on transient crates.io responses, and they now run on a shared-egress hosted pool — see inline on `.pipeline/OneBranch/OfficialPythonWheelsRelease.yml`.

### Nit

2. The shared template's `setvariable` side effect is load-bearing but undocumented — see inline on `.pipeline/templates/validate-release-crates.yml`.
3. `expand()` silently passes through unrecognized `- template:` references — see inline on `scripts/test_release_pipeline.py`.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Agreed that bounded retries could improve resilience, but I’m keeping the existing retry policy unchanged in this PR. The scope is moving the crate registry checks out of isolated containers and preserving publication ordering. A transient preflight failure stops before publication rather than permitting an incorrect release. Retrying the Absent path is a reasonable separate improvement, but it requires changing the script’s behavior as well as the attempt count; I’d prefer not to expand this fix to include that.

> [!NOTE]

.pipeline/templates/validate-release-crates.yml
saurabh500: Added a line to the template header in 8ef0b491 explaining that it also sets `mssqlTdsCrateVersion` and `mssqlMockTdsCrateVersion` from the manifest for later steps in the same job. No execution behavior changed.

> [!NOTE]

scripts/test_release_pipeline.py
saurabh500: I’m not adding the blanket assertion here. This helper also traverses the external `extends` template and unrelated templates in `stages.yml`, which it deliberately leaves unexpanded; rejecting every other `template` key would break those existing uses. Replacing or removing the required crate template already fails the download/validation assertions, so that case does not silently pass. A newly added opaque template could leave its contents outside those assertions, but addressing that needs a scoped check rather than the proposed blanket guard. I’d keep that broader harness improvement separate from this execution-context fix.

> [!NOTE]

Vahid-b (approved):

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -227,53 +227,78 @@ extends:
...
+
+        - ${{ if or(eq(parameters.publishMssqlTds, true), eq(parameters.publishMssqlMockTds, true)) }}:
+          - job: RegistryPreflight
+            displayName: 'Check selected crates.io versions'
+            dependsOn: ValidateCrates
+            pool:
+              type: windows
+              isCustom: true
+              name: Azure Pipelines
+              vmImage: windows-2022
```
Vahid-b, line 257: **Suggestion 1 — this job (and `CoreAvailable`, `MockAvailable`) omits `ob_outputDirectory`.**

Cited: every other `isCustom: true` job in this repo declares it — `stages.yml` lines 74, 118, 178, 246, 325, 406 and `PublishFeeds-Sandbox.yml` lines 117, 151, 202, 255, 308, 360, 407, 466, 529, 583. That is 16 for 16, and it includes the closest analogue: `MacOS_universal2` at `stages.yml:402`, which is the same `name: Azure Pipelines` hosted-pool shape you are introducing here. These three new jobs declare no `variables:` block at all.

Your six preview calls would only have caught this if the governed template consumes the variable at compile time. If it is consumed at runtime — by the output-collection / SDL step OneBranch appends — the first time it surfaces is inside a real release, and for `CoreAvailable` that is *after* an irreversible ESRP publish.

```yaml
            pool:
              type: windows
              isCustom: true
              name: Azure Pipelines
              vmImage: windows-2022
            variables:
              ob_outputDirectory: '$(Build.ArtifactStagingDirectory)'
            steps:
```

I could not read `Windows.Custom.Job.yml` to settle whether it is required — my ADO connection is scoped to `sqlclientdrivers`, so this is the repo convention talking, not the template. If it genuinely is optional for read-only custom jobs, a one-line comment saying so would stop the next person from "restoring" it.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -293,23 +318,42 @@ extends:
...
                 domaintenantid: '975f013f-7f24-47e8-a7d3-abc4752bf346'
 
+          - job: CoreAvailable
+            displayName: 'Verify published mssql-tds is available'
+            dependsOn: PublishCore
+            pool:
+              type: windows
+              isCustom: true
+              name: Azure Pipelines
+              vmImage: windows-2022
```
Vahid-b, line 330: Same as `RegistryPreflight`: no `ob_outputDirectory`. This is the one where a late runtime failure costs the most — it runs *after* `PublishCore`'s ESRP publish has already happened, and its failure skips `PublishMock`, leaving a half-published release.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -329,8 +373,19 @@ extends:
...
                 domaintenantid: '975f013f-7f24-47e8-a7d3-abc4752bf346'
 
+          - job: MockAvailable
+            displayName: 'Verify published mssql-mock-tds is available'
+            dependsOn: PublishMock
+            pool:
+              type: windows
+              isCustom: true
+              name: Azure Pipelines
+              vmImage: windows-2022
```
Vahid-b, line 385: Same as the other two: no `ob_outputDirectory`.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
```diff
@@ -227,53 +227,78 @@ extends:
...
       - stage: ReleaseCrates
         displayName: 'Release Rust Crates'
         dependsOn: []
         jobs:
-        - job: PublishRustCrates
-          displayName: 'Publish Rust crates to crates.io'
-          templateContext:
-            type: releaseJob
-            isProduction: true
+        - job: ValidateCrates
```
Vahid-b, line 240: **Nit 5 — say why this job is kept.**

`RegistryPreflight` is emitted for `or(publishMssqlTds, publishMssqlMockTds)` and re-runs this identical template, so `ValidateCrates` is duplicated work in every combination where a crate switch is on. It stands alone only in the `validateCratesOnly: true` with no crate switch case — which performs no registry check at all.

The design reason is good and I had to reconstruct it: integrity is proven *inside* the sealed container before the artifact is ever handed to a shared-tenant hosted agent, and a bad artifact skips everything downstream instead of reaching the network. Nothing in the file says that, so the next person optimizing this reads a redundant job and collapses it. One line above the job, in the same spirit as the header comment you added to the template in `8ef0b491`.

docs/release-management.md
```diff
@@ -292,16 +314,24 @@ ESRP, Git tags, or release-branch writes.
 NuGet and tag metadata come from the selected build's exact source commit, not
 the release pipeline's checkout or the current branch tip. A missing branch,
 commit, or required source file fails the operation without a checkout fallback.
-These jobs configure OneBranch's built-in checkout with `ob_git_fetchDepth` and
+Governed Git jobs configure OneBranch's built-in checkout with `ob_git_fetchDepth` and
 `ob_git_persistCredentials`; do not add a second `checkout: self`. A duplicate
 checkout can relocate the repository while governed task restrictions prevent
 updating the source path. Git commands use the explicit source directory, and
 wheel validation and packaging remain inside the governed build container.
+Custom registry jobs have one explicit `checkout: self`, since their template
```
Vahid-b: **Suggestion 3 — the justification here is contradicted by `stages.yml`.**

> Custom registry jobs have one explicit `checkout: self`, since their template does not inject a checkout.

Measured: `grep -rn checkout .pipeline/OneBranch/stages.yml` returns **zero** hits, yet every `isCustom: true` job in that file runs repo-relative scripts that require a working tree — `RustCrates` at `stages.yml:69` invokes `.pipeline/scripts/stamp-crate-versions.ps1` with a bare relative path, and `MacOS_universal2` at `stages.yml:402` is on the same hosted pool. Those jobs build the shipping product, so they are not broken.

The code is almost certainly fine: an explicit `checkout` suppresses ADO's implicit one, so you get exactly one either way, which is what your preview asserted. It is the stated reason that is wrong — and a wrong reason is expensive in *this* paragraph specifically, because the lines immediately above it warn that a duplicate checkout "can relocate the repository while governed task restrictions prevent updating the source path." Someone reading both will draw the wrong conclusion about when a second checkout is safe.

Suggest rewording to what was actually verified, e.g. *"Custom registry jobs declare an explicit `checkout: self`; preview confirms exactly one checkout per job."*

scripts/test_release_pipeline.py
```diff
@@ -270,6 +346,143 @@ def test_release_switch_graph(values):
...
+            content = manifest.encode("utf-8")
+            member = tarfile.TarInfo(f"{name}-0.1.0/Cargo.toml")
+            member.size = len(content)
+            archive.addfile(member, io.BytesIO(content))
+        entries.append({
+            "name": name, "version": "0.1.0", "file": f"{name}/{crate.name}",
+            "sha256": hashlib.sha256(crate.read_bytes()).hexdigest(),
+        })
+    if damage == "hash":
+        crate.write_bytes(b"replaced after preflight")
```
Vahid-b: **Nit 6 — the damage cases act on an implicit loop leftover.**

After `for name in ("mssql-tds", "mssql-mock-tds")`, `crate` is whichever file the last iteration built, so `damage == "hash"` here and `damage == "missing"` below both silently target the **mock** crate only. The core-crate equivalents are never exercised, and a reader cannot tell which crate is being damaged without replaying the loop.

An explicit binding would make that a choice rather than an accident:

```python
    damaged = tmp_path / "mssql-mock-tds" / "mssql-mock-tds-0.1.0.crate"
```

or parametrize over which crate gets damaged, since `validate-crate-release-artifact.ps1` hashes both.

scripts/test_release_pipeline.py
```diff
@@ -270,6 +346,143 @@ def test_release_switch_graph(values):
             assert all('git -C "$(Build.SourcesDirectory)"' in line for line in git_commands)
 
 
+@pytest.mark.parametrize("failed_gate", [
+    "PublishRelease", "ValidateCrates", "RegistryPreflight",
+    "PublishCore", "CoreAvailable", "PublishMock", "MockAvailable",
+])
+@pytest.mark.parametrize("result", ["Failed", "Canceled", "Skipped"])
+def test_release_gate_failure_propagation(failed_gate, result):
+    flags = dict.fromkeys(_SWITCHES, True) | {"validateCratesOnly": False}
```
Vahid-b: **Nit 7 — one switch combination.**

This pins skip-propagation only for the full six-job graph. The mock-only shape, where `PublishMock` depends on `RegistryPreflight` rather than `CoreAvailable`, never propagates through here. `test_release_switch_graph` does assert that `dependsOn` target — I mutated it to `ValidateCrates` and four cases failed — so the real exposure is small, but adding `(core, mock)` to the parametrization is a one-liner and would cover the branch that this PR actually introduces.

Vahid-b (changes requested):
## Summary

Withdrawing my earlier approval so the findings below get a decision before this lands — none of them are correctness defects in the job graph, which I re-walked and believe is right. The graph, the `@self` qualification, and the "no cross-job state" property all hold. What changed my mind is Suggestion 2: the run this PR cites reported a **second** error that this PR will duplicate rather than fix, and I'd rather that be a deliberate choice than an oversight.

Re-reviewed at head `8ef0b491` (one commit past the `f48da721` the description pins).

## Verification

**Read** — full `$BASE..HEAD` diff (`BASE=0d431b71`) in a dedicated worktree at `8ef0b491`; plus the unchanged `test-cratesio-package.ps1` and `validate-crate-release-artifact.ps1` these jobs orchestrate, and `stages.yml`, `PublishFeeds-Sandbox.yml`, `PyPIRelease.yml`, `.github/workflows/pr-linked-issue-check.yml` for convention comparison. All three prior-discussion endpoints read — David's three inline threads and your replies, both Copilot review bodies, and the coverage comment. I have not re-filed anything already answered there.

**Ran** —

- `pytest scripts/test_release_pipeline.py -q` → 55 passed / 36 failed here, against 33 passed / 13 failed on `$BASE`. Every failure on both sides is environmental: on WSL `pwsh` resolves to the Windows `pwsh.exe`, which cannot take the Linux script paths the harness builds. In CI this suite runs only on the `Build Windows` job (`validation-stages.yml:131`), where `pwsh` is native. Not a finding against the PR.
- **Four mutations, all caught** — this is the part I actually wanted to know, and the new tests do their job:

| Mutation | Result |
|---|---|
| Drop `@self` from one of the six template references | `test_crate_templates_resolve_in_self_repository` + 29 `switch_graph` cases fail |
| Revert `RegistryPreflight` to a governed (isolated) job | 24 `switch_graph` cases fail |
| `PublishMock` else-branch `dependsOn: RegistryPreflight` → `ValidateCrates` | 4 cases fail |
| Remove `checkout: self` from `CoreAvailable` | 8 `switch_graph` cases fail |

- ADO: `pipelines/builds/174224` status and logs; `git fetch ado stable && git show 64181c7a:.pipeline/OneBranch/OfficialPythonWheelsRelease.yml`.

**Not verified** — I could not read GovernedTemplates' `Windows.Custom.Job.yml` (different org; my ADO connection is scoped to `sqlclientdrivers`), so whether `ob_outputDirectory` is required in a custom job rests on this repo's convention rather than on the template source. I also did not isolate which of the four crate steps produced the socket error in 174224 — I confirmed from the expanded-YAML log that all four ran with the `(windows_build_container)` suffix, and from the build report that the job ended with `PowerShell exited with code '1'`, but not the step. No run or preview was queued from here.

## Blocking

None. Everything below is a Suggestion or a Nit; the `REQUEST_CHANGES` is about getting a decision on items 1 and 2, not about a defect in the graph.

## Suggestions

**2. Run 174224 also reported a OneBranch policy error that this PR is positioned to fix, and will instead duplicate.**

Measured. The build report for **174224** lists two errors under `Publish Rust crates to crates.io`, not one:

```
Error  PowerShell exited with code '1'.
Error  OneBranch PT Non-Blocking Error: Job 'PublishRustCrates' uses the release
       tasks ["esrprelease"] but is not marked as a release job. Release tasks will
       soon be disallowed outside of release jobs. ... Please address this issue to
       resolve the corresponding S360 item.
```

And `git show 64181c7a:.pipeline/OneBranch/OfficialPythonWheelsRelease.yml` — the exact commit that run used — shows the job **was** already marked:

```yaml
        - job: PublishRustCrates
          templateContext:
            type: releaseJob
            isProduction: true
```

So on that run, `templateContext: {type: releaseJob, isProduction: true}` alone did not satisfy the PT check. I am deliberately not asserting *why* — I could not read the governed template, so I have the observation without the mechanism. But `PublishCore` and `PublishMock` carry that identical shape, so the observable consequence of this PR is most likely two instances of the same S360 item where there was one.

I don't think this has to be fixed here, and "soon be disallowed" is a deadline rather than a today problem. What I'd like before merge is a decision: fix it, or file it and say so in the description. The body currently reads as though the release-job marking is settled ("Preserve ESRP@12, managed identity, service connection, ... and publishing isolation"), and 174224 says it isn't.

Inline items 1 and 3 below, plus the nits.

## Nits

**4. The description is stale in two ways.** It pins `Current commit: f48da721893ab59d1bd7ecf0f14e55cfa6fdd102`, but head is `8ef0b491` — the commit that answered David's template-header nit. And it closes with "Keep this PR draft pending authorized live validation" while the PR is not a draft. Either re-draft it or drop that line, so nobody assumes the live-validation gate is still enforced by the PR state. The six-preview evidence paragraph is also pinned to `f48da721`; worth one line saying whether `8ef0b491` (comment-only) was re-previewed or is exempt.

**7. `test_release_gate_failure_propagation` models a single switch combination.** `flags = dict.fromkeys(_SWITCHES, True) | {"validateCratesOnly": False}` means skip-propagation is only exercised on the full six-job graph; the mock-only shape, where `PublishMock` hangs off `RegistryPreflight` instead of `CoreAvailable`, never propagates through it. `test_release_switch_graph` does pin that `dependsOn` target — my M3 mutation fails there — so the exposure is small, but the parametrization is one line. Inline at `scripts/test_release_pipeline.py:355`.

## Things I agree should stay deferred

- **David's `-MaxAttempts 1` point.** Your deferral reasoning is right — it needs a script change, not just an attempt count, since the `Absent` path throws on the first non-404/200 response *and* on any transport exception regardless of `MaxAttempts`. But this PR is what makes that path reachable on a shared-egress hosted pool for the first time, so it should get a GitHub issue rather than living only in a closed thread.
- **The `expand()` opaque-template passthrough.** Your reasoning about `stages.yml` / `extends` traversal is correct, so the blanket assertion would indeed be wrong. A scoped version — assert no unhandled `template` reference *within* `extends.parameters.stages` — would close it without touching the others. Also follow-up material.
- **Suggestion 2** needs its own issue regardless of what you decide here.

David-Engel (commented):
**Automated review** — generated by GitHub Copilot on behalf of @David-Engel. This is **not an approval** and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

No findings at `8ef0b491`. Re-review since the previous automated pass at `f48da721`; the only new content is the one-line comment in `.pipeline/templates/validate-release-crates.yml` recording that `validate-crate-release-artifact.ps1` sets `mssqlTdsCrateVersion`/`mssqlMockTdsCrateVersion` for later steps in the same job. That is exactly the nit raised last pass, and the wording matches the actual `##vso[task.setvariable ...]` emission at the end of that script (job-scoped, no `isOutput`), which is why every crate job re-runs the template instead of consuming a prior job's variables.

I re-walked the full diff rather than only the delta:

- Job graph `ValidateCrates -> RegistryPreflight -> PublishCore -> CoreAvailable -> PublishMock -> MockAvailable` holds across all switch combinations; every `dependsOn` target is emitted under the same conditions as its dependent, including `PublishMock` falling back to `RegistryPreflight` in the mock-only case and `RegistryPreflight` being emitted whenever either crate switch is on.
- Registry jobs carry no ESRP variable group, no release service connection, and no `EsrpRelease@12`/`NuGetCommand@2` task; ESRP publication keeps `templateContext: releaseJob`, `isProduction: true`, `waitforreleasecompletion: true`, managed identity, and the same `folderlocation` under the downloaded `officialBuild` artifact. The `Release` job, NuGet feed settings (`continueOnConflict: true`), and tagging are unchanged from base.
- The availability wait is bounded: `test-cratesio-package.ps1` defaults to 40 attempts at 15s, so ~10 minutes plus request time — comfortably inside a hosted-agent job timeout.

**Verification**: read the complete diff (4 files, +373/-59) against merge base `0d431b71`, plus the unchanged `test-cratesio-package.ps1` and `validate-crate-release-artifact.ps1`, and the `isCustom` job conventions in `stages.yml`. Ran `python -m pytest scripts/test_release_pipeline.py -q` in an isolated worktree: **90 passed, 1 failed**. The failure is `test_missing_git_context_fails_explicitly`, which is untouched by this PR and fails here for an environmental reason (its fixture fetch resolved against the real remote in my sandbox), not because of this change. All new and modified cases pass. GitHub and ADO validation are green at this head. No ADO run was queued and no pipeline preview was performed from here, so runtime pool authorization and actual agent-to-crates.io reachability remain unverified by this review, as the PR body already states.

Not raising anything new: the open human threads on `ob_outputDirectory` for the three custom jobs, the `checkout` justification in `docs/release-management.md`, and the test nits are already on the PR and are the right place for that discussion. I could not independently settle the `ob_outputDirectory` question either — the successful template expansion in preview rules out a compile-time requirement, but not a runtime one.

This review was produced by an unattended automated run; findings were not checked by a human before posting.

Theekshna (commented):
Unattended hourly sweep: no new findings at head `8ef0b491`. Covered: full `BASE..HEAD` diff against merge base `0d431b71` (4 files, +373/-59) in a fresh worktree; all three prior-discussion endpoints (inline comments, review bodies incl. the collapsed Copilot detail, issue comments) plus `gh pr checks` (all green); the six-check list (msodbcsql parity and divergence-documentation are N/A -- no Rust/ODBC source touched); independently ran `python -m pytest scripts/test_release_pipeline.py -q` in this worktree (91 passed, matching the PR's own claim, re-deriving that evidence rather than accepting it). David-Engel's automated pass is clean at this same head, and Vahid-b's human `CHANGES_REQUESTED` review (also at this head, with its own 4-mutation verification table) already covers the open items -- the `ob_outputDirectory` question, the S360 release-task duplication, and the stale commit reference/draft line in the description -- awaiting the author's decision, so I'm not re-filing any of them. Nothing new to add.

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Leaving this unset for the same reason as `RegistryPreflight`: `Windows.Custom.Job.yml` does not require an output directory for these read-only jobs without artifact publication. `CoreAvailable` also completed successfully after ESRP publication in [174309](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=174309). Added the shared rationale in 389a5ee6; no execution settings changed.

> [!NOTE]

scripts/test_release_pipeline.py
saurabh500: Updated in 389a5ee6. Hash-corruption and missing-archive cases now each run for both `mssql-tds` and `mssql-mock-tds`, using an explicit parameterized path rather than the loop's final `crate` binding. The production validator is unchanged.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Leaving this unset for the same read-only custom-job reason described on `RegistryPreflight`. `MockAvailable` completed successfully after ESRP publication in [174316](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=174316) without an output directory. The shared comment added in 389a5ee6 explains why these jobs omit it.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Added a short comment above `ValidateCrates` in 389a5ee6 explaining that the isolated integrity gate is retained before artifacts reach a network job. The graph and validation behavior are unchanged.

> [!NOTE]

scripts/test_release_pipeline.py
saurabh500: Added mock-only failure propagation in 389a5ee6 alongside the existing both-crate cases, covering failed, canceled, and skipped gates. Confirmed that changing the mock-only dependency to bypass `RegistryPreflight` is rejected in all three cases. This changes tests only; the release graph is unchanged.

> [!NOTE]

docs/release-management.md
saurabh500: Agreed. Reworded in 389a5ee6: the explicit checkout in custom registry jobs replaces Azure Pipelines' implicit checkout rather than adding a second one, and native template preview confirms exactly one checkout per job. No checkout steps changed.

> [!NOTE]

.pipeline/OneBranch/OfficialPythonWheelsRelease.yml
saurabh500: Leaving `ob_outputDirectory` unset. I was able to inspect `Windows.Custom.Job.yml`: automatic artifact publication is gated by `ob_requirePublish` and `ob_pipelineartifacts_enabled`, and the associated post-job path is emitted only when artifacts are published. These read-only registry jobs do neither.

There is now runtime evidence as well: [174309](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=174309) completed the core-only preflight, publication, and availability check; [174316](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=174316) completed the mock-only flow. The registry jobs succeeded without this variable. Added a shared comment explaining the omission in 389a5ee6. No job settings changed.

> [!NOTE]

saurabh500:
Regarding suggestion 2: there is already an S360 work item in ADO tracking this.

David-Engel (commented):

## Summary

No findings at `389a5ee6`. This is a re-review since the previous automated pass at `8ef0b491`; the delta is the three review responses and nothing else. I re-walked the full diff against merge base `0d431b71` rather than only the delta, and the conclusions from the prior pass still hold: the job graph `ValidateCrates -> RegistryPreflight -> PublishCore -> CoreAvailable -> PublishMock -> MockAvailable` is emitted consistently across every switch combination, every `dependsOn` target is emitted under the same conditions as its dependent (including `PublishMock` falling back to `RegistryPreflight` in the mock-only case), registry jobs carry no ESRP variable group, no release service connection and no `EsrpRelease@12`/`NuGetCommand@2` task, and the `Release` job, NuGet settings (`continueOnConflict: true`) and tagging are unchanged from base.

On the delta specifically:

The `ob_outputDirectory` and `ValidateCrates` comments are accurate as written. `ValidateCrates` does set `ob_outputDirectory`, but the sentence that says it is not required is scoped to "read-only custom jobs" in the preceding line, so it reads correctly; the three jobs it describes are the ones that omit the variable.

The `docs/release-management.md` rewording is correct on the substance Vahid-b measured. Azure Pipelines injects an implicit `checkout: self` when a job declares no checkout step, so an explicit one replaces it rather than adding a second — which is consistent with the `isCustom` jobs in `stages.yml` that declare no checkout at all yet run repo-relative scripts.

The two test changes are not cosmetic; both actually guard the branch they name, which I verified by mutation rather than by reading:

Reverting the mock-only `dependsOn: RegistryPreflight` to `ValidateCrates` now fails `test_release_gate_failure_propagation[Failed-False-RegistryPreflight]` and its `Canceled`/`Skipped` siblings — three cases that did not exist before this commit, because the old parametrization only ever built the both-crate graph. Making the hash comparison in `validate-crate-release-artifact.ps1` blind to the core crate (`-and $entry.name -like "*mock*"`) fails only `test_crate_artifact_revalidation[hash-mssql-tds]`, the case this commit added; every other case in that test still passed under the mutation, which is exactly the blind spot Nit 6 described.

## Verification

Read the complete diff (4 files, +394/-59) against merge base `0d431b71` in the assigned worktree at `389a5ee6`, plus the unchanged `test-cratesio-package.ps1` and `validate-crate-release-artifact.ps1` these jobs orchestrate and the `isCustom` job conventions in `stages.yml`. Read all prior discussion endpoints live before drafting: both Copilot review bodies, Vahid-b's `CHANGES_REQUESTED` review, Theekshna's sweep, the previous automated pass, and all eleven review threads — every one is currently resolved with an author reply, and I am not re-filing anything from them.

Ran `python -m pytest scripts/test_release_pipeline.py -q` in the worktree: **108 passed** in 86s, matching the PR body's claim exactly. Then ran the two mutations described above and restored the worktree clean (`git status --short` empty) afterward.

`gh pr checks 548` is green on GitHub and on the ADO PR validation run 174391, with two macOS jobs still pending at the time of review. No Rust sources or dependencies changed, so no cargo run was warranted. No ADO run was queued and no pipeline preview was performed from here, so runtime pool authorization and actual agent-to-crates.io reachability remain outside what this review can confirm — the PR body already states that, and runs 174309/174316 are the author's evidence for it.

## Findings

None. This looks ready for human review, but this is **not an approval** — Vahid-b's `CHANGES_REQUESTED` is still the outstanding gate and only a human can clear it.

This review was produced by an unattended automated run; findings were not checked by a human before posting.

Theekshna (approved):

Vahid-b (approved):

PR #533, opened 2026-09-09
Convert varchar(max)/text PLP chunks from the server collation to UTF-8 (AB#47566)

Vahid-b:
## Description

`SQL_C_CHAR` output is UTF-8 on this driver, but PLP delivery of a `SingleByteText` column copied the wire bytes **verbatim**. Under a non-UTF-8 collation the caller got raw code page bytes labelled UTF-8. The non-PLP path was already correct — it decodes through the collation via `SqlString::to_utf8_string` — so the *same value* delivered inline and streamed disagreed byte for byte.

The offending line documented its own defect:

```rust
// Delivered verbatim today, so a non-UTF-8 server collation yields raw
// codepage bytes labelled UTF-8. Conversion attaches here: [AB#47566](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47566).
Some(PlpEncoding::SingleByteText) => copy_verbatim(),
```

### What changed

- **`get_data.rs` (`SQLGetData`)** — `SingleByteText` is decoded through the column's collation. Reuses the per-stream `encoding_rs::Decoder` that already served the `SQL_C_WCHAR` widening, so a multi-byte DBCS sequence split across a chunk boundary is carried rather than corrupted. `pending_utf8` holds back output the caller's buffer had no room for — decoding expands, e.g. CP1252 `0x80` is 3 UTF-8 bytes.
- **`fetch_scroll.rs` (`SQLBindCol` + `SQLFetch`)** — carried the byte-for-byte identical verbatim copy. Converted the same way. This arm was **latent**: no test read a `varchar(max)` under a non-UTF-8 collation through a bound fetch, so nothing failed on it.
- **The resolved encoding lives on `ActivePlpStream`**, with the decoder built lazily on first need. Keying it on the first call's target type meant a `SQL_C_BINARY` probe — the shape mssql-python issues per column — left a later `SQL_C_CHAR` read on the verbatim path, reintroducing the very defect being fixed.
- **`max_read` sized for expansion** via `narrow_max_read`, mirroring `utf16le_max_read`.
- **UTF-8 collations and `json` stay verbatim**, explicitly not double-converted — and now share a truncation gate. `trim_partial_utf8` was keyed on `json` alone, so a UTF-8-collated `varchar(max)` truncating in a bound slot could end mid-character. Pre-existing, but this change is what makes the two shapes equivalent, so they are generalized to one `verbatim_utf8_text` condition.
[...]

PR #552, opened 2026-09-10
Correct the stale bound-PLP size guidance in the agent instructions (AB#48048)

David-Engel:
Resolves [AB#48048](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/48048).

## The claim, verified

`.github/instructions/mssql-odbc.instructions.md` told agents that a `max`/LOB column only reaches the bound *streaming* path when it is too large for the transport to buffer whole, that "a few kilobytes still buffers", and that `deliver_bound_plp` is "reachable around a megabyte". That has been false since #446.

Checked mechanically rather than taken on faith:

- `try_read_row_column` attempts `try_begin_buffered_plp` for any column whose metadata says `is_plp()` and yields `CursorColumn::PlpStreaming`. The match arm is gated on `is_plp()` alone — no length appears in the decision at all.
- `plp_indicator` has exactly **one** production call site, inside `deliver_bound_plp`, and is the only producer of `SQL_NO_TOTAL` on the bound path.
- So `ABoundVarcharMaxTruncatedToWcharReportsNoTotal` already proves the point with **5,000** wire bytes — it asserts `SQL_NO_TOTAL`, which nothing else can produce. I ran it: it passes today.

I also measured it independently while working [AB#47240](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47240): a bound `varbinary(max)` of **10 bytes** reached `deliver_bound_plp`.

## The tail was wrong too

The bullet closed by saying a bound binary `max` column into a typed C target answers `22018` when buffered and `HYC00` when streamed. Measured on 18.06.0001:

| Case | mssql-odbc | msodbcsql |
| --- | --- | --- |
| `varbinary(8)` → `SQL_C_SLONG` | `07006` | `07006` |
| `varbinary(max)` (small) → `SQL_C_SLONG` | **`HYC00`** | **`07006`** |
| `varchar(8)` / `varchar(max)` `'42'` → `SQL_C_SLONG` | `42` | `42` |

Neither half held: non-`max` binary is `07006`, not `22018`. And the `max` row is a **real parity divergence** that the file was describing as a harmless artifact of buffering — see below.

## Also

Separated out `PLP_TYPED_MATERIALIZE_LIMIT`, a genuine and unrelated 1 MiB cap on how much a *typed* conversion will materialize, so the two 1 MiB numbers stop being conflated.

## Why this mattered

This is guidance that steers agents. It has already produced bad work: it misled a Copilot review on #533, and on #523 a review told me a 5,000-byte value was materialized, so I resized a test to 1,100,000 bytes on the strength of it. On #530 I then measured a 10-byte bound `varbinary(max)` reaching `deliver_bound_plp`, which contradicted it outright.

Documentation and test comments only; no behavior change. I left the existing 1.1 MB e2e cases alone — they are correct, just larger than they need to be.

## Found while verifying, not fixed here

- **`varbinary(max)` into a typed C target answers `HYC00` here and `07006` on msodbcsql.** The existing `ABoundStreamedVarbinaryMaxToTypedCTargetIsStillUnsupported` test covers this driver's `HYC00`; implementing the missing conversion is tracked by [AB#47239](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47239).
- **`ParamArrayTest.ParamsProcessedCountsTheFailingSet` fails on the msodbcsql leg on current `main`** (ours passes; expected 4, got 5). Pre-existing and unrelated to this change — flagging it for whoever owns [AB#47944](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/47944).

### DIFF
--- a/.github/instructions/mssql-odbc.instructions.md
+++ b/.github/instructions/mssql-odbc.instructions.md
@@ -708,21 +708,10 @@ Driver Manager (DM) provides serialization guarantees that the driver relies on
   - A success-path test.
   - A null-output-handle test.
   - An invalid-handle-type or invalid-input test.
-- **A `max`/LOB column reaches the bound *streaming* path whatever its size.**
-  `try_read_row_column` attempts `try_begin_buffered_plp` for any column whose
-  metadata says `is_plp()` and yields `CursorColumn::PlpStreaming`
-  (`tds_client.rs`); the decision is gated on the column being PLP, never on its
-  length. So an e2e that targets `deliver_bound_plp` needs no particular size -
-  a handful of bytes reaches it. `ABoundVarcharMaxTruncatedToWcharReportsNoTotal`
-  is the proof already in the suite: 5,000 wire bytes asserting `SQL_NO_TOTAL`,
-  which on the bound path only `plp_indicator` produces, and that has exactly one
-  call site - inside `deliver_bound_plp`.
-  Do not confuse this with `PLP_TYPED_MATERIALIZE_LIMIT`, a separate 1 MiB cap on
-  how much a *typed* conversion will materialize (see the parity note above).
-- **Bound binary into a typed C target is refused, and the SQLSTATE differs by
-  driver for `max`.** Measured on 18.06.0001: `varbinary(8)` into `SQL_C_SLONG`
-  is `07006` on both, and `varbinary(max)` is `07006` on msodbcsql but `HYC00`
-  here. `varchar(8)`/`varchar(max)` holding `'42'` convert to `42` on both.
+- PLP routing is gated by `is_plp()` alone, not size, so an e2e targeting
+  `deliver_bound_plp` does not need a large payload. Do not confuse this with
+  `PLP_TYPED_MATERIALIZE_LIMIT`, the separate 1 MiB cap on how much a typed
+  conversion will materialize (see the parity note above).
 - Use `cargo nextest` (via `cargo btest`), not `cargo test`.
 
 ## Code style
--- a/.github/instructions/mssql-odbc.instructions.md
+++ b/.github/instructions/mssql-odbc.instructions.md
@@ -708,16 +708,21 @@ Driver Manager (DM) provides serialization guarantees that the driver relies on
   - A success-path test.
   - A null-output-handle test.
   - An invalid-handle-type or invalid-input test.
-- **A `max`/LOB column only reaches the bound *streaming* path when it is too
-  large for the transport to buffer whole.** `try_read_buffered_column` decodes
-  any fully buffered column - PLP included - into a `ColumnValues`, so a small
-  `varchar(max)`/`varbinary(max)` is delivered by `deliver_bound`, exactly like a
-  non-max value, and never enters `deliver_bound_plp`. A few kilobytes still
-  buffers; the streaming path is reachable around a megabyte (the existing 1 MiB
-  cases in `fetch_scroll_test.cpp` prove it). Size an e2e that targets
-  `deliver_bound_plp` accordingly, or it will silently assert the non-PLP path -
-  a bound binary `max` column against a typed C target answers `22018` from the
-  typed converter when buffered, and `HYC00` when streamed.
+- **A `max`/LOB column reaches the bound *streaming* path whatever its size.**
+  `try_read_row_column` attempts `try_begin_buffered_plp` for any column whose
+  metadata says `is_plp()` and yields `CursorColumn::PlpStreaming`
+  (`tds_client.rs`); the decision is gated on the column being PLP, never on its
+  length. So an e2e that targets `deliver_bound_plp` needs no particular size -
+  a handful of bytes reaches it. `ABoundVarcharMaxTruncatedToWcharReportsNoTotal`
+  is the proof already in the suite: 5,000 wire bytes asserting `SQL_NO_TOTAL`,
+  which on the bound path only `plp_indicator` produces, and that has exactly one
+  call site - inside `deliver_bound_plp`.
+  Do not confuse this with `PLP_TYPED_MATERIALIZE_LIMIT`, a separate 1 MiB cap on
+  how much a *typed* conversion will materialize (see the parity note above).
+- **Bound binary into a typed C target is refused, and the SQLSTATE differs by
+  driver for `max`.** Measured on 18.06.0001: `varbinary(8)` into `SQL_C_SLONG`
+  is `07006` on both, and `varbinary(max)` is `07006` on msodbcsql but `HYC00`
+  here. `varchar(8)`/`varchar(max)` holding `'42'` convert to `42` on both.
 - Use `cargo nextest` (via `cargo btest`), not `cargo test`.
 
 ## Code style
--- a/mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
@@ -1323,9 +1323,9 @@ TEST_F(FetchScrollLiveTest, ABoundVarbinaryMaxDeliversAcrossARowset) {
 // refused before any conversion is attempted. Same AB#47239 gap as above, so it
 // likewise asserts our own answer rather than parity.
 //
-// PLP metadata selects the streaming path regardless of the value's size. A
-// binary PLP column is refused before typed conversion, so varbinary(max) into
-// SQL_C_SLONG answers HYC00 here even for a small value.
+// PLP metadata selects the streaming path regardless of the value's size. This
+// test keeps a large value to verify that refusing a binary PLP before typed
+// conversion still drains every chunk before the trailing column and next row.
 TEST_F(FetchScrollLiveTest, ABoundStreamedVarbinaryMaxToTypedCTargetIsStillUnsupported) {
     SKIP_IF_COMPARING_MSODBCSQL();
     ExecDirect(
--- a/mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
@@ -1276,9 +1276,9 @@ TEST_F(FetchScrollLiveTest, ABoundVarbinaryMaxDeliversAcrossARowset) {
     // buffer, and the value after it -- and the row after that -- still have to
     // decode, which is what a mis-sized drain would break.
     //
-    // 1,100,000 bytes, not a few thousand: a value small enough to arrive inside
-    // the already-buffered wire bytes is materialized and delivered by the bound
-    // non-PLP path, which would leave deliver_bound_plp untested.
+    // The large value is intentional to exercise repeated PLP chunk reads and
+    // draining across packets. PLP metadata selects deliver_bound_plp regardless
+    // of the value's size.
     ExecDirect(
         "SELECT n, CAST(REPLICATE(CAST(0x41 AS VARBINARY(MAX)), 1100000) AS VARBINARY(MAX)) "
         "AS lob, n * 11 AS tail "
@@ -1323,10 +1323,9 @@ TEST_F(FetchScrollLiveTest, ABoundVarbinaryMaxDeliversAcrossARowset) {
 // refused before any conversion is attempted. Same AB#47239 gap as above, so it
 // likewise asserts our own answer rather than parity.
 //
-// The value has to exceed what the transport can buffer for a whole column,
-// otherwise `try_read_buffered_column` materializes it and the row takes the
-// ordinary non-PLP delivery path instead of the streaming one under test -- a
-// small varbinary(max) here answers 22018 from the typed converter, not HYC00.
+// PLP metadata selects the streaming path regardless of the value's size. A
+// binary PLP column is refused before typed conversion, so varbinary(max) into
+// SQL_C_SLONG answers HYC00 here even for a small value.
 TEST_F(FetchScrollLiveTest, ABoundStreamedVarbinaryMaxToTypedCTargetIsStillUnsupported) {
     SKIP_IF_COMPARING_MSODBCSQL();
     ExecDirect(
--- a/mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
+++ b/mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
@@ -1321,7 +1321,9 @@ TEST_F(FetchScrollLiveTest, ABoundVarbinaryMaxDeliversAcrossARowset) {
 // The typed-conversion path reaches the same refusal by a different route: a
 // binary PLP column has no source encoding to convert from, so it is drained and
 // refused before any conversion is attempted. Same AB#47239 gap as above, so it
-// likewise asserts our own answer rather than parity.
+// likewise asserts our own answer rather than parity. Measured on msodbcsql
+// 18.06.0001, varbinary(max) into SQL_C_SLONG answers 07006 instead of this
+// driver's HYC00; AB#47239 tracks implementing the missing conversion here.
 //
 // PLP metadata selects the streaming path regardless of the value's size. This
 // test keeps a large value to verify that refusing a binary PLP before typed


.github/instructions/mssql-odbc.instructions.md
```diff
@@ -708,16 +708,21 @@ Driver Manager (DM) provides serialization guarantees that the driver relies on
...
-  buffers; the streaming path is reachable around a megabyte (the existing 1 MiB
-  cases in `fetch_scroll_test.cpp` prove it). Size an e2e that targets
-  `deliver_bound_plp` accordingly, or it will silently assert the non-PLP path -
-  a bound binary `max` column against a typed C target answers `22018` from the
-  typed converter when buffered, and `HYC00` when streamed.
+- **A `max`/LOB column reaches the bound *streaming* path whatever its size.**
+  `try_read_row_column` attempts `try_begin_buffered_plp` for any column whose
+  metadata says `is_plp()` and yields `CursorColumn::PlpStreaming`
+  (`tds_client.rs`); the decision is gated on the column being PLP, never on its
+  length. So an e2e that targets `deliver_bound_plp` needs no particular size -
```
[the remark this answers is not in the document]
David-Engel: Updated both comments to state that PLP metadata selects `deliver_bound_plp` regardless of payload size. The first now explains that the large payload exercises repeated chunk reads and cross-packet draining, while the typed-target comment records that even a small `varbinary(max)` returns `HYC00`.

> [!NOTE]

Theekshna (commented):

First pass from this automation on PR #552 (no prior marker). Docs-and-test-comments only (`.github/instructions/mssql-odbc.instructions.md` + comment updates in `fetch_scroll_test.cpp`), no production code changed. `gh pr checks` is fully green. Copilot's one review comment (stale contradictory comments still remaining in the test file) is already addressed in the current diff -- resolved, not re-filing.

**Verified the corrected claims mechanically rather than accepting them:**
- `try_read_row_column` (`tds_client.rs`) gates its PLP branch on `pause_state.columns().get(target).is_some_and(|metadata| metadata.is_plp())` alone -- no length check anywhere in that arm -- and unconditionally calls `try_begin_buffered_plp`, whose own result (`Some(Some(stream))`) drives `CursorColumn::PlpStreaming`, which `fetch_scroll.rs`'s bound-column loop (the `else` branch with no `buffered_prefix_len`) routes straight to `deliver_bound_plp`. Confirmed the new claim ("reaches the bound streaming path whatever its size") holds for the standard `SQLBindCol`+`SQLFetch` path this doc entry is about.
  - Noted for my own understanding, not a finding: there's a separate `try_read_row_plp_complete` fast path (256-byte fixed buffer) used only for the `buffered_prefix_len` + trailing-UTF16-PLP optimization -- a different code path from the one this doc entry describes, so it doesn't contradict the correction.
- `plp_indicator` has exactly one production call site (`fetch_scroll.rs:2005`, inside `deliver_bound_plp`) -- grepped the whole crate to confirm.
- `ABoundVarcharMaxTruncatedToWcharReportsNoTotal` exists exactly as described: 5,000 bytes, asserts `SQL_NO_TOTAL`.
- `PLP_TYPED_MATERIALIZE_LIMIT` (1 MiB) is real and distinct from the reachability claim, as the new doc text says.
- Did not fully re-derive the `07006`-vs-`HYC00` msodbcsql divergence against the reference tree within this pass's budget (it requires tracing msodbcsql's PLP-to-typed-C dispatch across `sqlccurs.cpp`/`sqlcdata.cpp`, not a single lookup table) -- but the PR discloses this as an unfixed, currently-untracked gap rather than asserting it as resolved, so I'm not treating the unverified half as a finding.

No new findings.

saurabh500 (commented):
This review came from an unattended sweep; its findings were not checked by a human before posting.

**Verdict: No blocking objection to this documentation correction.**

Checked both changed files against AB#48048 and the current implementation. The bound-column path selects PLP streaming without a minimum payload-size threshold, on both buffered-header and asynchronous reads. The separate whole-row optimization excludes PLP columns, and the 256-byte trailing-PLP shortcut requires no bindings, so it does not contradict the SQLBindCol/SQLFetch guidance. The existing 5,000-byte varchar(max) test asserts SQL_NO_TOTAL, and the 1 MiB typed-materialization cap is a separate decision.

For the binary-to-typed SQLSTATE statement, I checked the msodbcsql caller as well as its validator: `odbc/sqlccurs.cpp:1432-1458` validates bound conversions before reading values; `odbc/sqlcresl.cpp:407-498` rejects binary types for SQL_C_SLONG without distinguishing max payload lengths. The Rust typed-binary and PLP paths account for the documented 07006/HYC00 difference. The cited 18.06.0001 measurements remain the author's historical observations, not new measurements from this sweep.

### Blocking
No blocking findings.

### Suggestion
No new suggestions.

### Nit
No nits.

**Checks and critique:** This diff changes prose and C++ comments only, with no new or changed test behavior to mutation-test. Existing test assertions and production routing were read, not rerun. Head CI is green; the coverage report explicitly says there are no changed lines with coverage information. The earlier stale-test-comment finding is addressed and is not being duplicated. The rubber-duck critique changed neither findings nor verdict; its additional asynchronous-path evidence was checked directly before posting.

shiwanigupta0809 (commented):
## Summary

The correction is right, and I checked it against the code rather than taking the PR's word for it. **Blocking: 0 · Suggestion: 0 · Nit: 1.**

This matters more than a typical doc change, because the text being replaced was agent instructions that would actively mis-size future e2e tests. So I verified every load-bearing claim in the replacement:

- **The gate really is `is_plp()`, not length.** `try_read_row_column` (`tds_client.rs:6541-6551`) matches on `pause_state.columns().get(target).is_some_and(|metadata| metadata.is_plp())` and nothing else; the async `read_row_column` reaches `PlpStreaming` the same way at `:7069-7072`. No size predicate on either path. The claim being removed was wrong.

- **The cited proof holds.** `plp_indicator` has exactly one non-test call site — `fetch_scroll.rs:2005`, inside `deliver_bound_plp` — so a test asserting `SQL_NO_TOTAL` on the bound path can only have gone through that function. `ABoundVarcharMaxTruncatedToWcharReportsNoTotal` exists (`fetch_scroll_test.cpp:1007`), and at 5,000 wire bytes it is a genuine small-value witness.

- **`PLP_TYPED_MATERIALIZE_LIMIT` is a separate concern, as the new text says.** It is `1024 * 1024` at `fetch_scroll.rs:1610` and only ever reached through `typed_plp_chunk_fits` at `:1741`, well after the PLP path has been selected. Keeping these two 1 MiB-adjacent facts apart is the single most useful thing this change does — conflating them is almost certainly how the original guidance went wrong.

- **The new `HYC00` claim is correct, and for a reason worth recording.** `deliver_bound_plp` refuses a typed C target at `:1731-1735` when `column_info.text_encoding` is `None`, and `PlpEncoding::Binary => None` at `:1156`. That refusal is evaluated *before* any materialization, so a binary PLP column answers `HYC00` at any size. The old text's `22018` prediction followed from the old (wrong) belief that a small value took `deliver_bound`; both halves fall together.

The surviving `07006` / `HYC00` divergence note is the kind of measured-parity fact that belongs in these instructions.

## Nit

`mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp:1328` — the comment now asserts a small-value behaviour that this test does not exercise.

The new text says `varbinary(max)` into `SQL_C_SLONG` "answers `HYC00` here even for a small value", but the query below it still builds a 1,100,000-byte value. That statement is true — I traced it through the `text_encoding: None` refusal above — it is just not what the test measures.

Normally I would not raise this. I am raising it here because it is the same failure mode this PR exists to fix: the previous guidance survived precisely because the only bound-PLP e2e cases used megabyte values, so nothing in the suite contradicted the belief that size was what selected the path. Leaving the corrected claim unwitnessed re-creates that gap for the next reader.

The companion test at `:1276` doesn't have this problem — its updated comment now says the large payload is there to exercise repeated chunk reads and cross-packet draining, which is exactly what it does.

Either shrink the value here to something like `VARBINARY(MAX)` of a few bytes, or add a second short case beside it. If the large value is deliberate for this test too, saying so in one line would close it just as well.

David-Engel:
Addressed @shiwanigupta0809's nit in 1ca2f9db. The comment no longer claims this test exercises a small payload; it now records that the 1,100,000-byte value is deliberate to verify full multi-chunk draining before the trailing column and next row.

shiwanigupta0809:
## Summary  The correction is right, and I checked it against the code rather than taking the PR's word for it. **Blocking: 0 ┬╖ Suggestion: 0 ┬╖ Nit: 1.**  This matters more than a typical doc change, because the text being replaced was agent instructions that would actively mis-size future e2e tests. So I verified every load-bearing claim in the replacement:  - **The gate really is `is_plp()`, not length.** `try_read_row_column` (`tds_client.rs:6541-6551`) matches on `pause_state.columns().get(target).is_some_and(|metadata| metadata.is_plp())` and nothing else; the async `read_row_column` reaches `PlpStreaming` the same way at `:7069-7072`. No size predicate on either path. The claim being removed was wrong.  - **The cited proof holds.** `plp_indicator` has exactly one non-test call site ΓÇö `fetch_scroll.rs:2005`, inside `deliver_bound_plp` ΓÇö so a test asserting `SQL_NO_TOTAL` on the bound path can only have gone through that function. `ABoundVarcharMaxTruncatedToWcharReportsNoTotal` exists (`fetch_scroll_test.cpp:1007`), and at 5,000 wire bytes it is a genuine small-value witness.  - **`PLP_TYPED_MATERIALIZE_LIMIT` is a separate concern, as the new text says.** It is `1024 * 1024` at `fetch_scroll.rs:1610` and only ever reached through `typed_plp_chunk_fits` at `:1741`, well after the PLP path has been selected. Keeping these two 1 MiB-adjacent facts apart is the single most useful thing this change does ΓÇö conflating them is almost certainly how the original guidance went wrong.  - **The new `HYC00` claim is correct, and for a reason worth recording.** `deliver_bound_plp` refuses a typed C target at `:1731-1735` when `column_info.text_encoding` is `None`, and `PlpEncoding::Binary => None` at `:1156`. That refusal is evaluated *before* any materialization, so a binary PLP column answers `HYC00` at any size. The old text's `22018` prediction followed from the old (wrong) belief that a small value took `deliver_bound`; both halves fall together.  The surviving `07006` / `HYC00` divergence note is the kind of measured-parity fact that belongs in these instructions.  ## Nit  `mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp:1328` ΓÇö the comment now asserts a small-value behaviour that this test does not exercise.  The new text says `varbinary(max)` into `SQL_C_SLONG` "answers `HYC00` here even for a small value", but the query below it still builds a 1,100,000-byte value. That statement is true ΓÇö I traced it through the `text_encoding: None` refusal above ΓÇö it is just not what the test measures.  Normally I would not raise this. I am raising it here because it is the same failure mode this PR exists to fix: the previous guidance survived precisely because the only bound-PLP e2e cases used megabyte values, so nothing in the suite contradicted the belief that size was what selected the path. Leaving the corrected claim unwitnessed re-creates that gap for the next reader.  The companion test at `:1276` doesn't have this problem ΓÇö its updated comment now says the large payload is there to exercise repeated chunk reads and cross-packet draining, which is exactly what it does.  Either shrink the value here to something like `VARBINARY(MAX)` of a few bytes, or add a second short case beside it. If the large value is deliberate for this test too, saying so in one line would close it just as well.

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -708,16 +708,21 @@ Driver Manager (DM) provides serialization guarantees that the driver relies on
...
+  a handful of bytes reaches it. `ABoundVarcharMaxTruncatedToWcharReportsNoTotal`
+  is the proof already in the suite: 5,000 wire bytes asserting `SQL_NO_TOTAL`,
+  which on the bound path only `plp_indicator` produces, and that has exactly one
+  call site - inside `deliver_bound_plp`.
+  Do not confuse this with `PLP_TYPED_MATERIALIZE_LIMIT`, a separate 1 MiB cap on
+  how much a *typed* conversion will materialize (see the parity note above).
+- **Bound binary into a typed C target is refused, and the SQLSTATE differs by
```
Theekshna, line 722: [Suggestion] The divergence this bullet documents (`varbinary(max)` into a typed C target: `07006` on msodbcsql, `HYC00` here) is disclosed in the PR description as an unfixed gap "worth its own item," but there's no tracking issue for it yet. Consider filing one before this thread closes, so the follow-up isn't only findable by re-reading this PR.

Theekshna (commented):

Re-reviewed at the new head after the follow-up commit that addressed Copilot's comment (stale `deliver_bound`/`22018` wording in the two `fetch_scroll_test.cpp` comments). Both are now consistent with the corrected doc text -- confirmed by re-reading the diff, not re-verifying the underlying claim again (already mechanically verified against `tds_client.rs` in the prior pass on this PR, whose findings still hold: the PLP branch is gated on `is_plp()` alone, never on length).

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 2 |
| Nit | 0 |

**[Suggestion]** The PR description's `varbinary(max)`-into-typed-C-target `HYC00`-vs-`07006` divergence (also stated in the doc addition below) is flagged as "found while verifying, not fixed here... worth its own item," but no tracking issue exists for it (checked `gh issue list --search`). Worth filing one so it doesn't fall out of view once this PR closes.

**[Suggestion]** `gh pr checks` currently shows the overall pipeline red on the mssql-python macOS cross-repo build. This PR touches no Python or macOS-specific code (docs + C++ test comments only), so this is very likely the same class of pre-existing/infrastructure flake the PR description calls out for a different test (`ParamArrayTest` on AB#47944) rather than something this diff caused -- but that's an inference, not something I re-ran to confirm (unattended run, no CI re-triggering). Worth a human glance before merging to confirm it's not new.

.github/instructions/mssql-odbc.instructions.md
```diff
@@ -708,16 +708,21 @@ Driver Manager (DM) provides serialization guarantees that the driver relies on
...
-  large for the transport to buffer whole.** `try_read_buffered_column` decodes
-  any fully buffered column - PLP included - into a `ColumnValues`, so a small
-  `varchar(max)`/`varbinary(max)` is delivered by `deliver_bound`, exactly like a
-  non-max value, and never enters `deliver_bound_plp`. A few kilobytes still
-  buffers; the streaming path is reachable around a megabyte (the existing 1 MiB
-  cases in `fetch_scroll_test.cpp` prove it). Size an e2e that targets
-  `deliver_bound_plp` accordingly, or it will silently assert the non-PLP path -
-  a bound binary `max` column against a typed C target answers `22018` from the
-  typed converter when buffered, and `HYC00` when streamed.
+- **A `max`/LOB column reaches the bound *streaming* path whatever its size.**
```
Theekshna: This bullet doesn't fit the section's pattern — every other entry here is a terse, generalizable rule (e.g. "every new `SQLXxx` function must have X tests"), while this one narrates one function's call graph and cites a specific existing test by name as "proof," which will rot if that test is renamed/removed. Could this be trimmed to the actionable rule — "PLP routing is gated by `is_plp()` alone, not size, so an e2e targeting `deliver_bound_plp` doesn't need a large payload" — with the call-graph detail and test citation moved to a code comment near `deliver_bound_plp`/`try_read_row_column`?

.github/instructions/mssql-odbc.instructions.md
Theekshna: This looks like it should live in `## Deliberate deviations` instead — it's exactly the kind of application-observable driver difference (different SQLSTATE) that section is for, but it's missing the required format: no msodbcsql source citation, no `AB#` work item, no sign-off. Per that section's own rule, either it clears the bar and belongs there properly cited, or it's function-local and should be a code comment at the `deliver_bound_plp` typed-conversion check instead — not a loose bullet under Testing.

David-Engel: Agreed that this is not a deliberate deviation: it is an unimplemented conversion tracked by AB#47239. Removed the loose Testing bullet and moved the measured SQLSTATE difference to the owning `ABoundStreamedVarbinaryMaxToTypedCTargetIsStillUnsupported` test comment.

.github/instructions/mssql-odbc.instructions.md
David-Engel: Trimmed this to the actionable rule: PLP routing is gated by `is_plp()` rather than size, so tests targeting `deliver_bound_plp` do not need a large payload. The separate typed-materialization cap warning remains because it is the likely source of the original confusion.

.github/instructions/mssql-odbc.instructions.md
David-Engel: This gap is already tracked in Azure Boards as AB#47239, which the owning test comment now cites directly. I moved the measured msodbcsql 18.06.0001 SQLSTATE difference beside that test so the tracking context stays with the unimplemented conversion.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
```diff
@@ -1321,12 +1321,13 @@ TEST_F(FetchScrollLiveTest, ABoundVarbinaryMaxDeliversAcrossARowset) {
 // The typed-conversion path reaches the same refusal by a different route: a
 // binary PLP column has no source encoding to convert from, so it is drained and
 // refused before any conversion is attempted. Same AB#47239 gap as above, so it
-// likewise asserts our own answer rather than parity.
+// likewise asserts our own answer rather than parity. Measured on msodbcsql
+// 18.06.0001, varbinary(max) into SQL_C_SLONG answers 07006 instead of this
+// driver's HYC00; AB#47239 tracks implementing the missing conversion here.
```
Theekshna, line 1326: [Nit] The PR description's Found while verifying, not fixed here section says this gap has 
o test and is worth its own item -- both stale now that this comment cites AB#47239 and the pre-existing test already asserts HYC00. Worth a one-line edit to the description.

Theekshna (commented):

Re-reviewed at the new head (`dbe1c0e7`) after the follow-up commits that answered both prior suggestions and both inline nits from my last pass. Verified the author's fixes rather than trusting the commit messages:

- **AB#47239 now tracks the `07006`/`HYC00` divergence** (my prior Suggestion) -- cited directly on `ABoundStreamedVarbinaryMaxToTypedCTargetIsStillUnsupported` in `fetch_scroll_test.cpp:1326`, which already asserts `HYC00` and is skipped under `SKIP_IF_COMPARING_MSODBCSQL()`. Re-derived the msodbcsql side independently rather than taking the PR's table on faith: `IsValidColumnConversion` (`sqlcresl.cpp:466-472`, reference tree) returns `IDS_07_006` for any `SQLVARBINARY`/`SQLBIGVARBINARY`/`SQLIMAGE` source into the `SQL_C_SLONG` group, and `ProcessResults`'s per-column loop (`sqlccurs.cpp:1432-1453`) calls it before any data is pumped -- so the `07006` answer holds regardless of MAX-ness, confirming the table's `varbinary(max)` row.
- **CI (my other prior Suggestion)** -- `gh pr checks` is now fully green, including the previously-red mssql-python macOS cross-repo build.
- **Both inline nits from my second pass** (bullet not fitting the Testing section's terse-rule pattern; belonged in `## Deliberate deviations` or nowhere) -- author agreed it isn't a deliberate deviation (it's an unimplemented conversion), deleted the loose bullet, and moved the call-graph proof + measured SQLSTATE table into the owning test's comment. Confirmed by reading the current diff: `.github/instructions/mssql-odbc.instructions.md` now carries only the terse, generalizable `is_plp()` rule.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 1 |

**[Nit]** The PR description's "Found while verifying, not fixed here" section still says the `HYC00`/`07006` gap has "no test" and is "worth its own item" -- both are now stale: the test already existed pre-PR and now cites `AB#47239` directly. Worth a one-line edit so a reader doesn't go looking for a tracking item that already exists.

mssql-odbc/tests/e2e/tests/fetch_scroll_test.cpp
David-Engel: Updated the PR description to name the existing `ABoundStreamedVarbinaryMaxToTypedCTargetIsStillUnsupported` coverage and link AB#47239 as the tracking item. I also corrected the scope summary to “documentation and test comments only; no behavior change.”

Theekshna (approved):

PR #558, opened 2026-09-11
Add focused Miri buffer-safety tests for mssql-odbc

saurabh500:
## Description

Add 19 buffer-safety unit tests and extend the two UTF-16 regressions now on `main`. The tests cover the ODBC driver's string helpers, parameter readers, typed fetch converters, and parameter binding layouts using the production implementations directly, without Miri-specific stubs or disabled safety checks.

The shared UTF-16 reader alignment fix has landed on `main`. This branch preserves that implementation and consolidates the overlapping tests using the existing aligned-buffer fixture. Both the original leading/trailing isolated-surrogate case and the expanded alignment/length matrix remain covered. No public API changes.

- Check UTF-16 input at both alignments, including empty strings, odd byte counts, surrogate pairs, isolated surrogates, and unread uninitialized tails.
- Vary parameter value, indicator, and octet-length alignment independently across all eight combinations, including numeric and all five temporal input layouts.
- Exercise zero-length and SQL NULL input, binary BufferLength fallback, and staging/default sentinels. Leave slots ignored by the tested helper uninitialized; verify initialized input fields and metadata remain unchanged without inspecting struct padding.
- Check string capacities and terminators, typed output widths, optional null output pointers, and untouched outputs on conversion errors.
- Exercise production parameter-array address calculations with signed nonzero binding offsets, column-wise and packed row-wise layouts, independent displacements, distinct first/last values, and uninitialized gaps. Expected positions come from typed array/struct-field addresses, not the driver's stride helper.
- Assert fixture alignment and packed-stride properties, and reuse buffers across calls.

The SQL NULL case with an uninitialized octet-length slot exercises only `read_indicator`'s early return, not full execution. Execution's earlier data-at-execution probes still require readable, initialized non-null octet-length slots for input and input/output bindings; these helper-level tests do not relax that binding contract.

Add a focused `miri-odbc` nextest profile selecting 24 tests: the 19 new tests, two extended UTF-16 regressions, and three existing parameter-misalignment tests. Document pinned-nightly setup, dependency restoration, native execution, and the short build-directory setting needed for long Windows checkout paths.

### PR pipeline integration

Run the profile under Miri in **Windows x64 and Linux x64 PR validation only**. Both steps install `nightly-2026-09-06` with Miri/rust-src and run seed 0 against explicit x64 targets; Linux reuses the existing Ubuntu 22.04 build image. There are no Miri steps on ARM64, macOS, or Alpine. The native macOS ODBC validation newly added on `main` is retained unchanged.

Miri failures fail the job, retries remain disabled, and `--no-tests=fail` rejects an empty selection. Nextest emits a separate Miri JUnit report on each platform, which is published even after a failed run. Missing reports also fail publication. The normal native test suite still includes these tests.

All Miri invocations scope the driver through `--package mssqlodbc`. The shared profile contains only test-name predicates so it can also load in the reduced Kerberos workspace, which omits ODBC; an eager package predicate there prevented unrelated tests from starting. This repair preserves the 24-test Miri selection and does not skip or change Kerberos tests.

### Validation

- On the latest main merge, workspace formatting and all-feature/all-target clippy pass, as do formatting and clippy for the separately excluded `mssql-py-core` crate.
- After consolidating the overlapping UTF-16 tests, all 205 selected native string/scanner/buffer tests and all 24 Miri tests pass; Miri was run on Windows x64 with seeds 0 and 1.
- All 11 pipeline-configuration regression tests pass, covering both the x64-only Miri steps and main's native macOS ODBC step.
- Earlier platform validation executed the exact Windows pipeline step locally and the Linux payload in the actual CI build image (cargo-nextest 0.9.99): all 24 tests passed on each platform with seed 0 and valid JUnit reports.
- Both Miri execution and result-publication steps succeeded in hosted Azure build 175199. The reduced-workspace filter problem that build exposed was fixed in b23f3201, and required checks passed on that head before this main merge. Hosted checks will validate the new merge head again.
- Reduced/full fixture workspaces load the repaired shared config repeatedly on Windows and CI nextest 0.9.99 on Linux; explicit package scoping excludes other packages with identical test names. Empty selections still fail with exit 4.
- Simulated Windows native-command failures at installation, sysroot setup, and test execution each stop the step immediately.
- Before the UTF-16 fix, the strengthened regressions failed under Miri on both aligned dereference and unaligned slice construction. Test-sensitivity checks also detected an extra uninitialized byte, a misplaced terminator, and broken alignment/packed-stride fixtures. The filter probe fails if its package scoping is removed. All temporary mutations were restored.

The Miri profile intentionally excludes handle fixtures that start Tokio, socket-based mock servers, native TLS/authentication, and Driver Manager/DLL loading. Full `cargo btest`, authenticated Kerberos, live SQL Server, and C++ end-to-end suites were not run locally.

## Related Issues

[AB#45765](https://sqlclientdrivers.visualstudio.com/b95cf060-8083-439d-8ef1-405d5bf219d8/_workitems/edit/45765)

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes
- [x] Public API changes are documented (not applicable: no public API changes)

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -8,6 +8,12 @@ slow-timeout = { period = "60s", terminate-after = 3 }
 [profile.default]
 slow-timeout = { period = "60s", terminate-after = 3 }
 
+# Opt-in buffer tests that do not start Tokio or call native libraries.
+[profile.miri-odbc]
+default-filter = 'package(mssqlodbc) & (test(::memory_safety::) | test(conversion::param_buffer::tests::misaligned_))'
+retries = 0
+slow-timeout = { period = "120s", terminate-after = 3 }
+
 # Test groups for tests that need to run sequentially
 # TLS tests have race conditions with macOS Security.framework when creating
 # PKCS12 identities in parallel, so they must run one at a time.
--- a/mssql-odbc/README.md
+++ b/mssql-odbc/README.md
@@ -47,6 +47,50 @@ The `build.rs` script embeds platform-specific metadata:
 cargo btest -p mssqlodbc
 ```
 
+### Buffer safety with Miri
+
+The `miri-odbc` nextest profile selects the opt-in `memory_safety` unit-test
+modules and the parameter reader's existing misalignment tests. They cover
+unaligned values and indicators, initialized read extents, string terminators
+and capacities, fixed-width writes, untouched error outputs, and reuse of caller
+buffers. They also run as ordinary unit tests; no production code is replaced
+under Miri.
+
+From the repository root, with `cargo-nextest` installed:
+
+```powershell
+cargo fetch
+rustup toolchain install nightly-2026-09-06 --profile minimal --component miri,rust-src
+cargo +nightly-2026-09-06 miri nextest run --frozen -p mssqlodbc --lib --profile miri-odbc
+```
+
+`cargo fetch` creates the local, gitignored lockfile and restores dependencies
+before the frozen run. On Windows, a long checkout path can make Cargo use a
+compiler response file, which this Miri version does not support. In that case,
+set a short, dedicated build directory before running Miri:
+
+```powershell
+$env:CARGO_TARGET_DIR = Join-Path $env:TEMP "odbc-miri"
+```
+
+Run the same selection natively with:
+
+```powershell
+cargo nextest run --frozen -p mssqlodbc --lib --profile miri-odbc
+```
+
+The tests keep unread input tails uninitialized so Miri can detect over-reads
+that stay inside an allocation. Output sentinels additionally catch writes
+outside the declared slot even when those writes remain inside the backing
+allocation. Deliberate misalignment is asserted before the call, and C struct
+padding is not assumed to be initialized.
+
+This profile intentionally excludes handle fixtures (which start an I/O-enabled
+Tokio runtime), socket-based mock servers, native authentication/TLS, and the
+C++ Driver Manager tests. It does not test Windows DLL unloading or replace
+native end-to-end tests or fuzzing. Keep Miri's alignment and aliasing checks
+enabled; a passing run covers only the inputs and executions exercised.
+
 ### C++ e2e tests (Google Test)
 
 End-to-end tests that exercise the driver through the ODBC Driver Manager,
--- a/mssql-odbc/src/api/util.rs
+++ b/mssql-odbc/src/api/util.rs
@@ -373,6 +373,118 @@ mod tests {
     };
     use crate::api::odbc_types::{SQL_NTS, SqlInteger, SqlWChar};
 
+    mod memory_safety {
+        use super::*;
+        use crate::api::odbc_types::SqlLen;
+        use crate::test_support::AlignedBuffer;
+        use std::mem::MaybeUninit;
+
+        #[test]
+        fn wide_copies_respect_capacity_at_a_byte_offset() {
+            for text in ["", "a", "a\u{1f600}b"] {
+                let source: Vec<u16> = text.encode_utf16().collect();
+                for direct_encoding in [false, true] {
+                    for capacity in 0..=source.len() + 2 {
+                        let mut storage = AlignedBuffer([0xA5u8; 17]);
+                        assert!(capacity * size_of::<u16>() < storage.0.len());
+                        for _ in 0..2 {
+                            let ptr = storage.0.as_mut_ptr().wrapping_add(1).cast::<u16>();
+                            assert!(!ptr.is_aligned());
+                            // The displaced destination has room for every declared unit.
+                            let truncated = unsafe {
+                                if direct_encoding {
+                                    copy_utf16_with_nul(ptr, capacity, text)
+                                } else {
+                                    copy_with_nul(ptr, capacity, &source)
+                                }
+                            };
+                            assert_eq!(truncated, source.len() > capacity.saturating_sub(1));
+                            let mut expected = [0xA5; 17];
+                            if capacity != 0 {
+                                for (i, unit) in source
+                                    .iter()
+                                    .copied()
+                                    .take(capacity - 1)
+                                    .chain(std::iter::once(0))
+                                    .enumerate()
+                                {
+                                    expected[1 + i * 2..1 + (i + 1) * 2]
+                                        .copy_from_slice(&unit.to_ne_bytes());
+                                }
+                            }
+                            assert_eq!(storage.0, expected);
+                        }
+                    }
+                }
+            }
+        }
+
+        #[test]
+        fn wide_copies_initialize_only_the_written_prefix() {
+            let text = "a\u{1f600}b";
+            let source: Vec<u16> = text.encode_utf16().collect();
+            for direct_encoding in [false, true] {
+                for capacity in [0usize, 1, 2, 5] {
+                    let mut storage = [MaybeUninit::<u16>::uninit(); 5];
+                    let ptr = storage.as_mut_ptr().cast::<u16>();
+                    // The output is writable but has no initialized value to read.
+                    let truncated = unsafe {
+                        if direct_encoding {
+                            copy_utf16_with_nul(ptr, capacity, text)
+                        } else {
+                            copy_with_nul(ptr, capacity, &source)
+                        }
+                    };
+                    assert_eq!(truncated, source.len() > capacity.saturating_sub(1));
+                    if capacity != 0 {
+                        let expected = source
+                            .iter()
+                            .copied()
+                            .take(capacity - 1)
+                            .chain(std::iter::once(0));
+                        for (actual, expected) in storage.iter().zip(expected) {
+                            // Only the copied prefix and its terminator are initialized.
+                            assert_eq!(unsafe { actual.assume_init() }, expected);
+                        }
+                    }
+                }
+            }
+        }
+
+        #[test]
+        fn scalar_out_parameter_writes_stay_inside_the_unaligned_slot() {
+            let mut storage = AlignedBuffer([0xA5u8; 24]);
+            for value in [SqlLen::MIN, 42, SqlLen::MAX] {
+                let ptr = storage.0.as_mut_ptr().wrapping_add(1).cast::<SqlLen>();
+                assert!(!ptr.is_aligned());
+                // The slot is writable for one SqlLen, independently of its alignment.
+                unsafe { write_if_some(ptr, value) };
+                let mut expected = [0xA5; 24];
+                expected[1..1 + size_of::<SqlLen>()].copy_from_slice(&value.to_ne_bytes());
+                assert_eq!(storage.0, expected);
+            }
+        }
+
+        #[test]
+        fn utf16_reads_stop_at_the_initialized_extent() {
+            let mut storage = [MaybeUninit::<u16>::uninit(); 4];
+            storage[0].write(u16::from(b'a'));
+            storage[1].write(0);
+            let ptr = storage.as_ptr().cast::<u16>();
+            for _ in 0..2 {
+                // Two initialized units, including the terminator; the tail is unreadable.
+                unsafe {
+                    assert_eq!(read_utf16(ptr, SQL_NTS), "a");
+                    assert_eq!(read_utf16_long(ptr, SqlInteger::from(SQL_NTS)), "a");
+                    assert_eq!(read_utf16_attr(ptr, SqlInteger::from(SQL_NTS)), "a");
+                    assert_eq!(read_utf16(ptr, 2), "a\0");
+                    assert_eq!(read_utf16_long(ptr, 2), "a\0");
+                    assert_eq!(read_utf16_attr(ptr, 4), "a\0");
+                }
+            }
+        }
+    }
+
     #[test]
     fn rewrite_no_markers_is_unchanged() {
         let (out, n) = rewrite_param_markers("SELECT 1");
--- a/mssql-odbc/src/conversion/param_buffer.rs
+++ b/mssql-odbc/src/conversion/param_buffer.rs
@@ -447,6 +447,181 @@ mod tests {
     use crate::api::odbc_types::{SQL_C_DEFAULT, SQL_NO_TOTAL, SQL_PARAM_INPUT, SQL_TINYINT};
     use std::ffi::c_void;
 
+    mod memory_safety {
+        use super::*;
+        use crate::test_support::AlignedBuffer;
+        use std::mem::MaybeUninit;
+
+        fn check_fixed<T: Copy>(c_type: SqlSmallInt, value: T, expected: AppValue) {
+            let mut storage = AlignedBuffer([MaybeUninit::<u8>::uninit(); 32]);
+            assert!(size_of::<T>() < storage.0.len());
+            let ptr = storage.0.as_mut_ptr().wrapping_add(1).cast::<T>();
+            if align_of::<T>() > 1 {
+                assert!(!ptr.is_aligned());
+            }
+            // Initialize only the C type's storage, leaving any over-read visible to Miri.
+            unsafe { ptr.write_unaligned(value) };
+            let mut ind = 0;
+            let p = param(c_type, ptr.cast(), &mut ind);
+            for _ in 0..2 {
+                assert_eq!(read(&p).unwrap().as_ref(), Some(&expected));
+            }
+        }
+
+        #[test]
+        fn fixed_width_reads_use_only_the_initialized_unaligned_value() {
+            check_fixed(SQL_C_STINYINT, -7i8, AppValue::Integer(-7));
+            check_fixed(SQL_C_UTINYINT, u8::MAX, AppValue::Integer(255));
+            check_fixed(
+                SQL_C_SSHORT,
+                i16::MIN,
+                AppValue::Integer(i128::from(i16::MIN)),
+            );
+            check_fixed(
+                SQL_C_USHORT,
+                u16::MAX,
+                AppValue::Integer(i128::from(u16::MAX)),
+            );
+            check_fixed(
+                SQL_C_SLONG,
+                i32::MIN,
+                AppValue::Integer(i128::from(i32::MIN)),
+            );
+            check_fixed(
+                SQL_C_ULONG,
+                u32::MAX,
+                AppValue::Integer(i128::from(u32::MAX)),
+            );
+            check_fixed(
+                SQL_C_SBIGINT,
+                i64::MIN,
+                AppValue::Integer(i128::from(i64::MIN)),
+            );
+            check_fixed(
+                SQL_C_UBIGINT,
+                u64::MAX,
+                AppValue::Integer(i128::from(u64::MAX)),
+            );
+            check_fixed(SQL_C_BIT, 0xFFu8, AppValue::Bit(true));
+            check_fixed(SQL_C_FLOAT, 1.5f32, AppValue::Float(1.5));
+            check_fixed(SQL_C_DOUBLE, -2.5f64, AppValue::Double(-2.5));
+            let guid = SqlGuid {
+                data1: 0x12345678,
+                data2: 0x9ABC,
+                data3: 0xDEF0,
+                data4: [1, 2, 3, 4, 5, 6, 7, 8],
+            };
+            check_fixed(SQL_C_GUID, guid, AppValue::Guid(guid));
+            let numeric = SqlNumericStruct {
+                precision: 38,
+                scale: 2,
+                sign: 1,
+                val: [0x12; 16],
+            };
+            check_fixed(SQL_C_NUMERIC, numeric, AppValue::Numeric(numeric));
+        }
+
+        #[test]
+        fn explicit_byte_lengths_preserve_nuls_without_reading_the_tail() {
+            let mut storage = [MaybeUninit::<u8>::uninit(); 8];
+            for (slot, value) in storage.iter_mut().zip(*b"a\0b") {
+                slot.write(value);
+            }
+            for c_type in [SQL_C_CHAR, SQL_C_BINARY] {
+                let mut ind = 3;
+                let p = param(c_type, storage.as_mut_ptr().cast(), &mut ind);
+                let expected = if c_type == SQL_C_CHAR {
+                    AppValue::NarrowText(b"a\0b".to_vec())
+                } else {
+                    AppValue::Binary(b"a\0b".to_vec())
+                };
+                assert_eq!(read(&p).unwrap(), Some(expected));
+            }
+        }
+
+        #[test]
+        fn explicit_wide_lengths_read_only_complete_unaligned_units() {
+            let mut storage = AlignedBuffer([MaybeUninit::<u8>::uninit(); 16]);
+            let ptr = storage.0.as_mut_ptr().wrapping_add(1).cast::<u16>();
+            assert!(!ptr.is_aligned());
+            for (i, value) in [u16::from(b'a'), 0, u16::from(b'b')]
+                .into_iter()
+                .enumerate()
+            {
+                // Only these three complete units are initialized.
+                unsafe { ptr.wrapping_add(i).write_unaligned(value) };
+            }
+            for (length, expected) in [(6, &b"a\0\0\0b\0"[..]), (5, &b"a\0\0\0"[..])] {
+                let mut ind = length;
+                let p = param(SQL_C_WCHAR, ptr.cast(), &mut ind);
+                assert_eq!(
+                    read(&p).unwrap(),
+                    Some(AppValue::WideText(expected.to_vec()))
+                );
+            }
+        }
+
+        #[test]
+        fn nts_reads_stop_before_uninitialized_storage() {
+            let mut narrow = [MaybeUninit::<u8>::uninit(); 8];
+            narrow[0].write(b'a');
+            narrow[1].write(0);
+            let mut ind = SqlLen::from(SQL_NTS);
+            let p = param(SQL_C_CHAR, narrow.as_mut_ptr().cast(), &mut ind);
+            assert_eq!(read(&p).unwrap(), Some(AppValue::NarrowText(b"a".to_vec())));
+
+            let mut wide = AlignedBuffer([MaybeUninit::<u8>::uninit(); 16]);
+            let ptr = wide.0.as_mut_ptr().wrapping_add(1).cast::<u16>();
+            assert!(!ptr.is_aligned());
+            // The terminator is the last initialized unit in the caller's string.
+            unsafe {
+                ptr.write_unaligned(u16::from(b'a'));
+                ptr.wrapping_add(1).write_unaligned(0);
+            }
+            let p = param(SQL_C_WCHAR, ptr.cast(), &mut ind);
+            assert_eq!(read(&p).unwrap(), Some(AppValue::WideText(b"a\0".to_vec())));
+        }
+
+        #[test]
+        fn null_parameters_do_not_read_the_value_buffer() {
+            let mut value = MaybeUninit::<SqlTimestampStruct>::uninit();
+            for c_type in [SQL_C_SLONG, SQL_C_DOUBLE, SQL_C_TYPE_TIMESTAMP, SQL_C_WCHAR] {
+                let mut ind = SQL_NULL_DATA;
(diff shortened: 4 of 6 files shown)


mssql-odbc/src/api/util.rs
```diff
@@ -373,6 +373,118 @@ mod tests {
...
+            let ptr = storage.as_ptr().cast::<u16>();
+            for _ in 0..2 {
+                // Two initialized units, including the terminator; the tail is unreadable.
+                unsafe {
+                    assert_eq!(read_utf16(ptr, SQL_NTS), "a");
+                    assert_eq!(read_utf16_long(ptr, SqlInteger::from(SQL_NTS)), "a");
+                    assert_eq!(read_utf16_attr(ptr, SqlInteger::from(SQL_NTS)), "a");
+                    assert_eq!(read_utf16(ptr, 2), "a\0");
+                    assert_eq!(read_utf16_long(ptr, 2), "a\0");
+                    assert_eq!(read_utf16_attr(ptr, 4), "a\0");
```
saurabh500: Following up on the existing alignment comment: please carry over the unaligned-input regression cases from #560 (`a00b78d0`) when integrating these tests. That PR reproduced the macOS x64 abort in `read_utf16_long`, while native ARM64 passed the same C++ test.

For `read_utf16`, `read_utf16_long`, and `read_utf16_attr`, exercise byte-offset input with asserted misalignment under both `SQL_NTS` and explicit lengths. Include empty input, embedded NULs, odd byte counts that round down to whole UTF-16 units, surrogate pairs, and isolated surrogates with replacement-character decoding. Keep unread tails uninitialized to check read bounds as well as alignment.

The two new `api::util::tests::memory_safety` tests in #560 pass under Miri on `nightly-2026-09-06` with seeds 0 and 1. Their names fit this PR's filter; please preserve that selection and the existing initialized-extent coverage when reconciling the changes.

mssql-odbc/src/conversion/param_buffer.rs
```diff
@@ -447,6 +447,181 @@ mod tests {
...
+            let ptr = storage.0.as_mut_ptr().wrapping_add(1).cast::<T>();
+            if align_of::<T>() > 1 {
+                assert!(!ptr.is_aligned());
+            }
+            // Initialize only the C type's storage, leaving any over-read visible to Miri.
+            unsafe { ptr.write_unaligned(value) };
+            let mut ind = 0;
+            let p = param(c_type, ptr.cast(), &mut ind);
+            for _ in 0..2 {
+                assert_eq!(read(&p).unwrap().as_ref(), Some(&expected));
```
saurabh500: Could we expand this helper into an input-contract matrix rather than always using offset 1 with an aligned indicator? Keep the exact initialized value and uninitialized tail, but vary aligned/misaligned value addresses independently of aligned/misaligned indicator addresses. Add valid date/time/timestamp inputs alongside the numeric cases, and check that initialized input bytes remain unchanged after reads without inspecting struct padding.

The #560 failure shows why combining dimensions matters: testing unaligned outputs and bounded inputs separately did not exercise an unaligned UTF-16 input. This is a coverage suggestion, not a claim that these parameter readers have that bug.

mssql-odbc/src/conversion/param_buffer.rs
```diff
@@ -447,6 +447,181 @@ mod tests {
...
+                read(&p).unwrap(),
+                Some(AppValue::NarrowText(b"a\0b".to_vec()))
+            );
+            unsafe { indicator.write_unaligned(SQL_NULL_DATA) };
+            assert_eq!(read(&p).unwrap(), None);
+            unsafe {
+                indicator.write_unaligned(0);
+                length.write_unaligned(1);
+            }
+            assert_eq!(read(&p).unwrap(), Some(AppValue::NarrowText(b"a".to_vec())));
```
saurabh500: This already covers separate unaligned indicator/length slots and changing their values. Please extend the combinations to zero-length input, SQL NULL with an uninitialized value buffer, and supported length sentinels for each parameter type. Where the contract says a slot is ignored, leave it uninitialized so Miri detects an accidental read; initialize it for cases that require it.

Also vary the value pointer's alignment independently of the two metadata slots. Keep all required storage valid and live so a Miri failure identifies a reader regression rather than an invalid fixture.

.config/nextest.toml
```diff
@@ -8,6 +8,12 @@ slow-timeout = { period = "60s", terminate-after = 3 }
 [profile.default]
 slow-timeout = { period = "60s", terminate-after = 3 }
 
+# Opt-in buffer tests that do not start Tokio or call native libraries.
+[profile.miri-odbc]
+default-filter = 'package(mssqlodbc) & (test(::memory_safety::) | test(conversion::param_buffer::tests::misaligned_))'
```
saurabh500: As this profile grows, please include pure parameter-array/row-wise binding tests combining nonzero binding offsets, packed row strides, and independently displaced value/indicator storage. Exercise first and last elements with distinct values and uninitialized gaps, using production address-calculation helpers rather than calculating expected pointers with the same implementation.

Keep those cases in `memory_safety` modules or explicitly include them here, and verify the selected test list. The Driver Manager/native-library exclusion makes sense; cover these buffer-layout contracts below that boundary where possible. This is a suggested next coverage area, not a request to run the C++ suite under Miri.

saurabh500 (commented):
Additional input-coverage suggestions following the macOS x64 alignment failure fixed in #560. These are requests to broaden the Miri input matrix, not claims that each suggested path contains a bug.

mssql-odbc/README.md
```diff
@@ -47,6 +47,50 @@ The `build.rs` script embeds platform-specific metadata:
 cargo btest -p mssqlodbc
 ```
 
+### Buffer safety with Miri
```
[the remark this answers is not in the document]
saurabh500: The author explicitly requested leaving Related Issues blank for this draft and will supply the tracking link later. I have kept the PR in draft and left the required linked-issue check intact; the link is still needed before this moves forward.

mssql-odbc/src/api/util.rs
```diff
@@ -373,6 +373,118 @@ mod tests {
...
+                assert_eq!(storage.0, expected);
+            }
+        }
+
+        #[test]
+        fn utf16_reads_stop_at_the_initialized_extent() {
+            let mut storage = [MaybeUninit::<u16>::uninit(); 4];
+            storage[0].write(u16::from(b'a'));
+            storage[1].write(0);
+            let ptr = storage.as_ptr().cast::<u16>();
```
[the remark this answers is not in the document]
saurabh500: Fixed in 6965950e: the shared reader uses unaligned unit reads for both NUL-terminated and explicit-length input, with direct lossy decoding and no intermediate buffer. Both strengthened regressions failed under Miri before the fix; the aligned/misaligned cases now pass with both seeds, and the fixture itself fails if its odd offset is removed.

mssql-odbc/src/conversion/param_buffer.rs
saurabh500: Expanded check_fixed in 1b4596dc to all eight independent value/indicator/length alignment combinations, including valid inputs for all five temporal layouts and repeated reads. The tests check that input fields and metadata remain unchanged using typed comparisons that do not inspect struct padding.

mssql-odbc/src/api/util.rs
saurabh500: Carried over both regression cases from #560 in 1b4596dc, using the existing AlignedBuffer and retaining the initialized-extent test. They cover both alignments, explicit/NTS lengths, empty input, embedded NULs, odd byte counts, and surrogate decoding with uninitialized tails, and both names appear in the verified 24-test Miri selection.

.config/nextest.toml
saurabh500: Added pure column-array and packed row-wise parameter tests in 1b4596dc that call BoundParam::for_row and the actual readers, with nonzero offsets, independent displacements, distinct first/last values, and uninitialized gaps. Expected positions use typed array/struct-field addresses independently of the driver calculations, and both tests are included in the verified profile selection and pass under Miri with seeds 0 and 1.

mssql-odbc/src/conversion/param_buffer.rs
saurabh500: Added the combinations in 1b4596dc: zero-length uninitialized values, null/zero input, explicit/NTS lengths including binary BufferLength fallback, SQL NULL with uninitialized value and length storage, and staging/default sentinel errors before value reads. Each uses independent value/indicator/length displacements; the read_indicator safety documentation now states when the length slot is ignored.

.pipeline/templates/validation-stages.yml
```diff
@@ -135,6 +135,28 @@ stages:
...
+          if ($LASTEXITCODE -ne 0) { throw "Miri toolchain installation failed (exit $LASTEXITCODE)" }
+          cargo +nightly-2026-09-06 miri setup --target x86_64-pc-windows-msvc
+          if ($LASTEXITCODE -ne 0) { throw "Miri sysroot setup failed (exit $LASTEXITCODE)" }
+          cargo +nightly-2026-09-06 miri nextest run --frozen --package mssqlodbc --lib `
+            --target x86_64-pc-windows-msvc --profile miri-odbc --no-fail-fast --no-tests=fail
+          if ($LASTEXITCODE -ne 0) { throw "ODBC Miri tests failed (exit $LASTEXITCODE)" }
+        displayName: Run ODBC Miri tests (Windows x64)
+        condition: and(succeeded(), eq(variables['Build.Reason'], 'PullRequest'))
+        env:
+          MIRIFLAGS: -Zmiri-seed=0
```
saurabh500, line 152: What is this for ?

saurabh500: `MIRIFLAGS` passes options to Miri; `-Zmiri-seed=0` sets the random seed used for simulated allocation addresses and scheduling choices, making an execution reproducible for the same code, toolchain, and target. Zero is already [Miri’s default](https://github.com/rust-lang/miri#miri--z-flags-and-environment-variables)—this makes the common CI baseline explicit in both x64 jobs, without changing test selection or disabling any safety checks.

.pipeline/templates/validation-stages.yml
```diff
@@ -135,6 +135,28 @@ stages:
...
+      - task: PublishTestResults@2
+        displayName: Publish ODBC Miri test results (Windows x64)
+        condition: and(succeededOrFailed(), eq(variables['Build.Reason'], 'PullRequest'))
+        inputs:
+          testResultsFormat: JUnit
+          testResultsFiles: $(Build.SourcesDirectory)/target/nextest/miri-odbc/junit.xml
```
David-Engel, line 158: `Build_Windows` sets `CARGO_TARGET_DIR: C:\cargo_target_dir` (line 115), and nextest writes its JUnit report to `<target-dir>/nextest/<profile>/junit.xml`, so on that agent the Miri report should land under `C:\cargo_target_dir\...`, not under `$(Build.SourcesDirectory)/target/...`. The Linux step is fine for the opposite reason: the run happens inside the container at `/workspace`, and `docker-cargo-run.sh` does not forward the host `CARGO_TARGET_DIR`, so `/workspace/target/...` maps back to the sources tree.

Worth confirming from the passing run's log which file this task actually picked up. If the path does not resolve, one of two things is true: `failTaskOnMissingResultsFile` is honored and this fails the Windows job the first time it matters, or the installed task version ignores that input and the Windows Miri results are silently never published — which would make the "missing reports also fail publication" claim in the description untrue on Windows.

The repo has already been bitten by this: `build-template.yml:215` forces an explicit `-CoverageOutput` under `$(Build.SourcesDirectory)/target` with the comment that it keeps the publish path deterministic "regardless of the job's CARGO_TARGET_DIR". Deriving this path from `$(CARGO_TARGET_DIR)`, or copying the report to a fixed location after the run, would remove the ambiguity for both platforms.

.config/nextest.toml
```diff
@@ -8,6 +8,16 @@ slow-timeout = { period = "60s", terminate-after = 3 }
 [profile.default]
 slow-timeout = { period = "60s", terminate-after = 3 }
 
+# Opt-in buffer tests that do not start Tokio or call native libraries.
+# Scope via --package mssqlodbc; package predicates fail in workspaces without ODBC.
+[profile.miri-odbc]
+default-filter = 'test(::memory_safety::) | test(conversion::param_buffer::tests::misaligned_)'
```
David-Engel, line 14: The filter is name-only, so the profile is correctly scoped only while every caller remembers `--package mssqlodbc`. `cargo nextest run --profile miri-odbc` from the repo root picks up any `::memory_safety::` test in any workspace crate, which is a trap for someone following the README's native-run line with a different `-p`.

Since the reason for dropping the package predicate was that `package(mssqlodbc)` fails to parse in the reduced Kerberos workspace, `binary_id(mssqlodbc)` may be the middle ground: it is a name matcher rather than a workspace lookup, so it should still parse where the package is absent while keeping the profile self-scoping. Worth verifying against the reduced-workspace fixture you already built before adopting it.

.pipeline/templates/validation-stages.yml
```diff
@@ -135,6 +135,28 @@ stages:
           buildType: ${{ parameters.buildType }}
           testFilter: -E "not (test(connectivity))"
           enableOdbcE2E: true
+      - pwsh: |
+          $ErrorActionPreference = 'Stop'
+          rustup toolchain install nightly-2026-09-06 --profile minimal --component miri,rust-src
```
David-Engel: Nit: `nightly-2026-09-06` is pinned in four places — this step, the Linux step, `mssql-odbc/README.md`, and the assertion in `scripts/test_validation_pipeline.py`. A bump has to touch all four, and the first signal will be a config test failing on the YAML rather than anything pointing at the pin. A pipeline variable (say `miriToolchain`) referenced by both steps would cut that to two, with the config test asserting the variable instead of the literal.

mssql-odbc/src/conversion/param_buffer.rs
```diff
@@ -155,8 +155,9 @@ pub(crate) enum Indicator {
 /// buffers is now read correctly instead of one of the two being discarded.
 ///
 /// # Safety
-/// `param.strlen_or_ind_ptr` and `param.octet_length_ptr`, if non-null, must
-/// each point to one valid `SqlLen`.
+/// `param.strlen_or_ind_ptr`, if non-null, must point to one valid `SqlLen`.
+/// Unless it specifies `SQL_NULL_DATA`, `param.octet_length_ptr`, if non-null,
```
David-Engel: Nit: "Unless it specifies `SQL_NULL_DATA`" reads as though `it` were `octet_length_ptr`, which inverts the actual rule — `read_indicator` returns early when `*strlen_or_ind_ptr == SQL_NULL_DATA`, and only then is the octet-length slot never read. Something like "unless `*param.strlen_or_ind_ptr` is `SQL_NULL_DATA`, in which case the octet-length slot is not read" would be unambiguous. Worth the precision since this relaxes a safety contract that callers (and `null_parameters_do_not_read_the_value_buffer`) now rely on.

David-Engel (commented):

This came from an unattended run, so nothing here was checked by a human first.

## Summary

Test-only change to `mssql-odbc`, plus a nextest profile, a README section, PR-validation Miri steps, and pipeline-config regression tests. The only production edit is the `read_indicator` safety-doc rewrite; the unaligned UTF-16 reader fix itself is already on `main`. The new `memory_safety` modules drive the real helpers through displaced pointers, deliberately uninitialized tails, and sentinel-guarded outputs, which is the right shape for Miri, and the binding-layout cases derive expected addresses from typed array/struct fields rather than from the driver's own stride helper. No blocking issues. Three inline notes, all about CI wiring and docs rather than the tests themselves.

## Verification

Read the full diff plus the unchanged production code it exercises: the `api/util.rs` readers and copiers, `conversion/param_buffer.rs` `read_indicator`/`read_param_value`, and `params/bound_param.rs` `for_row`/`with_bind_offset`. Ran `cargo nextest run -p mssqlodbc --lib --profile miri-odbc` natively on the head commit: 24 tests selected and 24 passed, matching the selection claimed in the description.

Sensitivity probes, both reverted afterwards: changing `copy_with_nul`'s `min(buf_len - 1)` to `min(buf_len)` failed `wide_copies_respect_capacity_at_a_byte_offset` and `wide_copies_initialize_only_the_written_prefix`; making `for_row` use the value stride for the indicator and octet-length pointers failed `column_arrays_combine_binding_offsets_and_independent_alignments`. The new tests are not vacuous.

Miri itself was not run here (the pinned nightly is not installed in this environment), so alignment sensitivity rests on the CI jobs. The pipeline YAML was reviewed by reading only; no CI surface was executed. Required checks are green on this head.

## Findings

**Blocking:** none.

**Suggestion:** the Windows Miri JUnit publish path looks inconsistent with that job's `CARGO_TARGET_DIR`, and the profile is self-scoping only if every caller passes `--package mssqlodbc`. Both are inline.

**Nit:** the pinned nightly is duplicated across four files, and the reworded `read_indicator` safety clause has an ambiguous antecedent. Both are inline.

.pipeline/templates/validation-stages.yml
saurabh500: Centralized the pin in the Build stage's `miriToolchain` variable in 879eea59, with both jobs using it for installation, setup, and execution. The config test now checks those references and README version consistency; both expanded commands remain unchanged and passed all 24 Miri tests on their respective platforms.

.pipeline/templates/validation-stages.yml
saurabh500: Keeping this path: nextest's default `store.dir` is workspace-relative `target/nextest`, independent of `CARGO_TARGET_DIR`; I confirmed that with a separate build directory on Windows and CI's nextest 0.9.99 on Linux. The [Windows publication log for build 175217](https://dev.azure.com/SqlClientDrivers/904996cc-6198-4d39-8540-eca72bdf0b7b/_apis/build/builds/175217/logs/260?api-version=7.1) shows all 24 results published as test run 3397464, so deriving this path from `CARGO_TARGET_DIR` would point at the wrong location.

.config/nextest.toml
saurabh500: `binary_id(mssqlodbc)` also fails eagerly in the reduced workspace with `operator didn't match any binary IDs`, reproduced with nextest 0.9.99 and 0.9.122 even when selecting the `ci` profile. I’m keeping the name-only filter plus `--package mssqlodbc` in both CI jobs and the README; the pipeline regression tests require that argument, whereas adopting `binary_id` would restore the Kerberos startup failure.

mssql-odbc/src/conversion/param_buffer.rs
saurabh500: Clarified in 879eea59: the exception applies only when `param.strlen_or_ind_ptr` is non-null and points to `SQL_NULL_DATA`, in which case the octet-length slot is not read. No behavior changed, and the existing ignored-slot cases passed under Miri on both x64 platforms.

mssql-odbc/src/conversion/param_buffer.rs
```diff
@@ -155,8 +155,10 @@ pub(crate) enum Indicator {
 /// buffers is now read correctly instead of one of the two being discarded.
 ///
 /// # Safety
-/// `param.strlen_or_ind_ptr` and `param.octet_length_ptr`, if non-null, must
-/// each point to one valid `SqlLen`.
+/// `param.strlen_or_ind_ptr`, if non-null, must point to one valid `SqlLen`.
+/// `param.octet_length_ptr`, if non-null, must also point to one valid `SqlLen`,
+/// except when `param.strlen_or_ind_ptr` is non-null and points to `SQL_NULL_DATA`.
+/// In that case, the octet-length slot is not read.
```
Vahid-b: The rewording fixes the antecedent, but the exception is now scoped wider than the driver actually honours. `read_indicator` really does return at the `SQL_NULL_DATA` check before touching `octet_length_ptr`, so the clause is accurate *for this function* — the problem is that it is the only reader that behaves this way.

`param_convert::data_at_exec_indicator` (`param_convert.rs:185-188`) reads `*octet_length_ptr` behind nothing but an output-only test and a null check, and it runs for every bound parameter from `exec_common.rs:834` and `exec_direct.rs:212`, ahead of any indicator inspection. `execute.rs:1044` does the same read inline for the parameter-array DAE probe. So on the real execution path a SQL NULL binding whose octet-length slot is invalid or uninitialized is still dereferenced; the exception only holds if `read_indicator` is called in isolation.

That matters for what the new coverage proves. `null_parameters_do_not_read_the_value_buffer` leaves the length slot uninitialized and goes through `read`, so under Miri it pins the narrow contract, not the one the driver relies on end-to-end — worth knowing before someone reads this clause as licence to leave the slot invalid.

Suggest keeping the relaxation attached to the function rather than the binding, e.g. "this function does not read the octet-length slot in that case; other callers on the execution path (`data_at_exec_indicator`) still do", so the wider requirement stays visible.

Vahid-b (commented):

This came from an unattended run, so nothing here was checked by a human first.

## Summary

Re-review on `879eea59`, which only carries the two nit fixes from the previous pass: the pinned nightly is now the Build stage's `miriToolchain` variable (referenced three times in each x64 job, with `scripts/test_validation_pipeline.py` asserting the references and README consistency instead of the literal), and the `read_indicator` safety clause no longer has the ambiguous `it`. The two suggestions from that pass were answered with evidence and are not restated here. The test bodies, the nextest profile, and the pipeline wiring are otherwise unchanged. No blocking issues. One suggestion, about the scope of the reworded safety clause rather than its wording.

## Verification

Confirmed the selection and the run rather than taking the description's counts on trust: `cargo nextest list -p mssqlodbc --lib --profile miri-odbc` selects exactly the 24 claimed tests (19 `memory_safety` cases across `api::util`, `conversion::fetch_convert` and `conversion::param_buffer`, plus the three pre-existing `misaligned_*` cases), and `cargo nextest run` on the head commit reports 24 passed, 1631 skipped.

Read the full merge-base diff (`cc512057`) plus the production code it drives — `copy_with_nul` / `copy_utf16_with_nul` / `write_if_some` / `read_utf16{,_long,_attr}` in `api/util.rs`, `read_indicator` in `conversion/param_buffer.rs`, and `for_row` in `params/bound_param.rs`. Spot-checked the fixtures against their buffers: `check_column_array`'s `16 + 3 * size_of::<T>() < 64` bound holds for both instantiations with the extra displacement byte, and `packed_rows_...`'s 41-byte `#[repr(C, packed)] Row` gives the non-8-divisible stride its assertion claims. The column-wise case pins the real behaviour — `buffer_length` of 128 with `SQL_C_SLONG` still expects a 4-byte stride, so the fixed-width path cannot silently start using `buffer_length`.

Checked the YAML claims by reading: `Build_Windows`/`Build_Linux` are standalone jobs, not matrix jobs, so `--target x86_64-*` cannot leak onto the ARM64 legs, and the `miriToolchain` variable sits at Build-stage scope where both jobs resolve it. `$(miriToolchain)` inside the Linux `bash -lc '...'` payload is an agent-side macro expansion, so the single quotes do not block it.

Miri itself was not run: `rustup toolchain list` shows only `stable`, `1.88.0` and `1.95` on this host, no nightly, so alignment and uninit sensitivity rests on the CI jobs. `python`/`py` are both absent here, so `scripts/test_validation_pipeline.py` was verified by reading its assertions against the YAML and `.config/nextest.toml` (the `$(miriToolchain)` counts of 3, the three `$LASTEXITCODE` guards, the two `set -euo pipefail` occurrences, and the exact `default-filter` string all match). Required checks are pending, not failing, on this head.

## Findings

**Blocking:** none.

**Suggestion:** the relaxed octet-length exception is true of `read_indicator` alone, but reads like a driver-wide guarantee — inline.

**Nit:** none.

mssql-odbc/src/conversion/param_buffer.rs
saurabh500: Clarified in 60312035: the safety docs, test comment, and README now state that skipping the octet-length slot is local to `read_indicator`, not a full-execution guarantee. I checked the earlier probes in `exec_common`, `exec_direct`, and parameter-array execution; their initialized-length requirements and behavior are unchanged.

Theekshna (commented):
This is an unattended automated run; findings were not checked by a human first.

Covered: the delta since the last automated re-review (commit 60312035, "Clarify helper-only SQL NULL coverage"), which precisely resolves Vahid-b's open thread about read_indicator's SQL_NULL_DATA exception being scoped narrower than the full execution path - verified by reading data_at_exec_indicator (param_convert.rs:184-190) directly: it reads octet_length_ptr unconditionally once past the output-only/null checks, confirming the doc/test clarification is accurate and the intra-doc link to it resolves (param_buffer and param_convert are sibling modules under conversion). Also checked test sufficiency (already mutation-tested by David-Engel and Vahid-b on the prior heads), AI-slop scan of the ~25 added comment lines (invariant/why comments, not restating code), PR description/checklist currency, and current checks (all pending, no failures, no contradiction). Local native test verification was not possible in this environment: cargo nextest list -p mssqlodbc fails to build with a pre-existing host-toolchain codegen error (E9012/E9013, AVX-512 intrinsics) unrelated to this PR. No new findings.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

David-Engel (commented):

This came from an unattended run, so nothing here was checked by a human first.

## Summary

Re-review on `60312035`. The delta since the last automated pass is documentation only: the `read_indicator` safety clause now scopes the octet-length exception to that function and points at `param_convert::data_at_exec_indicator`, with a matching README paragraph and test comment. That is exactly what the open thread asked for, and it is accurate — `data_at_exec_indicator` (`param_convert.rs:184-190`) reads `*octet_length_ptr` behind only the output-only and null checks, so the relaxation genuinely belongs to `read_indicator` alone. The intra-doc link resolves: `param_convert` is a sibling module of `param_buffer` under `conversion`, and the target is `pub(crate)` in the same crate. Test bodies, the nextest profile, and the pipeline wiring are unchanged from the previously reviewed head. No findings.

## Verification

Read the full merge-base diff (`cc512057..60312035`) plus the production code the new tests drive: `copy_with_nul` / `copy_utf16_with_nul` / `write_if_some` / `read_utf16{,_long,_attr}` in `api/util.rs`, `read_indicator` / `read_param_value` / `read_wchar_bytes` in `conversion/param_buffer.rs`, and `write_fixed` in `conversion/fetch_convert.rs`.

Ran the selection rather than trusting the description: `cargo nextest list -p mssqlodbc --lib --profile miri-odbc` selects exactly 24 tests (21 `memory_safety` cases across `api::util`, `conversion::fetch_convert` and `conversion::param_buffer`, plus the three pre-existing `misaligned_*` cases), and `cargo nextest run` on this head reports 24 passed, 1631 skipped.

Two fresh mutation probes in areas the earlier passes had not touched, both reverted afterwards. Rounding `read_wchar_bytes`'s explicit byte count up instead of down failed `explicit_wide_lengths_read_only_complete_unaligned_units`. Making `write_fixed` report `size_of::<T>() + 1` in the indicator failed all four `check_write`-based `fetch_convert` cases (`integer_`, `floating_`, `guid_`, `temporal_targets_*`). Combined with the `copy_with_nul` and `for_row` probes from the previous passes, the new tests are not vacuous on the native path.

Spot-checked the fixtures against their buffers again after the reword: `check_fixed`'s 64-byte value/indicator/length buffers stay in bounds for every C type in `PARAMETER_C_TYPES`, `check_column_array`'s indicator and length arrays stay inside 64 bytes at both `±8` binding offsets, and the packed 41-byte `Row` in `packed_rows_*` keeps every field access inside the four-element array for `physical` rows 1 and 3. `null_parameters_do_not_read_the_value_buffer` is the case the new doc paragraph describes, and it does leave both the value and octet-length storage uninitialized, so Miri is what makes it meaningful.

Miri itself was not run here — no nightly toolchain is installed on this host — so alignment and uninit sensitivity continues to rest on the two x64 CI jobs, both of which passed on this head. The pipeline YAML and `scripts/test_validation_pipeline.py` were reviewed by reading only; no CI surface was executed. The Windows JUnit path question from the earlier pass is settled by evidence rather than argument: `Build_Windows` passed with `failTaskOnMissingResultsFile: true`, which it could not have done if the report had landed elsewhere.

## Findings

**Blocking:** none.

**Suggestion:** none.

**Nit:** none.

David-Engel (approved):

PR #565, opened 2026-09-15
Build C++ targets for official CodeQL snapshots

saurabh500:
## Description

Build the ODBC e2e and benchmark C++ targets in the official Windows x64 job for CodeQL, scoped to `cpp` with identifier `mssql_odbc_cpp`. Use clean Debug rebuilds, pin GoogleTest to its v1.17.0 commit, and fail on CMake errors. No SQL Server, test execution, manual CodeQL tasks, or packaging changes.

Validated: **162 pipeline tests pass**; two consecutive local native builds each recompiled all **55 first-party C++ sources**. Discovery and configure/build failure paths are covered.

[Official pipeline run 176122](https://dev.azure.com/SqlClientDrivers/mssql-rs/_build/results?buildId=176122) **succeeded** on `f6f64f15`, including the e2e and benchmark C++ build and all packaging jobs. The branch run had **no CodeQL Initialize/Finalize tasks**, so it validates compilation/packaging, not database upload.

**Pending:** confirm CodeQL finalization and upload from an official `stable` run. Leave the work item open for S360 automation (up to 48 hours).

## Related Issues

[AB#47581 — S360 CodeQL](https://dev.azure.com/SqlClientDrivers/mssql-rs/_workitems/edit/47581)

Local Rust checks not rerun for this pipeline/test-build-only change.


scripts/test_release_pipeline.py
```diff
@@ -113,6 +113,82 @@ def test_release_defaults_are_safe():
...
+    )
+    script = next(
+        step["pwsh"] for step in job["steps"]
+        if step.get("displayName") == "Build ODBC C++ targets for CodeQL"
+    )
+    stub = f"""
+    function cmake {{
+        Write-Output "cmake $args"
+        $global:LASTEXITCODE = if ($args[0] -eq '-S') {{ {configure_exit} }} else {{ {build_exit} }}
+    }}
```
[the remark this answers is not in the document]
saurabh500: Added cases that force the CMake lookup to fail and stub vswhere, checking that the discovered directory is prepended to PATH and an empty result fails before compilation. All 157 pipeline tests pass after merging main; removing either the PATH update or the missing-CMake error makes the new cases fail.

David-Engel (commented):

This is an unattended run, so nothing below was checked by a human first.

## Summary

Reviewed at `8a30d68a` against `origin/main`. Net diff is two files: an official-only CodeQL/C++ build step in the `Windows_x64` job, and 103 lines of pipeline tests. Nothing blocking — the change is correctly scoped to `isOfficial`, fails loudly on CMake errors, and the new tests are load-bearing. Four small items below, all non-blocking.

## What I verified

- **Scoping is right.** `isOfficial: true` is set only by `.pipeline/OneBranch/OfficialPythonWheelsBuild.yml:55` (trigger: `stable`, `pr: none`), so neither PR validation nor `NonOfficialPythonWheelsPublish.yml` picks up the CodeQL variables or the build step. The test's `for other in jobs: assert "Codeql." not in str(other)` pins that no other job inherits them.
- **The build stands alone.** `mssql-odbc/tests/e2e/CMakeLists.txt` links `odbc32`/`odbccp32` and pulls gtest via `FetchContent`; it does not consume the Rust driver, so the step is correct even in the `buildPythonWheels: false` / `buildOdbcNative: false` combination the test exercises. Native-command failures don't trip `$ErrorActionPreference = 'Stop'`, so the explicit `$LASTEXITCODE` checks are the right pattern here.
- **The tests actually guard the script.** Ran `python -m pytest scripts/test_release_pipeline.py -k cpp_codeql -q` → 9 passed. Mutation check: deleting the `throw "CMake configure failed..."` line from `stages.yml` makes `test_cpp_codeql_build_propagates_cmake_failures[1-0-path]` fail. The pwsh-subprocess pattern matches the rest of this file, and the tests run on the Windows agent (`validation-stages.yml:130`), where `pwsh` is available.
- **The Copilot thread on `test_release_pipeline.py:188` is genuinely addressed** — the `vs`/`missing` cases now force the `Get-Command` lookup to fail, stub `vswhere`, and assert the discovered directory is prepended to `PATH`.
- **CI is green on this head** (all checks pass, none failing or pending). The coverage comment reports no lines with coverage information, which is expected for a YAML/Python-script diff.

## Findings

### Suggestions

**`stages.yml:150` — the CodeQL C++ database will also contain googletest.** `cmake --build` with no `--target` builds every target the configure produced, and `FetchContent_MakeAvailable(googletest)` adds `gtest`/`gtest_main`/`gmock` to that set, so the traced compilations include `_deps/googletest-src`. Two consequences worth a decision before the first `stable` run rather than after: alerts from vendored third-party code land in the same snapshot as ours, and the official build now performs a live clone of `https://github.com/google/googletest.git` at a mutable tag (`v1.17.0`, `CMakeLists.txt:26`) during a release build. The clone itself is proven to work on this image — PR validation already builds e2e on the same `RUST-W22-SQL25` pool — so this is about snapshot hygiene and official-build supply chain, not about whether the step runs. If the S360 goal is just "a cpp database exists", it may be worth scoping the analysis away from `_deps/` (or pinning the tag to a commit SHA) so the alert list stays actionable.

**`stages.yml:141-145` — the `vswhere` fallback can fail with the wrong error.** If `vswhere.exe` is not at `${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe`, `& $vswhere` raises `CommandNotFoundException` under `$ErrorActionPreference = 'Stop'` and the intended `'CMake not found on PATH or in Visual Studio.'` message never prints. The new `missing` test case can't catch this, because it installs a function at that exact path, so `vswhere` is always callable under test. A `if (-not (Test-Path $vswhere)) { throw 'CMake not found on PATH or in Visual Studio.' }` guard would make the real agent behavior match what the test asserts.

### Nits

**`stages.yml:150` — `build-codeql` escapes `.gitignore`.** `mssql-odbc/.gitignore` ignores `tests/e2e/build/` and `tests/e2e/build_*/`; the hyphen in `build-codeql` matches neither, and `git check-ignore mssql-odbc/tests/e2e/build-codeql/CMakeCache.txt` confirms it is untracked-visible (only the `*.obj`/`*.pdb`-style entries catch anything inside). Harmless in CI, mildly annoying for anyone reproducing the command locally. Either rename the directory to `build_codeql` — which the existing `build_*/` rule already covers — or add the path to `mssql-odbc/.gitignore`.

**`stages.yml:150` — `-DCMAKE_BUILD_TYPE=Debug` is a no-op here.** `Visual Studio 17 2022` is a multi-config generator, so the configuration comes entirely from `--config Debug` on the build line. It's pinned by `test_release_pipeline.py:152`, so it will stick around; dropping both the flag and that assertion would remove a line that reads as load-bearing but isn't.

## Question

`Codeql.Language: cpp` (`stages.yml:120-121`) is a job-scoped override. On this job that's exactly what you want for the tracer, but `Windows_x64` also runs the Rust/maturin builds and the Python packaging steps — did you confirm that fixing the language list to `cpp` doesn't suppress anything the governed template would otherwise have auto-detected in this job? If auto-injected interpreted-language analysis runs in its own job, this is a non-issue and worth ignoring.

shiwanigupta0809 (commented):
This is an unattended automated run; no findings were checked by a human first.

Reviewed current head `8a30d68a9948ef5af7e032661936d5c83324b161` against merge base `aa2ce0e21bac3a44a901e022c9d7b78fbc787742`. No genuine new findings remained after deduplicating existing and resolved discussion.

Blocking: None
Medium: None
Low: None
Suggestion: None
Nit: None

.pipeline/OneBranch/stages.yml
```diff
@@ -131,6 +137,22 @@ stages:
...
+          if (-not (Get-Command cmake -ErrorAction SilentlyContinue)) {
+            $vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
+            $cmake = & $vswhere -latest -products '*' -requires Microsoft.VisualStudio.Component.VC.CMake.Project -find 'Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe' | Select-Object -First 1
+            if (-not $cmake) { throw 'CMake not found on PATH or in Visual Studio.' }
+            $env:PATH = "$(Split-Path $cmake);$env:PATH"
+          }
+          cmake -S mssql-odbc\tests\e2e -B mssql-odbc\tests\e2e\build-codeql -G "Visual Studio 17 2022" -A x64 -DCMAKE_BUILD_TYPE=Debug -DODBC_E2E_FORCE_UNICODE=ON
+          if ($LASTEXITCODE -ne 0) { throw "CMake configure failed with exit code $LASTEXITCODE" }
+          # Reused agents must compile again for each CodeQL database.
+          cmake --build mssql-odbc\tests\e2e\build-codeql --config Debug --clean-first
```
Theekshna: [Suggestion] `cmake --build` here has no `--target`, so it builds every target the configure step produced — including `gtest`/`gtest_main`/`gmock` from `FetchContent_MakeAvailable(googletest)` (`mssql-odbc/tests/e2e/CMakeLists.txt:19-27`). Two consequences for the CodeQL snapshot: vendored googletest sources (`_deps/googletest-src`) end up traced alongside ours, and the official build does a live clone of `https://github.com/google/googletest.git` at a mutable tag (`v1.17.0`) during a release run. Worth scoping the CodeQL analysis away from `_deps/`, or pinning the tag to a commit SHA, before the first `stable` run rather than after.

.pipeline/OneBranch/stages.yml
```diff
@@ -131,6 +137,22 @@ stages:
         skipPythonSetup: true
         skipRustupInstall: false
         enableJsDeps: false
+    - ${{ if eq(parameters.isOfficial, true) }}:
+      # Auto-injected CodeQL Initialize/Finalize tasks wrap this native build.
+      - pwsh: |
+          $ErrorActionPreference = 'Stop'
+          if (-not (Get-Command cmake -ErrorAction SilentlyContinue)) {
+            $vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
+            $cmake = & $vswhere -latest -products '*' -requires Microsoft.VisualStudio.Component.VC.CMake.Project -find 'Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe' | Select-Object -First 1
```
Theekshna, line 146: [Suggestion] If `vswhere.exe` isn't at `${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe`, `& $vswhere` raises `CommandNotFoundException` under `$ErrorActionPreference = 'Stop'` before the intended `'CMake not found on PATH or in Visual Studio.'` message ever prints. The new `missing`-case test can't catch this because it installs a function at that exact path, so `vswhere` is always callable under test. A `if (-not (Test-Path $vswhere)) { throw 'CMake not found on PATH or in Visual Studio.' }` guard would make the real agent's failure mode match what the test asserts.

.pipeline/OneBranch/stages.yml
```diff
@@ -131,6 +137,22 @@ stages:
...
+            if (-not $cmake) { throw 'CMake not found on PATH or in Visual Studio.' }
+            $env:PATH = "$(Split-Path $cmake);$env:PATH"
+          }
+          cmake -S mssql-odbc\tests\e2e -B mssql-odbc\tests\e2e\build-codeql -G "Visual Studio 17 2022" -A x64 -DCMAKE_BUILD_TYPE=Debug -DODBC_E2E_FORCE_UNICODE=ON
```
Theekshna: [Nit] `mssql-odbc/.gitignore` ignores `tests/e2e/build/` and `tests/e2e/build_*/`; the hyphen in `build-codeql` matches neither, so it's untracked-visible only via the generic `*.obj`/`*.pdb` rules deeper in the file. Harmless in CI, mildly annoying to reproduce locally. Rename to `build_codeql` (already covered by the existing `build_*/` rule) or add the path to `.gitignore`.

Theekshna: [Nit] `-DCMAKE_BUILD_TYPE=Debug` is a no-op here: `Visual Studio 17 2022` is a multi-config generator, so the configuration comes entirely from `--config Debug` on the build line (line 153). It's pinned by `scripts/test_release_pipeline.py:152` so it won't silently regress, but dropping both the flag and that assertion would remove a line that reads as load-bearing but isn't.

Theekshna (commented):
This is an unattended automated run; findings below were not checked by a human first.

## Summary

Official-only CodeQL/C++ build for `mssql-odbc`'s e2e CMake targets, plus 125 lines of pipeline tests. Scoped correctly to `isOfficial`, fails loudly on CMake errors. Reviewed at `8a30d68a` against merge-base `aa2ce0e2` on `main`.

## What I verified

- **Test count claim checks out exactly.** "157 pipeline tests" = `scripts/test_release_pipeline.py` (146, up from 137 at base — the 9 new parametrized cases) + `scripts/test_validation_pipeline.py` (11, unchanged). Ran both together: 157 passed. "52 C++ sources" also matches (`mssql-odbc/tests/e2e/tests/*.cpp` = 49 + 3 lib files = 52).
- **Mutation check.** Deleting the `throw "CMake configure failed..."` line makes `test_cpp_codeql_build_propagates_cmake_failures[1-0-path]` fail, so the new test is load-bearing, not vacuous.
- **Scoping and layering.** `isOfficial: true` only comes from `OfficialPythonWheelsBuild.yml` (trigger: `stable`), so PR validation never picks up the CodeQL variables or build step; the test's `assert "Codeql." not in str(other)` pins that.
- **No msodbcsql/ODBC-driver parity surface touched** (pipeline YAML + Python test script only) — N/A for that check.
- **PR description is current**: diff stat (125 additions, 0 deletions, 2 files) matches, AB#47581 is linked, and the checklist is reasonably omitted with an explicit note ("Rust checks not rerun for this pipeline/test-only change") rather than left unchecked — no drift against `gh pr checks` (all green).
- **No AI-slop comments** in the diff; the two added `#` comments (`stages.yml:141`, `:152`) each record a non-obvious *why* (CodeQL auto-injection wrapping, forced rebuild on reused agents), not restatement.

The four items below corroborate @David-Engel's open review at this same head SHA — I independently re-verified each against the actual `CMakeLists.txt` / `.gitignore` / CMake generator semantics rather than taking that review at face value; none are new.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 2 |
| Nit | 2 |

Vahid-b (commented):
**Automated review** — generated by GitHub Copilot on behalf of @Vahid-b. This is **not** an approval and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

## Summary

Reviewed `8a30d68a` against merge-base `aa2ce0e2` on `origin/main`. Net diff is 125 additions across two files: an `isOfficial`-gated CodeQL/C++ build step in the `Windows_x64` job, and nine new parametrized pipeline tests. The change does what it says — correctly scoped, fails loudly, and the new tests are load-bearing.

Prior reviews at this same head already cover googletest landing in the snapshot, the `vswhere` guard, the `build-codeql` gitignore gap, the `-DCMAKE_BUILD_TYPE=Debug` no-op, and the `Codeql.Language` override question. I am not repeating any of those. Only one item below is new, and it is non-blocking.

## Verification performed

- Confirmed repository is `microsoft/mssql-rs` and head SHA `8a30d68a…` before reading the diff, before running anything, and again immediately before posting.
- Computed merge base independently: `aa2ce0e21bac3a44a901e022c9d7b78fbc787742`. Diff stat matches the PR description exactly (2 files, +125, -0).
- Read both changed files in full, plus `mssql-odbc/tests/e2e/CMakeLists.txt`, `mssql-odbc-bench/CMakeLists.txt`, `.github/workflows/codeql.yml`, and the surrounding `Windows_x64` job definition in an isolated detached worktree (removed afterward; primary checkout untouched).
- Inventoried every C++/CMake path in the tree at the reviewed SHA to establish what the `cpp` database will and will not contain.
- Searched `.pipeline/`, `.github/`, and `scripts/` for any reference to `mssql-odbc-bench` / `odbc_bench` — zero matches, confirming nothing else compiles that tree.
- `gh pr checks`: all green at this head (ADO PR validation, CodeQL, coverage-report, CLA). Coverage comment reports no coverage-bearing lines, expected for a YAML/Python diff.
- Did **not** re-run the pytest suite: no real Python interpreter is available on this machine (only the Windows Store alias stubs), and two reviewers already ran it with a mutation check at this exact SHA. My finding below is established statically and does not depend on running it.

## Findings

### Blocking

None.

### Suggestion

**`mssql-odbc-bench/` is first-party C++ that no build traces, so it stays out of the `mssql_odbc_cpp` snapshot.** The new step configures only `-S mssql-odbc\tests\e2e`, so the database will contain the e2e test tree (plus vendored googletest, per the existing thread on line 153) and nothing else. But `mssql-odbc-bench/` is a second, independent CMake project with three Microsoft-copyrighted `.cpp` files and a `.hpp` (`src/odbc_bench.cpp`, `src/benchmark_main.cpp`, `src/admin_main.cpp`, `include/odbc_bench.hpp`), built with `/W4 /permissive-` under MSVC. It is compiled nowhere in CI — no match for `odbc-bench` or `odbc_bench` anywhere under `.pipeline/`, `.github/`, or `scripts/`.

This matters more than it might look, because `.github/workflows/codeql.yml` analyzes `languages: actions` only. That makes this ADO job the sole C++ coverage in the repo, so `mssql-odbc-bench` currently gets scanned by nothing at all. The net effect is a snapshot composed entirely of test/benchmark-adjacent code with a first-party C++ tree silently excluded — which is the shape of gap S360 tends to flag later.

It looks cheap to include: like the e2e project, the bench links only the platform Driver Manager (`odbc32` on Windows) and pulls its one dependency via `FetchContent`, needs no SQL Server to *compile* (only to *run*), and its README states it never links the Rust driver. A second configure/build pair against `-S mssql-odbc-bench`, or a decision recorded on AB#47581 that the bench is deliberately out of scope, would close this either way.

One piece of context that may help with @Theekshna's open line-153 thread rather than duplicate it: `mssql-odbc-bench/CMakeLists.txt:20` already pins its `FetchContent` dependency to a commit SHA with the tag in a trailing comment (`GIT_TAG c58e6d0710581e3a08d65c349664128a8d9a2461 # v1.9.1`). That is exactly the pinning pattern suggested there for gtest's mutable `v1.17.0`, so there is an in-repo precedent to copy.

### Nit

**`$(Split-Path $cmake)` collides with Azure Pipelines macro syntax on the untested fallback path.** In `stages.yml`, `$(...)` inside an inline `pwsh:` block is first processed by Azure DevOps as macro substitution — the same mechanism that expands `$(Build.SourcesDirectory)` elsewhere in this file. This works today: ADO leaves unrecognized macros literal, so PowerShell receives the text and evaluates it as a subexpression. Worth flagging only because this specific line sits on the branch that never executes when `cmake` is already on `PATH` — which, per the PR description's successful native build, is the case on `RUST-W22-SQL25` — so it has not actually run on an agent. Assigning `$cmakeDir = Split-Path $cmake` on its own line and interpolating `"$cmakeDir;$env:PATH"` sidesteps the ambiguity entirely. Distinct from @Theekshna's line-146 thread, which concerns `& $vswhere` throwing before the intended error message.

## Relevant required-CI failures

None. All required checks pass on `8a30d68a`.

saurabh500:
@David-Engel Thanks for the review. My take after checking the code and build logs:

- I agree with pinning GoogleTest v1.17.0 to its verified commit SHA, renaming the directory to `build_codeql` (already ignored), and removing the redundant `CMAKE_BUILD_TYPE` flag. These are recommendations for the next patch, not changes already in this head. Pinning and excluding third-party analysis solve different problems; I would not add a blanket `_deps` exclusion just to suppress potential findings.
- I would keep the current missing-vswhere behavior. I reproduced it: PowerShell stops immediately and names the missing executable. Replacing that useful error with the generic CMake message adds a check without improving correctness. The `missing` test covers a callable discovery tool returning no CMake, not a missing discovery tool.
- `Codeql.Language: cpp` does narrow this job's analysis. The unchanged Package Rust crates job in official stable build 175362 independently uploaded JavaScript, PowerShell, Python, and T-SQL databases (Finalize log 112). That supports the job-local override, though it is not proof of successful Actions/IaC coverage.

One validation limit worth making explicit: branch build 175428 succeeded and compiled the C++ targets, but its timeline has no CodeQL Initialize/Finalize tasks. It proves the build works, not database upload; the official stable snapshot still needs confirmation.

I also checked the newer reviews. @Vahid-b's benchmark coverage concern is useful: this step does not compile `mssql-odbc-bench`, and adding its compile-only CMake build would cover another first-party C++ tree. The claim that nothing in CI references/builds it is too broad: `.pipeline/odbc-perf-baseline-{windows,linux}-pipeline.yml` reference the perf-lab runners, and the Windows runner configures and builds it. That alone does not establish CodeQL coverage. The PowerShell `$(Split-Path $cmake)` nit does not need a fix: Azure Pipelines documents preserving unresolved macros specifically so shell expressions keep working.

.pipeline/OneBranch/stages.yml
saurabh500: Pinned GoogleTest v1.17.0 to its verified commit 52eb8108c5bdec04579160ae17225d66034bd723 in f6f64f15, matching the existing benchmark dependency pattern. I kept third-party analysis enabled: pinning addresses reproducibility, whereas blanket exclusions would hide findings rather than address them.

.pipeline/OneBranch/stages.yml
saurabh500: Keeping the existing failure: I reproduced the missing-executable case, and PowerShell stops immediately with an error naming the missing vswhere path. The test covers discovery returning no CMake; adding a filesystem check just to replace the more specific missing-tool error with a generic message does not improve correctness.

.pipeline/OneBranch/stages.yml
saurabh500: Renamed the build directory to build_codeql in f6f64f15; git check-ignore confirms both the e2e and benchmark build directories match their existing ignore rules.

saurabh500: Removed CMAKE_BUILD_TYPE and updated the assertion in f6f64f15. Both Visual Studio builds select Debug through --config Debug only.

saurabh500:
@David-Engel @Theekshna @Vahid-b Addressed the accepted review items in f6f64f15: GoogleTest is pinned to the verified v1.17.0 commit, the ignored build_codeql directory is used, the redundant CMAKE_BUILD_TYPE flag is removed, and the official step now also compiles mssql-odbc-bench. All four inline threads are answered and resolved; the missing-vswhere check was declined because the existing failure identifies the missing tool and already stops the job.

Keeping the PowerShell subexpression too: [Azure Pipelines documents preserving unresolved macros](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops#macro-syntax-variables) so shell expressions retain their meaning. There is no need to add a variable solely to avoid this supported syntax.

162 pipeline tests pass, and two consecutive native builds each recompiled all 55 first-party C++ files without running the executables. Pushed the identical commit to GitHub and ADO and queued [Official Python Wheels Build 176122](https://dev.azure.com/SqlClientDrivers/mssql-rs/_build/results?buildId=176122). ADO rejected the attempt with run-only CodeQL diagnostic overrides; the accepted run uses the saved pipeline settings. Remote validation and the stable-branch CodeQL upload are not yet confirmed.

Theekshna (commented):
Unattended automated re-review at `f6f64f15`: no new findings. The new commit (benchmark CodeQL build) is correctly tested and mirrors the e2e build; all four items from the prior sweep are either fixed in this commit or answered with a sound rationale I did not re-raise.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

saurabh500:
@David-Engel @Theekshna @Vahid-b [Official Python Wheels Build 176122](https://dev.azure.com/SqlClientDrivers/mssql-rs/_build/results?buildId=176122) succeeded on f6f64f15. The Windows x64 C++ step compiled both e2e and benchmark targets (log 133), and all build/packaging jobs passed.

The branch run has no CodeQL Initialize/Finalize tasks, so this confirms native compilation and packaging, not a finalized/uploaded database. That verification remains pending on the official stable run; AB#47581 stays open for S360 automation.

David-Engel (commented):

This is an unattended run, so nothing below was checked by a human first.

## Summary

Re-reviewed at `f6f64f15` against `origin/main` (merge base `aa2ce0e2`). Three files: the official-only CodeQL/C++ build step in `Windows_x64`, the GoogleTest SHA pin, and 118 lines of pipeline tests. **No findings.** Every item from my `8a30d68a` review is either fixed or reasonably declined with evidence, and the benchmark addition is correct.

## What I verified at this head

- **The four accepted items landed.** GoogleTest is pinned to `52eb8108` with the `# v1.17.0` provenance comment; the directory is now `build_codeql`, and `git check-ignore -v` confirms both paths are covered (`mssql-odbc/.gitignore:9 tests/e2e/build_*/` and `mssql-odbc-bench/.gitignore:3 build_*/`), so nothing lands untracked on a reused agent; the redundant `CMAKE_BUILD_TYPE` flag and its assertion are gone; `mssql-odbc-bench` is now configured and built.
- **First-party C++ coverage is complete.** Enumerating `*.cpp|*.cc|*.c|*.h` outside `target/`, `node_modules/`, `build*/`, and `_deps/` yields exactly three trees: `mssql-odbc/tests` (53), `mssql-odbc-bench/src` (3), `mssql-odbc-bench/include` (1). Both are now built, so there is no first-party C++ left out of the snapshot. `mssql-odbc-bench/perf-lab` holds only text files and shell/pwsh runners, so nothing is missed there.
- **The benchmark build stands alone and needs no extra flags.** `mssql-odbc-bench/CMakeLists.txt` links `odbc32` on `WIN32`, sets `UNICODE`/`_UNICODE` itself at line 45, and pins `google/benchmark` to `c58e6d07` — so omitting `-DODBC_E2E_FORCE_UNICODE=ON` on the benchmark configure is correct, not an oversight. Neither project consumes the Rust driver, so the step is valid in the `buildOdbcNative: false` combination the tests exercise.
- **Failure propagation is still exact after the benchmark was added.** I walked the new `expected_calls` arithmetic against the script for all 20 parametrizations: an e2e configure failure stops after 1 `cmake` call, an e2e build failure after 2, a benchmark configure failure after 3, a benchmark build failure after 4. Ran `python -m pytest scripts/test_release_pipeline.py -q -k codeql` in a clean worktree → **14 passed**. These tests run on the Windows agent (`validation-stages.yml:130`), where `pwsh` and Windows `Split-Path` semantics are available, so the `vs` stub's `PATH` assertion holds.
- **Scoping is unchanged and still correct.** The `Codeql.*` variables and the step remain behind `${{ if eq(parameters.isOfficial, true) }}`, and `for other in jobs: assert "Codeql." not in str(other)` keeps them off every other job. `.github/workflows/codeql.yml` analyzes only `languages: actions`, so there is no overlap with or regression of the GitHub-side analysis.
- **CI is green on this head** — all 20 checks pass, including ADO build 176121; nothing is failing or pending.

## Residual risk (not a finding, and already disclosed)

The one thing still unproven is the thing this PR exists for: branch run 176122 compiled both C++ targets but had no CodeQL Initialize/Finalize tasks, so `Codeql.Enabled`/`Codeql.Language`/`Codeql.BuildIdentifier` have not yet been observed producing and uploading a database. You already called this out in the description and in AB#47581, and it is only observable from a `stable` run, so it is not something this PR can close. Worth watching the first stable run for an actual `mssql_odbc_cpp` snapshot rather than assuming the variables took effect.

One note for whoever reads the resulting S360 alerts: the traced build includes `_deps/googletest-src` and `_deps/benchmark-src`, so third-party code will appear in the same database. You consciously declined a blanket `_deps` exclusion, which is a defensible call — just be ready for that noise on the first snapshot instead of treating it as a surprise.

shiwanigupta0809 (commented):
This is an unattended automated run; no findings were checked by a human first.

## Summary

Reviewed `f6f64f1562fa18814f36e5c1b9af13683bc17cad` against merge base `aa2ce0e21bac3a44a901e022c9d7b78fbc787742`. The PR correctly scopes the C++ CodeQL build to the official Windows x64 job, pins fetched dependencies, clean-builds both first-party CMake projects, and tests selection and failure propagation. Existing and resolved discussion was checked; no genuine new findings remain.

## Findings

- **Blocking:** None
- **Medium:** None
- **Low:** None
- **Suggestion:** None
- **Nit:** None

All current PR checks are passing. The remaining stable-run CodeQL upload verification is already disclosed in the PR description and tracked by the linked work item.

shiwanigupta0809:
Unattended automated code review posted: https://github.com/microsoft/mssql-rs/pull/565#pullrequestreview-5239114462

Category counts: Blocking 0, Medium 0, Low 0, Suggestion 0, Nit 0.

Vahid-b (commented):
**Unattended automated review** — generated by GitHub Copilot on behalf of @gargsaumya. This is **not an approval** and does not satisfy the human review requirement. Nothing below was checked by a human first; push back on anything that looks off.

## Summary

Reviewed `f6f64f15` against merge base `aa2ce0e2` on current `origin/main`. Net diff is 3 files / +145 −1: the `isOfficial`-gated CodeQL C++ build step in `Windows_x64`, the GoogleTest SHA pin, and 118 lines of pipeline tests. **No new findings.** Every item from the `8a30d68a` round is either fixed here or declined with a reason I independently checked and agree with, and the benchmark addition is correct and tested.

## What I verified independently

- **The GoogleTest pin is genuinely v1.17.0.** `gh api repos/google/googletest/git/ref/tags/v1.17.0` returns `52eb8108c5bdec04579160ae17225d66034bd723` with `type: commit`, exactly the SHA in `mssql-odbc/tests/e2e/CMakeLists.txt:25`. It is a lightweight tag, so there is no annotated-tag object to dereference and the SHA is the commit itself — the `# v1.17.0` comment is accurate, and this matches the `# v1.9.1` pattern already used by the benchmark project.
- **The SHA pin does not collide with shallow fetching.** Pinning `GIT_TAG` to a commit breaks `FetchContent` when `GIT_SHALLOW` is on, because a shallow clone only fetches ref tips. Neither `CMakeLists.txt` sets `GIT_SHALLOW`, so this configuration is safe — worth stating since it is the usual way a tag→SHA pin regresses.
- **Failure propagation, run against the real script rather than a transcription.** I extracted the `pwsh` block straight out of `stages.yml` and drove it under `pwsh 7.6.6` with a `cmake` stub. All five paths behave exactly as the new test asserts: all-ok → 4 calls / success; e2e configure fail → 1 call / fail; e2e build fail → 2 calls / fail; benchmark configure fail → 3 calls / fail; benchmark build fail → 4 calls / fail. The `expected_calls` arithmetic added for the benchmark is correct at every point.
- **Mutation check on the new benchmark guard.** Deleting `if ($LASTEXITCODE -ne 0) { throw "Benchmark CMake configure failed..." }` flips a failing benchmark configure from 3 calls / job failure to 4 calls / job **success** — so that guard is the only thing standing between a broken benchmark configure and a green official run, and the test that pins it is load-bearing rather than vacuous.
- **`$ErrorActionPreference = 'Stop'` does not cover this.** It does not trip on native-command exit codes, so the explicit `$LASTEXITCODE` check after every `cmake` invocation is the right pattern here, not redundancy.
- **Both build directories are ignored.** `git check-ignore -v` resolves `mssql-odbc/tests/e2e/build_codeql/CMakeCache.txt` to `mssql-odbc/.gitignore:9 tests/e2e/build_*/` and `mssql-odbc-bench/build_codeql/CMakeCache.txt` to `mssql-odbc-bench/.gitignore:3 build_*/`, so the rename off `build-codeql` covers the new benchmark tree too and nothing lands untracked on a reused agent.
- **Scoping is airtight.** `isOfficial: true` is set only at `OfficialPythonWheelsBuild.yml:55` (trigger `stable`); `NonOfficialPythonWheelsPublish.yml:80` sets `false`. Both the `Codeql.*` variables and the step sit behind `${{ if eq(parameters.isOfficial, true) }}`, and the test's `assert "Codeql." not in str(other)` pins that no sibling job inherits them.
- **Neither project needs the Rust driver or a live server to compile.** e2e links `odbc32`/`odbccp32` plus `gtest_main`; the benchmark links `odbc32` and sets `UNICODE`/`_UNICODE` itself at `mssql-odbc-bench/CMakeLists.txt:45`, which is why omitting `-DODBC_E2E_FORCE_UNICODE=ON` on its configure is correct rather than an oversight. `ODBC_E2E_FORCE_UNICODE` is a real option (`mssql-odbc/tests/e2e/CMakeLists.txt:71`). No `ctest`/`run_e2e` in the step, so nothing executes.
- **CI is green at this head** — all 20 required checks pass, none failing or pending. The coverage comment reporting no coverage-bearing lines is expected for a YAML/Python diff.

I did **not** re-run `pytest`: no real Python is on this machine (only the Windows Store alias stubs). That is why I exercised the extracted script under `pwsh` directly instead — it tests the agent behavior rather than the harness, and the findings above do not depend on the pytest run that two other reviewers already completed with a mutation check at this SHA.

## Findings

| Severity | Count |
|---|---|
| Blocking | 0 |
| Medium | 0 |
| Low | 0 |
| Suggestion | 0 |
| Nit | 0 |

Deduplicated against all five resolved threads and the prior review bodies; the GoogleTest `_deps` scoping, the `vswhere` `Test-Path` guard, the `build-codeql` ignore gap, the `CMAKE_BUILD_TYPE` no-op, the `Codeql.Language` override question and the `$(Split-Path $cmake)` macro nit are all already raised and answered, so I am not restating any of them.

## Residual risk (already disclosed, not a finding)

The one thing this PR cannot prove is the thing it exists for: branch run 176122 compiled both C++ targets but had no CodeQL Initialize/Finalize tasks, so `Codeql.Enabled` / `Codeql.Language` / `Codeql.BuildIdentifier` have not yet been observed producing and uploading a database. That is only observable from a `stable` run, and it is already called out in the description and on AB#47581. Worth confirming an actual `mssql_odbc_cpp` snapshot appears on the first stable run rather than assuming the variables took effect. Relatedly, since neither project excludes `_deps/`, expect `googletest-src` and `benchmark-src` in that first database — a consciously accepted trade-off, so treat the third-party alerts as known noise rather than a surprise.

David-Engel (approved):

PR #568, opened 2026-09-16
Fix benchmark LOB chunk length accounting

### LINKED ISSUE #567: Fix ODBC benchmark LOB chunk length accounting
### Describe the bug

The ODBC performance benchmark's LOB preflight inferred `SQLGetData` payload length by trimming NUL bytes backward from the reused 8192-byte chunk buffer. That assumes intermediate string chunks either fill the buffer or clear stale bytes after the current terminator.

After the Rust ODBC `VARCHAR(MAX)` conversion path started delivering shorter chunks, stale bytes from a previous `NVARCHAR(MAX)` read were counted as current `VARCHAR(MAX)` payload bytes. Pipeline run 20260915.1 failed at row 557 with `preflight value mismatch for LOB payload length`.

### Steps to reproduce

1. Run the Linux ODBC performance regression pipeline on commit `e8075729b3facd3691d5ed0ab36736e0c58384c6`.
2. Inspect the `Run tests on perf VM` log for build `175420`.
3. Observe the candidate driver failing during `getdata/rowwise_1k_c3_lob_max/chunked_8192` preflight.

### Expected behavior

The benchmark should assemble each generated LOB value from the bytes written by the current `SQLGetData` call only, ignoring stale bytes after the current string terminator. The LOB preflight should keep validating exact length, content, and continuation-call behavior.

### Actual behavior

The benchmark counts stale bytes after a short `SQL_C_CHAR` chunk's terminator, then fails with:

```text
ODBC benchmark failed: preflight value mismatch for LOB payload length at row id 557
```

Microsoft ODBC 18.6.2.1 passes the same preflight because its chunk pattern does not expose the stale-buffer assumption; the issue is in the benchmark reader.

### Version

Commit `e8075729b3facd3691d5ed0ab36736e0c58384c6`; fixed branch commit `54b8d5d468f3ca8c6a397c271e61350924fe8812`.

### Affected crate

mssql-odbc

### Environment

- OS: Ubuntu 22.04 perf VM
- Pipeline: GH-mssql-odbc Linux performance regression
- Microsoft ODBC Driver: 18.6.2.1

### Additional context

Validation run `175466` on branch `saurabh500-pipeline-failure` confirms the row-557 failure is fixed: the offline LOB chunk regression test passes and candidate, pinned baseline, and Microsoft ODBC all pass the LOB preflight with checksum `0x9d28f07ca94516fb`. That run still fails later on unrelated performance-regression gates.

saurabh500:
## Description

Fix the existing benchmark's LOB reader, not the SQL driver. A short `SQLGetData` read leaves old bytes after the new text's NUL terminator in the reused buffer. The old backward scan counted those stale bytes; the reader now scans forward to the first complete NUL character.

The change stays inside `read_lob_column()` with a related documentation/comment correction. No new tests, helper files, CMake changes, or pipeline-script changes. Existing exact length/content and continuation checks remain unchanged.

Validation:
- The simplified benchmark and admin executables build locally at commit `01e1ec7`.
- Azure DevOps run [175466](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=175466&view=logs), on the earlier commit `54b8d5d`, verified the same forward-scan logic: the Rust candidate, pinned Rust baseline, and Microsoft ODBC 18.6.2.1 all passed the previously failing LOB preflight with checksum `0x9d28f07ca94516fb`.
- That run failed later on four confirmed performance regressions in other workloads. The pipeline has not been rerun on the simplified commit; this PR does not change the product, pinned baseline, or performance gate.

## Related Issues

Fixes #567

## Checklist

- [ ] `cargo bfmt` passes
- [ ] `cargo bclippy` passes
- [ ] `cargo btest` passes
- [x] New/changed functionality has tests — existing benchmark correctness checks retained; no new tests added
- [x] Public API changes are documented — no public API changes; benchmark documentation updated


Theekshna (approved):

mssql-odbc-bench/src/odbc_bench.cpp
```diff
@@ -1858,23 +1857,19 @@ class WorkloadRunner::Impl {
...
-            // A driver that knows the remaining length reports it; one streaming
-            // an unbounded PLP value reports SQL_NO_TOTAL. Either way a value
-            // longer than the buffer means the call filled it.
-            std::size_t payload = kLobChunkBytes;
-            if (indicator >= 0 && static_cast<std::size_t>(indicator) < kLobChunkBytes) {
-                payload = static_cast<std::size_t>(indicator);
+            // Generated text has no embedded NUL. Short chunks can leave stale
+            // bytes after the current terminator in this reused buffer.
+            std::size_t payload = 0;
+            for (; payload + column.unit_bytes <= kLobChunkBytes;
```
David-Engel, line 1863: Optional, and it does not affect comparability — every driver in the matrix pays it equally — but this scan now sits inside the measured region, and it is more expensive than it looks. `std::all_of` over a 1- or 2-byte unit inside an early-exit loop does not vectorize, so it costs roughly a nanosecond per byte scanned.

Measured on this machine (MSVC 19.44, x64, `/O2`), scanning one full 8192-byte chunk:

| shape | this loop | `memchr` / 16-bit unit compare |
| --- | --- | --- |
| wide (`unit_bytes == 2`) | ~3.9 us | ~1.3 us |
| narrow (`unit_bytes == 1`) | ~7.7 us | below timer resolution |

For `getdata/rowwise_1k_c3_lob_max` that is about 30 us of harness CPU per row (three chunks per value, two LOB columns), roughly 30 ms per 1000-row iteration, added to a workload whose entire point is `SQLGetData` chunk cadence. GCC on the Linux perf VM may vectorize differently, so it is worth confirming there rather than taking these numbers as given.

Same semantics, including the missing-terminator throw, since `kLobChunkBytes` stays the not-found sentinel:

```cpp
std::size_t payload = kLobChunkBytes;
if (column.unit_bytes == 1) {
    if (const void* hit = std::memchr(chunk_.data(), 0, kLobChunkBytes)) {
        payload = static_cast<std::size_t>(
            static_cast<const unsigned char*>(hit) - chunk_.data());
    }
} else {
    for (std::size_t offset = 0; offset + 2 <= kLobChunkBytes; offset += 2) {
        std::uint16_t unit = 0;
        std::memcpy(&unit, chunk_.data() + offset, sizeof(unit));
        if (unit == 0) {
            payload = offset;
            break;
        }
    }
}
```

mssql-odbc-bench/src/odbc_bench.cpp
```diff
@@ -1858,23 +1857,19 @@ class WorkloadRunner::Impl {
...
-            // units recovers the exact payload, as mssql-python's
-            // FetchLobColumnData does; it is exact here because the generated
-            // text contains no embedded NUL.
-            payload -= payload % column.unit_bytes;
-            while (payload >= column.unit_bytes &&
-                   chunk_[payload - 1] == 0 &&
-                   (column.unit_bytes == 1 || chunk_[payload - 2] == 0)) {
-                payload -= column.unit_bytes;
+            if (payload + column.unit_bytes > kLobChunkBytes) {
+                throw std::runtime_error("SQLGetData returned LOB text without a terminator");
```
David-Engel, line 1872: Every other preflight failure in this file goes through `require_value(..., row_id)` and names the row, so a failure in a 1000-row preflight can be located. This one is a bare `runtime_error`: if a driver ever trips it, the log says which invariant broke but not where, and the row id is already in hand.

```cpp
throw std::runtime_error("SQLGetData returned LOB text without a terminator at row id " +
                         std::to_string(row_id));
```

David-Engel (commented):

## Summary

Reviewed the full diff at `01e1ec7b` against merge base `e8075729` in an isolated worktree, and read the existing Copilot bot review first (zero comments, nothing to re-file). The fix is correct: replacing the backward NUL trim with a forward scan to the first complete NUL unit is the right way to recover a short `SQLGetData` chunk out of a reused buffer, and the inlined loop is semantically identical to the `lob_chunk_payload_bytes()` helper that ADO run 175466 validated at `54b8d5d4`. The description matches the diff, and #567 is open and describes exactly this change. Two suggestions and one nit, none blocking.

## Verification

- Extracted the new loop into a standalone MSVC build (x64, `/O2`) and ran it over synthetic chunks: full wide chunk (4095 chars + terminator) gives 8190; full narrow chunk (8191 bytes + terminator) gives 8191; a 100-byte narrow chunk written over stale `'x'` bytes gives 100; a 200-byte wide chunk over stale `0xAB` gives 200; a buffer with no NUL throws. The full-buffer cases are the ones worth pinning: `kLobChunkBytes` is divisible by both unit sizes, so a legitimately filled chunk always ends with its terminator inside the last scanned unit and the new throw cannot misfire.
- `mssql-odbc/src/api/get_data.rs:1865-1873` reserves terminator bytes for every character target, so the driver under test does write the terminator the new throw now depends on.
- `is_lob_kind()` covers only `lob_character` and `lob_wide_character`, so no binary column reaches `read_lob_column()` and a forward NUL scan is safe for everything that does.
- `validate_lob_value()` already compares `lob_calls_[index] >= expected_calls`, so the README's shift from "three calls" to "at least three calls" matches the code that enforces it.
- I did not build the benchmark itself; no ODBC bench toolchain was configured in this environment.
- `gh pr checks 568` at this head: nothing failing. The MacOS test entry is a cancelled, superseded run (build 175580); the rest are pending on 175582.

## Suggestions

1. **The new scan is several microseconds per chunk and it runs inside the timed path** (`mssql-odbc-bench/src/odbc_bench.cpp:1863`) — measured numbers and a drop-in alternative are in the inline comment.
2. **Nothing in this PR's checks compiles the changed file.** `.pipeline/odbc-perf-baseline-linux-pipeline.yml` and its Windows counterpart are `trigger: none` / `pr: none`, and the PR validation pipeline does not build `mssql-odbc-bench`, so a compile break here would surface only the next time someone runs the perf lab. The simplification also dropped `src/lob_chunk.hpp` and its 100-line `tests/lob_chunk_test.cpp`, which was the only automated guard for this logic; that is defensible, since `main` has no test target under `mssql-odbc-bench` and keeping it means new CMake wiring. Combined with the description's honest note that run 175466 predates the simplification, the local build is the only evidence covering `01e1ec7`. Worth a manual perf-pipeline run on this exact commit before merge. (I did confirm the two commits' scan logic is semantically identical, so the run's LOB-preflight result should carry over.)

## Nit

1. **The new `runtime_error` carries no row or column context** (`mssql-odbc-bench/src/odbc_bench.cpp:1872`) — see the inline comment.

mssql-odbc-bench/src/odbc_bench.cpp
saurabh500: Limiting the scope. Will skip this suggestion

mssql-odbc-bench/src/odbc_bench.cpp
saurabh500: Thanks. Additional row context would be useful, but this is a diagnostic improvement rather than a correctness issue. This check also runs outside preflight, and the neighboring no-progress error follows the same direct `runtime_error` pattern without row context. I’d prefer to keep this PR focused on correcting chunk-length accounting and leave diagnostic enrichment out of scope.

shiwanigupta0809 (commented):
## Summary

Unattended automated review of head `01e1ec7be4c070601415031e6c9ebecd31644a3c` against the refreshed `origin/main` merge base. The forward scan correctly stops at the current chunk's first complete NUL unit, avoiding stale bytes in the reused buffer; the linked issue #567 matches the change. Existing discussion already covers the scan-cost, benchmark-build coverage, and diagnostic-context suggestions, so they are not duplicated here.

- **Blocking:** None.
- **Medium:** None.
- **Low:** None.
- **Suggestion:** None.
- **Nit:** None.

Validation reviewed: the full two-file diff, affected LOB generation/validation and driver terminator behavior, all existing inline threads/review bodies/top-level comments, and current checks. The benchmark could not be rebuilt locally because CMake is unavailable on this host; the PR's documented local build and prior perf-pipeline LOB preflight cover the same scan semantics.

PR #570, opened 2026-09-16
Stabilize Python connection-close session tests

### LINKED ISSUE #569: Make Python connection-close tests resilient to session cleanup timing
### Describe the bug

The Python connection-close integration tests identify a server session only by SPID and expect it to disappear from `sys.dm_exec_sessions` exactly 500 ms after `close()`. This couples correctness to SQL Server cleanup timing and allows a reused SPID to be mistaken for the original session.

### Steps to reproduce

Run `mssql-py-core/tests/test_connection.py` against SQL Server. The intermittent failure was observed in [pipeline run 20260916.1, build 175474](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=175474&view=logs&jobId=1dca5ef1-3811-5712-152d-1052a5815ba2&t=c4a41807-f882-55b6-8e4d-06ba7b4c1e43), during validation of #568. A deterministic reproduction of the original failure is not established.

### Expected behavior

Verify that the specific connection's server-side session disappears after close, allowing a bounded interval for asynchronous cleanup. A genuinely persistent session must still fail the test.

### Actual behavior

`test_connection_close_terminates_server_session` failed at `tests/test_connection.py:42` with `AssertionError: SPID should be gone after close` (`1 == 0`). The step reported 1 failed, 976 passed, 11 skipped, and 12 deselected. The bulk-copy close test, which uses the same fixed-delay/SPID pattern, passed in that run.

### Version

Pipeline merge commit `d0b00c595658bfff502daf891bdc86cd7f1b4a38` for #568.

### Affected crate

mssql-py-core

### Environment

Linux x64 build job using the Ubuntu 22.04 build container, Python 3.10.12, pytest 9.1.1, and the pipeline's SQL Server container.

### Additional context

Use a UUID-based `application_name` for each connection under test, confirm its presence through `sys.dm_exec_sessions.program_name` before closing, then poll for disappearance with a monotonic 10-second deadline rather than a fixed sleep. Apply this to both session-close tests and ensure cleanup runs on assertion failures.

The original log cannot distinguish delayed cleanup, SPID reuse, or a shutdown error. This change removes test ambiguity without claiming to prove the original root cause. #568 changes benchmark files only. The separate wheel-packaging failure in the same run is outside this issue's scope.

saurabh500:
## Description

Give each connection under test a UUID-based `application_name` and identify its server session using `sys.dm_exec_sessions.program_name`, rather than a reusable SPID. Confirm that the uniquely named session exists before closing, then poll every 100 ms for disappearance with a monotonic 10-second deadline instead of sleeping for a fixed 500 ms.

Apply the shared fixture and polling helper to both the normal close and bulk-copy close tests. Reuse the existing monitoring-connection fixture and ensure connections and the bulk-copy table are cleaned up on failures. Keep the existing cursor alive across close to preserve the original regression coverage. No production code or public API changes.

The failure in [pipeline build 175474](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=175474&view=logs&jobId=1dca5ef1-3811-5712-152d-1052a5815ba2&t=c4a41807-f882-55b6-8e4d-06ba7b4c1e43) does not establish whether cleanup timing, SPID reuse, or a shutdown failure caused the observed count of 1. This change removes the fixed-delay and session-identity assumptions without weakening the requirement that the session actually disappear. The separate wheel-packaging failure in that run is out of scope.

### Validation

Local validation on Windows with bindings built from this checkout and a local SQL Server:
- `python -m pytest mssql-py-core\tests\test_connection.py -q --tb=short`: **6 passed**.
- Both changed tests repeated in **20 separate pytest runs**: **40/40 passed**.
- One-off live-server negative check: deliberately leave a uniquely named session open; polling correctly fails after **10.08 seconds**. Close that same connection and verify the helper succeeds and its session count is zero. This was a manual validation experiment, not an automated regression test committed in this PR.
- `cargo bfmt` and `git diff --check`: passed locally.

Workspace clippy and the full Rust suite were not run locally. The checked lint/test checklist entries below reflect successful CI validation, not additional local runs.

In [CI build 175585](https://sqlclientdrivers.visualstudio.com/public/_build/results?buildId=175585), format checks, clippy (including workspace + mssql-py-core checks), Rust test jobs, the mssql-py-core Rust unit tests, and both Python-binding test steps passed. The platform build jobs and macOS test job also passed.

The PR is open for review, not a draft. Overall pipeline validation is **not green**: the separate cross-repo `Build mssql-python wheel` step failed. That packaging job is outside this Python-test-only change; the failure should not be confused with a connection-close test failure.

## Related Issues

Fixes #569

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes
- [x] Public API changes are documented (not applicable: no public API changes)


shiwanigupta0809 (commented):
## Summary

Reviewed head `0e0224db0b78876714b9adc5724b7c0e645eda68` against the refreshed `origin/main` merge base. The UUID-backed `application_name` identifies the exact server session, the bounded monotonic polling removes the fixed-delay race without weakening the close assertion, and cleanup preserves the existing connection/cursor regression coverage. No new findings.

- **Correctness:** None found.
- **Security:** None found.
- **Tests:** None found. The linked issue #569 matches the change, and the coverage report shows no coverable production lines in this test-only diff.
- **Readability / maintainability:** None found.
- **Performance / API surface:** None found; this changes integration tests only and introduces no public API behavior.

Validation reviewed: `git diff --check` passed; existing inline comments, review bodies, top-level comments, and checks were inspected. The current cross-repo Linux wheel failure occurs before test execution in packaging (`cp: cannot create regular file 'mssql_py_core/': Not a directory`) and is unrelated to this diff; other live-server CI jobs were still pending when this review was posted.

_This COMMENT review was posted by an unattended automated run and was not pre-approved by a human reviewer._

mssql-py-core/tests/test_connection.py
```diff
@@ -3,45 +3,59 @@
...
 
+def _session_count(cursor, application_name):
+    cursor.execute(
+        "SELECT COUNT(*) FROM sys.dm_exec_sessions "
+        f"WHERE program_name = '{application_name}'"
+    )
+    return cursor.fetchone()[0]
+
+
+def _wait_for_session_close(cursor, application_name):
```
Theekshna, line 20: [Suggestion] This helper's failure path (the `assert` firing when a session outlives the 10 s deadline) has no automated regression test — the PR description mentions only a manual local check ("polling correctly fails after 10.08 seconds"). A helper whose only job is to fail correctly under a specific condition is exactly the kind of logic the repo's "does this test still catch a reverted change" standard applies to. Consider a lightweight unit test that calls this with a stub cursor whose session count never reaches 0 and a short monkeypatched deadline, to pin the failure behavior without waiting 10 real seconds.

Theekshna (commented):
## Summary

Replaces SPID-based session identification and a fixed 500 ms sleep with a UUID-tagged `application_name` (a real, precedented context key — see `test_client_context.py:241` and `test_connstr_params_e2e.py:646`, which confirm it maps to `sys.dm_exec_sessions.program_name` via `APP_NAME()`) and a monotonic 10 s polling deadline. Verified the shared `monitored_connection` fixture correctly separates the connection under test (unique `application_name`) from the pre-existing `connection` fixture used purely for monitoring, and that `PyCoreConnection.close()` (`src/connection.rs:98`) is idempotent, so the fixture's `finally: conn.close()` after the test body already closed it is a safe no-op, not a double-close bug. The `Build mssql-python (cross-repo)` check failure on this PR's run is real but unrelated: it's the exact job removed by #573 (merged same day) as obsolete after microsoft/mssql-python#768 changed the installer; verified by cross-checking that PR's description rather than taking the claim in shiwanigupta0809's review at face value.

Could not independently reproduce the description's local-run figures (pytest "6 passed", "40/40" repeats, "10.08 seconds" negative-check) — this environment has no live SQL Server or built `mssql_py_core` bindings available to this run.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 2 |
| Nit | 0 |

### Suggestion

1. **PR description contradicts its own checklist.** The checklist marks `cargo bclippy` and `cargo btest` as passing, but the description text says "Workspace clippy and the full Rust suite were not run for this Python-test-only change; their checklist entries remain open." It also says "Keep this PR in draft pending CI and review," but the PR is not a draft. Please reconcile: either check only what was actually run, or update the stale draft note.
2. **[inline]** The failure path of the new polling helper (`_wait_for_session_close` timing out and raising) is exercised only by a one-off manual local check described in the PR body, not by an automated test in this diff. See inline comment.

mssql-py-core/tests/test_connection.py
```diff
@@ -69,43 +83,32 @@ def test_connection_close_rejects_query_on_existing_cursor(client_context):
...
+
+        conn.close()
+        _wait_for_session_close(mon_cursor, application_name)
+
+        # Data should persist (committed before close)
+        mon_cursor.execute(f"SELECT COUNT(*) FROM {table_name}")
+        assert mon_cursor.fetchone()[0] == 10
+    finally:
+        conn.close()
+        mon_cursor.execute(f"DROP TABLE {table_name}")
```
Vahid-b, line 111: [Suggestion] This `finally` now runs on the failure path, which the previous code did not: the old `DROP TABLE` sat after the SPID assertion, so a failing test skipped cleanup entirely.

Two consequences:

1. **Exception masking (unconditional).** If this `mon_cursor.execute(...)` raises for any reason, its exception replaces the `AssertionError` from `_wait_for_session_close` on line 104, so the `"Session for ... still exists 10 seconds after close"` message — the diagnostic this PR is specifically adding — never reaches the test report.
2. **Possible block instead of a fast failure (concern, not verified).** The failure this test exists to catch is a session that has *not* gone away. Such a session may still hold locks taken by the `table_lock=True` bulk copy, and this `DROP TABLE` has no lock timeout, so cleanup could hang rather than fail. I could not reproduce this — no live SQL Server was available to this run — so I am flagging the mechanism, not asserting the behavior.

Severity is bounded by the `IF OBJECT_ID(...) DROP TABLE` guard at the top of the test, which cleans up a leaked table on the next run.

Something like this keeps the cleanup while preserving the real failure:

```python
    finally:
        conn.close()
        try:
            mon_cursor.execute("SET LOCK_TIMEOUT 5000")
            mon_cursor.execute(f"DROP TABLE {table_name}")
        except Exception:  # cleanup must not mask the assertion above
            pass
```

Vahid-b (commented):
> **Automated review** — generated by GitHub Copilot on behalf of @Vahid-b. This is **not an approval** and does not satisfy the human review requirement. Findings may be incomplete or wrong; push back on anything that looks off.

## Summary

Test-only change (1 file, +57/-54) that replaces SPID identification and a fixed 500 ms sleep with a UUID-tagged `application_name` matched against `sys.dm_exec_sessions.program_name`, plus a bounded 10 s monotonic poll. The mechanism holds up under inspection: `application_name` is a real context key (`mssql-py-core/src/connection.rs:188`, assigned at `:410`); `PyCoreConnection.close()` is idempotent (`connection.rs:98-112`, guarded by `self.is_closed`), so the fixture's `finally: conn.close()` after the body already closed is a genuine no-op rather than a double-close bug; and the deadline check in `_wait_for_session_close` runs *after* the count query, so a genuinely persistent session still fails instead of being waived. Scope matches issue #569. No blocking issues.

## Verification performed

- Inspected the full diff in a detached worktree at `0e0224db0b78876714b9adc5724b7c0e645eda68` against merge-base `e8075729`.
- **Confirmed the changed tests actually execute in CI** — not established by the earlier reviews on this PR. `.pipeline/templates/build-template.yml:161-167` runs `./dev/test-python.sh` (no flags) on the Linux job under `Build.Reason == PullRequest`; `dev/test-python.sh:113` then runs `pytest tests/ -m "not longhaul and not smoke"`, which *selects* integration tests — `mssql-py-core/pytest.ini` declares `integration` as an opt-out marker and defines no `addopts` that deselects it. `Build Stage Build Linux` passed on build 175585, so both changed tests ran green against a live server. (Note `test-mssql-python-template.yml` runs `--mssql-python`, which only covers `tests/mssql_python/` and does *not* exercise this file.)
- Read the applicable policy files, all present: `.github/copilot-instructions.md`, `.github/instructions/pr-workflow.instructions.md`, `.github/skills/code-review/SKILL.md` and its `posting.md`. Limitation: the repo defines no canonical review-marker format, so the marker below follows the convention already used by prior automated reviews on this PR.
- Checked the fixtures in `mssql-py-core/tests/conftest.py`: `client_context` and `connection` are both function-scoped, and `connection` builds its own context via `get_client_context()`, so the monitoring connection never inherits the UUID `application_name`. The fixture's `== 1` pre-assert is therefore unambiguous.
- Independently verified the failing check rather than accepting the claim in an earlier review: #573 (merged 2026-09-16T04:41Z) deletes `.pipeline/templates/test-mssql-python-template.yml`. That file is absent from `origin/main` but still present at this PR's head, whose merge-base `e8075729` predates the merge.
- Could not run the tests locally: this host has no Python interpreter (`python` / `python3` resolve only to the Windows Store alias stubs and exit 9009), and no live SQL Server or built `mssql_py_core` bindings are available to this run. So the description's local figures ("6 passed", "40/40", "10.08 seconds") were not independently reproduced. Bare `cargo test` was not used.

## Required CI

`mssql-rs Pull request validation (Build mssql-python Build mssql-python (cross-repo))` is **failing**, which also fails the parent `mssql-rs Pull request validation` rollup. Per the verification above this is a stale-base artifact: the job's template was deleted from `main` by #573 *after* this branch's merge-base, so merging `main` into the branch should drop the job entirely. I could not read `repos/microsoft/mssql-rs/branches/main/protection` (HTTP 404 for this token), so I cannot confirm whether this check is formally required — please confirm before merge. Every other check passes.

## Findings

### Blocking

None.

### Suggestion

1. **`mssql-py-core/tests/test_connection.py:109-111` — the new `finally` can swallow the diagnostic this PR exists to produce.** Moving cleanup into `finally` is the right call for table hygiene, but it also means the unconditional `mon_cursor.execute(f"DROP TABLE {table_name}")` now runs *on the failure path*, which the previous code never did (the old `DROP` sat after the assertion, so a failure skipped it). If that `DROP` raises for any reason, its exception replaces the `AssertionError` from `_wait_for_session_close` and the "session still exists 10 seconds after close" message — the entire point of the change — is lost. The most plausible trigger is the exact regression being tested: a session that has not gone away may still hold locks from the `table_lock=True` bulk copy, and the `DROP` has no lock timeout, so cleanup can block instead of failing fast. I have not reproduced this (no live server available), so treat the blocking mechanism as a concern rather than a verified defect; the exception-masking itself is unconditional and follows directly from the control flow. Severity is limited by the `IF OBJECT_ID(...) DROP TABLE` guard at line 92-94, which self-heals a leaked table on the next run. Consider wrapping the cleanup so it cannot mask the real failure — e.g. `SET LOCK_TIMEOUT 5000;` before the `DROP`, and swallowing/warning on cleanup errors instead of propagating them. See inline comment.

### Nit

None.

---

Not duplicated here: the description/checklist contradiction raised by @Theekshna appears already addressed in the current PR body ("The checked lint/test checklist entries below reflect successful CI validation, not additional local runs", "The PR is open for review, not a draft"). Their still-open thread requesting an automated test for `_wait_for_session_close`'s failure path is not restated.

This PR appears ready for human review. It has **not** been approved, and this review does not satisfy the human review requirement.

mssql-py-core/tests/test_connection.py
saurabh500: Thanks—agreed that a stubbed-clock test would provide automated coverage of the timeout branch. For this focused integration-test fix, I’d prefer to defer that addition. We explicitly exercised the negative case against a live SQL Server: an open session failed after 10.08 seconds, and the same session disappeared after closing the connection. That checks the actual DMV lookup and failure behavior together. The integration tests also confirm the uniquely named session exists before close, preventing a mismatched lookup from passing vacuously. I’d treat the additional helper test as optional rather than a prerequisite for this fix.

mssql-py-core/tests/test_connection.py
saurabh500: Thanks for flagging this. A cleanup exception becomes the final exception, but Python preserves the original assertion through exception chaining, so the diagnostic is not unconditionally lost. Also, this bulk copy completes in autocommit mode; a session remaining present does not imply it retains the bulk-update lock. I verified locally that DROP TABLE succeeds from the monitor while the original connection and cursor remain open and its session is still present. A cleanup lock timeout could be useful defensive hardening, but I’d treat it as optional rather than a demonstrated defect here. I’d avoid broadly swallowing cleanup exceptions, since that could hide a real cleanup failure on an otherwise passing test.

Theekshna (commented):
## Summary

Unattended re-review of #570, re-triggered because my prior review thread on this PR was resolved. Re-verified the diff (`mssql-py-core/tests/test_connection.py` vs merge-base `e8075729`), the current PR description against the diff and `gh pr checks`, `PyCoreConnection::close()` idempotency in `connection.rs` (guarded by `is_closed`, never propagates the underlying close error), the `client_context`/`connection` fixtures in `conftest.py` (monitor connection defaults `application_name` to `"mssql-python"`, so it cannot collide with the UUID-tagged session under test), and re-derived the PR's own evidence (6 test functions in the file matches the claimed "6 passed"; 2 changed tests x 20 repeated runs matches "40/40"). `git diff --check` is clean. No ODBC surface is touched (parity check N/A) and no behavioral divergence from msodbcsql/SqlClient/mssql-jdbc is introduced (documentation check N/A).

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

Both findings raised in earlier rounds are resolved, not re-filed: the description/checklist contradiction I raised previously is fixed in the current body, and the missing automated test for `_wait_for_session_close`'s failure path (my suggestion) was discussed and reasonably declined by the author citing a manual live-server negative check, with the thread closed. Vahid-b's `finally`-block exception-masking concern was likewise discussed and declined with a concrete rebuttal (exception chaining, autocommit bulk copy), thread closed. No new issues from this pass.

The `Build mssql-python (cross-repo)` check is still red on the latest run; independently reconfirmed this is the same stale-merge-base artifact already identified: `.pipeline/templates/test-mssql-python-template.yml` was removed from `main` by #573, which post-dates this branch's merge-base (`e8075729`), and the file is still present in this PR's tree.

_This COMMENT review was posted by an unattended automated run and was not pre-approved by a human reviewer._

David-Engel (approved):

PR #575, opened 2026-09-16
Claim parked lookahead rows in the synchronous cursor path

### LINKED ISSUE #574: ODBC single-row fetch enters the async runtime once per row to acknowledge an already-buffered row
## Summary

`TdsClient::try_peek_past_current_row` reads the next row header off the wire, parks it in `active_row_read_state`, and sets `row_already_positioned`. The ODBC busy gate (`exec_common.rs::release_busy_if_row_exhausted`) performs this peek after every delivered row, so in steady state a row is essentially always parked.

`try_next_row_cursor` did not account for that state. It rejected any non-`Idle` `active_row_read_state` and returned `CursorPoll::Pending`, so the fetch paths in `fetch_scroll.rs` fell back to `dbc.runtime.block_on(client.next_row_cursor())`. The async method then immediately hit its own `row_already_positioned` branch and returned `Ok(true)` without touching the transport.

The result was one tokio runtime entry per row purely to acknowledge a row that was already buffered — no I/O, no decoding, pure fixed overhead on the hottest per-row path.

## Impact

Worst on rowset size 1, where the per-row cost is not amortized. Measured in the Linux perf lab (ADO run `175420`, main at `e8075729`), candidate vs Microsoft ODBC 18.6.2.1:

- `fetch/rowset_100k_c15_fixed/bound_rowset_1`: 216.0 ms vs Microsoft 185.1 ms — **1.167x slower**, the only benchmark exceeding the 5% acceptance gap.
- Same benchmark vs the pinned baseline `c453af2c`: **1.48x**.

Larger rowsets (64/1000) were unaffected, which is consistent with a fixed per-row cost.

## Expected behavior

The synchronous cursor should claim a parked row the same way `next_row_cursor` already does, without entering the runtime.

## Notes

Allocation counts are identical before and after, confirming the cost is the runtime transition rather than allocation.

saurabh500:
## Description

`TdsClient::try_peek_past_current_row` reads the next row header off the wire, parks it in `active_row_read_state`, and sets `row_already_positioned`. The ODBC busy gate (`release_busy_if_row_exhausted`) runs that peek after every delivered row, so in steady state a row is essentially always parked.

`try_next_row_cursor` did not account for that state: it rejected any non-`Idle` `active_row_read_state` and returned `CursorPoll::Pending`, so `fetch_scroll.rs` fell back to `dbc.runtime.block_on(client.next_row_cursor())`. The async method then immediately hit its own `row_already_positioned` branch and returned `Ok(true)` without touching the transport — one tokio runtime entry per row, with no I/O, purely to acknowledge a row that was already buffered.

This mirrors that acknowledgment synchronously, so a parked row is claimed without entering the runtime. Allocation counts are identical before and after, confirming the cost removed was the runtime transition rather than allocation.

Two regression tests cover the new path: one asserts the sync cursor claims a parked row and still decodes columns in order across a multi-row sequence, the other asserts the sync and async paths behave identically for a parked row under cancellation. Both fail before the fix.

### Measured results

Linux perf lab, ADO run [`175573`](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=175573), against run [`175420`](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=175420) (main at `e8075729`).

`fetch/rowset_100k_c15_fixed/bound_rowset_1` was the only benchmark slower than Microsoft ODBC 18.6.2.1, and is now faster than it:

| | main | this PR |
|---|---:|---:|
| Wall time | 216.0 ms | 159.6 ms |
| vs Microsoft 18.6.2.1 | 1.167x slower | **1.14x faster** |
| vs pinned baseline `c453af2c` | 1.48x | **1.07x** |

`fetch/rowset_20k_c15_fixed/bind_cycle_rowset_1` improves from 1.22x to 1.07x against the pinned baseline. Rowset 64/1000 benchmarks move by ≤2.4%, consistent with removing a fixed per-row cost that larger rowsets already amortize.

The candidate is now faster than Microsoft ODBC 18.6.2.1 on 13 of 14 lab benchmarks.

### Note on the regression gate

Run `175573` still reports ❌ on three confirmed regressions against the pinned baseline (`bound_rowset_1` 7.36%, `bind_cycle_rowset_1` 7.26%, `wide_10k_c600` 18.98%). All three **pre-exist on main** — run `175420` measures them at 1.48x, 1.22x, and 1.24x respectively — and this change reduces two of them substantially. `wide_10k_c600` is untouched by this PR and is tracked separately.

That lab run was measured on a branch that additionally carried the benchmark LOB fix from #568, which is required for the harness to get past its preflight and produce a complete comparison. This PR itself is perf-only and changes a single file.

## Related Issues

Fixes #574

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [x] `cargo btest` passes


shiwanigupta0809 (commented):
## Summary

Unattended automated review of `cafa309cbe0b99db3c391431d5d74b0b563446f3`. The synchronous fast path mirrors the existing asynchronous parked-row handoff, preserves the row state for column decoding, and addresses the runtime-entry regression described in #574. I found no new issues; the main validation and coverage jobs were still pending when reviewed.

- **Correctness:** No findings.
- **Security:** No findings.
- **Tests:** No findings. `sync_cursor_claims_parked_rows_without_advancing` and `sync_cursor_claims_cancelled_parked_rows_like_async` both passed with `cargo nextest`.
- **Performance:** No findings. The change removes only the unnecessary runtime fallback for an already-positioned row, consistent with the linked measurements.
- **API / maintainability:** No findings. No public API surface changes, and the new behavior follows the existing `row_already_positioned` invariant.

Theekshna (approved):

Theekshna (commented):
## Summary

Focused, well-mirrored fix: `try_next_row_cursor` now claims a row already parked by `try_peek_past_current_row` without falling through to the pre-existing `active_row_read_state != Idle` guard (which previously forced `CursorPoll::Pending`, sending `fetch_scroll.rs` into `dbc.runtime.block_on(client.next_row_cursor())` purely to re-acknowledge a buffered row). Confirmed the async `next_row_cursor` has the identical fast path ahead of its own checks, so this is parity with existing behavior, not new semantics. Traced every `row_already_positioned` reset site (lines `889`, `4825`, `7362`, `7749`) and confirmed each one also clears `active_row_read_state`/`current_metadata` in lockstep, so the new early return can't observe a stale state. Verified the caller pattern in `mssql-odbc/src/api/fetch_scroll.rs:1202` matches the PR's description of the runtime-entry regression. No `mssql-odbc` FFI surface is touched, so msodbcsql parity is N/A, and this is an internal perf fix rather than a driver-behavior divergence.

Local `cargo nextest` execution is blocked in this environment by an unrelated toolchain issue (this machine's `ms-prod` rustc backend can't codegen certain AVX512/`pclmulqdq` intrinsics pulled in transitively by `crc32fast` — unrelated to this diff). Verified the two new tests are non-vacuous by reading the pre-existing guard structure instead: without the fast path, `active_row_read_state` is `RowPaused` (not `Idle`) after a peek, so `try_next_row_cursor` would fall into the existing `!matches!(.., Idle) || cancelled` branch and return `Pending` — exactly what both new assertions would catch. CI (`gh pr checks`) shows the Windows/Linux/macOS build+test jobs passing; only the macOS test stage and coverage report were still pending at review time.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 0 |

### Suggestion

1. **Benchmark evidence is not independently verifiable, and the description aggregates several benchmarks without itemizing them.** The ADO build `175573`/`175420` links require authentication I don't have access to in this unattended run, so I could not re-derive the headline `216.0ms → 159.6ms` or `1.48x → 1.07x` figures myself. Separately, "Rowset 64/1000 benchmarks move by ≤2.4%" bundles several distinct benchmark configurations (rowset size × fixed/bound vs bind_cycle) into one aggregate bound, without listing individual before/after numbers the way the two headline benchmarks are. For a perf-only PR, itemizing all 14 lab benchmarks (or linking a table) would let a future reader confirm "13 of 14 faster than msodbcsql" without re-running the lab. Non-blocking — the mechanism change itself is correct and the two new regression tests genuinely guard it.

PR #586, opened 2026-09-17
Reduce ODBC fetch and packet-write overhead

saurabh500:
## Description

Reduce ODBC fetch and parameter-array RPC overhead without changing public APIs, wire semantics, transaction boundaries, packet size, benchmark workloads, or performance thresholds.

- Avoid unnecessary fetch metadata snapshots and binding copies; resume buffered decoding after packet boundaries.
- Fast-path UTF-16 validation and use checked 64-bit datetimeoffset arithmetic.
- Remove boxed packet-write/RPC-encoder futures, compact optional metadata, and batch small wire headers.
- Decode buffered completion tokens directly, skip clock reads for unlimited timeouts, and inline profiled conversion/dispatch paths. Preserve partial-read fallback, cancellation, and finite timeout accounting.

### Performance

[Run 175955](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=175955) passed the unchanged **5% Rust-baseline regression gate** on `f732d02c`.

**The <=5% Microsoft write-gap target is not met:** parameter-array writes measured **20.27 ms vs 18.86 ms (+7.45%)** for Microsoft ODBC 18.6.2.1. This reference is non-gating and uses initial five-sample medians; confirmation rounds compare only against Rust. The Rust write baseline uses sequential execution, not parameter arrays. Two-run confirmation of the Microsoft target remains incomplete.

Fetch gaps versus Rust: wide **+1.34%** after confirmation (0/4 regression rounds), bind/unbind **+3.95%**, variable-width rowsets 1000/64 **-0.49%/-0.77%** (initial measurements for the latter three).

### Validation

- 3,838 core/ODBC library tests passed on Windows; five live ODBC suites passed.
- Workspace and excluded Python bindings formatting/clippy passed.
- Regression coverage includes packet/header boundaries, Unicode, datetimeoffset ranges, RPC batches, partial tokens, truncation, cancellation, and timeouts.
- Full local `cargo btest` is not green: Windows is blocked by an incompatible `cargo-llvm-cov.exe` (OS error 216); an earlier Linux run had 3,832 passes, 372 integration-environment failures, and 24 skips.
- [Cross-platform validation 175956](https://sqlclientdrivers.visualstudio.com/mssql-rs/_build/results?buildId=175956) passed after retrying the macOS job whose agent disconnected during Colima startup. All other jobs passed on their first attempt.
- [Coverage reporting](https://github.com/microsoft/mssql-rs/actions/runs/35179733031) passed after retrying its artifact-wait timeout once the macOS coverage artifact was available. All current PR checks pass; neither retry required code changes.

Further tuning is paused with the Microsoft write-gap target still unmet.

## Related Issues

Related to #247; does not close the broader performance-tracking issue.

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes


Theekshna (commented):
This is an unattended automated run; findings were not checked by a human first.

The unattended sweep found no issues in this PR (protocol buffered-decode fast paths, boxed-future removal, ODBC bound-fetch/UTF-16/datetimeoffset changes, PR description, and evidence audit all checked).

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

shiwanigupta0809 (commented):
## Automated code review

This was an unattended automated run; the review was posted without human confirmation.

Reviewed the current `origin/main` merge-base diff, PR description and linked issue, all existing inline/review/top-level discussion, current checks and coverage, and directly affected code.

### Findings

- **Correctness:** None.
- **Security:** None.
- **Tests:** None.
- **Performance:** None.
- **API/readability:** None.

### Validation

Targeted `cargo nextest` runs passed: 10 `mssql-tds` tests covering packet boundaries, buffered token/row dispatch, cancellation/timeouts, RPC serialization, and metadata sizing; and 3 `mssqlodbc` tests covering UTF-16 validation and checked datetimeoffset arithmetic. All 19 current PR checks pass; reported diff coverage is 98%.

David-Engel (commented):
### 🟡 Automated review — suggestions, nothing blocking

Read the full diff against `origin/main` (merge-base `8cb052ed`) plus the directly affected code. All required checks are green on `f732d02c`, and diff coverage is reported at 98%, so I did not re-run the suites locally.

I specifically re-derived the risky equivalences and they hold:

- `datetimeoffset_parts`: the checked-i64 form rejects exactly the same inputs as the old i128 form. `time_nanoseconds > i64::MAX` and any `checked_add` overflow already imply more than 10<sup>7</sup> days, which is far outside `MAX_DAYS_SINCE_0001`; `days + utc_ticks.div_euclid(TICKS_PER_DAY)` cannot overflow `i64` (max ≈ 4.3e9 + 1.07e7), and `rem_euclid` is always non-negative so the `u64::try_from` never trips.
- `is_valid_utf16le`: the `bytes.is_ascii()` shortcut is sound — every byte ≤ 0x7F means every LE unit is ≤ 0x7F7F, below the 0xD800 surrogate range.
- `write_fixed_bytes`: `has_space(N + 1)` guarantees `position + N < max_payload_size`, so the batched write cannot skip a `handle_overflow_if_needed` that the per-byte loop would have performed. The boundary sweep test covers this well.
- RPC `ProcId` batching and the positional `[0, options]` batching are byte-identical to the previous sequences; `options_written` correctly excludes the streamed and encrypted paths that write the options byte themselves.
- The `buffered_control_token` fast path matches `await_read_or_interrupt`'s existing ordering: cancellation is checked first, and a read that is already ready returns before any timeout is applied. Consuming exactly 13/5 bytes matches the DONE and RETURNSTATUS layouts.
- `fetch_scroll_safe` re-locks STMT only after `desc_state` is dropped, so the documented "never hold STMT while acquiring DESC" ordering is preserved.
- `deliver_fixed_bound`'s new `size_of::<T>()` stride equals `element_stride` for every current call site (BIT/UTINYINT 1, SSHORT 2, SLONG/FLOAT 4, SBIGINT/DOUBLE 8, GUID 16, DATE/TIME 6, TIMESTAMP 16, SS_TIME2 12, SS_TIMESTAMPOFFSET 20).

#### Suggestions

1. **`mssql-odbc/src/api/fetch_scroll.rs` — `deliver_fixed_bound` now depends on an invariant nothing enforces.** Replacing `element_stride(binding.target_type, binding.buffer_length)` with `size_of::<T>()` is correct for every caller today, but a future `write_exact`/`write_temporal` call site passing a `T` that does not match the declared C type would silently mis-stride row-wise delivery across a rowset — no compile error, no panic, just wrong memory. The new doc line states the requirement; consider also enforcing it cheaply: `debug_assert_eq!(std::mem::size_of::<T>(), element_stride(binding.target_type, binding.buffer_length));`.

2. **`mssql-odbc/src/api/fetch_scroll.rs:~1726` — the `deliver_bound_plp` comment is now stale in a load-bearing way.** It still says "`plp_columns` snapshots the statement's copy, whose length also fixes `column_count`, so the lookup is always in range." With the new `current_result_supports_row_into()` short-circuit the vector can be empty, so the lookup is *never* in range on that path. The code is still correct — `metadata_supports_buffered_rows` requires `!column.is_plp()` for every column, so `CursorColumn::PlpStreaming` cannot occur when the snapshot is skipped — but that is exactly the reasoning the comment should now carry, since it is the only thing keeping the `None` arm from becoming a real drain-and-report-unsupported path.

3. **`mssql-tds/src/io/token_stream.rs` — `buffered_control_token`'s byte counts are hand-maintained in two places.** The function returns literal `13` and `5` while `buffered_done_token` independently hardcodes `12` and the RETURNSTATUS arm hardcodes `4`. The tests pin the current values, but deriving the consumed count (for example `1 + DONE_PAYLOAD_LEN`) would make drift impossible rather than merely detectable.

#### Nits

- `#[inline(always)]` on `serialize_value`, `serialize_value_inner`, `bound_param_to_rpc`, `bound_param_to_value_with_outcome`, and `RpcParameter::serialize` is a fairly heavy hammer for large `async fn` bodies; plain `#[inline]` keeps the cross-crate availability while leaving the final decision to LLVM. If the profile data genuinely shows `always` wins, a one-line note on the measurement would help the next person resist "simplifying" it.
- `ordinary_parameters_do_not_inline_streaming_and_encryption_metadata` asserts against `size_of::<SqlType>() + size_of::<Option<String>>() + 6 * size_of::<usize>()`, which is loose enough that it would still pass if one of the two boxes came back. Asserting the two fields are pointer-sized (`size_of::<Option<Box<SqlType>>>() == size_of::<usize>()`) would test the actual intent.

#### Process

The description still says "This PR remains draft" and leaves `cargo btest` unchecked, but the PR is marked ready for review. It also records the `<=5%` Microsoft write-gap target as unmet (`+7.45%`) with two-run confirmation incomplete. Before a human reviewer picks this up, please either return it to draft or refresh the title/description and checklist to match the state you actually want reviewed — right now the description argues against merging the change it describes.

<sub>Automated review. Not an approval, and no change request. Rustfmt/clippy status is taken from CI and not restated here.</sub>

shiwanigupta0809 (commented):
## Automated code review

This was an unattended automated run; the review was posted without human confirmation.

Reviewed the full merge-base diff against the current `origin/main`, directly affected code, PR title/description, linked issue #247, all inline threads, review bodies, top-level comments, coverage, and current checks. Existing same-head suggestions were not duplicated. I found no new issues.

### Findings

- **Blocking:** None.
- **Medium:** None.
- **Low:** None.
- **Suggestion:** None.
- **Nit:** None.

### Validation

All 19 current PR checks pass and reported diff coverage is 98%. No local test was run because no unresolved concern required targeted validation.

shiwanigupta0809:
Unattended automated code review posted: https://github.com/microsoft/mssql-rs/pull/586#pullrequestreview-5239113282

Category counts: Blocking 0, Medium 0, Low 0, Suggestion 0, Nit 0.

saurabh500:
@David-Engel Addressed the non-inlining suggestions from [your review](https://github.com/microsoft/mssql-rs/pull/586#pullrequestreview-5236518457) in 0cff06cf:

- Added the debug-only fixed-width stride assertion; the exact-type safety contract remains documented because equal widths do not establish type equivalence.
- Corrected the PLP comment: skipping the snapshot is safe because that path excludes all PLP columns.
- Shared DONE and RETURNSTATUS payload lengths between parsing and consumed-byte accounting.
- Strengthened the layout test using `size_of_val` on the actual `streamed_declaration` and `encrypted` fields, retaining the overall struct-size budget. Checking an independently spelled `Option<Box<...>>` would still pass if the real field were changed to an inline value.

The description's draft reference was already removed. The local `cargo btest` checkbox remains unchecked deliberately: its coverage executable still fails to start on this host (OS error 216). For this commit, 84 focused tests passed, along with workspace/Python formatting and all-feature/all-target clippy. Remote checks must rerun for the new commit.

Leaving `#[inline(always)]` unchanged for now. A separate local comparison is investigating the five suggested `#[inline]` mutations against `f732d02c`, keeping these maintenance changes out of the experiment. Results will be reported separately; no performance conclusion yet. The documented Microsoft write-gap target remains unmet.

saurabh500:
@David-Engel Follow-up on `#[inline(always)]`: completed a local comparison of `f732d02c` against exactly the five suggested `#[inline]` mutations. **Inconclusive; leaving the annotations unchanged, not claiming `always` is faster.**

Both were Windows x64 release/LTO builds with identical dependency locks, using the existing 2,000-row parameter-array workload against local SQL Server. After warmup, six ABBA blocks produced 180 samples per variant; all workload correctness checks passed.

| SQLExecute wall time | Current `inline(always)` | Five `inline` mutations |
|---|---:|---:|
| Median | 27.5916 ms | 28.0130 ms |
| IQR | 26.2289–30.9908 ms | 26.2744–30.6009 ms |

The pooled median difference is +1.53%, but block results change sign and the six-block t interval is approximately -7.54% to +13.71%. That does not establish equivalence within the preselected +/-2% tolerance, nor a reliable regression. CPU readings were quantized to 15.625 ms and measured the broader main-thread callback, not SQLExecute alone or all runtime threads, so they cannot settle this.

Limitations: shared Windows VM/local Docker SQL scheduling; the benchmark linked directly to the driver DLL after Driver Manager rejected an unregistered absolute DLL path. The workload and timed region were unchanged, but the Driver Manager route was not measured. This is diagnostic evidence, not the official regression gate. A quieter environment and higher-resolution CPU measurement are needed before making an evidence-based inlining change.

saurabh500:
@David-Engel Thanks for the recommendation to prefer `#[inline]` and leave the final decision to the compiler. Applied it to all five functions identified in your review in b80242f6: `serialize_value`, `serialize_value_inner`, `bound_param_to_rpc`, `bound_param_to_value_with_outcome`, and `RpcParameter::serialize`. Updated the README accordingly; other inlining annotations are unchanged.

This supersedes my earlier recommendation to retain `inline(always)`. The local comparison was inconclusive, not proof of equivalence, but it did not establish a benefit sufficient to justify forcing these functions inline. Ordinary inlining is the better default without that evidence.

214 focused conversion/serialization/RPC tests passed, as did workspace and Python-binding formatting and all-feature/all-target clippy. The previously documented local coverage-tool blocker remains; CI and the existing performance gate need to validate this new commit.

mssql-tds/src/connection/transport/network_transport.rs
```diff
@@ -2686,6 +2691,16 @@ impl NetworkTransport {
...
         self.attention_settlement = None;
+        // A ready read also wins over a zero timeout in await_read_or_interrupt.
+        // A pre-cancelled read must take its existing ATTENTION settlement path.
+        if !cancel_handle.is_some_and(|handle| handle.cancel_token.is_cancelled())
+            && let Some((token, consumed)) = crate::io::token_stream::buffered_control_token(
+                self.tds_read_buffer.get_buffered_slice(),
+            )
+        {
+            self.tds_read_buffer.consume_bytes(consumed)?;
+            return Ok(token);
```
David-Engel, line 2705: Nit, non-blocking. This early return skips 
eceive_token_internal, which is the only place that emits `debug!("Received token type: {:?} ({})")`. Since the fast path fires exactly when a DONE/DONEPROC/DONEINPROC/RETURNSTATUS token is already fully buffered — the common case once a result set's rows have been drained out of the same packet — statement completion effectively stops appearing in `RUST_LOG=mssql_tds=debug` output, while rarer tokens still log. That is the token whose status flags and row count are most worth seeing when reconstructing a stream from logs.

If the trace is worth keeping, a `debug!` (or `trace!`) next to `consume_bytes` here restores it for a few instructions on a path that otherwise does no I/O; if dropping it is deliberate, a one-line note here would stop the next reader from "restoring" it. Either way the PR description and README section currently describe the fast path without mentioning the logging change.

David-Engel (commented):

### Summary

Re-reviewed the full merge-base diff (`8cb052ed..b80242f6`) plus the surrounding code after `0cff06cf` and `b80242f6`. All three suggestions and both nits from the previous automated round are addressed: the `debug_assert_eq!` stride guard is in `deliver_fixed_bound`, the `deliver_bound_plp` comment now carries the real "snapshot skipped only when every column is non-PLP" reasoning, `DONE_PAYLOAD_LEN` / `RETURN_STATUS_PAYLOAD_LEN` are shared between parsing and consumed-byte accounting, the layout test asserts on the real fields via `size_of_val`, and the five hot functions moved from `#[inline(always)]` to `#[inline]` with the README updated to match. I found nothing blocking and one optional nit, inline below.

Equivalences I re-derived on this head:

- `buffered_control_token` is byte-for-byte equivalent to the async DONE/RETURNSTATUS parsers, and the consumed counts are now derived (`1 + DONE_PAYLOAD_LEN`, `1 + RETURN_STATUS_PAYLOAD_LEN`) rather than literal. The `receive_token` short-circuit checks cancellation first and only then a fully buffered token, which matches `await_read_or_interrupt`'s existing "cancelled → already-ready read → timeout" ordering, so the zero-timeout and pre-cancelled cases behave as before.
- The `drive_row_columns` buffered path cannot leak ciphertext: `try_decode_buffered_into` returns `None` for `crypto_metadata.is_some()` and for PLP, so encrypted columns still go through `decode_or_decrypt_column`. `get_buffered_slice()` is bounded by `buffer_length`, so stale bytes past the received data cannot make an incomplete value look complete.
- `write_fixed_bytes`'s `has_space(N + 1)` guarantees `position + N < max_payload_size`, so the batched write can never skip a `handle_overflow_if_needed` the per-byte loop would have performed. `write_collated_type_info`, the `ProcId` 6-byte batch and the positional `[0, options]` batch all reproduce the previous byte sequences, and `options_written` correctly excludes the streamed and encrypted paths that write the options byte themselves.
- `datetimeoffset_parts`'s checked-i64 form rejects exactly the same inputs as the old i128 form, and the new test asserts that against an i128 reference over the interesting `days`/`ticks`/`offset` corners.
- Skipping the `plp_columns` snapshot is safe because `current_result_supports_row_into()` requires `!column.is_plp()` for every column and `CursorColumn::PlpStreaming` is derived from the same `is_plp()` over the same COLMETADATA.
- Re-locking STMT for `column_sql_types` only when some binding is `SQL_C_DEFAULT` keeps the documented ordering ("never hold STMT while acquiring DESC") because `desc_state` is dropped first, and `STMT_STATE_FETCH_IN_PROGRESS` is already set so the metadata cannot move underneath it.

### Verification

Read-only. I did not rebuild or re-run the suites: the parent commit `f732d02c` had all 19 checks green with 98% reported diff coverage, and the delta since is confined to a `debug_assert`, comments, shared length constants, one strengthened test, and `#[inline(always)]` → `#[inline]`. Note that ADO validation and the coverage report are still **pending** on `b80242f6` — they have not failed, but they have not confirmed this head either.

### Findings

**Blocking:** none.

**Suggestion:** none.

**Nit:** one, inline — the `receive_token` fast path silently drops the `debug!("Received token type: …")` trace for the DONE family and RETURNSTATUS.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -5244,11 +5244,12 @@ impl TdsClient {
     /// A parked token was already timed and observed when it was first read, so
     /// it is replayed as-is; re-observing it would charge one token's evidence
     /// to the reset acknowledgement twice.
+    #[inline(always)]
```
Vahid-b: `#[inline(always)]` here is added by this PR, and it was not revisited when `serialize_value`, `serialize_value_inner`, `bound_param_to_rpc`, `bound_param_to_value_with_outcome` and `RpcParameter::serialize` moved to plain `#[inline]`.

The reasoning you gave for that move - the local comparison was inconclusive, so it "did not establish a benefit sufficient to justify forcing these functions inline" - applies here unchanged: there is no measurement attached to this annotation either. This is also the less comfortable of the two additions. `next_response_token` is a ~30-line `async fn` whose body holds an `.await` on `receive_token` plus `settle_interrupted_read`, `observe_response_token` and `observe_prepared_batch_done`, and it has seven non-test call sites in this file, so `always` forces that whole state machine into each caller's future instead of letting LLVM decide.

Plain `#[inline]` keeps cross-crate availability with the same argument you already applied elsewhere. If you would rather keep `always`, a one-line note pointing at the measurement would help: the README now says the large functions use `#[inline]`, which reads as though that were the crate-wide rule, so the next reader is likely to "fix" this to match.

(`BoundParam::for_row` is the other addition, and there `always` is easy to defend - small, non-async, pure pointer arithmetic. I would only change this one.)

mssql-odbc/src/api/fetch_scroll.rs
```diff
@@ -992,7 +980,21 @@ fn fetch_scroll_safe(
...
         let mut bindings = ColumnBinding::all_from_ard_state(&desc_state);
         drop(desc_state);
-        resolve_default_bindings(&mut bindings, &column_sql_types, odbc_version);
+        if bindings
+            .iter()
+            .any(|binding| binding.target_type == SQL_C_DEFAULT)
+        {
+            let Ok(stmt_state) = stmt.inner.lock() else {
+                error!("SQLFetchScroll: stmt mutex poisoned resolving default bindings");
+                return SQL_ERROR;
```
Vahid-b, line 990: Nit, non-blocking. The two sibling error exits about 30 lines above - the freed-ARD check and the poisoned-ARD lock - both run `if let Ok(mut stmt_state) = stmt.inner.lock() { stmt_state.clear_state(STMT_STATE_FETCH_IN_PROGRESS); post_sql_error(...) }` before returning `SQL_ERROR`, and the comment above them says why: "clearing STMT_STATE_FETCH_IN_PROGRESS so the statement is not left permanently stuck mid-fetch". This new exit returns `SQL_ERROR` with the flag still set and no diagnostic record posted.

I believe that is unavoidable here rather than a defect: the lock that failed *is* `stmt.inner`, so there is no way to clear the flag or post the record through it, and a poisoned STMT mutex leaves the statement unusable anyway - every later `lock()` fails too. But someone comparing the three exits in this function sees one that looks like it forgot, and this is also one of the two lines the coverage report flags as uncovered (988-989), so no test will ever demonstrate the difference.

A one-line comment would settle it, e.g. "cannot clear STMT_STATE_FETCH_IN_PROGRESS or post a diagnostic here: the poisoned lock is the one we would need."

Vahid-b (commented):
> **Unattended automated review.** These findings were not checked by a human first. This is not an approval and not a change request.

### Summary

Read the full merge-base diff (`8cb052ed..b80242f6`) plus the surrounding unchanged code, the PR description, linked issue #247, every existing review body, inline thread and top-level comment, the coverage report, and `gh pr checks`. Nothing blocking. The equivalences this change depends on hold, and I confirmed by mutation that the riskiest new tests are not vacuous.

### What I checked beyond the already-answered threads

- **The `receive_token` fast-path ordering really is equivalent.** `await_read_or_interrupt` checks cancellation first (`network_transport.rs:139`), then eagerly polls the read and returns it when ready (`:143-146`) *before* the timeout wraps it. The new short-circuit reproduces that order exactly. I also checked the one thing nobody had: the fast path skips the `already_dead` plumbing, but `already_dead` is only consulted in `read_to_attention_boundary`'s *interrupted* arm (`:815`), so a buffered token on a `known_dead` connection returns `Ok` either way.
- **`drive_row_columns`' "no write on `None`" dependency is pre-existing, not new.** Every arm of `try_decode_buffered_into` evaluates its `read!`s before the `writer.write_*` call, and `try_read_buffered_row_prefix_into` on `main` already relies on that same contract, so the new call site does not widen the risk.
- **`write_fixed_bytes` framing.** `handle_overflow_if_needed` always flushes exactly `packet_size` and copies the remainder forward, so the batched write, the per-byte fallback and the original `write_u16`/`write_u32` sequences produce an identical *packet* stream, not merely an identical payload. `has_space(N + 1)` implies `position + N < max_payload_size`, so the batched path cannot skip a flush the per-field path would have performed.
- **The `plp_columns` skip.** `metadata_supports_buffered_rows` (`tds_client.rs:7080-7086`) requires `!column.is_plp()` for every column, so `CursorColumn::PlpStreaming` cannot reach `deliver_bound_plp` on the skipped path. `column_count` comes from `stmt_state.column_metadata.len()`, not from `plp_columns.len()`, so the empty vector does not shrink the fill loop.
- **`element_stride` is documented as column-wise only** (`fetch_scroll.rs:1570`), so swapping in `size_of::<T>()` does not disturb row-wise binding, and the new `debug_assert_eq!` catches drift in test builds.

### Verification

`cargo nextest -p mssql-tds -p mssqlodbc --lib` in a dedicated worktree and target directory. The 11 new/changed tests pass. Mutation results:

| Mutation | Test | Outcome |
|---|---|---|
| `has_space(N + 1)` to `has_space(N)` | `fixed_bytes_preserve_packet_boundaries_and_reset_flags` | fails as it should |
| `bytes.is_ascii()` to an always-true byte check | both `utf16le_validation_*` tests | fail as they should |
| `checked_add` to `wrapping_add` | `datetimeoffset_checked_arithmetic_matches_wide_arithmetic` | still passes, but this is an *equivalent mutant* rather than a coverage gap: any overflow implies true ticks above `i64::MAX`, i.e. more than 10^10 days, and the wrapped value divides to roughly -1.067e10 days, so both forms land outside `0..=MAX_DAYS_SINCE_0001` and return `None`. `checked_add` is correctly defensive rather than load-bearing. |

ADO validation and the coverage report are still **pending** on `b80242f6` - they have not failed, but they have not confirmed this head either. The green results and the 98% diff coverage are from `f732d02c`.

### Findings

**Blocking:** none.

**Suggestion:** one, inline - the two `#[inline(always)]` annotations this PR *adds* were not revisited when the other five were downgraded.

**Nit:** one, inline - the new poisoned-`STMT`-lock exit is the only error path in that block that leaves `STMT_STATE_FETCH_IN_PROGRESS` set.

I deliberately did not re-raise the dropped `debug!` trace on the fast path, the description/checklist state, or the suggestions already addressed in `0cff06cf` and `b80242f6`; those threads exist and are answered.

Theekshna (commented):
This is an unattended automated run; findings were not checked by a human first.

Covered: the full diff since my last review (delta confined to addressing prior suggestions - debug_assert_eq! stride guard, comment fix, #[inline(always)] to #[inline]), msodbcsql-parity relevance (none: internal perf-only, no wire/attribute-table changes), test sufficiency (already mutation-tested on this exact head by David-Engel and Vahid-b), the benchmark claims in the description, PR description/checklist currency, and current checks. The one currently-failing check (mssql-python macOS cross-repo build) was traced to the pipeline's "Install and start Colima-based Docker" step timing out - the same pre-existing infra flake class already described in the PR body, not a regression from this change. No new findings beyond the three already-posted inline items (David-Engel's dropped debug! trace nit, Vahid-b's #[inline(always)] scope suggestion and poisoned-lock nit), which are not duplicated here.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 0 |
| Nit | 0 |

mssql-tds/src/connection/transport/network_transport.rs
saurabh500: Thanks for catching this. Fixed in 1d8ae153: both paths now use the same logging helper in token_stream, preserving the DEBUG level, original module target, and existing token-type/numeric-ID message. The buffered path uses the wire token identifier, so DONEPROC and DONEINPROC retain their distinct names. Formatting stays inside the debug macro.

Added a regression test covering all four completion tokens with no buffered data, a packet-split payload, and a fully buffered token. Each must emit the received-token event exactly once under the original module-specific debug filter. Six focused completion/cancellation/timeout tests passed, along with workspace and Python-binding formatting/clippy. No new status/row-count logging or intentional logging behavior change.

mssql-odbc/src/api/fetch_scroll.rs
saurabh500: Added the one-line explanation in 3beb9b58. Clearing the flag and posting a statement diagnostic both require the poisoned STMT lock. Kept the error log and SQL_ERROR behavior unchanged, following the repository's policy not to recover poisoned mutexes.

mssql-tds/src/connection/tds_client.rs
saurabh500: Agreed on preferring compiler discretion here too. Changed next_response_token to #[inline] in 3beb9b58; BoundParam::for_row remains unchanged. The earlier local experiment covered only the five previously named functions, so it provides no isolated evidence for this annotation. This is a default-policy choice, not a measured performance improvement or a claim of equivalence. 64 focused response/batch/timeout tests passed, plus workspace and Python-binding formatting/clippy.

David-Engel (approved):

David-Engel (commented):

### Summary

Re-reviewed the full merge-base diff (`8cb052ed..3beb9b58`) after the force-push, plus the surrounding unchanged code, the PR description, the linked issue #247, every prior review body and inline thread, and current checks. All three earlier threads (buffered-token debug logging, `#[inline(always)]` consistency, and the poisoned-STMT-lock exit in `SQLFetchScroll`) are resolved on this head, and I did not re-raise them.

No blocking issues, suggestions, or nits this round.

### Verification

The `receive_token` buffered fast path mirrors `await_read_or_interrupt` exactly: it checks pre-cancellation first, and a ready read already won over `Some(Duration::ZERO)` via the eager first poll, so the only prior behavior it could have changed is the `already_dead` branch — which is reachable only on an interrupted read, not a completed one. `get_buffered_slice()` is bounded by `buffer_position..buffer_length`, which `reset_to_length` resets per packet, so a 12-byte DONE or 4-byte RETURNSTATUS can never be assembled across a packet header, and `try_read_slice`'s `do_we_have_enough_data` uses the same window as `get_buffered_slice`, so the consume step cannot disagree with the peek.

The `drive_row_columns` buffered branch stays correct because `try_decode_buffered_into` returns `Ok(None)` for PLP and for `crypto_metadata.is_some()`, which is exactly the set `decode_or_decrypt_column` treats specially; NBCROW nulls, skips, and the PLP `stop_here` pause are all handled before it, and readers that only implement the default `buffered_slice()` keep the async path.

`write_fixed_bytes`'s `has_space(N + 1)` is the right bound: `handle_overflow_if_needed` flushes at `position() >= max_payload_size`, so `pos + N + 1 <= max` guarantees the fast path never lands on or past a boundary that the byte-wise path would have flushed, and `handle_overflow_if_needed` is a no-op otherwise, so no cancellation or timeout check is skipped. `PROC_ID_SWITCH`/proc id/`proc_options` and the collated `TYPE_INFO` batches all match the little-endian layout the replaced `write_u16_async`/`write_i16_async`/`write_u32_async` calls emitted. In `RpcParameter::serialize`, `options_written` is only set for positional, unencrypted, materialized parameters — precisely the case where nothing is emitted between the zero name-length byte and the options byte.

Skipping the `plp_columns` snapshot is safe because `metadata_supports_buffered_rows` requires every column to satisfy `!is_plp()`, so the snapshot would have been all-`None` anyway and `plp_columns.get(..).flatten()` yields the same answer; `column_count` comes from its own STMT-lock snapshot, not from `plp_columns.len()`. Gating `resolve_default_bindings` on an `SQL_C_DEFAULT` binding is equivalent since that function early-`continue`s on every other target type, and the new STMT lock is taken only after the ARD guard is dropped, so the "no STMT lock while acquiring a DESC lock" rule still holds. `datetimeoffset_parts` narrows to `i64` without losing rejections: a `time_nanoseconds` beyond `i64::MAX` already exceeds 10.6M days and the widest `i16` offset moves it by under 23 days, so `checked_add`/`try_from` returning `None` matches what the old `i128` day-range check rejected. `is_valid_utf16le`'s ASCII shortcut is sound because a surrogate code unit always has a high byte of at least `0xD8`, and `deliver_fixed_bound`'s `size_of::<T>()` stride is reached only through the two `binding.target_type == target_type` guards.

Verification was by reading. Every required check is green on `3beb9b58` (ADO validation across Windows/Linux/macOS/ARM, the cross-repo mssql-python suites, Kerberos, and merged coverage), with only `coverage-report` still pending, so I did not spend the review budget on a cold local rebuild.

### Findings

**Blocking:** none.

**Suggestion:** none.

**Nit:** none.

Nothing to change from my side; this looks ready for human review, but this comment is not an approval.

Vahid-b (approved):

PR #590, opened 2026-09-17
Fix autocommit outstanding request count

### LINKED ISSUE #587: mssql-tds sends increasing OutstandingRequestCount in autocommit mode, causing SQL Server error 3981
### Describe the bug

`mssql-tds` generates an increasing `OutstandingRequestCount` in the TDS
Transaction Descriptor header for successive autocommit requests. After one or
more successful requests, a later request that starts or completes a SQL Server
transaction can fail with SQL Server error 3981:

```text
The transaction operation cannot be performed because there are pending requests working on this transaction.
```

This occurs even when:

- requests are executed sequentially;
- Multiple Active Result Sets (MARS) is disabled;
- every response is fully consumed with `close_query().await`; and
- no application-level transaction is active before the failing request.

The MS-TDS Transaction Descriptor Header specification says:

> The TransactionDescriptor MUST be 0, and OutstandingRequestCount MUST be 1 if the connection is operating in AutoCommit mode.

Reference:
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/4257dd95-ef6c-4621-b75d-270738487d68

The current implementation instead uses a process-wide monotonically
increasing counter:

```rust
// Static counter for non-transaction request count
static NON_TRANSACTION_REQUEST_COUNT: AtomicU32 = AtomicU32::new(0);

pub fn create_non_transaction_header() -> Self {
    let count = NON_TRANSACTION_REQUEST_COUNT.fetch_add(1, Ordering::SeqCst);
    Self::new(0, count + 1)
}
```

Source at the time of reporting:
https://github.com/microsoft/mssql-rs/blob/aa2ce0e21bac3a44a901e022c9d7b78fbc787742/mssql-tds/src/message/headers.rs#L11-L58

Because this counter is static, its value increases across unrelated commands
and even across separate connections in the same process. SQL Server then
receives a value greater than `1` for an autocommit request and treats the
request as though other requests are pending.

### Steps to reproduce

Create a harmless procedure that opens and commits a transaction:

```sql
CREATE OR ALTER PROCEDURE dbo.mssql_tds_transaction_repro
AS
BEGIN
    SET NOCOUNT ON;

    BEGIN TRANSACTION;
    COMMIT TRANSACTION;
END;
GO
```

Using a normally connected `TdsClient` with the default MARS setting, execute
and drain at least one autocommit request before executing the procedure:

```rust
use mssql_tds::{
    connection::tds_client::{ExecuteOptions, TdsClient},
    message::parameters::rpc_parameters::RpcParameter,
};

async fn reproduce(client: &mut TdsClient) {
    let no_parameters: Vec<RpcParameter> = Vec::new();

    // First autocommit request: OutstandingRequestCount is sent as 1.
    client
        .execute_sp_executesql(
            "SELECT 1;".to_owned(),
            no_parameters,
            ExecuteOptions::new(),
        )
        .await
        .unwrap();
    client.close_query().await.unwrap();

    // Second autocommit request: OutstandingRequestCount is sent as 2.
    // SQL Server error 3981 is produced when the procedure performs its
    // transaction operation.
    let result = client
        .execute_sp_executesql(
            "EXEC dbo.mssql_tds_transaction_repro;".to_owned(),
            Vec::new(),
            ExecuteOptions::new(),
        )
        .await;

    // Drain the response even on a server-reported error.
    let drain_result = client.close_query().await;

    println!("execute result: {result:?}");
    println!("drain result: {drain_result:?}");
}
```

The same problem was observed in an application that sequentially performed:

1. connection and identity queries;
2. a candidate-selection query;
3. several parameterized `DELETE` commands;
4. an audit insert; and
5. a parameterized `EXEC` of a stored procedure containing
   `BEGIN TRANSACTION` and `COMMIT TRANSACTION`.

Every operation called `close_query().await` before the next operation. The
stored procedure nevertheless failed with:

```text
The transaction operation cannot be performed because there are pending requests working on this transaction.
```

The application-level error showed number 50000 only because the stored
procedure caught the original error text and rethrew it with `RAISERROR`. The
message corresponds to SQL Server database engine error 3981:

https://learn.microsoft.com/en-us/sql/relational-databases/errors-events/database-engine-events-and-errors-3000-to-3999

### Expected behavior

When the connection is in autocommit mode, every request should send:

```text
TransactionDescriptor = 0
OutstandingRequestCount = 1
```

Sequential commands whose prior responses have been completely drained should
not be treated as concurrently outstanding requests. The stored procedure
should begin and commit its transaction successfully.

### Actual behavior

Each autocommit request increments `NON_TRANSACTION_REQUEST_COUNT`, causing the
headers to contain values `1`, `2`, `3`, and so on. A later command that uses a
transaction fails with SQL Server error 3981, even though earlier responses
were drained and no request is actually pending.

### Version

mssql-tds 0.1.0 / main

### Affected crate

mssql-tds

### Environment

- OS: Windows Server 2025 Datacenter, build 26100, x64
- Rust: `rustc 1.98.0 (88d9e12ae 2026-08-18)`
- Rust target: `x86_64-pc-windows-msvc`
- SQL Server 2022
- MARS: disabled (`ClientContext` default)
- Query API: `TdsClient::execute_sp_executesql`

### Additional context

The likely correction for an autocommit header is to always use an outstanding
request count of `1`:

```rust
pub fn create_non_transaction_header() -> Self {
    Self::new(0, 1)
}
```

The static `NON_TRANSACTION_REQUEST_COUNT` would then be unnecessary.

A regression test could create two or more non-transaction headers and assert
that every header contains:

```text
transaction_descriptor = 0
outstanding_request_count = 1
```

An integration test could also execute several sequential autocommit commands,
fully drain each response, and then run `BEGIN TRANSACTION; COMMIT TRANSACTION;`
or a stored procedure containing those statements.

kzh125:
fix https://github.com/microsoft/mssql-rs/issues/587


kzh125:
Hi @saurabh500 @David-Engel  
I’ve created this PR and would appreciate it if you could take a look when you have a chance. Thanks.

saurabh500 (commented):
This review comes from an unattended sweep; no human checked it before posting.

**Verdict: clean — no findings. This is not an approval.**

The constructor now returns transaction descriptor `0` and outstanding request count `1` on every autocommit call, matching [MS-TDS 2.2.5.3.2](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-tds/4257dd95-ef6c-4621-b75d-270738487d68). I checked the SQL batch and RPC construction paths, including the first command of an RPC batch; the nonzero transaction-descriptor branches remain unchanged.

**Blocking:** None.

**Suggestion:** None.

**Nit:** None.

The performance pass found no substantiated issue in the affected header paths: the change removes the process-wide atomic increment without adding work to serialization. No timing improvement is claimed. Independent read-only critique raised no corrections.

This is a source-based assessment, not a live reproduction of error 3981. The successive-call test assertions were inspected; their effectiveness was not demonstrated by an executed mutation test.

saurabh500:
Thanks @kzh125 
We will review. Out of curiosity, what are you building on mssql-tds?

saurabh500:
Can you add an integration test? Your bug repro should be a good one.

shiwanigupta0809 (commented):
This review comes from an unattended automated run; no human checked it before posting.

## Summary
The outstanding-request accounting change matches the autocommit response lifecycle, and the targeted regression test fails when the expected count is deliberately changed.

- **Blocking: None.**
- **Medium: None.**
- **Low: None.**
- **Suggestion: None.**
- **Nit: None.**

Targeted validation: the affected `cargo nextest` selection passed, the count mutation was caught, and `cargo fmt -- --check` passed. `license/cla` is green; Azure Pipelines has not run and indicates an authorized `/azp run` may be required.

kzh125:
Hi @saurabh500 

I'm currently using mssql-tds to build standalone CLI tools that need to query SQL Server databases. I may also use it in other applications in the future.

I have implemented the integration tests.

@microsoft-github-policy-service agree

mssql-odbc/tests/e2e/tests/transaction_test.cpp
```diff
@@ -279,6 +279,38 @@ TEST_F(TransactionLiveTest, AutocommitInsertIsImmediatelyDurable) {
...
+// the wire (MS-TDS 2.2.5.3.2). A stale, ever-incrementing count previously
+// made a later request that begins/commits its own transaction fail with SQL
+// Server error 3981 ("The transaction operation cannot be performed because
+// there are pending requests working on this transaction"), even though every
+// prior response had been fully drained and autocommit was in effect the
+// whole time. Must pass identically on msodbcsql, which never had this bug.
+TEST_F(TransactionLiveTest, SequentialAutocommitQueriesToleratePriorTransactionProc) {
+    ExecDirectIgnoreError("DROP PROCEDURE IF EXISTS dbo.mssql_rs_587_repro");
+    Exec(
+        "CREATE PROCEDURE dbo.mssql_rs_587_repro AS "
```
shiwanigupta0809: **Low:** Please use a local temporary procedure (for example, `#mssql_rs_587_repro`) instead of a fixed `dbo` object. This name is shared by every session using the test database, so concurrent e2e/driver-comparison runs can drop or recreate it underneath this test; an assertion before the final cleanup also leaves it behind. The same file already uses session-scoped temp procedures such as `#rollback_proc`, which are automatically removed when the connection closes.

shiwanigupta0809 (commented):
This review comes from an unattended automated run; no human checked it before posting.

## Summary
The autocommit header fix matches MS-TDS 2.2.5.3.2, and the new unit, wire-capture, and mock-server tests all fail when the count is deliberately changed from `1` to `2`. One test-isolation issue remains in the live ODBC regression test.

- **Blocking:** None.
- **Medium:** None.
- **Low:** The live test creates a fixed persistent procedure that can collide across concurrent runs and survive early failures.
- **Suggestion:** None.
- **Nit:** None.

Targeted validation: `cargo clippy --offline -p mssql-tds -p mssql-mock-tds --all-targets -- -D warnings`; six focused unit tests; the mock-server integration test; and a mutation run where all three regression layers failed as expected. GitHub issue #587 is linked and matches the change. Head checks observed: CodeQL/check/license passed, coverage skipped, Analyze still running.

shiwanigupta0809:
Unattended automated code review posted: https://github.com/microsoft/mssql-rs/pull/590#pullrequestreview-5239183048

Category counts: Blocking 0, Medium 0, Low 1, Suggestion 0, Nit 0.

mssql-odbc/tests/e2e/tests/transaction_test.cpp
saurabh500: Good catch — switched to a session-scoped `#mssql_rs_587_repro` temp procedure, matching the `#rollback_proc` pattern already used in this file. Dropped the manual `DROP PROCEDURE` cleanup since temp objects are removed automatically when the connection closes. Verified the test still passes with the fix in place. Pushed in 62c7e277.

saurabh500: Resolving — fix pushed in 62c7e277.

saurabh500 (approved):

David-Engel (approved):

PR #599, opened 2026-09-17
Fix prepared handle cleanup before streamed execution

### LINKED ISSUE #598: Streamed parameter rebind leaks prepared handles and panics after repeated cancellation
### Describe the bug

A reused ODBC prepared statement retains an old pending unprepare through streamed data-at-execution. A completed stream creates another prepared handle; a later rebind attempts to overwrite the occupied cleanup slot. This requires neither concurrency nor a large workload.

### Steps to reproduce

On one connection, prepare an INSERT into a temporary table with a varchar(max) parameter once. Reuse the same statement for four cycles:
1. Bind input SQL_C_CHAR / SQL_VARCHAR, ColumnSize 0, BufferLength 0, stable token and SQL_DATA_AT_EXEC indicator.
2. SQLExecute and SQLParamData return SQL_NEED_DATA. Upload 65,537 deterministic nonuniform ASCII bytes in 8,191-byte chunks.
3. SQLCancel before the final SQLParamData; check that no row was inserted.
4. Execute/upload again and finalize with SQLParamData. Check one row and the complete payload, close/drain, truncate the table, then SQLFreeStmt(SQL_RESET_PARAMS).
5. Repeat without preparing again or recreating the statement.

### Expected behavior

All cycles succeed, cancelled attempts insert nothing, completed attempts preserve the full payload, and superseded prepared handles are released rather than accumulating until disconnect.

### Actual behavior

Observed by the initial isolated stress investigation at the commit below: the third bind in debug panics at handles/stmt.rs (`orphan_prepared_handle: a pending unprepare already exists`). The FFI boundary catches it and SQLBindParameter returns SQL_ERROR without an ODBC diagnostic. Release completes eight cycles but logs six `overwriting a pending unprepare — handle leaked until disconnect` messages; that is defective, not a pass.

A standalone regression and independent before/after validation are being added.

### Version

aa2ce0e21bac3a44a901e022c9d7b78fbc787742

### Affected crate

mssql-odbc

### Environment

Windows x64, non-administrator; Cargo 1.95.0 / MSVC 19.44; SQL Server 2025 container 17.0.4085.5.

### Additional context

The initial reduced four-cycle scenario passes Microsoft ODBC Driver 18, SQL_DRIVER_VER=18.06.0002. This is evidence for this scenario/build only. A broader reference mixed-stress run had a separate, unclassified rows-fetched assertion failure; no full reference-parity claim is made.

Suspected ownership path: orphan_prepared_handle -> SQLExecute streamed branch -> parked DAE orphan -> take_dae restores pending_unprepare. The lower-level begin_execute_prepared contract leaves orphan release to its caller.

saurabh500:
## Description

Fix the third-rebind panic and prepared-handle leak after streamed execution is cancelled and retried. Release the old handle before opening a stream; retain ownership if cancellation prevents sending, and retire the connection after an interrupted partial write.

### Validation

- After merging `main`: formatting, clippy and **2,009 unit tests** pass.
- Before the merge: debug/release/reference-driver regressions passed; CI passed on `0b0bee43` with **99% diff coverage**. Copilot feedback resolved. These CI results do not cover the merge commit.
- Original local full suite: **3,925 passed / 405 failed / 14 skipped** (390 missing-password, 12 missing-certificate, 3 TLS/login failures). Setup corrections cleared 27 cases; 378 were not rerun. Logs retained.

## Related Issues

Fixes #598

## Checklist

- [x] `cargo bfmt` passes
- [x] `cargo bclippy` passes
- [ ] `cargo btest` passes

Kept draft as requested.


mssql-tds/src/connection/tds_client.rs
```diff
@@ -3914,9 +3915,33 @@ impl TdsClient {
...
+                    orphan_id,
+                    true,
+                    ExecuteOptions {
+                        timeout: timeout_sec,
+                        ..opts.clone()
+                    },
+                )
+                .await;
+            if !self.prepared_handles.contains_key(&orphan_id) {
+                *orphaned = None;
```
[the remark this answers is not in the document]
saurabh500: Confirmed with a failing zero-byte cancellation test and fixed in 0b0bee43: cleanup now retains the orphan, server-handle map entry, and AE metadata until a send actually starts; batch cleanup shares the same rule, while an interrupted partial write consumes ownership and retires the uncertain connection without more writes. Repeated pre-send cancellation/retry, cancellation during writer creation, partial-write cancellation, successful release and server-error controls pass (1,983 affected unit tests plus both live regressions in debug/release); removing the injected partial bytes makes the new test fail as intended.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3633,6 +3635,21 @@ impl TdsClient {
         Ok(())
     }
 
+    async fn unprepare_orphan(
+        &mut self,
+        orphaned: &mut Option<StatementId>,
+        options: ExecuteOptions<'_>,
+    ) -> TdsResult<()> {
+        let Some(id) = *orphaned else {
+            return Ok(());
```
Theekshna, line 3644: [Suggestion] Both call sites of this private helper (	ds_client.rs:3832 and :3941-3943) already guard with if orphaned.is_some() before calling it, so this let Some(id) = *orphaned else { return Ok(()); } branch can't be reached today — it's exactly the one line the diff-coverage report shows as missing (99%, 1 line). Since it's private and both callers pre-check, either drop the redundant guard (let a caller misuse show up as a panic/logic bug instead of silently no-op'ing) or, if you want it kept as a safety net for a future third caller, swap it for a debug_assert!(orphaned.is_some()) so the coverage gap goes away and the intent (	his should never actually be reached) is explicit rather than implicit.

Theekshna (commented):
Unattended hourly review sweep. Checked out at `11eff55a` (merge-base `b758050e` against `main`, 5 files, +657/-51). Read all prior inline comments, reviews, and issue comments first — Copilot's earlier bot review flagged an unsent-cancellation handle leak on `execute_sp_unprepare`, and the author's reply plus the current diff show it fixed and mutation-tested (a failing zero-byte-cancellation test that a bytes-removed control makes pass): `message.send_attempted()`/`nothing_sent()` on `SuspendedMessage` distinguish "definitely never touched the wire" from "began writing, fate unknown," and I independently verified the mechanism against `tokio-util 0.7.19`'s `RunUntilCancelledFuture::poll` (biased to the future; an *already*-cancelled token short-circuits before the future is polled at all, so `send_attempted` staying `false` really does prove nothing was sent). That thread is resolved and I did not re-file it.

**Local build note:** `cargo nextest run -p mssql-tds --lib` fails to even compile in this sandbox (`E9012`/`E9013` codegen errors in `crc32fast`'s AVX-512/vpclmulqdq path against this toolchain, reproduced both with and without `-C target-feature=-avx512f,...`) — the same pre-existing environment problem noted on PR #620's review, unrelated to this diff. I relied on the Code Coverage Report bot comment (99% diff coverage, 1 missing line) plus static/call-site analysis instead of a local mutation pass. `gh pr checks` is still fully pending on this exact head (`11eff55a`, pushed ~2h ago after merging `main`); no failures yet, nothing to flag there, and I'm not triggering or waiting on it further.

**msodbcsql parity (check 1):** the reference tree's `DropPrepHandle` (`sqlcfunc.cpp:730-763`) and the piggyback sites in `sqlccmd.cpp:1642-1660` and `:8244-8262` piggyback a **same-statement** superseded handle onto the *next* `sp_prepare`/`sp_prepexec`'s by-reference `@handle` parameter — including for DAE/cursor-open RPCs (`sqlccmd.cpp:6722-6726`). I initially read this PR's new standalone `sp_unprepare`-before-stream as a divergence from that (an extra round trip), but `@handle` is a single by-reference slot carrying one statement's own old→new handle; it has no second slot for an *unrelated* statement's orphaned handle. Grepping this crate, `orphaned: &mut Option<StatementId>` is exactly that cross-identity case (a handle superseded by a *different* `PreparedStatement` id), and it is never piggybacked anywhere in this codebase, streamed or not — `execute_prepared_batch_inner`'s existing `None =>` arm (`tds_client.rs:3830-3832`, unchanged by this PR) already released a same-shaped `orphaned` via a standalone call before this PR touched the streamed path. So this isn't a new or streamed-specific divergence; it's the pre-existing convention extended consistently. No finding.

**Six-check accounting:** (3) Divergences — N/A per check 1, nothing to add to `.github/instructions/mssql-odbc.instructions.md`. (4) Description — see Nit below; linked issue #598 matches the diff's intent, checklist's unchecked `cargo btest` is honest given CI is still pending. (5) Verbose slop — none; new doc comments explain an invariant or the ownership contract (e.g. `send_attempted`'s doc, `unprepare`'s retry-safety note), and the e2e tests' `Benefits-from-mock-tds:` framing matches an existing convention (call_routing_test.cpp, describe_param_test.cpp, etc.), not new filler. (6) Evidence — re-derived the diff-coverage claim directly against the Code Coverage Report comment: 99%, exactly one missing line (`tds_client.rs:3644`), which is the same line my Suggestion below is about. I could not re-run the description's local full-suite numbers (2,009 unit tests; 3,925/405/14) — that requires a live SQL Server and local fixtures this sandbox doesn't have — so I'm not disputing them, only noting they're unverified here; 309 `#[test]`/`#[tokio::test]` functions in `tds_client.rs` alone make the total plausible.

| Severity | Count |
|---|---|
| Blocking | 0 |
| Suggestion | 1 |
| Nit | 1 |

[Nit] The description's checklist section ends with "Kept draft as requested," but this PR is not a draft (`isDraft: false`) and hasn't been since it was opened in this state. Leftover text like this reads as stale/contradictory to a reviewer checking currency — worth deleting.

This is an unattended automated review; findings may be incomplete or wrong, push back on anything that looks off.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3914,9 +3931,27 @@ impl TdsClient {
...
+        if orphaned.is_some() {
+            self.unprepare_orphan(
+                orphaned,
+                ExecuteOptions {
+                    timeout: timeout_sec,
+                    ..opts.clone()
+                },
+            )
+            .await?;
+            let remaining = Self::deduct_timeout(opts.timeout, started.elapsed()).into_timeout()?;
```
Theekshna, line 3950: **Blocking:** this re-charge makes `SQL_ATTR_QUERY_TIMEOUT = 1` fail deterministically whenever there is a live orphan to release.

`deduct_timeout` rounds elapsed **up** to whole seconds (`tds_client.rs:1446-1451`, `saturating_add(subsec_nanos() > 0)`), so any nonzero elapsed charges a full second. With a 1s budget this `into_timeout()?` is guaranteed `Exhausted` even when the release took microseconds — the user's statement never reaches the wire and `SQLExecute` returns a timeout for work that did no wire waiting.

Confirmed, not inferred — throwaway test on the mock transport (live orphan, `timeout_secs(1)`):

```
PROBE timeout_secs(1): Err(TimeoutError(String("command timeout exhausted")))
```

This is the hazard we already solved a layer up; `deduct_query_timeout` truncates *down* and says why (`exec_common.rs:562-572`): "Rounding that up would charge a full second against the budget for every step regardless of whether it touched the network, spuriously exhausting a small timeout (e.g. `1`) before any wire wait ever happened."

Pre-PR this was unreachable on the streamed path — the `started`/re-deduct block is new. `:3831` has the same shape for batch, but `SQLExecute`-with-DAE is the path applications actually hit.

Fix either way: truncate down as `deduct_query_timeout` does, or don't charge cleanup to the statement's budget at all (msodbcsql's `DropPrepHandle` bills `GetQueryTimeOut(lpstmtDvr)`; `flush_pending_unprepare` passes the same timeout to each step rather than a shrinking remainder). Needs a regression at `timeout_secs(1)` with a live orphan.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3871,17 +3887,18 @@ impl TdsClient {
...
-    /// The request stays open for the whole application-driven chunk sequence
-    /// and may be discarded server-side, which would evict the orphan from the
-    /// client's map while the server still held the plan. The caller keeps it
-    /// and releases it with [`unprepare`](Self::unprepare).
+    /// Unlike [`execute_prepared`](Self::execute_prepared), a streamed request
+    /// releases `orphaned` in a separate `sp_unprepare` before opening the RPC,
+    /// charging both operations to the command timeout. Piggybacking the drop
+    /// would let cancellation discard it while the server still held the plan.
+    /// An error before release leaves the orphan with the caller; after the
+    /// release send boundary it is consumed, even if the response fails.
```
Theekshna, line 3895: The behavior is documented here, but the departure from msodbcsql belongs in `.github/instructions/mssql-odbc.instructions.md` under "Deliberate deviations".

msodbcsql defers by default — `DropPrepHandle` (`sqlcfunc.cpp:732-763`), "defer dropping the sproc to save a round trip" — stashing into `hPrepDropDeferred`, which `BuildSPPrepExec` piggybacks onto the by-ref `@handle` (`sqlccmd.cpp:8244-8263`). On the DAE route that slot clears only *after* `SendRPCFromStmt` completes the whole request, in `ProcessDAEParam` (`sqlccmd.cpp:6717-6728`). So a cancelled DAE sequence keeps the handle for the next prepare at no extra round trip, and `VOID INTFUNC DropPrepHandle` structurally cannot fail the application.

Here a standalone `sp_unprepare` can fail the caller's execute: `execute.rs:429-437` maps the `Err` to `fail_with_tds`, and `begin_execute_prepared_release_server_error_preserves_connection_reuse` pins an 8179 about a handle the app never touched surfacing as the execute's diagnostic.

To be explicit: I think propagating is *correct* — swallowing would retain the orphan, completing the stream would mint a second handle, and the next rebind would hit the single `pending_unprepare` slot (`stmt.rs:1397-1415`), reviving #598. It meets both bars for an entry: an application can tell the difference, and the rule spans three functions. Record (a) the extra round trip vs the piggyback and (b) that the single-slot `pending_unprepare` forces propagation, so nobody "fixes" it back into a piggyback.

Note I reach the opposite conclusion from @Theekshna's sweep above, which read the same sites and called this no-finding. Their point that a standalone release is pre-existing convention (`:3831`) is right; my disagreement is narrower — it is the *failure propagation into the application*, not the standalone call, that an application can observe.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -15928,39 +15977,443 @@ mod tests {
...
+        for orphaned in [None, Some(sid(77))] {
+            let (mut client, sent) = create_capturing_client(vec![]);
+            let mut orphaned = orphaned;
+            let mut statement = PreparedStatement::new("INSERT INTO t(v) VALUES (@v)");
+            client
+                .begin_execute_prepared(
+                    &mut statement,
+                    vec![streamed_varbinary("@v")],
+                    &mut orphaned,
+                    ExecuteOptions::new().timeout_secs(1),
```
Theekshna, line 16070: This test is one variable away from catching the timeout bug I flagged on `:3950`.

It uses `timeout_secs(1)` — the exact budget that breaks — but only on the two arms where nothing is released (`None`, and an orphan absent from `prepared_handles`, cleared by the `is_some_and` guard). Both skip the `deduct_timeout` re-charge. Every arm that *does* release runs on the default unlimited budget, where `deduct_timeout` returns `None` and can never exhaust.

So the small budget and the release are each covered; their intersection — the only failing combination — is not. A third arm with a live handle in `prepared_handles` plus `timeout_secs(1)` would pin it. Worth noting this is why 99% diff coverage missed it: the gap is an untested *combination*, not an untested line.

Minor, separate: I could not reproduce the description's "2,009 unit tests" — `-p mssql-tds --lib` on this head gives 2,210 run / 2,203 passed (7 pre-existing `certificate_validator` / `win_tls` fixture failures that also fail on the merge base). Which invocation produced 2,009?

Theekshna (commented):
Reviewed at `11eff55a` (merge-base `b758050e`). Read the prior Copilot and @Theekshna reviews first; not re-filing the `:3644` guard or the draft-text nit. One blocking finding, reproduced locally, plus two suggestions. Mechanism itself checks out — I mutation-tested the `send_attempted` guard, the `retire_without_writing` call, and the set-point of `send_attempted` relative to the await; each mutation fails a test, so the new tests are not vacuous.

Vahid-b (commented):
_This is an unattended automated review. Findings were not checked by a human first — push back on anything that looks wrong._

## Summary

Reviewed at `11eff55a` (merge base `b758050e` against current `origin/main`, 5 files, +657/−51). The change is coherent and the mechanism holds up: `execute_sp_unprepare` stops evicting `prepared_handles` / `prepared_param_encryption` up front and instead evicts them only once `SuspendedMessage::send_attempted()` proves bytes were handed to the writer, retiring the connection when a send started but no packet flushed. `begin_execute_prepared` then releases a superseded handle in a standalone `sp_unprepare` before parking the cancellable stream, so the orphan is no longer carried into `DaeState` and restored by `take_dae` into an already-occupied `pending_unprepare` — which is exactly the `orphan_prepared_handle: a pending unprepare already exists` panic in #598. I traced that path end to end (`stmt.rs:1398` → `:1430`, `execute.rs:392-440`) and the fix matches the reported repro.

No blocking findings from me. One suggestion concerns the sibling release path the PR deliberately did not change, and I confirmed it with a probe rather than by reading.

I did not re-file the four findings already open from @Theekshna (`:3644` redundant guard, `:3950` `deduct_timeout` re-charge, `:3895` deviation-registry entry, `:16070` test gap) or the stale "Kept draft as requested" line. On `:3950` I independently confirmed the arithmetic is real — `deduct_timeout` (`tds_client.rs:1446-1451`) rounds elapsed **up** to whole seconds, and `CommandTimeoutBudget::Exhausted` maps to `TimeoutError` in `into_timeout` (`:1449`, `:1155`) — so with `timeout=1` any nonzero sub-second elapsed exhausts the budget. Worth noting that the same re-charge already ships on the materialized path (`execute_prepared_batch_inner`, unchanged by this PR), so it is a pre-existing shape this PR extends rather than introduces.

## Verification performed

- **Repos/policy.** Fetched `origin/main` (`120eab88`) and `refs/pull/599/head`; merge base `b758050e`. Read `.github/copilot-instructions.md`, `.github/instructions/pr-workflow.instructions.md`, `.github/instructions/mssql-odbc.instructions.md`, and `.github/skills/code-review/SKILL.md` + `posting.md` **from `origin/main`**, not the checkout. Note: the brief for this run said the code-review skill file was absent — it is in fact present at `origin/main:.github/skills/code-review/SKILL.md`, so I applied it.
- **Prior discussion.** Read all three endpoints (`pulls/599/comments`, `pulls/599/reviews`, `issues/599/comments`) plus every review thread and its resolution state before drafting. Read issue #598 in full.
- **Tests run locally** (`CARGO_TARGET_DIR` redirected; main worktree cache untouched):
  `cargo nextest run -p mssql-tds --lib -E 'test(orphan) + test(unprepare) + test(send_attempt) + test(retract_partial)'` → **24 passed, 0 failed**, including all six new ownership tests and `io::packet_writer::tests::send_attempt_survives_suspend_and_resume`.
- **Probe (the basis for the Suggestion below).** I added a temporary test driving `execute_sp_prepexec_for_test` with `cancel_on_writer_creation` — the same "cancelled before the writer's first send" hook this PR adds — and printed the resulting ownership state:
  `PROBE err=true bytes_sent=0 orphan=None map_has_handle=false dead=false`
  Restored afterwards; `git status --porcelain` is empty and the worktree is clean.
- **Flush-ordering check.** Confirmed `any_packet_flushed` / `message_complete` are set *after* `send_data_fut.await?` (`packet_writer.rs:459-461`) while `send_attempted` is set *inside* the cancellable future (`:445-448`), so the new `send_attempted && nothing_sent` pair genuinely distinguishes "never polled" from "polled, fate unknown". Verified the biased short-circuit semantics are what the `cancel_on_writer_creation` arm of `begin_execute_prepared_release_cancellation_tracks_send_attempt` relies on.
- **Not verified:** the two new C++ e2e tests — no live SQL Server or ODBC driver registration in this environment. They rest on the ADO validation leg, which passed on this head. I also did not re-run the description's local full-suite numbers (2,009 unit tests; 3,925/405/14); I am not disputing them, only noting they are unverified here.

## Findings

### Blocking

None.

### Suggestion

**1. `execute_sp_prepexec` still discards the orphan before the send boundary — the same window this PR closes for the streamed path.** (`mssql-tds/src/connection/tds_client.rs:4233-4236`)

This PR's own rationale on `begin_execute_prepared` says it best:

> Piggybacking the drop would let cancellation discard it while the server still held the plan.

That is a precise description of what the piggyback path still does. `orphan.take()` and both map evictions happen at RPC-build time — before the packet writer even exists — so a cancellation landing before the first flush loses the orphan. Measured, not inferred:

```
PROBE err=true bytes_sent=0 orphan=None map_has_handle=false dead=false
```

Zero bytes reached the server, yet the caller is told there is nothing left to release and the client has forgotten handle `7`. The server still holds that plan, so it leaks until disconnect. Contrast the new `execute_sp_unprepare` behaviour on the identical hook, which `begin_execute_prepared_release_cancellation_tracks_send_attempt` pins: `orphaned == Some(id)`, both map entries retained, connection reusable, release retryable.

Two supporting details:

- The doc contract at `:4122-4128` says the orphan is *"still `Some` if the call failed before serialization (the caller may retry the release)"*. The probe is a failure before anything was serialized to the wire, and it came back `None` — so the documented guarantee is narrower in practice than it reads.
- The pre-existing test `execute_sp_prepexec_consumes_orphan_after_send_begins` passes for a weaker reason than its name suggests. Because the take happens at build time, it would pass even if no send ever began; nothing currently pins the pre-send arm. `execute_sp_prepexec_preserves_orphan_on_pre_send_error` only covers the `has_open_batch` usage error, which returns before the take.

This is pre-existing on `main` and this PR strictly narrows the gap, so I am not treating it as blocking. But `send_attempted()` now exists precisely to close it, and leaving two of three release paths on the new contract and one on the old is the kind of asymmetry a later parity pass gets wrong. Suggest a follow-up issue rather than expanding this PR.

### Nit

**1. The retire comment describes only one of the two conditions that reach it.** (`mssql-tds/src/connection/tds_client.rs:3621-3623`)

```rust
if serialize_result.is_err() && message.nothing_sent() {
    // A cancelled write may have left a partial packet on the wire.
    self.retire_without_writing();
}
```

Since `any_packet_flushed` is set only after `send_data_fut.await?` succeeds, this branch is also taken when `network_writer.send()` returns a plain I/O error, not just on cancellation. Retiring is the right action in both cases, so this is wording only — but "a cancelled write" reads as an exhaustive description of when the connection gets condemned, and it is not.

## Required CI

`coverage-report` is **failing** on this head ([run 35689467030](https://github.com/microsoft/mssql-rs/actions/runs/35689467030/job/106623166443)). Reading the log, it is not a test or coverage regression — the job polls for the Cobertura artifact and gave up after 12 attempts:

```
Attempt 12: checking for coverage artifact...
⏳ Coverage artifact not ready yet (attempt 12)...
Available artifacts: none
❌ Timeout: coverage artifact not found within 75 minutes
```

Every other required check is green, including the full `mssql-rs Pull request validation` ADO pipeline (all five build legs, Test MacOS, Kerberos, and both cross-repo `mssql-python` legs) and `Merge Coverage`. Since the ADO validation now passed on this exact head, the unchecked `cargo btest` box in the description looks like it can be checked.

mssql-tds/src/connection/tds_client.rs
saurabh500: Keeping the no-op: this helper accepts an `Option<StatementId>`, and `None` means there is no resource to release, just as an absent handle is a no-op in `execute_sp_unprepare`. Both callers currently pre-check, but replacing this case with a panic/debug assertion solely to remove an uncovered line would not improve correctness.

mssql-tds/src/connection/tds_client.rs
saurabh500: The repeated live-orphan stream regression now uses `timeout_secs(1)`, asserts a finite parked timeout and the actual unprepare bytes, and covers four cancel/retry/rebind cycles; removing its live-handle setup makes it fail, and the sibling batch retry also uses one second. I could not substantiate the historical 2,009 count, so the description is being replaced with this run's exact results: 2,211 TDS unit tests and 3,911 workspace library tests passed after generating the repository's missing TLS fixtures.

mssql-tds/src/connection/tds_client.rs
saurabh500: Added deviation 15 in f30e7a7f, documenting the separate round trip and why release failures propagate instead of letting another handle occupy the single pending-unprepare slot. I checked `DropPrepHandle`, `BuildSPPrepExec`, and `ProcessDAEParam` in the local msodbcsql source; the entry explicitly distinguishes that source evidence from the still-outstanding build-specific reference measurement and human parity sign-off.

mssql-tds/src/connection/tds_client.rs
saurabh500: Fixed in f30e7a7f: shared `deduct_timeout` now charges whole elapsed seconds, matching `deduct_query_timeout`, while still rejecting exact/over-budget exhaustion and using cumulative elapsed time for multi-step execution. Both the live-orphan streamed path and the batched retry reproduced the one-second failure before the fix and pass afterward; the docs state the less-than-one-second rounding allowance.

saurabh500:
> `execute_sp_prepexec` still discards the orphan before the send boundary.

Confirmed the build-time `orphan.take()` and map eviction ordering and tracked the pre-existing materialized-path issue in #647, rather than expanding this streamed-cleanup fix. The interrupted-write wording is corrected in f30e7a7f as well.

mssql-tds/src/connection/tds_client.rs
saurabh500: The subsequent main merge moved the registry, so 503fc43f preserves this decision as entry 18 in `mssql-odbc/docs/parity-deviations.md`, following the updated engineering instructions. The policy and stated evidence limits are unchanged.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -4023,6 +4056,8 @@ impl TdsClient {
     /// and sends `sp_unprepare` only when the statement's connection id still
     /// matches the recovered connection. Take the id with
     /// [`PreparedStatement::take_id`] once the statement is done.
+    /// Keep that id if this call fails before sending; retrying it is safe,
+    /// since a previously attempted release has already removed its map entry.
```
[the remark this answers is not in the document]
saurabh500: I traced this caller, but the new unsent-cancellation case is not reachable here: `flush_pending_unprepare` passes only a timeout, not a cancellation handle, and ODBC `SQLCancel` does not interrupt this synchronous cleanup. The writer checks elapsed timeout after attempting the write, so this timeout path removes the TDS handle entry. A successful reconnect also clears the handle maps before `unprepare` checks the remaining budget.

I checked the timeout path with a temporary real-TCP mock-server probe: an eight-second RPC response delay and a one-second cleanup timeout returned in about 1.1 seconds, with both `pending_unprepare` and the TDS map entry cleared. The probe was removed afterward. Busy/failed-recovery errors are existing best-effort cleanup behavior, rather than a regression from the changed send ownership. Leaving this caller unchanged.

Theekshna (commented):
Unattended hourly review sweep found no findings. Covered: the delta since my last review at `11eff55a` (diffed against merge-base `cc7bb82e` of `main`; PR's own diff is 7 files, +718/-73) — the review-follow-up timeout fix in `TdsClient::deduct_timeout` (round-up to truncate-down, `tds_client.rs`) and its exec_common.rs doc-comment update, entry 18's divergence writeup in `parity-deviations.md`, `stmt.rs`/`execute.rs` comment updates, all 6 prior review threads (0/6 unresolved via GraphQL, single page), `gh pr checks` (16/16 green), the PR description/checklist against current diff and CI, and a fresh `cargo check -p mssql-tds --lib --tests` (sandbox blocks the usual mutation-test build on `crc32fast` codegen, so this is a compile-only substitute, not an executed test run).

Theekshna (approved):

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3914,9 +3929,27 @@ impl TdsClient {
...
+                ExecuteOptions {
+                    timeout: timeout_sec,
+                    ..opts.clone()
+                },
+            )
+            .await?;
+            let remaining = Self::deduct_timeout(opts.timeout, started.elapsed()).into_timeout()?;
+            timeout_sec = remaining.seconds();
+            self.remaining_request_timeout = remaining.duration();
+        }
```
Theekshna: **Please piggyback the orphan drop on the streamed `sp_prepexec` in this PR, rather than sending a standalone `sp_unprepare`**

The fix is correct and the tests are solid. My concern is the shape of the solution: it adds a server round trip that msodbcsql specifically avoids, and deviation entry 18 records that round trip as if it were forced. Once it lands as a documented deviation it becomes the baseline nobody revisits, so I'd like it addressed here rather than deferred.

**The slot is already there and deliberately empty.** In `begin_execute_prepared`, the streamed `sp_prepexec` hardcodes its by-ref `@handle` to NULL (`tds_client.rs:3993-3996`, "NULL by-reference `@handle`: prepare fresh, with no piggybacked drop"). The materialized path threads the orphan through that exact parameter (`tds_client.rs:4231-4237`). Same RPC, same slot — the streamed path just doesn't use it.

**What msodbcsql does.** It resolves the "the parked request may never reach the server" objection by *deferring eviction*, not by avoiding the piggyback:

| | msodbcsql | this PR |
|---|---|---|
| Where the drop rides | `@handle` by-ref on the next prepare — `BuildSPPrepExec`, `sqlccmd.cpp:8244-8263` | standalone `sp_unprepare` before the stream |
| When the orphan is released | only after the send succeeds — `ProcessDAEParam`, `sqlccmd.cpp:6717-6728` | at the release send boundary, before the stream opens |
| On send failure | `CHKRC_SERR_GOTO(retcode, ExitCS)` skips the clear, handle retained | release error fails the execute |
| Round trips on rebind-into-DAE | 1 | 2 |

`DropPrepHandle` (`sqlcfunc.cpp:732-763`) is explicit that deferral is the point — *"defer dropping the sproc to save a round trip"* — and it keeps a standalone path only for `fForceDrop = TRUE` (statement teardown). We're using the force-drop shape on the common rebind path.

Cancellation is actually *easier* for us than for msodbcsql. A parked streamed message is never `message_complete`, so withdrawal sends `EOM | IGNORE` and the server discards the entire message — the piggybacked drop included. There's no "may have landed" ambiguity to reason about, unlike the materialized case.

**One trap worth flagging up front, because the naive version reintroduces #598.** Retaining the orphan across the parked window is not sufficient on its own:

- `begin_execute_prepared` assigns the issued id *at park time* (`tds_client.rs:4032-4034`), so a parked stream leaves `prepared.stmt.id = Some(new_id)`, inert until the drain captures a handle.
- On cancel, `take_dae` restores `pending_unprepare = dae.orphaned` and never clears that id (`stmt.rs:1441-1446`).
- The next rebind calls `orphan_prepared_handle`, which takes the inert id unconditionally and hits `pending_unprepare.replace(...)` (`stmt.rs:1411-1419`) — two handles, one slot, `debug_assert!` fires. That is #598 again.

So the issued id needs to be assigned on successful `finalize()` rather than at park time. `cancel_streamed_write` can't do it today — it takes `&mut self` and never sees the `PreparedStatement` (`tds_client.rs:2486`), so this needs a small API change on the streamed surface.

**What I'd consider done:**
1. Streamed `sp_prepexec` carries the orphan in `@handle` when one is present.
2. Orphan and AE metadata are evicted only after the message completes, mirroring `ProcessDAEParam`.
3. Cancel retains the orphan *and* leaves no inert id behind — a cancel/rebind cycle must not trip the `pending_unprepare` assert.
4. A test asserting the rebind-into-DAE sequence emits **one** RPC with a non-NULL `@handle`, and no `sp_unprepare` (`0xFF 0xFF 0x0F 0x00`) — the inverse of the current `..._releases_each_orphan_with_one_second_timeout` assertion.
5. The standalone path stays for the cross-identity orphan case, which has no slot to ride (matching `fForceDrop = TRUE`).
6. Entry 18 either goes away or narrows to the cross-identity case.

If the id-assignment change turns out to ripple further than it looks, say so and I'll settle for splitting entry 18 and a tracked follow-up — but I'd like the piggyback attempted here first.

saurabh500: Implemented in a8602cc9: streamed `sp_prepexec` now piggybacks the orphan, and `end_execute_prepared_param` assigns the new ID and evicts the old handle/AE metadata only once the message is complete (including a completed send followed by an error); `SQLParamData` restores the resulting ownership on both success and failure. Repeated one-RPC/non-NULL-handle and ODBC cancel/rebind regressions pass, standalone cleanup remains for the cross-identity `sp_execute` case, and entry 18 is narrowed accordingly; 4,020 library tests and formatting/Clippy passed, but live SQL Server/reference-driver comparison was not rerun.

Theekshna (commented):
Unattended hourly re-review found no findings for this delta. Covered: the two new commits (`a8602cc9`, `6297939e`) implementing piggybacked prepared-handle cleanup and interrupted-send retirement in `mssql-tds`/`mssql-odbc` — cancellation/error-path ownership tracing across `param_data.rs`, `execute.rs`, `tds_client.rs`, and `packet_writer.rs`; msodbcsql parity independently confirmed against `sqlccmd.cpp` (`BuildSPPrepExec`'s by-ref `@handle` IN value from `hPrepDropDeferred`, matching this PR's own `parity-deviations.md` entry 18) and test sufficiency for the cancel/failed-send/completed-send-with-failed-response classes. Local `cargo nextest` could not run in this environment (a pre-existing AVX-512 codegen failure in `crc32fast`, reproduced identically on unmodified `main` HEAD — not caused by this PR); deferred to `gh pr checks 599`, which is green apart from stages still pending.

David-Engel (approved):

mssql-tds/src/connection/tds_client.rs
```diff
@@ -2615,9 +2640,12 @@ impl TdsClient {
         // transport is retired without being written to again.
         let ignored =
             tokio::time::timeout(CANCEL_TIMEOUT, packet_writer.cancel_current_message()).await;
-        drop(packet_writer);
+        let message = packet_writer.suspend();
+        let send_incomplete = message.send_incomplete();
```
David-Engel: **Suggestion:** `send_incomplete` is read *after* `cancel_current_message()` has already written the IGNORE packet, so this only catches a write interrupted by the withdrawal itself — never a message that arrived here already interrupted.

Today that is unreachable: `retract_partial_request` returns early on `message.send_incomplete()`, and a parked streamed message can only be re-parked after a write that completed. But `cancel_streamed_write` routes into `withdraw_sent_message` on `nothing_sent()` alone (`tds_client.rs:2495-2506`), so it relies on that invariant implicitly. If a message with `send_incomplete == true` ever reached this path, the IGNORE packet would be appended on top of a half-written packet/TLS record, and a successful write would clear the flag and report `Withdrawal::Done` — exactly the "neither IGNORE nor attention can repair it" case the rest of this PR is careful about.

Cheap defense-in-depth: check the incoming message on entry, before resuming a writer over it, e.g.

```rust
if message.send_incomplete() {
    message.abandon(self.transport.as_writer());
    return Withdrawal::WriteAbandoned(/* ... */);
}
```

and keep the post-write sample for the interruption that happens during the withdrawal.

mssql-tds/src/connection/tds_client.rs
```diff
@@ -3914,15 +4006,34 @@ impl TdsClient {
...
+                    ..opts.clone()
+                },
+            )
+            .await?;
+            let remaining = Self::deduct_timeout(opts.timeout, started.elapsed()).into_timeout()?;
+            timeout_sec = remaining.seconds();
+            self.remaining_request_timeout = remaining.duration();
+        }
 
+        let live_handle = statement
```
David-Engel: **Nit:** this recomputes `live_handle` from the same two inputs as the binding 17 lines above. `unprepare_orphan` only touches the orphan's entry, so the value can differ solely when a caller passes `statement.id == *orphaned` — which the ODBC layer cannot do, since `orphan_prepared_handle` takes the id out of the statement.

If the recompute is deliberate hardening for direct `mssql-tds` callers of this public API, a one-line comment saying so would stop a later reader from deleting it. Otherwise the first binding can just be the predicate:

```rust
let has_live_handle = statement
    .id
    .is_some_and(|id| self.prepared_handles.contains_key(&id));
```

leaving a single `live_handle` lookup here.

David-Engel (commented):
## Automated review (unattended run)

Reviewed the full diff against merge-base `cc7bb82e` (10 files, +1438/-144), focused on the streamed `sp_prepexec` piggyback, the send-ownership boundary in `PacketWriter`, and the ODBC `SQLParamData` plan restoration. This came from an unattended run, so the findings below were not checked by a human first.

The design holds up: ownership settles on `message_complete()` rather than at park time, cancellation retains the orphan without minting an inert identity, and the ODBC layer restores the plan on both success and failure. Verified the new tests are not vacuous with two mutations (`cargo nextest run -p mssql-tds --lib`, 35 prepared/streamed tests):

- forcing the streamed `@handle` back to `SqlType::Int(None)` fails `begin_execute_prepared_piggybacks_each_orphan_with_one_second_timeout`;
- inverting the `message.message_complete()` gate fails 9 tests, including `prepared_stream_partial_cancel_retains_orphan_and_metadata` and `prepared_stream_completed_send_consumes_orphan_on_error`.

Also re-ran `begin_execute_prepared_release_cancellation_tracks_send_attempt` 40x to check the new `send_attempted` flag for a race when the handle is already cancelled: 40/40 deterministic, because `CancellationToken::run_until_cancelled` has an `is_cancelled()` fast path that never polls the inner future. No flakiness there.

No blocking findings. Two non-blocking items are left inline:

1. **Suggestion** — `withdraw_sent_message` samples `send_incomplete` only after writing the IGNORE packet (`tds_client.rs:2644`).
2. **Nit** — duplicated `live_handle` lookup in `begin_execute_prepared` (`tds_client.rs:4034`).

**PR-wide (no attachable diff line):** the two red checks are infrastructure, not this PR. The macOS job ran `Run tests (mssql-tds and mssqlodbc packages)` to success; it then failed on `Publish Cobertura coverage artifact` with `##[error]Artifact CoberturaCoverageRust_MacOS already exists for build 177888`, which skipped the ODBC e2e steps and failed `Publish ODBC e2e test results (macOS)`. `coverage-report` then failed only because it refuses to report against a failed ADO build. Worth a re-run rather than a code change — but the e2e C++ cases added here (`RebindAfterCancelledStreamReleasesPriorHandle`, `RebindAfterDeferredStreamReleasesPriorHandle`) have therefore not actually executed on macOS in this build.

